PR #172, opened 2023-07-02
Toggle fullscreen on super/ctrl+return, only macOS for now

mrnugget:
This fixes or at least is the first step towards #171:

- it adds `cmd/super + return` as the default keybinding to toggle fullscreen for currently focused window.
- it adds a keybinding handler to the embedded apprt and then changes the macOS app to handle the keybinding by toggling currently focused window.

Let me know if the approach here is correct. I essentially copied what we ended up doing with `gotoTab`. But maybe you'd like the keybinding or handlers to be moved up/down in the respective files and so on.

### Demo

https://github.com/mitchellh/ghostty/assets/1185253/f02083fd-d564-4926-a99a-864db3037cca


mitchellh (changes requested):
This looks perfect. One little tweak but the rest looks good. I can shuffle things up/down later if I notice it.

PR #176, opened 2023-07-03
gtk: handle ctrl+return to toggle fullscreen

mrnugget:
This is a follow-up to #172 and adds fullscreen handling to the GTK apprt.

Works reliably for the current window and mimics exactly what happens when F11 is pressed, which seems to be the standard keybinding for GTK apps to toggle fullscreen.

### Demo

https://github.com/mitchellh/ghostty/assets/1185253/f7cbe0f3-5b87-4f0c-8a7e-e4d8b108f6a3

### DIFF
--- a/src/apprt/gtk.zig
+++ b/src/apprt/gtk.zig
@@ -459,6 +459,16 @@ const Window = struct {
         }
     }
 
+    /// Toggle fullscreen for this window.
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Perfect! Thanks so much, you're on a roll! I really appreciate it.

Issue #189, opened 2023-07-06
GTK: Search scrollback


mitchellh:
A major missing feature is the ability to search scrollback, i.e. `cmd+f` on Mac, `ctrl+F` on Linux. 

This issue can be implemented in multiple steps, not just one giant PR:

- [x] Core, single-thread search functionality in `src/terminal` (#2885, #9585)
- [x] Search thread, mailbox on top of the core system utilizing the big renderer lock to read the screen for small time slices
- [x] Hook up the search thread to Surface
- [x] Rendering search matches in our renderers (non-GUI work)
- [x] Search iteration (prev/next match) #9702
- [x] Glue in apprt, bindings #9709 
- [x] macOS GUI #9709 
- [ ] GTK GUI

**EDIT:** There was a previous approach here (see edit history) with concurrent PageList access. After various false starts, I think I've identified a better path forward.

GoldsteinE:
Prior art from foot: this dimming / highlighting looks really nice IMHO.

![Image](https://github.com/user-attachments/assets/8df7d2e7-98e5-46bc-974f-541f26778ad3)

zzhaolei:
Using `cmd+shift+j` and `cmd+shift+alt+j` does not work. 
<img width="800" alt="Image" src="https://github.com/user-attachments/assets/ff90e307-8c94-4e0b-8076-02c4b755c082" />

aofei:
OT: Genuinely think this issue should be pinned. 👀 (Been asked about this feature several times today)

mitchellh:
> OT: Genuinely think this issue should be pinned. 👀 (Been asked about this feature several times today)

Good feedback. Pinned!

weeheavy:
I'd love to see the elegance of Alacritty's search:
- `cmd`+`f` opens a small one-line search at the bottom
- `enter` scrolls foward through search results
- `shift`+`enter` scrolls backwards

nervenes:
I'd like the UI for this to be native, similar to Terminal.app
<img width="682" alt="Image" src="https://github.com/user-attachments/assets/7f20352a-480d-4168-869b-f49eaf65bdb1" />

sethgrid:
I just did a full grep of my file system looking for a breadcrumbs and then found out I can't search through my results. I was able to cmd+a cmd+c tab to vscode and cmd+p and then cmd+f. I am _constantly_ using cmd+f on my terminal :(

econner:
[...]

PR #200, opened 2023-07-09
Write scrollback to file, output to tty ("write_scrollback_file" binding)

mitchellh:
Fixes #196 
Helps #189 

This introduces a new keybinding `write_scrollback_file` bound to `Cmd+Shift+j` by default (Ctrl on Linux). This action will write the entire scrollback to a temporary file, and then output the path to that file to the terminal. This can be used to save your scrollback, search it, modify it (in the file, not the terminal), etc.

ANSI escape sequences are stripped -- we can add an option to add them later but I think its more useful without.

## Demo

https://github.com/mitchellh/ghostty/assets/1299/513b6c77-aacb-4371-b665-c8917571ec8d

### DIFF
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -23,8 +23,8 @@ const Command = @This();
 
 const std = @import("std");
 const builtin = @import("builtin");
-const TempDir = @import("TempDir.zig");
 const internal_os = @import("os/main.zig");
+const TempDir = internal_os.TempDir;
 const mem = std.mem;
 const os = std.os;
 const debug = std.debug;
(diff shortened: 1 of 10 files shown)


janos-r:
I don't use ghotty yet because I am waiting for a way how to simply search back and copy some output.
If this PR is already done, shouldn't it be simple to add a keybinding to open open the scrolback within the setup $EDITOR? That would be wonderful.

drone1:
Writing scrollback to files on disk opens up quite a large attack surface for malicious users, no? 

I would love an option to keep the buffer held in memory. I'd rather pay the cost in RAM, personally.

PR #266, opened 2023-08-09
Primary (selection) clipboard support

mitchellh:
Fixes #265 

This adds support for the "primary" clipboard (as its called in X11, otherwise also known as the "selection" clipboard). This is the clipboard whose contents are set automatically for any selection and can be pasted using the middle-click button.

This also adds a `copy-on-select` configuration that can have one of three values:

  - `false` - disable copy (and paste!) on select completely
  - `true` - enable copy/paste if the system supports a selection clipboard (default)
  - `clipboard` - enable copy/paste but use the system standard clipboard 

https://github.com/mitchellh/ghostty/assets/1299/f69b45de-b2f6-4635-939e-44c886cc8714

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -253,6 +253,7 @@ typedef void (*ghostty_runtime_toggle_fullscreen_cb)(void *, bool);
 
 typedef struct {
     void *userdata;
+    bool supports_selection_clipboard;
     ghostty_runtime_wakeup_cb wakeup_cb;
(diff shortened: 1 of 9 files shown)


00-kat:
Please don't bump ancient threads—this Pull Request is over one and a half years old.  The feature you are requesting is #4404 (while the title says right click to copy only, your requested behaviour is discussed in there too).

PR #296, opened 2023-08-17
macOS: inherit font size when creating new tab

mrnugget:
This is one part of #281. I'm not 100% edit about the code. I think the Swift code is largely fine (except for the names for the methods in `PrimaryWindowManager`), but I'm unsure about the Zig code and how to pass the `fontSize` around.

## Demo

https://github.com/mitchellh/ghostty/assets/1185253/007f9b8c-c14b-40ea-86b1-67874f0846ff


mitchellh (commented):
Everything actually looks pretty good. One big comment.

And then a question: does this preserve `cmd-T` working even if no windows are open?

PR #329, opened 2023-08-25
Implement cursor fg/bg for Metal renderer

hovsater:
An attempt to allow both foreground & background to be set for the cursor. In doing so, I break the current API for the configuration object and rename `cursor-color` to `cursor-background`. I also introduce a new configuration option, `cursor-foreground`, allowing you to set the foreground color.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -71,6 +71,7 @@ pub const Config = struct {
 
     /// The color of the cursor. If this is not set, a default will be chosen.
     @"cursor-color": ?Color = null,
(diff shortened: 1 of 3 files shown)


mitchellh (changes requested):
This looks really good, straightforward. I have a cosmetic question and then if you can follow my note above about impl this for OpenGl, that'd be great. Thanks!

Also, with most PRs, I try to attach a before/after screenshot so we can visually see what changed. If you can do that too that'd be awesome!

PR #342, opened 2023-08-27
macos: fix non-native fullscreen dock/menubar bugs

goonzoid:
I noticed a couple of minor bugs when using non-native fullscreen on a multi monitor setup:
- If the dock is displayed on the other screen (even persistently) it gets incorrectly hidden
- The menu bar on the other screen gets incorrectly hidden

This PR change aims to fix the first issue, and reduce occurrences of the second. I'll need to do some more digging to see if it's possible to fix the second issue completely (iterm2 appears to have the same problem, and I sorta suspect that it's an AppKit / NSApp related bug).

A couple of other things:
- These are literally the first lines of Swift I've ever written, so there's a good change I'm doing something silly, particularly when it comes to unwrapping optionals
- As far as I could tell, setting the window to "borderless" before going fullscreen wasn't actually doing anything, and not doing so allows us to get out of the game of storing a copy of the style mask while in fullscreen mode, so I got rid of that. It's possible that this is a mistake, though - @mrnugget do you remember specifically what the borderless setting was intended to accomplish? It might be one of those things that's different on different MacOS versions

Any and all feedback, comments, and suggestions welcome. This stuff gets a bit gross kinda fast, but I think that's just the nature of MacOS desktop programming sometimes unfortunately.

### DIFF
--- a/macos/Sources/Helpers/FullScreenHandler.swift
+++ b/macos/Sources/Helpers/FullScreenHandler.swift
@@ -4,7 +4,6 @@ class FullScreenHandler {
     var previousTabGroup: NSWindowTabGroup?
     var previousTabGroupIndex: Int?
     var previousContentFrame: NSRect?
-    var previousStyleMask: NSWindow.StyleMask?
     var isInFullscreen: Bool = false
     
     // We keep track of whether we entered non-native fullscreen in case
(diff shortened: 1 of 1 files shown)


mrnugget (approved):
Looks good to me! Tried it out locally and it works. Sorry for adding confusing code in the first place, I guess :) See comment for why I added it.

PR #380, opened 2023-09-02
macos: add hidden-menu/visible-menu non-native-fullscreen opts

goonzoid:
First part of #378.

Adds the following config values for `macos-non-native-fullscreen`:
- `visible-menu`
- `hidden-menu`

`true` and `false` continue to work as before.

Gave it a decent test run before opening the PR, but ideally someone else can test this out before merging... feels like this UI stuff is a little fragile, and it's easy to miss something.

~~Update: Also added a fix for the menu bar portion of #343 - the dock part may still be an issue.~~
Update: Also added a fix for #343


mitchellh (commented):
This looks great! Thank you! One comment

PR #385, opened 2023-09-02
Update font DPI when the content scale is updated

mgeist:
Fixes #372 

The content scale was already being updated when the terminal was moved between displays, so all that was left was to update the font size's DPI.

Added the logic under a `contentScaleCallback` since that seems like the primary mechanism that the embedded surface interacts with the `Surface`

Nits welcome! :D

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1192,6 +1192,15 @@ pub fn scrollCallback(
     try self.queueRender();
 }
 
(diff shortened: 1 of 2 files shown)


mitchellh (commented):
Looks good, I just took a look and the renderer padding also uses DPI but I want to take a stab at that myself since there are some bigger issues there I've noticed. Please just fix the `const` thing and ill merge this.

Thank you!

PR #387, opened 2023-09-02
content scale change events should also impact viewport padding

mitchellh:
This calculates the new padding pixel values and propogates those changes to the renderer.

@mgeist: can you verify everything still works with these changes?


mgeist (approved):
👍 everything looks great

on built-in display:
image
external display:
image

What's interesting here is that the screenshot looks kind of washed out for the external display when previewing this post on the built-in display, but it looks fine in reality. I believe this is a mac thing and not a ghostty thing - below is a screenshot of Alacritty with the same issue (although it's using a bolder font so it's not quite as drastic looking)

image

PR #391, opened 2023-09-02
Zoom/Unzoom Split

mitchellh:
Fixes #352 

This adds the ability to "zoom" and "unzoom" a split. A zoomed split is expanded to take up the full window. When unzoomed, the splits retain their existing layout. Focus state remains constant throughout.

This can be bound using the `toggle_split_zoom` action. The default binding is `super + shift + enter` (matches iTerm).

Some other details: when zoomed, the titlebar has a 🔍 emoji. Additionally, trying to create a new split while zoomed will unzoom and create the split.

## Demo

https://github.com/mitchellh/ghostty/assets/1299/33f3abd8-a502-430c-8a97-994b91c2a719


goonzoid (commented):
Looks good to me. Testing it out on macOS 13 and so far it's working perfectly! Thanks for adding this!

PR #443, opened 2023-09-13
macos: allow to hide decorations

SoraTenshi:
It's not as perfect as kitty's hide decorations (which also renders e.g. rounded corners) but i don't this is too bad.

It's also currently not able to hot-reload on config, but i also don't think anyone would want to switch between rendering decorations and not.

This pr also closes #319


mitchellh (commented):
This looks pretty good to me, let me test it locally tomorrow. I think stylistically there are a couple things I want to fix up but generally speaking its all good.

PR #453, opened 2023-09-14
Implement OSC 10 and OSC 11 default color queries

cryptocode:
These OSC commands report the default foreground and background colors.

Most terminals return the RGB components scaled up to 16-bit components, because some legacy software are unable to read 8-bit components. The PR follows this convention by default.

iTerm2 allow 8-bit reporting through a config option, and a similar option is added here. In addition to picking between scaled and unscaled reporting, the user can also turn off OSC 10/11 replies altogether.

Scaling is essentially `c / 1 * 65535`, where `c` is the 8-bit component, and reporting is left-padded with zeros if necessary. This format appears to stem from the XParseColor format.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -315,11 +315,11 @@ keybind: Keybinds = .{},
 /// Allowable values are:
 ///
 ///   * "none" - OSC 10/11 queries receive no reply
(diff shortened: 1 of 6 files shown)


mitchellh (commented):
This looks great! I've commented on some suggestions...

PR #461, opened 2023-09-15
Windows: implement tmpDir and expandPath

goonzoid:
You can't see the additional tests pass from `build run test` because of other compilation errors getting in the way, but if you run the tests for these files directly with `zig test` they now pass in some cases where they previously failed or didn't compile.

It might be nicer to separate the code by platform into separate functions and switch at the function call level rather than having these blocks of Windows specific code in the middle of existing functions. Very open to doing that if that would be preferred. For now I imitated what the zig standard library seems to do, though.

Am curious to hear from @marler8997 about the temp dir stuff, since there seem to be about 18 different env vars that you can use on Windows to get a temp dir path! A bit of searching lead me to believe that `$TMP` was the most commonly used one, but I could be missing something. Also, I have no idea what a sensible fallback path is if we don't get anything from `$TMP`... I left a `TODO` in there for that, but would love to figure it out.

As always, any feedback is more than welcome!

### DIFF
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -232,9 +232,19 @@ pub fn expandPath(alloc: Allocator, cmd: []const u8) !?[]u8 {
         return try alloc.dupe(u8, cmd);
     }
 
-    const PATH = os.getenvZ("PATH") orelse return null;
+    const PATH = switch (builtin.os.tag) {
+        .windows => blk: {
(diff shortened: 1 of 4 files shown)


mitchellh (commented):
The amount of windows-specific logic doesn't bother me, I think it'd be noisier at the moment to pull this out into dedicated functions. That's surely something we'll need to do eventually but not yet.

Some comments added.

SoraTenshi (changes requested):

SoraTenshi (approved):

PR #541, opened 2023-09-26
config: add term config option

rockorager:
Add a configuration key for the TERM environment variable. Default this to "ghostty". Most TEs are using their name as the default TERM value. Most modern termulators aren't even providing "xterm-" as an alias anymore, after some drama between kitty / ncurses.

Notably, this also has issues for tcell-based applications (I've submitted a PR to tcell to fix) because it fails if the TERM value doesn't match the _primary_ name of the terminal in the terminfo file.

Providing a config option allows users to modify-with-persistence if they have issues, but Ghostty should be known as Ghostty by default!

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -411,6 +411,9 @@ keybind: Keybinds = .{},
 /// Debug builds of Ghostty have a separate single-instance ID.
 @"gtk-single-instance": bool = true,
 
+/// This will be used to set the TERM environment variable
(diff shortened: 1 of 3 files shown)


mitchellh (commented):
I'd love to disown the `xterm-` prefix way of life, but it does spook me out. There's a lot of terrible StackOverflow answers out there telling people to just look for that to determine color and so on... I wonder if its worth it.

This looks good though, generally, one suggestion.

PR #566, opened 2023-09-28
macos: add tab index labels

gpanders:
I mentioned this on Discord but didn't get a response if this is a desired feature. I went ahead and made an attempt to implement it and it wasn't too difficult.

I don't have much (read: any) experience with macOS UI development, so if there's a better way to do this please tell me.

---

Adds labels to the window tabs that indicate the tab index and the key to press to activate that tab.

Screenshot:

Screenshot 2023-09-27 at 9 05 43 PM

Indices only go up to 9 (which is what is supported via the default keyboard shortcuts):

Screenshot 2023-09-27 at 9 09 27 PM

### DIFF
--- a/macos/Sources/Features/Primary Window/PrimaryWindowController.swift
+++ b/macos/Sources/Features/Primary Window/PrimaryWindowController.swift
@@ -1,6 +1,6 @@
 import Cocoa
 
-class PrimaryWindowController: NSWindowController {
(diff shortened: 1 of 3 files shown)


mitchellh (commented):
This looks really good! Especially for someone who has never done macOS dev before. I left a couple comments. I'll give this a go in a bit. I'm happy to take over for the comments just let me know.

PR #578, opened 2023-09-28
Kitty keyboard improvements

rockorager:
Partially fixes #561.

Fixes the reporting of alternate keys for alternate layouts in GTK. For all platforms, fixes reporting of base layout key.

This doesn't fix the case where we have a shifted non-alphabet character. We currently only check for this by using gdk_keyval_to_lower, but any shifted numeric or a semicolon will not have an unshifted value. Added a comment on what needs to be done.

Also the control flow is kind of ugly...open to suggestions for cleaner ways to do all the cases in zig.


mitchellh (commented):
Awesome work, thanks! I think the logic is mostly very sound, some small comments.

PR #634, opened 2023-10-07
Alternate scroll disabled by default, only sends up/down

mitchellh:
Fixes #633

A couple bugs with our alternate scroll (mode 1007) implementation:

  - This should be disabled by default. Verified with iTerm, Kitty, xterm
  - This should only send up/down arrows, not left/right. Verified with Kitty, xterm. Other terminals do send left/right but I don't think that is correct.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1195,15 +1195,6 @@ pub fn scrollCallback(
                 }
             }
 
-            if (x.delta_unsigned > 0) {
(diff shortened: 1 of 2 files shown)


danielrode:
Was this setting completely removed (rather than just disabled by default)? I tried setting `mouse_alternate_scroll = true` in my config and Ghostty now gives an error saying "mouse_alternate_scroll: unknown field".

PR #694, opened 2023-10-17
macos: add menu items to modify font size

gpanders:
Add a new "View" menu to the menu bar with items to increase, decrease, or reset the font size.

---

We also get the "Show Tab Bar" (to show the tab bar even when there is only one tab) and the "Show All Tabs" menu items for free.

Some screenshots:

Screenshot 2023-10-16 at 8 15 53 PM

Screenshot 2023-10-16 at 8 16 02 PM


mitchellh (approved):
Excellent job, this looks great. Thank you.

PR #704, opened 2023-10-19
Get `zig build test` working on Windows

der-teufel-programming:
Disable iconv on Windows by default (enabled via cli flag).
Skip various tests not implemented on windows.


mitchellh (changes requested):
Looks good, thanks! If this gets `zig build test` to work, maybe we can enable that on the CI now… (the windows CI is setup already). I can help look at that once some of the feedback is addressed.

PR #729, opened 2023-10-25
resolve simple crashes

truemedian:


mitchellh (commented):
Thanks! I made some stylistic tweaks and added a couple tests. Only one major change was the reversion of the log statements, see my comment there.

PR #735, opened 2023-10-27
macos: set window resizeIncrements when cell size changes

gpanders:
The resizeIncrements property is only modified when the cell size of the focused window changes. If two splits have the same cell size then the property is not modified when focusing between the two splits.

---

I implemented this at least 3 different ways, and this is the method I personally am most satisfied with, but curious to get feedback. None of the other runtimes (glfw or GTK) take advantage of the set cell size callback right now (neither has a "resize increment" API like Swift does[^1]), but they could in theory.

Note I'm also working on https://github.com/mitchellh/ghostty/issues/305. I can include that in this PR as well if we want to land these changes together, but wanted to get this code up first to make sure the approach looks ok.

Fixes: https://github.com/mitchellh/ghostty/issues/194

[^1]: GTK3 had such an [API](https://docs.gtk.org/gtk3/method.Window.set_geometry_hints.html), but it was removed in GTK4


mitchellh (commented):
This looks pretty fantastic. I have two really minor comments but overall this is excellent work. Want to address the couple bits of feedback and let me know when its ready to go?

mitchellh (approved):
Works great looks great. Thank you!

PR #753, opened 2023-10-28
feat: update undercurl with antialiasing

wintermi:
Updated the `undercurl` to be antialised to give the representation of a smooth curl.

The `undercurl` now starts at the midpoint of the wave which means the peak is always in the middle of the character.

The below screenshot compares a number of different options side by side with the 4 options that Kitty provides.

What has been committed as part of this pull request is Option 3.

image


mitchellh (changes requested):
Does this fix #699?

Looks amazing! I think there are some types I want to see if I could clean up but it may not be possible... hard to comment on in GH just have to try some Zig things.

PR #759, opened 2023-10-29
implement selecting output a `ScreenPoint` is in

dchinmay2:
This works by finding prompt markers provided by shell integration Does not yet close #752 as this is not exposed


mitchellh (commented):
Looking very good, just noting a couple potential crashers.

PR #767, opened 2023-10-30
fix window-padding-balance in combination with explicit padding

dvorakluk:
Computation of balanced padding was slightly off in combination with explicit paddings, because the "leftover" space was computed from the whole screen and not the padded one.

before:
before

after:
after


mitchellh (commented):
The only issue I see in this PR is I believe it makes balanced vs explicit mutually exclusive. Am I reading that correctly? It should be completely valid to enable balanced padding AND add some additional explicit padding.

The algorithm I think should be: (1) calculate the balanced padding given the screen size (2) add the explicit padding to the balanced calculations. That's what the old logic was trying to do. I'm on mobile so I didn't verify the bug or anything.

PR #778, opened 2023-11-01
gtk: add support for splits

mrnugget:
This adds support for splits to the GTK build.

I built on the work done by @nfisher1226 in his fork (introducing the `Tab` and `Paned` structs, the `Parent`/`Child` abstractions) and wired it together.

**What's included**:

- Terminals can be split with `Ctrl+shift+e`/`Ctrl+shift+o`
- Split terminals can be resized
- Focus can be changed by clicking on a surface
- When changing a Ghostty tab/window, focus is restored to the last-focused terminal in a split tab. *But* not if focus was changed by clicking on the tab label in the tab bar. See [my comment here](https://gitlab.gnome.org/GNOME/gtk/-/issues/4249#note_1904474) on the GTK issue tracker. It was an issue that I spent 4 days until I said that it's out of scope for this PR. Restoring focus _works_, as long as tabs are changed through keybindings (`Alt+[number]` or `Ctrl+shift+[left|right]`).

**What's not included**:

- Changing focus via keybindings. This still requires a bit of work, but shouldn't be too hard.

**Regressions**:

- This breaks the "drag tab to turn tab into separate window" behaviour. I talked with @mitchellh about this and we agreed that it's fine to fix this in a follow-up.

**What's not great, or what I'm not happy about**:

This PR contains the most Zig code I've ever written, so a lot of things might not be idiomatic. This is also the most I've ever worked with GTK, so same disclaimer applies.

To call out a few specific things that I'd love feedback on in review and/or suggestions:

- Focus handling is not great. I don't like `focus_child`, it feels like a crutch, but since GTK4 fires the focus enter/leave event weirdly, we can't rely on the callbacks to manage this field (which would make it better).
- I'm a bit unsure with some of the method receiver: should something be defined on `*Surface` or on `*Paned` or on `*Window`? This stuff is hard, tried my best, but maybe there's better ways.
- String handling (!) for surface titles. I'm not warmed-up with strings in Zig yet, so what I did is to keep a static buffer with a big question mark next to its size and then transform that into a cstring. The check whether the buffer has been set is especially bad. But I didn't want to separately heap-allocate this. Maybe that's better though? Happy to change it. 
- Naming of `parent` and `parentSurface`: so, this is due to a rebase. While this branch added `Surface.parent: Parent`, `main` got `Surface.parent: bool`. I renamed the latter to `parentSurface`, but this should probably be reworked.
- Idiomatic Zig. Again: first time writing Zig properly. Any pointers appreciated.
- *There is still a weird bug* I haven't easily reproduced: sometimes when you close splits, you can have "hidden" siblings pile up. You notice that the divider gets thicker and thicker and when you closed all the visible ones, the invisible ones pop up. I'll need to try to reproduce this.

## Demo

https://github.com/mitchellh/ghostty/assets/1185253/012e525b-214b-4e0d-92e0-4a5314626832


mitchellh (commented):
Yeah this is looking really good, I think I dropped some comments on some lines I wrote but in that case it's good notes for future me anyways. I think the last bit that needs to be looked at is the title text buffer 😄

PR #822, opened 2023-11-06
macos: implement split resizing

gpanders:
Fixes: https://github.com/mitchellh/ghostty/issues/305

The commits are broken out individually with commit messages. Below is more of a brain dump discussing the approach I took.

- macos: refactor SplitNode
- core: add support for bindings with multiple parameters
- core: add resize_split binding with default keys
- macos: support resizing splits
- macos: use Ghostty.AppState as @EnvironmentObject
---

This ended up being a lot more work than I originally anticipated, and I still think there is probably a lot that could be done better here. I'm opening the PR because it's time to get some external feedback and I've been sitting on this for far too long.

First, semantics: how do we describe "resizing a split"? I looked at prior art in iTerm2, kitty, and tmux and while the UI is different in all three, all have the concept of resizing a split by moving the "split divider" either left, right, up, or down relative to the focused split. This is the approach I took as well. This certainly _feels_ intuitive (at least to me) when used interactively with the keybindings, but in the future if Ghostty ever provides a programmatic API, we may also want to provide a way to resize splits in "absoltue" terms (e.g. "increase/decrease width/height", rather than moving the split divider).

Second, implementation: I implemented this by sending a notification to the currently focused `SplitNode.Leaf`. The leaf needs to know whether or not it is part of a split container, so we add a `parent` field to the Leaf so that it has access to its parent container, if any. If the user requests a resize along the same direction of the split (e.g. resizing left/right in a horizontal split or up/down in a vertical split) then we're done: the container resizes appropriately. But if not, we must continue to travel up the tree until we find a container that is split along the same direction as the requested resize.

The biggest challenge was figuring out how to take a resize amount (which is in points) and translate that into a change in the fractional split value that is used in the `SplitView`. The `SplitView` has a `split` field with a value between 0 and 1 which determines how much space is allocated to its top/left view versus the bottom/right view. I could not find a way to get the `SplitView`'s size outside of the view itself, so instead I did this by using a Publisher (`PassthroughSubject`) that is given to the `SplitView` by the `TerminalSplitContainer`. When the `SplitNode.Container` object has its `resize()` method called, we send a message to the Publisher which is received by the `SplitView`, which in turn updates its `split` field.

I am 100% open to better solutions here. I've been going over this for over a week trying to find the best, most idiomatic way to do it, and I'm out of ideas. So I just went with something that worked and am hoping somebody has a better idea.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -50,6 +50,13 @@ typedef enum {
 } ghostty_split_focus_direction_e;
 
 typedef enum {
+    GHOSTTY_SPLIT_RESIZE_UP,
+    GHOSTTY_SPLIT_RESIZE_DOWN,
+    GHOSTTY_SPLIT_RESIZE_LEFT,
+    GHOSTTY_SPLIT_RESIZE_RIGHT,
+} ghostty_split_resize_direction_e;
+
+typedef enum {
     GHOSTTY_INSPECTOR_TOGGLE,
     GHOSTTY_INSPECTOR_SHOW,
     GHOSTTY_INSPECTOR_HIDE,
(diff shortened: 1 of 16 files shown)


mitchellh (commented):
This all genuinely looks really good, especially considering the complexity. Excellent work! 😄 I downloaded it and it also works great.

I think my _only_ real concern or question is around the use of environment values in `SplitView`. Its been awhile since I used them so I guess a couple questions:

  1. Does this make them required? It'd be nice to be able to use `SplitView` without them set (i.e. the inspector for now, which doesn't want to support resize events yet, but also maybe future split view usages).

  2. Could the resize events and published not be sent in via normal means to the SplitView? I'm SURE you tried this and there is an explanation, I'd just love to understand it.

PR #838, opened 2023-11-07
macos: add key binding for equalizing split sizes

gpanders:
Fixes: https://github.com/mitchellh/ghostty/issues/831

Of course please feel free to tweak the default keybinding/name/etc.

Demo:

https://github.com/mitchellh/ghostty/assets/8965202/1abb00a9-4548-4b1a-bde0-b3952997d3eb


mitchellh (commented):
This looks pretty much perfect. Going to give it a test go now!

davidhellmann:
Can we have that as a config setting to set it as default?

PR #850, opened 2023-11-09
Fully implement OSC 4, 10, 11, 12, 104, 110, 111, and 112

gpanders:
Fixes: https://github.com/mitchellh/ghostty/issues/722

Commits are broken out individually.


mitchellh (changes requested):
Looks very good, as usual. 

There is one comment I have directly on the current implementation.

I think a broader issue I see (which doesn't have to block merge): currently if you reload config, it will override and reset the fg/bg/palette, I think? The behavior we probably want is that the config defines the _default_ values that are used if the terminal program hasn't set them. What do you think?

I'm happy to merge this first and address that later, but want to check if we should track that for the future.

PR #876, opened 2023-11-14
Add support for desktop notifications

gpanders:
Fixes: https://github.com/mitchellh/ghostty/issues/612

This PR adds support to core for OSC 9 and OSC 777 for sending desktop notifications from terminal applications. For now, the actual notifications are only implemented for macOS, but all of the core plumbing is there to make it work with GTK (it just needs updates to the apprt surface).

### DIFF
--- a/dist/linux/app.desktop
+++ b/dist/linux/app.desktop
@@ -9,6 +9,7 @@ Keywords=terminal;tty;pty;
 StartupNotify=true
 Terminal=false
 Actions=new-window;
+X-GNOME-UsesNotifications=true
 
 [Desktop Action new-window]
 Name=New Window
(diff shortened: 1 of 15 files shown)


mitchellh (commented):
This looks really good! Thanks for doing this. I think there are a few issues to chase down.

The big one I found is the nature of packing the pointer address into the user info of a notification. If you trigger a notification, then close the surface, then click the notification, you get a segfault (maybe, but in any case you're accessing illegal memory). That's an issue.

The second, I'm not sure if this is just a problem with my config. After approving notifications, notifications aren't showing up until I click the control center area. If I change my system settings to "alert", they show up, but it defaulted to "banners" for some reason. I can see you request alert privileges. 

The third is more minor, but it's unclear that a notification is from a terminal program and looks officially like Ghostty. That freaks me out a bit, I'm unsure if there is a way we can word the notification title or body in some way to better signal this...

PR #899, opened 2023-11-17
core: implement select all binding

Raiden1411:
Fixes (#892)


mitchellh (commented):
Basically perfect, just two requests for doc comments.

PR #909, opened 2023-11-18
macos: associate SurfaceViews with UUIDs

gpanders:
Add a unique ID (UUID) to each surface view and use it to associate
SurfaceViews with notifications. This is safer than using the address of
the surface view as the identifier, as there is no risk of addressing
invalid memory in the case where the view is freed while the
notification is still active.

---

This also fixes https://github.com/mitchellh/ghostty/issues/907, though is not mutually exclusive with https://github.com/mitchellh/ghostty/pull/908 as both are improvements.

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -39,8 +39,7 @@ class TerminalController: NSWindowController, NSWindowDelegate,
         super.init(window: nil)
         
         // Initialize our initial surface.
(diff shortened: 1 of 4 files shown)


mitchellh (commented):
This looks pretty good, and is definitely in line with what I was thinking the future direction of business logic should be within the Swift app. I wasn't sure, but I felt I'd have to go in this direction to enable things like dragging a split from one window to another (when we do split dragging). 

I want to think about this one a bit more but in general, this looks good.

PR #951, opened 2023-11-24
nix: add GitHub Actions workflow

vancluever:
This adds a nix workflow that does the following:

* Checks to see if the Zig cache fixed derivation needs its hash updated. It does this by downloading a new cache, calculating the hash, and comparing it against what has already been set. If the hash is updated, the new hash is committed.
* Runs a Nix build. This runs off of a hash from the last job to ensure it accounts for any updates to the hash during this workflow.

Note that this is currently set to run off of master as repository state can vary in GHA between pull requests from mainline branches and forks. Specifically, in the latter, GHA always seems to checkout as a detached HEAD, which prevent pushes from happening.

The cache workflow we use in the build job comes from:
  https://github.com/DeterminateSystems/magic-nix-cache-action/

Fixes #937.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,30 @@
+on: [push, pull_request]
+name: Nix
+jobs:
+  check-zig-cache-hash:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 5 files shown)


mitchellh (changes requested):
Looks really good, thanks!

PR #963, opened 2023-11-28
[DRAFT] Add rectangle select

vancluever:
This adds rectangle select mode; when dragging with ctrl+alt (or super+alt on MacOS), this allows you to select a rectangular region of the terminal instead of the full start-end points of the buffer.

https://github.com/mitchellh/ghostty/assets/10704423/5903eaba-b5c3-4a40-9922-998c3a5ba42e

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2209,6 +2209,17 @@ pub fn cursorPosCallback(
     }
 }
 
+// Checks to see if super is on in mods (MacOS) or ctrl. We use this for
+// rectangle select along with alt.
+//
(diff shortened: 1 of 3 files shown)


mitchellh (commented):
I like the idea! 

Okay the big one is **selectionString**. I actually have significant changes to this function in my hyperlinks branch ("urls"). My new implementation is a lot simpler and I think will get rid of some of your headaches, I wonder how we should reconcile that.... maybe I should cherry pick it over somehow or make a partial PR before I finish up the hyperlink work.

PR #969, opened 2023-11-30
gtk: make sure WM_CLASS is being set on X11

vancluever:
This ensures that WM_CLASS is being set on X11, looks like this was missed in #816 (not 100% sure if it always needs to be set in GDK or if it's a side effect of us not using `g_application_run`).

This also adds the "prgname" config attribute to control the instance name, which defaults to "ghostty" (or "ghostty-debug" on debug builds).

I've also added the default values to the config documentation.

Before:

```
WM_CLASS(STRING) = "", ""
```

After: 

```
WM_CLASS(STRING) = "ghostty-debug", "com.mitchellh.ghostty-debug"
```

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -156,6 +156,35 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         return error.GtkApplicationRegisterFailed;
     }
 
+    const display = c.gdk_display_get_default();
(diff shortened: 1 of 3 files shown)


mitchellh (commented):
Thanks for looking into this! You figured out something that we tried and failed awhile back to do. 

I guess my question is... should we just make `class` set the X11 class? Do we need to introduce a new configuration? _If we do_, should we name it something more specialized like `gtk-x11-class` or something? I also have some concerns about the `set_prgname` API I asked in the comments.

None of this is meant to discourage this PR. I want to get this PR in, just trying to figure out how it all relates.

PR #979, opened 2023-12-01
Surface: set crosshair, change event processing logic for mouse tracking

vancluever:
This work is mainly targeted at adding the crosshair for when ctrl/super+alt is pressed. We also add this for when mouse tracking is enabled so that we show the crosshair when ctrl/super+alt+shift is pressed at the same time.

I've also changed the event processing logic here because the amount of keys we have to process has greatly increased. Instead of processing each individual event, we now process the modifier state.

Additionally, some refactoring has been done geared at starting to re-work the mouse for the core surface into a something stateful. My hope is that we can continue to unravel some of this from the core
surface so that we can process key inputs, motion events, and anything else relevant as inputs to transitions for shape display, click behavior, etc.

This commit now also moves the `ctrlOrSuper` handlers to respective parts in the `Key` hierarchy, while also adding additional helpers for other modifiers.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -37,6 +37,7 @@ const input = @import("input.zig");
 const App = @import("App.zig");
 const internal_os = @import("os/main.zig");
 const inspector = @import("inspector/main.zig");
+const SurfaceMouse = @import("surface_mouse.zig");
 
 const log = std.log.scoped(.surface);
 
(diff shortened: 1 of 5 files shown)


mitchellh (changes requested):
Thanks, huge improvement. A couple I think easy pieces of feedback and this looks good!

PR #1001, opened 2023-12-06
Powerline: Add half-circle rendering

vancluever:
This adds in-terminal rendering for the powerline glyphs E0B4 and E0B6, similar to how we are rendering the triangle shapes currently.

The circle glyphs use a much more complex rendering due to the nuances of drawing them: we use a midpoint algorithm for drawing on a 4x supersampled matrix, fill, and then downsample. We use the same downsampling approach as is done in the arc box drawing code.

The midpoint variant we're using here is described by Dennis Yurichev: https://yurichev.com/news/20220322_circle/, although there are similar variants elsewhere (some cited at the bottom of his article).

### DIFF
--- a/src/font/sprite/Face.zig
+++ b/src/font/sprite/Face.zig
@@ -148,6 +148,8 @@ const Kind = enum {
 
             // Powerline fonts
             0xE0B0,
+            0xE0B4,
+            0xE0B6,
             0xE0B2,
             0xE0B8,
             0xE0BA,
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
I reviewed about half of it. 

Big picture: I really like it 😄 

Small picture, details: I think I have some concerns around the types, `@as` usage, etc. I left comments and hopefully that guides you in the right direction. Let me know if you have any questions.

PR #1062, opened 2023-12-12
gtk: implement equalize_splits

mrnugget:
This adds support for the equalize_splits feature that's already implemented for macOS.

It's essentially a port of the Swift implementation, using the same weights-mechanism to equalize split sizes.

I confirmed that the behaviour is the same by comparing various split configurations and their equalized result.

@mitchellh I'm not 100% sure about the `f16` and the conversion to/from int/float. I tried to match existing conventions I could spot in the codebase, but you have better eyes here when it comes to Zig.

### Demo

[equalize_splits_gtk.webm](https://github.com/mitchellh/ghostty/assets/1185253/47b144cb-f9c0-433f-8468-4aed7314407a)

### DIFF
--- a/src/apprt/gtk/Split.zig
+++ b/src/apprt/gtk/Split.zig
@@ -172,6 +172,46 @@ pub fn moveDivider(self: *Split, direction: input.SplitResizeDirection, amount: 
     c.gtk_paned_set_position(self.paned, new);
 }
 
+/// Equalize the splits in this split panel. Each split is equalized based on
+/// its weight, i.e. the number of Surfaces it contains.
+///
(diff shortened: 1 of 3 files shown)


mitchellh (changes requested):
Looks good. Maybe I'm missing something, but I don't see a reason to use `i16` and `f16` everywhere. I'd use `f64` and then convert that to `c_int` at the end as you're doing. Is there any reason the `equalize` function can't also return a float? You convert them almost immediately so I feel like if you just use `f64` everywhere and covert to `c_int` at the end then you'll eliminate most casts.

mitchellh (approved):
Looks perfect. Thanks!

PR #1077, opened 2023-12-13
gtk: close tabs with middle-click

meatcar:
Follow the behaviour of other tabbed interfaces like browsers, editors, and file-managers.

GTK code is based on `/src/apprt/gtk/ImguiWidget.zig`


ghuls:
Can this behavior be disabled with a config option?
I accidentally closed a tab a lot of times when trying to paste stuff with middle click.

PR #1094, opened 2023-12-14
termio: implement DECRQSS

gpanders:
Only SGR, DECSCUSR, DECSTBM, and DECSLRM are handled, as these are the only ones that Ghostty supports (as far as I can tell) and are the only ones that seem actually useful.


mitchellh (commented):
Looks great. One minor thing!

PR #1132, opened 2023-12-20
Use "ghostty" subdirectory for resources

gpanders:
Installing resources directly under `${prefix}/share` causes conflicts with other packages. This will become more problematic whenever Ghostty is opened and becomes packaged in distributions.

Instead, install all resources under a "ghostty" subdirectory (i.e. `${prefix}/share/ghostty`). This includes themes, shell integration, and terminfo files.

Only `/usr/share` style paths use the "ghostty" subdirectory. On macOS, Ghostty is already isolated within its app bundle, and if `$GHOSTTY_RESOURCES_DIR` is set then we assume that points to the actual resources dir (without needing to append "ghostty" to it).

This PR also adds a build time `-Dresources-dir` option. This was suggested by @jcollie in Discord as something that might be useful for packagers. If omitted (the default), then Ghostty's behavior is the same as it is today (it traverses the filesystem upward relative to the `ghostty` binary looking for the terminfo file).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -288,7 +288,7 @@ pub fn build(b: *std.Build) !void {
         const install = b.addInstallDirectory(.{
             .source_dir = .{ .path = "src/shell-integration" },
             .install_dir = .{ .custom = "share" },
-            .install_subdir = "shell-integration",
+            .install_subdir = b.pathJoin(&.{ "ghostty", "shell-integration" }),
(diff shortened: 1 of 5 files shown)


mitchellh (commented):
One comment that probably complicates this PR a bit. It's the main reason I didn't do this earlier! But, your work looks good and if you want you can consider that one complication its not _too bad_.

PR #1140, opened 2023-12-21
(DRAFT) Powerline: TinyVG additions, some double-width changes

vancluever:
This commit makes the following changes:

* Introduces TinyVG rendering for specific Powerline codepoints. There are a number of glyphs in the Powerline family that are too complex to be reasonably rendered with simple drawing commands alone, and are better represented as vectors. For these, we are leaning on TinyVG (https://tinyvg.tech/) to do the rasterization with the glyphs embedded as TinyVG within the resource directory. This initial commit includes the solid flame codepoints (E0C0 and E0C2).

* Updates the handling for Powerline glyphs that are supposed to be rendered as double-width. For these, we double the canvas width for the sprite rendering via a codepoint handler during the initial width calculation. We also remove the exemptions for all double-width Powerline codepoints from the fgMode handler in cell.zig (additionally renaming the handler to isPowerlineSingle to better communicate its purpose).

## Before

image

## After

image


mitchellh (commented):
Looks great, a couple comments!

PR #1176, opened 2023-12-28
Use adw_application_new() instead of adw_init()

tristan957:
This will enable further libadwaita improvements in the future, but will have no effect today.


mitchellh (commented):
Thanks, this seems fine to me, couple little nitpicks. I'd clean them up myself but since you have more stuff coming in may be helpful to see the feedback. Plus one question!

PR #1193, opened 2023-12-31
apprt/gtk: ensure modifier state matches current keypress under X11

vancluever:
This fixes an issue in that when running under X11, when a modifier key is pressed, the modifier state will "lag" behind what should be current. This is due to how X11 sends modifiers in events, i.e. it sends the state from right before the key press, and does not include the effects of the key press itself.

This is corrected by checking the X event queue directly for a pending XkbStateNotify event (we mask this on modifiers), and setting the modifiers off of that if we find one. If not, we fall back to the GDK call.


mitchellh (changes requested):

mitchellh (commented):
Thanks! I made some stylistic changes, and one major functional change: I moved the X11 function lookups into a struct so we can use comptime to make it easier on us. Can you verify everything is still working?

I also left some inline questions...

PR #1194, opened 2023-12-31
 Add exit code and runtime to abnormal exit error message, make threshold configurable.

jcollie:
Added runtime and exit code to the error message.

image


mitchellh (commented):
Looks great. Going to make the one type change.

PR #1204, opened 2024-01-03
Ability to set preferred fallback fonts

mitchellh:
Fixes #1196 
Fixes #886 

The `font-family` set of configurations are now repeatable. Repeating the value specifies a font fallback in case a font is not found.

This is a breaking change, since `font-family` would previously _overwrite_ the previous value so it was a way to _change_ the font. This is no longer possible. To facilitate this use case, I've added a special sentinel value that can be used with any `RepeatableString` type: if the value is empty (`""`) then the list will be reset.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -342,7 +342,7 @@ pub fn init(
             // A buffer we use to store the font names for logging.
             var name_buf: [256]u8 = undefined;
 
-            if (config.@"font-family") |family| {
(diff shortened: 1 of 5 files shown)


ilyagr:
Is there a way to specify Ghostty's default font-face as the first face, and then fallbacks from it (e.g. I wanted to try "Symbols Nerd Font Mono" and see how it affects https://github.com/ghostty-org/ghostty/issues/886#issuecomment-2570334389)?

PR #1253, opened 2024-01-08
implement enquiry/answerback

jcollie:
Needs a little more testing before merging. One question that I had would it be better for the configuration entry to be called "answerback-string" instead of "enquiry-string".

### DIFF
--- a/docs/sequences/enq.md
+++ b/docs/sequences/enq.md
@@ -10,11 +10,9 @@ operator.
 
 ## Implementation Details
 
-- ghostty always sends `""`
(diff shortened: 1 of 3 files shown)


mitchellh (commented):
Looks like a good start. The number of allocations required (I think... 3 dupes?) seems highly suspicious to me, but I haven't looked into more detail around lifetimes to think about it carefully. 

I don't see a mechanism for this to reload at runtime safely, as far as I can tell, either.

The core idea seems good.

PR #1267, opened 2024-01-09
input: manage application keypad state with mode 1035 

mitchellh:
Fixes https://github.com/mitchellh/ghostty/issues/1099

We previously applied application keypad mode logic (`ESC=` or mode 66)
whenever it was active. However, from looking at the behavior of other
terminals (xterm and foot) it appears this isn't correct.

For xterm, application keypad mode only applies unconditionally if the
keyboard mode is VT220 (`-kt vt220`). For modern terminals, application
keypad mode is only applied if mode 1035 is disabled.

Mode 1035 is the "ignore numpad state with keypad mode" mode. It
defaults to true on terminal startup. If this is true, keypads are
always encoded in numerical mode. If this is false, the numlock state
will be respected along with the desired keypad mode.


briancain (approved):
Fixed my issue! 🎉 👻

PR #1278, opened 2024-01-10
gtk: use modifier state from GTK key event on X11

gpanders:
Fixes: https://github.com/mitchellh/ghostty/issues/1274

On X11 `gdk_device_get_modifier_state` does not correctly return the
modifier state, while the modifier state passed to the key callback
does. On Wayland, the situation is exactly reversed.

Therefore on X11 we use the mods provided by the key callback and on
Wayland we continue to get the modifier state from the device.

`App.modifier_state_from_xkb` is removed since we can lift the
conditional out of the function call (we would need to make a second,
redundant check for the presence of `x11_xkb` otherwise).


vancluever (approved):
Looks great! :shipit:

PR #1279, opened 2024-01-10
add ability to specify RGB colors as names from the X11 rgb name list

jcollie:
fixes #1277


mitchellh (commented):
Thanks for the work here. I think I want to approach this a different way. I'll take over on this one because I can use some of the work you did here -- so it's not pointless and I do appreciate it!!

Mainly: there's no need to depend on the colors tar.gz. The file is MIT/X11 licensed so we can download rgb.txt (only 18 KB uncompressed) and store it directly in our source tree. Further, we don't need an external executable `generate_rgb_names` to generate Zig source. We can use comptime to `@embedFile` and process it directly at compile-time and generate the entries directly.

This will avoid the `build.zig` changes completely. The rest of the changes look pretty good, just personal style choices I'll probably apply but the core ideas are all good. 

I've got it from here. Thank you!

PR #1373, opened 2024-01-25
WIP / Discussion: D-Bus for GTK app runtime

jcollie:
This PR is posted for early feedback. D-Bus offers a lot in terms of integrating into the Gnome environment, offering things like search integrated from Gnome Shell, integration with other desktop programs by integrating with XDG desktop protocols, etc.

This code compiles and "runs" but the D-Bus functionality is nonfunctional at this point. Right now there's minimal changes to the "core" of Ghostty but some functionality (esp search) will require deeper changes to Ghostty eventually.


tristan957 (commented):
I think I would like to see more separation like:

- src/apprt/gtk/dbus/gnome.zig
- src/apprt/gtk/dbus/xdg-desktop-portal.zig

That way it is easy to see what files implement what dbus interfaces/requests

PR #1397, opened 2024-01-28
Add super+shift+<left, right, up, down> select expansion

ayatofrench:
## Change Log
- Add Shift+Super+<left, right> to expand selection by word
- Add Shift+Super+<up, down> to expand selection by paragraph

## Demo
[ghostty-super-shift-select-2024-01-27_23.05.17.webm](https://github.com/mitchellh/ghostty/assets/6732096/d8dfce76-5502-4ade-9c8b-cee379f43543)

## Notes
- Although `selectWord` considers both whitespace and non-whitespace words, selection expansion by word ignores whitespace words. This is because I am working with the assumption that a user would likely want to select the next non-whitespace word when using this feature.

Issue: https://github.com/mitchellh/ghostty/issues/1210#issuecomment-1890497607

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1293,10 +1293,10 @@ pub fn keyCallback(
         const sel = sel: {
             const old_sel = screen.selection orelse break :adjust_selection;
             break :sel old_sel.adjust(&screen, switch (event.key) {
-                .left => .left,
-                .right => .right,
-                .up => .up,
-                .down => .down,
+                .left => if (event.mods.super) .super_left else .left,
+                .right => if (event.mods.super) .super_right else .right,
+                .up => if (event.mods.super) .super_up else .up,
+                .down => if (event.mods.super) .super_down else .down,
                 .page_up => .page_up,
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Thanks! A couple points of feedback, one in comments and then:

1. Would love to see tests for this. The function you modified is well unit tested so we should cover the cases.

2. It'd be much preferred if we could reuse `selectWord` or `selectWordBetween` and `selectLine` for the super left/up/etc. behaviors rather than reimplementing it slightly differently in the adjust function. This way if we change the logic of either, it'll match up with adjustments, and vice versa.

pratyushseq:
@ayatofrench @mitchellh any update on this?

ayatofrench:
> @ayatofrench @mitchellh any update on this? Really miss this from iTerm.

Hey! Sorry, I dropped off the radar on this one. I got super busy with my firstborn. I have time now and if Mitchell is okay with it, I can work on this feature again.

PR #1411, opened 2024-01-29
Tint split view divider based on theme background

peteschaffner:
This creates a more subtle divider that better melds with the background:

CleanShot 2024-01-29 at 22 33 45@2x

CleanShot 2024-01-29 at 22 34 32@2x

CleanShot 2024-01-29 at 22 35 34@2x

CleanShot 2024-01-29 at 22 36 18@2x

CleanShot 2024-01-29 at 22 36 43@2x


mitchellh (changes requested):
Just one bit of feedback. Overall supportive of this approach though.

PR #1422, opened 2024-01-31
Add "Open with..." capability to GTK runtime.

jcollie:
This PR adds all of the necessary plumbing to use "Open with..." from Nautilus (and other Gnome/GTK apps). This will open any files in an editor in a new Ghostty window. The editor chosen can be configured either with the `EDITOR` environment variable or can be configured in the Ghostty config. If neither options are available it will fall back to `vi`.

image
[Screencast from 2024-01-31 15-05-48.webm](https://github.com/mitchellh/ghostty/assets/740022/0ce71ee6-1144-4ffb-83a1-ccba87425765)

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,6 +561,29 @@ on the search path for a lot of software (such as Gnome and KDE) and
 installing into a prefix with `-p` sets up a directory structure to ensure
 all features of Ghostty work.
 
+### Linux D-Bus Debugging Tips
+
+The GTK runtime uses D-Bus behind the scenes for sending signals and other features. Debug and release
+builds use different names and paths on the D-Bus bus so that they can run simultaneously.
+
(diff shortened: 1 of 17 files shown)


mitchellh (changes requested):
Thank you, excited to support this. Left various feedback.

mitchellh (changes requested):
Alright, left a couple easy comments, and a couple harder ones.

There are a couple major issues I have with this PR and at this point I think I need to context load and dive in to resolve them myself, or feel free to take another pass. I think the app pointer is a huge smell to me, the apprt message changes are also suspicious (I understand your point about the app not being setup yet), etc. 

It's possible to make it work would require a larger refactor. I'm not sure, but at this point it seems like I probably need to context load. That'll probably require a delay, sorry (until I have time to sit down and really sift through this)

PR #1474, opened 2024-02-06
Generate fish command completions for Ghostty.

jcollie:


mitchellh (changes requested):
Instead of a standalone exe, I believe we can generate this completely at comptime and export a string. See `src/config/vim.zig`. Let's do that.  Due to Zig's lazy analysis, if the string is not referenced, it won't be included in the binary so the actual Ghostty binary won't have it.

PR #1479, opened 2024-02-07
Remove titlebar tab shadows

peteschaffner:
In a light appearance there was a shadow cast on top of the inactive tabs that looked bad. In a dark appearance the shadow manifests as a solid line that required a faux border to look right. This removes both and improves the aesthetics of the tab bar in light themes.

This also adds some `NSView` extensions to help with the subview spelunking that this titlebar tabs project requires.

Before:
CleanShot 2024-02-07 at 11 22 20@2x

After:
CleanShot 2024-02-07 at 11 23 34@2x


qwerasd205 (commented):
I like this change! Could use some comments in the code though to explain what's going on at some points, since without context it otherwise seems quite mysterious. Also needs doc comments for the extension methods you've added.

Other than that-- there could be a nicer way to prevent the separator views, but if this doesn't seem to have a significant impact it's probably fine- other than the part where you use a DFS to check for the titlebar container, since if it doesn't exist (which it doesn't always due to stuff like disabling window decorations, non native fullscreen, fullscreen without tb tabs enabled, etc.) then you'll search the whole view hierarchy for no reason.

qwerasd205 (commented):
LGTM

PR #1484, opened 2024-02-08
macos: configurable titlebar fonts

jparise:
Add support for configurable fonts for window and tab titles. This is only implemented for macOS (and could be macOS-only if other platforms aren't able to support this using their windowing toolkits). It plays nicely with regular and titlebar tabs.

For example, with `window-title-font-family = "Iosevka Fixed"`:

### One Tab
image

### Regular Tabs
image

### Titlebar Tabs
image


qwerasd205 (approved):
LGTM! I looked over the code and everything seems in order, and I pulled it down and messed with it a bit and didn't spot any issues; seems to work as intended.

PR #1488, opened 2024-02-09
Replace ziglyph.codePointWidth with unicode.table.get(cp).width

erf:


mitchellh (commented):
Thanks! Preedit is an exceptionally rare codepath so this isn't super hot but we should probably just unify on one API anyways.

PR #1490, opened 2024-02-09
Keep shadows for native tab bar

tt:
#1479 removed tab shadows.

I think it was intended to only do so when `macos-titlebar-tabs` is enabled. However, it also applied to the native tab bar which makes Ghostty appear different from apps like Safari or Terminal.app.

I guess it's a preference if the shadows should be removed from both modes. I personally like when it looks like other apps.

## Screenshots

The two screenshots show before and after this change. The first tab is active and the last is hovered.

Tab bar without shadows

Tab bar with shadows


peteschaffner (changes requested):

peteschaffner (approved):
LGTM, thanks!

PR #1506, opened 2024-02-11
GTK: Add compiled-in GTK resources and use them for icons.

jcollie:
Use glib-compile-resources to compile CSS style sheets and icons into the Ghostty GTK binary. Makes for simpler access to icons and sets things up for customizing the look of Ghostty with CSS in the future. The CSS style sheets are blank for now so there will be no visual changes.


tristan957 (commented):
This is awesome. I had wanted to add gresource support to use GtkBuilder files. Lucky for me, you beat me to it.

It would be great if the GTK apprt could start moving its UI code into GtkBuilder files. I had started on that in a branch that I would be happy to share with you if it is something that interests you.

PR #1519, opened 2024-02-13
Native Win32 App Runtime

marler8997:
Adds a native win32 app runtime.  Uses OpenGL for rendering (same as the current glfw apprt).

Status:
* keyboard input: working
* mouse: position, left/right buttons and scroll wheel working
* clipboard: selection copy to clipboard working

image

image

The current implementation has some parts that are "solid" and should work long term while others will need more work.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -263,6 +263,7 @@ pub fn build(b: *std.Build) !void {
 
         if (target.result.os.tag == .windows) {
             exe.subsystem = .Windows;
+            exe.addIncludePath(.{ .path = "dist/windows/inc" });
             exe.addWin32ResourceFile(.{
                 .file = .{ .path = "dist/windows/ghostty.rc" },
             });
--- a/build.zig
+++ b/build.zig
@@ -1053,6 +1053,7 @@ fn addDeps(
     step.root_module.addImport("opengl", opengl_dep.module("opengl"));
     step.root_module.addImport("pixman", pixman_dep.module("pixman"));
     step.root_module.addImport("ziglyph", ziglyph_dep.module("ziglyph"));
+    const zigwin32_dep = b.dependency("zigwin32", .{});
 
     // Mac Stuff
     if (step.rootModuleTarget().isDarwin()) {
@@ -1166,6 +1167,10 @@ fn addDeps(
             .gtk => {
                 step.linkSystemLibrary2("gtk4", dynamic_link_opts);
                 if (config.libadwaita) step.linkSystemLibrary2("adwaita-1", dynamic_link_opts);
+            },
+
+            .win32 => {
+                step.root_module.addImport("zigwin32", zigwin32_dep.module("zigwin32"));
             },
         }
     }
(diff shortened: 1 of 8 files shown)


leroycep:
Would it make sense for a win32 runtime to use DirectX, since the MacOS runtime is using Metal? From my (very possibly flawed) recollection Windows doesn't officially support OpenGL; it is just common for GPU manufacturers to implement OpenGL and expose it once you install their GPU drivers.

That said, I'm not sure where or if this would matter in a practical sense.

ehartford:
I really love the idea of having a single terminal that I can run on all my machines whether they are Windows, Mac, or Linux.  I would certainly switch all of my machines to use ghostty.

Kazppa:
is this PR still in progress ?
I'm willing to give a hand though.

mitchellh:
Okay, it's time to close this PR. I'm 100% committed to Windows on the roadmap, and I appreciate all the work here. But I don't plan on bringing first class Windows support in the short term (hopefully before the end of the year though that's not a promise). I don't want an idle PR open in the mean time. 

The best course of action here for people who want to work on this I think would be to maintain a fork with a win32 (or other windows API) apprt. I'd be more than happy to discuss upstreaming that at some point but it would allow a group of people to work towards that goal in the mean time. 

Thanks everyone here, and I hope to reach back out sooner than later. ❤️

cavanaug:
@marler8997 are you going to maintain a fork somewhere for this, or is this a dead effort at the moment?

der-teufel-programming:
I wanted to set up a fork with the Win32 support if he doesn't

marler8997:
No plans to create/maintain a fork at this time.

I'll make myself available to consult when Mitchel is ready to bring windows up.  In addition, I can point you to a port I did of the "flow text editor" which is a terminal text editor into a native win32 GUI application.  The renderer could be lifted and re-used for ghostty, or, just used as a reference.

https://github.com/neurocyte/flow

Most of the relevant code will be in the `src/win32` directory.  It includes some nice window management code that snaps the size to a multiple of the cell size as well as all the input/keyboard code ghostty will need.  It's uses a GPU-accelerated renderer via Direct3D that is based on Casey Muratori's refterm.  At a high level, it's just a simple design for a terminal, which is just a texture to hold the rendered glyphs and a bit of shading code that takes an array of all the characters to render for each frame.  This implementation provides all the window management/setup code and also uses DirectWrite to render the text.  The text renderer is also easy to switch out via a comptime interface.

There's also https://github.com/marler8997/winterm.  This is a terminal I started and might develop on a bit more which uses the same renderer as flow, so, an extra reference in case it helps.

I'll also mention that I still plan on finishing a WinRT API generator which would allow Zig programs to access/use the WinUI API.  I haven't done so yet but once that's done it should be pretty easy to experiment with WinUI as well, although, based on some usage of WinUI I sometimes find the performance to be lacking especially when compared against a more "handmade" approach.  I've had good success with Direct2D for UI's but I'll admit it take more time/code to create.

der-teufel-programming:
@cavanaug I will open the branch and put it on Discord for everyone interested in working on it

mehmetkursataydin:
I can't wait for the Windows release :)

PR #1526, opened 2024-02-15
Avoid repainting new tab button for native tab bar

tt:
The new tab button was repainted in #1501.

However, it is also repainted when `macos-titlebar-tabs` is not enabled. This means that if, say, your theme's background color is dark but you're not in dark mode, the color of the button will not match the system appearance.

This pull request applies the same fix as in #1490 to avoid repainting.

## Screenshot
image


peteschaffner (commented):
Crap this is the second time I messed up the standard title/tab bar with my changes 🤦‍♂️ Really sorry about that and thank you for catching and fixing it!

I had one idea I wrote in a comment, but all-in-all this LGTM.

PR #1550, opened 2024-02-24
Add un-zoom button to titlebar and tabs

peteschaffner:
This change is one half of the proposal made in #1541, and adds a "un-zoom" button to the titlebar and individual tabs.

https://github.com/mitchellh/ghostty/assets/411395/cbd7e851-e608-4a75-aa2c-b686bd571d3d

Traditional title/tab bar:
CleanShot 2024-02-24 at 14 18 08@2x
CleanShot 2024-02-24 at 14 18 23@2x

Titlebar tabs:
CleanShot 2024-02-24 at 14 19 14@2x
CleanShot 2024-02-24 at 14 19 32@2x

Inactive tab (button doesn't respond to mouse-down in this state but focuses the tab):
CleanShot 2024-02-24 at 14 21 07@2x

In order to make this feature available for those using the default title/tab bar (i.e. not `macos-titlebar-tabs`), I added a toolbar to house the button in the event there is no tab bar visible. This could be considered a regression for some, however, as it makes the titlebar taller (from ~28pt to 38pt). We could make it possible to hide the toolbar, which could mitigate the "regression", but it would also make the button inaccessible if the tab bar isn't visible.

I also made the traditional title/tab bar adopt the same style as titlebar tabs, in that they use the background color of the terminal window. That is out of scope for this addition, but it seemed a quality-of-life change worth making while I was in the neighborhood :) Assuming most agree that this makes for a nicer window, it puts a bit more urgency on my finding a way to make the tab bar usable with [really dark/black backgrounds](https://github.com/mitchellh/ghostty/pull/1483#issuecomment-1939625405) (which I've created issue #1549 to track).

Lastly, I'm using the system accent color to tint the button's icon, but ultimately I'd like to pull one of the ANSI 16 colors defined by the terminal theme, but I'll need some help to expose the Zig `Palette` struct to the C API (cc @mitchellh).


jparise (commented):
I'm trying to think of a phrase other than "unZoom" to suggest instead, but I'm coming up blank on good ideas. Maybe something with "restore" in its name?

jparise (commented):
Latest revision isn't working for me ("Zoom Split" action doesn't appear to do anything), but I don't have the time right now to look more closely.

PR #1626, opened 2024-03-29
fix: remove deprecated to-pixdata gresource preprocessor

tristan957:
It was deprecated in gdk-pixbuf 2.32.

Link: https://docs.gtk.org/gio/struct.Resource.html


jcollie (approved):
LGTM

PR #1642, opened 2024-04-01
font/sprite: improve rendering of dashed lines

qwerasd205:
Previous implementation would draw dashes to the edge of the character cell, which would result in double-wide dashes at the point where they tiled. This fixes that, and also generally implements it in a cleaner way than before.

I made the decision that extra "left-over" length should be distributed in to dashes rather than gaps, since in my opinion it's a *lot* more noticeable when there are un-even gaps than when there are un-even dashes.

CC @vancluever

**Before** (left) and **after** (right):
image


mitchellh (commented):
Looks great, one comment would love your input on.

vancluever (approved):
LGTM here, comptime/codegen conversation notwithstanding!

PR #1690, opened 2024-04-17
Fix background-opacity for the GTK backend

TheOnlyMrCat:
The `background-opacity` config option previously applied transparency to the entire window, not just to the window's background, making it not very useful. I've made it now apply only to the background.

image

For the most part, this works by changing which widgets have the `background` CSS class. A small hack was needed to add this class to the tab bar, but it's necessary to be able to keep the theme's background colour.

This PR works as-is, but I've put the necessary CSS inline into the Zig code. Would you prefer it be added to the `style.css`, etc. files?


mitchellh (changes requested):
Some small requests but looks good so far. Thank you for doing this!

PR #1694, opened 2024-04-17
config: window-inherit-working-directory options

### LINKED ISSUE #1392: Allow different `window-inherit-working-directory` behavior for tabs and windows
I would like to have different behaviors for the `window-inherit-working-directory` option, either through a new option like `tab-inherit-working-directory` or a command that can be bound through a `keybind` option.

Use case would allow for news tabs to inherit the working directory while a new window would open the home directory, or vice versa.

jparise:
`window-inherit-working-directory` was previously a boolean value. This change makes it a more granular option, allowing you to enable this behavior for one or more surface "kinds": split, tab, window.

The full behavior is only implemented for the embedded runtime and macOS app. The other runtimes still need their own plumbing to track surface kind creation (stubbed as `.window` by default in this change).

Resolves #1392

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -32,6 +32,12 @@ typedef void* ghostty_inspector_t;
 
 // Enums are up top so we can reference them later.
 typedef enum {
+  GHOSTTY_SURFACE_KIND_SPLIT,
+  GHOSTTY_SURFACE_KIND_TAB,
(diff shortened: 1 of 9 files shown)


mitchellh (changes requested):
So far so good. Am I missing where the kind is actually set from the apprts?

PR #1729, opened 2024-05-05
shell-integration: automatic bash integration

jparise:
This change adds automatic bash shell detection and integration.

Unlike our other shell integrations, bash doesn't provide a built-in mechanism for injecting our ghostty.bash script into the new shell environment.

Instead, we start bash in POSIX mode and use the ENV environment variable to load our integration script, and the rest of the bash startup sequence becomes the responsibility of our script to emulate (along with disabling POSIX mode).

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,10 +201,10 @@ The currently supported shell integration features in Ghostty:
 
 #### Shell Integration Installation and Verification
 
-Ghostty will automatically inject the shell integration code for `zsh` and
-`fish`. `bash` does not support automatic injection but you can manually
-`source` the `ghostty.bash` file in `src/shell-integration`. Other shells are
-not supported. **If you want to disable this feature,** set
+Ghostty will automatically inject the shell integration code for `bash`, `zsh`
+and `fish`. Other shells do not have shell integration code written but will
+function fine within Ghostty with the above mentioned shell integration features
+inoperative. **If you want to disable automatic shell integration,** set
 `shell-integration = none` in your configuration file.
 
 **For the automatic shell integration to work,** Ghostty must either be run
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -408,7 +408,7 @@ palette: Palette = .{},
 /// Ghostty does not do any shell command parsing.
 ///
 /// If you're using the `ghostty` CLI there is also a shortcut to run a command
-/// with argumens directly: you can use the `-e` flag. For example: `ghostty -e
+/// with arguments directly: you can use the `-e` flag. For example: `ghostty -e
 /// fish --with --custom --args`.
 command: ?[]const u8 = null,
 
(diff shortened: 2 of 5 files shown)


akinomyoga:
The codes added in this PR seem to be a direct translation (i.e., replacing "kitty" with "ghostty") of [the one in kitty](https://github.com/kovidgoyal/kitty/blob/492ec3dfbf3be6f27603baabb5c2844f91436e04/shell-integration/bash/kitty.bash#L6-L59). However, Ghostty's license is the MIT license, which is incompatible with kitty's GPL-3.0. Wouldn't that be a problem?

jparise:
> I also thought we rewrote the bash integration at one point to be clean room anyways. But maybe not. Cc @jparise

It was originally a lot more basic, but when we were working through earlier compatibility challenges, I incorporated the additional flags, etc. from kitty.

Now that we've worked out those integration issues, I think we could remove some of that script complexity and just not support automatic script injection in those (rare) scenarios (e.g. `--noprofile`), but that's a separate discussion.

This is some preexisting recognition that some of this file's logic comes from kitty:

https://github.com/ghostty-org/ghostty/blob/bf3597a519164b81ca86f205d657557fc23fa8b7/src/shell-integration/bash/ghostty.bash#L1-L3

... but that doesn't call out the license aspect like we do in the zsh script.

akinomyoga:
Thank you for taking time to answer my question.

> The shell integration isn’t embedded or part of our binary

In my understanding, that is a *necessary* condition that the shell integration and the main binary are considered two separate programs, but that's not *sufficient* to claim they are separate. For example, consider a GPL-licensed project written in an interpreter programming language. If GPL only applies to a program compiled into a single binary, GPL doesn't make any sense for such a project.

What I wondered were the following two points:

The first is whether the shell integration is part of the Ghostty feature set or an external plug-in for Ghostty. To claim it is a separate product, we need to consider the shell integration *a plug-in* or *an extra feature*.

The second point is whether the "plug-in" establishes an [*intimate communication*](https://www.gnu.org/licenses/gpl-faq.html.en#GPLPlugins) with the main Ghostty process. By nature of the shell integration, once the shell-integration code is loaded in Bash, the instance of the shell integration continues to be inside Bash and establishes mutual communication with the main Ghostty process. This coupling seems stronger than the [*borderline case*](https://www.gnu.org/licenses/gpl-faq.html.en#GPLPlugins) invoking the plug-in and waiting for its final results.

- The part added by this PR just handles the initialization, but the latter part of `ghostty.bash` implements mutual communication, and it would be hard to claim the former and latter parts of a single file are two separate programs.
- Even in this case, it might be possible to claim that both Ghostty and the shell integration implement a public shell-integration protocol (including the injection part utilizing `ENV`) and make the shell-integration script replaceable with another implementation (such as `kitty.bash`) in the user configuration. But I'm not so sure about this.

Technically, even if the shell integration is not considered a separate program, we can continue to use the GPL-licensed code until the copyright holder contacts us. In this sense, it might not be urgent but I think it would be good to clarify the stance on those points.

> I also thought we rewrote the bash integration at one point to be clean room anyways. But maybe not. Cc @jparise

At least the codes merged in this PR are still present in the latest Ghostty. Actually, it's the other way around. I first found the shell-integration code in the latest version, and then I reached here after trying to find where to raise the discussion.

akinomyoga:
Yeah, the borderline of GPL is always difficult. Thank you for sharing your thoughts.

PR #1760, opened 2024-05-11
Rate limit desktop notifications

jcollie:
Add a once-per-second rate limit for all desktop notifications and a once-per-five-seconds rate limit for duplicate desktop notifications.

Fixes #1759.


mitchellh (changes requested):
One request, I also agree we should use a non-cryptographic hash and Wyhash is the typical one we use.

mitchellh (approved):

PR #1761, opened 2024-05-12
fix(renderer/Metal): dirty track link highlights

qwerasd205:
Partial fix for link highlight dirty tracking. Some issues remain with soft-wrapped links (see TODO in code), as well as pressing cmd while hovering without moving the mouse at all doesn't highlight until you do move the mouse for some reason.

More should be done to improve this, of course, but this is a small change that at least restores 99% of functionality/usability to the link highlighting.


mitchellh (commented):
My links are still pretty flaky with this PR so I want to take a more detailed look, but I agree this PR helps some so I'll merge as is. Thanks!

PR #1774, opened 2024-05-16
feat(shell-integration): add automatic integration for Elvish

ulzu:
Depends on PR #1772.

Fish automatic integration taken as an example.
Just like Fish, Elvish checks `XDG_DATA_DIRS` for its modules.
Thus, Fish integration in zig is reused, and integration in
Elvish now removes `GHOSTTY_FISH_XDG_DIR` environment variable
on launch.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3414,6 +3414,7 @@ pub const ShellIntegration = enum {
     bash,
     fish,
     zsh,
+    elvish,
 };
 
 /// Shell integration features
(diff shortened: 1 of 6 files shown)


jparise (commented):
Aside from the minor ordering (alphabetization) note, the `shell_integration` bits look simple-enough to me. Maybe we want to generalize the `XDG_DATA_DIRS` code into a common function instead of having `setupElvish` call `setupFish`?

Lastly, could you also update the main `README.md`?

PR #1782, opened 2024-05-20
Avoid coloring when using native tabs

tt:
As noted in https://github.com/mitchellh/ghostty/pull/1767#issuecomment-2119361525, the background color leaks through when using the native tab bar.

This is the same issue that was addressed in #1527. (Interestingly, the screenshot there is slightly darker so I guess it wasn't addressing it fully.)

## Screenshots

The two screenshots show before and after this change.

image

image


mitchellh (commented):
Thanks, this looks good. I flipped the logic a bit and added some more comments. Can you please verify my logic is correct? Its easy to flip this stuff...

PR #1872, opened 2024-06-21
Various Performance Changes

qwerasd205:
This PR comprises 4 separate performance-focused improvements. These are generally explained for the most part by the commit messages and comments in files, but I'll give a brief summary here:
1. Introduces CacheTable structure, a "forgetful" hashtable that can be generically used for caching various things; right now only applied to the shaper cache.
2. Introduces RefCountedSet structure, a generic structure for efficiently storing and querying reference counted items with associated IDs; right now only applied to `Style.Set`.
3. In the CoreText shaper, cache fonts between calls to `shape`, to avoid running the expensive `CTFontCreateCopyWithAttributes` method repeatedly.
4. Create dedicated thread to call `CFRelease` on CoreFoundation objects that are produced during shaping, since these calls can result in expensive cleanup in an object's release callback.

### Benchmarks
To validate the improvements from these changes, I ran a benchmark after each one, measuring `rebuildCells` times during a 5 second run of `cmatrix -u0` (first 15 samples are removed) - since `cmatrix` performance under the CoreText shaper being poor was the motivating issue that I aimed to address when looking in to these changes.

image

### Note
I have one more change in mind. The shaper interface as used by the renderer code currently results in some serious bottlenecks, wasted duplication of expensive work, and cache inefficiencies. I have a subtle but important change to the shaper interface planned which I've validated on my test branch to have a significant impact since it addresses the main problems quite well. I was originally going to include the change in this PR but life has been getting in the way lately so I decided it's better to get these changes in now since they can stand on their own. When I get around to implementing it, I expect the `rebuildCells` times under that benchmark to be around 2000µs, which ought to be good enough I think.

### DIFF
--- a/pkg/macos/foundation/string.zig
+++ b/pkg/macos/foundation/string.zig
@@ -19,6 +19,17 @@ pub const String = opaque {
         )))) orelse Allocator.Error.OutOfMemory;
     }
 
+    pub fn createWithCharactersNoCopy(
+        unichars: []const u16,
+    ) *String {
+        return @as(*String, @ptrFromInt(@intFromPtr(c.CFStringCreateWithCharactersNoCopy(
+            null,
+            @ptrCast(unichars.ptr),
+            @intCast(unichars.len),
+            foundation.c.kCFAllocatorNull,
+        ))));
+    }
+
     pub fn release(self: *String) void {
         c.CFRelease(self);
     }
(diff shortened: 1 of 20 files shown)


mitchellh (commented):
This looks extremely excellent. I didn't review the full thing yet, and the comments are mostly notes for me to follow as I add some changes, but this is very, very, very good work.

PR #1874, opened 2024-06-23
Make vim ftdetect file pattern more flexible

PeterCardenas:
The current file detect pattern is not flexible to custom `XDG_CONFIG_HOME` locations, but for my use case it's not flexible to editing the config files with [chezmoi](https://chezmoi.io), which renames the `.config` directory to `dot_config`. This change makes the file pattern only match `ghostty/config` with any parent directory.


jparise (approved):
Looks good!

PR #1880, opened 2024-06-25
Fix a few RefCountedSet problems

qwerasd205:
Fixes #1873 and a few problems that made themselves known after/while fixing that one, specifically:
- Memory leak in `PageList.clone`
- Under-utilization of possible `Id` space because capacity size limited by `Id`
- Bad input causes repeated `OutOfMemory` errors to grow the set beyond what it can grow to or what is necessary-- fixed by introducing `NeedsRehash` error for cases where new IDs are not accessible to the Set but at least 10% of the currently used IDs belong to dead items.

### DIFF
--- a/src/inspector/page.zig
+++ b/src/inspector/page.zig
@@ -37,7 +37,7 @@ pub fn render(page: *const terminal.Page) void {
         }
         {
             _ = cimgui.c.igTableSetColumnIndex(1);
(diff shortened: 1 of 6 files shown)


mitchellh (commented):
Thanks! I want to add tests for ref counted set still but im satisfied eyeballing this to approve it. I added a test for the PageList change though!

PR #1929, opened 2024-07-09
Reflow rework, various fixes

qwerasd205:
- Fixes #1857
- *Probably* fixes #1851
- Also fixes various other reflow-related page content corruptions.

The new reflow impl passes all tests (some of which were testing for incorrect behavior before, which I fixed) and passes my manual sanity check where I run the terminal, create a bunch of scrollback with various things in it, then resize a bunch to see if any of it ends up getting messed up or any asserts are hit.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1753,7 +1753,7 @@ pub fn scrollCallback(
         };
     };
 
-    log.info("scroll: delta_y={} delta_x={}", .{ y.delta, x.delta });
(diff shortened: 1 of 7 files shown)


mitchellh (commented):
Only one comment and question.

All commits up to the final commit I very carefully reviewed and look good. The final commit is the big one and is hard to really deeply understand without redoing this work myself, but I didn't see anything obviously wrong. 

Thank you! 😄

PR #1955, opened 2024-07-17
terminal: implement in-band size reports (Mode 2048)

mitchellh:
Fixes #1953 

https://gist.github.com/rockorager/e695fb2924d36b2bcf1fff4a3704bd83


rockorager (approved):
Looks good, Tested it out and works as expected. Thanks!

PR #1960, opened 2024-07-17
fix: remove dollar sign from code snippets

o-az:
currently when copying code snippets one has to manually remove the dollar sign first. Consider:

```sh
$ echo "this can't be copy pasted into terminal because of the initial $"
```
`bash: $: command not found`
`fish: Expected a variable name after this $.`
vs.

```sh
echo "this is fine"
```


jparise (commented):
Might also be nice to consistently use `shell-session` for all of these markdown code blocks as part of this change.

PR #1979, opened 2024-07-20
apprt/gtk: implement context menu

kareigu:
Implements a context menu for GTK:

## Without selection:
image

## With selection:
image

Tried to match behaviour in other GTK apps with the menu preferring to open down and right from the cursor, if possible.

## A couple notes:
Left a FIXME in place for disabling copy while nothing is selected. Currently I'm just setting the button action to `noop` which feels really hackish, but from trying multiple different approaches this was the best one by far. (Suggestions welcome).

Setting sensitivity to disabled on the menu item didn't seem to work from what I tried, while removing the menu item completely causes some nasty artifacts where the menu is displayed when it still hasn't finished doing layout again.

**Which brings me to the second note:**
Currently we're storing a reference to the same `PopoverMenu` and reusing it when we could be doing the same approach as with AppKit and just creating it on the fly as needed. 

This however causes some problems like the menu becoming a scrollable list for some reason (this I suspect to be just a skill issue on my end, since I can't find other GTK apps doing the same thing) and having this ugly relayout happening when you open the menu. 

*Nautilus creates a new `PopoverMenu` every time you right click on something*

https://github.com/user-attachments/assets/c49730ae-04eb-4a3f-9b07-b88b32ee7ef1

The video doesn't really show it well on anything other than the longer menus, but in person it's really noticeable.

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -797,13 +797,7 @@ fn initContextMenu(self: *App) void {
     const menu = c.g_menu_new();
     errdefer c.g_object_unref(menu);
 
-    {
-        const section = c.g_menu_new();
-        defer c.g_object_unref(section);
-        c.g_menu_append_section(menu, null, @ptrCast(@alignCast(section)));
-        c.g_menu_append(section, "Copy", "win.copy");
-        c.g_menu_append(section, "Paste", "win.paste");
-    }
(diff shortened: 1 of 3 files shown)


mitchellh (changes requested):
Thanks for doing this! Looks pretty good. 

Noted on your gotchas, I'll take a look but I'd be fine merging the logic as-is today and looking into improving it later too.

PR #2021, opened 2024-08-01
gtk: allow running in the background

jcollie:
This patch fixes #2010 by implementing `quit-after-last-window-closed` for the GTK apprt. It also adds the ability for the GTK apprt to exit after a delay once all surfaces have been closed and adds the ability to start Ghostty without opening an initial window.


mitchellh (changes requested):
One fairly sizable request but everything else looks good.

jparise (commented):
This looks good to me! Just calling out a few things I noticed while reading through the code.

PR #2033, opened 2024-08-03
Include file/line context for config file errors

wpaulino:
Closes #1063.

As per my [comment](https://github.com/ghostty-org/ghostty/issues/1063#issuecomment-2259601763) in the issue, I opted to only include context for file errors. Happy to discuss alternatives if more context for CLI config errors is desired. I also improved the error formatting a bit along the way.


mitchellh (commented):
Not doing reviewing but a couple simple things so far. Thank you!

PR #2062, opened 2024-08-07
Metal optimizations

qwerasd205:
This PR significantly reworks the Metal shaders for memory efficiency. Additionally I've adjusted the formatting of the shader code for readability, and normalized the naming conventions.

There are two main changes that affect memory efficiency.
1. Background cells are now a static sized buffer of `cols * rows` total `uchar4`s representing the cell colors, which are drawn with a single fullscreen shader.
2. I've taken steps to optimize the foreground (text) cell vertex data, namely:
    - Background color (4 bytes) has been removed, since it can be fetched from the bg color buffer based on grid position.
    - I demoted the glyph bearings from `[2]i32` to `[2]i16`, it seems like a sane expectation to me that the bearings will never be more than 32K pixels in a given direction.
    - I have reorganized the structs to be as compact as possible in memory, so that the size (with padding) is now 32 bytes, as opposed to the prior size which was something like 56 bytes.

Beyond this, I have added a couple other small optimizations, such as a fullscreen vertex shader that uses a single triangle, and using pixel coordinates to sample the glyph texture so that conversion to normalized coordinates isn't necessary.

### Results
Experimentally I saw a ~20% reduction in total GPU memory use, and a ~80% reduction in max allocated buffer size.

- [ ] Verify correct behavior on an Intel mac under macOS 13. **DO NOT MERGE WITHOUT TESTING**

### DIFF
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -1371,9 +1371,9 @@ fn drawPostShader(
         void,
         objc.sel("drawPrimitives:vertexStart:vertexCount:"),
         .{
-            @intFromEnum(mtl.MTLPrimitiveType.triangle_strip),
+            @intFromEnum(mtl.MTLPrimitiveType.triangle),
             @as(c_ulong, 0),
-            @as(c_ulong, 4),
+            @as(c_ulong, 3),
         },
     );
 }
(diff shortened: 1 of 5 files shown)


mitchellh (changes requested):
Looks amazing, some really tiny comments. I'll try to figure out the Intel builds.

PR #2071, opened 2024-08-10
gtk: add resize overlay

jcollie:
This adds a transient overlay that shows the size of the surface while you are resizing the window or the surfaces.

[Screencast from 2024-08-09 21-20-20.webm](https://github.com/user-attachments/assets/d75ee853-c44b-49ef-8d14-cdb517a29d8a)


mitchellh (commented):
Really awesome. I didn't dig into the GTK APIs but did a high level review with some requests but this is exciting to see!

PR #2113, opened 2024-08-17
Implement Kitty Color Protocol (OSC 21)

jcollie:
Kitty 0.36.0 added support for a new OSC escape sequence for quering, setting, and resetting the terminal colors. Details can be found [here](https://sw.kovidgoyal.net/kitty/color-stack/#setting-and-querying-colors).

This fully parses the OSC 21 escape sequences, but only supports actually querying and changing the foreground color, the background color, and the cursor color because that's what Ghostty currently supports. Adding support for the other settings that Kitty supports changing ranges from easy (cursor text) to difficult (visual bell, second transparent background color).

### DIFF
--- a/src/inspector/termio.zig
+++ b/src/inspector/termio.zig
@@ -259,6 +259,11 @@ pub const VTEvent = struct {
                 }
             },
 
+            .Struct => try md.put(
+                key,
+                try alloc.dupeZ(u8, @typeName(Value)),
+            ),
+
             else => switch (Value) {
                 u8 => try md.put(
                     key,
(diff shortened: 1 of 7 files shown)


mitchellh (changes requested):
Very excellent work! Truly! I've noted some changes I'd like to see. Thank you!

mitchellh (commented):
Thanks! I want to play around with one thing but the rest looks good.

mitchellh (changes requested):
Couple more changes requested. I can get to them soon or you can, either way. They're small.

PR #2115, opened 2024-08-18
Index should create scrollback anytime top scroll region is top line

mitchellh:
Fixes #2112. See that issue for details.

~~This slows down this case by 3x using a custom vtebench benchmark I made (see below)~~. With the custom vtebench benchmark below, speed has improved in this PR over `main`. All standard vtebench benchmarks appear unaffected, they must not test this case.

## vtebench

vtebench doesn't have a benchmark covering this case, so I wrote one:

setup:

```sh
#!/bin/sh

# tty="/dev/$(ps -o tty= -p $$)"
# lines=$(tput lines < $tty)
# the above wasn't working, change this to whatever your terminal is
lines=23

printf "\e[$lines;1H"
printf "X"
printf "\e[1;$((lines - 1))r"
printf "\e[$((lines - 1));1H"
```

benchmark:

```sh
#!/bin/sh

# Scroll all lines up as quickly as possible.

printf "y\n"
```

Results:

image

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -757,45 +757,8 @@ pub fn cursorScrollAbove(self: *Screen) !void {
     assert(self.cursor.y < self.pages.rows - 1);
 
     const old_pin = self.cursor.page_pin.*;
-    if (try self.pages.grow()) |new_page_node| {
-        // We allocated a new page and went to it. In this case, our new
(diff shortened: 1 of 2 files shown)


mitchellh (commented):
Adding a couple TODOs for myself that I forgot to consider.

PR #2120, opened 2024-08-19
macos: jump to last_tab

edmz:
(Somehow, I can't make super+0 trigger the action. Other bindings do work).

(No changes were made to GTK.)


mitchellh (changes requested):
Thanks. Noting some changes. I can get around to them if you don't.

PR #2126, opened 2024-08-20
osc: add parsing for Kitty desktop notifications (OSC 99)

jcollie:
Adds the ability to parse [Kitty desktop notifications](https://sw.kovidgoyal.net/kitty/desktop-notifications/) but doesn't do anything with them yet.

### DIFF
--- a/src/inspector/termio.zig
+++ b/src/inspector/termio.zig
@@ -199,18 +199,37 @@ pub const VTEvent = struct {
             void => {},
             []const u8 => try md.put("data", try alloc.dupeZ(u8, v)),
             else => |T| switch (@typeInfo(T)) {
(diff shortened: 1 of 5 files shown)


mitchellh (changes requested):
Thanks. One comment specific to this but see some of the changes I made to the Kitty Color Protocol parser. 

Additionally, I'm starting to think the Kitty-specific structs should PROBABLY move to `src/terminal/kitty/` since that's where we put everything else. Including the color protocol. I'll make that move later.

That'll make it easier to read these structs (less indentation) and easier to isolate tests for them for any logic we might have.

mitchellh (changes requested):
A bunch more relatively easy comments. I still haven't deeply studied the format of the protocol to understand some of the decisions here bigger picture.

PR #2169, opened 2024-08-30
renderer: implement blinking cells

pluiedev:
Tested on OpenGL and it works just fine, though Metal is completely untested at the moment — I don't even know if it compiles. Need a hand from Mac users


anna328p:
Are there any plans to implement this still? This PR seems to have gotten overlooked.

pluiedev:
I am planning to implement this but it kinda got nuked in the process of going public.. Will open a new PR later

PR #2178, opened 2024-09-01
kitty graphics: performance enhancements

jcollie:
Improve the performance of Kitty graphics by switching to the WUFFS library for decoding PNG images and for "swizzling" G, GA, and RGB data to RGBA formats needed by the renderers.

WUFFS claims 2-3x performance benefits over other implementations, as well as memory-safe operations.

Although not thorougly benchmarked, performance is on par with Kitty's graphics decoding.

https://github.com/google/wuffs

This is a `perf` report of using `mpv` to play a video back using the Kitty image protocol from before the patch:

before

After the patch:

after


mitchellh (changes requested):
Couple things I want to look at but a few big ones to request!

Thanks for doing this, the results look great.

PR #2201, opened 2024-09-06
Wide cell boundary conditions in ECH & DCH + soft-wrap reset correctness

qwerasd205:
Adds tests for boundary conditions involving wide chars for ECH & DCH and fixes them with a helper function `Screen.splitCellBoundary`.

Also introduces `Screen.cursorResetWrap` which resets the wrap state for the cursor row (and the cursor's pending wrap) in a correct way, accounting for spacer heads and the previous and next row's wrap flags as well. A few functions in `Terminal` previously just did `self.screen.cursor_row.wrap = false` and called it a day which can result in orphaned spacer heads and improper `wrap` or `wrap_continuation` flags on the prev and next rows.

An important `TODO` I left is in `eraseChars`, technically the boundary condition handling is still insufficient there, since it's not accounting for protected cells creating additional boundaries, or eliminating the two boundaries that are actually being handled at all.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -3172,34 +3172,6 @@ pub const Pin = struct {
         set.set(self.y);
     }
 
-    /// Resets the soft-wrap state of the row this pin is on, appropriately
-    /// handling the wrap and wrap_continuation flags for the previous and
-    /// next row as well.
-    ///
-    /// DOES NOT handle clearing spacer heads.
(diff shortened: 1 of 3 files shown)


mitchellh (changes requested):
Thanks, the tests look good, some feedback on the other parts.

mitchellh (changes requested):

PR #2228, opened 2024-09-12
apprt/gtk: add option to color header bars with theme colors

Pangoraw:
This adds an option to color the header bar with the config (theme) colors. Implements #2150 only for libadwaita.

I am not sure `window-theme=theme` is the most descriptive name for the variant here, maybe something like `window-theme=config` would be better?

Here is an example with `adw-toolbar-style=flat` and `theme=WarmNeon`

image


mitchellh (changes requested):
Awesome. This looks great. I don't mind this being Adw only, we should just document it as such.

PR #2231, opened 2024-09-12
gtk: terminal bell

jcollie:
Use GTK's media streaming APIs to implement the terminal bell. By default it's disabled since it's likely that only a small subset of people will want this feature. Currently only audio support is implemented but there's scaffolding for several other methods of alerting the user.

NOTE: GTK needs to have access to GStreamer and the "base" and "good" plugins or it crashes when you try to play a sound. I've added the necessary packages to the NixOS developer shell and package but I haven't found a method yet to detect the situation and avoid the crash.


mitchellh (changes requested):
I think my biggest issue is the complexity Wayland introduces. 

A few requests:

* Can we remove Wayland from this PR? 
* In the future, can we use the Wayland C API rather than Zig? That'll make it easier for us to upgrade Zig versions?

PR #2243, opened 2024-09-13
config: interpret leading ? in config-file as an optional file

gpanders:
For example, `config-file = ?config.foo` will not report an error if `config.foo` does not exist.


mitchellh (changes requested):
Big thinker comment.

PR #2270, opened 2024-09-19
macOS Secure Input

mitchellh:
Fixes #1325 

This implements macOS secure input as a global application option. macOS secure input is an OS feature that makes it so that applications using event tapping functionality cannot capture keystrokes, making it useful for typing in sensitive environments. 

To mimic other applications (Terminal, iTerm), I've added the global option to the application menu bar. I've also added a `toggle_secure_input` binding action that is unbound by default. If it is bound, it is synced with the menu item and shown as an accelerator.

There are still some TODO items before I'm ready to merge, but the core of this is already in a reviewable state.

image

cc @qwerasd205 

## TODO

- [x] UI indication that secure input is enabled for the given surface.
- [x] Automatic secure entry for password input (#2262)
- [x] Persist across restarts to match Terminal.app
- [x] Introduce configuration for hiding the overlay
- [x] Introduce configuration to disable secure input automatically on at password inputs

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -464,6 +464,7 @@ typedef void (*ghostty_runtime_show_desktop_notification_cb)(void*,
 typedef void (
     *ghostty_runtime_update_renderer_health)(void*, ghostty_renderer_health_e);
 typedef void (*ghostty_runtime_mouse_over_link_cb)(void*, const char*, size_t);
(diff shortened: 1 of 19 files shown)


evanrelf (commented):
Saw this activate at a `sudo` password prompt and came searching for the change out of curiosity. Very cool!

Leaving a comment because I think there might be a bug with the popover.

PR #2276, opened 2024-09-20
macOS: add `macos-titlebar-style = hidden`

qwerasd205:
A titlebar style which hides the titlebar without removing the other typical window frame elements.

> [!NOTE]
> This is a fairly slapdash implementation and I didn't bother making live reload of this config option play nice, it causes some strange behaviors.
>
> Also, I didn't bother testing this with a variety of configurations, there are almost certain to be some little bugs and weird incompatibilities that need to be tracked down.

Feel free to make any changes you feel necessary before merging.


mitchellh (approved):
Looks good to me.

PR #2295, opened 2024-09-24
Various underline (and strikethrough) improvements

qwerasd205:
Improvements to underline and strikethrough rendering, including:
- Reworked the underline sprite font rendering for saner results in terms of positioning
- Adjusted strikethrough position metric calculations
- Made underlines draw *before* text, so that they don't harm legibility as much for text with descenders

### CoreText
|main|this PR|
|-|-|
|image|image|

### FreeType
|main|this PR|
|-|-|
|image|image|

### TextEdit
image

### DIFF
--- a/pkg/macos/text/font.zig
+++ b/pkg/macos/text/font.zig
@@ -188,6 +188,14 @@ pub const Font = opaque {
         return c.CTFontGetUnderlineThickness(@ptrCast(self));
     }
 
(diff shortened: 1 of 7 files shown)


mitchellh (changes requested):
This all looks very good, I have a few very minor pieces of feedback.

Besides this, are you happy with this? This seems like a good improvement thank you.

PR #2296, opened 2024-09-24
cli: "fancy" theme preview

jcollie:
This adds a "fancy" theme preview to the `+list-themes` CLI action. By default, if the command is connected to a TTY, it will display the fancy preview. If it is not connected to a TTY, or the user specifies `--plain` on the command line, a simple list of themes will be printed to stdout.

An update to `libvaxis` is required for system theme detection to work.

While in the preview `F1` or `?` will show a help screen.
Screenshot from 2024-09-23 21-16-43 
Screenshot from 2024-09-23 21-16-51

https://github.com/user-attachments/assets/e413bdb6-acdf-444b-a17a-3bb531188fce

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1043,6 +1043,10 @@ fn addDeps(
         .target = target,
         .optimize = optimize,
     });
+    const zf_dep = b.dependency("zf", .{
+        .target = target,
+        .optimize = optimize,
+    });
 
     // Wasm we do manually since it is such a different build.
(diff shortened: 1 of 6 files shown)


mitchellh (commented):
This looks great. Thanks for doing this. 

In debug builds, vaxis logs are showing up and breaking the TUI. I'm not sure why they're not using our logger actually since it should be global to the Zig process... but we may want to just disable the `--preview` mode for now in debug builds.

PR #2313, opened 2024-09-28
sgr: add support for legacy double underline

rockorager:
SGR 21 is defined to be a double underline. This behavior is common
among many terminals, notably xterm.

References: [https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_)


creack (approved):
LGTM

PR #2318, opened 2024-09-29
gtk: add "search with selection" keybind action and context menu item

jcollie:
This adds the ability to launch an internet search with any selected text on the active surface. If there is no selected text, the action is a no-op. Search providers are configured using the `search-provider` configuration entry and can be chosen from a set of predefined options or a supplied custom URI.

This only implements the context menu for GTK but extending this to macOS should be fairly easy as most of the logic is in the core. The keybind action should work on macOS as-is although it is not assigned a keybind by default.

Screenshot from 2024-09-28 17-19-30
Screenshot from 2024-09-28 17-20-52

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -252,6 +252,7 @@ const DerivedConfig = struct {
     window_padding_balance: bool,
     title: ?[:0]const u8,
     links: []Link,
+    search_provider: configpkg.SearchProvider,
 
     const Link = struct {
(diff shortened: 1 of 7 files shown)


gitmpr:
I created this for myself as well. seems like there is at least some shared interest into this after all, especially on linux where this is not a part of the native gtk4 ui library. An optional, off by default feature would be nice.

PR #2324, opened 2024-09-29
terminal: select whole links if regex matches

ryzech:
Still needs tests added, and a general lookover but it does work as intended.

Added a new function under `src/terminal/Screen.zig` called `selectLink`, this checks the selection against a regex to see if it is a link, otherwise returns null. 

Updated the logic in `src/Surface.zig` so it checks the link first, and if it catches null it selects a word instead.

Feel free to nitpick anything, might've missed something not 100% familiar with the codebase yet.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2735,9 +2735,16 @@ pub fn mouseButtonCallback(
             // word selection where we clicked.
         }
 
-        const sel = screen.selectWord(pin) orelse break :sel;
-        try self.setSelection(sel);
-        try self.queueRender();
+        // Check if the selection contains a link, if it does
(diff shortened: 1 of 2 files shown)


mitchellh (commented):
I would consider modifying `selectWord` to behave this way rather than introduce a `selectLink`. If `selectWord` finds a boundary that is `:` or `/` we can go into "is this a URL" mode. 

It's possible you may want to keep `selectLink` how it is and just fall back to it but I can also see just building this into `selectWord` (and if its ever something we want configurable, introducing an option on the function).

This would have the benefit of expanding selection by word to also work by jumping over hyperlinks, which I _think_ is also desired behavior.

PR #2328, opened 2024-09-30
cli: handle mouse scroll before setting window

Pangoraw:
Scrolling would go past the end so that selection is not seen in the list (see video).

[Screencast from 2024-09-30 11-04-23.webm](https://github.com/user-attachments/assets/69bd175d-8b4e-40a2-8d5d-368d277f46f5)

Now that mouse effects are handled before clamping the selection is correctly set at the bottom as when using the keyboard to scroll.


jcollie (approved):
LGTM

PR #2338, opened 2024-09-30
Add an equivalent to iTerm's "show/hide all windows" hotkey (Fix for: #2205)

rolandpeelen:
First time contributor, so if I'm violating something, please let me know.

This feature should fix: https://github.com/ghostty-org/ghostty/issues/2205 (and is coincidentally basically the only thing holding me back from daily driving Ghostty).

The pr adds an action called `toggle_visibility`, which essentially hides and shows all Ghostty windows with a global key (and from the menubar).


mitchellh (changes requested):
Looks great! Some minor requests...

PR #2399, opened 2024-10-06
Undercurl fixes/improvements

qwerasd205:
Fixes #2381 

Additionally, improves the undercurl rendering to look nicer.

- Undercurl wave amplitude is now based on the cell width and not the underline thickness, meaning it will be more uniform.
- The wave now always has enough room and won't be clipped off the bottom.
- The wave is now based on a loose approximation of an offset curve for the (co)sine wave, rather than just another vertically offset wave, this improves the perceived uniformity of thickness and looks a lot better.
- **(Bonus)** Minimum spacing of 2px between the double underline to avoid legibility issues with particularly thin underlines.

### Comparisons
||main|this PR|
|-|-|-|
|**6pt**|image|image|
|***+100% thickness***|image|image|
|**12pt**|image|image|
|***+100% thickness***|image|image|
|**48pt**|image|image|
|***+100% thickness***|image|image|

### DIFF
--- a/src/font/sprite/underline.zig
+++ b/src/font/sprite/underline.zig
@@ -158,18 +158,22 @@ fn drawDashed(alloc: Allocator, width: u32, thickness: u32) !CanvasAndOffset {
 /// the basic math structure for this since I was lazy with the
 /// geometry.
 fn drawCurly(alloc: Allocator, width: u32, thickness: u32) !CanvasAndOffset {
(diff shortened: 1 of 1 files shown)


gnohj:
this seems to be broken again on the nightly release

00-kat:
@gnohj open a new discussion → .https://github.com/ghostty-org/ghostty/discussions/new?category=issue-triage.

PR #2417, opened 2024-10-09
Renderer `rebuildCells` rework

qwerasd205:
This was a long time coming and I went ahead and did it while fixing a bug with background colors behind ligatures.

This is a significant rework of the renderer cell rebuilding, removing a lot of duplicated work from inside of `updateCell` and making things generally slightly more sane.

Remember this performance graph from the last time I improved the shaper performance?

image

:)

> [!NOTE]
> I am unable to test the preedit cell logic for the OpenGL renderer because I'm on macOS.
> Someone else will need to do that before this can be merged.

### DIFF
--- a/src/font/shape.zig
+++ b/src/font/shape.zig
@@ -44,10 +44,7 @@ pub const Cell = struct {
     /// this cell is available in the text run. This glyph index is only
     /// valid for a given GroupCache and FontIndex that was used to create
     /// the runs.
-    ///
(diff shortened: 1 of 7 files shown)


mitchellh (changes requested):
I haven't tried to _understand_ it yet, but here is some superficial feedback since I think you're taking a look at it anyways. Thanks for this!

PR #2440, opened 2024-10-14
gtk: use CSS variables and color calcs introduced in 4.16

jcollie:
This takes advantage of CSS variables and color expressions to improve the `window-theme=ghostty` support. The only visible difference from the previous implementation is that the header bar will darken if the Ghostty window is in the background, which is standard for GTK apps.

This is conditional at runtime. If Ghostty detects that you're running against GTK 4.16 or newer it will use the CSS variables and color calcs. If you're running against older versions it will use CSS classes to achieve nearly the same effect.

cc: @vancluever @tristan957

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -852,45 +852,72 @@ fn syncActionAccelerator(
 }
 
 fn loadRuntimeCss(config: *const Config, provider: *c.GtkCssProvider) !void {
+    var buf: [4096]u8 = undefined;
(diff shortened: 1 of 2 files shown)


tristan957 (commented):
Nice stuff. Did you read all of https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/css-variables.html#header-bar-colors? I see that there are some colors that have relations to the ones you set. Do we need to mess with those at all?

PR #2491, opened 2024-10-24
System packaging guide, build.zig updates, source tarball generation with signature

mitchellh:
This PR introduces a few separate things:

**build.zig updates for system integration modes.** This allows building Ghostty using the standard `zig` flags for using dynamically linked system packages rather than statically build versions. The packaging guide describes how to do this. As part of this, the existing `-Dstatic` flag is _removed_ since this is now solved by the system integration flags.

**A `PACKAGING.md` file targeted to package maintainers.** This has the information I believe system packagers will need to successfully package Ghostty. 

**CI updates for generating a source tarball with a signature.** This is currently just a `git archive` but in the future we can also remove test files and such to make it smaller. The signature is signed via mini sign, with the public key in the `PACKAGING.md` file. 

## Example

Here is an example set of commands to build Ghostty for packaging on Linux:

```sh
$ ZIG_GLOBAL_CACHE_DIR=/tmp/offline-cache ./nix/build-support/fetch-zig-cache.sh

$ DESTDIR=/tmp/ghostty \
zig build \
  --prefix /usr \
  --system /tmp/offline-cache/p \
  -Doptimize=ReleaseFast \
  -Dcpu=baseline
```

The first step requires internet access but the second step does not (satisfying the need of many packaging environments). The first step can have all of its content hashes verified with `build.zig.zon`. 

The second step creates a fully dynamically linked (implied by `--system`) build in `/tmp/ghostty` that can then be "installed" to the proper chroot. Example, the binary is at `/tmp/ghostty/usr/bin/ghostty` in this example.

Packagers could play around with `DESTDIR` and `--prefix` as necessary.


vancluever (approved):
Looks good over here!

BTW, I think out of all of the packaging systems out there, Nix is probably the most restrictive when it comes to sandboxing, in fact I think it's the only one that network sandboxes, or at least it's the only one I've ever ran into network sandboxing issues on; chroots on the other hand are more common. My packaging chops on non-Nix systems are quite rusty... but I'd expect that we could refine the process if need be.

PR #2512, opened 2024-10-28
Make sure a potential port component is considered during hostname validation for OSC 7 handling and shell integration

reykjalin:
## Description

Fixes https://github.com/ghostty-org/ghostty/issues/2484

The reason shell integration would break when macOS users enable the Private Wi-Fi Address is because the hostname is changed to something like `12:34:56:78:90:12`. The hostname passed via OSC 7 is assumed to be a URI, and the URI parser in the Zig standard library (correctly, I think) assumes the last portion before the URI path preceded by a colon (`:`) is a port.

This set of changes makes sure the port is included when comparing the hostname passed in via OSC 7 control sequences to what we get from `posix.gethostname`.

## Testing instructions

1. On your Mac, open Wi-Fi settings &rarr; click the meatball menu for the currently connected network &rarr; click "Network settings…" &rarr; Enable "Private Wi-Fi address". I've set mine to "Rotating", but I don't think that matters so long as it's on.
2. Checkout `trunk` and build the full Mac app by running `zig build && cd mac && xcodebuild && cd ..`
    * **Note:** you **MUST** use the full mac build, the issue doesn't occur with the GLFW runtime.
3. Run the built app `open mac/build/ReleaseLocal/Ghostty.app`.
4. `cd` to any directory that's not the default directory.
5. Open a new tab or split, and notice that the new tab or split is opened in the default directory used when you open a new terminal instead of your current working directory (CWD).
6. Checkout this branch, repeat steps (1)-(5), but make sure the CWD is preserved in step (5).

### DIFF
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -43,6 +43,50 @@ pub fn isLocalHostname(hostname: []const u8) !bool {
     return std.mem.eql(u8, hostname, ourHostname);
 }
 
+test "bufPrintHostnameFromFileUri succeeds with ascii hostname" {
+    const uri = try std.Uri.parse("file://localhost/");
+
+    var buf: [posix.HOST_NAME_MAX]u8 = undefined;
+    const actual = try bufPrintHostnameFromFileUri(&buf, uri);
+
+    try std.testing.expectEqualStrings("localhost", actual);
+}
+
+test "bufPrintHostnameFromFileUri succeeds with hostname as mac address" {
+    const uri = try std.Uri.parse("file://12:34:56:78:90:12");
+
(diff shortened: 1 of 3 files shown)


mitchellh (changes requested):
Looks much better. A couple more requests:

I think we should move `bufPrintHostnameFromFileUri` to `src/os`. It might make sense to just be in a new file maybe `src/os/hostname.zig`. From there, I think we should add a few test cases (`test {} ` blocks) to cover each of the cases.

Additionally, we should make that an explicit error set.

I think `isLocalHostname` could also move in there and be tested.

Thank you!

mitchellh (approved):
Looks great! Just did some stylistic changes. Thank you so much for this!

PR #2514, opened 2024-10-28
apprt: propagate OSC10/11 (set term fore/background color) to apprt

neurocyte:
This is to allow the running apprt to set the UI theme to match the terminal application coloring.


mitchellh (changes requested):
Thanks for separating out the PRs, doing the core work first is a great separation.

Some minor feedback.

Issue #2509, opened 2024-10-27
Feature Request: ability to set title and colour of a tab


markpapadakis:
This is a really useful feature of iTerm2; I use different tabs for various projects and such, and being able to name them/set their title, and doubly so, being able to set a color, makes it a lot easier to switch to the right tabs and work with a tabs-based workflow.

![Image](https://github.com/user-attachments/assets/e23f213a-0fc3-4804-a10c-1cb37cb8743e)

![Image](https://github.com/user-attachments/assets/10b824db-ca2d-4679-a907-a2b4b918d455)

![Image](https://github.com/user-attachments/assets/64800935-e06b-4966-ad81-f7b01482332f)

xiaket:
I assume this goes without saying, but having this as an action that can be triggered by a keybind would be tremendously helpful.

edrozenberg:
Related but a bit different, I'm hoping to be able to use a generic fixed title for windows ("Ghostty"), but allow the title of each tab to automatically reflect the activity/contents. For ex. a window would be named "Ghostty", but each tab would have a different dynamic name based on which server I've ssh'd to in that tab. This is currently possible in Terminal.app for example:

Terminal.app title for windows:
![Image](https://github.com/user-attachments/assets/4c0f3f07-4eae-4589-a90f-5b5d075fa6a3)

Terminal.app title for tabs:
![Image](https://github.com/user-attachments/assets/20945f31-0687-4d4a-8e2e-916e9f0d3226)

felixkreuk:
maybe i missed it in the docs, but is there an option to use escape codes to set the tab title? similarly to:
https://iterm2.com/documentation-escape-codes.html

awhitford:
Should this issue be split into 2 issues:
1. Ability to set the **Tab Title**
2. Ability to set the **Tab Color**

One may be tricker than the other.  I'm personally more concerned about the title.

kbd:
FWIW here's a current example of how I use kitty's tabs and its remote control to make it easy to find the tab I'm looking for:

![Image](https://github.com/user-attachments/assets/7a44efd8-d0bd-4ace-b481-e448408c12b7)

[...]

PR #2515, opened 2024-10-28
WIP: apprt/gtk: style gtk titlebar by osc10/11

neurocyte:
This updates the dynamically loaded titlebar css to match the colors requested by the running terminal application.

This depends on #2514

TODO
- [X] #2514
- [x] #2562
- [ ] Style just the window belonging to the surface (currently styles all windows)
- [ ] Style just the current tab if a tab bar is showing
- [ ] Add a configuration option to disable (enable?) this behaviour

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -473,6 +473,7 @@ pub fn performAction(
         .toggle_tab_overview => self.toggleTabOverview(target),
         .toggle_window_decorations => self.toggleWindowDecorations(target),
         .quit_timer => self.quitTimer(value),
+        .color_change => try self.colorChange(target, value),
 
         // Unimplemented
         .close_all_windows,
@@ -485,7 +486,6 @@ pub fn performAction(
         .key_sequence,
         .render_inspector,
         .renderer_health,
-        .color_change,
         => log.warn("unimplemented action={}", .{action}),
     }
 }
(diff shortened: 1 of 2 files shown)


karlhorky:
@neurocyte I'm not sure if this PR does what I am interpreting that it does. Here's what I interpret currently:

This PR allows user to configure setting a tab background color when a foreground program is running, eg:

1. Nothing running in tab -> default tab background color
2. User starts a foreground blocking progress (eg. a development server)
3. The terminal tab changes background color automatically to a (user-configurable) color
4. User stops the process
5. The terminal tab changes background color automatically to the default tab background color

If that is accurate, then I think there is probably others interested in a feature like this, if you look through some of the preliminary comments in https://github.com/ghostty-org/ghostty/issues/2509

I would also be interested. Currently, I'm [setting a 🚀 (rocket emoji) in iTerm using zsh `precmd` / `preexec`](https://github.com/karlhorky/dotfiles/blob/7665aa84d8493bc2272605ef976751e0e7e64da6/settings/.zshrc#L53-L63) to show at a quick glance that a foreground process is running in a tab. But a background color would be much more visually prominent.

What's important for my use case is the part in parentheses for 3 above - that the color is user-configurable, not only possible to be defined somehow by the program which is running.

jcollie:
@neurocyte any plans to update this and move it forward?

mitchellh:
Thanks for working on this. Going to close this until we can find a more actionable path forward, its been stale for too long.

PR #2523, opened 2024-10-28
macOS: prevent native window drag by top region when titlebar hidden

qwerasd205:
Currently `macos-titlebar-style = hidden` doesn't prevent the native window drag gesture in the top region of the window- in the original PR it did, but there were issues with how it went about it so it was removed (see c6bbdfb). I figured out how to safely and simply remove the gesture, and as a bonus it opens up potential future enhancements.

The native window drag region is driven ultimately by the window's `contentLayoutRect`, so we can just override it in `TerminalWindow` to return a rect the size of the full window, disabling the gesture without causing any side effects by altering the responder chain.

This makes `macos-titlebar-style = hidden` a much nicer experience. The window can still be resized, managed by the OS and third party window managers, and dragged by the edges, but the native drag gesture in the titlebar region is fully avoided.

### Future work
We may consider adjusting this to produce a `contentLayoutRect` that doesn't include the padding area of the terminal grid(s), so that the window can be dragged from a larger region around the edges (and not just the resize region).

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -22,7 +22,7 @@ class TerminalController: BaseTerminalController {
     private var restorable: Bool = true
 
     /// The configuration derived from the Ghostty config so we don't need to rely on references.
-    private var derivedConfig: DerivedConfig
+    private(set) var derivedConfig: DerivedConfig
 
     /// The notification cancellable for focused surface property changes.
     private var surfaceAppearanceCancellables: Set<AnyCancellable> = []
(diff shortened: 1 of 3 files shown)


devshane:
Just wanted to chime in and say that this is actually a feature to me. I was pleasantly surprised that I didn't have to hunt for the window edge to drag it.

mitchellh:
> Just wanted to chime in and say that this is actually a feature to me. I was pleasantly surprised that I didn't have to hunt for the window edge to drag it.

Understood. That wasn't the original intention of hidden titlebars (only incidental) so we should discuss how we can accommodate this desire elsewhere.

00-kat:
> we should discuss how we can accommodate this desire elsewhere

N.B.: someone has opened #5513; I couldn't find any other discussions on this so I guess that discussion would be the preferable place to discuss this.

Issue #2487, opened 2024-10-24
block cursor should grow with `adjust-cell-height`


jesseleite:
Hey @mitchellh! Again, just wanted to thank you for all your incredible work on Ghostty. Super impressed with it, and really appreciate all the love you've put into it! 🔥

Anyway, re: #1072... I know it was intentional that `adjust-cell-height` does not also adjust cursor height, but for those of us using it to make things super spacious, I find it looks weird when my cursor doesn't match the cell height. 

For example, in both Kitty and WezTerm, I was able this kind of look when selecting text in Neovim...

![Image](https://github.com/user-attachments/assets/398842bb-8190-4aee-8f47-7258acf8cc19)

But in Ghostty, look how cursor height is handled on top of my selection in Neovim...

![Image](https://github.com/user-attachments/assets/e761b95b-c2ca-4b70-8b88-58aa666589ca)

Do you think it might be worth adding a config for `adjust-cursor-height` so that we can set cursor height individually from `adjust-cell-height`? Or maybe even `adjust-line-height`, which could act like `adjust-cell-height`, but with different behaviour around things like cursor height? I'm not sure what's cleanest though. What are your thoughts? 🙏

jesseleite:
Solved by https://github.com/ghostty-org/ghostty/pull/3062 ❤️❤️❤️

PR #2580, opened 2024-11-03
Add adjust-cursor-height metric

idr4n:
Hi,

My attempt to address issue #2487 by adding `adjust-cursor-height`, which increases or decreases the cursor height by a percentage of the current one which is linked to the font size.

https://github.com/user-attachments/assets/ea44558f-e4f5-4480-9c70-0b5728e4d74d


mitchellh (changes requested):
Thanks. I want to think about this further in the context of the linked issue. I think having such a configuration is good and I'm willing to merge this separate to the other changes suggested in the original issue, but before doing so want to think about the broader goals of our adjustments.

PR #2626, opened 2024-11-08
feat: implement configurable freetype load flags

nfejzic:
Hello everyone, this is my first contribution to Ghostty. I am neither really familiar with the code-base, nor with Zig language, so please take that into consideration. I am thankful for any and every advice! 

That said, this PR aims to support configuration of (some basic) freetype load flags. In particular, the same freetype load flags as in what [Wezterm supports]. I also borrowed the explanation of the flags from Wezterm's documentation.

Fixes #2622, closes #2069.

[Wezterm supports]: https://wezfurlong.org/wezterm/config/lua/config/freetype_load_flags.html

### DIFF
--- a/src/build/fish_completions.zig
+++ b/src/build/fish_completions.zig
@@ -12,7 +12,7 @@ pub const fish_completions = comptimeGenerateFishCompletions();
 
 fn comptimeGenerateFishCompletions() []const u8 {
     comptime {
-        @setEvalBranchQuota(17000);
+        @setEvalBranchQuota(18000);
         var counter = std.io.countingWriter(std.io.null_writer);
         try writeFishCompletions(&counter.writer());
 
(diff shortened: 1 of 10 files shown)


mitchellh (changes requested):
Much, much better! This feels right. I'm going to go ahead and run CI on this just so you can see if there are any CI issues. I've made some comments/requests but they're overall quite minor.

mitchellh (approved):
Looks good. I made a couple tweaks on style, the only "bug" I found was that `collection()` should use `key.load_flags` rather than the config. It probably wouldn't have resulted in an actual bad behavior but it's the correct thing to do since the key is what is hashed.

I also removed the `bitmap` configuration. That seems a bit dangerous to expose and I want to wait until we have a real use case for it.

PR #2674, opened 2024-11-14
Update shebangs in installed shellscripts to meet `rpmlint` and match…

AnthonyZhOon:
Closes #2672

Update shebangs in scripts that will be installed, previously they failed `rpmlint -i ghostty` for a packaged ghostty installation. 
```
ghostty.x86_64: E: script-without-shebang /usr/share/ghostty/shell-integration/bash/bash-preexec.sh
ghostty.x86_64: E: script-without-shebang /usr/share/ghostty/shell-integration/elvish/lib/ghostty-integration.elv
ghostty.x86_64: E: env-script-interpreter /usr/share/ghostty/shell-integration/bash/ghostty.bash /usr/bin/env bash
```

I chose the `/usr/bin/<script>` form over `/bin<script>` as this is documented as the filepath provided by scripting language packages on [pkgs.org](https://fedora.pkgs.org/41/fedora-x86_64/elvish-0.21.0-4.fc41.x86_64.rpm.html).

This lets me execute the scripts as executables, however testing on the `elvish` config had errors when executing it directly. I don't use `elvish` personally and I'm not if this is fine here.
```
$ ./ghostty-integration.elv
Multiple compilation errors:
  variable $edit:before-readline not found
```


jparise (commented):
On macOS, the (system) shell binary locations are `/bin/bash` and `/bin/zsh`, so changing to `/usr/bin` could cause problems.

In general, though, I believe these are all meant to be sourced rather than executed as standalone scripts, so I don't know how important any of this is in practice.

PR #2696, opened 2024-11-16
macos: support linkable AboutView properties

jparise:
This allows us to enrich the build's commit property as a GitHub link.

This change also displays the property values using a monospaced font, which I think looks a little nicer (especially the commit SHA).

Screenshot 2024-11-15 at 7 57 44 PM


mitchellh (changes requested):
Thanks, looks good to me! Some requests... at least to try or hear your thoughts.

PR #2729, opened 2024-11-19
apprt/gtk: add gtk-tabs-location=hidden

tristan957:
This is only supported on libadwaita. Instead of the normal tab overview button, we will use an AdwTabButton[0].

Link: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.6/class.TabButton.html [0]


mitchellh (changes requested):
Starting to review some of your GTK PRs sorry for the delay.

Some requests but also: I ran this locally with libadw and it didn't work... the buttons in the top menu bar didn't do anything. I'm not sure what's going on. Do you have any ideas?

PR #2749, opened 2024-11-21
apprt/gtk: abstract AdwHeaderBar and GtkHeaderBar

tristan957:
This will make further changes a little bit more readable.


mitchellh (changes requested):
Some superficial requests but I'm good with this after that.

PR #2797, opened 2024-11-24
Support loading config from "Application Support" directory on macOS

kyswtn:
Closes #2634.


mitchellh (changes requested):
Looking good, I have a few I think easy requests. Thank you!

PR #2529, opened 2024-10-29
gtk: add context menu to libadwaita tab headers

jcollie:
Re-uses the context menu from the main app.

PR #2798, opened 2024-11-24
gtk: refactor gtk & adw notebook implementations

jcollie:
Put GTK and libadwaita notebook implementations into separate structs/ files for clarity.

This breaks out the GTK notebook refactor from #2529 and does it without adding an allocation. If/when this is accepted I'll redo #2529 and #2641 to be based on top of this. A PR for #2629 will depend on this too.


tristan957 (approved):
If this goes in, I'll adjust my headerbar PR to be similar.

jcollie:
whoops, accidentally closed...

PR #2809, opened 2024-11-25
[WIP] Add menu item for making Ghostty the default terminal

iheanyi:
I looked at the [source code of iTerm2](https://github.com/gnachman/iTerm2/blob/803acf37f9e016e7d52a7ec5e9028759de8ad491/sources/iTermLaunchServices.m#L241-L247) to figure out how they do this, but this takes the same logic and implements it within Ghostty. There's also functions for setting the Terminal back to the default macOS Terminal, but these are not hooked up to the UI just yet (was unsure if they were wanted). 

Only thing blocking this pull request from being ready is disabling the menu item when Ghostty is already the default. 😅

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -690,10 +690,10 @@ class AppDelegate: NSObject,
         return bundleID.isEqual(defaultTerminal())
     }
     
-    func defaultTerminal() -> String {
+    func defaultTerminal() -> String? {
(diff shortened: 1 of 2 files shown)


jparise (commented):
Just some thoughts after a quick look:

It might be nice to represent this using an `enum` (e.g. `.apple` and `.ghostty`?) and a `defaultTerminal` property with `set` and `get` implementations.

The code could also maybe go in a new `Helper/DefaultTerminal.swift` source file to keep the `AppDelegate` a little leaner.

PR #2818, opened 2024-11-26
Support changing the cursor style for unfocused windows

ginnyTheCat:
This adds the `cursor-style-unfocused` config option (called `cursor-shape-unfocused` in kitty).

Setting it to `bar` for example looks like this when then the window is not focused.
a terminal prompt with a bar cursor

Before it would always show a `block_hollow` like this:
a terminal prompt with a outlined/hollow block at the end

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -442,6 +442,10 @@ palette: Palette = .{},
 ///
 @"cursor-style": terminal.CursorStyle = .block,
 
+/// The style of the cursor when the window is not focused. All valid values
+/// supported by `cursor-style` are accepted here as well.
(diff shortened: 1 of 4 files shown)


mitchellh (changes requested):
Thanks for the PR, I'm fine with making this configurable, but the approach taken in this PR isn't correct. 

The approach I think that would be better is making the unfocused cursor style a part of the renderer `DerivedConfig` (you'll have to update both OpenGL/metal) and then sending that in as a param to the `cursor.style()` function.

mitchellh:
Closing this due to lack of activity. I provided a review with changes I wanted to see. Feel free to make them and reopen, or anyone else for that matter. I appreciate the initial effort though and I'm still supportive of this feature. 😄

PR #2821, opened 2024-11-26
termio: fixes to kitty color reporting

gpanders:
The kitty color report response is an OSC, not a CSI, so change `[` to `]`. Colors which are unset should be reported with `{key}=`.

Ref: https://sw.kovidgoyal.net/kitty/color-stack/

>For example, if the client sends:
>
>```
><OSC> 21 ; foreground=? ; cursor=? <ST>
>```
>
>The terminal responds:
>
>```
><OSC> 21 ; foreground=rgb:ff/00/00 ; cursor= <ST>
>```
>
>This indicates that the foreground color is red and the cursor color is undefined


qwerasd205 (approved):
LGTM

PR #2842, opened 2024-11-27
macos: disable background opacity/blur in native fullscreen

mitchellh:
See #2840


henryhchchc:
image

There is a regression since 1.1.0. The same issue occurs with the latest tip (f0d27606).

PR #2849, opened 2024-11-29
Add ghostty right-click menu item for Plasma users

andrejdaskalov:
Added a right click menu action that launches Ghostty in the given directory, same as how Konsole handles it.
It's added at build time in the ```$prefix/share/kio/servicemenus``` directory. It is ignored by other DEs.
image


mitchellh (changes requested):
Thanks! One question in the code and another question: do you agree to relicense your commits to MIT?

PR #2870, opened 2024-12-02
Add alt keybindings for ctrl+ins = Copy and shift+ins = Paste

deftdawg:
On non-MacOS desktop environments (Windows, Gnome, KDE, Xfce, VS Code, ...), `ctrl+ins` and `shift+ins` are system-wide alternate keybindings (except for terminals) for `Copy` and `Paste` respectively.   This PR explicitly defines them as such in Ghostty's default keybindings.

Using `ctrl+ins` as an alt-keybinding for `Copy` allows static/context-unaware remapping of `Copy` to `cmd+c` for Linux systems using Mac keyboards via [keyd](https://github.com/NixOS/nixpkgs/issues/137698#issuecomment-2404192700); with the default `ctrl+shift+c` keybinding for copy this is basically impossible (because that binding only applies to terminals).


deftdawg:
The sibling PR to this was [merged into Konsole](https://invent.kde.org/utilities/konsole/-/merge_requests/1035) today.  Happy New Year :clinking_glasses:

rockorager:
This PR has somehow broken toasts when copying with ctrl+shift+c. Toasts still display if I use the ctrl+ins bind. Reverting this commit does fix the issue, but I cannot for the life of me see how.

rockorager:
Another datapoint: adding `keybind = ctrl+insert=unbind` also fixes this.

Adding *any other* keybind which does the `copy_to_clipboard` action breaks the original bind. The underlying text *is* copied though

rockorager:
Oh I see the issue. We only set a single accelerator for `copy_to_clipboard`

deftdawg:
Is there a function preferable to set.put that would be better or is this a data structure thing? (Apologies if this a dumb question)

mitchellh:
> Oh I see the issue. We only set a single accelerator for `copy_to_clipboard`

The real bug is we should probably trigger the toast for any of the bindings. We should also probably put this binding before so that for menus we use the more traditional one

PR #2889, opened 2024-12-05
macos: Add "Use Option as Meta Key" menu item

nico:
Fixes #2520.

Makes state per surface, to match Terminal.app

Adds menu item to xib and wires it up.

Clicking the menu item sends an action into libghostty, which toggles state on the surface and sends a notification back. The menu item state then updates in response to that notification.

The menu item also updates when the current surface changes, either if the current split changes, or when tabs change, or when the quick terminal is opened or closed.

---

I went with a slight tweak to what you proposed in #2520 (when the option is set to left/right, it toggles that instead), I hope that's alright.

This is my first real PR in all of zig, swift, and ghostty, so it's well possible I did something silly.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -399,6 +399,14 @@ typedef enum {
   GHOSTTY_FULLSCREEN_NON_NATIVE_VISIBLE_MENU,
 } ghostty_action_fullscreen_e;
 
+// apprt.action.OptionAsAlt
+typedef enum {
+  GHOSTTY_OPTION_AS_ALT_OFF,
+  GHOSTTY_OPTION_AS_ALT_ON,
+  GHOSTTY_OPTION_AS_ALT_LEFT,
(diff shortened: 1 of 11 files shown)


iandyh:
Thanks for this PR. Looking forward it to be merged. 
This feature is what I'm looking for once I downloaded Ghostty! (Emacs users will love it)

mitchellh (changes requested):
Hi! Thanks for this PR and apologies for the delayed review. This looks quite good and I'm okay with the logical changes.

Unfortunately, we've had many changes since this PR. Can you rebase and fix it up? The major change is our option as alt default detection. I think to work around this we probably will have to think through a slightly different strategy.

Are you interested in that? 

Thank you!

mitchellh:
No response on feedback. Happy to take a look again in the future.

PR #2895, opened 2024-12-05
macos: Make "Settings…" menu item open config file in Application Sup…

nico:
…port

...unless ~/.config/ghostty/config already exists, then that is opened. (Or whatever $XDG_CONFIG_HOME points to.)

If both files exists, ghostty reads first the one in ~/.config/ghostty/config and then the one in Application Support, and merges the settings. In that case, the menu item opens the file at ~/.config.

Fixes #2890.


jparise (approved):

martinengelke:
Maybe the user should get asked if a ~/.config/ghostty/config shall be created the first time he is opening the 'settings' rather than expecting him to create it by himself.

PR #2930, opened 2024-12-11
macOS: "option-as-alt" defaults to "true" for US keyboard layouts

mitchellh:
A common issue for US-centric users of a terminal is that the "option" key on macOS is not treated as the "alt" key in the terminal.

## Background

macOS does not have an "alt" key, but instead has an "option" key. The "option" key is used for a variety of purposes, but the troublesome behavior for some (and expected/desired behavior for others) is that it is used to input special characters.

For example, on a US standard layout, `option-b` inputs `∫`. This is not a typically desired character when using a terminal and most users will instead expect that `option-b` maps to `alt-b` for keybinding purposes with whatever shell, TUI, editor, etc. they're using.

On non-US layouts, the "option" key is a critical modifier key for inputting certain characters in the same way "shift" is a critical modifier key for inputting certain characters on US layouts.

We previously tried to change the default for `macos-option-as-alt` to `left` (so that the left option key behaves as alt) because I had the wrong assumption that international users always used the right option key with terminals or were used to this. But very quickly beta users with different layouts (such as German, I believe) noted that this is not the case and broke their idiomatic input behavior. This behavior was therefore reverted.

## Solution

This confusing behavior happened frequently enough that I decided to implement the more complex behavior in this commit. The new behavior is that when a US layout is active, `macos-option-as-alt` defaults to true if it is unset. When a non-US layout is active, `macos-option-as-alt` defaults to false if it is unset. This happens live as users change their keyboard layout.

**An important goal of Ghostty is to have zero-config defaults** that satisfy the majority of users. Fiddling with configurations is -- for most -- an annoying task and software that works well enough out of the box is delightful. Based on surveying beta users, I believe this commit will result in less configuration for the majority of users.

## Other Terminals

This behavior is unique amongst terminals as far as I know. Terminal.app, Kitty, iTerm2, Alacritty (I stopped checking there) all default to the default macOS behavior (option is option and special characters are inputted).

All of the aforementioned terminals have a setting to change this behavior, identical to Ghostty (or, Ghostty identical to them perhaps since they all predate Ghostty).

I couldn't find any history where users requested the behavior of defaulting this to something else for US based keyboards. That's interesting since this has come up so frequently during the Ghostty beta!

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -245,7 +245,7 @@ const DerivedConfig = struct {
     mouse_scroll_multiplier: f64,
     mouse_shift_capture: configpkg.MouseShiftCapture,
     macos_non_native_fullscreen: configpkg.NonNativeFullscreen,
-    macos_option_as_alt: configpkg.OptionAsAlt,
+    macos_option_as_alt: ?configpkg.OptionAsAlt,
     vt_kam_allowed: bool,
     window_padding_top: u32,
     window_padding_bottom: u32,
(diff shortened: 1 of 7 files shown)


jwr:
I don't think there is a "right" solution. I just wanted to provide a data point: I use a Mac with the PL (Polish) keyboard layout. I installed ghostty and was surprised that the ctrl+alt+f shortcut which normally brought up fzf in fish doesn't work anymore. It produced "ń" instead, which is what I normally get in every app if I press alt-n.

It took some searching and trying things out to encounter the `macos-option-as-alt` setting. After I set it to `true`, I got the behavior I expected. I am also unable to type Polish characters in a shell in ghostty (so, alt-n does not produce "ń"), but I'd rather have that breakage than having ctrl+alt+f broken.

Of course, ideally I'd like the best of both worlds. I tried configuring bindings to override (`keybind="alt+n=text:ń"`), but that didn't work as expected.

tomaszwostal:
@jwr I have exactly the same problem. Did you manage to solve it?

jwr:
With the 1.0.1 update the workaround with `keybind="alt+n=text:ń"` started working, so I guess I'll create those bindings for all PL characters with diacritics and that will solve my problem!

kemko:
> The “ABC” keyboard is not considered US standard.

Do you want to ends with good zeroconfig defaults or good standards compatibility? Because of using ABC layout I must change this setting instead of relying on defaults.

cyxr001:
I use the macbook and  Manico app to switch app . And the shortcut is `option + 1`, `option + 2`...
When I set macos-option-as-alt: false , it will print some chars. 
When I set macos-option-as-alt: true, it won't do anything.
Is there any workaround ?

PR #2985, opened 2024-12-16
Handle fonts with missing or invalid OS/2 tables

qwerasd205:
It turns out this happens more than I thought. Most notably, apple's fallback font for a bunch of symbols, `Apple Gothic`, has no OS/2 table 🙃. I've gone ahead and fixed it so that both CoreText and FreeType handle this gracefully.

Speaking of graceful handling, this problem came to attention due to our very *un*-graceful handling of when a font fails to load due to a metrics calculation error. It results in a bunch of blanked out text, very bad, we should fix that.

*At some point I'll try to collect a group of test fonts with all sorts of weirdnesses to exercise the weird edge cases in our handling.*

### DIFF
--- a/pkg/macos/text/font.zig
+++ b/pkg/macos/text/font.zig
@@ -217,6 +217,10 @@ pub const FontOrientation = enum(c_uint) {
 
 pub const FontTableTag = enum(u32) {
     svg = c.kCTFontTableSVG,
+    os2 = c.kCTFontTableOS2,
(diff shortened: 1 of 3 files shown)


mitchellh (commented):
Looks good. I'm going to make some stylistic changes but otherwise looks great. 

It definitely would be nice if we could find some permissively licensed fonts (or make our own custom ones) to test this stuff since we have all the machinery already.

PR #3011, opened 2024-12-19
bash: avoid accidents on macos when /bin/bash is being used

anund:
If completions are sourced under bash missing `mapfile` (macos default /bin/bash) users get an error rather than a completion suggestion. Move the error to point of source rather than point of use. It is possible to write these completions to support bash without mapfile if someone is so inclined.


jparise (commented):
Maybe we should just use an explicit bash version check? That would also be easier to document.

PR #2963, opened 2024-12-14
bash: add bash completion generation

anund:
Bash completions on par with fish and zsh completions. There's a lot of room to add additional custom completions in all three languages. (see `key) return;;` for cases)

I've noticed a few mistakes with the other completions I'll raise as a separate PR.

<details>
  <summary>Generated ghostty.bash - updated</summary>

```bash

# -o nospace requires we add back a space when a completion is finished
# and not part of a --key= completion
addSpaces() {
  for idx in "${!COMPREPLY[@]}"; do
    [ -n "${COMPREPLY[idx]}" ] && COMPREPLY[idx]="${COMPREPLY[idx]} ";
  done
}

_fonts() {
  local IFS=$'\n'
  mapfile -t COMPREPLY < <( compgen -P '"' -S '"' -W "$($ghostty +list-fonts | grep '^[A-Z]' )" -- "$cur")
}

_themes() {
  local IFS=$'\n'
  mapfile -t COMPREPLY < <( compgen -P '"' -S '"' -W "$($ghostty +list-themes | sed -E 's/^(.*) \(.*$/\1/')" -- "$cur")
}

_files() {
  mapfile -t COMPREPLY < <( compgen -o filenames -f -- "$cur" )
  for i in "${!COMPREPLY[@]}"; do
    if [[ -d "${COMPREPLY[i]}" ]]; then
      COMPREPLY[i]="${COMPREPLY[i]}/";
    fi
    if [[ -f "${COMPREPLY[i]}" ]]; then
      COMPREPLY[i]="${COMPREPLY[i]} ";
    fi
  done
}

_dirs() {
  mapfile -t COMPREPLY < <( compgen -o dirnames -d -- "$cur" )
  for i in "${!COMPREPLY[@]}"; do
    if [[ -d "${COMPREPLY[i]}" ]]; then
      COMPREPLY[i]="${COMPREPLY[i]}/";
    fi
  done
  if [[ "${#COMPREPLY[@]}" == 0 && -d "$cur" ]]; then
    COMPREPLY=( "$cur " )
  fi
}

config="--help"
config+=" --version"
config+=" --font-family="
config+=" --font-family-bold="
config+=" --font-family-italic="
config+=" --font-family-bold-italic="
config+=" --font-style="
config+=" --font-style-bold="
config+=" --font-style-italic="
config+=" --font-style-bold-italic="
config+=" --font-synthetic-style="
config+=" --font-feature="
config+=" --font-size="
config+=" --font-variation="
[...]

PR #3044, opened 2024-12-21
Bash complex completions

anund:
Given https://github.com/ghostty-org/ghostty/discussions/2994 revert https://github.com/ghostty-org/ghostty/pull/2963/commits/f1728f594a681941b85a8d2fa8a136d625d9b633. Add documentation to help point people towards the discussion and needed changes to handle `COMP_WORDBREAKS` not including expected characters. 

We know macOS and nixos defaults to including `=` so explicit handling to change behaviour should only be required if it turns out some systematically used tool brings in a completion function that changes `COMP_WORDBREAKS` away from the default. (something they should not be doing)

If necessary I can create a issue to track handling unexpected word breaks usage. This change improves the current completions.

### DIFF
--- a/src/build/bash_completions.zig
+++ b/src/build/bash_completions.zig
@@ -37,12 +37,47 @@ fn writeBashCompletions(writer: anytype) !void {
     const pad4 = pad3 ++ pad1;
 
     try writer.writeAll(
+        \\
         \\# -o nospace requires we add back a space when a completion is finished
         \\# and not part of a --key= completion
         \\addSpaces() {
         \\  for idx in "${!COMPREPLY[@]}"; do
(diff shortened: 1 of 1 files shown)


jparise (approved):
This works well. Nice work!

I think there are further opportunities for improvements that we can work on as follow-ups, but let's roll with this.

jparise (commented):
Separately, we should revise this script to be friendlier to the global namespace. It leaves these functions (and `addSpaces`) available outside of the completion context.

We should either nest them within a single `_ghostty` function or prefix them with `__ghostty_`.

This also goes for the top-level variables such as `$topLevel`.

PR #3046, opened 2024-12-21
core: don't try opening config files that are anything but files

jcollie:
In a [Discord thread](https://discord.com/channels/1005603569187160125/1319727495473397791) someone was having problems running Ghostty because they had accidentally created a directory called `~/.config/ghostty/config`. Instead of erroring out Ghostty would hang trying to "read" the directory. Crashes can also happen if the argument to `--config-file` on the CLI or a recursively loaded config file.

This patch prevents those hangs or crashes by refusing to read anything but a plain file (symbolic links to plain files continue to work as well).

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2546,8 +2546,20 @@ pub fn loadIter(
 pub fn loadFile(self: *Config, alloc: Allocator, path: []const u8) !void {
     assert(std.fs.path.isAbsolute(path));
 
-    var file = try std.fs.cwd().openFile(path, .{});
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
There's some unrelated changes in this PR that are preventing me from snap-merging this because I'm not sure I agree with the approach/style of them (around the error messages on open, changes to optional handling, etc.).

Please keep the PR focused to one small task and PRs get merged much more quickly. Can you modify this one to _only_ do the stat change for dir detection?

PR #3060, opened 2024-12-22
bash: move functions and variables out of global scope

anund:
Leave only `_ghostty` as a bash global.


jparise (approved):
Thanks!

PR #3062, opened 2024-12-22
font: add `cursor-height` metric, and `adjust-` config for it.

qwerasd205:
Subsumes #2580 (which has multiple conflicts with main due to recent changes to metrics); I figured it'd be easier to just implement it this way.

#2580 claimed to solve #2487 but I don't think it really does- ideally we can think of a good way to configure each individual cursor type, but I don't wanna just do something ad hoc and add a bunch of config keys blindly so I limited the scope of this.


jesseleite:
Yesss, thank you so much for this @qwerasd205 @mitchellh ❤️

PR #3099, opened 2024-12-24
terminal: implement support for ConEmu OSC 9 progress report

Jan200101:
ConTerm and iTerm2 have both taken OSC 9 to implement different things: iTerm2 uses it to implement desktop notifications
ConTerm uses it to implement lots of different instructions

Terminals such as Kitty or GhosTTY have implemented OSC 9 like iTerm2 did meanwhile other projects such as Windows Terminal, systemd and flatpak have decided to use OSC 9;4 to report the progress state

https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
https://iterm2.com/documentation-escape-codes.html

For this PR I've implemented support for the OSC 9;4 command by ConEmu with a fallback to desktop notification if anything doesn't match what is expected

### DIFF
--- a/src/terminal/osc.zig
+++ b/src/terminal/osc.zig
@@ -158,6 +158,12 @@ pub const Command = union(enum) {
     /// End a hyperlink (OSC 8)
     hyperlink_end: void,
 
+    /// Set progress state (OSC 9;4)
+    progress: struct {
+        state: ProgressState,
+        progress: ?u8 = null,
+    },
+
     pub const ColorKind = union(enum) {
         palette: u8,
         foreground,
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
I haven’t fully reviewed the implementation but no matter what this file requires unit tests!

mitchellh (commented):
Okay I reviewed a bit more thoroughly. I think we should move to parse the ConEmu OSC 9 sequences (we can start with just `9;4`) separately and just note that as being impossible to send via OSC 9 desktop notifications. 

They should emit a new `Command`. And we should add unit tests for all that.

mitchellh (changes requested):
This PR is going way beyond scope, at least from what I originally thought.

I want to get ConEmu progress report _ignoring_ in for 1.0, but I don't want any of the apprt handling in yet. I want to discuss that further to figure out what we're going to accept or not. There has been no associated discussion as noted in CONTRIBUTING.md.

Issue #2384, opened 2024-10-04
macOS: Customize initial size for quick terminal


lucymhdavies:
In macOS, it looks like we're setting the default sizes here:
https://github.com/ghostty-org/ghostty/blob/main/macos/Sources/Features/QuickTerminal/QuickTerminalPosition.swift#L10-L29

For example, when showing at the top of the screen, we're setting the height to be 1/4 the height of the screen.
(specifically, this is whichever screen quick terminal first loads on. there's some edge cases around quick-terminal-screen=mouse if you have monitors of varying sizes, which I'm not going to go into right now)

I couldn't figure out from https://github.com/ghostty-org/ghostty/pull/2320/files how this is implemented on other systems. Am I missing something, or is this a macOS-only feature for now?

There are two options we could go for here:
* Allow the user to specify a multiplier in their config, which would default to 0.25
* Allow the user to specify a fixed height

(plus the option of both)

In my case, I would want my multiplier to be 1. i.e. the quick terminal takes up the entire screen. For my specific example, I see no value in specifying a fixed height. My instinct is that most people who use quick terminal have a preference for it to fill "a bit of the screen", "a lot of the screen" or "all of the screen". I imagine pixel-precision sizing of the quick terminal would be a niche desire.
(I could of course be wildly wrong these assumptions)

So were I to implement this myself (and I'm thinking I probably will), my proposal is:
* Add a new config item, e.g. `quick-terminal-size-multiplier`, with default value of 0.25
* Update where we set the window size, for example `height: screen.frame.height * multiplier`

There are a few edge cases I'd have to consider (e.g. making sure the multiplier is > 0, and <= 1), but otherwise this feels like it would be a fairly simple thing to implement.

If you're happy with the proposed implementation, I might get a PR raised for it over the weekend.

sohsatoh:
This feature has been the biggest hurdle for me in fully replacing iTerm2 with Ghostty, so I’m currently working on implementing it.

It’s just a proof of concept for now, but the changes aren’t too significant.

If everything looks good, I’m going to submit a pull request in a few days (hopefully).

https://github.com/user-attachments/assets/1510f20d-5b53-4066-90d5-bd1aa8c3130e

donaldguy:
I see the PR is in flight, so maybe preaching to the choir but

I would strongly appreciate some ability to dial in quick-terminal width when in top b/c

[...]

PR #3100, opened 2024-12-24
feat: add support for middle position quick terminal

dmehala:
# Description

This commit introduces the ability to launch the quick terminal in the middle position.

Screen Recording 2024-12-24 at 14 59 46

## Note to reviewer
- The quick terminal is currently centered. Should we consider adding a top offset to better align with the screenshot in [the issue](#2494 )? Should it be configurable?
- On large monitors, half the visible frame might be excessively large. To mitigate, I am planning to implement #2384 but we should probably agree on a good default maximum width/height in middle position.
- I also figured out, reloading the configuration does not update the quick terminal configuration. That is also an issue I will try to fix.

Resolves #2494

I agree to relicense my commit to MIT.

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalPosition.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalPosition.swift
@@ -5,6 +5,7 @@ enum QuickTerminalPosition : String {
     case bottom
     case left
     case right
+    case middle
 
     /// Set the loaded state for a window.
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Thanks this looks great. I requested a couple small changes, but I think for an initial go this is fine. Your size issues are noted, I think we can continue to improve and iterate on those post-merge.

PR #3102, opened 2024-12-24
Fix UB in style hashing by using autoHash, but keep XxHash3

mitchellh:
Back out "perf(styles): greatly improve style.hash performance" This backs out commit 3bfe4cd25ca7a5ae4d4084818b86ada9236b3bb5, but keeps the hash algorithm as XxHash3 which showed improvements in performance.


qwerasd205 (approved):
LGTM

PR #3103, opened 2024-12-24
optimize `Style` hashing to be single-shot

Rexicon226:
The idea is to serialize `Style` into a packed format as quickly as possible. 
Following an SOA pattern here yields the best result.

Using the Doom-Fire benchmark, 
New: 780FPS
Old (f46a6769): 627FPS

Memory usage isn't affected as the layout changes cancel each other out.

### DIFF
--- a/src/terminal/color.zig
+++ b/src/terminal/color.zig
@@ -95,7 +95,7 @@ pub const Name = enum(u8) {
 };
 
 /// RGB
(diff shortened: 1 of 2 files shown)


qwerasd205 (approved):
My only feedback was about adding comments to clarify why this is done the way it is, so now that they've been added this seems good to me!

PR #3110, opened 2024-12-25
Fuzzer fixes

qwerasd205:
I fuzzed things a little bit with integrity tests enabled and found 2 bugs which I fixed (I think one of them is responsible for the crash reported in #2497 after it was closed) and 1 other potential bug in a similar vein.

I also found a violation of the LTR shaping assertion in the Metal renderer and a style ref mismatch (off by 1) integrity violation in `Terminal.index` -> `PageList.eraseRowBounded` -> `Page.clonePartialRowFrom`, neither of which I've had time to fix.

I didn't have time to make tests for the fixed issues, but they shouldn't be *too* hard to make? Hopefully.

While fixing a test failure I found a bug, revealed by my `cursorReload` fix, where during reflow tracked pins could be lost. I fixed that, but there's still some test failures remaining, specifically for `Screen.cursorScrollAbove` - because my fix to `cursorScrollAboveRotate` changed the dirtying behavior (`cursorChangePin` unconditionally dirties the source and destination row for the cursor, because cursor position can change shaping). I took a look and it seems like what's being tested for that behavior is incorrect.

I don't have enough time to investigate more thoroughly tonight, I might have a little time tomorrow, but I'm opening the PR in case you want to take a look.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -244,6 +244,7 @@ pub fn init(
 
     // We always track our viewport pin to ensure this is never an allocation
     const viewport_pin = try pool.pins.create();
+    viewport_pin.* = .{ .node = page_list.first.? };
     var tracked_pins: PinSet = .{};
     errdefer tracked_pins.deinit(pool.alloc);
     try tracked_pins.putNoClobber(pool.alloc, viewport_pin, {});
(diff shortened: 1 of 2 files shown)


mitchellh:
@qwerasd205 Do you want me to keep this open? Or do you want me to abandon this for now (until one of us can either take a look or we address things more fundamentally)

qwerasd205:
This can probably be closed for now yeah

PR #3111, opened 2024-12-25
Improve URL Regex Handling and Tests in Config Module & Improve locale handling and add robust fallbacks

plyght:
This PR updates the URL regex to handle more edge cases, such as nested parentheses, brackets, and invalid trailing characters. 
It also expands test coverage for multiple schemes and complex URLs while ensuring compatibility with stable Zig versions by replacing `@concat` and `@intCast`. 
A documentation comment error in the test block was fixed.

This PR refactors ensureLocale to improve readability, error handling, and robustness. It enhances macOS locale handling by extracting system preferences via Cocoa when LANG is unset. Fallbacks are streamlined, with en_US.UTF-8 as the default when locale determination fails. Improved logging ensures better debugging and maintainability.


mitchellh (changes requested):
There are good changes in here and changes I don't agree with. For example, linking to Zig docs all over isn't necessary since it's a Zig codebase there is some general assumption you understand Zig when reading it.

I also don't like long variable names if the block isn't nested or larger than a dozen or so lines (i.e. replacing `v` with something more descriptive) because when the context is small enough its easy enough to see in context what that means. I find longer variable names make my code scanning ability more noisy. As I said though, for larger blocks or nested blocks I definitely do it.

I'm not a big fan of what I call "flower boxes" either (comments starting and ending with dashes). I use them in very specific scenarios but try to avoid them if possible. 

I'm happy to see some of this go in but please focus on the actual change. Any style suggestions please open a discussion first and then we can agree or not if we want them.

Thanks!

src/config/url.zig
```diff
@@ -1,50 +1,64 @@
...
+    "|git://" ++
+    "|ssh://" ++
+    "|tel:" ++
+    "|magnet:" ++
+    "|ipfs://" ++
+    "|ipns://" ++
+    "|gemini://" ++
+    "|gopher://" ++
+    "|news:";
+
```
Bammy1542, line 42: smpt

mitchellh:
I apologize but theres been changes to the regex that cause conflicts. I still didn't see enough clarity in value for this PR so it was languishing. If you want to reopen it please keep the scope small and explain what you're trying to achieve. I don't really want refactors unless previously discussed. Thanks!

PR #3121, opened 2024-12-25
fix: correct handling of CTC and DECST8C

aymanbagabas:
Cursor Tab Control (CTC) `CSI <n> W` has a default value of 0, which is the same as setting a tab stop at the current cursor position. Set Tab at every 8th Column (DECST8C) `CSI ? 5 W` is a DEC private sequence that resets the tab stops to every 8th column.

Reference: https://vt100.net/docs/vt510-rm/DECST8C.html
Reference: https://wezfurlong.org/ecma48/07-control.html?highlight=ctc#7210-ctc---cursor-tabulation-control
Reference: https://gitlab.gnome.org/GNOME/vte/-/blob/master/src/parser-seq.py#L596-L601
Fixes: https://github.com/ghostty-org/ghostty/issues/3120

### DIFF
--- a/src/terminal/stream.zig
+++ b/src/terminal/stream.zig
@@ -601,30 +601,37 @@ pub fn Stream(comptime Handler: type) type {
                 // Cursor Tabulation Control
                 'W' => {
                     switch (input.params.len) {
-                        0 => if (input.intermediates.len == 1 and input.intermediates[0] == '?') {
-                            if (@hasDecl(T, "tabReset"))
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Thanks. Can you add unit test coverage for this too? Should pretty simple (see bottom of the file). A lot of the original work predates any test coverage but as we iterate and make changes it'd be good to add so we can refactor this all in the future.

PR #3130, opened 2024-12-26
Handle execveZ failures in `Command.zig`

anund:
Bit of a rabbit hole came up while trying to package ghostty for nixos. `zig build test` would just hang when run as part of `checkPhase` in a nix build. (rather than the `nix develop ...` command run in CI).

Here's what I understand so far:

`/usr/bin/env` does not exist in a nix build sandbox. This only works as a test binary when running in `nix develop` as it shares its environment with the user where this compatibility crutch exists.
When executing `Command.zig` tests that reference `/usr/bin/env` the eventual call to fork+execevZ will duplicate the running test process as execevZ _returns_ rather than dissapearing into the new exec'd process. Duplicating a test process via fork does unexepected things. `zig build test` will hang for <reasons?>. A test binary created via `-Demit-test-exe` will run two copies of the test suite producing two different failure outputs for the same test. (or ~4 copies of the test framework, one extra for each test that fails this way)

`/bin/sh` can be used and an alternative test target. It isn't amazing as it's relying on stdenv creating `/bin/sh` and it just existing on user systems. Other alternatives I can think of would require build flags or some sort of contract with packaging around what binary will exist for the `Command.zig` tests.

This is ~probably right, either way leave it until after release. I agree to re-license my commits as MIT.

### DIFF
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -588,8 +588,8 @@ test "createNullDelimitedEnvMap" {
 test "Command: pre exec" {
     if (builtin.os.tag == .windows) return error.SkipZigTest;
     var cmd: Command = .{
-        .path = "/usr/bin/env",
-        .args = &.{ "/usr/bin/env", "-v" },
+        .path = "/bin/sh",
+        .args = &.{ "/bin/sh", "--help" },
         .pre_exec = (struct {
             fn do(_: *Command) void {
                 // This runs in the child, so we can exit and it won't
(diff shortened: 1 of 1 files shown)


anund:
I can't see how this fails in CI. 
- https://github.com/ghostty-org/ghostty/blob/7fd20cadb0364cdfc75a17ccb6b201f7f817e180/src/Command.zig#L705 sets cwd to `/bin/`
- https://github.com/ghostty-org/ghostty/blob/7fd20cadb0364cdfc75a17ccb6b201f7f817e180/src/Command.zig#L703 prints pwd
- https://github.com/ghostty-org/ghostty/blob/7fd20cadb0364cdfc75a17ccb6b201f7f817e180/src/Command.zig#L722 some how sees `/usr/bin` ? 

Ci appears to be using the right commits, no stale grabs of commits. But the fork is now busted post public.

anund:
Github fork isn't updating and no longer sees this repo as a target. Reforking, reopening pr

PR #3132, opened 2024-12-26
fix: handle intermediate bytes in CSI and ESC sequences

### LINKED ISSUE #3122: Proper escape sequence intermediate handling
Many escape sequences ignore intermediate bytes. For example, Erase Characters (ECH) `CSI <n> X` doesn't have any intermediate bytes, and something like `CSI ? X` should be ignored. Instead, Ghostty treat `CSI ? X` as if it was `CSI X` and erases a character at the current cursor position.

```sh
printf 'ABC\b\x1b[?X'
```

Current output:
```
AB_
```

The cursor is at the 3rd column

Expected output (invalid sequence is ignored):
```
ABC
```

The cursor is at the 4th column

aymanbagabas:
This adds missing handling for CSI and ESC commands.

Fixes: https://github.com/ghostty-org/ghostty/issues/3122


apatterndarkly:
Real nice! 

As someone who exits insert mode with `ctrl+[` on several apps in the terminal, I'd been spewing `;5u` all over the place for the past few hours. Fortunately that little clue led me to a work around, but yeah I'm super happy to see this fix on its way, so thank you @aymanbagabas.

For the slim chance anyone lands here with my same issue in the meantime, my workaround was to add the following to my config:

```
keybind = ctrl+left_bracket=text:\x1b
```
✌️

aymanbagabas:
I accidentally closed the PR after force pushing the updated changes. Apparently, GitHub doesn't carry forked repository visibility settings :/

Could someone reopen the PR? Otherwise, I can create a new one

CC/ @mitchellh

PR #3173, opened 2024-12-26
Update help comment for backslash

lcrownover:
Such a little commit, but I banged my head on it for a few minutes!


mitchellh:
Thanks!

PR #3176, opened 2024-12-26
Redo: Handle execveZ failures in `Command.zig` tests

anund:
https://github.com/ghostty-org/ghostty/pull/3130 do over. Github seems to have lost track of the fork status for the previous work. (perhaps permissions?). Rebased the PR to main as I can't see why it could possibly fail.

<details>
  <summary>Previous description</summary>
Bit of a rabbit hole came up while trying to package ghostty for nixos. zig build test would just hang when run as part of checkPhase in a nix build. (rather than the nix develop ... command run in CI).

Here's what I understand so far:

/usr/bin/env does not exist in a nix build sandbox. This only works as a test binary when running in nix develop as it shares its environment with the user where this compatibility crutch exists.
When executing Command.zig tests that reference /usr/bin/env the eventual call to fork+execevZ will duplicate the running test process as execevZ returns rather than dissapearing into the new exec'd process. Duplicating a test process via fork does unexepected things. zig build test will hang for <reasons?>. A test binary created via -Demit-test-exe will run two copies of the test suite producing two different failure outputs for the same test. (or ~4 copies of the test framework, one extra for each test that fails this way)

/bin/sh can be used and an alternative test target. It isn't amazing as it's relying on stdenv creating /bin/sh and it just existing on user systems. Other alternatives I can think of would require build flags or some sort of contract with packaging around what binary will exist for the Command.zig tests.
</details>

### DIFF
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -182,9 +182,10 @@ fn startPosix(self: *Command, arena: Allocator) !void {
     _ = posix.execveZ(pathZ, argsZ, envp) catch null;
 
     // If we are executing this code, the exec failed. In that scenario,
-    // we return a very specific error that can be detected to determine
-    // we're in the child.
-    return error.ExecFailedInChild;
+    // terminate so we don't duplicate the original process
+    // note: returning to test code from this point would run 2 copies of the test suite
+    std.debug.print("failed to execveZ as child process, terminating", .{});
+    std.process.exit(1);
 }
 
 fn startWindows(self: *Command, arena: Allocator) !void {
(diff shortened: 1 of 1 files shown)


anund:
I can't see how this fails in CI. 
- https://github.com/ghostty-org/ghostty/blob/7fd20cadb0364cdfc75a17ccb6b201f7f817e180/src/Command.zig#L705 sets cwd to `/bin/`
- https://github.com/ghostty-org/ghostty/blob/7fd20cadb0364cdfc75a17ccb6b201f7f817e180/src/Command.zig#L703 prints pwd
- https://github.com/ghostty-org/ghostty/blob/7fd20cadb0364cdfc75a17ccb6b201f7f817e180/src/Command.zig#L722 some how sees `/usr/bin` ? 

Ci appears to be using the right commits, no stale grabs of commits. But the fork is now busted post public.

But I can trigger CI again with the new, public?, fork.

auscompgeek:
`/bin` is a symlink to `/usr/bin` on many Linux distros these days, including Debian/Ubuntu by default.

anund:
It's unclear to me what a safe universal replacement for `/usr/bin/env` would be in the tests at the moment. Simplified the pr to just address cleaning up the child process in `Command.zig` for this PR.

anund:
Fixed now for `nix build`. `/tmp` being `/private/tmp` gives fixes the result across ~ci/nix.

anund:
@jcollie car to have a second look as the previous PR needed to be closed?

PR #3177, opened 2024-12-26
Fix notification handling from publishing while View is updating

trevor-e:
I was originally looking into this issue: https://github.com/ghostty-org/ghostty/issues/3109

When running the logic and triggering a config reload, Xcode warns us about updating publishable properties from within the notification callback functions:
Screenshot 2024-12-26 at 5 46 19 PM

I believe this is because `SurfaceView` is being used as both a bridged NSView (inside `Surface`) and also an `ObservableObject`, so it's possible for the notification callback to happen while a SwiftUI render loop is occurring. The notification delivery happens on whatever thread posted the message. The better solution long-term is likely to separate the `ObservableObject` logic into its own class to avoid mixing with the View logic.

The solution here is to simply move the publishable mutation out of the current loop via `DispatchQueue.main.async`. I confirmed the warning goes away with this, and I didn't notice any odd behavior while reloading config changes.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -423,7 +423,9 @@ extension Ghostty {
             // We also just trigger a backing property change. Just in case the screen has
             // a different scaling factor, this ensures that we update our content scale.
             // Issue: https://github.com/ghostty-org/ghostty/issues/2731
-            viewDidChangeBackingProperties()
+            DispatchQueue.main.async { [weak self] in
+                self?.viewDidChangeBackingProperties()
+            }
         }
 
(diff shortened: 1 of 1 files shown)


mitchellh:
> I believe this is because SurfaceView is being used as both a bridged NSView (inside Surface) and also an ObservableObject, so it's possible for the notification callback to happen while a SwiftUI render loop is occurring. The notification delivery happens on whatever thread posted the message. The better solution long-term is likely to separate the ObservableObject logic into its own class to avoid mixing with the View logic.

Ah hah! Thank you so much for explaining this to me. I knew these warnings existed, and I knew how to fix them (DispatchQueue), but I didn't know exactly why. This makes perfect sense.

Will run CI then merge.

PR #3217, opened 2024-12-27
Re-add nix-compat flake input

### LINKED ISSUE #3216: `shell.nix` is currently non-functional
Somewhere along the way, the flake-compat entry was dropped from `flake.lock`, so `shell.nix` no longer works:

```
$ nix-shell
error:
       … while evaluating '(import (let flake-compat = (((builtins).fromJSON ((builtins).readFile /home/chip/src/ghostty/flake.lock))).nodes.flake-compat; in (fetchTarball { sha256 = (flake-compat).locked.narHash; url = ("https://github.com/edolstra/flake-compat/archive/" + (flake-compat).locked.rev + ".tar.gz"); })) { src = /home/chip/src/ghostty; })' to select 'shellNix' on it
         at /home/chip/src/ghostty/shell.nix:1:1:
            1| (import
             | ^
            2|   (

       … while calling the 'import' builtin
         at /home/chip/src/ghostty/shell.nix:1:2:
            1| (import
             |  ^
            2|   (

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'flake-compat' missing
       at /home/chip/src/ghostty/shell.nix:4:22:
            3|     let
            4|       flake-compat = (builtins.fromJSON (builtins.readFile ./flake.lock)).nodes.flake-compat;
             |                      ^
            5|     in
```

It's easy to re-add, but depending on preference, I'd suggest there are a few reasonable ways to do so:
 * Explicitly add it as an input in `flake.nix`. This is how it was originally done and seems to be [accidentally?] removed by #1025
 * Statically insert a `fetchurl` or something into `shell.nix`. This has perhaps a slight benefit of being less likely to be accidentally removed again, but it also pushes it out of view of the `nix flake ...` commands.

chipb:
Its entry in flake.lock is required for shell.nix to operate as it's been written. Hash values are restored to where they last existed.

Fixes #3216.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,21 @@
 {
   "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1696426674,
+        "narHash": "sha256-kvjfFW7WAETZlt09AgDn1MrtKzP7t90Vf7vypd3OL1U=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "0f9255e01c2351cc7d116c072cb317785dd33b33",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
     "flake-utils": {
       "inputs": {
         "systems": "systems"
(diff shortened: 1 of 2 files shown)


mitchellh:
Thank you!

PR #3259, opened 2024-12-27
Use doc comments for focus-follows-mouse

anmolw:
Fixes the description of `focus-follows-mouse` not showing up in the autogenerated docs and the website


mitchellh:
Thank you!

PR #3260, opened 2024-12-27
ci: lint zig code with ZLint

DonIsaac:
## What This PR Does

Adds a CI workflow to lint Zig source code with [zlint](https://github.com/DonIsaac/zlint).

ZLint checks for:
- Compiler errors that are never emitted because they occur in dead code
- Potentially unsafe logic
- More idiomatic ways to do the same thing (e.g. `foo() catch |e| return e` -> `try foo()`
- Dead code

This workflow can be configured to only check files changed in a pull request. I have turned this off for demonstration purposes. Please let me know if you would like it re-enabled.

Here's some sample output from `zlint --quiet`
image

### DIFF
--- a/.github/workflows/zlint.yml
+++ b/.github/workflows/zlint.yml
@@ -1,0 +1,23 @@
+name: ZLint
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  zlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Lint with ZLint
+        uses: DonIsaac/zlint-action@v0
+        with:
+          # When true (the default value), only files changed in a
+          # PR get linted. Has no effect on pushes.
+          diff-only: false


qwerasd205:
Aren't all 3 of those in your example screenshot false positives?

rockorager:
Yeah, you can use `try` in comptime code...if it fails you get a compile error.

Uncomment the second print line to bring the failing code into the compilation unit

https://zigbin.io/9af430

DonIsaac:
> Yeah, you can use `try` in comptime code...if it fails you get a compile error.
> 
> Uncomment the second print line to bring the failing code into the compilation unit
> 
> https://zigbin.io/9af430

I had no idea you could do that...

Thanks for the info, I'll add checks for that.

mitchellh:
Thanks for this. I'm not interested in adding this for now, but please feel free to continue developing ZLint against the Ghostty codebase and if you find issues I'm happy to address specific issues. Once there is some more stability I'm happy to take another look.

DonIsaac:
@rockorager thanks for the info, [this has been fixed](https://github.com/DonIsaac/zlint/pull/180).

PR #3273, opened 2024-12-27
macos: disable auto-updates for local (source) builds

### LINKED ISSUE #3179: macOS: ability to disable auto-update on source-built builds
OS: OSX 14.2 Sonoma
Commit Built From: [4b4d4062](https://github.com/ghostty-org/ghostty/commits/4b4d4062dfed7b37424c7210d1230242c709e990)

I built Ghostty from source using the commit above but on first start, it checks for version, seems to find 0.1 and says there's a new build.
Image
Is there a way to turn it off if I intend to build it from source?

jparise:
The auto-update prompt isn't useful for local (source) builds so disable both update checks and automatic downloads for the Debug and Release configurations.

Fixes #3179


macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -485,10 +485,16 @@ class AppDelegate: NSObject,
         }
 
         // Sync our auto-update settings
-        updaterController.updater.automaticallyChecksForUpdates =
-            config.autoUpdate == .check || config.autoUpdate == .download
-        updaterController.updater.automaticallyDownloadsUpdates =
-            config.autoUpdate == .download
+        // Local (source) builds always disable the updater
+        if let commit = Bundle.main.infoDictionary?["GhosttyCommit"] as? String, !commit.isEmpty {
```
jparise, line 489: We could also wrap this in a more generic "isLocalBuild" property under e.g. `Ghostty.Info` if we think it would be useful elsewhere or improve the clarity of this check.

mitchellh: I'm not sure this is right. Because tip builds should have a commit too, no? I'd prefer to use something a bit more robust. I'm not totally sure what that is yet (there are multiple options...).

The one I'm leaning towards the most is inserting a key into the plist file for releases only (`release-tip` `release-tag` jobs) and only enabling sparkle at all if its present.

jparise: > Because tip builds should have a commit too, no?

Yes, `GhosttyCommit` is set by all of the release workflows: pr, tag, tip. It is _not_ set as part of the local build process.

> The one I'm leaning towards the most is inserting a key into the plist file for releases only (release-tip release-tag jobs) and only enabling sparkle at all if its present.

I think that'll work. It's most similar to the third alternative I listed above. Do you have a preference for (ab)using [Sparkle's `SUEnableAutomaticChecks`](https://sparkle-project.org/documentation/customization/) for this or introducing our own `GhosttyAutoUpdate` key?

mitchellh: Oh my mistake I misread the boolean logic.

And yes! `SUEnableAutomaticChecks` is what I was thinking of! I think for debug and release local schemes we should just set this to NO and for other schemes we should leave it unset (to allow users to override it).

We can do this in the `Ghostty-Info.plist` in Xcode.

jparise: We unfortunately can't vary an Info.plist value on a per-scheme basis without taking one of the more elaborate approaches I listed above.

I think the compromise might be to default it to NO (which would apply to all build schemes) but toggle it back on in the release workflows where we want it.

mattwidmann: You can also set `INFOPLIST_PREPROCESS` to run the C preprocessor on `Info.plist`. Paired with `INFOPLIST_OTHER_PREPROCESSOR_FLAGS` conditionalized on the config, you could try something like this:

```
INFOPLIST_PREPROCESS = YES
INFOPLIST_OTHER_PREPROCESSOR_FLAGS[config=ReleaseLocal] = -DLOCAL=1
INFOPLIST_OTHER_PREPROCESSOR_FLAGS[config=Debug] = -DLOCAL=1
```

And then add this snippet to the `Info.plist`:

```
	<key>SUEnableAutomaticChecks</key>
#if LOCAL
	</false>
#else
	</true>
#endif
```

jparise: Thanks for the suggestion, @mattwidmann! I used a `INFOPLIST_PREPROCESSOR_DEFINITIONS`-based approach in my latest revision.

archcorsair:
Would it be difficult to an option for the config file?
```
auto-update = true/false
```
That way a user can still manually check for updates, but the auto-check for updates would be disabled.

mitchellh:
> Would it be difficult to add an option for the config file?

This already exists.

jparise:
I'm happy enough with this approach, but I think I'd still have a small preference for:

- Always disabling Sparkle by default in `Ghostty-Info.plist` (`SUEnableAutomaticChecks=NO`)
- Using `PlistBuddy` in the release workflows to "enable" Sparkle by deleting that plist entry.

Three reasons:

1. This Info.plist preprocessing still feels a little hacky, but it's at least using a standard mechanism.
2. We already modify the plist in those workflows to write `GhosttyCommit`.
3. It ties the "turn on Sparkle" logic to the release process instead of the build scheme, which may or may not be the better approach.

I don't feel very strongly but wanted to note my thoughts after sitting with it for a bit.

mitchellh:
This looks super well done thank you.

macos/Ghostty.xcodeproj/project.pbxproj
```diff
@@ -843,6 +844,8 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
+				INFOPLIST_PREPROCESS = YES;
+				INFOPLIST_PREPROCESSOR_DEFINITIONS = "SPARKLE_CHECKS_DISABLED=1";
```
mitchellh, line 848: Wait, actually, did I misread this and is this in Release?

mitchellh:
Temporarily reverted this since I think the definition is set in `Release`. I didn't fully verify but I don't want to risk it quite yet. Sorry @jparise. I still want to get this in but want to verify...

jparise:
> Temporarily reverted this since I think the definition is set in `Release`. I didn't fully verify but I don't want to risk it quite yet. Sorry @jparise. I still want to get this in but want to verify...

Yes, that was my mistake. I included Release instead of ReleaseLocal. I can fix that up later today unless you get to it first.

jparise:
> > Using PlistBuddy in the release workflows to "enable" Sparkle by deleting that plist entry.
> 
> Yeah, I think my biggest fear is we cut a release that disables automatic updates by accident. This would solve that. I'd be happy to take a PR to rework this to do it that way.

#3943 implements this safer approach. It's also simpler than what I had here because it doesn't involve plist preprocessing and is conceptually aligned with the release process, which is already updating the plist with Sparkle's public key.

Issue #3275, opened 2024-12-27
Certain paths in the docs contain spaces when the shouldn't


jorisguex:
As per the title, certain paths contain spaces when the shouldn't which make copy-and-pasting not work as expected.

Link: https://ghostty.org/docs/config/reference#theme
Screenshot: ![Image](https://github.com/user-attachments/assets/80e3f7ff-7387-42f4-a623-2136b15e7291)

mitchellh:
These are all auto-generated by a build script so should be easy to resolve!

mitchellh:
Fixes by #3276 thank you

PR #3276, opened 2024-12-27
Fix certain paths in the docs containing spaces when the shouldn't

jorisguex:
This addresses issue #3275


mitchellh:
I'd love to see a fix in the autogen scripts as well but this is a good stopgap. Thank you!

PR #3332, opened 2024-12-27
Fix zsh shell integration docs

lixiaoyan:
- There is no need to quote variables in zsh.
- The shell integration file is not executable; we should `source` it instead.

### DIFF
--- a/src/shell-integration/README.md
+++ b/src/shell-integration/README.md
@@ -73,6 +73,6 @@ sequence occurs.
 
 ```bash
 if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
-  "$GHOSTTY_RESOURCES_DIR"/shell-integration/zsh/ghostty-integration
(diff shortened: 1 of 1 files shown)


stschulte:
I was courious about your first claim, since the quoting should prevent issues when the variable value itself contains spaces. Looks like this is indeed differently handled by zsh vs bash.
Setup:

```
% mkdir "foo dir"
% echo "echo Hello" > "foo dir/integration"
% RESOURCE_DIR="foo dir"
```

In zsh the variable is expanded as if the value is put in quotes, e.g. `ls` will see one `foo dir` argument not `foo` and `dir`.
```
% ls $RESOURCE_DIR 
integration
% source $RESOURCE_DIR/integration
Hello
```

In bash the variable is expanded but if not put in quotes the value will be treated unquoted, so `ls` will see two parameters.
```
% ls $RESOURCE_DIR
ls: cannot access 'foo': No such file or directory
ls: cannot access 'dir': No such file or directory
% source $RESOURCE_DIR/integration
bash: foo: No such file or directory
```

Good to know.

PR #3353, opened 2024-12-27
fix: use `open -t` to open text editor

### LINKED ISSUE #3284: macOS: `open_config` action should use `open -t`
Excellent details in #3282

> minimally: changing open.zig to use the -t flag for this so you at least get whats configured for plain text files (e.g. associated for .txt) [also what I tried to do with defaults write after first searching "macos default editor" today] (provided open isn't being used elsewhere that is incompatible with that)

dacheng-gao:
This fixes #3284

### DIFF
--- a/src/os/open.zig
+++ b/src/os/open.zig
@@ -12,7 +12,7 @@ pub fn open(alloc: Allocator, url: []const u8) !void {
     // the process to exit to collect stderr.
     const argv, const wait = switch (builtin.os.tag) {
         .linux => .{ &.{ "xdg-open", url }, false },
-        .macos => .{ &.{ "open", url }, true },
+        .macos => .{ &.{ "open -t", url }, true },
         .windows => .{ &.{ "rundll32", "url.dll,FileProtocolHandler", url }, false },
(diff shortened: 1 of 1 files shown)


src/os/open.zig
```diff
@@ -12,7 +12,7 @@ pub fn open(alloc: Allocator, url: []const u8) !void {
     // the process to exit to collect stderr.
     const argv, const wait = switch (builtin.os.tag) {
         .linux => .{ &.{ "xdg-open", url }, false },
-        .macos => .{ &.{ "open", url }, true },
+        .macos => .{ &.{ "open -t", url }, true },
```
liby, line 15: Thanks for working on this! While using `open -t` directly in the implementation works, it might be better to keep the `open` function generic and pass the `-t` flag as a parameter instead. This way:

1. The `open` function remains flexible for different use cases
2. Other callers of `open` won't be forced to use a text editor
3. The specific opening behavior stays controlled by the caller

mitchellh:
Superseded by #4004. This addresses the comments that @liby made as well, incidentally. Thank you!

PR #3363, opened 2024-12-27
Linux: Added install scripts for bash and nushell

ThatSealgair:
Added two scripts to help users install ghostty before it's available from a user's preferred package manager.

### DIFF
--- a/install_linux.nu
+++ b/install_linux.nu
@@ -1,0 +1,26 @@
+#!/usr/bin/env nu
+
+# Check if running as root
+if ((id -u | into int) != 0) {
+    echo "Please run as root (sudo)"
+    exit 1
+}
+
+# Build and install (errors will automatically cause exit in nu)
+zig build -Doptimize=ReleaseFast --prefix /usr install
+
+# Create desktop entry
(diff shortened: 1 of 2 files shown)


ThatSealgair:
This is my first open source PR. Just wanted to share scripts to make life easier. Let me know if any information is missing, ect.

BWindey:
In what way does this improve over the manual `zig build -p ~/.local -Doptimize=ReleaseFast`, or the sudo version with `-p /usr`?
The above already makes a desktop entry (at least on my machine), and is in the docs: https://ghostty.org/docs/install/build#installation-directory

ThatSealgair:
Yeah, I didn't realise it created a desktop entry. I had to restart my OS for it to show up.
Creating the desktop entry and having a simpler installation process then what the [docs](https://ghostty.org/docs/install/build) provide was the goal. But it's probably better to had the website docs updated to `zig build -Doptimize=ReleaseFast --prefix /usr install` or as you suggested `zig build -p ~/.local -Doptimize=ReleaseFast` if you want it to be local.

Issue #3412, opened 2024-12-27
Graphical artifacts with fractional scaling on Linux


lbrame:
I am using Ghostty built on Linux on a Fedora 41 desktop running GNOME 47, Wayland session, and the scaling level set to 175%. My build is compiled with the optimizer on from commit [a8e5eef](https://github.com/ghostty-org/ghostty/commit/a8e5eef11cc67f87f445626f9ca2993373774bf8). On top of the already known issue #1938 , I have also used numerous artifacts that haappen during regular use. An example is random white or transparent lines everywhere:

![Image](https://github.com/user-attachments/assets/f623b568-5b5a-4df2-9263-9a93378933f1)

![Image](https://github.com/user-attachments/assets/6d90af1c-f874-4933-a5b7-2568ff018270)

Another issue I have seen is that, when the terminal is tiled to an edge of the screen on GNOME, it does not fully adhere to the panel and neighbor windows, but it has a visible gap.

![Image](https://github.com/user-attachments/assets/fe096902-98dd-4367-b266-a617732a14c5)

This only happens on the GTK build. The basic `glfw` output produced by `zig build -Dapp-runtime=glfw run` is unaffected:

![Image](https://github.com/user-attachments/assets/4b09280b-cc47-42d2-a8c0-dd0559902abb)

Issue #1938, opened 2024-07-10
Blurry font rendering under fractional scaling on Linux


190n:
I use 125% scaling on my laptop display, and Ghosty exhibits blurry font rendering. At 100% or 200% scaling, the rendering is crisp, and at 150% or 175% it looks blurry but not as pronounced as at 125%. Here are screenshots for comparison:

100% (note that this one looks blurry in HiDPI mode simply because there aren't many pixels, but it looks sharp when rendered with 100% scaling)
![image](https://github.com/ghostty-org/ghostty/assets/7763597/3b4d4154-4d0f-4213-9194-bf608c22681a)

125%
![image](https://github.com/ghostty-org/ghostty/assets/7763597/7b3585ff-ede4-45d8-a89d-ceb9ac6e63fc)

200%
![image](https://github.com/ghostty-org/ghostty/assets/7763597/2b4242a1-61bd-47a5-9389-879573867e14)

I tried `font-thicken = true` mentioned in #661 and it didn't make much difference.

I'm running Ghostty built from commit b1677b890049cc81ba5848e3f45dc48ab78081ff, on GNOME Wayland 46.2 and Fedora 40 Silverblue. I see in the logs that it does identify that fractional scaling should be used: `Using. fractional scale 1.25 for EGL window (772 665 => 965 831)`.

lbrame:
I can confirm everything here. Framework 16, Fedora 41 with GNOME 47 and 175% scale.

CnTeng:
I have the same problem.

CnTeng:
Use
```shell
GSK_RENDERER=gl xxxx
```
to start other gtk app like foliate or libadwata demo will show the same problem.
These app use vulkan as defualt, but ghostty not.
So, I think is about gtk and opengl.

lbrame:
Seeing this is related to GTK, I tried to reproduce this against GNOME's Ptyxis terminal - which uses GTK 4 and libadwaita - trying to make the font configuration and size roughly close. Ghostty to the left, Ptyxis to the right. Here in the terminal area, Ghostty's text looks fuzzier. What's even stranger is that it Ghostty does not tile to its half of the screen all the way, and there is a scaled pixel's gap of space between the CSD and GNOME's panel. This is a different bug, but this might be related because we are talking about renderers.

![Image](https://github.com/user-attachments/assets/bd762f56-5835-481d-8fc1-eb5d92436fc2)

[...]

PR #3419, opened 2024-12-27
Partially fix issue #1938, fix issue #3412, by enabling the vulkan-based ngl renderer in GTK

lbrame:
## Summary of this change
In this draft pull request, I try changing the GSK renderer to the newer, Vulkan-based `ngl` renderer in the GTK build.

## Context
The motivation behind this PR stemmed from issue #1938 and #3412. Currently, Ghostty has several issues when running on a Wayland desktop with fractional scaling enabled, such as graphical artifacts and blurry font rendering.

## Scope
This change interests source file `src/apprt/gtk/App.zig` and it should only affect Linux systems, for builds compiled with GTK support enabled.

## Implementation details
This is admittedly a pretty quick and dirty attempt that needs more polish. I am open to trying my best to improve it. In the context of the GTK App initialization, I removed the code that disables Vulkan and I switched the `GSK_RENDERER` from `opengl` to `ngl`.

What this does is switch the GSK renderer from the old `gl` renderer to [the new Ngl renderer](https://docs.gtk.org/gsk4/class.NglRenderer.html).

## Impact
This new renderer greatly improves positioning and anti-aliasing in hi-dpi setups with fractional scaling, as better documented in the article [New renderers for GTK](https://blog.gtk.org/2024/01/28/new-renderers-for-gtk/) from the GNOME blog. Several other Libadwaita-based applications are migrating to this new renderer.

There are some immediate benefits to this change.

* I am no longer able to reproduce issue #3412 and there are no more graphical artifacts. Even the window now tiles perfectly, with no gaps.
* It greatly improves font rendering in the GTK interface (title bar and menus). The fonts go from fuzzy to very crisp. This is an improvement for #1938 

## Examples
This is how the fonts on the main menu and title bar look with the `opengl` renderer:
image

image
There is a fuzzy halo around the fonts.

Here is the same thing with `ngl`:
image
image

Here is how the "About Ghostty" dialog compares:

image

image

However, I am not sure this change impacts the contents of the terminal window as of now. I left a screenshot to help validate that. Left: stock build. Right: build from this PR. Below: output from `zig build -Dapp-runtime=glfw run`. Note that the latter `glfw` build runs through the XWayland socket, so it does not have access to the native resolution framebuffer as GTK under Wayland can as per [Wayland fractional-scale-v1 protocol](https://wayland.app/protocols/fractional-scale-v1), but rather it uses a supersampling trick which makes rendering different.

image

All of this is being tested on a machine running Fedora Workstation 41, `gnome-shell` 47.2, `mutter` 47.3,  `gtk4` 4.16.5 with a Wayland session and fractional scaling set to 175%. Base build is from commit a8e5eef.

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -100,40 +100,10 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         c.gtk_get_micro_version(),
     });
 
-    // Disabling Vulkan can improve startup times by hundreds of
-    // milliseconds on some systems. We don't use Vulkan so we can just
-    // disable it.
-    if (version.atLeast(4, 16, 0)) {
-        // From gtk 4.16, GDK_DEBUG is split into GDK_DEBUG and GDK_DISABLE.
-        // For the remainder of "why" see the 4.14 comment below.
-        _ = internal_os.setenv("GDK_DISABLE", "gles-api,vulkan");
-        _ = internal_os.setenv("GDK_DEBUG", "opengl");
-    } else if (version.atLeast(4, 14, 0)) {
-        // We need to export GDK_DEBUG to run on Wayland after GTK 4.14.
-        // Older versions of GTK do not support these values so it is safe
-        // to always set this. Forwards versions are uncertain so we'll have to
-        // reassess...
-        //
-        // Upstream issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/6589
-        //
-        // Specific details about values:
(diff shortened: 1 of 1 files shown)


lbrame:
I have looked further into the history of this change and came across this: https://github.com/ghostty-org/ghostty/pull/1891

It seems like completely reverting this change is not an acceptable solution, because it breaks on some systems. For what it's worth, I am unable to reproduce the flickers on `ngl`, but I understand this might be a behavior specific to my GPU.

I'll close this for now, and go back to the drawing board and open a discussion.

PR #3426, opened 2024-12-27
Fix `DESTDIR` handling for terminfo installation

### LINKED ISSUE #3152: DESTDIR not respected during zig build
Using the `DESTDIR` as advertised in PACKAGING.md does *not* work, it still tries to write things directly to the prefix directory. Only using `--prefix "$DESTDIR/usr"` seems to work for me.

liby:
## Description:

Fix `DESTDIR` handling when installing terminfo database files by using `install_path` instead of `install_prefix`. This ensures files are correctly installed under `$DESTDIR/$prefix` during packaging.

## Changes:

- Replace `b.install_prefix` with `b.install_path` for terminfo database installation paths

- This change properly respects the `DESTDIR` environment variable during installation

## Testing:

I've verified this fix by:

1. Setting `DESTDIR=/tmp/ghostty`

2. Building with: 
```bash
DESTDIR=/tmp/ghostty \
zig build \
  --prefix /usr \
  --system /tmp/offline-cache/p \
  -Doptimize=ReleaseFast \
  -Dcpu=baseline
```

3. Confirming files are correctly installed to:

```
/tmp/ghostty/usr/share/terminfo/ghostty.terminfo
/tmp/ghostty/usr/share/terminfo/ghostty.termcap
```

The files are now properly installed under `$DESTDIR/$prefix` path structure as expected.

cc @BratishkaErik - Since you suggested this fix in #3152, would you mind reviewing this implementation?

Fixes #3152

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -481,19 +481,25 @@ pub fn build(b: *std.Build) !void {
             run_step.step.dependOn(&src_install.step);
 
             {
+                // Use cp -R instead of Step.InstallDir because we need to preserve
+                // symlinks in the terminfo database. Zig's InstallDir step doesn't
+                // handle symlinks correctly yet.
                 const copy_step = RunStep.create(b, "copy terminfo db");
                 copy_step.addArgs(&.{ "cp", "-R" });
                 copy_step.addFileArg(path);
-                copy_step.addArg(b.fmt("{s}/share", .{b.install_prefix}));
+                copy_step.addArg(b.fmt("{s}/share", .{b.install_path}));
                 b.getInstallStep().dependOn(&copy_step.step);
             }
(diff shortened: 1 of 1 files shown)


BratishkaErik:
> cc @BratishkaErik - Since you suggested this fix in #3152, would you mind reviewing this implementation?

I'd suggest to also add comment on why `cp` is used here instead of Zig step. Other than that seems fine. Smth like:

> Step.InstallDir does not copies symlinks in this Zig version,
> so use `cp -R` here instead as a hack.

BratishkaErik (approved):

mitchellh:
Thank you!

Issue #2866, opened 2024-12-01
Inconsistency in `goto_split` between Linux and macOS


rafikdraoui:
After `new_split:right` followed by `new_split:down`, the layout looks like this, with the focus in split C:

```
┌─────┬─────┐
│     │  B  │
│  A  ├─────┤
│     │  C  │
└─────┴─────┘
```

After `goto_split:left`, on macOS the focus is in split A, but in GTK the focus is in split B.

I believe that the GTK behavior should be changed to match the one from macOS.

isgkm:
Can confirm on linux the focus goes from C -> B with goto_split:left instead of C -> A, also while at B and doing goto_split:right the focus is moved to C, which probably isn't the desired behavior, it should either move focus back to A and loop or not do anything at all, which is the scenario with goto_split:right while at C.

pthieu:
Can confirm what @cosmicboots's scenario on mac
I have a 2x3 (2 cols, 3 rows each) setup:
```
+---------------+---------------+
|               |               |
|       A       |       D       |
|               |               |
+---------------+---------------+
|               |               |
|       B       |       E       |
|               |               |
+---------------+---------------+
|               |               |
|       C       |       F       |
|               |               |
+---------------+---------------+
```

It's not intuitive triggering `goto_split:left/right` on any of the splits, going left on E could hit C and going left on D could also hit C, depending on how the splits were created

nerozero:
Confirm that, experiencing same issue

ariasmn:
Any updates on this? If it's open for work, I'm sure either someone or I could (try to) contribute.

jcollie:
Closed by #5654

rawkode:
I don't think this is fixed at all.

```
┌─────┬─────┐
│  A  │  B  │
|─────├─────┤
│  C  │  D  │
└─────┴─────┘
```

With this configuration:

- When on A, using goto_split:down goes to B (Should be C)
- When on B, using goto_split:down goes to C (Should be D)
- When on D, using goto_split:up   goes to C (Should be B)


```
❯ ghostty +version
Ghostty 1.1.4-4aa2898-nix

Version
  - version: 1.1.4-4aa2898-nix
  - channel: tip
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.gtk
  - font engine   : font.main.Backend.fontconfig_freetype
[...]

PR #3427, opened 2024-12-27
Rename `goto_split` top/bottom directions to up/down.

### LINKED ISSUE #3237: Keybinds: Inconsistent Direction Parameters for `new_split` and `goto_split`
#### Description

When configuring keybinds, the direction parameters for `new_split` and `goto_split` are inconsistent:

- `new_split` uses: `up | down | left | right`
- `goto_split` uses: `previous | next | top | left | bottom | right`

The goto_split direction parameters include top and bottom, which can be misleading as they might imply focusing on the topmost or bottommost split in the layout, rather than simply moving to an adjacent split. This inconsistency with new_split (up and down) may confuse users and create ambiguity in the configuration.

#### Steps to Reproduce

1. Configure `new_split` keybinds:
   ```plaintext
   keybind = ctrl+s>h=new_split:left
   keybind = ctrl+s>j=new_split:down
   keybind = ctrl+s>k=new_split:up
   keybind = ctrl+s>l=new_split:right
   ```
2. Configure `goto_split` keybinds:
   ```plaintext
   keybind = ctrl+h=goto_split:left
   keybind = ctrl+j=goto_split:bottom
   keybind = ctrl+k=goto_split:top
   keybind = ctrl+l=goto_split:right
   ```

#### Expected Behavior

The direction parameters should be consistent across both actions.

#### Actual Behavior

`new_split` and `goto_split` use different sets of direction parameters.

#### Suggested Change

Standardize the direction parameters to a single set, such as `previous | next | up | left | down | right`.

This would improve clarity and ease of configuration.

dpatterbee:
Renames the top/bottom directions of `goto_split` to up/down. I have tested this on linux (nixos) but given that `goto_split` is broken on linux anyway (#2866) there's not a whole lot to test.

I have no way to build on macOS so I can't verify that I've changed everything correctly for that.

Closes #3237


mitchellh:
Thanks this looks good but I'm a bit worried now that we're released about backwards compatibility. Let me think about that.

dpatterbee:
Here's a small patch which would retain "top" and "bottom" as aliases of the new "up" and "down" directions if that's the direction you'd like to go. I agree that breaking backwards compatibility 2 days post-launch is probably not ideal.

```
index 1442af86..b4e9e593 100644
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4046,6 +4046,8 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
             .{ .surface = self },
             .goto_split,
             switch (direction) {
+                .top => apprt.action.GotoSplit.up,
+                .bottom => apprt.action.GotoSplit.down,
                 inline else => |tag| @field(
                     apprt.action.GotoSplit,
                     @tagName(tag),
diff --git a/src/input/Binding.zig b/src/input/Binding.zig
index f8cc71d0..1c318852 100644
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -473,6 +473,11 @@ pub const Action = union(enum) {
         left,
         down,
         right,
+
+        // For backwards compatilbility purposes we also support "top" and "bottom" as aliases for
+        // "up" and "down"
+        top,
+        bottom,
     };
 
     pub const SplitResizeDirection = enum {
```

mitchellh:
Thanks. It might be easier to actually implement a custom `parseCLI` function for this enum and translate the old values to the new values. That would clean up everything else. If you do this I will merge.

Please add tests for that as well.

Thank you!

gigamonster256:
Documentation for goto_split was just added in #4388 which is out of date as of this being merged (assuming new names should be in the documentation instead of legacy top/bottom)

PR #3460, opened 2024-12-27
feat(config): generate default template when config file is not found

### LINKED ISSUE #3203: Default config file contents
Config files created automatically by the `open_config` action should have an explanation for users, because an empty file just staring you in the face doesn't leave you much of a path forward without turning around and looking at the documentation. While the user *should* be reading documentation, it's also preferable for them to not need to proactively do so in most cases.

The default config file then should be a header which explains the config and references the documentation. Important elements of this explanation:
- **Where is this file located?** So that the user knows how to edit it in their preferred editor if they don't like the one that was opened when they used the action. This should prevent issues where a user creates a config on macOS in the `Application Support` location and then a second one in the XDG location which conflicts.
- **What is this file format?** So that the user knows how to specify config options.
- **What config options are available?** A link to the docs should be included as well as a reference to `+show-config`.

### Here is my suggestion:
```
# </path/to/ghostty/config>
#
# Docs: https://ghostty.org/docs/config
#
# This is the configuration file for Ghostty. This file is created automatically
# if the `open_config` action is invoked and Ghostty can't find any config files
# on your system. It has been placed at the path listed at the top of the file.
#
# This file is not initialized with any default options set because the majority
# of users should only need to set a few things. While it is recommended to read
# the docs, a brief explanation of the syntax for this file can be found below.
#
# Syntax
# ---
# # Any line beginning with a # is a comment. It's not possible to put
# # a comment after a config option, it will be interpreted as a part
# # of the value. For example, this will have a value of "#123abc":
# background = #123abc
#
# # Spacing around the equals sign does not matter.
# # All of these are identical:
# key=value
# key= value
# key =value
# key = value
#
# # Empty values are used to reset config keys to default
# key =
#
# # Some config options have unique syntax for their value,
# # which is explained in the docs for that config option.
# ---
#
# Tip: You can run `ghostty +show-config --default --docs` to view a
#      list of all available config options and their default values.
#
# See also: `man ghostty`
#
# Please consider reading the full docs.
```

pluiedev:
Closes #3203

### DIFF
--- a/src/config/edit.zig
+++ b/src/config/edit.zig
@@ -48,13 +48,7 @@ pub fn open(alloc_gpa: Allocator) !void {
 ///
 /// The allocator must be an arena allocator. No memory is freed by this
 /// function and the resulting path is not all the memory that is allocated.
-///
-/// NOTE: WHY IS THIS INLINE? This is inline because when this is not
-/// inline then Zig 0.13 crashes [most of the time] when trying to compile
-/// this file. This is a workaround for that issue. This function is only
-/// called from one place that is not performance critical so it is fine
-/// to be inline.
-inline fn configPath(alloc_arena: Allocator) ![]const u8 {
+fn configPath(alloc_arena: Allocator) ![]const u8 {
     const paths: []const []const u8 = try configPathCandidates(alloc_arena);
     assert(paths.len > 0);
 
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2668,18 +2668,40 @@ pub fn loadFile(self: *Config, alloc: Allocator, path: []const u8) !void {
     try self.expandPaths(std.fs.path.dirname(path).?);
 }
 
+pub const OptionalFileAction = enum { loaded, not_found, @"error" };
+
 /// Load optional configuration file from `path`. All errors are ignored.
-pub fn loadOptionalFile(self: *Config, alloc: Allocator, path: []const u8) void {
-    self.loadFile(alloc, path) catch |err| switch (err) {
-        error.FileNotFound => std.log.info(
-            "optional config file not found, not loading path={s}",
-            .{path},
-        ),
-        else => std.log.warn(
-            "error reading optional config file, not loading err={} path={s}",
-            .{ err, path },
-        ),
-    };
+///
+/// Returns the action that was taken.
+pub fn loadOptionalFile(
+    self: *Config,
+    alloc: Allocator,
+    path: []const u8,
+) OptionalFileAction {
+    if (self.loadFile(alloc, path)) {
+        return .loaded;
+    } else |err| switch (err) {
+        error.FileNotFound => return .not_found,
+        else => {
+            std.log.warn(
+                "error reading optional config file, not loading err={} path={s}",
+                .{ err, path },
+            );
+
+            return .@"error";
+        },
+    }
+}
+
+fn writeConfigTemplate(path: []const u8) !void {
+    log.info("creating template config file: path={s}", .{path});
+    const file = try std.fs.createFileAbsolute(path, .{});
+    defer file.close();
+    try std.fmt.format(
+        file.writer(),
+        @embedFile("./config-template"),
+        .{ .path = path },
+    );
 }
 
 /// Load configurations from the default configuration files. The default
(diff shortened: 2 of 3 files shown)


qwerasd205:
Notably, this PR automatically initializes the file on the first attempted load, rather than as a part of `open_config` (which is handled by `open` in `./src/config/edit.zig`). I'm not sure how I feel about this.

qwerasd205:
As for the contents of the template- there are some changes to my template that are definitely improvements, but some of the new wording is a little sloppy, the text jumps around between moods in a strange way, extraneous commentary is included which will reduce the change of a user properly reading it (users could not care less about the philosophical basis of the configuration file, they just want to change their font and color scheme), and not leading with the docs link front and center is definitely worse.

mitchellh:
> Notably, this PR automatically initializes the file on the first attempted load, rather than as a part of open_config (which is handled by open in ./src/config/edit.zig). I'm not sure how I feel about this.

I agree, I don't think we should be doing this in the load portion.

pluiedev:
> Notably, this PR automatically initializes the file on the first attempted load, rather than as a part of `open_config` (which is handled by `open` in `./src/config/edit.zig`). I'm not sure how I feel about this.

I feel like I prefer this over only generating when `open_config` is run, since another common point of feedback is that some people *didn't even know* that you can open the config by pressing `Ctrl/Cmd-,`.

Plus, when they follow the docs and try to find the config file at `~/.config/ghostty/config` they get quite confused for one reason or another (some people mistake it for a directory, some people ask "wait, where's the `ghostty` directory in `~/.config`?", etc.) Therefore I'd argue that only generating the file upon `open_config` only addresses half the problem.

pluiedev:
> As for the contents of the template- there are some changes to my suggested template that are definitely improvements, but some of the new wording is a little sloppy, the text jumps around between moods in a strange way, 

Proofreads & wording suggestions welcome, but an issue I had with the original version is that it feels too cold/robotic for an introductory document — I tried to rephrase it with a more welcoming tone but maybe now it seems a bit disjointed

> extraneous commentary is included which will reduce the chance of a user properly reading it (users could not care less about the philosophical basis of the configuration file, they just want to change their font and color scheme), 

Zero-config is a great selling point of Ghostty — heck, it's even everywhere on the website. I think it's a needed point of justification as to *why* we're not like, say, Kitty, where the default config is a full list of options.

> and not leading with the docs link front and center is definitely worse.

That can definitely be moved

edrozenberg:
Love this:
```
This template file has been automatically created at path
#   {[path]s} 
```

As a new user, Ghostty created the config file for me and I had a bit of a hl of a time figuring out where it put it, Application Support (Mac) was not my first second or fifth guess, I looked  in ~/.config and then no clue. Eventually I found it. This template text will be a big help.

src/config/edit.zig
```diff
@@ -48,13 +48,7 @@ pub fn open(alloc_gpa: Allocator) !void {
...
 /// The allocator must be an arena allocator. No memory is freed by this
 /// function and the resulting path is not all the memory that is allocated.
-///
-/// NOTE: WHY IS THIS INLINE? This is inline because when this is not
-/// inline then Zig 0.13 crashes [most of the time] when trying to compile
-/// this file. This is a workaround for that issue. This function is only
-/// called from one place that is not performance critical so it is fine
-/// to be inline.
-inline fn configPath(alloc_arena: Allocator) ![]const u8 {
+fn configPath(alloc_arena: Allocator) ![]const u8 {
```
mitchellh, line 51: This is just an unrelated change I snuck in here because I noticed it.

mitchellh (approved):
Alright, I made some edits. I'm still not super happy with how this works but it does the job and I think it'll help users quite a lot. Thanks @pluiedev

HeySora:
There might be an issue in this PR, considering that this happened on my first launch after updating to 1.0.1.

image

This will actually fail if the `~/Library/Application Support/com.mitchellh.ghostty` directory does not exist, which can be fixed by doing `cmd+,`, but this creates an empty `config` file, without the template. Because of this, ghostty will not attempt to create a template file, considering the file exists (despite being empty). Closing ghostty, deleting that file while keeping the folder, then restarting ghostty properly creates the template file this time!

A workaround for this would be to also generate the template by doing `cmd+,` for the first time, in the code path that creates the folder and file.

edrozenberg:
I've not been able to ever get a template to show up with 1.0.1. Always blank. `Cmd ,` always opens a blank config, in my editor TextMate, regardless of any directory/missing directory in App Supp.

mitchellh:
Does the directory exist? I just deleted my config files (moved them) and it gets created fine. But @HeySora is correct that it requires the dir to exist which is unfortunate. I'm making an issue to track that now.

edrozenberg:
It worked after these steps:

1. Close Ghostty windows, Quit Ghostty

2. `rm -rf ~/Library/Application\ Support/com.mitchellh.ghostty`

3. Open Ghostty, `Cmd ,`

4. Blank `config` file opens in TextMate

5. `cat ~/Library/Application\ Support/com.mitchellh.ghostty/config` is blank

6. `rm -f ~/Library/Application\ Support/com.mitchellh.ghostty/config`

7. Open Ghostty, `Cmd ,`

8. Template config file shows up in TextMate ✅

liby:
Hi @HeySora,

> image

Interesting observation! While trying to reproduce this issue, I noticed something different. I can only see the error message in the logs: 

```
2025-01-01 17:30:11.159045+0000 Info ghostty: [com.mitchellh.ghostty:config] creating template config file: path=/Users/bryan/Library/Application Support/com.mitchellh.ghostty/config
2025-01-01 17:30:11.159075+0000 Error ghostty: [com.mitchellh.ghostty:config] error creating template config file err=error.FileNotFound
```

But I don't see any UI modal/notification like the one in your screenshot. I'm curious:

- Did you have any special configuration to enable these error notifications?

- Or perhaps you're running on a different platform? I'm on macOS with:
  ```
  Ghostty 1.0.1
  - channel: stable
  - renderer: renderer.Metal
  ```

Would be helpful to know your setup as it seems to provide a better user experience with visible error notifications!

PR #3490, opened 2024-12-27
CI: Update release-tag.yml to include the version in the source archive and prefix within

AnthonyZhOon:
Continuing from #3043 I agree that it seems idiomatic to have an archive with format <name>-<version>.tar.gz and matching prefix for packaging, RPM and Debian packaging guides seem to assume this format and the automated extract tooling assumes it too.

# Testing
I haven't tested running this workflow,  and am unsure about the yaml substitution at lines 105-106

# Breaking changes
This would break existing packaging scripts, not sure how we want to version it

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -87,21 +87,21 @@ jobs:
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
       - name: Create Tarball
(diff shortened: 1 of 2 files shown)


mitchellh:
Yeah. I think we need to rip the bandaid and probably just break compatibility on this one. This just seems overall more idiomatic.

mitchellh:
I decided to be a bit more graceful. I kept the old style as well but updating our PACKAGING.md to only mention the new one.

PR #3511, opened 2024-12-27
Fix example of using ZSH integration manually

Tenzer:
The ZSH integration file should be sourced so it's read into the current process, rather than executed in a separate ZSH process, in which case it doesn't have any impact on the current ZSH process.


lishaduck:
I think this is a dup of #3332.

00-kat:
Actually this wasn't a dupe since it was incorrect in two different places (the other PR updated README.md, this updates the comments in the Zsh shell integration source code), but I myself forgot to check closed PRs for dupes when i created #3680 (I'll leave it to Mitchell to decide which PR to merge).

lishaduck:
> Actually this wasn't a dupe since it was incorrect in two different places

Oh, yup. My apologies, @Tenzer. 😔

PR #3526, opened 2024-12-27
make target specific dependencies lazy

Jan200101:
spiritual follow-up to #1447

turns zig_objc and zig_js into lazy dependencies so they aren't required when building for different targets using the system package mode 

I don't have a mac to test this so I'm at the mercy of someone else doing that for me

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1210,11 +1210,11 @@ fn addDeps(
     }
 
     // Wasm we do manually since it is such a different build.
-    if (step.rootModuleTarget().cpu.arch == .wasm32) {
-        const js_dep = b.dependency("zig_js", .{
+    if (step.rootModuleTarget().cpu.arch == .wasm32) wasm: {
+        const js_dep = b.lazyDependency("zig_js", .{
             .target = target,
             .optimize = optimize,
-        });
+        }) orelse break :wasm;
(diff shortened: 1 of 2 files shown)


der-teufel-programming:
Great idea

Jan200101:
It appears as if making a dependency lazy stops zig fetch from automatically downloading it.
This hasn't been an issue with GLFW since its not used as default anywhere but making zig_objc and zig_js into lazy dependencies will possibly break Nix builds on macOS.

mitchellh:
I'm conceptually fine with this but the nix CI jobs and various other checks need to be fixed.

mitchellh:
> maybe its best to wait out until https://github.com/ziglang/zig/pull/19975 gets merged?

I'd prefer this.

mitchellh:
Can you reopen your PR when we upgrade Zig? (and that is merged). Thanks.

jcollie:
> the alternative would be to include zig_objc specifically in the fetch script but that would then result in 2 places where the dependencies need to be maintained.

Yes, unfortunately maintaining the dependencies in two places is what we have to deal with now for lazy and transitive dependencies. We can look at switching to what the "official" nix package did - using `zig2nix` to convert the `.zig.zon` to a Nix expression. It's not that much harder to maintain than the current system and it handles lazy and transitive dependencies.

PR #3560, opened 2024-12-27
github: fix prettier check for preapproved issue template

luk1337:


luk1337:
saw that CI was failing on that ^^

mitchellh:
Ah just fixed myself. Thanks.

PR #3569, opened 2024-12-27
apprt/gtk: color popovers when window-theme=ghostty

tristan957:
This looks better than the regular dark color. It also happens to match what Ptyxis does. It does not support non-libadwaita builds.

Before:
image

After:
image

It will look even better whenever we fix the separator colors :smile:


mitchellh:
Lovely thanks.

hedgieinsocks:
@tristan957 hi, I wonder whether you got an idea how to force a themed background color in the tab overview ;)

tristan957:
I do after reading the documentation. I will explore that today :smile:

PR #3570, opened 2024-12-27
apprt/gtk: use a subtitle to mark the current working directory

tristan957:
If the title is already the current working directory, hide the subtitle. Otherwise show the current working directory, like if a command is running for instance.

This is a re-opening of my original PR because I had to delete my fork and re-fork it.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1118,6 +1118,15 @@ keybind: Keybinds = .{},
 /// required to be a fixed-width font.
 @"window-title-font-family": ?[:0]const u8 = null,
 
+/// The text that will be displayed in the subtitle of the window. Valid values:
+///
+///   * `false` - Disable the subtitle.
+///   * `working-directory` - Set the subtitle to the working directory of the
+///      surface.
+///
+/// This feature is only supported on GTK with Adwaita enabled.
+@"window-subtitle": WindowSubtitle = .false,
+
 /// The theme to use for the windows. Valid values:
 ///
 ///   * `auto` - Determine the theme based on the configured terminal
@@ -3966,6 +3975,11 @@ pub const WindowPaddingColor = enum {
     background,
     extend,
     @"extend-always",
+};
+
+pub const WindowSubtitle = enum {
+    false,
+    @"working-directory",
 };
 
 /// Color represents a color using RGB.
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -347,6 +347,11 @@ cursor: ?*c.GdkCursor = null,
 /// pass it to GTK.
 title_text: ?[:0]const u8 = null,
 
+/// Our current working directory. We use this value for setting tooltips in
+/// the headerbar subtitle if we have focus. When set, the text in this buf
+/// will be null-terminated because we need to pass it to GTK.
+pwd: ?[:0]const u8 = null,
+
 /// The timer used to delay title updates in order to prevent flickering.
 update_title_timer: ?c.guint = null,
 
@@ -628,6 +633,7 @@ fn realize(self: *Surface) !void {
 pub fn deinit(self: *Surface) void {
     self.init_config.deinit(self.app.core_app.alloc);
     if (self.title_text) |title| self.app.core_app.alloc.free(title);
+    if (self.pwd) |pwd| self.app.core_app.alloc.free(pwd);
 
     // We don't allocate anything if we aren't realized.
     if (!self.realized) return;
@@ -876,7 +882,7 @@ fn updateTitleLabels(self: *Surface) void {
             // I don't know a way around this yet. I've tried re-hiding the
             // cursor after setting the title but it doesn't work, I think
             // due to some gtk event loop things...
-            c.gtk_window_set_title(window.window, title.ptr);
+            window.setTitle(title);
         }
     }
 }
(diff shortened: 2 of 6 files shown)


gpanders:
This is a nice touch, but are there any plans to make this optional?

Fish sets the window title to my current working directory, but in a "condensed" form, so e.g. `/home/ganders/src/ghostty` is `~/s/ghostty`. But since this string doesn't match the pwd exactly, I see this:

image

This is obviously redundant and so just ends up adding visual noise that's unhelpful.

Disabling fish's title setting is not really a good option either, since then I wouldn't have those titles in the tab titles (where the subtitle isn't present).

tristan957:
Yeah we can make this optional. Do you have a preferred setting name? `gtk-cwd-subtitle`?

gpanders:
>In theory, we could make it free-form using an strftime like interface.

I had the same thought, which would also make disabling it easy (just leave it empty). But longer term that would allow it to show not only PWD, but also things like current running command (using shell integration) or hostname (using OSC 7).

tristan957:
Maybe, I should shoot for the moon, and just add something like that in this PR. What kind of syntax should we choose? Probably something like `"{cwd}: {hostname} {{these are escaped}}"?

jcollie:
> Maybe, I should shoot for the moon, and just add something like that in this PR. What kind of syntax should we choose? Probably something like `"{cwd}: {hostname} {{these are escaped}}"?

I wonder how soon we'll get a request to have `{time} {date}`?. Actually, NGL, I'd probably use something like that.

mitchellh (changes requested):
I'm ready to get merging on this one, thanks for sticking with me. My final request is that this be optional. I'm willing to default it to true for now and see how it feels on my day to day usage but I think as @gpanders noted an option to turn this off would be good.

I would suggest `window-subtitle` and make it an `enum { false, @"working-directory" }`. This will let us add more in the future if we want. It also lets us make it cross-platform in the future if we want. And an enum is forward compatible to a `union(enum)` if we ever want to introduce a formatted string as this PR comment notes.

FantomeBeignet:
> > Maybe, I should shoot for the moon, and just add something like that in this PR. What kind of syntax should we choose? Probably something like `"{cwd}: {hostname} {{these are escaped}}"?
> 
> I wonder how soon we'll get a request to have `{time} {date}`?. Actually, NGL, I'd probably use something like that.

That's kinda what I requested in #4096

tristan957:
> I would suggest `window-subtitle` and make it an `enum { false, @"working-directory" }`. This will let us add more in the future if we want. It also lets us make it cross-platform in the future if we want. And an enum is forward compatible to a `union(enum)` if we ever want to introduce a formatted string as this PR comment notes.

I almost feel like instead of `working-directory`, it would be better to see something like `{working-directory}`, and just make it super future proof, but I'm obviously more than happy to defer :smile:.

mitchellh:
> I almost feel like instead of working-directory, it would be better to see something like {working-directory}, and just make it super future proof, but I'm obviously more than happy to defer 😄.

I think enums are going to be a lot simpler (and type safe for free) to add new values for. Then we can use this enum later to also build up the format strings. At least that's what I'm thinking. I'm also not 100% sold on the format string yet so this lets us get this merged more quickly rather than combining two features in one PR.

tristan957:
Thanks for coming in with a great counter-example @gpanders!

tristan957:
I'm not really sure the best way to fix this issue.

```
src/apprt/gtk/notebook_adw.zig:161:21: error: expected type '[:0]const u8', found '[*c]const u8'
    window.setTitle(title);
                    ^~~~~
src/apprt/gtk/Window.zig:431:39: note: parameter type declared here
pub fn setTitle(self: *Window, title: [:0]const u8) void {
                                      ^~~~~~~~~~~~
referenced by:
    init: src/apprt/gtk/notebook_adw.zig:41:86
    init: src/apprt/gtk/notebook.zig:26:61
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
src/apprt/gtk/notebook_gtk.zig:260:21: error: expected type '[:0]const u8', found '[*c]const u8'
    window.setTitle(label_text);
                    ^~~~~~~~~~
src/apprt/gtk/Window.zig:431:39: note: parameter type declared here
pub fn setTitle(self: *Window, title: [:0]const u8) void {
```

jcollie:
> I'm not really sure the best way to fix this issue.
> 
> ```
> src/apprt/gtk/notebook_adw.zig:161:21: error: expected type '[:0]const u8', found '[*c]const u8'
>     window.setTitle(title);
>                     ^~~~~
> src/apprt/gtk/Window.zig:431:39: note: parameter type declared here
> pub fn setTitle(self: *Window, title: [:0]const u8) void {
>                                       ^~~~~~~~~~~~
> referenced by:
>     init: src/apprt/gtk/notebook_adw.zig:41:86
>     init: src/apprt/gtk/notebook.zig:26:61
>     remaining reference traces hidden; use '-freference-trace' to see all reference traces
> src/apprt/gtk/notebook_gtk.zig:260:21: error: expected type '[:0]const u8', found '[*c]const u8'
>     window.setTitle(label_text);
>                     ^~~~~~~~~~
> src/apprt/gtk/Window.zig:431:39: note: parameter type declared here
> pub fn setTitle(self: *Window, title: [:0]const u8) void {
> ```

`std.mem.span` will take a `[*c]u8` as returned from a C function and convert it to a `[:0]u8`. It's used a few places in the GTK apprt.

src/config/Config.zig
```diff
@@ -1076,6 +1076,15 @@ keybind: Keybinds = .{},
...
 @"window-title-font-family": ?[:0]const u8 = null,
 
+/// The text that will be displayed in the subtitle of the window. Valid values:
+///
+///   * `false` - Disable the subtitle.
+///   * `working-directory` - Set the subtitle to the working directory of the
+///      surface.
+///
+/// This feature is only supported on GTK with Adwaita enabled.
+@"window-subtitle": WindowSubtitle = .false,
```
tristan957, line 1086: I made the default false. Easier for me to just enable than make a whole bunch of people disable :smile:

mitchellh: I'm happy to reassess "zero config defaults" as more feedback arrives when something is shipped. I'm very supportive of the least impactful config being the default to start, and if we find many people are turning it on then we switch course.

jcollie (approved):

PR #3587, opened 2024-12-27
Modify cmd+9 behaviour

### LINKED ISSUE #3528: macOS: default `cmd+9` binding should be `last_tab`, not `goto_tab`
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3361

<sup>Originally posted by **Samasaur1** December 27, 2024</sup>
This is a combination of a question about changing the default behavior and something that is kind of bug-like.

#### Behavior

In Terminal.app and every other Mac app I can think of, ⌘9 goes to the last tab instead of the 9th tab when more than 9 tabs exist. In Ghostty, however, ⌘9 selects the 9th tab in that scenario.

#### Bug

I was able to override this behavior with
```
keybind = command+nine=last_tab
```
but this did not change the little hints that Ghostty inserts for each tab:
screenshot of the tab bar with 10 tabs showing

#### Next steps

I'm not really sure what exactly I'm asking for here. Mostly I guess I'm saying that I'm surprised that ⌘9 *didn't* go to the last tab with no configuration, since that's the "usual" Mac app behavior. It would also be nice if the hints in the tab bar updated if you set keybindings like I have done, although I imagine that the current hints are hardcoded and getting this to work would be a not insignificant amount of work for very little benefit.</div>

br0h1t:
Fixes #3528 

**Note:**
Not sure if this is what OP requested regarding displaying user defined shortcuts as hints, but we currently relabel all tabs, but only consider labels for tabs from 1 to 9. Applying this simple patch should allow for displaying shortcut hints for tabs beyond 9
<details><summary>Patch</summary>

```diff
diff --git a/macos/Sources/Features/Terminal/TerminalController.swift b/macos/Sources/Features/Terminal/TerminalController.swift
index 7fd1802d..242e5711 100644
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -155,13 +155,6 @@ class TerminalController: BaseTerminalController {
         tabListenForFrame = windows.count > 1
 
         for (tab, window) in zip(1..., windows) {
-            // We need to clear any windows beyond this because they have had
-            // a keyEquivalent set previously.
-            guard tab <= 9 else {
-                window.keyEquivalent = ""
-                continue
-            }
-
             let action = "goto_tab:\(tab)"
             if let equiv = ghostty.config.keyEquivalent(for: action) {
                 window.keyEquivalent = "\(equiv)"
```

</details>

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2303,18 +2303,18 @@ pub fn default(alloc_gpa: Allocator) Allocator.Error!Config {
         );
     }
     {
+        // On macOS we default to super but everywhere else
+        // is alt.
+        const mods: inputpkg.Mods = if (builtin.target.isDarwin())
+            .{ .super = true }
+        else
+            .{ .alt = true };
+
         // Cmd+N for goto tab N
         const start = @intFromEnum(inputpkg.Key.one);
-        const end = @intFromEnum(inputpkg.Key.nine);
+        const end = @intFromEnum(inputpkg.Key.eight);
         var i: usize = start;
         while (i <= end) : (i += 1) {
-            // On macOS we default to super but everywhere else
-            // is alt.
-            const mods: inputpkg.Mods = if (builtin.target.isDarwin())
-                .{ .super = true }
-            else
-                .{ .alt = true };
-
             try result.keybind.set.put(
                 alloc,
                 .{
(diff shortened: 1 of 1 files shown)


mitchellh:
Made some changes but looked great. Thank you. once CI is green will merge.

PR #3619, opened 2024-12-28
Wasm shaping and atlas rendering again

gabydd:
Restarting the wasm pr cause github closed it starting with just getting things building for wasm freestanding and improving some metrics, also added an example of shaped glyph being rendered on the canvas
image

I think some things happened in the rebase in the build.zig which I will fix

I have locally opengl rendering, termio and apprt enough to get https://codeberg.org/zenith-editor/zenith to work in the browser but I think incrementally we can get to a better solution that enables improving libghostty

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -15,6 +15,8 @@ pub fn build(b: *std.Build) !void {
     // Ghostty dependencies used by many artifacts.
     const deps = try buildpkg.SharedDeps.init(b, &config);
     const exe = try buildpkg.GhosttyExe.init(b, &config, &deps);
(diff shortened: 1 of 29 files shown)


mitchellh:
Going to close this since we don't plan on merging it as-is. I appreciate the work though and we'll surely take some of this to bring our warm builds back up to speed eventually.

gabydd:
No problem, sorry this isn't in the greatest shape, happy to review, discuss or implement when there is a clearer idea of the direction wasm support goes in.

PR #3664, opened 2024-12-28
docs: update guides on handling number identifier

hisamafahri:
Related: https://github.com/ghostty-org/website/pull/170#discussion_r1898746444


mitchellh:
For now I'll accept this but I plan on adding aliases for these.

Issue #3400, opened 2024-12-27
adjust-underline-thickness does not support fractional values


pingiun:
I don't think I fully understand how the adjust-underline-thickness config option works, but I've tried setting it to 1, and to 0.5. On the value of 1, it seems thicker than the default option. On 0.5 I get an error that the value could not be read.

The default thickness looks like this:

![Image](https://github.com/user-attachments/assets/574983ea-e3a8-4d13-a42d-f82ff6f9494d)

A value of 1 looks like this:

![Image](https://github.com/user-attachments/assets/9a591482-78fb-4e7f-b0ba-302749ae3730)

(Screenshots on macOS, where I think it makes sense to have a 0.5 pixel line on retina screens)

A value of 0.5 gives an error:

```
adjust-underline-thickness: unknown error error.InvalidFormat
```

Current behaviour:
- It seems impossible to set a thinner than default underline thickness

Desired behaviour:
- I want to have thinner underlines than the default

I understand that text decorations aren't fully fleshed out yet, but could not find an issue yet for the request of thinner underlines, so that's why I'm making this issue

pingiun:
I've found that I can set the value to `adjust-underline-thickness = -80%`, which makes the underline thinner. It was not clear from the docs that this value allows percentages

pingiun:
As this was a misunderstanding from the docs, I created a PR for the docs: https://github.com/ghostty-org/website/pull/169

PR #3672, opened 2024-12-28
docs: Add info about adjustments to adjustment config options

pingiun:
Fix for misunderstanding in https://github.com/ghostty-org/ghostty/issues/3400

Adds info about the MetricModifier config options that they can be used with percentages, and links to the config option `adjust-cell-width` that has more info


mitchellh:
I removed the linking since that'll be awkward for our other generated docs. I think we should update the website source generator to find pre spans with config keys and auto link them.

Issue #3646, opened 2024-12-28
macOS Natural Text Editing: "command delete" should delete whole line, not one character


mitchellh:
Discussion: #3232 

Add: `keybind = cmd+backspace=esc:w` as a default for macOS. This is similar to #3106.

mitchellh:
Note: this may have to be `\x15`. Me or whoever does this should do some research to figure out the right input here.

datwaft:
`\x15` is working for me. https://github.com/datwaft/dotfiles/blob/16504b331bffaa6b56f65012132330c2124323b8/.config/ghostty/config#L12

mitchellh:
Fixed by #3679

PR #3679, opened 2024-12-28
Add default keybinding cmd+backspace=esc:w for macOS

korayfeyzula:
regarding https://github.com/ghostty-org/ghostty/issues/3646

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2557,6 +2557,11 @@ pub fn default(alloc_gpa: Allocator) Allocator.Error!Config {
         );
         try result.keybind.set.put(
             alloc,
+            .{ .key = .{ .translated = .backspace }, .mods = .{ .super = true } },
+            .{ .esc = "w" },
+        );
+        try result.keybind.set.put(
+            alloc,
             .{ .key = .{ .translated = .left }, .mods = .{ .alt = true } },
             .{ .esc = "b" },
         );
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2558,7 +2558,7 @@ pub fn default(alloc_gpa: Allocator) Allocator.Error!Config {
         try result.keybind.set.put(
             alloc,
             .{ .key = .{ .translated = .backspace }, .mods = .{ .super = true } },
-            .{ .esc = "w" },
+            .{ .esc = "\x15" },
         );
         try result.keybind.set.put(
             alloc,


mitchellh:
Looks good, I just want to study a bit more on which is correct.

Koray08:
`\x15` seems to be universally accepted across terminals

vr7bd:
Just updated to 1.0.1. and it doesn't seem to be working
I reverted back my config to `keybind = super+backspace=text:\x15`(working) which the code seems to be doing
Maybe something is missing?

mitchellh:
This may have to be escaped. There's a discrepancy here. It worked on my system but I'll look into it.

jinalskothari:
can confirm, does not clear line by default in 1.01 . suggested binging works.

lilyball:
in 1.0.1 for me in fish shell with vi keybindings ⌘⌫ both clears the current line and switches from input mode to normal mode. Fish is telling me that the ⌘⌫ sequence emits <kbd>Esc</kbd>, <kbd>Ctrl</kbd>+<kbd>U</kbd>.

`ghostty +show-config` tells me that the keybinding in effect here is

```
keybind = super+backspace=esc:^U
```

That said, I might suggest that Ghostty actually skip the keybinding here and instead implement `-deleteToBeginningOfLine:`. Ghostty passes text input through the Cocoa text input system, and if the text input client implements this selector then the text input system will invoke it in response to ⌘⌫. The reason for doing it this way is users who customize the text input system via DefaultKeyBindings.dict will then get expected behavior if they rebind this action. I might recommend this approach for any other similar standard macOS text editing actions that you're trying to set up keybinds for. By implementing the standard text input selector for the action instead you'll play nicely with users who rebind this, and you'll also know if you do the selector that you haven't missed any potential alternative keybinds that are supposed to do the same thing.

You can run `plutil -p /System/Library/Frameworks/AppKit.framework/Resources/StandardKeyBinding.dict` to see the standard set of Cocoa text input keybinds, this will tell you all the selectors they invoke, though the selectors should probably all be found in the [`NSStandardKeyBindingResponding`  protocol](https://developer.apple.com/documentation/appkit/nsstandardkeybindingresponding).

rrotter:
> in 1.0.1 for me in fish shell with vi keybindings ⌘⌫ both clears the current line and switches from input mode to normal mode. Fish is telling me that the ⌘⌫ sequence emits Esc, Ctrl+U.

That's the problem, the expected output is `^U`, not `esc:^U`. fish accepts either as equivalent, but `esc:^U` doesn't work on zsh, or rather, it would be an entirely different keyboarding that won't do anything unless something is mapped to it. `esc:^U` also does nothing on bash.

DrewLandgraveCbsi:
> keybind = super+backspace=text:\x15

This worked for me on mac 1.0.1 using zsh

seeker-3:
on version 1.0.1 for me. Out of the box

command+backspace did not work with zsh and bash

but it did work for

fish and cat

but putting this in my config `keybind = super+backspace=text:\x15`

makes it work for all 4.

I'm on sequoia 15.2

PR #3680, opened 2024-12-28
Use `source` in the example for sourcing Zsh shell integration

00-kat:
A few people were copying that snippet and were facing issues such as "permission denied", and (after a subsequent `chmod`) "syntax error".

-----
From the [CONTRIBUTING.md]:

> Pull requests should be associated with a previously accepted issue.

I decided to ignore this for a single-word change, since creating a discussion then waiting for it to be promoted to an issue seemed pointless for such a minor change.

[CONTRIBUTING.md]: https://github.com/ghostty-org/ghostty/blob/main/CONTRIBUTING.md#pull-requests-implement-an-issue

### DIFF
--- a/src/shell-integration/zsh/ghostty-integration
+++ b/src/shell-integration/zsh/ghostty-integration
@@ -25,7 +25,7 @@
 # Ghostty in all shells should add the following lines to their .zshrc:
 #
 #   if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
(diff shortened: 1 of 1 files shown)


00-kat:
This is actually a dupe of #3511 (I forgot to check closed PRs), I'll leave it up to someone else to decide which one to merge.

And no, this is not a dupe of #3332, that fixes a different file.

Issue #3648, opened 2024-12-28
Add `confirm-close-surface` option `always`


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3620

This would always show a confirmation dialog even if shell integration says a process isn't running.

```
confirm-close-surface = always
```

PR #3688, opened 2024-12-28
Allow `always` in `confirm-close-surface` config option

jay-dee7:
implements #3648


liby:
This PR is implemented quite well, why should it be closed?

jay-dee7:
Hey @liby, I just saw two other PRs doing the same thing after I added the PR and didn't want to add unnecessary overhead for a small fix like this one.

PR #3700, opened 2024-12-28
Add `always` in `confirm-close-surface` config option

### LINKED ISSUE #3648: Add `confirm-close-surface` option `always`

ErfanTech:
Fixes #3648


mitchellh:
Made some stylistic Zig changes but otherwise looks great. Thank you very much!

PR #3703, opened 2024-12-28
pass always config option to `confirm-close-surface`

### LINKED ISSUE #3648: Add `confirm-close-surface` option `always`

umegbewe:
closes #3648


umegbewe:
I just noticed that #3700  addresses this issue in a similar way to my submission. Since there's no significant difference between our approaches, I'll close this PR to reduce redundancy.

PR #3714, opened 2024-12-28
fix probable typo in Config.zig

apatterndarkly:
Based on both context and interaction with ghostty, "now" was intended to be "not" in the paragraph describing the behavior of `focus-follows-mouse` is set to `true`.


mitchellh:
Aye thanks

apatterndarkly:
My pleasure, truly.

PR #3727, opened 2024-12-28
Improve invalid keybind error messages

otherJL0:
Draft for [Better keybind error messages](https://github.com/ghostty-org/ghostty/discussions/3732) discussion


mitchellh:
The history here is a bit crazy and impossible to read. Can you try reopening a clean PR? Thanks

PR #3262, opened 2024-12-27
updated logic for grouping actions

salfel:
Update logic for generating `webgen_actions`. 

Before the change, the following grouping was produced:
```md
## `copy_to_clipboard`
Copy and paste.


## `paste_from_clipboard`
## `paste_from_selection`
## `increase_font_size`
Increase/decrease the font size by a certain amount.
```

After the change, the following grouping is being produced
```md
## `copy_to_clipboard`
## `paste_from_clipboard`
## `paste_from_selection`
Copy and paste.

## `increase_font_size`
## `decrease_font_size`
Increase/decrease the font size by a certain amount.
```

Please note that this is my first time ever writing zig, so forgive me violating zig best practices and feel free to make suggestions

Successor of [this](https://github.com/ghostty-org/website/pull/147) pr in the website repo

PR #3736, opened 2024-12-28
deinit buffer

salfel:
successor of #3262


mitchellh:
Yeah I noticed this in review but didn’t say anything because it’s not that important. This runs as a one shot program so quitting without freeing is totally fine. I’ll merge this though thanks!

PR #3742, opened 2024-12-28
feat: customize quick terminal size

### LINKED ISSUE #2384: macOS: Customize initial size for quick terminal

dmehala:
# Description

This commit introduce `quick-terminal-size` option which allows to define the size of the quick terminal.

It also fixes an issue where the quick terminal position was not properly updated when reloading the configuration.

Resolves #2384

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -8,7 +8,7 @@ class QuickTerminalController: BaseTerminalController {
     override var windowNibName: NSNib.Name? { "QuickTerminal" }
 
     /// The position for the quick terminal.
-    let position: QuickTerminalPosition
+    private var position: QuickTerminalPosition
 
     /// The current state of the quick terminal
     private(set) var visible: Bool = false
@@ -72,7 +72,7 @@ class QuickTerminalController: BaseTerminalController {
         syncAppearance(ghostty.config)
 
         // Setup our initial size based on our configured position
-        position.setLoaded(window)
+        derivedConfig.quickTerminalSize.apply(window, position)
 
         // Setup our content
         window.contentView = NSHostingView(rootView: TerminalView(
@@ -306,6 +306,9 @@ class QuickTerminalController: BaseTerminalController {
 
     private func syncAppearance(_ config: Ghostty.Config) {
         guard let window else { return }
+        
+        // Update the quick terminal size right away
+        config.quickTerminalSize.apply(window, config.quickTerminalPosition)
 
         // If our window is not visible, then delay this. This is possible specifically
         // during state restoration but probably in other scenarios as well. To delay,
@@ -390,7 +393,8 @@ class QuickTerminalController: BaseTerminalController {
 
         // Update our derived config
         self.derivedConfig = DerivedConfig(config)
-
+        self.position = self.derivedConfig.quickTerminalPosition
+        
         syncAppearance(config)
     }
 
@@ -398,17 +402,23 @@ class QuickTerminalController: BaseTerminalController {
         let quickTerminalScreen: QuickTerminalScreen
         let quickTerminalAnimationDuration: Double
         let quickTerminalAutoHide: Bool
+        let quickTerminalPosition: QuickTerminalPosition
+        let quickTerminalSize: QuickTerminalSize
 
         init() {
             self.quickTerminalScreen = .main
             self.quickTerminalAnimationDuration = 0.2
             self.quickTerminalAutoHide = true
+            self.quickTerminalPosition = .top
+            self.quickTerminalSize = .init()
         }
 
         init(_ config: Ghostty.Config) {
             self.quickTerminalScreen = config.quickTerminalScreen
             self.quickTerminalAnimationDuration = config.quickTerminalAnimationDuration
             self.quickTerminalAutoHide = config.quickTerminalAutoHide
+            self.quickTerminalPosition = config.quickTerminalPosition
+            self.quickTerminalSize = config.quickTerminalSize
         }
     }
 }
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalSize.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalSize.swift
@@ -1,0 +1,81 @@
+import Cocoa
+import GhosttyKit
+
+class QuickTerminalSize {
+    enum Size {
+        case percent(value: Double)
+        case pixel(value: UInt)
+        
+        init?(c_dimension: ghostty_config_quick_terminal_dimension_s) {
+            switch(c_dimension.unit) {
+            case GHOSTTY_QUICK_TERMINAL_PIXEL_UNIT:
+                self = .pixel(value: UInt(c_dimension.value))
+            case GHOSTTY_QUICK_TERMINAL_PERCENTAGE_UNIT:
+                self = .percent(value: Double(c_dimension.value) / 100.0)
+            default:
+                return nil
+            }
+        }
+        
+        func apply(value: CGFloat) -> CGFloat {
+            switch(self) {
+            case .pixel(let fixed_size):
+                return CGFloat(fixed_size);
+            case .percent(let pct):
+                return value * pct;
+            }
+        }
+    }
+
+    var mainDimension: Size;
+    var secondDimension: Size;
+
+    init() {
+        self.mainDimension = Size.percent(value: 0.25)
+        self.secondDimension = Size.percent(value: 0.25)
+    }
+
+    init(config: ghostty_config_quick_terminal_size_s) {
+        switch (config.len) {
+        case 1:
+            self.mainDimension = Size(c_dimension: config.dimensions[0]) ?? Size.percent(value: 0.25)
+            self.secondDimension = Size.percent(value: 0.25)
+        case 2:
+            self.mainDimension = Size(c_dimension: config.dimensions[0]) ?? Size.percent(value: 0.25)
+            self.secondDimension = Size(c_dimension: config.dimensions[1]) ?? Size.percent(value: 0.25)
+        default:
+            self.mainDimension = Size.percent(value: 0.25)
+            self.secondDimension = Size.percent(value: 0.25)
+        }
+    }
+    
+    /// Set the window size.
+    func apply(_ window: NSWindow, _ position: QuickTerminalPosition) {
+        guard let screen = window.screen ?? NSScreen.main else { return }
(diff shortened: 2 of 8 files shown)


src/config/Config.zig
```diff
@@ -1411,6 +1411,31 @@ keybind: Keybinds = .{},
...
+///  * Two comma separated value specifies the width and height.
+///
+/// Examples:
+///
+/// ```
+/// quick-terminal-size = 25%       // 25% of the maximum size for the growable dimension
+/// quick-terminal-size = 42px      // 42 pixels for the growable dimension
+/// quick-terminal-size = 25%,75%   // 25% for the primary dimension, 75% for the secondary
+/// quick-terminal-size = 300px,80% // 300px for the primary dimension, 80% for the secondary
+/// ```
```
dmehala, line 1436: I am not a native English speaker, so feel free reword 😅.

macos/Sources/Features/QuickTerminal/QuickTerminalSize.swift
```diff
@@ -0,0 +1,81 @@
...
+                self = .percent(value: Double(c_dimension.value) / 100.0)
+            default:
+                return nil
+            }
+        }
+        
+        func apply(value: CGFloat) -> CGFloat {
+            switch(self) {
+            case .pixel(let fixed_size):
+                return CGFloat(fixed_size);
```
dmehala, line 23:
```suggestion
                return min(value, CGFloat(fixed_size));
```

src/config/Config.zig
```diff
@@ -5294,6 +5319,240 @@ pub const QuickTerminalScreen = enum {
...
+                        },
+                        .pixel => |v| {
+                            break :res @intCast(v);
+                        },
+                    }
+                },
+            };
+        }
+
+        fn parseValue(input: []const u8) !Size {
```
dmehala, line 5360: [Nitpick] Parsing could be improve, `quick-terminal-size: 20%, 30%` will probably fail due to the space after the comma. I could/should fix it, but other config suffers from the same issue and should probably be fixed in a subsequent PR.

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -306,6 +306,9 @@ class QuickTerminalController: BaseTerminalController {
 
     private func syncAppearance(_ config: Ghostty.Config) {
         guard let window else { return }
+        
+        // Update the quick terminal size right away
+        config.quickTerminalSize.apply(window, config.quickTerminalPosition)
```
dmehala, line 311: [Suggestion] Update the window size here because the call is delayed until the quick terminal becomes visible, which is too late. Not ideal, since this function is called repeatedly until it can be consumed. How about synchronizing the appearance when `toggle` is called?

macos/Sources/Features/QuickTerminal/QuickTerminalSize.swift
```diff
@@ -0,0 +1,81 @@
...
+
+class QuickTerminalSize {
+    enum Size {
+        case percent(value: Double)
+        case pixel(value: UInt)
+        
+        init?(c_dimension: ghostty_config_quick_terminal_dimension_s) {
+            switch(c_dimension.unit) {
+            case GHOSTTY_QUICK_TERMINAL_PIXEL_UNIT:
+                self = .pixel(value: UInt(c_dimension.value))
```
dmehala, line 12: [TODO] Use screen scale (IIRC retina has a scale of 2-3 while standard disply is 1)

Jaycedam:
Hi I've been trying this out, thank you for working on this.
I have some issues but I could be doing something wrong, when using:
```
quick-terminal-size = 80%
```
only the width is modified

image

using two values works as expected 
`quick-terminal-size = 80%,80%`

image

Also the [global keybind](https://ghostty.org/docs/config/keybind/reference#toggle_quick_terminal) stopped working but maybe that's a different issue?

dmehala:
Hey @Jaycedam 

> I have some issues but I could be doing something wrong, when using:
> quick-terminal-position = center
quick-terminal-size = 80%
only the width is modified

This is the expected behaviour, howver, I agree for `center` position it seems natural to expect the size to change on both dimensions unless a second dimension is specified. I am planning to push this change:

Screen Recording 2024-12-29 at 23 36 12

**NOTE**: I have an ultrawide monitor.

> Also the [global keybind](https://ghostty.org/docs/config/keybind/reference#toggle_quick_terminal) stopped working but maybe that's a different issue?

The PR doesn't include any changes related to global keybindings, so that would be a surprising side effect. Unfortunately, I am not able to reproduce the issue. Could you share the steps to reproduce it with you configuration? Thank you.

Jaycedam:
> This is the expected behaviour, howver, I agree for center position it seems natural to expect the size to change on both dimensions

I was confused by the wording of the documentation added here, but as long as it reflects the usage I think it will be easy to understand.

> The configuration accept one or two dimensions:
> * A single value specifies the dimension that is growable based on the quick terminal position:
>     * For `center`, the size applied to both the width and height.

The global keybind was a problem on my end so ignore it. Re-enabling the accesibility toggle fixed it.

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
dmehala: addressed in #4055

YANOUSHek:
I know I'm a little late to the party but I just started using ghostty today. Shouldn't we also allow to specify the size of the quick terminal in rows and cols? So
```
quick-terminal-size = 120,50
```
would create a window that has 120 columns and 50 rows.

Dolvur:
> I know I'm a little late to the party but I just started using ghostty today. Shouldn't we also allow to specify the size of the quick terminal in rows and cols? So
> 
> ```
> quick-terminal-size = 120,50
> ```
> 
> would create a window that has 120 columns and 50 rows.

This was mentioned as a future enhancement due to it being more complex https://github.com/ghostty-org/ghostty/issues/2384#issuecomment-2401129283

Also thank you dmehala for working on this. I always use a fullscreen hotkey toggle for my terminal needs

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -390,25 +393,32 @@ class QuickTerminalController: BaseTerminalController {
 
         // Update our derived config
         self.derivedConfig = DerivedConfig(config)
-
+        self.position = self.derivedConfig.quickTerminalPosition
+        
         syncAppearance(config)
```
donaldguy, line 398:
```suggestion
        syncAppearance()
```

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -390,25 +393,32 @@ class QuickTerminalController: BaseTerminalController {
...
     }
 
     private struct DerivedConfig {
         let quickTerminalScreen: QuickTerminalScreen
         let quickTerminalAnimationDuration: Double
         let quickTerminalAutoHide: Bool
+        let quickTerminalPosition: QuickTerminalPosition
+        let quickTerminalSize: QuickTerminalSize
```
donaldguy, line 406:
```suggestion
        let quickTerminalPosition: QuickTerminalPosition
        let quickTerminalSize: QuickTerminalSize
        let windowColorspace: String
        let backgroundOpacity: Double
```

or your preferred order

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -390,25 +393,32 @@ class QuickTerminalController: BaseTerminalController {
...
+            self.quickTerminalPosition = .top
+            self.quickTerminalSize = .init()
         }
 
         init(_ config: Ghostty.Config) {
             self.quickTerminalScreen = config.quickTerminalScreen
             self.quickTerminalAnimationDuration = config.quickTerminalAnimationDuration
             self.quickTerminalAutoHide = config.quickTerminalAutoHide
+            self.quickTerminalPosition = config.quickTerminalPosition
+            self.quickTerminalSize = config.quickTerminalSize
```
donaldguy, line 421:
```suggestion
            self.quickTerminalPosition = config.quickTerminalPosition
            self.quickTerminalSize = config.quickTerminalSize
            self.windowColorspace = config.windowColorspace
            self.backgroundOpacity = config.backgroundOpacity
```

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -390,25 +393,32 @@ class QuickTerminalController: BaseTerminalController {
...
         let quickTerminalAutoHide: Bool
+        let quickTerminalPosition: QuickTerminalPosition
+        let quickTerminalSize: QuickTerminalSize
 
         init() {
             self.quickTerminalScreen = .main
             self.quickTerminalAnimationDuration = 0.2
             self.quickTerminalAutoHide = true
+            self.quickTerminalPosition = .top
+            self.quickTerminalSize = .init()
```
donaldguy, line 413:
```suggestion
            self.quickTerminalPosition = .top
            self.quickTerminalSize = .init()
            self.windowColorspace = ""
            self.backgroundOpacity = 1.0
```

donaldguy (commented):
locally rebased this onto tip (at 8f5f432ab61fa73b77cb61239c0cff34250f77bb)

there were some detected-conflicts in `QuickTerminalController.swift` (vs #4055 )

they are all fairly obvious resolutions but to (potentially) save you the trouble (modulo if you have strong feelings about a different order)

[see below for one ~substantive question]

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
donaldguy: FYI, since you removed the function param this symbol no longer resolves and tanks the swift compilation. 

not sure if the `derivedConfig.` invocation up at 75 (72 in my rebased version) is adequate or if this wants to stay but go:

```suggestion
        // Update the quick terminal size right away
        self.derivedConfig.quickTerminalSize.apply(window, self.derivedConfig.quickTerminalPosition)
```

or something else

donaldguy, line 311: FYI, since you removed the function param this symbol no longer resolves and tanks the swift compilation. 

not sure if the  invocation up at 75 (72 in my rebased version) is adequate or if this wants to stay but go:

```suggestion
        
        // Update the quick terminal size right away
        self.derivedConfig.quickTerminalSize.apply(window, self.position)
```

or something else

PR #3746, opened 2024-12-28
GTK: add delay before updating the title

### LINKED ISSUE #2503: Change title after a small delay to avoid flicker
First of all this is not really an issue, I think it is the desried behaviour, but I found it distracting.

When I execute a fast command like `ls` or `pwd`, the title bar text changes from the `directory name -> command(ls) -> directory name`. The problem is Ghostty is fast enough that this text transition is just a flicker, I cannot see which command I executed looking at the title bar and tbh it's not important because I just executed.

For long running commands like `htop`, `watch`... it's interesting to see the command in the title bar.

My suggestion is to wait a few milliseconds to change the title bar, so fast commands don't change it.
I don't know how hard is to make this change or if other found it distracting too, it's just a suggestion...

https://github.com/user-attachments/assets/12553c37-c6b4-4c3b-bf2f-27e426b6a61b

b6k-dev:
This pr fixes https://github.com/ghostty-org/ghostty/issues/2503 for GTK. 
The implementation is quite similar to what was done in https://github.com/ghostty-org/ghostty/pull/2929 for the macOS version.

Before (i was able to reproduce the issue by just invoking `ls`):

https://github.com/user-attachments/assets/011acb1d-de71-46a1-8a14-45e8eb932183

After:

https://github.com/user-attachments/assets/b749cd1c-355e-47de-a976-62d98a56f966

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -650,6 +650,7 @@ pub fn deinit(self: *Surface) void {
     // and therefore the unfocused_overlay has been destroyed as well.
     c.g_object_unref(self.im_context);
     if (self.cursor) |cursor| c.g_object_unref(cursor);
+    if (self.update_title_timer) |timer| _ = c.g_source_remove(timer);
     self.resize_overlay.deinit();
 }
 
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -346,6 +346,9 @@ cursor: ?*c.GdkCursor = null,
 /// pass it to GTK.
 title_text: ?[:0]const u8 = null,
 
+/// The timer used to delay title updates in order to prevent flickering.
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/Surface.zig
```diff
@@ -346,6 +346,9 @@ cursor: ?*c.GdkCursor = null,
 /// pass it to GTK.
 title_text: ?[:0]const u8 = null,
 
+/// The timer used to delay title updates in order to prevent flickering.
+update_title_timer: ?c.guint = null,
```
mitchellh, line 350: This probably needs to be invalidated if the surface is closed.

mitchellh (changes requested):

src/apprt/gtk/Surface.zig
maciekbartczak: Added the invalidation in `deinit` function.

Issue #3477, opened 2024-12-27
build: support building with GTK compiled without x11-backend


ldv:
**Build environment**: Fedora 41, _gtk4_ package compiled with _x11-backend_ option disabled.

I tried to build the package with:
_zig build -fsys=simdutf -Dtarget=native  -Doptimize=ReleaseFast -Dpie_

but it fails with the following errors:

```
install
└─ install ghostty
   └─ zig build-exe ghostty ReleaseFast native 2 errors
src/apprt/gtk/c.zig:2:15: error: C import failed
pub const c = @cImport({
              ^~~~~~~~
referenced by:
    gtk: src/cli/version.zig:8:80
    run: src/cli/version.zig:42:13
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
/home/ldv/Downloads/ghostty-1.0.0/.zig-cache/o/cfe3a1b5077a346ccf90ce9c54a404e4/cimport.h:3:10: error: 'gdk/x11/gdkx.h' file not found
#include <gdk/x11/gdkx.h>
```

Is there a plan to support "pure" Wayland GTK backend? Or am I missing something?

mitchellh:
This file is required for compilation but we don't use it at runtime if Wayland is used. We could introduce a build option to not complete the X11 support at all but I'm not sure if it's worth it... still I've tagged this to look into it.

sin-ack:
We also hit this issue in Gentoo packaging, as GTK 4 can be built without X. https://bugs.gentoo.org/947032

mitchellh:
Thanks for the additional data point. I'd like to fix this.

tristan957:
I tried fixing this one time, but I don't understand Zig's C support enough to fix it. If this was a C project, I would have it done, but I'm still new to Zig. Very interested to see what solutions can be come up with!

mitchellh:
> I tried fixing this one time, but I don't understand Zig's C support enough to fix it.

We need to introduce a build option that we then use to avoid ever referencing the `cImport` decl. Since Zig tree shakes, this will avoid trying to load and compile it.

mitchellh:
This should be fixed on main with #3748. I'd appreciate any confirmation, it should happen automatically now.

ldv:
I tried to build from [master (c8950d3):](https://github.com/ghostty-org/ghostty/commit/c8950d376ab3be13fbd9a19317163075f4feddbc):

[...]

PR #3748, opened 2024-12-28
gtk: add option to not link against libX11

jcollie:
Possible fix for #3477. Needs testing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -342,6 +342,7 @@ jobs:
       matrix:
         adwaita: ["true", "false"]
         x11: ["true", "false"]
+    name: GTK adwaita=${{ matrix.adwaita }} x11=${{ matrix.x11 }}
     runs-on: namespace-profile-ghostty-sm
     needs: test
     env:
@@ -357,6 +358,15 @@ jobs:
           path: |
             /nix
             /zig
+
+      # Install Nix and use that to run our tests so our environment matches exactly.
+      - uses: cachix/install-nix-action@v30
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - uses: cachix/cachix-action@v15
+        with:
+          name: ghostty
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
       - name: Test GTK Build
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -375,7 +375,6 @@ jobs:
           name: ghostty
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
-
   test-macos:
     runs-on: namespace-profile-ghostty-macos
     needs: test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -329,24 +329,52 @@ jobs:
       - name: Test GTK Build
         run: nix develop -c zig build -Dapp-runtime=gtk -Dgtk-adwaita=true -Demit-docs
 
-      - name: Test GTK Build (No Libadwaita)
-        run: nix develop -c zig build -Dapp-runtime=gtk -Dgtk-adwaita=false -Demit-docs
-
-      - name: Test GTK build with explicit X11
-        run: nix develop -c zig build -Dapp-runtime=gtk -Dgtk-adwaita=true -Demit-docs -Dgtk-x11=true
-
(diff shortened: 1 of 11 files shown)


tristan957:
Does the zig build system have a notion of Meson's feature options where you can detect if GTK was built without X11, and then build ghostty without X11 support too?

build.zig
```diff
@@ -105,6 +105,12 @@ pub fn build(b: *std.Build) !void {
         "Enables the use of Adwaita when using the GTK rendering backend.",
     ) orelse true;
 
+    config.x11 = b.option(
+        bool,
+        "gtk-x11",
+        "Enables the linking against X11 libraries when using the GTK rendering backend.",
```
tristan957, line 111:
```suggestion
        "Enables linking against X11 libraries when using the GTK rendering backend.",
```

src/apprt/gtk/ghostty_gtk_compat.h
```diff
@@ -12,3 +12,11 @@
...
 #define G_APPLICATION_DEFAULT_FLAGS G_APPLICATION_FLAGS_NONE
 #endif
+
+#ifndef False
+#define False 0
+#endif
+
+#ifndef True
+#define True 1
+#endif
```
tristan957, line 22: Why is this necessary?

jcollie: Somewhere in `<gdk/x11/gdkx.h>` or `<X11/XKBlib.h>` C macros for `True` and `False` get defined and they are used in a couple of places in `gtk/Window.zig` and `gtk/Surface.zig`. I suppose it would be just as easy to replace the uses in the Zig code.

jcollie:
> Does the zig build system have a notion of Meson's feature options where you can detect if GTK was built without X11, and then build ghostty without X11 support too? Essentially, just automatically do the right thing.

If there's a command that we can run to determine if GTK was built without X11, we should be able to take advantage of that in `build.zig`. I'm not aware of a native feature in Zig that does that.

src/apprt/gtk/ghostty_gtk_compat.h
tristan957: Yeah I would recommend that

tristan957:
```
$ pkg-config --variable=targets gtk4
broadway wayland x11
```

tristan957:
`pkgconf` is also another tool for querying, so you may want to check both. Not sure if `pkg-config` is guaranteed to exist.

mitchellh:
Ghostty already requires pkg-config to build so it’s fine to rely on it.

jcollie:
> `pkgconf` is also another tool for querying, so you may want to check both. Not sure if `pkg-config` is guaranteed to exist.

Pretty sure that the Zig build system requires `pkg-config` to exist to find system libraries, so I think that we can depend on it being available.

tristan957:
Even better!

jcollie:
> ```
> $ pkg-config --variable=targets gtk4
> broadway wayland x11
> ```

That should be easy enough to check for.

jcollie:
OK, the latest commit _should_ detect X11 support in GTK4 but I don't have a system handy that has GTK4 without X11 support. I _thought_ it would be possible to set that up in the devshell but I wasn't able to get that to work.

So if someone has a system without X11 installed it would be great if they can test it out.

mitchellh:
This looks good to me. I want to make a couple stylistic tweaks but this is well done!

mitchellh:
Moved the GTK tests to a matrix. If that's good will merge.

.github/workflows/test.yml
```diff
@@ -329,25 +329,53 @@ jobs:
...
       - name: Test System Build
         run: nix develop -c zig build --system ${ZIG_GLOBAL_CACHE_DIR}/p
 
+  test-gtk:
+    strategy:
+      fail-fast: false
+      matrix:
+        adwaita: ["true", "false"]
+        x11: ["true", "false"]
+    runs-on: namespace-profile-ghostty-md
```
tristan957: I would suggest using enum like strings. Otherwise you won't know which job failed because it will look like `test-gtk(true, false)` or something

mitchellh: I just named them using the `name:` key. Is there some other way? That's what I'm used to doing.

tristan957: It is just a UX problem in my opinion of GitHub that have to be worked around. No biggie

PR #3806, opened 2024-12-28
apprt: add window-titlebar-{background,foreground} config options

tristan957:
This gives people finer-grained control over the coloring of their titlebars. Currently only implemented for GTK.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1149,6 +1149,16 @@ keybind: Keybinds = .{},
 ///   * `end` - Insert the new tab at the end of the tab list.
 @"window-new-tab-position": WindowNewTabPosition = .current,
 
+/// Background color for the window titlebar. This only takes effect if
+/// window-theme is set to ghostty. Currently only supported in the GTK app
+/// runtime.
+@"window-titlebar-background": ?Color = null,
+
+/// Foreground color for the window titlebar. This only takes effect if
+/// window-theme is set to ghostty. Currently only supported in the GTK app
+/// runtime.
+@"window-titlebar-foreground": ?Color = null,
+
 /// This controls when resize overlays are shown. Resize overlays are a
 /// transient popup that shows the size of the terminal while the surfaces are
 /// being resized. The possible options are:
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -966,8 +966,8 @@ fn loadRuntimeCss(
     const config: *const Config = &self.config;
     const window_theme = config.@"window-theme";
     const unfocused_fill: Config.Color = config.@"unfocused-split-fill" orelse config.background;
-    const headerbar_background = config.background;
-    const headerbar_foreground = config.foreground;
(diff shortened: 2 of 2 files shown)


src/config/Config.zig
```diff
@@ -400,9 +400,15 @@ theme: ?Theme = null,
 /// Background color for the window.
 background: Color = .{ .r = 0x28, .g = 0x2C, .b = 0x34 },
 
+/// Background color for the titlebar.
+@"background-titlebar": ?Color = null,
+
```
jcollie, line 405: `background-titlebar` => `titlebar-background`

Precedent: `selection-background`

src/config/Config.zig
```diff
@@ -400,9 +400,15 @@ theme: ?Theme = null,
...
 /// Foreground color for the window.
 foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 
+/// Background color for the titlebar.
+@"foreground-titlebar": ?Color = null,
+
```
jcollie, line 411: `foreground-titlebar` => `titlebar-foreground`

Precedent: `selection-foreground`

src/config/Config.zig
jcollie: Probably also need to note that this only takes effect if `window-theme=ghostty`.

jcollie (changes requested):
LGTM other than the config variable names.

tristan957:
Thanks for your suggestions @jcollie. They make sense to me.

jcollie (approved):

src/config/Config.zig
```diff
@@ -400,9 +400,17 @@ theme: ?Theme = null,
 /// Background color for the window.
 background: Color = .{ .r = 0x28, .g = 0x2C, .b = 0x34 },
 
+/// Background color for the titlebar. This is only takes effect if
+/// window-theme is set to ghostty.
+@"titlebar-background": ?Color = null,
```
mitchellh, line 405: I think to fit in with the rest of our configs this should be `window-titlebar-background` (and likewise for foreground). And they should be down in the `window-`-prefixed section.

src/config/Config.zig
```diff
@@ -1149,6 +1149,14 @@ keybind: Keybinds = .{},
 ///   * `end` - Insert the new tab at the end of the tab list.
 @"window-new-tab-position": WindowNewTabPosition = .current,
 
+/// Background color for the window titlebar. This only takes effect if
+/// window-theme is set to ghostty.
+@"window-titlebar-background": ?Color = null,
+
+/// Foreground color for the window titlebar. This only takes effect if
+/// window-theme is set to ghostty.
+@"window-titlebar-foreground": ?Color = null,
```
mitchellh, line 1158: Easy change but we should note for each of this that this configuration only applies to GTK currently.

tristan957: I'll go ahead and add that. Was hoping I could snipe someone into adding the feature

mitchellh: It's non-trivial for macOS since we have a variety of titlebar styles that all need to be implemented separately I think.

PR #3808, opened 2024-12-28
Clarify how to read logs on macOS

nalzok:
I removed the sentence `See the Mac App section for more information.` because I cannot find the relevant information. It would be great to have a separate log file and indicate its location in the document.

For the record, I just spent 15 minute searching for the location of a log file in the document, which is painful because the website does not have a search button. Eventually I gave up and looked into the source file, and finally found the following comment (btw do we really need `sudo`?):

https://github.com/ghostty-org/ghostty/blob/ca7c9547129cb156417f7aa50cb390edecb685d7/src/main_ghostty.zig#L129-L132

Hopefully, this commit can make other new users' lives easier!


mitchellh:
Yes, great, thank you!

PR #3811, opened 2024-12-28
feat: expand tildes `~` in config file paths to `HOME`

### LINKED ISSUE #3328: File paths in config file should expand tilde (`~`) to HOME

z-jxy:
Fixes https://github.com/ghostty-org/ghostty/issues/3328

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4364,33 +4364,24 @@ pub const RepeatablePath = struct {
 
             // Check if the path starts with a tilde and expand it to the home directory on linux/mac
             if (std.mem.startsWith(u8, path, "~/")) {
-                const home_var = try internal_os.home(&buf); // cache this?
-                if (home_var) |home_dir| {
-                    const rest = path[1..]; // Skip the ~
-                    const expanded_len = home_dir.len + rest.len;
-                    if (expanded_len > buf.len) {
-                        try diags.append(alloc, .{
-                            .message = try std.fmt.allocPrintZ(
-                                alloc,
-                                "error resolving file path {s}: path too long after expanding home directory",
-                                .{path},
-                            ),
-                        });
-                        self.value.items[i] = .{ .required = "" };
-                        continue;
-                    }
-
-                    @memcpy(buf[home_dir.len..expanded_len], rest);
-
+                if (try internal_os.expandHome(path, &buf)) |expanded_path| {
                     log.debug(
                         "expanding file path from home directory: path={s}",
-                        .{buf[0..expanded_len]},
+                        .{expanded_path},
                     );
-
                     switch (self.value.items[i]) {
-                        .optional, .required => |*p| p.* = try alloc.dupeZ(u8, buf[0..expanded_len]),
+                        .optional, .required => |*p| p.* = try alloc.dupeZ(u8, expanded_path),
                     }
-
+                    continue;
+                } else {
+                    try diags.append(alloc, .{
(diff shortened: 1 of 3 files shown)


src/config/Config.zig
```diff
@@ -4256,6 +4256,33 @@ pub const RepeatablePath = struct {
             var buf: [std.fs.max_path_bytes]u8 = undefined;
             const abs = dir.realpath(path, &buf) catch |err| abs: {
                 if (err == error.FileNotFound) {
+                    // Check if the path starts with a tilde and expand it to the home directory on linux/mac
+                    if (path[0] == '~') {
```
jcollie: This test needs to be a little more sophisticated, it should see if `path[1]` is the path separator because `~abc/` is the home directory of user `abc`. Only `~/` is the home directory of the current user.

src/config/Config.zig
```diff
@@ -4256,6 +4256,33 @@ pub const RepeatablePath = struct {
...
+                        const home_env_var = switch (builtin.os.tag) {
+                            .linux, .macos => std.posix.getenv("HOME"),
+                            .windows => null,
+                            else => null,
+                        };
+                        if (home_env_var) |home_dir| {
```
jcollie: Instead of directly reading the `HOME` env var, there's a function called `home` in `src/os/homedir.zig` that should be re-used. It works on more platforms and also handles cases where `HOME` is not set by checking in other platform-specific places for a user's home directory.

src/config/Config.zig
```diff
@@ -4254,6 +4254,59 @@ pub const RepeatablePath = struct {
...
             // to the base.
             var buf: [std.fs.max_path_bytes]u8 = undefined;
+
+            // Check if the path starts with a tilde and expand it to the home directory on linux/mac
+            if (path[0] == '~') {
+                const home_env_var = switch (builtin.os.tag) {
+                    .linux, .macos => std.posix.getenv("HOME"),
+                    .windows => null,
+                    else => null,
+                };
```
jparise, line 4264: We should use `home` (from `os/homedir`) for this. It provides a more complete implementation.

I think this new code could also be generalized as an `expandHome` or similar function in that module, which would make it easier to reuse and unit test.

jcollie (changes requested):

src/config/Config.zig
z-jxy: Updated this to now check for `~/`

src/config/Config.zig
z-jxy: Thanks, didn't know that function was there. I updated the implementation to use it and moved the code into `os/homedir` as `expandHomeUnix()`

src/os/homedir.zig
```diff
@@ -110,6 +110,31 @@ fn trimSpace(input: []const u8) []const u8 {
...
+    if (!std.mem.startsWith(u8, path, "~/")) return null;
+    const home_dir = try home(buf) orelse return null;
+    const rest = path[1..]; // Skip the ~
+    const expanded_len = home_dir.len + rest.len;
+
+    if (expanded_len > buf.len) return Error.BufferTooSmall;
+    @memcpy(buf[home_dir.len..expanded_len], rest);
+
+    return buf[0..expanded_len];
+}
```
tristan957, line 136: I think we could add a test for these functions.

z-jxy: Added a test for the unix implementation with a few cases

mitchellh:
Made some stylistic tweaks, auto-merge on.

PR #3814, opened 2024-12-28
fix: quick terminal `focus-follows-mouse` behaviour

### LINKED ISSUE #3337: Quick Terminal - no 'focus-follows-mouse' functionality
Ghostty Version 1.0.0
Build 8469
Commit 4b4d4062

(MacOS)

I noticed when using splits in the Quick Terminal, `focus-follows-mouse = true` does not have an effect in the Quick Terminal. This may be working as intended though, and having focus follow inside splits inside the Quick Terminal is a pretty niche use case tbh.

dmehala:
# Description

Quick Terminal now focuses on the surface under the mouse pointer when `focus-follows-mouse` is enabled.

Fixes #3337

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -45,6 +45,11 @@ class BaseTerminalController: NSWindowController,
         didSet { surfaceTreeDidChange(from: oldValue, to: surfaceTree) }
     }
 
+    /// Whether the terminal surface should focus when the mouse is over it.
+    var focusFollowsMouse: Bool {
+        self.derivedConfig.focusFollowsMouse
+    }
+
     /// Non-nil when an alert is active so we don't overlap multiple.
     private var alert: NSAlert? = nil
 
@@ -262,7 +267,6 @@ class BaseTerminalController: NSWindowController,
         
         // Set the main window title
         window.title = to
-
     }
     
     func pwdDidChange(to: URL?) {
@@ -604,15 +608,18 @@ class BaseTerminalController: NSWindowController,
     private struct DerivedConfig {
         let macosTitlebarProxyIcon: Ghostty.MacOSTitlebarProxyIcon
         let windowStepResize: Bool
+        let focusFollowsMouse: Bool
 
         init() {
             self.macosTitlebarProxyIcon = .visible
             self.windowStepResize = false
+            self.focusFollowsMouse = false
         }
 
         init(_ config: Ghostty.Config) {
             self.macosTitlebarProxyIcon = config.macosTitlebarProxyIcon
             self.windowStepResize = config.windowStepResize
+            self.focusFollowsMouse = config.focusFollowsMouse
         }
     }
 }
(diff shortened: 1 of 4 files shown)


macos/Sources/Features/QuickTerminal/QuickTerminalWindow.swift
```diff
@@ -1,6 +1,8 @@
 import Cocoa
 
 class QuickTerminalWindow: NSWindow {
+    var focusFollowsMouse: Bool = false
```
mitchellh, line 4: I think we should add this to `BaseTerminalController` and use that instead of a window property. That would simplify everything else in SurfaceView if we introduce new terminal types.

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -617,12 +617,19 @@ extension Ghostty {
...
+                }
+            }()
+            
             // If focus follows mouse is enabled then move focus to this surface.
-            if let window = self.window as? TerminalWindow,
-               window.isKeyWindow &&
-                window.focusFollowsMouse &&
-                !self.focused
-            {
+            if (self.window!.isKeyWindow && !self.focused && focusFollowsMouse) {
```
mitchellh, line 632: I'd prefer to guard against window being null here. It is possible (we detach views at certain times).

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -67,6 +67,9 @@ class QuickTerminalController: BaseTerminalController {
         // The quick window is not restorable (yet!). "Yet" because in theory we can
         // make this restorable, but it isn't currently implemented.
         window.isRestorable = false
+        
+        // Propagate `focus-follows-mouse` options to the quick terminal.
+        window.focusFollowsMouse = ghostty.config.focusFollowsMouse
```
mitchellh, line 72: BaseTerminalController has a windowDidLoad you can add this to. It might make sense to add a function there to call on config change too (notification is already there too).

mitchellh (changes requested):

mitchellh:
Appreciate this. 😄 Suggested some changes.

dmehala:
I completely reverted https://github.com/dmehala/ghostty/commit/0ec402f926c66b2c94e217c18a30f55e36a5fe02 and addressed your comments. Now `BaseTerminalController` have a new `followFocusMouse` property which is used by `SurfaceView`. I'm not a Swift or AppKit expert, so this might not be the best approach 😅, but please feel free to take over if that works better for you.

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -62,6 +62,8 @@ class BaseTerminalController: NSWindowController,
 
     /// The configuration derived from the Ghostty config so we don't need to rely on references.
     private var derivedConfig: DerivedConfig
+    
+    private(set) var followFocusMouse: Bool = false
```
dmehala, line 66:
```suggestion
    /// Keep track of whether the controller should focus on the pane currently being hovered.
    private(set) var followFocusMouse: Bool = false
```

mitchellh:
Works great. Made some stylistic changes feel free to take a look. Will merge once green. Thanks

mitchellh (approved):

Issue #1649, opened 2024-04-04
Keybinding to reopen last closed tab 


pnodet:
Nice to have enhancement: being able to reopen the last closed tab (in the same directory).

It's something I use on web browser and sometimes I happen to close a tab quickly, I'd love to be able to reopen it as quickly.

nskazki:
Hello, Mitchell! Appreciate the support for "quick terminal" and jump_to_prompt! The only other feature I need to migrate from Iterm2 is this. Is this something you plan on doing in the coming year, or is the feature a low priority?

xwillq:
+1 for this feature.

To add some info, when you undo, iTerm doesn't simply create new session.
* If a pane was closed, it will be restored in the same tab at the same position and it will keep it's configuration, e.g. font size, background color, size, etc.
* If a tab was closed, it will be restored in the same window at the same position in the tab bar. All panes inside that tab will be restored, keeping their configurations.
* If a window was closed, it will be restored at the same position and with the same dimensions. All tabs inside that window will be restored, keeping their configurations.
* When you close a session, iTerm doesn't kill process running inside of it. So, when you restore a session, you will keep all of the state. Your editor will be running, your ssh connection will be intact. This is the main reason why iTerm has a time limit for session restoration - it has to kill these processes at some point.
* When you restore a session, it will keep it's scrollback buffer. This is a second reason for time limit.

As to how this feature should work in Ghostty, I don't think that simply opening a new surface in the same working directory as previously closed one is very useful. Sure, it can be convenient if you don't want to `cd` again, but you can solve this problem using shell (I believe fish allows you to type `cd` and hit up to bring last cd from history, also there is a plugin for zsh that does the same). 
For me, the most important thing with this feature is that you can restore scrollback buffer. What if I set up some service that randomly generates a password for me and prints it to the terminal and I accidentally closed the tab? Or run some destructive command that outputs some info? In this cases, I don't want to or can't redo what I did to get the same info, so restoring scrollback buffer is very important.
Process restoration seems less useful to me, because confirmation for closing surfaces mostly alleviates the problem of accidentally closing a surface with some important process running.

mitchellh:
Due to the amount of apprt-specific metadata required (tab vs split vs window, positions, etc.), I'm going to adjust my previous implementation recommendations and say this should be implemented totally custom at the aport-level. 

For macOS, it may be desirable to integrate this directly with [UndoManager](https://developer.apple.com/documentation/foundation/undomanager).

[...]

PR #3822, opened 2024-12-29
feat: add keybinding to reopen last closed tab

pnodet:
I am NOT familiar with either Zig or Swift. This is a rough implementation and it might be wrong on many levels.

#1649

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -557,6 +557,10 @@ typedef struct {
   bool soft;
 } ghostty_action_reload_config_s;
 
+// apprt.action.ReopenLastTab
+typedef struct {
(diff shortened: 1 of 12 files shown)


mitchellh:
Thank you for your attempt at this and I'm sorry I didn't review this sooner. Your work here is quite good. I would have some review comments but I've decided to pursue this in a different direction: since there is so much apprt-specific state to involve in this feature (tab vs split, positions, etc.) I'm going to prefer to implement this purely at the apprt level rather than a core libghostty feature. 

As such, I'll close this. I'm sorry!

pnodet:
No hards feelings, I had fun going deeper into ghostty

PR #3831, opened 2024-12-29
macos: always have terminal fullscreen style

### LINKED ISSUE #3553: macOS: exiting fullscreen does not restore background-opacity
When entering fullscreen mode with `cmd-ctrl-f` the background is not transparent anymore, as expected. But when I exit fullscreen mode the same way again, the original `background-opacity` value is not restored, or at least the background is not transparent anymore.

### Before fullscreen
Image

### Fullscreen
Image

### After exiting fullscreen
Image

## Config
Device: MacBook Pro 13" M1
OS: Macos 15.2
Version: Ghostty 1.0.0

`~/.config/ghostty/config`:
```
background-opacity = 0.7
background-blur-radius = 20

window-padding-x = 8
window-padding-y = 8
window-colorspace = display-p3
macos-titlebar-style = hidden
macos-icon = custom-style
macos-icon-frame = plastic
macos-icon-ghost-color = ffffff
macos-icon-screen-color = 262335
command = /bin/zsh -c /opt/homebrew/bin/zellij

foreground = ffffff
background = 262335
cursor-color = ffffff
selection-foreground = ffffff
selection-background = 262335

palette = 0=#262335
palette = 1=#fe4450
palette = 2=#72f1b8
palette = 3=#f3e70f
palette = 4=#03edf9
palette = 5=#ff7edb
palette = 6=#03edf9
palette = 7=#ffffff
palette = 8=#614d85
palette = 9=#fe4450
palette = 10=#72f1b8
palette = 11=#fede5d
palette = 12=#03edf9
palette = 13=#ff7edb
palette = 14=#03edf9
palette = 15=#ffffff
```

gigamonster256:
~~When processing toggle_fullscreen actions, multiple FullscreenStyle derived objects would register observers on the same window and subsequently remove them when the old style was deinited.~~

~~This change passes ownership of the fullscreenNotification observers to TerminalController while still allowing FullscreenStyles access to the window for style specific actions.~~

We need to have at least one fullscreen style (even if it is not being used) to make sure that observers for fullscreen state changes are set up. An alternative approach would be the first version of this PR (for which my reasoning was wrong but the end change was correct) which registers the fullscreen observers at either the TerminalController or BaseTerminalController level.

Fixes #3553

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -309,11 +309,11 @@ class BaseTerminalController: NSWindowController,
             // We consider our mode changed if the types change (obvious) but
             // also if its nil (not obvious) because nil means that the style has
             // likely changed but we don't support it.
-            if newStyle == nil || type(of: newStyle) != type(of: oldStyle) {
+            if newStyle == nil || type(of: newStyle!) != type(of: oldStyle) {
                 // Our mode changed. Exit fullscreen (since we're toggling anyways)
-                // and then unset the style so that we replace it next time.
+                // and then set the new style for future use
                 oldStyle.exit()
-                self.fullscreenStyle = nil
+                self.fullscreenStyle = newStyle
 
                 // We're done
                 return


gigamonster256:
reading @a7mdDev's https://github.com/ghostty-org/ghostty/issues/3553#issuecomment-2564956925, the fix is much simpler than I originally thought...
also fixes inforrect nullable type asssertion.

mitchellh:
Perfect. Thank you!

PR #3837, opened 2024-12-29
font/freetype: Enable bitmap glyphs for non-color faces

sin-ack:
This allows for crisp bitmap font rendering once again. Tested with Terminus (TTF), and at both 1x and 2x DPI the font renders perfectly (click for 1:1 size):

Terminus (TTF), size 12, 96 DPI

Terminus (TTF), size 12, 192 DPI


sin-ack:
@mitchellh Since you were the author of #1428, can you elaborate on what you mean by "issues with cell metrics and rasterization"?

src/font/face/freetype_convert.zig
```diff
@@ -43,26 +43,14 @@ pub fn monoToGrayscale(alloc: Allocator, bm: Bitmap) Allocator.Error!Bitmap {
...
-                buf[target_i] = bitval;
-                target_i += 1;
-            }
+    for (0..bm.rows) |y| {
+        const row_offset = y * @as(usize, @intCast(bm.pitch));
+        for (0..bm.width) |x| {
+            const byte_offset = row_offset + @divTrunc(x, 8);
+            const mask = @as(u8, 1) << @intCast(7 - (x % 8));
+            const bit: u8 = @intFromBool((bm.buffer[byte_offset] & mask) != 0);
+            buf[y * bm.width + x] = bit * 255;
```
mitchellh, line 52: On the same line of thinking I talked about in Discord, it'd be great if we added Terminus (license allowing) to `src/font/res` and add a unit test that failed before and passes now so that we don't regress this.

src/font/face/freetype.zig
```diff
@@ -323,14 +322,6 @@ pub const Face = struct {
             // glyph properties before render so we don't render here.
             .render = !self.synthetic.bold,
 
-            // Disable bitmap strikes for now since it causes issues with
-            // our cell metrics and rasterization. In the future, this is
-            // all fixable so we can enable it.
-            //
-            // This must be enabled for color faces though because those are
-            // often colored bitmaps, which we support.
-            .no_bitmap = !self.face.hasColor(),
```
mitchellh, line 332: Just noting I'm going to consider the implications of this a bit further. I don't remember the full history here.

src/font/face/freetype_convert.zig
sin-ack: Added Terminus TTF, and added a test that ensures that bitmap fonts keep being rendered crisply.

mitchellh:
Okay, I've been running locally with this pulled in for a couple days, and I can't remember why that original comment was there. I think it's likely that Qwer's work on metrics just resolved that for us (or some other work along the way, since that line is _ancient_). Merging.

PR #3843, opened 2024-12-29
Use foreground color in macOS tab titles

BlakeWilliams:
I'm not a Zig or a Swift/macOS developer but I've been really liking GhostTTY and wanted to open this PR to either contribute back, or at least share the progress/learnings when tackling https://github.com/ghostty-org/ghostty/issues/2764 (which I think this should close, if this is a valid approach). I have a vimscript that syncs the terminal colors with my vim theme, so I was pretty motivated to implement this feature.

This change was relatively straightforward (but required a lot of learning). In general it:

- Tracks `foregroundColor` in the SurfaceView. It's pre-populated with the config defined foreground color. It is also updated via the color change events (which I added foreground support to)
- Adds a new `titleForegroundColor` property to use in `TerminalWindow` titles so the colors can be updated dynamically.
- Updates each window's tab foreground color when `syncAppearance` is called. (I'm honestly not sure if this is the best approach, but it's the solution I found. happy to update if there's a better approach)
- Calls `syncAppearance` when a window becomes the key window to sync the colors across each tab.
- Updates some fonts to use 11pt sizes instead of `systemFontSize` to match the existing behavior/size of tabs (happy to provide screenshots to elaborate)

Here's a quick screencap of the new behavior:

https://github.com/user-attachments/assets/ba97432e-193a-4afd-95cd-faf23694530a

<details>
  <summary>I also quickly validated that this works with `macos-titlebar-style=transparent` in addition to `tabs`</summary>

Screenshot 2024-12-28 at 10 01 24 PM
</details>

<details>
  <summary>and just for completions sake, bg transparency:</summary>
  Screenshot 2024-12-28 at 10 04 20 PM
</details>

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -253,20 +253,25 @@ class TerminalController: BaseTerminalController {
         // If it does, we match the focused surface. If it doesn't, we use the app
         // configuration.
         let backgroundColor: OSColor
+        let foregroundColor: OSColor
         if let surfaceTree {
             if let focusedSurface, surfaceTree.doesBorderTop(view: focusedSurface) {
                 // Similar to above, an alpha component of "0" causes compositor issues, so
                 // we use 0.001. See: https://github.com/ghostty-org/ghostty/pull/4308
                 backgroundColor = OSColor(focusedSurface.backgroundColor ?? surfaceConfig.backgroundColor).withAlphaComponent(0.001)
+                foregroundColor = OSColor(focusedSurface.foregroundColor ?? surfaceConfig.foregroundColor)
             } else {
                 // We don't have a focused surface or our surface doesn't border the
                 // top. We choose to match the color of the top-left most surface.
                 backgroundColor = OSColor(surfaceTree.topLeft().backgroundColor ?? derivedConfig.backgroundColor)
+                foregroundColor = OSColor(surfaceTree.topLeft().foregroundColor ?? derivedConfig.foregroundColor)
             }
         } else {
             backgroundColor = OSColor(self.derivedConfig.backgroundColor)
+            foregroundColor = OSColor(self.derivedConfig.foregroundColor)
         }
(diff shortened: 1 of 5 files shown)


macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -268,6 +273,11 @@ class TerminalController: BaseTerminalController {
             // so we only call this if we are opaque.
             window.updateTabBar()
         }
+
+        guard let windows = self.window?.tabbedWindows as? [TerminalWindow] else { return }
+        for (w) in windows {
+            w.titleForegroundColor = window.titleForegroundColor
+        }
```
mitchellh, line 280: It feels like this should be moved to a private func and this should be called here and in `windowDidBecomeKey` instead of a _full `syncAppearance`_ which is rather expensive. Thoughts?

mitchellh (changes requested):
Honestly, very good job. I have one question/request.

macos/Sources/Features/Terminal/TerminalController.swift
BlakeWilliams: That sounds good and reasonable to me.  I'll make that change later today and do some more manual testing to make sure all the existing cases are still handled.

BlakeWilliams: Updated in https://github.com/ghostty-org/ghostty/pull/3843/commits/3c17b9176e6c6cbfca6b83ce5943c5b185480ebb

BlakeWilliams:
I noticed this while testing another color scheme, but some color schemes with light backgrounds can still be difficult to read. I think this is due to lighter tab backgrounds being shaded when dimmed.

Happy to address it in this PR if necessary, but also happy to try finding some time to look into it as a follow-up if it's not a blocker. 

Screenshot 2025-01-08 at 10 32 38 PM

mitchellh:
Thanks for this. I apologize for not getting this in. I'm going to go ahead and close this. There haven't been further requests for this and with Tahoe on the way this is all changing pretty dramatically.

PR #3844, opened 2024-12-29
cgroup: change suffix to .scope

ruihe774:
I think it's more consistent with the name convention used by systemd and other terminal emulators (e.g. gnome-terminal).

See also: https://systemd.io/CGROUP_DELEGATION/#systemds-unit-types


tristan957 (approved):

heftig:
FTR, I've found this confusing. The cgroups created aren't actually scopes.

PR #3857, opened 2024-12-29
gtk/x11: link directly to libX11, no more dlopen

jcollie:
As a follow-up to #3477 and #3748, this eliminates the use of dlopen to access `libX11` functions by directly linking `libX11` if X11 is enabled. This should also fix problems with systems like NixOS and Void Linux that have reported problems using Ghostty on X11 when using the distribution packages.


tristan957 (approved):

NOTtheMessiah:
Just ran this with `nix run github:ghostty-org/ghostty/pull/3857/head`, and it seems to fix the issue I was having with `nix run github:ghostty-org/ghostty`

PR #3903, opened 2024-12-29
surface: handle hyperlinks more reliably

rockorager:
We refresh the link hover state in two (generic) cases

1. When the modifiers change
2. When the cursor changes position

Each of these have additional state qualifiers. Modify the qualifiers
such that we refresh links under the following scenarios:

1. Modifiers change
  - Control is pressed (this is handled in the renderer)
  - Mouse reporting is off
    OR
    Mouse reporting is on AND shift is pressed AND we are NOT reporting
    shift to the terminal

2. Cursor changes position
  - Control is pressed (this is handled in the renderer)
  - We previously were over a link
  - The position changed (or we had no previous position)
  - Mouse reporting is off
    OR
    Mouse reporting is on AND shift is pressed AND we are NOT reporting
    shift to the terminal

This fixes a few issues with the previous implementation:

1. If mouse reporting was on and you were over a link, pressing ctrl
   would enable link hover state. If you moved your mouse, you would
   exit that state. The logic in the keyCallback and the
   cursorPosCallback was not the same. Now, they both check for the same
   set of conditions
2. If mouse reporting was off, you could hold control and move the mouse
   to discover links. If mouse reporting was on, holding control + shift
   would not allow you to discover links. You had to be hovering one
   when you pressed the modifiers. Previously, we only refreshed links
   if we *weren't* reporting the mouse event. Now, we refresh links even
   even if we report a mouse event (ie a mouse motion event with the
   shift modifier pressed *will* hover links and also report events)

## Old Behavior

Notice that the state of the hyperlink is erratic in `comlink`. When I am over it and press ctrl the link is underlined and the url hint in the lower left shown for one frame, but then the state is dropped.

https://github.com/user-attachments/assets/52d6a8c8-8459-4d67-85eb-5d91f9833771

## New Behavior

State is retained when holding ctrl+shift. And the link only underlines if I press both ctrl+shift. If I move the mouse around while holding these keys, I can discover new links.

https://github.com/user-attachments/assets/78fa8e97-eb0c-4618-bd96-fe40d6bc67ce

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1703,16 +1703,37 @@ pub fn keyCallback(
         // Update our modifiers, this will update mouse mods too
         self.modsChanged(event.mods);
 
-        // Refresh our link state
-        const pos = self.rt_surface.getCursorPos() catch break :mouse_mods;
-        self.mouseRefreshLinks(
-            pos,
-            self.posToViewport(pos.x, pos.y),
-            self.mouse.over_link,
-        ) catch |err| {
-            log.warn("failed to refresh links err={}", .{err});
-            break :mouse_mods;
-        };
+        // We only refresh links if
+        // 1. mouse reporting is off
+        // OR
+        // 2. mouse reporting is on and we are not reporting shift to the terminal
+        if (self.io.terminal.flags.mouse_event == .none or
+            (self.mouse.mods.shift and !self.mouseShiftCapture(false)))
+        {
+            // Refresh our link state
+            const pos = self.rt_surface.getCursorPos() catch break :mouse_mods;
+            self.mouseRefreshLinks(
+                pos,
+                self.posToViewport(pos.x, pos.y),
+                self.mouse.over_link,
+            ) catch |err| {
+                log.warn("failed to refresh links err={}", .{err});
+                break :mouse_mods;
+            };
+        } else if (self.io.terminal.flags.mouse_event != .none and !self.mouse.mods.shift) {
+            // If we have mouse reports on and we don't have shift pressed, we reset state
+            try self.rt_app.performAction(
+                .{ .surface = self },
+                .mouse_shape,
+                self.io.terminal.mouse_shape,
+            );
+            try self.rt_app.performAction(
+                .{ .surface = self },
+                .mouse_over_link,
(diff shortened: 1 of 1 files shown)


rockorager:
cc @tristan957

tristan957:
Thanks for finding the proper fix. Next release is gonna be great!

mitchellh:
It's hard to make sense of this logic since it's so poorly tested (something I'm working on). A brief glance this looks okay, and I think it should only affect links if there are bugs in this logic anyways, and links are already a bit buggy. So overall I'm just going to merge this and try to refactor in tests later. Thank you!

rockorager:
I completely agree. I tried to document *when* we do this the best I could, but there are lot of moving pieces, as well as opinions. This currently works how I intuitively would think it should, but even then I had some reservations on which behavior is preferred. Ping me once you are ready with your refactor and I can put in some concrete tests to get the behavior exact (and make it easier to discuss the ideal behavior).

HartBlanc:
This change has triggered some discussion here that may help to inform the decision about what the correct behaviour is:
https://github.com/ghostty-org/ghostty/discussions/4382#discussioncomment-11716566

Personally, I preferred when I didn't need to press shift. This is partly because this is how iTerm behaves (the terminal emulator I was using before Ghostty), but also because I didn't need to be aware of when mouse reporting was enabled when deciding which keys to press.

rockorager:
> Personally, I preferred when I didn't need to press shift. This is partly
> because this is how iTerm behaves (the terminal emulator I was using before
> Ghostty), but also because I didn't need to be aware of when mouse reporting
> was enabled when deciding which keys to press.

I tested a couple terminals before deciding on this behavior, but I don't have
any on my system that have a behavior I like (except for foot, with it's hints -
but this is entirely different). The closest in features is kitty, which has
hover action *without pressing anything*. Having used this for a few days now, I
would agree that it would be nicer to have the functionality be the same whether
capture is on or not, especially since there isn't an obvious way for the user
to know if the application has mouse events on.

mitchellh:
The plan is to introduce mouse bindings (there's a discussion about it) and let people customize their hyperlink behavior to their heart's content.

PR #3920, opened 2024-12-29
apprt/gtk: style the tab overview when window-theme=ghostty

tristan957:
This requires libadwaita main (to be 1.8).

Signed-off-by: Tristan Partin <tristan@partin.io>

https://github.com/user-attachments/assets/057abec5-f85b-4fbb-980c-c58a20e7ddc7

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -987,13 +987,15 @@ fn loadRuntimeCss(
         switch (window_theme) {
             .ghostty => try writer.print(
                 \\:root {{
-                \\  --headerbar-fg-color: rgb({d},{d},{d});
-                \\  --headerbar-bg-color: rgb({d},{d},{d});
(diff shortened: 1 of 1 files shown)


jcollie (approved):

mitchellh:
I assume the variable here has no effect on prior versions hence we don't need a version check. I trust you enough to assume you thought that through so I'll merge but if you can confirm that'd be great.

tristan957:
> I assume the variable here has no effect on prior versions hence we don't need a version check. I trust you enough to assume you thought that through so I'll merge but if you can confirm that'd be great.

The only issue it could present is if somewhere else in our GTK stack, someone was using the same variable, but it is safe to assume that neither GTK nor libadwaita prior to 1.8 has ever used this variable. Remember that CSS variables are new in GTK as of 4.16, too.

PR #3923, opened 2024-12-29
Fix bash script without conditional function

donnybeelo:
This is a compatibility improvement

The bash script that runs when ghostty is launched produces the following error:
```
bash: /home/dan/git/ghostty/zig-out/share/ghostty/shell-integration/bash/ghostty.bash: line 120: syntax error near unexpected token `('
bash: /home/dan/git/ghostty/zig-out/share/ghostty/shell-integration/bash/ghostty.bash: line 120: `        sudo() {'
```

This change should alleviate that by defining the original sudo function outside and aliasing sudo to that function. This should allow ghostty to work without issues on systems with sudo already aliased.

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -6,26 +6,6 @@
 # my simple bash usage this is working. If a bash expert wants to
 # improve this please do!
 
-sudo_function() {
-  builtin local sudo_has_sudoedit_flags="no"
-  for arg in "$@"; do
-    # Check if argument is '-e' or '--edit' (sudoedit flags)
-    if [[ "$arg" == "-e" || $arg == "--edit" ]]; then
-      sudo_has_sudoedit_flags="yes"
-      builtin break
-    fi
-    # Check if argument is neither an option nor a key-value pair
-    if [[ "$arg" != -* && "$arg" != *=* ]]; then
-      builtin break
-    fi
-  done
-  if [[ "$sudo_has_sudoedit_flags" == "yes" ]]; then
-    builtin command sudo "$@";
-  else
-    builtin command sudo TERMINFO="$TERMINFO" "$@";
-  fi
-}
-
 # We need to be in interactive mode and we need to have the Ghostty
 # resources dir set which also tells us we're running in Ghostty.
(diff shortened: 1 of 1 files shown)


tristan957:
cc @jparise

src/shell-integration/bash/ghostty.bash
```diff
@@ -6,6 +6,26 @@
 # my simple bash usage this is working. If a bash expert wants to
 # improve this please do!
 
+sudo_function() {
```
jparise, line 9: I don't think we want to leave this function in the global namespace. That was an advantage of the previous approach, along with only defining it when the `sudo` shell feature is enabled.

If we can't find another approach, we should rename this with a `__ghostty_` prefix like our other global functions.

src/shell-integration/bash/ghostty.bash
```diff
@@ -117,25 +137,7 @@ function __ghostty_precmd() {
...
-              builtin break
-            fi
-          done
-          if [[ "$sudo_has_sudoedit_flags" == "yes" ]]; then
-            builtin command sudo "$@";
-          else
-            builtin command sudo TERMINFO="$TERMINFO" "$@";
-          fi
-        }
+        alias sudo="sudo_function"
```
jparise, line 140: I'll also have to think a little bit about the differences between using an `alias` versus function override.

jparise:
I thought about this a little more. I think we should also consider there two alternate approaches:

1. If `sudo` is already a function or an alias, skip our integration code.
2. If `sudo` is already a function or an alias, save a reference to that existing symbol, replace it with our function, and invoke that existing symbol instead of `builtin command sudo`. We can't assume compatibility, but users can always disable our 'sudo' shell feature in that case.

mitchellh:
I like both approaches, so I'm open to either.

donnybeelo:
It's more of an issue with bash checking for syntax errors in the ghostty script before it even runs, and erroring when it sees sudo being used as a function name when it exists as an alias. I implemented the second

jparise:
I spent some more time looking at this and am proposing an alternate solution in #4080.

It uses the `function sudo` syntax, which won't conflict with an existing 'sudo' alias in a syntax-error-producing way.

PR #3929, opened 2024-12-29
Set an initial start position

### LINKED ISSUE #3362: Configure initial window position
I didn't see a configuration option that will allow me to specify the coordinates for the initial window position, but I would like for such an option to exist.

For example, the initial window is always created in a centered position after the app is launched…

Image

…but perhaps I'd like for it to be created at the top left instead:

Image

aw1875:
Allow the ability to set an initial start position from the config. Adds `window-initial-position-{x,y}` to the config as an optional i16 value (see swift docs in [this comment](https://github.com/ghostty-org/ghostty/pull/3929#discussion_r1899266607) for the reasoning behind this if needed) and handles setting the position when the initial window is created

Closes https://github.com/ghostty-org/ghostty/issues/3362


src/apprt/gtk/Surface.zig
```diff
@@ -840,6 +840,12 @@ pub fn setInitialWindowSize(self: *const Surface, width: u32, height: u32) !void
     );
 }
 
+pub fn setInitialWindowPosition(self: *const Surface, x: i32, y: i32) !void {
+    // We need the surface's window to set the position.
+    const window = self.container.window() orelse return;
+    c.gtk_window_move(@ptrCast(window.window), x, y);
```
jcollie, line 846: `gtk_window_move` is not a valid GTK4 API. If this is PR is accepted for macOS, `setInitialWindowPosition` should be a no-op on GTK.

jcollie (changes requested):

src/apprt/gtk/Surface.zig
aw1875: yup! I made a note of that in the "Work in progress" to-do list. I'd personally like to explore if there are any compatibility workarounds for GTK4 so that this can still be an option but I don't currently have the time today to dig deeper into it so for now this is just a draft. If I don't end up finding any "good" way to do this, I'll be sure to make it clear that this config option only works on macOS and X11 (IIRC there is a build config option to build using GTK3 for X11)

src/config/Config.zig
```diff
@@ -1113,9 +1113,12 @@ keybind: Keybinds = .{},
 /// to be a certain position to fit within the grid. There is nothing Ghostty
 /// will do about this, but it will make an effort.
 ///
+/// Important note: Setting this value will only work on macOs and glfw builds
```
jcollie, line 1116: `macOS`

src/config/Config.zig
```diff
@@ -1113,9 +1113,12 @@ keybind: Keybinds = .{},
...
+/// on Linux. GTK 4.0 does not support setting the window position.
+///
 /// This will default to the top-left corner of the screen if not set (0, 0).
-@"window-position-x": i32 = 0,
-@"window-position-y": i32 = 0,
+@"start-position-x": i32 = 0,
+@"start-position-y": i32 = 0,
```
jcollie, line 1122: `window-initial-position-x` and `window-initial-position-y` I think would be clearer as to their purpose.

src/config/Config.zig
aw1875: oops thank you

src/config/Config.zig
aw1875: cool sounds good, I'll revert that change

src/apprt/gtk/Surface.zig
jcollie: The GTK devs are pretty adamant that positioning windows in that way is out of scope for GTK. Ghostty will not build against GTK3. GTK4 can use X11 but even in that case you can't position windows. It's possible that there's a Wayland protocol to do that but I don't think that we're ready to be talking to Wayland directly. And compositor specific solutions I think are not a viable option either (there's just too many of them).

aw1875: yeah I actually just did some digging around for a bit and nothing really seemed that great to me either, I removed the GTK implementation of `setInitialWindowPosition` in [826bc18](https://github.com/ghostty-org/ghostty/pull/3929/commits/826bc18204ae1aa580340f9aa8ef96e565de5e15) and updated the config docs appropriately

jcollie:
After some thinking, I think that this is the wrong approach. There shouldn't be any need to store the position in the core Surface struct or to use an apprt action to position the window. First, only the first surface that initiates the window creation should control the position of the window. If a new tab or split can affect the position of the window you'll have windows jumping back to their initial position after the user has moved them around and then creates a new tab or split.

When a new window is created it should check to see if these settings are configured and position the window at window creation time. There's no need for a message to be sent to the window at a later time.

src/config/Config.zig
jcollie: The should also be unsigned and optional. That way you can tell if the user did not set these in their config. 32 bits is probably too much, I think that a `u16` should be fine.

jcollie:
I don't program in Swift but there should be some existing machinery from macOS's state restoration that can be reused to set the initial window position from the config.

src/config/Config.zig
aw1875: Correct me if I'm wrong but you can't you have negative coordinates depending on your monitor layout? I might just be mixing this up with something else but I thought that was a possibility

jcollie: I can't say for 100% sure, you'd need to check the documentation of the various APIs used to position the window. We definitely want these to be optional though. If the user didn't set them in the config, we don't want to force them into a specific location.

aw1875: I'm going to drop this here for later but macOS allows negative values (my understanding is that this value is based on your primary monitor's bottom left corner). I'll also dig into the Linux side later to see if I can find any docs supporting this behaviour.

https://developer.apple.com/documentation/appkit/nswindow/setframeorigin(_:)#Discussion

jsejcksn:
> I'm considering renaming `window-position-{x,y}` to `start-position-{x,y}` for clarity

I have no strong opinion about the name, but because existing configuration keys already use the term `initial` instead of `start` (e.g. [`initial-window`](https://ghostty.org/docs/config/reference#initial-window)), I suggest `initial-window-position` — or, as @jcollie [mentioned](https://github.com/ghostty-org/ghostty/pull/3929#discussion_r1899256090), `window-initial-position`.

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -461,6 +461,31 @@ class TerminalController: BaseTerminalController {
...
         data.encode(with: state)
     }
+    
+    func setInitialWindowPosition(_ window: NSWindow, x: Int16?, y: Int16?, windowDecorations: Bool) {
+        if let primaryScreen = NSScreen.screens.first {
+            let frame = primaryScreen.visibleFrame
+
+            if let windowPositionX = x, let windowPositionY = y {
+                // Offset titlebar if needed, otherwise use default padding of 12
+                // NOTE: Not 100% certain where this extra padding comes from but I'd love
```
aw1875, line 471: Highlighting this note here in case a reviewer has more knowledge in the mac window space, I couldn't find anything helpful here but my hunch is this is default gap/padding/margin for NSWindow frames

mitchellh (approved):
I made some modifications (see my commit), but overall great job!

I renamed the config to `window-position-x` and y. The `initial` was a bit redundant there since the position is only ever relevant at start. I don't see a future where we have any other position values.

I also removed the titlebar math you had. I'm not fully sure I understood it, and the magic numbers were freaking me out. If there is a bug here in some circumstances we should open a discussion and figure that out in a more robust manner.

ls-comh:
Might it be possible to support the tiling positions introduced with macOS Sequoia?

To give an illustration:
_Open Ghostty, then from the menubar select Window -> Move -> Right_

Explicit x,y coordinates are valid and enable full customisation, but also supporting the built-in options would likely cover most cases, feel native, avoid working out coordinates, and work if/when the screen size changes.

e.g. `window-position: right` would be elegant 👌

ls-comh:
I was so happy to see this rolled out today ❤️ 

It _would_ be nice to replace:
```
window-height = 60
window-width = 101
window-position-x = 864
window-position-y = 0
```

With:
`window-position: right`

But I'm happy regardless. My thanks goes out to everyone involved 🙏

PR #3931, opened 2024-12-29
Added snap packaging

kenvandine:
Added snap packaging, fixes #3153

If whoever registered the name in the snap store could add me as a collaborator, I can handle getting it released in the store, setup automated builds, and request the necessary classic permissions in the store.


snap/snapcraft.yaml
```diff
@@ -0,0 +1,77 @@
...
+  zig:
+    plugin: nil
+    build-packages:
+      - curl
+    override-pull: |
+      set -ex
+      case "$CRAFT_TARGET_ARCH" in
+        amd64)  arch=x86_64    ;;
+        arm64)  arch=aarch64  ;;
+        *)      arch=""     ;;
```
tristan957, line 45: Presumably it would be good to exit non-zero in this case with a good error message before curl tries to download something that doesn't exist.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+  ghostty:
+    source: .
+    after: [ zig ]
+    plugin: nil
+    build-packages:
+      - libgtk-4-dev
+      - libadwaita-1-dev
+      - git
+    stage-packages:
```
tristan957, line 68: I'm not too familiar with Snap. What are stage packages?

tristan957 (commented):
How do all of the shell completion files and vim/neovim files get exported? I don't see any logic related to them here.

I'd like to see a GitHub Actions workflow added, so this doesn't regress :).

kenvandine:
> How do all of the shell completion files and vim/neovim files get exported? I don't see any logic related to them here.
> 
> I'd like to see a GitHub Actions workflow added, so this doesn't regress :).

I've added a workflow to ensure the snap builds.  I'll work on the shell completion. I hadn't done anything for that, but will

snap/snapcraft.yaml
kenvandine: stage-packages are ubuntu packages that satisfy runtime dependencies which get bundle in the snap.

tristan957: Hmm, why would we want to bundle shells? Do classic snaps not have access to the host binaries?

.github/workflows/snap.yaml
```diff
@@ -0,0 +1,14 @@
+name: Snap
+
+on:
+  push: {}
```
tristan957, line 4: Let's restrict this to certain branches, at least main for now

.github/workflows/snap.yaml
```diff
@@ -0,0 +1,14 @@
...
+  pull_request: {}
```
tristan957, line 5: Similar comment here

.github/workflows/snap.yaml
```diff
@@ -0,0 +1,14 @@
...
+  workflow_dispatch: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: snapcore/action-build@v1
+      id: snapcraft
```
tristan957, line 14: Let's remove the ide since it isn't being referenced.

.github/workflows/snap.yaml
```diff
@@ -0,0 +1,14 @@
+name: Snap
+
+on:
+  push: {}
+  pull_request: {}
+  workflow_dispatch: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
```
tristan957, line 10: We will probably want to use one of the project's runners. I'll let @mitchellh give a better review here.

snap/local/launcher
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+export XDG_CONFIG_HOME="$SNAP_REAL_HOME/.config"
+export XDG_DATA_HOME="$SNAP_REAL_HOME/.local/share"
+
+exec "$@"
```
tristan957, line 6: It seems like we could just use `/bin/sh` for the shebang.

snap/snapcraft.yaml
kenvandine: It does have access to anything on the host, so these could be removed.

.github/workflows/snap.yaml
kenvandine: Fixed

.github/workflows/snap.yaml
kenvandine: Fixed

.github/workflows/snap.yaml
kenvandine: Done

snap/local/launcher
kenvandine: Done

kenvandine:
@tristan957 I think I've responded to all the feedback so far.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+        exit 1
+      fi
+
+      tar xf zig-lin*xz
+      rm -f *xz
+      mv zig-linux*/* .
+    prime:
+      - -*
+
+  ghostty:
```
matifali, line 66: I wonder why you did not choose to use the pre-built packages available at https://github.com/ghostty-org/ghostty/blob/main/PACKAGING.md?

kenvandine: That's more desirable for third party packagers, but if the ghostty project is publishing the snap it's nice to keep the packaging in the source and stable releases can be easily maintained by building from stable branches or specific tags.

mitchellh (changes requested):
I'm still not 100% sure I want to maintain a snap but it does seem to be more generally accepted. I'm going to think about that more.

My main request here is does `snap/` have to be a top-level folder? I think it makes more sense to put into `dist/`

tristan957:
I think there is slight value in maintaining a flatpak and snap. They are "universal" formats that can run on any (if snap is supported) distro. It also gives us the ability to have an official Linux distribution channel(s) that we can recommend for people in the event downstreams can't properly package Ghostty for users.

Of note, the author of this PR works at Canonical, so perhaps they could be interested in helping to maintain the Snap into the future :)

mitchellh:
Yeah. I'm leaning towards it. 😄 

@kenvandine if you'd be happy to hang around and help out I'd absolutely love that too.

kenvandine:
snapcraft looks for the packaging in a few places.  We'd have to jump through some hoops if we wanted it under dist.  We could do it, but I would like to keep things as simple as possible for automated builds.

I am happy to help look after the snap! I maintain quite a few already and I plan to be a daily user of ghostty too.

Who registered the name in the snap store?

mitchellh:
> snapcraft looks for the packaging in a few places. We'd have to jump through some hoops if we wanted it under dist. We could do it, but I would like to keep things as simple as possible for automated builds.

Understood. We can keep it where it is.

> Who registered the name in the snap store?

Not me. We may have to register a dispute.

kenvandine:
Want me to go ahead and start the dispute?

kenvandine:
I created a name dispute request in the store.

mitchellh:
> I created a name dispute request in the store.

Appreciate it. After the dispute, do I need to transfer it to my account?

kenvandine:
Yeah, once we get everything in place we'll get it transferred.

kenvandine:
I've requested permission for classic confinement at https://forum.snapcraft.io/t/request-classic-confinement-for-ghostty/44523

mitchellh:
I updated the builders to use Namespace, but it looks like there are issues there. I'll reach out to support for that. 

I tested this locally and works great, but will be important for CI to be able to run this. 😄

mitchellh:
FYI Namespace has a fix in the pipeline so we'll probably have CI working soon.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+platforms:
+  amd64:
+  arm64:
+
+apps:
+  ghostty:
+    command: bin/ghostty
+    command-chain: [ bin/launcher ]
+    completer: share/bash-completion/completions/ghostty.bash
```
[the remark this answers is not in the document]
tristan957: to be fair, I don't much about snap, so take this comment with a grain of salt.

dobicinaitis:
Hi! I was about to request classic confinement for my own Ghostty snap when I noticed the request to approve the upstream one. Awesome! I'll stop working on my implementation, but you might find something useful in the GitHub Actions workflows I set up ([link](https://github.com/dobicinaitis/ghostty-snap/tree/main/.github/workflows)). They were designed to periodically check for new Ghostty releases and create snaps for both the stable and tip channels. However, all the version change detection would be redundant for workflows in this repository.

mitchellh:
Your Git history here is getting messed up, btw. You probably see it already but just in case you don't...

kenvandine:
> Your Git history here is getting messed up, btw. You probably see it already but just in case you don't...

Ugh, thanks I hadn't noticed. I've been rebasing while I iterate. Currently working to make sure the classic snap works well on non-ubuntu distros, which can be tedious. I'll get the history cleaned up soon.

kenvandine:
> Your Git history here is getting messed up, btw. You probably see it already but just in case you don't...

I'm struggling to see why the history is whacky. I've been rebasing my branch from main regularly and the number of files showing up in the PR as changed is growing. Since my rebases have been from main, those changes shouldn't show in my diff. I've never seen this happen before. Any pointers?

mitchellh:
I'll take a look and try to force push a fix onto the branch tomorrow if I can 😄

mitchellh:
@kenvandine I cleaned up the git history. I'm not sure if I did a couple conflicts correctly so please review the files... but the conflicts were very small so I think they'd be minor.

snap/snapcraft.yaml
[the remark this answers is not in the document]
kenvandine: I think the snap completer logic only knows bash for now.

kenvandine:
> @kenvandine I cleaned up the git history. I'm not sure if I did a couple conflicts correctly so please review the files... but the conflicts were very small so I think they'd be minor.

Thanks, that looks much better. There was one small change missing, I just pushed that as a new commit. I'll avoid rebasing for now to prevent that weirdness again.

mitchellh:
Got a bit further on CI, that's the major blocker for this. I'm going to keep plugging away at it and working with Namespace on a resolution. :)

kenvandine:
@mitchellh, I am really happy with the state of the snap now, so from my perspective, this is ready to land whenever you are ready for that. Once this is merged, we can chat about the process to transfer the snap over to you in the store and get automated builds going.

snap/local/launcher
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+export XDG_CONFIG_HOME="$SNAP_REAL_HOME/.config"
+export XDG_DATA_HOME="$SNAP_REAL_HOME/.local/share"
```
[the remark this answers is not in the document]
tristan957: I still have a concern here.

tristan957 (changes requested):
I don't know much about snaps, so please forgive me. Ghostty also includes various support files for tools like bat, neovim, and vim. Is it possible for the snap to export those for the various tools to find or is it up to the user to make sure the tools can find those files?

Edit: Did not mean to request changes.

kenvandine:
> I don't know much about snaps, so please forgive me. Ghostty also includes various support files for tools like bat, neovim, and vim. Is it possible for the snap to export those for the various tools to find or is it up to the user to make sure the tools can find those files?
> 
> Edit: Did not mean to request changes.

If you mean if neovim is installed separately, that might be tricky. How do tools like neovim discover ghostty is installed?

kenvandine:
About XDG_CONFIG_HOME and XDG_DATA_HOME, those are to ensure the config files are exposed properly.

tristan957:
> > I don't know much about snaps, so please forgive me. Ghostty also includes various support files for tools like bat, neovim, and vim. Is it possible for the snap to export those for the various tools to find or is it up to the user to make sure the tools can find those files?
> > Edit: Did not mean to request changes.
> 
> If you mean if neovim is installed separately, that might be tricky. How do tools like neovim discover ghostty is installed?

The locations of the files which Ghostty installs seem to be on the default paths that vim and neovim pickup.

tristan957:
> About XDG_CONFIG_HOME and XDG_DATA_HOME, those are to ensure the config files are exposed properly.

Right, but the paths might not match the user's XDG_CONFIG_HOME and XDG_DATA_HOME so you would be exporting them into a random location?

kenvandine:
> > About XDG_CONFIG_HOME and XDG_DATA_HOME, those are to ensure the config files are exposed properly.
> 
> Right, but the paths might not match the user's XDG_CONFIG_HOME and XDG_DATA_HOME so you would be exporting them into a random location?



> > About XDG_CONFIG_HOME and XDG_DATA_HOME, those are to ensure the config files are exposed properly.
> 
> Right, but the paths might not match the user's XDG_CONFIG_HOME and XDG_DATA_HOME so you would be exporting them into a random location?

I made those exports conditional

kenvandine:
> > > I don't know much about snaps, so please forgive me. Ghostty also includes various support files for tools like bat, neovim, and vim. Is it possible for the snap to export those for the various tools to find or is it up to the user to make sure the tools can find those files?
> > > Edit: Did not mean to request changes.
> > 
> > 
> > If you mean if neovim is installed separately, that might be tricky. How do tools like neovim discover ghostty is installed?
> 
> The locations of the files which Ghostty installs seem to be on the default paths that vim and neovim pickup.

With the snap, the read-only filesystem will appear below /snap/ghostty/current

snap/local/launcher
```diff
@@ -0,0 +1,46 @@
...
+
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+$LD_LIBRARY_PATH:}${SNAP}/usr/lib/${ARCH}:${SNAP}/usr/lib/${ARCH}/vdpau
+export LIBGL_DRIVERS_PATH=${LIBGL_DRIVERS_PATH:+$LIBGL_DRIVERS_PATH:}${SNAP}/usr/lib/${ARCH}/dri/
+export LIBVA_DRIVERS_PATH=${LIBVA_DRIVERS_PATH:+$LIBVA_DRIVERS_PATH:}${SNAP}/usr/lib/${ARCH}/dri/
+export __EGL_VENDOR_LIBRARY_DIRS=${__EGL_VENDOR_LIBRARY_DIRS:+$__EGL_VENDOR_LIBRARY_DIRS:}${SNAP}/usr/share/glvnd/egl_vendor.d
+export __EGL_EXTERNAL_PLATFORM_CONFIG_DIRS=${__EGL_EXTERNAL_PLATFORM_CONFIG_DIRS:+$__EGL_EXTERNAL_PLATFORM_CONFIG_DIRS:}${SNAP}/usr/share/egl/egl_external_platform.d
+export DRIRC_CONFIGDIR=${SNAP}/usr/share/drirc.d
+export VK_LAYER_PATH=${VK_LAYER_PATH:+$VK_LAYER_PATH:}${SNAP}/usr/share/vulkan/implicit_layer.d/:${SNAP}/usr/share/vulkan/explicit_layer.d/
+export XDG_DATA_DIRS=${XDG_DATA_DIRS:+$XDG_DATA_DIRS:}${SNAP}/usr/share
+export XLOCALEDIR="${SNAP}/usr/share/X11/locale"
```
tristan957, line 31: Are these paths supposed to exist on other distributions too or do they exist with the snap itself, regardless of distribution?

tristan957:
> > > > I don't know much about snaps, so please forgive me. Ghostty also includes various support files for tools like bat, neovim, and vim. Is it possible for the snap to export those for the various tools to find or is it up to the user to make sure the tools can find those files?
> > > > Edit: Did not mean to request changes.
> > > 
> > > 
> > > If you mean if neovim is installed separately, that might be tricky. How do tools like neovim discover ghostty is installed?
> > 
> > 
> > The locations of the files which Ghostty installs seem to be on the default paths that vim and neovim pickup.
> 
> With the snap, the read-only filesystem will appear below /snap/ghostty/current

Thanks for bearing with me on these questions and clarifications! I think it would be good to add some documentation, somewhere, that says what to do if a user installs Ghostty through a snap and wants their various tools to pickup the support files. That documentation most likely should live in the website repository, and I'm not saying you should write it either. More just making a note!

Sounds like it is as easy as telling the tools to look in `/snap/ghostty/...`.

snap/local/launcher
kenvandine: These are available inside the snap runtime. It's to ensure the snap works on any distro, not just a single version of Ubuntu. It works fine on debian, fedora, etc.

kenvandine:
> > > > > I don't know much about snaps, so please forgive me. Ghostty also includes various support files for tools like bat, neovim, and vim. Is it possible for the snap to export those for the various tools to find or is it up to the user to make sure the tools can find those files?
> > > > > Edit: Did not mean to request changes.
> > > > 
> > > > 
> > > > If you mean if neovim is installed separately, that might be tricky. How do tools like neovim discover ghostty is installed?
> > > 
> > > 
> > > The locations of the files which Ghostty installs seem to be on the default paths that vim and neovim pickup.
> > 
> > 
> > With the snap, the read-only filesystem will appear below /snap/ghostty/current
> 
> Thanks for bearing with me on these questions and clarifications! I think it would be good to add some documentation, somewhere, that says what to do if a user installs Ghostty through a snap and wants their various tools to pickup the support files. That documentation most likely should live in the website repository, and I'm not saying you should write it either. More just making a note!
> 
> Sounds like it is as easy as telling the tools to look in `/snap/ghostty/...`.

Yes, it should be. /snap/ghostty/current will always point to the currently enabled version of the snap.

mitchellh:
Alright! Our CI seems to be hung up on this issue that our CI provider found: https://github.com/canonical/craft-providers/issues/719 I'm not sure if there is a workaround we can apply to that yet but that's the major issue right now...

kenvandine:
This is my first time touching zig, check out my changes to src/termio/Exec.zig which should help ensure the snap isn't leaking environment variables that were necessary to launch ghostty but shouldn't be set in there resulting terminal shell.

src/termio/Exec.zig
```diff
@@ -732,6 +732,19 @@ const Subprocess = struct {
...
+        if (env.get("SNAP") != null) {
+            env.remove("SNAP");
+            env.remove("DRIRC_CONFIGDIR");
+            env.remove("__EGL_EXTERNAL_PLATFORM_CONFIG_DIRS");
+            env.remove("__EGL_VENDOR_LIBRARY_DIRS");
+            env.remove("LD_LIBRARY_PATH");
+            env.remove("LIBGL_DRIVERS_PATH");
+            env.remove("LIBVA_DRIVERS_PATH");
+            env.remove("VK_LAYER_PATH");
+            env.remove("XLOCALEDIR");
```
tristan957, line 745: This would unconditionally remove these environment variables from the shell. Reading your launcher file above, you append the various snap paths if the environment variable exists. Wouldn't we just want to remove the snap paths from the environment variables if the user already had them set?

kenvandine: No, the problem is the resulting shell in ghostty uses these library paths, drivers, etc for any processes started from ghostty which won't work mixing libs from in the container with host libs. This effectively drops everything needed to start ghostty without leaking those paths into the resulting shell. If a user is setting any of them in /etc or .bashrc, etc those will be honored by the shell anyway.

src/termio/Exec.zig
```diff
@@ -732,6 +732,19 @@ const Subprocess = struct {
             try env.put("GHOSTTY_RESOURCES_DIR", dir);
         }
 
+        // Unset environment varies set by the snap
+        if (env.get("SNAP") != null) {
```
jcollie, line 736: The "Zig" way to do this is:

```
if (env.get("SNAP")) |_| {
```

lwbt:
- https://github.com/ghostty-org/ghostty/releases/tag/v1.1.0 has been released 5 days ago
- https://snapcraft.io/ghostty is at `latest/stable 1.0.2-tip-675-g4da51fa5`

I don't understand how snap packaging actually works and I don't expect same day updates to the snap, but please update it.

kenvandine:
> * https://github.com/ghostty-org/ghostty/releases/tag/v1.1.0 has been released 5 days ago
> * https://snapcraft.io/ghostty is at `latest/stable 1.0.2-tip-675-g4da51fa5`
> 
> I don't understand how snap packaging actually works and I don't expect same day updates to the snap, but please update it.

This will be automated once this PR is merged, but in the mean time I need to rebase which I've been avoiding because it caused some issues in the git history. I'll carefully do that now to get ghostty in the store updated.

mitchellh:
And just to note, I'm waiting on the upstream issue to get resolved in snap, otherwise our CI is flaky (and fails more than 50% of the time for me) due to the snap race condition. That's the major blocker now.

kenvandine:
> And just to note, I'm waiting on the upstream issue to get resolved in snap, otherwise our CI is flaky (and fails more than 50% of the time for me) due to the snap race condition. That's the major blocker now.

Understood, they are working on that.

mitchellh:
Snap 8.6.2 contains the required fix. Trying to the CI jobs again.

mitchellh:
@kenvandine Noticing this in the logs:

```
Building ghostty
:: ++ git describe --abbrev=8
:: fatal: No names found, cannot describe anything.
:: + craftctl set version=
```

They're still running though...

mitchellh:
Okay! They almost completed, just the version field issue. 😄

kenvandine:
> Okay! They almost completed, just the version field issue. 😄 

Can you make the job pull with --tags?

mitchellh:
> Can you make the job pull with --tags?

It does by default, I'm using `actions/checkout@v4`. I think you have to opt-in? Or am I mistaken?

mitchellh:
Tests are green! I moved the env var setting from termio to the apprt since that's where they go now.

One more green run and I'll merge this then we can consider release management and other things in future work.

Thank you @kenvandine

mitchellh (approved):

kenvandine:
> Tests are green! I moved the env var setting from termio to the apprt since that's where they go now.
> 
> One more green run and I'll merge this then we can consider release management and other things in future work.
> 
> Thank you @kenvandine 

That's awesome! I think the snap is in pretty good shape, I've tested on several Ubuntu versions, Debian 11 and 12, and fedora 41 myself.

PR #3934, opened 2024-12-29
core: add build option to disable sentry

jcollie:
This disables compiling/linking Sentry automatically on platforms other than macOS, which removes a potential blocker for getting Ghostty running on *BSD or other systems.

This is also useful for the security paranoid that don't want any chance that sensitive information could be captured in a crash dump.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -152,6 +152,17 @@ pub fn build(b: *std.Build) !void {
         }
     };
 
+    config.sentry = b.option(
+        bool,
+        "sentry",
+        "Build with Sentry crash reporting. Default for macOS is true, false for any other system.",
+    ) orelse sentry: {
+        switch (target.result.os.tag) {
+            .macos, .ios => break :sentry true,
+            else => break :sentry false,
+        }
+    };
+
     const pie = b.option(
         bool,
         "pie",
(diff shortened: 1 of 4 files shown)


build.zig
```diff
@@ -152,6 +152,18 @@ pub fn build(b: *std.Build) !void {
...
     };
 
+    config.sentry = b.option(
+        bool,
+        "sentry",
+        "Build with Sentry crash reporting. Default for Linux and macOS is true, false for any other system.",
+    ) orelse sentry: {
+        if (target.result.isDarwin()) break :sentry true;
+        switch (target.result.os.tag) {
+            .linux, .macos, .ios => break :sentry true,
```
mitchellh, line 162: We should probably just disable Sentry on Linux. It actually doesn't capture symbols today so its not useful.

mitchellh (changes requested):

build.zig
jcollie: Done. I left the CI task that builds Sentry support on Linux to prevent any regressions.

mitchellh:
Updated to remove the macOS CI test. macOS builds are already slow and expensive as it is, I don't think this is a common enough config on macOS to warrant it. Kept Linux.

Also removed the Windows check since we default to false on Windows anyways.

mitchellh (approved):

PR #3943, opened 2024-12-29
macos: disable auto-updates for local builds

### LINKED ISSUE #3179: macOS: ability to disable auto-update on source-built builds

jparise:
The auto-update prompt isn't useful for local (source) builds. Disable it by default by setting Sparkle's SUEnableAutomaticChecks Info.plist key to NO (false) for all build configurations.

We then selectively re-enable it by deleting that Info.plist key from our release workflows. We delete the key instead of setting its value to YES (true) to give us Sparkle's default behavior of prompting the user to enable update checks on the second application launch. (YES tells Sparkle to skip that prompt and silently enable update checks.)

See also: https://sparkle-project.org/documentation/customization/

(This is a safer alternative to #3273.)

Fixes: #3179

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -110,6 +110,7 @@ jobs:
 
           # Updater
           /usr/libexec/PlistBuddy -c "Set :SUPublicEDKey $SPARKLE_KEY_PUB" "macos/build/Release/Ghostty.app/Contents/Info.plist"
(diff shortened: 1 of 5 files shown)


mitchellh:
Perfect than you.

PR #3945, opened 2024-12-29
add option to strip build regardless of optimization

Jan200101:
adds the option "strip" which can be used to override the default strip setting, which is based on the optimization mode.

Useful for a distro setting where you want a release build but still keep symbols.

Also reuses the option for the shared and static library

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -161,7 +161,7 @@ pub fn build(b: *std.Build) !void {
     const strip = b.option(
         bool,
         "strip",
-        "Build the website data for the website.",
+        "Strip the final executable. Default true for fast and small releases",
     ) orelse null;
 
     const conformance = b.option(
--- a/build.zig
+++ b/build.zig
@@ -162,7 +162,11 @@ pub fn build(b: *std.Build) !void {
         bool,
         "strip",
         "Strip the final executable. Default true for fast and small releases",
-    ) orelse null;
+    ) orelse switch (optimize) {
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -158,6 +158,12 @@ pub fn build(b: *std.Build) !void {
         "Build a Position Independent Executable. Default true for system packages.",
     ) orelse system_package;
 
+    const strip = b.option(
+        bool,
+        "strip",
+        "Build the website data for the website.",
```
jparise, line 164: I think this might have been a copy-and-pasted description. 😉

Jan200101: woops, I edited it locally but forgot to stage it :grimacing:

build.zig
```diff
@@ -158,6 +158,12 @@ pub fn build(b: *std.Build) !void {
         "Build a Position Independent Executable. Default true for system packages.",
     ) orelse system_package;
 
+    const strip = b.option(
+        bool,
+        "strip",
+        "Strip the final executable. Default true for fast and small releases",
+    ) orelse null;
```
Rexicon226, line 165: `orelse null` doesn’t make a lot of sense, since `option` already returns an optional.

i think the best solution is to move that switch into the orelse here.

Jan200101: good point, addressed in 061a730dd37188c6b6aca7eac66f8d0a9bc108b3

PR #3954, opened 2024-12-29
macos: restore 0 blur-radius when reloading config

gigamonster256:
If a blur radius config value was previously set but then removed or set to 0, the new blur radius would not take effect on config reload due to the early return clause.


mitchellh:
Perfect thanks

Issue #3899, opened 2024-12-29
macOS: Cmd + Shift + Plus should increase font size too


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3888

<div type='discussions-op-text'>

<sup>Originally posted by **thiagowfx** December 29, 2024</sup>
Out-of-the-box, the shortcut to increase/decrease font sizes in Ghostty on macOS is `Cmd + +` and `Cmd + -`, respectively (no Shift).

With Shift:

**Observed**: `Cmd + Shift + +` inserts a literal '+', and `Cmd + Shift + -` decreases the font size.

**Expected**: `Cmd + Shift + +` should increase the font size as well, consistently.

Why is the behavior for `Cmd + Shift + +` different? This seems unintentional.

Note that this is the out-of-the-box behavior in both Terminal.app and iTerm2.</div>

mitchellh:
Note: the solution here doesn't appear on the surface to be a missing binding (but may be). Some research is required why `cmd+shift+-` works but not plus on macOS currently.

alexhollinghead:
I'm not able to reproduce this on 15.1 Sequoia; both `CMD + Shift + +` and `CMD + Shift + -` adjust the font size for me. I do already have font size manually configured in my config file. Could this account for your issue?

curtbushko:
I can reproduce this on Sequoia 15.2.  

`Cmd + Shift + + ` gives me `=` in my shell.
`Cmd + Shift + - ` reduces the font size.


From: [Mac keyboard shortcuts](https://support.apple.com/en-ca/102650)

> Shift–Command–Minus sign (-): Decrease the size of the selected item.
>
> Shift–Command–Plus sign (+): Increase the size of the selected item. Command–Equal sign (=) performs the same function.```

Also, the `Cmd + Shift` shortcuts are not listed in the [Apple Terminal guide](https://support.apple.com/en-ca/guide/terminal/trmlshtcts/mac)

mdznr:
You should be able to just set ⌘+ as the menu item’s key equivalent and AppKit will do the keyboard localization for you for this well-known key equivalent. This is inspectable in other apps that have similar menu items (Safari, TextEdit, Preview, Terminal, and others).

I verified locally that removing:
```
    try result.keybind.set.put(
        alloc,
        .{ .key = .{ .translated = .equal }, .mods = inputpkg.ctrlOrSuper(.{}) },
        .{ .increase_font_size = 1 },
[...]

PR #3956, opened 2024-12-29
macOS: Cmd + Shift + Plus should increase font size too

curtbushko:
Addresses [#3899](https://github.com/ghostty-org/ghostty/issues/3899).

It looks like `Command + Shift + minus` is a mac built-in keybinding while `Command + Shit + plus` is not. In my testing, the combos work (Command + minus, Command + Shift + minus, Command + plus, Command + Shift + plus)

From: [Mac keyboard shortcuts](https://support.apple.com/en-ca/102650)

> Shift–Command–Minus sign (-): Decrease the size of the selected item.
>
> Shift–Command–Plus sign (+): Increase the size of the selected item. Command–Equal sign (=) performs the same function.```

(Note: I'm trying to find where iTerm2 does it in their code)

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2091,7 +2091,7 @@ pub fn default(alloc_gpa: Allocator) Allocator.Error!Config {
     // set the expected keybind for the menu.
     try result.keybind.set.put(
         alloc,
-        .{ .key = .{ .translated = .plus }, .mods = inputpkg.ctrlOrSuper(.{}) },
+        .{ .key = .{ .translated = .plus }, .mods = inputpkg.ctrlOrSuper(.{ .shift = true }) },
         .{ .increase_font_size = 1 },
     );
     try result.keybind.set.put(


mitchellh:
> (Note: I'm trying to find where iTerm2 does it in their code)

Beware that iTerm2 is GPL so if you find this, you can't copy the code.

curtbushko:
> Beware that iTerm2 is GPL so if you find this, you can't copy the code.

Thanks. I couldn't find the exact spot in their code anyway.

I think Config.zig is the place to put this as that was where you put the 'Natural text editing' keybinds the other day and it is kind of the same. I don't think we want it in Ghostty.Input.swift or in the swift code?

mdznr:
The key equivalents for these menu items are just `"+"`. AppKit does some menu item key equivalent localization by default.

peterjc:
Will this handle command with the number pad add/plus (which works on the Apple Terminal, Firefox, etc - see issue for details)?

curtbushko:
No, I don't think so because that is a different key.

mitchellh:
There's an ongoing discussion on how best to handle this so I'm going to close this for now and get back to it later. Thanks @curtbushko though!

PR #3970, opened 2024-12-29
copy_or_interrupt action

IrishBruse:
Windows terminal and kitty smart copy keybind action
Discussion here #2811

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3870,22 +3870,35 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
         },
 
         .copy_to_clipboard => {
-            // We can read from the renderer state without holding
-            // the lock because only we will write to this field.
-            if (self.io.terminal.screen.selection) |sel| {
-                const buf = self.io.terminal.screen.selectionString(self.alloc, .{
-                    .sel = sel,
-                    .trim = self.config.clipboard_trim_trailing_spaces,
-                }) catch |err| {
-                    log.err("error reading selection string err={}", .{err});
-                    return true;
-                };
+            if (copyToClipboard(self)) {
+                return true;
+            }
+        },
+
+        .copy_or_interrupt => {
+            if (copyToClipboard(self)) {
+                return true;
+            } else {
+                const buf = try self.alloc.alloc(u8, 1);
                 defer self.alloc.free(buf);
 
-                self.rt_surface.setClipboardString(buf, .standard, false) catch |err| {
-                    log.err("error setting clipboard string err={}", .{err});
-                    return true;
-                };
+                buf[0] = 3;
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/Window.zig
```diff
@@ -284,7 +284,7 @@ pub fn init(self: *Window, app: *App) !void {
         const toolbar_view: *c.AdwToolbarView = @ptrCast(c.adw_toolbar_view_new());
 
         const header_widget: *c.GtkWidget = self.header.?.asWidget();
-        c.adw_toolbar_view_add_top_bar(toolbar_view, header_widget);
```
jcollie: This seems like an unrelated change that should not be part of this PR.

jcollie (changes requested):
This will be easier to review if these two commits are squashed together. Especially the churn of adding/removing debug logging.

src/apprt/gtk/Window.zig
IrishBruse: Fixed

src/Surface.zig
```diff
@@ -3870,22 +3870,35 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
...
-                    log.err("error setting clipboard string err={}", .{err});
-                    return true;
-                };
+                buf[0] = 3;
+
+                self.io.queueMessage(try termio.Message.writeReq(
+                    self.alloc,
+                    buf,
+                ), .unlocked);
+
```
jcollie, line 3891: It would be slightly more efficient to directly use a small write request since we're only sending a single byte. See an example here:

https://github.com/ghostty-org/ghostty/blob/3059d9036b233d672e6dc7cdfb91c822978cc1ef/src/Surface.zig#L2646-L2658

src/Surface.zig
```diff
@@ -3870,22 +3870,35 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
...
+            if (copyToClipboard(self)) {
+                return true;
+            }
+        },
+
+        .copy_or_interrupt => {
+            if (copyToClipboard(self)) {
+                return true;
+            } else {
+                const buf = try self.alloc.alloc(u8, 1);
```
jcollie, line 3882: You can reduce the nesting level since the `then` branch of the `if` always returns.

src/Surface.zig
IrishBruse: Im not sure which I need .locked vs .unlocked ive it left as unlocked

src/Surface.zig
IrishBruse: Fixed

mitchellh:
Superseded by #4328, the accepted solution for that discussion. Thanks though!

PR #3976, opened 2024-12-30
app support config file now supersedes xdg when opening preferences on OSX

### LINKED ISSUE #3953: macOS: `open_config` action should prefer AppSupport over XDG
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3952

Our priority when loading configs is well documented: https://ghostty.org/docs/config#macos-specific-path-(macos-only):

But `open_config` does the opposite. This makes it too easy for people to have configs that are being overridden without them realizing it.

We should swap the logic in `open_config`.

MauriceElliott:
added conditional to edit.open so that when a config file exists in the app support directory on osx, it will prioritise opening that one over the one in the xdg config folder.

fixes: #3953

### DIFF
--- a/src/config/edit.zig
+++ b/src/config/edit.zig
@@ -13,11 +13,17 @@ pub fn open(alloc_gpa: Allocator) !void {
         if (comptime builtin.os.tag == .macos) macos: {
             // On macOS, use the XDG path if the app support path doesn't exist.
             const app_support_path = try internal_os.macos.appSupportDir(alloc_gpa, "config");
+
+            // If no configuration file currently exist, it should be created in app support.
+            var no_config_file = false;
+
             if (std.fs.accessAbsolute(app_support_path, .{})) {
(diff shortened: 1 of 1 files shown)


MauriceElliott:
fixes: #3953

MauriceElliott:
Nevermind, this is wrong as it will default to creating new configuration files in .config, and only opening the appsupport one if it already exists. Will fix and push.

MauriceElliott:
Now resolved.

The logic goes: 
if an app support config file exists, cmd+, will open it.
if a .config config file exists and no app support config file exists, cmd+, will the .config file.
if a .config config file exists and a app support config file exists, cmd+, will open the app support config file.
if neither exist, cmd+, will create a new config file in the app support directory.

mitchellh:
Superseded by #4004. I apologize, I was already working on this. I made one additional improvement over this PR (now that I looked at it): I prefer non-empty files over empty ones.

PR #3977, opened 2024-12-30
Correct comptime GTK atLeast() version comparison

iainh:
The comptime path of the GTK `atLeast()` version function fails to take the proceeding portion of the version into account. For example version 5.1.0 is incorrectly marked as less than 4.16.7 due to the minor version (1) being less than the minor we are comparing against (16).

This update required that the major versions be equal when comparing minor versions and the major and minor versions be equal when comparing micro versions.

For example, building against GTK 4.17.1:
Before: version.atLeast(4,16,2) -> false
After:  version.atLeast(4,16,2) -> true

### DIFF
--- a/src/apprt/gtk/version.zig
+++ b/src/apprt/gtk/version.zig
@@ -44,15 +44,15 @@ test "atLeast" {
     const std = @import("std");
     const testing = std.testing;
 
-    try testing.expectEqual(true, atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
+    try testing.expect(atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
 
-    try testing.expectEqual(false, atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION + 1));
-    try testing.expectEqual(false, atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION + 1, c.GTK_MICRO_VERSION));
-    try testing.expectEqual(false, atLeast(c.GTK_MAJOR_VERSION + 1, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
+    try testing.expect(!atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION + 1));
+    try testing.expect(!atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION + 1, c.GTK_MICRO_VERSION));
+    try testing.expect(!atLeast(c.GTK_MAJOR_VERSION + 1, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
 
-    try testing.expectEqual(true, atLeast(c.GTK_MAJOR_VERSION - 1, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
-    try testing.expectEqual(true, atLeast(c.GTK_MAJOR_VERSION - 1, c.GTK_MINOR_VERSION + 1, c.GTK_MICRO_VERSION));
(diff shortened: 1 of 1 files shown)


jcollie (changes requested):
Looks good, but it would be nice to get some unit tests so that this doesn't get broken in the future.

iainh:
I've added some unit tests which should cover the issue and the fix but unfortunately I don't have a strong enough grasp in zig/the build system to know if they are being executed.

Running `zig test` on the file directly fails due to the build_config module (referenced by apprt/gtk/c.zig) missing so I'm not sure if that is a viable method of running tests on a project of ghostty's complexity. I'd be glad for any advice you can offer.

mitchellh:
> I don't have a strong enough grasp in zig/the build system to know if they are being executed.

A trick I do is just add:

```
try testing.expect(false);
```

To ensure they're being run. And you can do `zig build test -Dtest-filter="name"` to make it much faster to filter to a subset of tests.

jcollie:
> I've added some unit tests which should cover the issue and the fix but unfortunately I don't have a strong enough grasp in zig/the build system to know if they are being executed.
> 
> Running `zig test` on the file directly fails due to the build_config module (referenced by apprt/gtk/c.zig) missing so I'm not sure if that is a viable method of running tests on a project of ghostty's complexity. I'd be glad for any advice you can offer.

Basically two ways:

1. Use `std.log.warn` to print something out (lower logging levels won't be shown).
2. Change the test so that it fails, and then change it back.

The logging should be removed before submitting the final PR, there's already too much logging from the tests.

`zig build test -Dtest-filter=atLeast` can be used to try and run just one specific test to speed up development.

src/apprt/gtk/version.zig
```diff
@@ -39,3 +39,20 @@ pub inline fn atLeast(
 
     return false;
 }
+
+test "atLeast" {
+    const std = @import("std");
+    const testing = std.testing;
+
+    try testing.expectEqual(true, atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
```
jcollie, line 47: If you're checking a boolean, you can do:
```
try testing.expect(atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION);
try testing.expect(!atLeast(c.GTK_MAJOR_VERSION + 1, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION);
```

jcollie (approved):

src/apprt/gtk/version.zig
iainh: Thanks for the suggestion, I've updated accordingly.

src/apprt/gtk/version.zig
```diff
@@ -44,15 +44,15 @@ test "atLeast" {
...
 
-    try testing.expectEqual(true, atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
+    try testing.expect(atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
 
-    try testing.expectEqual(false, atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION + 1));
-    try testing.expectEqual(false, atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION + 1, c.GTK_MICRO_VERSION));
-    try testing.expectEqual(false, atLeast(c.GTK_MAJOR_VERSION + 1, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
+    try testing.expectEqual(!atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION + 1));
+    try testing.expectEqual(!atLeast(c.GTK_MAJOR_VERSION, c.GTK_MINOR_VERSION + 1, c.GTK_MICRO_VERSION));
+    try testing.expectEqual(!atLeast(c.GTK_MAJOR_VERSION + 1, c.GTK_MINOR_VERSION, c.GTK_MICRO_VERSION));
```
jcollie, line 52: These should be `testing.expect` as well and not `testing.expectEqual`

jcollie (approved):

PR #3989, opened 2024-12-30
Add back quotes for variables in the zsh shell integration to improve compatibility

lixiaoyan:
https://github.com/ghostty-org/ghostty/pull/3332#issuecomment-2564526118
According to this comment, using variables without quoting may not work with `setopt SH_WORD_SPLIT`.

We don't need to quote variables in `[[ ]]` because it works differently.


mitchellh:
Thanks!

PR #3999, opened 2024-12-30
apprt/gtk: make window-decoration=false with gtk-titlebar=true look better

tristan957:
Before this change, there seemed to be some artifacting in the window corners due to the window border no longer outlining the content properly. By detecting the situation, we can turn the window border radius off.

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -972,6 +972,9 @@ fn loadRuntimeCss(
     const headerbar_foreground = config.@"window-titlebar-foreground" orelse config.foreground;
 
     try writer.print(
+        \\window.without-window-decoration-and-with-titlebar {{
+        \\  border-radius: 0 0;
+        \\}}
         \\widget.unfocused-split {{
         \\ opacity: {d:.2};
         \\ background-color: rgb({d},{d},{d});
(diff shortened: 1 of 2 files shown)


tristan957:
I'd love to provide pictures, but it is hard to capture. Any reviewer should test it for themselves.

mitchellh:
Thanks for the attention to detail. I trust you saw and fixed this.

tristan957:
See previous discussion in https://github.com/ghostty-org/ghostty/discussions/3290.

src/apprt/gtk/App.zig
```diff
@@ -972,6 +972,9 @@ fn loadRuntimeCss(
     const headerbar_foreground = config.@"window-titlebar-foreground" orelse config.foreground;
 
     try writer.print(
+        \\window.without-window-decoration-and-with-titlebar {{
+        \\  border-radius: 0 0;
+        \\}}
         \\widget.unfocused-split {{
```
jcollie, line 978: I think that this would be better in the static `style.css` since it never changes.

mitchellh: Probably right. I'd accept a follow up on it, but also fine with it here.

tristan957: Whoops my bad. I forgot about that. I will do that.

rhodes-b:
@tristan957 can confirm this does fix it

PR #4004, opened 2024-12-30
config: edit opens AppSupport over XDG on macOS, prefers non-empty paths

### LINKED ISSUE #3284: macOS: `open_config` action should use `open -t`

### LINKED ISSUE #3953: macOS: `open_config` action should prefer AppSupport over XDG

mitchellh:
Fixes #3953
Fixes #3284

This fixes two issues. In fixing one issue, the other became apparent so I fixed both in this one commit.

The first issue is that on macOS, the `open` command should take the `-t` flag to open text files in a text editor. To do this, the `os.open` function now takes a type hint that is used to better do the right thing.

Second, the order of the paths that we attempt to open when editing a config on macOS is wrong. Our priority when loading configs is well documented: https://ghostty.org/docs/config#macos-specific-path-(macos-only). But open_config does the opposite. This makes it too easy for people to have configs that are being overridden without them realizing it.

This commit changes the order of the paths to match the documented order. If neither path exists, we prefer AppSupport.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3195,7 +3195,7 @@ fn processLinks(self: *Surface, pos: apprt.CursorPos) !bool {
                 .trim = false,
             });
             defer self.alloc.free(str);
-            try internal_os.open(self.alloc, str);
+            try internal_os.open(self.alloc, .unknown, str);
         },
 
         ._open_osc8 => {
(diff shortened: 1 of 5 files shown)


mitchellh:
Okay this PR is fighting up a Zig 0.13 compiler bug (it's been a long time!) somehow in the `configPath` function.

PR #4026, opened 2024-12-30
cli/list_themes: Sort themes by background color

sin-ack:
This sorts theme background colors into 16 buckets, from dark to light in ascending order (with the name as the final tie-breaker as before). It's intended to prevent the jarring (epilepsy-inducing even?) flashes that happen when you hold down the down/up arrow in the list-themes UI. It's still not perfect since the foreground colors can vary wildly between themes, but it should be significantly better than the white-to-black (and vice-versa) flashes before this change.

Because we need to determine the theme's background color when sorting, this change moves the loading of the theme files to the initialization step instead of doing it just-in-time. To preserve the message box displayed when the theme cannot be read, we store the error message as part of the ThemeListElement.

Before (epilepsy warning):

https://github.com/user-attachments/assets/152b71d2-16e2-4faa-a125-64ef52687fde

After:

https://github.com/user-attachments/assets/15b8da3e-193f-4659-a48d-8af4650e5fc9

### DIFF
--- a/src/cli/list_themes.zig
+++ b/src/cli/list_themes.zig
@@ -33,9 +33,62 @@ const ThemeListElement = struct {
     location: themepkg.Location,
     path: []const u8,
     theme: []const u8,
-    rank: ?f64 = null,
+    config: Config,
+    config_error: ?anyerror,
+    rank: ?f64,
+
+    pub fn init(
+        alloc: std.mem.Allocator,
+        location: themepkg.Location,
+        path: []const u8,
+        theme: []const u8,
+    ) !ThemeListElement {
+        var self: ThemeListElement = undefined;
+
+        self.location = location;
+        self.rank = null;
+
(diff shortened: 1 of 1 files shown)


mitchellh:
The non-alphabetic left column is triggering me. I see the value in this but want to think about whether it should be a flag...

sin-ack:
I thought about adding an order option in addition to the existing filter but didn't know if the additional complexity is worth it. If you want I can do that as well.

mitchellh:
Thoughts @jcollie ?

mitchellh:
Sorry @sin-ack, I'm going to reject this. I haven't heard a vocal enough number of complaints about this to think the non-alphabetical list is worth it. I've heard plenty of praises (in general -- not of the sorting) about the theme picker though, so my judgement call is that this isn't negatively affecting too many people.

PR #4033, opened 2024-12-30
close_tab keybind (gtk apprt only)

lg28literconvectionmicrowaveoven:
Title. Adds a close_tab keybind that essentially behaves the exact same as clicking the tab close button on the tab bar.


lg28literconvectionmicrowaveoven:
Relates to #4331

mitchellh:
This works great. Thanks for the patience. Auto-merge on.

novadev94:
This also changes the default `cmd + w = close_surface` on macOS, which isn't ideal.

mitchellh:
Ah I forgot to clean that up. Thanks for pointing it out will look now.

mitchellh:
Resolved, it's now bound to `cmd+opt+w` to match iTerm2. It doesn't work yet though (other PR).

PR #4035, opened 2024-12-30
fix(gtk): fix issue detecting preferred color scheme on older systems

leigholiver:
On my system (pop-os 22.04 LTS) with system theme and seperate light and dark themes, ghostty always defaults to light mode. Switches between light and dark mode are properly handled.

In the logs this error is reported:

error(gtk): unable to get current color scheme: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such method “ReadOne”

The spec notes that the other functions are "[Deprecated, use ReadOne instead.](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html)" so using ReadOne is cerainly the correct path.

I've managed to fix this on my system by checking for the error and falling back to an implementation using the deprecated Read.

Discussion: https://github.com/ghostty-org/ghostty/discussions/3704
Issue: https://github.com/ghostty-org/ghostty/issues/4038

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -1403,7 +1403,15 @@ pub fn getColorScheme(self: *App) apprt.ColorScheme {
         null,
         &err,
     ) orelse {
-        if (err) |e| log.err("unable to get current color scheme: {s}", .{e.message});
+        if (err) |e| {
+            // If ReadOne is not yet implemented, fall back to deprecated "Read" method
+            // Error code: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such method “ReadOne”
+            if (e.code == 19) {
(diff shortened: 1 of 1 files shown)


mitchellh:
Well done, well written. Thank you.

radioactiveman:
For the Audacious media player I have also implemented similar code to follow the dark theme settings.

Your approach helped a lot to understand how the D-Bus calls work, even though I have not heard of Zig before. :slightly_smiling_face:
So thank you for that. You can see my implementation below. It checks the API version and uses `Read` or `ReadOne` conditionally which is cleaner and more convenient in my opinion.

https://github.com/audacious-media-player/audacious/commit/c53e40cd22ec53ef705516ef574776e64d54a90e

PR #4039, opened 2024-12-30
Handle short boolean flags in zsh/fish completions

### LINKED ISSUE #2992: Add short form boolean option handling to zsh/fish completions
Zsh and fish completions do long form completion for `--<key>=true` `--<key>=false`. Follow the idea in https://github.com/ghostty-org/ghostty/pull/2963/commits/79d2f508a9eb767f50ef2c30fbf5282357705e60 to add valueless form handling for zsh and fish.  `--<key>=true` -> `--<key>`

b6k-dev:
Closes: https://github.com/ghostty-org/ghostty/issues/2992

### DIFF
--- a/src/build/fish_completions.zig
+++ b/src/build/fish_completions.zig
@@ -56,7 +56,6 @@ fn writeFishCompletions(writer: anytype) !void {
         else {
             try writer.writeAll(if (field.type != Config.RepeatablePath) " -f" else " -F");
             switch (@typeInfo(field.type)) {
-                .Bool => try writer.writeAll(" -a \"true false\""),
                 .Enum => |info| {
                     try writer.writeAll(" -a \"");
                     for (info.fields, 0..) |f, i| {
@@ -114,7 +113,6 @@ fn writeFishCompletions(writer: anytype) !void {
             } else try writer.writeAll(" -f");
 
             switch (@typeInfo(opt.type)) {
-                .Bool => try writer.writeAll(" -a \"true false\""),
                 .Enum => |info| {
                     try writer.writeAll(" -a \"");
                     for (info.fields, 0..) |f, i| {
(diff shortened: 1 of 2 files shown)


mitchellh:
cc @anund @jparise for review if you can

src/build/fish_completions.zig
```diff
@@ -114,7 +113,6 @@ fn writeFishCompletions(writer: anytype) !void {
             } else try writer.writeAll(" -f");
 
             switch (@typeInfo(opt.type)) {
```
anund, line 115:
```suggestion
            switch (@typeInfo(opt.type)) {
               .Bool => {},
```

Explicitly do nothing rather than forget boolean values need no arguements?

src/build/zsh_completions.zig
```diff
@@ -47,50 +47,52 @@ fn writeZshCompletions(writer: anytype) !void {
...
-        else if (std.mem.eql(u8, "theme", field.name))
-            try writer.writeAll("_themes")
-        else if (std.mem.eql(u8, "working-directory", field.name))
-            try writer.writeAll("{_files -/}")
-        else if (field.type == Config.RepeatablePath)
-            try writer.writeAll("_files") // todo check if this is needed
-        else {
-            try writer.writeAll("(");
-            switch (@typeInfo(field.type)) {
-                .Bool => try writer.writeAll("true false"),
```
anund, line 63: It would be good to still handle the `.Bool` case in the switch like the suggestion above for fish and avoid the extra nesting. A different change would be needed roughly: 

- create a shared prefix ~`const equals_required="=-:::"` 
- add the prefix to all the cases where the previous the key has some required value

anund (changes requested):
Changes work :+1:, Two style requests.

src/build/fish_completions.zig
maciekbartczak: Agreed, explicit handling is better.

src/build/zsh_completions.zig
maciekbartczak: Implemented your suggestion.

anund (approved):

PR #4049, opened 2024-12-30
fix (macOS): Show quick terminal on full-screen app

### LINKED ISSUE #2721: macOS: Quick Terminal not visible when pressing global keybind and full-screen app is active
Discussed in https://github.com/ghostty-org/ghostty/discussions/2691

From testing iTerm2 (the only other macOS terminal I know that supports this), it appears it switches to another space and opens the quick terminal there. We should follow suit and do that.

sohsatoh:
closes #2721  

This PR resolves the issue where the Quick Terminal was not visible when pressing the global keybind while a full-screen app was active.  

### Changes  
- Added new configuration options for `quick-terminal-space-behavior`
  - The Quick Terminal will now overlay properly on top of full-screen applications

##### `quick-terminal-space-behavior = remain`

- The Quick Terminal will be remain open on the space when switching spaces.

##### `quick-terminal-space-behavior = move`

- The Quick Terminal will be moved to active space when switching spaces.

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -102,6 +102,7 @@
 		C159E89D2B69A2EF00FDFE9C /* OSColor+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = C159E81C2B66A06B00FDFE9C /* OSColor+Extension.swift */; };
 		C1F26EA72B738B9900404083 /* NSView+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = C1F26EA62B738B9900404083 /* NSView+Extension.swift */; };
 		C1F26EE92B76CBFC00404083 /* VibrantLayer.m in Sources */ = {isa = PBXBuildFile; fileRef = C1F26EE82B76CBFC00404083 /* VibrantLayer.m */; };
+		CFBB5FEA2D231E5000FD62EE /* QuickTerminalSpaceBehavior.swift in Sources */ = {isa = PBXBuildFile; fileRef = CFBB5FE92D231E5000FD62EE /* QuickTerminalSpaceBehavior.swift */; };
 		FC5218FA2D10FFCE004C93E0 /* zsh in Resources */ = {isa = PBXBuildFile; fileRef = FC5218F92D10FFC7004C93E0 /* zsh */; };
 		FC9ABA9C2D0F53F80020D4C8 /* bash-completion in Resources */ = {isa = PBXBuildFile; fileRef = FC9ABA9B2D0F538D0020D4C8 /* bash-completion */; };
 /* End PBXBuildFile section */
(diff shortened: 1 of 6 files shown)


sohsatoh:
I just noticed that the same fix is mentioned in the comment.  
Please feel free to close this PR if @thecrypticace's fix is better than mine.
https://github.com/ghostty-org/ghostty/issues/2721#issuecomment-2564935255

thecrypticace:
The main things I see here are:

1: (buggy behavior)
- Be on a desktop space
- Have zero windows open
- Open the quick terminal
- Switch to a fullscreen app / space
- Notice that it does not switch because of the "previous app" activation behavior

2: (buggy behavior)
- Sometimes this behavior also happens when switching between fullscreen apps / spaces
- for me it works fine when switching between safari / arc in fullscreen;
- but not when switching from Safari to Xcode

3: (idea)
The terminal probbly shouldn't dismiss when switching apps since it can stay open with this change. If that is done it'll need to be made key after the space switch has completed — I believe that takes a few turns of dispatch async. But it also seems fine to let it dismiss. The user can re-open it with a global hotkey.

3: (idea)
Should this "all spaces" behavior depend on whether or not the hotkey is a global hotkey? Since it can't be triggered with a local one when a full screen app is active. Though if the dismiss behavior is kept then this probably doesn't really matter.

sohsatoh:
@thecrypticace 

Thanks,

> 1, 2

Could you share the screen record and your setup?  I could not reproduce the issue on my setup.

https://github.com/user-attachments/assets/a2fcb06a-21fa-4f1f-9bf1-3a17b236b4ca

```
# font
font-family = Menlo
font-style = Regular
font-size = 18
font-thicken = true

# window
window-padding-x = 20
window-padding-y = 5
window-padding-balance = true
window-theme = ghostty
background-opacity = 0.75
background-blur-radius = 18

# quick terminal
quick-terminal-position = "center"
quick-terminal-autohide = false
keybind = global:cmd+ctrl+o=toggle_quick_terminal
```

> 3: (idea)
> The terminal probbly shouldn't dismiss when switching apps since it can stay open with this change. If that is done it'll need to be made key after the space switch has completed — I believe that takes a few turns of dispatch async. But it also seems fine to let it dismiss. The user can re-open it with a global hotkey.

Yes, that seems to be the ideal behavior.
However, there’s already an option `quick-terminal-autohide`. As mentioned in the PR, setting it to false allows the window to remain even when switching spaces. This behavior is quite close to what’s expected, and considering the name `quick-terminal-autohide` , I think it’s reasonable for the behavior to depend on this option. Therefore, I believe the current implementation is fine as it is.
(That said, when quick-terminal-autohide is set to true, the Quick Terminal display briefly glitches, so the fix you’re suggesting might still be necessary regardless.)
https://ghostty.org/docs/config/reference#quick-terminal-autohide


https://github.com/user-attachments/assets/6ae43662-fb10-4b1f-ac99-c018dbe5b104

> 4: (idea)
> Should this "all spaces" behavior depend on whether or not the hotkey is a global hotkey? Since it can't be triggered with a local one when a full screen app is active. Though if the dismiss behavior is kept then this probably doesn't really matter.

Yes, that’s the expected behavior.
It seems natural for it to work across all spaces only when it’s set as a global hotkey.

sohsatoh:
If you use .moveToActiveSpace instead of .canJoinAllSpaces, the quick terminal should stay in the current space even when you switch spaces.
This behavior seems similar to iTerm2, but I’m unsure which would feel more natural.

https://github.com/user-attachments/assets/9aac0520-298d-41a3-8609-eb465588318b

thecrypticace:
Ah only `quick-terminal-autohide = true` is affected (which is the default). when it's set to false it works fine. Going *from* Xcode to Safari works fine as well. It's the other way around that does not. I think it depends on having a previous app already set as well.

And Xcode seems to be involved in both scenarios and only bugged when it's running / debugging the app. If the app is launched outside of the Xcode debugger it works fine. So maybe this is a non-issue.

Finder -> Xcode:

https://github.com/user-attachments/assets/92d6b7af-1d97-4340-904f-610a5ce7f9d0

Safari <-> Xcode:

https://github.com/user-attachments/assets/a74b6a06-2c84-4011-8242-c320c9dd9b17

thecrypticace:
Personally, I think `.canJoinAllSpaces` is the more intuitive option. I'd always set iTerm to the "all spaces" option when I used it previously.

Either one is fine but usually if I trigger the quick terminal I want it to show, I type something, I close it and go back to what I'm doing. Space switching feels like it breaks me out of that flow even though technically you end up back on the same app after closing (just takes a second b/c of the animation).

sohsatoh:
@thecrypticace 
Thanks, I added the option `quick-terminal-space-behavior` to select `.canJoinAllSpaces` / `moveToActiveSpace`.

okuuva:
> @thecrypticace Thanks, I added the option `quick-terminal-space-behavior` to select `.canJoinAllSpaces` / `moveToActiveSpace`.

I think this is The Right Way™ here since it replicates iTerm behavior and it is the only terminal app on macOS that offers this functionality without resorting to third part apps and / or dirty hacks.

But exactly how much it should behave like in iTerm by default? I agree with @thecrypticace that the `.canJoinAllSpaces` is the more intuitive behavior, even though iTerm uses the `.moveToActiveSpace` equivalent by default. Personally I find the space switching jarring but to each their own.

sohsatoh:
@okuuva 
I have never encountered that issue. If you are using a build of the app different from the one you initially granted permission for, you will need to reauthorize Ghostty from Privacy&Security -> Accessibility. However, if the cause lies elsewhere, I don’t think it’s related to a bug in this implementation.

okuuva:
@sohsatoh 
> If you are using a build of the app different from the one you initially granted permission for, you will need to reauthorize Ghostty from Privacy&Security -> Accessibility.

That was it 🤦🏼 Thanks! And sorry for bothering 😓

mitchellh:
Excellent work! Thank you so much and thanks for the patience. I made some minor opinionated changes but the behavior is otherwise mostly retained.

PR #4081, opened 2024-12-30
bash: remove "request for experts" comment

jparise:
We now have a multiple folks who have pitched in to improve this script.


mitchellh:
Definitely thanks.

PR #4092, opened 2024-12-30
Deb package for debian-base distro

ronit18:
[DO NOT APPROVE TO RUN WORKFLOW] 
I have created this `deb/build.sh` script by which we can build a .deb package which will install ghostty binary in `/usr/bin/ghostty` and desktop app in `/usr/share/applications/ghostty`. The main problem with this is i need help in github-actions. I am not very good at it. 

I tried to run it in my repo and im getting this error, i looked the code in build.zig it checks whether its a git repo if yes then checks for `v1.0.0`, i tried everything still getting error in that. 
image

Need help in this part.


charyouz:
Can you try to output what the found version is, if it's not v1.0.0?

deb/build.sh
```diff
@@ -0,0 +1,92 @@
...
+[Desktop Entry]
+Version=1.0
+Name=Ghostty
+Comment=A fast, feature-rich, and cross-platform terminal emulator that uses platform-native UI and GPU acceleration.
+Exec=$APP_NAME
+Icon=$APP_NAME
+Terminal=true
+Type=Application
+Categories=Utility;System;
+EOF
```
rhodes-b, line 74: this is already packaged with ghostty

rhodes-b (commented):
You should read this, Ghostty does not plan on supporting packages built in the repo for Linux

https://ghostty.org/docs/install/binary#linux-(official)

deb/build.sh
ronit18: so this means i dont have to create a .desktop file myself, Just copy from this and use right?

ronit18:
> Can you try to output what the found version is, if it's not v1.0.0?

Yes. Following are the version injections i tried:
- Taking version from GITHUB_REF
```bash
 if [ -n "$GITHUB_REF" ]; then
    VERSION=$(echo "$GITHUB_REF" | sed 's|refs/tags/v||') # Extract version after 'v'
else
    VERSION="1.0.0"
fi
echo "Using version: $VERSION"
```
image

As you can see i added echo using version. and it is using 10.0.0 but still got error in build.zig

- Making a env from actions/workflow
```yml
- name: Print GitHub Tag
              run: |
                  echo "GitHub Ref: $GITHUB_REF"
                  echo "Tag: ${GITHUB_REF##*/}"

            - name: Build .deb Package
              env:
                  ZIG_GLOBAL_CACHE_DIR: /tmp/offline-cache
                  VERSION: ${GITHUB_REF##*/}

              run: |
                  chmod +x ./deb/build.sh
                  ./deb/build.sh
```

This will show tag in `Print Github tag` but inside script it wont show anything. a blank variable.

ronit18:
edit: i have removed deb/build.sh and integrated that whole code in actions.yml by taking code snippets from `.github/workflow/release-tag.yml`

ronit18:
edit: i have removed deb/build.sh and integrated that whole code in actions .yml by taking code snippets from `.github/workflow/release-tag.yml`

Also have figured out version error. now build is working but getting error of `bzip2`(i've already added this in dependency building' but still getting package not found. This error is only occuring when i try to build on workflow/actions

SCR-20241231-jyay

ronit18:
FInallyyyyyy got everything working😭, now its building on github-actions im getting a deb file for amd64 architecture.
Now have to figure out how can i merge deb packaging code with existing `.github/workflows/release-tag.yml`.

clayrisser:
I've packaged ghostty for debian if anyone wants to reference it.

https://github.com/clayrisser/debian-ghostty

The [`debian/unstable`](https://github.com/clayrisser/debian-ghostty/tree/debian/unstable/debian) branch contains the rules for compiling ghostty with zig. Because zig is not currently available on debian, I had to download zig in the rules during the debian build. This breaks the rules of proper debian packaging which makes this approach impossible to release on a debian repository (such as the openSUSE Build Service).

I worked around the above issues by using the pre-built binaries in the [`debian/bookworm`](https://github.com/clayrisser/debian-ghostty/tree/debian/bookworm/debian) branch. Until zig is supported on debian this is the only way to get ghostty in a debian repository.

https://github.com/clayrisser/debian-ghostty/tree/debian/bookworm/debian/prebuilt

ronit18:
Closing pr, because deb package is already made for ghostty.

PR #4110, opened 2024-12-30
gtk: don't use gtk_window_set_titlebar if adwaita is enabled but it's older than 1.4.0

jcollie:
Fix #4097

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -216,6 +216,14 @@ pub fn init(self: *Window, app: *App) !void {
         }
     }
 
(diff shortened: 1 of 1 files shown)


tristan957 (approved):
Thanks for fixing my mistake!

tristan957:
It would be good to mention the discussion too.

jcollie:
> It would be good to mention the discussion too.

Yeah, once I get confirmation from the folks in #4097 that this actually works for them I'll update the description with relevant info.

mitchellh:
Failing CI is a flappy test. The rest looks good. Just advise when you want me to merge.

jcollie:
OK, I've gotten two reports that this fixes the issue so it's ready to merge.

PR #4115, opened 2024-12-30
Enable bitmap font usage under CoreText

### LINKED ISSUE #2168: `+list-fonts` action should not list bitmap fonts
A couple of bitmap fonts, particularly those with the .otb format do not work in ghostty. Other bitmap fonts that use the .ttf file format do work though. 
So I suggest, at least until supporting .otb is a priority: remove any .otb fonts to be listed when running `ghostty +list-fonts` so as to not give false hopes

qwerasd205:
macOS bitmap-only fonts are a poorly documented format, which are often distributed as `.dfont` or `.dfon` files. They use a 'bhed' table in place of the usual 'head', but the table format is byte-identical, so enabling the use of bitmap-only fonts only requires us to properly fetch this table while calculating metrics.

ref: https://fontforge.org/docs/techref/bitmaponlysfnt.html

Reverts #3550 for obvious reasons, and may close issue #2168 because this should now mean that bitmap fonts are properly supported under both font backends due to #3837 - unless `otb` fonts are still not properly supported under FreeType (they are not supported under CoreText because CoreText does not know how to handle them).

I tested this change with the `.dfont` distribution of [Cozette](https://github.com/slavfox/Cozette) v1.25.2 and saw no visual issues.

### DIFF
--- a/src/font/discovery.zig
+++ b/src/font/discovery.zig
@@ -362,16 +362,9 @@ pub const CoreText = struct {
         const list = set.createMatchingFontDescriptors();
         defer list.release();
 
-        // Bring the list of descriptors in to zig land
-        var zig_list = try copyMatchingDescriptors(alloc, list);
-        errdefer alloc.free(zig_list);
-
-        // Filter them. We don't use `CTFontCollectionSetExclusionDescriptors`
-        // to do this because that requires a mutable collection. This way is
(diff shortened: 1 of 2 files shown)


mitchellh:
Excellent thank you. ❤️ CI failure is unrelated to you and is fixed on main.

PR #4128, opened 2024-12-30
macOS: weak self for event monitor to avoid retain cycle for controllers

### LINKED ISSUE #3219: Ghost-Windows on macOS with global `toggle-visibility` keybind.
Thanks for building Ghostty. I'm trying to migrate from iTerm2 which has a "Show/hide all windows with a system-wide hotkey". Here is my current config:

```
command = /opt/homebrew/bin/fish
theme = Builtin Light
font-family = "Fira Code"
keybind = global:ctrl+y=toggle_visibility
```

Every time I toggle visibility, it creates a non-interactive "ghost window" behind the current window. Here, I'm toggling between iTerm and Ghostty by pressing `ctrl+y`:

https://github.com/user-attachments/assets/bb42bedd-9985-4cbc-83ef-9017b9264ce2

Separately, when visibility is hidden using this shortcut, clicking the Ghostty icon in the macOS dock does not show the Ghostty windows until triggering toggle visibility action again.

Next, pressing the toggle visibility button sometimes creates tons of empty windows:
Image

Finally, when Ghostty gets into this state, `cmd+w` may close the session but leaves a ghost window behind:
Image

I'm using the latest versions of everything on macOS.

mitchellh:
Fixes #3219

We were holding a reference cycle to the base terminal controller. This was preventing the window from ever being fully deallocated.

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -111,8 +111,8 @@ class BaseTerminalController: NSWindowController,
         // Listen for local events that we need to know of outside of
         // single surface handlers.
         self.eventMonitor = NSEvent.addLocalMonitorForEvents(
-            matching: [.flagsChanged],
-            handler: localEventHandler)
+            matching: [.flagsChanged]
+        ) { [weak self] event in self?.localEventHandler(event) }
     }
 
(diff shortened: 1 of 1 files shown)


vpontis:
Thank you! This was my biggest pain point in migrating from iTerm

PR #4130, opened 2024-12-30
gtk: make sure that window-decoration is honored on all paths

jcollie:
Fix a regression from #4110 .

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -514,13 +514,19 @@ pub fn toggleWindowDecorations(self: *Window) void {
     const new_decorated = !old_decorated;
     c.gtk_window_set_decorated(self.window, @intFromBool(new_decorated));
 
+    // Fix any artifacting that may occur in window corners.
+    if (new_decorated) {
+        c.gtk_widget_add_css_class(@ptrCast(self.window), "without-window-decoration-and-with-titlebar");
+    } else {
+        c.gtk_widget_remove_css_class(@ptrCast(self.window), "without-window-decoration-and-with-titlebar");
+    }
+
     // If we have a titlebar, then we also show/hide it depending on the
     // decorated state. GTK tends to consider the titlebar part of the frame
     // and hides it with decorations, but libadwaita doesn't. This makes it
     // explicit.
-    if (self.header) |v| {
-        const widget = v.asWidget();
-        c.gtk_widget_set_visible(widget, @intFromBool(new_decorated));
+    if (self.header) |headerbar| {
+        headerbar.setVisible(new_decorated);
     }
 }
 
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -156,6 +156,9 @@ pub fn init(self: *Window, app: *App) !void {
     if (app.config.@"gtk-titlebar") {
         const header = HeaderBar.init(self);
 
+        // If we are not decorated then we hide the titlebar.
+        header.setVisible(app.config.@"window-decoration");
+
         {
             const btn = c.gtk_menu_button_new();
             c.gtk_widget_set_tooltip_text(btn, "Main Menu");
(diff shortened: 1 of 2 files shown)


adriandelgado:
I confirm that this fixes the regression.

artist-artix:
With this config the problem of the absent window decoration still reproduces:

```
keybind = ctrl+z=toggle_window_decorations
window-decoration = true
gtk-titlebar = false
```

The GTK titlebar is not shown: correct
The native WM decoration with its titlebar is not shown: incorrect

Pressing ctrl-z shows some minor change to the window border, but also then the WM decoration with its titlebar is not displayed.

jcollie:
> With this config the problem of the absent window decoration still reproduces:
> 
> ```
> keybind = ctrl+z=toggle_window_decorations
> window-decoration = true
> gtk-titlebar = false
> ```
> 
> The GTK titlebar is not shown: correct The native WM decoration with its titlebar is not shown: incorrect
> 
> Pressing ctrl-z shows some minor change to the window border, but also then the WM decoration with its titlebar is not displayed. And pressing ctrl-t to create a new tab crashes ghostty.

I don't _think_ that this is directly a result of this PR. I'm getting the following logs when I try and toggle window decorations:

```
debug(surface): key event binding flags=input.Binding.Flags{ .consumed = true, .all = false, .global = false } action=toggle_window_decorations
warning(gtk): unimplemented action=apprt.action.Action.Key.key_sequence
```

artist-artix:
Indeed this PR is not the cause, but it is related. The problem is originally from commit a7c93cd I think, as there are multiple commits modifying Window.zig it's not easy to determine, but if this PR is to "make sure that window-decoration is honored on all paths" the mentioned issue also needs to be solved.
Thank you.

jcollie:
> With this config the problem of the absent window decoration still reproduces:
> 
> ```
> keybind = ctrl+z=toggle_window_decorations
> window-decoration = true
> gtk-titlebar = false
> ```

Ah, I think I understand. When `gtk-titlebar=false` the titlebar will _never_ be shown, even when toggling window decorations. I'm not sure if that's a regression from previous behavior or not, those are not options that I use.

The decorations are toggling but it's very hard to see as it's basically just a shadow.

artist-artix:
I'm not sure how to address this, but without a fix for the regression ghostty is no longer usable in case gtk-titlebar=false.

jcollie:
> Indeed this PR is not the cause, but it is related. The problem is originally from commit [a7c93cd](https://github.com/ghostty-org/ghostty/commit/a7c93cdfb16b61137ce086f3b6eed9bf7b94d73e) I think, as there are multiple commits modifying Window.zig it's not easy to determine, but if this PR is to "make sure that window-decoration is honored on all paths" the mentioned issue also needs to be solved. Thank you.

Hmm... The user has explicitly told Ghostty that they do not want a titlebar via `--gtk-titlebar=false` and that's not what we are toggling. In any case, I don't think that this is going to be a easy/quick fix as we don't even create a titlebar if `--gtk-titlebar=false`.

artist-artix:
Indeed gtk-titlebar is false, but window-decoration is true so the native window title bar should be show just like it was before today;s commits.

jcollie:
I checked out 1.0.0 and tested:

`gtk-titlebar: false`
`window-decoration: true`

Titlebar was shown on launch but could be hidden with `toggle_window_decorations`

`gtk-titlebar: true`
`window-decoration: true`

Worked as expected.

`gtk-titlebar: true`
`window-decoration: false`

Worked as expected.

`gtk-titlebar: false`
`window-decoration: false`

`toggle_window_decorations` had no effect, titlebar and decorations never appeared

So I don't think that this was ever working 100% as one would expect.

jcollie:
@artist-artix Please open a new discussion about the interaction of `gtk-titlebar`, `window-decoration`, and `toggle_window_decorations`. It's broken now, but it was broken before, just in a different way. Part of the problem is that the interaction between these options isn't well defined/documented so the behavior has changed as different people have worked on the code. We'll use the new discussion to iron out those details and update the documentation and the code to match.

artist-artix:
I have opened https://github.com/ghostty-org/ghostty/discussions/4172

> I checked out 1.0.0 and tested:
> 
> gtk-titlebar: false
> window-decoration: true
> 
> Titlebar was shown on launch but could be hidden with toggle_window_decorations

This is exactly as expected and how it is supposed to work when reading the documentation. So in 1.0 this was NOT broken.

jcollie:
> This is exactly as expected and how it is supposed to work when reading the documentation. So in 1.0 this was NOT broken.

That is _not_ how I would expect things to work, and yes I've read the documentation. So as I said there's both documentation problems and code problems. Please open a new discussion so that we can figure out what the interactions should be between these options rather than continuing the discussion here.

PR #4139, opened 2024-12-31
font: allow non-boolean font feature settings

### LINKED ISSUE #3128: CoreText: parse features equivalently to HarfBuzz (currently we only parse binary features)
The CoreText shaper parses the provided font features using *overly* simplified assumptions about how font features are specified:

https://github.com/ghostty-org/ghostty/blob/7027147c7695064bf86f084e316e19dc604fcaaf/src/font/shaper/coretext.zig#L131-L140

The HarfBuzz shaper uses `Feature.fromString` which is a loose wrapper around `hb_feature_from_string`:

https://github.com/ghostty-org/ghostty/blob/7027147c7695064bf86f084e316e19dc604fcaaf/src/font/shaper/harfbuzz.zig#L48

Despite the CoreText `FeatureList.append`'s doc comments claiming to use "the same" feature syntax as HarfBuzz, this is not the case; the syntax for features passed to `hb_feature_from_string` can be found in the [HarfBuzz docs](https://harfbuzz.github.io/harfbuzz-hb-common.html#hb-feature-from-string).

The practical impact of this is that many configs that will work under HarfBuzz will not under CoreText, and it's actually *impossible* to specify non-boolean features (such as a specific character alternative).

I am willing to implement a parser with proper parity to the HarfBuzz one, it shouldn't be too difficult, but I figure'd I'd open this issue first to keep track of it since I might not get around to it for a couple days.

qwerasd205:
\+ much more flexible syntax and lenient parser
\+ allows comma-separated list as a single config value

This allows, e.g. `cv01 = 2` to select the second variant of `cv01`.

Resolves #3128 

Parser could probably be a little smaller than it is- would be a lot cleaner with the labeled switch continue pattern from Zig 0.14. Maybe should've put it in its own file too...

I spent *much* too long trying to test this with `cv01` with [monaspace](https://github.com/githubnext/monaspace) before realizing that the README refers to v1.2 but the latest released version (and hence the one I had installed) was v1.101 -- I installed the v1.2 version and tested with both CoreText and HarfBuzz and successfully set `cv01 = 2` and got the expected result.

Feel free to make any stylistic changes you feel necessary before merging.

### DIFF
--- a/src/font/shaper/harfbuzz.zig
+++ b/src/font/shaper/harfbuzz.zig
@@ -3,6 +3,10 @@ const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
 const harfbuzz = @import("harfbuzz");
 const font = @import("../main.zig");
+const terminal = @import("../../terminal/main.zig");
+const Feature = font.shape.Feature;
+const FeatureList = font.shape.FeatureList;
+const default_features = font.shape.default_features;
 const Face = font.Face;
 const Collection = font.Collection;
 const DeferredFace = font.DeferredFace;
@@ -10,7 +14,6 @@ const Library = font.Library;
 const SharedGrid = font.SharedGrid;
 const Style = font.Style;
 const Presentation = font.Presentation;
-const terminal = @import("../../terminal/main.zig");
 
 const log = std.log.scoped(.font_shaper);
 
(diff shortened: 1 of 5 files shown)


qwerasd205:
Actually I'll go ahead and fix those typos and extract the `Feature` stuff to its own file while I'm at it.

src/font/shaper/harfbuzz.zig
```diff
@@ -3,14 +3,17 @@ const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
 const harfbuzz = @import("harfbuzz");
 const font = @import("../main.zig");
+const terminal = @import("../../terminal/main.zig");
```
qwerasd205, line 6: ... *sigh* this is a copy/paste mistake. It can be removed. (I don't have time right now)

qwerasd205: Wait no it isn't I just reordered it from below. I'm too tired for this 🙃

mitchellh:
I'm going to hold this for after 1.0.1, just due to the type of change this is and risk 😄 Appreciate it though.

mitchellh (approved):
No comments this work is perfect.

PR #4140, opened 2024-12-31
doc: add background-opacity needs restart

ferrgo:
The section informing that some configs need a restart points that the specific config should have details about this. This adds the information that `background-opacity` on MacOS needs a restart
 
> Some configuration options cannot be reloaded at runtime; others may only apply to newly created terminals. See the specific configuration option for details.

_PS: lovin ghostty_

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -560,6 +560,7 @@ palette: Palette = .{},
 /// On macOS, background opacity is disabled when the terminal enters native
 /// fullscreen. This is because the background becomes gray and it can cause
 /// widgets to show through which isn't generally desirable.
+/// On macOs, this setting cannot be reloaded and needs a restart
 @"background-opacity": f64 = 1.0,
 
 /// A positive value enables blurring of the background when background-opacity
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -560,7 +560,7 @@ palette: Palette = .{},
 /// On macOS, background opacity is disabled when the terminal enters native
 /// fullscreen. This is because the background becomes gray and it can cause
 /// widgets to show through which isn't generally desirable.
-/// On macOs, this setting cannot be reloaded and needs a restart
+/// On macOS, changing this configuration requires restarting Ghostty completely.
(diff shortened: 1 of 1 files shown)


src/config/Config.zig
```diff
@@ -560,6 +560,7 @@ palette: Palette = .{},
 /// On macOS, background opacity is disabled when the terminal enters native
 /// fullscreen. This is because the background becomes gray and it can cause
 /// widgets to show through which isn't generally desirable.
+/// On macOs, this setting cannot be reloaded and needs a restart
```
akx, line 563:
```suggestion
/// On macOS, changing this configuration requires restarting Ghostty completely.
```
would echo the spelling of `macOS` used just above, and the verbiage used for e.g. `quick-terminal-position` :)

ferrgo: Thanks @akx ! I wrote macOs, but not sure if this is specific, may be necessary for linux as well, I'll try to get it tested, but if anyone knows how this goes on linux, I'd appreciate the inputs!

gigamonster256:
Are y'all sure changing background-opacity needs a restart to work? I doesn't for me, hot-reloads work just fine.

ferrgo:
> @ferrgo @akx Are y'all sure changing background-opacity needs a restart to work? I doesn't for me, hot-reloads work just fine. Is there a specific state you need to be in where the config reload does not work?

For me opening Ghostty without this setting and trying to enable would require a restart. If `background-opacity` is set and I try to change it's value then the reload is enough. Is this ur scenario? Try deleting the option on config file, restarting Ghostty and trying to set it again.

gigamonster256:
> PS.: Sorry for ghosttying u, have been off for a few days.

Haha no worries. You were right about weirdness happening if a window was created with an opacity of 1 and then adjusted but it looks like it doesn't matter anymore: https://github.com/ghostty-org/ghostty/pull/5083

ferrgo:
I see, indeed I had an issue that going fullscreen and back would not set opacity correctly, don't usually go fullscreen but testing now and its fixed indeed

PR #4159, opened 2024-12-31
fix: respect notch in MacOS non-native fullscreen mode

aslushnikov:
This patch adds notch padding to respect macOS notch in non-native fullscreen mode.

I do not have any experience with swift & swiftui, so please let me know if there's a better way to implement this!

References #378

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -333,7 +333,15 @@ class BaseTerminalController: NSWindowController,
         }
     }
 
-    func fullscreenDidChange() {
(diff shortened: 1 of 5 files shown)


aslushnikov:
@mitchellh hi Mitchell,

Is there anything I could do to have this landed?

mitchellh:
> Is there anything I could do to have this landed?

I'm sorry, the main blocker to this is I don't like the delegate change and it requires me to sit down and really think if that's the right approach. If you refactor this to not use that then it'd probably be merged sooner.

I'm not saying it'd be rejected in its current form, it just requires more time than I've had to dedicate to it. 😄

mitchellh:
Thanks for the work here. I felt #5750 is a better path forward.

PR #4171, opened 2024-12-31
termio/exec: fix 100% CPU usage after wait-after-command process exits

kdrag0n:
read(2) returning 0 means that the other end of the pipe/pty has been
closed (EOF), so there cannot be any more output to read on the pipe, and
the io reader thread can just exit.

If exec.wait_after_command=false, the read thread's quit pipe is
immediately written to after the child process dies, so all is well.
However, if wait_after_command=true (which is the case when using Ghostty
to run a .command/.sh file on macOS), the read thread keeps spinning,
causing persistent 100% CPU usage per exited process.

Fix it by exiting the reader thread on EOF.

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -179,8 +179,11 @@ pub fn threadExit(self: *Exec, td: *termio.Termio.ThreadData) void {
     // Quit our read thread after exiting the subprocess so that
     // we don't get stuck waiting for data to stop flowing if it is
     // a particularly noisy process.
-    _ = posix.write(exec.read_thread_pipe, "x") catch |err|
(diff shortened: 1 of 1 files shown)


mitchellh:
I haven't looked into it more closely, but I'm still seeing 100% CPU usage even with your changes in. 

Reproduction:

1. `zig build run -Dapp-runtime=glfw -- --wait-after-command=true`
2. Type `exit` in the launched window
3. See CPU usage

mitchellh:
Oh, my rebase... removed your change. Looking at it now.

mitchellh:
Alright this looks clean to me. Thank you

PR #4185, opened 2024-12-31
config: Add the option `toast_on_clipboard_copy`

YotamGurfinkel:
Add a config option to enable/disable the toast shown on clipboard copy

Also suggested in https://github.com/ghostty-org/ghostty/discussions/4165

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -894,7 +894,9 @@ fn gtkActionCopy(
         return;
     };
 
-    self.sendToast("Copied to clipboard");
+    if (self.app.config.@"toast-on-clipboard-copy") {
+        self.sendToast("Copied to clipboard");
+    }
 }
 
 fn gtkActionPaste(
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -894,7 +894,7 @@ fn gtkActionCopy(
         return;
     };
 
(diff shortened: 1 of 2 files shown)


mitchellh (approved):
I changed the config to `adw-toast` and to a packed struct. This will let us introduce other on/off in the future as we add more toasts. Packed struct configs also let you set it to "false" to disable all or "true" to enable all.

tristan957:
Maybe we should rename it to `in-app-notifications` and just document that it currently only works with GTK + Adwaita?

mitchellh:
> Maybe we should rename it to `in-app-notifications` and just document that it currently only works with GTK + Adwaita?

Yeah I considered this. But I'm not even sure how we're going to surface this information yet in macOS. It may be in such a way that doesn't make sense to ever disable. I'm not sure. But... there's really no downside to doing so so I'll rename it now.

jcollie:
> > Maybe we should rename it to `in-app-notifications` and just document that it currently only works with GTK + Adwaita?
> 
> Yeah I considered this. But I'm not even sure how we're going to surface this information yet in macOS. It may be in such a way that doesn't make sense to ever disable. I'm not sure. But... there's really no downside to doing so so I'll rename it now.

I think that `in-app-notifications` could be confused with OSC 9 (or 99 or...) notifications.

mitchellh:
I made #4460.

PR #4200, opened 2024-12-31
Implement loading custom css in the GTK app

### LINKED ISSUE #4089: GTK: ability to load custom CSS
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4088

<sup>Originally posted by **jcollie** December 30, 2024</sup>
Add a config entry `gtk-custom-css` that, when set, will load custom CSS from a file. The setting can be repeated to load multiple files and files can be marked as optional by prefixing it with a `?`. The custom CSS will be added as priority USER so it should take precedence over any CSS set by Ghostty or the system

</div>

b6k-dev:
Closes https://github.com/ghostty-org/ghostty/issues/4089
Gave it a shot and implemented the custom css loading.
My general idea is to use a provider for each stylesheet the user wants to load and then when the config changes unload them and create new providers.
A separate provider has to be used for each stylesheet the user wants to load, since when the provider loads the css it clears all the previously loaded styles, so in effect we cannot use one provider to load multiple stylesheets, but maybe there is a better way to overcome this limitation which I'm not seeing.

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -82,7 +82,7 @@ transient_cgroup_base: ?[]const u8 = null,
 css_provider: *c.GtkCssProvider,
 
 /// Providers for loading custom stylesheets defined by user
-custom_css_providers: std.ArrayList(*c.GtkCssProvider),
+custom_css_providers: std.ArrayListUnmanaged(*c.GtkCssProvider),
 
 /// The timer used to quit the application after the last window is closed.
 quit_timer: union(enum) {
@@ -428,7 +428,7 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         // our "activate" call above will open a window.
         .running = c.g_application_get_is_remote(gapp) == 0,
         .css_provider = css_provider,
-        .custom_css_providers = std.ArrayList(*c.GtkCssProvider).init(core_app.alloc),
+        .custom_css_providers = .{},
     };
 }
 
@@ -448,7 +448,7 @@ pub fn terminate(self: *App) void {
     for (self.custom_css_providers.items) |provider| {
         c.g_object_unref(provider);
     }
-    self.custom_css_providers.deinit();
+    self.custom_css_providers.deinit(self.core_app.alloc);
 
     self.config.deinit();
 }
@@ -909,11 +909,8 @@ fn syncConfigChanges(self: *App) !void {
             .{},
         ),
     };
-    self.loadCustomCss() catch |err| switch (err) {
-        error.OutOfMemory => log.warn(
-            "out of memory loading custom CSS, no custom CSS applied",
-            .{},
-        ),
+    self.loadCustomCss() catch |err| {
+        log.warn("Failed to load custom CSS, no custom CSS applied, err={}", .{err});
     };
 }
 
@@ -1048,14 +1045,10 @@ fn loadRuntimeCss(
     }
 
     // Clears any previously loaded CSS from this provider
-    c.gtk_css_provider_load_from_data(
-        self.css_provider,
-        buf.items.ptr,
-        @intCast(buf.items.len),
-    );
+    loadCssProviderFromData(self.css_provider, buf.items);
 }
 
-fn loadCustomCss(self: *App) Allocator.Error!void {
+fn loadCustomCss(self: *App) !void {
     const display = c.gdk_display_get_default();
 
     // unload the previously loaded style providers
@@ -1073,23 +1066,50 @@ fn loadCustomCss(self: *App) Allocator.Error!void {
             .optional => |path| .{ path, true },
             .required => |path| .{ path, false },
         };
-        std.fs.accessAbsolute(path, .{}) catch |err| {
+        const file = std.fs.openFileAbsolute(path, .{}) catch |err| {
             if (err != error.FileNotFound or !optional) {
                 log.err("error opening gtk-custom-css file {s}: {}", .{ path, err });
             }
             continue;
         };
+        defer file.close();
+
+        log.info("loading gtk-custom-css path={s}", .{path});
+        var buf_reader = std.io.bufferedReader(file.reader());
+        const contents = try buf_reader.reader().readAllAlloc(
+            self.core_app.alloc,
+            5 * 1024 * 1024 // 5MB
+        );
+        defer self.core_app.alloc.free(contents);
 
         const provider = c.gtk_css_provider_new();
-
         c.gtk_style_context_add_provider_for_display(
             display,
             @ptrCast(provider),
             c.GTK_STYLE_PROVIDER_PRIORITY_USER,
         );
-        c.gtk_css_provider_load_from_path(provider, path);
 
-        try self.custom_css_providers.append(provider);
+        loadCssProviderFromData(provider, contents);
+
+        try self.custom_css_providers.append(self.core_app.alloc, provider);
+    }
+}
+
+fn loadCssProviderFromData(provider: *c.GtkCssProvider, data: []const u8) void {
+    if (version.atLeast(4, 12, 0)) {
+        const g_bytes = c.g_bytes_new(data.ptr, data.len);
+        defer c.g_bytes_unref(g_bytes);
+
+        c.gtk_css_provider_load_from_bytes(
+            provider,
+            g_bytes
+        );
+    } else {
+        c.gtk_css_provider_load_from_data(
+            provider,
+            data.ptr,
+            @intCast(data.len),
+        );
     }
 }
 
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -82,7 +82,7 @@ transient_cgroup_base: ?[]const u8 = null,
 css_provider: *c.GtkCssProvider,
 
 /// Providers for loading custom stylesheets defined by user
-custom_css_providers: std.ArrayListUnmanaged(*c.GtkCssProvider),
+custom_css_providers: std.ArrayListUnmanaged(*c.GtkCssProvider) = .{},
 
 /// The timer used to quit the application after the last window is closed.
 quit_timer: union(enum) {
@@ -428,7 +428,6 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         // our "activate" call above will open a window.
         .running = c.g_application_get_is_remote(gapp) == 0,
         .css_provider = css_provider,
-        .custom_css_providers = .{},
     };
 }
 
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/App.zig
```diff
@@ -1040,6 +1055,44 @@ fn loadRuntimeCss(
...
+        const path, const optional = switch (p) {
+            .optional => |path| .{ path, true },
+            .required => |path| .{ path, false },
+        };
+        std.fs.accessAbsolute(path, .{}) catch |err| {
+            if (err != error.FileNotFound or !optional) {
+                log.err("error opening gtk-custom-css file {s}: {}", .{ path, err });
+            }
+            continue;
+        };
```
maciekbartczak, line 1081: I'm not sure if this is the place where this kind of validation should happen. I'm not that familiar with the codebase yet, but maybe this should be validated when the config is loaded, so that the message could be appended to `_diagnostics`?

jcollie: There's a bit of a race condition between the access test and the usage here. I was playing around with this and well and I think that the better option is to open and read the file using Zig functions and then use `gtk_css_provider_load_from_data` or `gtk_css_provider_load_from_bytes`.

We probably should have a helper function that takes a `[]const u8` and then uses `gtk_css_provider_load_from_data` or `gtk_css_provider_load_from_bytes` depending on GTK version (one is deprecated in 4.12 and one is not available until 4.12). That can then be used in all the places that we load CSS.

src/apprt/gtk/App.zig
```diff
@@ -81,6 +81,9 @@ transient_cgroup_base: ?[]const u8 = null,
 /// CSS Provider for any styles based on ghostty configuration values
 css_provider: *c.GtkCssProvider,
 
+/// Providers for loading custom stylesheets defined by user
+custom_css_providers: std.ArrayList(*c.GtkCssProvider),
+
```
jcollie, line 86: I'd recommend using a `std.ArrayListUnmanaged` instead.

src/apprt/gtk/App.zig
```diff
@@ -425,6 +428,7 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         // our "activate" call above will open a window.
         .running = c.g_application_get_is_remote(gapp) == 0,
         .css_provider = css_provider,
+        .custom_css_providers = std.ArrayList(*c.GtkCssProvider).init(core_app.alloc),
     };
```
jcollie, line 432: If a `std.ArrayListUnmanaged` used then it can have a default value of `.{}` and then doesn't need to be initialized here.

src/apprt/gtk/App.zig
jcollie: > maybe this should be validated when the config is loaded, so that the message could be appended to `_diagnostics`?

Perhaps when the config is loaded just do an access check and add a diagnostic if the file isn't optional, but I think that's not necessary for this PR.

maciekbartczak: Then I'll leave the scope of this PR as is and won't mess with the validation during config load.

src/apprt/gtk/App.zig
```diff
@@ -1073,23 +1066,50 @@ fn loadCustomCss(self: *App) Allocator.Error!void {
...
         };
+        defer file.close();
 
-        const provider = c.gtk_css_provider_new();
+        log.info("loading gtk-custom-css path={s}", .{path});
+        var buf_reader = std.io.bufferedReader(file.reader());
+        const contents = try buf_reader.reader().readAllAlloc(
+            self.core_app.alloc,
+            5 * 1024 * 1024 // 5MB
+        );
```
jcollie, line 1082: Should note in the config entry comments about the file size limit. I mean 5MiB is _probably_ enough, but it would be good to document that limitation.

src/apprt/gtk/App.zig
```diff
@@ -82,7 +82,7 @@ transient_cgroup_base: ?[]const u8 = null,
 css_provider: *c.GtkCssProvider,
 
 /// Providers for loading custom stylesheets defined by user
-custom_css_providers: std.ArrayList(*c.GtkCssProvider),
+custom_css_providers: std.ArrayListUnmanaged(*c.GtkCssProvider),
```
jcollie, line 86: You can give this a default value as well if you want.

src/apprt/gtk/App.zig
```diff
@@ -428,7 +428,7 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         // our "activate" call above will open a window.
         .running = c.g_application_get_is_remote(gapp) == 0,
         .css_provider = css_provider,
-        .custom_css_providers = std.ArrayList(*c.GtkCssProvider).init(core_app.alloc),
+        .custom_css_providers = .{},
     };
```
jcollie, line 432: Not necessary if you add a default value above, but it's more of a nitpick on my part.

src/apprt/gtk/App.zig
```diff
@@ -1073,23 +1066,50 @@ fn loadCustomCss(self: *App) Allocator.Error!void {
...
                 log.err("error opening gtk-custom-css file {s}: {}", .{ path, err });
             }
             continue;
         };
+        defer file.close();
 
-        const provider = c.gtk_css_provider_new();
+        log.info("loading gtk-custom-css path={s}", .{path});
+        var buf_reader = std.io.bufferedReader(file.reader());
+        const contents = try buf_reader.reader().readAllAlloc(
```
jcollie, line 1079: I think that the buffered reader is unnecessary.

src/apprt/gtk/App.zig
```diff
@@ -909,11 +909,8 @@ fn syncConfigChanges(self: *App) !void {
             .{},
         ),
     };
-    self.loadCustomCss() catch |err| switch (err) {
-        error.OutOfMemory => log.warn(
-            "out of memory loading custom CSS, no custom CSS applied",
-            .{},
-        ),
+    self.loadCustomCss() catch |err| {
+        log.warn("Failed to load custom CSS, no custom CSS applied, err={}", .{err});
```
jcollie, line 913: We are trying to be more explicit about catching exceptions, so unless something added dozens of errors we should keep the switch. It helps catch regressions if something accidentally starts returning more errors.

jcollie (changes requested):

src/apprt/gtk/App.zig
maciekbartczak: Added the info in the config.

src/apprt/gtk/App.zig
maciekbartczak: I implemented your suggestion, however i have some doubts - maybe there is a simpler way to handle the Reader errors. At first I thought handling all errors from an error set would be possible with one case, something like:
```
    self.loadCustomCss() catch |err| switch (err) {
        std.fs.File.Reader.Error => log.warn(
            "failed to load custom CSS, no custom CSS applied - encountered error while reading file: {}",
            .{err},
        ),
    };
```
I tried a few different ways to handle all the errors in one case, however the compiler complained and forced me to handle each error from the set separately, so i just did a one big case:
```
        std.fs.File.Reader.Error.InputOutput,
        std.fs.File.Reader.Error.SystemResources,
        std.fs.File.Reader.Error.IsDir,
        std.fs.File.Reader.Error.OperationAborted,
        std.fs.File.Reader.Error.BrokenPipe,
        std.fs.File.Reader.Error.ConnectionResetByPeer,
        std.fs.File.Reader.Error.ConnectionTimedOut,
        std.fs.File.Reader.Error.NotOpenForReading,
        std.fs.File.Reader.Error.SocketNotConnected,
        std.fs.File.Reader.Error.WouldBlock,
        std.fs.File.Reader.Error.AccessDenied => log.warn(
            "failed to load custom CSS, no custom CSS applied - encountered error while reading file: {}",
            .{err},
        ),
```
If this is not a standard way to handle error sets in switch statements please point me to the correct solution :smile:

src/apprt/gtk/App.zig
maciekbartczak: switched to a non-buffered reader

src/apprt/gtk/App.zig
maciekbartczak: Sure, initializing it here with a default value is fine.

src/apprt/gtk/App.zig
```diff
@@ -909,8 +908,33 @@ fn syncConfigChanges(self: *App) !void {
...
+        std.fs.File.Reader.Error.IsDir,
+        std.fs.File.Reader.Error.OperationAborted,
+        std.fs.File.Reader.Error.BrokenPipe,
+        std.fs.File.Reader.Error.ConnectionResetByPeer,
+        std.fs.File.Reader.Error.ConnectionTimedOut,
+        std.fs.File.Reader.Error.NotOpenForReading,
+        std.fs.File.Reader.Error.SocketNotConnected,
+        std.fs.File.Reader.Error.WouldBlock,
+        std.fs.File.Reader.Error.AccessDenied => log.warn(
+            "failed to load custom CSS, no custom CSS applied - encountered error while reading file: {}",
```
jcollie, line 935: OK, that's maybe a case that you don't need the switch, sorry!

src/apprt/gtk/App.zig
jcollie: Yeah, I didn't think that it would get that out of hand. Feel free to revert this part.

maciekbartczak: Reverted.

jcollie (approved):

src/config/Config.zig
```diff
@@ -1925,6 +1925,14 @@ keybind: Keybinds = .{},
 /// Adwaita support.
 @"gtk-adwaita": bool = true,
 
+/// Custom CSS files to be loaded.
+///
+/// This configuration can be repeated multiple times to load multiple files.
+/// Prepend a ? character to the file path to suppress errors if the file does
+/// not exist. If you want to include a file that begins with a literal ?
+/// character, surround the file path in double quotes (").
+@"gtk-custom-css": RepeatablePath = .{},
```
jcollie, line 1935: I was doing some testing and I had a crash if I didn't explicitly specify an absolute path to the custom CSS file, so for some reason the paths are not being expanded properly.

jcollie (changes requested):

src/config/Config.zig
rhodes-b: https://github.com/ghostty-org/ghostty/pull/3811 should fix expansion not sure about the crash

jcollie: > #3811 should fix expansion not sure about the crash

The crash is specifically because we have an assert in place to check for non-absolute paths. And #3811 won't fix it because I wasn't using a `~`, just a plain filename `--gtk-custom-css=custom.css`. `RepeatablePath` _should_ resolve that relative to Ghostty's config dir I thought but it obviously didn't.

mitchellh (approved):

src/config/Config.zig
jcollie: Actually, I think that I figured out my problem, if you specify `--background-image=custom.css` on the CLI, Ghostty tries to resolve that based on the `$PWD`. If it can't be resolved, I think that the code leaves it as is without any warnings and then we eventually get a panic. So that's not a blocker for this (even though Mitchell has already merged this).

PR #4217, opened 2024-12-31
Add tab title renaming feature to macos

### LINKED ISSUE #2509: Feature Request: ability to set title and colour of a tab

aswinmprabhu:
Tries to resolve #2509 partially by enabling tab title renaming feature for MacOS.

* Used the existing set_title action
* Added the set_title action to Bindings so that a keybind can be set for this feature (defaulted this to cmd+i -- same as iTerm2)
* 3 ways to trigger this feature: "Set Title" option in the right-click menu for the window, "Set Title" option in the "View" menu in the title bar, and the keybind (cmd+i default)
* Prevented esc sequences from the different programs from resetting the tab title once it is set manually using a new state var called `titleSetManually` in the SurfaceView
* The input box is a basic native macos dialog with "OK" and "Cancel". Leaving the title text box empty resets the title to automatic / from the `title` property in the config

**Demo**:

https://github.com/user-attachments/assets/5cacc389-4486-46c4-8cd5-dda347e9c663

Need some feedback. Thanks!


qwerasd205:
Haven't reviewed the code, but the demo looks good. I'd prefer if the UI were in a popover that drops down from the tab title, but that can come later, this is *fine* as it is.

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -352,7 +357,51 @@ extension Ghostty {
                 withTimeInterval: 0.075,
                 repeats: false
             ) { [weak self] _ in
-                self?.title = title
+                // Create a dialog and ask the uesr if
```
jbayer, line 360: Looks like a minor typo here

francescarpi:
I love this feature! It would also be nice to be able to change the title by double-clicking the tab title. This could be addressed in a separate PR, depending on the difficulty.
Thanks!

mitchellh:
Sorry for the delay! Great work with this. I made some modifications that I felt were important:

* SurfaceView keeps track of the last title libghostty sent, so when we reset the title we can immediately set it to the correct value. Previously, it waited for the next. 
* The above also removes the `reload_config` which we shouldn't ever call unless the config actually wants to be reloaded.
* I changed the binding action to `change_title_prompt`
* I also made an apprt action this can trigger

mitchellh (approved):

mitchellh (approved):

mitchellh (approved):

mitchellh (dismissed):

mitchellh:
> I'm going to bikeshed this a bit but what's the rationale for some things being called set_title_prompt and some things being called change_title_prompt

Yeah I will unify this.

> Also why is "prompt" part of the verb anyway?

Because the user is being prompted

jcollie:
> Because the user is being prompted

Then it should be `prompt...` and not `set_...`

mitchellh:
> Then it should be prompt... and not set_...

I'm good with this

macos/Sources/App/macOS/MainMenu.xib
```diff
@@ -232,6 +233,13 @@
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="L3L-I8-sqk"/>
+                            <menuItem title="Change Title..." id="24I-xg-qIq">
```
mdznr, line 236:
```suggestion
                            <menuItem title="Change Title…" id="24I-xg-qIq">
```

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -1117,6 +1166,8 @@ extension Ghostty {
             menu.addItem(.separator())
             menu.addItem(withTitle: "Reset Terminal", action: #selector(resetTerminal(_:)), keyEquivalent: "")
             menu.addItem(withTitle: "Toggle Terminal Inspector", action: #selector(toggleTerminalInspector(_:)), keyEquivalent: "")
+            menu.addItem(.separator())
+            menu.addItem(withTitle: "Change Title...", action: #selector(changeTitle(_:)), keyEquivalent: "")
```
mdznr, line 1170:
```suggestion
            menu.addItem(withTitle: "Change Title…", action: #selector(changeTitle(_:)), keyEquivalent: "")
```

mdznr (commented):
The ellipsis character should be used in menu items like these

https://support.apple.com/guide/applestyleguide/e-apsg076a7313/1.0/web/1.0#apd4768067dad604

mvolkmann:
This issue is marked as merged, but I have version 1.1.2 and when I right-click in a split in macOS I don't see the "Set Title" menu option. Is this change currently available for macOS?

mitchellh:
No it’s on tip. See the milestone this got attached to

felipevolpone:
I'm on `tip`, but I'm not able to setup a keybind for this. I've tried:
```
keybind = cmd+i=change_title_prompt
```

but when I reload the config, I get the `keybind: unknown error error.InvalidAction` error.
Am I missing something?

00-kat:
It's not `change_title_prompt` anymore, you want `prompt_surface_title`.

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -384,6 +385,7 @@ class AppDelegate: NSObject,
         syncMenuShortcut(config, action: "increase_font_size:1", menuItem: self.menuIncreaseFontSize)
         syncMenuShortcut(config, action: "decrease_font_size:1", menuItem: self.menuDecreaseFontSize)
         syncMenuShortcut(config, action: "reset_font_size", menuItem: self.menuResetFontSize)
+        syncMenuShortcut(config, action: "change_title_prompt", menuItem: self.menuChangeTitle)
```
00-kat, line 388: This line is very suspicious though (it's still like this on the current main); every other instance is `prompt_surface_title`. (Corrected in #7149.)

felipevolpone:
Thanks, that made it 🙌 

One small feedback [1]: At least in mac os, when the popup appears the focus of the cursor is not on the text field, but on the "Cancel" button, which is kinda weird.

Not sure if it's possible to see with this screenshot:
Screenshot 2025-04-16 at 08 43 43

[1] If this is not the place for it, I'm happy to open an issue for it.

00-kat:
Yeah, open a new Discussion.

felipevolpone:
For reference: https://github.com/ghostty-org/ghostty/discussions/7105

SachaG:
I'm running version 1.1.3 and not seeing a way to trigger the dialog. Has this been rolled back since? Is this feature documented anywhere?

jcollie:
> I'm running version 1.1.3 and not seeing a way to trigger the dialog. Has this been rolled back since? Is this feature documented anywhere?

It is not part of 1.1.3 or any other released version. It will be part of 1.2 when it is released. Please read the comments more carefully before commenting on such an old closed PR.

SachaG:
@jcollie sorry but it's not obvious which release this feature is attached to at first glance, so I think given that the PR was approved a couple months ago it's reasonable to assume the feature might already be part of the latest public release. Looking forward to 1.2 though :)

jcollie:
> @jcollie sorry but it's not obvious which release this feature is attached to at first glance, so I think given that the PR was approved a couple months ago it's reasonable to assume the feature might already be part of the latest public release. Looking forward to 1.2 though :)

Check the "milestone" that this PR is attached to. It's up near the top on the right. It's even mentioned in a comment above.

mitchellh:
CleanShot 2025-06-24 at 22 04 20@2x

You can always tell where an issue or PR made it via this. This is managed by CI.

PR #4225, opened 2024-12-31
Set the paste button as default in the GTK paste confirmation dialog

danudey:
A simple change to make it so that, in the GTK4 paste confirmation dialog, the user can just hit enter/space to confirm the paste.

After some playing around, it seems as though GTK4 needs you to set the focus on a widget after the entire view has been configured, meaning that we have to create more references to pass the `confirm_button` GtkWidget up and up. Not 100% certain if this is the best way to do so, but:

1. Add the `ButtonsView` to the `PrimaryView` struct (as `PrimaryView.buttons`)
2. Add the `confirm_button` to the `ButtonsView` struct (as `ButtonsView.confirm_button`)
3. Call `c.gtk_widget_grab_focus` on (the now-accessible) `view.buttons.confirm_button`

This seems to work as expected, but I'm not sure if:

1. We should also make `cancel_button` available?
2. There's a better way to expose `confirm_button` to `PrimaryView`?
3. I did a good Zig?

I've never written (or read) Zig code before today so I hope this makes sense. Feedback welcome!

### DIFF
--- a/src/apprt/gtk/ClipboardConfirmationWindow.zig
+++ b/src/apprt/gtk/ClipboardConfirmationWindow.zig
@@ -89,7 +89,7 @@ fn init(
     const view = try PrimaryView.init(self, data);
     self.view = view;
     c.gtk_window_set_child(@ptrCast(window), view.root);
-    _ = c.gtk_widget_grab_focus(view.buttons.confirm_button);
+    _ = c.gtk_widget_grab_focus(view.buttons.cancel_button);
 
     c.gtk_widget_show(window);
 
@@ -162,6 +162,7 @@ const PrimaryView = struct {
 const ButtonsView = struct {
     root: *c.GtkWidget,
     confirm_button: *c.GtkWidget,
+    cancel_button: *c.GtkWidget,
 
     pub fn init(root: *ClipboardConfirmation) !ButtonsView {
         const cancel_text, const confirm_text = switch (root.pending_req) {
@@ -174,9 +175,6 @@ const ButtonsView = struct {
 
         const confirm_button = c.gtk_button_new_with_label(confirm_text);
         errdefer c.g_object_unref(confirm_button);
-
-        // TODO: Focus on the paste button
-        _ = c.gtk_widget_grab_focus(confirm_button);
 
         // Create our view
         const view = try View.init(&.{
@@ -202,7 +200,7 @@ const ButtonsView = struct {
             c.G_CONNECT_DEFAULT,
         );
 
-        return .{ .root = view.root, .confirm_button = confirm_button };
+        return .{ .root = view.root, .confirm_button = confirm_button, .cancel_button = cancel_button };
     }
 
     fn gtkCancelClick(_: *c.GtkWidget, ud: ?*anyopaque) callconv(.C) void {
--- a/src/apprt/gtk/ClipboardConfirmationWindow.zig
+++ b/src/apprt/gtk/ClipboardConfirmationWindow.zig
@@ -89,6 +89,8 @@ fn init(
     const view = try PrimaryView.init(self, data);
     self.view = view;
     c.gtk_window_set_child(@ptrCast(window), view.root);
+    _ = c.gtk_widget_grab_focus(view.buttons.confirm_button);
+
     c.gtk_widget_show(window);
 
     // Block the main window from input.
@@ -104,6 +106,7 @@ fn gtkDestroy(_: *c.GtkWidget, ud: ?*anyopaque) callconv(.C) void {
 const PrimaryView = struct {
     root: *c.GtkWidget,
     text: *c.GtkTextView,
+    buttons: ButtonsView,
 
     pub fn init(root: *ClipboardConfirmation, data: []const u8) !PrimaryView {
         // All our widgets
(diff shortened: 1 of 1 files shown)


jcollie:
I haven't looked at the code, but I'm against this change. The whole point of putting up a confirmation dialog is to get the user to stop and think before continuing. If you just can just blindly click through it kind of defeats the point.

If you don't want to be bothered with the window, there are the `clipboard-paste-protection` and `clipboard-paste-bracketed-safe` settings that can disable the confirmation button entirely.

danudey:
I had originally thought the same thing, but I did notice that in `src/apprt/gtk/ClipboardConfirmationWindow.zig` line 172 (on mobile so I can't get the link easily) there's a TODO for this exact functionality, so I assumed it was intended but not yet implemented.

Also, the paste dialog still serves its original purpose - not having the user inadvertently dump a bunch of lines of text into the terminal if they didn't mean to - but now allows the user to easily confirm. As someone who often pastes into a terminal, being able to easily confirm is important, but, as someone who often pastes into a terminal, making sure I'm pasting the right thing when I mean to is also important.

My preference for this functionality would be, similar to Kitty, separate keys bound to actions. In kitty, you hit P to confirm paste, for example, which is less likely to be hit accidentally.

I'm open to suggestions on how better to implement this, but I think the intent of the dialog is a confirmation of what you're doing and not a flow-breaking barrier to overcome every time you paste.

mitchellh:
cc @tristan957 for any feedback on this. I don't really have a super strong opinion other than we should be consistent.

tristan957:
Haven't reviewed the video, but I would just test what other GTK apps do for similar type dialogs. For instance, which action is focused when you try to exit GNOME Text Editor without saving your file. Also, we should be internally consistent. What button is focused if you try to close a surface that is currently running a program and Ghostty asks you to confirm?

jcollie:
> For instance, which action is focused when you try to exit GNOME Text Editor without saving your file?

Gnome Text Editor focuses the "Save" button.

image

tristan957:
Thanks. This tells me that we should indeed focus the "affirmative" action.

danudey:
> Thanks. This tells me that we should indeed focus the "affirmative" action.

Not to argue against my own patch, but I think it could be argued that we should focus the 'safe' action - which saving a document is. Likewise if Ghostty detects that you have a program running in a tab and you close the window, the default choice for the prompt is the 'safe' action - to not close the window.

image

In view of this, I've changed the PR to instead focus the cancel button. Now the workflow when someone pastes is:

1. The window pops up. They can hit space/enter to cancel or (Tab, space) to confirm
2. The default if they hit space/enter by accident is the safe option - no harm, no action
3. If they cancel by accident they can just paste again

I think this solves all the points raised:

1. Pasting was kind of a pain (my original pain point)
2. Accidentally pasting and then accidentally confirming was too easy (a good counterpoint)
3. We default to the safe option, as we do in other places, for consistency

Feedback on this new approach?

jcollie:
> > Thanks. This tells me that we should indeed focus the "affirmative" action.
> 
> Not to argue against my own patch, but I think it could be argued that we should focus the 'safe' action - which saving a document is. Likewise if Ghostty detects that you have a program running in a tab and you close the window, the default choice for the prompt is the 'safe' action - to not close the window.
> 
> image
> 
> In view of this, I've changed the PR to instead focus the cancel button. Now the workflow when someone pastes is:
> 
>     1. The window pops up. They can hit space/enter to cancel or (Tab, space) to confirm
> 
>     2. The default if they hit space/enter by accident is the safe option - no harm, no action
> 
>     3. If they cancel by accident they can just paste again
> 
> 
> I think this solves all the points raised:
> 
>     1. Pasting was kind of a pain (my original pain point)
> 
>     2. Accidentally pasting and then accidentally confirming was too easy (a good counterpoint)
> 
>     3. We default to the safe option, as we do in other places, for consistency
> 
> 
> Feedback on this new approach?

Making the "safe" option the default is good with me.

jcollie:
I thought that we had come to the consensus that the "safe" option should be the default, so that should be the "cancel" button.

danudey:
> I thought that we had come to the consensus that the "safe" option should be the default, so that should be the "cancel" button.

Sorry, the commits are confusing; because I rebased, it shows both of my commits - the one that sets confirm as default, and then the one (less clear) that sets cancel as default instead, overriding the first.

jcollie:
The "cancel" button should have the "suggested-action" CSS class added to it, and the "paste" button should have the "destructive-action" CSS class added to it.

jcollie:
> > I thought that we had come to the consensus that the "safe" option should be the default, so that should be the "cancel" button.
> 
> Sorry, the commits are confusing; because I rebased, it shows both of my commits - the one that sets confirm as default, and then the one (less clear) that sets cancel as default instead, overriding the first.

Yes, I saw that after I sent the message, sorry.

jcollie (approved):

danudey:
Added the new classes. The 'suggested' button looks weird (the dark grey makes it look disabled?) but that could just be my current GTK theme.

image

mitchellh (approved):

jcollie:
> Added the new classes. The 'suggested' button looks weird (the dark grey makes it look disabled?) but that could just be my current GTK theme.

Yeah, I'd check without a custom GTK theme.

jcollie:
Looks pretty good with the default dark theme!

Screenshot From 2025-01-06 14-10-06

PR #4226, opened 2024-12-31
Add background image support for OpenGL

### LINKED ISSUE #3645: Window background image configuration option
Discussion: #3639 

Add a new configuration option: `background-image` which takes a file path to an image to set as the terminal background. 

`background-opacity` should affect this. 

We can add additional configurations such as whether to stretch it etc. later. For the first version, we should always stretch it to fit the window.

I suggest implementing this in the renderers rather than the apprts because it'll make it easier to make cross-platform and consistent.

yunusey:
Partially fixes #3645.

https://github.com/user-attachments/assets/8d61e979-3cad-481a-81fc-bc8a3f8a9573

_I apologize for the video quality - GitHub limits videos to be less than 10MB._

This PR adds support for setting background image dynamically for OpenGL only. Metal still needs its own implementation.

Because the only way to read the image data and decode to RGB/RGBA so that it is ready to be uploaded to the GPU is to use `kitty.terminal.graphics` with the project's current state, we only support PNG.


src/renderer/OpenGL.zig
```diff
@@ -2439,6 +2562,46 @@ fn drawCellProgram(
...
 
+fn drawBackgroundImage(
+    self: *OpenGL,
+    gl_state: *const GLState,
+) !void {
+    // If we don't have a background image, just return
+    if (self.current_background_image == null) {
+        return;
+    }
+    // Bind our background image program
```
jcollie, line 2573: The Zig way to deal with optionals in a situation like this is to "strip" the optional like this:

```
const current_background_image = self.current_background_image orelse return;
```
Then you can refer to `current_background_image` without needing the awkward `.?` syntax.

src/renderer/OpenGL.zig
```diff
@@ -2305,6 +2400,31 @@ pub fn drawFrame(self: *OpenGL, surface: *apprt.Surface) !void {
         }
     }
 
+    // Check if we need to update our current background image
+    if (self.current_background_image != null) {
+        switch (self.current_background_image.?) {
+            .ready => {},
```
jcollie, line 2406: I'd recommend:
```
if (self.current_background_image) |current_background_image| {
  switch (current_background_image) {
```

src/renderer/OpenGL.zig
```diff
@@ -1199,6 +1235,57 @@ fn prepKittyImage(
...
+                .image_id = 0,
+            },
+        },
+        .data = try self.alloc.dupe(u8, path),
+    };
+    defer command.deinit(self.alloc);
+
+    // Load the iamge
+    var loading = try terminal.kitty.graphics.LoadingImage.init(self.alloc, &command);
+    defer loading.deinit(self.alloc);
```
jcollie, line 1264: Probably not in this commit, but the graphics loading code should be factored out of the kitty graphics code as it's pretty awkward to use here. We can also add JPEG support.

src/renderer/OpenGL.zig
yunusey: Oops, thanks for catching that! I will fix it!

src/renderer/OpenGL.zig
yunusey: I agree! I just didn't want to implement a whole system for that - wanted to make sure that this is what we want to do. And I totally agree that we should add JPEG support - it is very crucial!

ShivamRawat0l (approved):

src/renderer/opengl/BackgroundImageProgram.zig
```diff
@@ -0,0 +1,121 @@
...
+const gl = @import("opengl");
+
+pub const Input = extern struct {
+    /// vec2 terminal_size
+    terminal_width: u32 = 0,
+    terminal_height: u32 = 0,
+
+    /// uint mode
+    mode: BackgroundMode = .aspect,
+};
```
jcollie, line 14: This should be moved up a directory level so that I can be shared with the Metal renderer.

src/renderer/OpenGL.zig
jcollie: Hmm looking at this, we really don't want to drag the Kitty image protocol code into this. The PNG decoding code we have is pretty simple to use, it's basically just:

```
const wuffs = @import("wuffs");
const pngdata = <read data>;
const imgdata = try wuffs.png.decode(allocator, pngdata);
```
You'll end up with a struct that has the image height, width and data in RGBA format which I think is what the OpenGL code is expecting anyway.

I pushed PR #4250 which adds JPEG decoding support so that should soon be available.

src/config/Config.zig
```diff
@@ -403,6 +404,24 @@ background: Color = .{ .r = 0x28, .g = 0x2C, .b = 0x34 },
 /// Foreground color for the window.
 foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 
+/// Background image for the window.
+@"background-image": RepeatablePath = .{},
```
jcollie, line 408: This should not be a `RepeatablePath` as only one image can ever be shown.

jcollie (changes requested):
This is starting to look very good! I'm sure others will have more to say about the rendering code as they have more experience, but I like it!

Screenshot From 2025-01-01 21-32-10

src/renderer/opengl/BackgroundImageProgram.zig
```diff
@@ -0,0 +1,121 @@
...
+
+pub const BackgroundMode = enum(u8) {
+    aspect = 0,
+    scaled = 1,
+    _,
+};
```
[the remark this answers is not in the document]
yunusey: I will work on the first one; I think you are completely right, it doesn't make any sense for the mode to be defined there (I was just following `ImageProgram`'s implementation, but we are dealing with different things here, so it should be moved to an upper level module).

For the second one, I was planning to add more modes. It is just that the issue says, for the first version, we need the image to be scaled, and I didn't want to take initiative before getting more insight, but I will try implementing them now.

src/config/Config.zig
```diff
@@ -403,6 +404,24 @@ background: Color = .{ .r = 0x28, .g = 0x2C, .b = 0x34 },
...
+
+/// Background image opactity
+@"background-image-opacity": f32 = 0.0,
+
```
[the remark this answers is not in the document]
yunusey: Nice catch, will do!

src/config/Config.zig
yunusey: I also thought about the type we should use for `background-image`. Since, currently, all the config keys related to a path are handled with `RepeatablePath`, I just followed the patterns. What would be your recommendation to use instead of `RepeatablePath`, `[]u8` directly?

Oh, I didn't know about `wuffs`, I will try using it instead. Thanks a lot for letting me know!

jcollie: We probably should have a `Path` struct that basically does the same stuff as `RepeatablePath` except it only allows one path.

src/renderer/opengl/BackgroundImageProgram.zig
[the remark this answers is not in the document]
yunusey: Done! Check out the new modes below:

## Centered
Centered
## Stretched
Stretched
## Tiled
Tiled
## Zoomed
Zoomed

src/config/Config.zig
yunusey: So, apparently, `Path` is used in `Config.zig` as an enum inside `RepeatablePath`, so I thought I would name it `SinglePath`, but if you have any other recommendations, we can most definitely change its name. Also, I thought it would be better if we defined `BackgroundImageMode` directly in `Config.zig` and pass it to lower modules, but if you think we should define it in `renderer`, we can change it as well.

yunusey: And I got rid of `kitty.graphics` stuff - we are using `wuffs` directly, and I think it will be easier for us to integrate `JPEG` with your PR.

src/config/Config.zig
```diff
@@ -4113,6 +4113,84 @@ pub const Palette = struct {
     }
 };
 
+/// Path is a path to a single file.
+pub const SinglePath = struct {
+    const Self = @This();
+
+    /// The actual value that is updated as we parse.
+    value: []const u8 = "",
+
```
jcollie, line 4122: This should be an optional and default to null.

src/config/Config.zig
```diff
@@ -4113,6 +4113,84 @@ pub const Palette = struct {
...
+        log.debug(
+            "expanding file path relative={s} abs={s}",
+            .{ path, abs },
+        );
+
+        self.value = try alloc.dupeZ(u8, abs);
+    }
+};
+
```
jcollie, line 4193: This looks good, except that you need to check for SinglePath here:

https://github.com/ghostty-org/ghostty/blob/600e4171544f2ed29ba7a95afe1a10bde51b0030/src/config/Config.zig#L3055-L3074

src/renderer/OpenGL.zig
```diff
@@ -1244,52 +1247,63 @@ fn prepKittyImage(
...
-        .rgba => .{ .pending_rgba = pending },
-        .png => unreachable, // should be decoded by now
+    // Store the image
+    self.current_background_image = .{ .pending_rgba = pending };
+}
+
+/// Reads the content of the given image path and returns it
+pub fn readImageContent(self: *OpenGL, path: []const u8) ![]u8 {
+    // Open the file
+    var file = std.fs.cwd().openFile(path, .{}) catch |err| {
```
jcollie, line 1277: There should be an assert here checking for an absolute path and then using `std.fs.openFileAbsolute` to open the file.

src/renderer/shaders/bgimage.f.glsl
```diff
@@ -1,13 +1,25 @@
 #version 330 core
 
+const uint MODE_ZOOMED = 0u;
+const uint MODE_STRETCHED = 1u;
+const uint MODE_TILED = 2u;
+const uint MODE_CENTERED = 3u;
+
```
jcollie, line 7: I don't know if we can do `#include` in GLSL but if we can we should do that so that the modes never get out of sync between the two files.

src/renderer/OpenGL.zig
```diff
@@ -1244,52 +1247,63 @@ fn prepKittyImage(
...
+    // Read the file
+    var managed = std.ArrayList(u8).init(self.alloc);
+    errdefer managed.deinit();
+    const size: usize = max_image_size;
+    reader.readAllArrayList(&managed, size) catch |err| {
+        log.warn("failed to read file: {}", .{err});
+        return error.InvalidData;
     };
+
+    return managed.items;
```
jcollie, line 1306: Don't return `managed.items` directly, return `managed.toOwnedSlice()`.

src/renderer/OpenGL.zig
```diff
@@ -1244,52 +1247,63 @@ fn prepKittyImage(
...
+    }
+
+    var buf_reader = std.io.bufferedReader(file.reader());
+    const reader = buf_reader.reader();
+
+    // Read the file
+    var managed = std.ArrayList(u8).init(self.alloc);
+    errdefer managed.deinit();
+    const size: usize = max_image_size;
+    reader.readAllArrayList(&managed, size) catch |err| {
```
jcollie, line 1301: It's not necessary to copy `max_image_size` to a new variable.

jcollie (changes requested):

src/config/Config.zig
jcollie: Yep, glad you were able to get rid of that.

jcollie: > So, apparently, `Path` is used in `Config.zig` as an enum inside `RepeatablePath`, so I thought I would name it `SinglePath`, but if you have any other recommendations, we can most definitely change its name. Also, I thought it would be better if we defined `BackgroundImageMode` directly in `Config.zig` and pass it to lower modules, but if you think we should define it in `renderer`, we can change it as well.

That all makes sense.

src/config/Config.zig
yunusey: Done!

src/config/Config.zig
yunusey: Sorry for the oversight, done!

src/renderer/OpenGL.zig
yunusey: Done!

src/renderer/shaders/bgimage.f.glsl
yunusey: Unfortunately, I don't think that is possible. You can see another application [here](https://github.com/ghostty-org/ghostty/blob/f60068eabd94e6784f3b557c7287c49dd36bb24c/src/renderer/shaders/cell.f.glsl#L27C1-L31C36).

src/renderer/OpenGL.zig
yunusey: Done!

src/renderer/OpenGL.zig
yunusey: Oops, thanks for noticing that!

src/renderer/shaders/bgimage.f.glsl
yunusey: I've checked the shader code again, and I don't think we actually need to pass `mode` to the fragment shader. Though this might change in the case of the addition of new image modes that require the fragment shader to access this variable.

src/config/Config.zig
```diff
@@ -4094,6 +4113,91 @@ pub const Palette = struct {
...
+        };
+    }
+
+    /// Used by Formatter
+    pub fn formatEntry(self: Self, formatter: anytype) !void {
+        const value = self.value orelse return;
+        try formatter.formatEntry([]const u8, value);
+    }
+
+    pub fn expand(
```
mitchellh, line 4151: Given the duplication here with `RepeatablePath`, I'd like to see this pulled out to a shared common function. It'll probably require some parameter changes. This is especially necessary since we just merged home path expansion and this is now missing it.

src/config/Config.zig
```diff
@@ -5416,6 +5520,14 @@ pub const GraphemeWidthMethod = enum {
     unicode,
 };
 
+/// See background-image-mode
+pub const BackgroundImageMode = enum(u8) {
```
mitchellh, line 5524: Please add to docs why this has to be a `u8` and note that these values must always match our shaders exactly.

mitchellh (changes requested):
Fantastic work. I've left some review comments but overall this is looking good.

src/config/Config.zig
yunusey: Done!

jcollie (approved):
This is looking good to me!

src/config/Config.zig
yunusey: I've added a note both for the shader code and for the definition of the `BackgroundImageMode` regarding the synchronization and the use of `u8`.

src/config/Config.zig
```diff
@@ -414,6 +415,24 @@ background: Color = .{ .r = 0x28, .g = 0x2C, .b = 0x34 },
...
+@"background-image-opacity": f32 = 1.0,
+
+/// Background image mode to use.
+///
+/// Valid values are:
+///
+///   * `zoomed` - Image is scaled to fit the window, preserving aspect ratio.
+///   * `scaled` - Image is scaled to fill the window, not preserving aspect ratio.
+///   * `tiled` - Image is repeated horizontally and vertically to fill the window.
+///   * `centered` - Image is centered in the window and displayed 1-to-1 pixel
```
r0wdiggity, line 431: would it be possible to add an image position offset? it might make sense as parameters to to the `centered mode` (i.e `centered 1 -1` would move the image 1 in the positive x direction and -1 in the y direction based on the units in the shader)

jcollie: It would be _possible_ but it would make the config parser more complex for certain. I'd suggest for now convert your image to a PNG and extend the image with a transparent border on the sides opposite to the way you want to shift the image.

It could be useful to have options to anchor the image in the corners, so `upper_left`, `upper_right`, `lower_left`, `lower_right`.

r0wdiggity: yeah thats what I currently have for kitty, but the aspect ratio of the actual image gets messed up because it scales the border as well. The `window_icon` in kitty has the anchors you mentioned which are awesome but using that makes the image appear in each pane with split panes. thanks for the response and the work on this feature!

jcollie: `centered` should not alter the aspect ratio of the image, at least as I imagined it. I'd have to double-check the code.

yunusey: Okay, I've implemented anchor options as well. You can check them out now. Here are some demos using

## Centered
2025-01-06-140639_hyprshot
## Upper Left
2025-01-06-140648_hyprshot
## Upper Right
2025-01-06-140654_hyprshot
## Lower Right
2025-01-06-140700_hyprshot
##  Lower Left
2025-01-06-140707_hyprshot

jcollie (dismissed):

src/config/Config.zig
r0wdiggity: this looks good thank you!

yunusey:
@mitchellh, could you take a look at this? Are there any other changes needed here, or is it ready to go?

src/config/Config.zig
```diff
@@ -20,6 +20,7 @@ const global_state = &@import("../global.zig").state;
 const fontpkg = @import("../font/main.zig");
 const inputpkg = @import("../input.zig");
 const terminal = @import("../terminal/main.zig");
+const BackgroundImageProgram = @import("../renderer/opengl/BackgroundImageProgram.zig");
```
Rohit-Bevinahally, line 23: Is this import needed ?

yunusey: Thanks for pointing that out! I removed it.

jarodtaylor:
Ohhhh I can't wait for this to get in!

sagg0t:
Amazing work here, thank you very much!

And though it's kinda late to ask for such changes may I suggest another option for background-image-mode.
The reason being is that of all the modes implemented so far, if I want an image to fill my whole screen, I'm pretty much choosing between "scaled" mode, which deforms the image or "zoomed" mode, which leaves parts of the window uncovered by the image. And I personally don't find any of those aesthetically pleasing.

The way it should work is that the image is placed in the center, scaled to fill all the window background while preserving aspect ratio, and if the aspect ratio of the image doesn't match one of the terminal window, the part of the image that exceeds the window gets cut off (maybe name it "cropped" mode.)

For reference I can suggest MacOS wallpaper modes:
- "zoomed" mode would be "Fit to Screen";
- "scaled" mode would be "Stretch to Fill Screen";
- and what I'm suggesting is "Fill screen".

I know it's a lot and I'd gladly implement it myself, but I'm barely familiar with Zig and have zero experience with shaders.
So please consider.

yunusey:
> Amazing work here, thank you very much!

Thank you so much!

> And though it's kinda late to ask for such changes may I suggest another option for background-image-mode. The reason being is that of all the modes implemented so far, if I want an image to fill my whole screen, I'm pretty much choosing between "scaled" mode, which deforms the image or "zoomed" mode, which leaves parts of the window uncovered by the image. And I personally don't find any of those aesthetically pleasing.
> 
> The way it should work is that the image is placed in the center, scaled to fill all the window background while preserving aspect ratio, and if the aspect ratio of the image doesn't match one of the terminal window, the part of the image that exceeds the window gets cut off (maybe name it "cropped" mode.)

I think that's a pretty nice mode. I've just implemented it, and as you requested, the mode's name is `cropped`. In order to use it, assuming you are on MacOS from your comment, you will probably need to wait until the same feature is implemented for Metal, which MacOS uses. @Rohit-Bevinahally probably will take care of that as well!

Here is a video demonstrating how the mode acts when the terminal is resized:

https://github.com/user-attachments/assets/c2b5df39-954d-4fc9-8404-7bf325756af4

_The video quality is pretty bad, but hope is sufficient for demonstration purposes._



> I know it's a lot and I'd gladly implement it myself, but I'm barely familiar with Zig and have zero experience with shaders.
> So please consider.

Thanks a lot for commenting again, I think it is a pretty nice feature!

src/config/Config.zig
```diff
@@ -450,6 +450,34 @@ background: Color = .{ .r = 0x28, .g = 0x2C, .b = 0x34 },
...
+///   * `centered` - Image is centered in the window and displayed 1-to-1 pixel
+///     scale, preserving both the aspect ratio and the image size.
+///   * `upper-left` - Image is anchored to the upper left corner of the window,
+///     preserving the aspect ratio.
+///   * `upper-right` - Image is anchored to the upper right corner of the window,
+///     preserving the aspect ratio.
+///   * `lower-left` - Image is anchored to the lower left corner of the window,
+///     preserving the aspect ratio.
+///   * `lower-right` - Image is anchored to the lower right corner of the window,
+///     preserving the aspect ratio.
```
pluiedev, line 477: I would recommend using more standardized terminology like `fit`, `contain`, `cover`, etc. from CSS's [`object-fit`](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) property

src/config/Config.zig
```diff
@@ -3338,6 +3366,86 @@ fn expandPaths(self: *Config, base: []const u8) !void {
     }
 }
 
+/// Expand a relative path to an absolute path. This function is used by
```
pluiedev, line 3369: Needs to be rebased to use configpkg.Path

src/renderer/OpenGL.zig
```diff
@@ -1201,6 +1246,82 @@ fn prepKittyImage(
...
+
+    // Decode the png (currently, we only support png)
+    const decoded_image: wuffs.ImageData = blk: {
+        // Extract the file extension
+        const ext = std.fs.path.extension(path);
+        const ext_lower = try std.ascii.allocLowerString(self.alloc, ext);
+        defer self.alloc.free(ext_lower);
+
+        // Match based on extension
+        if (std.mem.eql(u8, ext_lower, ".png")) {
```
pluiedev, line 1266: Use `std.ascii.eqlIgnoreCase` to avoid allocation

src/renderer/OpenGL.zig
```diff
@@ -2318,6 +2447,31 @@ pub fn drawFrame(self: *OpenGL, surface: *apprt.Surface) !void {
...
+
+            .pending_gray,
+            .pending_gray_alpha,
+            .pending_rgb,
+            .pending_rgba,
+            .replace_gray,
+            .replace_gray_alpha,
+            .replace_rgb,
+            .replace_rgba,
+            => try self.current_background_image.?.upload(self.alloc),
```
pluiedev, line 2463: Why not just use the `current_background_image` that's already been extracted via the if-capture?

src/renderer/opengl/BackgroundImageProgram.zig
```diff
@@ -0,0 +1,116 @@
...
+
+    // Element buffer (EBO)
+    const ebo = try gl.Buffer.create();
+    errdefer ebo.destroy();
+    var ebobind = try ebo.bind(.element_array);
+    defer ebobind.unbind();
+    try ebobind.setData([6]u8{
+        0, 1, 3, // Top-left triangle
+        1, 2, 3, // Bottom-right triangle
+    }, .static_draw);
```
pluiedev, line 51: Nit: you could use a triangle strip here to save 2 vertices (and really, avoid using an EBO entirely)

pluiedev (changes requested):

src/config/Config.zig
yunusey: I think this makes sense, would you mind, maybe, writing what each one should be mapped to? For instance, what should `upper-left` be mapped to in such format? I was also thinking perhaps we could use [Kitty's layout modes](https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.background_image_layout).

src/config/Config.zig
yunusey: Done!

src/renderer/OpenGL.zig
```diff
@@ -810,6 +842,19 @@ pub fn updateFrame(
             try self.prepKittyGraphics(state.terminal);
         }
 
+        if (self.current_background_image == null and
+            self.background_image.value != null)
```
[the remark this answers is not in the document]
yunusey: Thanks for catching that, done!

src/renderer/OpenGL.zig
yunusey: Thanks for catching that! I don't even know what I was thinking; allocating space to lower down a string only to compare it with another string is pretty crazy 😅

src/renderer/OpenGL.zig
yunusey: So, the issue is that `current_background_image` is `const`. Hence, we can't really call the `upload` function with `current_background_image`. If we capture it as a pointer, it will only complicate things for no need.

src/renderer/opengl/BackgroundImageProgram.zig
qwerasd205: This is how it is in all the current OpenGL shaders but is something I actually fix in my rework (which is... getting close to finished finally)

yunusey: As @qwerasd205 also pointed out, this is how we are handling the shaders at the moment (see [ImageProgram](https://github.com/ghostty-org/ghostty/blob/a2f52b08e5ad43420aab4e5cf5b2733cefb05845/src/renderer/opengl/ImageProgram.zig#L58C1-L61C22)). I think it is better if we keep it consistent for now.

src/config/Config.zig
pluiedev: That would be [`background-position`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position) in CSS land I think, as a separate attribute from fit/contain/cover. FWIW I don't think we need to implement options that are as complex as CSS's, but it's nice to have as a reference.

src/renderer/opengl/BackgroundImageProgram.zig
yunusey: @pluiedev, I am marking this as resolved, but if you have any suggestions, don't hesitate to let me know!

src/renderer/OpenGL.zig
yunusey: Can I mark this as resolved?

src/config/Config.zig
yunusey: Okay, so this is how I changed it:
* `zoomed`: `contain`
* `stretched`: `fill`
* `cropped`: `cover`
* `upper-left`: `top-left`
* `upper-right`: `top-right`
* `lower-left`: `bottom-left`
* `lower-right`: `bottom-right`

_I have to say that switching to `top/bottom` from `upper/lower` was a great idea! Apparently, that's what we are using for most other stuff (e.g. [`resize_overlay_position`](https://github.com/ghostty-org/ghostty/blob/26a42fac0ec8f612a3ddce60bab9842c79a2756a/src/config/Config.zig#L1476C1-L1488C61)). Thanks for the suggestion!_

So, for the `tiled` and `centered`;
*  `tiled` doesn't have any mode corresponding to it, unless I am mistaken, so leaving it unchanged for now.
* we could set `centered` to be `none`, but I think it makes it confusing a little bit, we could rename it to `center` like `resize_overlay_position`, or just leave it this way. Please let me know what you think!

pluiedev (changes requested):

src/config/Config.zig
pluiedev: I personally think that we should split this into two settings. Right now as you can see in the GLSL code, the positions and `centered` don't really make sense when put parallel to the other options, because they *imply* `contain` while putting the image at an offset. Instead, we should have one `background-image-fit` setting that has `contain`, `fill`, `cover` and `tiled`, and a `background-image-position` setting.

We should allow users to position their background in the four cardinal directions too, I think. In the future we could also add pixel/percentage offsets to the setting, but I think 8 directions + center works fine for now

src/config/Config.zig
```diff
@@ -4162,6 +4270,63 @@ pub const Palette = struct {
     }
 };
 
+/// SinglePath is a path to a single file. When loading the configuration
+/// file, always the last one will be kept and be automatically expanded
+/// relative to the path of the config file.
+pub const SinglePath = struct {
```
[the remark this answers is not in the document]
yunusey: This was added prior to the introduction of the `Path` struct (see https://github.com/ghostty-org/ghostty/pull/4226#discussion_r1900534393), and I didn't know that we had this addition. I've made the necessary changes with https://github.com/ghostty-org/ghostty/pull/4226/commits/50cf2fee6acb0eee9f557738011e32633b9cf09f.

src/config/Config.zig
yunusey: Okay, I've implemented this feature with https://github.com/ghostty-org/ghostty/pull/4226/commits/624a8c812ce98209a4c260f3c3979b3a4990782c. Now, we have support for `contain`, `fill`, `cover`, `none`, and `tiled` as `background-image-mode`s and we have support for all the 9 directions with `background-image-position`s. `fill` doesn't get affected from the change in directions, naturally. And when the mode `tiled` is selected, it starts tiling from `top-left` regardless as it has a different mechanism and `top-center` or `center-left` doesn't make sense in this context at all. But `contain`, `cover`, and `none` respect the `background-image-position` setting. Let me know what you think!

cschar:
Cannot wait for this!

yunusey:
Hi @pluiedev, did you get a chance to check the recent changes?

pluiedev:
I'm on vacation right now but I'll check whenever I have time

src/renderer/OpenGL.zig
```diff
@@ -302,6 +327,9 @@ pub const DerivedConfig = struct {
         // Copy our shaders
         const custom_shaders = try config.@"custom-shader".clone(alloc);
 
+        // Copy our background image
+        const background_image = try config.@"background-image".?.clone(alloc);
```
pluiedev: Why is this using `.?`? Looks like it will just crash whenever the user doesn't specify a background image

```suggestion
        const background_image = if (config.@"background-image") |v| try v.clone(alloc) else null;
```

src/renderer/OpenGL.zig
```diff
@@ -795,6 +833,27 @@ pub fn updateFrame(
             try self.prepKittyGraphics(state.terminal);
         }
 
+        if (self.current_background_image == null) {
+            if (self.background_image) |background_image| {
+                const img_path, const optional = switch (background_image) {
+                    .optional => |path| .{ path, true },
+                    .required => |path| .{ path, false },
+                };
```
pluiedev: I think it's better to flip this logic and use `required` instead

```suggestion
                const img_path, const required = switch (background_image) {
                    .optional => |path| .{ path, false },
                    .required => |path| .{ path, true },
                };
```

src/renderer/OpenGL.zig
```diff
@@ -795,6 +833,27 @@ pub fn updateFrame(
...
+                self.prepBackgroundImage(img_path) catch |err| {
+                    switch (err) {
+                        error.InvalidData => {
+                            if (!optional) {
+                                log.err("error loading background image {s}: {}", .{ img_path, err });
+                            }
+                        },
+                        else => return err,
+                    }
+                };
```
pluiedev:
```suggestion
                self.prepBackgroundImage(img_path) catch |err| switch (err) {
                    error.InvalidData => if (!optional) {
                        log.err("error loading background image {s}: {}", .{ img_path, err });
                    },
                    else => return err,
                };
```

src/renderer/OpenGL.zig
```diff
@@ -1160,6 +1219,77 @@ fn prepKittyImage(
     gop.value_ptr.transmit_time = image.transmit_time;
 }
 
+/// Prepares the current background image for upload
+pub fn prepBackgroundImage(self: *OpenGL, path: []const u8) !void {
+    // Read the file content
+    const file_content = try self.readImageContent(path);
+    defer self.alloc.free(file_content);
+
+    // Decode the png (currently, we only support png)
```
pluiedev: Aren't we also parsing jpegs below?

src/renderer/OpenGL.zig
```diff
@@ -1160,6 +1219,77 @@ fn prepKittyImage(
...
+        } else {
+            log.warn("unsupported image format: {s}", .{ext});
+            return error.InvalidData;
+        }
+    };
+    defer self.alloc.free(decoded_image.data);
+
+    // Copy the data into the pending state
+    const data = try self.alloc.dupe(u8, decoded_image.data);
+    errdefer self.alloc.free(data);
```
pluiedev: I'm slightly confused as to why this is necessary. We own the data contained within `decoded_image` and we're using the same allocator, so why are we making a copy of it and freeing the old data?

src/renderer/OpenGL.zig
```diff
@@ -1160,6 +1219,77 @@ fn prepKittyImage(
...
+    // File must be a regular file
+    if (file.stat()) |stat| {
+        if (stat.kind != .file) {
+            log.warn("file is not a regular file kind={}", .{stat.kind});
+            return error.InvalidData;
+        }
+    } else |err| {
+        log.warn("failed to stat file: {}", .{err});
+        return error.InvalidData;
+    }
```
pluiedev: Why must this be a regular file? The user could use a symlink and it would be a perfectly valid use case, especially if they use something like Nix to manage dotfiles. As long as we can get data out of it, it's fine IMO. Directories are already filtered out by `openFileAbsolute` when the `directory` flag is unset anyway.

src/renderer/OpenGL.zig
```diff
@@ -1160,6 +1219,77 @@ fn prepKittyImage(
...
+
+    // Read the file
+    var managed = std.ArrayList(u8).init(self.alloc);
+    errdefer managed.deinit();
+    reader.readAllArrayList(&managed, max_image_size) catch |err| {
+        log.warn("failed to read file: {}", .{err});
+        return error.InvalidData;
+    };
+
+    return managed.toOwnedSlice();
```
pluiedev: This is just [`Reader.readAllAlloc`](https://ziglang.org/documentation/master/std/#std.io.GenericReader.readAllAlloc)

src/renderer/OpenGL.zig
```diff
@@ -2298,6 +2437,31 @@ pub fn drawFrame(self: *OpenGL, surface: *apprt.Surface) !void {
...
+
+            .pending_gray,
+            .pending_gray_alpha,
+            .pending_rgb,
+            .pending_rgba,
+            .replace_gray,
+            .replace_gray_alpha,
+            .replace_rgb,
+            .replace_rgba,
+            => try self.current_background_image.?.upload(self.alloc),
```
pluiedev: I still feel I'm a tad allergic to `.?`s. Sorry.

```suggestion
            => try current_background_image.upload(self.alloc),
```

src/renderer/OpenGL.zig
```diff
@@ -2298,6 +2437,31 @@ pub fn drawFrame(self: *OpenGL, surface: *apprt.Surface) !void {
         }
     }
 
+    // Check if we need to update our current background image
+    if (self.current_background_image) |current_background_image| {
+        switch (current_background_image) {
```
pluiedev:
```suggestion
    if (self.current_background_image) |*current_background_image| {
        switch (current_background_image.*) {
```

src/renderer/OpenGL.zig
```diff
@@ -2298,6 +2437,31 @@ pub fn drawFrame(self: *OpenGL, surface: *apprt.Surface) !void {
...
+            .replace_gray_alpha,
+            .replace_rgb,
+            .replace_rgba,
+            => try self.current_background_image.?.upload(self.alloc),
+
+            .unload_pending,
+            .unload_replace,
+            .unload_ready,
+            => {
+                self.current_background_image.?.deinit(self.alloc);
```
pluiedev:
```suggestion
                current_background_image.deinit(self.alloc);
```

src/renderer/shaders/bgimage.v.glsl
```diff
@@ -0,0 +1,81 @@
...
+
+	// Calculate the aspect ratio of the terminal and the image
+	vec2 aspect_ratio = vec2(
+		terminal_size.x / terminal_size.y,
+		image_size.x / image_size.y
+	);
+
+	switch (mode) {
+	case MODE_CONTAIN:
+		// If zoomed, we want to scale the image to fit the terminal
```
pluiedev: Comments need updating

src/renderer/shaders/bgimage.v.glsl
```diff
@@ -0,0 +1,81 @@
...
+		// We don't need to do anything for stretched or tiled
+		break;
+	}
+
+	vec2 final_image_size = terminal_size * position * scale;
+	vec2 offset = vec2(0.0, 0.0);
+
+	uint y_pos = position_index / 3u; // 0 = top, 1 = center, 2 = bottom
+	uint x_pos = position_index % 3u; // 0 = left, 1 = center, 2 = right
+	offset = ((terminal_size * (1.0 - scale)) / 2.0) * vec2(x_pos, y_pos);
```
pluiedev: Why not just declare the variable here directly? This isn't C

```suggestion
	vec2 offset = ((terminal_size * (1.0 - scale)) / 2.0) * vec2(x_pos, y_pos);
```

pluiedev (changes requested):

src/renderer/OpenGL.zig
yunusey: Thanks for catching this.

src/renderer/OpenGL.zig
yunusey: I've updated it, but just in case you are curious, we are using the other way around [here](https://github.com/ghostty-org/ghostty/blob/1ff91625981ef92add92624c01b6b9095c365ed1/src/config/Config.zig#L2829C1-L2832C11), [here](https://github.com/ghostty-org/ghostty/blob/1ff91625981ef92add92624c01b6b9095c365ed1/src/renderer/shadertoy.zig#L27C1-L30C11), or [here](https://github.com/ghostty-org/ghostty/blob/1ff91625981ef92add92624c01b6b9095c365ed1/src/apprt/gtk/App.zig#L1232C1-L1235C11). So, I had implemented it using `optional` as it was more common than the `required` implementation.

src/renderer/OpenGL.zig
yunusey: Done.

src/renderer/OpenGL.zig
yunusey: Yes, thanks for catching that.

src/renderer/OpenGL.zig
```diff
@@ -1160,6 +1215,64 @@ fn prepKittyImage(
...
+        .data = data.ptr,
+    };
+
+    // Store the image
+    self.current_background_image = .{ .pending_rgba = pending };
+}
+
+/// Reads the content of the given image path and returns it
+pub fn readImageContent(self: *OpenGL, path: []const u8) ![]u8 {
+    assert(std.fs.path.isAbsolute(path));
```
pluiedev: `openFileAbsolute` already asserts that the path is absolute, so this check is redundant

src/renderer/OpenGL.zig
yunusey: I was trying to keep things as similar to [`kitty.graphics_image` implementation](https://github.com/ghostty-org/ghostty/blob/1ff91625981ef92add92624c01b6b9095c365ed1/src/terminal/kitty/graphics_image.zig#L239C1-L248C10) as possible, but I've changed it so that we are completely ignoring any of these checks.

src/renderer/OpenGL.zig
yunusey: Done.

src/renderer/OpenGL.zig
yunusey: Done.

src/renderer/shaders/bgimage.v.glsl
yunusey: Thanks for pointing that out.

src/renderer/shaders/bgimage.v.glsl
yunusey: We used to have an offset that changed based on different modes, so it must have stayed from then. Thanks for pointing that out. Done!

src/renderer/OpenGL.zig
yunusey: Again, I was trying to keep things as similar to [kitty's implementation](https://github.com/ghostty-org/ghostty/blob/1ff91625981ef92add92624c01b6b9095c365ed1/src/terminal/kitty/graphics_image.zig#L260C1-L267C11) as possible, but I've changed it.

src/renderer/OpenGL.zig
yunusey: Done.

src/renderer/OpenGL.zig
pluiedev: The thing is that Kitty's implementation has different ["transmission mediums"](https://sw.kovidgoyal.net/kitty/graphics-protocol/#the-transmission-medium) and the code you're copying from deals with only the "simple file" case, which explicitly forbids anything other than regular files. Make sure to check the context!

src/renderer/OpenGL.zig
pluiedev: In that case it's because we need to be able to *add* data onto an existing pile, because Kitty's image data is streamed and chunked. Again ***please*** make sure to read the surrounding context before blindly copying anything

src/renderer/OpenGL.zig
yunusey: I see, thank you for pointing that out. I apologize for my oversight.

src/renderer/OpenGL.zig
yunusey: You are right, https://github.com/ghostty-org/ghostty/pull/4226/commits/5f849b2060f0539db1da327bfb43a86c8f2cb02d removes the redundant `assert`.

src/renderer/OpenGL.zig
yunusey: https://github.com/ghostty-org/ghostty/pull/4226/commits/5f849b2060f0539db1da327bfb43a86c8f2cb02d removes the allocation and uses `@constCast` to pass the already owned `decoded_image.data`.

src/renderer/OpenGL.zig
```diff
@@ -1160,6 +1215,61 @@ fn prepKittyImage(
...
+            return error.InvalidData;
+        }
+    };
+    errdefer self.alloc.free(decoded_image.data);
+
+    // Copy the data into the pending state
+    const pending: Image.Pending = .{
+        .width = decoded_image.width,
+        .height = decoded_image.height,
+        .data = @constCast(decoded_image.data).ptr,
```
pluiedev: This *looks* very sus, but it should be fine. Honestly I don't know why our wuffs bindings return the decoded data as a `[]const u8` as we in fact own its backing memory, so it should be fine to treat it as a `[]u8` for all intents and purposes...

pluiedev (approved):
Thank you so much for putting up with my BS. I'll let someone more qualified to comment on the quality of the shader code, but the Zig part looks great now. Thanks again!

yunusey:
> Thank you so much for putting up with my BS. I'll let someone more qualified to comment on the quality of the shader code, but the Zig part looks great now. Thanks again!

Haha, no BS at all! I appreciate all the feedback you provided; let me know if you have any other suggestions!

src/renderer/OpenGL.zig
yunusey: > This _looks_ very sus

I agree :)



> Honestly I don't know why our wuffs bindings return the decoded data as a `[]const u8` as we in fact own its backing memory, so it should be fine to treat it as a `[]u8` for all intents and purposes...

I am not really sure either. It looks like both `wuffs.png.decode` and `wuffs.jpeg.decode` have the data as `[]u8`, but for some reason, `ImageData.data` is declared to be `[]const u8` and hence they are cast as `const`. But as far as the safety is concerned, I think it should be fine.

src/renderer/OpenGL.zig
```diff
@@ -795,6 +833,23 @@ pub fn updateFrame(
...
 
+        if (self.current_background_image == null) {
+            if (self.background_image) |background_image| {
+                const img_path, const required = switch (background_image) {
+                    .optional => |path| .{ path, false },
+                    .required => |path| .{ path, true },
+                };
+                if (single_threaded_draw) self.draw_mutex.lock();
+                defer if (single_threaded_draw) self.draw_mutex.unlock();
+                self.prepBackgroundImage(img_path) catch |err| switch (err) {
```
pluiedev, line 844: Oops... I think you forgot to expand paths here. Currently using a path relative to home (prefixed with `~/`) crashes the app 😅

pluiedev (changes requested):

mitchellh:
FYI I’ve been holding off on this and the cursor animation PRs because I don’t want to make any renderer changes until Qwers refactor is in. After that we should be able to rebase and I’m happy to take that on. 

Until then I’m holding off review for a moment

pluiedev:
image

The current rendering of the image is also way too dark for some reason. It's supposed to be the same image as my desktop wallpaper:

image

src/renderer/OpenGL.zig
yunusey: Thanks for pointing that out; it is really interesting. I think `Config.expandPaths` should handle all of these. I'll look into it in a few hours!

yunusey:
> The current rendering of the image is also way too dark for some reason. It's supposed to be the same image as my desktop wallpaper:

Oh, I am only now noticing the issue (I was testing with `background-image-opacity = 0.4` all the time). Thank you so much for letting me know.

src/renderer/OpenGL.zig
yunusey: So, I think I found the issue: as you know, we are using `?Path` for `background-image` which is not handled in the `Config.expandPath` function (only `RepeatablePath` and `Path` are). So, we need an additional check for this type. Can you check https://github.com/ghostty-org/ghostty/pull/4226/commits/3784f84c30ad8f6c92341c45b51cbde89d682e96 whenever you have time, please?

yunusey:
https://github.com/ghostty-org/ghostty/pull/4226/commits/304a67c3e21970cbf12da466943d9f93b6c3d389 should resolve the brightness issue of the background images now. I think `wuff.png.decode` and `wuff.jpeg.decode` had been updated recently that basically changed the way the images are stored (the problem starts occurring after the rebase). There exists an `unlinearize` function that's used in [`image.f.glsl`](https://github.com/ghostty-org/ghostty/blob/1ff91625981ef92add92624c01b6b9095c365ed1/src/renderer/shaders/image.f.glsl#L9C1-L16C2) that converts the image to the `sRGB` with gamma compression that outputs the appropriate colors needed, and I think it solves the issue! Please let me know if it works for you!

src/renderer/OpenGL.zig
pluiedev: Honestly we should just change the type in the wuffs binding to be `[]u8` then

pluiedev (approved):
image

Gorgeous

src/renderer/OpenGL.zig
yunusey: > Honestly we should just change the type in the wuffs binding to be `[]u8` then

Hi again! I think (https://github.com/ghostty-org/ghostty/pull/4226/commits/5034880dde13dcb55966c8e5b12af261dbe34216) should fix the issue now.

qwerasd205:
This PR is superseded by #7686. Thanks for the work on this though! Even if this won't be merged, the bike shedding of the config options in particular helped me decide on what config options to provide pretty quickly instead of having to work it out from scratch.

Props for dealing with implementing this under the old renderer architecture- seems like it was annoyingly in the weeds because of the lack of abstraction.

PR #4235, opened 2024-12-31
fix(gtk): add close confirmation for tabs

### LINKED ISSUE #4234: No confirmation when closing tab (GTK)
On the discord this bit of feedback came up from [here](https://ncurses.scripts.mit.edu/?p=ncurses.git;a=blob;f=misc/terminfo.src;h=ccd4c32099cf740d331eeaa3955f48f177435878;hb=a28a11d84d969cfdc876e158deae7870e8948a24#l8323)
```
8323 # - ghostty has tabs (imitating gnome-terminal); when closing a tab with a
8324 #   running process (e.g., a hung vttest), ghostty does not prompt about the 
8325 #   process to be killed.
```

On investigation there are two places in the GTK apprt where tabs are directly closed/removed without checking for confirmation/running processes ([1](https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/Tab.zig#L127), [2](https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/Tab.zig#L149))

leigholiver:
On the discord this bit of feedback came up from [here](https://ncurses.scripts.mit.edu/?p=ncurses.git;a=blob;f=misc/terminfo.src;h=ccd4c32099cf740d331eeaa3955f48f177435878;hb=a28a11d84d969cfdc876e158deae7870e8948a24#l8323)
```
8323 # - ghostty has tabs (imitating gnome-terminal); when closing a tab with a
8324 #   running process (e.g., a hung vttest), ghostty does not prompt about the 
8325 #   process to be killed.
```

This PR adds confirmation to the places where tabs are closed directly 

Fixes: https://github.com/ghostty-org/ghostty/issues/4234

### DIFF
--- a/src/apprt/gtk/Tab.zig
+++ b/src/apprt/gtk/Tab.zig
@@ -122,14 +122,12 @@ pub fn remove(self: *Tab) void {
 }
 
 /// Helper function to check if any surface in the split hierarchy needs close confirmation
-const needsConfirm = struct {
-    fn check(elem: Surface.Container.Elem) bool {
-        return switch (elem) {
-            .surface => |s| s.core_surface.needsConfirmQuit(),
-            .split => |s| check(s.top_left) or check(s.bottom_right),
-        };
-    }
-}.check;
+fn needsConfirm(elem: Surface.Container.Elem) bool {
+    return switch (elem) {
+        .surface => |s| s.core_surface.needsConfirmQuit(),
+        .split => |s| needsConfirm(s.top_left) or needsConfirm(s.bottom_right),
+    };
+}
 
 /// Close the tab, asking for confirmation if any surface requests it.
 fn closeWithConfirmation(tab: *Tab) void {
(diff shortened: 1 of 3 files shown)


src/apprt/gtk/Tab.zig
```diff
@@ -121,10 +121,71 @@ pub fn remove(self: *Tab) void {
...
 
+/// Helper function to check if any surface in the split hierarchy needs close confirmation
+const needsConfirm = struct {
+    fn check(elem: Surface.Container.Elem) bool {
+        return switch (elem) {
+            .surface => |s| s.core_surface.needsConfirmQuit(),
+            .split => |s| check(s.top_left) or check(s.bottom_right),
+        };
+    }
+}.check;
```
mitchellh, line 132: I'm aware of this pattern but its not something we generally do within this project.

I don't think we need to wrap this in a struct at all and can just make it a top-level (non-pub) `fn`.

mitchellh:
Thanks! This only did GTK and didn't work for Adwaita builds. My add-on commit makes this work for Adwaita as well.

PR #4241, opened 2025-01-01
core: if we change RLIMIT_NOFILE, reset it when executing commands

### LINKED ISSUE #4232: Restore default file handle limits for executed children
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4215

<sup>Originally posted by **tbiehn** December 31, 2024</sup>
On my machine Ghostty sets a soft stack size limit of 
`-s: stack size (kbytes)             16384` 
and a file handle limit of 
`-n: file descriptors                1073741816`. 
[src/file.zig](https://github.com/ghostty-org/ghostty/blob/47cf5cbb4023d70fc1009d1953fc9348fc1b60c8/src/os/file.zig#L7) mentions that file handle limits are maximized, ostensibly to allow a single parent process to own many children PTYs without hitting the handle limit. Cool.

Perhaps ghostty should not pass these soft limits to child PTYs.
Maybe we should save the unmodified limits and set them on child PTYs?</div>

jcollie:
Fixes #4232 .


src/os/main.zig
```diff
@@ -32,7 +32,9 @@ pub const getenv = env.getenv;
 pub const setenv = env.setenv;
 pub const unsetenv = env.unsetenv;
 pub const launchedFromDesktop = desktop.launchedFromDesktop;
+pub const rlimit = file.rlimit;
 pub const fixMaxFiles = file.fixMaxFiles;
+pub const resrtoreMaxFiles = file.restoreMaxFiles;
```
auscompgeek, line 37: typo:
```suggestion
pub const restoreMaxFiles = file.restoreMaxFiles;
```

tbiehn:
The only question I have is whether this is better in Command.zig or specified in Exec.zig's pre_exec handler.

To me Exec.zig feels 'closer' to the code that changes the rlimits in the first place, and the existence of a pre_exec handler feels like its meant to implement this sort of fixup.

Practically that's just moving the global import to Exec.zig instead and throwing the fixup in around line ~1160.

jcollie:
> The only question I have is whether this is better in Command.zig or specified in Exec.zig's pre_exec handler.
> 
> To me Exec.zig feels 'closer' to the code that changes the rlimits in the first place, and the existence of a pre_exec handler feels like its meant to implement this sort of fixup.
> 
> Practically that's just moving the global import to Exec.zig instead and throwing the fixup in around line ~1160.

I think that the argument for doing it in `src/Command.zig` is that's where we do other things like setting up Linux cgroups and the like. `Command` is only used in `src/termio/Exec.zig` AFAIR so there's no real practical difference. There are other locations where we fork/exec besides `src/Command.zig` so perhaps those should receive the same treatment.

src/os/main.zig
jcollie: Whoops, I'll fix that up soon.

mitchellh (approved):
Thanks! I made some modifications. Namely:

I moved restoration before `preexec`. If a caller wants to set the rlimits in a future preexec we don't want to overwrite those.

I also used a struct to better encapsulate our logic.

PR #4242, opened 2025-01-01
Restore original file handle soft limits in child processes.

tbiehn:
Addresses #4215


tbiehn:
@jcollie addresses this in Command.zig -- this addresses it in Exec.zig, Command.zig looks OK to me.

PR #4255, opened 2025-01-01
Partial fix for #1938, add GDK_DEBUG=gl-no-fractional

morganabc:
GSK_RENDERER=opengl does not properly handle fractional scaling and thus we need to set GDK_DEBUG=gl-no-fractional.

Potential fix for #1938

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -108,7 +108,7 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         // From gtk 4.16, GDK_DEBUG is split into GDK_DEBUG and GDK_DISABLE.
         // For the remainder of "why" see the 4.14 comment below.
         _ = internal_os.setenv("GDK_DISABLE", "gles-api,vulkan");
-        _ = internal_os.setenv("GDK_DEBUG", "opengl");
+        _ = internal_os.setenv("GDK_DEBUG", "opengl,gl-no-fractional");
     } else if (version.atLeast(4, 14, 0)) {
         // We need to export GDK_DEBUG to run on Wayland after GTK 4.14.
         // Older versions of GTK do not support these values so it is safe
@@ -123,7 +123,7 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         //   - "vulkan-disable" - disable Vulkan, Ghostty can't use Vulkan
         //     and initializing a Vulkan context was causing a longer delay
         //     on some systems.
-        _ = internal_os.setenv("GDK_DEBUG", "opengl,gl-disable-gles,vulkan-disable");
+        _ = internal_os.setenv("GDK_DEBUG", "opengl,gl-disable-gles,vulkan-disable,gl-no-fractional");
     } else {
         // Versions prior to 4.14 are a bit of an unknown for Ghostty. It
         // is an environment that isn't tested well and we don't have a


CnTeng:
It works! Thinks!

agriffis:
This makes a significant difference for me. Kitty is my benchmark, and unpatched ghostty is blurrier (look especially at the bold fonts in the vim status line). Your patch makes ghostty the crispest of the three!

### kitty

kitty

### ghostty unpatched

ghostty-before

### ghostty patched

ghostty-after

agriffis:
Just wanted to follow up to acknowledge a drawback of this fix. It seems to result in a quantization that sometimes affects letter shapes. For example "p" might render nicely most of the time, but occasionally the vertical stem is skinnier by a pixel, or maybe the right side of the bowl appears pinched.

IMHO this is objectively better than the previous fuzziness, but it's a matter of opinion whether it's better than kitty. Some people might feel that kitty's rendering still strikes the best balance.

Ideally I think that ghostty would:

1. fix the blurriness to achieve kitty-like rendering
2. offer a configuration option to quantize pixels as this patch achieves

so that the two "better" outcomes are both available by user preference. Now that I've used this patch, I might actually stick with it personally, because I really like the crispness even if it sometimes results in a skinny stem.

I would include a screenshot of the oddities, but... every time I engage GNOME's screenshot tool, the oddities go away!

morgan9e:
> Just wanted to follow up to acknowledge a drawback of this fix. It seems to result in a quantization that sometimes affects letter shapes. For example "p" might render nicely most of the time, but occasionally the vertical stem is skinnier by a pixel, or maybe the right side of the bowl appears pinched.
> 
> 
> 
> IMHO this is objectively better than the previous fuzziness, but it's a matter of opinion whether it's better than kitty. Some people might feel that kitty's rendering still strikes the best balance.
> 
> 
> 
> Ideally I think that ghostty would:
> 
> 
> 
> 1. fix the blurriness to achieve kitty-like rendering
> 
> 2. offer a configuration option to quantize pixels as this patch achieves
> 
> 
> 
> so that the two "better" outcomes are both available by user preference. Now that I've used this patch, I might actually stick with it personally, because I really like the crispness even if it sometimes results in a skinny stem.
> 
> 
> 
> I would include a screenshot of the oddities, but... every time I engage GNOME's screenshot tool, the oddities go away!

Thanks for detailed follow up! 
Yeah it still has traditional pixel problem, but I think GTK wont be ever able to fix it..

I dont know how kitty renders font but integrating different font renderer with it would be hard anyways so..

morgan9e:
> Just wanted to follow up to acknowledge a drawback of this fix. It seems to result in a quantization that sometimes affects letter shapes. For example "p" might render nicely most of the time, but occasionally the vertical stem is skinnier by a pixel, or maybe the right side of the bowl appears pinched.
> 
> 
> 
> IMHO this is objectively better than the previous fuzziness, but it's a matter of opinion whether it's better than kitty. Some people might feel that kitty's rendering still strikes the best balance.
> 
> 
> 
> Ideally I think that ghostty would:
> 
> 
> 
> 1. fix the blurriness to achieve kitty-like rendering
> 
> 2. offer a configuration option to quantize pixels as this patch achieves
> 
> 
> 
> so that the two "better" outcomes are both available by user preference. Now that I've used this patch, I might actually stick with it personally, because I really like the crispness even if it sometimes results in a skinny stem.
> 
> 
> 
> I would include a screenshot of the oddities, but... every time I engage GNOME's screenshot tool, the oddities go away!

Thanks for detailed follow up!
Yeah it still has traditional pixel problem, but I think GTK wont ever be able to fix it

I guess fractional scaling is pain but with font rendering its much more pain..

morgan9e:
> Just wanted to follow up to acknowledge a drawback of this fix. It seems to result in a quantization that sometimes affects letter shapes. For example "p" might render nicely most of the time, but occasionally the vertical stem is skinnier by a pixel, or maybe the right side of the bowl appears pinched.
> 
> 
> 
> IMHO this is objectively better than the previous fuzziness, but it's a matter of opinion whether it's better than kitty. Some people might feel that kitty's rendering still strikes the best balance.
> 
> 
> 
> Ideally I think that ghostty would:
> 
> 
> 
> 1. fix the blurriness to achieve kitty-like rendering
> 
> 2. offer a configuration option to quantize pixels as this patch achieves
> 
> 
> 
> so that the two "better" outcomes are both available by user preference. Now that I've used this patch, I might actually stick with it personally, because I really like the crispness even if it sometimes results in a skinny stem.
> 
> 
> 
> I would include a screenshot of the oddities, but... every time I engage GNOME's screenshot tool, the oddities go away!

Thanks for detailed follow up!
Yeah it still has traditional pixel problem, but I think GTK wont ever be able to fix it.

We dont have that much options unless they add more mature renderer for fractional scaling.. they said newGL would fix it, but obviously it still cant fix it for some cases.

I think there is no answer for this fractional font scaling problem. One say the macOS way is right, other say that way is accurate or this way is more appealing..
Fractional scaling is pain already but with font rendering its much more pain..

mitchellh:
Thanks for researching this and thanks to others for confirming this fix. As noted above, even if there is a better long term solution this seems like a strict improvement, so let's go with it. I verified locally (non-fractional) that this doesn't negatively impact anything for me.

agriffis:
@morgan9e ghostty font rendering seems to have reverted to blurry for me. It's not a ghostty problem... it happens even when I install the older builds from around the timeframe of this PR.

Is it possible that `GDK_DEBUG=gl-no-fractional` stopped working in a recent update of gtk?

morgan9e:
> @morgan9e ghostty font rendering seems to have reverted to blurry for me. It's not a ghostty problem... it happens even when I install the older builds from around the timeframe of this PR.
> 
> Is it possible that `GDK_DEBUG=gl-no-fractional` stopped working in a recent update of gtk?
> 
> EDIT: Yes, it's gone: https://gitlab.gnome.org/GNOME/gtk/-/commit/a99224bac0

Thanks!
why did they removed it...?

So now we have to change renderer or pin gtk version
I've seen some project do that since GTK's "decisions" breaks a lot of things.

I'll have look at it soon.

teohhanhui:
> why did they removed it...?

The `_DEBUG` part of the env var might give you a hint. It's never meant to be something they support.

morgan9e:
> > why did they removed it...?
> 
> 
> 
> The `_DEBUG` part of the env var might give you a hint. It's never meant to be something they support.

well that makes sense
but their fix for blurry fractional scaling of old renderers was using that debug feature, so its bit sad.

blurry font was expected behavior so we can't do anything other than that.

PR #4298, opened 2025-01-01
support different base for palette keys in config

roshal:
motivated by the desire to align Aonfig lines

https://ziglang.org/doAumentation/master/std/#std.fmt.parseInt

unaligned

```ini
palette = 0=#000000

palette = 1=#FFAAAA
palette = 2=#AAFFAA
palette = 4=#AAAAFF

palette = 3=#FFFF00
palette = 5=#FF00FF
palette = 6=#00FFFF

palette = 7=#AAAAAA
palette = 8=#555555

palette = 9=#FF55AA
palette = 10=#AAFF55
palette = 12=#55AAFF

palette = 11=#FFAA55
palette = 13=#AA55FF
palette = 14=#55FFAA

palette = 15=#FFFFFF
```

expecting

```ini
palette = 0x0=#000

palette = 0x1=#FAA
palette = 0x2=#AFA
palette = 0x4=#AAF

palette = 0x3=#FF0
palette = 0x5=#F0F
palette = 0x6=#0FF

palette = 0x7=#AAA
palette = 0x8=#555

palette = 0x9=#F5A
palette = 0xA=#AF5
palette = 0xC=#5AF

palette = 0xB=#FA5
palette = 0xD=#A5F
palette = 0xE=#5FA

palette = 0xF=#FFF
```

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -447,6 +447,10 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 /// the 256 colors in the terminal color table) and `COLOR` is a typical RGB
 /// color code such as `#AABBCC` or `AABBCC`, or a named X11 color.
 ///
+/// The palette index can be in decimal, binary, octal, or hexadecimal.
+/// Decimal is assumed unless a prefix is used: `0b` for binary, `0o` for octal,
+/// and `0x` for hexadecimal.
+///
 /// For definitions on the color indices and what they canonically map to,
 /// [see this cheat sheet](https://www.ditig.com/256-colors-cheat-sheet).
 palette: Palette = .{},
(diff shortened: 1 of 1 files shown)


src/config/Config.zig
```diff
@@ -4036,7 +4036,7 @@ pub const Palette = struct {
         const eqlIdx = std.mem.indexOf(u8, value, "=") orelse
             return error.InvalidValue;
 
-        const key = try std.fmt.parseInt(u8, value[0..eqlIdx], 10);
+        const key = try std.fmt.parseInt(u8, value[0..eqlIdx]);
```
mitchellh, line 4039: We use Zig 0.13 so the final arg has to be `0` to imply the base based on prefix.

mitchellh (changes requested):

mitchellh:
This is a bit cursed but I don't see a reason to disallow it.

jcollie:
> This is a bit cursed but I don't see a reason to disallow it.

If we do allow it, it should be documented.

jcollie:
Also, the commit messages are just "update". I don't think that we need to get _super_ formal here with our commit messages but I think that the 1st line should be a little more descriptive. @roshal can you squash these commits together and provide a better commit message?

roshal:
@jcollie

commits will squash when merging a pull request and the commit message will contain the pull request title

jcollie:
> @jcollie
> 
> my assumption is that commits will squash when merging and the commit message will contain the pull request title
> 
> should i squash all commits together to provide the pull request title in the commit message ?

The commit that merges this PR will contain the title and text from this PR but the individual commits will not be merged and their commit messages will not be altered.

Issue #4302, opened 2025-01-01
Feature Request: option to start Ghostty maximized on Linux


ComicShrimp:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3642

Implement a new configuration flag to allow Ghostty to start in a maximized window state on Linux systems using GTK.

Configuration Option:
  Name: `gtk-maximize`
  Type: `boolean`

When set to true, Ghostty will launch with a maximized window on startup. This option will be evaluated before the existing `fullscreen` setting. If both `gtk-maximize` and `fullscreen` are enabled, `fullscreen` will take precedence.

ComicShrimp:
Got It, Thanks for the quick response !

PR #4305, opened 2025-01-01
feature: option to start Ghostty maximized on Linux

ComicShrimp:
This PR is related to #4302 adding the option to start Ghostty maximized in linux systems adding a new flag in config.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -265,10 +265,12 @@ pub fn init(self: *Window, app: *App) !void {
     c.gtk_popover_set_has_arrow(@ptrCast(@alignCast(self.context_menu)), 0);
     c.gtk_widget_set_halign(self.context_menu, c.GTK_ALIGN_START);
 
-    if (app.config.@"gtk-maximize") c.gtk_window_maximize(self.window);
-
-    // If we are in fullscreen mode, new windows start fullscreen.
-    if (app.config.fullscreen) c.gtk_window_fullscreen(self.window);
+    // Set window mode based on the config
+    if (app.config.fullscreen) {
+        c.gtk_window_fullscreen(self.window);
+    } else if (app.config.maximize) {
+        c.gtk_window_maximize(self.window);
+    }
 
     // We register a key event controller with the window so
     // we can catch key events when our surface may not be
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -265,6 +265,8 @@ pub fn init(self: *Window, app: *App) !void {
     c.gtk_popover_set_has_arrow(@ptrCast(@alignCast(self.context_menu)), 0);
     c.gtk_widget_set_halign(self.context_menu, c.GTK_ALIGN_START);
 
+    if (app.config.@"gtk-maximize") c.gtk_window_maximize(self.window);
+
     // If we are in fullscreen mode, new windows start fullscreen.
     if (app.config.fullscreen) c.gtk_window_fullscreen(self.window);
 
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1855,8 +1855,9 @@ keybind: Keybinds = .{},
 /// does not apply to tabs, splits, etc. However, this setting will apply to all
 /// new windows, not just the first one.
 ///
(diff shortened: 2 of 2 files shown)


src/config/Config.zig
```diff
@@ -1851,6 +1851,13 @@ keybind: Keybinds = .{},
 /// must always be able to move themselves into an isolated cgroup.
 @"linux-cgroup-hard-fail": bool = false,
 
+/// Start new windows maximized. This setting applies to new windows and
+/// does not apply to tabs, splits, etc. However, this setting will apply to all
+/// new windows, not just the first one.
+///
+/// This setting is overwritten by `fullscreen` option.
+@"gtk-maximize": bool = false,
```
jcollie, line 1859: I don't think that we want to prefix this config with `gtk-` as this could be used on macOS as well.

src/apprt/gtk/Window.zig
```diff
@@ -265,6 +265,8 @@ pub fn init(self: *Window, app: *App) !void {
     c.gtk_popover_set_has_arrow(@ptrCast(@alignCast(self.context_menu)), 0);
     c.gtk_widget_set_halign(self.context_menu, c.GTK_ALIGN_START);
 
+    if (app.config.@"gtk-maximize") c.gtk_window_maximize(self.window);
+
     // If we are in fullscreen mode, new windows start fullscreen.
     if (app.config.fullscreen) c.gtk_window_fullscreen(self.window);
```
jcollie, line 272: A warning should be logged if both `maximize` and `fullscreen` are set to `true`. That can be done in `finalize()` in `src/config/Config.zig`.  `fullscreen` should take precedence, so don't call `gtk_window_maximize` if `fullscreen=true`.

src/config/Config.zig
```diff
@@ -1851,6 +1851,13 @@ keybind: Keybinds = .{},
 /// must always be able to move themselves into an isolated cgroup.
 @"linux-cgroup-hard-fail": bool = false,
 
+/// Start new windows maximized. This setting applies to new windows and
+/// does not apply to tabs, splits, etc. However, this setting will apply to all
+/// new windows, not just the first one.
+///
+/// This setting is overwritten by `fullscreen` option.
```
jcollie, line 1858: I think a better wording would be is that `maximized` is ignored if `fullscreen=true`.

jcollie (changes requested):

src/config/Config.zig
ComicShrimp: Sure, it makes more sense, I removed `gtk-` from the name.

src/config/Config.zig
ComicShrimp: Changed the config description and added a note, to indicate that this feature is only for Linux, for now.

src/apprt/gtk/Window.zig
ComicShrimp: Thanks for guiding me, I added the evaluation to `finalize()` function.

mitchellh (changes requested):
This is kind of funky but this feature already existed without introducing a configuration. 

We should be treating `window-width` and `window-height` with overly large values and clamping them to the screen, effectively maximizing the window. This is even documented:

```
/// Sizes larger than the screen size will be clamped to the screen size.
/// This can be used to create a maximized-by-default window size.
```

I'd like to request that we follow this same approach for now in GTK and remove the explicit `maximize`.

ComicShrimp:
> This is kind of funky but this feature already existed without introducing a configuration.
> 
> We should be treating `window-width` and `window-height` with overly large values and clamping them to the screen, effectively maximizing the window. This is even documented:
> 
> ```
> /// Sizes larger than the screen size will be clamped to the screen size.
> /// This can be used to create a maximized-by-default window size.
> ```
> 
> I'd like to request that we follow this same approach for now in GTK and remove the explicit `maximize`.

This behavior is already present in GTK and works as expected. Since we don't need explicit `maximize`, I think is better to close this PR.

mitchellh:
I agree. I will note that it is _funky_ like I said, but it avoids a separate confusion of how to handle `window-width/height` AND `maximize` being set, which would at the very least have to be handled and moreso have to be documented as an edge case.

That kind of exists already for `fullscreen` but I feel like `fullscreen` is an entire different app state already that its a bit less surprising how things should go.

PR #4308, opened 2025-01-01
macos: fix window borders on dark mode

gantoreno:
After updating to 1.0.1 I noticed something different in the terminal, which turned out being the window borders - it appeared as if Ghostty was using light-mode style borders (dark/black outline with a thin light stroke at the top) instead of the entire light outline from before:

| 1.0.0 | 1.0.1 |
| - | - |
| Screenshot 2025-01-01 at 2 28 12 PM | Screenshot 2025-01-01 at 2 29 07 PM |

After digging a bit, I found #3834, which fixes fullscreen background colors through alpha channels by appending a `withAlphaComponent(0.0)` to `backgroundColor` - for reasons I may be entirely unaware of (since I'm not a Swift developer), this seems to cause the dark-mode border style to go away.

Some lines above that, I noticed the `.clear` callout from line 266, which talks about matching Terminal.app's styles, and it _also_ has a `withAlphaComponent` but set to `0.001` - if I understand correctly, and the fix from #3834 works by setting the alpha component to a _practically_ zero value, then I thought perhaps a really small number like `0.001` could do the trick as well. This ended up working and bringing back the right borders again.

Not sure again if this may make a difference anywhere else in the app or bring any undesired behavior, but if anyone who is well-versed in Swift would like chime in with more details or perhaps a better approach, I'd greatly appreciate it!

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -244,7 +244,9 @@ class TerminalController: BaseTerminalController {
         let backgroundColor: OSColor
         if let surfaceTree {
             if let focusedSurface, surfaceTree.doesBorderTop(view: focusedSurface) {
-                backgroundColor = OSColor(focusedSurface.backgroundColor ?? surfaceConfig.backgroundColor).withAlphaComponent(0.0)
+                // Similar to above, an alpha component of "0" causes compositor issues, so
+                // we use 0.001. See: https://github.com/ghostty-org/ghostty/pull/4308
+                backgroundColor = OSColor(focusedSurface.backgroundColor ?? surfaceConfig.backgroundColor).withAlphaComponent(0.001)
             } else {
                 // We don't have a focused surface or our surface doesn't border the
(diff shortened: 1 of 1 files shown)


dmehala:
This is the second regression introduced by [fa831405](https://github.com/ghostty-org/ghostty/commit/fa8314058552bda5c7c09f769f7fef7c97aa39f4). At this point, it might be worth reverting and finding another solution?

mitchellh:
> This is the second regression introduced by [fa831405](https://github.com/ghostty-org/ghostty/commit/fa8314058552bda5c7c09f769f7fef7c97aa39f4). At this point, it might be worth reverting and finding another solution?

What is the other regression?

mitchellh:
For now I'll accept this, thank you!!! But if we decide that modifying alpha here AT ALL is wrong, I'm happy to revert that fully.

dmehala:
@mitchellh I can't find the discussion/issue but it was reported on [discord](https://discord.com/channels/1005603569187160125/1318616867576746065/1318616867576746065). It causes flickering on new splits.

PR #4313, opened 2025-01-01
Don't install 1024x1024 icons for Flatpak

yorickpeterse:
Per the Flatpak spec (https://docs.flatpak.org/en/latest/conventions.html#application-icons) the maximum icon size is 512x512. Trying to build a Flatpak with an icon larger than this will fail.

To solve this, installing the icon is skipped when building with -Dflatpak=true.


jcollie (approved):

mitchellh:
This is probably why I actually didn't install the 1024 version initially because I previously had a working (buggy though) Flatpak build.

Issue #3125, opened 2024-12-25
Parse ConEmu OSC9


mitchellh:
See: https://conemu.github.io/en/AnsiEscapeCodes.html#OSC_Operating_system_commands

Some other background: #3124, #3119 

ConEmu OSC9 conflicts with iTerm notifications. We've supported iTerm notifications for a long time and I'd like to continue to do so. As a workaround, I think we should say that any ConEmu OSC9 commands can _never_ be desktop notifications. This feels like a good tradeoff to me because why are you making those notifications anyways... and you can work around it by using OSC777.

We should expand our OSC parser to parse all ConEmu OS commands (OSC9).

As part of this, we should do better research on how ConEmu handles specific erroneous inputs because the table does not make it clear and there is no spec. 

**NOTE: This issue is only tracking _parsing_. We should not _do anything_ with these commands.** I'm not sure if we ever want to support some of these and for the others we should discuss in a GH discussion about what to do.

- [X] 9;1 #4327
- [X] 9;2 #4447
- [X] 9;3 #4644
- [X] 9;4 #3124
- [X] 9;5 #4948
- [x] 9;6 #8410 
- [X] 9;7 #10399
- [X] 9;8 #10399
- [X] 9;9 #10399 
- [X] 9;10 #10399
- [X] 9;11 #10399
- [X] 9;12 #10399

Also of note:
#4727 #8413

rhodes-b:
https://github.com/ghostty-org/ghostty/pull/7975

This added progress bar support? is that sufficient to close this (or I guess limit to macos it hasnt been added there?)

pluiedev:
We only support OSC 9;1 to OSC 9;5, whereas a full ConEmu parser would need to parse OSC 9;6 to OSC 9;12 as well

jcollie:
Would be helpful to edit the 1st comment or add some sub-tasks to keep track of what's missing. Also, we only do something with 9;4 (progress bars). We should also document which ones we _definitely_ do not plan to implement and which ones we'd be open to accepting patches for.

Some of the ConEmu OSCs look like security nightmares - OSC 9;7 is insta-RCE and 9;6 and 9;8 are sus although 9;6 is probably too ConEmu-specific anyway.

9;1 seems pointless, not sure why you'd need to use an esc sequence to sleep when basically every language offers a sleep function.
9;2 could basically be implemented by hooking it up to existing desktop notifications.
9;3 is similar to the "Change Title" but remote controllable.
[...]

PR #4327, opened 2025-01-01
feat: parse ConEmu OSC9;1

dmehala:
# Description

This PR implements support for the [ConEmu OSC9;1 escape sequence](https://conemu.github.io/en/AnsiEscapeCodes.html#OSC_Operating_system_commands).

Based on my understanding of [ConEmu's source code](https://github.com/Maximus5/ConEmu/blob/740b09c363cb16fbb730d72c53eaca1c530a016e/src/ConEmuCD/ConAnsiImpl.cpp#L705-L724):
- The default timeout is set to `100` milliseconds if no value is specified.  
- The timeout value is clamped to a maximum of `10000` milliseconds.

#3125


src/terminal/osc.zig
```diff
@@ -158,6 +158,9 @@ pub const Command = union(enum) {
     /// End a hyperlink (OSC 8)
     hyperlink_end: void,
 
+    /// Sleep (OSC 9;1) in ms
+    sleep: u16,
```
dmehala, line 162: I considered using the [Duration type](https://github.com/ghostty-org/ghostty/blob/94599102e9fb8247af08cbbbcb7ee25e3d31e1bd/src/config/Config.zig#L5546) but this would introduce a dependency on Config. Would that be an issue?

src/terminal/osc.zig
```diff
@@ -788,6 +796,21 @@ pub const Parser = struct {
...
+                    self.command = .{ .sleep = 100 };
+                    self.buf_start = self.buf_idx;
+                    self.complete = true;
+                    self.state = .conemu_sleep_value;
+                },
+                else => self.state = .invalid,
+            },
+
+            .conemu_sleep_value => switch (c) {
+                '0'...'9' => {},
```
dmehala, line 810: I could construct the number directly here, but I thought it would be safer to handle potential overflow correctly by using `std.fmt.parseUnsigned`.

src/terminal/osc.zig
```diff
@@ -1147,6 +1170,22 @@ pub const Parser = struct {
...
 
+    fn endConEmuSleepValue(self: *Parser) void {
+        switch (self.command) {
+            .sleep => |*v| {
+                const str = self.buf[self.buf_start..self.buf_idx];
+                if (str.len != 0) {
+                    if (std.fmt.parseUnsigned(u16, str, 10)) |num| {
+                        v.* = @min(num, 10000);
+                    } else |_| {
+                        v.* = 10000;
```
dmehala, line 1181: Not sure if we need to replicate ConEmu's behaviour here.

src/terminal/osc.zig
jcollie: I don't think that the Duration type is necessary here, unless there's a need to display it with units or parse it from a string units. You might consider scaling it to nanoseconds and using a `u64` since that's what Zig tends to use for that sort of thing.

src/terminal/osc.zig
jcollie: `parseUnsigned` allows `_` as in `1_000_000` is a valid Zig number so I always feel a little icky when using it to parse data. I think that doing it like this is fine. Performance shouldn't be a concern.

src/terminal/osc.zig
jcollie: What's ConEmu's behavior? We should either replicate ConEmu's behavior or document the differences.

src/terminal/osc.zig
```diff
@@ -1634,6 +1674,48 @@ test "OSC: set palette color" {
...
+
+    const input = "9;1;12345";
+    for (input) |ch| p.next(ch);
+
+    const cmd = p.end('\x1b').?;
+
+    try testing.expect(cmd == .sleep);
+    try testing.expectEqual(10000, cmd.sleep);
+}
+
```
jcollie, line 1718: You should test invalid codes as well, perhaps parameters that contain something other than digits.

src/terminal/osc.zig
dmehala: https://github.com/Maximus5/ConEmu/blob/740b09c363cb16fbb730d72c53eaca1c530a016e/src/ConEmuCD/ConAnsiImpl.cpp#L705-L724

jcollie: OK, looks like we match ConEmu's behavior. Please be careful linking to code though as we can't use GPL code since we are MIT. ConEmu is BSD 3 clause so I think that's OK (IANAL).

src/terminal/osc.zig
dmehala: addressed in c98d207eb98b1a0d822e148c79792f1b17c3a98a

src/terminal/osc.zig
```diff
@@ -1634,6 +1676,61 @@ test "OSC: set palette color" {
...
+    var p: Parser = .{};
+
+    const input = "9;1;foo";
+    for (input) |ch| p.next(ch);
+
+    const cmd = p.end('\x1b');
+
+    try testing.expect(cmd == null);
+}
```
dmehala, line 1732: Note to myself: well that should default to `100` to fully match conemu's behaviou

src/terminal/osc.zig
mitchellh: Since the duration is in the ConEmu spec as `ms` I'm fine keeping this as `u16` and documenting it (as @dmehala already has) as a millisecond unit.

src/terminal/osc.zig
dmehala: addressed in 9d9fa60ece3736001be39ec2ac80b0e6c4b44f17

dmehala:
Alright, I think the logic matches ConEmu's now.

jcollie (approved):

mitchellh:
Please rebase. I can't merge :)

PR #4341, opened 2025-01-01
Install icons during postInstall phase of Nix flake

toasterrepairman:
Small fix for Nix packaging that installs the app icon and desktop file (icon_512 by default) during postInstall phase. There was no issue for this, but I'm not really sure it was large enough to warrant a full triage. 

Before:

image

After: 

image


jcollie:
Those are not the standard locations/names to install the `.desktop` file or the icon. The `.desktop` file is installed to `$out/share/applications/com.mitchellh.ghostty.desktop` and the icons are installed to `$out/share/icons/hicolor/<size>/apps/com.mitchell.ghostty.png`. If those file location aren't working for you then I suspect that you have something wrong locally on your system.

PR #4343, opened 2025-01-02
core: detect what desktop environment the user is using

jcollie:


tristan957 (commented):
It's XDG_CURRENT_DESKTOP

jcollie:
> It's XDG_CURRENT_DESKTOP

`XDG_CURRENT_DESKTOP` is wierd, it can actually be a list of up to 3 desktop names. I have no idea why you would want that but that's what it's documented to be. `XDG_SESSION_DESKTOP` is just the current DE.

rhodes-b (commented):
this could also potentially be a good time / place to also print what OS someone is using 

its typically found in `/etc/*release` on many distros not sure how much this covers

I know you talked about that before

jcollie:
> this could also potentially be a good time / place to also print what OS someone is using
> 
> its typically found in `/etc/*release` on many distros not sure how much this covers
> 
> I know you talked about that before

Yeah I've thought about that but `/etc/os-release` is not super easy to parse (I believe basically it's a list of shell environment variables). So I don't think that it's necessary to do that at this point. It's definitely out of scope for this PR.

anund:
May want to grab `XDG_SESSION_TYPE` as additional information in the other case. Wandering around a few vms with default, perhaps incorrect, configuration `XDG_SESSION_DESKTOP` wasn't always set. (plasma,gnome,hyprland seem ok).

jcollie:
> May want to grab `XDG_SESSION_TYPE` as additional information in the other case. Wandering around a few vms with default, perhaps incorrect, configuration `XDG_SESSION_DESKTOP` wasn't always set. (plasma,gnome,hyprland seem ok).

We're mainly concerned with gnome/not gnome at this point. If we ever need to detect a specific DE other than Gnome we can refine this check. We can add XDG_SESSION_TYPE in the future if it's needed.

src/cli/version.zig
```diff
@@ -36,6 +37,7 @@ pub fn run(alloc: Allocator) !u8 {
     try stdout.print("  - font engine: {}\n", .{build_config.font_backend});
     try stdout.print("  - renderer   : {}\n", .{renderer.Renderer});
     try stdout.print("  - libxev     : {}\n", .{xev.backend});
+    try stdout.print("  - desktop env: {s}\n", .{@tagName(internal_os.desktopEnvironment())});
```
mitchellh, line 40: I'd only write this out for the GTK apprt. macOS has no others so its just noise.

src/os/desktop.zig
```diff
@@ -59,3 +59,29 @@ pub fn launchedFromDesktop() bool {
...
+
+/// Detect what desktop environment we are running under. This is mainly used on
+/// Linux to enable or disable GTK client-side decorations but there may be more
+/// uses in the future.
+pub fn desktopEnvironment() DesktopEnvironment {
+    return switch (comptime builtin.os.tag) {
+        .macos => .macos,
+        .windows => .windows,
+        .linux => de: {
+            if (@inComptime()) @compileError("Checking for the desktop environment on Linux must be done at runtime.");
```
mitchellh, line 78: This is unnecessary. The functions called below won't work in comptime, the compiler will error for us.

mitchellh (changes requested):

src/os/desktop.zig
jcollie: The call to `posix.getenv` does indeed fail, but it fails more because of an implementation detail in Zig that could change in the future rather than something more fundamental so I would prefer to be explicit about why we can't check for DE at comptime (and provide a more informative error): https://github.com/ziglang/zig/issues/4524

`std.ascii.eqlIgnoreCase` works just fine at comptime.

mitchellh:
Can you please rebase, there are conflicts

jcollie:
> Can you please rebase, there are conflicts

Done, and done.

mitchellh (approved):

PR #4345, opened 2025-01-02
performable: prefix

### LINKED ISSUE #4328: Introduce `performable:` prefix for keybind triggers to passthrough if they would do nothing
Discussion: #2811 

The discussion was focused on `copy` but this could apply to many different actions so I think we should generalize it. Ghostty already has a mechanism in `performBindingAction` to determine if a binding action was _performed_ or not. We can use that. 

The new prefix `performable:` (similar to other prefixes like `unconsumed:`, `all:`, `global:`) would be a new modifier with the following effect:

## Details

`performable:`-prefixed triggers will only apply at all if they can be performed. If they can't be performed, they act as if the keybinding doesn't exist. For example, this can be used with copy to copy or send an interrupt signal:

```ini
keybind = performable:ctrl+c=copy_to_clipboard
```

This trigger interacts with the other prefixes in the following way:

  - `global:` - No effect. Global actions are always considered performed.
  - `all:` - No effect. All actions are always considered performed.
  - `unconsumed:` - Only applies if the action is performable. These two sort of cancel each other out so it's odd to use them both, but it doesn't cause problems.

IrishBruse:
closes #4328
closes #3970

makes this possible now
```
keybind = performable:ctrl+c=copy_to_clipboard # copy if theres a selection else send sigint
keybind = ctrl+v=paste_from_clipboard
```

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1941,10 +1941,16 @@ fn maybeHandleBinding(
         return .closed;
     }
 
+    // If we have the performable flag and the
+    // action was not performed do nothing at all
+    if (leaf.flags.performable and !performed) {
+        return null;
+    }
+
     // If we consume this event, then we are done. If we don't consume
     // it, we processed the action but we still want to process our
     // encodings, too.
-    if (performed and consumed) {
+    if (consumed) {
         // If we had queued events, we deinit them since we consumed
         self.endKeySequence(.drop, .retain);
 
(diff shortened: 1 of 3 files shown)


mitchellh:
Thanks! You mostly got it. I made some modifications feel free to look at my commits.

IrishBruse:
Awesome thanks for the help im new to zig.

PR #4387, opened 2025-01-02
GTK: crash reports notice

### LINKED ISSUE #2183: GUI notice there are crash reports
#2173 is merged. As noted in that PR, we currently only dump crash reports to disk and the user has to know that to realize they're there.

We should show a GUI notice that crash reports exist and what path they exist at and encourage bug reporting. In the future, it may be nice to have a "send report" button too. This is also related to #2182 because we may in the future want to show some information about the crash as well.

I think an initial success would be simply noting crashes exist, and what their file paths are.

b6k-dev:
This pr implements the GUI notice informing about existing crash reports in the GTK application, and partially resolves https://github.com/ghostty-org/ghostty/issues/2183.

As proposed in https://github.com/ghostty-org/ghostty/issues/2183#issuecomment-2325155949 I implemented a banner similiar to the debug mode notice.
The message embeds clickable links leading to the reports directory and reporting instructions.

`gtk-adwaita=true`:
image
`gtk-adwaita=false`:
image

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -44,6 +44,9 @@ const log = std.log.scoped(.gtk);
 
 pub const Options = struct {};
 
+/// A global variable indicating if user has dismissed the notice about existing crash reports
+pub var crash_reports_notice_dismissed = false;
+
 core_app: *CoreApp,
 config: Config,
 
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -44,8 +44,6 @@ const log = std.log.scoped(.gtk);
 
 pub const Options = struct {};
 
-/// A global variable indicating if user has dismissed the notice about existing crash reports
-pub var crash_reports_notice_dismissed = false;
 
 core_app: *CoreApp,
 config: Config,
@@ -55,6 +53,9 @@ ctx: *c.GMainContext,
 
 /// True if the app was launched with single instance mode.
 single_instance: bool,
+
+/// A flag indicating if user has dismissed the notice about existing crash reports
+crash_reports_notice_dismissed: bool = false,
 
 /// The "none" cursor. We use one that is shared across the entire app.
 cursor_none: ?*c.GdkCursor,
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -44,7 +44,6 @@ const log = std.log.scoped(.gtk);
 
 pub const Options = struct {};
 
-
 core_app: *CoreApp,
 config: Config,
 
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -14,6 +14,7 @@ const configpkg = @import("../../config.zig");
 const font = @import("../../font/main.zig");
 const input = @import("../../input.zig");
 const CoreSurface = @import("../../Surface.zig");
+const crash = @import("../../crash/main.zig");
 
 const App = @import("App.zig");
 const Color = configpkg.Config.Color;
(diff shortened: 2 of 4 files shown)


src/apprt/gtk/Window.zig
```diff
@@ -247,9 +248,83 @@ pub fn init(self: *Window, app: *App) !void {
...
+
+        if (try it.next()) |_| {
+            const warning_text = try std.fmt.allocPrintZ(
+                self.app.core_app.alloc,
+            \\ ⚠️ There are existing crash reports located at <a href="file://{s}">{s}</a>.
+                 \\ If possible, <a href="https://github.com/ghostty-org/ghostty?tab=readme-ov-file#crash-reports">report them to the developers</a> to help improve the application.
+                    ,
+                .{crash_dir.path, crash_dir.path}
+            );
+            defer self.app.core_app.alloc.free(warning_text);
```
maciekbartczak, line 270: After giving it some thought this probably needs improvements to memory management. 
I think `warning_text` should be freed only after the banner is dismissed. What's more we should probably destroy and free the banner widget and its container when it is dismissed, since there is no way to bring it back after that.
In order to simplify those operations I'm thinking of extracting all that banner logic and state into a separate struct so that we can access the references in the dismiss callback just by casting the data pointer to self. 
However I'll wait with introducing any changes for feedback, since I have a feeling I'm trying to overengineer the implementation here.

maciekbartczak: If the widget and container need to be destroyed, I’ll need some guidance, as I couldn’t find out how do do it from the GTK docs.

mitchellh:
This looks pretty good to me. I'd like @jcollie or @tristan957 to review this. 

We're now disabling Sentry by default on Linux because its not providing useful information yet, but having this in for when its enabled would be good.

src/apprt/gtk/Window.zig
tristan957: GTK/libadwaita will copy the text, so you should feel free to deallocate memory as soon as possible.

tristan957:
I don't like that we seem to be controlling the line break. I'd rather let GTK/Pango handle that, so the text can just flow as needed.

jcollie:
> I don't like that we seem to be controlling the line break. I'd rather let GTK/Pango handle that, so the text can just flow as needed.

I tried that out and it doesn't look very good, plus it takes up a lot more vertical space because it's not packed very well. Unless we get fancier with the layout I think it's a good compromise.

image

src/apprt/gtk/Window.zig
```diff
@@ -247,9 +248,94 @@ pub fn init(self: *Window, app: *App) !void {
         }
         c.gtk_widget_add_css_class(@ptrCast(gtk_window), "devel");
         c.gtk_widget_add_css_class(@ptrCast(warning_box), "background");
+        c.gtk_widget_add_css_class(@ptrCast(warning_box), "banner-box");
         c.gtk_box_append(@ptrCast(box), warning_box);
     }
 
+    // Display a banner informing about crash reports if any exist
+    {
```
jcollie, line 256: The `Window.init` function is getting quite long so I think that we need to break this out into a separate function and add the banner there.

src/apprt/gtk/Window.zig
```diff
@@ -247,9 +248,94 @@ pub fn init(self: *Window, app: *App) !void {
...
+        defer self.app.core_app.alloc.free(crash_dir.path);
+        var it = try crash_dir.iterator();
+        defer it.deinit();
+
+        if (try it.next()) |_| {
+            const warning_text = try std.fmt.allocPrintZ(
+                self.app.core_app.alloc,
+            \\ There are existing crash reports located at <a href="file://{s}">{s}</a>.
+                 \\ If possible, <a href="https://github.com/ghostty-org/ghostty?tab=readme-ov-file#crash-reports">report them to the developers</a> to help improve the application.
+                    ,
```
jcollie, line 267: It's wonderful that clicking the links takes you someplace useful. I was thinking that we'd immediately be inundated with requests to get some kind of GUI to manage crash reports (and we may still need that eventually) but clicking the link brings up Gnome's file manager with the crash reports directory so it's pretty easy for users to manage their crash reports.

mitchellh:
> I don't like that we seem to be controlling the line break. I'd rather let GTK/Pango handle that, so the text can just flow as needed.

Fwiw on the macOS we manually control line breaking all over to be more aesthetically pleasing.

src/apprt/gtk/Window.zig
jcollie, line 256: Also, we need to set a global variable (probably in `src/gtk/App.zig`) to indicate if the user has been shown the banner and has dismissed it. Right now every new window gets the banner, even if you have dismissed it previously. I don't think that it needs to be persisted on disk but as long as I'm using the same instance of Ghostty the banner should only be shown once.

src/apprt/gtk/Window.zig
```diff
@@ -247,9 +248,94 @@ pub fn init(self: *Window, app: *App) !void {
         }
         c.gtk_widget_add_css_class(@ptrCast(gtk_window), "devel");
         c.gtk_widget_add_css_class(@ptrCast(warning_box), "background");
+        c.gtk_widget_add_css_class(@ptrCast(warning_box), "banner-box");
         c.gtk_box_append(@ptrCast(box), warning_box);
     }
 
+    // Display a banner informing about crash reports if any exist
+    {
+        var crash_dir = try crash.defaultDir(self.app.core_app.alloc);
```
jcollie, line 257: This can be a `const`.

jcollie (changes requested):
The code also needs to have `zig fmt` run against it. I've made a few comments, but otherwise is looking good. `prettier` also needs to be run to format the `.css` files.

src/apprt/gtk/Window.zig
maciekbartczak: Done.

maciekbartczak: Implemented the global variable approach.
I think that ideally we should only show the notice if new crash reports are generated since the last dismissal time, but that requires some sort of disk persistence and i feel like that is beyond the scope of this pr.

maciekbartczak:
Hey @jcollie, just following up on this PR - I've addressed the requested changes and resolved the conflicts, I would appreciate a re-review.

src/apprt/gtk/App.zig
```diff
@@ -44,6 +44,9 @@ const log = std.log.scoped(.gtk);
 
 pub const Options = struct {};
 
+/// A global variable indicating if user has dismissed the notice about existing crash reports
+pub var crash_reports_notice_dismissed = false;
+
```
pluiedev, line 49: Global states are a bit yikes and we should avoid them. I think setting this as a field of GTK App works just fine.

src/apprt/gtk/Window.zig
```diff
@@ -575,6 +579,72 @@ pub fn sendToast(self: *Window, title: [:0]const u8) void {
...
+    const crash_dir = try crash.defaultDir(self.app.core_app.alloc);
+    defer self.app.core_app.alloc.free(crash_dir.path);
+    var it = try crash_dir.iterator();
+    defer it.deinit();
+
+    if (try it.next()) |_| {
+        const warning_text = try std.fmt.allocPrintZ(self.app.core_app.alloc,
+            \\ ⚠️ There are existing crash reports located at <a href="file://{s}">{s}</a>.
+            \\ If possible, <a href="https://github.com/ghostty-org/ghostty?tab=readme-ov-file#crash-reports">report them to the developers</a> to help improve the application.
+        , .{ crash_dir.path, crash_dir.path });
```
pluiedev, line 596: FYI you can avoid repeating the parameter by specifying `{0s}` in the format string

src/apprt/gtk/Window.zig
```diff
@@ -961,6 +1031,18 @@ fn gtkActionReset(
...
+fn gtkHideWidget(_: ?*c.GtkWidget, data: ?*anyopaque) void {
+    const widget: *c.GtkWidget = @ptrCast(@alignCast(data));
+    c.gtk_widget_set_visible(widget, c.FALSE);
+    App.crash_reports_notice_dismissed = true;
+}
+
+fn adwHideBanner(_: ?*c.GtkWidget, data: ?*anyopaque) void {
+    const banner: *c.AdwBanner = @ptrCast(data);
+    c.adw_banner_set_revealed(banner, c.FALSE);
+    App.crash_reports_notice_dismissed = true;
```
pluiedev, line 1043: I think it would be better if this propagates to all windows, which we can do via the apprt action system

pluiedev (changes requested):

src/apprt/gtk/Window.zig
jcollie: I don't know that we want to do it via the apprt action system, but definitely this should propagate to all windows.

maciekbartczak: @jcollie what alternative solution would you propose to propagate the dismissal to multiple windows?

jcollie: The Core app structure contains a list of apprt surfaces. Basically need to loop over the list, get the associated window and take the steps to clear the banner.

src/apprt/gtk/Window.zig
```diff
@@ -575,6 +581,74 @@ pub fn sendToast(self: *Window, title: [:0]const u8) void {
...
+            const close_button = c.gtk_button_new_from_icon_name("window-close-symbolic");
+            c.gtk_widget_set_margin_top(@ptrCast(close_button), 10);
+            c.gtk_widget_set_margin_bottom(@ptrCast(close_button), 10);
+            c.gtk_widget_set_margin_end(@ptrCast(close_button), 10);
+            _ = c.g_signal_connect_data(close_button, "clicked", c.G_CALLBACK(&gtkHideWidget), self, null, 0);
+
+            c.gtk_center_box_set_center_widget(@ptrCast(center_box), message);
+            c.gtk_center_box_set_end_widget(@ptrCast(center_box), close_button);
+
+            c.gtk_box_append(@ptrCast(warning_box), @ptrCast(center_box));
```
pluiedev, line 643: FYI since we're dropping non-Adwaita GTK builds you can remove this branch (or maybe this can serve as a fallback for truly ancient libadwaita before 1.3)

src/apprt/gtk/Window.zig
```diff
@@ -961,6 +1035,28 @@ fn gtkActionReset(
     };
 }
 
+fn gtkHideWidget(_: ?*c.GtkWidget, data: ?*anyopaque) void {
+    const self = userdataSelf(data.?);
+    for (self.app.core_app.surfaces.items) |surface| {
+        const gtk_surface: *Surface = @ptrCast(surface);
```
pluiedev, line 1041: Is this cast necessary? The items should be `apprt.Surface`, which is equal to the GTK `Surface` at compile time.

pluiedev (changes requested):

src/apprt/gtk/Window.zig
maciekbartczak: It is not necessary :+1:

src/apprt/gtk/Window.zig
maciekbartczak: I've removed the non-Adwaita fallback and moved the Adwaita version check to the top of the function.

src/apprt/gtk/Window.zig
```diff
@@ -582,6 +582,13 @@ pub fn sendToast(self: *Window, title: [:0]const u8) void {
 }
 
 fn displayCrashReportsNoticeIfNeeded(self: *Window, app: *App, parent_widget: *c.GtkWidget) !void {
+    if (!((comptime adwaita.versionAtLeast(1, 3, 0)) and
+        adwaita.enabled(&app.config) and
```
jcollie, line 586: As of this morning Adwaita is no longer optional so this function no longer exists.

src/apprt/gtk/Window.zig
```diff
@@ -1032,15 +1036,25 @@ fn gtkActionReset(
...
 
 fn adwHideBanner(_: ?*c.GtkWidget, data: ?*anyopaque) void {
-    const banner: *c.AdwBanner = @ptrCast(data);
-    c.adw_banner_set_revealed(banner, c.FALSE);
-    App.crash_reports_notice_dismissed = true;
+    const self = userdataSelf(data.?);
+    for (self.app.core_app.surfaces.items) |surface| {
+        const gtk_surface: *Surface = @ptrCast(surface);
+        if (gtk_surface.container.window()) |window| {
+            c.adw_banner_set_revealed(@ptrCast(window.crash_notice_widget), c.FALSE);
```
jcollie, line 1023: Even though it probably doesn't matter to the GTK code, `window.crash_notice_widget` should be checked for `null`.

jcollie (changes requested):

jcollie:
After some discussion on Discord, I'm going to close this PR. You've done some good work and have been responsive to all of our comments! Unfortunately displaying crash report banners on Linux is unnecessary at this point since Sentry is disabled on Linux by default right now. Since 99.9% of Linux users will never generate a crash report having the machinery to report them is extra code to maintain that's never used. Thanks again and I hope that you will continue to contribute to Ghostty! There's lots of work to be done yet!

PR #4395, opened 2025-01-02
Add `close_tab` keybinding action for macOS

### LINKED ISSUE #4331: Add `close_tab` keybinding action
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3778

<sup>Originally posted by **Qtechinnovation** December 28, 2024</sup>
I noticed that https://ghostty.org/docs/config/keybind/reference does not mention an option to set a keybind to close a tab, but only to open one. I assumed that since opening a new tab by default uses the rather sensible ctrl+shift+t, that closing the tab would use ctrl+shift+w, but it does not. Is there an option to set a close tab keybind? I found a discussion about a person who forked ghostty to add one, but no interaction with the discussion, so I think this is warranted, especially for all the people who use keybinds as a way to increase productivity.</div>

### Details

This would close a tab _and all splits within that tab_. This should behave like `close_window` does for tabs and splits.

liby:
This PR adds support for a dedicated `close_tab` keybinding action, allowing users to bind specific keys for closing tabs. The implementation:

- Adds `close_tab` as a new keybinding action
- Preserves all existing confirmation dialogs for running processes
- Works seamlessly with macOS native tab system

### Testing
- [x] Tested with single tabs
- [x] Tested with multiple tabs
- [x] Tested with running processes (confirmation dialog)
- [x] Tested with splits within tabs

image

https://github.com/user-attachments/assets/155210f7-20fe-4a96-8800-6969df214871

Partially resolved #4331

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -30,6 +30,7 @@ class AppDelegate: NSObject,
     @IBOutlet private var menuSplitRight: NSMenuItem?
     @IBOutlet private var menuSplitDown: NSMenuItem?
     @IBOutlet private var menuClose: NSMenuItem?
+    @IBOutlet private var menuCloseTab: NSMenuItem?
     @IBOutlet private var menuCloseWindow: NSMenuItem?
     @IBOutlet private var menuCloseAllWindows: NSMenuItem?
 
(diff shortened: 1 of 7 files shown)


mitchellh:
> Resolves https://github.com/ghostty-org/ghostty/issues/4331

Just noting this is not true, since we have to also implement it for GTK. But I think there's a PR for that too.

Thanks for doing this, I'll review soon!

mitchellh:
Awesome work! I made some minor modifications to fit in with the GTK work done. Auto-merge is on and then this is supported on both platforms.

PR #4403, opened 2025-01-02
gtk(wayland): add support for background blur on KDE Plasma

pluiedev:
Also establishes a foundation for Wayland support and fixes a minor bug (GTK windows remaining opaque when `background-opacity` is set to 1 on startup and later updated to less than 1 with a config reload)

Can't update the Zig cache hash myself since I'm currently in China and my proxy's broken for some reason :(

See also #4361, part of #4626

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -24,6 +24,8 @@ const XCFrameworkStep = @import("src/build/XCFrameworkStep.zig");
 const Version = @import("src/build/Version.zig");
 const Command = @import("src/Command.zig");
 
+const Scanner = @import("zig_wayland").Scanner;
+
 comptime {
     // This is the required Zig version for building this project. We allow
     // any patch version but the major and minor must match exactly.
@@ -105,19 +107,19 @@ pub fn build(b: *std.Build) !void {
         "Enables the use of Adwaita when using the GTK rendering backend.",
     ) orelse true;
 
-    config.x11 = b.option(
-        bool,
-        "gtk-x11",
-        "Enables linking against X11 libraries when using the GTK rendering backend.",
-    ) orelse x11: {
-        if (target.result.os.tag != .linux) break :x11 false;
+    var x11 = false;
+    var wayland = false;
 
+    if (target.result.os.tag == .linux) pkgconfig: {
         var pkgconfig = std.process.Child.init(&.{ "pkg-config", "--variable=targets", "gtk4" }, b.allocator);
 
         pkgconfig.stdout_behavior = .Pipe;
         pkgconfig.stderr_behavior = .Pipe;
 
-        try pkgconfig.spawn();
+        pkgconfig.spawn() catch |err| {
+            std.log.warn("failed to spawn pkg-config - disabling X11 and Wayland integrations: {}", .{err});
+            break :pkgconfig;
+        };
 
         const output_max_size = 50 * 1024;
 
@@ -139,18 +141,31 @@ pub fn build(b: *std.Build) !void {
         switch (term) {
             .Exited => |code| {
                 if (code == 0) {
-                    if (std.mem.indexOf(u8, stdout.items, "x11")) |_| break :x11 true;
-                    break :x11 false;
+                    if (std.mem.indexOf(u8, stdout.items, "x11")) |_| x11 = true;
+                    if (std.mem.indexOf(u8, stdout.items, "wayland")) |_| wayland = true;
+                } else {
+                    std.log.warn("pkg-config: {s} with code {d}", .{ @tagName(term), code });
+                    return error.Unexpected;
                 }
-                std.log.warn("pkg-config: {s} with code {d}", .{ @tagName(term), code });
-                break :x11 false;
             },
             inline else => |code| {
                 std.log.warn("pkg-config: {s} with code {d}", .{ @tagName(term), code });
                 return error.Unexpected;
             },
         }
-    };
+    }
+
+    config.x11 = b.option(
+        bool,
+        "gtk-x11",
+        "Enables linking against X11 libraries when using the GTK rendering backend.",
+    ) orelse x11;
+
+    config.wayland = b.option(
+        bool,
+        "gtk-wayland",
+        "Enables linking against Wayland libraries when using the GTK rendering backend.",
+    ) orelse wayland;
 
     config.sentry = b.option(
         bool,
@@ -1458,6 +1473,17 @@ fn addDeps(
                 step.linkSystemLibrary2("gtk4", dynamic_link_opts);
                 if (config.adwaita) step.linkSystemLibrary2("adwaita-1", dynamic_link_opts);
                 if (config.x11) step.linkSystemLibrary2("X11", dynamic_link_opts);
+
+                if (config.wayland) {
+                    const scanner = Scanner.create(b, .{});
+
+                    const wayland = b.createModule(.{ .root_source_file = scanner.result });
+
+                    scanner.generate("wl_compositor", 1);
+
+                    step.root_module.addImport("wayland", wayland);
+                    step.linkSystemLibrary2("wayland-client", dynamic_link_opts);
+                }
 
                 {
                     const gresource = @import("src/apprt/gtk/gresource.zig");
--- a/build.zig
+++ b/build.zig
@@ -1479,7 +1479,14 @@ fn addDeps(
 
                     const wayland = b.createModule(.{ .root_source_file = scanner.result });
 
+                    const plasma_wayland_protocols = b.dependency("plasma_wayland_protocols", .{
+                        .target = target,
+                        .optimize = optimize,
+                    });
+                    scanner.addCustomProtocol(plasma_wayland_protocols.path("src/protocols/blur.xml"));
+
                     scanner.generate("wl_compositor", 1);
+                    scanner.generate("org_kde_kwin_blur_manager", 1);
 
                     step.root_module.addImport("wayland", wayland);
                     step.linkSystemLibrary2("wayland-client", dynamic_link_opts);
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -10,10 +10,6 @@
   oniguruma,
   zlib,
   libGL,
-  libX11,
-  libXcursor,
-  libXi,
-  libXrandr,
   glib,
   gtk4,
   libadwaita,
@@ -26,7 +22,17 @@
   pandoc,
   revision ? "dirty",
   optimize ? "Debug",
-  x11 ? true,
+
+  enableX11 ? true,
+  libX11,
+  libXcursor,
+  libXi,
+  libXrandr,
+
+  enableWayland ? true,
+  wayland,
+  wayland-protocols,
+  wayland-scanner,
 }: let
   # The Zig hook has no way to select the release type without actual
   # overriding of the default flags.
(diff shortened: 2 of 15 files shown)


src/config/Config.zig
```diff
@@ -563,11 +563,27 @@ palette: Palette = .{},
...
+/// "Blur" plugin. If disabled, enable it by ticking the checkbox to the left.
+/// Then click on the "Configure" button and there will be two sliders that
+/// allow you to set background blur and noise strengths for all apps,
+/// including Ghostty.
+///
+/// All other Linux desktop environments are as of now unsupported. Users may
+/// need to set environment-specific settings and/or install third-party plugins
+/// in order to support background blur, as there isn't a unified interface for
+/// doing so.
 @"background-blur-radius": u8 = 0,
```
jcollie, line 587: I think that it would be better to introduce a `background-blur` boolean to enable/disable background blur and then `background-blur-radius` would then only have an effect on macOS. Or make `background-blur-radius` an optional that defaults to `null`, but I think that the explicit boolean would be clearer.

src/apprt/gtk/Window.zig
```diff
@@ -79,6 +83,7 @@ pub fn init(self: *Window, app: *App) !void {
         .notebook = undefined,
         .context_menu = undefined,
         .toast_overlay = undefined,
+        .wayland = undefined,
```
gabydd, line 86: I think this needs to be set to null, as Wayland.init can return an error and if it does then wayland won't be set and will stay undefined. And if configChange is called then the undefined memory will be accessed

gabydd (changes requested):
We should consider resource management here for example do we need to destroy the blur manager on close. And do we need to handle the display disconnect or does gtk do that I think you can use WAYLAND_DEBUG=1 to see the Wayland messages which should have the destroy and disconnect stuff there

src/apprt/gtk/Window.zig
jcollie: If we try and initialize a connection to Wayland and it fails, perhaps that should be a fatal error?

hqnna: > If we try and initialize a connection to Wayland and it fails, perhaps that should be a fatal error?

This would probably be the ideal approach.

nix/package.nix
```diff
@@ -27,6 +27,10 @@
   revision ? "dirty",
   optimize ? "Debug",
   x11 ? true,
+  enableWayland ? true,
```
mitchellh, line 30: I think we should ensure `x11` and `enableWayland` are using the same style. So we should probably change `x11` to `enableX11`. I'm not sure what is idiomatic for Nix here, but whatever is we should do.

src/apprt/gtk/Window.zig
```diff
@@ -629,6 +647,18 @@ fn gtkRefocusTerm(v: *c.GtkWindow, ud: ?*anyopaque) callconv(.C) bool {
     return true;
 }
 
+fn gtkRealize(v: *c.GtkWindow, ud: ?*anyopaque) callconv(.C) bool {
+    const self = userdataSelf(ud.?);
+
+    self.wayland = Wayland.init(v) catch return false;
+
+    if (self.wayland) |*wayland| {
+        wayland.setBlur(self.app.config.@"background-blur-radius" > 0) catch return false;
```
mitchellh, line 656: We shouldn't duplicate this. In the macOS app it's common for there to be a central `syncAppearance` function that we call from everywhere where we need to setup appearance. We may want to copy that pattern here.

src/apprt/gtk/wayland.zig
```diff
@@ -0,0 +1,89 @@
+const std = @import("std");
+const c = @import("c.zig").c;
+const wayland = @import("wayland");
+const wl = wayland.client.wl;
+const org = wayland.client.org;
+
+const log = std.log.scoped(.gtk_wayland);
+
+const Wayland = @This();
```
mitchellh, line 9: In this codebase we put this at the very top always (first line).

src/apprt/gtk/wayland.zig
```diff
@@ -0,0 +1,89 @@
...
+        c.gdk_wayland_surface_get_type(),
+    ) == 0)
+        return null;
+
+    const wl_surface: *wl.Surface = @ptrCast(c.gdk_wayland_surface_get_wl_surface(surface) orelse return null);
+    const wl_display: *wl.Display = @ptrCast(c.gdk_wayland_display_get_wl_display(display) orelse return null);
+
+    const registry = try wl_display.getRegistry();
+
+    var self = Wayland{ .surface = wl_surface };
```
mitchellh, line 32: Prefer RLS: `var self: Wayland = .{`

src/apprt/gtk/wayland.zig
```diff
@@ -0,0 +1,89 @@
...
+
+    registry.setListener(*Wayland, registryListener, &self);
```
mitchellh, line 34: I'm not familiar with Wayland protocols [yet] but is this safe to do because the callback happens just below this? Otherwise, `&self` is a pointer to the stack.

src/apprt/gtk/wayland.zig
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    if (self.blur_token) |*tok| {
+        // Only release token when transitioning from blurred -> not blurred
+        if (!blurred) {
+            mgr.unset(self.surface);
+            tok.*.release();
+            self.blur_token = null;
+        }
+    } else {
```
mitchellh, line 57: Nitpick but I'd prefer to return above this and not have the `else` nested.

src/config/Config.zig
mitchellh: I disagree about introducing a new config. macOS uses 0 to disable blur. We should do the same here. The uglier part is that the magnitude of value doesn't matter for Linux.

I'd propose introduce a new type here `BackgroundBlurRadius` that wraps a `u8` but parses `false` to 0 and `true` to something else (I recommend 20, since that's a reasonable value for macOS). That way this works on both macOS and Linux with "true" but also can be set to a specific number to have an effect on macOS and still enable on Linux.

And ultimately, this keeps a single, backwards compatible config.

mitchellh (changes requested):
This is looking extremely good. The code quality here is good and the complexity is very manageable. Great work. I've requested some changes and left some comments.

nix/package.nix
jcollie: The nixpkgs package has `x11Support` so probably `waylandSupport`?

src/apprt/gtk/wayland.zig
gabydd: this is fine for now because the roundtrip blocks on the events being processed https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1ab60f38c2f80980ac84f347e932793390 but in the future we might want a stable pointer here, for example in my apprt it is very useful that the surface is a stable pointer on init so you can do things like this https://github.com/gabydd/ghostty-wayland/blob/ab63027d485eeee31124bb1cd52a0e877ef5ea13/src/wayland.zig#L772-L773 without much song and dance, whereas things at the app level are more complicated though that is mostly because i have to get the Surface from the id of the wayland surface which shouldnt be a problem here

nix/package.nix
pluiedev: It's usually withX11 or enableX11 in my experience

src/apprt/gtk/wayland.zig
pluiedev: This is fine if we do a roundtrip (which is blocking). I had run into bugs where a listener is accidentally using stale pointers, so we should definitely be cautious with that (maybe with a comment explaining why this is fine)

src/config/Config.zig
pluiedev: The config key's name is definitely a bit awkward though... `background-blur-radius = true` doesn't *quite* make sense

src/apprt/gtk/wayland.zig
mitchellh: If a future use case is going to require a stable pointer anyways, we should go ahead and do it now. Bolting on a pointer stability later kind of sucks. We've done it a lot in the codebase.

src/config/Config.zig
mitchellh: That's right. Let me think about this a bit more. I'm not ready to accept two configs.

nix/package.nix
mitchellh: I think pluie is a Nixpkgs maintainer, no? In that case, I trust your judgement more on this one. I don't care what the wording is as long as its idiomatic and consistent.

src/config/Config.zig
pluiedev: To me the best option is to just have one option called `background-blur`: `background-blur = true` and `background-blur = 20` both make sense IMO

nix/package.nix
Mic92: Can confirm that camel casing and enable prefixes are commonly used to differentiate boolean options from packages.

pluiedev:
> We should consider resource management here for example do we need to destroy the blur manager on close.

You can't do that, actually. There isn't a destructor. I've looked at how [winit](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/wayland/window/state.rs#L1080) does it (Apache 2.0 code so I think it's fine) and they only clean up the blur token.

> And do we need to handle the display disconnect or does gtk do that I think you can use WAYLAND_DEBUG=1 to see the Wayland messages which should have the destroy and disconnect stuff there

They're all managed by GTK (GDK, to be more precise), so we don't have to worry about that at all

build.zig
```diff
@@ -139,18 +138,31 @@ pub fn build(b: *std.Build) !void {
         switch (term) {
             .Exited => |code| {
                 if (code == 0) {
-                    if (std.mem.indexOf(u8, stdout.items, "x11")) |_| break :x11 true;
-                    break :x11 false;
+                    if (std.mem.indexOf(u8, stdout.items, "x11")) |_| x11 = true;
+                    if (std.mem.indexOf(u8, stdout.items, "wayland")) |_| wayland = true;
+                } else {
+                    std.log.warn("pkg-config: {s} with code {d}", .{ @tagName(term), code });
+                    return error.Unexpected;
```
mitchellh, line 145: This should not return an error since it makes `zig build --help` fail if `pkg-config` is not available.

src/apprt/gtk/wayland.zig
```diff
@@ -0,0 +1,125 @@
...
+const build_options = @import("build_options");
+
+const log = std.log.scoped(.gtk_wayland);
+
+/// Wayland state that contains application-wide Wayland objects (e.g. wl_display).
+pub const AppState = struct {
+    display: *wl.Display,
+    blur_manager: ?*org.KdeKwinBlurManager = null,
+
+    pub fn create(display: ?*c.GdkDisplay) ?AppState {
```
mitchellh, line 15: The `create` function idiomatically in Zig allocates and returns a pointer. This should be `init`.

src/apprt/gtk/wayland.zig
```diff
@@ -0,0 +1,125 @@
...
+            return null;
+
+        const wl_display: *wl.Display = @ptrCast(c.gdk_wayland_display_get_wl_display(display_) orelse return null);
+
+        return .{
+            .display = wl_display,
+        };
+    }
+
+    pub fn init(self: *AppState) !void {
```
mitchellh, line 35: This should assume the `self` pointer is unintiailzed memory and use `self.* = .{ ... }` to set it up. This will let you call `init` with uninitialized memory to setup the stable pointer.

mitchellh (changes requested):
I apologize but a few more requests here. This is close! Sorry for the delay.

src/config/Config.zig
mitchellh: I agree with this. I think we need to introduce a mechanism to rename. I can work on that. For the sake of this PR, please override `background-blur-radius`. We'll add the mechanism to rename and do the rename in a follow up PR. I don't want to block this for you.

mitchellh (approved):

mitchellh:
Automerge is on. Send it.

PR #4407, opened 2025-01-02
chore: fix typos

nhld:
Fixes:

- te -> the
- require you restart -> require you to restart
- neovim -> Neovim


qwerasd205:
"... may require you restart" is a perfectly valid construction. It's a subjunctive clause. It could do with a "that" after "require", but it's valid even without it.

PR #4421, opened 2025-01-02
gtk: fix non-notebook separator colors

jcollie:
Before: 
Screenshot From 2025-01-02 12-25-55
After:
Screenshot From 2025-01-02 12-28-02

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -99,6 +99,7 @@ pub fn init(self: *Window, app: *App) !void {
     self.window = gtk_window;
     c.gtk_window_set_title(gtk_window, "Ghostty");
     c.gtk_window_set_default_size(gtk_window, 1000, 600);
+    c.gtk_widget_add_css_class(@ptrCast(gtk_window), "terminal-window");
 
     // GTK4 grabs F10 input by default to focus the menubar icon. We want
     // to disable this so that terminal programs can capture F10 (such as htop)
--- a/src/apprt/gtk/notebook_adw.zig
+++ b/src/apprt/gtk/notebook_adw.zig
@@ -23,6 +23,7 @@ pub const NotebookAdw = struct {
         assert(adwaita.enabled(&app.config));
 
         const tab_view: *c.AdwTabView = c.adw_tab_view_new().?;
+        c.gtk_widget_add_css_class(@ptrCast(@alignCast(tab_view)), "notebook");
 
         if (comptime adwaita.versionAtLeast(1, 2, 0) and adwaita.versionAtLeast(1, 2, 0)) {
             // Adwaita enables all of the shortcuts by default.
--- a/src/apprt/gtk/notebook_gtk.zig
+++ b/src/apprt/gtk/notebook_gtk.zig
@@ -20,6 +20,8 @@ pub const NotebookGtk = struct {
 
         // Create a notebook to hold our tabs.
         const notebook_widget: *c.GtkWidget = c.gtk_notebook_new();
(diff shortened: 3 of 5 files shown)


tristan957:
Why do we have the `separator` CSS in the first place?

jcollie:
> Why do we have the `separator` CSS in the first place?

Because the separator disappears when you have `background-opacity` < 1 and the desktop behind leaks through. #2242

src/apprt/gtk/notebook.zig
```diff
@@ -27,6 +27,8 @@ pub const Notebook = union(enum) {
 
         // Create a notebook to hold our tabs.
         const notebook_widget: *c.GtkWidget = c.gtk_notebook_new();
+        c.gtk_widget_add_css_class(notebook_widget, "notebook");
```
tristan957, line 30: I think a good refactor here would be to add this call in the `create()` function.

src/apprt/gtk/Window.zig
```diff
@@ -99,6 +99,7 @@ pub fn init(self: *Window, app: *App) !void {
     self.window = gtk_window;
     c.gtk_window_set_title(gtk_window, "Ghostty");
     c.gtk_window_set_default_size(gtk_window, 1000, 600);
+    c.gtk_widget_add_css_class(@ptrCast(gtk_window), "ghostty");
```
tristan957, line 102: Hmm, I wonder if a more specific CSS name would be more useful.

tristan957 (approved):

src/apprt/gtk/Window.zig
mitchellh: I agree, I think `terminal-window` may be better (because there may be other types of windows in the future).

src/apprt/gtk/notebook_adw.zig
```diff
@@ -23,6 +23,7 @@ pub const NotebookAdw = struct {
         assert(adwaita.enabled(&app.config));
 
         const tab_view: *c.AdwTabView = c.adw_tab_view_new().?;
+        c.gtk_widget_add_css_class(@ptrCast(@alignCast(tab_view)), "notebook");
```
mitchellh, line 26: I'm shocked this is needed, but I'm guessing it's to unify between GTK/Adw? Btw I'm fine with generic "notebook" here because I think the specificity will come from the parent containing the notebook.

mitchellh (changes requested):

src/apprt/gtk/notebook_adw.zig
jcollie: It's so that I can distinguish separators inside the notebook (i.e. between splits) and separators elsewhere in the UI (i.e. in menus).

mitchellh: No I mean I'm more shocked that you can't select on element type without a class, i.e.:

```
.ghostty Notebook {}
```

But I guess GTK doesn't expose them like this?

jcollie: Oh, yeah you can select based on element type. Maybe there's a better way to do it but I'm not a CSS expert by any means.

mitchellh:
Due to the other approvals, queued for merge once CI passes. Thanks!

mitchellh (approved):

src/apprt/gtk/notebook_adw.zig
jcollie: Ah, I remember, this _is_ because of the gtk/adw split - we use different elements depending on the flavor so we need a class.

mitchellh: Yep that's what I figured, makes sense!

PR #4422, opened 2025-01-02
surface: add timer-based scrolling during selection

moni-dz:
Adds a timer to continuously scroll during selection when outside the viewport, 15ms per line.

Currently the scrolling behavior requires you to jiggle the mouse to continuously scroll upwards/downwards when selecting text.

### Before

https://github.com/user-attachments/assets/18e6c547-ed04-4098-88b4-35360f8c8c3c

### After

https://github.com/user-attachments/assets/46d5a6fc-b38e-46cf-b00f-52c8bc289f52

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -138,6 +138,9 @@ child_exited: bool = false,
 /// to let us know.
 focused: bool = true,
 
+/// Used to determine whether to continuously scroll.
+selection_scroll_active: bool = false,
+
 /// The effect of an input event. This can be used by callers to take
 /// the appropriate action after an input event. For example, key
 /// input can be forwarded to the OS for further processing if it
@@ -944,6 +947,34 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
             ) catch |err| {
                 log.warn("apprt failed to ring bell={}", .{err});
             };
+        },
+
+        .selection_scroll => |active| {
+            self.selection_scroll_active = active;
+
+            if (self.selection_scroll_active) {
+                const pos = try self.rt_surface.getCursorPos();
+                const pos_vp = self.posToViewport(pos.x, pos.y);
+                const screen = &self.renderer_state.terminal.screen;
+                const delta: isize = if (pos.y < 0) -1 else 1;
+
+                try self.io.terminal.scrollViewport(.{ .delta = delta });
+
+                // Always the case, but doesn't hurt to check
+                if (self.mouse.left_click_count == 1) {
+                    const pin = screen.pages.pin(.{
+                        .viewport = .{
+                            .x = pos_vp.x,
+                            .y = pos_vp.y,
+                        },
+                    }) orelse {
+                        if (comptime std.debug.runtime_safety) unreachable;
+                        return;
+                    };
+
+                    try self.dragLeftClickSingle(pin, pos.x);
+                }
+            }
         },
     }
 }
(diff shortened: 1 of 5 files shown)


mitchellh (changes requested):
Thanks for taking this on.

I do not like the idea that the renderer thread is controlling the viewport in this way or modifying the terminal at all. The renderer thread should, philosophically, be a read-only consumer of the terminal.

I like that by putting it in a thread with libxev you've made this cross-platform and just work, though.

I wonder if this should perhaps go into the IO thread instead, which spiritually owns the terminal anyways.

moni-dz:
> I do not like the idea that the renderer thread is controlling the viewport in this way or modifying the terminal at all. The renderer thread should, philosophically, be a read-only consumer of the terminal.
> 
> I wonder if this should perhaps go into the IO thread instead, which spiritually owns the terminal anyways.

I have moved the logic to the Termio thread, that seems to be the right call as sending mailbox messages became cleaner.

moni-dz:
Now the selection region keeps up with the scrolling.

src/Surface.zig
```diff
@@ -3383,6 +3389,12 @@ pub fn cursorPosCallback(
         self.renderer_state.terminal.screen.dirty.hyperlink_hover = true;
     }
 
+    // Stop selection scrolling when inside the viewport within a 1px buffer
+    // for fullscreen windows, but only when selection scrolling is active.
+    if (pos.x >= 1 and pos.y >= 1 and self.io_thread.scroll_active) {
```
mitchellh, line 3394: Accessing the IO thread state here is a data race and is unsafe. I think we need to store this state on the Surface itself.

src/termio/Thread.zig
```diff
@@ -419,3 +440,65 @@ fn stopCallback(
...
+        error.Canceled => {},
+        else => {
+            log.warn("error during selection scroll callback err={}", .{err});
+            return .disarm;
+        },
+    };
+
+    const cb = cb_ orelse return .disarm;
+    const surface = cb.io.surface_mailbox.surface;
+    const pos = try surface.rt_surface.getCursorPos();
```
mitchellh, line 472: These methods on RT surface expect to be called from the Surface thread, so they aren't thread safe. We can't call this.

src/termio/Thread.zig
```diff
@@ -419,3 +440,65 @@ fn stopCallback(
...
+            .viewport = .{
+                .x = pos_vp.x,
+                .y = pos_vp.y,
+            },
+        }) orelse {
+            if (comptime std.debug.runtime_safety) unreachable;
+            return .rearm;
+        };
+
+        surface.dragLeftClickSingle(pin, pos.x) catch {};
```
mitchellh, line 491: Similarly, this is unsafe to call from this thread.

mitchellh (changes requested):
Thanks for the work. There are some major thread safety issues with this invoking unsafe or undefined behavior. We need to rethink how to approach this. Unfortunately, this isn't the way. 

One approach we may need to take is to run the timer on the IO thread but send a message to the Surface (GUI thread) to invoke the scroll rather than doing it directly. It's a bit of a message soup but this isn't a super common activity so I think it won't negatively impact performance too much.

mitchellh (approved):
Thanks for doing this and addressing feedback. I rebased it and resolved some issues I noticed:

* We were accessing shared state in Surface.zig without the lock
* We shouldn't do a renderer wakeup from Termio because it may actually render before our Surface thread updates so we should trigger that from the surface
* Micro optimizations around conditionals and shared locks being held

PR #4436, opened 2025-01-02
macos: make auto-update optional

### LINKED ISSUE #4433: Change `auto-update` to optional, default to null
Change the `auto-update` config[^1] to an optional `?AutoUpdate` and default it to "null".

The behavior when "null" (unset) is that we defer to the Sparkle default behavior. This will allow users to override this using standard `defaults`. I think this will also incidentally change it so that on first run Sparkle asks the user for permission to check for updates instead of us pinging right away (which should alleviate some users privacy concerns although we don't track that data). 

This will also make it so the checkbox during update dialogs work.

[^1]: https://ghostty.org/docs/config/reference#auto-update

jparise:
When unset, we use Sparkle's default behavior, which is based on the user's preference stored in the standard user defaults.

The rest of the previous behavior is preserved:
- When SUEnableAutomaticChecks is explicitly false, auto-updates are disabled.
- When 'auto-update' is set, use its value to set Sparkle's auto-update behavior.

Fixes #4433

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -486,15 +486,16 @@ class AppDelegate: NSObject,
 
         // Sync our auto-update settings. If SUEnableAutomaticChecks (in our Info.plist) is
         // explicitly false (NO), auto-updates are disabled. Otherwise, we use the behavior
-        // defined by our "auto-update" configuration.
-        if Bundle.main.infoDictionary?["SUEnableAutomaticChecks"] as? Bool != false {
-            updaterController.updater.automaticallyChecksForUpdates =
-                config.autoUpdate == .check || config.autoUpdate == .download
-            updaterController.updater.automaticallyDownloadsUpdates =
-                config.autoUpdate == .download
-        } else {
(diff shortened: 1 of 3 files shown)


macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -486,15 +486,16 @@ class AppDelegate: NSObject,
...
-        // defined by our "auto-update" configuration.
-        if Bundle.main.infoDictionary?["SUEnableAutomaticChecks"] as? Bool != false {
-            updaterController.updater.automaticallyChecksForUpdates =
-                config.autoUpdate == .check || config.autoUpdate == .download
-            updaterController.updater.automaticallyDownloadsUpdates =
-                config.autoUpdate == .download
-        } else {
+        // defined by our "auto-update" configuration (if set) or fall back to Sparkle
+        // user-based defaults.
+        if Bundle.main.infoDictionary?["SUEnableAutomaticChecks"] as? Bool == false {
```
jparise, line 491: Unlike some other languages, it's nice that Swift will only evaluate the lhs to `== false` if it's non-nil.

mitchellh:
Perfect

PR #4443, opened 2025-01-02
WIP: apprt/gtk: fix equalizing when doubleclicking near the divider

### LINKED ISSUE #4354: GTK: Splits are equalized when double clicking anywhere, not just the divider
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4352

<sup>Originally posted by **Opposite34** January  1, 2025</sup>
Referring to #3557

Some people brought up that the splits are equalized when double clicked anywhere on the window, which of course shouldn't be done as some TUIs rely on double clicks to work properly (selecting words in some editors, etc). 

In the PR above, the author mentioned double clicking on a handle bar (aka divider) to equalize the splits. However, the `gtkMouseDown` callback they wrote seems to just check for double clicks without checking the mouse position:

```zig
fn gtkMouseDown(
    _: *c.GtkGestureClick,
    n_press: c.gint,
    _: c.gdouble,
    _: c.gdouble,
    ud: ?*anyopaque,
) callconv(.C) void {
    if (n_press == 2) {
        const self: *Split = @ptrCast(@alignCast(ud));
        _ = equalize(self);
    }
}
```</div>

### Solution

Double-clicking should only apply to the split divider.

Opposite34:
This PR aims to fix #4354 .

## Before
[Before](https://github.com/user-attachments/assets/447c61d4-2562-4cc7-8d14-950dacf06d08)

## After
[After](https://github.com/user-attachments/assets/2f4b10df-94fa-4973-8055-f991c4529ad5)

As you can see in the fix, one problem is that sometimes the resizing is bounced back to the original sizing (I didn't show it here, but this happens prior to this fix as well). My suspicion is with the race condition of doubleclick to equalize (this PR) and with the dragging of the divider to resize the panes. However, the latter might be a GTK feature we just use (?) as I couldn't find it anywhere in the codebase...

Apart from that, the function `equalizeTopSplit` I added in this fix, which climbs up the split parents to the topmost one, seems to partially function similarly to `equalizeSplits` inside `apprt/gtk/App.zig` (in the app we can just get the topmost split, while here we have to climb up as the Splits only have the knowledge of its parent). I want to prevent code repetition, so some advice here will also be nice as well.

### DIFF
--- a/src/apprt/gtk/Split.zig
+++ b/src/apprt/gtk/Split.zig
@@ -111,6 +111,15 @@ pub fn init(
     // Keep a long-lived reference, which we unref in destroy.
     _ = c.g_object_ref(paned);
 
+    // Clicks
+    const gesture_click = c.gtk_gesture_click_new();
+    errdefer c.g_object_unref(gesture_click);
+    c.gtk_event_controller_set_propagation_phase(@ptrCast(gesture_click), c.GTK_PHASE_CAPTURE);
+    c.gtk_gesture_single_set_button(@ptrCast(gesture_click), 1);
+    c.gtk_widget_add_controller(paned, @ptrCast(gesture_click));
+    // Signals
+    _ = c.g_signal_connect_data(gesture_click, "pressed", c.G_CALLBACK(&gtkMouseDown), self, null, c.G_CONNECT_DEFAULT);
+
     // Update all of our containers to point to the right place.
     // The split has to point to where the sibling pointed to because
     // we're inheriting its parent. The sibling points to its location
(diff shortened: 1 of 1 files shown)


Opposite34:
In the [Discord server](https://discord.com/channels/1005603569187160125/1320882404717625374/1324556131867164724), Kat has confirmed that this work on their system without the size rebouncing back I and some other people are having. I am wondering if a new discussion should be created regarding that issue or not, as it seems like that requires more of knowing how to work with gtk which I currently don't have.

Still also want some comments regarding `equalizeTopSplit` and the `equalizeSplits` inside `apprt/gtk/App.zig` on what approach we should do to organize these properly.

mitchellh:
Note I reverted https://github.com/ghostty-org/ghostty/commit/bec46fc2fcb8d0ec02b56472b4bf9f21139e5f45

I'd still like double clicking dividers to equalize. That is something I've actually tried before and wanted to work, and something that I've found generally works in other "divided" programs.

Opposite34:
> Note I reverted [bec46fc](https://github.com/ghostty-org/ghostty/commit/bec46fc2fcb8d0ec02b56472b4bf9f21139e5f45)
> 
> I'd still like double clicking dividers to equalize. That is something I've actually tried before and wanted to work, and something that I've found generally works in other "divided" programs.

Good to know. In that case I will keep the issue as one then. Will rebase this later and figure out a way for double clicking the divider itself to not be buggy.

Opposite34:
I have some updates. The bug with the rebouncing only occurs when double-clicking at `GtkPanedHandle`, which is supposedly the divider itself. 

However, when looking through the GTK debug inspector (`GTK_DEBUG=interactive ./zig-out/bin/ghostty`), this `GtkPanedHandle` not actually centered to the divider, but skewed to the right or the bottom of the split depending on the splitted direction. 

Moreover, `GtkPanedHandle` itself doesn't seems to response to gesture click properly. I am not sure whether this was due a logical error in the code I tried it with (I'm gonna assume it is this one for now, and will try to make it work later), or whether `GtkPanedHandle` really can't emit signals from gesture click. Either way, because of the hitbox of this widget itself being skewed to one side, it might be a bit weird to double click this properly even if we can get it working later on.

I have also been chatting with the author of the original PR in [this thread](https://discord.com/channels/1005603569187160125/1321941528213717022) of the Ghostty Discord as they suggested trying to bind the gesture click signal they did to `GtkPanedHandle` above but to no avail.

mitchellh:
What's the status of this PR? Should it be reviewed still?

mitchellh:
Closing due to no response

PR #4444, opened 2025-01-02
feat: Improve invalid keybind error messages

otherJL0:
Implementation of [Better keybind error messages](https://github.com/ghostty-org/ghostty/discussions/3732) discussion

### DIFF
--- a/src/cli/args.zig
+++ b/src/cli/args.zig
@@ -7,6 +7,7 @@ const diags = @import("diagnostics.zig");
 const internal_os = @import("../os/main.zig");
 const Diagnostic = diags.Diagnostic;
 const DiagnosticList = diags.DiagnosticList;
+const binding = @import("../input/Binding.zig");
 
 // TODO:
 //   - Only `--long=value` format is accepted. Do we want to allow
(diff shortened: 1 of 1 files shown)


otherJL0:
This PR is a clean implementation of https://github.com/ghostty-org/ghostty/pull/3727 after mucking up the git history

mitchellh (changes requested):
Thanks for this PR. I'm not happy with this approach. I recognize it works, and I recognize the effort, so it does pain me to say this but I'm going to reject this PR. My notes below...

First, I think the errors in general are too generic to add to `cli/args.zig`. I would recommend re-wrapping them with something like `BindingInvalidFormat`, `BindingInvalidAction`, etc.

Still, even with that, it doesn't feel like `cli/args` should _know_ about these errors. Just spitballing in my head (not sure if its a good approach) but one thing I would try is having a function on `T` (in `parse`) that can take an error type `E` and format it. For example:

```
pub fn formatError(alloc: Allocator, err: FormatableError) { ... }
```

And then using `hasDecl` and some basic Zig reflection to determine if the error matches this error.

PR #4451, opened 2025-01-03
kittygfx: Ensure temporary files are named per spec

dgl:
Temporary files used with Kitty graphics must have "tty-graphics-protocol" somewhere in their full path.

https://sw.kovidgoyal.net/kitty/graphics-protocol/#the-transmission-medium


mitchellh:
Was this a recent change? I'm shocked I never ran into this. I believe we are able to unit test this so I want to look into that but otherwise looks good.

dgl:
@mitchellh the tests didn't use a magic name in them and blame says they haven't been touched since June. Added a negative test too.

mitchellh:
Thank you!

PR #4456, opened 2025-01-03
Correct typos and update typos.toml

liby:
I noticed that the version of `typos` in nixpkgs is still at [1.28.4](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/ty/typos/package.nix#L12) (as of [2 weeks ago](https://github.com/crate-ci/typos/releases/tag/v1.28.4)), while the latest version is [1.29.3](https://github.com/crate-ci/typos/releases/tag/v1.29.3). For tools like this where the dictionary is constantly being updated, we might want to consider using `uses: crate-ci/typos@master` directly in our CI workflow instead of the version from nixpkgs.

I understand the current setup using `devShell.nix` helps maintain consistency between local development and CI environments. However, I found some typos that can be detected by the new version of typos, but were missed by the old version of typos on nixpkgs.

image

## Proposal:

1. Use `crate-ci/typos` action in CI while keeping nixpkgs version for local development (catching more typos in CI but introducing some inconsistency)

### DIFF
--- a/src/renderer/OpenGL.zig
+++ b/src/renderer/OpenGL.zig
@@ -146,7 +146,7 @@ image_bg_end: u32 = 0,
 image_text_end: u32 = 0,
 image_virtual: bool = false,
 
-/// Defererred OpenGL operation to update the screen size.
+/// Deferred OpenGL operation to update the screen size.
 const SetScreenSize = struct {
     size: renderer.Size,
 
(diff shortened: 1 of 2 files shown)


mitchellh:
> I understand the current setup using devShell.nix helps maintain consistency between local development and CI environments. However, I found some typos that can be detected by the new version of typos, but were missed by the old version of typos on nixpkgs.

The issue with that is I can't run `typos` locally (without installing software outside of Nix which is bleh). I'm happy to just stay on the old version. We can overlay `Nixpkgs-unstable` for `typos` which will get us at least the latest that Nix supports....

PR #4464, opened 2025-01-03
gtk: add option for numeric icons on tabs

jcollie:
Screenshot From 2025-01-02 22-59-06


tristan957:
I don't love this particular solution, but would be interested in unlocking this feature for people. I think templating in the window (surface) title would make more sense.

jcollie:
> I don't love this particular solution, but would be interested in unlocking this feature for people. I think templating in the window (surface) title would make more sense.

The big reason that I switched to an icon was that I had better control over the looks of the icon. I initially tried using unicode chars like "①" but they were a little small and didn't look very good in the default Cantarell font. By using an icon we can have much better control over it's appearance, maybe even dynamically generating it since we _do_ have the machinery for that.

tristan957:
But why not just use ascii numbers? Imagine a config like:

```
tab-title-prefix = {tab-index} {emdash}
```

Something stupid like that.

jcollie:
> But why not just use ascii numbers? Imagine a config like:
> ```
> tab-title-prefix = {tab-index} {emdash}
> ```
> Something stupid like that.

Because I wanted something that would be visually distinct from the tab title. If it's just text in the same font as the tab title that's much harder to do.

Also, there's the possibility of dynamically generating the icons. Some possible uses are spinners/beachballs/progress meters for the ConEmu OSC 9;4 codes. Probably a lot of other creative uses we can come up with.

mitchellh (commented):
Fwiw, I'm supportive of this as long as it's default-off. I think the non-standard icons and visual noise might be a bit jarring to people. But also, this feature has been supported by a few users so I don't see issue in bringing it in.

I leave it to the subsystem maintainers, just want to note its not a veto for me.

jcollie:
OK, this PR has been updated for zig-gobject and also I've added a keybind action for toggling the icons. The config is much simpler too, with just a boolean (default false).

pluiedev (changes requested):

src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp
```diff
@@ -67,6 +67,11 @@ menu menu {
         label: _("Close Tab");
         action: "win.close-tab";
       }
+
+      item {
+        label: _("Toggle Tab Icons");
+        action: "win.toggle-tab-icons";
+      }
```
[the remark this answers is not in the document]
jcollie: Removed this menu item.

src/config/Config.zig
```diff
@@ -2254,6 +2254,17 @@ keybind: Keybinds = .{},
...
+/// If `true`, show a numeric icon for the first 10 tabs. Tabs are numbered
+/// 1-10.
+///
+/// By default, Ghostty has keybinds for directly switching to tabs 1-8, using
+/// `Alt+N` (`Cmd+N` on macOS). `Alt+9` (`Cmd+9` on macOS) is configured by
+/// default to go to the last tab. These keybinds always work and are not
+/// affected by the state of the tabs icon.
+///
+/// The default is `false` (no icons will be shown).
+@"gtk-tab-icons": bool = false,
```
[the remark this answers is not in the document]
jcollie: Switched to `window-tab-icons`

src/apprt/gtk/TabView.zig
```diff
@@ -90,10 +105,37 @@ pub fn asWidget(self: *TabView) *gtk.Widget {
...
+                    .{position + 1},
+                ) catch unreachable;
+                const icon = gio.ThemedIcon.new(name);
+                page.setIcon(icon.as(gio.Icon));
+            } else {
+                page.setIcon(null);
+            }
+        } else {
+            page.setIcon(null);
+        }
```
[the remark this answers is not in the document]
jcollie: Done.

src/apprt/gtk/gresource.zig
```diff
@@ -53,6 +53,20 @@ const icons = [_]struct {
     },
 };
 
+const numeric_icons = [_][]const u8{
+    "0",
```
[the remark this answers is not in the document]
jcollie: Removed.

jcollie:
In addition to addressing @pluiedev 's comments, I added the ability to generate custom icons if your version of GTK is new enough. That way all tabs are numbered, not just the first 10.

src/apprt/gtk/gresource.zig
```diff
@@ -121,12 +134,18 @@ pub fn main() !void {
...
             "    <file alias=\"{s}/apps/com.mitchellh.ghostty.png\">images/icons/icon_{s}.png</file>\n",
             .{ icon.alias, icon.source },
         );
     }
+    for (numeric_icons) |icon| {
+        try writer.print(
+            "    <file alias=\"scalable/status/numeric-{0s}-circle-outline-symbolic.svg\">images/icons/numeric-{0s}-circle-outline-symbolic.svg</file>\n",
+            .{icon},
+        );
+    }
```
pluiedev, line 148: Why not: (you could extract the 10 out to another variable if you want)
```suggestion
    for (0..10) |idx| {
        try writer.print(
            "    <file alias=\"scalable/status/numeric-{0}-circle-outline-symbolic.svg\">images/icons/numeric-{0}-circle-outline-symbolic.svg</file>\n",
            .{idx + 1},
        );
    }
```

src/apprt/gtk/TabView.zig
```diff
@@ -90,10 +107,83 @@ pub fn asWidget(self: *TabView) *gtk.Widget {
...
+                        }),
+                        .number = number,
+                    },
+                ) catch unreachable;
+
+                const bytes = glib.Bytes.new(svg.ptr, svg.len);
+                defer bytes.unref();
+
+                break :icon gio.BytesIcon.new(bytes).as(gio.Icon);
+            }
```
pluiedev, line 167: I think for clarity it's better if this is separated out into its own function. Bonus points if the SVG template can be used by the gresource generation code to generate the 10 default icons, so that we don't have to commit them to the repo :D

pluiedev (approved):
All I have are suggestions for minor improvements/nitpicks now. GJ

pluiedev:
Also FYI libadwaita's tab button widget uses a custom [`adw-tab-counter-symbolic` icon](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/icons/scalable/status/adw-tab-counter-symbolic.svg) that looks more like a rounded square - it's also limited to double digits, with 100+ tabs appearing like [an infinity symbol](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/icons/scalable/status/adw-tab-overflow-symbolic.svg). I'm not saying that we should copy their designs wholesale (I'm not even sure if we can do that), but having that design uniformity would be nice. A rounded square/squircle would make the text more legible anyway.

src/apprt/gtk/gresource.zig
jcollie: Actually, after doing a bit more research, I _think_ that we can support generating the SVGs even back to Debian Bookworm. Difficult to verify 100%. I just opted to drop the static SVGs altogether.

src/apprt/gtk/TabView.zig
jcollie: After dropping the static SVGs the code is much cleaner. I don't think that we need to break this out into a separate function at this point unless we need to reuse the code somewhere else.

src/config/Config.zig
```diff
@@ -1381,6 +1381,22 @@ keybind: Keybinds = .{},
 ///   * `end` - Insert the new tab at the end of the tab list.
 @"window-new-tab-position": WindowNewTabPosition = .current,
 
+/// If `true`, show a numeric icon on the tabs that indicate the position of the
+/// tab. Tabs are numbered starting at 1. On GTK the icons appear as a number
+/// inside a circle on the left side of the tab. On macOS the icons appear as
+/// `⌘N` on the right side of the tab.
```
mitchellh, line 1387: Note this isn't exactly correct for macOS. On macOS, the tab accessory label (as it's called) shows the keyboard shortcut bound to `goto_tab:N`. It has nothing to do with the numeric position of the tab except by relation to that binding _usually_ being mapped that way.

src/config/Config.zig
```diff
@@ -1381,6 +1381,22 @@ keybind: Keybinds = .{},
...
+/// By default, Ghostty has keybinds for directly switching to tabs 1-8, using
+/// `Alt+N` (`Cmd+N` on macOS). `Alt+9` (`Cmd+9` on macOS) is configured by
+/// default to go to the last tab. These keybinds always work and are not
+/// affected by the state of the tabs icon.
+///
+/// The default is `false` (no icons will be shown) except on macOS where the
+/// default is `true`.
+///
+/// Disabling the icons does not work on macOS.
+@"window-tab-icons": bool = if (builtin.os.tag.isDarwin()) true else false,
```
mitchellh, line 1398: If we really want this feature, I kind of feel we should make it a `gtk-`-prefixed feature. It doesn't make sense to me on macOS, I don't know any application that has anything like this.

mitchellh (changes requested):
I still don't really love this feature, but I've seen enough requests for it on the Linux side I'm willing to accept it. Note my comments though.

src/config/Config.zig
jcollie: Switched back to `gtk-` prefix.

src/config/Config.zig
jcollie: Removed references to macOS.

tristan957:
> I still don't really love this feature, but I've seen enough requests for it on the Linux side I'm willing to accept it. Note my comments though.

Maybe we should start at the beginning with answering, "what is this trying to solve".

jcollie:
> Maybe we should start at the beginning with answering, "what is this trying to solve".

I think basically it makes it easier to use `Alt+N` to switch to a specific tab. You can quickly see what position a tab is in, esp if the window is narrow enough that not all of the tab headers are shown so you can't just count them.

tristan957:
> > Maybe we should start at the beginning with answering, "what is this trying to solve".
> 
> I think basically it makes it easier to use `Alt+N` to switch to a specific tab. You can quickly see what position a tab is in, esp if the window is narrow enough that not all of the tab headers are shown so you can't just count them.

Do the `alt+<num>` shortcuts work with numbers larger than 9?

mitchellh:
> Do the alt+<num> shortcuts work with numbers larger than 9?

No, keyboard shortcuts have to map to keys (or code points) that a keyboard can produce, especially since #7320 but that doesn't really change the core point here.

For this reason, macOS doesn't even label tabs beyond 9. Example:

CleanShot 2025-05-13 at 15 41 44@2x

> [!NOTE]
>
> It actually doesn't label beyond _8_, because on macOS its standard that cmd+9 goes to the _last tab_ no matter what.

jcollie:
> > > Maybe we should start at the beginning with answering, "what is this trying to solve".
> > 
> > 
> > I think basically it makes it easier to use `Alt+N` to switch to a specific tab. You can quickly see what position a tab is in, esp if the window is narrow enough that not all of the tab headers are shown so you can't just count them.
> 
> Do the `alt+<num>` shortcuts work with numbers larger than 9?

Not directly, but you can do hacky things like:

 ```
keybind=alt+1>1:goto_tab:11
keybind=alt+1>2:goto_tab:12
keybind=alt+1>3:goto_tab:13
…
```

No one would really probably do that IRL though. Perhaps down the road we can have a keybind action that pops up a dialog and requests input, or some other better way of doing that.

pluiedev:
One of the ideas I've had for a while is a "navigation mode" that would allow contextual display of indices to jump to, which generalizes over both tab indices and split indices. 

For example when you enter tab navigation with some keybind (say alt+shift+t), the tab titles become replaced by the tab index so you know where to jump to via any of the number keys. This can be undone by just pressing Esc. Similarly if you enter split navigation (perhaps alt+shift+s), there'll be an overlay over each split in the current tab, and pressing any number between 1-9 will take you to that split. This UI pattern could then be extended into jumping based on title search (a request I've seen before) or some other special marker.

jcollie:
I think that I'm going to close this one out. I rarely use tabs myself so I have little need for it and using the tab icons seems to have sparked little joy in anyone else.

chenzhiwei:
Hi @jcollie, thank you for this PR, it is really useful to switch tabs by `alt+num`.

Is there an alternative way to implement this? Such as using `title = "{index} {title}"` to add an index to the tab?

PR #4465, opened 2025-01-03
apprt/gtk: continue cleanup of window-decoration code

tristan957:
Remove all window corner artifacting. Now we also respond to changes when the window becomes decorated or not.

### DIFF
--- a/src/apprt/gtk/style.css
+++ b/src/apprt/gtk/style.css
@@ -33,7 +33,11 @@ label.size-overlay.hidden {
   opacity: 0;
 }
 
-window.without-window-decoration-and-with-titlebar {
+window.ssd.no-border-radius {
+  /* Without clearing the border radius, at least on Mutter with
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/style.css
```diff
@@ -33,7 +33,11 @@ label.size-overlay.hidden {
   opacity: 0;
 }
 
-window.without-window-decoration-and-with-titlebar {
+window.ssd.no-border-radius {
```
tristan957, line 36: I could also just make this `window.no-border-radius`

jcollie:
Interesting. I don't see anything obviously wrong but then this seems to be a deep dark corner of GTK that is not well documented. Tomorrow I'm going to resurrect my project to try out `river` and experiment with this.

mitchellh (approved):
Looks good to me. This is still marked as WIP so let me know when you're happy with it and I'll press the merge.

PR #4501, opened 2025-01-03
fix: macos incorrect quick terminal position

acehinnnqru:
As this discussion: #4353


mitchellh:
Thanks for doing this, the math looks right to me and the quick terminal works on my display.

PR #4511, opened 2025-01-03
[macOS] Adopt @Observable in SwiftUI code

harlanhaskins:
This moves off of `ObservableObject` and `@Published` over to `@Observable`, which is preferred as it allows views to observe changes on a per-property basis rather than an entire object at a time.

I went through the app and tried to compare behavior where I could, and I couldn't spot any visual differences, but I am sure there's good chance I have missed something.

This also requires updating the minimum macOS deployment target to 14.0. If that is unacceptable, I am happy just closing this PR. Increasing the deployment target required updating some `onChange` callsites to avoid depreciation warnings.

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalView.swift
+++ b/macos/Sources/Features/Terminal/TerminalView.swift
@@ -28,18 +28,18 @@ protocol TerminalViewDelegate: AnyObject {
 /// The view model is a required implementation for TerminalView callers. This contains
 /// the main state between the TerminalView caller and SwiftUI. This abstraction is what
 /// allows AppKit to own most of the data in SwiftUI.
-protocol TerminalViewModel: ObservableObject {
+protocol TerminalViewModel: AnyObject, Observable {
     /// The tree of terminal surfaces (splits) within the view. This is mutated by TerminalView
-    /// and children. This should be @Published.
+    /// and children.
     var surfaceTree: Ghostty.SplitNode? { get set }
 }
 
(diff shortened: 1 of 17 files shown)


macos/Sources/Features/Terminal/TerminalView.swift
```diff
@@ -86,32 +86,32 @@ struct TerminalView<ViewModel: TerminalViewModel>: View {
...
+                        self.delegate?.pwdDidChange(to: pwdURL)
                     }
-                    .onChange(of: cellSize) { newValue in
-                        guard let size = newValue else { return }
+                    .onChange(of: cellSize) {
+                        guard let size = cellSize else { return }
                         self.delegate?.cellSizeDidChange(to: size)
                     }
-                    .onChange(of: viewModel.surfaceTree?.hashValue) { _ in
+                    .onChange(of: viewModel.surfaceTree?.hashValue) {
```
harlanhaskins, line 107: Given that `Container` is now `@Observable`, this may no longer be required. A view will observe the properties of all `@Observable` instances it reads from during `body`, which means we no longer need to manually propagate changes up from child objects. I am not confident enough to remove this, however.

mitchellh:
macOS 14 as a minimum deployment target is unacceptable. We want to continue supporting all supported Apple platforms. That’s the reason I was holding back. Sorry! But thanks for trying…

harlanhaskins:
No worries, this was still a fun way to get acquainted with the codebase. Thanks for the quick reply!

PR #4531, opened 2025-01-03
font(coretext): add config to adjust strength of `font-thicken`.

qwerasd205:
This is achieved by rendering to an alpha-only context rather than a normal single-channel context, and adjusting the brightness at which CoreText thinks it's drawing the glyph, which affects how it applies font smoothing (which is what `font-thicken` enables).

### DIFF
--- a/src/font/face.zig
+++ b/src/font/face.zig
@@ -100,6 +100,15 @@ pub const RenderOptions = struct {
     ///
     /// This only works with CoreText currently.
     thicken: bool = false,
+
+    /// "Strength" of the thickening, between `0` and `255`.
+    /// Only has an effect when `thicken` is enabled.
+    ///
+    /// `0` does not correspond to *no* thickening,
+    /// just the *lightest* thickening available.
+    ///
+    /// CoreText only.
+    thicken_strength: u8 = 255,
 };
 
 test {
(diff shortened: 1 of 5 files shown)


qwerasd205:
# Screenshots

## `font-thicken = false`
no-thicken

## `font-thicken = true`

### `font-thicken-strength = 0`
thicken-0

### `font-thicken-strength = 127`
thicken-127

### `font-thicken-strength = 255`
thicken-255

src/font/face.zig
```diff
@@ -100,6 +100,15 @@ pub const RenderOptions = struct {
...
     thicken: bool = false,
+
+    /// "Strength" of the thickening, between `0` and `255`.
+    /// Only has an effect when `thicken` is enabled.
+    ///
+    /// `0` does not correspond to *no* thickening,
+    /// just the *lightest* thickening available.
+    ///
+    /// CoreText only.
+    thicken_strength: u8 = 255,
```
mitchellh, line 111: My only thought here is should we use a f32 here instead of a u8. It feels like this would be more user-friendly to configure for a range of "min" to "max".

mitchellh (changes requested):
One question/comment. I'm happy to do that work myself later.

src/font/face.zig
qwerasd205: I considered it but it felt like false precision to me-- I'm not particularly bothered about it either way though so I'm fine with changing it to a float if you want.

mitchellh: I can see the argument, but the argument types match and the difference between subtle changes should not be noticeable enough.

mitchellh: I take it back, this causes hashing issues and I don't want to deal with it lol.

PR #4544, opened 2025-01-03
core: function to detect the display type

jcollie:


src/os/desktop.zig
```diff
@@ -85,3 +85,30 @@ pub fn desktopEnvironment() DesktopEnvironment {
...
+
+/// Detect what desktop environment we are running under. This is mainly used on
+/// Linux to enable or disable GTK client-side decorations but there may be more
+/// uses in the future.
+pub fn displayType() DisplayType {
+    return switch (comptime builtin.os.tag) {
+        .macos => .macos,
+        .windows => .windows,
+        .linux => dt: {
+            if (@inComptime()) @compileError("Checking for the display type on Linux must be done at runtime.");
```
AlexJuca, line 105: @jcollie *suggestion*: Should word this as follows: "On Linux, the display type must be determined dynamically at runtime."

mitchellh:
What would we use this for? (Genuine question)

jcollie:
> What would we use this for? (Genuine question)

@pluiedev is working on patches to apply window blur on X11 and Wayland. If Ghostty is compiled with both X11 and Wayland support we need a runtime method to determine which blur methods to apply. I'm sure more will come up as we decide to talk more native Wayland protocols.

mitchellh:
Since we're using GTK, I'm not sure if this is the right way for us to detect this. 

I think getting the GDK display and using these defines might be better:

https://github.com/GNOME/gtk/blob/867d178e932d9872b853697ef3b59ed769410198/gdk/x11/gdkx11display.h#L46-L51

And the associated Wayland equivalents.

I think these are only available if GTK is compiled with X11/Wayland, so we'll have to use our build config to also assist. I'd feel more comfortable dumping this responsibility onto GTK since it's actually the one setting up our displays. No?

jcollie:
Hmm yeah. This would be especially important if the user for some reason forced Ghostty to run on XWayland.

PR #4549, opened 2025-01-03
Feat: display memory size in Terminal Inspector using bytes and kibibytes

AlexJuca:
This PR ensures that we can view all memory-related information in the Terminal Inspector in bytes and kibibytes for improved readability at a glance. 

Screenshot 2025-01-03 at 23 32 07


src/inspector/utils.zig
```diff
@@ -0,0 +1,3 @@
+pub fn toKiloBytes(bytes: usize) usize {
```
JakeWharton, line 1: This is kibibytes not kilobytes, and as a result should be displayed as KiB not KB.

src/inspector/page.zig
```diff
@@ -25,7 +27,7 @@ pub fn render(page: *const terminal.Page) void {
         }
         {
             _ = cimgui.c.igTableSetColumnIndex(1);
-            cimgui.c.igText("%d bytes", page.memory.len);
+            cimgui.c.igText("%d bytes (%d Kb)", page.memory.len, utils.toKiloBytes(page.memory.len));
```
JakeWharton, line 30: The "b" is cased incorrectly here.

src/inspector/utils.zig
AlexJuca: @JakeWharton Thanks for the heads up on this one. Since this is representing memory size I think kibibyte is the most appropriate.

AlexJuca:
This is how it looks after changing how we represent memory size (kilobytes to kibibytes)

Screenshot 2025-01-04 at 09 01 25

AlexJuca:
@mitchellh I wasn't too sure about using `utils` here. Do you have any suggestions on the naming and location of that file?

src/inspector/Inspector.zig
```diff
@@ -14,6 +14,7 @@ const input = @import("../input.zig");
 const renderer = @import("../renderer.zig");
 const terminal = @import("../terminal/main.zig");
 const inspector = @import("main.zig");
+const utils = @import("utils.zig");
```
AlexJuca, line 17: @mitchellh  I wasn't too sure about using utils here. Do you have any suggestions on the naming and location of that file?

mitchellh, line 17: Thanks for calling this out. I try to avoid `utils` packages since it becomes a dumping ground for anything and everything. I'd prefer naming this `units.zig`

mitchellh (changes requested):
Simple requests. Thanks!

src/inspector/Inspector.zig
AlexJuca: @mitchellh I had the same feeling initially. Cool, I've adjusted based on your feedback. 

Thanks a ton.

PR #4550, opened 2025-01-03
Migrate packaging guide to the website repository

Secrus:
See also: ghostty-org/website#256


mitchellh:
No, sorry, for reasons stated in the Discord. I think its important to keep this as text alongside the source.

PR #4560, opened 2025-01-04
gtk: enable window-title-font-family

jcollie:


src/apprt/gtk/App.zig
```diff
@@ -995,6 +995,31 @@ fn loadRuntimeCss(
...
+            \\
+            \\.inspector-window headerbar {{
+            \\  font-family: "{[font_family]s}";
+            \\}}
+            \\
+            \\.clipboard-confirmation-window headerbar {{
+            \\  font-family: "{[font_family]s}";
+            \\}}
+            \\
+            \\.config-errors-window headerbar {{
```
mitchellh, line 1012: Instead of all this duplication, maybe we should add another class ".window" to every window so that we can target _all windows_ easily too.

src/config/Config.zig
```diff
@@ -1073,7 +1073,12 @@ keybind: Keybinds = .{},
 
 /// The font that will be used for the application's window and tab titles.
 ///
-/// This is currently only supported on macOS.
+/// If this setting is left unset, the default font will be used. On macOS
+/// this is "SF Pro." On Gnome it is usually "Cantarell" but it is possible to
```
mitchellh, line 1077: I recommend removing what the default font is. Any system could easily change this and I don't want to deal with maintaining it. Your note that it is the system default is good enough I think.

mitchellh (changes requested):
Thanks for doing this, bringing parity between systems is great. 

Simple requests!

PR #4564, opened 2025-01-04
Handle `super+period` key combination on macOS

### LINKED ISSUE #4540: macOS: super+period doesn't reach Surface keyCallback
Discussion: #2881 

If you put a log into `Surface.keyCallback`, `super+period` doesn't create a key event. This is causing some keybinds to not work.

liby:
## Description

This PR fixes an issue where `super+period` key combination was not working properly on macOS.

The `performKeyEquivalent` method in the macOS implementation was restricting key combinations to only those that included the [Control modifier](https://github.com/ghostty-org/ghostty/blob/main/macos/Sources/Ghostty/SurfaceView_AppKit.swift#L813-L818). This meant that other valid key combinations like `super+period` were being blocked before they could be properly processed.

## Solution

Removed the Control modifier check in `performKeyEquivalent`, allowing other key combinations like `super+period` to be processed correctly. The key combination is now explicitly handled in the switch statement, which provides better control over which combinations are allowed.

https://github.com/user-attachments/assets/186ba3c8-aeac-478a-ab3d-88d078c5e69a

Fixes #4540

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -810,49 +810,65 @@ extension Ghostty {
                 return false
             }
 
-            // Only process keys when Control is active. All known issues we're
-            // resolving happen only in this scenario. This probably isn't fully robust
-            // but we can broaden the scope as we find more cases.
-            if (!event.modifierFlags.contains(.control)) {
-                return false
-            }
+            // First check if this is a special key combination we want to handle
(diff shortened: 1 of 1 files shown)


mitchellh:
This works but I do have a much more robust solution in the works that solves a bunch of issues at once (since they all stem from similar `performKeyEquivalent` handling). So, I'll keep this PR open in case that doesn't pan out but lets wait for that...

mitchellh:
Superceded by #4591

PR #4574, opened 2025-01-04
apprt/gtk: Add Close Window action to context menu

tdgroot:
This action already exists in the main menu. The Close Window action allows for better workflow and consistency.

Fixes #4198


mitchellh:
Rejected pending further discussion.

PR #4582, opened 2025-01-04
fix: `config-file` diagnostics when theme is set

### LINKED ISSUE #4509: Non-existent config-file does not show an error when `theme` is set
As reported on Discord, using a non-existent `config-file` without a leading `?` character when `theme` is set does not show an error message or log message (reproduced on both Linux and macOS).

I noted the same thing in https://github.com/ghostty-org/ghostty/pull/2258, but never made a formal issue to track it.

## Steps to reproduce

1. Create a config file with the following contents:

   ```
   # Doesn't matter which theme is used here
   theme = nord
   config-file = some_nonexistent_file
   ```

2. Restart Ghostty or reload the configuration

## Expected behavior

A "Configuration Errors" window should show up saying that `some_noneexistent_file` doesn't exist.

## Actual behavior

No window appears and no message appears in the log.

dmehala:
Fix a bug when diagnostics are lost when a theme is loaded.

### The issue
When a them is loaded, it initializes a new empty `Config`, apply the theme, then replays the user configuration on top of it. This behaviour is explained in detail here:

https://github.com/ghostty-org/ghostty/blob/8f5f432ab61fa73b77cb61239c0cff34250f77bb/src/config/Config.zig#L3192-L3205

The issue is because the `loadTheme` function overwrites the existing `Config` with its internal state, including diagnostics. As a result, previously generated diagnostics from [LoadRecursiveFiles](https://github.com/ghostty-org/ghostty/blob/8f5f432ab61fa73b77cb61239c0cff34250f77bb/src/config/Config.zig#L2991) are lost.

NOTE: This is one solution among others and may not be the most optimal approach.

Fixes #4509

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3207,6 +3207,11 @@ fn loadTheme(self: *Config, theme: Theme) !void {
     var new_config = try self.cloneEmpty(alloc_gpa);
     errdefer new_config.deinit();
 
+    // Clone our diagnostics if any.
+    if (!self._diagnostics.empty()) {
+        new_config._diagnostics = try self._diagnostics.clone(alloc_gpa);
+    }
+
     // Load our theme
     var buf_reader = std.io.bufferedReader(file.reader());
     const reader = buf_reader.reader();


src/config/Config.zig
```diff
@@ -3207,6 +3207,11 @@ fn loadTheme(self: *Config, theme: Theme) !void {
     var new_config = try self.cloneEmpty(alloc_gpa);
     errdefer new_config.deinit();
 
+    // Clone our diagnostics if any.
+    if (!self._diagnostics.empty()) {
+        new_config._diagnostics = try self._diagnostics.clone(alloc_gpa);
```
mitchellh, line 3212: I'm suspicious about this. I think it could result in duplicates since our replay should also contain invalid elements. I want to think about this more.

dmehala: Hmm, you're right—`try result.loadDefaultFiles(alloc_gpa);` is called before `finalize`, which can lead to duplicate diagnostics. I overlooked this possibility.

mitchellh: The goal with replay is to reproduce the exact same behavior as if it weren't a replay. So we should look into how we're doing optional files and ensure our replay is preserved.

anund:
Found this issue for a different reason: Found things for a different reason https://github.com/ghostty-org/ghostty/compare/main...anund:ghostty-reforked:validate_config_test

What I haven't had time to track down is whether it's necessary to teach locations to replay steps instead of just :crossed_fingers: remembering to copy diagnostics in all the places.

anund:
`zig build test -Dtest-filter="config.Config"` is useful for cleaning up the test output.

anund:
@dmehala, pushed a conditional state test. I'm happy if you just take what I learned and continue working on this.

dmehala:
Another possible solution would be to move the file existence check into `RepeatablePath.parseCLI`. This way, when the step is replayed, the diagnostic will be generated. However, this approach could be susceptible to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issues.

That said, I haven't found a satisfying solution. By its nature, `config-file` isn't replayable, as the configuration files may no longer exist after they've been loaded. IMO, diagnostics generated by [LoadRecusriveFiles](https://github.com/ghostty-org/ghostty/blob/7a27af8bfce9007e31be10aeafe7a968a66aa0b5/src/config/Config.zig#L2998) should be handled differently.

As for the issue reported by @anund, adding a replay location isn't very useful for the end user. A better approach would be to add the location in `Replay.Steps`, thus, diagnostics generated in a replay would be able to pick the original location.

mitchellh:
Closing since this isn't the right solution. Discussion can continue the original issue.

PR #4583, opened 2025-01-04
Improve the documentation for move_tab keybind action

komapa:
As a new user of ghostty, it was not intuitive to figure out how to provide the `offset` parameter. It makes sense when you look more at the rest of the options but I think we can still make these docs cleaner, like I have done in this PR.

Thanks!

### DIFF
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -329,7 +329,7 @@ pub const Action = union(enum) {
     goto_tab: usize,
 
     /// Moves a tab by a relative offset.
-    /// Adjusts the tab position based on `offset` (e.g., -1 for left, +1 for right).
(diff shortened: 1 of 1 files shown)


src/input/Binding.zig
```diff
@@ -341,7 +341,8 @@ pub const Action = union(enum) {
     /// the direction given. For example `new_split:up`. Valid values are left, right, up, down and auto.
     new_split: SplitDirection,
 
-    /// Focus on a split in a given direction. For example `goto_split:up`. Valid values are left, right, up, down, previous and next.
```
komapa, line 344: This is the only docs line that crosses the 120 char boundary. I know this is probably not a hard requirement but I thought it will make this more readable in the IDE.

mitchellh:
Thanks!

Issue #4506, opened 2025-01-03
Load configuration from XDG_CONFIG_DIRS


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4477

<div type='discussions-op-text'>

<sup>Originally posted by **vikanezrimaya** January  3, 2025</sup>
I want to be able to drop a system-wide config for Ghostty into `/etc/xdg`. `XDG_CONFIG_DIRS` is a variable listing additional locations for config files that could be checked if the appropriate config in `$XDG_CONFIG_HOME`/`~/.config` is not found, and by default it includes `/etc/xdg` (at least on my distribution of choice).

If ghostty would respect this directory (by either loading this config if no config exists in the home directory, or loading it before the user config, allowing user to override system-wide defaults), that would be nice. It looks like ghostty creates an empty default config if it doesn't exist in the home directory, so perhaps the second option is the most appropriate.</div>

### Solution

We should load from _both_ `XDG_CONFIG_DIRS` and `XDG_CONFIG_HOME` and override along the way.

jcollie:
For anyone that works on this, consider creating an iterator interface similar to the `LocationIterator` in `src/config/theme.zig` that iterates through all of the possible locations in order of preference (including the macOS Application Support directory). Eventually we'll want to include Windows' equivalent locations.

danudey:
It's worth noting that, [per the spec](https://specifications.freedesktop.org/basedir-spec/latest/#basics), `$XDG_CONFIG_DIRS` can be a colon-delimited array of paths (similar to `PATH`). Technically this means that we should check each of the directories specified there (if there are more than one) for config files.

Currently the code in `internal_os.xdg.dir()` seems to assume:
1. That the directories are a subdirectory of `$HOME`)
2. That there's only one going to be one directory in the environment variable passed in
3. That you only want one directory returned.

I think it's worth deciding if ghostty should just use the first directory in `$XDG_CONFIG_DIRS` or if the xdg directory lookups should return an array of zero or more directory paths based on what's in the environment variable provided.

mitchellh:
My expectation was we'd read all the directories. I agree that with @jcollie that the proper approach is to implement an iterator for this to avoid unnecessary allocations.

jcollie:
So to be a little more specific, we need an iterator that returns these directories in this order:
 
1. `$XDG_CONFIG_HOME/ghostty` iff `$XDG_CONFIG_HOME` is set
[...]

PR #4584, opened 2025-01-04
core: add iterator to handle XDG_*_DIRS

jcollie:
Start of implementing #4506 .

### DIFF
--- a/src/os/xdg.zig
+++ b/src/os/xdg.zig
@@ -193,3 +193,77 @@ test parseTerminalExec {
         try testing.expectEqualSlices([*:0]const u8, actual, &.{ "a", "-e", "b", "c" });
     }
 }
+
+/// https://specifications.freedesktop.org/basedir-spec/latest/
+pub const Dir = enum {
+    config,
+    data,
+
+    pub fn key(self: Dir) [:0]const u8 {
+        return switch (self) {
+            .config => "XDG_CONFIG_DIRS",
+            .data => "XDG_DATA_DIRS",
+        };
+    }
+
+    pub fn default(self: Dir) [:0]const u8 {
+        return switch (self) {
+            .config => "/etc/xdg",
+            .data => "/usr/local/share:/usr/share",
+        };
+    }
+};
+
+pub const DirIterator = struct {
+    data: []const u8,
+    iterator: std.mem.SplitIterator(u8, .scalar),
+
+    /// https://specifications.freedesktop.org/basedir-spec/latest/
+    pub fn init(key: Dir) DirIterator {
+        const data = data: {
+            if (posix.getenv(key.key())) |data| {
+                if (std.mem.trim(u8, data, &std.ascii.whitespace).len > 0) break :data data;
+            }
+
+            break :data key.default();
+        };
+
+        return .{
+            .data = data,
+            .iterator = std.mem.splitScalar(u8, data, ':'),
+        };
+    }
+
+    pub fn next(self: *DirIterator) ?[]const u8 {
+        return self.iterator.next();
+    }
+};
+
+test "xdg dirs" {
+    const c = @cImport({
+        @cInclude("stdlib.h");
+    });
+
+    const testing = std.testing;
+    {
+        _ = c.unsetenv(Dir.config.key());
+        var it = DirIterator.init(.config);
+        try testing.expectEqualStrings("/etc/xdg", it.next().?);
+        try testing.expect(it.next() == null);
+    }
+    {
+        _ = c.unsetenv(Dir.data.key());
+        var it = DirIterator.init(.data);
+        try testing.expectEqualStrings("/usr/local/share", it.next().?);
+        try testing.expectEqualStrings("/usr/share", it.next().?);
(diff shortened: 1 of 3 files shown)


src/os/xdg.zig
```diff
@@ -193,3 +193,75 @@ test parseTerminalExec {
         try testing.expectEqualSlices([*:0]const u8, actual, &.{ "a", "-e", "b", "c" });
     }
 }
+
+pub const Dirs = enum {
+    config,
+    data,
+};
```
jparise, line 200: What do you think about somehow storing the env var key (string) and default value (string) here in `Dirs`?

Then we could reference those common values from places like:

https://github.com/ghostty-org/ghostty/blob/e2f9eb6a6f4dc2108f91293938374c0ed314dcb8/src/termio/shell_integration.zig#L479

jcollie: Excellent idea!

jparise: One more opportunity: https://github.com/ghostty-org/ghostty/blob/e2f9eb6a6f4dc2108f91293938374c0ed314dcb8/src/termio/Exec.zig#L806-L813

jcollie: Fixed.

src/os/xdg.zig
```diff
@@ -193,3 +193,87 @@ test parseTerminalExec {
         try testing.expectEqualSlices([*:0]const u8, actual, &.{ "a", "-e", "b", "c" });
     }
 }
+
+/// https://specifications.freedesktop.org/basedir-spec/latest/
+pub const Dirs = enum {
```
mitchellh, line 198: Rename to `Dir`, usually singular.

src/os/xdg.zig
```diff
@@ -193,3 +193,87 @@ test parseTerminalExec {
...
+    pub fn default(self: Dirs) []const u8 {
+        return switch (self) {
+            .config => "/etc/xdg",
+            .data => "/usr/local/share:/usr/share",
+        };
+    }
+};
+
+pub const config_dirs = Dirs.config;
+pub const data_dirs = Dirs.data;
```
mitchellh, line 218: I don't think you should export these like these. Callers should fully qualify it.

src/os/xdg.zig
```diff
@@ -193,3 +193,87 @@ test parseTerminalExec {
...
+    }
+
+    pub fn next(self: *DirsIterator) ?[]const u8 {
+        return self.iterator.next();
+    }
+};
+
+/// https://specifications.freedesktop.org/basedir-spec/latest/
+pub fn dirsIterator(alloc: std.mem.Allocator, key: Dirs) std.mem.Allocator.Error!DirsIterator {
+    const env = std.process.getEnvMap(alloc);
```
mitchellh, line 247: Allocating when we don't have to on systems like posix feels bad. We shouldn't allocate on posix.

src/os/xdg.zig
```diff
@@ -193,3 +193,87 @@ test parseTerminalExec {
...
+
+pub const config_dirs = Dirs.config;
+pub const data_dirs = Dirs.data;
+
+pub const DirsIterator = struct {
+    data: []const u8,
+    iterator: std.mem.SplitIterator(u8, .scalar),
+
+    /// https://specifications.freedesktop.org/basedir-spec/latest/
+    pub fn init(env: std.process.EnvMap, key: Dirs) DirsIterator {
```
mitchellh, line 225: EnvMap as the arg here feels wrong. EnvMap requires allocation and is also mutable in general. I think we just want the key and then to use the `getenv` machinery thereafter. To test it we can `setenv` our process.

mitchellh (changes requested):
Looking at this, the API doesn't feel right. 

A major issue is posix requires allocation. Posix `getenv` doesn't require allocation so we shouldn't force allocation for this either. Since we have to call this in the config loading path too, lowering allocation as much as possible is a win.

Bigger picture, I know I said this iterator would be helpful, but I'm not fully sure. I know I like small PRs but I'm not actually sure this is the right design. I'm not sure I could know the right design without doing it myself but I'm suspicious looking at this...

src/os/xdg.zig
jcollie: Fixed.

src/os/xdg.zig
jcollie: Removed.

src/os/xdg.zig
jcollie: Now non-allocating. Zig doesn't expose `setenv` so testing resorts to bringing `setenv`/`unsetenv` through `@cImport()`.

src/os/xdg.zig
jcollie: Allocations removed.

src/os/xdg.zig
jcollie: Removed.

mitchellh:
I may come back to use this code and I'll cherry pick your commit if so, but I want to take a stab at implementing #4506 fully to see if this is required. I started to at one point and realized this iterator approach wasn't needed, but didn't finish to know for sure.

PR #4591, opened 2025-01-04
macOS: Input Improvements

### LINKED ISSUE #4522: GTK: `performable` keys that map to menu accelerators always perform their action
When a keybinding is available as a menu item, we set the keybinding as that menu items accelerator. When that is pressed, we go directly through the `performBinding` machinery. Unfortunately, since this happens before reaching the surface, we never call `keyCallback` and get all the input method inputs for this. 

We need to somehow hopefully tell GTK when an action is not performed that we want it to _fall through_ to the normal input event propagation stack.

### LINKED ISSUE #4540: macOS: super+period doesn't reach Surface keyCallback

### LINKED ISSUE #4590: macOS: Keybindings should take precedence over standard menu shortcuts
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3187

<sup>Originally posted by **cortesi** December 26, 2024</sup>
For example, bind a key like this:

```
keybind = cmd+h=goto_split:left
```

Observe that `cmd-h` still hides the window, which is the default menu shortcut, rather than taking the configured action.</div>

mitchellh:
Sorry for the vague title. This PR addresses multiple issues:

1. Fixes #4540 
2. #4522 is fixed for macOS only
3. Fixes #4590 
4. Fixes an untracked issue where `command+key` events will not send release events for Kitty keyboard protocol, something I only noticed while working on this.

There are multiple components to this PR.

## Part 1: `App/Surface.keyEventIsBinding`

This new API (also available in libghostty as `ghostty_surface_key_is_binding`) returns a boolean true if the given key event would match a binding trigger if it was the next key event sent. It does not process the binding now.

This can be used by event handlers that intercept key events to determine if it should send the event to Ghostty. This helps resolve #4590 for us but is also part of all resolved issues.

## Part 2: macOS `performKeyEquivalent` changes

macOS calls `performKeyEquivalent` for any key combination that may trigger a key equivalent. if this returns `true` then it is handled and macOS ceases processing the event. 

We were already using this to intercept things like `Ctrl+/` which triggers a context menu in macOS Sequoia. But we now expand this to intercept all events to check for bindings. This lets us fix #4590.

Additionally, it's been changed to special case `cmd+period`. I'm sure more need to be added.

## Part 3: NSEvent local listener for command keyUp events

macOS simply doesn't send `keyUp` events for key events with command pressed. The only way to work around this is to register an `NSEvent` local listener. We now do this. This fixes the untracked issue noted above.

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -88,6 +88,7 @@
 		A5CEAFDE29B8058B00646FDA /* SplitView.Divider.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CEAFDD29B8058B00646FDA /* SplitView.Divider.swift */; };
 		A5CEAFFF29C2410700646FDA /* Backport.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CEAFFE29C2410700646FDA /* Backport.swift */; };
 		A5CF66D42D289CEE00139794 /* NSEvent+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CF66D32D289CEA00139794 /* NSEvent+Extension.swift */; };
+		A5CF66D72D29DDB500139794 /* Ghostty.Event.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CF66D62D29DDB100139794 /* Ghostty.Event.swift */; };
 		A5D0AF3B2B36A1DE00D21823 /* TerminalRestorable.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5D0AF3A2B36A1DE00D21823 /* TerminalRestorable.swift */; };
 		A5D0AF3D2B37804400D21823 /* CodableBridge.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5D0AF3C2B37804400D21823 /* CodableBridge.swift */; };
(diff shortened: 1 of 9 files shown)


AlexJuca:
@mitchellh macOS prioritizes global or application-level shortcuts (e.g., ⌘C for copy, ⌘V for paste). The keyDown event is handled, but the keyUp is ignored by the O.S., so NSEvent is the only other viable solution. 

Your solution is pretty good.

mitchellh:
> Your solution is pretty good.

Awesome. Thanks!

macos/Ghostty.xcodeproj/project.pbxproj
```diff
@@ -405,7 +411,7 @@
 				A5985CE52C33060F00C57AD3 /* man */,
 				A5A1F8842A489D6800D1E8BC /* terminfo */,
 				FC5218F92D10FFC7004C93E0 /* zsh */,
-				9351BE8E2D22937F003B3499 /* nvim */,
+				9351BE8E2D22937F003B3499 /* vim */,
```
liby, line 414: Hi! I noticed that in this PR, both vim and nvim folders are renamed to "vim" in the Xcode project references. While the final build output still correctly maintains the distinction between vim and nvim folders, I'm curious about the motivation behind this change in the project structure.

I see that both folders contain identical content at the moment, but having two identically named "vim" folders in the Resources group (as shown in the Xcode project navigator) might be a bit confusing for future maintenance. Could you share the reasoning behind this unification of names in the Xcode project?

image

mitchellh: Hm. I didn’t do this on purpose at all. So this was somehow a mistake. I’ll take a look!

PR #4593, opened 2025-01-04
gtk: add `split-separator-color` config

### LINKED ISSUE #4326: Add `split-divider-color` configuration
Discussion: #3301 

Let's introduce a `split-divider-color` configuration so that the color is completely customizable. We should do this for both GTK and macOS. The default should be the automatic split divider color behavior that exists now.

I'd love to also make the logic for the automatic version better so most people do not have to touch this. I'm just accepting this feature in the short term so people have an escape hatch.

jcollie:
Fixes #4326

image

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -610,6 +610,10 @@ palette: Palette = .{},
 /// Specified as either hex (`#RRGGBB` or `RRGGBB`) or a named X11 color.
 @"unfocused-split-fill": ?Color = null,
 
+/// The color of the divider between splits. If unset the default system color
+/// will be used. GTK only.
+@"split-divider-color": ?Color = null,
+
 /// The command to run, usually a shell. If this is not an absolute path, it'll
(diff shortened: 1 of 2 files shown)


src/config/Config.zig
```diff
@@ -610,6 +610,10 @@ palette: Palette = .{},
 /// Specified as either hex (`#RRGGBB` or `RRGGBB`) or a named X11 color.
 @"unfocused-split-fill": ?Color = null,
 
+/// The color of the separator between splits. If unset the default system color
+/// will be used. GTK only.
+@"split-separator-color": ?Color = null,
```
mitchellh, line 615: GTK may use this terminology but our internals have always used "divider" terminology. So that's why I chose that.

mitchellh (changes requested):
Please rename it back to the approved name, I chose it for a reason (noted here).

This is a simple enough change we should also implement it for macOS. I can handle that later.

PR #4595, opened 2025-01-04
Implement configuration option split-divider-color for macOS

chschneider:
Addresses #4326 for macOS only, should be easy to combine PR #4593 doing the same for GTK

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.Config.swift
+++ b/macos/Sources/Ghostty/Ghostty.Config.swift
@@ -375,8 +375,6 @@ extension Ghostty {
             )
         }
 
-        // This isn't actually a configurable value currently but it could be done day.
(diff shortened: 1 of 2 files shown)


macos/Sources/Ghostty/Ghostty.Config.swift
```diff
@@ -381,7 +381,20 @@ extension Ghostty {
             let backgroundColor = OSColor(backgroundColor)
             let isLightBackground = backgroundColor.isLightColor
             let newColor = isLightBackground ? backgroundColor.darken(by: 0.08) : backgroundColor.darken(by: 0.4)
-            return Color(newColor)
+
```
jcollie, line 384: The comments at the beginning of the function should be fixed up to indicate that this is now configurable.

Issue #4554, opened 2025-01-04
Zombie process leaks


goonzoid:
When surfaces close, we sometimes fail to call `wait`/`waitpid` on the associated process, which results in it hanging around as a zombie for the lifetime of the Ghostty application. There is a related discussion which I had missed until making this issue at https://github.com/ghostty-org/ghostty/discussions/3596.

I can reproduce this reliably on MacOS. Strangely, while we also fail to call `waitpid` on the subprocess on linux, it still gets reaped. I did some pretty deep digging trying to figure out what is causing this, but came up empty handed. Happy to discuss this aspect further and share what I tried if it becomes relevant.

### Reproduction Steps

I've been able to reproduce this in both the GLFW and SwiftUI builds on MacOS.

1. Launch Ghostty
1. Create a new tab (so now you have two)
1. Close the tab using Ctrl-D at the prompt, or by running `exit` (closing the tab directly does not trigger this)
1. Take a look at the process tree. You should see a zombie process that is a direct child of the core ghostty process. On MacOS, this is the `login(1)` wrapper process. Depending on what you use to inspect the process tree, it may appear as `<defunct>`.

### What's happening?

There are some longer explanations (largely correct, but perhaps a bit confused) in the edit history of this issue. Here's the short version after more debugging:

When the surface's subprocess exits independent of Ghostty, we null out the `Command` before calling `stop` on it, which means we never call `waitpid`. Simple as that.

The fix ended up being a bit more complicated, due to a quirk in the implementation of `waitpid` we were using in the Zig standard library, and because of the mysterious unknown force that's somehow causing the subprocess to be reaped already on linux. I'll discuss the details of the fix further in [my PR](https://github.com/ghostty-org/ghostty/pull/4605) to make reviewing easier.

basile-henry:
So you don't think the process hanging around after Ctrl-D on Linux is related to this bug on MacOS?
There's more detail about the Linux version in this other discussion: https://github.com/ghostty-org/ghostty/discussions/3495 (it's a shame they don't automatically cross link like issues do)

goonzoid:
> So you don't think the process hanging around after Ctrl-D on Linux is related to this bug on MacOS? There's more detail about the Linux version in this other discussion: [#3495](https://github.com/ghostty-org/ghostty/discussions/3495) (it's a shame they don't automatically cross link like issues do)

[...]

PR #4605, opened 2025-01-05
Fix leaked zombie processes when closing surfaces

goonzoid:
Issue description can be found at #4554, which this fixes.

I'm not thrilled about this solution, but it's the pragmatic option. I reached my wits' end trying to figure out why things weren't quite the same on linux, and we'd need to revisit that if we want to solve this in a different way.

Some notes:
- we were failing to call `waitpid` on subprocesses that exit independently of ghostty (i.e. they terminate, rather than us terminating them by e.g. closing a tab)
- `std.posix.waitpid` [panics](https://github.com/ziglang/zig/blob/0.13.0/lib/std/posix.zig#L4254) if `waitpid` returned `ECHILD`, which happens if the process has already been reaped when we call `waitpid`
- on linux, something mysterious is causing the subprocesses to be reaped without us calling `waitpid`, and so adding in a call to `waitpid` caused the IO thread to panic (this is what @slovdahl was seeing below with an older version of my fix, I believe)

Details of the fix:
- to enable us to always be sure to call `waitpid`, but to avoid panicking on linux, I switched us over to using `std.c.waitpid` instead of `std.posix.waitpid`
- to make this change easier, I reviewed our use of the `Command.wait` method, and discovered we could simplify it a bit
- this also provided an opportunity to de-duplicate some of the looping logic around non-blocking calls to `waitpid` that had been introduced

The only other option (other than getting to the bottom of the linux reaping mystery, which may still be worth doing), is to introduce a comptime check around the additional call to `waitpid` so that it happens on MacOS only. Would be happy to switch to that approach if it's preferred.

I definitely would understand if you don't want to merge this as is and would prefer to really get to the bottom of what's happening on linux. I just don't have more time to dedicate to that right now. I could look more next week, but would probably benefit from a second pair of eyes to ensure I'm not missing something obvious.

### DIFF
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -306,6 +306,7 @@ fn setupFd(src: File.Handle, target: i32) !void {
 }
 
 /// Wait for the command to exit and return information about how it exited.
+/// "block" is currently not respected on Windows.
 pub fn wait(self: Command, block: bool) i32 {
     if (comptime builtin.os.tag == .windows) {
         // Block until the process exits. This returns immediately if the
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -81,34 +81,6 @@ pub const LinuxCgroup = if (builtin.os.tag == .linux) ?[]const u8 else void;
 pub const linux_cgroup_default = if (LinuxCgroup == void)
 {} else null;
 
-/// The various methods a process may exit.
-pub const Exit = if (builtin.os.tag == .windows) union(enum) {
-    Exited: u32,
-} else union(enum) {
-    /// Exited by normal exit call, value is exit status
-    Exited: u8,
-
-    /// Exited by a signal, value is the signal
-    Signal: u32,
-
-    /// Exited by a stop signal, value is signal
-    Stopped: u32,
-
-    /// Unknown exit reason, value is the status from waitpid
-    Unknown: u32,
-
-    pub fn init(status: u32) Exit {
-        return if (posix.W.IFEXITED(status))
-            Exit{ .Exited = posix.W.EXITSTATUS(status) }
-        else if (posix.W.IFSIGNALED(status))
-            Exit{ .Signal = posix.W.TERMSIG(status) }
-        else if (posix.W.IFSTOPPED(status))
-            Exit{ .Stopped = posix.W.STOPSIG(status) }
-        else
-            Exit{ .Unknown = status };
-    }
-};
-
 /// Start the subprocess. This returns immediately once the child is started.
 ///
 /// After this is successful, self.pid is available.
@@ -334,7 +306,7 @@ fn setupFd(src: File.Handle, target: i32) !void {
 }
 
 /// Wait for the command to exit and return information about how it exited.
-pub fn wait(self: Command, block: bool) !Exit {
+pub fn wait(self: Command, block: bool) i32 {
     if (comptime builtin.os.tag == .windows) {
         // Block until the process exits. This returns immediately if the
         // process already exited.
@@ -349,24 +321,14 @@ pub fn wait(self: Command, block: bool) !Exit {
             return windows.unexpectedError(windows.kernel32.GetLastError());
         }
 
-        return .{ .Exited = exit_code };
+        // If Windows exit codes can be larger than i32-MAX, this is potentially
+        // unsafe, but that seems very unlikely
+        return @intCast(exit_code);
     }
 
-    const res = if (block) posix.waitpid(self.pid.?, 0) else res: {
-        // We specify NOHANG because its not our fault if the process we launch
-        // for the tty doesn't properly waitpid its children. We don't want
-        // to hang the terminal over it.
-        // When NOHANG is specified, waitpid will return a pid of 0 if the process
-        // doesn't have a status to report. When that happens, it is as though the
-        // wait call has not been performed, so we need to keep trying until we get
-        // a non-zero pid back, otherwise we end up with zombie processes.
-        while (true) {
-            const res = posix.waitpid(self.pid.?, std.c.W.NOHANG);
-            if (res.pid != 0) break :res res;
-        }
-    };
-
-    return Exit.init(res.status);
+    var options: c_int = 0;
+    if (!block) options = std.c.W.NOHANG;
+    return std.c.waitpid(self.pid.?, null, options);
 }
 
 /// Sets command->data to data.
@@ -605,9 +567,8 @@ test "Command: pre exec" {
 
     try cmd.testingStart();
     try testing.expect(cmd.pid != null);
-    const exit = try cmd.wait(true);
-    try testing.expect(exit == .Exited);
-    try testing.expect(exit.Exited == 42);
+    const pid = cmd.wait(true);
+    try testing.expectEqual(cmd.pid, pid);
 }
 
 fn createTestStdout(dir: std.fs.Dir) !File {
@@ -641,9 +602,8 @@ test "Command: redirect stdout to file" {
 
     try cmd.testingStart();
     try testing.expect(cmd.pid != null);
-    const exit = try cmd.wait(true);
-    try testing.expect(exit == .Exited);
-    try testing.expectEqual(@as(u32, 0), @as(u32, exit.Exited));
+    const pid = cmd.wait(true);
+    try testing.expectEqual(cmd.pid, pid);
 
     // Read our stdout
     try stdout.seekTo(0);
@@ -676,9 +636,8 @@ test "Command: custom env vars" {
 
     try cmd.testingStart();
     try testing.expect(cmd.pid != null);
-    const exit = try cmd.wait(true);
-    try testing.expect(exit == .Exited);
-    try testing.expect(exit.Exited == 0);
+    const pid = cmd.wait(true);
+    try testing.expectEqual(cmd.pid, pid);
 
     // Read our stdout
     try stdout.seekTo(0);
@@ -712,9 +671,8 @@ test "Command: custom working directory" {
 
     try cmd.testingStart();
     try testing.expect(cmd.pid != null);
-    const exit = try cmd.wait(true);
-    try testing.expect(exit == .Exited);
-    try testing.expect(exit.Exited == 0);
+    const pid = cmd.wait(true);
+    try testing.expectEqual(cmd.pid, pid);
 
     // Read our stdout
     try stdout.seekTo(0);
@@ -754,9 +712,8 @@ test "Command: posix fork handles execveZ failure" {
 
     try cmd.testingStart();
     try testing.expect(cmd.pid != null);
-    const exit = try cmd.wait(true);
-    try testing.expect(exit == .Exited);
-    try testing.expect(exit.Exited == 1);
+    const pid = cmd.wait(true);
+    try testing.expectEqual(cmd.pid, pid);
 }
 
 // If cmd.start fails with error.ExecFailedInChild it's the _child_ process that is running. If it does not
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -22,7 +22,6 @@ const inputpkg = @import("../input.zig");
 const terminal = @import("../terminal/main.zig");
 const internal_os = @import("../os/main.zig");
 const cli = @import("../cli.zig");
-const Command = @import("../Command.zig");
 
 const conditional = @import("conditional.zig");
 const Conditional = conditional.Conditional;
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -1297,13 +1297,10 @@ const Subprocess = struct {
                             },
                         }
 
-                        // See Command.zig wait for why we specify WNOHANG.
-                        // The gist is that it lets us detect when children
-                        // are still alive without blocking so that we can
-                        // kill them again.
-                        const res = posix.waitpid(pid, std.c.W.NOHANG);
-                        log.debug("waitpid result={}", .{res.pid});
-                        if (res.pid != 0) break;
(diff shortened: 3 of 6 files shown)


slovdahl:
I checked out your branch on my Ubuntu 24.04 and built it. It's not working very well though. When I press `ctrl-d` in a tab, the `ghostty` window hangs and I have to press "Force Quit" in the popup that appears after a few seconds.

Log output after pressing `ctrl-d`:

```
Destroying EGL context
warning(io_exec): error getting pgid for kill
```

image

goonzoid:
Yikes! That’s no good. I think I know what the issue is though. Thanks for checking it out!

I’ll try to get a Linux VM set up soon so that I can fix it correctly for both platforms.

goonzoid:
Okay, thanks to @slovdahl's feedback and a fresh look at the code, I realized there's a much simpler fix available, which I've just pushed.

I think there's a higher than average risk of future bugs here with any slight changes in this area, so I still think it may be worth a rethink of much of subprocess management code. That said, this PR should fix at least some (possibly all) zombie leak cases without introducing any risk of us getting stuck in a loop and hanging.

slovdahl:
I'm afraid the ghostty process still hangs when pressing `ctrl-d` on my Ubuntu 24.04 at least. :thinking:

goonzoid:
Hmm, that's surprising. Is it possible that you're still building the previous version of my branch? Note that I did a force-push in order to update this PR, so you may need to do a hard reset or similar (a simple pull likely won't do it).

jorismak:
Doing a first pull of your repo/branch and build now, let's see.
But in the issue / discussion, there was talk of the same happening on MacOS. Could be a complete different issue of course, but maybe it's not a Linux specific thing...

jorismak:
```
info: ghostty version=1.0.2-HEAD+d58b4353
info: runtime=apprt.Runtime.gtk
info: font_backend=font.main.Backend.fontconfig_freetype
info: dependency harfbuzz=8.4.0
info: renderer=renderer.OpenGL
info: libxev backend=main.Backend.io_uring
```

When exiting the shell with ctrl+d or `exit` still this and then nothing:
```
info(io_exec): started subcommand path=/bin/sh pid=46244
info(io_exec): subcommand cgroup=/user.slice/user-1000.slice/user@1000.service/app.slice/app-ghostty-transient-46203.scope/surfaces/6802C70.scope
info(grid): reallocating GPU buffer old=0 new=8
warning(stream): unimplemented OSC callback: terminal.osc.Command{ .end_of_command = terminal.osc.Command.Command__struct_94772{ .exit_code = null } }
warning(stream): unimplemented OSC command: end_of_command
Using integer scale 2 for EGL window (1384 930 => 2768 1860)
Destroying EGL context
```

Tried again with gtk-single-instance forced to false, but still the same.

And with closing the window through Gnome / window manager:
```
info(io_exec): started subcommand path=/bin/sh pid=46855
info(io_exec): subcommand cgroup=-
info(grid): reallocating GPU buffer old=0 new=8
warning(stream): unimplemented OSC callback: terminal.osc.Command{ .end_of_command = terminal.osc.Command.Command__struct_94772{ .exit_code = null } }
warning(stream): unimplemented OSC command: end_of_command
Using integer scale 2 for EGL window (1384 930 => 2768 1860)
Destroying EGL context
Destroying EGL context
info(io_exec): read thread got quit signal
info(surface): surface closed addr=2ace9060
```

During the gtk2 era when I used glib, it was also tricky to get threads to nicely quit. Resulted in 'the thread must always exit itself', you can't force-abort it from the main thread, or it will stay behind.

But, as I said, that was ages ago, and I hope that modern Zig doesn't have the same issues as an ageing glib version :).

slovdahl:
> Hmm, that's surprising. Is it possible that you're still building the previous version of my branch? Note that I did a force-push in order to update this PR, so you may need to do a hard reset or similar (a simple pull likely won't do it).

Yeah, I deleted the local branch before fetching your branch again. Re-did it once more and verified I have https://github.com/ghostty-org/ghostty/pull/4605/commits/d58b43536f42be1a5e248374ebb92799e36fc429 locally as my HEAD.

image

Noticed something interesting now though. This is `ctrl-d` with more than one tab open (which still hangs):

image

_But_, if I press `ctrl-d` with only one tab open, the window is closed as expected. The process keeps running in the background though, and nothing else than `Destroying EGL context` is logged.

jorismak:
Never tested more tabs, so I don't know the 'original' 1.0.0 / 1.0.1 behaviour. But this branch indeed 'hangs' / stalls when opening multiple tabs and closing one with ctrl+d. Funnily, if I 'split right' for example, it seems to close OK with ctrl+d.

I don't know if it's related. But if I have an alacritty terminal running with `top` inside it, I see very high CPU idle with 0% wait. The moment I start a ghostty shell, the waits reach up to just over 10% (and thus my idle CPU goes down, while I don't have the feeling it's doing anything more).

If I then close the ghostty window with ctrl+d, the waits keep showing. If I type `killall ghostty` in another alacritty terminal, the waits instantly drop back to 0% again.

image

slovdahl:
> I see very high CPU idle with 0% wait. The moment I start a ghostty shell, the waits reach up to just over 10% (and thus my idle CPU goes down, while I don't have the feeling it's doing anything more).

Likely unrelated: https://github.com/ghostty-org/ghostty/discussions/3224

slovdahl:
> Never tested more tabs, so I don't know the 'original' 1.0.0 / 1.0.1 behaviour. But this branch indeed 'hangs' / stalls when opening multiple tabs and closing one with ctrl+d. Funnily, if I 'split right' for example, it seems to close OK with ctrl+d.

FWIW, windows close normally with `ctrl-d` for me on latest `main`.

jorismak:
> > Never tested more tabs, so I don't know the 'original' 1.0.0 / 1.0.1 behaviour. But this branch indeed 'hangs' / stalls when opening multiple tabs and closing one with ctrl+d. Funnily, if I 'split right' for example, it seems to close OK with ctrl+d.
> 
> FWIW, windows close normally with `ctrl-d` for me on latest `main`.

Yeah, for me as well. The problem was never with windows not closing. The problem is with the ghostty process not exiting when the shell decides to quit.

slovdahl:
Built myself a debug build. All cases with only a single tab.

On `main` (but using `epoll` instead of `io_uring`) when I press `ctrl-d` and the window is closed but the process remains:

```
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(io_thread): mailbox message=termio.message.Message{ .write_small = termio.message.MessageData(u8,38).Small{ .data = { 4, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170 }, .len = 1 } }
debug(io_exec): child process exited status=127 runtime=21997ms
debug(app): mailbox message=surface_message
debug(gtk_surface): gl surface unrealized
debug(opengl): program destroyed id=165
debug(opengl): program destroyed id=162
debug(renderer_thread): mailbox message=renderer.message.Message{ .reset_cursor_blink = void }
Destroying EGL context
debug(app): focus event focused=false
debug(io_exec): termios change mode=pty.Mode{ .canonical = true, .echo = true }
debug(renderer_thread): mailbox message=renderer.message.Message{ .focus = false }
debug(io_thread): mailbox message=termio.message.Message{ .focused = false }
debug(gtk): window destroy
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
```

Pressing the X button in the top right corner to close the window:

```
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(gtk): window close request
debug(app): focus event focused=false
debug(gtk_surface): gl surface unrealized
debug(opengl): program destroyed id=165
debug(opengl): program destroyed id=162
Destroying EGL context
Destroying EGL context
debug(renderer_thread): mailbox message=renderer.message.Message{ .focus = false }
debug(io_thread): mailbox message=termio.message.Message{ .focused = false }
debug(gtk): tab box destroy
debug(gtk_surface): gl destroy
debug(renderer_thread): starting renderer thread shutdown
debug(renderer_thread): renderer thread exited
debug(io_thread): starting IO thread shutdown
debug(io_exec): process group killed pgid=576934
debug(io_exec): waitpid result=0
debug(io_exec): process group killed pgid=576934
debug(io_exec): waitpid result=576934
info(io_exec): read thread got quit signal
debug(io_thread): IO thread exited
info(surface): surface closed addr=7e5e51a62000
debug(gtk): window destroy
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
```

This PR when pressing `ctrl-d`:

```
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(io_thread): mailbox message=termio.message.Message{ .write_small = termio.message.MessageData(u8,38).Small{ .data = { 4, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170 }, .len = 1 } }
debug(io_exec): child process exited status=0 runtime=14748ms
debug(app): mailbox message=surface_message
debug(renderer_thread): mailbox message=renderer.message.Message{ .reset_cursor_blink = void }
debug(gtk_surface): gl surface unrealized
debug(opengl): program destroyed id=165
debug(opengl): program destroyed id=162
debug(io_exec): termios change mode=pty.Mode{ .canonical = true, .echo = true }
Destroying EGL context
debug(app): focus event focused=false
debug(renderer_thread): mailbox message=renderer.message.Message{ .focus = false }
debug(io_thread): mailbox message=termio.message.Message{ .focused = false }
debug(gtk): window destroy
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
```

This PR when pressing the X button in the top right corner:

```
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
debug(gtk): window close request
debug(app): focus event focused=false
debug(gtk_surface): gl surface unrealized
debug(opengl): program destroyed id=165
debug(opengl): program destroyed id=162
Destroying EGL context
Destroying EGL context
debug(io_thread): mailbox message=termio.message.Message{ .focused = false }
debug(renderer_thread): mailbox message=renderer.message.Message{ .focus = false }
debug(gtk): tab box destroy
debug(gtk_surface): gl destroy
debug(renderer_thread): starting renderer thread shutdown
debug(renderer_thread): renderer thread exited
debug(io_thread): starting IO thread shutdown
debug(io_exec): process group killed pgid=580075
debug(io_exec): waitpid result=0
debug(io_exec): process group killed pgid=580075
debug(io_exec): waitpid result=580075
info(io_exec): read thread got quit signal
debug(io_thread): IO thread exited
info(surface): surface closed addr=73b675df0000
debug(gtk): window destroy
debug(app): mailbox message=redraw_surface
debug(app): mailbox message=redraw_surface
```

goonzoid:
Thanks for all the info folks. Clearly I need to get a Linux test setup before proceeding much further. This all works well on MacOS (at least for me). I have some theories about what might be causing the stuck process, but don't want to keep guessing when I can't test it out locally.

slovdahl:
I tried to debut the GTK side of this a little bit. First time I'm touching both Zig and GTK, so bear with me.. :grin: 

I'm not sure if it's _the_ problem, but one thing that I find a bit strange is that [`App.deleteSurface`](https://github.com/ghostty-org/ghostty/blob/f4a9b65f788ac437649fea95a905bc9e92c0e668/src/App.zig#L194) is never called when the window is closed after pressing `ctrl-d` with only a single tab open.. The only place the method is called from is `errdefer self.app.core_app.deleteSurface(self);` in the `realize` method in [`apprt/gtk/Surface.zig`](https://github.com/ghostty-org/ghostty/blob/f4a9b65f788ac437649fea95a905bc9e92c0e668/src/apprt/gtk/Surface.zig#L592). AFAICT, this means that `quit_timer` is never called.

cc'ing @jcollie as the original author of the `quit_timer` code paths.

goonzoid:
Small progress here: managed to get a linux environment stood up today, and can reproduce the behavior exactly as described above. Ran out of time to debug but should be able to get to that soon.

goonzoid:
This isn't quite ready to merge. I accidentally pushed to the PR branch rather than my backup branch. I can already see that this won't build on Windows, for example.

Will add more context on what's going on to the issue soon.

goonzoid:
I need to boot up my Windows machine to test these changes, since something is obviously amiss there. Will be able to do that tomorrow I hope.

jorismak:
Can I try on Linux already? Or do you have something like 'just wait and sit till I say it is ready' :).

goonzoid:
> Can I try on Linux already? Or do you have something like 'just wait and sit till I say it is ready' :).

No harm in trying. I don't think it fixes anything there, but I never dug deep in to the linux specific issues so it's possible that it does.

mitchellh:
Superseded by #5353. Thanks for the initial work @goonzoid ❤️

PR #4609, opened 2025-01-05
macos: detect IME input source change as part of keyDown event

### LINKED ISSUE #4539: macOS: AquaSKK Japanese input behaves poorly
### Discussed in https://github.com/ghostty-org/ghostty/discussions/2628

<sup>Originally posted by **lonsagisawa** November  8, 2024</sup>
## Description

I'm using AquaSKK for Japanese input.
AquaSKK uses unusual keybindings, like Ctrl-j to turn on hiragana input, q to toggle hiragana and katakana, l to return to latin input.

In standard macOS applications, keybinds are not transferred to app.

https://github.com/user-attachments/assets/7e530f5c-2995-4d84-acf3-61e18c53a59c

However,  in Ghostty, IME and Ghostty both receive keybinds.

https://github.com/user-attachments/assets/a53a547a-a6b3-4a25-b3cc-ec620adbf4c8

</div>

### More Details:

Specific reproduction step: https://github.com/ghostty-org/ghostty/discussions/2628#discussioncomment-11194834

mitchellh:
Fixes #4539

AquaSKK is a Japanese IME (Input Method Editor) for macOS. It uses keyboard inputs to switch between input modes. I don't know any other IMEs that do this, but it's possible that there are others. Prior to this change, the keyboard inputs to switch between input modes were being sent to the terminal, resulting in erroneous characters being written.

This change adds a check during keyDown events to see if the input source changed _during the event_. If it did, we assume an IME captured it and we don't pass the event to the terminal.

This makes AquaSKK functional in Ghostty.


mitchellh:
cc @lonsagisawa if you can verify this fix works then I'll merge. Your reproduction no longer fails on my machine.

lonsagisawa:
@mitchellh Confirmed to work with AquaSKK and macSKK. Also it won't harm basic keybindings in default IM.

PR #4612, opened 2025-01-05
Disable macOS alt-left remappings if kitty disambiguate flag is set

krobelus:
(Originally reported at https://github.com/fish-shell/fish-shell/issues/11004)

Commit 5bb2c62f (config: revert cmd/alt+left/right to legacy encoding
on macOS by default, 2024-12-24) translates keys like alt-left to alt-b
on macOS.  That behavior is part of the default Terminal.app and allows
alt-left to do something reasonable on the default macOS shell (zsh).

Other applications want to disambiguate alt-left from alt-b. Ror
example, fish maps them to different commands.

fish requests the kitty keyboard protocol,
specifically the [disambiguate flag
(0b1)](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#disambiguate-escape-codes).

I don't know if the protocol adjudicates this specific question but I
think the disambiguate flag is a good signal that helpful remappings
are not desired. Let's ignore these bindings while disambiguate is set.
In future we could try to find a better criteria.

To test, it should be enough to run (in bash or zsh)

	printf '\x1b[=1u'

and check if the keys change accordingly.

Unfortunately I can't get XCode to run on my Mac right now,
but I was able to test the change by applying the 
"Natural text editing" keybinds on Linux.

Here's me typing the affected keys into
"ghostty -e fish_key_reader -V",
before and after.
Looks good, except that fish does not support the super modifier yet.

<details>

super-left

	# decoded from: \x01
	bind ctrl-a 'do something'
	# decoded from: \e\[1\;9D
	bind left 'do something'

super-right

	# decoded from: \x05
	bind ctrl-e 'do something'
	# decoded from: \e\[1\;9C
	bind right 'do something'

super-backspace

	# decoded from: \e\x15
	bind ctrl-alt-u 'do something'
	# decoded from: \e\[127\;9u
	bind backspace 'do something'

alt-left

	# decoded from: \eb
	bind alt-b 'do something'
	# decoded from: \e\[1\;3D
	bind alt-left 'do something'

alt-right

	# decoded from: \ef
	bind alt-f 'do something'
	# decoded from: \e\[1\;3C
	bind alt-right 'do something'

</details>

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1833,6 +1833,22 @@ pub fn keyCallback(
     return .consumed;
 }
 
+pub fn isBindingMaskedByKittyDisambiguate(
+    self: *Surface,
+    action: input.Binding.Action,
+) bool {
+    switch (action) {
+        .esc_unless_kitty_disambiguate,
+        .text_unless_kitty_disambiguate => {
+            if (self.io.terminal.screen.kitty_keyboard.current().disambiguate) {
+                return true;
+            }
+        },
+        else => {},
+    }
+    return false;
+}
+
 /// Maybe handles a binding for a given event and if so returns the effect.
 /// Returns null if the event is not handled in any way and processing should
 /// continue.
@@ -1915,6 +1931,10 @@ fn maybeHandleBinding(
         .leaf => |leaf| leaf,
     };
     const action = leaf.action;
+
+    if (self.isBindingMaskedByKittyDisambiguate(action)) {
+        return null;
+    }
 
     // consumed determines if the input is consumed or if we continue
     // encoding the key (if we have a key to encode).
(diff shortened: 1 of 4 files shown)


src/input/Binding.zig
```diff
@@ -233,13 +233,14 @@ pub const Action = union(enum) {
     /// CSI header (`ESC [` or `\x1b[`).
     csi: []const u8,
 
-    /// Send an `ESC` sequence.
-    esc: []const u8,
+    /// Unless kitty disambiguate is active, send an `ESC` sequence.
+    esc_unless_kitty_disambiguate: []const u8,
```
krobelus, line 237: oh these names are also used by config -- should keep the old ones around then if we stick with this approach

src/Surface.zig
```diff
@@ -1833,6 +1833,22 @@ pub fn keyCallback(
...
 }
 
+pub fn isBindingMaskedByKittyDisambiguate(
+    self: *Surface,
+    action: input.Binding.Action,
+) bool {
+    switch (action) {
+        .esc_unless_kitty_disambiguate,
+        .text_unless_kitty_disambiguate => {
+            if (self.io.terminal.screen.kitty_keyboard.current().disambiguate) {
```
mitchellh, line 1843: For future note: this requires holding "the big lock" (expensive and slows everything down). I don't think the lock is held here and introducing it in this path kind of sucks. Its not the worst since its gated on the binding action but I still don't love it.

mitchellh (changes requested):
Thanks for taking a stab at this. Ultimately, I don't love this approach. 

It's not backed by a discussion otherwise I'd put my thoughts there. If you (or someone else) wants to continue pursuing a solution to this please open a discussion and I'm happy to go back and forth on design.

I think the stated intention here is to have a keybinding that only evaluates under certain keyboard protocol conditions in order to solve the issue that when a running application supports richer keyboard protocols, you want to allow it through.

I suppose some questions: is the Kitty disambiguate flag critically necessary to gate this? Or can we safely assume anyone using non-legacy encoding can accept this? If it's not critically necessary, I'd feel better about a sequence such as `text_legacy` that only worked in legacy modes. 

This approach in here of course _works_ but it feels a bit like duct tape rather than a more thoughtful approach to the problem. I may be wrong about that and we may want to swing back to this but I'm not willing to snap merge this PR without further thought, I'm sorry.

I appreciate the attempt though.

mitchellh:
I

PR #4645, opened 2025-01-05
vim compiler plugin for ghostty filetype

beaumccartney:
Basically integrates `ghostty +validate-config` with vim's compiler feature. This allows validating the config from vim and navigating to errors e.g. with the quickfix list.

### DIFF
--- a/src/config/vim.zig
+++ b/src/config/vim.zig
@@ -26,7 +26,7 @@ pub const ftplugin =
     \\
     \\if !exists('current_compiler')
     \\  compiler ghostty
-    \\  let b:undo_ftplugin .= "| compiler make"
+    \\  let b:undo_ftplugin .= " makeprg< errorformat<"
     \\endif
     \\
 ;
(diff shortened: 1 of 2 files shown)


mitchellh:
cc @gpanders - thoughts?

src/config/vim.zig
```diff
@@ -24,6 +24,21 @@ pub const ftplugin =
     \\
     \\let b:undo_ftplugin = 'setl cms< isk< ofu<'
     \\
+    \\if !exists('current_compiler')
+    \\  compiler ghostty
+    \\  let b:undo_ftplugin .= "| compiler make"
```
gpanders, line 29: I think we can just reset makeprg and errorformat here

```suggestion
    \\  let b:undo_ftplugin .= ' makeprg< efm<'
```

gpanders (approved):
I’m on mobile so can’t verify that the error format is correct, but as long as it is this looks like a nice addition

src/config/vim.zig
beaumccartney: done!

PR #4652, opened 2025-01-05
config: store non-reproducible diagnostics in replay steps

### LINKED ISSUE #4509: Non-existent config-file does not show an error when `theme` is set

mitchellh:
Fixes #4509

Our config has a replay system so that we can make changes and reproduce the configuration as if we were reloading all the files. This is useful because it lets us "reload" the config under various conditions (system theme change, etc.) without risking failures due to world state changing (i.e. config files change or disappear).

The replay system assumed that all diagnostics were reproducible, but this is not the case. For example, we don't reload `config-file` so we can't reproduce diagnostics that come from it.

This commit adds a new `diagnostic` replay step that can be used to store non-reproducible diagnostics and `config-file` is updated to use it.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3080,25 +3080,31 @@ pub fn loadRecursiveFiles(self: *Config, alloc_gpa: Allocator) !void {
 
         // We must only load a unique file once
         if (try loaded.fetchPut(path, {}) != null) {
-            try self._diagnostics.append(arena_alloc, .{
+            const diag: cli.Diagnostic = .{
                 .message = try std.fmt.allocPrintZ(
                     arena_alloc,
                     "config-file {s}: cycle detected",
                     .{path},
                 ),
-            });
+            };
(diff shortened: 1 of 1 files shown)


mitchellh:
cc @gpanders @dmehala

Issue #4643, opened 2025-01-05
bash: use \w instead of $PWD for the title


jparise:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4601

<div type='discussions-op-text'>

<sup>Originally posted by **piki** January  4, 2025</sup>
The `PS1=` variable in the bash integration for the title bar could use [`\w`](https://www.man7.org/linux/man-pages//man1/bash.1.html#PROMPTING) rather than `$PWD` to indicate the working directory.

The effect is to replace any leading instance of `$HOME` with a tilde, so `/Users/patrick/code` becomes `~/code`.  It also respects [`$PROMPT_DIRTRIM`](https://www.man7.org/linux/man-pages//man1/bash.1.html#PARAMETERS), if any.  It looks cleaner to me, especially if there are multiple tabs open.

Implemented in [this branch](https://github.com/ghostty-org/ghostty/compare/main...piki:ghostty:ps1-working-dir?expand=1).</div>

---

From @jparise:

This sounds like a good improvement to me, and it looks like we do something similar in the zsh shell integration: https://github.com/ghostty-org/ghostty/blob/f4a9b65f788ac437649fea95a905bc9e92c0e668/src/shell-integration/zsh/ghostty-integration#L200

The "set title" command is purely informational, so `$PROMPT_DIRTRIM`'s abbreviation behavior shouldn't cause any functional issues. (We can rely on `\e]7;kitty-shell-cwd://..` to report the full current working directory.)

I think it's possible some people might like to further customize their title, but they can do that in through their own shell configuration and disable our "title" shell integration feature.

jparise:
@piki feel free to create a pull request based on your branch. If you're short on time, I can also make this change.

piki:
> [@piki](https://github.com/piki) feel free to create a pull request based on your branch. If you're short on time, I can also make this change.

Opened as #4656

PR #4656, opened 2025-01-06
Use \w instead of $PWD for title bar

piki:
The `PS1=` variable in the bash integration for the title bar could use [`\w`](https://www.man7.org/linux/man-pages//man1/bash.1.html#PROMPTING) rather than `$PWD` to indicate the working directory.

The effect is to replace any leading instance of `$HOME` with a tilde, so `/Users/patrick/code` becomes `~/code`.  It also respects [`$PROMPT_DIRTRIM`](https://www.man7.org/linux/man-pages//man1/bash.1.html#PARAMETERS), if any.  It looks cleaner to me, especially if there are multiple tabs open.

Implements #4643 (cc discussion #4601).


jparise (approved):

mitchellh:
Thank you!

PR #4662, opened 2025-01-06
gtk: add config option to disable GTK OpenGL debug logging

jcollie:
The only change from default should be that when running a ReleaseFast binary you won't get OpenGL debugging.

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -187,8 +187,9 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
             }
         }
         try writer.writeByte(0);
-        log.warn("setting GDK_DEBUG={s}", .{fmt.getWritten()});
-        _ = internal_os.setenv("GDK_DEBUG", buf[0 .. fmt.pos - 1 :0]);
+        const value = fmt.getWritten();
+        log.warn("setting GDK_DEBUG={s}", .{value[0 .. value.len - 1]});
+        _ = internal_os.setenv("GDK_DEBUG", value[0 .. value.len - 1 :0]);
     }
 
     {
@@ -204,8 +205,9 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
             }
         }
         try writer.writeByte(0);
-        log.warn("setting GDK_DISABLE={s}", .{fmt.getWritten()});
-        _ = internal_os.setenv("GDK_DISABLE", buf[0 .. fmt.pos - 1 :0]);
+        const value = fmt.getWritten();
+        log.warn("setting GDK_DISABLE={s}", .{value[0 .. value.len - 1]});
+        _ = internal_os.setenv("GDK_DISABLE", value[0 .. value.len - 1 :0]);
     }
 
     if (version.runtimeAtLeast(4, 14, 0)) {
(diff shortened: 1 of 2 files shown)


src/config/Config.zig
```diff
@@ -1952,6 +1952,15 @@ keybind: Keybinds = .{},
 /// must always be able to move themselves into an isolated cgroup.
 @"linux-cgroup-hard-fail": bool = false,
 
+/// Enable or disable GTK's OpenGL debugging logs. The default for depends
+/// on the optimization level that Ghostty was built with:
```
tristan957, line 1956:
```suggestion
/// Enable or disable GTK's OpenGL debugging logs. The default depends on
/// the optimization level that Ghostty was built with:
```

src/apprt/gtk/App.zig
```diff
@@ -165,6 +129,93 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
...
+        try writer.writeByte(0);
+        log.warn("setting GDK_DISABLE={s}", .{fmt.getWritten()});
+        _ = internal_os.setenv("GDK_DISABLE", buf[0 .. fmt.pos - 1 :0]);
+    }
+
+    if (version.atLeast(4, 14, 0)) {
+        // We need to export GSK_RENDERER to opengl because GTK uses ngl by
+        // default after 4.14
+        _ = internal_os.setenv("GSK_RENDERER", "opengl");
+    }
```
tristan957, line 217: it would be nice to override this from the config because personally, I just want to use whatever Fedora ships by default.

jcollie: I went with `default` and `opengl`.

src/config/Config.zig
jcollie: Fixed.

tristan957:
I like the second commit. Thanks!

mitchellh (changes requested):
Couple small requests but overall excellent work

src/config/Config.zig
```diff
@@ -1954,6 +1954,23 @@ keybind: Keybinds = .{},
...
+/// - `ReleaseFast`: `false`
+@"gtk-opengl-debug": bool = build_config.slow_runtime_safety,
+
+/// After GTK 4.14.0, we need to force the GSK renderer to OpenGL as the default
+/// GSK renderer is broken on some systems. If you would like to override
+/// that bekavior, set `gtk-gsk-renderer=default` and either use your system's
+/// default GSK renderer, or set the GSK_RENDERER environment variable to your
+/// renderer of choice before launching Ghostty. This setting has no effect when
+/// using versions of GTK earlier than 4.14.0.
+@"gtk-gsk-renderer": GtkGskRenderer = .opengl,
```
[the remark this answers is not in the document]
jcollie: Hmm, that doesn't really work with our current config parser, as we have no way to set a value to a Zig `null` if the default isn't `null`. The default should be to set `GSK_RENDERER=opengl`. If it's an optional we can only do `gtk-gsk-renderer=opengl` or `gtk-gsk-renderer=`, both of which set `GSK_RENDERER=opengl`.

src/config/Config.zig
```diff
@@ -1954,6 +1954,23 @@ keybind: Keybinds = .{},
...
 @"linux-cgroup-hard-fail": bool = false,
 
+/// Enable or disable GTK's OpenGL debugging logs. The default depends on the
+/// optimization level that Ghostty was built with:
+///
+/// - `Debug`: `true`
+/// - `ReleaseSafe`: `true`
+/// - `ReleaseSmall`: `true`
+/// - `ReleaseFast`: `false`
+@"gtk-opengl-debug": bool = build_config.slow_runtime_safety,
```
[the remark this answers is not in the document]
jcollie: Done.

PR #4668, opened 2025-01-06
macOS: Add Bluetooth permission description.

evieluvsrainbows:
Adds the missing Bluetooth permission to Ghostty's application playlist manifest file, as it was missing. Tweaks have also been made to existing permission descriptions to be more readable. Should fix an abrupt crash in Fastfetch and any other apps that may request Bluetooth permissions.


evieluvsrainbows:
Do *not* merge yet. I didn't realize there were some permissions already in the manifest; I will update the PR later to only add the missing permissions. Don't know how I overlooked that there were already some in the manifest. Oops.

mitchellh:
Thanks!

PR #4676, opened 2025-01-06
Add keybind action `copy_url_to_clipboard`

### LINKED ISSUE #4633: Add keybind action `copy_url_to_clipboard`
Discussion: #4616 

We should introduce a new keybind action `copy_url_to_clipboard` that copies the URL below the mouse to the clipboard. This should work with both normal detection and OSC8. 

With the new `performable:` binding prefix, its easy to make this only have an affect while there is a link present.

liby:
## Description

This PR implements the `copy_url_to_clipboard` keybind action. This action allows users to copy URLs directly to the clipboard without needing to select them first.

### Features

- Works with both regex-matched URLs and OSC8 hyperlinks
- Copies only the URL portion, not the surrounding text
- Respects the `clipboard_trim_trailing_spaces` configuration
- Provides clear error feedback
- Follows the same patterns as other clipboard operations

https://github.com/user-attachments/assets/c9c82e3d-dfc5-4171-b367-d6799305d87f

Resolves https://github.com/ghostty-org/ghostty/issues/4633

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3936,6 +3936,33 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
             return false;
         },
 
+        .copy_url_to_clipboard => {
+            // If the mouse isn't over a link, nothing we can do.
+            if (!self.mouse.over_link) return false;
+
+            const pos = try self.rt_surface.getCursorPos();
+            if (try self.linkAtPos(pos)) |link_info| {
+                // Get the URL text from selection
+                const url_text = (self.io.terminal.screen.selectionString(self.alloc, .{
+                    .sel = link_info[1],
+                    .trim = self.config.clipboard_trim_trailing_spaces,
+                })) catch |err| {
+                    log.err("error reading url string err={}", .{err});
+                    return false;
+                };
+                defer self.alloc.free(url_text);
+
+                self.rt_surface.setClipboardString(url_text, .standard, false) catch |err| {
+                    log.err("error copying url to clipboard err={}", .{err});
+                    return true;
+                };
+
+                return true;
+            }
+
+            return false;
+        },
+
         .paste_from_clipboard => try self.startClipboardRequest(
             .standard,
             .{ .paste = {} },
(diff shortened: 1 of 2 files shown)


src/Surface.zig
```diff
@@ -4244,6 +4244,39 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
...
+                return false;
+            }
+
+            const pos = try self.rt_surface.getCursorPos();
+            const point = self.posToViewport(pos.x, pos.y);
+            const screen = &self.renderer_state.terminal.screen;
+
+            const pin = screen.pages.pin(.{ .viewport = point }) orelse {
+                return false;
+            };
```
AlexJuca, line 4259: @liby We should probably leave a log if failed to get a pin for the clicked point.

log.warn("failed to get pin for clicked point", .{});

AlexJuca:
@liby There is a bug in this that copies more than the URL. It copies the entire line of text to the clipboard.

peterjc:
See also #4678 which also implements this, but with slightly different logic

Grazfather:
While this is nice, it still requires the mouse. A way to pop up hints for urls, and then take a different action (e.g. open in browser or copy) would be nicer. Nicer yet, being able to specify different regexes would be nice, so that we can bind different select actions for e.g. git hashes vs URLs.

See the [discussion](https://discord.com/channels/1005603569187160125/1322076989384818778/1322078469823135764) here.

With this, how is it different than cmd-clicking, other than it copying? I would argue that this behaviour is stranger than the cmd-click to open, since you're using the mouse but not clicking. It would make more sense (imo) to e.g. shift-cmd-click to copy, and cmd-click to open.

mitchellh:
Indeed, this is an orthogonal issue @Grazfather and it doesn't invalidate or delay any of the other discussions you've noted.

Grazfather:
I just hope that these sort of actions are not baked in too deeply, making flexibility more difficult down the line.

For example, `self.mouse.over_link` makes a link some sort of first-class entity, while to my mind it might make sense to instead e.g. have `self.mouse.over_match`, so that different types of hover-able things on screen can be defined.

Grazfather:
Updated my link above, accidentally linked to a discord discussion instead of the github discussion

Grazfather:
That's nice to hear :) I am only going by a superficial level reading of the code.

Thanks!

PR #4678, opened 2025-01-06
feat: implement copy_url_to_clipboard keybind action functionality

### LINKED ISSUE #4633: Add keybind action `copy_url_to_clipboard`

AlexJuca:
This PR resolves #4633 

Here is a quick demo:

https://github.com/user-attachments/assets/62ae56e5-c3f1-45b6-a1df-3d1c459004ac

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3932,6 +3932,27 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
             return false;
         },
 
+        .copy_url_to_clipboard => {
+            const pos = try self.rt_surface.getCursorPos();
+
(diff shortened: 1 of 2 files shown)


AlexJuca:
@mitchellh  Any edge cases I might have missed?

peterjc:
See also #4676 which also implements this, but with slightly different logic

mitchellh:
Going with #4676. Thank you

PR #4686, opened 2025-01-06
renderer/metal: blend rendered text in linear RGB space (aka. Gamma Correction)

aljoscha:
Touches #2125

Before, we were blending rendered text onto the background in sRGB
space, which is linear in perceived color/brightness but _not_ linear in
physical terms.

Correct alpha blending requires decoding from sRGB space into linear RGB
space, performing the blending, and then encoding back to sRGB space for
presenting. See here for a very good article on the whole topic of gamma
and how it relates to sRGB and text rendering:
https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/#antialiasing

In this demonstation, I do the decoding "by hand" in the shader, and
rely on Metal do re-encode as sRGB, by setting the right pixelFormat on
the colorAttachment.

One thing to note is that correct alpha blending in linear space makes
the rendered fonts look even thinner than they already are. When Kitty
introduced alpha blending in linear space, they introduced a
configuration that allows adjusting the gamma/contrast, which will
thicken up the fonts again. Fonts being thinner when doing correct
blending is also something that the article linked above explains, and
you can confirm this behavior for youself by running through an example
alpha blending calculation by hand, on pen and paper, say.

Just letting you know, I'm not at all wedded to this PR. 🕊️  I got interested in the topic because font rendering on Ghostty looked somewhat "flimsy" to me, so I dug down into the rabbit hole of sRGB, gamma correction, etc. If you have other plans or someone is already working on this I can close the PR. _If_ you're keen to help me shepherd this in, though, I'd be very happy to respond to any comments/suggestions!

TODO/Notes before merging:

 - Add config that allows thickening fonts
 - Check performance, maybe introduce a LUT for doing the
   conversions/decode. **Which benchmarks should I be running?**
 - Follow up: I've only looked at Metal, but the Linux/OpenGL renderer
   might have the same issues. I could work on that one next.

## Screenshots

You might have to zoom in/pixel peep on these.

This shows how fonts look even thinner now, as the above article also describes:
Screenshot 2025-01-06 at 14 49 48

We can see there's no more dark color artifacts around the edged:
Screenshot 2025-01-06 at 14 49 10

And here I demonstrate how we could thicken the fonts again by adding adjustable "gain" (name totally TBD) to the rendered font's alpha:
Screenshot 2025-01-06 at 14 54 22

Here's a yet more thorough comparison. We have browsers up top (Firefox, Safari, Chrome), of which I'd say Firefox does _not_ blend in linear space but Safari and Chrome _do_. And I'd say Safari looks best to my eyes. In the bottom row we have Ghostty main and Terminal, which blend in sRGB space (without gamma correction), and then Kitty and this branch on the bottom right which _do_ blend in linear space (with gamma correction). Here it also really helps to click through to the screenshot and zoom in. But the changes are most obvious when going in there with Digital Colour Meter.
Screenshot 2025-01-07 at 13 52 19

### DIFF
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -2113,6 +2113,7 @@ pub fn setScreenSize(
                 const id_init = id_alloc.msgSend(objc.Object, objc.sel("init"), .{});
                 break :init id_init;
             };
+            // NOTE: Maybe these need to be changed as well, so custom shaders correctly decode/encode from/to sRGB?
             desc.setProperty("pixelFormat", @intFromEnum(mtl.MTLPixelFormat.bgra8unorm));
             desc.setProperty("width", @as(c_ulong, @intCast(size.screen.width)));
             desc.setProperty("height", @as(c_ulong, @intCast(size.screen.height)));
@@ -2143,6 +2144,7 @@ pub fn setScreenSize(
                 const id_init = id_alloc.msgSend(objc.Object, objc.sel("init"), .{});
                 break :init id_init;
             };
+            // NOTE: Maybe these need to be changed as well, so custom shaders correctly decode/encode from/to sRGB?
             desc.setProperty("pixelFormat", @intFromEnum(mtl.MTLPixelFormat.bgra8unorm));
             desc.setProperty("width", @as(c_ulong, @intCast(size.screen.width)));
             desc.setProperty("height", @as(c_ulong, @intCast(size.screen.height)));
@@ -2859,6 +2861,7 @@ fn addGlyph(
         .powerline => .fg_powerline,
     };
 
+    // TOOD: Could convert from sRGB color to linear here, possibly using a LUT.
     try self.cells.add(self.alloc, .text, .{
         .mode = mode,
         .grid_pos = .{ @intCast(x), @intCast(y) },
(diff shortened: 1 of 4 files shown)


qwerasd205:
There's nothing wrong with this implementation, however it isn't "correct" for macOS apps. After thorough investigation and a very helpful hint from @lunacookies, I've determined that these days what Apple seemingly does for their native apps (e.g. Terminal.app, TextEdit), is that they simply ignore gamma correction, but they avoid having particularly glaring fringing thanks to the fact that they render in the display's color space, which is typically a wide gamut color space more or less equal to Display P3. By rendering in a wide gamut color space, fully saturated sRGB colors are no longer at the very edges of the space, so blending issues are reduced quite a bit with no special extra handling required.

In experimenting with trying to exactly replicate Terminal.app's rendering (which I have successfully achieved now), I've already got a branch locally which has most of the code required for doing things that way, so I'll probably clean it up a bit and PR it some time tomorrow. I intend to have things implemented in a way that can be generalized to the OpenGL renderer and Linux, and which will allow the user to configure their desired alpha blending, whether they want "native" style rendering or proper gamma correction (or possibly a special third option I cooked up, which might give the best of both worlds).

Sorry to take this from you 😅 -- this PR is definitely on the right track if *just* doing proper gamma corrected blending was what was needed, but it's not exactly, and I already have most of the details of what *is* necessary worked out on my local branch.

aljoscha:
> Sorry to take this from you 😅 -- this PR is definitely on the right track if _just_ doing proper gamma corrected blending was what was needed, but it's not exactly, and I already have most of the details of what _is_ necessary worked out on my local branch.

No problem at all! ☺️ I'm glad you're working on it, you seem to have a good handle on things in general! Looking forward to your fix.

Closing now...

PR #4695, opened 2025-01-06
ignore ending w/o-newline delimiter @ link.regex

xerc:
@ 1.0.1 | 8696

`printf "https://google.com\nhttps://google.com"`

hightlight the second link (click is blocked anyway)
```
https://google.com
https://google.com% << the link ends with %
```

### DIFF
--- a/src/config/url.zig
+++ b/src/config/url.zig
@@ -26,7 +26,7 @@ pub const regex =
     "(?:" ++ url_schemes ++
     \\)(?:
 ++ ipv6_url_pattern ++
-    \\|[\w\-.~:/?#@!$&*+,;=%]+(?:[\(\[]\w*[\)\]])?)+(?<![,.])|(?:\.\.\/|\.\/*|\/)[\w\-.~:\/?#@!$&*+,;=%]+(?:\/[\w\-.~:\/?#@!$&*+,;=%]*)*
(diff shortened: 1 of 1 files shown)


src/config/url.zig
```diff
@@ -24,7 +24,7 @@ const oni = @import("oniguruma");
 /// handling them well requires a non-regex approach.
 pub const regex =
     "(?:" ++ url_schemes ++
-    \\)(?:[\w\-.~:/?#@!$&*+,;=%]+(?:[\(\[]\w*[\)\]])?)+(?<![,.])
+    \\)(?:[\w\-.~:/?#@!$&*+,;=%]+(?:[\(\[]\w*[\)\]])?)+(?<![,.%])
```
mitchellh, line 27: Any changes to the regular expression must be accompanied by changes to unit tests.

mitchellh (changes requested):
Thanks, see comments.

src/config/url.zig
xerc: @mitchellh is there a location where this without-newline character is defined?

mitchellh:
Sorry someone else made a URL regex change. Can you rebase your changes?

mitchellh:
> @mitchellh is there a location where this without-newline character is defined?

I'm not sure what character you're talking about.

xerc:
the `%` on the end of the line if there was ne new-line char definied in the output

mitchellh:
> the `%` on the end of the line if there was ne new-line char definied in the output

The shell does this. Not Ghostty

mitchellh:
I'm not sure this is the right approach. I think this will exclude valid URLs. Your test cases are already a bit suspicious. Your original motivation to exclude the end of line marker is also a bit suspicious because that's shell-customizable. For example, mine is not a `%` at all.

Rejected. 

Open to the idea but not happy with the approach.

Issue #4494, opened 2025-01-03
bash: seeing extra leading character '[' in command output


jparise:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3566

<div type='discussions-op-text'>

<sup>Originally posted by **jharshman** December 27, 2024</sup>
After compiling from source and running I am seeing an extra character in the output of every command. Is anyone else experiencing this? Shell is Bash and no modifications to the PS1. 

```bash
[josh@serenity ghostty]$ echo hi
[hi
[josh@serenity ghostty]$ pwd
[/home/josh/Projects/ghostty
```

```
info: ghostty version=1.0.1-dev+0000000
info: ghostty build optimize=ReleaseFast
info: runtime=apprt.Runtime.gtk
info: font_backend=font.main.Backend.fontconfig_freetype
info: dependency harfbuzz=8.4.0
info: dependency fontconfig=21402
info: renderer=renderer.OpenGL
info: libxev backend=main.Backend.io_uring
info(os): setlocale from env result=en_US.UTF-8
info(gtk): GTK version build=4.12.4 runtime=4.16.5
```</div>

**Comment by jparise** [comment]
> [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/10311?s=64&amp;v=4">@jparise](https://github.com/jparise?rgh-link-date=2025-01-03T17%3A20%3A35.000Z) Thanks for converting. I can't tell from the discussion but did you get a lead on what is causing this? More importantly: do you have an easy reproduction?

Unfortunately, no to both. I tried a bunch of things myself (details in the discussion), but I haven't been able to reproduce or identify the root cause. Multiple people have seen it in various Linux environments though.

**Comment by jharshman** [comment]
@jparise I'm happy to assist in creating an environment in which this is reproducible. Let me know if you want an extra set of hands on this.

**Comment by jparise** [comment]
> @jparise I'm happy to assist in creating an environment in which this is reproducible. Let me know if you want an extra set of hands on this.

Yes, that would be very helpful. Thanks for the offer!

**Comment by vhodges** [comment]
I have replicated the issue outside of Ghostty (eg in Gnome Terminal) on NixOS.

Setup:

```
mkdir -p temp/share/ghostty/shell-integration/bash/
cp $GHOSTTY_RESOURCES_DIR/shell-integration/bash/* temp/share/ghostty/shell-integration/bash/ #(or from src/shell-integration/bash)

[...]

PR #4710, opened 2025-01-06
termio: don't leak VTE_VERSION into child processes

jparise:
This variable is set by gnome-terminal and other VTE-based terminals. We don't want our child processes to think we're running under VTE.

See #4494

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -856,9 +856,11 @@ const Subprocess = struct {
             env.remove("GHOSTTY_MAC_APP");
         }
 
-        // Don't leak these environment variables to child processes.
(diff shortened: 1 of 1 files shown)


jcollie (approved):

src/termio/Exec.zig
```diff
@@ -857,6 +857,8 @@ const Subprocess = struct {
         }
 
         // Don't leak these environment variables to child processes.
+        env.remove("VTE_VERSION");
```
mitchellh, line 860: We should add a comment explaining why

mitchellh (changes requested):
One request. Thanks for chasing this down!

mitchellh (approved):

PR #4713, opened 2025-01-06
feat: open files without `file://` protocol

amustaque97:
**Overview**: add support for file paths starts with `../` `./` and `/`

To implement this I extended the existing regex variable in the `src/config` dir. In a few test cases, extra space is added intentionally to verify we don't include space in the URL as it looks & feels odd.

Here is the text file content used for testing
```console
➜  ~ cat test.txt
https://google.com
file:///Users/ABC/oss/ghostty/build.zig
file:///Users/ABC/oss/ghostty/debug.sh


/Users/ABC/oss/ghostty/src/../debug.sh
../../Applications/Zed.app/Contents/Info.plist
[link](/home/user/ghostty.user/example) -- non-existent
[link](/Users/ABC/oss/ghostty/src/../debug.sh)
first time ../../Applications/Zed.app/Contents/Info.plist contributor
➜  ~
```

Here is the screen recording of the changes.
[Screencast](https://www.youtube.com/watch?v=Q8qdwdBVbWk)

### DIFF
--- a/src/config/url.zig
+++ b/src/config/url.zig
@@ -24,7 +24,7 @@ const oni = @import("oniguruma");
 /// handling them well requires a non-regex approach.
 pub const regex =
     "(?:" ++ url_schemes ++
-    \\)(?:[\w\-.~:/?#@!$&*+,;=%]+(?:[\(\[]\w*[\)\]])?)+(?<![,.])
+    \\)(?:[\w\-.~:/?#@!$&*+,;=%]+(?:[\(\[]\w*[\)\]])?)+(?<![,.])|(?:\.\.\/|\.\/*|\/)[\w\-.~:\/?#@!$&*+,;=%]+(?:\/[\w\-.~:\/?#@!$&*+,;=%]*)*
 ;
 const url_schemes =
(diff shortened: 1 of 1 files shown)


mitchellh:
Well done. Thanks.

varyform:
While it's a good improvement, there are some issues:
`↳ app/views/albums/index.html.haml:1`
will be recognized as `/views/albums/index.html.haml:1`

Which is neither a correct path nor a valid file protocol URL (b/c of line number)

I think these kind of things will behave much better once `link` configuration option is available.

Temp solution for my personal needs (Rails+Zed):
https://github.com/ghostty-org/ghostty/compare/main...varyform:ghostty:extra-protocols?expand=1

```ruby
module CustomPathLogger
  private def from_rails_root(string)
    "zed://file#{string}"
  end
end

ActiveSupport.on_load(:action_view) do
  ActionView::LogSubscriber.prepend(CustomPathLogger)
end

```

mitchellh:
> While it's a good improvement, there are some issues:
↳ app/views/albums/index.html.haml:1
will be recognized as /views/albums/index.html.haml:1

Indeed, it's not a perfect solution but it's strictly an improvement. This is a well unit tested file, feel free to add additional test cases and resolve them. That'd be very helpful!

Issue #814, opened 2023-11-04
Inspector Improvements Tracking Issue


mitchellh:
This is a tracking issue for multiple improvements that can be made to the inspector. If you start working on any of these, feel free to open a new dedicated issue for the task. Otherwise, we can use this as a holding ground for future work.

**A lot of these are very contributor friendly. This is a great place to start hacking on Ghostty if you're interested.**

- [ ] macos: support opening the inspector in a pop-up window (as Linux does now)
- [ ] macos: support opening the inspector in a horizontal split
- [ ] macos: support opening in a split by extending the window instead of shrinking the terminal
- [ ] macos: scrolling with touchpad is too fast (https://github.com/ghostty-org/ghostty/issues/1375)
- [ ] linux/macos: support monitor DPI for font scaling
- [ ] add ability to change font family and size (#1097, #1095)
- [ ] list traversal (i.e. in Terminal IO tab) should support arrow keys (#1096)
- [ ] add copy to clipboard button for terminal IO (#1092)
- [ ] add semantic prompt row flags to the inspector
- [ ] add visual indicator of some sort to show what terminal an inspector is associated with (#1386)
- [ ] show termios configuration (#2587)
- [ ] show indication that echo is off for keyboard input (#2587)

bartgrantham:
minor inspector UI suggestion (macOS, maybe others): show a checkmark next to "Terminal Inspector" under the View menu if the inspector is currently open.  This provides feedback for users who open the inspector for the first time and are looking how to close it.  Alternatively or in addition, change the menu item to "Toggle Terminal Inspector".

jcollie:
@bartgrantham please don't revive such old issues (that should have been closed long ago). Please open a new discussion with your suggestion.

PR #4719, opened 2025-01-06
Inspector: support monitor DPI for font scaling

pedrocarlo:
This is my first contribution to the project. This pull request should solve the #814: linux/macos: support monitor DPI for font scaling. I am not an expert in terminals nor fonts, but after researching and reading ghostty's source code, I saw that the current font dpi was already stored in surface.font_size struct. 

After that it was just a matter of updating the imgui to add the two more rows to show the dpi. At least, this is what I understood by the request "support monitor DPI for font scaling". If I am wrong it would be great to have some more guidance, on what to do.

Thanks in advance!

### DIFF
--- a/src/inspector/Inspector.zig
+++ b/src/inspector/Inspector.zig
@@ -745,6 +745,35 @@ fn renderSizeWindow(self: *Inspector) void {
                 );
             }
         }
+
+        {
+            cimgui.c.igTableNextRow(cimgui.c.ImGuiTableRowFlags_None, 0);
+
+            {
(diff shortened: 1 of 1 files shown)


mitchellh:
Thanks for the contribution. Unfortunately this isn't what we meant by that. The issue is that all the font sizes and so on in the inspector don't take into account DPI so they appear really small (or really big) on some monitors. They're pixel sized instead of point sized. 

Thanks for giving it a shot though, what you did here looks good but not sure we need this right now. ❤️

Issue #4626, opened 2025-01-05
Background blur on Linux


pluiedev:
As previously discussed in #4361, background blur can be implemented on Linux... with *various* caveats.

Let's break this down for each desktop environment/compositor, and X11/Wayland platforms:

| Compositor | DE | X11 | Wayland |
|-|-|-|-|
| KWin | KDE, Deepin[^deepin] | ✅ `_KDE_NET_WM_BLUR_BEHIND_REGION`<br>#4723 | ✅ [`org_kde_kwin_blur_manager-v1`/`ext-blur-v1`](https://wayland.app/protocols/kde-blur)<br>#4403 |  |
| Mutter | GNOME, Cinnamon | ❌  [^blur-my-shell] | ❌ |

As you can see, this is a pretty sad state of affairs. Alas. Note that only GNOME w/ Blur My Shell supports adjusting the *intensity* of the blur - KDE and Deepin (?) both use global blur settings that can't be overridden by applications.

[^blur-my-shell]: [Blur My Shell](https://github.com/aunetx/blur-my-shell) used to allow applications to request blurring via `_MUTTER_HINTS=blur-provider=sigma`, although that is removed by v60 (see aunetx/blur-my-shell#559)

[^deepin]: Deepin <20 used to use a custom fork of Mutter known as deepin-wm which supported the custom attribute `_NET_WM_DEEPIN_BLUR_REGION_ROUNDED`. I don't think we should try to support Deepin versions that old, however...

pluiedev:
After doing some more research it seems that KWin is the only compositor remaining that allows applications to opt into window blurring. All other ones I've seen require you to set compositor/window manager-specific settings for custom window classes (e.g. GNOME with Blur my Shell, Hyprland with its builtin blur, ...), which makes sense IMO.

We've done what we can — let's put this issue to rest

ayykamp:
`ext-background-effect-v1` was recently [added to wayland staging](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/272), so hopefully more compositors will support it soon.

pluiedev:
> `ext-background-effect-v1` was recently [added to wayland staging](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/272), so hopefully more compositors will support it soon.

Looks cool, although we'll have to wait for a new wayland-protocols and KWin release at the very least. I'll add that to my TODOs

bassemkaroui:
[...]

PR #4723, opened 2025-01-06
gtk: unify Wayland and X11 platforms, implement background blur for KDE X11

pluiedev:
Part of #4626. Please let me know if the architecture looks alright — I'm fairly convinced that I'm being unorthodox here.


src/apprt/gtk/Surface.zig
```diff
@@ -807,9 +806,6 @@ pub fn getContentScale(self: *const Surface) !apprt.ContentScale {
         c.g_object_get_property(@ptrCast(@alignCast(settings)), "gtk-xft-dpi", &value);
         const gtk_xft_dpi = c.g_value_get_int(&value);
 
-        // As noted above gtk-xft-dpi is multiplied by 1024, so we divide by
-        // 1024, then divide by the default value (96) to derive a scale. Note
-        // gtk-xft-dpi can be fractional, so we use floating point math here.
         const xft_dpi: f32 = @as(f32, @floatFromInt(gtk_xft_dpi)) / 1024;
```
jcollie, line 809: This comment should remain I think.

src/apprt/gtk/key.zig
```diff
@@ -105,32 +105,25 @@ pub fn keyvalUnicodeUnshifted(
...
     event: *c.GdkEvent,
     physical_key: input.Key,
     gtk_mods: c.GdkModifierType,
-    x11_xkb: ?*x11.Xkb,
+    app_platform: platform.AppState,
 ) input.Mods {
     const device = c.gdk_event_get_device(event);
 
-    var mods = mods: {
+    var mods = switch (app_platform.inner) {
```
jcollie, line 115: Seems a little icky to be accessing `.inner` maybe move this to a function called `platform.AppState.getMods()`?

jcollie (changes requested):

src/apprt/gtk/App.zig
```diff
@@ -36,8 +36,7 @@ const c = @import("c.zig").c;
 const version = @import("version.zig");
 const inspector = @import("inspector.zig");
 const key = @import("key.zig");
-const x11 = @import("x11.zig");
-const wayland = @import("wayland.zig");
+const platform = @import("platform.zig");
```
mitchellh, line 39: I'll just put this right at the beginning since it's everywhere: I'm not sure if "platform" is the right noun for this. Platform is a catch-all of a pluggable thing (similar to widget or plugin or extension). But I feel like in this case there's probably a better noun to describe _what_ we're actually abstracting. 

For example, is it:

  * Windowing protocol
  * Display server
  * etc.

(Making those one word is harder...)

But this is pretty core to things so I'd like to make sure we get this right.

src/apprt/gtk/App.zig
```diff
@@ -364,46 +359,9 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
...
-
-        // Set up Xkb
-        break :x11_xkb try x11.Xkb.init(display);
-    };
-
-    // Initialize Wayland state
-    var wl = wayland.AppState.init(display);
-    if (wl) |*w| try w.register();
+    var app_platform: platform.AppState = undefined;
+    app_platform.init(display, &config, app_id);
```
mitchellh, line 363: These two lines are extremely suspicious. The undefined-followed-by-init pattern is usually used when we need a stable pointer. Otherwise, we just use the `init` pattern to return a value. The way this is modeled here doesn't feel right to me.

src/apprt/gtk/App.zig
```diff
@@ -364,46 +359,9 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
...
+    try app_platform.setup();
```
mitchellh, line 364: If `setup` is always called directly after `init`, then we should combine `init` and `setup` into a fail-able `init`.

src/apprt/gtk/key.zig
mitchellh: Agree we should try to contain all the switching on "platforms" in one place if possible.

src/apprt/gtk/platform.zig
```diff
@@ -0,0 +1,135 @@
...
+const build_options = @import("build_options");
+const input = @import("../../input.zig");
+const apprt = @import("../../apprt.zig");
+const Config = @import("../../config.zig").Config;
+const adwaita = @import("adwaita.zig");
+const builtin = @import("builtin");
+
+const log = std.log.scoped(.gtk_platform);
+
+pub const AppState = struct {
```
mitchellh, line 14: Bikeshed: Name this `App`

src/apprt/gtk/platform.zig
```diff
@@ -0,0 +1,135 @@
...
+    gdk_display: *c.GdkDisplay,
+    derived_config: DerivedConfig,
+
+    inner: union(enum) {
```
mitchellh, line 18: Bikeshed: name this whatever we decide on for "platform"

mitchellh (changes requested):
It's not as icky as I expected honestly. The thing that definitely cannot be right is our `init` approach (I commented in more detail). And I think we should bike shed "Platform".

src/apprt/gtk/key.zig
pluiedev: Yeah I tried to encapsulate that inside platform, but then we have to move `translateMods` into there too..

src/apprt/gtk/App.zig
pluiedev: I like "protocol". Calling it "backend" like GTK/GDK does feels *okay* though that might be confused for app runtimes

PR #4724, opened 2025-01-06
gtk(wayland): implement server-sided decorations

### LINKED ISSUE #4630: Server-side decorations
The one thing GNOME is infamously not fond of. Sadly, it's the only way we can manage to make Ghostty feel *truly* native as different desktop environments enforce different but uniform decoration styles, that preserve the overall design language of the system — disabling Adwaita is a possible alternative today but that also removes way too many quality-of-life features (tab overviews, toasts, etc.)

On Wayland, as discussed in #4361, we can use the [`xdg-decoration`](https://wayland.app/protocols/xdg-decoration-unstable-v1) protocol that everyone except GNOME implements (of course). I'm not sure what the correct X11 APIs are for this, but having it at all on Wayland would be great.

pluiedev:
Fixes #4630 fully unless there exists an X11 API I haven't found yet :p 

~~Depends on first commit of #4723, which is duplicated here for now~~

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -282,7 +282,7 @@ const c = @cImport({
 /// For example, a value of `1` increases the value by 1; it does not set it to
 /// literally 1. A value of `20%` increases the value by 20%. And so on.
 ///
-/// There is little to no validation on these values so the wrong values (i.e.
+/// There is little to no validation on these values so the wrong values (e.g.
 /// `-100%`) can cause the terminal to be unusable. Use with caution and reason.
 ///
 /// Some values are clamped to minimum or maximum values. This can make it
@@ -467,7 +467,7 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 
 /// The minimum contrast ratio between the foreground and background colors.
 /// The contrast ratio is a value between 1 and 21. A value of 1 allows for no
-/// contrast (i.e. black on black). This value is the contrast ratio as defined
+/// contrast (e.g. black on black). This value is the contrast ratio as defined
 /// by the [WCAG 2.0 specification](https://www.w3.org/TR/WCAG20/).
 ///
 /// If you want to avoid invisible text (same color as background), a value of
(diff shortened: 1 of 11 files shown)


build.zig
```diff
@@ -1484,9 +1484,13 @@ fn addDeps(
                         .target = target,
                         .optimize = optimize,
                     });
+
+                    // FIXME: replace with `zxdg_decoration_v1` once GTK merges https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/6398
+                    scanner.addCustomProtocol(plasma_wayland_protocols.path("src/protocols/server-decoration.xml"));
```
jcollie, line 1489: Is there a way to detect if the server supports `zxdg_decoration_v1` and if it doesn't, fall back to the old KDE server decoration protocol?

src/config/Config.zig
```diff
@@ -5780,6 +5780,27 @@ pub const BackgroundBlur = union(enum) {
     }
 };
 
+/// See window-decoration
+pub const WindowDecoration = enum {
+    true,
```
jcollie, line 5785: I think that it would be better to call this `client` and have `parseCLI` below translate `true` => `client`.

src/config/Config.zig
```diff
@@ -5780,6 +5780,27 @@ pub const BackgroundBlur = union(enum) {
...
+    server,
+    false,
```
jcollie, line 5787: Same as `true` above, except call it `none`. This matches the protocol more directly.

jcollie (changes requested):

build.zig
pluiedev: The thing is that that doesn't really make sense, since *GTK* only ever listens to the old one — I've tried using `zxdg_decoration_v1` but it's not very useful if the client doesn't understand it

jcollie: Ah that's annoying.

src/config/Config.zig
pluiedev: That wouldn't really make sense on macOS, though

mitchellh:
Alright I got KDE setup (optionally) in my NixOS environment, so I'm going to pull this down and hack on it soon.

mitchellh:
Made some changes:

* macOS project changes since its not a boolean anymore
* I did make the `window-decoration` value an enum of `none, client, server`. I think this does make more sense. It does allow for `false` and `true` will which makes it behave nicely with macOS. But the docs are pretty clear about all this.
* Whenever possible, always avoid comparing an enum to exactly one value. Always use a switch so we can catch future cases we need to update when changing enum values.
* Docs for `window-decoration`

PR #4733, opened 2025-01-07
Implement "Paste Selection" on macOS like Terminal.app

### LINKED ISSUE #2722: macOS: support selection pasteboard
Discussed in https://github.com/ghostty-org/ghostty/discussions/2670

> <div type='discussions-op-text'>
> 
> <sup>Originally posted by **lilyball** November 13, 2024</sup>
> One of my favorite Terminal.app features is it has a menuitem ⇧⌘V for Paste Selection. What this does is whatever the last text I selected in Terminal.app was, it pastes that. So this is kind of like `copy-on-select` behavior except it stores it in a separate pasteboard and only pasted when I use the Paste Selection action. This can be used across windows too, and it's also persisted across relaunch. I'd really love to see Ghostty pick up the same feature.
> 
> Implementation note: my initial guess here is Terminal.app uses `+[NSPasteboard pasteboardWithName:]` to create a private pasteboard for this purpose, which would also explain how it persists it (since I don't see it writing the string to preferences). I don't know what name it's using for this. If Ghostty does this as a macOS-only feature then it could perhaps do the same thing.</div>

I believe we should be able to implement this already since we support a selection pasteboard on Linux. I think this is probably almost purely a Swift thing, maybe some embedded apprt work but not much.

Westacular:
As discussed in #2670 and #2722

Fixes #2722

- This uses an NSPasteboard with the name `com.mitchellh.ghostty.selection` as a dedicated 'selection' clipboard
- Sets `supports_selection_clipboard` to true for macOS
- Sets the default `copy-on-select` config to `.true` for macOS
- Adds a "Paste Selection" menu item and default cmd+shift+v key binding for macOS (to match Terminal.app)

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -62,7 +62,7 @@ extension Ghostty {
             // uses to interface with the application runtime environment.
             var runtime_cfg = ghostty_runtime_config_s(
                 userdata: Unmanaged.passUnretained(self).toOpaque(),
-                supports_selection_clipboard: false,
+                supports_selection_clipboard: true,
                 wakeup_cb: { userdata in App.wakeup(userdata) },
                 action_cb: { app, target, action in App.action(app!, target: target, action: action) },
                 read_clipboard_cb: { userdata, loc, state in App.readClipboard(userdata, location: loc, state: state) },
@@ -320,13 +320,13 @@ extension Ghostty {
             let surfaceView = self.surfaceUserdata(from: userdata)
             guard let surface = surfaceView.surface else { return }
 
-            // We only support the standard clipboard
-            if (location != GHOSTTY_CLIPBOARD_STANDARD) {
+            // Get our pasteboard
+            guard let pasteboard = NSPasteboard.ghostty(location) else {
                 return completeClipboardRequest(surface, data: "", state: state)
             }
 
             // Get our string
-            let str = NSPasteboard.general.getOpinionatedStringContents() ?? ""
+            let str = pasteboard.getOpinionatedStringContents() ?? ""
             completeClipboardRequest(surface, data: str, state: state)
         }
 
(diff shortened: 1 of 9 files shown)


macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -364,14 +373,20 @@ extension Ghostty {
...
-            if (location != GHOSTTY_CLIPBOARD_STANDARD) { return }
+            let pasteboard: NSPasteboard
+            switch location {
+            case GHOSTTY_CLIPBOARD_STANDARD:
+                pasteboard = NSPasteboard.general
+            case GHOSTTY_CLIPBOARD_SELECTION:
+                pasteboard = selectionPasteboard
+            default:
+                return
+            }
```
jparise, line 384: I think you could assign the `pasteboard` to the result of the `switch` expression to make this a little more concise.

jparise (commented):
I'm not sure this should become the default behavior on macOS, given that it is uncommon. I don't have a strong opinion though.

That being said, it's great that this makes it available!

mitchellh (commented):
Excellent work! This all looks really good. I want to pull it down and run it and stuff but the code level review passes.

mitchellh:
Made some changes, most notably the menu item is now disabled if there are no selection paste contents.

luisdavim:
Is there a way to make copy on select go back to use the main clipboard like before this change, since i updated CMD+P and middle click result in pasting different things, I'd prefer for them to be on sync.

onnos:
I'm with @luisdavim on this one, started a [discussion](https://github.com/ghostty-org/ghostty/discussions/4898) for options.

Westacular:
@luisdavim @onnos There's already a config option for that: `copy-on-select = clipboard`

onnos:
My bad! Though I do feel the docs for `copy-on-select` are now subtly wrong.

> The value clipboard will always copy text to the selection clipboard (for supported systems) as well as the system clipboard. This is sometimes a preferred behavior on Linux.

> Middle-click paste will always use the selection clipboard on Linux and the system clipboard on macOS. Middle-click paste is always enabled even if this is false.

"Middle-click paste will always use [..] the system clipboard on macOS" is now incorrect, I believe.

2c2c:
perhaps deprecate old values when making changes like this? pretty low stakes feature I guess, but I did end up spinning my wheels for a bit wondering how I broke copying.

PR #4742, opened 2025-01-07
feat: add new noop action for key bindings (#4680)

sassanh:
Introduces a “noop” action for key bindings, allowing shortcuts to be defined without triggering any action, while still passing the key event to the pseudoterminal (pty).

- Unlike “ignore,” it does not block the key event from reaching the pty.
- Unlike “unbind,” it overrides OS-level shortcuts, effectively disabling the default OS-level action.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1968,7 +1968,10 @@ fn maybeHandleBinding(
 
     // If we have the performable flag and the action was not performed,
     // then we act as though a binding didn't exist.
-    if (leaf.flags.performable and !performed) {
+    if (leaf.flags.performable and !performed or switch (action) {
+        .noop => true,
+        else => false,
+    }) {
         // If we're in a sequence, we treat this as if we pressed a key
         // that doesn't exist in the sequence. Reset our sequence and flush
         // any queued events.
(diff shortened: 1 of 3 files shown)


src/Surface.zig
```diff
@@ -1968,7 +1968,10 @@ fn maybeHandleBinding(
 
     // If we have the performable flag and the action was not performed,
     // then we act as though a binding didn't exist.
-    if (leaf.flags.performable and !performed) {
+    if (leaf.flags.performable and !performed or switch (action) {
+        .noop => true,
+        else => false,
+    }) {
```
tristan957, line 1974:
```suggestion
    if ((leaf.flags.performable and !performed) or action == .noop) {
```

mitchellh:
Thanks for giving this a shot! Your implementation is pretty good. But, as noted in #4680, I realized this is not necessary because `noop` is equivalent to `unconsumed:trigger=ignore`. Example:

```
keybind = unconsumed:cmd+h=ignore
```

Will encode to the pty. This is obscure enough that we should probably document it. I think it's also worth considering if we want some special syntax to basically map to this. For example: `keybind = cmd+h=noop` is just syntactic sugar that generates the previous one or something.

In any case, thanks for giving this a shot. 😄

PR #4745, opened 2025-01-07
apprt/gtk: fix window colors when window-theme=ghostty

tristan957:
Before:
image

After:
image


jcollie:
hmm i'm mixed about this one. how does it look on the config errors, paste confirmation, and inspector windows?

tristan957:
I'm not sure. Are those even AdwWindows/dialogs?

PR #4783, opened 2025-01-07
apprt/gtk: Move most version checks to runtime

sin-ack:
### apprt/gtk: Add version.runtimeAtLeast

This will be used for version checks that are independent of the version
of GTK we built against.

### apprt/gtk: Move most version checks to runtime

Unless we are guarding against symbols added in new versions we now
check against the runtime version of GTK to handle them even when we
didn't build against that version.

### DIFF
--- a/src/apprt/gtk/version.zig
+++ b/src/apprt/gtk/version.zig
@@ -7,6 +7,11 @@ const c = @import("c.zig").c;
 /// in the headers. If it is run in a runtime context, it will
 /// check the actual version of the library we are linked against.
 ///
+/// This function should be used in cases where the version check
+/// would affect code generation, such as using symbols that are
+/// only available beyond a certain version. For checks which only
+/// depend on GTK's runtime behavior, use `runtimeAtLeast`.
+///
 /// This is inlined so that the comptime checks will disable the
 /// runtime checks if the comptime checks fail.
 pub inline fn atLeast(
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/version.zig
```diff
@@ -26,6 +26,16 @@ pub inline fn atLeast(
     // If we're in comptime then we can't check the runtime version.
     if (@inComptime()) return true;
 
+    return runtimeAtLeast(major, minor, micro);
+}
+
+/// Verifies that the GTK version at runtime is at least the given
+/// version.
```
mitchellh, line 33: I think documentation here and in `atLeast` could be better to clarify when either should be used. I'm happy to cover that myself but for example clarifying that:

* `atLeast` should be used when there is a symbol being linked
* `runtimeAtLeast` should be used when there is a _behavior_ being checked.

sin-ack: Added clarification on when each function should be used. Please let me know if they are inappropriate.

jcollie:
Coincidentally, looks like someone ran into this exact scenario. [Here's](https://discord.com/channels/1005603569187160125/1325693636834230393) the link to discord. They were running 4.16 but had compiled with 4.14.

src/apprt/gtk/version.zig
```diff
@@ -7,6 +7,11 @@ const c = @import("c.zig").c;
...
 /// check the actual version of the library we are linked against.
 ///
+/// This function should be used in cases where the version check
+/// would affect code generation, such as using symbols that are
+/// only available beyond a certain version. For checks which only
+/// depend on GTK's runtime behavior, use `runtimeAtLeast`.
+///
 /// This is inlined so that the comptime checks will disable the
 /// runtime checks if the comptime checks fail.
 pub inline fn atLeast(
```
jcollie, line 17: Can we rename this to `buildAtLeast` or `compiledAtLeast`? Some thing a little more descriptive in the function itself.

mitchellh: No because it checks both. I think a general `atLeast` is fine here.

sin-ack: Sorry, was off to bed. My original intention was to use `buildAtLeast` and `runtimeAtLeast`. Feel free to fix it in a followup.

PR #4785, opened 2025-01-07
bash: drop automatic shell integration with --posix

jparise:
'--posix' starts bash in POSIX mode (like /bin/sh). This is rarely used for interactive shells, and removing automatic shell integration support for this option allows us to simply/remove some exceptional code paths.

Users are still able to manually source the shell integration script.

Also fix an issue where we would still inject GHOSTTY_BASH_RCFILE if we aborted the automatic shell integration path _after_ seeing an --rcfile or --init-file argument.


mitchellh:
Looks good to me. Assuming it works 😄 Thanks

PR #4789, opened 2025-01-07
fix: ensure terminal windows don't part from parent window when toggling visibility

### LINKED ISSUE #4329: macOS: Tabs are split into windows when using `toggle_visibility` action
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4240

<sup>Originally posted by **writes-of-spring** December 31, 2024</sup>
With 1.0.1 when running the toggle visibility command a window with tabs are made into multiple windows.

The toggle_visibility command is no longer configuration documentation in case it's no longer a featured option.

</div>

AlexJuca:
As of version 1.0.1 (macOS build) when running the toggle visibility action a window with tabs is made into multiple windows. 

This PR ensures terminal tabs are reconstructed and correctly placed into its parent window.

# Demo

https://github.com/user-attachments/assets/44f14bca-15a1-4717-ba0a-44f0767feec3

FYI: I will create another PR to ensure the right tab is focused after the main window is restored.

Fixes #4329

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -72,6 +72,7 @@
 		A5A6F72A2CC41B8900B232A5 /* Xcode.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5A6F7292CC41B8700B232A5 /* Xcode.swift */; };
 		A5B30539299BEAAB0047F10C /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = A5B30538299BEAAB0047F10C /* Assets.xcassets */; };
 		A5CA378C2D2A4DEB00931030 /* KeyboardLayout.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CA378B2D2A4DE800931030 /* KeyboardLayout.swift */; };
+		A5CA378E2D31D6C300931030 /* Weak.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CA378D2D31D6C100931030 /* Weak.swift */; };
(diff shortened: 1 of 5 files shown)


mitchellh (commented):
I haven't tested this but the source looks a bit suspicious to me. Does this force all windows back into the same main window? Or does it reconstruct multiple tabbed windows properly? If its the former then we need to go back to the drawing board on this one...

AlexJuca:
@mitchellh You're right. Let me take another jab at this one.

mitchellh:
Thanks. Closing now

AlexJuca:
> Thanks. Closing now

@mitchellh Actually, you do not need to close this one as I intend to use the same PR for the new approach. Is that fine?

mitchellh:
I'd prefer you open a new PR when ready otherwise it's unclear what is sitting and what is WIP. I can perhaps convert it to a draft though let me see.

mitchellh:
I don't think I can convert it, but perhaps you can convert it to a draft? Link should be under "Reviewers"

AlexJuca:
@mitchellh A side effect of .orderOut() approach is that child windows are removed from their NSTabWindowGroup when hidden, and calling .addTabbedWindow() when unhiding does not successfully re-associate them with their parent group. This appears to be a limitation of how NSTabWindowGroup manages windows in macOS.

I've implemented a new approach, which simplifies the process of hiding and showing windows and avoids the use of the .orderOut() function. This approach ensures windows remain part of the macOS window hierarchy.

The current solution uses alpha value manipulation (alphaValue = 0 to hide and alphaValue = 1 to show) for all windows. 
This ensures Windows remain part of their original NSTabGroup: The alpha value approach keeps the windows logically attached to their parent group, avoiding group fragmentation.

	
Here is a quick demo:

https://github.com/user-attachments/assets/b62a6c5b-3a77-4c65-adfb-dc0ca213b023

Let me know what you think of this approach.

mitchellh:
Thanks! Your work worked but then I realized that `toggle_visibility` has a bunch of weird behaviors that didn't match iTerm2. I reworked that now and fixed those issues. See my commit. Thanks for your help!

PR #4797, opened 2025-01-08
build.zig Improvements

mitchellh:
This is a major refactor of `build.zig` aimed at improving maintainability while also fixing some bugs.

The major idea behind the refactor is to split the `build.zig` file up into distinct `src/build/*.zig` files. By doing so, we can improve readability of the primary `build.zig` while also enabling better reuse of steps. Our `build.zig` is now less than 150 lines of code (of course, it calls into a lot more lines but they're neatly organized now).

Ignore the commit messages I'm going to squash this whole thing.

Improvements:

  * `build.zig` is less than 150 lines of readable code.
  * Help strings and unicode table generators are only run once when multiple artifacts are built since the results are the same regardless of target.
  * Metal lib is only built once per architecture (rather than once per artifact)
  * Resources (shell integration, terminfo, etc.) and docs are only built/installed for artifacts that need them
  * Wayland protocols no longer required for source build

Breaking changes:

  * Removed broken wasm build (@gabydd will re-add)
  * Removed conformance files, shell scripts are better and we don't run these anymore
  * Removed macOS app bundle creation, we don't use this anymore since we use Xcode

## Extra: Some History

Our `build.zig` hasn't been significantly refactored since the project started, when Zig was _version 0.10_. Since then, the build system has changed significantly. We've only ever duct taped the `build.zig` as we needed to support new Zig versions, new features, etc. It was a mess.

The major improvement is adapting the entire Ghostty `build.zig` to the Step and LazyPath changes introduced way back in Zig 0.12. This lets us better take advantage of parallelism and the dependency graph so that steps are only executed as they're needed.

As such, you can see in the build.zig that we initialize a lot of things, but unless a final target (i.e. install, run) references those steps, _they'll never be executed_. This lets us clean up a lot.


src/build/GhosttyBench.zig
```diff
@@ -0,0 +1,70 @@
+//! GhosttyWebdata generates all the Ghostty website data that is
```
anund, line 1: Comment no longer matches the file.

src/build/GhosttyResources.zig
```diff
@@ -0,0 +1,257 @@
...
+        _ = wf.add("ftplugin/ghostty.vim", config_vim.ftplugin);
+        _ = wf.add("compiler/ghostty.vim", config_vim.compiler);
+
+        const install_step = b.addInstallDirectory(.{
+            .source_dir = wf.getDirectory(),
+            .install_dir = .prefix,
+            .install_subdir = "share/nvim/site",
+        });
+        try steps.append(&install_step.step);
+    }
```
tristan957, line 167: Seems like we code do some code-reuse here if you wanted. Might also be useful in their own files. Would it be useful to add a build option for vim/neovim support files (probably also bat), which defaulted to true?

src/build/GhosttyResources.zig
```diff
@@ -0,0 +1,257 @@
...
+        try steps.append(&b.addInstallFile(
+            b.path("dist/linux/app.desktop"),
+            "share/applications/com.mitchellh.ghostty.desktop",
+        ).step);
+
+        // Right click menu action for Plasma desktop
+        try steps.append(&b.addInstallFile(
+            b.path("dist/linux/ghostty_dolphin.desktop"),
+            "share/kio/servicemenus/com.mitchellh.ghostty.desktop",
+        ).step);
```
tristan957, line 201: I wouldn't mind personally if this moved behind a build option as a GNOME user. Something like `-Dkde-dolphin-action=true` or something. Again, maybe it makes sense to default to true. Happy to turn it off for my own builds.

src/build/GhosttyResources.zig
mitchellh: We could but I think now that its in a separate file it's pretty clean to have them split out like this (wet, not dry) because it makes it very clear what each thing is doing. It's all simple code. 

We could now very easily add build flags to control what is included but unless someone actually has a use case for that I don't think we have to yet.

src/build/GhosttyResources.zig
mitchellh: Yeah I believe all this stuff should be defaulted to true, because when Ghostty is packaged, it could be used by any desktop environment downstream really.

tristan957:
Anyway, looks good. I mainly made the vim/neovim comment because it wasn't immediately obvious based on code structure where they were located now.

mitchellh:
Thanks! Yeah, I think there are still further improvements. The goal was to give us a foundation we could iterate on those improvements more cleanly, since everything that touched `build.zig` before was kind of a nightmare and full of ick. 

There's still ick that's now in separate files but it should be easier to review cleanup now.

Going to merge this and let CI chew through some releases.

PR #4804, opened 2025-01-08
Reduce ghost emoji flash in title bar

### LINKED ISSUE #4799: Delay "👻" in title bar to prevent FOUC
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4798

<sup>Originally posted by **jasongitmail** January  7, 2025</sup>
Tiny suggestion... remove the default ghost emoji 👻 in the title bar of new windows please.

When a new window is opened, a flash of white which is this emoji appears for 1/2 sec, and just acts like a FOUC (flash of unstyled content) especially on dark themes. It'd feel more refined without it.

This suggestion is not solvable by [hiding the title bar](https://github.com/ghostty-org/ghostty/discussions/3343) because I want the title to contain the directory path.</div>

### Proposal

When a new window is created, use a blank title and delay the ghost emoji in the titlebar for something like 500ms, and only set it if another title hasn't been set. That should give the shell enough time to startup and set a title if there is one.

liby:
Fixes #4799

This PR attempts to reduce the flash caused by the ghost emoji in the title bar when opening new windows.

## Changes:

- Initialize `SurfaceView.title` with empty string instead of ghost emoji

- Simplify title computation logic in `TerminalView`

- Adding a 500ms fallback timer for "👻"

	- Canceling timer if title is set

## Current Status:

While these changes reduce the initial ghost emoji flash, there's still a brief moment where a folder emoji appears alone in the title bar when opening a new window. This suggests there might be a race condition or timing issue with how the title is being set and updated.

https://github.com/user-attachments/assets/3688c9f3-1727-4379-b04d-0bd6ac105728

Would appreciate feedback on the remaining flash issue and suggestions for further improvements.

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalView.swift
+++ b/macos/Sources/Features/Terminal/TerminalView.swift
@@ -10,7 +10,7 @@ protocol TerminalViewDelegate: AnyObject {
 
     /// The title of the terminal should change.
     func titleDidChange(to: String)
-    
+
     /// The URL of the pwd should change.
     func pwdDidChange(to: URL?)
 
@@ -56,15 +56,10 @@ struct TerminalView<ViewModel: TerminalViewModel>: View {
 
     // The title for our window
     private var title: String {
-        var title = "👻"
-
-        if let surfaceTitle = surfaceTitle {
-            if (surfaceTitle.count > 0) {
-                title = surfaceTitle
-            }
+        if let surfaceTitle = surfaceTitle, !surfaceTitle.isEmpty {
+            return surfaceTitle
         }
-
-        return title
+        return "👻"
     }
 
     // The pwd of the focused surface as a URL
@@ -72,7 +67,7 @@ struct TerminalView<ViewModel: TerminalViewModel>: View {
         guard let surfacePwd, surfacePwd != "" else { return nil }
         return URL(fileURLWithPath: surfacePwd)
     }
-    
+
     var body: some View {
         switch ghostty.readiness {
         case .loading:
--- a/macos/Sources/Features/Terminal/TerminalView.swift
+++ b/macos/Sources/Features/Terminal/TerminalView.swift
@@ -56,7 +56,7 @@ struct TerminalView<ViewModel: TerminalViewModel>: View {
 
     // The title for our window
     private var title: String {
-        if let surfaceTitle = surfaceTitle, !surfaceTitle.isEmpty {
+        if let surfaceTitle, !surfaceTitle.isEmpty {
             return surfaceTitle
         }
         return "👻"
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -12,7 +12,7 @@ extension Ghostty {
         // The current title of the surface as defined by the pty. This can be
         // changed with escape codes. This is public because the callbacks go
         // to the app level and it is set from there.
-        @Published private(set) var title: String = "👻"
+        @Published private(set) var title: String = ""
 
         // The current pwd of the surface as defined by the pty. This can be
         // changed with escape codes.
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -12,7 +12,14 @@ extension Ghostty {
         // The current title of the surface as defined by the pty. This can be
         // changed with escape codes. This is public because the callbacks go
         // to the app level and it is set from there.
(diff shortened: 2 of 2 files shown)


macos/Sources/Features/Terminal/TerminalView.swift
```diff
@@ -56,23 +56,18 @@ struct TerminalView<ViewModel: TerminalViewModel>: View {
 
     // The title for our window
     private var title: String {
-        var title = "👻"
-
-        if let surfaceTitle = surfaceTitle {
-            if (surfaceTitle.count > 0) {
-                title = surfaceTitle
-            }
+        if let surfaceTitle = surfaceTitle, !surfaceTitle.isEmpty {
```
jparise, line 59: Note that you can shorten the first `let` assignment here:

```suggestion
        if let surfaceTitle, !surfaceTitle.isEmpty {
```

jparise (approved):
Code looks good to me! I'm not sure if we should also implement the delay that was proposed in #4799:

> When a new window is created, use a blank title and delay the ghost emoji in the titlebar for something like 500ms, and only set it if another title hasn't been set. That should give the shell enough time to startup and set a title if there is one.

... but this change looks like a step in the right direction improvement-wise.

liby:
> Code looks good to me! I'm not sure if we should also implement the delay that was proposed in #4799:
> 
> > When a new window is created, use a blank title and delay the ghost emoji in the titlebar for something like 500ms, and only set it if another title hasn't been set. That should give the shell enough time to startup and set a title if there is one.
> 
> ... but this change looks like a step in the right direction improvement-wise.


Thank you for the review! 

I noticed that the flickering issue is more complex than initially thought. The current flickering isn't just about the ghost emoji - it's also related to how the title (folder emoji + path) gets updated.

Even with the 500ms delay for the ghost emoji, I still see a brief flash where the folder emoji appears before the path. This suggests that the issue might be more related to the title update mechanism itself rather than just the ghost emoji timing.

I believe this current implementation is a good first step as it addresses the original issue about the ghost emoji's white flash. For the broader title update flickering, we might want to:

1. First validate if this improvement helps with the original issue
2. Create a separate issue to track and investigate the title update mechanism if needed

What do you think?

jparise:
> I believe this current implementation is a good first step as it addresses the original issue about the ghost emoji's white flash. For the broader title update flickering, we might want to:
> 
>     1. First validate if this improvement helps with the original issue
>     2. Create a separate issue to track and investigate the title update mechanism if needed
> 
> What do you think?

I think that sounds right. Maybe do (1) as part of this PR's scope before merging it?

liby:
> > I believe this current implementation is a good first step as it addresses the original issue about the ghost emoji's white flash. For the broader title update flickering, we might want to:
> 
> > 
> 
> >     1. First validate if this improvement helps with the original issue
> 
> >     2. Create a separate issue to track and investigate the title update mechanism if needed
> 
> > 
> 
> > What do you think?
> 
> 
> 
> I think that sounds right. Maybe do (1) as part of this PR's scope before merging it?

Yes, but if there are better suggestions or directions, I don’t mind improving this PR further.

mitchellh:
The direction looks good. I think we should continue improving this PR though. For one, I don't like that if nothing ever sets a title then the window is just blank. To be fair, MOST shells (esp with our shell integration) should set the title, but if you run a command or something it will not. That's why I suggested the timer to fallback to a default eventually.

liby:
> but if you run a command or something it will not.

Interesting case, I hadn't considered a similar scenario before.

I've made a [commit](https://github.com/ghostty-org/ghostty/pull/4804/commits/5bfb3925baf895f78c5b3233dbe9251bd1839700) to improve this scenario by:
 - Adding a 500ms fallback timer for "👻"
 - Canceling timer if title is set

 While I haven't been able to observe this edge case in my testing environment, the improvement should provide a more robust solution for such scenarios.

AlexJuca (approved):
@liby This looks good. Do we have an updated video to visualize how things appear after adding the fallback timer?

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -139,6 +149,13 @@ extension Ghostty {
             // can do SOMETHING.
             super.init(frame: NSMakeRect(0, 0, 800, 600))
 
+            // Set a timer to show the ghost emoji after 500ms if no title is set
+            titleFallbackTimer = Timer.scheduledTimer(withTimeInterval: 0.5, repeats: false) { [weak self] _ in
+                if let self = self, self.title.isEmpty {
```
mitchellh, line 154: `if let self` is shorthand for this btw.

PR #4816, opened 2025-01-08
linux: add "Open in Ghostty" shortcut for nautilus

GabMus:


GabMus:
what about making a separate package for it in its own repo? it's perhaps a bit silly considering it's just a simple script but it would circumvent the issue of having gpl code in this repo. alternatively I can ask permission to relicense to the original author

mitchellh:
> alternatively I can ask permission to relicense to the original author

How about we try this first and see what happens. Thanks!

dist/linux/ghostty_nautilus.py
```diff
@@ -0,0 +1,97 @@
...
+        if isdir('/run/systemd/system/'):
+            self._systemd = Gio.DBusProxy.new_sync(session,
+                    Gio.DBusProxyFlags.NONE,
+                    None,
+                    "org.freedesktop.systemd1",
+                    "/org/freedesktop/systemd1",
+                    "org.freedesktop.systemd1.Manager", None)
+
+    def _open_terminal(self, path):
+        cmd = ['ghostty', f'--working-directory={path}', '--gtk-single-instance=false']
```
jcollie, line 40: Probably should set a unique `--class` and `--x11-instance-name` so that DBus doesn't get confused with multiple instances of Ghostty running.

swsnr:
Hello 👋 

I'm that original author, dropping by to point out that I can't legally permit you to use my code under any license other than the current one, because the underlying Nautilus plugin library it uses is distributed under GPL 2 or later.

You can't legally distribute Nautilus plugins under any license other than GPL 2 or later, not in Python at least.

mitchellh:
Alright thanks. It looks like we'll have to license this file separately. We already do that in other places so that isn't too bad. I'll review this soon!

Bananas-Are-Yellow:
The 1.1.0 release notes list "Open in Ghostty" as a new feature, but I'm on 1.1.2 and I can't see any such command on the Nautilus context menu. Am I missing something? Do I need to enable this?

Pop!_OS 22.04 LTS

2zqa:
@Bananas-Are-Yellow It appears `nautilus-python` package is required.

Bananas-Are-Yellow:
> @Bananas-Are-Yellow It appears `nautilus-python` package is required.

Looks like I'm out of luck then. I'm using Pop!_OS 22.04 LTS.

```
$ sudo apt install nautilus-python
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package nautilus-python
```
In any case, according to [this](https://gitlab.gnome.org/GNOME/nautilus-python), I need Nautilus ≥ 43.beta, but I'm at a lower version.
```
$ nautilus --version
GNOME nautilus 42.6
```

kerbrek:
> @Bananas-Are-Yellow It appears `nautilus-python` package is required.

In Ubuntu/Debian (and its derivatives) it is [python3-nautilus](https://packages.ubuntu.com/noble/python3-nautilus)

bittner:
Is this already being taken care of by the NixOS configuration? ([nix/package.nix](https://github.com/ghostty-org/ghostty/blob/main/nix/package.nix), [nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/gh/ghostty/package.nix))

I have a hard time spotting what needs to be installed as dependencies, let alone configured to make this work in Nautilus. :worried:

rhodes-b:
> Is this already being taken care of by the NixOS configuration? ([nix/package.nix](https://github.com/ghostty-org/ghostty/blob/main/nix/package.nix), [nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/gh/ghostty/package.nix))
> 
> I have a hard time spotting what needs to be installed as dependencies, let alone configured to make this work in Nautilus. :worried: 

Please dont comment on a 9 month old pr open a issue and triage discussion post.

Im not sure what its called but you need to install the Nautilus python plugin package
https://github.com/ghostty-org/ghostty/discussions/7344#discussion-8315413
the name seems to be listed here

it was also answered right above your post and it is easy to look at the package inputs which do not contain that package so no its not handled automatically by the flake or in nixpkgs
https://search.nixos.org/packages?channel=25.05&show=nautilus-python&query=nautilus-python

PR #4818, opened 2025-01-08
chore: replace adwaita-1 to libadwaita-1 for better pkg-config handling

lungsin:
The reasoning for this PR is discussed at https://github.com/ghostty-org/ghostty/discussions/3667

But in short, `pkg-config` queries `libadwaita-1` successfully, but not for `adwaita-1`, hence renaming it would result in better integration with pkg-config.

### DIFF
--- a/src/build/SharedDeps.zig
+++ b/src/build/SharedDeps.zig
@@ -431,7 +431,7 @@ pub fn add(
 
             .gtk => {
                 step.linkSystemLibrary2("gtk4", dynamic_link_opts);
-                if (self.config.adwaita) step.linkSystemLibrary2("adwaita-1", dynamic_link_opts);
+                if (self.config.adwaita) step.linkSystemLibrary2("libadwaita-1", dynamic_link_opts);
                 if (self.config.x11) step.linkSystemLibrary2("X11", dynamic_link_opts);
 
                 if (self.config.wayland) {


lungsin:
Hmm interesting, there seems to be something weird going on with adwaita package lol. According to @00-kat and @jcollie, pkg-config seems to recognize `libadwaita-1` instead of `adwaita-1` in both Arch and Nixos (In this discussion [here](https://github.com/ghostty-org/ghostty/discussions/3667#discussioncomment-11693923) and [here](https://github.com/ghostty-org/ghostty/discussions/3667#discussioncomment-11693985)), i.e. `pkg-config --cflags --libs adwaita-1` doesn't work. Not sure how to investigate this issue though.

Should I close this PR and continue the discussion somewhere else?

jcollie:
NixOS should definitely allow libadwaita-1:

```
$ ls -l (nix eval -f '<nixpkgs>' --raw libadwaita.dev)/lib/pkgconfig
.r--r--r-- root root 433 B 55 years ago  libadwaita-1.pc
```

I'll pull the PR down in a bit and give it a try.

jcollie:
Pulled down the PR and it built just fine both using `zig build` and `nix build .#default`. This will definitely need to be rebased though with the recent `build.zig` changes.

mitchellh:
Oh you're right actually sorry, I must've typod something (and copy/pasted since it didnt' work on either). Rip.

PR #4840, opened 2025-01-08
Implement a size-limit function for GTK

### LINKED ISSUE #4836: There's no "size_limit" implementation for GTK
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4790

<sup>Originally posted by **gtjoseph** January  7, 2025</sup>
Without a "size_limit" implementation for GTK, if gtk-adwaita is enabled (the default), you can't size a window below about 482x322 pixels because it's left to gtk to decide what the minimum is.  That's an issue. 

If gtk-adwaita is disabled, you can actually size the window down to 0 lines and if gtk-titlebar is also disabled, you can also size it down to 0 columns.  I'm not sure if this is an issue or not :)

Either way, I'l be submitting a PR to implement size_limit by calling gtk_widget_set_size_request().  There's still some minimum that adwaita imposes (about 357x248) but at least you can get reasonably close to the published minimums.

</div>

gtjoseph:
A "size-limit" function has been implemented for GTK which calls
gtk_widget_set_size_request() to set the minimum widget/window size. Without
this function, it's left to GTK to set the minimum size which is usually
a lot larger than the documented 10x4 cell minimum size.  This doesn't fix the
issue completely as GTK retains the final say in how small a window can be
but it gets closer.

Resolves: #4836

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -493,6 +493,7 @@ pub fn performAction(
         .pwd => try self.setPwd(target, value),
         .present_terminal => self.presentTerminal(target),
         .initial_size => try self.setInitialSize(target, value),
+        .size_limit => try self.setSizeLimit(target, value),
         .mouse_visibility => self.setMouseVisibility(target, value),
         .mouse_shape => try self.setMouseShape(target, value),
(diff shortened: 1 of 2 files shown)


mitchellh:
Looks perfect to me. Thank you!

PR #4845, opened 2025-01-09
Config: apply xdg system config files

### LINKED ISSUE #4506: Load configuration from XDG_CONFIG_DIRS

mjrochford:
fixes: #4506 

Applies the config files inside the XDG_CONFIG_DIRS with the first file taking precedence.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2930,33 +2930,21 @@ fn writeConfigTemplate(path: []const u8) !void {
 /// On macOS, `$HOME/Library/Application Support/$CFBundleIdentifier/config`
 /// is also loaded.
 pub fn loadDefaultFiles(self: *Config, alloc: Allocator) !void {
-    const config_subdir = "ghostty";
+    const config_dir = "ghostty";
     const config_file = "config";
-    // Load XDG system config first
-    var it = internal_os.xdg.DirIterator.init(.config);
-    // We must reverse the order of the iterator because the first xdg config
-    // dir must takes importance per the spec https://specifications.freedesktop.org/basedir-spec/latest/#variables
-    var xdg_config_dirs = std.ArrayList([]const u8).init(alloc);
-    defer xdg_config_dirs.deinit();
-    while (it.next()) |d| {
-        try xdg_config_dirs.insert(0, d);
+    const config_subdir = try std.fs.path.join(alloc, &[_][]const u8{
+        config_dir,
+        config_file,
+    });
+    defer alloc.free(config_subdir);
+    var it = internal_os.xdg.Dir.config.iter(alloc, .{ .subdir = config_subdir });
+    var xdg_action: ?OptionalFileAction = null;
+    var xdg_path: ?[]const u8 = null;
+    while (try it.next()) |dir| {
+        defer alloc.free(dir);
+        xdg_path = dir;
+        xdg_action = self.loadOptionalFile(alloc, dir);
     }
-    // Apply the system configs from last to first
-    for (xdg_config_dirs.items) |xdg_dir| {
-        const config_path = try std.fs.path.join(alloc, &[_][]const u8{
-            xdg_dir,
-            config_subdir,
-            config_file,
-        });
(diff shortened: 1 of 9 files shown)


src/config/Config.zig
```diff
@@ -2911,20 +2913,44 @@ fn writeConfigTemplate(path: []const u8) !void {
...
-    // Load XDG first
-    const xdg_path = try internal_os.xdg.config(alloc, .{ .subdir = "ghostty/config" });
+    const config_subdir = "ghostty";
+    const config_file = "config";
+    // Load XDG system config first
+    var it = internal_os.xdg.DirIterator.init(.config);
+    // We must reverse the order of the iterator because the first xdg config
+    // dir must takes importance per the spec https://specifications.freedesktop.org/basedir-spec/latest/#variables
+    var xdg_config_dirs = std.ArrayList([]const u8).init(alloc);
+    defer xdg_config_dirs.deinit();
```
mitchellh, line 2930: Instead of all this machinery, the `DirIterator` could do a reverse iteration using `std.mem.lastIndexOf` to implement your own iterator. I would recommend this route.

mitchellh (changes requested):
Thanks for building on @jcollie's work here and keeping the commit credits. I made some notes.

mitchellh:
Also please note my feedback on @jcollie's PR, it applies here as well.

mjrochford:
Re. API design
### Current API
> os.xdg.config(alloc, .{ .subdir = "ghostty/config" });
> os.xdg.cache(alloc, .{ .subdir = "ghostty/config" });

### This API
To get all the xdg dirs in order from least to most important
user dirs come last after all others have been consumed
> var it = os.xdg.Dir.config.iter(alloc, .{ .subdir = "ghostty/config" });
Single directory access is possible ala
> os.xdg.UserDir.config.path(alloc, .{ .subdir = "ghostty" });
> os.xdg.UserDir.data.path(alloc, .{ .subdir = "ghostty" });
> os.xdg.UserDir.cache.path(alloc, .{ .subdir = "ghostty" });
> os.xdg.UserDir.state.path(alloc, .{ .subdir = "ghostty" });
Ditto for system directories
> os.xdg.SystemDir.config.iter(alloc, .{ .subdir = "ghostty" });

I've just gone ahead and used splitBackwardsScalar instead of splitScalar, I think wrapping an std iterator is better here.
Least important first makes sense to me for the Config use case, not sure about other cases.

### Better API
> os.programdirs.config.iter(alloc)
I think splitting XDG/Windows/MacOS known directories logic into their own modules and unifying these into a programdirs enum could make sense something akin to ProjectDirs from [this](https://stackoverflow.com/questions/43853548/xdg-basedir-directories-for-windows) stack overflow post.

neutralinsomniac:
FYI, with this branch, I every time I launch ghostty a random 4-byte garbage-named directory ends up in my home directory. An example of three of these:
```
''$'\231\032\004'
''$'\033''w'$'\001'
''$'\307''0'$'\001'
```

tristan957:
Would you mind rebasing this? It sounds interesting for Linux packagers.

mitchellh:
Closing due to lack of rebase, we can come back to this. Sorry.

PR #4849, opened 2025-01-09
build: update zig-wayland to use new LazyPath API

mitchellh:
This is a more idiomatic way to handle build paths in Zig 0.13 and later.

### DIFF
--- a/src/build/SharedDeps.zig
+++ b/src/build/SharedDeps.zig
@@ -435,15 +435,13 @@ pub fn add(
                 if (self.config.x11) step.linkSystemLibrary2("X11", dynamic_link_opts);
 
                 if (self.config.wayland) {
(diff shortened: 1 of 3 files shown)


src/build/SharedDeps.zig
```diff
@@ -435,15 +435,13 @@ pub fn add(
                 if (self.config.x11) step.linkSystemLibrary2("X11", dynamic_link_opts);
 
                 if (self.config.wayland) {
-                    const scanner = Scanner.create(b, .{
+                    const scanner = Scanner.create(b.dependency("zig_wayland", .{}), .{
                         // We shouldn't be using getPath but we need to for now
```
jcollie, line 439: This comment should be removed.

PR #4850, opened 2025-01-09
GTK: refactor headerbar into separate Adwaita & GTK structs

jcollie:
There's one behavioral change here. Before this patch, if `gtk-titlebar=false` we _never_ created a headerbar. This explicitly contradicted the comments in the source, and the documentation for `gtk-titlebar` imply that if a window starts out without a titlebar it can be brought back later with the `toggle_window_decorations` keybind action.

After this patch, a headerbar is always created, but if `gtk-titlebar=false` or `window-decoration=false` it's immediately hidden.

I'm not sure how this interacts with the current SSD/CSD detection that seems to happen when running Ghostty on non-Gnome DEs so it'll be important to get #4724 merged (plus any follow ups) to enable more explicit control of SSD/CSD.

### DIFF
--- a/src/apprt/gtk/headerbar.zig
+++ b/src/apprt/gtk/headerbar.zig
@@ -4,93 +4,58 @@ const c = @import("c.zig").c;
 const Window = @import("Window.zig");
 const adwaita = @import("adwaita.zig");
 
-const AdwHeaderBar = if (adwaita.versionAtLeast(0, 0, 0)) c.AdwHeaderBar else void;
+const HeaderBarAdw = @import("headerbar_adw.zig");
+const HeaderBarGtk = @import("headerbar_gtk.zig");
 
 pub const HeaderBar = union(enum) {
-    adw: *AdwHeaderBar,
-    gtk: *c.GtkHeaderBar,
+    adw: HeaderBarAdw,
+    gtk: HeaderBarGtk,
 
-    pub fn init(window: *Window) HeaderBar {
-        if ((comptime adwaita.versionAtLeast(1, 4, 0)) and
-            adwaita.enabled(&window.app.config))
-        {
-            return initAdw(window);
+    pub fn init(self: *HeaderBar) void {
+        const window: *Window = @fieldParentPtr("headerbar", self);
+        if ((comptime adwaita.versionAtLeast(1, 4, 0)) and adwaita.enabled(&window.app.config)) {
+            HeaderBarAdw.init(self);
+        } else {
+            HeaderBarGtk.init(self);
         }
 
-        return initGtk();
-    }
-
-    fn initAdw(window: *Window) HeaderBar {
-        const headerbar = c.adw_header_bar_new();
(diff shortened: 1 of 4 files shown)


tristan957 (approved):
Love this! I agree that checking with the SSD/window manager folks would be good before merging this PR.

src/apprt/gtk/headerbar_adw.zig
```diff
@@ -0,0 +1,74 @@
...
+
+const Window = @import("Window.zig");
+const adwaita = @import("adwaita.zig");
+
+const HeaderBar = @import("headerbar.zig").HeaderBar;
+
+const AdwHeaderBar = if (adwaita.versionAtLeast(0, 0, 0)) c.AdwHeaderBar else anyopaque;
+const AdwWindowTitle = if (adwaita.versionAtLeast(0, 0, 0)) c.AdwWindowTitle else anyopaque;
+
+pub const HeaderBarAdw = struct {
```
pluiedev, line 12: Wouldn't it be easier for downstream importers for this to be a file-level struct?

src/apprt/gtk/headerbar.zig
```diff
@@ -4,93 +4,64 @@ const c = @import("c.zig").c;
...
-        return .{ .gtk = @ptrCast(headerbar) };
+        if (!window.app.config.@"gtk-titlebar" or !window.app.config.@"window-decoration")
+            self.setVisible(false);
     }
 
     pub fn setVisible(self: HeaderBar, visible: bool) void {
-        c.gtk_widget_set_visible(self.asWidget(), @intFromBool(visible));
+        switch (self) {
+            .adw => |adw| adw.setVisible(visible),
+            .gtk => |gtk| gtk.setVisible(visible),
```
pluiedev, line 29: Could use the `inline else` pattern in order to reduce typing

pluiedev (commented):
I don't think unconditionally adding the headerbar would be a problem — as long as it is invisible

src/apprt/gtk/headerbar_adw.zig
jcollie: Fixed.

src/apprt/gtk/headerbar.zig
jcollie: Fixed.

pseudocc:
Hi @jcollie , this PR introduces a bug when:
1. window-decoration=false and gtk-adwaita=true.
2. Enter full screen mode then exit it.
3. The header shows up.

 window-decoration=false and gtk-adwaita=false is okay.

jcollie:
> Hi @jcollie , this PR introduces a bug when:
> 
>     1. window-decoration=false and gtk-adwaita=true.
> 
>     2. Enter full screen mode then exit it.
> 
>     3. The header shows up.
> 
> 
> window-decoration=false and gtk-adwaita=false is okay.

It's not really a "bug" per se, but a difference in opinion on what the various options mean and how they interact. Without a fuller discussion and clarification of how the various options interact I'm not going to be making more changes to how these options work.

Issue #4732, opened 2025-01-07
GTK: Menu should be in context menu if no titlebar


mitchellh:
For our GTK build, we should put the menu into the context menu (as a `Menu =>` submenu) if the titlebar is disabled. This will allow users to access the menu even when the titlebar is disabled or SSD is used.

tristan957:
We could use https://docs.gtk.org/gtk4/class.PopoverMenuBar.html.

tristan957:
We could also use a `GtkBox` styled with [`.toolbar`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/style-classes.html#toolbars).

tristan957:
I guess you would also wrap it in an [`AdwToolbarView`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.ToolbarView.html) like what we currently have.

mitchellh:
Fixed by #4864

PR #4864, opened 2025-01-09
feat(GTK): show menu in context menu if titlebar is disabled

ariasmn:
This PR addresses https://github.com/ghostty-org/ghostty/issues/4732.

While @tristan957 suggested alternative approaches, this implementation provides a straightforward way to make the menu accessible when the window decoration is disabled. It follows patterns seen in other GTK apps for handling submenus, though not strictly in the context menu format truth be told.

If there’s a better way to approach this or further refinements needed, I’m happy to discuss and iterate. This has been a minor issue I’ve encountered personally, and I’d like to help improve the experience for others as well.

Small video of how it looks:

https://github.com/user-attachments/assets/59548fef-f11c-421f-b05b-be81eab6ce06

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -831,12 +831,12 @@ fn setSizeLimit(
     switch (target) {
         .app => {},
         .surface => |v| try v.rt_surface.setSizeLimits(.{
-                    .width = value.min_width,
-                    .height = value.min_height,
-                }, if (value.max_width > 0) .{
-                    .width = value.max_width,
-                    .height = value.max_height,
-                } else null),
+            .width = value.min_width,
+            .height = value.min_height,
+        }, if (value.max_width > 0) .{
+            .width = value.max_width,
+            .height = value.max_height,
+        } else null),
     }
 }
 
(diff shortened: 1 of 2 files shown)


mitchellh:
This looks great to me and was exactly what I had in mind. I'll wait for a @jcollie or @tristan957 review. Even if we can improve this in some way later I think getting this in sooner than later would be good; its something a lot of users hit.

jcollie:
This is fine as a quick fix. I have plans to make the menu/context menu more integrated but it may be the weekend before I can get enough time to implement it.

tristan957:
I don't love it, but I also won't be using it.

mitchellh:
In the limited other software I've looked at where you can disable titlebars or decorations, this seems fairly idiomatic. It's also been the most common ask by Gnome-haters just behind SSD support (explicitly asking for menu bars in the context menu). We'll have to remember when SSD support is merged in to verify that we also enables the menu in the right click. I plan on taking over this PRs myself so I'll check into it.

mitchellh:
(My above comment isn't to invalidate any other approaches, its just that I'm happy to have this in for now and we can discuss its merit when other options pop up -- pun intended!)

ariasmn:
I agree with you all. This might not be the best approach, and there’s definitely room for improvement, but I’m glad I could contribute something (even as a quick fix) in the meantime.

Thank you all for your amazing work!

tristan957:
> In the limited other software I've looked at where you can disable titlebars or decorations, this seems fairly idiomatic.

If that's the case, then so be it. I definitely think a GtkPopoverMenuBar with a GtkRevealer is a better solution, similar to what Firefox does if you hit the `Alt` key. You never have to leave the keyboard. Can you access this context menu only via keyboard?

tristan957:
> I’m glad I could contribute something (even as a quick fix) in the meantime.

Welcome to the club :rocket: :fireworks:.

PR #4871, opened 2025-01-09
Remove unused MetalView

jacobsandlund:
I wanted to see where metal rendering happens and found this file, but I noticed it's not being used.


mitchellh:
I'm sort of keeping this on purpose in case it comes in handy... we did use it at some point and I felt it could be useful. Thanks though!

PR #4876, opened 2025-01-09
cli: add +boo command

rockorager:
Add a `+boo` command to show the animation from the website. The data
for the frames is compressed during the build process. This build step
was added to the SharedDeps object because it is used in both
libghostty and in binaries.

The compression is done as follows:

- All files are concatenated together using \x01 as a combining byte
- The files are compressed to a cached build file
- A zig file is written to stdout which `@embedFile`s the compressed
  file and exposes it to the importer
- A new anonymous module "framedata" is added in the SharedDeps object

Any file can import framedata and access the compressed bytes via
`framedata.compressed`. In the `boo` command, we decompress the slice and
split it into frames for use in the animation.

The overall addition to the binary size is 348k.


rockorager:
I also tried using the zig `std.compress.zlib` namespace, even though this is using `flate` under the hood. No appreciable change in binary size (~10 bytes smaller)

jcollie:
Would it make sense to add `https://github.com/ghostty-org/website` to `build.zig.zon` and get the frames from there, rather than copying them into the Ghostty source code? Won't help the binary size but seems "cleaner" to me.

rockorager:
> Would it make sense to add `https://github.com/ghostty-org/website` to
> `build.zig.zon` and get the frames from there, rather than copying them into
> the Ghostty source code? Won't help the binary size but seems "cleaner" to me.

Sounds like a good idea to me.

tristan957:
> I think this should be totally optional and standalone, and shouldn't be included in the binary

Please explain why.

rhodes-b:
> I think this should be totally optional and standalone, and shouldn't be included in the binary

patches are always something optional you can apply as well

mitchellh:
Sorry for the delay on this @rockorager. Can you rebase this and I'll bring it in. 😄

rockorager:
Rebased, but the nix cache is failing and I don't seem to have the tools to update the cache. @mitchellh could you push a commit with the cache update?

PR #4885, opened 2025-01-09
[GTK4] Disallow tab bar focussing

danudey:
~Adds configuration to allow/~Disallow the Adwaita tab bar from gaining focus with an additional click. This is the default behaviour of the tab bar, and it allows users to use the arrow keys to navigate the tabs and then the tab key to move to the window surface.

Personally, I hate when this happens and don't ever want my tabs focussed by accident, so I thought it might make sense to add ~a configuration option~ functionality to prevent this from happening.

Works great with adwaita 1.5 (Ubuntu 24.04) and adwaita 1.1 (Ubuntu 22.04). Tab bars in 22.04 are showing up above the window title bar, but ~~I'm not sure if that's related to this change~~ ~~that seems to be unrelated to this change~~ that is resolved separately in #5410.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -294,6 +294,10 @@ pub fn init(self: *Window, app: *App) !void {
             const tab_bar = c.adw_tab_bar_new();
             c.adw_tab_bar_set_view(tab_bar, self.notebook.adw.tab_view);
 
+            if (!app.config.@"gtk-tabs-can-focus") {
+                c.gtk_widget_set_can_focus(@ptrCast(@alignCast(tab_bar)), 0);
+            }
+
             if (!app.config.@"gtk-wide-tabs") c.adw_tab_bar_set_expand_tabs(tab_bar, 0);
 
             const tab_bar_widget: *c.GtkWidget = @ptrCast(@alignCast(tab_bar));
@@ -347,6 +351,10 @@ pub fn init(self: *Window, app: *App) !void {
                     .hidden => unreachable,
                 }
                 c.adw_tab_bar_set_view(tab_bar, adw.tab_view);
+
+                if (!app.config.@"gtk-tabs-can-focus") {
+                    c.gtk_widget_set_can_focus(@ptrCast(@alignCast(tab_bar)), 0);
+                }
 
                 if (!app.config.@"gtk-wide-tabs") c.adw_tab_bar_set_expand_tabs(tab_bar, 0);
             },
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -294,9 +294,7 @@ pub fn init(self: *Window, app: *App) !void {
             const tab_bar = c.adw_tab_bar_new();
             c.adw_tab_bar_set_view(tab_bar, self.notebook.adw.tab_view);
 
-            if (!app.config.@"gtk-tabs-can-focus") {
-                c.gtk_widget_set_can_focus(@ptrCast(@alignCast(tab_bar)), 0);
-            }
+            c.gtk_widget_set_can_focus(@ptrCast(@alignCast(tab_bar)), 0);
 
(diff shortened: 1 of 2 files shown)


tristan957:
I can't say that I agree with the premise. Why can't you just use keybinds with `next_tab/previous_tab` actions to fit your workflow?

tristan957:
> Tab bars in 22.04 are showing up above the window title bar, but I'm not sure if that's related to this change that seems to be unrelated to this change.

Known problem if you're interested in fixing it in a separate PR. Would be good to have a fix before the next release. I tried reproducing the issue but never could.

jcollie:
> > Tab bars in 22.04 are showing up above the window title bar, but I'm not sure if that's related to this change that seems to be unrelated to this change.
> 
> Known problem if you're interested in fixing it in a separate PR. Would be good to have a fix before the next release. I tried reproducing the issue but never could.

I _think_ that I solved that along with my menu reorganization PR (#4952), but that'll have to wait until Mitchell finishes his GTK App/Window refactor because I need to touch some of the same code.

mitchellh:
I don't love the configuration on this.

I think my question instead would be: is it desirable for the tab bar to EVER have keyboard focus? Is that a common thing? Tab navigation is already heavily bindable in Ghostty and I do agree with the premise that the tab bar getting focus from an extra click is not ideal.

tristan957:
> I don't love the configuration on this.
> 
> I think my question instead would be: is it desirable for the tab bar to EVER have keyboard focus? Is that a common thing? Tab navigation is already heavily bindable in Ghostty and I do agree with the premise that the tab bar getting focus from an extra click is not ideal.

You know I check what GNOME Text Editor does, and they seem to make the tab bar not focusable if you click on a tab. Though you can still tab through the widget tree to focus it if you try hard enough. I would say matching that behavior makes sense. Instead of call `gtk_widget_can_focus()` or whatever. The tab page contents should instead grab the focus.

mitchellh:
> The feature request makes sense to me, and we should definitely get it in for 1.1!

Fantastic. Thanks for doing the research. I agree. I think at least for 1.1, making the tab bar not focusable at all (as this PR does -- but without the config) seems fine. I think its far far far less desirable for that to ever gain focus than the incidental case you actually want it to via the keyboard or something.

danudey:
Sorry for the delay, but I've taken out the configuration option and made it so that the "don't focus" functionality is the only behavior. PTAL and let me know if there's anything else you think could be improved.

danudey:
> > Tab bars in 22.04 are showing up above the window title bar, but I'm not sure if that's related to this change that seems to be unrelated to this change.
> 
> Known problem if you're interested in fixing it in a separate PR. Would be good to have a fix before the next release. I tried reproducing the issue but never could.

@tristan957 Fixed in #5410

danudey:
@mitchellh I think this PR is ready for review, and pairs well with the recently merged PR to fix the Adwaita <1.4 tab bar issue as well. Hoping we can get this in for 1.1, whenever that is planned for.

tristan957:
How come when I click tabs on my system the tab bar is not focused?

danudey:
> How come when I click tabs on my system the tab bar is not focused? Libadwaita version is 1.6.2. Is it possible that this was a bug in a previous version of libadw?

Do you mean without this patch? I'm not certain, it's possible that the default behavior was changed. I'll see if I can reproduce. Either way, I think this PR makes sense for people on 22.04 and below.

tristan957:
> I also see very little down side to the one-line patch of explicitly making the tab bar non-focusable with GTK

It breaks the tab focus hierarchy if you do that.

mitchellh:
> It breaks the tab focus hierarchy if you do that.

mitchellh:
> It breaks the tab focus hierarchy if you do that.

Then let's just close this? I haven't ever seen any other discussion complaining about this issue so while I'm appreciative of the PR, I'd be willing to just wait until there's something more apparent here.

jcollie:
Closing, per the discussion.

PR #4893, opened 2025-01-10
vim: update configuration to include theme files

anund:
Extends vim validation and file type detection to theme files

cc @gpanders @beaumccartney  (as you were involved on the previous vim PR)

### DIFF
--- a/src/config/vim.zig
+++ b/src/config/vim.zig
@@ -3,7 +3,16 @@ const Config = @import("Config.zig");
 
 /// This is the associated Vim file as named by the variable.
 pub const syntax = comptimeGenSyntax();
-pub const ftdetect = "au BufRead,BufNewFile */ghostty/config set ft=ghostty\n";
+pub const ftdetect =
+    \\" Vim filetype detect file
+    \\" Language: Ghostty config file
+    \\" Maintainer: Ghostty <https://github.com/ghostty-org/ghostty>
+    \\"
+    \\" THIS FILE IS AUTO-GENERATED
+    \\
+    \\au BufRead,BufNewFile */ghostty/config,*/ghostty/themes/* set ft=ghostty
+    \\
+;
 pub const ftplugin =
     \\" Vim filetype plugin file
     \\" Language: Ghostty config file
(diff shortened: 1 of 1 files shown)


jparise (approved):

src/config/vim.zig
```diff
@@ -31,12 +40,18 @@ pub const ftplugin =
...
+    \\"
+    \\" THIS FILE IS AUTO-GENERATED
+    \\
     \\if exists("current_compiler")
     \\  finish
     \\endif
     \\let current_compiler = "ghostty"
     \\
-    \\CompilerSet makeprg=ghostty\ +validate-config
+    \\CompilerSet makeprg=ghostty\ +validate-config\ --config-file='%:p'
```
gpanders, line 54: `%:S` escapes special characters and is common in compiler files.

```suggestion
    \\CompilerSet makeprg=ghostty\ +validate-config\ --config-file=%:S
```

src/config/vim.zig
```diff
@@ -31,12 +40,18 @@ pub const ftplugin =
...
     \\CompilerSet errorformat=%f:%l:%m
```
gpanders, line 55: I didn't take a closer look at this before, but is this efm actually correct? When I run `ghostty +validate-config` with an invalid option it does not print out the filename or line number, only the message. In which case `'errorformat'` should just be `%m`. If `validate-config` ever does print the filename and line number in `filename:line:message` format, then this would be correct.

anund: It's mostly correct. Ghostty has ~3 or more output formats depending on where an error happens. The most common error you'll see is something like:
- `/home/<user>/.config/ghostty/themes/testing:4:not-a-field: unknown field`
- `/home/<user>/.config/ghostty/themes/testing:4:palette: invalid value "not right"`
- `<file>:<number>: <message>"`(looking at possibilities in `diagnostic.zig`.`write`, may not be possible atm)

Those examples can go wrong in the current code:
- `not-a-field: unknown field` (location = none, anytime a key is replayed)
- `error opening config-file /home/<user>/.config/ghostty/themes/nope: error.FileNotFound` (non-reproducible steps generated in `loadRecursiveFiles`)
- `cli:4:palette: invalid value "not right"` (not technically a worry here)

anund:
```suggestion
    \\CompilerSet errorformat=%f:%l:%m
    \\CompilerSet errorformat+=%m
```

I think will work but I was unsure where to verify that second step in vim.

anund: I'll do (1) for this PR. (2) is something I've been poking at. It's just taking a while to unwind the current code.

gpanders (approved):

PR #4902, opened 2025-01-10
feat: add save instruction to +list-themes

### LINKED ISSUE #4731: Pressing `<enter>` in `+list-themes` should explain how to save theme
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4708

<sup>Originally posted by **jcollie** January  6, 2025</sup>
Pressing `<enter>` in `+list-themes` should pop up a message with instructions on how to edit the config file and reload. I've seen many people confused that there's no way to set the theme from the TUI.</div>

### Solution

Approved as proposed until #2353 is complete and we can actually set the theme.

Aaron-212:
Close #4731 

image

### DIFF
--- a/src/cli/list_themes.zig
+++ b/src/cli/list_themes.zig
@@ -192,6 +192,7 @@ const Preview = struct {
         normal,
         help,
         search,
+        save,
     },
     color_scheme: vaxis.Color.Scheme,
     text_input: vaxis.widgets.TextInput,
@@ -377,6 +378,8 @@ const Preview = struct {
                             self.mode = .help;
                         if (key.matches('/', .{}))
                             self.mode = .search;
+                        if (key.matchesAny(&.{ vaxis.Key.enter, vaxis.Key.kp_enter }, .{}))
+                            self.mode = .save;
                         if (key.matchesAny(&.{ 'x', '/' }, .{ .ctrl = true })) {
                             self.text_input.buf.clearRetainingCapacity();
                             try self.updateFiltered();
@@ -430,6 +433,12 @@ const Preview = struct {
                         }
                         try self.text_input.update(.{ .key_press = key });
                         try self.updateFiltered();
+                    },
+                    .save => {
+                        if (key.matches('q', .{}))
+                            self.should_quit = true;
+                        if (key.matchesAny(&.{ vaxis.Key.escape, vaxis.Key.enter, vaxis.Key.kp_enter }, .{}))
+                            self.mode = .normal;
                     },
                 }
             },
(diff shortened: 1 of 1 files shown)


jcollie (approved):

jainhitesh9998:
@Aaron-212  @mitchellh.

ghostty doesnt create a config file by default right?
I'm using it in mac and had to create it manually at either XDG or ~/.config/ghostty/config.

if this specific reference can be added in the +list-themes functionality or a default empty config is created during installation, it'll make things more accessible to a lot of people.

themes is probably only customization most do. 

i have had a few of my friends try ghostty. 
the terminal enthusiasts customized it to their liking really fast. the occasional users took some time going through the docs or needed help.

pluiedev:
> ghostty doesnt create a config file by default right? I'm using it in mac and had to create it manually at either XDG or ~/.config/ghostty/config.

It does. #3460

mitchellh:
Note @pluiedev there's some bugs in 1.0.1 where if the dir doesn't exist we don't create the file. I think these were resolved on tip though.

src/cli/list_themes.zig
```diff
@@ -724,6 +733,51 @@ const Preview = struct {
                 child.fill(.{ .style = self.ui_standard() });
                 self.text_input.drawWithStyle(child, self.ui_standard());
             },
+            .save => {
+                const theme = self.themes[self.filtered.items[self.current]];
+
+                const width = 90;
+                const height = 12;
```
pluiedev, line 740: Just curious — where do these values come from?

src/cli/list_themes.zig
```diff
@@ -724,6 +733,51 @@ const Preview = struct {
...
+                        },
+                    },
+                );
+
+                child.fill(.{ .style = self.ui_standard() });
+
+                const save_instructions = [_][]const u8{
+                    "To apply this theme, add the following line to your Ghostty configuration:",
+                    "",
+                    try std.fmt.allocPrint(alloc, "theme = {s}", .{theme.theme}),
```
pluiedev, line 759: This leaks memory. I think just using a fixed-size buffer works (set it to something big like 128 or 256 bytes)

pluiedev (changes requested):

src/cli/list_themes.zig
Aaron-212: I'm pretty new to Zig. Do you mean the following?
```zig
var buffer: [128]u8 = undefined;
const theme_line = try std.fmt.bufPrint(&buffer, "theme = {s}", .{theme.theme});
```

src/cli/list_themes.zig
Aaron-212: I tested a few value and this one looks the best.

src/cli/list_themes.zig
jcollie: This line _would_ leak memory if the allocator passed to the draw function wasn't an arena allocator. Using a fixed size stack-allocated buffer in this case would cause problems because the memory needs to outlive the `draw()` function and eventually get used by the `libvaxis` `render()` function. All the memory is cleaned up at the end of each run through the event loop.

https://github.com/ghostty-org/ghostty/blob/ff9414d9ea7b16a375d41cde8f6f193de7e5db72/src/cli/list_themes.zig#L248-L262

mitchellh (approved):

Issue #4516, opened 2025-01-03
macOS: background bleeds/flicker when resizing splits, creating tabs


mitchellh:
Discussion: #4182, #4181, #4180, #3508

I'm not sure what is causing this yet but it'd be great to nail this down and clean this up. 

https://github.com/user-attachments/assets/d596061c-f638-4850-8ea3-b5ecc78f813b

andrewcrook:
I get this as well 

![Image](https://github.com/user-attachments/assets/73a42d90-72b2-4ea5-b4ca-8729720b6d29)

dmehala:
I suspect what we're observing is the latency between the surface redraw and the window resize with a transparent background. Since https://github.com/ghostty-org/ghostty/commit/fa8314058552bda5c7c09f769f7fef7c97aa39f4 the background is fully transparent.

Here's a video, top is without https://github.com/ghostty-org/ghostty/commit/fa8314058552bda5c7c09f769f7fef7c97aa39f4 and bottom 1.01+f4a9b65f:

https://github.com/user-attachments/assets/b3502207-a42f-4d5d-a934-a3074e950d87

Unfortunately, the solution isn't as simple as reverting fa8314058552bda5c7c09f769f7fef7c97aa39f4, because the issue will persist when using `background-opacity = 0.x`.

andrewcrook:
I find its not as bad having background-opacity < 1 like 0.9 than having it as  1.

I wonder if there could be or is an option to turn off redrawing during resizing and dragging? 
Just have the overlay and redraw after such an action is completed? 

I know this can be done on macOS via an OS wide setting, however,  can it be done per app i.e Ghostty’s configuration?

andreafspeziale:
Hello everyone, sorry to jump in the conversation. I was looking for an existing "flickering" issue and I found this one. I'm experiencing something similar, not while resizing but while downloading smth.

Here is a quick video:

https://github.com/user-attachments/assets/cd6ce186-5b14-4cc0-9b18-0ff4fd74aefb

Let me know if it can be related to the current issue of if I should fill a new one.

TY for your hard work 🤗

andreafspeziale:
I also recently noticed that opening a new tab with `cmd+t` has some kind of flickering too

https://github.com/user-attachments/assets/895d96ab-42d5-41f1-a696-9b3eac37d38f

I'm on the latest version

edrozenberg:
[...]

PR #4913, opened 2025-01-10
Metal alpha blending fixes + color handling improvements

qwerasd205:
This PR addresses #2125 for the Metal renderer. Both options are available: "Apple-style" blending where colors are blended in a wide gamut color space, which reduces but does not eliminate artifacts; and linear blending where colors are blended in linear RGB.

Because this doesn't add support for linear blending on Linux, I don't know whether the issue should be closed or not.

### List of changes in no particular order
- We now set the layer's color space in the renderer not in the apprt
- We always set the layer to Display P3 color spaces
- If the user hasn't configured their `window-colorspace` to `display-p3` then terminal colors are automatically converted from sRGB to the corresponding Display P3 color in the shader
- Background color is not set with the clear color anymore, instead we explicitly set all bg cell colors since this is needed for minimum contrast to not break with dark text on the default bg color (try it out, it forces it fully white right now), and we just draw the background as a part of the bg cells shader. Note: We may want to move the main background color to be the `backgroundColor` property on the `CAMetalLayer`, because this should fix the flash of transparency during startup (#4516) and the weirdness at the edge of the window when resizing. I didn't make that a part of this PR because it requires further changes and my changes are already pretty significant, but I can make it a follow-up.
- Added a config option for changing alpha blending between "native" blending, where colors are just blended directly in sRGB (or Display P3) and linear blending, where colors are blended in linear space.
- Added a config option for an experimental technique that I think works pretty well which compensates for the perceptual thinning and thickening of dark and light text respectively when using linear blending.
- Custom shaders can now be hot reloaded with config reloads.
- Fixed a bug that was revealed when I changed how we handle backgrounds, page widths weren't being set while cloning the screen.

### Main takeaways
Color blending now matches nearly identically to Apple apps like Terminal.app and TextEdit, not *quite* identical in worst case scenarios, off by the tiniest bit, because the default color space is *slightly* different than Display P3.

Linear alpha blending is now available for mac users who prefer more accurate color reproduction, and personally I think it looks very nice with the alpha correction turned on, I will be daily driving that configuration.

### Future work
- Handle primary background color with `CALayer.backgroundColor` instead of in shader, to avoid issues around edges when resizing.
- Parse color space info directly from ICC profiles and compute the color conversion matrix dynamically, and pass it as a uniform to the shaders.
- Port linear blending option to OpenGL.
- Maybe support wide gamut images (right now all images are assumed to be sRGB).

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -264,23 +264,15 @@ const c = @cImport({
 ///   This is also sometimes known as "gamma correction".
 ///   (Currently only supported on macOS. Has no effect on Linux.)
 ///
-///   To prevent the uneven thickness caused by linear blending, you can use
-///   the `experimental-linear-correction` option which applies a correction
-///   curve to the text alpha depending on its brightness, which compensates
-///   for the thinning and makes the weight of most text appear very similar
-///   to when it's blendeded non-linearly.
+/// * `linear-corrected` - Corrects the thinning/thickening effect of linear
+///   by applying a correction curve to the text alpha depending on its
+///   brightness. This compensates for the thinning and makes the weight of
+///   most text appear very similar to when it's blended non-linearly.
 ///
 /// Note: This setting affects more than just text, images will also be blended
 /// in the selected color space, and custom shaders will receive colors in that
 /// color space as well.
 @"text-blending": TextBlending = .native,
-
-/// Apply a correction curve to text alpha to compensate for uneven apparent
-/// thickness of different colors of text, roughly matching the appearance of
-/// text rendered with non-linear blending.
-///
-/// Has no effect if not using linear `text-blending`.
-@"experimental-linear-correction": bool = false,
 
 /// All of the configurations behavior adjust various metrics determined by the
 /// font. The values can be integers (1, -1, etc.) or a percentage (20%, -15%,
@@ -5787,6 +5779,14 @@ pub const GraphemeWidthMethod = enum {
 pub const TextBlending = enum {
     native,
     linear,
+    @"linear-corrected",
+
+    pub fn isLinear(self: TextBlending) bool {
+        return switch (self) {
+            .native => false,
+            .linear, .@"linear-corrected" => true,
+        };
+    }
 };
 
 /// See freetype-load-flag
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -248,6 +248,40 @@ const c = @cImport({
 /// This is currently only supported on macOS.
 @"font-thicken-strength": u8 = 255,
 
+/// What color space to use when performing alpha blending.
+///
+/// This affects how text looks for different background/foreground color pairs.
+///
+/// Valid values:
+///
+/// * `native` - Perform alpha blending in the native color space for the OS.
+///   On macOS this corresponds to Display P3, and on Linux it's sRGB.
+///
+/// * `linear` - Perform alpha blending in linear space. This will eliminate
+///   the darkening artifacts around the edges of text that are very visible
+///   when certain color combinations are used (e.g. red / green), but makes
+///   dark text look much thinner than normal and light text much thicker.
+///   This is also sometimes known as "gamma correction".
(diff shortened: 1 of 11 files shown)


src/config/Config.zig
```diff
@@ -248,6 +248,40 @@ const c = @cImport({
...
+/// in the selected color space, and custom shaders will receive colors in that
+/// color space as well.
+@"text-blending": TextBlending = .native,
+
+/// Apply a correction curve to text alpha to compensate for uneven apparent
+/// thickness of different colors of text, roughly matching the appearance of
+/// text rendered with non-linear blending.
+///
+/// Has no effect if not using linear `text-blending`.
+@"experimental-linear-correction": bool = false,
```
mitchellh, line 283: Not ready to review the whole thing but knee jerk reaction: should this be an alternate enum value for `text-blending`, i.e. `linear-experimental`?

qwerasd205: Probably. I set up the config options and then sorta just left them how I put them at first. It might also make sense to rename `text-blending` to `alpha-blending`. I'd also kinda like to rename `window-colorspace` to `terminal-colorspace` since what it does is define how *terminal colors* should be interpreted for rendering, but that felt out-of-scope for this PR.

src/font/face/coretext.zig
```diff
@@ -343,13 +343,12 @@ pub const Face = struct {
...
-            .space = try macos.graphics.ColorSpace.createDeviceGray(),
-            .context_opts = @intFromEnum(macos.graphics.BitmapInfo.alpha_mask) &
-                @intFromEnum(macos.graphics.ImageAlphaInfo.only),
+            .space = try macos.graphics.ColorSpace.createNamed(.linearGray),
+            .context_opts = @intFromEnum(macos.graphics.ImageAlphaInfo.only),
         } else .{
             .color = true,
             .depth = 4,
-            .space = try macos.graphics.ColorSpace.createDeviceRGB(),
+            .space = try macos.graphics.ColorSpace.createNamed(.displayP3),
```
AlexJuca, line 351: @qwerasd205 *suggestion*: Should we make this `...ColorSpace.createFromName(.displayP3)`?

qwerasd205: Why? I could understand changing it to `createWithName` if we wanted to match the name of the CoreGraphics function, but this really feels like splitting hairs either way. I'm not devoted to the name, but if I changed it I'd be more likely to change it to `createWithName` than `createFromName` unless there's particularly good reasoning...

AlexJuca: @qwerasd205 The way I see it, we are creating a `ColorSpace` from a `ColorSpace.Name` called `.displayP3`.

From that perspective, it would logically flow into `ColorSpace.createFromName(.displayP3)`.

I do think `createWithName` is also a great option.

mitchellh: Technically our `macos` pkg is a thin layer over Carbon/AppKit APIs and we just mirror their names, so _technically_ the correct name here is `createWithName` but I'm not going to lose sleep over it and don't really care.

src/terminal/PageList.zig
```diff
@@ -520,6 +520,7 @@ pub fn clone(
         assert(node.data.capacity.rows >= chunk.end - chunk.start);
         defer node.data.assertIntegrity();
         node.data.size.rows = chunk.end - chunk.start;
+        node.data.size.cols = chunk.node.data.size.cols;
```
mitchellh, line 523: lol.

src/config/Config.zig
mitchellh: I moved the linear correction into `text-blending`.

src/terminal/PageList.zig
qwerasd205: This took me over an hour to find 😭

src/font/face/coretext.zig
AlexJuca: @mitchellh @qwerasd205 Im also happy if we leave it the way it is. 

Resolving comment.

PR #4927, opened 2025-01-11
Handle setting _NET_WM_STATE

### LINKED ISSUE #4646: GTK/X11: set `_NET_WM_STATE(ATOM)` as maximized when clamping to window size
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4570

<sup>Originally posted by **KotonBads** January  4, 2025</sup>
with [this pr](https://github.com/ghostty-org/ghostty/pull/4305) being closed, setting `window-width` and `window-height` to a size larger than the screen should set it's wm state to maximized

currently it just resizes the window to the screen size
image

instead of having a separate option, it would be nice if it just does it automatically
image</div>

aw1875:
Adds the logic to set the `_NET_WM_STATE` appropriately based on whether the window is maximized or not. I wasn't sure about where some of this code should live after the unified winproto code was added so definitely open to moving logic around if needed.

It might be nice to move the direct X server communication to a more central location. Doing this would allow us to individually set `_NET_WM_STATE_MAXIMIZED_VERT` and `_NET_WM_STATE_MAXIMIZED_HORZ` on the first load depending on what matches (currently it is all or nothing)

Closes https://github.com/ghostty-org/ghostty/issues/4646


jcollie:
I don't think that this is the correct approach to handling "maximized" windows in GTK/Gnome. GTK already manages the X11 `_NET_WM_STATE` property (see [here](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/x11/gdksurface-x11.c) and this just feels like we're trying to work against GTK rather than with it.

In GTK "maximized" is a unique state that is different from merely _really big_. The main differences are changes in the window decoration and being able to toggle between maximized and the pre-maximized window size.

You can see how this affects Ghostty by enabling the maximize/minimize buttons using Gnome tweaks or by double-clicking in the header bar. Compare that to how the window looks when you just manually resize the window to the limits of the display.

I think that a better approach would be to add a `maximize` config entry so that Ghostty windows can start out maximized and a `toggle_maximize` keybind action so that windows can be maximized with a keyboard shortcut.

If we _really must_ have the maximized state set when windows are created with a size that exceeds the size of the display we shouldn't be manipulating the X11 state directly, we should be using the GTK `gtk_window_maximize`/`gtk_window_unmaximize` APIs.

aw1875:
@jcollie you know what, I just realized that I only tested the props in gnome when I had the `gtk-titlebar` disabled before :facepalm:, so all my original testing around this issue was wrong.

I feel like a config option for starting maximized a shortcut toggle would be a great solution here. I'm not sure if you think it makes the most sense to just fix these changes here or create a separate PR?

jcollie:
> I'm not sure if you think it makes the most sense to just fix these changes here or create a separate PR?

I would start a new PR I think.

aw1875:
Okay I've got another branch put together with the proposed implementation so going to close this PR out.

PR #4930, opened 2025-01-11
declare `StartupWMClass` in .desktop

Samueru-sama:
This allows the app to be pinned in the dock/task manager in several common desktop environments. 

When not set this happens when you open the pinned application: 

image


jcollie (approved):

mitchellh:
Thanks!

PR #4949, opened 2025-01-11
Correct IME position calculation with window padding

### LINKED ISSUE #4933: macOS: IME incorrect position when using `window-padding-x`/`window-padding-y`
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4883

<sup>Originally posted by **grant0417** January  9, 2025</sup>
When using either of the configs `window-padding-x`/`window-padding-y` the IME positions is incorrectly reported on macOS. This was discovered as part of this issue: https://github.com/aws/amazon-q-developer-cli/issues/395 

https://github.com/user-attachments/assets/9e80be9a-3c7e-4f71-bba3-e5596f088437

</div>

liby:
This PR addresses two IME-related issues:

1. Fix IME position calculation with window padding, fixed https://github.com/ghostty-org/ghostty/issues/4933

   - Add padding offset to IME position calculation to correctly position the IME window
   
   - This ensures IME window appears at the right position when `window-padding-x/window-padding-y` are set

2. Clear selection when IME input starts

   - Previously, when using CJK IME with text selected, the IME window would appear near the selection
   
   - Now it clears selection and shows IME window at cursor position
   
   - This matches the behavior of iTerm2 and Terminal.app

## Before

https://github.com/user-attachments/assets/cd5c7b55-2083-40ce-a528-9a98898a6498

## After

https://github.com/user-attachments/assets/9f2c17c6-e885-45f3-9ab1-a9c3858690ec

### DIFF
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -1550,11 +1550,6 @@ pub const CAPI = struct {
         return selection.len;
     }
 
-    /// Clear the current selection in the surface.
-    export fn ghostty_surface_clear_selection(surface: *Surface) void {
-        surface.core_surface.io.terminal.screen.clearSelection();
-    }
-
     /// Tell the surface that it needs to schedule a render
     export fn ghostty_surface_refresh(surface: *Surface) void {
         surface.refresh();
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -1550,6 +1550,11 @@ pub const CAPI = struct {
         return selection.len;
     }
 
+    /// Clear the current selection in the surface.
+    export fn ghostty_surface_clear_selection(surface: *Surface) void {
+        surface.core_surface.io.terminal.screen.clearSelection();
+    }
+
     /// Tell the surface that it needs to schedule a render
     export fn ghostty_surface_refresh(surface: *Surface) void {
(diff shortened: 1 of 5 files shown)


src/apprt/embedded.zig
```diff
@@ -1550,6 +1550,11 @@ pub const CAPI = struct {
         return selection.len;
     }
 
+    /// Clear the current selection in the surface.
+    export fn ghostty_surface_clear_selection(surface: *Surface) void {
+        surface.core_surface.io.terminal.screen.clearSelection();
```
mitchellh, line 1555: This is unsafe because the IO thread owns the terminal, not the GUI thread. This requires a lock. But the lock shouldn't be held outside of Surface.zig. We should add a `clearSelection` function to `Surface.zig`.

mitchellh (changes requested):
Looks good, confirmed the selection behavior with other apps. One request.

mitchellh:
Made the necessary changes. I removed the libghostty API call. It's not necessary since we can do this all in core in a place we have the lock anyways. 

The logic wasn't also strictly correct as far as I can tell when comparing to Terminal.app. I believe the updated logic is correct. We now only unselect when IME text starts. The previous one also cleared when IME began (no input yet).

PR #4952, opened 2025-01-11
gtk: rework Windows and menus

jcollie:
1. Rework the GTK Window code to clean up the if/else spaghetti. This
  _should_ fix issues with older versions of Adwaita getting the
  titlebar and tab bar out of order.
2. Consolidate code for menus into one file and switch to using
  GtkPopupMenus built from GTK Builder XML files. This changes menus
  so that there is one per window and one per surface. This results
  in more memory usage, but more correct behavior. Previously context
  menus would pop up at the wrong location, due to not being attached
  to the correct GTK widget. Using GTK Builder XML files reduces the
  amount of code to create the menus and will make future changes to
  the menu structure easier.
3. Add a "top menu" that can be shown/hidden with a keybind action.
  This will be useful for people that use SSD and thus don't have the
  hamburger menu from the title bar.

Screenshot From 2025-01-12 14-06-01
Screenshot From 2025-01-12 14-06-14


tristan957:
I don't agree with the premise of this PR. I think we should continue to follow patterns established by other GTK/GNOME apps.

unikitty37:
> I don't agree with the premise of this PR. I think we should continue to follow patterns established by other GTK/GNOME apps. "About Ghostty" is not useful in a context menu.

It is if you have `gtk-titlebar` set to `false` and can't access the hamburger menu, though, as there doesn't seem to be any other way to get to it short of adding a keybinding, which is completely excessive :)

Leaving `gtk-titlebar` set to `true` makes Ghostty completely fail to fit in with any WM that _isn't_ Gnome.

src/apprt/gtk/ui/menu-window.ui
```diff
@@ -0,0 +1,92 @@
+<interface domain="com.mitchellh.ghostty">
```
tristan957, line 1: Might be good to add the `<xml>` declaration line.

tristan957:
```
 <?xml version="1.0" encoding="UTF-8">
```

mitchellh:
There is **a lot** I like in this PR, but there are some fundamental things I don't really like and want more discussion on.

Things I like:

- Menu showing up in the right place
- Cleaned up main menu i.e. supporting all split directions

Things I'm indifferent about (don't care either way):

- Increased memory usage. Its probably small given the benefit.
- GTK builder files.

Things I don't like:

- The context menu and hamburger menu being the same. I don't think they serve the same purpose and long term I don't think it makes sense to always have those match. I also agree with @tristan957 that some times like "About Ghostty" are nonsense in the context menu.
- The context menu is huge (number of items). I'd prefer to make it smaller. To address the non-titlebar people, I like our current approach of nesting a "Menu" submenu. 
- "Top menu" terminology. I don't know what a "top menu" is. Maybe it's common maybe it's not, but Googling at least didn't reveal a lot of usage. To me, there's really just a menu and a context menu. No need to introduce "top".

If we can discuss or address the "don't like" list then I'd be happy to head towards merging this.

jcollie:
OK, I think that I've addressed all of the feedback so far. "Top bar" was an invention of mine because I couldn't think of anything else, renamed to just "menubar". The context menu is slimmed down. I've also added compile-time checks of
the GTK builder files so it should be much harder to crash Ghostty with an invalid or missing file.

I don't like the idea of embedding the hamburger menu in the context menu, as there's _a lot_ of duplication. It would be much easier to have section(s) of the context menu that get show/hidden based on the state of the titlebar. Not sure what is really "needed" but IMHO that can be bikeshedded later. Since the menus are declared as XML rather than code it's pretty easy to update them.

Screenshot From 2025-01-24 21-36-12

tristan957:
Thanks for the latest updates. I can't say that I enjoy using the popover menu for the context menu. I still think we should just use the regular rectangular one like all other applications that I can think of.

jcollie:
> Thanks for the latest updates. I can't say that I enjoy using the popover menu for the context menu. I still think we should just use the regular rectangular one like all other applications that I can think of.

Is it the arrow that you object to? That's easily removed.

Screenshot From 2025-01-27 21-15-47 Screenshot From 2025-01-27 21-18-13

jcollie:
The latest patches add the ability to manually rename tabs. You can restore a manually renamed tab to normal by renaming the tab to a "blank" title. At some point I'd like to add a keybind action to rename the tab (so in theory a tab could be renamed from the keyboard only).

Screenshot From 2025-01-27 21-22-43
Screenshot From 2025-01-27 21-22-55
Screenshot From 2025-01-27 21-23-01

tristan957:
Should this PR maybe be split up for easier reviewing? For instance, fixing the location of the context menu.

jcollie:
Going to overhaul this once we've removed non-adwaita support.

PR #4956, opened 2025-01-11
macos: paste multiple files separated by space

mitchellh:
https://github.com/ghostty-org/ghostty/discussions/4892#discussioncomment-11808631


jcollie:
This is a behavior difference from GTK - multiple files are separated by a newline.

mitchellh:
> This is a behavior difference from GTK - multiple files are separated by a newline.

I think for now having difference here is probably fine. This seems like a cultural divide.

## macOS

- Terminal.app: space
- iTerm2: space
- WezTerm: space
- Warp: space
- Kitty: newline
- Alacritty: no support

## Linux

- Gnome: newline
- Xfce: newline
- xterm: no support (or doesn't work for me)

mitchellh:
Fwiw I do prefer spaces, I think that's far more useful in the likely scenario you're pasting onto a CLI command. I don't understand the benefit of a newline.

PR #4962, opened 2025-01-11
feat: ensure text, files and URLs can be drag and dropped to terminal

### LINKED ISSUE #4932: macOS: Dragged text should insert text
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4857

<sup>Originally posted by **danieleavitabile** January  8, 2025</sup>
In MacOS Ghostty accepts drag and drop of files from Finder. Ghostty, however,  doesn't seem to accept dragged text. In MacOS several applications support dragging text (Firefox, other browsers, Zotero for bigliographies for instance), but Ghostty does not accept that this text is dropped, while iTerm lets you do that.  

I attach a screen recording that shows what this allows to do under iTerm, and how it fails on Ghostty. 

https://github.com/user-attachments/assets/d0091f34-251a-4036-9b07-c94f5e99346f

To me this seems a useful feature to have on MacOS, and it's part of the integration in that ecosystem.

</div>

### Security Note

We should be very careful with this to ensure that newlines and bracketed paste escape do not happen. It may be useful to run this through the unsafe paste sanitization.

AlexJuca:
This PR ensures one can drag and drop the following things to the terminal window:

- File (Inserts the file path into the terminal)
- URL
- Text

This resolves #4932

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView.swift
+++ b/macos/Sources/Ghostty/SurfaceView.swift
@@ -92,22 +92,6 @@ extension Ghostty {
                                 windowFocus = false
                             }
                         }
-                        .onDrop(of: [.fileURL], isTargeted: nil) { providers in
-                            providers.forEach { provider in
-                                _ = provider.loadObject(ofClass: URL.self) { url, _ in
-                                    guard let url = url else { return }
-                                    let path = Shell.escape(url.path)
-                                    DispatchQueue.main.async {
-                                        surfaceView.insertText(
-                                            path,
-                                            replacementRange: NSMakeRange(0, 0)
-                                        )
-                                    }
-                                }
-                            }
-
-                            return true
-                        }
                     #endif
 
                     // If our geo size changed then we show the resize overlay as configured.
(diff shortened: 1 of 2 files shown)


qwerasd205:
Note, we already handle dragging *files* via https://github.com/ghostty-org/ghostty/blob/bdbd0263a1c55e656408c772718370c92374bedb/macos/Sources/Ghostty/SurfaceView.swift#L95-L107 and the additional support of dragged text should probably be added *there* instead.

mitchellh:
> and the additional support of dragged text should probably be added there instead.

I think the view is actually a better place and we should think about removing the SwiftUI side. I tried implementing this on the SwiftUI side and ran into SwiftUI issues where when specifying `fileURL, plainText`, files would ONLY report `plainText`. I could never find a way around this. So if this works it's more promising.

Pulling it down now.

mitchellh:
Thank you. I made a cleanup commit feel free to take a look. Notes:

* Moved drag implementation out to a separate extension
* Handle dropping multiple files the same way we paste multiple files
* Use `copy` dragging operation so we get a cursor icon by AppKit
* Unify draggingEntered and registration to share the same underlying array of accepted types

PR #4973, opened 2025-01-12
For GTK runtime, don't call cursorPosCallback if cursor pos didn't actually change.

### LINKED ISSUE #3345: Mouse is shown when title changes with mouse-hide-while-typing enabled
When setting the option `mouse-hide-while-typing` to `true` anytime the title changes for any reason the mouse is shown again.
For example: open Neovim, set the `title` option with `:set title` and then open a split to a different file. When switching between these to windows with `Ctrl + w + w` the mouse is shown again.

I did some digging and found that commenting out [this](https://github.com/ghostty-org/ghostty/blob/a8e5eef11cc67f87f445626f9ca2993373774bf8/src/Surface.zig#L3308) `self.showMouse()` call fixes this specific issue.

Ghostty version:
```
Ghostty 1.0.0-4b4d406-nix

Version
  - version: 1.0.0-4b4d406-nix
  - channel: tip
Build Config
  - Zig version: 0.13.0
  - GTK version:
    build      : 4.16.3
    runtime    : 4.16.3
  - libadwaita : enabled
    build      : 1.6.2
    runtime    : 1.6.2
```

jnichols0:
Fixes #3345

Piggybacking on the logic introduced PR https://github.com/ghostty-org/ghostty/pull/3997, this patch prevents mouse motion events with the same cursor position triggered by the window title updating from un-hiding the mouse even when `mouse-hide-while-typing` config item is true.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -1496,31 +1496,37 @@ fn gtkMouseMotion(
         .y = @floatCast(scaled.y),
     };
 
-    // When the GLArea is resized under the mouse, GTK issues a mouse motion
-    // event. This has the unfortunate side effect of causing focus to potentially
-    // change when `focus-follows-mouse` is enabled. To prevent this, we check
-    // if the cursor is still in the same place as the last event and only grab
-    // focus if it has moved.
+    // There seem to be at least two cases where GTK issues a mouse motion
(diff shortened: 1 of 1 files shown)


mitchellh:
Works well for me, thank you

Scutua (approved):

PR #4977, opened 2025-01-12
Adding $EDITOR check for open

ScriptAlchemist:
Allowing for users to set the $EDITOR variable and use it to open a file instead of the default. I don't know zig and I'm testing this out. I just wanted to allow others to look at it as well.

This should solve [#2504](https://github.com/ghostty-org/ghostty/issues/2504)

Still figuring out the builds at the moment to really test out better

### DIFF
--- a/src/os/open.zig
+++ b/src/os/open.zig
@@ -19,27 +19,12 @@ pub fn open(
     url: []const u8,
 ) !void {
     const cmd: OpenCommand = switch (builtin.os.tag) {
-        .linux => .{ .child = std.process.Child.init(
-            &.{ "xdg-open", url },
-            alloc,
-        ) },
-
+        .linux => try determineOpenCommandLinux(alloc, url),
         .windows => .{ .child = std.process.Child.init(
             &.{ "rundll32", "url.dll,FileProtocolHandler", url },
             alloc,
         ) },
-
-        .macos => .{
-            .child = std.process.Child.init(
-                switch (typ) {
-                    .text => &.{ "open", "-t", url },
-                    .unknown => &.{ "open", url },
-                },
-                alloc,
-            ),
-            .wait = true,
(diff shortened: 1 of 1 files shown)


xzbdmw:
Hi, does this pr open $EDITOR in another ghostty window?

ScriptAlchemist:
> Hi, does this pr open $EDITOR in another ghostty window?

I believe that it's like running `nvim ./filepath.txt` so it would be in the current buffer screen

I have not successfully built this on mac. I'm not trying that hard on the weekend.

ScriptAlchemist:
Until I fix the build this does not work

xzbdmw:
I'll suggest opening a new window and run $EDITOR there, because the current shell may have running process ;)

jcollie:
This is the wrong approach to opening up a new Ghostty window and running `$EDITOR $config`. I realize that this is something that people _really_ want but this really needs to wait for some of the API work for this to work properly.

ScriptAlchemist:
> This is the wrong approach to opening up a new Ghostty window and running `$EDITOR $config`. I realize that this is something that people _really_ want but this really needs to wait for some of the API work for this to work properly.

Okay, @jcollie then just close this pr out. It doesn't even build right now. So obviously this isn't the correct method. I don't know zig and I haven't spend enough time to figure it out. Well enough apparently.

I just found the file that does the thing I want. Then was going to add the feature that I wanted. Instead of just complaining, but figured having it more public here would give me better insight.

I really just want it to open in place like everything else that works in the terminal.

imadtg:
> I'll suggest opening a new window and run $EDITOR there, because the current shell may have running process ;)

could you not have a conditional check if a current process is running and only open a new window/buffer if one is already running? this seems better than doing either unconditionally

PR #4995, opened 2025-01-12
config: overhaul help string generation

pluiedev:
Overhauls the way help strings are generated by also generating documentation for the possible values allowed for each type, adds the default value of each key, and deduplicating type-specific documentation (e.g. `Color` and `Duration`).

Still WIP: Webgen works but other forms of documentation are untested. Action and KeybindAction also likely have to be rewritten.


pluiedev:
Will close this in favor of a more comprehensive config introspection API that's required for things like GUI settings

PR #5000, opened 2025-01-12
macos: only set quick terminal level to popUpMenu during animation

### LINKED ISSUE #4999: macOS: Quick terminal covers IM widget
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4979

<sup>Originally posted by **rmGFW** January 12, 2025</sup>
image
</div>

mitchellh:
Fixes #4999

We need to set the level to popUpMenu so that we can move the window offscreen and animate it over the main menu, but we must reset it back to floating after the animation is complete so that other higher-level windows can be shown on top of it such as IME windows.

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalWindow.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalWindow.swift
@@ -28,23 +28,5 @@ class QuickTerminalWindow: NSPanel {
 
         // We don't want to activate the owning app when quick terminal is triggered.
         self.styleMask.insert(.nonactivatingPanel)
-
-        // We need to set our window level to a high value. In testing, only
-        // popUpMenu and above do what we want. This gets it above the menu bar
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/QuickTerminal/QuickTerminalWindow.swift
```diff
@@ -28,23 +28,5 @@ class QuickTerminalWindow: NSPanel {
...
-        // we just want the behavior of.
-        self.collectionBehavior = [
-            // We want this to be part of every space because it is a singleton.
-            .canJoinAllSpaces,
-
-            // We don't want to be part of command-tilde
-            .ignoresCycle,
-
-            // We want to show the window on another space if it is visible
-            .fullScreenAuxiliary]
```
mitchellh, line 48: All of this now gets managed by the controller so we don't need to set it on the window.

PR #5006, opened 2025-01-12
[macOS] feat: add `bring_all_to_front` keybinding

### LINKED ISSUE #4704: Keybinding Action to "Bring all to front"
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4696

<sup>Originally posted by **massimopalmieri** January  6, 2025</sup>
It would be great to be able to have a keybind action to focus the terminal ("Bring all to front"), just like `toggle_visibility` Shows/hides all windows.</div>

### Proposal

Add a new keybinding `bring_all_to_front` that does this. I believe we can make this work on GTK too.

dmehala:
Can't consider the feature complete until the Linux (GTK) implementation :/ .

Fixes #4704

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -41,6 +41,7 @@ class AppDelegate: NSObject,
 
     @IBOutlet private var menuToggleVisibility: NSMenuItem?
     @IBOutlet private var menuToggleFullScreen: NSMenuItem?
+    @IBOutlet private var menuBringAllToFront: NSMenuItem?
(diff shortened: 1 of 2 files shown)


mitchellh:
I was ready to merge this but actually @pluiedev's comment here makes sense: https://github.com/ghostty-org/ghostty/issues/4704#issuecomment-2609288470

There's some complications to that idea though but thanks to the 1.1.1 release cycle we actually have the core machinery to make this work. I'm working through it now.

mitchellh:
Since this is a pretty standard macOS action, I removed all the binding stuff and just added the macOS menu item and behavior. This will let us ship this behavior and then deal with the binding later.

Issue #3381, opened 2024-12-27
GTK: Add an option to hide window decorations when the window is maximized


ansipunk:
I find window decorations useless when the window is maximized, so I'd like to hide them to preserve screen space, since I use tmux for multiplexing the terminal. Foot has an option for that and I imagine it wouldn't be too difficult to implement it here.

ohmsl:
Could use
https://ghostty.org/docs/config/reference#window-decoration
if you want to always disable it

ansipunk:
@ohmsl then I'd lose the ability to pull edges and corners of the window, move it around when it's not maximized and would have to use a keybinding to toggle. I think it could be improved.

ohmsl:
Agreed, it should at least remove the window decorations when full screened, whether this needs to be an option I don't know but +1 from me

mitchellh:
I prefer a discussion for feature requests but I'm willing to accept this one on the spot. It's a good idea. 

This should only apply to GTK. macOS already has its own hidden decorations when fullscreen (that appear when you hover the mouse up top).

isgj:
IMO it should hide the decorations when it goes full-screen,  not when the window is maximized.

Pangoraw:
@ansipunk would something like this help ?

ansipunk:
@Pangoraw thanks, I'll keep on eye on that discussion, it indeed does look great.

yorickpeterse:
The trick I was using was to use `ghostty --gtk-titlebar=false` as this would result in a basic GTK titlebar, which _did_ get hidden when enabling fullscreen mode. It seems that per 1.0.1 this approach no longer works, as `--gtk-titlebar=false` completely removes the title bar on Gnome.

In my case I use Neovim in Ghostty in fullscreen mode, and not being able to hide the title bar in full screen mode is rather annoying due to the amount of space it takes up.

Doom4535:
I might add that an auto hide type behavior would be nice (like how the Windows taskbar can be configured to disappear when the mouse isn't near it), as the GUI buttons are convenient, but for smaller (and lower res) screens real-estate is at a premium and the decorations bar takes up a good amount of space (or at least on my Ubuntu system).  I am using the `window-decoration = false` setting currently to work around this, but sometimes miss the buttons.

ansipunk:
Please do not confuse full screen and maximized modes.

jcollie:
[...]

PR #5008, opened 2025-01-13
gtk: hide titlebar if fullscreened

jcollie:
Partially addresses #3381

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -204,6 +204,7 @@ pub fn init(self: *Window, app: *App) !void {
     }
 
     _ = c.g_signal_connect_data(gtk_window, "notify::decorated", c.G_CALLBACK(&gtkWindowNotifyDecorated), self, null, c.G_CONNECT_DEFAULT);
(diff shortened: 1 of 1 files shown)


mitchellh:
Does this make sense to do unconditionally or should this be a config? I'm not suggesting it should be a config, I'm just genuinely curious what would be expected here. For example, on macOS, the OS handles this for you and unconditionally hides the title (but shows it when the mouse goes near the top -- provided by the OS so not suggesting we do that here).

jcollie:
> Does this make sense to do unconditionally or should this be a config?

Every other app that I know of that supports fullscreen unconditionally hides the titlebar if you fullscreen it. The "canonical" examples would be browsers like firefox, chrome, etc. Firefox temporarily unhides the titlebar if you move the mouse near the top of the screen and chromium-based browsers show a "X" icon that lets you exit fullscreen. I agree that we don't necessarily need to do something like that.

We might consider adding F11 (in addition to ctrl+enter) as a default keybind for toggling fullscreen as that's the default keybind in browsers for entering fullscreen.

Issue #4800, opened 2025-01-08
`copy-on-select = clipboard` causes "Copied to clipboard" toast spam


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4766

<div type='discussions-op-text'>

<sup>Originally posted by **dumbasPL** January  7, 2025</sup>
host: Arch Linux, GNOME 47 on Wayland
version: 1.0.2-main+a3837a1e (built from source)
simplified config (without theming):
```
shell-integration-features = no-cursor
copy-on-select = clipboard
scrollback-limit = 134217728
adw-toolbar-style = raised-border
auto-update = off
term = xterm-256color
```
steps to reproduce:
1. enable `copy-on-select = clipboard`
2. select something by clicking and dragging the mouse
3. observe `Copied to clipboard` spam at the bottom

expected behavior:  
`Copied to clipboard` shows once the selection has ended (when the mouse button is lifted)</div>

### Solution

I think we should also look into the core here (`src/Surface.zig`) to see if we should only apply the selection when a mouse release event happens. 

We should probably also change the GTK implementation to only show the toast after a very short (50ms or so) delay.

lcx:
this happens with MacOS as well, it's constantly copying to clipboard. When using something like alfred which has an option to merge clipboards if cmd+c is pressed twice, the result is very bizzare merging of multiple entries to the clipboard.
Turning this of in alfred still shows the multiple entries of the same data copied to clipboard in the clipboard history.

gegoune:
I have enquired about this a while ago on discord. See link below for screenshot of what's described by @lcx above. 

https://discord.com/channels/1005603569187160125/1313506853765779528

JeffreyVdb:
This also means that when you're using a clipboard manager, the list of entries is going to get cluttered very quickly. A selection of 10 characters basically means that you'll have 10 entries in your clipboard manager, every subsequent entry being one character longer than the one preceding it. Your last entry will be the one you want but you'll quickly erase other clipboard entries unless your history is really big.

otomist:
Going to take a shot at fixing this

otomist:
> I think we should also look into the core here (`src/Surface.zig`) to see if we should only apply the selection when a mouse release event happens.

For mouse release there's functionality for following links. Is it safe to assume that following links should take precedence? 
Having both happen works too but my general thought is if someone is holding Ctrl they are probably looking to click a link and not copy text.

jnichols0:
[...]

PR #5010, opened 2025-01-13
Fix copy_on_select GTK toast spam

jnichols0:
Fixes core issue for #4800, moves copy_on_select clipboard logic in `src/Surface.zig` out of `setSelection` and into `handleMouseButton`. When left mouse is released and the current selection is not empty, saves selection to clipboard and sends an adw toast. The selection will still be updated as the user drags, however the clipboard will be written to just once when the mouse button is released. 

This patch doesn't add a delay to sending the toast as suggested, wasn't sure how to best do that without throwing it in an async function. Open to suggestions. :)

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1393,67 +1393,7 @@ fn clipboardWrite(self: *const Surface, data: []const u8, loc: apprt.Clipboard) 
 ///
 /// This must be called with the renderer mutex held.
 fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
-    const prev_ = self.io.terminal.screen.selection;
     try self.io.terminal.screen.select(sel_);
-
-    // If copy on select is false then exit early.
-    if (self.config.copy_on_select == .false) return;
-
-    // Set our selection clipboard. If the selection is cleared we do not
-    // clear the clipboard. If the selection is set, we only set the clipboard
-    // again if it changed, since setting the clipboard can be an expensive
-    // operation.
-    const sel = sel_ orelse return;
-    if (prev_) |prev| if (sel.eql(prev)) return;
-
-    const buf = self.io.terminal.screen.selectionString(self.alloc, .{
-        .sel = sel,
-        .trim = self.config.clipboard_trim_trailing_spaces,
-    }) catch |err| {
-        log.err("error reading selection string err={}", .{err});
-        return;
-    };
-    defer self.alloc.free(buf);
-
-    // Set the clipboard. This is not super DRY but it is clear what
-    // we're doing for each setting without being clever.
-    switch (self.config.copy_on_select) {
-        .false => unreachable, // handled above with an early exit
-
-        // Both standard and selection clipboards are set.
-        .clipboard => {
-            const clipboards: []const apprt.Clipboard = &.{ .standard, .selection };
-            for (clipboards) |clipboard| self.rt_surface.setClipboardString(
-                buf,
-                clipboard,
-                false,
-            ) catch |err| {
-                log.err(
-                    "error setting clipboard string clipboard={} err={}",
-                    .{ clipboard, err },
-                );
-            };
-        },
-
-        // The selection clipboard is set if supported, otherwise the standard.
-        .true => {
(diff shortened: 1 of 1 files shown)


src/Surface.zig
```diff
@@ -2879,6 +2819,64 @@ pub fn mouseButtonCallback(
...
+        action == .release and
+        self.config.copy_on_select != .false)
+    set_clipboard: {
+        self.renderer_state.mutex.lock();
+        defer self.renderer_state.mutex.unlock();
+
+        // Only update clipboard if something is actually selected.
+        const selection = self.io.terminal.screen.selection orelse break :set_clipboard;
+
+        const buf = self.io.terminal.screen.selectionString(self.alloc, .{
```
jcollie, line 2833: This line is leaking memory.

src/Surface.zig
```diff
@@ -1393,67 +1393,7 @@ fn clipboardWrite(self: *const Surface, data: []const u8, loc: apprt.Clipboard)
...
-    if (prev_) |prev| if (sel.eql(prev)) return;
-
-    const buf = self.io.terminal.screen.selectionString(self.alloc, .{
-        .sel = sel,
-        .trim = self.config.clipboard_trim_trailing_spaces,
-    }) catch |err| {
-        log.err("error reading selection string err={}", .{err});
-        return;
-    };
-    defer self.alloc.free(buf);
```
jcollie, line 1416: Leak is caused because this line didn't get moved to the new location.

src/Surface.zig
```diff
@@ -2879,6 +2819,64 @@ pub fn mouseButtonCallback(
...
+        const selection = self.io.terminal.screen.selection orelse break :set_clipboard;
+
+        const buf = self.io.terminal.screen.selectionString(self.alloc, .{
+            .sel = selection,
+            .trim = self.config.clipboard_trim_trailing_spaces,
+        }) catch |err| {
+            log.err("error reading selection string err={}", .{err});
+            break :set_clipboard;
+        };
+        log.debug("copying {} to clipboards", .{selection});
```
jcollie, line 2840: Remove this log, especially since it could potentially leak sensitive data into the logs (like passwords or security tokens).

jcollie (changes requested):

jcollie:
I think adding a delay is unnecessary since the toast spam is fixed in a better way.

src/Surface.zig
jnichols0: Oops! Thanks for catching that.

jnichols0:
Latest commit fixes memory leak and removes log line.

jcollie (approved):

src/Surface.zig
```diff
@@ -2879,6 +2819,64 @@ pub fn mouseButtonCallback(
         }
     }
 
+    // Set the clipboard if configured with copy_on_select
+    if (button == .left and
```
pluiedev, line 2823: Do we want to restrict it only to the left mouse button? Sounds overly restrictive...

jnichols0: As far as I can tell, right now we only ever call setSelection with a non-null selection (and thus potentially copying to the clipboard) after one or more left clicks. I think adding that behavior for other mouse buttons might be out of the scope of this bug fix.

mitchellh:
The issue I have with this PR is that it makes it so that non-mouse selection (i.e. selection keybinds) do not update the selection clipboard. I'm not sure if that's intended behavior (I haven't checked other terminals that support it), but the way I currently expect selection to work is that any selection method would result in the selection keyboard being updated.

I think an alternate approach is needed here where maybe `setSelection` isn't changed, but our mouse events need to stop using it and set `terminal.screen.select` directly and then we call `setSelection` at the end (with the existing selection) in order to persist it once to the clipboard and avoid the toast spam.

jnichols0:
Ah I see. I'll take another look.

unknown:
@jnichols0 Any progress on this? If not, I'd be happy to work on it.

jnichols0:
> @jnichols0 Any progress on this? If not, I'd be happy to work on it.
Please go for it, I probably don't have time to tackle this soon.

mitchellh:
Thanks, superseded by #5995

PR #5021, opened 2025-01-13
Fix backslash comment in ghostty.h

askielboe:
The other half of https://github.com/ghostty-org/ghostty/pull/3173. :)


pluiedev (approved):
10\10

PR #5028, opened 2025-01-13
gtk: add config entry to hide titlebar when the window is maximized

### LINKED ISSUE #3381: GTK: Add an option to hide window decorations when the window is maximized

jcollie:
Fixes #3381

Note that #4936 will need to be merged or you'll need to rely on Gnome's default keybinding for unmaximizing a window (super+down).

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -204,6 +204,7 @@ pub fn init(self: *Window, app: *App) !void {
     }
 
     _ = c.g_signal_connect_data(gtk_window, "notify::decorated", c.G_CALLBACK(&gtkWindowNotifyDecorated), self, null, c.G_CONNECT_DEFAULT);
+    _ = c.g_signal_connect_data(gtk_window, "notify::maximized", c.G_CALLBACK(&gtkWindowNotifyMaximized), self, null, c.G_CONNECT_DEFAULT);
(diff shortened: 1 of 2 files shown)


tristan957 (approved):
Simple enough. Nice job. Didn't really realize people would use such a feature.

jcollie:
> Simple enough. Nice job. Didn't really realize people would use such a feature.

The people that want it seem quite _passionate_ about it.

src/apprt/gtk/Window.zig
```diff
@@ -588,6 +589,17 @@ fn gtkRealize(v: *c.GtkWindow, ud: ?*anyopaque) callconv(.C) bool {
     return true;
 }
 
+fn gtkWindowNotifyMaximized(
+    _: *c.GObject,
+    _: *c.GParamSpec,
+    ud: ?*anyopaque,
+) callconv(.C) void {
+    const self = userdataSelf(ud orelse return);
+    if (self.app.config.@"gtk-titlebar-hide-when-maximized") {
```
mitchellh, line 598: I think there's potentially a bug here, but haven't verified:

1. Set `gtk-titlebar-hide-when-maximized`
2. Maximize
3. Unset and reload config
4. Unmaximize, header bar is now missing.

I think the fix is easy enough: unconditionally show the header bar when the window is unmaximized.

mitchellh (changes requested):
One bug, but simple, thanks!

mitchellh (approved):

PR #5033, opened 2025-01-13
terminal: keep track of colon vs semicolon state in CSI params

### LINKED ISSUE #5022: Kakoune SGR sequences fails to parse
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4959

<sup>Originally posted by **ygrek** January 11, 2025</sup>
When I trigger undercurl in kakoune editor I get the following message in ghostty log
```
warning(io_handler): unimplemented or unknown SGR attribute: terminal.sgr.Attribute.Attribute__struct_127656{ .full = { 0, 4, 3, 38, 2, 208, 208, 208, 58, 2, 0, 190, 80, 70 }, .partial = { 70 } }
```
and it looks like green underline 

it works as expected with kakoune in the xfce-terminal (red undercurl)

ftr orange undercurl as expected is displayed when I do `echo -e '\e[4:3m\e[58:2::240:143:104mfoobarbaz\e[59m\e[4:0m\n'` in the shell

kakoune doesn't use curses and instead I believe has custom code to parse terminfo
it could be kakoune bug or combination of issues, I lack understanding of terminfo stuff to figure it out

PS built from v1.0.1-379-g918ccdba with ` zig build -p $HOME/.local -Doptimize=ReleaseFast` running with empty config
</div>

mitchellh:
Fixes #5022

The CSI SGR sequence (CSI m) is unique in that its the only CSI sequence that allows colons as delimiters between some parameters, and the colon vs. semicolon changes the semantics of the parameters.

Previously, Ghostty assumed that an SGR sequence was either all colons or all semicolons, and would change its behavior based on the first delimiter it encountered.

This is incorrect. It is perfectly valid for an SGR sequence to have both colons and semicolons as delimiters. For example, Kakoune sends the following:

    ;4:3;38;2;175;175;215;58:2::190:80:70m

This is equivalent to:

  - unset (0)
  - curly underline (4:3)
  - foreground color (38;2;175;175;215)
  - underline color (58:2::190:80:70)

This commit changes the behavior of Ghostty to track the delimiter per parameter, rather than per sequence. It also updates the SGR parser to be more robust and handle the various edge cases that can occur. Tests were added for the new cases.

### DIFF
--- a/src/terminal/Parser.zig
+++ b/src/terminal/Parser.zig
@@ -6,6 +6,7 @@ const Parser = @This();
 
 const std = @import("std");
 const builtin = @import("builtin");
+const assert = std.debug.assert;
 const testing = std.testing;
 const table = @import("parse_table.zig").table;
 const osc = @import("osc.zig");
@@ -81,11 +82,15 @@ pub const Action = union(enum) {
     pub const CSI = struct {
         intermediates: []u8,
         params: []u16,
+        params_sep: SepList,
         final: u8,
-        sep: Sep,
+
+        /// The list of separators used for CSI params. The value of the
+        /// bit can be mapped to Sep.
+        pub const SepList = std.StaticBitSet(MAX_PARAMS);
 
         /// The separator used for CSI params.
-        pub const Sep = enum { semicolon, colon };
+        pub const Sep = enum(u1) { semicolon = 0, colon = 1 };
 
         // Implement formatter for logging
         pub fn format(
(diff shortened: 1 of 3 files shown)


jerch:
In your recent blog article about libghostty you say:

> And I still haven't found a single web console that renders [this complex style sequence](https://github.com/ghostty-org/ghostty/blob/8cb52323e5575877eef44029c153eda006cece80/src/terminal/Parser.zig#L672-L722) correctly.

I guess that most webconsoles use xterm.js under the hood. As far as I know, there is nothing wrong in xterm.js with mixing colon and semicolon notation in SGR, this is supposed to work since colon notation got implemented. If there is indeed an issue, plz let us know. Thanks.

PR #5036, opened 2025-01-13
fix: Ensure file paths are properly escaped during pasteboard paste operations

AlexJuca:
This PR ensures file paths derived from pasteboard operations are properly escaped:

## Before fix:
Attempting to cd into these paths rendered the following error: `cd: too many arguments` because the directory name includes spaces or special characters, which need to be handled correctly when running the cd command.

https://github.com/user-attachments/assets/4d03a94a-fa14-4ac8-9c1e-1ad8b0b939a7

## Afer fix

File paths are now properly escaped and we can correctly `cd` into these paths.

https://github.com/user-attachments/assets/72e794c0-31e4-43b8-bebf-76c161924bd1

This change ensures Ghostty has the same behaviour as Iterm2.

### DIFF
--- a/macos/Sources/Helpers/NSPasteboard+Extension.swift
+++ b/macos/Sources/Helpers/NSPasteboard+Extension.swift
@@ -9,14 +9,14 @@ extension NSPasteboard {
 
     /// Gets the contents of the pasteboard as a string following a specific set of semantics.
     /// Does these things in order:
-    /// - Tries to get the absolute filesystem path of the file in the pasteboard if there is one.
+    /// - Tries to get the absolute filesystem path of the file in the pasteboard if there is one and ensures the file path is properly escaped.
     /// - Tries to get any string from the pasteboard.
     /// If all of the above fail, returns None.
     func getOpinionatedStringContents() -> String? {
         if let urls = readObjects(forClasses: [NSURL.self]) as? [URL],
            urls.count > 0 {
             return urls
-                .map { $0.isFileURL ? $0.path : $0.absoluteString }
+                .map { $0.isFileURL ? Ghostty.Shell.escape($0.path) : $0.absoluteString }
                 .joined(separator: " ")
         }
 


Mil0dV:
Ok so I don't know this code base enough to figure out if this also applies to linux (I searched and found bits of safepaste code, but I tested it, and it seems to fail exactly like you described (vids won't load for me). 

So I'd say: why not also fix this for linux?

AlexJuca:
> Ok so I don't know this code base enough to figure out if this also applies to linux (I searched and found bits of safepaste code, but I tested it, and it seems to fail exactly like you described (vids won't load for me).
> 
> So I'd say: why not also fix this for linux?

@Mil0dV I don't know if this issue exists for Linux.

Can you reproduce the issue on Linux? That would be nice.

Unfortunately, I can't because I don't use Linux, and the fix will likely involve very platform-specific changes.

Mil0dV:
@AlexJuca ah my edit and your reply crossed eachother. Yes, I reproduced this on linux. Shouldn't the same Ghostty.Shell.escape also work there, if the right file is found to insert it?

AlexJuca:
> @AlexJuca ah my edit and your reply crossed eachother. Yes, I reproduced this on linux. Shouldn't the same Ghostty.Shell.escape() also work there, if the right file is found to insert it?

@Mil0dV The `Ghostty.Shell.escape()` function is a Swift only API, so it won't work on Linux.

jcollie:
This was recently added to the Zig code:

https://github.com/ghostty-org/ghostty/blob/d1fd22ae80f764aea5c05987871a58a2fb776862/src/os/shell.zig#L9-L41

It's not exposed to macOS at the moment though.

AlexJuca:
@jcollie @mitchellh Quick question, I have contributed several times to ghostly prior to the existence of the new vouching system. 

Would I automatically be vouched?
Or would any PR I create be considered a first time contribution even though it's technically not?

jcollie:
!vouch @AlexJuca

PR #5050, opened 2025-01-14
fix(gtk/toggle_fullscreen): header bar visibility

pseudocc:
While unfullscreen the window, header.setVisible(true) will be called. For .gtk it's okay, because it will follows the window-decoration, while .adw doesn't.

After #4850, the header is always created and added to the window, so setting visibility to the correct value is vital for .adw.

Fixes: #4850

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -538,8 +538,8 @@ pub fn toggleMaximize(self: *Window) void {
 
 /// Toggle fullscreen for this window.
 pub fn toggleFullscreen(self: *Window) void {
-    const is_fullscreen = c.gtk_window_is_fullscreen(self.window);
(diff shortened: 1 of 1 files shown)


jcollie:
Unfortunately, the interaction between `window-decoration`, `gtk-titlebar` and the various options for toggling window state are poorly specified. Because my reading of the documentation comments is different, and it could be seen as a bug that the titlebar _doesn't_ reappear after toggling fullscreen. We need to have a larger discussion on these options and come up with ways that users can _clearly_ express their intentions for window decoration and titlebars. Until that happens, any PRs that change the behavior are just going to cause additional churn when people with differing interpretations report bugs or submit PRs.

mitchellh:
I ended up fixing this in #5090.

PR #5064, opened 2025-01-14
gtk: request initial color scheme asynchronously

### LINKED ISSUE #4632: Ghostty should not delay startup if dbus is unreasonably slow
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4620

<sup>Originally posted by **daveadams** January  5, 2025</sup>
Hello, I'm testing out Ghostty on Linux and I'm seeing an unusably long delay in opening the initial terminal window when starting ghostty on the order of one to two minutes. The logs indicate a timeout when querying DBus for the system theme.

There are other topics about "significant delays" but those seem to be on the order of a second or two rather than multiple minutes.

My system is Debian 12. I'm using Sway as my window manager, so Gnome services aren't fully available, which I suspect is the source of the problem.

Relevant log snippet (with timestamps added):

```
06:28:25 info: ghostty version=1.0.2-main+0306c592
06:28:25 info: ghostty build optimize=ReleaseFast
06:28:25 info: runtime=apprt.Runtime.gtk
06:28:25 info: font_backend=font.main.Backend.fontconfig_freetype
06:28:25 info: dependency harfbuzz=8.4.0
06:28:25 info: dependency fontconfig=21402
06:28:25 info: renderer=renderer.OpenGL
06:28:25 info: libxev backend=main.Backend.io_uring
06:28:25 info(os): setlocale from env result=LC_CTYPE=en_US.UTF-8;LC_NUMERIC=en_US.UTF-8;LC_TIME=C.UTF-8;LC_COLLATE=en_US.UTF-8;LC_MONETARY=en_US.UTF-8;LC_MESSAGES=en_US.UTF-8;LC_PAPER=en_US.UTF-8;LC_NAME=en_US.UTF-8;LC_ADDRESS=en_US.UTF-8;LC_TELEPHONE=en_US.UTF-8;LC_MEASUREMENT=en_US.UTF-8;LC_IDENTIFICATION=en_US.UTF-8
06:28:25 info(gtk): GTK version build=4.8.3 runtime=4.8.3
06:28:25 info: reading configuration file path=/home/dadams/.config/ghostty/config
06:28:25 info(config): default shell source=env value=/bin/bash
06:28:25 info(gtk): libadwaita version build=1.2.2 runtime=1.2.2
06:29:40 error(gtk): unable to get current color scheme: Timeout was reached
06:29:40 info(grid): loaded OpenGL 4.6
06:29:40 info(font_shared_grid_set): font regular: DejaVu Sans Mono
06:29:40 info(font_shared_grid_set): font bold: DejaVu Sans Mono Bold
06:29:40 info(font_shared_grid_set): font italic: DejaVu Sans Mono Oblique
06:29:40 info(font_shared_grid_set): font bold_italic: DejaVu Sans Mono Bold Oblique
06:29:40 info(io_exec): found Ghostty resources dir: /home/dadams/src/github.com/ghostty-org/ghostty/zig-out/share/ghostty
06:29:40 info(io_exec): shell integration automatically injected shell=termio.shell_integration.Shell.bash
06:29:40 warning(gtk): unimplemented action=apprt.action.Action.Key.cell_size
06:29:40 warning(gtk): unimplemented action=apprt.action.Action.Key.size_limit
06:29:40 info(io_exec): started subcommand path=/bin/sh pid=155494
06:29:40 info(io_exec): subcommand cgroup=-
06:30:05 error(gtk): unable to get current color scheme: Timeout was reached
06:30:05 info(grid): reallocating GPU buffer old=0 new=38
```

You can see the 75 second delay between reporting the libadwaita version and the "Timeout was reached message" and a second time, there's a 25 second delay in the penultimate log line.

Watching dbus-monitor, I see a lot of business going on as the DBus lib tries various ways to get to the API it's trying to call to get the color scheme, but I think ultimately this is the meaningful error:

```
method call time=1736081549.289863 sender=:1.68 -> destination=org.freedesktop.DBus serial=41 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=AddMatch
   string "type='signal',sender='org.freedesktop.DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',path='/org/freedesktop/DBus',arg0='org.freedesktop.IBus'"
method return time=1736081549.289893 sender=org.freedesktop.DBus -> destination=:1.68 serial=36 reply_serial=41
method call time=1736081549.289899 sender=:1.68 -> destination=org.freedesktop.DBus serial=42 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=GetNameOwner
   string "org.freedesktop.IBus"
error time=1736081549.289906 sender=org.freedesktop.DBus -> destination=:1.68 error_name=org.freedesktop.DBus.Error.NameHasNoOwner reply_serial=42
   string "Could not get owner of name 'org.freedesktop.IBus': no such name"
```

So it's pretty clear to me that if the correct Gnome services aren't running, there's a massive delay in showing the initial window.

There are some config settings related to forcing a particular theme. I tried setting those to no effect. The code doesn't really have any branch points around fetching this setting, so I wasn't surprised. It would be nice to have an option to tell ghostty not to bother trying to talk to such specific services and just let me tell it what theming I want.</div>

pluiedev:
Requesting the initial color scheme on systems where the D-Bus interface is nonexistent would delay Ghostty startup by 1-2 minutes. That's not acceptable. Our color scheme events are already async-friendly anyway.

Fixes #4632

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -73,6 +73,11 @@ clipboard_confirmation_window: ?*ClipboardConfirmationWindow = null,
 /// This is set to false when the main loop should exit.
 running: bool = true,
 
+/// If we should retry querying D-Bus for the color scheme with the deprecated
+/// Read method, instead of the recommended ReadOne method. This is kind of
+/// nasty to have as struct state but its just a byte...
+dbus_color_scheme_retry: bool = true,
+
 /// The base path of the transient cgroup used to put all surfaces
 /// into their own cgroup. This is only set if cgroups are enabled
 /// and initialization was successful.
@@ -1271,16 +1276,14 @@ pub fn run(self: *App) !void {
         self.transient_cgroup_base = path;
     } else log.debug("cgroup isolation disabled config={}", .{self.config.@"linux-cgroup"});
 
-    // Setup our D-Bus connection for listening to settings changes.
+    // Setup our D-Bus connection for listening to settings changes,
+    // and asynchronously request the initial color scheme
     self.initDbus();
 
     // Setup our menu items
     self.initActions();
     self.initMenu();
     self.initContextMenu();
-
-    // Setup our initial color scheme
-    self.colorSchemeEvent(self.getColorScheme());
 
     // On startup, we want to check for configuration errors right away
     // so we can show our error window. We also need to setup other initial
@@ -1328,6 +1331,22 @@ fn initDbus(self: *App) void {
         &gtkNotifyColorScheme,
         self,
         null,
+    );
+
+    // Request the initial color scheme asynchronously.
+    c.g_dbus_connection_call(
+        dbus,
+        "org.freedesktop.portal.Desktop",
+        "/org/freedesktop/portal/desktop",
+        "org.freedesktop.portal.Settings",
+        "ReadOne",
+        c.g_variant_new("(ss)", "org.freedesktop.appearance", "color-scheme"),
+        c.G_VARIANT_TYPE("(v)"),
+        c.G_DBUS_CALL_FLAGS_NONE,
+        -1,
+        null,
+        dbusColorSchemeCallback,
+        self,
     );
 }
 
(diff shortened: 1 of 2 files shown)


mitchellh:
> The optimization here, imo, is hard depend on libadwaita

What part is a hard dependency on adw? 

> Seems to me that the people who complain about this are people who don't actually know how to make functioning Linux systems.

This isn't a good reason to reject this. Ghostty should aim to do what its can to just work as much as possible without requiring perfect configuration. That could mean mitigation, detection, etc. Most people just want a working terminal and I consider it our job to provide that to the best of our abilities. 

In all projects I've ever run, my general philosophy is maintainers pick up complexity so consumers don't have to. The complexity will fall on the project.

tristan957:
> > The optimization here, imo, is hard depend on libadwaita
> 
> What part is a hard dependency on adw?

We should just listen to this property: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/property.StyleManager.color-scheme.html, which is already determined by the same D-Bus request we are already making here.

https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-settings-impl-portal.c#L247-256

mitchellh:
I see, I wonder if that'd break for people who have sufficiently old systems. I don't see libadw doing the fallback that we do, and that hit at least a couple people with the 1.0.0 release. I suppose this would fall in your category of "broken" systems but I'm not sure yet. Maybe so, maybe not. 

> The XDG Desktop portal is becoming a standard part of a Linux system. I'm not sure I understand why we cannot just depend on it existing.

Yeah. I'm not sure. 

It's hard on the Linux side because it's all a personal judgement call of what where one draws the line of broken vs not. On Windows/macOS we have a mega corp saying "this is EOL," which amongst its many warts, is extremely helpful for 3PD.

pluiedev:
I think WM users often just skip over it since it's not "technically" required to use the WM, and instructions on how to install `xdg-desktop-portal-{gtk,gnome}` is often put in postinstall instructions specifically for say, when Flatpak doesn't work. If we do decide to hard depend on that then we should make the error more decipherable.

tristan957:
> It's hard on the Linux side because it's all a personal judgement call of what where one draws the line of broken vs not.

It's not really personal judgement, but what app developers want. App developers want to target a platform, and Linux is not a platform. The XDG Desktop Portal helps app developers and platform providers achieve that.

The developers of Ghostty can easily just say the xdg-desktop-portal with the GTK backend is required. Thusly, any system that doesn't provide them is broken and should not expect Ghostty to work properly.

mitchellh:
> It's not really personal judgement, but what app developers want

I think this is personal judgment lol, since app developers are generally people.

> The developers of Ghostty can easily just say the xdg-desktop-portal with at least the GTK backend is required. Thusly, any system that doesn't provide them is broken and should not expect Ghostty to work properly.

Indeed. I'm not sure yet if I want to make this statement.

bphenriques:
Just observer and newish to linux customisation: I thought that people would like to keep libadwaita sort of optional specially those that prefer no decorations on their Windows (me personally). Hence, I guess, preferring the xdg standardisation

bphenriques:
Thank 

> The thing is that the whole no-window-decoration thing is separate (see #4724) and disabling it by disabling Adwaita is more of a hack (since you lose on _many_ features like the tab overview and popup toasts). If you use any GTK4 app chances are your system already pulls in libadwaita, which makes the argument against it somewhat weak.
> 
> There's also a good argument to argue for making Adwaita mandatory, which is that it allows us to drop a lot of duplicate code we have to maintain for GTK and Adwaita separately (e.g. the different header bar types, see #4850). Continuing to support configurations with Adwaita disabled feels like unnecessary burden if the only justification behind it is that Adwaita is technically optional and "bloat".

Thank you, wanted to learn a bit about the subject 👍

mitchellh:
Regardless of our long term approach, the initial startup delay found in #4632 needs to be resolved. I'm not opposed to going ads-only and thus relying only on adw style manager (unclear if it has this issue too though), but for a Ghostty 1.1 release we still should strive to resolve this issue given the current architecture.

This PR doesn't introduce any new architectural choices, it's just changing the approach we use to do what we always did. I'm going to review this now and plan to merge it. I'm happy in the future to rip it all out if we chase a different direction.

Issue #4957, opened 2025-01-11
Triple-clicking empty line and dragging does not set selection start


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4938

<div type='discussions-op-text'>

<sup>Originally posted by **injust** January 10, 2025</sup>
https://github.com/user-attachments/assets/70628017-20c6-4121-99ec-be7cad062c87</div>

### Reproduction

1. `cat README.md` (Ghostty's works)
2. Triple click any blank line
3. Drag up or down

rhodes-b:
This seems to be working for me on GTK, so it should have macos label?

otomist:
I'll pick this up

otomist:
Opened a PR at #5068 which should fix this. 

Tested on:
       OS: Pop 22.04 jammy
       Shell: bash 5.1.16                       
       Resolution: 2560x1440
       WM: Awesome
    



https://github.com/user-attachments/assets/4afaa863-0637-4362-a494-f22445f07e41

injust:
This was fixed by https://github.com/ghostty-org/ghostty/pull/5068 (I don't have permissions to close this issue).

PR #5068, opened 2025-01-14
Fix triple clicking empty line and dragging

otomist:
Fixes core issue #4957. Adds a bool to `SelectLine` allowing the `selectLine` function to select lines that are empty. When starting a triple selection on an empty line the initial `selectLine` returns null because we don't see any characters, in this case we rerun `selectLine` but short circuit with the `allow_empty_lines`. We need to run `selectLine` with out allowing empty lines once because if there are characters on the line we don't want to select empty space.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3563,17 +3563,15 @@ fn dragLeftClickTriple(
     const screen = &self.io.terminal.screen;
     const click_pin = self.mouse.left_click_pin.?.*;
 
-    // Get the line selection under our current drag point. If there isn't a line, do nothing.
+    // Get the line selection under our current drag point. If there isn't a
+    // line, do nothing.
     const line = screen.selectLine(.{ .pin = drag_pin }) orelse return;
 
-    // get the selection under our click point.
-    var sel_ = screen.selectLine(.{ .pin = click_pin });
-
-    // We may not have a selection if we started our triple-click in an area
-    // that had no data, in this case recall selectLine with allow_empty_lines.
-    if (sel_ == null) {
-        sel_ = screen.selectLine(.{ .pin = click_pin, .allow_empty_lines = true });
-    }
+    // Get the selection under our click point. We first try to trim
+    // whitespace if we've selected a word. But if no word exists then
+    // we select the blank line.
+    const sel_ = screen.selectLine(.{ .pin = click_pin }) orelse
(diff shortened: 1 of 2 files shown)


src/terminal/Screen.zig
```diff
@@ -2215,6 +2215,7 @@ pub const SelectLine = struct {
     /// state changing a boundary. State changing is ANY state
     /// change.
     semantic_prompt_boundary: bool = true,
+    allow_empty_lines: bool = false,
```
mitchellh, line 2218: Needs docs

src/terminal/Screen.zig
```diff
@@ -2292,6 +2293,11 @@ pub fn selectLine(self: *const Screen, opts: SelectLine) ?Selection {
         return null;
     };
 
+    // If we allow empty lines, we don't need to do any further checks.
+    if (opts.allow_empty_lines) {
+        return Selection.init(start_pin, end_pin, false);
+    }
```
mitchellh, line 2299: This needs to be unit tested. Everything in Screen should be unit tested (for historical reasons some parts aren't but going forward we should be adding it it all)

src/Surface.zig
```diff
@@ -3563,22 +3563,23 @@ fn dragLeftClickTriple(
...
+    const line = screen.selectLine(.{ .pin = drag_pin }) orelse return;
+
+    // get the selection under our click point.
+    var sel_ = screen.selectLine(.{ .pin = click_pin });
+
+    // We may not have a selection if we started our triple-click in an area
+    // that had no data, in this case recall selectLine with allow_empty_lines.
+    if (sel_ == null) {
+        sel_ = screen.selectLine(.{ .pin = click_pin, .allow_empty_lines = true });
+    }
```
mitchellh, line 3576: Use a labeled block to put all of this (line 3570 to here) in a `const` assignment.

mitchellh (changes requested):
The approach looks good, some requests.

src/terminal/Screen.zig
otomist: As I was making this change I realized setting whitespace to null does the equivalent to my new allow_empty_lines flag in screen. I adjusted the code to use whitespace instead so there actually aren't any new changes to `Screen.zig`

mitchellh (approved):

PR #5075, opened 2025-01-14
bash: less intrusive automatic shell integration

jparise:
We now use a temporary function (__ghostty_bash_startup) to perform the
bash startup sequence. This gives us a local function scope in which to
store some temporary values (like rcfile). This way, they won't leak
into the sourced files' scopes.

Also, use `~/` instead of `$HOME` for home directory paths as a simpler
shorthand notation.

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -20,14 +20,13 @@
 if [[ "$-" != *i* ]] ; then builtin return; fi
 if [ -z "$GHOSTTY_RESOURCES_DIR" ]; then builtin return; fi
 
-# When automatic shell integration is active, we need to manually
-# load the normal bash startup files based on the injected state.
+# When automatic shell integration is active, we were started in POSIX
+# mode and need to manually recreate the bash startup sequence.
 if [ -n "$GHOSTTY_BASH_INJECT" ]; then
-  builtin declare ghostty_bash_inject="$GHOSTTY_BASH_INJECT"
-  builtin unset GHOSTTY_BASH_INJECT ENV
-
-  # At this point, we're in POSIX mode and rely on the injected
-  # flags to guide is through the rest of the startup sequence.
+  # Store a temporary copy of our startup flags and unset these global
+  # environment variables so we can safely handle reentrancy.
+  builtin declare __ghostty_bash_flags="$GHOSTTY_BASH_INJECT"
+  builtin unset ENV GHOSTTY_BASH_INJECT
 
   # Restore bash's default 'posix' behavior. Also reset 'inherit_errexit',
   # which doesn't happen as part of the 'posix' reset.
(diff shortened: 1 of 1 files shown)


src/shell-integration/bash/ghostty.bash
```diff
@@ -20,14 +20,13 @@
...
-# load the normal bash startup files based on the injected state.
-if [ -n "$GHOSTTY_BASH_INJECT" ]; then
-  builtin declare ghostty_bash_inject="$GHOSTTY_BASH_INJECT"
-  builtin unset GHOSTTY_BASH_INJECT ENV
-
-  # At this point, we're in POSIX mode and rely on the injected
-  # flags to guide is through the rest of the startup sequence.
+# When automatic shell integration is active, we were started in POSIX
+# mode and need to manually recreate the bash startup sequence.
+if [ -z "${GHOSTTY_BASH_STARTUP:+GHOSTTY_BASH_STARTUP}" ]; then
```
anund, line 25: If the value _is_ set replace it with the literal string "GHOSTTY_BASH_STARTUP"? I'm not sure what this particular expansion is helping with.

jparise: > `${parameter:+word}`
> If parameter is null or unset, nothing is substituted, otherwise the expansion of word is substituted.

This `-z` test will succeed when `$GHOSTTY_BASH_STARTUP` is set to any string (even empty). It will fail (and not run the startup sequence) when `$GHOSTTY_BASH_STARTUP` is missing/null/unset. The literal string `GHOSTTY_BASH_STARTUP` is arbitrary here; it could be anything.

This lets us remove the `1` sentinel value from the environment variable, which was previously needed to pass the `-n` (non-empty) test. We could go back to that approach, too; I don't have strong feelings either way. The minor advantage to what I have here is that `$GHOSTTY_BASH_STARTUP` _only_ contains startup flags.

anund: More specifically, isn't `[ -z ${var} ]` equivalent to `[ -z "${var:+var}" ]`. If the value is set replacing it doesn't change the result of the test unless I'm missing something.

jparise: I'm going to take another pass it this to avoid it being too confusing.

jparise: I decided to drop the `GHOSTTY_BASH_INJECT` changes from this pull request. It now just focuses on the scope-related change.

JacobCallahan:
Hi @jparise, 
     I just updated my installation of ghostty today and noticed that the startup (including new tab) behavior has changed, specifically not correctly loading my `.bashrc`. In the picture below, you can see the initial (default) prompt followed by the expected prompt after re-sourcing my .bashrc.

Screenshot From 2025-01-17 15-32-46

I don't want to jump the gun with a premature issue, but if you have alternative ideas, then I'd be interested in hearing them. For the record, this worked before today's update and also works in Gnome's default terminal as well as ptyxis.

Thanks

jparise:
> Hi @jparise, I just updated my installation of ghostty today and noticed that the startup (including new tab) behavior has changed, specifically not correctly loading my `.bashrc`. In the picture below, you can see the initial (default) prompt followed by the expected prompt after re-sourcing my .bashrc.

You're the second person to report this (#5169), so something definitely seems amiss. I haven't been able to reproduce the problem, though, so we might need to just revert this change and revisit it some other time. I wish I knew what was causing the issue though. ☹️

JacobCallahan:
@jparise ahh thanks for the reference. I really like what you're doing for us bash users!

src/shell-integration/bash/ghostty.bash
```diff
@@ -40,35 +39,40 @@ if [ -n "$GHOSTTY_BASH_INJECT" ]; then
...
+        # The location of the system bashrc is determined at bash build
+        # time via -DSYS_BASHRC and can therefore vary across distros:
+        #  Arch, Debian, Ubuntu use /etc/bash.bashrc
+        #  Fedora uses /etc/bashrc sourced from ~/.bashrc instead of SYS_BASHRC
+        #  Void Linux uses /etc/bash/bashrc
+        #  Nixos uses /etc/bashrc
+        for rcfile in /etc/bash.bashrc /etc/bash/bashrc /etc/bashrc; do
+          [ -r "$rcfile" ] && { builtin source "$rcfile"; break; }
+        done
+        if [[ -z "$GHOSTTY_BASH_RCFILE" ]]; then GHOSTTY_BASH_RCFILE=~/.bashrc; fi
```
keszybz, line 65: I have doubts about the expansion of the tilde here. Bash does expand ~ in the middle, but pure sh does not:
```console
[zbyszek@x1c ghostty]$ sh -c 'echo VAR=~'
VAR=~
[zbyszek@x1c ghostty]$ bash -c 'echo VAR=~'
VAR=/home/zbyszek
```
It'd be good to verify that it is expanded as expected.

leepa:
Add me to the list of people impacted by this (on Fedora). I've locally reverted this PR so I can still work.

leepa:
> > Hi @jparise, I just updated my installation of ghostty today and noticed that the startup (including new tab) behavior has changed, specifically not correctly loading my `.bashrc`. In the picture below, you can see the initial (default) prompt followed by the expected prompt after re-sourcing my .bashrc.
> 
> You're the second person to report this (#5169), so something definitely seems amiss. I haven't been able to reproduce the problem, though, so we might need to just revert this change and revisit it some other time. I wish I knew what was causing the issue though. ☹️

Although you can't pinpoint the issue, I would propose reverting this change before it makes its way into a release.

jparise:
> Although you can't pinpoint the issue, I would propose reverting this change before it makes its way into a release.

#5196

jparise:
#5249 reverts the majority of this change. I believe the root cause of most people's issues was the change to function-level scope, which was a flawed attempt to provide a more hermetic environment in which to source the startup files.

PR #5077, opened 2025-01-14
fix: building on systems with older adwaita

prplz:
3cdb9a7 made ghostty unable to compile on my debian system (adwaita 1.2.2) due to `adw_tab_overview_set_show_start_title_buttons` and `adw_tab_overview_set_show_end_title_buttons` which are since 1.3.0. I increased the comptime check to 1.4.0, this is fine since the assignment of `tab_overview` is [already comptime guarded by 1.4.0](https://github.com/ghostty-org/ghostty/blob/3cdb9a7dfe59fa46dde27b5ada5932bbe42094b8/src/apprt/gtk/Window.zig#L130)


mitchellh:
Thanks! This is something I'm also working on now...

PR #5083, opened 2025-01-15
fix(macos): prevent transparency leakage/flash in new/resized surfaces

### LINKED ISSUE #4516: macOS: background bleeds/flicker when resizing splits, creating tabs

qwerasd205:
Fixes #4516

By using the `CAMetalLayer`'s `backgroundColor` property instead of drawing the background color in our shader, it is always stretched to cover the full surface, even when live-resizing, and it doesn't require us to draw a frame for it to be initialized so there's no transparent flash when a new surface is created (as in a new split/tab).

This commit also allows for hot reload of `background-opacity`, `window-vsync`, and `window-colorspace`.

### DIFF
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -148,6 +148,9 @@ layer: objc.Object, // CAMetalLayer
 /// a display link.
 display_link: ?DisplayLink = null,
 
+/// The `CGColorSpace` that represents our current terminal color space
+terminal_colorspace: *graphics.ColorSpace,
+
 /// Custom shader state. This is only set if we have custom shaders.
 custom_shader_state: ?CustomShaderState = null,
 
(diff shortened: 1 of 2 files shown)


mitchellh:
Looks great. Thank you!

edrozenberg:
I built`ff9414d9` on MacOS, can confirm all flashing is gone, no flashing at all creating/closing windows/tabs, no flashing resizing splits. A fantastic QOL (Quality of Legendary Status) improvement. Thank you! 🌟 🔭 👻

Issue #5047, opened 2025-01-14
Add `toggle_background_opacity` keybind action


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4664

<div type='discussions-op-text'>

<sup>Originally posted by **yanskun** January  5, 2025</sup>
I like to use the terminal with a transparent background.
However, when sharing a screen for pair programming, etc., it seems that noise can be introduced depending on the other person, so I would like to be able to easily turn off transparency.

iTerm has a Toggle Transparency feature that allows you to switch between custom opacity and non-transparent by pressing `cmd + u`.
https://stackoverflow.com/questions/34159559/toggle-opacity-with-hotkey-in-terminal-iterm#:~:text=cmd%20%2B%20u%20toggles%20the%20transparency.

Does Ghostty have it?

Since Alacritty doesn't seem to have this feature, I prepared a shell script to change the config directly.

If Ghostty doesn't have this feature, I'll prepare a Shell Script.
However, if Ghostty has it, I'd like to use it.
(If I get excited, I'd like to implement it directly in Ghostty and submit a PR.)</div>

### Proposal

Add a new keybind action `toggle_background_opacity`. This does nothing if `background-opacity` is 1+. 

If `background-opacity` is less than 1, then this will either set the window transparent or not depending on its current state.

When reloading the config, it is acceptable that background opacity is reapplied. This will make implementing this easier.

This should be implemented for macOS, X11, and Wayland.

imkgh:
It would be even better if the transparency adjustment could be triggered with a shortcut, because sometimes when working on a small screen, I want to see what's behind my terminal.

yujinqiu:
Please add it, I really want it from my muscle memory :-)

00-kat:
…the entire point of this Issue is to track its addition.  Subscribe if you want to know when it's done (you can even choose to subscribe to specifically when it is closed; it's under “Notifications” on the sidebar → “Customize” → “Custom”).

> +1

Use the 👍️ button, you're just sending all subscribers an email and using up space; the 👍️ button also makes it much easier to gauge interest (it has a number!!).

![A screenshot of hovering over the 👍️ button on the Issue starter, with multiple crude arrows pointed at it.](https://github.com/user-attachments/assets/f3893084-2bb7-4642-b910-830032f431e3)

ShunsukeMatsuno:
[...]

PR #5086, opened 2025-01-15
Add toggle_background_opacity option

aw1875:
Linux implementation for https://github.com/ghostty-org/ghostty/issues/5047 (mentioning it rather than considering it closed due to the fact that I don't have access to a Mac currently to implement and test that side of this card)


AlexJuca:
@aw1875 If you feel comfortable I could implement the macOS functionality that builds on your changes.

aw1875:
@AlexJuca, yeah that'd be great if you wanted to do that. I'm pretty sure I know how it should look so you can definitely work on that!

AlexJuca:
> @AlexJuca, yeah that'd be great if you wanted to do that. I'm pretty sure I know how it should look so you can definitely work on that!

Awesome, I'll work on this tomorrow.

src/apprt/gtk/Window.zig
```diff
@@ -559,6 +559,17 @@ pub fn toggleWindowDecorations(self: *Window) void {
...
+pub fn toggleBackgroundOpacity(self: *Window) void {
+    if (self.app.config.@"background-opacity" >= 1) return;
+
+    if (c.gtk_widget_has_css_class(@ptrCast(self.window), "background") == 1) {
+        c.gtk_widget_remove_css_class(@ptrCast(self.window), "background");
+    } else {
+        c.gtk_widget_add_css_class(@ptrCast(self.window), "background");
+    }
+}
+
```
pluiedev, line 572: This doesn't actually update the config for the OpenGL renderer, so the actual terminal screen will still render as if they're transparent (and likely have wrong colors)

pluiedev (changes requested):

src/apprt/gtk/Window.zig
aw1875: Good point, didn't even think about that. Let me take a look at that as well

jcollie:
@aw1875 is this something that you plan on updating? There has been a lot of code changes recently so this needs a rebase plus addressing the rest of the feedback that you've gotten so far.

aw1875:
> @aw1875 is this something that you plan on updating? There has been a lot of code changes recently so this needs a rebase plus addressing the rest of the feedback that you've gotten so far.

I've been so busy I haven't had time to take a look at the codebase in a while. I'll try to see if I can put some time aside tomorrow to re-familiarize myself with everything and look into fixing this up/updating things. If I can't find the time I'll probably close this pr out and come back to it when I do have more time if nobody has done it

aw1875:
Rebased and updated the code to fix the issue with not actually notifying OpenGL. It might make more sense to move the check for the value of `background-opacity` in the config to the App level so we can exit early rather running through the whole sync process when the value is already one. If I were to make this change, this is what I had in mind:

```diff
diff --git a/src/apprt/gtk/App.zig b/src/apprt/gtk/App.zig
index 22c3064f..a7a5bdfe 100644
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -772,6 +772,8 @@ fn toggleBackgroundOpacity(
     switch (target) {
         .app => {},
         .surface => |v| {
+            if (self.config.@"background-opacity" >= 1) return;
+
             const window = v.rt_surface.container.window() orelse {
                 log.info(
                     "toggleBackgroundOpacity invalid for container={s}",

diff --git a/src/apprt/gtk/Window.zig b/src/apprt/gtk/Window.zig
index 9b1ba4c3..6ad094de 100644
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -643,8 +643,6 @@ pub fn toggleWindowDecorations(self: *Window) void {

 /// Toggle the background opacity for this window.
 pub fn toggleBackgroundOpacity(self: *Window) !f64 {
-    if (self.app.config.@"background-opacity" >= 1) return self.app.config.@"background-opacity";
-
     self.config.background_opacity = if (self.config.background_opacity == 1.0) self.app.config.@"background-opacity" else 1.0;
     try self.syncAppearance();
```

src/apprt/gtk/Window.zig
```diff
@@ -457,12 +457,6 @@ pub fn syncAppearance(self: *Window) !void {
     self.winproto.syncAppearance() catch |err| {
         log.warn("failed to sync winproto appearance error={}", .{err});
     };
-
-    toggleCssClass(
-        @ptrCast(self.window),
-        "background",
-        self.config.background_opacity >= 1,
-    );
```
aw1875, line 465: Just as a note to reviewers, this was a duplicate call to the same toggle we do on line 422

src/apprt/gtk/App.zig
```diff
@@ -796,6 +797,31 @@ fn toggleQuickTerminal(self: *App) !bool {
...
+                );
+                return;
+            };
+
+            const opacity = try window.toggleBackgroundOpacity();
+
+            // Cycle through all surfaces and set the background opacity so they stay in sync
+            for (self.core_app.surfaces.items) |surface| {
+                surface.core_surface.setBackgroundOpacity(opacity);
+            }
```
pluiedev, line 820: Wait what? That's not right. It should only affect the core surface of the current apprt surface

src/apprt/gtk/Window.zig
```diff
@@ -661,6 +661,16 @@ pub fn toggleWindowDecorations(self: *Window) void {
     };
 }
 
+/// Toggle the background opacity for this window.
+pub fn toggleBackgroundOpacity(self: *Window) !f64 {
+    if (self.app.config.@"background-opacity" >= 1) return self.app.config.@"background-opacity";
```
pluiedev, line 666: What's this for?

src/apprt/gtk/Window.zig
```diff
@@ -661,6 +661,16 @@ pub fn toggleWindowDecorations(self: *Window) void {
...
+
+    self.config.background_opacity = if (self.config.background_opacity == 1.0) self.app.config.@"background-opacity" else 1.0;
```
pluiedev, line 668: Doing == on a float isn't a good idea. Use the `a - b < epsilon` method.

src/renderer/OpenGL.zig
```diff
@@ -653,6 +653,17 @@ pub fn setVisible(self: *OpenGL, visible: bool) void {
...
 
+/// Set the backgrond opacity in the config.
+///
+/// Must be called on the render thread.
+pub fn setBackgroundOpacity(self: *OpenGL, opacity: f64) void {
+    if (single_threaded_draw) self.draw_mutex.lock();
+    defer if (single_threaded_draw) self.draw_mutex.unlock();
+
+    // We don't want to go below 0 or above 1
+    self.config.background_opacity = @max(0, @min(1, opacity));
```
pluiedev, line 664: std.math.clamp

pluiedev (changes requested):

src/apprt/gtk/App.zig
aw1875: Wouldn't we want these changes reflected across the app? If I have several tabs open and I toggle the opacity wouldn't the expected behavior be to toggle it for them all?

pluiedev: Then this should be an app-level action instead of a surface-level one

src/apprt/gtk/Window.zig
pluiedev: Well if it is global state then it shouldn't be stored per-window. Should be part of the App's DerivedConfig instead.

aw1875: okay cool that sounds good. I wasn't sure if I should be modifying that for this logic but that makes sense, I can update that

src/apprt/gtk/App.zig
aw1875: would we still want to loop through the surfaces then? Or are you suggesting avoiding that?

src/apprt/gtk/Window.zig
aw1875: I might just be blind but I don't see a Derived Config at the app level, only the package config that you can set from modifying your ghostty config

src/apprt/gtk/App.zig
jcollie: Do we, in fact, want `toggle_background_opacity` to be an app-level toggle, rather than a per surface? If we had a per surface toggle, you can always make it an "app" toggle by prefixing the keybind will `all:` but there's no way to do the opposite.

jcollie: Or, must this affect all surfaces in a window?

pluiedev: I mean it makes the most sense for me for this to be a "window-level" action, but since we've rejected (I think?) that for core that means we have to stick with either app or surface level for now.

jcollie: yeah i think it makes most sense to have the toggle affect only one surface. if users want different they can add `all:` to their keybind.

aw1875: Okay, so in the case of the `all` flag, we'd iterate through the surfaces as we do now; otherwise, we'd change the current surface?

jcollie: Having the `all:` flag should mean the the Ghostty core will iterate though all of the surfaces for you, we only need to handle the targeted surface.

aw1875: One caveat I noticed here: since the changes are only handled at a surface level, it's pretty easy to get a mismatch of tabs. For example, if I have 2 tabs open and I use the toggle with the `all` flag, it will toggle those two open tabs to 100 opacity. Then, if I open a new tab, it will be at my default opacity and will work opposite to the two other tabs since the state is part of the derived config at the surface level.

src/apprt/gtk/App.zig
```diff
@@ -52,6 +52,9 @@ pub const Options = struct {};
 core_app: *CoreApp,
 config: Config,
 
+/// Derived configuration values that are computed from the main. Temporarily prefixed with `_` per https://discord.com/channels/1005603569187160125/1349556080090288232/1349569138846203955
+_config: DerivedConfig,
```
pluiedev, line 56: I'm not sure about the _ prefix - it usually indicates internal data other code shouldn't use, but we do intend to switch everything over to use derived config in the future. `derived_config` perhaps?

pluiedev (changes requested):

aw1875:
Been a little too busy to keep catching up on rebasing, plus still waiting to hear back on some implementation details in [this discussion](https://github.com/ghostty-org/ghostty/discussions/6686), so going to close this pr out for the time being.

PR #5090, opened 2025-01-15
fix: gtk titlebar being restored if it shouldn't be

### LINKED ISSUE #5262: GTK: Titlebar Reappears When Maximizing and Minimizing Ghostty Window to Normal
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5137

<sup>Originally posted by **dagimg-dot** January 16, 2025</sup>
Even with `gtk-titlebar = false` set in the Ghostty configuration file, the title bar reappears when the Ghostty window is maximized and then minimized to normal window. This behavior persists despite the explicit setting to disable the title bar.

**Environment:**
- **Operating System:** Fedora 41
- **Desktop Environment:** GNOME 47.2

**Ghostty Information:**
- **Version:** 1.0.2-main+ff9414d9
- **Build Channel:** tip
- **Build Config:**
  - **Zig Version:** 0.13.0
  - **Build Mode:** ReleaseFast
  - **App Runtime:** GTK
  - **Font Engine:** Fontconfig + Freetype
  - **Renderer:** OpenGL
  - **Libxev Backend:** io_uring
  - **Desktop Environment:** GNOME
  - **GTK Version:**
    - **Build:** 4.16.5
    - **Runtime:** 4.16.5
  - **Libadwaita:** Enabled
    - **Build:** 1.6.2
    - **Runtime:** 1.6.2
  - **LibX11:** Enabled
  - **Libwayland:** Enabled

**Steps to Reproduce:**
1. Set `gtk-titlebar = false` in the Ghostty configuration file.
2. Launch Ghostty.
3. Maximize the Ghostty window.
4. Minimize the Ghostty window.
5. Observe that the title bar reappears even though it was initially disabled.

**Expected Behavior:**
The title bar should remain hidden at all times, regardless of window state (maximized, minimized, or normal).

**Actual Behavior:**
The title bar reappears after maximizing and then minimizing the Ghostty window, contrary to the configuration setting.</div>

## Mitchell's Notes

Note: this is unconfirmed on my end. But I know we recently implemented the maximized/minimized header hiding code and so I can easily see how we weren't properly checking this config on the restoration.

If this ends up being a separate issue we can close this, move to discussion, and go from there.

aw1875:
I feel like we should be respecting the value of `gtk-titlebar` in the `gtkWindowNotifyMaximized` callback as mentioned in discord (https://discord.com/channels/1005603569187160125/1005603569711452192/1328976978094723163)

Would close https://github.com/ghostty-org/ghostty/issues/5262

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -675,7 +675,13 @@ fn gtkWindowNotifyFullscreened(
     ud: ?*anyopaque,
 ) callconv(.C) void {
     const self = userdataSelf(ud orelse return);
-    self.headerbar.setVisible(c.gtk_window_is_fullscreen(@ptrCast(object)) == 0);
+    const fullscreened = c.gtk_window_is_fullscreen(@ptrCast(object)) != 0;
+    if (!fullscreened) {
+        self.headerbar.setVisible(self.app.config.@"gtk-titlebar");
+        return;
+    }
+
+    self.headerbar.setVisible(false);
 }
 
 // Note: we MUST NOT use the GtkButton parameter because gtkActionNewTab
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -646,7 +646,7 @@ fn gtkWindowNotifyMaximized(
     if (!self.winproto.clientSideDecorationEnabled()) return;
 
     if (!maximized) {
-        self.headerbar.setVisible(true);
+        self.headerbar.setVisible(self.app.config.@"gtk-titlebar");
         return;
     }
     if (self.app.config.@"gtk-titlebar-hide-when-maximized") {
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -639,16 +639,20 @@ fn gtkWindowNotifyMaximized(
     ud: ?*anyopaque,
 ) callconv(.C) void {
     const self = userdataSelf(ud orelse return);
-    const maximized = c.gtk_window_is_maximized(self.window) != 0;
 
     // Only toggle visibility of the header bar when we're using CSDs,
     // and actually intend on displaying the header bar
     if (!self.winproto.clientSideDecorationEnabled()) return;
 
+    // If we aren't maximized, we should show the headerbar again
+    // if it was originally visible.
+    const maximized = c.gtk_window_is_maximized(self.window) != 0;
     if (!maximized) {
         self.headerbar.setVisible(self.app.config.@"gtk-titlebar");
         return;
     }
+
+    // If we are maximized, we should hide the headerbar if requested.
     if (self.app.config.@"gtk-titlebar-hide-when-maximized") {
         self.headerbar.setVisible(false);
     }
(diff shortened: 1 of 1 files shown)


aw1875:
Just a note from the discussion but this PR may not be relevant at this point, but just wanted to get this put together in case it is

aw1875:
Just a note from the discussion that this PR may not be relevant at this point, but I just wanted to get this put together in case it is

rudrabhoj:
Tested these commits on Gentoo with Hyrland. This resolves the problem on the fullscreen mode.

jcollie:
I feel that we should not be changing Ghostty's behavior with respect to `gtk-titlebar` and `window-decoration` until we update the documentation surrounding these configuration entries (perhaps adding/modifying options in the process). In particular here are the last two lines of the `gtk-titlebar` documentation that indicate that `gtk-titlebar` doesn't have any effect after initial window creation. This patch also does not take into account situations when `window-decoration` might affect the display of the titlebar.

Everyone seems to want to modify the behavior to suit _their particular interpretation_ of what `gtk-titlebar`  _should_ mean and how it interacts with `window-decoration`. We need to put more work first in clarifying the documentation so that _everyone_ agrees on the interpretation and it's possible to clearly specify a user's wishes.

For example, there is no documented way currently to turn off window decorations (rounded corners and drop shadows) _and_ keep the GTK titlebar.

https://github.com/ghostty-org/ghostty/blob/ddf7173ae92b4992c4506b6df4681c0687cc29c3/src/config/Config.zig#L2069-L2078

mitchellh:
> I feel that we should not be changing Ghostty's behavior with respect to gtk-titlebar and window-decoration until we update the documentation surrounding these configuration entries (perhaps adding/modifying options in the process).

This PR and the associated issue are not to change behavior. The intention of maximize and minimize is that the window should return back to the _same state it was in_. 

The quoted value you linked in the config also only applies to _config reloading_. It is not apriori behavior that means "once we unset this at runtime, it'll never be set again."

mitchellh:
Tested this locally and it's good. I found another issue where we didn't respect CSD vs SSD on the fullscreen change either so I fixed that up too.

jcollie:
> This PR and the associated issue are not to change behavior. The intention of maximize and minimize is that the window should return back to the _same state it was in_.

Which is not what this PR does. Consider this sequence of events if this PR is merged:

1. Start Ghostty with `gtk-titlebar=true`.
2. Window is created with a titlebar.
3. Maximize window.
4. Window now has no titlebar.
5. Edit the config, change `gtk-titlebar=false`, reload config.
6. Unmaximize the window.
7. Titlebar doesn't reappear.

One could certainly argue persuasively that that's what the user _intended_ when they edited the config file. But it's _not_ the documented behavior ("Changing this value at runtime and reloading the configuration will only affect new windows") and it's not what the state of the window was _before_ maximization.

mitchellh:
The `gtk-titlebar` config right now is runtime reloadable, the docs are now outdated (this PR didn't introduce this). I think it makes sense to make things runtime reloadable if they can, most of the non-runtime reloadable things are simply because we didn't introduce the logic to do it at the time the feature was introduced.

I do agree we need to discuss the semantics of the interaction between gtk-titlebar, window decorations, CSDs/SSDs, etc. but I don't think this issue is trying to address that particular pain point. 

To me, the goal of this is to restore what the visual style should be. Namely, the bug is:

1. Start with `gtk-titlebar = false`
2. Maximize
3. Minimize
4. Titlebar appears

PR #5122, opened 2025-01-15
[macOS] feat: Add setting to hide icon from dock/cmd-tab

### LINKED ISSUE #4538: macOS: Introduce config to hide Ghostty from the dock and cmd-tab
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3810

<sup>Originally posted by **frolvanya** December 28, 2024</sup>
image

I'm using ghostty primarily in quick-terminal mode, so I don't want to see an icon in the dock</div>

### Proposal

Introduce a new configuration `macos-hidden` (boolean). When set to true, the Ghostty app will be hidden. I think concretely, this means turning the app into an "accessory" app:

```swift
NSApp.setActivationPolicy(.accessory)
```

**This has a number of downsides,** so it must be carefully documented. Notably:

- App no longer appears in dock
- App no longer appears in cmd-tab
- Looks like keyboard layout changes are [no longer automatic](https://gitlab.com/gnachman/iterm2/-/issues/10973)

I think as a future improvement we can introduce a menu bar icon that can bring the app to focus when click (optional). But introducing this option as default false will help get some quick-terminal-only users going more happily.

xalbd:
Resolves #4538 

Adds boolean configuration option `macos-hidden` which toggles the NSApp's activation policy appropriately on config change.

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -533,6 +533,15 @@ class AppDelegate: NSObject,
         // AppKit mutex on the appearance.
         DispatchQueue.main.async { self.syncAppearance(config: config) }
 
+        // Decide whether to hide/unhide app from dock and app switcher
+        switch (config.macosHidden) {
+        case .never:
+            NSApp.setActivationPolicy(.regular)
+
+        case .always:
+            NSApp.setActivationPolicy(.accessory)
+        }
+
         // If we have configuration errors, we need to show them.
         let c = ConfigurationErrorsController.sharedInstance
         c.errors = config.errors
--- a/macos/Sources/Ghostty/Ghostty.Config.swift
+++ b/macos/Sources/Ghostty/Ghostty.Config.swift
@@ -302,6 +302,16 @@ extension Ghostty {
             return buffer.map { .init(ghostty: $0) }
         }
 
+        var macosHidden: MacHidden {
+            guard let config = self.config else { return .never }
+            var v: UnsafePointer<Int8>? = nil
+            let key = "macos-hidden"
+            guard ghostty_config_get(config, &v, key, UInt(key.count)) else { return .never }
+            guard let ptr = v else { return .never }
+            let str = String(cString: ptr)
+            return MacHidden(rawValue: str) ?? .never
+        }
+
         var focusFollowsMouse : Bool {
             guard let config = self.config else { return false }
             var v = false;
@@ -514,6 +524,11 @@ extension Ghostty.Config {
         case off
         case check
         case download
+    }
+
+    enum MacHidden : String {
+        case never
+        case always
     }
 
     enum ResizeOverlay : String {
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1974,6 +1974,26 @@ keybind: Keybinds = .{},
 /// find false more visually appealing.
 @"macos-window-shadow": bool = true,
 
+/// If true, the macOS icon in the dock and app switcher will be hidden. This is
+/// mainly intended for those primarily using the quick-terminal mode.
+///
+/// Note that setting this to true means that keyboard layout changes
+/// will no longer be automatic.
+///
+/// Control whether macOS app is excluded from the dock and app switcher,
+/// a "hidden" state. This is mainly intended for those primarily using
+/// quick-terminal mode, but is a general configuration for any use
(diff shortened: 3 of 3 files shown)


macos/Sources/Ghostty/Ghostty.Config.swift
```diff
@@ -300,6 +300,14 @@ extension Ghostty {
             return buffer.map { .init(ghostty: $0) }
         }
 
+        var macosHidden: Bool {
```
AlexJuca, line 303: @xalbd We probably want to name this `isAppIconHiddenFromMacOS`.

src/config/Config.zig
```diff
@@ -1909,6 +1909,13 @@ keybind: Keybinds = .{},
 /// you may want to disable it.
 @"macos-secure-input-indication": bool = true,
 
+/// If true, the macOS icon in the dock and cmd-tab will be hidden. This is
+/// mainly intended for those primarily using the quick-terminal mode.
+///
+/// Note that setting this to true means that keyboard layout changes
+/// will no longer be automatic.
+@"macos-hidden": bool = false,
```
AlexJuca, line 1917: @xalbd Should we name this `is-app-icon-hidden-from-macos`?

AlexJuca (changes requested):

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -523,6 +523,9 @@ class AppDelegate: NSObject,
         // this is called as part of application launch it can deadlock with an internal
         // AppKit mutex on the appearance.
         DispatchQueue.main.async { self.syncAppearance(config: config) }
+        
+        // Decide whether to hide/unhide app from dock and cmd-tab
```
AlexJuca, line 527: @xalbd  // Decide whether to hide/unhide app from dock and app switcher

src/config/Config.zig
```diff
@@ -1909,6 +1909,13 @@ keybind: Keybinds = .{},
 /// you may want to disable it.
 @"macos-secure-input-indication": bool = true,
 
+/// If true, the macOS icon in the dock and cmd-tab will be hidden. This is
```
AlexJuca, line 1912: /// If true, the macOS icon in the dock and app switcher will be hidden.

AlexJuca (changes requested):

src/config/Config.zig
xalbd: What about `macos-icon-hide` to keep naming consistent with other `macos-` options?

AlexJuca: Those are also good options.

macos/Sources/App/macOS/AppDelegate.swift
AlexJuca: @xalbd Also, why do you call it cmd tab and not its official name "app switcher"?

AlexJuca (approved):
This looks good. Thanks

mitchellh:
I changed the config to an enum so we can easily accommodate the `quick-terminal` value in the future (as noted in #4538). Otherwise, looks great. Running CI then will merge.

mitchellh (approved):

bolechen:
sorry, how can i use this feta at macOS?
I set `macos-hidden=true` it say 

```
com.mitchellh.ghostty/config:55:macos-hidden: unknown field
```

00-kat:
PRs aren't for support, you should open a Discussion next time.

This one is easy though: this is only be available in the next version, so if you want it now you should switch to tip with [`auto-update-channel`] or some other means[^1].

[`auto-update-channel`]: https://ghostty.org/docs/config/reference#auto-update-channel

[^1]: Such as installing `ghostty@tip` if you use Homebrew.

bolechen:
> PRs aren't for support, you should open a Discussion next time.
> 
> This one is easy though: this is only be available in the next version, so if you want it now you should switch to tip with [`auto-update-channel`](https://ghostty.org/docs/config/reference#auto-update-channel) or some other means[1](#user-content-fn-1-a47eb474357eb0f3c1467bc4841289df).
> 
> ## Footnotes
> 1. Such as installing `ghostty@tip` if you use Homebrew. [↩](#user-content-fnref-1-a47eb474357eb0f3c1467bc4841289df)

I appreciate you taking the time to explain the solution, and I hope this thread will be useful for others who might run into the same issue while searching. Thanks again!

PR #5123, opened 2025-01-15
core/gtk/opengl: report the cursor position to the apprt

jcollie:
Not sure if this is the best/right way to do this. Wasn't sure if the de-duplication should happen in the renderer or the surface thread. Or maybe there's a better way to do this that I missed because I don't know this part of the code very well.


tristan957 (dismissed):

mitchellh (changes requested):
I'd do the de-duping within the renderer. The way this is written today will result in a lot of wasted CPU cycles across threads since every single frame would result in a mailbox message and subsequent thread wakeup (which is likely on another core).

I'd also like to pair this with an actual use case, because I'm not convinced an apprt action and evented system is even necessary or correct for this. It's all just potentially very noisy and I'm concerned about power/cpu usage with all of this.

I think your designated use case is to move the URL overlay when the screen cursor is around it, right? Let me think about that...

jcollie:
Closing this as there doesn't seem to be much of a need for this at this time.

tristan957:
I still think it would be good to do this so we can show the url overlay on the other side of the terminal if the cursor is in the way

mitchellh:
I think eventually yes but I wasn't happy enough with this approach to merge this. I want to think more critically about how to surface this information.

PR #5124, opened 2025-01-15
gtk(wayland): respect compositor SSD preferences

pluiedev:
Compositors can actually tell us whether they want to use CSD or SSD!

(Ignore the context menu changes — they will most likely be unified after #4952 anyway)


src/apprt/gtk/winproto/wayland.zig
```diff
@@ -57,6 +59,12 @@ pub const App = struct {
         registry.setListener(*Context, registryListener, context);
         if (display.roundtrip() != .SUCCESS) return error.RoundtripFailed;
 
+        if (context.kde_decoration_manager != null) {
+            // FIXME: Roundtrip again because we have to wait for the decoration
```
mitchellh, line 63: This is fine for now. Ideally we can add the fd to the GTK event loop so the event loop wakes up when we have data to process.

mitchellh:
Oops I forgot to fix macOS. Doing this now.

PR #5125, opened 2025-01-15
config: add appstream spec for linux

psadi:
This PR adds the `ghostty.metainfo.xml` file for linux, the spec provides metadata and essential information (name, description, categories, screenshots) to [AppStream](https://www.freedesktop.org/wiki/Distributions/AppStream) systems, ensuring that Ghostty can be properly recognized and presented in software catalogs.

With this metadata, users can find more relevant details about the application before installation, improving discoverability and user trust.


tristan957:
This is a great idea. We can also make use of this in the libadwaita side of things: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/ctor.AboutDialog.new_from_appdata.html

dist/linux/ghostty.appdata.xml
```diff
@@ -0,0 +1,70 @@
...
+  <description>
+    <p>
+      Ghostty is a terminal emulator that differentiates itself by being fast,
+      feature-rich, and native. While there are many excellent terminal
+      emulators available, they all force you to choose between speed,
+      features, or native UIs.  Ghostty provides all three.
+    </p>
+
+    <p>
+      In all categories, I am not trying to claim that Ghostty is the best
```
jcollie, line 19: I realize that this is mostly copied from Ghostty's own documentation, but I find the first-person point of view here odd.

jcollie:
> This is a great idea. We can also make use of this in the libadwaita side of things: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/ctor.AboutDialog.new_from_appdata.html

Yes, it would be good to have a single "source of truth" for this information.

dist/linux/ghostty.appdata.xml
psadi: Yeah agreed, will change this to keep it more generic.

dist/linux/ghostty.appdata.xml
```diff
@@ -0,0 +1,54 @@
...
+  <releases>
+    <release version="" date="">
+      <description>
+        <p>
+          release notes
+        </p>
+      </description>
+    </release>
+  </releases>
+-->
```
pluiedev, line 53: I wonder if we should generate releases with code instead... Sounds kinda tedious to do this by hand

jcollie: Including the actual release notes creates a catch 22 situation. Would certainly be easy to add a link to the release notes on the website, since that's easily predictable.

pluiedev: Yeah no adding the full release notes is overkill. Just direct people to the website

psadi: agreed, will add the release notes URL in the description. Please confirm if it is possible to update the release version along with the build timestamp for attribute ` <release version="" date="">` ?

appsstream spec mandates this values if release attribute enabled.

AnthonyZhOon:
I believe the file extension should be `.metainfo.xml` as `.appdata.xml` is listed as a legacy format [AppStream Specification](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html)

I see a few more fields we can provide

```
<provides>
  <binary>ghostty</binary>
</provides>
```

[Recommends-Control](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-control)
Keyboard and mouse seem reasonable
We can use similar fields as the .desktop file https://github.com/ghostty-org/ghostty/blob/main/dist/linux/app.desktop
[Categories](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-categories)
Our .desktop file which has `Categories=System;TerminalEmulator;`
[Keywords](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-keywords)
Also shared with the .desktop

psadi:
> I believe the file extension should be `.metainfo.xml` as `.appdata.xml` is listed as a legacy format [AppStream Specification](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html)
> 
> I see a few more fields we can provide
> 
> ```
> <provides>
>   <binary>ghostty</binary>
> </provides>
> ```
> 
> [Recommends-Control](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-control) Keyboard and mouse seem reasonable We can use similar fields as the .desktop file https://github.com/ghostty-org/ghostty/blob/main/dist/linux/app.desktop [Categories](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-categories) Our .desktop file which has `Categories=System;TerminalEmulator;` [Keywords](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-keywords) Also shared with the .desktop

renamed to metainfo, added additional attributes. appstream-ci lint passes with the config at its current state.

tristan957:
> > I believe the file extension should be `.metainfo.xml` as `.appdata.xml` is listed as a legacy format [AppStream Specification](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html)
> > I see a few more fields we can provide
> > ```
> > <provides>
> >   <binary>ghostty</binary>
> > </provides>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [Recommends-Control](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-control) Keyboard and mouse seem reasonable We can use similar fields as the .desktop file https://github.com/ghostty-org/ghostty/blob/main/dist/linux/app.desktop [Categories](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-categories) Our .desktop file which has `Categories=System;TerminalEmulator;` [Keywords](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-keywords) Also shared with the .desktop
> 
> renamed to metainfo, added additional attributes. appstream-ci lint passes with the config at its current state.

can you add a gh workflow so we never fail the lint?

src/build/GhosttyResources.zig
```diff
@@ -199,7 +199,13 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyResources {
...
         try steps.append(&b.addInstallFile(
-            b.path("dist/linux/ghostty.appdata.xml"),
+            b.path("dist/linux/ghostty.metainfo.xml"),
+            "share/metainfo/com.mitchellh.ghostty.metainfo.xml",
+        ).step);
+
+        // for legacy systems
+        try steps.append(&b.addInstallFile(
+            b.path("dist/linux/ghostty.metainfo.xml"),
             "share/metainfo/com.mitchellh.ghostty.appdata.xml",
```
jcollie, line 209: If `.metainfo.xml` is the standard, is there a reason to keep `.appdata.xml` around? Is there any significant software left that supports the old extension, but not the new one?

psadi: I currently maintain ghosty AppImage and the appimagetool expects the metadata at `usr/share/metainfo/AppName.appdata.xml`

The support for `metainfo.xml` is widely discussed but is yet to be supported. Hence figured to keep both the specs so the relevant systems can pick one of these as required.

joebonrichie:
`appstream-builder` from https://github.com/hughsie/appstream-glib currently fails to parse this due:

```xml
<icon type="remote">https://raw.githubusercontent.com/ghostty-org/ghostty/refs/heads/main/images/icons/icon_256.png</icon>
```

As `appstream-builder` does not support remote icons. `appstream-glib` is in maintenance mode but is still used by Fedora, Solus and other distributions AFAIK. Removing this line allows generation to succeed. 

It also fails as both a `.metainfo.xml` and a `.appdata.xml` are provided causing a duplicate of error.

psadi:
> `appstream-builder` from https://github.com/hughsie/appstream-glib currently fails to parse this due:
> 
> ```xml
> <icon type="remote">https://raw.githubusercontent.com/ghostty-org/ghostty/refs/heads/main/images/icons/icon_256.png</icon>
> ```
> 

Thanks for letting know.

> As `appstream-builder` does not support remote icons. `appstream-glib` is in maintenance mode but is still used by Fedora, Solus and other distributions AFAIK. Removing this line allows generation to succeed. 
> 
> It also fails as both a `.metainfo.xml` and a `.appdata.xml` are provided causing a duplicate of error.

Yeah there is still some work to be done here, the failure should be mostly to release tag not having version and the timestamp attributes which needs to be parsed and updated during the build process.

PR #5162, opened 2025-01-16
core: add context menu key

jcollie:
I have this key on a cheap Lenovo keyboard at work and would like to put it to some use.

### DIFF
--- a/src/input/key.zig
+++ b/src/input/key.zig
@@ -401,7 +401,8 @@ pub const Key = enum(c_int) {
     kp_delete,
     kp_begin,
 
-    // TODO: media keys
+    // special keys
+    context_menu,
 
     // modifiers
     left_shift,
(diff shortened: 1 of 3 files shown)


src/input/key.zig
```diff
@@ -330,6 +330,7 @@ pub const Key = enum(c_int) {
     backspace,
     print_screen,
     pause,
+    context_menu,
```
mitchellh, line 333: This should go in the "function key" section.

mitchellh (changes requested):
I also believe that any changes to the Key enum require changes to libghostty's header and possibly some Swift. I can check the Swift but definitely check the header please.

jcollie:
> I also believe that any changes to the Key enum require changes to libghostty's header and possibly some Swift. I can check the Swift but definitely check the header please.

I didn't touch the swift code, as I'm not exactly sure what the values in `Ghostty.Input.swift` are supposed to correspond to.

mitchellh (approved):
I have a better understanding of our keyboard key future and this fits in. 

(https://www.w3.org/TR/uievents-code/)

PR #5164, opened 2025-01-16
[macOS] fix: Closing tab with running process while in Stage Manager no longer causes focus to be lost

### LINKED ISSUE #5108: MacOS: Closing tab causes window to lose focus when Stage Manager is enabled.
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4036

<sup>Originally posted by **kenvunz** December 30, 2024</sup>
When Stage Manager (SM) is enabled, closing one tab will lose the entire app focus, rather stay focused on the remaining tab.

In this particular example, I have 2 windows, one Chrome and one Ghostty, closing one tab and the app will immediately losing the focus.

CleanShot 2024-12-30 at 11 33 51

</div>

xalbd:
Resolves #5108 

Reproduction: While Stage Manager is activated, create a new tab in Ghostty, run `sleep 500` in that tab, and try and close it. Make sure to try both closing the tab with cmd-w and clicking the x; also try accepting the modal with a mouse click and pressing enter. 

Relevant AppKit documentation: [NSAlert beginSheetModal](https://developer.apple.com/documentation/appkit/nsalert/beginsheetmodal(for:completionhandler:))

What I believe is happening is that the alert modal ("there is still a process running") is closed automatically after the completion handler is run, which closes the window on which the alert is attached. Something seems to go wrong in Stage Manager at this point; perhaps it detects the alert closing as a "full" window closing and moves focus to a different window. 

The fix I've identified is to manually call the `orderOut` function to dismiss the alert _before_ the code that closes the window is run.

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -449,6 +449,7 @@ class BaseTerminalController: NSWindowController,
         alert.addButton(withTitle: "Cancel")
         alert.alertStyle = .warning
         alert.beginSheetModal(for: window, completionHandler: { response in
+            self.alert = nil
             switch (response) {
             case .alertFirstButtonReturn:
                 alert.window.orderOut(nil)
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -449,9 +449,9 @@ class BaseTerminalController: NSWindowController,
         alert.addButton(withTitle: "Cancel")
         alert.alertStyle = .warning
         alert.beginSheetModal(for: window, completionHandler: { response in
-            self.alert = nil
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -449,9 +449,9 @@ class BaseTerminalController: NSWindowController,
         alert.addButton(withTitle: "Cancel")
         alert.alertStyle = .warning
         alert.beginSheetModal(for: window, completionHandler: { response in
-            self.alert = nil
```
mitchellh, line 452: Doesn't this at least have to be set in the `default` case? Otherwise another modal will never be shown if you reject it, right?

xalbd: Yeah, you're right; honestly not quite sure why I removed that line

PR #5170, opened 2025-01-17
Add `toggle_background_opacity` action for macOS

### LINKED ISSUE #5047: Add `toggle_background_opacity` keybind action

liby:
## Changes

- Added `ghosttyBackgroundOpacityDidToggle` notification for handling opacity toggle events

- Implemented opacity toggle logic in `SurfaceView_AppKit.swift`

- If `background-opacity` is 1 or greater, do nothing

## Testing

1. Set `background-opacity` < 1 in config

2. Bind a key to `toggle_background_opacity` action

3. Press the key to toggle between transparent and opaque states

4. Verify that:

   - Window properly toggles between transparent and opaque states
   
   - Background color is correct in both states
   
   - Action does nothing when `background-opacity` is 1 or more 
   
   - Action does nothing in fullscreen mode

https://github.com/user-attachments/assets/04f114e0-4035-4b8c-860f-71ef696f2258

Resolves #5047


mitchellh (changes requested):
Sorry for the very long delay in review. I can see this is a highly requested feature based on the upvotes. I think the approach here works but not where I'd like to see this done. I think this should be done at the window level (our `NSWindow` subclasses). And this needs to be updated for all the latest changes since. My fault for letting it bitrot. 

If you'd like to reopen this PR with the rebased changes, please do that.

PR #5173, opened 2025-01-17
gtk: always set the title on the underlying window when using adwaita

jcollie:
Fixes #5140


tristan957:
Really interesting that libadwaita doesn't proxy this. Anyway looks like a good fix. Thanks!

tristan957 (approved):

tristan957:
For reference in case anyone else is interested: https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-window-title.c

PR #5189, opened 2025-01-17
fix(wuffs): don't premul alpha when loading png

qwerasd205:
This was causing excessively faint/dark areas anywhere there was transparency, since our image shaders for both Metal and OpenGL assume the image data has straight alpha (which, I think should be the case for data transmitted directly with kitty graphics protocol rather than png-encoded), and do an additional multiplication.

After this change, our image blending looks *like most other apps*, even though it's still technically wrong since it "should" be done in linear space. Kitty, for example, does do linear blending for images, which is *more correct* even though it doesn't look like most apps. On macOS if `text-blending` is set to `linear` we do get linear blending, and match Kitty exactly with this change.

In order to have linear blending of images while maintaining non-linear blending of text for a "native" look on macOS we'd have to use swap textures with different pixel formats (`*_srgb` / not `*_srgb`) or else the API validator yells at us (even though it *does* work correctly...) -- either that or perform our blending in a shader instead of relying on Metal, which is my plan ultimately whenever I get around to making the deferred render pipeline.

### DIFF
--- a/pkg/wuffs/src/jpeg.zig
+++ b/pkg/wuffs/src/jpeg.zig
@@ -55,7 +55,7 @@ pub fn decode(alloc: Allocator, data: []const u8) Error!ImageData {
 
     c.wuffs_base__pixel_config__set(
         &image_config.pixcfg,
-        c.WUFFS_BASE__PIXEL_FORMAT__RGBA_PREMUL,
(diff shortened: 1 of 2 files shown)


qwerasd205:
Good catch, forgot about JPEG

PR #5192, opened 2025-01-17
gtk: don't toggle headerbar on (un)maximize while using SSDs

pluiedev:
See #5137. We should never display the header bar when using SSDs anyway


jcollie (approved):

pluiedev:
I love spurious Git failures breaking CI

AlexJuca (approved):

AlexJuca:
@pluiedev I think re-running the CI pipeline might sort things out for you.

PR #5196, opened 2025-01-18
bash: prefer $HOME over tilde notation

### LINKED ISSUE #5206: bash: profile files not being sourced
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5169

<sup>Originally posted by **tristan957** January 16, 2025</sup>
I find that when the Bash shell integration is enabled, direnv stops working. Has anyone else experienced this issue?
</div>

---

In the follow-on discussion, this was narrowed down to show that profile scripts weren't being sourced during automatic shell integration, and the #5075 was identified as the source of the regression.

jparise:
This reverts part of #5075. While the ~ notation is more concise, I think the consensus is that using $HOME is more explicit, so prefer it.

We've also seen reports (#5169) that something about #5075 breaks the startup sequence in some environments. Reverting this part of the change will help us narrow that down.

Fixes #5206

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -47,7 +47,7 @@ if [ -n "$GHOSTTY_BASH_INJECT" ]; then
     if builtin shopt -q login_shell; then
       if [[ $__ghostty_bash_flags != *"--noprofile"* ]]; then
         [ -r /etc/profile ] && builtin source "/etc/profile"
-        for rcfile in ~/.bash_profile ~/.bash_login ~/.profile; do
+        for rcfile in "$HOME/.bash_profile" "$HOME/.bash_login" "$HOME/.profile"; do
           [ -r "$rcfile" ] && { builtin source "$rcfile"; break; }
         done
       fi
@@ -62,7 +62,7 @@ if [ -n "$GHOSTTY_BASH_INJECT" ]; then
         for rcfile in /etc/bash.bashrc /etc/bash/bashrc /etc/bashrc; do
           [ -r "$rcfile" ] && { builtin source "$rcfile"; break; }
         done
-        if [[ -z "$GHOSTTY_BASH_RCFILE" ]]; then GHOSTTY_BASH_RCFILE=~/.bashrc; fi
+        if [[ -z "$GHOSTTY_BASH_RCFILE" ]]; then GHOSTTY_BASH_RCFILE="$HOME/.bashrc"; fi
         [ -r "$GHOSTTY_BASH_RCFILE" ] && builtin source "$GHOSTTY_BASH_RCFILE"
       fi
     fi


tristan957:
I did some testing on my side. I think this is the reason for the regressions that I discovered. Only time will tell I guess after this is merged.

tristan957 (approved):
on line 59, there is a comment which includes a tilde. Might as well use $HOME there too

jparise:
> on line 59, there is a comment which includes a tilde. Might as well use $HOME there too

I think I'll let that one be for line length reasons.

jparise:
> I did some testing on my side. I think this is the reason for the regressions that I discovered. Only time will tell I guess after this is merged.

Great! I'm still unclear why the `~/` wouldn't be expanded to `$HOME/` here. (I know it would be treated as a literal if it were quoted, but there weren't any cases of that here.) And I don't think we're changing `$HOME` to something different than what bash would expand. Still a mystery!

tristan957:
On my side, all I did was remove the tildes from the variables, reinstalled Ghostty, checked that the edited bash integration was in place, and reverted my Ghostty config and bashrc changes. 

Now, I could have made a mistake somewhere along the way. I guess we will see after this PR merges for more confirmation.

src/shell-integration/bash/ghostty.bash
```diff
@@ -47,7 +47,7 @@ if [ -n "$GHOSTTY_BASH_INJECT" ]; then
     if builtin shopt -q login_shell; then
       if [[ $__ghostty_bash_flags != *"--noprofile"* ]]; then
         [ -r /etc/profile ] && builtin source "/etc/profile"
-        for rcfile in ~/.bash_profile ~/.bash_login ~/.profile; do
+        for rcfile in "$HOME/.bash_profile" "$HOME/.bash_login" "$HOME/.profile"; do
```
keszybz, line 50: This has the advantage that it'll behave correctly if $HOME includes whitespace. Of course it's a niche case, but it's better to be resilient.

src/shell-integration/bash/ghostty.bash
```diff
@@ -62,7 +62,7 @@ if [ -n "$GHOSTTY_BASH_INJECT" ]; then
         for rcfile in /etc/bash.bashrc /etc/bash/bashrc /etc/bashrc; do
           [ -r "$rcfile" ] && { builtin source "$rcfile"; break; }
         done
-        if [[ -z "$GHOSTTY_BASH_RCFILE" ]]; then GHOSTTY_BASH_RCFILE=~/.bashrc; fi
+        if [[ -z "$GHOSTTY_BASH_RCFILE" ]]; then GHOSTTY_BASH_RCFILE="$HOME/.bashrc"; fi
```
keszybz, line 65: As I wrote in the original PR, this is iffy:

> I have doubts about the expansion of the tilde here. Bash does expand ~ in the middle, but pure sh does not:
```console
[zbyszek@x1c ghostty]$ sh -c 'echo VAR=~'
VAR=~
[zbyszek@x1c ghostty]$ bash -c 'echo VAR=~'
VAR=/home/zbyszek
```

keszybz (approved):
LGTM.

tristan957:
@JacobCallahan to work around, you can probably source the shell integration manually in your bashrc.

JacobCallahan:
@tristan957 can you elaborate?

PR #5202, opened 2025-01-18
chore: update stb_image.h

eltociear:
exitting -> exiting


mitchellh:
It's likely this should go upstream since we just copy this and may reintroduce it if it's not fixed upstream. But I'll merge since its so basic.

PR #5212, opened 2025-01-18
termio: revise macOS-specific .hushlogin note

jparise:
login(1)'s .hushlogin logic was "fixed" in macOS Sonoma 14.4, so this
comment (and our workaround) is only relevant for versions earlier than
that.

The relevant change to login/login.c is part of system_cmds-979.100.8.

> login.c: look for .hushlogin in home directory (112854361)

- https://github.com/apple-oss-distributions/system_cmds/commit/1bca46ecc5b76432f42eb23ec39fe63e8159f251
- https://github.com/apple-oss-distributions/distribution-macOS/tree/macos-144


mitchellh:
Hah! Very interesting.

PR #5214, opened 2025-01-18
feat: initial implementation of i18n/l10n

pluiedev:
See #2357. Tech-wise, we're just using regular ol' `gettext`, except that we use Zig ports of `xgettext` and `msgfmt` that actually understand Zig source files. We also generate `.mo` files from `.po` files automatically, and one can run `zig build update-translations` to update the `.pot` and `.po` files (this is not run automatically as it alters source files).

I am terrible at translating. Please don't take my amateur 5-second Chinese translations seriously :p


pluiedev:
Superseded by #6004

PR #5217, opened 2025-01-19
ansi: Add support blinking text

thecharlesjenkins:
Add support for the blinking text ansi character. Both "slow" (code 5) and "fast" (code 6) blink at the same interval as the cursor, as it seems implementations often do this.

Blinking of text stops when the screen loses focus, same as the cursor. To keep blinking of text and cursor in sync, typing a character no longer resets the cursor blinking timer. Instead of reseting the timer, the cursor is immediately set to visible and will stay visible for at least the length of one interval to ensure that blinking does not happen while the user is actively typing. This has the side-effect that at startup, it takes two intervals until the first cursor blink happens when ghostty is launched.

Also introduce a config "blink-interval" which can set the blinking speed of the cursor and text to an arbitrary time. This config code was originally written by @pluiedev.

Discussion in: https://github.com/ghostty-org/ghostty/discussions/4258#discussion-7767836

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -478,6 +478,41 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 /// This value does not apply to Emoji or images.
 @"minimum-contrast": f64 = 1,
 
+/// The amount of time it takes for blinking text and cursors to toggle between
+/// being invisible and visible.
+/// Any cell on the screen could be set to blink by setting SGR attribute 5,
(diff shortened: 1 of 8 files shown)


mitchellh:
I'm sorry for not getting this in. Since this PR, the renderer has had a major refactor which makes considerable conflicts here. I'm still interested in this and it's on me for dropping the ball on not reviewing and fixing this up earlier to get it in. I'm sorry about that. If you're interested, please fix up the conflicts, rebase, and make a new PR.

PR #5219, opened 2025-01-19
feat: support selection foreground being cell foreground

### LINKED ISSUE #2685: Support selection foreground being cell foreground
From #2656 

Unlike #2656, the final approach I am going to suggest for this is to change `selection-foreground` to a tagged union in Zig and take some additional types. As part of this, I also propose removing `selection-invert-fg-bg` (cc @gpanders) and making it part of that union. Proposal:

```ini
# default
selection-foreground = window-background

# discussion 2656
selection-foreground = cell-foreground

# explicit color
selection-foreground = ABC123

# replacing selection-invert-fg-bg
selection-foreground = cell-background
selection-background = cell-foreground
```

dal-liu:
This resolves #2685.

## Changes

* Created `SelectionColor` tagged union that can take a `Color`, `cell-foreground`, or `cell-background`
* Used the new union to implement the feature for Metal and OpenGL
* Removed the use of `invert_selection_fg_bg` during rendering as suggested in the issue

## Demo

macOS:

https://github.com/user-attachments/assets/b5b2db83-bb62-4929-8e3c-870a1e1a7a5c

Ubuntu:

https://github.com/user-attachments/assets/259dd707-d9d9-4590-8f3c-a67ccdef34bc

Any feedback would be helpful, I'm sure there's a lot of room for improvement here.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -62,6 +62,13 @@ pub const compatibility = std.StaticStringMap(
     // Ghostty 1.2 removed the `hidden` value from `gtk-tabs-location` and
     // moved it to `window-show-tab-bar`.
     .{ "gtk-tabs-location", compatGtkTabsLocation },
+
+    // Ghostty 1.2 lets you set `cell-foreground` and `cell-background`
+    // to match the cell foreground and background colors, respectively.
+    // This can be used with `cursor-color` and `cursor-text` to recreate
+    // this behavior. This applies to selection too.
+    .{ "cursor-invert-fg-bg", compatCursorInvertFgBg },
+    .{ "selection-invert-fg-bg", compatSelectionInvertFgBg },
 });
 
 /// The font families to use.
@@ -597,18 +604,6 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 @"selection-foreground": ?DynamicColor = null,
 @"selection-background": ?DynamicColor = null,
 
-/// Swap the foreground and background colors of cells for selection. This
-/// option overrides the `selection-foreground` and `selection-background`
-/// options.
-///
-/// If you select across cells with differing foregrounds and backgrounds, the
-/// selection color will vary across the selection.
-///
-/// Warning: This option has been deprecated as of version 1.2.0. Instead,
-/// users should set `selection-foreground` and `selection-background` to
-/// `cell-background` and `cell-foreground`, respectively.
-@"selection-invert-fg-bg": bool = false,
-
 /// Whether to clear selected text when typing. This defaults to `true`.
 /// This is typical behavior for most terminal emulators as well as
 /// text input fields. If you set this to `false`, then the selected text
@@ -651,19 +646,20 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 palette: Palette = .{},
 
 /// The color of the cursor. If this is not set, a default will be chosen.
-/// Specified as either hex (`#RRGGBB` or `RRGGBB`) or a named X11 color.
-/// Since version 1.2.0, this can also be set to `cell-foreground` to match
-/// the cell foreground color, or `cell-background` to match the cell
-/// background color.
-@"cursor-color": ?DynamicColor = null,
-
-/// Swap the foreground and background colors of the cell under the cursor. This
-/// option overrides the `cursor-color` and `cursor-text` options.
 ///
-/// Warning: This option has been deprecated as of version 1.2.0. Instead,
-/// users should set `cursor-color` and `cursor-text` to `cell-foreground` and
-/// `cell-background`, respectively.
-@"cursor-invert-fg-bg": bool = false,
+/// Direct colors can be specified as either hex (`#RRGGBB` or `RRGGBB`)
+/// or a named X11 color.
+///
+/// Additionally, special values can be used to set the color to match
+/// other colors at runtime:
+///
+///   * `cell-foreground` - Match the cell foreground color.
+///     (Available since version 1.2.0)
+///
+///   * `cell-background` - Match the cell background color.
+///     (Available since version 1.2.0)
+///
+@"cursor-color": ?DynamicColor = null,
 
 /// The opacity level (opposite of transparency) of the cursor. A value of 1
 /// is fully opaque and a value of 0 is fully transparent. A value less than 0
@@ -3843,10 +3839,6 @@ pub fn parseManuallyHook(
     return true;
 }
 
-/// parseFieldManuallyFallback is a fallback called only when
-/// parsing the field directly failed. It can be used to implement
-/// backward compatibility. Since this is only called when parsing
-/// fails, it doesn't impact happy-path performance.
 fn compatGtkTabsLocation(
     self: *Config,
     alloc: Allocator,
@@ -3862,6 +3854,51 @@ fn compatGtkTabsLocation(
     }
 
     return false;
+}
+
+fn compatCursorInvertFgBg(
+    self: *Config,
+    alloc: Allocator,
+    key: []const u8,
+    value_: ?[]const u8,
+) bool {
+    _ = alloc;
+    assert(std.mem.eql(u8, key, "cursor-invert-fg-bg"));
+
+    // We don't do anything if the value is unset, which is technically
+    // not EXACTLY the same as prior behavior since it would fallback
+    // to doing whatever cursor-color/cursor-text were set to, but
+    // I don't want to store what that is separately so this is close
+    // enough.
+    //
+    // Realistically, these fields were mutually exclusive so anyone
+    // relying on that behavior should just upgrade to the new
+    // cursor-color/cursor-text fields.
+    const set = cli.args.parseBool(value_ orelse "t") catch return false;
+    if (set) {
+        self.@"cursor-color" = .@"cell-foreground";
+        self.@"cursor-text" = .@"cell-background";
+    }
+
+    return true;
+}
+
+fn compatSelectionInvertFgBg(
+    self: *Config,
+    alloc: Allocator,
+    key: []const u8,
+    value_: ?[]const u8,
+) bool {
+    _ = alloc;
+    assert(std.mem.eql(u8, key, "selection-invert-fg-bg"));
+
+    const set = cli.args.parseBool(value_ orelse "t") catch return false;
+    if (set) {
+        self.@"selection-foreground" = .@"cell-background";
+        self.@"selection-background" = .@"cell-foreground";
+    }
+
+    return true;
 }
 
 /// Create a shallow copy of this config. This will share all the memory
@@ -4437,28 +4474,41 @@ pub const DynamicColor = union(enum) {
 
     pub fn parseCLI(input_: ?[]const u8) !DynamicColor {
         const input = input_ orelse return error.ValueRequired;
-
         if (std.mem.eql(u8, input, "cell-foreground")) return .@"cell-foreground";
         if (std.mem.eql(u8, input, "cell-background")) return .@"cell-background";
-
-        return DynamicColor{ .color = try Color.parseCLI(input) };
+        return .{ .color = try Color.parseCLI(input) };
     }
 
     /// Used by Formatter
     pub fn formatEntry(self: DynamicColor, formatter: anytype) !void {
         switch (self) {
             .color => try self.color.formatEntry(formatter),
-            .@"cell-foreground", .@"cell-background" => try formatter.formatEntry([:0]const u8, @tagName(self)),
+
+            .@"cell-foreground",
+            .@"cell-background",
+            => try formatter.formatEntry([:0]const u8, @tagName(self)),
         }
     }
 
     test "parseCLI" {
         const testing = std.testing;
 
-        try testing.expectEqual(DynamicColor{ .color = Color{ .r = 78, .g = 42, .b = 132 } }, try DynamicColor.parseCLI("#4e2a84"));
-        try testing.expectEqual(DynamicColor{ .color = Color{ .r = 0, .g = 0, .b = 0 } }, try DynamicColor.parseCLI("black"));
-        try testing.expectEqual(DynamicColor{.@"cell-foreground"}, try DynamicColor.parseCLI("cell-foreground"));
-        try testing.expectEqual(DynamicColor{.@"cell-background"}, try DynamicColor.parseCLI("cell-background"));
+        try testing.expectEqual(
+            DynamicColor{ .color = Color{ .r = 78, .g = 42, .b = 132 } },
+            try DynamicColor.parseCLI("#4e2a84"),
+        );
+        try testing.expectEqual(
+            DynamicColor{ .color = Color{ .r = 0, .g = 0, .b = 0 } },
+            try DynamicColor.parseCLI("black"),
+        );
+        try testing.expectEqual(
+            DynamicColor{.@"cell-foreground"},
+            try DynamicColor.parseCLI("cell-foreground"),
+        );
+        try testing.expectEqual(
+            DynamicColor{.@"cell-background"},
+            try DynamicColor.parseCLI("cell-background"),
+        );
 
         try testing.expectError(error.InvalidValue, DynamicColor.parseCLI("a"));
     }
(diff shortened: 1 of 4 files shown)


kjmph:
I like this patch. A couple of minor things, when there are inverted cells, this patch shows the original foreground, rather than the inverted foreground. If you compare to Terminal.app, it will always show the inverted foreground. Additionally, very oddly, the cursor-text seems to not be honored from the theme.

If any example code could help, here is my personal patch (from the 1.0.1 tag) that resolves all these, with the exception that I have a custom theme with no `selection-foreground` specified. I also do something similar with cursor text so that the cell-foreground is used. Can we make that change there as well?

```patch
diff --git a/src/renderer/Metal.zig b/src/renderer/Metal.zig
index b37f440f..a2ef6c16 100644
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -2435,8 +2435,17 @@ fn rebuildCells(
                 if (selected and !self.config.invert_selection_fg_bg) {
                     // If we don't have invert selection fg/bg set
                     // then we just use the selection foreground if
-                    // set, otherwise the default bg color.
-                    break :fg self.config.selection_foreground orelse self.background_color orelse self.default_background_color;
+                    // set, otherwise the exising fg color.
+                    if (style.flags.inverse) {
+                        break :fg self.config.selection_foreground orelse bg_style orelse self.background_color orelse self.default_background_color;
+                    } else {
+                        break :fg self.config.selection_foreground orelse fg_style;
+                    }
+                    // Except, all themes set selection_foreground! We
+                    // want a theme that doesn't set
+                    // selection_foreground. We could always force
+                    // fg_style, but that isn't a general solution
+                    //break :fg fg_style;
                 }
 
                 // Whether we need to use the bg color as our fg color:
@@ -2671,8 +2680,14 @@ fn rebuildCells(
                 break :blk sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color;
             } else if (self.config.cursor_text) |txt|
                 txt
-            else
-                self.background_color orelse self.default_background_color;
+            else base: {
+                const sty = screen.cursor.page_pin.style(screen.cursor.page_cell);
+                if (sty.flags.inverse) {
+                    break :base sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color;
+                } else {
+                    break :base sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color;
+                }
+            };
 
             self.uniforms.cursor_color = .{
                 uniform_color.r,
```

dal-liu:
@kjmph Thanks for catching the inverted cell issue, it should hopefully be fixed now. I've also made `cursor-text` and `cursor-color` accept `cell-foreground` and `cell-background` the same way that the selection colors do.

kjmph:
Haha, great improvement! Except for two things. 1) If the `cursor-color` is `cell-foreground` then it wipes out the cell's text. If `cursor-color` is `cell-background` then it can't be seen because it blends in. I think it wise to remove the support from `cursor-color` since it will lead people astray into odd configuration territory. 2) The cursor foreground is wiped out with inverted cells. I don't understand why that happens with your PR, unless it exists in the upstream `main` branch? I'm personally working on the `1.0.1` tag.

Here is an image showing this patch (and `cursor-text = cell-foreground`) with inverted cells:
Screenshot 2025-01-21 at 10 43 44 PM

Here is an image showing the patch I shared against `1.0.1` with inverted cells:
Screenshot 2025-01-21 at 10 44 10 PM

I'll give a test against `main`.

kjmph:
Oh!!! This is a problem upstream, it isn't a regression with this patch. I bet most people allow the shell integration for cursors and because it is a bar, don't notice the color mishap.

dal-liu:
To me, changing `cursor-color` to accept cell colors makes sense since we should have more control over the colors, plus it was suggested in the original issue as well. That being said, if a 3rd opinion disagrees then I'll gladly revert the changes, so I will leave it for now.

As for the `cursor-text` thing, it's definitely worth opening a discussion about it.

src/renderer/Metal.zig
```diff
@@ -2833,19 +2803,21 @@ fn rebuildCells(
...
+            break :color switch (color) {
+                // Use the color set by cursor-color, if any. If the cell
+                // is reversed, use the opposite cell color instead.
+                .color => color.color.toTerminalRGB(),
+                .@"cell-foreground" => if (sty.flags.inverse) bg_style else fg_style,
+                .@"cell-background" => if (sty.flags.inverse) fg_style else bg_style,
+            };
+        } else
+        // Otherwise, use the foreground color.
+        self.foreground_color orelse self.default_foreground_color;
```
gpanders, line 2820: Indentation looks off here, run this through `zig fmt` if you haven't already.

src/renderer/Metal.zig
```diff
@@ -2833,19 +2803,21 @@ fn rebuildCells(
...
-                else
-                    (sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color);
-            } else {
-                break :color self.foreground_color orelse self.default_foreground_color;
-            }
-        };
+        const cursor_color = self.cursor_color orelse if (self.default_cursor_color) |color| color: {
+            const sty = screen.cursor.page_pin.style(screen.cursor.page_cell);
+            const fg_style = sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color;
+            const bg_style = sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color;
```
gpanders, line 2809: It's probably not a big deal, but we are computing these styles even if they're not needed when the cursor color is set to an explicit style. I don't think there is a significant runtime cost though. If we wanted to avoid that we could first check to see if the enum has the `.color` tag and only compute the styles if it does not (this would be a perfect use case for https://github.com/ziglang/zig/issues/12863)

src/config/Config.zig
```diff
@@ -492,8 +494,10 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
 palette: Palette = .{},
 
 /// The color of the cursor. If this is not set, a default will be chosen.
-/// Specified as either hex (`#RRGGBB` or `RRGGBB`) or a named X11 color.
-@"cursor-color": ?Color = null,
+/// Specified as either hex (`#RRGGBB` or `RRGGBB`), a named X11 color,
+/// `cell-foreground` to match the cell foreground color, or `cell-background`
+/// to match the cell background color.
+@"cursor-color": ?DynamicColor = null,
```
gpanders, line 500: Probably not for this PR (and maybe never), but I wonder if we should unify the terminology for cursor and selection, so that instead of `cursor-text` and `cursor-color` we use `cursor-foreground` and `cursor-background`.

gpanders (commented):
Nice work for a first time contributor :) This is imo a nice simplification of some of this code which was starting to get a bit hairy.

This PR makes `selection-invert-fg-bg` and `cursor-invert-fg-bg` no-ops, which is a breaking change. I'll leave it to Mitchell to determine how to approach that, but at the least the doc strings for those options should be updated to indicate that they no longer do anything.

src/renderer/Metal.zig
dal-liu: `zig fmt` didn't change anything so I just moved the comment somewhere else, hopefully it still makes sense

src/renderer/Metal.zig
dal-liu: Added a check before these calls which should skip them if the union has the `.color` tag

kjmph:
@cwrau It looks like it could match gnome-terminal, but I don't have your precise setup. Here is the best replication I can draw:
Screenshot 2025-01-28 at 11 48 32 AM
Screenshot 2025-01-28 at 11 48 45 AM

But again, I'm using `block_hollow` to show this. Otherwise, the block will wipe out the foreground because it matches the foreground color. It won't invert the text color like gnome-terminal.

I think you would want a `cursor-invert-fg-bg = true` corollary to still be to get the same behavior as gnome-terminal (although it will have to be blink aware). I notice my patch also has this problem, and I get away with it via a non-standard cursor color by default. I fixed it in my personal patch, which I attached the diff from the 1.0.1 tag. Note, if you wanted to try it you would have to be on a Mac, and you would need to remove `selection-foreground` from your theme in: `Ghostty.app/Contents/Resources/ghostty/themes/<your_theme>`

Then, set the config:

```
cursor-style = block
cursor-style-blink = true
cursor-invert-fg-bg = true
```

That should match gnome terminal in your example. See attached video:

https://github.com/user-attachments/assets/b1f85db6-6efc-4a8a-86a9-e3298b503001

Patch:
```diff
diff --git a/src/renderer/Metal.zig b/src/renderer/Metal.zig
index b37f440f..db41d413 100644
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -2435,8 +2435,17 @@ fn rebuildCells(
                 if (selected and !self.config.invert_selection_fg_bg) {
                     // If we don't have invert selection fg/bg set
                     // then we just use the selection foreground if
-                    // set, otherwise the default bg color.
-                    break :fg self.config.selection_foreground orelse self.background_color orelse self.default_background_color;
+                    // set, otherwise the existing fg color.
+                    if (style.flags.inverse) {
+                        break :fg self.config.selection_foreground orelse bg_style orelse self.background_color orelse self.default_background_color;
+                    } else {
+                        break :fg self.config.selection_foreground orelse fg_style;
+                    }
+                    // Except, all themes set selection_foreground! We
+                    // want a theme that doesn't set
+                    // selection_foreground. We could always force
+                    // fg_style, but that isn't a general solution
+                    //break :fg fg_style;
                 }
 
                 // Whether we need to use the bg color as our fg color:
@@ -2638,7 +2647,11 @@ fn rebuildCells(
         const cursor_color = self.cursor_color orelse self.default_cursor_color orelse color: {
             if (self.cursor_invert) {
                 const sty = screen.cursor.page_pin.style(screen.cursor.page_cell);
-                break :color sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color;
+                if (sty.flags.inverse) {
+                    break :color sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color;
+                } else {
+                    break :color sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color;
+                }
             } else {
                 break :color self.foreground_color orelse self.default_foreground_color;
             }
@@ -2668,11 +2681,21 @@ fn rebuildCells(
 
             const uniform_color = if (self.cursor_invert) blk: {
                 const sty = screen.cursor.page_pin.style(screen.cursor.page_cell);
-                break :blk sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color;
+                if (sty.flags.inverse) {
+                    break :blk sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color;
+                } else {
+                    break :blk sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color;
+                }
             } else if (self.config.cursor_text) |txt|
                 txt
-            else
-                self.background_color orelse self.default_background_color;
+            else base: {
+                const sty = screen.cursor.page_pin.style(screen.cursor.page_cell);
+                if (sty.flags.inverse) {
+                    break :base sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color;
+                } else {
+                    break :base sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color;
+                }
+            };
 
             self.uniforms.cursor_color = .{
                 uniform_color.r,
```

kjmph:
Actually, I stand corrected, with this PR, I can get close to gnome-terminal @cwrau:
```
cursor-style = block
cursor-text = cell-background
cursor-style-blink = true
cursor-color = cell-foreground
```
But the foreground colors look saturated incorrectly? (see the S and the E here:)
Screenshot 2025-01-28 at 3 35 17 PM
Screenshot 2025-01-28 at 3 34 31 PM

cwrau:
> Actually, I stand corrected, with this PR, I can get close to gnome-terminal @cwrau:
> 
> ```
> shell-integration-features = no-cursor
> cursor-style = block
> cursor-text = cell-background
> cursor-style-blink = true
> cursor-color = cell-foreground
> ```
> 
> But the foreground colors look saturated incorrectly? (see the S and the E here:) Screenshot 2025-01-28 at 3 35 17 PM Screenshot 2025-01-28 at 3 34 31 PM

Amazing! Looking forward to this PR 😁

kjmph:
Anything I can assist with this PR? It would be nice to stop maintaining a private patch.

kjmph:
I will say, my personal patches are reduced after 1.1.0 was released:

```patch
diff --git a/src/renderer/Metal.zig b/src/renderer/Metal.zig
index 866f9682..7561f1a7 100644
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -2635,8 +2635,17 @@ fn rebuildCells(
                 if (selected and !self.config.invert_selection_fg_bg) {
                     // If we don't have invert selection fg/bg set
                     // then we just use the selection foreground if
-                    // set, otherwise the default bg color.
-                    break :fg self.config.selection_foreground orelse self.background_color orelse self.default_background_color;
+                    // set, otherwise the existing fg color.
+                    if (style.flags.inverse) {
+                        break :fg self.config.selection_foreground orelse bg_style orelse self.background_color orelse self.default_background_color;
+                    } else {
+                        break :fg self.config.selection_foreground orelse fg_style;
+                    }
+                    // Except, all themes set selection_foreground! We
+                    // want a theme that doesn't set
+                    // selection_foreground. We could always force
+                    // fg_style, but that isn't a general solution
+                    //break :fg fg_style;
                 }
 
                 // Whether we need to use the bg color as our fg color:
@@ -2879,8 +2888,13 @@ fn rebuildCells(
                     (sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color);
             } else if (self.config.cursor_text) |txt|
                 txt
-            else
-                self.background_color orelse self.default_background_color;
+            else base: {
+                const sty = screen.cursor.page_pin.style(screen.cursor.page_cell);
+                break :base if (sty.flags.inverse)
+                    (sty.bg(screen.cursor.page_cell, color_palette) orelse self.background_color orelse self.default_background_color)
+                else
+                    (sty.fg(color_palette, self.config.bold_is_bright) orelse self.foreground_color orelse self.default_foreground_color);
+            };
 
             self.uniforms.cursor_color = .{
                 uniform_color.r,
```

mitchellh:
Apologies for the large delay on this. I'm open to this but can you rebase it on the latest renderer refactor and I'm happy to take another look.

src/config/Config.zig
```diff
@@ -591,15 +591,22 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
...
 /// Swap the foreground and background colors of cells for selection. This
 /// option overrides the `selection-foreground` and `selection-background`
 /// options.
 ///
 /// If you select across cells with differing foregrounds and backgrounds, the
 /// selection color will vary across the selection.
+///
+/// Warning: This option has been deprecated as of version 1.2.0. Instead,
+/// users should set `selection-foreground` and `selection-background` to
+/// `cell-background` and `cell-foreground`, respectively.
```
mitchellh: Awesome, I can handle writing the backwards compatibility handler for this so we can remove this.

src/config/Config.zig
```diff
@@ -4409,6 +4426,54 @@ pub const Color = struct {
     }
 };
 
+/// Represents the color values that can be set to a non-static value.
+///
+/// Can either be a Color or one of the special values
+/// "cell-foreground" or "cell-background".
+pub const DynamicColor = union(enum) {
```
mitchellh: Bikeshed I don't know if `DynamicColor` is the right name for this. Maybe it is. I'll think about it.

src/termio/Termio.zig
```diff
@@ -265,8 +263,11 @@ pub fn init(self: *Termio, alloc: Allocator, opts: termio.Options) !void {
     // Create our stream handler. This points to memory in self so it
     // isn't safe to use until self.* is set.
     const handler: StreamHandler = handler: {
-        const default_cursor_color = if (!opts.config.cursor_invert and opts.config.cursor_color != null)
-            opts.config.cursor_color.?.toTerminalRGB()
+        const default_cursor_color = if (opts.config.cursor_color) |color|
+            switch (color) {
+                .color => color.color.toTerminalRGB(),
+                else => null,
```
mitchellh: We should use an exhaustive switch here instead of `else` so the compiler can error if we add more cases so we can consider whether to update this.

src/termio/stream_handler.zig
```diff
@@ -121,8 +121,11 @@ pub const StreamHandler = struct {
         self.default_background_color = config.background.toTerminalRGB();
         self.default_cursor_style = config.cursor_style;
         self.default_cursor_blink = config.cursor_blink;
-        self.default_cursor_color = if (!config.cursor_invert and config.cursor_color != null)
-            config.cursor_color.?.toTerminalRGB()
+        self.default_cursor_color = if (config.cursor_color) |color|
+            switch (color) {
+                .color => color.color.toTerminalRGB(),
+                else => null,
```
mitchellh: Same feedback here, we should use an exhaustive enum.

mitchellh (changes requested):
Thanks so much, this looks really good. I left some notes. I plan on picking this up and carrying this over the finish line myself but if you want to address the easy ones feel free. Appreciate you!

mitchellh:
Pushed up backwards compatibility handlers for the old fields. I'm going to address some other feedback but have to step away for now.

kjmph:
Apologies, @mitchellh, I know this is considered closed. Yet, if `cursor-text = cell-foreground` is set, I can see the `cursor-text` looks different. I posted here: https://github.com/ghostty-org/ghostty/pull/5219#issuecomment-2620007145

The function `cell_text_vertex` in `src/renderer/shaders/shaders.metal` seems to uses linear blending for all other cells, but the cursor cell uses the Display P3 color directly. If I set it to use linear coloring then it matches all other cells. Like so:

```patch
diff --git a/src/renderer/shaders/shaders.metal b/src/renderer/shaders/shaders.metal
index 4797f89e4..4e02b6336 100644
--- a/src/renderer/shaders/shaders.metal
+++ b/src/renderer/shaders/shaders.metal
@@ -665,13 +665,13 @@ vertex CellTextVertexOut cell_text_vertex(
   // If this cell is the cursor cell, but we're not processing
   // the cursor glyph itself, then we need to change the color.
   if ((in.bools & IS_CURSOR_GLYPH) == 0 && is_cursor_pos) {
     out.color = load_color(
       uniforms.cursor_color,
       uniforms.use_display_p3,
-      false
+      true
     );
   }
 
   return out;
 }
 
```

Would you prefer if this was a new issue?

00-kat:
> Would you prefer if this was a new issue?

That probably would be a good idea, I'm not sure if anyone is going to see your comment.

PR #5222, opened 2025-01-19
fix: quick terminal hidden by macos menu bar

dmehala:
ghostty#5000 changed the window level from `.popupMenu` to `.floating` to improve IME support. However, this introduced a side effect which render the Quick Terminal (QT) below the macOS menu bar, whereas previously it would cover it.

When positioned on `right` and `left`, the top of the QT becomes partially hidden. This PR adjust the size of the QT to ensure it remains fully visible and stays below the menu bar.

Before #5000:
Screen Recording 2025-01-19 at 11 14 48
With #5000:
Screen Recording 2025-01-19 at 11 13 27
This branch:
Screen Recording 2025-01-19 at 11 17 32


mitchellh:
Perfect thank you, this looks right!

PR #5233, opened 2025-01-19
Add background image support for Metal

br0h1t:
- Implements #3645 for Metal.
- Uses foundation / logic from the OpenGL counterpart (https://github.com/ghostty-org/ghostty/pull/4226) making changes only when necessary

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -30,10 +30,11 @@ pub const RepeatableFontVariation = Config.RepeatableFontVariation;
 pub const RepeatableString = Config.RepeatableString;
 pub const RepeatableStringMap = @import("config/RepeatableStringMap.zig");
 pub const RepeatablePath = Config.RepeatablePath;
-pub const SinglePath = Config.SinglePath;
+pub const Path = Config.Path;
 pub const ShellIntegrationFeatures = Config.ShellIntegrationFeatures;
 pub const WindowPaddingColor = Config.WindowPaddingColor;
 pub const BackgroundImageMode = Config.BackgroundImageMode;
+pub const BackgroundImagePosition = Config.BackgroundImagePosition;
 
 // Alternate APIs
 pub const CAPI = @import("config/CAPI.zig");
--- a/src/config.zig
+++ b/src/config.zig
@@ -30,8 +30,10 @@ pub const RepeatableFontVariation = Config.RepeatableFontVariation;
 pub const RepeatableString = Config.RepeatableString;
 pub const RepeatableStringMap = @import("config/RepeatableStringMap.zig");
 pub const RepeatablePath = Config.RepeatablePath;
+pub const SinglePath = Config.SinglePath;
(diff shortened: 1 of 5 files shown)


sagg0t:
Hey @Rohit-Bevinahally, thank you for this PR, I can't wait to have it merged!
While using these changes in my local build, I noticed that the background image is separate for every split. I think many ghostty users (including me) would appreciate being able to have a single image for the whole window.

What's you opinion on this? Thanks again for the PR!

P.S. also noticed that the image doesn't cover vertical padding 
image
image

jcollie:
> While using these changes in my local build, I noticed that the background image is separate for every split. I think many ghostty users (including me) would appreciate being able to have a single image for the whole window.

This is because we don't have a renderer layer (Metal or OpenGL) that covers the entire window, only each individual surface. Adding that would require _major_ surgery in how rendering is implemented and out-of-scope for this PR, and not something to be attempted without some in-depth discussion.

Rohit-Bevinahally:
> P.S. also noticed that the image doesn't cover vertical padding

Hi @sagg0t. Sorry I missed your comment.
The current design on opengl and metal draws the image on the area covered by the terminal (i.e. excluding padding). Whether it is better to cover the full screen instead or add another config option to choose between the two will need some discussion.
To cover the full screen you can apply this patch
<details><summary>Patch</summary>

```diff
diff --git a/src/renderer/Metal.zig b/src/renderer/Metal.zig
index dc3cf26b..54c155dd 100644
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -1725,8 +1725,8 @@ fn drawBackgroundImage(
     const Buffer = mtl_buffer.Buffer(mtl_shaders.BgImage);
     var buf = try Buffer.initFill(self.gpu_state.device, &.{.{
         .terminal_size = .{
-            @as(f32, @floatFromInt(self.size.terminal().width)),
-            @as(f32, @floatFromInt(self.size.terminal().height)),
+            @as(f32, @floatFromInt(self.size.screen.width)),
+            @as(f32, @floatFromInt(self.size.screen.height)),
         },
         .mode = self.background_image_mode,
     }}, .{
diff --git a/src/renderer/shaders/cell.metal b/src/renderer/shaders/cell.metal
index 3d18c646..34b88e22 100644
--- a/src/renderer/shaders/cell.metal
+++ b/src/renderer/shaders/cell.metal
@@ -780,6 +780,8 @@ vertex BgImageVertexOut bg_image_vertex(
         break;
   }
 
+  offset -= uniforms.grid_padding.wx;
+
   out.position = uniforms.projection_matrix * float4(final_image_size + offset, 0.0, 1.0);
   out.tex_coord = position;
   if (in.mode == MODE_TILED) {

```

</details>

yunusey (approved):
Thank you so much for this PR! I don't really know anything about the Metal shaders, but from what I can understand, it looks pretty nice.

qwerasd205:
This PR is superseded by #7686. Thanks for this work, regardless!

PR #5249, opened 2025-01-20
bash: revert automatic shell integration changes

### LINKED ISSUE #5206: bash: profile files not being sourced

jparise:
The intention of #5075 was to create a less intrusive, more hermetic environment in which to source the bash startup files. This caused problems for multiple people, and I believe that's because the general expectation is that these files are sourced at global (not function) scope.

For example, when a file is sourced from within a function scope, any variables that weren't explicitly exported into the global environment won't be available outside of the scope of the function. Most system and personal startup files aren't written with that constraint because it's not how bash itself loads these files.

As a small improvement over the original code, `rcfile` has been renamed to `__ghostty_rcfile`. Avoiding leaking this variable while sourcing these files was a goal of #5075, and prefixing it make it much less of a potential issue.

This change also reverts the $HOME -> ~/ change. While the ~/ notation is more concise, using $HOME is more common and easier to implement safely with regard to quoting.

Fixes #5206

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -41,37 +41,31 @@ if [ -n "$GHOSTTY_BASH_INJECT" ]; then
 
   # Manually source the startup files. See INVOCATION in bash(1) and
   # run_startup_files() in shell.c in the Bash source code.
-  function __ghostty_bash_startup() {
-    builtin local rcfile
-
-    if builtin shopt -q login_shell; then
-      if [[ $__ghostty_bash_flags != *"--noprofile"* ]]; then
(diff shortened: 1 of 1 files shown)


tristan957:
:superhero:

JacobCallahan:
@jparise I've applied the patch locally and can confirm that the issue has been resolved. Thanks!

PR #5265, opened 2025-01-21
fix(renderer): clip terminal contents to expected grid size (#4523)

### LINKED ISSUE #4523: Renderers (Metal and OpenGL) should clip terminal if terminal grid size mismatches physical window
Both Metal and OpenGL have something like the following:

https://github.com/ghostty-org/ghostty/blob/7eb35d72758a984c6c028272e6f53b7694e42b60/src/renderer/OpenGL.zig#L749-L760

In this case, we just stop rendering which makes it look like the terminal is hanging. We should clip the terminal area (if it's bigger than our window) or blank the padding (if it's smaller) in this scenario. 

The only known way to trigger this is DECCOLM, so it's low priority since almost nothing uses this. But notably `vtetest` uses this and it makes it appear like we perform much worse than we do.

qwerasd205:
Resolves #4523

More notably this fixes a memory corruption crash that can occur while resizing the font under Metal while there's a lot of active changes occurring (e.g. while running DOOM fire). The change where all background colors are explicitly written exposed this issue, though it was technically a problem the whole time I'm fairly sure, just that the corruption it caused before was benign.

This significantly improves the robustness of the renderers since it prevents synchronization issues from causing memory corruption due to out of bounds read/writes while building the cells.

TODO: when viewport is narrower than renderer grid size, fill blank margin with bg color- currently appears as black, this only affects DECCOLM right now, and possibly could create single-frame artefacts during poorly managed resizes, but it's not ideal regardless.

### DIFF
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -1046,19 +1046,6 @@ pub fn updateFrame(
             }
         }
 
-        // If our terminal screen size doesn't match our expected renderer
-        // size then we skip a frame. This can happen if the terminal state
-        // is resized between when the renderer mailbox is drained and when
-        // the state mutex is acquired inside this function.
-        //
-        // For some reason this doesn't seem to cause any significant issues
-        // with flickering while resizing. '\_('-')_/'
-        if (self.cells.size.rows != state.terminal.rows or
-            self.cells.size.columns != state.terminal.cols)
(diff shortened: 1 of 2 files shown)


mitchellh:
Thanks! I didn't verify this but the logic looks good to me.

PR #5267, opened 2025-01-21
Prevent hyperlink hover state when mouse is outside viewport

### LINKED ISSUE #5252: macOS: hyperlink in top corner of window gets fake mouse-over when holding Cmd key
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5236

<sup>Originally posted by **rrotter** January 19, 2025</sup>
This is a little hard to describe, so I'll start by listing the steps to reproduce:

1. put a hyperlink in the top left of a window (`fd --hyperlink` is a quick way to do this)
2. move mouse out of the window, hold Cmd
3. hyperlink in top left acts like it's getting a mouse-over whenever the mouse is outside the window (whether window has focus or not)

Reproduced this with current (72d0855) tip build, as well as v1.0.1.
I don't see anything about this in existing discussions or issues.</div>

liby:
## Description

Fixed an issue where hyperlinks would maintain their hover state when the mouse is outside the viewport while holding the Cmd key. This created inconsistent behavior with the window's standard hover state clearing logic.

## Changes

- Added viewport boundary check in `mouseRefreshLinks` function to prevent link processing when cursor is outside the window

## Testing

To reproduce and verify the fix:

1. Run `fd --hyperlink "\.zig$" src/apprt` to create hyperlinks

2. Move mouse over a hyperlink (it should highlight)

3. Move mouse outside window

4. Hold Cmd key

	- Before: Link would show hover state

	- After: Link remains in non-hover state

Fixes #5252

@rrotter Could you please try this to see if it solves your issue?

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1041,6 +1041,9 @@ fn mouseRefreshLinks(
     pos_vp: terminal.point.Coordinate,
     over_link: bool,
 ) !void {
+    // If the position is outside our viewport, do nothing
+    if (pos.x < 0 or pos.y < 0) return;
+
     self.mouse.link_point = pos_vp;
 
     if (try self.linkAtPos(pos)) |link| {
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -638,7 +638,7 @@ pub const Surface = struct {
                 .y = @floatCast(opts.scale_factor),
             },
             .size = .{ .width = 800, .height = 600 },
-            .cursor_pos = .{ .x = 0, .y = 0 },
+            .cursor_pos = .{ .x = -1, .y = -1 },
             .keymap_state = .{},
         };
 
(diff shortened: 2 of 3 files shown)


rrotter:
Thanks for looking into this. Your fix partially works. For it to work the mouse has to have ever exited the window, but if it's never been in the window the unexpected hover state still occurs.

Steps to reproduce:
1. move mouse away from current window, so it won't be over next window created
2. cmd-n
3. make a hyperlink w/ keyboard (I just type `fd --hyperlink` in my home directory a couple of times)
4. press and hold cmd: top left hyperlink is again in hover state

liby:
> if it's never been in the window the unexpected hover state still occurs

Thank you for the detailed bug report. I've investigated this issue and found the root cause:

The issue occurs because of how mouse position is initialized in new windows. When a new window is created, the mouse position is initially set to (0, 0), which can trigger hover effects on hyperlinks at that position even though the mouse has never physically entered the window.

The fix is to initialize the mouse position to invalid coordinates (-1, -1) in both the embedded runtime and GTK.

The previous behavior was problematic because it assumed (0, 0) as a valid initial position, which could trigger hover effects on any hyperlinks at the top-left corner of the window.

Let me know if you notice any other edge cases or if you have any questions about the implementation.

Issue #3496, opened 2024-12-27
write_scrollback_file doesn't include what's on the current screen


Tenzer:
First off, great work on the achievement of this project!

The `write_scrollback_file` action was suggested in https://github.com/ghostty-org/ghostty/issues/189#issuecomment-1824543708 to be an alternative to the currently missing scrollback search functionality.

When testing this out, I however noticed it's only contents of the scrollback which is outside of the current screen which get written to the file, which means - if you want to search through the scrollback - you will only be able to search through the lines which are not visible on the screen.

Is this intended? Would it not be better to output everything in the scrollback, including what is visible on the current screen?

I can't think of a use case where you'd only want the contents that aren't visible on the screen to be written in the file.

As a quick way of testing this, you can open a new tab, enter `for i in $(seq 1 100); do echo $i; done` to print out numbered lines, and then perform the `write_scrollback_file` action by pressing `super + shift + j` and notice it doesn't include all the lines outputted from the loop.

mitchellh:
`write_screen_file` is what you want. `write_scrollback_file` only writes the _scrollback_ (not the active screen)

Tenzer:
Sorry! Right after creating this, I noticed the `write_screen_file` ([link](https://ghostty.org/docs/config/keybind/reference#write_screen_file)) action is available which does include the contents of the current screen.

jcollie:
FWIW, `write_screen_file` and `write_scrollback_file` are absolutely _awful_ names for these operations given what they do. I think that we should re-open this issue as a springboard to rename them and to change the default keybinding.

mitchellh:
This is the terminology we use internally:

https://github.com/ghostty-org/ghostty/blob/1fa2e699d142b9bc2795a4d3db98b035cc5a825c/src/terminal/point.zig#L6-L48

mitchellh:
I’m not trying to suggest it. I’m just posting it as information.

PR #5285, opened 2025-01-22
Change default key bindings to capture full screen contents (`write_screen_file`)

MadLittleMods:
The current default keybinding on `Ctrl-Shift-J`/`Cmd-Shift-J` utilize [`write_scrollback_file`](https://ghostty.org/docs/config/keybind/reference#write_scrollback_file) which only captures text that's scrolled off-screen. This can be confusing, as I would expect it to capture everything on and off-screen.

Per the docs, [`write_screen_file`](https://ghostty.org/docs/config/keybind/reference#write_screen_file) is the "Same as `write_scrollback_file` but writes the full screen contents" which is perfect to solve this problem and aligns exactly with the expected behavior.

This change addresses the friction that people have reported in the discussions below and specifically coming from the [search scrollback feature discussion](https://github.com/ghostty-org/ghostty/issues/189), where `Ctrl-Shift-J`/`Cmd-Shift-J` is one of the interim decent workarounds but then you also have to figure out the extra intricacy to use `write_screen_file` instead of `write_scrollback_file`.

Previous discussions:

 - https://github.com/ghostty-org/ghostty/discussions/3652
 - https://github.com/ghostty-org/ghostty/issues/3496
 - https://github.com/ghostty-org/ghostty/discussions/4911
 - https://github.com/ghostty-org/ghostty/discussions/4390
 - https://github.com/ghostty-org/ghostty/discussions/2363#discussioncomment-11735957
 - https://github.com/ghostty-org/ghostty/issues/189#issuecomment-2564719973
 - https://github.com/ghostty-org/ghostty/pull/2040
 
### Workaround

Before this PR is merged, you can achieve the same result by updating your Ghostty config:

`.config/ghostty`
```sh
# So we have more scrollback history (the size of the scrollback buffer in bytes)
scrollback-limit = 100000000

# The default keybindings use `write_scrollback_file` but that only includes what is
# offscreen. `write_screen_file` includes what's on screen and offscreen.
keybind = ctrl+shift+j=write_screen_file:paste
keybind = ctrl+shift+alt+j=write_screen_file:open
```

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2370,13 +2370,13 @@ pub fn default(alloc_gpa: Allocator) Allocator.Error!Config {
     try result.keybind.set.put(
         alloc,
         .{ .key = .{ .translated = .j }, .mods = inputpkg.ctrlOrSuper(.{ .shift = true }) },
-        .{ .write_scrollback_file = .paste },
+        .{ .write_screen_file = .paste },
     );
 
     try result.keybind.set.put(
         alloc,
         .{ .key = .{ .translated = .j }, .mods = inputpkg.ctrlOrSuper(.{ .shift = true, .alt = true }) },
-        .{ .write_scrollback_file = .open },
+        .{ .write_screen_file = .open },
     );
 
     // Expand Selection


mitchellh:
Thanks, I think this is a very reasonable change.

PR #5286, opened 2025-01-22
Fix text selection when dragging window with hidden titlebar

### LINKED ISSUE #5259: macOS: Dragging with `macos-titlebar-style=hidden` should not invoke selection
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5208

<sup>Originally posted by **korikhin** January 18, 2025</sup>
👋

I noticed a small inconvenience while dealing with window dragging.

*Sorry if this is a duplicate, but I couldn't find any discussions/issues related to this matter, nor any mentions that this behaviour will be fixed in future releases.*

1. You lose text selection when dragging the window
2. Selection occurs when you drag the window with too much passion

See the recording below. Thanks!

OS: macOS 15.2.0
Version: 1.0.1
Build 8696
Commit: 3f7c3afa

```ini
font-family = "JetBrainsMono Nerd Font Mono"
font-size = 15
adjust-cell-height = 12
theme = catppuccin-macchiato
#minimum-contrast = 1.2
background-opacity = 0.9
background-blur-radius = 50
unfocused-split-fill = #1E1E2E
window-padding-x = 10
window-padding-y = 10
window-padding-balance = true
confirm-close-surface = false
#macos-non-native-fullscreen = visible-menu
macos-titlebar-style = hidden
macos-option-as-alt = true
#macos-icon = custom-style
```

https://github.com/user-attachments/assets/7c67be9c-f612-410a-8023-da3b531e821c</div>

## Reproduction

```
macos-titlebar-style = hidden
```

Print some text, select it, and start jiggling around your mouse near the titlebar.

liby:
When `macos-titlebar-style i`s set to `hidden`, dragging the window near the
titlebar area would inadvertently trigger text selection. This was caused
by the `mouseDown` event being passed to the surface even when in the
titlebar region.

The fix improves the logic in `mouseDown` handler to properly handle window
dragging without triggering text selection when clicking in the titlebar
area. This provides a smoother window dragging experience while preserving
expected text selection behavior in the terminal area.

https://github.com/user-attachments/assets/cd66b644-4945-4d4b-884c-6126b52e7c1d

Fixes #5259

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -124,6 +124,12 @@ extension Ghostty {
         // A timer to fallback to ghost emoji if no title is set within the grace period
         private var titleFallbackTimer: Timer?
 
+        // The height of the titlebar in points
+        private var titlebarHeight: CGFloat {
+            guard let window = self.window else { return 0 }
+            return window.frame.height - window.contentLayoutRect.height
+        }
+
         /// Event monitor (see individual events for why)
         private var eventMonitor: Any? = nil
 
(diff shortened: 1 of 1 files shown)


macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -124,6 +124,9 @@ extension Ghostty {
         // A timer to fallback to ghost emoji if no title is set within the grace period
         private var titleFallbackTimer: Timer?
 
+        // The height of the titlebar in points, updated when window properties change
+        private var titlebarHeight: CGFloat = 0
```
mitchellh, line 128: I don't think there's any reason to cache this. It seems fairly cheap to calculate on the fly, and that removes a whole lot of complexity around keeping it up to date during various events that I suspect we could get wrong at some point.

mitchellh (changes requested):
One comment. I want to think about this approach a bit more and just explore if there's any other way to detect a titlebar area click but this seems agreeable otherwise.

macos/Sources/Ghostty/SurfaceView_AppKit.swift
liby: I agree that caching the titlebar height isn't necessary. The calculation is very lightweight and removing the cache eliminates the complexity of keeping it in sync across various window events.

Regarding alternative approaches for detecting titlebar clicks, I've thought about it and here are some considerations:

   - Using hitTest might be less reliable since the titlebar area is part of our view when using hidden titlebar style
   
   - Adding a separate view for the titlebar would add complexity without clear benefits
   
   - Using window delegate methods would be too late as we need to intercept the click before window dragging

I'm open to exploring other approaches if you have specific ideas in mind!

mitchellh:
Thanks, upon further review I think there are issues with this but more broadly I think the entire approach is actually wrong.  First, specifically: mouse move and up events are still sent to the pty while dragging is occuring. More broadly though, I think #2523 and removing the drag area at the top completely is the correct approach.

PR #5309, opened 2025-01-22
core: add `env` config option

### LINKED ISSUE #5257: Add `env` configuration to specify environment variables to launch command with
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5197

<sup>Originally posted by **networkhermit** January 17, 2025</sup>
There are occasions when users want to inject environment variables to the terminal emulator, like [the env config provided by alacritty](https://alacritty.org/config-alacritty.html#env).

Currently there are two ways in ghostty to achieve similar outcome:

1. Tweak the `command` configuration like `command = /usr/bin/env WINIT_X11_SCALE_FACTOR="1.0" "${SHELL}" -l`. Personally I think it is verbose and platform dependent.
2. Add environment variables in .zshrc/.bashrc/config.fish guarded by `if [[ "$TERM_PROGRAM" = ghostty ]];` blocks. It works but I hope to use a more non-intrusive method.

Past closed disscussion in #3776.</div>

### Proposal

Introduce an `env` configuration (similar in its naming simplicity to `command`). The configuration value is syntax `KEY=VALUE` and only one may be specified per key, but the key may be repeated. Example:

```toml
env = foo=bar
env = bar=baz
```

Just like any other repeatable string type configuration we support, setting it to empty will reset the value to the default (which in our case is... empty).

jcollie:
Fixes #5257

Specify environment variables to pass to commands launched in a terminal surface. The format is `env=KEY=VALUE`.

`env = foo=bar`
`env = bar=baz`

Setting `env` to an empty string will reset the entire map to default (empty).

`env =`

Setting a key to an empty string will remove that particular key and corresponding value from the map.

`env = foo=bar`
`env = foo=`

will result in `foo` not being passed to the launched commands. Setting a key multiple times will overwrite previous entries.

`env = foo=bar`
`env = foo=baz`

will result in `foo=baz` being passed to the launched commands.

These environment variables _will not_ be passed to commands run by Ghostty for other purposes, like `open` or `xdg-open` used to open URLs in your browser.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -531,6 +531,7 @@ pub fn init(
         var io_exec = try termio.Exec.init(alloc, .{
             .command = command,
             .env = env,
+            .env_override = config.env,
             .shell_integration = config.@"shell-integration",
             .shell_integration_features = config.@"shell-integration-features",
             .working_directory = config.@"working-directory",
--- a/src/config.zig
+++ b/src/config.zig
@@ -27,6 +27,7 @@ pub const OptionAsAlt = Config.OptionAsAlt;
 pub const RepeatableCodepointMap = Config.RepeatableCodepointMap;
 pub const RepeatableFontVariation = Config.RepeatableFontVariation;
 pub const RepeatableString = Config.RepeatableString;
+pub const RepeatableStringMap = @import("config/RepeatableStringMap.zig");
 pub const RepeatablePath = Config.RepeatablePath;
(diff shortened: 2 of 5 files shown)


mitchellh:
Looks good! I made a couple changes:

* I renamed the internal fields to `env_override` and moved the env setting to be the last thing. This lets `env` override any automatically set Ghostty env vars which can be useful.
* I removed a bunch of memory management from `RepeatableStringMap` because its part of the config contract that all allocs are arenas.
* I fixed the memory leak @rockorager pointed out

PR #5311, opened 2025-01-23
build: options to enable/disable terminfo & termcap install

### LINKED ISSUE #5253: Remove `ghostty.termcap` and `ghostty.terminfo` from install
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5229

<sup>Originally posted by **titaniumtraveler** January 19, 2025</sup>
When compiling `ghostty` with `zig build` it generates the files `share/terminfo/ghostty.termcap` and `share/terminfo/ghostty.terminfo`, which are not useful for the purpose of resolving terminal capabilities on linux.

Would it be a good idea to have an option to exclude them from the package? Or at least document for packagers what their purpose is and that they can be excluded?</div>

jcollie:
Fixes #5253

Add `-Demit-terminfo` and `-Demit-termcap` build options to enable/disable installtion of source terminfo and termcap files.

### DIFF
--- a/src/build/Config.zig
+++ b/src/build/Config.zig
@@ -55,6 +55,8 @@ emit_helpgen: bool = false,
 emit_docs: bool = false,
 emit_webdata: bool = false,
 emit_xcframework: bool = false,
+emit_terminfo: bool = false,
+emit_termcap: bool = false,
 
 /// Environmental properties
 env: std.process.EnvMap,
@@ -306,11 +308,32 @@ pub fn init(b: *std.Build) !Config {
         break :emit_docs path != null;
     };
 
+    config.emit_terminfo = b.option(
+        bool,
+        "emit-terminfo",
+        "Install Ghostty terminfo source file",
+    ) orelse switch (target.result.os.tag) {
+        .windows => true,
+        else => switch (optimize) {
+            .Debug => true,
+            .ReleaseSafe, .ReleaseFast, .ReleaseSmall => false,
+        },
+    };
+
+    config.emit_termcap = b.option(
+        bool,
+        "emit-termcap",
+        "Install Ghostty termcap file",
+    ) orelse false;
+
     config.emit_webdata = b.option(
         bool,
         "emit-webdata",
         "Build the website data for the website.",
-    ) orelse false;
+    ) orelse switch (optimize) {
+        .Debug => true,
+        .ReleaseSafe, .ReleaseFast, .ReleaseSmall => false,
+    };
 
     config.emit_xcframework = b.option(
         bool,
--- a/src/os/resourcesdir.zig
+++ b/src/os/resourcesdir.zig
@@ -21,7 +21,10 @@ pub fn resourcesDir(alloc: std.mem.Allocator) !?[]const u8 {
 
     // This is the sentinel value we look for in the path to know
     // we've found the resources directory.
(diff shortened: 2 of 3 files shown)


PACKAGING.md
```diff
@@ -118,6 +118,14 @@ relevant to package maintainers:
   version, glibc, etc. Run `zig targets` to a get a full list of available
   targets.
 
+- `-Demit-terminfo`: Install the Ghostty terminfo source file. Defaults to
+  `false` except on Windows where it defaults to `true`. Except for Windows,
+  this file is not required for proper operation as compiled versions of the
+  terminfo file are installed.
+
+- `-Demit-termcap`: Install the Ghostty termcap file. This may be useful for
```
mitchellh, line 126: No need to document these in PACKAGING.md. We only document the most relevant configs. It's unlikely a packager will ever care here since our defaults are so sane. The goal of PACKAGING.md is to only give the most relevant actionable info so a packager succeeds, not to be a reference.

src/build/Config.zig
```diff
@@ -306,6 +308,21 @@ pub fn init(b: *std.Build) !Config {
         break :emit_docs path != null;
     };
 
+    config.emit_terminfo = b.option(
+        bool,
+        "emit-terminfo",
+        "Install Ghostty terminfo source file",
+    ) orelse switch (target.result.os.tag) {
+        .windows => true,
+        else => false,
```
mitchellh, line 317: I think this and termcap both should emit for Debug builds. It could be useful for dev builds.

src/os/resourcesdir.zig
```diff
@@ -21,7 +21,16 @@ pub fn resourcesDir(alloc: std.mem.Allocator) !?[]const u8 {
 
     // This is the sentinel value we look for in the path to know
     // we've found the resources directory.
-    const sentinel = "terminfo/ghostty.termcap";
+    const sentinels = switch (comptime builtin.target.os.tag) {
```
mitchellh, line 24: Why was introducing multiple necessary? Shouldn't a single work?

I understand the platform difference just not why this is now an array.

mitchellh (changes requested):
Thanks, some asks

PACKAGING.md
jcollie: Removed.

src/build/Config.zig
jcollie: Done.

src/os/resourcesdir.zig
jcollie: Belt and suspenders I guess. I've switched back to a single file.

src/os/resourcesdir.zig
```diff
@@ -21,7 +21,10 @@ pub fn resourcesDir(alloc: std.mem.Allocator) !?[]const u8 {
 
     // This is the sentinel value we look for in the path to know
     // we've found the resources directory.
-    const sentinel = "terminfo/ghostty.termcap";
+    const sentinel = switch (comptime builtin.target.os.tag) {
+        .windows => "terminfo/ghostty.terminfo",
+        else => "terminfo/x/xterm-ghostty",
```
qwerasd205, line 26: This sentinel is incorrect and causes a complete failure to find the directory, on macOS at least-- that `x` is not right. Perhaps we should just put a dedicated sentinel file in the directory.

PR #5319, opened 2025-01-23
bash: handle additional command arguments

jparise:
A '-' or '--' argument signals the end of bash's own options. All remaining arguments are treated as filenames and arguments. We shouldn't perform any additional argument processing once we see this signal.

We could also assume a non-interactive shell session in this case unless the '-i' (interactive) shell option has been explicitly specified, but let's wait on that until we know that doing so would solve a real user problem (and avoid any false negatives).

### DIFF
--- a/src/termio/shell_integration.zig
+++ b/src/termio/shell_integration.zig
@@ -203,8 +203,6 @@ test "setup features" {
 /// our script's responsibility (along with disabling POSIX
 /// mode).
 ///
-/// This approach requires bash version 4 or later.
-///
 /// This returns a new (allocated) shell command string that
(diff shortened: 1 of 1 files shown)


src/termio/shell_integration.zig
```diff
@@ -203,8 +203,6 @@ test "setup features" {
 /// our script's responsibility (along with disabling POSIX
 /// mode).
 ///
-/// This approach requires bash version 4 or later.
```
jparise, line 206: This was never quite true. The problem was with Apple's bash version of bash 3.2 (#2966).

PR #5324, opened 2025-01-23
apprt/gtk: when text is committed, end the preedit state

### LINKED ISSUE #3567: GTK: Spanish dead key input breaks input buffer
I've tried to write with an spanish keyboard layout, and every time I pressed a [dead key](https://en.wikipedia.org/wiki/Dead_key) (like: `´` or `¨`) the input buffer stops working.

Image

The same thing happends using the 'Compose Key' option of the keyboard configuration on Ubuntu.

**Steps to reproduce:**

1. Change your keyboard layout to other with dead keys, like Spanish or French.
2. Write something with a dead key, like `á` or `ä`.
3. Try to continue writing.

**Bug:**

- After pressing a dead key, the characters introduced will be overwritten one after another and special keys like Enter or Backspace won't work as expected.
- The bug only affects the tab in which you've pressed the dead key.

```
Ubuntu 24.04.1 LTS

❯ ghostty --version
Ghostty 1.0.0

Version
  - version: 1.0.0
  - channel: stable
Build Config
  - Zig version: 0.13.0
  - GTK version:
    build      : 4.14.2
    runtime    : 4.14.2
  - libadwaita : enabled
    build      : 1.5.0
    runtime    : 1.5.0
```

mitchellh:
Fixes #3567

ibus 1.5.29 doesn't trigger a preedit end state when text is committed. This is fixed in ibus 1.5.30, but we need to handle this case for older versions which are shipped on LTS distributions such as Ubuntu.

Every other input method engine I've tried thus far also triggers a preedit end state when text is committed, and none would expect preedit to continue after text is committed. So I think it's safe to assume that this is the expected behavior.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -1961,6 +1961,12 @@ fn gtkInputCommit(
     // Committing ends composing state
     self.im_composing = false;
 
+    // End our preedit state. Well-behaved input methods do this for us
+    // by triggering a preedit-end event but some do not (ibus 1.5.29).
+    self.core_surface.preeditCallback(null) catch |err| {
+        log.err("error in preedit callback err={}", .{err});
+    };
+
     // Send the text to the core surface, associated with no key (an
     // invalid key, which should produce no PTY encoding).
     _ = self.core_surface.keyCallback(.{


tomaspinho:
Hey @mitchellh , thanks a lot for Ghostty. :heart: 

1.1.0 breaks dead keys on ibus `1.5.31`. The initial modifier character works but the final character is inputted instead, with no modifier. Example:

```
`  + a = a
´ + e = e
¨ + space = space (instead of ")

and so on...
```

I had to downgrade to 1.0.1. I'm running on Arch with the following:

```
Ghostty 1.0.1-arch

Version
  - version: 1.0.1-arch
  - channel: tip
Build Config
  - Zig version: 0.13.0
  - GTK version:
    build      : 4.16.12
    runtime    : 4.16.12
  - libadwaita : enabled
    build      : 1.6.3
    runtime    : 1.6.4
  - libX11     : enabled
  ```

ibus versions:

```
pacman -Q | grep ibus
ibus 1.5.31-2
libibus 1.5.31-2
```

Other GTK4/Adwaita apps are working fine, so are browsers, etc. I'm running on Hyprland, but I assume this happens in other places too. I'll try GNOME and report back.

tomaspinho:
So now I'm convinced the culprit of this is actually https://github.com/ghostty-org/ghostty/commit/eb21a58aa4dc59dd2a0fbd026cf7cd842f0f07ed

Some assumptions are made there regarding indirect interactions with something that is not running - ibus.

4lm1r:
In my case, under Archlinux, I realized that pressing ALT GR before worked for me.

PR #5326, opened 2025-01-23
gtk: implement custom audio bell

jcollie:
Fixes #2710 

Adds support for custom audio bell notifications.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -2408,9 +2408,9 @@ pub fn ringBell(self: *Surface) !void {
     if (features.audio) audio: {
         // Play a user-specified audio file.
 
-        const pathname, const optional = switch (self.app.config.@"bell-audio-path" orelse break :audio) {
-            .optional => |path| .{ path, true },
-            .required => |path| .{ path, false },
+        const pathname, const required = switch (self.app.config.@"bell-audio-path" orelse break :audio) {
+            .optional => |path| .{ path, false },
+            .required => |path| .{ path, true },
         };
 
         const volume = std.math.clamp(self.app.config.@"bell-audio-volume", 0.0, 1.0);
@@ -2418,7 +2418,7 @@ pub fn ringBell(self: *Surface) !void {
         std.debug.assert(std.fs.path.isAbsolute(pathname));
         const media_file = gtk.MediaFile.newForFilename(pathname);
 
-        if (!optional) {
+        if (required) {
             _ = gobject.Object.signals.notify.connect(
                 media_file,
                 ?*anyopaque,
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -2413,13 +2413,7 @@ pub fn ringBell(self: *Surface) !void {
             .required => |path| .{ path, false },
         };
 
-        const volume: f64 = @min(
-            @max(
-                0.0,
-                self.app.config.@"bell-audio-volume",
-            ),
-            1.0,
-        );
+        const volume = std.math.clamp(self.app.config.@"bell-audio-volume", 0.0, 1.0);
 
         std.debug.assert(std.fs.path.isAbsolute(pathname));
         const media_file = gtk.MediaFile.newForFilename(pathname);
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -2405,6 +2405,47 @@ pub fn ringBell(self: *Surface) !void {
         surface.beep();
     }
 
+    if (features.audio) audio: {
+        // Play a user-specified audio file.
+
+        const pathname, const optional = switch (self.app.config.@"bell-audio-path" orelse break :audio) {
+            .optional => |path| .{ path, true },
+            .required => |path| .{ path, false },
+        };
+
+        const volume: f64 = @min(
+            @max(
+                0.0,
+                self.app.config.@"bell-audio-volume",
+            ),
+            1.0,
+        );
+
+        std.debug.assert(std.fs.path.isAbsolute(pathname));
+        const media_file = gtk.MediaFile.newForFilename(pathname);
+
+        if (!optional) {
+            _ = gobject.Object.signals.notify.connect(
+                media_file,
+                ?*anyopaque,
+                gtkStreamError,
+                null,
+                .{ .detail = "error" },
+            );
+        }
+        _ = gobject.Object.signals.notify.connect(
+            media_file,
+            ?*anyopaque,
+            gtkStreamEnded,
+            null,
+            .{ .detail = "ended" },
+        );
+
+        const media_stream = media_file.as(gtk.MediaStream);
+        media_stream.setVolume(volume);
+        media_stream.play();
+    }
+
     // Mark tab as needing attention
     if (self.container.tab()) |tab| tab: {
         const page = window.notebook.getTabPage(tab) orelse break :tab;
@@ -2412,4 +2453,28 @@ pub fn ringBell(self: *Surface) !void {
         // Need attention if we're not the currently selected tab
         if (page.getSelected() == 0) page.setNeedsAttention(@intFromBool(true));
     }
+}
+
+/// Handle a stream that is in an error state.
+fn gtkStreamError(media_file: *gtk.MediaFile, _: *gobject.ParamSpec, _: ?*anyopaque) callconv(.c) void {
+    const path = path: {
+        const file = media_file.getFile() orelse break :path null;
+        break :path file.getPath();
+    };
+    defer if (path) |p| glib.free(p);
+
+    const media_stream = media_file.as(gtk.MediaStream);
+    const err = media_stream.getError() orelse return;
+
+    log.warn("error playing bell from {s}: {s} {d} {s}", .{
+        path orelse "<<unknown>>",
+        glib.quarkToString(err.f_domain),
+        err.f_code,
+        err.f_message orelse "",
+    });
+}
+
+/// Stream is finished, release the memory.
+fn gtkStreamEnded(media_file: *gtk.MediaFile, _: *gobject.ParamSpec, _: ?*anyopaque) callconv(.c) void {
+    media_file.unref();
 }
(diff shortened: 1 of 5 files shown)


mitchellh:
Just want to note I'm going to not review this pending further discussion (I know there's a discussion open). One thing I noticed right away though if you care is this uses the old `@hasDecl` method of surface callbacks and we should instead switch to the apprt actions system.

NovaViper:
Gave the PR and test and the bell works perfectly on NixOS 25.05 unstable (built from source)

NovaViper (approved):

include/ghostty.h
```diff
@@ -600,6 +600,7 @@ typedef enum {
   GHOSTTY_ACTION_COLOR_CHANGE,
   GHOSTTY_ACTION_RELOAD_CONFIG,
   GHOSTTY_ACTION_CONFIG_CHANGE,
+  GHOSTTY_ACTION_BELL,
```
tristan957, line 603: Feels like we should be more consistent with action names.

```
TOGGLE_SPLIT_ZOOM
PWD
RELOAD_CONFIG
BELL
```

Maybe `RING_BELL` would be more consistent?

src/config/Config.zig
```diff
@@ -2309,6 +2309,30 @@ term: []const u8 = "xterm-ghostty",
...
+///   * `system` - Use a system function to play an audible sound. This differs
+///     from the `audio` feature in that the sound played is not customizable
+///     from within Ghostty. Your system may allow for the sound to be
+///     customized externally to Ghostty.
+///   * `audio` - Play a custom sound. (GTK only).
+///
+/// Example: `audio`, `no-audio`, `system`, `no-system`:
+///
+/// By default, no bell features are enabled.
+@"bell-features": BellFeatures = .{},
```
00-kat, line 2329: Should there be a note that this is currently only supported on Linux?

src/config/Config.zig
```diff
@@ -2309,6 +2309,30 @@ term: []const u8 = "xterm-ghostty",
...
+/// default value is used, so you must explicitly disable features you don't
+/// want.
+///
+/// Available features:
+///
+///   * `system` - Use a system function to play an audible sound. This differs
+///     from the `audio` feature in that the sound played is not customizable
+///     from within Ghostty. Your system may allow for the sound to be
+///     customized externally to Ghostty.
+///   * `audio` - Play a custom sound. (GTK only).
```
00-kat, line 2324: > (GTK only)

(continuation of previous comment) and if so, should this be removed?

src/apprt/gtk/Surface.zig
```diff
@@ -1264,6 +1265,81 @@ pub fn showDesktopNotification(
...
+                                    log.warn("unable to determine media config subdir: {}", .{err});
+                                    break :audio;
+                                },
+                                filename,
+                            }) catch |err| {
+                                log.warn("unable to allocate space for bell audio pathname: {}", .{err});
+                                break :audio;
+                            };
+                    };
+                    std.fs.accessAbsoluteZ(pathname, .{ .mode = .read_only }) catch {
```
00-kat, line 1302: Does GTK not return an error if the media file can't be found?  This would be a possible TOCTOU bug I think.

00-kat (approved):

src/apprt/gtk/Surface.zig
jcollie: I removed the access check in the Zig code and enhanced the error message that you get from GTK by at least including the filename. But you still get some pretty obscure error messages:

```
error(gtk_surface): error playing bell from /dev/null: gst-play-error-quark 0 Error from element /GstPlayBin3:playbin3/GstURIDecodeBin3:uridecodebin3/GstURISourceBin:urisourcebin0/GstTypeFindElement:typefindelement0: Could not determine type of stream.
Stream doesn't contain enough data.
../plugins/elements/gsttypefindelement.c(1012): gst_type_find_element_chain_do_typefinding (): /GstPlayBin3:playbin3/GstURIDecodeBin3:uridecodebin3/GstURISourceBin:urisourcebin0/GstTypeFindElement:typefindelement0:
Can't typefind stream
```

src/apprt/gtk/Surface.zig
```diff
@@ -1264,6 +1265,90 @@ pub fn showDesktopNotification(
...
+                    media_stream.play();
+                },
+            }
+        }
+    }
+}
+
+/// Handle a stream that is in an error state.
+fn gtkStreamError(media_stream: *gtk.MediaStream, _: *gobject.ParamSpec, _: ?*anyopaque) callconv(.C) void {
+    const media_file = gobject.ext.cast(gtk.MediaFile, media_stream) orelse return;
```
pluiedev, line 1336: FYI you can avoid the cast by just passing in the media file directly above without `.as(gtk.MediaStream)`, then the first parameter would just be a `gtk.MediaFile`

src/apprt/surface.zig
```diff
@@ -81,6 +81,9 @@ pub const Message = union(enum) {
     /// The terminal has reported a change in the working directory.
     pwd_change: WriteReq,
 
+    /// Bell
+    bell: void,
```
pluiedev, line 85: `: void` isn't necessary for a tagged union variant without a value anymore. We should probably clean up all the `: void`s inside the apprt actions at some point

src/config/Config.zig
```diff
@@ -2309,6 +2309,30 @@ term: []const u8 = "xterm-ghostty",
...
+///
+/// Example: `audio`, `no-audio`, `system`, `no-system`:
+///
+/// By default, no bell features are enabled.
+@"bell-features": BellFeatures = .{},
+
+/// If `audio` is an enabled bell feature, this is a path to an audio file.
+/// If the path is not absolute, it is considered relative to the `media`
+/// subdirectory of the Ghostty configuration directory.
```
pluiedev, line 2333: I'm really not sure whether we should do this. I think people should just always specify absolute paths so that it works regardless of Ghostty configs

src/apprt/gtk/Surface.zig
```diff
@@ -1264,6 +1265,90 @@ pub fn showDesktopNotification(
     c.g_application_send_notification(g_app, body.ptr, notification);
 }
 
+/// Handle bell features.
+pub fn bell(self: *Surface) void {
+    inline for (std.meta.fields(configpkg.Config.BellFeatures.Features)) |field| {
+        const feature = std.meta.stringToEnum(configpkg.Config.BellFeatures.Features, field.name) orelse unreachable;
+        const enabled = @field(self.app.config.@"bell-features", field.name);
+        if (enabled) {
+            switch (feature) {
```
pluiedev, line 1274: Honestly I think we should just do it the "dumb" way and use individual if clauses... This is a bit too hard to read even for me

pluiedev (changes requested):

src/apprt/gtk/Surface.zig
jcollie: Actually, you can't (I tried). Since the property definition is on `gtk.MediaStream` the signal is expecting a function with a `*gtk.MediaStream` as the first parameter.

src/config/Config.zig
gpanders: How does it work for `config-file` or `custom-shader`? IIRC in both cases a relative path resolves relative to the directory of the config file, which seems reasonable to do here too

jcollie: Hmm it's all a bit convoluted but it looks like `config-file`, `custom-shader`, and `gtk-custom-css` are always treated relative to the directory of the file that they are referenced from, or in the case of CLI args, the current working directory. Re-using that logic I think would be fine.

src/apprt/surface.zig
jcollie: Fixed.

src/config/Config.zig
jcollie: No, I think that mentioning support should be on a per-feature basis as there may be features that are not supported by all platforms.

src/apprt/gtk/Surface.zig
jcollie: Fixed.

src/config/Config.zig
jcollie: I pushed PR #6622 to refactor RepeatablePath and add NonRepeatablePath (because I don't think it makes any sense to have multiple audio files). This way `bell-audio` has the same behavior as other configuration entries with regards to relative directories or expanding `~/`.

src/apprt/gtk/Surface.zig
pluiedev: That doesn't make much sense... child classes inherit all properties from their parent class. For example, you can still listen for changes of the `GtkWindow:decorated` property on an AdwApplicationWindow since it subclasses GtkWindow

jcollie: Well, I coulda _swore_ that this broke earlier, but ¯\_(ツ)_/¯. Fixed now.

jcollie:
Converted to draft until #6622 is merged.

src/apprt/action.zig
```diff
@@ -244,6 +244,7 @@ pub const Action = union(Key) {
     /// Closes the currently focused window.
     close_window,
 
+    /// The system has received a request to activate the bell.
```
tristan957, line 247:
```suggestion
    /// The system has received a request to ring the bell.
```

tristan957 (approved):

include/ghostty.h
jcollie: #7087 switched to RING_BELL

pluiedev (changes requested):

src/apprt/gtk/Surface.zig
```diff
@@ -2453,6 +2453,47 @@ pub fn ringBell(self: *Surface) !void {
...
+        const volume: f64 = @min(
+            @max(
+                0.0,
+                self.app.config.@"bell-audio-volume",
+            ),
+            1.0,
+        );
+
+        std.debug.assert(std.fs.path.isAbsolute(pathname));
+        const media_file = gtk.MediaFile.newForFilename(pathname);
```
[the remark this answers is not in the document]
jcollie: That might be good for a follow-up. The audio files _should_ be short so I think just loading them into memory when the config is reloaded would be relatively trivial.

pluiedev (approved):
This has been waiting for long enough

mitchellh (approved):
I’m good with this too. We can do the macOS side another time.

PR #5327, opened 2025-01-23
Add draggable surfaces

### LINKED ISSUE #5259: macOS: Dragging with `macos-titlebar-style=hidden` should not invoke selection

korikhin:
1. Fixes #5259.
2. Each surface now has its own draggable titlebar region except the topmost ones when the titlebar is visible.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -579,10 +579,34 @@ extension Ghostty {
             return true
         }
 
-        override func mouseDown(with event: NSEvent) {
-            guard let surface = self.surface else { return }
+        private func handleSurfaceMouseDown(_ surface: ghostty_surface_t, _ event: NSEvent) {
             let mods = Ghostty.ghosttyMods(event.modifierFlags)
             ghostty_surface_mouse_button(surface, GHOSTTY_MOUSE_PRESS, GHOSTTY_MOUSE_LEFT, mods)
+        }
+
+        override func mouseDown(with event: NSEvent) {
+            guard let surface = self.surface else { return }
+            guard let window = self.window else {
+                handleSurfaceMouseDown(surface, event)
+                return
+            }
+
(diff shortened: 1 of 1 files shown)


mitchellh:
This looks... very similar to #5286. I'll work through that PR to get merged in. Thank you! ❤️

mitchellh:
It seems like there are additional cases you added here, can you explain that?

korikhin:
Yeah, it is **exactly** the same. The only difference I was about to point out in the original PR is that we get a redundant region when `macos-titlebar-style` is **not** `hidden`. If it is not a bad thing, you've already closed the PR. 🙂

mitchellh:
I actually think this whole approach (not your fault) is wrong. I'm going to comment on the other PR to note.

PR #5336, opened 2025-01-24
macos: fix finding the resources dir

jcollie:
Fix breakage on macOS from #5311


mitchellh:
Reverted #5311. We can address this back in the original diff probably.

PR #5340, opened 2025-01-24
build: options to enable/disable terminfo & termcap install (take 2)

### LINKED ISSUE #5253: Remove `ghostty.termcap` and `ghostty.terminfo` from install

jcollie:
Fixes #5253

Add -Demit-terminfo and -Demit-termcap build options to enable/disable installation of source terminfo and termcap files.

Replacement for #5311

### DIFF
--- a/src/build/GhosttyResources.zig
+++ b/src/build/GhosttyResources.zig
@@ -16,15 +16,6 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyResources {
 
     // Terminfo
     terminfo: {
-        const mkdir_step = RunStep.create(b, "make share/terminfo directory");
-        switch (cfg.target.result.os.tag) {
-            // windows mkdir shouldn't need "-p"
-            .windows => mkdir_step.addArgs(&.{"mkdir"}),
-            else => mkdir_step.addArgs(&.{ "mkdir", "-p" }),
-        }
-        mkdir_step.addArg(b.fmt("{s}/share/terminfo", .{b.install_path}));
-        try steps.append(&mkdir_step.step);
-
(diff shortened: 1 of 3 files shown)


src/build/GhosttyResources.zig
```diff
@@ -16,16 +16,29 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyResources {
...
         // Write it
         var wf = b.addWriteFiles();
-        const src_source = wf.add("share/terminfo/ghostty.terminfo", str.items);
-        const src_install = b.addInstallFile(src_source, "share/terminfo/ghostty.terminfo");
-        try steps.append(&src_install.step);
+        const source = wf.add("ghostty.terminfo", str.items);
+
+        if (cfg.emit_terminfo) {
+            const source_install = b.addInstallFile(source, "share/terminfo/ghostty.terminfo");
+            source_install.step.dependOn(&mkdir_step.step);
```
mitchellh, line 39: I think only the `copy` step that uses `cp` should depend on `mkdir`. `addInstallFile` does create the directory but the problem is since we're using `cp -r` (for the reasons stated in the comment) the directory isn't made.

PR #4446, opened 2025-01-02
docs: Cleanup and improve the actions documentation with arguments and examples

erf:
Improve the documentation of [actions](https://ghostty.org/docs/config/keybind/reference#text) generated from `Binding.zig` to make it more clear and consistent.

Key improvements:
- Added consistent documentation format for all actions
- Added clear descriptions of arguments and their values
- Included complete examples using full `keybind = key=action:value` syntax
- Cleaned and simplified some action descriptions
- Made platform limitations (macOS, Linux) explicitly visible

Related to this [discussion](https://github.com/ghostty-org/ghostty/discussions/3404).

Note: The documentation was improved with the help of AI (Claude and Copilot)

PR #5344, opened 2025-01-24
Docs: improve doc structure for action bindings with args and examples

erf:
Related to #4446 , as suggested i've made a smaller chunk of updates for the action binding docs, including arguments and examples. I did this only for enum types this time.

I've also included the argument type name in the Argument (e.g. AdjustSelection).

Not sure if we should include the keybindings or not in the example?

I also added a format of some unrelated code change.


src/input/Binding.zig
```diff
@@ -280,8 +280,15 @@ pub const Action = union(enum) {
     scroll_page_fractional: f32,
     scroll_page_lines: i16,
 
-    /// Adjust an existing selection in a given direction. This action
-    /// does nothing if there is no active selection.
+    /// Adjust the current selection in a given direction. Does nothing if no 
+    /// selection exists.
+    ///
+    /// Arguments:
+    ///   - AdjustSelection: left, right, up, down, page_up, page_down, home, end,
```
mitchellh, line 287: I don't think the type is helpful to end users. If you look at our config docs we tend to just give the valid values.

mitchellh (changes requested):
Thanks, much better 😄 One note.

PR #5361, opened 2025-01-24
macos: autohide dock if quick terminal would conflict with it

mitchellh:
Fixes #5328

The dock sits above the level of the quick terminal, and the quick terminal frame typical includes the dock. Hence, if the dock is visible and the quick terminal would conflict with it, then part of the terminal is obscured.

This commit makes the dock autohide if the quick terminal would conflict with it. The autohide is disabled when the quick terminal is closed.

We can't set our window level above the dock, as this would prevent things such as input methods from rendering properly in the quick terminal window.

iTerm2 (the only other macOS terminal I know of that supports a dropdown mode) frames the terminal around the dock. I think this looks less aesthetically pleasing and I prefer autohiding the dock instead.

We can introduce a setting to change this behavior if desired later.

Additionally, this commit introduces a mechanism to safely set app-global presentation options from multiple sources without stepping on each other.

## Demo

https://github.com/user-attachments/assets/1f5bb945-dca4-49e7-8bcb-95b55524cfb5

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -69,6 +69,8 @@
 		A59FB5CF2AE0DB50009128F3 /* InspectorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A59FB5CE2AE0DB50009128F3 /* InspectorView.swift */; };
 		A59FB5D12AE0DEA7009128F3 /* MetalView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A59FB5D02AE0DEA7009128F3 /* MetalView.swift */; };
 		A5A1F8852A489D6800D1E8BC /* terminfo in Resources */ = {isa = PBXBuildFile; fileRef = A5A1F8842A489D6800D1E8BC /* terminfo */; };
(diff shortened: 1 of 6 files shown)


mitchellh:
Received approval from original issue reporter, going to give it a go!

milgrim:
Works great when a Ghostty window is active, but not when other applications are in focus:

https://github.com/user-attachments/assets/a341214f-2b11-49d9-8fa7-f9d486734e41

mitchellh:
@milgrim #5363

bluekeybo:
> We can introduce a setting to change this behavior if desired later.

I keep the dock on the left, and prefer it to not hide. All my windows fill the screen up to the dock. If set to position "top", the quick terminal window should start right where the dock ends, and drop from the top, without hiding it.

dzirtusss:
Looks weird now for me as well, I keep the dock on the right. Older behavior to just cover the dock w/o hiding was nicer. IMO, if possible, better this be an option.

mitchellh:
> Older behavior to just cover the dock w/o hiding was nicer. IMO, if possible, better this be an option.

Please read the background discussion on why this isn't an option. This breaks IME input (i.e. Japanese).

xavdid:
@mitchellh Are you interested / willing to add a setting for the dock-covering behavior?

I read all the discussions and I appreciate that covering the dock breaks IME input. But, as someone who doesn't type Japanese characters in my terminal and has _really_ gotten used to the exact way the iTerm2 quick terminal works, the discrepancy is making the switch difficult.

PR #5363, opened 2025-01-25
macos: hide dock globally if the dock conflicts

mitchellh:
Related to #5361

The fix in 5361 wasn't sufficient since it only applied if our app was in the foreground. Our quick terminal is a non-activating NSPanel to allow it to work on any space (fullscreen included). This means that Ghostty doesn't become the active app when the quick terminal is shown and another app is in the foreground.

To work around this, we now hide the dock globally when the quick terminal is shown AND the dock is in a conflicting position. We restore this state when the quick terminal is hidden, loses focus, or Ghostty is quit.


milgrim:
Works as expected now, thanks!

taylrfnt:
It looks like this PR introduced some potentially sticky Dock visibility settings where Ghostty overrides a user's Dock visibility preference without resetting it on Quick Terminal close - see #5617 for reproduction, bisecting, etc.

PR #5367, opened 2025-01-25
fix(gtk): detect modifier keys without cursor movement

### LINKED ISSUE #5191: regression(gtk): Pressing Ctrl when mouse is over a hyperlink does not detect the link until cursor is moved
Same as https://github.com/ghostty-org/ghostty/issues/2114, but GTK only. That issue was resolved but has since regressed.

Bisected to ed81b62ec24790fea79877d2e2124d82271ee3df.

## Steps to reproduce

1. `echo 'https://example.com'`
2. Hover the mouse over the `https://example.com` text and press and hold the Control key.

## Expected behavior

The cursor should immediately change to the "pointer" shape, the link should be highlighted, and the URL preview should appear in the lower left corner of the window.

## Actual behavior

Nothing happens until the mouse is moved by at least one cell (at which point all of the above does happen).

ghost:
This fixes a regression where pressing Ctrl while hovering over a link wouldn't detect
the link until the cursor moved.

Tested on both x11 / wayland

Fixes #5191

### DIFF
--- a/src/apprt/gtk/key.zig
+++ b/src/apprt/gtk/key.zig
@@ -49,6 +49,22 @@ pub fn translateMods(state: c.GdkModifierType) input.Mods {
     return mods;
 }
 
+pub fn deviceMods(device: ?*c.GdkDevice, gtk_mods: c.GdkModifierType) ?input.Mods {
+    if (device) |dev| {
+        const device_mods = c.gdk_device_get_modifier_state(dev);
+        if (device_mods != gtk_mods) {
+            return input.Mods{
+                .shift = (device_mods & c.GDK_SHIFT_MASK) != 0,
+                .ctrl = (device_mods & c.GDK_CONTROL_MASK) != 0,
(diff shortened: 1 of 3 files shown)


mitchellh (changes requested):
Since your changes are both GTK-specific and not window protocol specific, I believe it'd make more sense to put them in `apprt/gtk/key.zig`. What do you think?

I haven't vetted the actual approach yet, just something I immediately noticed.

nnyyxxxx:
> Since your changes are both GTK-specific and not window protocol specific, I believe it'd make more sense to put them in `apprt/gtk/key.zig`. What do you think?

yes, that makes sense. I'll start working on that

nnyyxxxx:
Any ETA on when this will get merged? or are there still issues with this MR.

mitchellh:
Superseded by #5420, an alternate approach I took that I think will result in less input latency since we only run logic we're already been running. If you see any problems with this let me know. Thanks!

PR #5373, opened 2025-01-25
fix: cmd+triple-click select all command output when first line wraps

### LINKED ISSUE #4693: cmd+triple_click doesn't select full contents of the output for certain screen sizes
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4666

<sup>Originally posted by **msharran** January  5, 2025</sup>
# Issue

In my Macbook pro 14 inch (Apple M1 Pro Chip) with font size of 15 by default, cmd+triple_click selection leaves out the first line. Refer below image.
 
Screenshot 2025-01-06 at 9 29 38 AM

But when I do cmd+minus till it wraps the contents in a single line, then it works as expected. Refer below image.

Screenshot 2025-01-06 at 9 30 09 AM

# Feature used is Shell Integration

Link: https://ghostty.org/docs/features/shell-integration#features

Refer selected portion in the screenshot below

image

</div>

### Reproduction

I haven't narrowed down a reproduction yet but I can easily believe and visualize how there would be off-by-N errors here when it comes to the prompt marking. Converting to issue for visibility.

dpatterbee:
I found this bug was easily reproduced with any command that wrapped to multiple rows on the first line of its output. The cause is that we stop searching for rows once we reach the first one where `row.semantic_prompt = .command`, which means that we reach the bottom line of wrapped output and stop there.

This PR makes it so that we continue iterating until we reach a row where `semantic_prompt != .command` and then return the previous one (or the last one if we run out of rows).

I also updated the test cases to include this.

I considered that this bug would also be avoided if we didn't propagate the `command` semantic prompt to additional rows on wrapped lines, but I don't know enough about the shell integration to make a call on that.

Closes #4693

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -2596,6 +2596,15 @@ pub fn selectOutput(self: *Screen, pin: Pin) ?Selection {
         while (it.next()) |p| {
             it_prev = p;
             const row = p.rowAndCell().row;
+            switch (row.semantic_prompt) {
+                .command => break,
+
+                .unknown,
+                .prompt,
+                .prompt_continuation,
+                .input,
+                => {},
+            }
             if (row.semantic_prompt == .command) {
                 break;
             }
@@ -2606,8 +2615,14 @@ pub fn selectOutput(self: *Screen, pin: Pin) ?Selection {
         // yield the previous row.
         while (it.next()) |p| {
             const row = p.rowAndCell().row;
-            if (row.semantic_prompt != .command) {
-                break :boundary it_prev;
+            switch (row.semantic_prompt) {
+                .command => {},
+
+                .unknown,
+                .prompt,
+                .prompt_continuation,
+                .input,
+                => break :boundary it_prev,
             }
             it_prev = p;
         }
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -2591,13 +2591,24 @@ pub fn selectOutput(self: *Screen, pin: Pin) ?Selection {
     const start: Pin = boundary: {
         var it = pin.rowIterator(.left_up, null);
         var it_prev = pin;
+
+        // First, iterate until we find the first line of command output
+        while (it.next()) |p| {
+            it_prev = p;
+            const row = p.rowAndCell().row;
+            if (row.semantic_prompt == .command) {
+                break;
+            }
+        }
+
+        // Because the first line of command output may span multiple visual rows we must now
+        // iterate until we find the first row of anything other than command output and then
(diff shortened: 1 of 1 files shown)


src/terminal/Screen.zig
```diff
@@ -2591,13 +2591,24 @@ pub fn selectOutput(self: *Screen, pin: Pin) ?Selection {
...
         var it_prev = pin;
+
+        // First, iterate until we find the first line of command output
         while (it.next()) |p| {
+            it_prev = p;
             const row = p.rowAndCell().row;
-            switch (row.semantic_prompt) {
-                .command => break :boundary p,
-                else => {},
+            if (row.semantic_prompt == .command) {
```
mitchellh, line 2599: We should use an exhaustive `switch` here. Even when we're matching only one enum element, I prefer to use an exhaustive switch so that the compiler can error and tell us all the places a switch is used when we add an element. This makes it easy to consider if we need to change our behavior.

(I know the old code didn't use an exhaustive switch, that was a mistake)

src/terminal/Screen.zig
```diff
@@ -2591,13 +2591,24 @@ pub fn selectOutput(self: *Screen, pin: Pin) ?Selection {
...
+                break;
             }
+        }
 
+        // Because the first line of command output may span multiple visual rows we must now
+        // iterate until we find the first row of anything other than command output and then
+        // yield the previous row.
+        while (it.next()) |p| {
+            const row = p.rowAndCell().row;
+            if (row.semantic_prompt != .command) {
```
mitchellh, line 2609: Same here

mitchellh (changes requested):
Thank you! This is looking good, just a simple request.

src/terminal/Screen.zig
dpatterbee: Done :)

jparise:
> I considered that this bug would also be avoided if we didn't propagate the `command` semantic prompt to additional rows on wrapped lines, but I don't know enough about the shell integration to make a call on that.

One thing that can get tricky is, in our current row-based approach, the last prompt marker is dominant, overwriting any other markers that were emitted for that row. I'm not sure if that complicates the `command` case, but it is relevant for multiline prompts and `input` regions.

I wrote some related notes in 63747d1e228021e6b09d670ca72b3604f5ffdb74:

> Going forward, we should revisit our semantic prompt implementation. Our
row-based approach is too limiting; lines can have multiple markers, and
markers should be recorded with their full coordinates so they can form
ranges.
>
> See: https://per.bothner.com/blog/2019/shell-integration-proposal/

PR #5397, opened 2025-01-27
macos: add titled-visible-menu option to macos-non-native-fullscreen

sassanh:
Non-native fullscreen has certain limitations at the moment regarding being truly fullscreen (taking all screen surface) and losing no functionality when activated. Currently, tab functionality is lost when non-native fullscreen is activated.

This commit introduces the `titled-visible-menu` mode for macOS non-native fullscreen. Like `visible-menu` mode, it hides the dock and uses its surface, leaving the menubar visible. This mode makes full use of the screen (except for the menubar) while retaining the tabbar’s functionality.

While a truly fullscreen non-native mode without feature loss is ideal, this implementation provides a functional alternative in the meantime.


sassanh:
This is how it looks:
image

include/ghostty.h
```diff
@@ -412,6 +412,7 @@ typedef enum {
   GHOSTTY_FULLSCREEN_NATIVE,
   GHOSTTY_FULLSCREEN_NON_NATIVE,
   GHOSTTY_FULLSCREEN_NON_NATIVE_VISIBLE_MENU,
+  GHOSTTY_FULLSCREEN_TITLED_NON_NATIVE_VISIBLE_MENU,
```
mitchellh, line 415: I think this naming doesn't match the pattern of other variants. It should be `GHOSTTY_FULLSCREEN_NON_NATIVE_TITLED_VISIBLE_MENU` probably along with all the other enums and so on throughout this PR.

macos/Sources/Helpers/Fullscreen.swift
```diff
@@ -141,6 +145,7 @@ class NonNativeFullscreen: FullscreenBase, FullscreenStyle {
 
     struct Properties {
         var hideMenu: Bool = true
+        var untitled: Bool = true
```
mitchellh, line 148: I think having an inverted bool here makes the cognitive load slightly higher versus matching the config. I'd change this instead to `titled` and invert all the logic so we match the config.

mitchellh (changes requested):
I'm okay with this compromise. One request that is mostly busy work...

sassanh:
> I'm okay with this compromise. One request that is mostly busy work...

@mitchellh Thanks! I applied the changes.

sassanh:
@mitchellh Kindly review it again when you got the time.

mitchellh:
Hey @sassanh, sorry I delayed this for so long. I'm ready to review this and merge it quickly if you can rebase it on top of the latest changes. The macOS app has changed somewhat but I don't think the changes will be difficult to fit in (it should be less code than it was previously).

sassanh:
Nice! I believe I have a POC of a truly fullscreen, occupying the entire screen area edge to edge including the menubar while keeping the native tab bar (it's a regular window).

I'll attempt to clean up the code and update this pull request soon. The idea it is override `constrainFrameRect` like this:
```swift
    override func constrainFrameRect(_ frameRect: NSRect,
                                     to screen: NSScreen?) -> NSRect {
        frameRect
    }
```
This allows the window to be moved to the menubar area.

image

Rounded top corners are fine on MacBook, as it has rounded top corners, but I should find a way to get rid of the rounded corners in the bottom. I think I will find something in padded-notch probably. I will check it in my iMac which has square corners to see how it looks there.

sassanh:
With the tab-bar
image

mitchellh:
I'm willing to take a look at that option. The hard part is just figuring out how it interacts with everything else. 😄

sassanh:
@mitchellh I It is mostly functioning as expected in my narrow tests, kindly review it when you got the time.

All non-native fullscreen modes so far, didn't support tabs, so they solely changed the state of the current tab/window. But a non-native fullscreen that supports tabs, should apply changes to other tabs in the tab group too.

This is needed cause otherwise when you switch tabs in the tab group, you see some inconsistencies. For example you would see the green macOS fullscreen button is correctly disabled in the tab you triggered the fullscreen action on, but it is enabled in other tabs. Or when you exit fullscreen and switch to other tabs you would see the window size changes. Or since other tabs do not know they are in fullscreen mode, triggering the full-screen action while they are active, would just reapply fullscreen logic, instead of exiting fullscreen.

You can see these inconsistencies if you comment out this if and its content: https://github.com/ghostty-org/ghostty/pull/5397/files#diff-490b1864abbc90619e0ea991499485e5a886be260489253a982095b690686a89R789

That "if" is basically responsible for syncing the state of tabs in the tab group. I have a feeling it's a bit hacky and probably not the cleanest implementation. But with my limited knowledge about swift and ghostty codebase, I would need help to be able to implement it better. Please feel free to add a commit or to edit my commit if needed.

sassanh:
I moved redundant code into a new `syncNonNativeFullscreenState` method of the `BaseTerminalController`. Additionally, I ensured that new tabs are also synchronized with the tab group regarding their `fullscreenStyle`.

mitchellh:
Non-native fullscreen has a lot of complexity already, and I'm hesitant to add more. I know I said earlier I'm okay with the compromise but I'd like to walk that back. This hasn't been heavily requested (tabs have, though) and I'd rather focus in on a full custom tabs solution because that'll satisfy everyone, despite itself being complicated. 

I appreciate the work here that you did, I hope you can continue to maintain your fork for yourself, but I'm going to pass on this for merging into main. Sorry.

sassanh:
No worries, you’ve got the big picture.

I’ll likely keep my fork updated, since having a fullscreen terminal makes a huge difference to my concentration. In the past 25 years I never worked in a windowed terminal. I want it to resemble the old Linux TTY.

PR #5439, opened 2025-01-30
nix: use zig2nix to manage nix cache/deps

jcollie:
This brings the internal package more in line with how the nixpkgs package is built. It also handles recursive dependencies better than the current system.


mitchellh:
Going to push this out of 1.1 just due to the diff count and plans to release... now-ish lol.

mitchellh (changes requested):
Thanks! I'm about ready to merge this, I think the only thing we need to do is add `build.zig.zon.nix` to our `.gitattributes` file as auto-generated. Its a simple change.

jcollie:
> Thanks! I'm about ready to merge this, I think the only thing we need to do is add `build.zig.zon.nix` to our `.gitattributes` file as auto-generated. Its a simple change.

This file needs to be managed in git as it's the replacement for `nix/zigCacheHash.nix`

mitchellh:
I know. I suggested git attributes not git ignore :)

mitchellh:
I made the change, rebased, and updated for the new gobject reps.

jcollie:
> I know. I suggested git attributes not git ignore :)

ah. my bad. reading while in a hurry is bad for one's ego.

PR #5463, opened 2025-01-31
gtk(wayland): respect window-decoration=none on GNOME

pluiedev:
This is, admittedly, a very silly bug. On GNOME the SSD protocol is not available and past me just decided to always enable CSDs in that case, *even when* `window-decoration = none`. I now question my own intelligence.


brentdurksen:
After upgrading to 1.1.0 on GNOME on Wayland, window decorations started appearing despite having `window-decoration = none` in my config. Cherry-picking this commit onto 1.1.0 fixes the issue for me. Thanks.

tristan957 (approved):

PR #5472, opened 2025-01-31
Toggling visibility is now ignored when in fullscreen mode.

shaps80:
This PR ensures visibility toggling is ignored when the currently focused surface is fullscreen. This includes the following:

- Show/Hide All Terminals is ignored
- Using a binding to toggle is ignored

This ensures Ghostty doesn't unexpectedly lose focus or disappear and is the expected behavior on macOS.

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -709,11 +709,21 @@ class AppDelegate: NSObject,
 
     /// Toggles visibility of all Ghosty Terminal windows. When hidden, activates Ghostty as the frontmost application
     @IBAction func toggleVisibility(_ sender: Any) {
+        // Toggle visibility doesn't do anything if the focused window is native
+        // fullscreen.
+        guard let keyWindow = NSApp.keyWindow,
+              !keyWindow.styleMask.contains(.fullScreen) else { return }
+
         // If we have focus, then we hide all windows.
         if NSApp.isActive {
             // We need to keep track of the windows that were visible because we only
             // want to bring back these windows if we remove the toggle.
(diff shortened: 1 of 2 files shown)


shaps80:
One open question I have @mitchellh is regarding the Show/Hide menu. With this implementation the menu does nothing, including show/hide of any other Ghostty windows. I think this is valid/correct from my POV since the user can only see the currently fullscreen windows anyway. If you disagree, I'll need to update such that we only ignore the fullscreen windows but my preference would as implemented. Thought multi-screen setups may prefer an alternative where we still hide non-fullscreen windows. Let me know which way you'd like to go and I can update 👌

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -708,7 +708,11 @@ class AppDelegate: NSObject,
     }
 
     /// Toggles visibility of all Ghosty Terminal windows. When hidden, activates Ghostty as the frontmost application
+    ///
+    /// - Note: When the focused surface is fullscreen, this method does nothing.
```
mitchellh, line 712: We should add this note to the binding docs in `src/input/Binding.zig`

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -708,7 +708,11 @@ class AppDelegate: NSObject,
...
     @IBAction func toggleVisibility(_ sender: Any) {
+        guard terminalManager.focusedSurface?.window?.styleMask.contains(.fullScreen) == false else { return }
```
mitchellh, line 714: I think we can use `NSApp.keyWindow` (or whatever that accessor is)

mitchellh (changes requested):

shaps80:
@mitchellh addressed all comments, should be ready to go 👍

PR #5489, opened 2025-01-31
update zig-wayland

mitchellh:
This fixes a somewhat jank workaround they had to do. No functional changes. I also took this time to cache it in our mirror. The content hash always matches upstream so you can verify it yourself.


mitchellh:
This regresses https://codeberg.org/ifreund/zig-wayland/issues/67 so going to close and wait on that.

PR #5505, opened 2025-02-01
gtk: don't copy resize overlay

jcollie:


mitchellh (changes requested):
Why? What's the benefit here?

I personally try to avoid `@fieldParentPtr` if at all possible because it (1) ties a structure to directly being a member of another structure and (2) is somewhat confusing to understand for people. 

The only use case I know of where `@fieldParentPtr` is needed is in callback-based code that might pass a "self" pointer but doesn't support userdata.

jcollie:
Closing in favor of #5941

PR #5509, opened 2025-02-01
core: don't copy App and apprt.App

jcollie:
Besides avoiding copying, this allows consumers to choose to allocate these structs on the stack or to allocate on the heap. It also gives the apprt.App a stable pointer sooner in the process.

### DIFF
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -1318,6 +1318,7 @@ pub const CAPI = struct {
         config: *const Config,
     ) !*App {
         var core_app = try global.alloc.create(CoreApp);
+        try core_app.init(global.alloc);
         errdefer {
             core_app.deinit();
             global.alloc.destroy(core_app);
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -117,10 +117,11 @@ pub const App = struct {
     config: Config,
 
     pub fn init(
+        self: *App,
         core_app: *CoreApp,
         config: *const Config,
         opts: Options,
-    ) !App {
+    ) !void {
         // We have to clone the config.
         const alloc = core_app.alloc;
         var config_clone = try config.clone(alloc);
@@ -129,7 +130,7 @@ pub const App = struct {
         var keymap = try input.Keymap.init();
         errdefer keymap.deinit();
 
-        return .{
+        self.* = .{
             .core_app = core_app,
             .config = config_clone,
             .opts = opts,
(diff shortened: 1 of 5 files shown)


jcollie:
> I'm not against it but again... why now? Is this change for change sake or is there a motivating thing you're working on?

Yeah, as I was working on some of the changes needed to get the GTK apprt structs to use DerivedConfigs instead of using the "raw" config this made some things easier. Plus it's more consistent with how the other GTK structs are handled (Window, Tab, Surface).

src/main_ghostty.zig
```diff
@@ -96,11 +96,13 @@ pub fn main() !MainReturn {
     }
 
     // Create our app state
-    var app = try App.create(alloc);
-    defer app.destroy();
+    var app: App = undefined;
+    defer app.deinit();
+    try app.init(alloc);
```
gpanders, line 101: If `init` fails, `app.deinit()` will be called on an undefined object. The `defer` should come after the `init`.

```suggestion
    try app.init(alloc);
    defer app.deinit();
```

jcollie: Fixed.

src/apprt/embedded.zig
```diff
@@ -1346,13 +1347,16 @@ pub const CAPI = struct {
         opts: *const apprt.runtime.App.Options,
         config: *const Config,
     ) !*App {
-        var core_app = try CoreApp.create(global.alloc);
-        errdefer core_app.destroy();
+        var core_app = try global.alloc.create(CoreApp);
+        errdefer {
+            core_app.deinit();
+            global.alloc.destroy(core_app);
+        }
```
gabydd, line 1354: I believe core_app needs to be inited after this line seems like it's initialization was removed

gabydd (changes requested):
Just a little thing. Also i agree having app be a stable pointer would have been useful for me at some point(though it's also not that bad to have to allocate a couple objects for callbacks) but is there a reason that CoreApp needs a stable pointer?

src/apprt/embedded.zig
jcollie: Fixed.

jcollie:
> is there a reason that CoreApp needs a stable pointer

I don't have a particular use case or need at this time but made the change now for consistency's sake.

src/apprt/embedded.zig
```diff
@@ -1346,13 +1347,17 @@ pub const CAPI = struct {
         opts: *const apprt.runtime.App.Options,
         config: *const Config,
     ) !*App {
-        var core_app = try CoreApp.create(global.alloc);
-        errdefer core_app.destroy();
+        var core_app = try global.alloc.create(CoreApp);
+        try core_app.init(global.alloc);
+        errdefer {
+            core_app.deinit();
+            global.alloc.destroy(core_app);
```
gabydd, line 1354: This part of the error defer has to be above the init so if the init fails then it will be destroyed. The ordering will still be correct as the second errdefer runs first see this as an example 
```zig
const std = @import("std");

fn fail() !void {
    return error.Fail;
}
pub fn main() !void {
    errdefer std.debug.print("first\n", .{});
    errdefer std.debug.print("second\n", .{});
    try fail();
}
```

gabydd (changes requested):
Sorry about the churn on this part

mitchellh (approved):
Rebased, reintroduce `create` to App there's no reason not to have that and its idiomatic Zig.

PR #5529, opened 2025-02-02
Remember last focused window position for next startup

### LINKED ISSUE #4233: macOS: Remember last focused window location for next startup
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4204

<sup>Originally posted by **evertonstz** December 31, 2024</sup>
I'm one of those guys that keeps closing and reopening their terminal during the day and that's one of the first things I noticed when using Ghostty: it doesn't remembers the size and position of the last closed window like most other MacOS apps.
It's been a long time since I used Linux with modern gnome, but I think that's also the expected behavior for apps there, right?</div>

## Research on Other Behavior

`Terminal.app` and `iTerm` both remember the position of the last focused window (not last created window) and restore it when restarting the app.  Only the position is remembered, not the size.

**Note this is different from `window-save-state`.** This isn't _restoring_ any window. This is just remembering the _position_.

This behavior isn't typical with Gnome applications, so this is tagged macOS only.

## Implementation

Given this is a macOS only behavior, I'm inclined to leverage macOS APIs for this and put this information into `defaults`. I suspect that's where these programs are also doing it but verification required.

I think we should solve this completely in the Swift code.

There are no configs currently to specify an initial window position, but in the future if/when we have those, those should always override this.

liby:
Implements window position persistence on macOS, similar to Terminal.app and iTerm2. The window position is saved when it becomes main or moves, and restored on next startup. Window position is kept within visible screen bounds.

cc @jsumners - Would you mind giving it a try? Any feedback would be appreciated!

Resolves #4233

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -72,6 +72,7 @@
 		A5A2A3CA2D4445E30033CF96 /* Dock.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5A2A3C92D4445E20033CF96 /* Dock.swift */; };
 		A5A2A3CC2D444ABB0033CF96 /* NSApplication+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5A2A3CB2D444AB80033CF96 /* NSApplication+Extension.swift */; };
 		A5A6F72A2CC41B8900B232A5 /* Xcode.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5A6F7292CC41B8700B232A5 /* Xcode.swift */; };
+		A5AEB1652D5BE7D000513529 /* LastWindowPosition.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5AEB1642D5BE7BF00513529 /* LastWindowPosition.swift */; };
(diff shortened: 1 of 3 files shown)


jsumners:
Is it available in some sort of nightly build?

jcollie:
> Is it available in some sort of nightly build?

You'll need to build from source with this PR applied to test this out.

jsumners:
I be up for giving it a whirl if I could download a binary from this project's CI. I suspect setting up a Zig environment isn't too burdensome, but I don't currently have one and have other things to focus on.

mitchellh:
Thanks! This was on the right path. I made some changes:

* Moved the WindowPositionManager out to a Helper file
* Moved the position logic from SurfaceView to the NSWindowDelegate (TerminalController)
* We already have a `setInitialPosition` function that handles the various initial positions, I use that now
* As a note: putting it on TerminalController ensures this only applies to main terminal windows, not quick terminal and so on

mitchellh (approved):

martimlobao:
Is there a setting to always set the initial position at the center of the screen? This could probably be achieved with `window-position-x`/`-y`, but you'd have to take into account the window size, and it would only work for a specific screen size. I was thinking of something along the lines of `window-position-x = center`, but I don't know if some other setting does this.

iamart:
how to enable it? or it just remembers position and not whether window was maximized or not

00-kat:
Pull requests are not for technical support. Open a [new discussion].

[new discussion]: https://github.com/ghostty-org/ghostty/discussions/new?category=q-a

PR #5530, opened 2025-02-02
core: add explicit errors to src/os/env.zig

jcollie:


src/os/env.zig
```diff
@@ -3,13 +3,15 @@ const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 const posix = std.posix;
 
+pub const Errors = Allocator.Error;
```
mitchellh, line 6: This should be singular

PR #5531, opened 2025-02-02
core: add explicit errors to src/pty.zig

jcollie:


src/pty.zig
```diff
@@ -41,12 +41,14 @@ pub const Mode = packed struct {
 // a termio that doesn't use a pty. This isn't used in any user-facing
 // artifacts, this is just a stopgap to get compilation to work on iOS.
 const NullPty = struct {
+    pub const Errors = error{GetModeFailed};
```
mitchellh, line 44: Idiomatically should be named `Error`

Secondly though I don't see this const (in any of the implementations) used anywhere. Is it?

mitchellh (changes requested):
Thanks, see note

PR #5533, opened 2025-02-02
gtk(x11): support server-side decorations

pluiedev:
Remind me to never touch Xlib code ever again.

### DIFF
--- a/src/apprt/gtk/winproto/x11.zig
+++ b/src/apprt/gtk/winproto/x11.zig
@@ -13,7 +13,7 @@ const log = std.log.scoped(.gtk_x11);
 pub const App = struct {
     display: *c.Display,
     base_event_code: c_int,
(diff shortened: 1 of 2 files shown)


1
```diff
@@ -0,0 +1,379 @@
+info: ghostty version=1.1.1-HEAD+3470f33c
```
gtjoseph, line 1: A log file that didn't get deleted perhaps?

gtjoseph (changes requested):

1
gtjoseph: I'd approve the PR but it looks like it's failing to compile again, both in CI and for me locally.

gtjoseph (approved):

mitchellh (changes requested):
Really excellent work, as usual! A few notes/requests.

PR #5539, opened 2025-02-02
New editor action for write_screen_file

AdrielVelazquez:
This PR adds an additional option to open write_screen_file in your defined terminal `EDITOR` instead of the desktop environments default with `xdg-open` or `open`

[Screencast From 2025-02-02 15-22-56.webm](https://github.com/user-attachments/assets/5b184372-1fb6-4f95-9b6f-bd69e6ee0134)

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -18,6 +18,7 @@ pub const Message = apprt.surface.Message;
 const std = @import("std");
 const builtin = @import("builtin");
 const assert = std.debug.assert;
+const getenv = @import("os/env.zig").getenv;
 const Allocator = std.mem.Allocator;
 const ArenaAllocator = std.heap.ArenaAllocator;
 const global_state = &@import("global.zig").state;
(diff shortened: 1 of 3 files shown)


src/Surface.zig
```diff
@@ -4317,6 +4317,22 @@ fn closingAction(action: input.Binding.Action) bool {
...
 }
 
+fn openScreenFile(
+    self: *Surface,
+    file_path: []const u8,
+) !void {
+    const editor = std.posix.getenv("EDITOR") orelse {
+        std.log.debug("EDITOR environment variable not set", .{});
+        return error.EnvironmentVariableNotFound;
+    };
```
pluiedev, line 4327: In core code we should not assume the OS to be POSIX. Either use `std.process.getEnvVarOwned` or our own internal `getenv`. Also, use the scoped logger (`log`)

src/Surface.zig
```diff
@@ -4317,6 +4317,22 @@ fn closingAction(action: input.Binding.Action) bool {
...
+
+    const command = try std.fmt.allocPrint(self.alloc, "{s} {s}\n", .{ editor, file_path });
```
pluiedev, line 4329: We should not be just injecting stuff into the user's shell...

src/Surface.zig
```diff
@@ -4694,3 +4711,17 @@ fn presentSurface(self: *Surface) !void {
...
+    const allocator = testing.allocator;
+
+    var surface = Surface{ .alloc = allocator, .size = undefined, .app = undefined, .rt_app = undefined, .rt_surface = undefined, .font_grid_key = undefined, .font_size = undefined, .font_metrics = undefined, .renderer = undefined, .renderer_state = undefined, .renderer_thr = undefined, .renderer_thread = undefined, .mouse = undefined, .keyboard = undefined, .io = undefined, .io_thr = undefined, .io_thread = undefined, .config = undefined, .config_conditional_state = undefined };
+    const file_path = "test.txt";
+
+    _ = unsetenv("EDITOR");
+
+    const err = surface.openScreenFile(file_path);
+    try testing.expectError(error.EnvironmentVariableNotFound, err);
+}
```
pluiedev, line 4727: Initializing a blank `Surface` like that is just silly and overkill.

pluiedev (changes requested):

jcollie:
Thanks for your work on this, but due to the security implications of injecting a command into the user's buffer this isn't a PR that we're going to accept. This _is_ functionality that we want to have, it just needs to be implemented in a different manner.

PR #5550, opened 2025-02-03
apprt/gtk: handle input methods that end preedit before commit

### LINKED ISSUE #5494: GTK regression: ibus 1.5.31 dead keys on Wayland broken
1.1.0 breaks dead keys on ibus `1.5.31` in Hyprland (non-GNOME). The initial modifier character shows but the final character eats up input. Example:

```
`  + a = both are eaten up
´ + e = both are eaten up
¨ + space = both are eaten up

and so on...
```

I had to downgrade to 1.0.1. I'm running on Arch with the following:

```
Ghostty 1.0.1-arch

Version
  - version: 1.0.1-arch
  - channel: tip
Build Config
  - Zig version: 0.13.0
  - GTK version:
    build      : 4.16.12
    runtime    : 4.16.12
  - libadwaita : enabled
    build      : 1.6.3
    runtime    : 1.6.4
  - libX11     : enabled
  ```

ibus versions:

```
pacman -Q | grep ibus
ibus 1.5.31-2
libibus 1.5.31-2
```

Other GTK4/Adwaita apps are working fine, so are browsers, etc. I'm running on Hyprland, but I assume this happens in other places too. ~I'll try GNOME and report back.~ EDIT: 1.1.0 works in GNOME, but not Hyprland.

So, I kept digging. The difference between GNOME and Hyprland on the same system is that Hyprland doesn't start ibus, so the changes between 1.0.3 and 1.1.0 are dependent on ibus being running. Digging further...

_Originally posted by @tomaspinho in https://github.com/ghostty-org/ghostty/issues/5324#issuecomment-2628459180_

mitchellh:
Fixes #5494

When ibus/fcitx is not running (the GTK "simple" input method is active), the preedit end event triggers _before_ the commit event. For both ibus/fcitx, the opposite is true. We were relying on this ordering.

This commit changes the GTK input handling to not rely on this ordering. Instead, we encode our composing state into the boolean state of whether a key event is pressed. This happens before ANY input method events are triggered.

Tested dead key handling on: X11/Wayland, ibus/fcitx/none.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,6 +78,89 @@ pull request will be accepted with a high degree of certainty.
 > not open a WIP pull request to discuss a feature. Instead, use a discussion
 > and link to your branch.
 
+# Developer Guide
+
+> [!NOTE]
+>
+> **The remainder of this file is dedicated to developers actively
+> working on Ghostty.** If you're a user reporting an issue, you can
+> ignore the rest of this document.
+
+## Input Stack Testing
+
+The input stack is the part of the codebase that starts with a
+key event and ends with text encoding being sent to the pty (it
+does not include _rendering_ the text, which is part of the
+font or rendering stack).
+
+If you modify any part of the input stack, you must manually verify
+all the following input cases work properly. We unfortunately do
+not automate this in any way, but if we can do that one day that'd
+save a LOT of grief and time.
+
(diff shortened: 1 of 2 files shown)


mitchellh:
This is a draft, I still need to verify ibus 1.5.31 and run through some other input methods.

mitchellh:
Verified CJK, Emoji, Hex still work.

PR #5560, opened 2025-02-03
gtk: introduce Zig bindings for GTK/GObject

pluiedev:
`zig-gobject` is a set of GObject bindings that allow us to write GTK-facing code in Zig instead of getting hands dirty with C. It's been tested and refined in real-life applications and several GTK contributors agree that it is a marked improvement over using the C API directly, such as allowing method call syntax and avoiding many manual `@ptrCast`s.

This PR doesn't actually contain any changes to our preexisting GTK code — the migration process is intended to begin in chunks, firstly in self-contained components (e.g. the header bar, overlays, etc.), and then full-scale migration can begin when we remove non-Adwaita GTK builds for 1.2. (After all, why port code that you'll remove later either way?)

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -41,6 +41,10 @@
             .url = "git+https://github.com/natecraddock/zf/?ref=main#ed99ca18b02dda052e20ba467e90b623c04690dd",
             .hash = "1220edc3b8d8bedbb50555947987e5e8e2f93871ca3c8e8d4cc8f1377c15b5dd35e8",
         },
+        .gobject = .{
+            .url = "https://github.com/ianprime0509/zig-gobject/releases/download/v0.2.2/bindings-gnome47.tar.zst",
+            .hash = "12208d70ee791d7ef7e16e1c3c9c1127b57f1ed066a24f87d57fc9f730c5dc394b9d",
+        },
 
         // C libs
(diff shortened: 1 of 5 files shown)


mitchellh:
I'm totally on board with cleaning this up, but I always hesitate to add new Zig dependencies because it makes it that much harder to upgrade Zig minor versions. Is the maintainer in the Zig community? Can they commit to keeping up with minor releases?

pluiedev:
I don't know what their credentials in the community would be, but I've used one of their other libraries, zig-gettext, for the i18n PR, and I think it's pretty solid work. As of now, [the project README](https://github.com/ianprime0509/zig-gobject?tab=readme-ov-file#usage) claims that the generated bindings aim to be compatible with both 0.13 and master (0.14), and evidently a lot of work has been done lately to make that happen.

mochalins:
I've seen the author (Ian) active on various Zig communities (GitHub, Discord, etc.) for at least the past two years, if that helps assuage any worries.

mitchellh (approved):
Okay, let's bring it in. I agree that the sanity this brings is extremely high so it's a very high value dependency. We'll see when Zig 0.14 is released how much our dependencies (in general) cause us any grief, since our upgrade path will be dependent on them. Merging.

PR #5570, opened 2025-02-04
Fix cursor character not visible when cursor color matches background

### LINKED ISSUE #5554: Metal (regression): cannot see character under the cursor if block cursor and text are the same
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5480

<sup>Originally posted by **kpovel** January 31, 2025</sup>
https://github.com/user-attachments/assets/94262c57-d6de-412f-b88e-5863562136e5

In my ghostty config, I have: 

```
background = 000000
shell-integration-features = no-cursor # makes wide cursor
```

I found a quick fix for this bug, but I am not sure if it is optimal for every case: https://github.com/kpovel/ghostty/commit/a1afcc7ec59573bf1a493d9ab253e66ae178cc02</div>

### Notes

This is probably Metal-only so I tagged it as such.

liby:
When cursor color matches the background color, the optimization that skips rendering identical colors causes the character under cursor to become invisible. This PR ensures characters at cursor position are always rendered, maintaining cursor visibility regardless of color settings.

image

cc @kpovel - Would you mind giving it a try? Any feedback would be appreciated!

Fixes #5554

### DIFF
--- a/src/renderer/shaders/cell.metal
+++ b/src/renderer/shaders/cell.metal
@@ -471,16 +471,15 @@ vertex CellTextVertexOut cell_text_vertex(
     out.color = contrasted_color(uniforms.min_contrast, out.color, out.bg_color);
   }
 
-  // If this cell is the cursor cell, then we need to change the color.
-  if (
-    in.mode != MODE_TEXT_CURSOR &&
-    (
+  // Check if current position is under cursor (including wide cursor)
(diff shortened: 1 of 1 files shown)


kpovel:
It works as expected, the character under the cursor is rendering

mitchellh (approved):

PR #5579, opened 2025-02-05
[macOS] Improve split and tab navigation behavior

### LINKED ISSUE #5552: Split and tab navigation keybind actions should be performable
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5490

<sup>Originally posted by **shinoxzu** January 31, 2025</sup>
Hi! Before the 1.1.0 update, I was able to select the next/previous split using Ctrl+Tab and Ctrl+Shift+Tab bindings, as long as there were no other tabs. If there were any, Ghostty would ignore the splits and select the tabs instead. I liked this behavior.

Now, in 1.1.0, this no longer works. I explored the new performable prefix and tried using it like this:
```
keybind = performable:ctrl+tab=goto_split:next
keybind = performable:ctrl+shift+tab=goto_split:previous
```
So as far as I understand, this should restore the previous behavior, but it doesn’t: I cannot select the next/previous tab even if there are no splits (in any of the tabs). However, moving between splits works just fine. I suppose this happens because Ghostty considers the goto_split action performable even when there are no other splits.

May I do it another way, or can we perform the goto_split action only if there are some splits?</div>

liby:
This PR improves the split and tab navigation experience by making `gotoSplit` actions (next/previous) automatically fall back to `gotoTab` actions (next/previous)  when no splits exist in the current window, providing more intuitive navigation between splits and tabs.

The implementation is currently macOS-only as I don't have access to Linux hardware for testing.

cc @shinoxzu - Would you mind giving it a try? Any feedback would be appreciated!

Partially fixes #5552

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -107,7 +107,7 @@ extension Ghostty {
         deinit {
             // This will force the didSet callbacks to run which free.
             self.app = nil
-            
+
 #if os(macOS)
             NotificationCenter.default.removeObserver(self)
 #endif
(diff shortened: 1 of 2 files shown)


shinoxzu:
Hello! I tried it and it works just fine. Thank you!

mitchellh:
Unfortunately this isn't the right approach. #5644 is closer. `goto_split` was never meant to fallback to `goto_tab`, it only worked previously for the specific non-default keybinding specified in #5552 because that keybinding is the same as the system bindings. We should not be making `goto_split` fallback to `goto_tab`.

Instead, we need a way for `goto_split` to work in the new performable system so that we can ignore the key input when its not performable (and the binding is marked performable).

PR #5590, opened 2025-02-05
gtk: add option to always display the tab bar

pluiedev:
Also fixes crashes in both vanilla GTK and Adwaita implementations of `closeTab`, which erroneously close windows twice when there are no more tabs left (we probably already handle it somewhere else).

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1410,6 +1410,27 @@ keybind: Keybinds = .{},
 ///   * `end` - Insert the new tab at the end of the tab list.
 @"window-new-tab-position": WindowNewTabPosition = .current,
 
+/// Whether to show the tab bar.
+///
+/// Valid values:
+///
+///  - `always`
+///
+///    Always display the tab bar, even when there's only one tab.
+///
+///  - `auto` *(default)*
+///
+///    Automatically show and hide the tab bar. The tab bar is only
+///    shown when there are two or more tabs present.
+///
+///  - `never`
+///
+///    Never show the tab bar. Tabs are only accessible via the tab
+///    overview or by keybind actions.
+///
+/// Currently only supported on Linux (GTK).
+@"window-show-tab-bar": WindowShowTabBar = .auto,
+
 /// Background color for the window titlebar. This only takes effect if
 /// window-theme is set to ghostty. Currently only supported in the GTK app
 /// runtime.
@@ -5747,7 +5768,6 @@ pub const GtkSingleInstance = enum {
 pub const GtkTabsLocation = enum {
     top,
     bottom,
-    hidden,
 };
 
 /// See gtk-toolbar-style
@@ -5793,6 +5813,13 @@ pub const WindowSaveState = enum {
 pub const WindowNewTabPosition = enum {
     current,
     end,
+};
+
+/// See window-show-tab-bar
+pub const WindowShowTabBar = enum {
+    always,
+    auto,
+    never,
 };
 
 /// See resize-overlay
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -54,6 +54,9 @@ window: *adw.ApplicationWindow,
 /// The header bar for the window.
 headerbar: HeaderBar,
 
+/// The tab bar for the window.
+tab_bar: *adw.TabBar,
+
 /// The tab overview for the window. This is possibly null since there is no
 /// taboverview without a AdwApplicationWindow (libadwaita >= 1.4.0).
 tab_overview: ?*adw.TabOverview,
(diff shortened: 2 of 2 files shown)


jcollie (dismissed):

src/config/Config.zig
```diff
@@ -1336,6 +1336,11 @@ keybind: Keybinds = .{},
 ///   * `end` - Insert the new tab at the end of the tab list.
 @"window-new-tab-position": WindowNewTabPosition = .current,
 
+/// Whether tabs should always be displayed, regardless of how many tabs
+/// are open. If false (default), the tab bar is only displayed when there
+/// are two or more tabs.
+@"window-always-show-tabs": bool = false,
```
mitchellh, line 1342: This should note it's only GTK for now. This has been requested on macOS as well so I'm interested in making it work there eventually.

mitchellh (changes requested):
One request.

I also want to consider a bike shed on the config option. I wonder if we want to rename it to something like `window-tab-bar` and support like `auto, always` right now. There's some requests to have `never` already (weird but exist) and while I dont' want to support it now it'd let us reuse that config one day.

pluiedev:
Why would you want to have it as `never`... 🤦‍♀️ Though I guess it's not that hard to implement

pluiedev:
So I've been thinking about this for a while, and I think it might be worthwhile to reorganize the config options pertaining to tab bars:

- `window-tab-bar` = `auto`, `always`, `none`

  - `auto` is behavior we have right now, and `none` has roughly the same effect as `gtk-tabs-location=hidden`, but the tab bar is merely *hidden* instead of not being attached
  
- `window-tab-bar-position` = `top`, `bottom`, (`left`, `right`?)

 - AFAICT macOS currently only supports `top` due to native tabbing limitations, but `bottom` could be supported when we move to a custom tabbing solution. `left` and `right` aren't included by default as it's not a design that libadwaita supports natively, but we *could* devise a hack later using something like [`AdwOverlaySplitView`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.OverlaySplitView.html) with the sidebar always visible.

mitchellh:
> My personal issue with libadwaita is the vertical space it requires for tabs (CSD and SSD), and that it doesn't respect (prob can't) what I have otherwise configured as GTK theme.

Note you can also attempt to slim them down using our `gtk-custom-css` configuration option. I haven't verified if its possible or how easy it is, but on paper it should be possible.

src/config/Config.zig
```diff
@@ -1336,6 +1336,26 @@ keybind: Keybinds = .{},
 ///   * `end` - Insert the new tab at the end of the tab list.
 @"window-new-tab-position": WindowNewTabPosition = .current,
 
+/// Whether to show the tab bar.
+/// Currently only supported on Linux (GTK).
```
mitchellh, line 1340: A very small nit: this text is usually shown in a separate line at the very bottom of the config docs idiomatically.

mitchellh (changes requested):
Small change request but the rest looks good. Feel free to merge at your desire if you feel its stable.

pluiedev:
I'm still not sure about the `never` case as it does basically the same (or even *exactly* the same when libadwaita is disabled) thing as `gtk-tabs-location=hidden`... Though I'm fine with only merging the `always` option

pluiedev:
Yeah - the difference is that on libadwaita `gtk-tabs-location=hidden` we display the number of tabs in a tab button, whereas `window-tab-bar=never` doesn't. I find that kinda silly...

src/config/Config.zig
```diff
@@ -1336,6 +1336,27 @@ keybind: Keybinds = .{},
...
+///    Automatically show and hide the tab bar. The tab bar is only
+///    shown when there are two or more tabs present.
+///
+///  - `never`
+///
+///    Never show the tab bar. Tabs are only accessible via the tab
+///    overview or by keybind actions.
+///
+/// Currently only supported on Linux (GTK).
+@"window-show-tab-bar": WindowShowTabBar = .auto,
```
tristan957, line 1358: Should we just remove `gtk-tabs-location=hidden`?

pluiedev: IMO probably yeah, but then we need to have a migration mechanism for people who are using that config. (See, I did tell everyone that we might run into more complex config migrations in the future :sweat_smile:)

jcollie (changes requested):
This LGTM except that it needs rebased now that adwaita is required.

chenzhiwei:
@pluiedev Hi, could you please rebase this PR?

pluiedev:
Rebased and replaced `gtk-tabs-location=hidden` with `window-show-tab-bar=never`. This is obviously a breaking change but we can add a migration mechanism in a different PR

jcollie (approved):

chenzhiwei:
aha, conflicts again.

chenzhiwei:
@mitchellh could you take a look at this long waited feature?

mitchellh (approved):

Ev357:
> Rebased and replaced `gtk-tabs-location=hidden` with `window-show-tab-bar=never`. This is obviously a breaking change but we can add a migration mechanism in a different PR

found why it broke my config haha

mitchellh:
Oh good catch.

mitchellh:
Made #7706 to track that.

PR #5593, opened 2025-02-05
GTK: disable color management

eifr:
fixes: https://github.com/ghostty-org/ghostty/discussions/4976

Before: 
image

After:
image

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -146,14 +146,14 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
 
     var gdk_disable: struct {
         @"gles-api": bool = false,
-        /// current gtk implementation for color management is not good enough. 
-        /// see: https://bugs.kde.org/show_bug.cgi?id=495647
-        @"color-mgmt": bool = true,
+        @"color-mgmt": bool,
         /// Disabling Vulkan can improve startup times by hundreds of
         /// milliseconds on some systems. We don't use Vulkan so we can just
         /// disable it.
         vulkan: bool = false,
-    } = .{};
+    } = .{
+        .@"color-mgmt" = config.@"disable-gtk-color-mgmt",
+    };
 
     environment: {
         if (version.runtimeAtLeast(4, 16, 0)) {
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -148,16 +148,18 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         @"gles-api": bool = false,
         /// current gtk implementation for color management is not good enough.
         /// see: https://bugs.kde.org/show_bug.cgi?id=495647
-        @"color-mgmt": bool,
+        @"color-mgmt": bool = true,
         /// Disabling Vulkan can improve startup times by hundreds of
         /// milliseconds on some systems. We don't use Vulkan so we can just
         /// disable it.
         vulkan: bool = false,
-    } = .{
-        .@"color-mgmt" = config.@"gtk-gdk-disable-color-mgmt",
-    };
+    } = .{};
 
     environment: {
+        if (version.runtimeAtLeast(4, 17, 0)) {
+            gdk_disable.@"color-mgmt" = false;
+        }
+
         if (version.runtimeAtLeast(4, 16, 0)) {
             // From gtk 4.16, GDK_DEBUG is split into GDK_DEBUG and GDK_DISABLE.
             // For the remainder of "why" see the 4.14 comment below.
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -148,6 +148,7 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         @"gles-api": bool = false,
         /// current gtk implementation for color management is not good enough.
         /// see: https://bugs.kde.org/show_bug.cgi?id=495647
+        /// gtk issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/6864
         @"color-mgmt": bool = true,
         /// Disabling Vulkan can improve startup times by hundreds of
         /// milliseconds on some systems. We don't use Vulkan so we can just
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -146,6 +146,9 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
 
     var gdk_disable: struct {
         @"gles-api": bool = false,
+        /// current gtk implementation for color management is not good enough. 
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/App.zig
```diff
@@ -146,6 +146,9 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
 
     var gdk_disable: struct {
         @"gles-api": bool = false,
+        /// current gtk implementation for color management is not good enough. 
+        /// see: https://bugs.kde.org/show_bug.cgi?id=495647
+        @"color-mgmt": bool = true,
```
pluiedev, line 151: I don't think this should be an unconditional workaround. Maybe restrict it to certain GTK versions?

pluiedev (changes requested):

src/apprt/gtk/App.zig
eifr: Seems like it's GTK's impl and not a bug introduce in a specific version.

pluiedev: From the upstream issue it reads that GTK's implementation was broken and is intended to be fixed in future versions. Therefore I would recommend to only disable color management explicitly for those broken GTK versions (IIRC 4.16 was the series that introduced color management, and from there you can track down the patch releases)

jcollie: Rather than tracking down version numbers, I'd just default it to false and add a config option to enable it.

eifr: I ended up switching to config

src/config/Config.zig
```diff
@@ -1350,6 +1350,10 @@ keybind: Keybinds = .{},
 /// Specified as either hex (`#RRGGBB` or `RRGGBB`) or a named X11 color.
 @"window-titlebar-foreground": ?Color = null,
 
+/// current gtk implementation for color management is not good enough.
+/// see: https://bugs.kde.org/show_bug.cgi?id=495647
+@"disable-gtk-color-mgmt": bool = false,
```
jcollie, line 1355: A better name would be `gtk-gdk-disable-color-mgmt`. Any GTK-specific config entries should be prefixed with `gtk`.

src/apprt/gtk/App.zig
```diff
@@ -146,14 +146,14 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
 
     var gdk_disable: struct {
         @"gles-api": bool = false,
-        /// current gtk implementation for color management is not good enough. 
-        /// see: https://bugs.kde.org/show_bug.cgi?id=495647
-        @"color-mgmt": bool = true,
```
jcollie: I'd keep the comments here, even though it's a duplicate.

src/config/Config.zig
jcollie: I'd expand the comment here and explain the purpose and effect of the setting a little more clearly (including the default and under what situations you'd want to enable this). While fine for an inline code comment, this will appear in user-facing documentation.

jcollie (changes requested):

src/config/Config.zig
pluiedev: I still am of the opinion that this should not be a config option. We already work around numerous GTK quirks that were tied to specific versions in the code base and IMHO that's necessary for a good OOTB experience. If we go down this route, expect tons of Plasma 6.3 users (whenever it comes out) on buggy GTK versions to complain...

jcollie: I agree in general that defaulting to "doing the right thing" automatically is good, wouldn't this necessitate conditioning based on both the GTK version _and_ the Plasma version? We'd need to add detection for fixed versions of GTK (easy if we know the versions) as well as KDE version (not sure how hard this would be for a GTK app).

pluiedev: In my understanding it's that Plasma is implementing the xx_color_management_v4 protocol completely correctly and GTK is the one that's doing it wrong, so we shouldn't really need to care about the Plasma version

src/config/Config.zig
```diff
@@ -1350,9 +1350,18 @@ keybind: Keybinds = .{},
...
+///
+/// By default this is set to `false`, meaning color management is enabled.
+/// You may want to enable this setting (set to `true`) if you experience:
+/// - Incorrect or washed out colors in your terminal
+/// - Color inconsistencies between GTK applications
+/// - Performance issues related to color management
+///
+/// This is a workaround for known issues with GTK's color management implementation,
+/// particularly affecting applications running under Wayland.
+/// See: https://bugs.kde.org/show_bug.cgi?id=495647
```
jcollie, line 1363: This looks good, but is there a corresponding GTK bug that we can link to?

src/config/Config.zig
eifr: I ran Fedora Rawhide on a VM with GTK 4.17.4, and it’s working fine without disabling `color-mgmt`. Honestly, I’m not sure why—I checked the GTK changelog, but nothing specifically points to these fixes.  

Because of that, I decided to remove the config. But anyway, maybe we should allow users to pass env vars for `GTK_DISABLE`.  

[[This issue](https://gitlab.gnome.org/GNOME/gtk/-/issues/6864)] tracks their fixes.

pluiedev (approved):
Can't test on my end (Nixpkgs's GTK is tied to GNOME major releases, and 48 won't come till next month at the earliest), but the idea sounds good. We can always tweak this later.

src/apprt/gtk/App.zig
```diff
@@ -146,13 +146,21 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
...
         /// Disabling Vulkan can improve startup times by hundreds of
         /// milliseconds on some systems. We don't use Vulkan so we can just
         /// disable it.
         vulkan: bool = false,
     } = .{};
 
     environment: {
+        if (version.runtimeAtLeast(4, 17, 0)) {
+            gdk_disable.@"color-mgmt" = false;
+        }
```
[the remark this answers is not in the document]
pluiedev: Hm yeah. I hope by 4.18 they can get their stuff together for GNOME 48...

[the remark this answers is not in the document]
eifr: updated to 4.18 for now

eifr:
@pluiedev Could you please take another look? I think your requested changes are blocking progress.

eifr:
What's the reason this pull request is blocked?

src/config/Config.zig
pluiedev: After a *lot* of digging it appears that this is the corresponding GTK MR: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7612, fixed in [4.15.6](https://gitlab.gnome.org/GNOME/gtk/-/blob/15edfe4ffa7a800ab67717e6a6be81d0b37c1813/NEWS#L28)

PR #5625, opened 2025-02-07
Metal: use "Managed" resource storage mode on discrete GPUs

qwerasd205:
Discrete GPUs cannot use the "shared" storage mode. This causes undefined behavior right now, and I believe it's what's causing a problem on Intel systems with discrete GPUs with "inverted" cells. (Observed in discussion #5597)

This commit also sets the CPU cache mode to "write combined" for our resources since we don't read them back so Metal can optimize them further with this hint.

Will want to get verification from someone with an Intel + discrete GPU system that this fixes the issue before this is merged.

### DIFF
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -182,15 +182,34 @@ pub const GPUState = struct {
     /// This buffer is written exactly once so we can use it globally.
     instance: InstanceBuffer, // MTLBuffer
 
+    /// The default storage mode to use for resources created with our device.
+    ///
+    /// This is based on whether the device is a discrete GPU or not, since
+    /// discrete GPUs do not have unified memory and therefore do not support
+    /// the "shared" storage mode, instead we have to use the "managed" mode.
+    default_storage_mode: mtl.MTLResourceOptions.StorageMode,
+
     pub fn init() !GPUState {
         const device = try chooseDevice();
         const queue = device.msgSend(objc.Object, objc.sel("newCommandQueue"), .{});
         errdefer queue.release();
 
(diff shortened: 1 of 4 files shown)


src/renderer/Metal.zig
```diff
@@ -182,27 +182,46 @@ pub const GPUState = struct {
...
     pub fn init() !GPUState {
         const device = try chooseDevice();
         const queue = device.msgSend(objc.Object, objc.sel("newCommandQueue"), .{});
         errdefer queue.release();
 
+        // We determine whether our device is a discrete GPU based on these:
+        // - We're on macOS (iOS, iPadOS, etc. are guaranteed to be integrated).
+        // - We're not on aarch64 (Apple Silicon, therefore integrated).
+        // - The device reports that it does not have unified memory.
+        const is_discrete =
```
mitchellh, line 201: This is super minor and we don't need to address it if you don't want, but it may be helpful to make a top-level const like `const discrete_supported` with the comptime parts of this. 

I'm not sure we'd use it anywhere else but I think it'd also clean this boolean up a bit.

mitchellh (approved):
This looks great to me, thanks for chasing this down. 

I'd be fine merging without confirmation. I don't have any Intel machines, but if we want to ping specific people that may work. I wonder if this will fix the "red reverse E" issue that Intel folks have had and we thought was a Metal bug.

src/renderer/Metal.zig
qwerasd205: I feel like it's fine for now. In my renderer rework I'll probably go ahead and do that though.

PR #5632, opened 2025-02-08
core: protect against crashes and hangs when themes are not files

jcollie:
If a theme was not a file or a directory you could get a crash or a hang (depending on platform) if the theme references a directory. This patch also prevents attempts to load from other non-file sources.

Fixes: #5596

### DIFF
--- a/src/cli/args.zig
+++ b/src/cli/args.zig
@@ -8,6 +8,8 @@ const internal_os = @import("../os/main.zig");
 const Diagnostic = diags.Diagnostic;
 const DiagnosticList = diags.DiagnosticList;
 
+const log = std.log.scoped(.cli);
+
 // TODO:
 //   - Only `--long=value` format is accepted. Do we want to allow
 //     `--long value`? Not currently allowed.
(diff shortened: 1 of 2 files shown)


src/config/Config.zig
```diff
@@ -3381,6 +3381,14 @@ fn loadTheme(self: *Config, theme: Theme) !void {
         name,
         &self._diagnostics,
     )) orelse return;
+    const stat = try themefile.file.stat();
+    switch (stat.kind) {
+        .file => {},
+        else => {
+            log.warn("not loading file \"{s}\" because it is a: {s}", .{ themefile.path, @tagName(stat.kind) });
```
00-kat, line 3388: Useless nitpick: "it is" here but "it's" in `theme.zig`.
```suggestion
            log.warn("not loading file \"{s}\" because it's a: {s}", .{ themefile.path, @tagName(stat.kind) });
```

00-kat (approved):

src/config/Config.zig
jcollie: Actually, i just removed the stat test from this location. Going to rely on `themepkg.open` never returning anything but a file.

PR #5599, opened 2025-02-05
Introduce subsystem maintainers via CODEOWNERS

mitchellh:
This commit introduces the proposed subsystem maintainer system for Ghostty. This commit doesn't assign anyone yet to the subsystems, but defines the CODEOWNERS file, creates the GitHub teams, and documents the system.

We can discuss this in Discord, but any feedback is also welcome here.

PR #5644, opened 2025-02-09
Performable tab and split navigation

### LINKED ISSUE #5552: Split and tab navigation keybind actions should be performable

jcollie:
Partial addresses #5552 (for GTK).

This PR adds the core infrastructure for keybind actions that are implemented as runtime app actions to be performable. This is done by having `rt_app.performAction` return a boolean. By default all runtime app actions return `true` (the action was performed) unless they are modified to return `true`/`false` as appropriate.

The GTK apprt is modified so that `goto_split`, `previous_tab`, `next_tab`, `last_tab`, and `goto_tab` are performable. macOS support will need to be added in a subsequent commit.

This doesn't completely solve the issue for the OP because if the `goto_split` isn't performable there is no fallback to`previous_tab` or `next_tab`.

I don't think that the approach taken in #5579 is the right one as it conflates split and tab navigation unconditionally which I don't think is what everyone would want. Either a separate action that explicitly combines the actions or a solution to #3175 will be the ultimate solution I believe.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -644,7 +644,7 @@ typedef void (*ghostty_runtime_write_clipboard_cb)(void*,
                                                    ghostty_clipboard_e,
                                                    bool);
 typedef void (*ghostty_runtime_close_surface_cb)(void*, bool);
-typedef void (*ghostty_runtime_action_cb)(ghostty_app_t,
+typedef bool (*ghostty_runtime_action_cb)(ghostty_app_t,
                                           ghostty_target_s,
                                           ghostty_action_s);
 
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -423,7 +423,7 @@ extension Ghostty {
 
         // MARK: Actions (macOS)
 
-        static func action(_ app: ghostty_app_t, target: ghostty_target_s, action: ghostty_action_s) {
+        static func action(_ app: ghostty_app_t, target: ghostty_target_s, action: ghostty_action_s) -> Bool {
             // Make sure it a target we understand so all our action handlers can assert
             switch (target.tag) {
             case GHOSTTY_TARGET_APP, GHOSTTY_TARGET_SURFACE:
@@ -431,7 +431,7 @@ extension Ghostty {
 
             default:
                 Ghostty.logger.warning("unknown action target=\(target.tag.rawValue)")
-                return
+                return false
             }
 
             // Action dispatch
@@ -541,10 +541,15 @@ extension Ghostty {
                 fallthrough
             case GHOSTTY_ACTION_QUIT_TIMER:
                 Ghostty.logger.info("known but unimplemented action action=\(action.tag.rawValue)")
-
+                return false
             default:
                 Ghostty.logger.warning("unknown action action=\(action.tag.rawValue)")
+                return false
             }
+
+            // If we reached here then we assume performed since all unknown actions
+            // are captured in the switch and return false.
+            return true
         }
 
         private static func quit(_ app: ghostty_app_t) {
--- a/src/App.zig
+++ b/src/App.zig
@@ -161,7 +161,7 @@ pub fn updateConfig(self: *App, rt_app: *apprt.App, config: *const Config) !void
     const applied: *const configpkg.Config = if (applied_) |*c| c else config;
 
     // Notify the apprt that the app has changed configuration.
-    try rt_app.performAction(
+    _ = try rt_app.performAction(
         .app,
         .config_change,
         .{ .config = applied },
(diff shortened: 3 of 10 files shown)


mitchellh:
Just noting I want to review this one due to the core implications.

jcollie:
> Just noting I want to review this one due to the core implications.

🫡

liby:
> I don't think that the approach taken in https://github.com/ghostty-org/ghostty/pull/5579 is the right one as it conflates split and tab navigation unconditionally which I don't think is what everyone would want.

As the author of PR #5579, I'd like to share some thoughts inspired by PR #5644. The elegant solution proposed in #5644 highlights a challenge that passionate contributors like myself often face: without access to higher-level architectural discussions and design principles, it's difficult to envision comprehensive solutions purely from issue descriptions. While I can implement functional fixes, arriving at more elegant, system-wide solutions requires a deeper understanding of the project's architectural vision.

After making numerous contributions to Ghostty through PRs and consistent engagement, I've been reflecting on how we could enhance our collaboration model. While the Discord development channel provides a great space for general discussions, I believe there's potential to create clearer pathways for long-term contributors to grow within the project. The recent discussion about [Subsystem Maintainers](https://discord.com/channels/1005603569187160125/1336783540993396808) seemed like a promising step in this direction, though I understand there may not be plans to expand the team at this time.

I share this perspective because I genuinely believe that creating growth opportunities for experienced contributors would help build a more sustainable project while maintaining Ghostty's high standards.

pluiedev:
My 2c would be that we should lean into the fact that a "tab" is just the root node of a binary tree that has "splits"/"surfaces" as leaf nodes. A tab without splits is just a root node with only one leaf attached. I don't see why they should be treated especially differently.

liby:
The tree structure perspective of tabs and splits is indeed insightful - it provides a clearer mental model for the system design.

@pluiedev You're right that I might have overemphasized the organizational structure. In fact, the point I would like to express more is to promote more collaborative technical discussions.

For example, this PR raises some valid points about the implementation in #5579 and leads to discussions about more elegant approaches. It would be wonderful if we could have had these architectural discussions directly in #5579. Such discussions would help contributors like myself better understand the project's technical direction and evolve our implementations accordingly. I believe this kind of direct technical dialogue and guidance would be more valuable than any formal roles in helping contributors grow and align better with the project's vision.

00-kat:
This is getting really off-topic at this point, can this conversation be moved to #5599 or https://discord.com/channels/1005603569187160125/1336783540993396808 :).

(Feel free to disagree if the conversation is related to this thread and I misinterpreted something.)

mitchellh:
> After making numerous contributions to Ghostty through PRs and consistent engagement, I've been reflecting on how we could enhance our collaboration model. While the Discord development channel provides a great space for general discussions, I believe there's potential to create clearer pathways for long-term contributors to grow within the project.

I believe the right place to have this discussion is within the discussions or issues. If you want to test whether you're angling at the right design or propose a core change to better handle future changes, a discussion or issue (if already open) is best. We sometimes kick over to Discord to have a more realtime conversation, but I try to enshrine those decisions into issues if we have them.

In this particular case, we never discussed anything, so all of these PRs are taking various stabs at the right solution. I don't even know the right solution yet. I noted in your PR why I disagreed with it but I also noted in this PR why I'm also suspicious of the approach, although it is closer to what I envisioned (and noted in the original issue as quoted below):

> We don't really have a way for the action system to query information or results out of it yet. Given actions are synchronous, we should probably add a result type to them and allow for stuff like this to return output.

mitchellh:
On further thought, another benefit of `bool` is that we can return false for unimplemented actions and therefore also detect that. That's something I've wanted in the past (to fallback to some default behavior) so I think that'll come in handy. That further suggests to me that simply flagging whether something is performed or not is a good thing.

Thanks @jcollie. I made some stylistic changes, made sure unimplemented actions return false, and got libghostty accepting bool return values (but haven't ported the macOS logic yet).

Will enable auto-merge.

jcollie:
I think that the macOS build was broken by this PR:

image

mitchellh:
Was that an old build? Latest is building fine. Oh it may be iOS.

PR #5646, opened 2025-02-09
Equalize Splits invalid keybind

rhodes-b:
refer to #4007

there are also quite a few posts about this in the discord help section

Basically for some reason for the equal key shift is not a typical modifier where it would be the original key + shift modifier but it actually just changes equal to respond as the plus key this is the only key on the keyboard I can see exhibit this behavior

it seems all 3 major operating systems report the key this way (mac, linux, and windows)

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2622,7 +2622,7 @@ pub fn default(alloc_gpa: Allocator) Allocator.Error!Config {
         );
         try result.keybind.set.put(
             alloc,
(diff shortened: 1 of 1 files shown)


rhodes-b:
as a side note I wanted to see if any others are messed up

increase_font_size is binded as `ctrl+plus` which is impossible because of this quirk and `ctrl+equal` (looking at the inspector) is the valid keybind to increase font size

mitchellh (approved):

PR #5647, opened 2025-02-09
Fix confirm-close-surface not working for hidden quick terminal

### LINKED ISSUE #5450: macOS: `confirm-close-surface` ignores hidden quick terminal
Discussion: #5445 

That discussion contains two issues. This is focused on one of the two. Reproduction:

1. `confirm-close-surface = always` (to make it easy to trigger)
2. Open Ghostty
3. View => Quick Terminal
4. Repeat to hide the quick terminal.
5. Cmd+Q

Expected: confirmation

Actual: no confirmation

liby:
Fixes #5450.

https://github.com/user-attachments/assets/7090fe55-dd1f-4517-9f8a-ffc3807283db

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -245,7 +245,13 @@ class AppDelegate: NSObject,
         // This probably isn't fully safe. The isEmpty check above is aspirational, it doesn't
         // quite work with SwiftUI because windows are retained on close. So instead we check
         // if there are any that are visible. I'm guessing this breaks under certain scenarios.
-        if (windows.allSatisfy { !$0.isVisible }) { return .terminateNow }
+        //
+        // NOTE(mitchellh): I don't think we need this check at all anymore. I'm keeping it
+        // here because I don't want to remove it in a patch release cycle but we should
+        // target removing it soon.
+        if (self.quickController == nil && windows.allSatisfy { !$0.isVisible }) {
+            return .terminateNow
+        }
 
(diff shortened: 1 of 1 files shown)


macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -245,7 +245,14 @@ class AppDelegate: NSObject,
         // This probably isn't fully safe. The isEmpty check above is aspirational, it doesn't
         // quite work with SwiftUI because windows are retained on close. So instead we check
         // if there are any that are visible. I'm guessing this breaks under certain scenarios.
-        if (windows.allSatisfy { !$0.isVisible }) { return .terminateNow }
+        if (windows.allSatisfy { !$0.isVisible }) {
```
mitchellh, line 248: Huh, this check is actually very problematic in general. If I'm reading this right, having hidden windows (Cmd+H) would bypass close confirmation as well.

I suggest we look into this line and redo this to maybe be based on `BaseTerminalController` inspection.

mitchellh (changes requested):
See feedback

macos/Sources/App/macOS/AppDelegate.swift
mitchellh: I was browsing the history of this line and its very suspicious, I think its to workaround an only retain cycle we had. I don't think this visible check needs to apply anymore. BUT, I don't want to rock the boat for 1.1.1 so I'm going to add a note to look at for 1.2 and get this in.

mitchellh (approved):

PR #5650, opened 2025-02-09
gtk(x11): set `WINDOWID` env var for subprocesses

pluiedev:
`WINDOWID` is the conventional environment variable for scripts that want to know the X11 window ID of the terminal, so that it may call tools like `xprop` or `xdotool`. We already know the window ID for window protocol handling, so we might as well throw this in for convenience.

Originally suggested by #4299

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -519,9 +519,17 @@ pub fn init(
     // This separate block ({}) is important because our errdefers must
     // be scoped here to be valid.
     {
+        var env_ = rt_surface.defaultTermioEnv() catch |err| env: {
+            // If an error occurs, we don't want to block surface startup.
+            log.warn("error getting env map for surface err={}", .{err});
+            break :env null;
+        };
+        errdefer if (env_) |*env| env.deinit();
+
         // Initialize our IO backend
         var io_exec = try termio.Exec.init(alloc, .{
             .command = command,
+            .env = env_,
             .shell_integration = config.@"shell-integration",
             .shell_integration_features = config.@"shell-integration-features",
             .working_directory = config.@"working-directory",
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -12,6 +12,7 @@ const objc = @import("objc");
 const apprt = @import("../apprt.zig");
 const font = @import("../font/main.zig");
 const input = @import("../input.zig");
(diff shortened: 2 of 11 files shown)


src/Surface.zig
```diff
@@ -540,6 +540,8 @@ pub fn init(
         });
         errdefer io_exec.deinit();
 
+        try self.rt_surface.addSubprocessEnv(&io_exec.subprocess.env);
```
mitchellh, line 543: I don't like that this traverses 3 completely separate areas of the codebase: core (Surface), termio, and apprt. Further `termio.Exec` isn't meant to be dug into like this, because we may have non-Exec termio implementations in the future (i.e. tmux control protocol).

I suggest we find an alternate approach to this one. Following an approach similar to the cgroup mechanism above is probably _better_, although I don't like that either much and I understand how that's messy with the existing process map.

I don't really have a better solution to consider right now, but just noting this so we don't merge yet.

mitchellh (changes requested):
See my comment on the `addSubprocessEnv` call...

src/Surface.zig
pluiedev: IMHO we already do way too much OS-specific/apprt specific things inside, such as removing GTK-specific env vars that wouldn't be present for non-GTK apprts, removing XCode env vars, handling .hushlogin, etc. It's a huge smorgasbord of platform checks and hacks that already make exposing libghostty as a neutral platform difficult. We should offload all that logic to apprts, even if it does mean there's a three-layer jump in between.

mitchellh: Indeed, but there’s a difference between isolated but perhaps domain inconsistent logic and directly accessing fields and decls across components. I’d like to think about a better approach to solve the latter. If it solves the former great but no need to boil the ocean to merge this PR. Just want to avoid the direct coupling.

mitchellh: Also just for fun as a point of history, the reason we set platform specific env cars today in termio is because it predates apprts :) it even predates being cross platform and started Linux only!

mitchellh:
@pluiedev Okay, cleaned it all up! Removed the coupling, made it so that apprts can advertise default base environments. This also cleans up all the macOS/GTK-specific env stuff in `termio/Exec`.

pluiedev:
Looks much more sensible and less like a hack! TYSM ❤️

Issue #5623, opened 2025-02-07
Command palette


pluiedev:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4195

<div type='discussions-op-text'>

<sup>Originally posted by **Xophmeister** December 31, 2024</sup>
In tmux, `Prefix+:` will open a command line input to allow you to enter arbitrary tmux commands. This is similar to Vim, where `:` enters command line mode.

It would be useful to have such a feature in Ghostty to try out actions, without having to define a keybinding for them.</div>

**Maintainer's Note**: Instead of a text interface with commands, it would be perhaps more idiomatic for a GUI application to display a pop up command palette instead, à la VS Code, IntelliJ IDEA, Blender, etc. There, we can also display default keybindings, which can be more easily learned by new users.

tristan957:
I agree with what you said. Have you looked at GNOME Builder's UX for this for the GTK apprt? I'm hoping this is something libadwaita will provide in the future, but should not block an initial implementation.

pluiedev:
Prior art:
- VS Code's command palette: https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette
- Superhuman's blog post on how they designed their command palette: https://blog.superhuman.com/how-to-build-a-remarkable-command-palette/

Working demo (GTK):

https://github.com/user-attachments/assets/e162c101-4f61-4ce0-a82c-75905ceb0fca

pluiedev:
> Have you looked at GNOME Builder's UX for this for the GTK apprt?

I haven't — there are some technical issues behind this that would make its inclusion in libadwaita somewhat difficult. (Essentially libadwaita doesn't support ListViews for its boxed list pattern, and ListBoxes don't really react dynamically enough to updating queries or are scalable enough for potentially a large number of entries.)

jcollie:
> Working demo (GTK):
> 2025-02-07_02-43-47.mp4

Can we fix the width of the window? I find the changing size as the results are narrowed down distracting.

mitchellh:
Great work on the GTK side, looks good!

How would parameterized commands work with this? I suggest we just keep it simple and follow the same format as expected for `keybind` because that is also the underlying core API we invoke. But how should we show it in the palette?

It'd be great to comptime generate a lot of the metadata that this uses (if we aren't already).

jcollie:
[...]

PR #5681, opened 2025-02-11
gtk: add command palette

pluiedev:
GTK part of #5623

Depends on #5560 because writing custom GObject subclasses in Zig without a nice wrapper is like hitting your head with a sturdy piece of pickled mackerel

Currently still WIP: I think the only way for someone not to go insane with implementing complex widgets like this is to use either GTK Builder `.ui` files or to use [Blueprint](https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/) to generate UI files as a resource


pluiedev:
Accidentally requested reviews from code owners: **do not review this yet please!**

src/input/Binding.zig
```diff
@@ -475,6 +475,8 @@ pub const Action = union(enum) {
     /// This currently only works on macOS.
     toggle_visibility: void,
 
+    toggle_action_palette,
```
mitchellh, line 478: Docs

src/apprt/gtk/ActionPalette.zig
```diff
@@ -0,0 +1,692 @@
...
+    title.setText("");
+    bind.setAccelerator("");
+}
+fn itemFactoryTeardown(_: *gtk.SignalListItemFactory, object: *gobject.Object, _: *ActionPalette) callconv(.C) void {
+    const item = gobject.ext.cast(gtk.ListItem, object).?;
+    item.setChild(null);
+}
+
+/// A selection of commands/actions that actually make sense in a GUI context
+const Command = enum(c_int) {
```
mitchellh, line 541: Looking at the shape of this, we can and should definitely generate this via comptime in core. I can take a stab at it. I don't see any reason this would be a burden and I think my approach as suggested would be _close_ to what works here.

libghostty also requires that any enum that makes it out into C ABI space is backed by C types and all strings are null-terminated so even if some mangling is needed in the GTK apprt, its probably minimal and now that I see how this works I can help shape it to that anyways.

src/apprt/action.zig
```diff
@@ -226,6 +226,8 @@ pub const Action = union(Key) {
     /// for changes.
     config_change: ConfigChange,
 
+    toggle_action_palette,
```
mitchellh, line 229: Docs

mitchellh (changes requested):
Surprisingly simple! I didn't heavily review the real meat of the action palette GTK stuff, just a cursory look. But I looked at the rest. 

I think we should try to find a way to limit scope on this and split it up to make it a bit easier. We already have the gobject PR that's separate which is nice. I think I want to do the command palette core work in a separate PR that you can then try to rebase this on top of. 

I also think introducing Blueprints (something I've always been supportive of!) as part of this makes it slightly more complicated too. Can we introduce blueprints for something simpler (say... the unsafe paste dialog or something) and get that in a separate PR so we can get all the machinery we need to work with blueprints within the project separate from a massive feature?

And finally, this should all of course wait for 1.2 😄

src/apprt/action.zig
pluiedev: It was just a placeholder

mitchellh:
Closing this due to tons of conflicts, but I'd love to see this revived and rebased on #7153 😄

PR #5696, opened 2025-02-11
macos: add a variety of artist-drawn alternate icons

mitchellh:
This is just a fun change to add a bunch of alternate icons. We don't want to add too many since this increases the final bundle size but we also want to have some fun. :)

Ghostty-More-Alts


cmomodo:
is there a command to list the available icons ?

corsinconzett:
@cmomodo in the commit you can see the names of the newly added icons:

    blueprint,
    chalkboard,
    microchip,
    glass,
    holographic,
    paper,
    retro,
    xray,

brabli:
@corsinconzett That isn't all of the available options though, from this error I got when using an incorrect icon name we can see that there is also "official" and "custom-style" so it makes sense that a command is added to show all the available options if one does not exist already.

ghostty_icons_error

corsinconzett:
@brabli sure 👍🏻 I wasn't debating whether or not a list command would make sense. I merely posted the new icon names (which is why I omitted the "official" and "custom-style" options) because I was looking for them as well.

JakeWharton:
They are available under the normal +show-config command for the 'macos-icon' property, or the website: https://ghostty.org/docs/config/reference#macos-icon

brabli:
Running `+show-config` by itself shows the current config values rather than showing all the available options.

Running `+show-config --docs` lists all the available options for everything, which is closer to what's being asked. Is that what you mean when you say the 'normal' `+show-config` command or am I being daft and not realising you can run `+show-config` (or another command) against specific properties to list all their available options?

JakeWharton:
Yes, sorry. By "under" I was referring to the fact that you can using the existing `+show-config` subcommand to obtain the list rather than Ghostty needing a new subcommand. But you will indeed need `--docs` (and probably also want `--default`) flags when running that subcommand to actually reach the documentation.

brabli:
Cool, thanks for clarifying!

PR #5699, opened 2025-02-12
macOS: use Nix to get Zig deps

jcollie:
This should help with API rate limits being hit by macOS builders since they can't use the Namespace cache that Linux builders can use to cache Zig dependencies. It will need to be run by Mitchell once so that the Cachix action can push everything up to the cache and then the full benefits should be seen. Not sure how using `--system` on all the macOS builds will affect things overall but it doesn't seem to have affected the CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -202,10 +202,14 @@ jobs:
       - name: XCode Select
         run: sudo xcode-select -s /Applications/Xcode_16.0.app
 
+      - name: get the Zig deps
+        id: deps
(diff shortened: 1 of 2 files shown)


mitchellh:
Awesome, this is a good idea and I'm willing to give it a try.

> It will need to be run by Mitchell once so that the Cachix action can push everything up to the cache and then the full benefits should be seen.

The CI just needs to successfully run once. The Cachix action automatically pushes the diffed store back up at the end of every run. Let's give it a shot!

PR #5710, opened 2025-02-12
gtk: fix context menu computed location

AnthonyZhOon:
When a tab bar is displayed, the context menu opened with right click is offset from the cursor.

This was due to using the incorrect coordinate space for describing where to draw the context menu.

[gtk_popover_set_pointing_to](https://docs.gtk.org/gtk4/method.Popover.set_pointing_to.html) uses the coordinate space of the popover's parent, however we used the active window's coordinate space which was noticeably different when the tab bar is visible.

Before:
Screenshot_12-Feb_22-39-07_com mitchellh ghostty
After:
Screenshot_12-Feb_22-41-41_com mitchellh ghostty-debug


pluiedev (approved):
Ouch. Makes sense.

pluiedev:
Open to merge if tests pass

PR #5720, opened 2025-02-12
terminal: reflow the saved cursor if we have one

### LINKED ISSUE #5718: Terminal resize with reflow doesn't reflow the saved cursor (ESC 7)
From #2783

The saved cursor (DECSC) should be properly reflowed. We properly reflow the primary cursor already. 

## Terminal Comparison

| Terminal      | Cursor Restoration Behavior                                      |
|--------------|------------------------------------------------------------------|
| Ghostty 1.1.0 :ghost: | Does not reflow the saved cursor |
| Kitty  | Reflows the cursor but its off by one, surely a bug                              |
| Gnome Console (libvte) | Reflows saved cursor      |
| iTerm2       | Does not reflow the saved cursor |
| WezTerm      | Does not reflow the saved cursor                                     |
| Terminal.app | Restores it in a nonsensical location 🤡                     |
| foot         | Reflows saved cursor                                                     |
| Alacritty    | Restores it in a nonsensical location (but different from Terminal.app!)        |

Note: I couldn't compare to xterm because xterm doesn't reflow the text.

## Background

When a terminal is resized with text reflow (i.e. soft-wrapped text), the cursor is generally reflowed with it. 

For example, imagine a terminal window 5-columns wide and you type the following without pressing enter. The cursor is on the X.

```
OOOOO
OOX
```

If you resize the window now to 8 or more columns, this happens, as expected:

```
OOOOOOOX
```

As expected, the cursor remains on the "X". This behaves like any other text input...

Terminals also provide an escape sequence to [save the cursor (ESC 7 aka DECSC)](https://ghostty.org/docs/vt/esc/decsc). This includes, amongst other things, the cursor position. The cursor can be restored with [DECRC](https://ghostty.org/docs/vt/esc/decrc). 

The behavior of the position of the _saved cursor_ in the context of text reflow is unspecified and varies wildly between terminals (see previous table above). Ghostty does this right now (as do many other terminals):

```
OOOOOOOO
   X
```

But this leads to broken things as noted in #2783. This issue proposes (and accepts) that we reflow the saved the cursor.

## Reproduction

```bash
#!/usr/bin/env bash

function csi() {
  echo -ne "\033[$1"
}

COLS=$(tput cols)

# Clear the screen and write enough text to wrap.
csi "2J"
csi "H"
for i in $(seq 1 $((COLS + 10))); do
  echo -n "O"
done

# Put the cursor over the "X"
echo -ne "X"
csi "D"

# Save the cursor
echo -ne "\x1b7"

# Move the cursor to the top-left
csi "H"

# Maximize the window at this point then press enter
read

# Restore the cursor
echo -ne "\x1b8"

read
```

mitchellh:
Fixes #5718

When a terminal is resized with text reflow (i.e. soft-wrapped text), the cursor is generally reflowed with it.

For example, imagine a terminal window 5-columns wide and you type the following without pressing enter. The cursor is on the X.

```
OOOOO
OOX
```

If you resize the window now to 8 or more columns, this happens, as expected:

```
OOOOOOOX
```

As expected, the cursor remains on the "X". This behaves like any other text input...

Terminals also provide an escape sequence to
[save the cursor (ESC 7 aka DECSC)](https://ghostty.org/docs/vt/esc/decsc). This includes, amongst other things, the cursor position. The cursor can be restored with [DECRC](https://ghostty.org/docs/vt/esc/decrc).

The behavior of the position of the _saved cursor_ in the context of text reflow is unspecified and varies wildly between terminals Ghostty does this right now (as do many other terminals):

```
OOOOOOOO
   X
```

This commit changes the behavior so that we reflow the saved cursor.

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -1590,6 +1590,18 @@ fn resizeInternal(
         self.cursor.hyperlink = null;
     }
 
+    // We need to insert a tracked pin for our saved cursor so we can
+    // modify its X/Y for reflow.
+    const saved_cursor_pin: ?*Pin = saved_cursor: {
+        const sc = self.saved_cursor orelse break :saved_cursor null;
+        const pin = self.pages.pin(.{ .active = .{
+            .x = sc.x,
+            .y = sc.y,
+        } }) orelse break :saved_cursor null;
+        break :saved_cursor try self.pages.trackPin(pin);
+    };
+    defer if (saved_cursor_pin) |p| self.pages.untrackPin(p);
+
     // Perform the resize operation.
     try self.pages.resize(.{
         .rows = rows,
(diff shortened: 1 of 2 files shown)


qwerasd205 (approved):
LGTM!

PR #5726, opened 2025-02-13
fix(termio): heap-allocate Subprocess arena so alloc doesn't get lost

qwerasd205:
As of #5715 the glfw apprt would crash on start because of a bad memory access when trying to deinit the env map because the glfw apprt doesn't return a default termio env, which means a new one is constructed with the arena alloc, and it ends up storing that alloc which itself stores a pointer to the stack-allocated arena object which is invalid after exiting `init`.

This allocates the arena object on the heap instead to fix that.

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -702,7 +702,7 @@ const Subprocess = struct {
         @cInclude("unistd.h");
     });
 
-    arena: std.heap.ArenaAllocator,
+    arena: *std.heap.ArenaAllocator,
     cwd: ?[]const u8,
     env: ?EnvMap,
     args: [][]const u8,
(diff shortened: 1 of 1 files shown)


rockorager (dismissed):

mitchellh:
This is kind of gross, because for the apprts that provide an env map (which is, all our production apprts), it forces an additional heap allocation for the arena now. A very minor cost, to be sure, but it still doesn't sit right with me.

Since both of the major apprts provide an env map already, I wonder if it'd be much cleaner to just move forward to changing the env map to be non-optional.

rockorager:
Forcing an apprt to return an env is much cleaner for libghostty consumers, too.

mitchellh:
Superseded by #5742

PR #5728, opened 2025-02-13
fix: inconsistent behavior from legacy and kitty Key encoding

exploitJ:
this fix addresses issue occurring in nvim/vim where Backspace deletes previous letter while in preedit state.

related to #1638

> same issue reproducible
> However, that only fixed for legacy encoding.

Unanswered Discussion https://github.com/ghostty-org/ghostty/discussions/5312<div type='discussions-op-text'> can be closed afterwards

tested on macos sequioa IME '2-set Korean', 'Japanese Romaji'


mitchellh (approved):

mitchellh:
Thanks! Added a unit test and tested it locally. Sorry for the delay.

PR #5733, opened 2025-02-13
build: add nix script to make cache tarball

jcollie:
Run `nix run .#make-cache` to create a directory `blob/` with a tarball of the Zig deps. Potentially suitable for uploading to R2 to make getting the Zig deps easier for people with bad connectivity to certain sites or for packagers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ glad.zip
 /Box_test_diff.ppm
 /ghostty.qcow2
 /build.zig.zon2json-lock
+/blob
--- a/flake.nix
+++ b/flake.nix
@@ -45,6 +45,8 @@
     builtins.foldl' nixpkgs-stable.lib.recursiveUpdate {} (
       builtins.map (
         system: let
+          version = "1.1.1";
+          revision = self.shortRev or self.dirtyShortRev or "dirty";
           pkgs-stable = nixpkgs-stable.legacyPackages.${system};
           pkgs-unstable = nixpkgs-unstable.legacyPackages.${system};
         in {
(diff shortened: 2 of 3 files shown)


flake.nix
```diff
@@ -99,6 +101,31 @@
             x11-gnome = runVM ./nix/vm/x11-gnome.nix;
             x11-plasma6 = runVM ./nix/vm/x11-plasma6.nix;
             x11-xfce = runVM ./nix/vm/x11-xfce.nix;
+            make-cache = let
```
mitchellh, line 104: I think we should move this into a file in `nix/` and import it rather than dumping it all in our flake.

flake.nix
```diff
@@ -99,6 +101,31 @@
...
+                  rm -rf "$WORK_DIR"
+                }
+                trap cleanup EXIT
+                ln -s "${self.packages.${system}.deps}" "$WORK_DIR/cache"
+                pushd "$WORK_DIR/cache" > /dev/null
+                tar --create --file "$WORK_DIR/temp.tar.gz" */*
+                popd > /dev/null
+                HASH=$(sha256sum "$WORK_DIR/temp.tar.gz" | awk "NR == 1 { print \$1 }; NR>1 { exit }")
+                mv "$WORK_DIR/temp.tar.gz" "blob/$HASH.tar.gz"
+                ln -sf "$HASH.tar.gz" "blob/ghostty-cache-${version}-${revision}.tar.gz"
```
mitchellh, line 123: I haven't run this yet, but if I'm reading this correctly it's named two things. I think I'd prefer to have this named `zig-cache.tar.gz` and let the caller deal with renaming it to what they see fit, its just so much easier with scripting that way.

.gitignore
```diff
@@ -19,3 +19,4 @@ glad.zip
 /Box_test_diff.ppm
 /ghostty.qcow2
 /build.zig.zon2json-lock
+/blob
```
mitchellh, line 22: Any reason we can't just dump this into the pwd and ignore the filename? "Blob" seems a bit random.

mitchellh (changes requested):

flake.nix
jcollie: It's done that way so that there isn't a 500+ MB file for each git commit. By naming it after the hash of the contents and then adding a symlink from the git commit space usage is minimized.

mitchellh: Sorry I mean just outputting one file and letting the release automation deal with what it wants to do otherwise. For uploading to R2 I think we need the physical file anyways, I'm not sure if the CI script (that we don't own/maintain) follows symlinks.

mitchellh:
If this file is 500+ MB like you said, that's not something I want to be mirroring and hosting for people. Let's go with the other approach of parsing the JSON and providing a download script. This is all just temporary anyways until `zig build --fetch` works properly.

PR #5738, opened 2025-02-13
macos: don't remove ctrl modifier for text input

### LINKED ISSUE #5448: macOS: Ctrl+L with Ergo-L keyboard layout doesn't produce a ctrl encoding
Steps to reproduce:

1. Install Ergo-L: https://ergol.org
2. Launch Ghostty
3. Run any keyboard input tracker, i.e. `kitty +kitten show_key -m kitty` (or any other)
4. Switch to Ergo-L
5. Ctrl+L (On US physical: Ctrl+H)

Notice the control key doesn't show up.

I've tracked this down to the code below, but this code exists for a reason. We should investigate why we do this, fix this, and ensure we don't regress the reason we did this in the first place.

https://github.com/ghostty-org/ghostty/blob/c5508e7d1922842ecd3160ea73b97da1282168b3/macos/Sources/Ghostty/SurfaceView_AppKit.swift#L853-L858

mitchellh:
Fixes #5448

We previously removed the ctrl modifier for text commit (IME-style) to workaround a libghostty quirk (as noted in the comment in the diff). But this broke other keyboard layouts.

This commit attempts to clean this up slightly -- but not completely -- by removing that hack, and only modifying the ctrl behavior for the UCKeyTranslate call.

Long term, I plan to remove UCKeyTranslate completely, as noted in the todo comment already written just below this diff.

This fixes the aforementioned issue and hopefully doesn't regress any other behavior. I tested the following:

  1. Dvorak Ctrl characters
  2. Ergo-L Ctrl characters
  3. US standard Ctrl characters
  4. Japanese IME input Ctrl input to modify IME state

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -849,28 +849,8 @@ extension Ghostty {
             var handled: Bool = false
             if let list = keyTextAccumulator, list.count > 0 {
                 handled = true
-
-                // This is a hack. libghostty on macOS treats ctrl input as not having
-                // text because some keyboard layouts generate bogus characters for
-                // ctrl+key. libghostty can't tell this is from an IM keyboard giving
-                // us direct values. So, we just remove control.
-                var modifierFlags = event.modifierFlags
-                modifierFlags.remove(.control)
-                if let keyTextEvent = NSEvent.keyEvent(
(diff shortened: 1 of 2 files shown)


mitchellh:
This caused #5743, so there was a case I didn't test. I haven't nailed it down completely, but its something to do with Kitty keyboard input since it only affects that protocol.

PR #5404, opened 2025-01-28
gtk: apply all window appearance changes in syncAppearance

pluiedev:
The GTK side of appearance code is kind of a mess with several different functions all having the responsibility of interacting with each other and setting the appropriate window appearance. It should solely be the responsibility of the `syncAppearance` function to apply appearance changes, with other callbacks/functions calling it instead: much like what we already do for the macOS apprt.

~~I also took the time to refactor the libadwaita version checks since calling `versionAtLeast(0, 0, 0)` does get old after a while. Now almost all checks are given human-readable names and contributors need not memorize what the relevant version checks all are.~~ Moved to another PR

PR #5741, opened 2025-02-13
gtk: use AdwAlertDialog for close dialogs, fix incorrect close dialogs

pluiedev:
AdwAlertDialog is the recommended way to do alert/message dialogs starting from libadwaita 1.5, and is much easier to manage than GtkMessageDialog. (The latter is also deprecated since GTK 4.10, but this PR does not migrate it to use GtkAlertDialog, mostly because of its obtuse interface and that we'll remove the GtkMessageDialog code anyway in 1.2 when we remove non-Adwaita builds.)

We also had two bugs where tabs with only one split would display the "close surface" confirmation dialog, and windows would do the same when closed via the "Close Window" menu item or by the `close_window` keybind action. (The "close window" dialog only appears when the user clicks on the close button on the titlebar.) Initially I was very confused by this, but it turns out that we don't have any apprt action related to closing a window, and it was simply closing surfaces...


src/apprt/gtk/App.zig
```diff
@@ -1508,7 +1517,6 @@ fn quitNow(self: *App) void {
         }
     }.callback, null);
 }
-
```
tristan957, line 1511: Oops!

tristan957 (commented):
I decided to take a look at what GNOME Console does in these situations:

image

image

Maybe we could align our wording a little more closely with what they do, except without the run-on sentence :smile:. One quick suggestion is to remove the `this`.

What do you think about waiting to merge this PR until after we remove vanilla GTK and add zig-gobject?

tristan957:
Maybe the second commit in this series should be its own PR? It doesn't seem related to GTK other than implementing the action which we could then do in a subsequent PR.

tristan957:
Could you also provide some screenshots so we can see everything put together?

pluiedev:
We've already added zig-gobject, and honestly IMO this can be the first step to removing non-Adwaita builds — an advantage of zig-gobject is that because the bindings are all statically generated extern functions, we don't actually need to link against libadwaita and it would not complain about missing symbols. Removing vanilla GTK support is as simple as just removing the vanilla GTK code (though, we might need to provide a fallback for older libadwaita versions using `AdwMessageDialog`, but the API is essentially the same)

pluiedev:
image
image
image
image

tristan957:
What's the difference between "close window" and "quit ghostty"? Also what are your thoughts on `Close` and `Cancel` terminology?

pluiedev:
Close window closes one window, quit Ghostty quits all windows. I think Close and Cancel would probably look better, but the current labels are just for the sake of being the same as vanilla GTK, which used Yes/No buttons.

tristan957:
> Close window closes one window, quit Ghostty quits all windows. I think Close and Cancel would probably look better, but the current labels are just for the sake of being the same as vanilla GTK, which used Yes/No buttons.
> 
> EDIT: The [GNOME HIG](https://developer.gnome.org/hig/patterns/feedback/dialogs.html) suggests using Cancel/Close. Will follow that then.

Loving the prose updates! I also like the idea of the `CloseTarget` you've added. Nice little abstraction.

src/Surface.zig
```diff
@@ -4242,7 +4242,11 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
 
         .close_surface => self.close(),
 
-        .close_window => self.app.closeSurface(self),
+        .close_window => return try self.rt_app.performAction(
```
mitchellh, line 4245: Ah! I remember the history here. Close window was implemented for macOS only for a time because its a common macOS-ism. The handling is all done in the apprt and not here, so I just put whatever here since nothing ever triggered it. That is, until GTK added a menu item to do so. 😄

src/apprt/gtk/App.zig
```diff
@@ -1438,6 +1440,16 @@ fn newWindow(self: *App, parent_: ?*CoreSurface) !void {
     try window.newTab(parent_);
 }
 
+fn closeWindow(_: *App, target: apprt.action.Target) void {
```
mitchellh, line 1443: Might as well make this performable and do a boolean return value from this and `performAction`. No reason not to.

src/apprt/gtk/Window.zig
```diff
@@ -598,6 +599,28 @@ pub fn focusCurrentTab(self: *Window) void {
...
+pub const CloseTarget = enum {
+    window,
+    tab,
+    surface,
+
+    fn title(self: CloseTarget) [:0]const u8 {
+        return switch (self) {
+            .window => "Close this Window?",
+            .tab => "Close this Tab?",
+            .surface => "Close this Surface?",
```
mitchellh, line 611: Recommend changing this to "Terminal" (the copy, not the enum). It will be more instantly recognizable to users I think than surface...

src/apprt/gtk/close_dialog.zig
```diff
@@ -0,0 +1,149 @@
...
+const Tab = @import("Tab.zig");
+const Surface = @import("Surface.zig");
+const adwaita = @import("adwaita.zig");
+
+const gobject = @import("gobject");
+const gio = @import("gio");
+const adw = @import("adw");
+const gtk = @import("gtk");
+
+pub const CloseDialog = extern struct {
```
mitchellh, line 14: I want to see more comments in this file, at the very least this structure and what its purpose is but the private data and so on could also use docs.

mitchellh (changes requested):
Looks great, just wanted to add a few of my notes.

src/apprt/gtk/Window.zig
pluiedev: "Terminal" is kind of a confusing term since it can refer to the entire terminal... That's part of the terminology problem I've been bringing up in discussions lately — we need a better term than "split" or "terminal" since the former isn't exactly accurate (a surface directly under a tab wouldn't be considered as a "split") and the latter is too vague. Maybe we should lean into iTerm2's terminology of panes and such?

src/apprt/gtk/close_dialog.zig
mitchellh: No I understand that, I mean the higher level docs on what this is and what fields should be set and so on. Just general structure docs.

mitchellh: I should add, just more generally, the coding style I have is _very comment heavy_. There are reasons for this and I can explain them another time, but making sure the code stays idiomatically the same throughout the project is important. 😄

src/apprt/gtk/Window.zig
tristan957: I agree that with Mitchell that surface probably shouldn't be surfaced to users. I'm not sure terminal is the right word to use, but I like panel over surface.

src/apprt/gtk/adwaita.zig
```diff
@@ -73,3 +73,8 @@ test "versionAtLeast" {
     try testing.expect(versionAtLeast(c.ADW_MAJOR_VERSION - 1, c.ADW_MINOR_VERSION, c.ADW_MICRO_VERSION + 1));
     try testing.expect(versionAtLeast(c.ADW_MAJOR_VERSION, c.ADW_MINOR_VERSION - 1, c.ADW_MICRO_VERSION + 1));
 }
+
+// Whether AdwDialog, AdwAlertDialog, etc. are supported (1.5+)
+pub fn supportsDialogs() bool {
+    return versionAtLeast(1, 5, 0);
+}
```
tristan957, line 80: I like this pattern you've introduced here. It will be beneficial to add similar functions throughout the codebase.

src/apprt/gtk/Window.zig
```diff
@@ -774,12 +775,15 @@ fn gtkRefocusTerm(v: *c.GtkWindow, ud: ?*anyopaque) callconv(.C) bool {
...
-
 fn gtkCloseRequest(v: *c.GtkWindow, ud: ?*anyopaque) callconv(.C) bool {
     _ = v;
     log.debug("window close request", .{});
     const self = userdataSelf(ud.?);
+    self.closeWithConfirmation();
+    return true;
+}
 
+pub fn closeWithConfirmation(self: *Window) void {
```
tristan957, line 786: I don't love the name of this function given the next comment. What about just close with documentation in a function level comment?

src/apprt/gtk/close_dialog.zig
```diff
@@ -0,0 +1,173 @@
...
+                .surface => "Close Surface?",
+            };
+        }
+
+        pub fn body(self: Target) [:0]const u8 {
+            return switch (self) {
+                .app => "All terminal sessions will be terminated.",
+                .window => "All terminal sessions in this window will be terminated.",
+                .tab => "All terminal sessions in this tab will be terminated.",
+                .surface => "The currently running process in this surface will be terminated.",
```
tristan957, line 129: Do we currently know what the running process is? See the GNOME Console pictures above.

src/apprt/gtk/close_dialog.zig
```diff
@@ -0,0 +1,173 @@
...
+    };
+};
+
+fn findActiveWindow(data: ?*const anyopaque, _: ?*const anyopaque) callconv(.C) c_int {
+    const window: *gtk.Window = @ptrCast(@alignCast(@constCast(data orelse return -1)));
+
+    // Confusingly, `isActive` returns 1 when active,
+    // but we want to return 0 to indicate equality.
+    // Abusing integers to be enums and booleans is a terrible idea, C.
+    return if (window.isActive() > 0) 0 else -1;
```
tristan957, line 172: `window.isActive()` returns an int in zig-gobject? I also don't love the `> 0`. Why not just compare to `1` or `c.TRUE`.

src/apprt/gtk/close_dialog.zig
```diff
@@ -0,0 +1,173 @@
...
+                .window => |v| gtk.Window.destroy(@ptrCast(v.window)),
+                .tab => |v| v.remove(),
+                .surface => |v| v.container.remove(),
+            };
+        }
+    };
+};
+
+fn findActiveWindow(data: ?*const anyopaque, _: ?*const anyopaque) callconv(.C) c_int {
+    const window: *gtk.Window = @ptrCast(@alignCast(@constCast(data orelse return -1)));
```
tristan957, line 167: These 3 casts are so ugly haha

src/apprt/gtk/Surface.zig
```diff
@@ -717,7 +719,12 @@ pub fn redraw(self: *Surface) void {
 }
 
 /// Close this surface.
-pub fn close(self: *Surface, processActive: bool) void {
+pub fn close(self: *Surface, process_active: bool) void {
+    self.closeWithConfirmation(process_active, .{ .surface = self });
+}
```
tristan957, line 724: Do we need this function to satisfy an interface or something? It doesn't seem to add much much over the next function.

src/apprt/gtk/Surface.zig
```diff
@@ -717,7 +719,12 @@ pub fn redraw(self: *Surface) void {
...
+
+/// Close this surface.
+pub fn closeWithConfirmation(self: *Surface, process_active: bool, target: CloseDialog.Target) void {
```
tristan957, line 727: Given this is the surface, is our target not always the surface?

src/apprt/gtk/close_dialog.zig
mitchellh: Not yet. That’s some core work that needs to be done. It’ll require some upstream changes in xev too. We need this for many planned features I want to build…

src/apprt/gtk/close_dialog.zig
pluiedev: I mean, you could write it as `@intFromBool(!(window.isActive() == 0))` but that's just even harder to understand what it does

src/apprt/gtk/Surface.zig
pluiedev: Nope! When a tab with just one split is closed it'll call this function with the target set to the tab

src/apprt/gtk/adwaita.zig
pluiedev: It came from #5404 :p

src/apprt/gtk/close_dialog.zig
tristan957: > I mean, you could write it as `@intFromBool(!(window.isActive() == 0))` but that's just even harder to understand what it does. GLib's boolean semantics are "0 if false, anything else if true", so it's theoretically possible that a truthy value won't be exactly 1.

If you want to be pedantic, sure, butthe function is documented as returning `TRUE`: https://docs.gtk.org/gtk4/method.Window.is_active.html

jcollie: I generally prefer `!= 0` when comparing C truthy values, as `0` is `false` and anything else is `true`.

jcollie:
@pluiedev if you can resolve the conflicts and take a look at the open comments I'd like to review this one next.

src/apprt/gtk/Window.zig
pluiedev: Gonna call it "Close Split?" for now as it fits with the rest of the GUI actions (we could also add it to the menus under the Split category)

src/apprt/gtk/CloseDialog.zig
```diff
@@ -0,0 +1,164 @@
...
+) callconv(.C) void {
+    const dialog = gobject.ext.cast(DialogType(), source_object.?).?;
+    const target: *Target = @ptrCast(@alignCast(ud.?));
+    const resp = dialog.chooseFinish(result);
+
+    if (std.mem.orderZ(u8, resp, "close") == .eq) {
+        target.close();
+    }
+
+    target.allocator().destroy(target);
```
jcollie, line 84: Getting the following crashes when I close a split or a tab:

```
info(surface): surface closed addr=7fffdc074000
General protection exception (no address available)
/home/jeff/dev/ghostty/src/apprt/gtk/CloseDialog.zig:142:33: 0x32d0042 in allocator (ghostty)
            .tab => |v| v.window.app.core_app.alloc,
                                ^
/home/jeff/dev/ghostty/src/apprt/gtk/CloseDialog.zig:84:21: 0x32d02b9 in closeCallback (ghostty)
    target.allocator().destroy(target);
                    ^
```

```
info(surface): surface closed addr=7fffdc0c0000
Segmentation fault at address 0x7fffdc0cf648
/home/jeff/dev/ghostty/src/apprt/gtk/CloseDialog.zig:143:30: 0x32d005c in allocator (ghostty)
            .surface => |v| v.app.core_app.alloc,
                             ^
/home/jeff/dev/ghostty/src/apprt/gtk/CloseDialog.zig:84:21: 0x32d02b9 in closeCallback (ghostty)
    target.allocator().destroy(target);
                    ^
???:?:?: 0x7ffff70c1333 in ??? (libgio-2.0.so.0)
Unwind information for `libgio-2.0.so.0:0x7ffff70c1333` was not available, trace may be incomplete

???:?:?: 0x0 in ??? (???)
```

src/build/SharedDeps.zig
```diff
@@ -441,6 +441,7 @@ pub fn add(
                     .{ "gobject", "gobject2" },
                     .{ "gio", "gio2" },
                     .{ "glib", "glib2" },
+                    .{ "gio", "gio2" },
```
jcollie, line 444: This is a duplicate.

jcollie (changes requested):

src/apprt/gtk/App.zig
```diff
@@ -1400,60 +1402,19 @@ fn quit(self: *App) void {
     // If we're already not running, do nothing.
     if (!self.running) return;
 
-    // If we have no toplevel windows, then we're done.
```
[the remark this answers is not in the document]
pluiedev: I think I know why. Somehow `quitNow` never sets `self.running` to false???

src/apprt/gtk/CloseDialog.zig
pluiedev: I somehow never encountered this, but I guess I can just get the allocator before we close anything

jcollie (approved):

mitchellh (approved):

PR #5749, opened 2025-02-14
gtk: require libadwaita

jcollie:
This commit removes support for building without libadwaita. (Y'all knew that I just had this sitting in my back pocket). This will need some serious review to ensure that we haven't lost any functionality.


src/apprt/gtk/App.zig
```diff
@@ -237,23 +244,15 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
     }
 
     c.gtk_init();
+    c.adw_init();
```
tristan957, line 247: Don't call `gtk_init()` https://gitlab.gnome.org/GNOME/libadwaita/-/blob/649967044e269a0118ae09a9ff0ecba90cc58300/NEWS?page=2#L1549

src/apprt/gtk/Window.zig
```diff
@@ -87,37 +83,28 @@ pub fn init(self: *Window, app: *App) !void {
...
-    self.window = gtk_window;
-    c.gtk_window_set_title(gtk_window, "Ghostty");
-    c.gtk_window_set_default_size(gtk_window, 1000, 600);
-    c.gtk_widget_add_css_class(@ptrCast(gtk_window), "window");
-    c.gtk_widget_add_css_class(@ptrCast(gtk_window), "terminal-window");
+    self.window = @ptrCast(@alignCast(gtk_widget));
+
+    c.gtk_window_set_title(self.window, "Ghostty");
+    c.gtk_window_set_default_size(self.window, 1000, 600);
+    c.gtk_widget_add_css_class(gtk_widget, "adw");
```
tristan957, line 93: I think we should remove this CSS class.

src/apprt/gtk/Window.zig
```diff
@@ -407,19 +365,16 @@ pub fn syncAppearance(self: *Window, config: *const configpkg.Config) !void {
...
-                @ptrCast(tab_overview),
-                @intFromBool(csd_enabled),
-            );
-            c.adw_tab_overview_set_show_end_title_buttons(
-                @ptrCast(tab_overview),
-                @intFromBool(csd_enabled),
-            );
-        }
+    if (self.tab_overview) |tab_overview| {
+        if (!adwaita.versionAtLeast(1, 4, 0)) unreachable;
```
tristan957, line 369: Should you move this if outside the other one?

src/apprt/gtk/headerbar.zig
```diff
@@ -1,58 +1,60 @@
...
-        }
-    }
-
-    pub fn setSubtitle(self: HeaderBar, subtitle: [:0]const u8) void {
-        switch (self) {
-            inline else => |v| v.setSubtitle(subtitle),
-        }
-    }
-};
+// const adwaita = @import("adwaita.zig");
```
tristan957, line 7: remove the comment

src/apprt/gtk/notebook.zig
```diff
@@ -1,169 +1,193 @@
+/// An abstraction over the GTK notebook and Adwaita tab view to manage
+/// all the terminal tabs in a window.
+const Notebook = @This();
```
tristan957, line 3: Instead of notebook which is GTK parlance. Let's just use TabView. Maybe do this in a separate commit to help with diff size.

src/apprt/gtk/notebook.zig
```diff
@@ -1,169 +1,193 @@
...
-pub const Notebook = union(enum) {
-    adw: NotebookAdw,
-    gtk: NotebookGtk,
+/// Set to true so that the adw close-page handler knows we're forcing
+/// and to allow a close to happen with no confirm. This is a bit of a hack
+/// because we currently use GTK alerts to confirm tab close and they
+/// don't carry with them the ADW state that we are confirming or not.
+/// Long term we should move to ADW alerts so we can know if we are
+/// confirming or not.
+forcing_close: bool = false,
```
tristan957, line 24: cc @pluiedev here

src/config/Config.zig
```diff
@@ -2153,28 +2147,14 @@ keybind: Keybinds = .{},
...
-/// This requires a GTK4 desktop with a GTK4 theme.
-///
-/// If you are running GTK3 or have a GTK3 theme, you may have to set this
-/// to false to get your theme picked up properly. Having this set to true
-/// with GTK3 should not cause any problems, but it may not work exactly as
-/// expected.
-///
-/// This configuration only has an effect if Ghostty was built with
-/// Adwaita support.
-@"gtk-adwaita": bool = true,
```
tristan957, line 2176: Maybe it would be easier for users if we just blackholed this option and warned in the logs?

src/config/Config.zig
```diff
@@ -5752,13 +5732,11 @@ pub const GtkSingleInstance = enum {
 pub const GtkTabsLocation = enum {
     top,
     bottom,
-    left,
-    right,
```
tristan957, line 5756: Again, thinking about users, maybe we should keep these around and default them to top?

src/config/Config.zig
jcollie: I'd bet that a lot of users will never see the logs, so I think that we need something noisier than that. I think that options that just silently do nothing will frustrate more people that don't read the release notes. I think that it's going to become increasingly more common to have people that install Ghostty from their package manager and launch it from a GUI desktop icon so they may not even know where to look for the logs.

src/config/Config.zig
jcollie: As above, I'd rather be noisy about it so that people know why their tabs aren't showing up where they expect them to.

src/apprt/gtk/notebook.zig
jcollie: +1 on switching to Adwaita alerts.

src/apprt/gtk/notebook.zig
jcollie: +1 on keeping diff size down.

src/apprt/gtk/Window.zig
mitchellh: Normally I'd say for the sake of backwards compatibility we should just keep it since it costs nothing, but I don't think anyone is actively relying on it and the fix is simple enough (remove the selector).

src/apprt/gtk/Window.zig
mitchellh: Also as a note `if (x) unreachable` is equivalent to `assert(!x)`, so we should just use an assert.

src/config/Config.zig
mitchellh: Again, normally I'm very much in the camp of bending over backwards to support backwards compatibility. In this case, we've explicitly warned for a couple releases now that this is coming and that we will be breaking compatibility, so I think in this case, we should hard remove as @jcollie has done and have the config warnings popup come up.

src/apprt/gtk/notebook.zig
pluiedev: Agreed

src/apprt/gtk/notebook.zig
jcollie: Even better, let's make it a separate PR. Maybe at the same time switch to zig-gobject.

src/apprt/gtk/headerbar.zig
jcollie: done

src/apprt/gtk/Window.zig
jcollie: switched to `assert()` where appropriate.

src/apprt/gtk/Window.zig
jcollie: Done.

src/apprt/gtk/App.zig
jcollie: Done.

jcollie:
OK, assuming that CI passes, I think that this is ready to merge. This is really going to f*** up most existing PRs that touch GTK code. Better to rip that bandaid off so to speak IMHO.

mitchellh:
I defer to you all. I'm ready to see it happen when you are.

pluiedev (approved):
Just did a quick skim and I think the diff itself looks good in terms of just removing vanilla GTK mechanisms. I think there should be more follow up PRs that clean up other stuff (e.g. renaming notebook to tab view, respecting system light/dark preferences in AdwStyleManager, etc.) but we need not do them now.

PR #5750, opened 2025-02-14
macos: add `padded-notch` option for `macos-non-native-fullscreen`

mitchellh:
Finishes #378
Supercedes #4159

This adds a new enum value for `macos-non-native-fullscreen`: `padded-notch`. This value will add padding to the top of the window to account for the notch on applicable devices while still hiding the menu.

This value is preferred over "visible-menu" by some people because for screens without a notch, the window will take up the full height.

The plan in the future is that we may color the padded area when a notch is present. In this commit it appears as transparent.

EDIT: Hah, the video below is pretty useless since the screen recording software omits the notch. 😄 

https://github.com/user-attachments/assets/4b1e84dd-3b3a-44a8-a83b-0f51e44f6cf8


aslushnikov:
This is awesome, thank you!

krzkrzkrz:
Thank you! This is very much appreciated!

PR #5763, opened 2025-02-14
build: script to download deps

jcollie:
Run `python src/build/deps.py`. That will create a directory `deps` that contain tarballs of all the Zig deps that don't use `git+https`. Suitable for uploading to R2.


mitchellh:
Superseded by #5764 (your commit went there)

PR #5765, opened 2025-02-14
build: mirror most of our direct dependencies

mitchellh:
This adds a new script `update-mirror.sh` which generates the proper blob format for R2 (or any blob storage) to mirror all of our dependencies.

It doesn't automate updating build.zig.zon but on an ongoing basis this should be easy to do manually, and we can strive to automate it in the future.

I omitted iTerm2 color themes because we auto-update that via CI and updating all of the machinery to send it to our mirror and so on is a pain. Additionally, this doesn't mirror transitive dependencies because Zig doesn't have a way to fetch those from a mirror instead (unless you pre-generate a full cache like packagers but that's not practical for day to day development).

It's hugely beneficial just to get most of our dependencies mirrored.


build.zig.zon
```diff
@@ -4,46 +4,56 @@
...
+            // natecraddock/zf
+            .url = "https://deps.files.ghostty.org/zf-1220edc3b8d8bedbb50555947987e5e8e2f93871ca3c8e8d4cc8f1377c15b5dd35e8.tar.gz",
             .hash = "1220edc3b8d8bedbb50555947987e5e8e2f93871ca3c8e8d4cc8f1377c15b5dd35e8",
         },
         .gobject = .{
-            .url = "https://github.com/ianprime0509/zig-gobject/releases/download/v0.2.2/bindings-gnome47.tar.zst",
-            .hash = "12208d70ee791d7ef7e16e1c3c9c1127b57f1ed066a24f87d57fc9f730c5dc394b9d",
+            // ianprime0509/zig-gobject
+            .url = "https://deps.files.ghostty.org/gobject-12202912acaf613bf664b33148e4dcb45ab660f8688ae8f8f1eb2dea8e2833810f12.tar.gz",
+            .hash = "12202912acaf613bf664b33148e4dcb45ab660f8688ae8f8f1eb2dea8e2833810f12",
```
mitchellh, line 56: This is the only one where the hash changed since we switched from `tar.zst` to `tar.gz` to make our mirror script easier. The content hash matches upstream.

mitchellh: Nvm the release is actually processed so we need to support zst here.

PR #5775, opened 2025-02-15
macos: add +list-mac-app-icons cli action

sanyer:
This adds a new CLI command `+list-mac-app-icons` that lists all available macOS app icons added in #5696 which can be used with the `macos-icon` configuration option.

## Details
- Adds new `list-mac-app-icons` command to list available macOS app icons
- Only available on macOS and returns an error message on other platforms
- Follows existing CLI command patterns and code structure

## Example Usage
```sh
Ghostty.app/Contents/MacOS/ghostty +list-mac-app-icons
official
blueprint
chalkboard
microchip
...
```

## Testing
- Tested on macOS: successfully lists available icons
- Tested on Linux: displays appropriate "only supported on macOS" message
- Verified help text functionality works as expected

### DIFF
--- a/src/cli/action.zig
+++ b/src/cli/action.zig
@@ -9,6 +9,7 @@ const list_keybinds = @import("list_keybinds.zig");
 const list_themes = @import("list_themes.zig");
 const list_colors = @import("list_colors.zig");
 const list_actions = @import("list_actions.zig");
+const list_mac_app_icons = @import("list_mac_app_icons.zig");
 const show_config = @import("show_config.zig");
 const validate_config = @import("validate_config.zig");
 const crash_report = @import("crash_report.zig");
@@ -39,6 +40,9 @@ pub const Action = enum {
 
     /// List keybind actions
     @"list-actions",
+
+    /// List available macOS app icons
+    @"list-mac-app-icons",
 
     /// Dump the config to stdout
     @"show-config",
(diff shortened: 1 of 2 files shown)


00-kat:
But... why?  Do the docs of [`macos-icon`] not suffice?

[`macos-icon`]: https://ghostty.org/docs/config/reference#:~:text=blueprint%2C%20chalkboard%2C%20microchip%2C%20glass%2C%20holographic%2C%20paper%2C%20retro%2C%20xray

mitchellh:
Thanks for the PR.

I'm also not so sure about this. As it stands, this PR isn't doing much more than the docs can do. There are already plans underway to make `+show-config` be able to show possible enum values for multiple choice configs, so that would be another place this is solved, imo.

I think if this command were expanded to use the Kitty image protocol or something to show the possible icons visually, similar to the theme list, then it'd be more interesting. But as a plaintext list... I'm not sure I see the value, sorry.

Feel free to correct me if I'm wrong in my interpretation, I'll read the followups.

sanyer:
> show possible enum values for multiple choice configs

this sounds like a good plan, eliminating the need for this PR, yes!

sanyer:
> use the Kitty image protocol

oh, this one would indeed be better, let me see what I can do

PR #5781, opened 2025-02-15
gtk: update menus to use popovers and builder ui files

### LINKED ISSUE #5779: GTK: Add "Split Left" and "Split Up" actions to the menubar/context menu
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5535

<sup>Originally posted by **edrozenberg** February  2, 2025</sup>
Ghostty 1.1.0

Currently on Mac, Ghostty shows "Split Right" and "Split Down" in the menubar.

Suggestion:
Add "Split Left" and "Split Up" to the menubar even if by default they might have no keybindings assigned.

In my Ghostty `config` I'm using the following on Mac. I decided to use **e** and **o** bindings, similar to Linux, bc I personally remember these more easily (v **e** rtical) / (h **o** rizontal).
```
keybind = cmd+e=new_split:right
keybind = shift+cmd+e=new_split:left
keybind = cmd+o=new_split:down
keybind = shift+cmd+o=new_split:up
```
The modified "Right" and "Down" bindings show up properly in the Mac menubar, but "Left" and "Up" don't show up at all, even though they are active and function as defined in the `config`.

Motivation: in many cases, the user may need to split to the left or up to organize their workspace around existing content (an already open editor etc). If these options are not available in the menubar, the user may not realize this is possible, unless they specifically enable keybindings for these actions in the `config` file. And even when they add these bindings to the `config` file, they are not currently exposed in the menubar and may be overlooked/forgotten.

I assume but haven't checked that something similar might make sense on Linux, if these menu options are also missing there.</div>

<hr/>

**Maintainer's note:** This is missing on Linux also. #4952 attempted to fix it but it was caught in a much wider array of changes. I think it would be better if we can have one or two focused PRs that would fix this for Linux and/or macOS instead.

jcollie:
Menus (context menus and the window hamburger menu) now use popovers and are defined using GTK builder UI files. This is a bit more "modern" and reduces the amount of code to define menus.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -1083,6 +1083,8 @@ fn gtkTitlebarMenuActivate(
     _: *c.GParamSpec,
     ud: ?*anyopaque,
 ) callconv(.C) void {
+    // debian 12 is stuck on GTK 4.8
+    if (!version.atLeast(4, 10, 0)) return;
     const active = c.gtk_menu_button_get_active(btn) != 0;
     const self = userdataSelf(ud orelse return);
     if (active) {
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -18,6 +18,7 @@ const CoreSurface = @import("../../Surface.zig");
 const App = @import("App.zig");
 const Color = configpkg.Config.Color;
 const Surface = @import("Surface.zig");
+const Menu = @import("menu.zig").Menu;
 const Tab = @import("Tab.zig");
 const c = @import("c.zig").c;
 const adwaita = @import("adwaita.zig");
@@ -46,7 +47,8 @@ tab_overview: ?*c.GtkWidget,
 /// The notebook (tab grouping) for this window.
 notebook: TabView,
 
-context_menu: *c.GtkWidget,
+/// The "main" menu that is attached to a button in the headerbar.
+titlebar_menu: Menu(Window, "titlebar_menu", true),
 
 /// The libadwaita widget for receiving toast send requests.
 toast_overlay: *c.GtkWidget,
@@ -112,7 +114,7 @@ pub fn init(self: *Window, app: *App) !void {
         .headerbar = undefined,
         .tab_overview = null,
         .notebook = undefined,
-        .context_menu = undefined,
+        .titlebar_menu = undefined,
         .toast_overlay = undefined,
         .winproto = .none,
     };
@@ -136,6 +138,9 @@ pub fn init(self: *Window, app: *App) !void {
 
     // Create our box which will hold our widgets in the main content area.
     const box = c.gtk_box_new(c.GTK_ORIENTATION_VERTICAL, 0);
+
+    // Set up the menus
+    self.titlebar_menu.init(self);
 
     // Setup our notebook
     self.notebook.init(self);
@@ -174,7 +179,15 @@ pub fn init(self: *Window, app: *App) !void {
         const btn = c.gtk_menu_button_new();
         c.gtk_widget_set_tooltip_text(btn, "Main Menu");
         c.gtk_menu_button_set_icon_name(@ptrCast(btn), "open-menu-symbolic");
-        c.gtk_menu_button_set_menu_model(@ptrCast(btn), @ptrCast(@alignCast(app.menu)));
+        c.gtk_menu_button_set_popover(@ptrCast(btn), @ptrCast(@alignCast(self.titlebar_menu.asWidget())));
+        _ = c.g_signal_connect_data(
+            btn,
+            "notify::active",
+            c.G_CALLBACK(&gtkTitlebarMenuActivate),
+            self,
+            null,
+            c.G_CONNECT_DEFAULT,
+        );
         self.headerbar.packEnd(btn);
     }
 
@@ -259,11 +272,6 @@ pub fn init(self: *Window, app: *App) !void {
         c.adw_tab_overview_set_view(@ptrCast(tab_overview), @ptrCast(@alignCast(self.notebook.tab_view)));
     }
 
-    self.context_menu = c.gtk_popover_menu_new_from_model(@ptrCast(@alignCast(self.app.context_menu)));
-    c.gtk_widget_set_parent(self.context_menu, box);
-    c.gtk_popover_set_has_arrow(@ptrCast(@alignCast(self.context_menu)), 0);
-    c.gtk_widget_set_halign(self.context_menu, c.GTK_ALIGN_START);
-
     // We register a key event controller with the window so
     // we can catch key events when our surface may not be
     // focused (i.e. when the libadw tab overview is shown).
@@ -272,7 +280,6 @@ pub fn init(self: *Window, app: *App) !void {
     c.gtk_widget_add_controller(gtk_widget, ec_key_press);
 
     // All of our events
-    _ = c.g_signal_connect_data(self.context_menu, "closed", c.G_CALLBACK(&gtkRefocusTerm), self, null, c.G_CONNECT_DEFAULT);
     _ = c.g_signal_connect_data(self.window, "realize", c.G_CALLBACK(&gtkRealize), self, null, c.G_CONNECT_DEFAULT);
     _ = c.g_signal_connect_data(self.window, "close-request", c.G_CALLBACK(&gtkCloseRequest), self, null, c.G_CONNECT_DEFAULT);
     _ = c.g_signal_connect_data(self.window, "destroy", c.G_CALLBACK(&gtkDestroy), self, null, c.G_CONNECT_DEFAULT);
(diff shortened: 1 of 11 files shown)


src/apprt/gtk/notebook.zig
```diff
@@ -1,5 +1,5 @@
-/// An abstraction over the GTK notebook and Adwaita tab view to manage
-/// all the terminal tabs in a window.
+/// An abstraction over the Adwaita tab view to manage all the terminal tabs in
```
pluiedev, line 1: Why is this diff here lol

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,136 @@
...
+                Window => "window",
+                Surface => "surface",
+                else => unreachable,
+            };
+            const parent: *T = @alignCast(@fieldParentPtr(variant, self));
+
+            var builder = Builder.init("menu-" ++ name ++ "-" ++ variant, .ui);
+            defer builder.deinit();
+
+            const object = builder.getObject("menu") orelse unreachable;
```
pluiedev, line 44: orelse unreachable is just .?

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,136 @@
...
+            const menu_model: *gio.MenuModel = @ptrCast(@alignCast(object));
```
pluiedev, line 45: Use gobject.ext.cast

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,136 @@
...
+                    _ = gtk.Popover.signals.closed.connect(
+                        menu_widget,
+                        *Self,
+                        gtkRefocusTerm,
+                        self,
+                        .{},
+                    );
+                    break :brk menu_widget;
+                },
+                .popover_menu_bar => brk: {
```
pluiedev, line 62: Why is this block necessary?

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,136 @@
...
+                .parent = parent,
+                .menu_widget = menu_widget,
+            };
+        }
+
+        pub fn setParent(self: *const Self, widget: *c.GtkWidget) void {
+            self.menu_widget.as(gtk.Widget).setParent(@ptrCast(@alignCast(widget)));
+        }
+
+        pub fn asWidget(self: *const Self) *c.GtkWidget {
```
pluiedev, line 77: We should try to avoid c.Gtk* types in new code anyway. Any old code that still expects those types should call @ptrCast themselves

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,136 @@
...
+            };
+
+            const action_map: *gio.ActionMap = @ptrCast(@alignCast(window));
+            const action: *gio.SimpleAction = @ptrCast(@alignCast(action_map.lookupAction("copy") orelse return));
+            action.setEnabled(@intFromBool(has_selection));
+        }
+
+        pub fn popupAt(self: *const Self, x: f64, y: f64) void {
+            const rect: gdk.Rectangle = .{
+                .f_x = @intFromFloat(x),
```
pluiedev, line 114: We should let the caller convert the float to an int on their side. Accepting a floating point value and immediately converting it to an integer is a bit disingenuous.

pluiedev (changes requested):

src/apprt/gtk/notebook.zig
jcollie: ¯\_(ツ)_/¯

src/apprt/gtk/menu.zig
jcollie: yeah, i guess i just find `.?` ugly but it doesn't matter to me.

src/apprt/gtk/menu.zig
jcollie: That's leftover from when I was testing menu bars. Removed.

src/apprt/gtk/menu.zig
jcollie: Done.

src/apprt/gtk/menu.zig
jcollie: Done.

src/apprt/gtk/menu.zig
jcollie: TIL.

src/apprt/gtk/ui/menu-surface-context_menu.ui
```diff
@@ -0,0 +1,90 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
tristan957, line 1: is it possible to use double quotes here too?

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,125 @@
...
+const Builder = @import("Builder.zig");
+
+const gtk = @import("gtk");
+const gdk = @import("gdk");
+const gio = @import("gio");
+const gobject = @import("gobject");
+
+const log = std.log.scoped(.gtk_menu);
+
+pub fn Menu(
```
mitchellh, line 16: Too few (zero!) comments in this entire file. The trivial functions don't need doc comments but almost everything else should.

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,125 @@
...
+const apprt = @import("../../apprt.zig");
+const App = @import("App.zig");
+const Window = @import("Window.zig");
+const Surface = @import("Surface.zig");
+const Builder = @import("Builder.zig");
+
+const gtk = @import("gtk");
+const gdk = @import("gdk");
+const gio = @import("gio");
+const gobject = @import("gobject");
```
mitchellh, line 12: A nit but these imports don't follow the idiom we follow in the rest of the codebase which is generally ordered as:

```
<std>
<packages>
<local>
```

src/apprt/gtk/menu.zig
```diff
@@ -0,0 +1,125 @@
...
+        parent: *T,
+        menu_widget: *MenuWidget,
+
+        pub fn init(self: *Self) void {
+            const name = switch (T) {
+                Window => "window",
+                Surface => "surface",
+                else => unreachable,
+            };
+            const parent: *T = @alignCast(@fieldParentPtr(variant, self));
```
mitchellh, line 37: I implore us to reconsider `@fieldParentPtr` if we can. The use of it is a huge smell that something is wrong. I noted this in another PR. I don't have enough context on the gobject work to know if we must do this, but let's think about it. I'll take a closer look another time (feel free to not block this merge on it).

mitchellh (changes requested):

src/apprt/gtk/menu.zig
jcollie: Added some comments.

src/apprt/gtk/menu.zig
jcollie: Removed.

src/apprt/gtk/menu.zig
jcollie: Fixed.

src/apprt/gtk/ui/menu-surface-context_menu.ui
jcollie: Yes.

gtjoseph:
BONUS!  Just FYI...  This PR eliminated the `Gtk-CRITICAL **: gtk_widget_unparent: assertion 'GTK_IS_WIDGET' failed` message when the context menu had never been shown when ghostty shuts down.

tristan957 (approved):

src/apprt/gtk/Window.zig
```diff
@@ -883,11 +880,7 @@ fn gtkActionClose(
     ud: ?*anyopaque,
 ) callconv(.C) void {
     const self: *Window = @ptrCast(@alignCast(ud orelse return));
-    const surface = self.actionSurface() orelse return;
-    _ = surface.performBindingAction(.{ .close_surface = {} }) catch |err| {
```
pluiedev, line 887: I'm speculating as to why this is here, but if you're also running into the issue where the window is closed and freed twice, then I would probably suggest to remove the logic for closing the window inside the tab view instead

src/apprt/gtk/ui/menu-surface-context_menu.ui
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface domain="com.mitchellh.ghostty">
+  <requires lib="gtk" version="4.0"/>
+  <menu id="menu">
+    <section>
+      <item>
+        <attribute name="label" translatable="yes">Copy</attribute>
+        <attribute name="action">win.copy</attribute>
+      </item>
```
pluiedev, line 9: Is this all hand-written? I feel like we might want some kind of mechanism to link this and the various actions in code together

pluiedev:
BTW Blueprint has builtin syntax for menus which I find a lot easier to read and maintain (I just really, really dislike XML lol) https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/reference/menus.html

src/apprt/gtk/ui/menu-surface-context_menu.ui
jcollie: Yes, it's hand written. Yes, I think in a follow-up PR we'll want to convert it to a blueprint.

src/apprt/gtk/Window.zig
jcollie: It's because `gtkActionClose` is associated with the "Close Window" menu item, not "Close a Surface". Nothing about windows being close/freed twice.

src/build/docker/debian/Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN DEBIAN_FRONTEND="noninteractive" apt-get -qq update && \
     libonig-dev \
     lintian \
     lsb-release \
+    libxml2-utils \
```
pluiedev, line 13: Why is this here?

pluiedev (changes requested):

src/build/docker/debian/Dockerfile
jcollie: `xmllint` is needed when compiling builder ui files into resources (for the `xml-stripblanks` operation).

tristan957 (approved):

jcollie:
@mitchellh @pluiedev can I get an updated review? Merging is blocked right now.

pluiedev (approved):

gtjoseph:
@jcollie I think there's an issue with gtk-single-instance.  When true, and more than 1 window is opened, typing "exit" in any window except the last one opened causes the shell to exit leaving a blank window with title and borders.  When you type "exit" in the last window, the earlier exited ones close.  The "main" branch as of this morning doesn't have this issue.

jcollie:
> @jcollie I think there's an issue with gtk-single-instance. When true, and more than 1 window is opened, typing "exit" in any window except the last one opened causes the shell to exit leaving a blank window with title and borders. When you type "exit" in the last window, the earlier exited ones close. The "main" branch as of this morning doesn't have this issue.

Yeah, this was fixed in #5837. The branch for this PR hasn't been rebased to include the fix.

gtjoseph:
> Yeah, this was fixed in #5837. The branch for this PR hasn't been rebased to include the fix.

Ah gotcha.

tristan957 (approved):

mitchellh (approved):

PR #5790, opened 2025-02-15
Replace the alert dialog for tab title editing with popover

### LINKED ISSUE #5768: macOS: "Change Title" feature should use a popover instead of an NSAlert
The feature merged in #4217 should use a popover instead of an alert.

liby:
Resolves #5768

https://github.com/user-attachments/assets/08d78c3e-5fb3-475e-a5b8-c29ff79ba7e9

During testing, it was found when `window-decoration` is set to `none`, the "Change Title..." in both the context menu and menu bar doesn't have any effect since there's no title bar to display or modify. Perhaps in this case we should hide this feature to avoid confusion.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -387,40 +387,45 @@ extension Ghostty {
 
         /// Set the title by prompting the user.
         func promptTitle() {
-            // Create an alert dialog
-            let alert = NSAlert()
-            alert.messageText = "Change Terminal Title"
-            alert.informativeText = "Leave blank to restore the default."
-            alert.alertStyle = .informational
-
-            // Add a text field to the alert
-            let textField = NSTextField(frame: NSRect(x: 0, y: 0, width: 250, height: 24))
-            textField.stringValue = title
-            alert.accessoryView = textField
-
-            // Add buttons
-            alert.addButton(withTitle: "OK")
-            alert.addButton(withTitle: "Cancel")
-
-            let response = alert.runModal()
-
-            // Check if the user clicked "OK"
-            if response == .alertFirstButtonReturn {
-                // Get the input text
-                let newTitle = textField.stringValue
-
-                if newTitle.isEmpty {
-                    // Empty means that user wants the title to be set automatically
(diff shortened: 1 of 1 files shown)


mitchellh:
> During testing, it was found when window-decoration is set to none, the "Change Title..." in both the context menu and menu bar doesn't have any effect since there's no title bar to display or modify. Perhaps in this case we should hide this feature to avoid confusion.

I don't know a good way to conditionally disable this menu item. Do you know an easy way, do you want to take a stab at this? I agree with it.

liby:
> I don't know a good way to conditionally disable this menu item. Do you know an easy way, do you want to take a stab at this? I agree with it.

I've handled the "Change Title..." menu item in two ways:

1. Context menu: Hide the menu item when window-decoration is "none" or in full-screen mode (since the popover can't be positioned correctly without a title bar)
2. Menu bar: Disable the menu item when window-decoration is "none" or in full-screen mode

Let me know if you'd like me to make any adjustments.

<details>

<summary>Screenshots</summary>

image

image

image

</details>

mitchellh:
A couple things I noticed:

* For `macos-titlebar-style = tabs`, I get an exception.

* For native tabs, when the titlebar is being changed, the title turns into a ghost and we probably shouldn't do that:

CleanShot 2025-02-16 at 12 13 26@2x

* The popover X offset is slightly off center.

liby:
> For `macos-titlebar-style = tabs`, I get an exception.

I haven't been able to reproduce the exception with `macos-titlebar-style = tabs`.

> For native tabs, when the titlebar is being changed, the title turns into a ghost and we probably shouldn't do that

This is a tricky issue I've also noticed. After investigating, it appears to be related to window focus - when the popover is shown, the main window loses focus, causing Ghostty to lose its context and [fall back to the default "👻" emoji](https://github.com/ghostty-org/ghostty/blob/main/macos/Sources/Features/Terminal/TerminalView.swift#L62). 

I've tried several approaches but haven't found a clean solution yet, as this seems to be tied to how macOS handles window focus and popover interactions.

mitchellh:
Thanks for doing this. I haven't heard any requests to do this recently and there's been significant change to the titlebar since so I think I'm going to close this and resolve that issue as unplanned.

PR #5791, opened 2025-02-15
gtk: fix building on Debian 12

### LINKED ISSUE #5788: gtk: support libadwaita at least back to version 1.2 for Debian 12
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5787

<sup>Originally posted by **kyleam** February 15, 2025</sup>
With the merge of 25c5ecf5 (gtk: require libadwaita, 2025-02-14) from gh-5749, building on Debian stable (12, bookworm) fails with

```
src/apprt/gtk/Window.zig:173:30: error: root struct of file 'cimport' has no member named 'adw_tab_button_new'
                const btn = c.adw_tab_button_new();
                            ~^~~~~~~~~~~~~~~~~~~
```

and

```
src/apprt/gtk/Window.zig:368:10: error: root struct of file 'cimport' has no member named 'adw_tab_overview_set_show_start_title_buttons'
        c.adw_tab_overview_set_show_start_title_buttons(
```

`adw_tab_button_new` and `adw_tab_overview_set_show_start_title_buttons` were not introduced until libadwaita **1.3.0**.

Debian 12 has libadwaita **1.2.2**.

```
$ apt-cache policy libadwaita-1-dev
libadwaita-1-dev:
  Installed: 1.2.2-1
  Candidate: 1.2.2-1
  Version table:
 *** 1.2.2-1 500
        500 http://deb.debian.org/debian bookworm/main amd64 Packages
        100 /var/lib/dpkg/status
```

It may be that the intent of gh-5749 was to drop support for libadwaita versions before 1.3.0, but, glancing through the PR, I don't see that explicitly state.  @jcollie, can you confirm?

---

Related discussion (from prior to the merge of gh-5749): gh-5072

cc: @jneuhaus
</div>

jcollie:
`std.debug.assert(x)` _is not_ the same as `if (!x) unreachable` because the function call is not `inline`. Since it's not inline the Zig compiler will try to compile any code that might otherwise be unreachable.

Also, added a CI test that compiles Ghostty in a Debian 12 container to ensure that regressions do not happen.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -628,11 +628,7 @@ jobs:
           nix develop -c sh -c "cd pkg/${{ matrix.pkg }} ; zig build test"
 
   test-debian-12:
-    strategy:
-      fail-fast: false
-      matrix:
-        platform: ["linux/amd64", "linux/arm64"]
-    name: Test build on Debian 12 ${{ matrix.platform }}
+    name: Test build on Debian 12
     runs-on: namespace-profile-ghostty-sm
     needs: test
     steps:
@@ -651,4 +647,3 @@ jobs:
           context: .
           file: src/build/docker/debian/Dockerfile
           build-args: DISTRO_VERSION=12
-          platform: ${{ matrix.platform }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -647,4 +647,4 @@ jobs:
         with:
           context: .
           file: src/build/docker/debian/Dockerfile
-          build-args: DISTRO_VERSION=12
+          build-args: DISTRO_VERSION=12,ZIG_VERSION=0.13.0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -626,3 +626,29 @@ jobs:
       - name: Test ${{ matrix.pkg }} Build
         run: |
           nix develop -c sh -c "cd pkg/${{ matrix.pkg }} ; zig build test"
+
+  test-debian-12:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: ["linux/amd64", "linux/arm64"]
+    name: Test build on Debian 12 ${{ matrix.platform }}
+    runs-on: namespace-profile-ghostty-sm
+    needs: test
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install and configure Namespace CLI
+        uses: namespacelabs/nscloud-setup@v0
+
+      - name: Configure Namespace powered Buildx
+        uses: namespacelabs/nscloud-setup-buildx-action@v0
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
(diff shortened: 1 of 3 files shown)


src/apprt/gtk/Window.zig
```diff
@@ -152,7 +152,7 @@ pub fn init(self: *Window, app: *App) !void {
 
     // If we're using an AdwWindow then we can support the tab overview.
     if (self.tab_overview) |tab_overview| {
-        assert(adwaita.versionAtLeast(1, 4, 0));
+        if (!adwaita.versionAtLeast(1, 4, 0)) unreachable;
```
mitchellh, line 155: Ah my mistake! That's right.

.github/workflows/test.yml
```diff
@@ -626,3 +626,29 @@ jobs:
       - name: Test ${{ matrix.pkg }} Build
         run: |
           nix develop -c sh -c "cd pkg/${{ matrix.pkg }} ; zig build test"
+
+  test-debian-12:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: ["linux/amd64", "linux/arm64"]
```
mitchellh, line 634: We probably only need to test amd64, I can't see a scenario where one works and one fails since what we're really protecting against with this test is package versions.

.github/workflows/test.yml
```diff
@@ -626,3 +626,29 @@ jobs:
...
+
+      - name: Configure Namespace powered Buildx
+        uses: namespacelabs/nscloud-setup-buildx-action@v0
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: src/build/docker/debian/Dockerfile
+          build-args: DISTRO_VERSION=12
```
mitchellh, line 653: Let's pass the Zig version in here as well and put that as a top-level workflow env. The number of places we hardcode the Zig version we use is too damn high. 

Longer term it'd probably be better to make a `setup` job that extracts it from our source code (eventually the build.zig.zon but we don't use that yet).

mitchellh (changes requested):
A couple small points. I don't love the idea of maintaining a Dockerfile for this but it is probably the most practical.

.github/workflows/test.yml
```diff
@@ -626,3 +626,29 @@ jobs:
       - name: Test ${{ matrix.pkg }} Build
         run: |
           nix develop -c sh -c "cd pkg/${{ matrix.pkg }} ; zig build test"
+
+  test-debian-12:
```
mitchellh, line 630: Oh also, this key needs to be added to the "checks required" job at the top of this file.

.github/workflows/test.yml
jcollie: Fixed.

.github/workflows/test.yml
jcollie: Fixed.

.github/workflows/test.yml
jcollie: Fixed.

mitchellh (approved):
Assuming CI passes, approved, thank you!

PR #5801, opened 2025-02-16
feat: add support for relative paths

amustaque97:
**Changes Summary**

To make the relative path clickable we'll check whether we're able to
resolve the path in the local system or not. We check the line and it's all
substrings separated by <space>. If a path is resolvable then we will build
the selection and return the result i.e. `ActionSelection` struct. `ActionSelection`
is containsthe same `action` and `selection` properties.

Related to: https://github.com/ghostty-org/ghostty/issues/1972


src/Surface.zig
```diff
@@ -3223,6 +3225,47 @@ fn linkAtPos(
...
+                if (sel) |s| {
+                    const result = .{
+                        .action = input.Link.Action{ .open = {} },
+                        .selection = s,
+                    };
+                    return result;
+                }
+            }
+        }
+    }
```
mitchellh, line 3267: This is complicated enough that it'd require unit tests.

mitchellh (changes requested):
Sorry for the delay. Here's the review. I think it's acceptable but one part will need unit tests. Please reopen the PR when that is done, I'm going to close just because its been awhile (my fault).

PR #5803, opened 2025-02-16
[macOS] feat: Add "quick-terminal" option to "macos-hidden" config

### LINKED ISSUE #5784: macOS: Add "quick-terminal" option to "macos-hidden" config
Builds on #4538

Add a `quick-terminal` enum value to `macos-hidden` that only hides the macOS app if there are no active terminal windows except the quick terminal. Importantly, I think we should keep the macOS app visible if there is any other active non-terminal window such as an update notification. 

This will require some changes to the macOS app, I think we should add a notification from our TerminalManager whenever a terminal window is added/removed that AppDelegate can listen to so that our activation policy is fully managed at the AppDelegate level.

McNight:
Fixes #5784

I started with a notification for listening to Ghostty terminal windows but then I saw the remark about update windows so I implemented `applicationDidUpdate` which is called whenever a window is added/removed.

Tell me if if you want me to remove the notification for Ghostty terminals!

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -403,6 +403,12 @@ class AppDelegate: NSObject,
         reloadDockMenu()
     }
 
+    func applicationDidUpdate(_ notification: Notification) {
+        guard derivedConfig.shouldSwitchBetweenActivationPolicies else { return }
+        // Are we presenting any regular windows ?
+        NSApp.setActivationPolicy(NSApp.visibleRegularWindows.isEmpty ? .accessory : .regular)
+    }
+
     /// Syncs a single menu shortcut for the given action. The action string is the same
     /// action string used for the Ghostty configuration.
     private func syncMenuShortcut(_ config: Ghostty.Config, action: String, menuItem: NSMenuItem?) {
@@ -540,6 +546,9 @@ class AppDelegate: NSObject,
 
         case .always:
             NSApp.setActivationPolicy(.accessory)
+
+        case .quick_terminal:
+            NSApp.setActivationPolicy(NSApp.visibleRegularWindows.isEmpty ? .accessory : .regular)
         }
 
         // If we have configuration errors, we need to show them.
@@ -786,17 +795,20 @@ class AppDelegate: NSObject,
         let initialWindow: Bool
         let shouldQuitAfterLastWindowClosed: Bool
         let quickTerminalPosition: QuickTerminalPosition
+        let shouldSwitchBetweenActivationPolicies: Bool
 
         init() {
             self.initialWindow = true
             self.shouldQuitAfterLastWindowClosed = false
             self.quickTerminalPosition = .top
+            self.shouldSwitchBetweenActivationPolicies = false
         }
 
         init(_ config: Ghostty.Config) {
             self.initialWindow = config.initialWindow
             self.shouldQuitAfterLastWindowClosed = config.shouldQuitAfterLastWindowClosed
             self.quickTerminalPosition = config.quickTerminalPosition
+            self.shouldSwitchBetweenActivationPolicies = config.macosHidden == .quick_terminal
         }
     }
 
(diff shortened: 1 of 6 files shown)


macos/Sources/Features/Terminal/TerminalManager.swift
```diff
@@ -17,7 +17,13 @@ class TerminalManager {
     var focusedSurface: Ghostty.SurfaceView? { mainWindow?.controller.focusedSurface }
 
     /// The set of windows we currently have.
-    var windows: [Window] = []
+    private(set) var windows: [Window] = [] {
+        didSet {
```
mitchellh, line 21: Yeah let's remove all this machinery since we aren't using it. 

I think using Combine in the future is a better approach than this anyways.

macos/Sources/Ghostty/Package.swift
```diff
@@ -247,6 +247,10 @@ extension Notification.Name {
 
     /// Close tab
     static let ghosttyCloseTab = Notification.Name("com.mitchellh.ghostty.closeTab")
+
+    /// Managed windows did change.
+    static let ghosttyWindowsChanged = Notification.Name("com.mitchellh.ghostty.windowsChanged")
```
mitchellh, line 252: Remove this

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -786,17 +795,20 @@ class AppDelegate: NSObject,
         let initialWindow: Bool
         let shouldQuitAfterLastWindowClosed: Bool
         let quickTerminalPosition: QuickTerminalPosition
+        let shouldSwitchBetweenActivationPolicies: Bool
```
mitchellh, line 798: I think it's less confusing generally to keep these derived configs as dumb as possible. I recommend changing this to just copy the `macosHidden` value instead.

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -540,6 +546,9 @@ class AppDelegate: NSObject,
 
         case .always:
             NSApp.setActivationPolicy(.accessory)
+
+        case .quick_terminal:
+            NSApp.setActivationPolicy(NSApp.visibleRegularWindows.isEmpty ? .accessory : .regular)
```
mitchellh, line 551: I think its cleaner to make something like a `syncActivationPolicy` function and call that from here and the `applicationDidUpdate` callback.

mitchellh (changes requested):
I haven't actually run it to test it yet but reviewing the code superficially I have some requests. I do think we should remove the notification machinery for now, thanks for that though. 😄

src/config/Config.zig
```diff
@@ -5762,6 +5762,7 @@ pub const MacTitlebarProxyIcon = enum {
 pub const MacHidden = enum {
     never,
     always,
+    @"quick-terminal",
```
mitchellh, line 5765: Oh we need to add docs for this to the config key in this file.

McNight: I added a comment for this new case, but maybe all the related paragraphs above need to be updated as well?

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -403,6 +403,10 @@ class AppDelegate: NSObject,
         reloadDockMenu()
     }
 
+    func applicationDidUpdate(_ notification: Notification) {
```
mitchellh, line 406: Okay I tested the logic of this and imo this callback gets called _way_ too much. On initial launch it's called 4 times and when I close the window once it gets called something like a dozen or more times (I didn't count, but it was a lot). 

I think we need to do something different. Maybe "any open window" isn't the right logic and we need to do some more research into how iTerm2 behaves instead.

McNight: Yes that's what I noticed as well :/
I saw this method from a blog post about the challenge to track managed windows...
With AppKit Notifications, you get when windows are about to close, and when they become key/main but that's it.

Maybe we could intercept (via swizzling?) when windows are about to be set to the `NSApp.windows` property; or even add KVO to this property (seems that it is not)...

Do you think that `applicationDidUpdate(_:)` may cause performance issues?

mitchellh:
Going to close this, I want to find another way to achieve this. If we can't, then we may resurrect this but I'm not ready to merge this as-is. Thank you for trying though.

PR #5813, opened 2025-02-16
Fix tilde path handling in terminal links

Pym:
This PR improves the handling of tilde (`~`) in clickable paths within the terminal, making it more consistent with standard terminal behavior.

Related to issue #1972 and PR #5801

### Core Functionality
- Enhance path handling in `Surface.zig` to properly expand tilde paths
- Update URL regex pattern in `config/url.zig` to recognize standalone tilde
- Clean up path processing in `renderer/link.zig` for better maintainability

### Test Coverage
Added comprehensive test cases in `url.zig` to verify:
- Standalone tilde (`~`) detection and expansion
- Tilde with path (`~/path`) handling
- Tilde in text context (`cd ~ to go home`)

## Impact
These changes ensure:
1. Clicking on a standalone `~` opens the home directory
2. Paths starting with `~/` correctly expand to the full home directory path
3. Existing path patterns and URL detection remain unaffected

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3275,7 +3275,21 @@ fn processLinks(self: *Surface, pos: apprt.CursorPos) !bool {
                 .trim = false,
             });
             defer self.alloc.free(str);
-            try internal_os.open(self.alloc, .unknown, str);
+
+            // Handle paths starting with ~
+            if (str.len > 0 and str[0] == '~') {
+                const home = std.posix.getenv("HOME") orelse "";
+                const path = if (str.len == 1)
+                    try std.fmt.allocPrint(self.alloc, "{s}", .{home})
+                else if (str[1] == '/')
+                    try std.fmt.allocPrint(self.alloc, "{s}{s}", .{ home, str[1..] })
+                else
+                    try std.fmt.allocPrint(self.alloc, "{s}", .{str});
+                defer self.alloc.free(path);
+                try internal_os.open(self.alloc, .unknown, path);
+            } else {
+                try internal_os.open(self.alloc, .unknown, str);
+            }
         },
 
         ._open_osc8 => {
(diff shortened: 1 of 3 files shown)


src/Surface.zig
```diff
@@ -3275,7 +3275,21 @@ fn processLinks(self: *Surface, pos: apprt.CursorPos) !bool {
                 .trim = false,
             });
             defer self.alloc.free(str);
-            try internal_os.open(self.alloc, .unknown, str);
+
+            // Handle paths starting with ~
+            if (str.len > 0 and str[0] == '~') {
+                const home = std.posix.getenv("HOME") orelse "";
```
mitchellh, line 3281: We should be using `internal_os.expandHome` instead of repeating this logic everywhere.

mitchellh (changes requested):
There's a lot of repeated logic we should stop doing since it is very platform-specific. i.e. the `HOME` env var is not correct to use even on macOS (if its not set) nor Windows when we get there. Additionally, `std.posix` won't be available on Windows so that's no good either.

I'm also curious how this interacts with OSC8 hyperlinks and how other terminals handle this scenario. Can any terminal program sending tilde-paths via OSC8 and have it open? (I can't even recall if you can set file URLs with OSC8 to be honest it's been awhile since I looked at that spec).

Pym:
Thanks @mitchellh for this detailed feedback.

Indeed, I am new to Zig and I went a bit too fast in my implementation, without taking into account the cross-platform aspects and the reuse of existing code.

I will take the time to study the OSC8 specification in more detail. I will come back with an improved version of this PR!

mitchellh:
Appreciate it. Please reopen this PR when you do so or open a new one, I'd rather not a have a stale PR dangling. Thank you!

PR #5834, opened 2025-02-17
macOS: Fix new window focus when created from quick terminal

### LINKED ISSUE #5688: macOS: new window (cmd+n) from quick terminal global keybind doesn't focus Ghostty
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5663

<sup>Originally posted by **leojweda** February 10, 2025</sup>
OS: macOS.
Version: 1.1.0

Steps to reproduce:
* Open the quick terminal.
* Press the shortcut for a new window (`cmd +n`).

Expected results: The new window is focused.

Actual results: Because the quick terminal doesn't focus Ghsotty, the focused app stays as what you had before. Using `alt+tab` to switch to Ghostty doesn't focus the newly created window. You have to switch to another app and back for it to focus. Clicking the new Ghostty window using the mouse doesn't work either.</div>

### Extra Notes

* This requires a global keybind to reproduce

liby:
## Root Cause

The issue has two aspects:

1. The window creation process didn't explicitly force focus on the new window after showing it.
2. More fundamentally, we were relying on `NSApp.isActive` to determine whether to activate the application, which is problematic because:
   - When creating a window from quick terminal, the application is already "active" but this active state is owned by the quick terminal
   - The [`NSApp.isActive` check](https://github.com/ghostty-org/ghostty/blob/4cfe5522db9f073cef9695e9edb1d71f02f2edb1/macos/Sources/Features/Terminal/TerminalManager.swift#L100) doesn't accurately reflect our intent - creating a new window is an explicit user action that should always result in that window gaining focus

## Solution

Removing the `NSApp.isActive` check.

```swift
// Before
if !NSApp.isActive {
    NSApp.activate(ignoringOtherApps: true)
}

// After
NSApp.activate(ignoringOtherApps: true)
```

Fixes #5688

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalManager.swift
+++ b/macos/Sources/Features/Terminal/TerminalManager.swift
@@ -95,11 +95,8 @@ class TerminalManager {
             }
         }
 
-        // If our app isn't active, we make it active. All new_window actions
-        // force our app to be active.
-        if !NSApp.isActive {
-            NSApp.activate(ignoringOtherApps: true)
-        }
+        // All new_window actions force our app to be active.
+        NSApp.activate(ignoringOtherApps: true)
 
         // We're dispatching this async because otherwise the lastCascadePoint doesn't
         // take effect. Our best theory is there is some next-event-loop-tick logic


mitchellh:
This is unfortunately not working for me. Were you sure to try this with a global keybind and another app focused? That's where the more difficult issue is...

liby:
> This is unfortunately not working for me. Were you sure to try this with a global keybind and another app focused? That's where the more difficult issue is...

Indeed, in this case, only adopting the proposed improvement can be work.

```swift
// Before
if !NSApp.isActive {
    NSApp.activate(ignoringOtherApps: true)
}

// After
NSApp.activate(ignoringOtherApps: true)
```
https://github.com/user-attachments/assets/d983e664-c2b5-43b8-8c15-33a4c3b4fd7b

mitchellh (approved):

PR #5846, opened 2025-02-18
gtk: add process scanner

jcollie:
This is _very_ much incomplete, as there are race conditions regarding accessing the list of surfaces, plus there needs to be a lot of discussion regarding the design of various interfaces. But since there was some interest in seeing the code in it's current state I thought that I'd push this up here for better visibility.


jcollie:
Closing this, as having a "global" process scanner is the wrong solution I believe.

PR #5866, opened 2025-02-19
GTK: propagate config updates to the GTK apprt surfaces

jcollie:
This will be needed to the convert GTK apprt surfaces to a DerivedConfig model rather than accessing the global config stored in the app. It's a no-op for now, but I have a PR in the works to update the resize overlay that will take advantage of this.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -361,7 +361,7 @@ pub fn updateConfig(
     config: *const configpkg.Config,
 ) !void {
     // avoid multiple reconfigs when we have many surfaces contained in this
-    // surface using the integer value of config as a simple marker to know if
+    // window using the integer value of config as a simple marker to know if
     // we've "seen" this particular config before
     const this_config = @intFromPtr(config);
     if (self.last_config == this_config) return;
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -31,6 +31,10 @@ const log = std.log.scoped(.gtk);
 
 app: *App,
 
+/// Used to deduplicate updateConfig invocations
+last_config: usize,
+
+/// Local copy of any configuration
 config: DerivedConfig,
 
 /// Our window
@@ -107,6 +111,7 @@ pub fn init(self: *Window, app: *App) !void {
     // Set up our own state
     self.* = .{
         .app = app,
+        .last_config = @intFromPtr(&app.config),
         .config = DerivedConfig.init(&app.config),
         .window = undefined,
         .headerbar = undefined,
@@ -355,6 +360,13 @@ pub fn updateConfig(
     self: *Window,
     config: *const configpkg.Config,
 ) !void {
+    // avoid multiple reconfigs when we have many surfaces contained in this
+    // surface using the integer value of config as a simple marker to know if
+    // we've "seen" this particular config before
+    const this_config = @intFromPtr(config);
+    if (self.last_config == this_config) return;
+    self.last_config = this_config;
+
     self.config = DerivedConfig.init(config);
 
     // We always resync our appearance whenever the config changes.
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -892,6 +892,9 @@ fn configChange(
 ) void {
     switch (target) {
         .surface => |surface| surface: {
+            surface.rt_surface.updateConfig(new_config) catch |err| {
+                log.err("unable to update surface config: {}", .{err});
+            };
             const window = surface.rt_surface.container.window() orelse break :surface;
             window.updateConfig(new_config) catch |err| {
                 log.warn("error updating config for window err={}", .{err});
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -682,6 +682,12 @@ pub fn deinit(self: *Surface) void {
     self.resize_overlay.deinit();
 }
 
+/// Update our local copy of any configuration that we use.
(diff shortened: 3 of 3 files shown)


mitchellh (changes requested):
This isn't the right approach. I'm trying to move away from any decl-based callbacks and move to the apprt action system. And the apprt action system already sends a `config_change` event targeting the surface on line 1216. 

The macOS apprt uses this exactly in the way this PR does: it then maps that back to a SurfaceView and triggers a `configDidChange` notification on it. 

I would expect other appts to do the same. This would be an excellent use case for `@fieldParentPtr` if necessary 😄 (a callback that targets the field that triggers it and using that to find a containing structure). Although, there are other approaches as well.

mitchellh:
> This would be an excellent use case for @fieldParentPtr if necessary

On further thought, it's not necessary at all, because you have access to `rt_surface` directly from that. (macOS doesn't because it goes through libghostty which doesn't have access to Zig structs). 

So the `performAction` in the apprt could easily trigger `target.surface.rt_surface.configDidChange` if that was the desired logic.

mitchellh:
Doing a bit more research, it looks like we currently only send that event to the window:

CleanShot 2025-02-18 at 20 47 20@2x

The preferential approach is that this is probably sent to the surface, and window changes should probably only take effect when app-level config changes are made. (Reload config is an app-level config change).

jcollie:
> Doing a bit more research, it looks like we currently only send that event to the window for GTK:

Yeah, that's probably why I didn't think of doing there earlier. One problem with the current approach is that if a window contains multiple surfaces (both tabs and splits) it will have it's `updateConfig` method called multiple times, once for each surface.

jcollie:
OK, this is now a GTK-only change by taking advantage of the apprt action. I also added some simple code to dedup the number of times that the GTK window will update it's config. Before it would reload the config once for every surfaces contained within the window, now it will just reload once.

src/apprt/gtk/Window.zig
```diff
@@ -355,6 +360,13 @@ pub fn updateConfig(
     self: *Window,
     config: *const configpkg.Config,
 ) !void {
+    // avoid multiple reconfigs when we have many surfaces contained in this
+    // surface using the integer value of config as a simple marker to know if
```
tristan957, line 364: s/surface/window? or is that what you meant

jcollie: DOH!

src/apprt/gtk/Window.zig
```diff
@@ -355,6 +360,13 @@ pub fn updateConfig(
     self: *Window,
     config: *const configpkg.Config,
 ) !void {
+    // avoid multiple reconfigs when we have many surfaces contained in this
+    // window using the integer value of config as a simple marker to know if
+    // we've "seen" this particular config before
+    const this_config = @intFromPtr(config);
+    if (self.last_config == this_config) return;
+    self.last_config = this_config;
```
mitchellh, line 368: This whole thing smells fishy. 

What's the cause of this being called multiple times?

mitchellh (commented):
There's something fishy about the dedup. The fix is fine but it fails the "why" test for me (as in, if we ask why enough times I feel like we don't need it at all). 

If we only update the window config when we reload the app (not the surface target), then would that resolve it? Is there a downside to that?

src/apprt/gtk/Window.zig
jcollie: It's because the core app loops through every surface and sends a config changed message to the surface. That in turn calls the Window update config function. That happens once for each surface. So if a Window has five splits it will get called five times.

I'm on mobile right now but I'm pretty sure that we do not keep a list of windows anywhere (except buried in GTK's object tree) so we can't send the config changed message to the window except through the individual surfaces.

jcollie: https://github.com/ghostty-org/ghostty/blob/38908e01268298ced33b7b6835e49b8c6d437e12/src/App.zig#L146

tristan957: https://docs.gtk.org/gtk4/method.Application.get_windows.html will give you the list of Windows.

mitchellh (approved):
I'm going to approve this to get you going but I think the last config intFromPtr is still the wrong long term solution and there's something more fundamental in the message flow within the GTK apprt that should be done to prevent this.

pluiedev (approved):
Agreed with Mitchell that the whole `@intFromPtr` business is fishy, but apparently Zig doesn't offer any methods of pointer equality? Coming from Rustland with lots of talks about strict provenance (aka "no, pointers aren't just integers") this whole thing *reeks*. This is okay as a bandaid fix but I think in the long run we need to rethink our dispatch model to deduplicate window targets somewhere in App.zig. Increasingly it's starting to feel like it's a bit nonsensical to make actions like toggle window decorations or toggle tab overview be "surface" actions when they are in fact "window" actions.

jcollie:
> but apparently Zig doesn't offer any methods of pointer equality

Even if it did, I specifically chose to store an integer to try and avoid the temptation to de-reference the pointer.

> This is okay as a bandaid fix but I think in the long run we need to rethink our dispatch model to deduplicate window targets somewhere in App.zig.

Yes, the flow of config changes is _very_ confusing. It crosses back and forth between core and apprt several times and then gets pushed to every surface which pushes it to every window and in a separate action. I'm going to open a discussion about the topic.

pluiedev:
I meant pointer equality as in comparing the pointers themselves as opposed to their contents (which would require dereferencing), which is equivalent to comparing their addresses and provenance, on systems that have hardware provenance checking.

PR #5885, opened 2025-02-19
Change default log level to warning.

gtjoseph:
Ghostty's default log level is now "warning" unless one of the following
conditions is met...
* A "+action" is being run in which case it's "error".
* There's no app_runtime (lib mode) in which case it's "error".
* The GHOSTTY_LOG environment variable isn't empty in which case it's "info".
* Ghostty was started from a desktop/launcher icon in which case it's "info".
* A debug build is being run in which case it's "debug".

The conditons are evaluated from top to bottom and the last one met wins.

### DIFF
--- a/src/global.zig
+++ b/src/global.zig
@@ -26,18 +26,12 @@ pub const GlobalState = struct {
     gpa: ?GPA,
     alloc: std.mem.Allocator,
     action: ?cli.Action,
-    logging: Logging,
+    log_level: std.log.Level,
     rlimits: ResourceLimits = .{},
 
     /// The app resources directory, equivalent to zig-out/share when we build
     /// from source. This is null if we can't detect it.
     resources_dir: ?[]const u8,
-
-    /// Where logging should go
-    pub const Logging = union(enum) {
-        disabled: void,
-        stderr: void,
-    };
 
     /// Initialize the global state.
     pub fn init(self: *GlobalState) !void {
@@ -56,7 +50,7 @@ pub const GlobalState = struct {
             .gpa = null,
             .alloc = undefined,
             .action = null,
-            .logging = .{ .stderr = {} },
+            .log_level = .warn,
             .rlimits = .{},
             .resources_dir = null,
         };
@@ -92,11 +86,11 @@ pub const GlobalState = struct {
         // If we have an action executing, we disable logging by default
         // since we write to stderr we don't want logs messing up our
         // output.
-        if (self.action != null) self.logging = .{ .disabled = {} };
+        if (self.action != null) self.log_level = .err;
 
         // For lib mode we always disable stderr logging by default.
         if (comptime build_config.app_runtime == .none) {
-            self.logging = .{ .disabled = {} };
+            self.log_level = .err;
         }
 
         // I don't love the env var name but I don't have it in my heart
@@ -107,8 +101,16 @@ pub const GlobalState = struct {
         if ((try internal_os.getenv(self.alloc, "GHOSTTY_LOG"))) |v| {
             defer v.deinit(self.alloc);
             if (v.value.len > 0) {
-                self.logging = .{ .stderr = {} };
+                self.log_level = .info;
             }
+        }
+
+        if (internal_os.launchedFromDesktop()) {
+            self.log_level = .info;
+        }
+
+        if (builtin.mode == .Debug) {
+            self.log_level = .debug;
         }
 
         // Setup our signal handlers before logging
--- a/src/main_ghostty.zig
+++ b/src/main_ghostty.zig
@@ -119,6 +119,12 @@ fn logFn(
     comptime format: []const u8,
     args: anytype,
 ) void {
+
+    // If we're not logging messages at this level, we can just stop now.
+    if (@intFromEnum(level) > @intFromEnum(state.log_level)) {
+        return;
+    }
+
     // Stuff we can do before the lock
     const level_txt = comptime level.asText();
     const prefix = if (scope == .default) ": " else "(" ++ @tagName(scope) ++ "): ";
@@ -147,15 +153,9 @@ fn logFn(
         logger.log(std.heap.c_allocator, mac_level, format, args);
     }
 
-    switch (state.logging) {
-        .disabled => {},
-
-        .stderr => {
-            // Always try default to send to stderr
-            const stderr = std.io.getStdErr().writer();
-            nosuspend stderr.print(level_txt ++ prefix ++ format ++ "\n", args) catch return;
-        },
-    }
+    // Always try default to send to stderr
+    const stderr = std.io.getStdErr().writer();
+    nosuspend stderr.print(level_txt ++ prefix ++ format ++ "\n", args) catch return;
 }
 
 pub const std_options: std.Options = .{


00-kat:
Cross-referencing the original discussion: #5723.

src/cli/action.zig
```diff
@@ -68,17 +76,33 @@ pub const Action = enum {
...
 
     /// Detect the action from CLI args.
-    pub fn detectCLI(alloc: Allocator) !?Action {
+    /// Pass in a pointer to a LoggingOptions structure if you want to retrieve
+    /// any `-q/--quiet` or `-v/--verbose` options or `null` if you don't need them.
+    pub fn detectCLI(alloc: Allocator, loggingOptions: ?*LoggingOptions) !?Action {
         var iter = try std.process.argsWithAllocator(alloc);
         defer iter.deinit();
-        return try detectIter(&iter);
+        return try detectIter(&iter, loggingOptions);
```
pluiedev, line 84: Out-pointers are generally somewhat frowned upon in the codebase, but I'm not exactly sure what's a better alternative

src/global.zig
```diff
@@ -26,19 +27,14 @@ pub const GlobalState = struct {
     gpa: ?GPA,
     alloc: std.mem.Allocator,
     action: ?cli.Action,
-    logging: Logging,
     rlimits: ResourceLimits = .{},
+    log_level: std.log.Level,
+    log_enabled: bool,
```
pluiedev, line 32: I don't know if we want to use a boolean here. I think errors should be logged even in quiet mode, so that users may know when something truly *bad* happens. We don't use them nearly as much as warnings in the codebase, and the places where they're found are usually situations that have gone quite wrong and should never happen.

src/main_ghostty.zig
```diff
@@ -147,15 +151,9 @@ fn logFn(
...
-
-        .stderr => {
-            // Always try default to send to stderr
-            const stderr = std.io.getStdErr().writer();
-            nosuspend stderr.print(level_txt ++ prefix ++ format ++ "\n", args) catch return;
-        },
-    }
+    // Always try default to send to stderr
+    const stderr = std.io.getStdErr().writer();
+    nosuspend stderr.print(level_txt ++ prefix ++ format ++ "\n", args) catch return;
```
pluiedev, line 156: Zig doesn't even have async anymore. The `nosuspend` keyword doesn't even do anything to my best knowledge.

src/main_ghostty.zig
```diff
@@ -123,6 +123,10 @@ fn logFn(
     const level_txt = comptime level.asText();
     const prefix = if (scope == .default) ": " else "(" ++ @tagName(scope) ++ "): ";
 
+    if (!state.log_enabled or @intFromEnum(level) > @intFromEnum(state.log_level)) {
```
pluiedev, line 126: IMO comparing log levels via their integer representation is bad style. Write an explicit switch case.

src/cli/action.zig
```diff
@@ -68,17 +76,33 @@ pub const Action = enum {
...
     }
 
     /// Detect the action from any iterator, used primarily for tests.
-    pub fn detectIter(iter: anytype) Error!?Action {
+    /// Pass in a pointer to a LoggingOptions structure if you want to retrieve
+    /// any `-q/--quiet` or `-v/--verbose` options or `null` if you don't need them.
+    pub fn detectIter(iter: anytype, loggingOptions: ?*LoggingOptions) Error!?Action {
```
pluiedev, line 90: Is there any reason `loggingOptions` would ever be null outside of test cases?

pluiedev (changes requested):

pluiedev:
Also I disagree that the default should be error. We should preserve current behavior and allow users who are bothered by it to adjust Ghostty in accordance to their wishes.

src/cli/action.zig
gtjoseph: Searching revealed out-pointers were the most recommended alternative.

There were a couple of other options I tested...
* Change Action from an enum to a struct so when it's returned to global it can contain more than just the action.  That's a fairly invasive change however because other modules like bash_completion.zig would need to be changed.
* Store the results of the parsing in a top-level cli.action structure and add a getter for it.  Kinda clunky.
* Don't use the existing iterator in Action and add one in main_ghostty just to get the logging options.  Worked fine but I thought using the existing iterator would save a bit.

src/global.zig
gtjoseph: Fair enough.  I can remove the `-q/--quiet` option.

src/main_ghostty.zig
gtjoseph: That was existing code.  I can remove it.

src/main_ghostty.zig
gtjoseph: That was the most recommended solution for changing log levels at runtime and is how log.zig itself does it.  I suppose I could change it to a simple equality test.  Only log if the message is an error or if verbose is true.

src/cli/action.zig
gtjoseph: Nope.  I just didn't want to require the tests to pass in a structure when they didn't need to.  I suppose I could create a static one just for testing and have each test use that.

gtjoseph:
> Also I disagree that the default should be error. We should preserve current behavior and allow users who are bothered by it to adjust Ghostty in accordance to their wishes.

I suppose I could leave the default level at info and just keep `--quiet` which would suppress everything except errors.  I just don't get it though.  Why would your average non-ghostty-developer want those messages?  What info do they convey?  Shouldn't the program be as friendly as possible to the larger audience especially when it's fairly trivial to do so and it doesn't compromise any other features?

00-kat:
The main issue with having opt-in as opposed to opt-out logging is that people will paste the logs without `warning`s and `info`s into any GitHub discussions they create.  Though there are many who think that the default is too noisy, though, especially when compared to every other terminal emulator out there.

src/cli/action.zig
gtjoseph: I added a quick arguments iterator in global just before the call to Action.detectCLI().  This removes the need to pass the pointer.

src/main_ghostty.zig
gtjoseph: The Zig 0.13.0 documentation still has nosuspend as does the example logFn in the std.log source so I'm going to leave this as is to be safe.

src/cli/action.zig
gtjoseph: I've removed the pointer argument.

src/global.zig
gtjoseph: `-q/--quiet` has been removed.

src/main_ghostty.zig
gtjoseph: Since I've removed the `-q/--quiet` option, there's no longer a need to compare the levels.  It's just...

```
if (!state.verbose_logging and level != .err) {
    return;
}
```

gtjoseph:
> Also I disagree that the default should be error. We should preserve current behavior and allow users who are bothered by it to adjust Ghostty in accordance to their wishes.

I'm going to leave it at error.  The point of the PR was to make the default logging behavior like that of most other terminal emulators. I.E. quiet.  If the direction is to leave the default at info then the PR can just be closed.

mitchellh:
I'm not fully sure about this change yet (PR without an issue so its at your own risk anyways).

But, I will say the default should be "warn." The "error" log level in Zig is meant to indicate a fatal, non-recoverable error where the program either exits immediately or is likely in a severely broken state. A warning is meant to indicate an unexpected, possibly problematic (but maybe not) issue. 

If we have spurious "warn" logs we should quiet those rather than raise the default to error.

mitchellh:
Another thought that I want to make sure I get down, I've thought it a few times but keep forgetting to put it down so here seems to be best: I think its generally helpful to have verbose logs when starting from a desktop location, because if something goes wrong, its easier to have ready access to some log course (i.e. Console.app on macOS, syslog-style locations on Linux). 

So we may want to consider only defaulting to quiet if we're not launched from the desktop (we have an API for this already `internal_os.launchedFromDesktop()`). And we may want to just keep the existing logic `GHOSTTY_LOG` env var otherwise. That would simplify this PR and minimize any changes.

gtjoseph:
> I'm not fully sure about this change yet (PR without an issue so its at your own risk anyways).

Heh, that's on @pluiedev.  She recommended I open the PR. :) 

> But, I will say the default should be "warn."

Makes sense.

> If we have spurious "warn" logs we should quiet those rather than raise the default to error.

Agreed.

> I think its generally helpful to have verbose logs when starting from a desktop location, because if something goes wrong, its easier to have ready access to some log course (i.e. Console.app on macOS, syslog-style locations on Linux).

AFAIK stdout and stderr go nowhere in Linux when started from a .desktop file.  Well, unless it's a Gnome thing.  Was that the reason for the macos-specific logging in logFn()?
 
> So we may want to consider only defaulting to quiet if we're not launched from the desktop (we have an API for this already `internal_os.launchedFromDesktop()`). And we may want to just keep the existing logic `GHOSTTY_LOG` env var otherwise. That would simplify this PR and minimize any changes.

Can do.

So to recap:

Default is warn unless started from the desktop in which case it's info.
If you start from a command line and want info, you can use `-v/--verbose` or GHOSTTY_LOG.
GHOSTTY_LOG would always apply now, instead of only when an action was being run or do you want to leave it for actions only?

00-kat:
> AFAIK stdout and stderr go nowhere in Linux when started from a .desktop file. Well, unless it's a Gnome thing.

I've never checked my own system log but some launchers seem to redirect it to the system log as this user found: https://github.com/ghostty-org/ghostty/discussions/3614#discussioncomment-11683027.

mitchellh:
> AFAIK stdout and stderr go nowhere in Linux when started from a .desktop file. Well, unless it's a Gnome thing. Was that the reason for the macos-specific logging in logFn()?

Yep! macOS has a builtin GUI app (Console.app) and CLI (`log`) to view macOS system logs and this way desktop app logs go to a place users can view, export, etc.

> So to recap:
> 
> Default is warn unless started from the desktop in which case it's info.
> If you start from a command line and want info, you can use -v/--verbose or GHOSTTY_LOG.
> GHOSTTY_LOG would always apply now, instead of only when an action was being run or do you want to leave it for actions only?

Yeah I think this all makes sense, or at least I would take a look at that. I'm also suggesting we remove `-v` to avoid the arg iteration on startup. At least for now. The changes I suggested I'd merge now, I'm willing to discuss and consider the CLI flags later but I'm not sure.

jcollie:
> AFAIK stdout and stderr go nowhere in Linux when started from a .desktop file.

Depending on the system, that all probably goes to the systemd user journal. Well, assuming you are using systemd. Use
`journalctl --user -a -f` to see if it shows up. The logs would be easier to find if we launched Ghostty from a systemd user service but that's a different discussion.

gtjoseph:
Yeah, I already checked journalctl --user.  Nothing.
Fedora 41 XFCE.

gtjoseph:
PR updated.  See the PR description for new status.

@mitchellh Can I submit a new PR now to squash the warnings (or convert to info)?  It doesn't depend on this PR.

jcollie:
> Can I submit a new PR now to squash the warnings (or convert to info)? It doesn't depend on this PR.

What about instead of changing the logging level of some of these, we condition the log on the release level (i.e. they would be suppressed in anything but Debug level)?

gtjoseph:
> What about instead of changing the logging level of some of these, we condition the log on the release level (i.e. they would be suppressed in anything but Debug level)?

I'm not following.  Wouldn't that be the same as changing just their message levels to debug?

jcollie:
> > What about instead of changing the logging level of some of these, we condition the log on the release level (i.e. they would be suppressed in anything but Debug level)?
> 
> I'm not following. Wouldn't that be the same as changing just their message levels to debug?

The main difference is that warnings are emitted during a `zig build test` and debugs are not.

gtjoseph:
That didn't really clarify anything. :) It's late though and my brain is fried so let me read that again in the morning.

gtjoseph:
@jcollie Sleep didn't help.  Can you give me a code snippet to demonstrate the result you're proposing?

PR #5905, opened 2025-02-20
Gtk: change title prompt

### LINKED ISSUE #5769: GTK: Add "Change Title" feature
Implement #4217, #2509 for GTK.

This can be implemented in whatever GUI method is most idiomatic for GTK. 

The apprt action to listen for is `prompt_title` targeting a surface. 

Similar to macOS, we should ensure that there is a mechanism to reset back to the pty-provided title. On macOS, we do this by leaving the prompt blank and pressing okay. Some accounting has to be done to keep track of the pty-provided title (via the `set_title` apprt action).

b6k-dev:
This PR implements the title change functionality for the GTK app and closes https://github.com/ghostty-org/ghostty/issues/5769

Demo:

https://github.com/user-attachments/assets/cad611b3-b7bf-40ac-8d0f-11d2095fe525

For now I came up with a basic UI that i believe is sufficient, but I'm open to feedback and happy to iterate on it further. 
https://github.com/ghostty-org/ghostty/issues/5769#issuecomment-2660341107 - Regarding Mitchell's comment I checked Gnome Console and I could not find a similar feature.


src/apprt/gtk/Surface.zig
```diff
@@ -346,6 +346,11 @@ cursor: ?*c.GdkCursor = null,
 /// pass it to GTK.
 title_text: ?[:0]const u8 = null,
 
+/// The title of the surface as reported by the terminal.
```
jcollie, line 349: See my comment here: https://github.com/ghostty-org/ghostty/issues/5769#issuecomment-2672676710

But I think we should not be changing the _surface_ title. This is about changing the _tab_ title. If we change the surface title the tab title will change every time that you switch focus to a different split if you have multiple splits in a tab.

jcollie (changes requested):

src/apprt/gtk/Surface.zig
jcollie: After discussing this with Mitchell on Discord that's how he wants the feature to work right now. The current implementation seems a little awkward to me but maybe I'd have to take another look. So ignore this comment for now.

src/apprt/gtk/Surface.zig
```diff
@@ -978,6 +978,31 @@ pub fn getTitle(self: *Surface) ?[:0]const u8 {
...
+    self: *Surface,
+};
+
+pub fn promptTitle(self: *Surface) !void {
+    const window = self.container.window() orelse return;
+
+    const context = try self.app.core_app.alloc.create(PromptTitleDialogContext);
+    context.self = self;
+
+    const dialog = c.gtk_message_dialog_new(window.window, c.GTK_DIALOG_MODAL, c.GTK_MESSAGE_QUESTION, c.GTK_BUTTONS_OK_CANCEL, "Set Tab Title");
```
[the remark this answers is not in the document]
maciekbartczak: Moved the dialog structure to a blueprint file.

src/apprt/gtk/Surface.zig
```diff
@@ -969,15 +989,45 @@ fn updateTitleTimerExpired(ctx: ?*anyopaque) callconv(.C) c.gboolean {
...
+
+pub fn promptTitle(self: *Surface) !void {
+    const window = self.container.window() orelse return;
+
+    var builder = Builder.init("prompt-title-dialog", .blp);
+    defer builder.deinit();
+
+    const dialog: *adw.AlertDialog = @ptrCast(builder.getObject("prompt_title_dialog"));
+    dialog.addResponse("cancel", "Cancel");
+    dialog.addResponse("ok", "OK");
```
maciekbartczak, line 1026: This could possibly be done from the `.blp` since there is a responses property, however I could not get this to work. Also there are no examples of how do define the responses from the blueprint, so I'm just leaving the definitions in the code right now.

src/apprt/gtk/Surface.zig
```diff
@@ -1013,27 +1018,14 @@ const PromptTitleDialogContext = struct {
...
-    const buffer = c.gtk_entry_get_buffer(@ptrCast(entry));
-    c.gtk_entry_buffer_set_text(buffer, self.getTitle() orelse "", -1);
-    c.gtk_box_append(@ptrCast(content_area), entry);
-    c.gtk_entry_set_activates_default(@ptrCast(entry), 1);
-    c.gtk_widget_show(entry);
-
-    _ = c.g_signal_connect_data(dialog, "response", c.G_CALLBACK(&gtkPromptTitleResponse), context, null, c.G_CONNECT_DEFAULT);
-
-    c.gtk_widget_show(dialog);
+    dialog.choose(@ptrCast(window.window), null, @ptrCast(&gtkPromptTitleResponse), self);
```
jcollie, line 1025: Just a quick note since I don't have time for a full review right now, but the `@ptrCast` on the callback shouldn't be necessary using zig-gobject.

maciekbartczak: Fixed.

maciekbartczak:
Two things that I modified from the original dialog version:
* the input box has focus by default
* the ok response has `suggested` apperance

image

src/apprt/gtk/Surface.zig
pluiedev: https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/reference/extensions.html#adw-alertdialog-responses

src/apprt/gtk/Surface.zig
```diff
@@ -2298,3 +2348,38 @@ fn g_value_holds(value_: ?*c.GValue, g_type: c.GType) bool {
     }
     return false;
 }
+
+fn gtkPromptTitleResponse(source_object: ?*gobject.Object, result: *gio.AsyncResult, ud: ?*anyopaque) callconv(.C) void {
+    const dialog: *adw.AlertDialog = @ptrCast(source_object.?);
+    const self = userdataSelf(ud.?);
+
+    const response = dialog.chooseFinish(result);
+    if (glib.strEqual("ok", response) != 0) {
```
pluiedev, line 2357: Why not std.mem.orderZ(.., ..) == .eq?

src/apprt/gtk/Surface.zig
```diff
@@ -2298,3 +2348,38 @@ fn g_value_holds(value_: ?*c.GValue, g_type: c.GType) bool {
...
+        const title_entry: *gtk.Entry = @ptrCast(dialog.getExtraChild());
```
pluiedev, line 2358: Use gobject.ext.cast

src/apprt/gtk/Surface.zig
```diff
@@ -2298,3 +2348,38 @@ fn g_value_holds(value_: ?*c.GValue, g_type: c.GType) bool {
...
+        const title = std.mem.span(title_entry.getBuffer().getText());
+
+        // if the new title is empty and the user has set the title previously, restore the terminal provided title
+        if (title.len == 0 and self.title_from_terminal != null) {
+            if (self.getTerminalTitle()) |terminal_title| {
+                self.setTitle(terminal_title, .user) catch {};
+                self.app.core_app.alloc.free(self.title_from_terminal.?);
```
pluiedev, line 2365: The .? here feels like a bit of a smell... I think what I would do is to wrap this inside an if-capture

src/apprt/gtk/Surface.zig
```diff
@@ -2298,3 +2348,38 @@ fn g_value_holds(value_: ?*c.GValue, g_type: c.GType) bool {
...
+
+    const response = dialog.chooseFinish(result);
+    if (glib.strEqual("ok", response) != 0) {
+        const title_entry: *gtk.Entry = @ptrCast(dialog.getExtraChild());
+        const title = std.mem.span(title_entry.getBuffer().getText());
+
+        // if the new title is empty and the user has set the title previously, restore the terminal provided title
+        if (title.len == 0 and self.title_from_terminal != null) {
+            if (self.getTerminalTitle()) |terminal_title| {
+                self.setTitle(terminal_title, .user) catch {};
```
pluiedev, line 2364: Log error

src/apprt/gtk/ui/prompt-title-dialog.blp
```diff
@@ -0,0 +1,10 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.AlertDialog prompt_title_dialog {
+  heading: "Change Terminal Title";
+  body: "Leave blank to restore the default title.";
```
pluiedev, line 6: Prefer [translatable strings](https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/reference/values.html#translated-strings) for user-facing text

pluiedev (changes requested):

src/apprt/gtk/Surface.zig
maciekbartczak: Fixed.

src/apprt/gtk/Surface.zig
maciekbartczak: Added a log statement.

src/apprt/gtk/Surface.zig
maciekbartczak: I guess since getTerminalTitle already checks for the null this top level != null is redundant, so the simplest solution for me is to just remove it.

pluiedev: Yea sure

src/apprt/gtk/Surface.zig
```diff
@@ -969,15 +989,40 @@ fn updateTitleTimerExpired(ctx: ?*anyopaque) callconv(.C) c.gboolean {
...
+        title[zoom_title_prefix.len..];
+}
+
+pub fn promptTitle(self: *Surface) !void {
+    const window = self.container.window() orelse return;
+
+    var builder = Builder.init("prompt-title-dialog", .blp);
+    defer builder.deinit();
+
+    const entry: *gtk.Entry = @ptrCast(builder.getObject("title_entry"));
```
pluiedev, line 1019: Never ptrCast between GObject objects - always do gobject.ext.cast

src/apprt/gtk/Surface.zig
```diff
@@ -2298,3 +2343,39 @@ fn g_value_holds(value_: ?*c.GValue, g_type: c.GType) bool {
...
     return false;
 }
+
+fn gtkPromptTitleResponse(source_object: ?*gobject.Object, result: *gio.AsyncResult, ud: ?*anyopaque) callconv(.C) void {
+    const dialog: *adw.AlertDialog = @ptrCast(source_object.?);
+    const self = userdataSelf(ud.?);
+
+    const response = dialog.chooseFinish(result);
+    if (std.mem.orderZ(u8, "ok", response) == .eq) {
+        const title_entry: *gtk.Entry = gobject.ext.cast(gtk.Entry, dialog.getExtraChild().?).?;
```
pluiedev, line 2353: The explicit type annotation isn't necessary

src/apprt/gtk/Surface.zig
```diff
@@ -2298,3 +2343,39 @@ fn g_value_holds(value_: ?*c.GValue, g_type: c.GType) bool {
...
+        const title = std.mem.span(title_entry.getBuffer().getText());
+
+        // if the new title is empty and the user has set the title previously, restore the terminal provided title
+        if (title.len == 0) {
+            if (self.getTerminalTitle()) |terminal_title| {
+                self.setTitle(terminal_title, .user) catch |err| {
+                    log.err("Failed to set title: {}", .{err});
```
pluiedev, line 2360: This isn't really *formalized* anywhere but our log messages usually look like
```suggestion
                    log.err("failed to set title={}", .{err});
```

pluiedev (changes requested):

jcollie:
@mitchellh @kenvandine since I had to touch the snap build I'd appreciate a quick once-over.

jcollie:
@maciekbartczak Thanks for this work! I added a few changes to get this compiling on older distros.

src/apprt/gtk/Surface.zig
maciekbartczak: Fixed all log messages.

src/apprt/gtk/Surface.zig
maciekbartczak: fixed in https://github.com/ghostty-org/ghostty/pull/5905/commits/34c347dbbb4d77f89f2b89cdea8e59d9bd72cdc4

src/apprt/gtk/Surface.zig
maciekbartczak: fixed in https://github.com/ghostty-org/ghostty/pull/5905/commits/34c347dbbb4d77f89f2b89cdea8e59d9bd72cdc4

jcollie (approved):
OK, I think that this is ready to go!

pluiedev (approved):
The diff is a bit noisy due to all the blueprint fixes (frankly I think it would be better done in a separate PR) but the die's already been cast. Serve it forth

pluiedev:
Oh, merge conflicts :p

PR #5916, opened 2025-02-21
Update libxev to use dynamic backend, support Linux configurability

mitchellh:
Related to #3224

Previously, Ghostty used a static API for async event handling: io_uring on Linux, kqueue on macOS. This commit changes the backend to be dynamic on Linux so that epoll will be used if io_uring isn't available, or if the user explicitly chooses it.

This introduces a new config `async-backend` (default "auto") which can be set by the user to change the async backend in use. This is a best-effort setting: if the user requests io_uring but it isn't available, Ghostty will fall back to something that is and that choice is up to us.

Basic benchmarking both in libxev and Ghostty (vtebench) show no noticeable performance differences introducing the dynamic API, nor choosing epoll over io_uring.

For platforms that don't have multiple choices (currently macOS), the dynamic API is literally the static API so there should be no performance or behavioral changes whatsoever.


src/config/Config.zig
```diff
@@ -2235,6 +2235,34 @@ term: []const u8 = "xterm-ghostty",
...
+///
+/// Available options:
+///
+///   * `auto` - Automatically choose the best backend for the platform
+///     based on available options.
+///   * `epoll` - Use the `epoll` API
+///   * `io_uring` - Use the `io_uring` API
+///
+/// If the selected backend is not available on the platform, Ghostty will
+/// fall back to an automatically chosen backend that is available.
```
00-kat, line 2258: Should a warning be logged?
Currently the only way to know if the requested backend was unavailable is to check the line
> ```
> info: libxev auto backend=XXX
> ```
and see if it shows the requested one.

00-kat (approved):

src/config/Config.zig
mitchellh: This is also confusingly wrong, this only shows the _auto_ backend. The manual one shows up a little later. A warning would probably be good. We unfortunately have to have some lag between this log and the real one because we have to load config and do a bunch of other setup before we have that available.

jcollie (approved):
Did a test build on one system and no blowups so far.

src/config/Config.zig
00-kat: > The manual one shows up a little later.

I assume that by “manual one” you mean `kqueue` on macOS?  Because `ghostty --async-backend=epoll` still shows the same `libxev auto backend` log.

mitchellh: No, look later there is a `libxev manual backend=XXX` log message.

00-kat: That one is the only log message which says `xev`; I'll try the new change you just force pushed though.

mitchellh: That's because on success its a debug log level, I should raise it.

mitchellh: Done

00-kat: Ah okay, that would explain it since I was running a ReleaseFast build.

00-kat (approved):

src/apprt/gtk/App.zig
```diff
@@ -137,6 +138,27 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
...
+        };
+
+        if (result) {
+            log.info(
+                "libxev manual backend={s}",
+                .{@tagName(xev.backend)},
+            );
+        } else {
+            log.warn(
+                "libxev manual backend failed, using default={s}",
```
00-kat, line 156: This doesn't seem to actually work, on my machine with io_uring disabled these are the only logs that are printed when I request io_uring:
```console
$ zig-out/bin/ghostty --async-backend=io_uring 2>&1 | grep xev
info: ghostty version=1.1.3-dynamic-libxev+aee46266
info: libxev default backend=epoll
info(gtk): libxev manual backend=epoll
```

mitchellh: Oh yeah that's a libxev bug. Fixing it now.

mitchellh: Alright should be good.

00-kat: Yep that works, thanks!

00-kat (approved):

PR #5920, opened 2025-02-22
Docs change: Note instead of simple text and Removing two times showing info.

ronit18:
In docs option reference, for macos-icon info. It is showing two times similar thing. And Note looks good instead of just text because it wont work without `custom-style`. So a Big Note is a good highlight.


mitchellh (approved):
Thanks!

PR #5926, opened 2025-02-22
Add support for whitespace in color and palette parsing

### LINKED ISSUE #5921: config: `Palette` and `Color` types should support spaces
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5798

<sup>Originally posted by **cdwilson** February 16, 2025</sup>
When I open my Ghostty theme file in VS Code, I don't see any color previews for the `palette` colors because there is no space between the ANSI # and the color. However, the color preview does show up for `background`, `foreground`, etc because there is a space.

image

For comparison, here's how the color previews show up in VS Code's settings file:

image

If the syntax for the theme files is revisited at some point in the future, I just wanted to mention this in case there is a way to make it compatible with VS Code's color preview.</div>

## Notes

* `Palette` should support spaces around the `=` sign.
* `Color` should be modified to support whitespace beginning/end
* Unit tests should cover all of this

liby:
Resolves #5921

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4204,14 +4204,16 @@ pub const Color = struct {
 
     pub fn parseCLI(input_: ?[]const u8) !Color {
         const input = input_ orelse return error.ValueRequired;
+        // Trim any whitespace before processing
+        const trimmed = std.mem.trim(u8, input, " \t");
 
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Thanks!

PR #5933, opened 2025-02-22
macOS: fix crosshair verticalText and contextMenu pointerStyles

monsterkodi:
fixes three mouse pointer styles that weren't working on macOS.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -360,14 +360,14 @@ extension Ghostty {
                 pointerStyle = .resizeLeftRight
 
             case GHOSTTY_MOUSE_SHAPE_VERTICAL_TEXT:
-                pointerStyle = .default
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Hm, modifying this in the back port isn't exactly right because back port is only supposed to match the real version directly. We may have to introduce a non-backport structure/view helper to add our pointer style.

Sorry to make this complicated, but the way it's implemented now if we ever dropped support for macOS versions that don't support `pointerStyle` then we'd get breakage.

monsterkodi:
unfortunately, i don't understand what you are saying :)
are you talking about the `contextMenu` style being implemented differently?
if yes, we could just remove the fix for the `contextMenu`style and still get the other two styles working?

monsterkodi:
i honestly tried to understand the problem, but i couldn't figure it out :(

since i don't want to spoil your PR list (and my mind with this unfinished business ;) i will close it. i am sure someone more capable than me will be able to solve this properly at some point or you will find the time to do it yourself.

Issue #4404, opened 2025-01-02
Introduce option to right click to copy


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4381

<div type='discussions-op-text'>

<sup>Originally posted by **7ijme** January  2, 2025</sup>
When I right-click, I don't want a context menu to open. I want it to copy the current selection.

This is what it looks like now:
![image](https://github.com/user-attachments/assets/cd18a301-6400-4576-a99a-ef4d69c411cd)
</div>

### Solution

We should introduce a new configuration option `copy-on-right-click` (similar to `copy-on-selection`), boolean default to "false" that can be set to true by those who wish for this behavior.

When true, context menus are disabled and text is copied. When mouse reporting is enabled, this should also copy so long as there is an active selection.

We should make this work on both macOS and Linux. This is more prevalent in the Linux world but I see no reason to disallow it on macOS since it's opt-in.

peterhowe-adwanted:
I use the middle mouse button (via Logitech Options+) for Mac's Mission Control, so it would be great to be able to have paste on right click.

jzetterman:
I don't mean to hijack, but would it make sense to do right click to paste while you're doing this work as well? See [3783](https://github.com/ghostty-org/ghostty/discussions/3783#discussioncomment-11873022).

han5:
In other terminals, you right click on a URL and it gives the option to Open Link. 
At least quick select / copy will work as well.

chenzhiwei:
There is already a `copy-on-selection`, I think a `paste-on-right-click` is more useful.

In Windows Terminal, the mouse behavior have more optoins:

> Left-click and drag your mouse to create a selection. Double-click expands the selection by word, whereas triple-click expands by line.
> 
> If you are holding the Alt key, you will create a block selection (as opposed to a line selection). Block selections create a rectangular region that do not wrap to the end of the line.
>
> If you are holding the Shift key, you can explicitly expand the selection to a specific point on the terminal without the need to click and drag.
> 
> Once you have a selection present, you have a few options. A single left-click will clear your selection. If you actually want to use it, you can right-click to copy the selected text to your clipboard and clear the selection. If you right-click again, the contents of your clipboard will then be pasted into the terminal.

https://learn.microsoft.com/en-us/windows/terminal/selection#mouse-support

IamTheLime:
[...]

PR #5935, opened 2025-02-22
Right click to copy support

IamTheLime:
Adds right click to copy support to ghostty.

Only gtk context menu is currently being disabled.

Issue [4404](https://github.com/ghostty-org/ghostty/issues/4404)

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1465,7 +1465,7 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
     // If copy on select is false then exit early.
     if (self.config.copy_on_select == .false) return;
 
-    // clear the clipboard. If the selection is set, we only set the clipboard
+    // Clear the clipboard. If the selection is set, we only set the clipboard
     // again if it changed, since setting the clipboard can be an expensive
     // operation.
     const sel = sel_ orelse return;
@@ -3109,8 +3109,10 @@ pub fn mouseButtonCallback(
         // If we already have a selection and the selection contains
         // where we clicked then we don't want to modify the selection.
         if (self.io.terminal.screen.selection) |prev_sel| {
-            try self.copyOnMouseAction(prev_sel, self.config.copy_on_right_click);
-            if (prev_sel.contains(screen, pin)) break :sel;
+            if (prev_sel.contains(screen, pin)) {
+                try self.copyOnMouseAction(prev_sel, self.config.copy_on_right_click);
+                break :sel;
+            }
             // The selection doesn't contain our pin, so we create a new
             // word selection where we clicked.
         }
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -235,7 +235,8 @@ const DerivedConfig = struct {
     clipboard_trim_trailing_spaces: bool,
     clipboard_paste_protection: bool,
     clipboard_paste_bracketed_safe: bool,
-    copy_on_select: configpkg.CopyOnSelect,
+    copy_on_select: configpkg.CopyOnMouseAction,
+    copy_on_right_click: configpkg.CopyOnMouseAction,
     confirm_close_surface: configpkg.ConfirmCloseSurface,
     cursor_click_to_move: bool,
     desktop_notifications: bool,
@@ -297,6 +298,7 @@ const DerivedConfig = struct {
             .clipboard_paste_protection = config.@"clipboard-paste-protection",
             .clipboard_paste_bracketed_safe = config.@"clipboard-paste-bracketed-safe",
             .copy_on_select = config.@"copy-on-select",
+            .copy_on_right_click = config.@"copy-on-right-click",
             .confirm_close_surface = config.@"confirm-close-surface",
             .cursor_click_to_move = config.@"cursor-click-to-move",
             .desktop_notifications = config.@"desktop-notifications",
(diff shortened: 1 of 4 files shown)


src/config/Config.zig
```diff
@@ -1506,12 +1506,18 @@ keybind: Keybinds = .{},
...
 ///
 /// The default value is true on Linux and macOS.
-@"copy-on-select": CopyOnSelect = switch (builtin.os.tag) {
+@"copy-on-select": CopyOnMouseAction = switch (builtin.os.tag) {
     .linux => .true,
     .macos => .true,
     else => .false,
 },
 
+@"copy-on-right-click": CopyOnMouseAction = switch (builtin.os.tag) {
```
pluiedev, line 1515: Docs

src/config/Config.zig
```diff
@@ -1506,12 +1506,18 @@ keybind: Keybinds = .{},
...
+    .linux => .false,
+    .macos => .false,
+    else => .false,
```
pluiedev, line 1518: If it's just disabled everywhere then... just make it false?

pluiedev (changes requested):

jcollie:
Before we start merging pull requests, we should really spend some more time thinking about the best config for this.
Having a dedicated `copy-on-right-click` config seems way too specialized. Things will quickly get out of hand as we want to do other things with mouse button actions. In particular, pasting (instead of copying) with right click seems like a viable option as well. If we don't develop full-blown mouse button remapping as discussed in #3848 we should at least make the right click config a little more extensible.

I'd recommend something like this:

```
right-click-action = open-context-menu
right-click-action = copy
right-click-action = paste
```

`open-context-menu` would be the default.

src/Surface.zig
```diff
@@ -3103,14 +3109,15 @@ pub fn mouseButtonCallback(
         // If we already have a selection and the selection contains
         // where we clicked then we don't want to modify the selection.
         if (self.io.terminal.screen.selection) |prev_sel| {
+            try self.copyOnMouseAction(prev_sel, self.config.copy_on_right_click);
```
jcollie, line 3112: If you already have a selection and then right click outside the selection you get two "Copied to Clipboard" toasts (with `--copy-on-right-click=clipboard`) so one of these calls to `copyOnMouseAction` seems redundant or misplaced.

jcollie (changes requested):

src/config/Config.zig
jcollie: I think that this will partially work on macOS, except that it will _both_ copy the selection _and_ open up the context menu. I can't test, but that's my intuition.

src/Surface.zig
```diff
@@ -1469,6 +1458,22 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
...
 }
 
+fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
+    const prev_ = self.io.terminal.screen.selection;
+    try self.io.terminal.screen.select(sel_);
+
+    // If copy on select is false then exit early.
+    if (self.config.copy_on_select == .false) return;
+
+    // clear the clipboard. If the selection is set, we only set the clipboard
```
00-kat, line 1468: Insanely minor but
```suggestion
    // Clear the clipboard. If the selection is set, we only set the clipboard
```

src/config/Config.zig
IamTheLime: Yes, unfortunately I don't have a macbook at hand to test this on but the copy should work however the menu will still pop up. I will add the option to the docs

src/Surface.zig
IamTheLime: You are absolutely correct, I moved this inside the subsequent conditional, where you are clicking inside an existing selection.

src/config/Config.zig
```diff
@@ -1512,6 +1512,20 @@ keybind: Keybinds = .{},
...
+/// The value `clipboard` will always copy text to the selection clipboard
+/// as well as the system clipboard.
+///
+/// Middle-click paste will always use the selection clipboard. Middle-click
+/// paste is always enabled even if this is `false`.
+///
+/// The default value is false for all systems.
+///
+/// At present this will only disable the context menu on Linux in the future
+/// this shoul also disable the context menu on macOS
```
IamTheLime, line 1528: Kept the copy of this identical to `copy-on-select` Is  the preference to reference  the existing config for things like the behaviour of clipboard?

src/config/Config.zig
IamTheLime: Good point

IamTheLime:
> Before we start merging pull requests, we should really spend some more time thinking about the best config for this. Having a dedicated `copy-on-right-click` config seems way too specialized. Things will quickly get out of hand as we want to do other things with mouse button actions. In particular, pasting (instead of copying) with right click seems like a viable option as well. If we don't develop full-blown mouse button remapping as discussed in #3848 we should at least make the right click config a little more extensible.
> 
> I'd recommend something like this:
> 
> ```
> right-click-action = open-context-menu
> right-click-action = copy
> right-click-action = paste
> ```
> 
> `open-context-menu` would be the default.

This is a reasonable enough ask, and I think it makes sense.

I just read through @mitchellh's [response](https://github.com/ghostty-org/ghostty/issues/4404#issuecomment-2676560200)  on the original issue to this same suggestion

Should I close this MR and re-open one with this behaviour or even close it on a permanent basis until there is a proposal for a more permanent solution?

src/config/Config.zig
IamTheLime: [Website changes](https://github.com/ghostty-org/website/pull/302)

tjatjagithub:
I desperately need paste-on-right-click :-)

00-kat:
@tjatjagithub luckily GitHub was considerate enough to add a button you can press to voice your interest.
A screenshot of the thumbs up reaction on the original post.

IamTheLime:
@tjatjagithub You might want to have a look at the issue this is targeting. There is some meaningful discussion regarding how this is not perhaps the path to take. (In a pinch, should you really desperately need it, this MR adds minimal changes to ghostty and the instructions provided in the repo make it incredibly easy to build from source so you can always merge main into this branch and do a quick build of ghostty).

@00-kat  Should I close this MR to avoid any further confusion?

00-kat:
> Should I close this MR to avoid any further confusion?

No need; I was only saying that to tell them not to add useless +1 comments.

mitchellh:
If you're willing to rebase this, I agree with @jcollie. Ultimately _that too_ will be deprecated when we introduce bindable mouse actions, but I think `right-click-action` is a good stopgap. I don't want to block this PR/issue on full bindable mouse actions because this is a heavily requested feature, I'd rather deprecate later.

So, if you're willing to follow up, let's move to `right-click-action` and for this PR you can just do `context-menu` (default) and `copy`. Someone else can do paste if they want.

mitchellh (changes requested):
Note my comment above.

mitchellh:
Resolved by #8254

PR #5949, opened 2025-02-23
terminal: increase CSI max params to 24 to accept Kakoune sequence

mitchellh:
See #5930

Kakoune sends a real SGR sequence with 17 parameters. Our previous max was 16 so we threw away the entire sequence. This commit increases the max rather than fundamentally addressing limitations.

Practically, it took us this long to witness a real world sequence that exceeded our previous limit. We may need to revisit this in the future, but this is an easy fix for now.

In the future, as the comment states in this diff, we should probably look into a rare slow path where we heap allocate to accept up to some larger size (but still would need a cap to avoid DoS). For now, increasing to 24 slightly increases our memory usage but shouldn't result in any real world issues.

### DIFF
--- a/src/terminal/Parser.zig
+++ b/src/terminal/Parser.zig
@@ -86,7 +86,9 @@ pub const Action = union(enum) {
         final: u8,
 
         /// The list of separators used for CSI params. The value of the
-        /// bit can be mapped to Sep.
+        /// bit can be mapped to Sep. The index of this bit set specifies
+        /// the separator AFTER that param. For example: 0;4:3 would have
+        /// index 1 set.
         pub const SepList = std.StaticBitSet(MAX_PARAMS);
 
         /// The separator used for CSI params.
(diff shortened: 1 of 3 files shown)


rockorager (approved):

rockorager:
24 seems like a good cap. I think it's reasonably *possible* to hit, but agree that since it's taken so long to get past our 16 max it's unlikely.

0. Reset at beginning ("0")
1. RGB with colorspace is 6. Set fg, bg, and ul would be 18 total ("38:2::0:0:0")
2. Assume we set underline since we set underline color, that's another 2 with styled underline
3. Set up to three more attributes. There are 7 commonly available, but not all would make sense at the same time (eg invisible is a weird combo).

That puts us at a max of 28 if you were to try set everything possible at once.

mitchellh:
> That puts us at a max of 28 if you were to try set everything possible at once.

That's a good analysis and it may be worth going to 28 just for that but a part of me also just wants to wait and see what happens with 24. Theoretically it's just 28 for a reasonable, well-formed user. It can be infinite since you can set the same SGR multiple times in a single sequence. 😢 I don't think there is anything restricting that, anyways.

PR #5954, opened 2025-02-23
gtk: ensure that the content scale is always positive

### LINKED ISSUE #5927: GTK: content scale should never be negative
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5906

<sup>Originally posted by **ChausseBenjamin** February 20, 2025</sup>
When building the ghostty binary from source on a gentoo system. I am greeted with a blank screen.
Here are logs printed to stdout when launched from another terminal:
```
info: ghostty version=1.1.3-main+38908e01
info: runtime=apprt.Runtime.gtk
info: font_backend=font.main.Backend.fontconfig_freetype
info: dependency harfbuzz=8.4.0
info: renderer=renderer.OpenGL
info: libxev backend=main.Backend.io_uring
info(os): setlocale from env result=en_CA.utf8
info(gtk): GTK version build=4.16.12 runtime=4.16.12
info(gtk): libadwaita version build=1.6.2 runtime=1.6.2
info: reading configuration file path=/home/master/.config/ghostty/config
info(config): default shell source=env value=/bin/zsh
warning(gtk): setting GDK_DEBUG=gl-no-fractional
warning(gtk): setting GDK_DISABLE=gles-api,color-mgmt,vulkan
warning(gtk): setting GSK_RENDERER=opengl
Unrecognized value "color-mgmt". Try GDK_DISABLE=help
info(grid): loaded OpenGL 4.6
info(font_shared_grid_set): font regular: VulfMono-Regular
info(font_shared_grid_set): font bold: VulfMono-Bold
info(font_shared_grid_set): font italic: VulfMono-Italic
info(font_shared_grid_set): font bold_italic: VulfMono-BoldItalic
info(io_exec): found Ghostty resources dir: /home/master/.local/src/ghostty/zig-out/share/ghostty
info(io_exec): shell integration automatically injected shell=termio.shell_integration.Shell.zsh
warning(gtk): unimplemented action=apprt.action.Action.Key.cell_size
warning(surface): WARNING: very small terminal grid detected with padding set. Is your padding reasonable?
info(io_exec): started subcommand path=/bin/sh pid=24880
info(io_exec): subcommand cgroup=-
warning(gtk): unable to get current color scheme: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop
info(grid): reallocating GPU buffer old=0 new=8
```

Note that this occurs even on a blank config (no fonts or background). 
Also, the visuals work fine when compiling with the following command (albeit the app is slow then):
```
zig build -Dapp-runtime=glfw run
```

</div>

## Further Discussion

I don't know what causes this, but clearly in the discussion comments someone is getting a negative xscale/yscale from GTK. I can't think of any reason this should ever be allowed and it's causing problems later since we convert these to unsigned integers at some point after doing some math.

I'm not sure what the right behavior is but if a scale is negative, we should probably assume it's invalid and just revert back to `1`?

jcollie:
Fixes #5927

If GTK reports a negative scale, we ignore that and use 1.0 for the scale.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -852,8 +852,8 @@ pub fn getContentScale(self: *const Surface) !apprt.ContentScale {
         // Future: detect GTK version 4.12+ and use gdk_surface_get_scale so we
         // can support fractional scaling.
         const scale = widget.getScaleFactor();
-        if (scale < 0) {
-            log.warn("gtk_widget_get_scale_factor returned a negative number: {d:.3}", .{scale});
+        if (scale <= 0) {
+            log.warn("gtk_widget_get_scale_factor returned a non-positive number: {}", .{scale});
             break :scale 1.0;
         }
         break :scale @floatFromInt(scale);
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -847,28 +847,41 @@ pub fn shouldClose(self: *const Surface) bool {
 }
 
 pub fn getContentScale(self: *const Surface) !apprt.ContentScale {
-    // Future: detect GTK version 4.12+ and use gdk_surface_get_scale so we
-    // can support fractional scaling.
-    const gtk_scale: f32 = @floatFromInt(c.gtk_widget_get_scale_factor(@ptrCast(self.gl_area)));
+    const gtk_scale: f32 = scale: {
+        const widget: *gtk.Widget = @ptrCast(@alignCast(self.gl_area));
+        // Future: detect GTK version 4.12+ and use gdk_surface_get_scale so we
+        // can support fractional scaling.
+        const scale = widget.getScaleFactor();
+        if (scale < 0) {
+            log.warn("gtk_widget_get_scale_factor returned a negative number: {d:.3}", .{scale});
+            break :scale 1.0;
+        }
+        break :scale @floatFromInt(scale);
+    };
 
     // Also scale using font-specific DPI, which is often exposed to the user
     // via DE accessibility settings (see https://docs.gtk.org/gtk4/class.Settings.html).
     const xft_dpi_scale = xft_scale: {
         // gtk-xft-dpi is font DPI multiplied by 1024. See
         // https://docs.gtk.org/gtk4/property.Settings.gtk-xft-dpi.html
-        const settings = c.gtk_settings_get_default();
+        const settings = gtk.Settings.getDefault() orelse break :xft_scale 1.0;
+        var value = std.mem.zeroes(gobject.Value);
+        defer value.unset();
+        _ = value.init(gobject.ext.typeFor(c_int));
+        settings.as(gobject.Object).getProperty("gtk-xft-dpi", &value);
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/Surface.zig
```diff
@@ -847,28 +847,41 @@ pub fn shouldClose(self: *const Surface) bool {
...
-        const gtk_xft_dpi = c.g_value_get_int(&value);
+        const settings = gtk.Settings.getDefault() orelse break :xft_scale 1.0;
+        var value = std.mem.zeroes(gobject.Value);
+        defer value.unset();
+        _ = value.init(gobject.ext.typeFor(c_int));
+        settings.as(gobject.Object).getProperty("gtk-xft-dpi", &value);
+        const gtk_xft_dpi = value.getInt();
+
+        // ensure that the scale is never negative
+        if (gtk_xft_dpi < 0) {
```
pluiedev, line 875: Here's a crazy idea... what if the negative content scale problem was caused only by gtk-xft-dpi being negative? [The docs](https://docs.gtk.org/gtk4/property.Settings.gtk-xft-dpi.html) say apparently it can be -1 to signify the "default" value (whatever that means). We should corroborate this with users, obviously, but I suspect this is what's happening.

jcollie: That begs the question of what the "default" value is. From what I've googled it seems as if the default is whatever works out to a scale factor of 1.0, so basically no code changes necessary. Would be good to add that as a default.

jcollie: These commits to libadwaita seems to confirm `gtk-xft-dpi` <= 0 should be treated as 96 * 1024:

https://gitlab.gnome.org/GNOME/libadwaita/-/commit/a7738a4d269bfdf4d8d5429ca73ccdd9b2450421
https://gitlab.gnome.org/GNOME/libadwaita/-/merge_requests/1372/diffs?commit_id=9759d3fd81129608dd78116001928f2aed974ead

src/apprt/gtk/Surface.zig
```diff
@@ -847,28 +847,44 @@ pub fn shouldClose(self: *const Surface) bool {
...
 pub fn getContentScale(self: *const Surface) !apprt.ContentScale {
-    // Future: detect GTK version 4.12+ and use gdk_surface_get_scale so we
-    // can support fractional scaling.
-    const gtk_scale: f32 = @floatFromInt(c.gtk_widget_get_scale_factor(@ptrCast(self.gl_area)));
+    const gtk_scale: f32 = scale: {
+        const widget: *gtk.Widget = @ptrCast(@alignCast(self.gl_area));
+        // Future: detect GTK version 4.12+ and use gdk_surface_get_scale so we
+        // can support fractional scaling.
+        const scale = widget.getScaleFactor();
+        if (scale < 0) {
```
pluiedev, line 855: A scale of 0 is similarly nonsensical here I think

pluiedev (approved):

PR #5955, opened 2025-02-23
fix(macos): prevent performing newTab shortcut on QuickTerminalWindow…

### LINKED ISSUE #5939: macOS: Quick Terminal - Cmd-T should show error, not create new window/tab
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5642

<sup>Originally posted by **edrozenberg** February  8, 2025</sup>
While focused on Quick Terminal:

- Menu **File** -> **New Tab** shows error: "Cannot Create New Tab" ✅
- Keyboard shortcut "Cmd+T": creates a new window the same size and location as the Quick Terminal, or a new tab in an existing window. ❌

Suggestion: for consistency, when focused on Quick Terminal, both the menu action AND keyboard shortcut should have the same effect i.e. show the error dialog.</div>

McNight:
Fixes #5939

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -442,7 +442,7 @@ class QuickTerminalController: BaseTerminalController {
         }
     }
 
-    func showNoNewTabAlert() {
+    private func showNoNewTabAlert() {
         guard let window else { return }
         let alert = NSAlert()
         alert.messageText = "Cannot Create New Tab"
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -59,6 +59,11 @@ class QuickTerminalController: BaseTerminalController {
             selector: #selector(ghosttyConfigDidChange(_:)),
             name: .ghosttyConfigDidChange,
             object: nil)
+        center.addObserver(
+            self,
+            selector: #selector(onNewTab),
+            name: Ghostty.Notification.ghosttyNewTab,
+            object: nil)
     }
 
     required init?(coder: NSCoder) {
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/Terminal/TerminalManager.swift
```diff
@@ -125,6 +125,12 @@ class TerminalManager {
     }
 
     private func newTab(to parent: NSWindow, withBaseConfig base: Ghostty.SurfaceConfiguration?) {
+        // If the parent window is a QuickTerminalWindow, we early return with an alert.
+        if let controller = parent.windowController as? QuickTerminalController {
```
mitchellh, line 129: I think we should move this logic out to the `onNewTab` notification and instead of binding us to QuickTerminalController instead test that the controller is a `TerminalController` and ignore all else. This way if we introduce more terminal types then the manager only responds to that.

mitchellh (changes requested):
Requesting one slight redo.

I think in addition to what I said, moving QuickTerminalController to listen for the new tab notification and invert the logic from TerminalManager: if the parent is a QuickTerminalController then show the error.

This would keep all the QT-logic constrained to the QT.

mitchellh (approved):
This looks great. Thank you!

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -437,6 +442,16 @@ class QuickTerminalController: BaseTerminalController {
         }
     }
 
+    private func showNoNewTabAlert() {
+        guard let window else { return }
+        let alert = NSAlert()
+        alert.messageText = "Cannot Create New Tab"
+        alert.informativeText = "Tabs aren't supported in the Quick Terminal."
+        alert.addButton(withTitle: "OK")
+        alert.alertStyle = .warning
```
mdznr, line 451: https://developer.apple.com/design/human-interface-guidelines/alerts#macOS

> **Use a caution symbol sparingly.** Using a caution symbol… too frequently in your alerts diminishes its significance. Use the symbol only when extra attention is really needed, as when confirming an action that might result in unexpected loss of data. Don’t use the symbol for tasks whose only purpose is to overwrite or remove data, such as a save or empty trash.

PR #5962, opened 2025-02-24
feat: respect maximize config on macOS

### LINKED ISSUE #5928: macOS: Implement `maximize` config (works for GTK)
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5881

<sup>Originally posted by **taylrfnt** February 19, 2025</sup>
Porting over from Discord...

# Issue Details
When using the [`maximize` configuration option](https://ghostty.org/docs/config/reference#maximize), the Ghostty window is not drawn to the maximized size (see recording below) as expected.
maximize-repro

# Reproduction Steps
1. Create a minimal Ghostty config of `maximize = true`
2. Restart Ghostty (`super`+`q`)
3. Open a new Ghostty window, observe it is not maximized.

## Ghostty Configuration
```
maximize = true
```
## Update Channel
`tip`

## Ghostty Version
```
Ghostty 1.1.3-main+da32534e

Version
  - version: 1.1.3-main+da32534e
  - channel: tip
Build Config
  - Zig version: 0.13.0
  - renderer   : renderer.Metal
```

> [!NOTE]
> This might be tangentially related to #5783, but it seemed distinct enough (different configuration options) to warrant its own Discussion item.
</div>

Aaron-212:
Resolve #5928

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -375,7 +375,11 @@ class TerminalController: BaseTerminalController {
         // If we have only a single surface (no splits) and that surface requested
         // an initial size then we set it here now.
         if case let .leaf(leaf) = surfaceTree {
-            if let initialSize = leaf.surface.initialSize,
+            if config.maximize {
+                if let screen = window.screen ?? NSScreen.main {
+                    window.setFrame(screen.visibleFrame, display: true)
+                }
+            } else if let initialSize = leaf.surface.initialSize,
                let screen = window.screen ?? NSScreen.main {
                 // Get the current frame of the window
                 var frame = window.frame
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -377,27 +377,31 @@ class TerminalController: BaseTerminalController {
         if case let .leaf(leaf) = surfaceTree {
             if let initialSize = leaf.surface.initialSize,
```
mitchellh, line 378: I haven't looked into this yet, but just on the surface: I think this may block this from working sometimes? I thought a surface only has an initial size if `window-width/height` is set? And we want maximize to always work.

Aaron-212: Yeah you are right.

mitchellh (approved):

PR #5966, opened 2025-02-24
gtk: apply initial appearance settings before presenting

### LINKED ISSUE #5934: `title` config should be respected and set immediately on new windows, tabs
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5836

<sup>Originally posted by **vaygr** February 17, 2025</sup>
Steps to reproduce:

```
yabai -m rule --add app="Ghostty" title="^scratchterm$" scratchpad="scratchterm" grid=5:5:1:0:3:3
open -na Ghostty.app --args --title="scratchterm"
```

Even though the title is reported correctly once ghostty is launched:

```
{
        "id":5281,
        "pid":13867,
        "app":"Ghostty",
        "title":"scratchterm",
        "scratchpad":"",
        "frame":{
                "x":12.0000,
                "y":50.0000,
                "w":1488.0000,
                "h":920.0000
        },
        "role":"AXWindow",
        "subrole":"AXStandardWindow",
        "root-window":true,
        "display":1,
        "space":1,
        "level":0,
        "sub-level":-20,
        "layer":"normal",
        "sub-layer":"below",
        "opacity":1.0000,
        "split-type":"vertical",
        "split-child":"second_child",
        "stack-index":0,
        "can-move":true,
        "can-resize":true,
        "has-focus":true,
        "has-shadow":true,
        "has-parent-zoom":false,
        "has-fullscreen-zoom":true,
        "has-ax-reference":true,
        "is-native-fullscreen":false,
        "is-visible":true,
        "is-minimized":false,
        "is-hidden":false,
        "is-floating":false,
        "is-sticky":false,
        "is-grabbed":false
}
```

But the rule never gets triggered, that is unless I remove `title` argument for `yabai -m rule` or set it to `^.*$`.

I tested it with kitty to see if it could be yabai's issue, but no, it worked as expected:

```
yabai -m rule --add app="kitty" title="^scratchterm$" scratchpad="scratchterm" grid=5:5:1:0:3:3
kitty --title "scratchterm"
```

My assumption is ghostty could be setting the title a bit too late?</div>

## Notes

Right now, the core Surface sends a `set_title` apprt action AFTER the surface is initialized, which -- at least for macOS -- happens to be after the Window (tabs are also windows in macOS) is created. 

If the user sets an explicit `title` config, we should set that as part of the window creation process rather than wait for this message.

pluiedev:
Fixes #5934 (only on the GTK side), #5960

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -7,6 +7,10 @@ const Window = @This();
 
 const std = @import("std");
 const builtin = @import("builtin");
+
+const gtk = @import("gtk");
+const gobject = @import("gobject");
+
 const build_config = @import("../../build_config.zig");
 const Allocator = std.mem.Allocator;
 const assert = std.debug.assert;
@@ -73,11 +77,12 @@ pub const DerivedConfig = struct {
     gtk_wide_tabs: bool,
     gtk_toolbar_style: configpkg.Config.GtkToolbarStyle,
 
+    title: ?[:0]const u8,
     maximize: bool,
     fullscreen: bool,
     window_decoration: configpkg.Config.WindowDecoration,
 
-    pub fn init(config: *const configpkg.Config) DerivedConfig {
+    pub fn init(config: *const configpkg.Config, alloc: Allocator) !DerivedConfig {
         return .{
             .background_opacity = config.@"background-opacity",
             .background_blur = config.@"background-blur",
@@ -88,10 +93,15 @@ pub const DerivedConfig = struct {
             .gtk_wide_tabs = config.@"gtk-wide-tabs",
             .gtk_toolbar_style = config.@"gtk-toolbar-style",
 
+            .title = if (config.title) |t| try alloc.dupe(u8, t) else null,
             .maximize = config.maximize,
             .fullscreen = config.fullscreen,
             .window_decoration = config.@"window-decoration",
         };
+    }
+
+    pub fn deinit(self: *DerivedConfig, alloc: Allocator) void {
+        if (self.title) |t| alloc.free(t);
     }
 };
 
@@ -114,7 +124,7 @@ pub fn init(self: *Window, app: *App) !void {
     self.* = .{
         .app = app,
         .last_config = @intFromPtr(&app.config),
-        .config = DerivedConfig.init(&app.config),
+        .config = try DerivedConfig.init(&app.config, app.core_app.alloc),
         .window = undefined,
         .headerbar = undefined,
         .tab_overview = null,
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/Window.zig
```diff
@@ -14,6 +14,8 @@ const configpkg = @import("../../config.zig");
 const font = @import("../../font/main.zig");
 const input = @import("../../input.zig");
 const CoreSurface = @import("../../Surface.zig");
+const gtk = @import("gtk");
+const gobject = @import("gobject");
```
jcollie, line 18: Mitchell's standard order for imports is anything from `std`, 3rd party libs, internal imports.

src/apprt/gtk/Window.zig
```diff
@@ -88,6 +91,7 @@ pub const DerivedConfig = struct {
             .gtk_wide_tabs = config.@"gtk-wide-tabs",
             .gtk_toolbar_style = config.@"gtk-toolbar-style",
 
+            .title = config.title,
```
jcollie, line 94: This should be copied, and then freed in a `deinit` function, or when reconfigured.

jcollie (changes requested):

gtjoseph:
Doesn't compile for me...
```
zig build -Doptimize=ReleaseFast \
	-Demit-terminfo -Demit-termcap -freference-trace=20
install
└─ install ghostty
   └─ zig build-exe ghostty ReleaseFast native
      └─ run glib-compile-resources (ghostty_resources.c)
         └─ run generate_gresource_xml
            └─ run gtk_blueprint_compiler (menu-surface-context_menu.ui) failure
error: FileNotFound
/usr/lib/zig/std/process/Child.zig:423:5: 0x103e8b8 in waitPosix (gtk_blueprint_compiler)
    return self.term.?;
    ^
```

pluiedev:
You need to install `blueprint-compiler` and `libxml2` (for `xmllint`) - it's a new dependency for 1.2

jcollie:
> Doesn't compile for me...

You need to install `blueprint-compiler`. It's a new build requirement.

gtjoseph:
I hacked around the compile issue with the alloc.dupe in DerivedConfig.init() but the window still draws with the default size before using the configured size.

pluiedev:
I could try to move it even earlier into the initialization process, but then I don't know if it'd help either. Might be just GTK being GTK.

gtjoseph:
How about my idea in the discussion of starting the window minimized then unminimizing it after all setup is done?

tristan957:
> How about my idea in the discussion of starting the window minimized then unminimizing it after all setup is done?

Can't you just hide the widget (window) and show it after everything is done?

pluiedev:
This PR was an attempt to fix your issue, but it's not *just* for your issue. You're free to use your own fix.

mitchellh:
> Fixes https://github.com/ghostty-org/ghostty/issues/5934 (only on the GTK side), https://github.com/ghostty-org/ghostty/discussions/5960

GitHub is going to auto-close #5934 when this merges so just be aware to reopen it and just relabel for macOS. ❤️

PR #5968, opened 2025-02-24
gtk: switch clipboard confirmation to zig-gobject and blueprints

jcollie:
Note that for Debian 12, the blueprints must be compiled on a distro with a newer version of `blueprint-compiler` and the raw UI XML committed to git. Debian 12 includes `blueprint-compiler` 0.6.0 which doesn't support compiling `adw.MessageDialog` even though the version of `libadwaita` supports it.


src/apprt/gtk/ui/ccw-osc-52-write-15.blp
```diff
@@ -0,0 +1,23 @@
+using Gtk 4.0;
+using Adw 1;
+translation-domain "com.mitchellh.ghostty";
+
+Adw.AlertDialog clipboard_confirmation_window {
```
pluiedev, line 5: IIRC Blueprint files can contain multiple *objects*, just not multiple *templates*. I think the OSC 52 variants at least can and should be merged with their non-OSC 52 variants, and then we decide which object to use at runtime.

pluiedev (changes requested):

tristan957:
I agree with everything @pluiedev said. Good work so far.

src/apprt/gtk/ui/ccw-osc-52-write-15.blp
jcollie: Unfortunately for multiple objects in the same file it won't quite work, you get conflicting IDs on the responses:

image

I'll have a think about folders/naming.

mitchellh (commented):
I agree with the naming convention (as in, I think we can do better). The rest looks great.

jcollie:
> I agree with the naming convention (as in, I think we can do better).

Yeah, I think that I have it nailed down. Gotta do some testing and add some docs.

src/apprt/gtk/ui/ccw-osc-52-write-15.blp
jcollie: Fixed.

pluiedev (changes requested):

pluiedev (approved):
:shipit:

PR #5974, opened 2025-02-24
"Return to Default Size" implementation

### LINKED ISSUE #1328: "Return to Default Size" feature
I apologize if this already exists and I missed it, but I think a feature to return the window to the configured startup size would be nice. The macOS terminal has this and I used it regularly when it was my main terminal.

image

CrazyMillerGitHub:
## Added Support for "Return To Default Size"

This update introduces support for the **"Return To Default Size"** feature.

### Fixes  
- Resolves [#1328](https://github.com/ghostty-org/ghostty/issues/1328)

### Screenshots  

| Description | Screenshot |
|------------|------------|
| **Ghostty** | Screenshot 2025-02-24 at 21 15 38 |
| **After changing window size** | Screenshot 2025-02-24 at 21 16 00 |
| **Native Terminal App (for reference)** | Screenshot 2025-02-24 at 21 16 20 |

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -753,6 +753,16 @@ class AppDelegate: NSObject,
         quickController.toggle()
     }
 
+    @IBAction func returnToDefaultSize(_ sender: Any) {
+        guard let window = NSApp.keyWindow else { return }
+        let currentOrigin = window.frame.origin
+        let newFrame = NSRect(
+            origin: currentOrigin,
+            size: WindowConfig.defaultSize
+        )
+        window.setFrame(newFrame, display: true)
+    }
+
     /// Toggles visibility of all Ghosty Terminal windows. When hidden, activates Ghostty as the frontmost application
     @IBAction func toggleVisibility(_ sender: Any) {
         // If we have focus, then we hide all windows.
@@ -804,6 +814,10 @@ class AppDelegate: NSObject,
         }
     }
 
+    private enum WindowConfig {
+        static let defaultSize = CGSize(width: 800, height: 600)
+    }
+
     private struct ToggleVisibilityState {
         let hiddenWindows: [Weak<NSWindow>]
         let keyWindow: Weak<NSWindow>?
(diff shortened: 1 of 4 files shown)


macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -804,6 +814,10 @@ class AppDelegate: NSObject,
         }
     }
 
+    private enum WindowConfig {
+        static let defaultSize = CGSize(width: 800, height: 600)
```
CrazyMillerGitHub, line 818: Extracted to separate config, if there's a better place, can move there

mitchellh: This isn't correct a static value. We should be using the default size for the window that might be set with the config. 

There's a lot of logic in `TerminalWindow.windowDidLoad` I think around this. That should probably be extracted in some way so it can be reused by AppD.

mitchellh (changes requested):
Notes in the comments.

macos/Sources/App/macOS/AppDelegate.swift
CrazyMillerGitHub: I see that window-height && window.width described only in docs and there's no connection between initial height/width now, but I suppose that it should be(based on doc), will support gathering this info from config

macos/Sources/Ghostty/Ghostty.Config.swift
```diff
@@ -140,7 +140,21 @@ extension Ghostty {
             guard let ptr = v else { return "" }
             return String(cString: ptr)
         }
-        
+
+        var windowHeight: Int16? {
+            guard let config = self.config else { return nil }
+            var v: Int16 = 0
+            let key = "window-height"
```
CrazyMillerGitHub, line 147: Reference: 
https://ghostty.org/docs/config/reference#window-height
https://ghostty.org/docs/config/reference#window-width

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -574,6 +574,23 @@ class TerminalController: BaseTerminalController {
         window.close()
     }
 
+    @IBAction func returnToDefaultSize(_ sender: Any) {
+        guard
+            let window = NSApp.keyWindow,
+            let initialWidth = ghostty.config.windowWidth,
```
mitchellh, line 580: These are in grid cells. So we can't simply read the config. This is unfortunately not the correct approach.

mitchellh (changes requested):
Thanks again. I think this is not the right approach unfortunately, please note my comment. I think the right approach is probably something like extracting our initial size out to a function and calling that so we run the same logic on windowDidLoad. 

One issue is we wouldn't respect changes to `window-width/height` though... but we can justify that initially by saying we return the default size when the window was created...

mitchellh:
Okay, I made some changes to the `initial_size` apprt action is triggered anytime that would change, so all surfaces can be aware of what their initial size would be, that should help this significantly.

mitchellh:
I'm going to pull down the code and make some updates myself... thanks for taking it this far. ❤️

mitchellh (approved):

mitchellh:
Done, redid stuff but also kept a good bit of what you did. Feel free to take a look!

CrazyMillerGitHub:
> Done, redid stuff but also kept a good bit of what you did. Feel free to take a look!

Checked diff, significant improvement was made by you, @mitchellh. Glad that part of code was helpful. Looks like true. And also thanking for detailed feedback regarding code. Hope to be helpful maybe later with some tasks to improve my main terminal app))

PR #5977, opened 2025-02-25
Fix empty keybind setting

### LINKED ISSUE #5936: `keybind =` (blank) should restore default keybinds
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5816

<sup>Originally posted by **levito** February 16, 2025</sup>
According to https://ghostty.org/docs/config#syntax, options with empty values should set them to the default. But with `keybind =`, all keybinds are removed, including the default ones.

Also, according to https://ghostty.org/docs/config#file-location, multiple config files are applied after one another, with the macOS config in `~/Library` overwriting the `~/.config` one.

With this behavior, I wanted to create two config files to have dotfiles working on both Linux and macOS:
1. `~/.config/ghostty/config` with options for both systems and keybinds for Linux
2. `~/Library/Application\ Support/com.mitchellh.ghostty/config` should then unset the Linux keybinds with `keybinds =` and set different ones for macOS

Did I miss anything, or is the current behavior that `keybind =` removes all keybinds including the default ones a bug?</div>

unknown:
Fixes https://github.com/ghostty-org/ghostty/issues/5936.

Extracts the default keybind setting to an `init` function. Add logic to call `init` in `parseIntoField` if it is defined for the type.

### DIFF
--- a/src/cli/args.zig
+++ b/src/cli/args.zig
@@ -767,6 +767,29 @@ test "parseIntoField: ignore underscore-prefixed fields" {
     try testing.expectEqualStrings("12", data._a);
 }
 
+test "parseIntoField: struct with default func" {
+    const testing = std.testing;
+    var arena = ArenaAllocator.init(testing.allocator);
+    defer arena.deinit();
+    const alloc = arena.allocator();
+
+    var data: struct {
+        a: struct {
+            const Self = @This();
+
+            v: []const u8,
+
+            pub fn setToDefault(self: *Self, _alloc: Allocator) !void {
+                _ = _alloc;
+                self.v = "HELLO!";
+            }
+        },
+    } = undefined;
+
+    try parseIntoField(@TypeOf(data), alloc, &data, "a", "");
+    try testing.expectEqual(@as([]const u8, "HELLO!"), data.a.v);
+}
+
 test "parseIntoField: string" {
     const testing = std.testing;
     var arena = ArenaAllocator.init(testing.allocator);
--- a/src/cli/args.zig
+++ b/src/cli/args.zig
@@ -247,16 +247,6 @@ pub fn parseIntoField(
 
     inline for (info.Struct.fields) |field| {
         if (field.name[0] != '_' and mem.eql(u8, field.name, key)) {
-            // If the value is empty string (set but empty string),
-            // then we reset the value to the default.
-            if (value) |v| default: {
-                if (v.len != 0) break :default;
-                const raw = field.default_value orelse break :default;
-                const ptr: *const field.type = @alignCast(@ptrCast(raw));
-                @field(dst, field.name) = ptr.*;
-                return;
-            }
-
             // For optional fields, we just treat it as the child type.
             // This lets optional fields default to null but get set by
             // the CLI.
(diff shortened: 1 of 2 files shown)


src/config/Config.zig
```diff
@@ -5130,6 +5134,13 @@ pub const Keybinds = struct {
             return;
         }
 
+        if (std.mem.eql(u8, value, "")) {
```
unknown, line 5137: I didn't copy the above comment regarding clearing memory given the proximity but happy to copy it here if needed.

src/config/Config.zig
```diff
@@ -1074,7 +1074,7 @@ class: ?[:0]const u8 = null,
 /// Ghostty will attempt to request these permissions. If the permissions are
 /// not granted, the keybind will not work. On macOS, you can find these
 /// permissions in System Preferences -> Privacy & Security -> Accessibility.
-keybind: Keybinds = .{},
```
jcollie, line 1077: All of the fields of `Config` should have a default value so that it can be instantiated as `.{}`.

src/config/Config.zig
jcollie: Just to be sure I'd copy it.

src/config/Config.zig
```diff
@@ -5133,6 +5134,13 @@ pub const Keybinds = struct {
             return;
         }
 
+        if (std.mem.eql(u8, value, "")) {
+            log.info("config has 'keybind =', using default keybind", .{});
```
jcollie, line 5138: `keybind` at the end should be plural.

src/config/Config.zig
jcollie: A slight optimization would to be compare `value.len == 0`.

jcollie (changes requested):

src/config/Config.zig
unknown: This requires thinking about how to change to [parseIntoField](https://github.com/unknown/ghostty/blob/5dc016e0d087439627c8816dd7d204e05074aa22/src/cli/args.zig#L238). The current logic is to use the default field value as a fallback when no value is provided. Not sure what the best option here would be.

jcollie: Move `setDefaultKeybinds` to a method on `Keybinds` called `setToDefault`. In [`parseIntoFields`](https://github.com/unknown/ghostty/blob/5dc016e0d087439627c8816dd7d204e05074aa22/src/cli/args.zig#L254) check for that method with `@hasDecl` and call it instead of using the raw default value.

unknown: Makes sense, thanks for the help!

jcollie (approved):

mitchellh (changes requested):
One easy request. 

A couple notes for the future (we should NOT do them in this PR, just to keep things simple, but can follow up):

1. We should move the keybindings out to a dedicated file. They've been causing too much noise in `Config.zig` for too long and this reminded me. `src/config/keybinding.zig` or something.

2. We should make a `src/cli/args.init` function that takes a struct and uses comptime to find all the fields that have an `init` and call them so that we have one place we can always use to default any "config" structure (we use args for other stuff too).

src/cli/args.zig
```diff
@@ -761,6 +767,29 @@ test "parseIntoField: ignore underscore-prefixed fields" {
...
+
+    var data: struct {
+        a: struct {
+            const Self = @This();
+
+            v: []const u8,
+
+            pub fn init(self: *Self, _alloc: Allocator) !void {
+                _ = _alloc;
+                self.v = "HELLO!";
```
mitchellh, line 784: FYI this isn't safe, you should be doing `self.* = .{ .v = "HELLO!" };`

mitchellh (changes requested):
Apologies, one small thing. I'd fix it myself but not at a PC rn.

mitchellh (approved):

mitchellh:
Perfect. Thanks!

unknown:
> Apologies, one small thing. I'd fix it myself but not at a PC rn.



> Perfect. Thanks!

No worries! I can work on the followups when I get the chance.

PR #5978, opened 2025-02-25
gtk: convert window actions to use zig-gobject

jcollie:


src/apprt/gtk/Window.zig
```diff
@@ -475,36 +481,38 @@ fn toggleCssClass(
 /// menus and such. The menu is defined in App.zig but the action is defined
 /// here. The string name binds them.
 fn initActions(self: *Window) void {
+    // FIXME: when rest of file is converted to gobject
+    const window: *gtk.Window = @ptrCast(@alignCast(self.window));
+    const action_map = gobject.ext.cast(gio.ActionMap, window).?;
```
pluiedev, line 486: I think the window should instead be casted to `gtk.ApplicationWindow`, then you can do `window.as(gio.ActionMap)`, and that would be entirely safe and checked at compile time.

pluiedev (changes requested):

pluiedev (approved):
:shipit:

PR #5984, opened 2025-02-25
Fix Cmd+H shortcut sending 'h' character when app is last visible

### LINKED ISSUE #5929: macOS: Cmd+H should not send "h" if its the last visible window
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5870

<sup>Originally posted by **Samasaur1** February 19, 2025</sup>
On macOS, there are two easy shortcuts for hiding applications. Command-H hides the current application, while Option-Command-H hides all other applications. However, you must always have at least one unhidden application (enforced by the operating system). With all other apps, if the given app is the last unhidden app and you attempt to hide it with command-H, nothing happens. However, Ghostty handles the failing Command-H shortcut as a normal shortcut, which results in a normal `h` being sent to the terminal (when using the default config).

To reproduce:
- Make sure at least one Finder window is open (I'm not 100% sure this is necessary to reproduce, but hiding Finder behaves oddly when there are no windows)
- Open Ghostty
- Hide all other apps with Option-Command-H
- Attempt to hide Ghostty with Command-H
- Notice that you are typing `h`s into the terminal</div>

## Notes

I'm not sure how easy this is to fix, but we should match other application behavior in this case because `Cmd+H` is a system shortcut. I didn't do full research to see what other terminals do, but assuming Terminal.app and iTerm behave this way (h is not shown) we should do the same.

liby:
Fixes #5929

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -956,6 +956,30 @@ extension Ghostty {
 
             let equivalent: String
             switch (event.charactersIgnoringModifiers) {
+            case "h":
+                // Special handling for Cmd+H (hide application)
+                // Only handle if it's exactly Cmd+H with no other modifiers
+                if (event.modifierFlags.contains(.command) &&
+                    event.modifierFlags.isDisjoint(with: [.shift, .control, .option])) {
+
+                    let visibleWindowCount = NSApp.windows.filter { $0.isVisible }.count
+                    let isLastVisibleApp = NSWorkspace.shared.runningApplications
+                        .filter { $0.activationPolicy == .regular && $0.isHidden == false }
+                        .count <= 1
+
+                    // If we're likely the last visible app, consume the event and don't pass it to the terminal
+                    if isLastVisibleApp {
+                        return true
+                    }
+
+                    // Otherwise, let macOS handle it normally
+                    return false
+                }
+
+                // For other combinations with 'h', continue with normal processing
+                // This allows combinations like Shift+Cmd+H to be processed normally
+                equivalent = "h"
+
             case "/":
                 // Treat C-/ as C-_. We do this because C-/ makes macOS make a beep
                 // sound and we don't like the beep sound.
@@ -1252,7 +1276,7 @@ extension Ghostty {
                 AppDelegate.logger.warning("action failed action=\(action)")
             }
         }
-        
+
         @IBAction func changeTitle(_ sender: Any) {
             promptTitle()
         }


macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -956,6 +956,30 @@ extension Ghostty {
 
             let equivalent: String
             switch (event.charactersIgnoringModifiers) {
+            case "h":
+                // Special handling for Cmd+H (hide application)
+                // Only handle if it's exactly Cmd+H with no other modifiers
+                if (event.modifierFlags.contains(.command) &&
+                    event.modifierFlags.isDisjoint(with: [.shift, .control, .option])) {
```
mitchellh, line 963: I thought about this, but `Cmd+H` is probably remappable via system settings yeah? If so, we probably need to figure out a way to block it in that case. I just want to look into that before merging this.

mitchellh (changes requested):
One note

macos/Sources/Ghostty/SurfaceView_AppKit.swift
liby: Thank you for your review. I've researched this issue and found that `Cmd+H` is a system-level shortcut in macOS that [cannot be directly remapped through system settings](https://github.com/warpdotdev/Warp/issues/2973#issuecomment-1631728610). Users would need third-party tools like [Karabiner-Elements](https://karabiner-elements.pqrs.org/) to modify such system shortcuts.

I've also explored an alternative approach to avoid directly checking for `Cmd+H` in the performKeyEquivalent method:

1. I tried removing the special handling for `Cmd+H` in `performKeyEquivalent` and letting the system handle all `Cmd+H` key events normally.

2. I implemented the `applicationWillHide` method in `AppDelegate`, which should be called when the application is about to be hidden, regardless of how the hide operation is triggered.

3. In `applicationWillHide`, I checked if Ghostty is the last visible application and set a flag `isHandlingHideOperation`.

4. In the `keyDown` method, I checked this flag and prevented any keyboard events from being processed if the flag was `true`.

However, during testing, I discovered that when Ghostty is the last visible window, macOS doesn't trigger the `applicationWillHide` method at all. This suggests that we can't avoid checking for `Cmd+H` directly in our keyboard event handling code.

mitchellh:
Thanks for the detailed follow up to my comment. 

In my own investigation, I think the Cmd+H issue is only one small part of a larger problem (#6057) and I think this is actually a key encoding issue. If I'm wrong I'll reopen this.

Thank you!

Issue #5256, opened 2025-01-20
macOS: Use system API to open links


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5200

<div type='discussions-op-text'>

<sup>Originally posted by **sindresorhus** January 18, 2025</sup>
Currently, it uses the `open` command. The problem with using that is that it makes it impossible for browser pickers apps to detect that the link is opened from Ghostty, because by the time the app checks for the source app, the `open` command has exited. It should also be slightly faster to use the system API: https://developer.apple.com/documentation/appkit/nsworkspace/open(_:)?language=objc

I'm referring to: https://github.com/ghostty-org/ghostty/blob/72d085525b22d66468c5969a4d507a0fa68d4a04/src/os/open.zig#L32C9-L41C11

---

Specifically, [Velja](https://sindresorhus.com/velja) uses the source app to let users create advanced rules for which browser to use depending on the source app.</div>

shaps80:
Given this is a native app we could actually use SwiftUI's `openURL` from the `Environment`. This also ensures future compatibility with iOS too. This also simplifies URL overriding (if required) in SwiftUI.

Specifically I'm referring to:
https://developer.apple.com/documentation/swiftui/environmentvalues/openurl

matthebrown:
I was interested in taking a crack at this feature but was looking for some more clarity on the required changes. Would the approach here involve invoking the `NSWorkspace.open` method from the existing `open.zig` file like below? Or are we talking about handling links in the native SwiftUI?

```zig
const NSWorkspace = objc.getClass("NSWorkspace");
const shared = NSWorkspace.msgSend(objc.Object, objc.sel("shared"));
 _ = shared.msgSend(bool, objc.sel("open:"), .{ url });
```

shaps80:
@matthebrown personally my preference would be the safer high-level APIs provided by SwiftUI.

Was there any reason specifically you wanted implement it using older Obj-C `msgSend` APIs?

matthebrown:
That makes sense, I'm just a little unsure what to do with this existing logic in `open.zig`. I guess we need to keep it as-is since it's also used to open text.

jcollie:
Perhaps we need to convert opening URLs into an apprt action. That way it would be easier for each apprt to implement native experiences, plus make it easier for macOS to use native Swift code to implement the action.

shaps80:
I agree, I like the idea that it allows each platform to handle things more appropriately as well 👍

[...]

PR #5988, opened 2025-02-25
core/gtk: open urls using an apprt action instead of doing it directly

jcollie:
Partial implementation of #5256

This implements the core changes necessary to open urls using an apprt action rather than doing it directly from the core.

Implements the open_url action in the GTK apprt.

Note that this should not be merged until a macOS-savvy developer can add an implementation of the open_url action for the macOS apprt.

### DIFF
--- a/src/apprt/action.zig
+++ b/src/apprt/action.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const assert = std.debug.assert;
 const apprt = @import("../apprt.zig");
 const configpkg = @import("../config.zig");
@@ -267,6 +268,11 @@ pub const Action = union(Key) {
 
     check_for_updates,
 
+    /// Open a URL using the native OS mechanisms. On macOS this might be `open`
+    /// or on Linux this might be `xdg-open`. The exact mechanism is up to the
+    /// apprt.
+    open_url: OpenUrl,
+
     /// Sync with: ghostty_action_tag_e
     pub const Key = enum(c_int) {
         quit,
@@ -317,6 +323,7 @@ pub const Action = union(Key) {
         undo,
         redo,
         check_for_updates,
+        open_url,
     };
 
     /// Sync with: ghostty_action_u
@@ -357,7 +364,13 @@ pub const Action = union(Key) {
         // For ABI compatibility, we expect that this is our union size.
         // At the time of writing, we don't promise ABI compatibility
         // so we can change this but I want to be aware of it.
-        assert(@sizeOf(CValue) == 16);
+        assert(@sizeOf(CValue) == switch (builtin.target.os.tag) {
+            .windows => switch (builtin.target.cpu.arch) {
+                .x86 => 16,
+                else => 24,
+            },
+            else => 24,
+        });
     }
 
     /// Returns the value type for the given key.
@@ -611,6 +624,40 @@ pub const ConfigChange = struct {
     pub fn cval(self: ConfigChange) C {
         return .{
             .config = self.config,
+        };
+    }
+};
+
+/// The type of the data at the URL to open. This is used as a hint to
+/// potentially open the URL in a different way.
+/// Sync with: ghostty_action_open_url_kind_s
+pub const OpenUrlKind = enum(c_int) {
+    text,
+    unknown,
+};
+
+/// Open a URL
+pub const OpenUrl = struct {
+    /// The type of data that the URL refers to.
+    kind: OpenUrlKind,
+    /// The URL.
+    url: []const u8,
+
+    // Sync with: ghostty_action_open_url_s
+    pub const C = extern struct {
+        /// The type of data that the URL refers to.
+        kind: OpenUrlKind,
+        /// The URL (not zero terminated).
+        url: [*]const u8,
+        /// The number of bytes in the URL.
+        len: usize,
+    };
+
+    pub fn cval(self: OpenUrl) C {
+        return .{
+            .kind = self.kind,
+            .url = self.url.ptr,
+            .len = self.url.len,
         };
     }
 };
--- a/src/apprt/action.zig
+++ b/src/apprt/action.zig
@@ -1,5 +1,4 @@
 const std = @import("std");
-const builtin = @import("builtin");
 const assert = std.debug.assert;
 const apprt = @import("../apprt.zig");
 const configpkg = @import("../config.zig");
@@ -364,12 +363,10 @@ pub const Action = union(Key) {
         // For ABI compatibility, we expect that this is our union size.
         // At the time of writing, we don't promise ABI compatibility
         // so we can change this but I want to be aware of it.
-        assert(@sizeOf(CValue) == switch (builtin.target.os.tag) {
-            .windows => switch (builtin.target.cpu.arch) {
-                .x86 => 16,
-                else => 24,
-            },
-            else => 24,
+        assert(@sizeOf(CValue) == switch (@sizeOf(usize)) {
+            4 => 16,
+            8 => 24,
+            else => unreachable,
         });
     }
 
--- a/src/apprt/action.zig
+++ b/src/apprt/action.zig
@@ -625,28 +625,35 @@ pub const ConfigChange = struct {
     }
 };
 
-/// The type of the data at the URL to open. This is used as a hint to
-/// potentially open the URL in a different way.
-/// Sync with: ghostty_action_open_url_kind_s
-pub const OpenUrlKind = enum(c_int) {
-    text,
-    unknown,
-};
-
 /// Open a URL
 pub const OpenUrl = struct {
     /// The type of data that the URL refers to.
-    kind: OpenUrlKind,
+    kind: Kind,
+
     /// The URL.
     url: []const u8,
 
+    /// The type of the data at the URL to open. This is used as a hint to
+    /// potentially open the URL in a different way.
+    ///
+    /// Sync with: ghostty_action_open_url_kind_e
+    pub const Kind = enum(c_int) {
+        /// The type is unknown. This is the default and apprts should
+        /// open the URL in the most generic way possible. For example,
+        /// on macOS this would be the equivalent of `open` or on Linux
+        /// this would be `xdg-open`.
+        unknown,
+
+        /// The URL is known to be a text file. In this case, the apprt
+        /// should try to open the URL in a text editor or viewer or
+        /// some equivalent, if possible.
+        text,
+    };
+
     // Sync with: ghostty_action_open_url_s
     pub const C = extern struct {
-        /// The type of data that the URL refers to.
-        kind: OpenUrlKind,
-        /// The URL (not zero terminated).
+        kind: Kind,
         url: [*]const u8,
-        /// The number of bytes in the URL.
         len: usize,
     };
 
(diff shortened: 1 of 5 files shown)


tristan957 (approved):

tristan957:

src/os/open.zig
```diff
@@ -77,3 +81,131 @@ const OpenCommand = struct {
...
+        &.{ "xdg-open", url },
+        alloc,
+    );
+
+    // We're only interested in stderr
+    exe.stdin_behavior = .Ignore;
+    exe.stdout_behavior = .Ignore;
+    exe.stderr_behavior = .Pipe;
+
+    exe.spawn() catch |err| {
```
mitchellh, line 123: Having bunch of `catch` in this is nasty and makes it so `err defer` can't be used (if we even wanted to). I recommend the following pattern instead:

```zig
fn openUrlLinuxThread(...) void {
    // Handle error once
    openUrlLinuxThreadError(...) catch |err| {}
}

fn openUrlLinuxThreadError(...) !void {
    // Return errors like normal, use errdefer.
}
```

mitchellh (changes requested):
Thanks for starting this, I can take a look for libghostty.

src/os/open.zig
jcollie: Fixed.

src/apprt/action.zig
```diff
@@ -555,3 +561,27 @@ pub const ConfigChange = struct {
...
+};
+
+pub const OpenUrl = struct {
+    kind: OpenUrlKind,
+    url: [:0]const u8,
+
+    // Sync with: ghostty_action_open_url_s
+    pub const C = extern struct {
+        kind: OpenUrlKind,
+        url: [*:0]const u8,
```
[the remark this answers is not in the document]
jcollie: Fixed.

jcollie:
> Did you see https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.OpenURI.html? Seems like we should use that and then fallback to xdg-open.

That looks interesting but I'd rather take that on in a follow-up PR after this is merged.

src/apprt/action.zig
```diff
@@ -341,7 +348,13 @@ pub const Action = union(Key) {
...
         // At the time of writing, we don't promise ABI compatibility
         // so we can change this but I want to be aware of it.
-        assert(@sizeOf(CValue) == 16);
+        assert(@sizeOf(CValue) == switch (builtin.target.os.tag) {
+            .windows => switch (builtin.target.cpu.arch) {
+                .x86 => 16,
+                else => 24,
+            },
+            else => 24,
+        });
```
jcollie, line 357: This is really sus but I don't know of a better way to handle this.

pluiedev: My incredibly crackpot guess would be that this might be because Windows is LLP64 (sizeof(long) == 4) whereas both Linux and macOS are LP64 (sizeof(long) == 8). If this is not it then I legit don't have any more ideas

jcollie: OK, I think that I have a better check now:
```
-        assert(@sizeOf(CValue) == switch (builtin.target.os.tag) {
-            .windows => switch (builtin.target.cpu.arch) {
-                .x86 => 16,
-                else => 24,
-            },
-            else => 24,
+        assert(@sizeOf(CValue) == switch (@sizeOf(usize)) {
+            4 => 16,
+            8 => 24,
+            else => unreachable,
```

src/apprt/action.zig
```diff
@@ -601,3 +611,37 @@ pub const ConfigChange = struct {
         };
     }
 };
+
+/// The type of the data at the URL to open. This is used as a hint to
+/// potentially open the URL in a different way.
+/// Sync with: ghostty_action_open_url_kind_s
+pub const OpenUrlKind = enum(c_int) {
+    text,
+    unknown,
```
pluiedev, line 620: Need more doc comments explaining what each of these variants means

src/os/open.zig
```diff
@@ -77,3 +75,125 @@ const OpenCommand = struct {
     child: std.process.Child,
     wait: bool = false,
 };
+
+/// Use `xdg-open` to open a URL using the default application.
+///
+/// Any output on stderr is logged as a warning in the application logs. Output
+/// on stdout is ignored.
+pub fn openUrlLinux(
```
pluiedev, line 83: I don't really love the name here... Maybe `openUrlXdg` makes more sense - this would also allow us to add something like `openUrlPortal`. Or maybe, we should just add a "method" enum as the parameter to `open`, that looks something like

```zig
pub const OpenMethod = enum {
    /// Use macOS's builtin `open` utility to open URLs.
    open,
    /// Use the `xdg-open` utility to open URLs. Available on most Linux distributions.
    xdg_open,
    /// Use the OpenURL method of the XDG Desktop Portal on Linux.
    xdg_desktop_portal,
}
```

src/os/open.zig
```diff
@@ -2,20 +2,18 @@ const std = @import("std");
...
 ///
 /// Any output on stderr is logged as a warning in the application logs.
 /// Output on stdout is ignored.
 pub fn open(
     alloc: Allocator,
-    typ: Type,
+    kind: apprt.action.OpenUrlKind,
     url: []const u8,
 ) !void {
     const cmd: OpenCommand = switch (builtin.os.tag) {
```
pluiedev, line 19: Also it seems like the Linux path is still extant in the original `open` function - is this intended? This looks a bit redundant and not really used anymore

pluiedev (changes requested):

tristan957 (commented):
Would `open_uri` be a better name than `open_url`?

src/os/open.zig
```diff
@@ -77,3 +75,125 @@ const OpenCommand = struct {
...
+
+    // We're only interested in stderr
+    exe.stdin_behavior = .Ignore;
+    exe.stdout_behavior = .Ignore;
+    exe.stderr_behavior = .Pipe;
+
+    exe.spawn() catch |err| {
+        switch (err) {
+            error.FileNotFound => {
+                log.err("Unable to find xdg-open. Please install xdg-open and ensure that it is available on the PATH.", .{});
```
00-kat, line 131: The package name is [`xdg-utils`] pretty much [everywhere]; maybe the hint here should tell the user to install `xdg-utils` instead of `xdg-open`.

[`xdg-utils`]: https://www.freedesktop.org/wiki/Software/xdg-utils
[everywhere]: https://repology.org/project/xdg-utils/versions

src/apprt/action.zig
```diff
@@ -614,3 +624,37 @@ pub const ConfigChange = struct {
         };
     }
 };
+
+/// The type of the data at the URL to open. This is used as a hint to
+/// potentially open the URL in a different way.
+/// Sync with: ghostty_action_open_url_kind_s
+pub const OpenUrlKind = enum(c_int) {
```
tristan957: I don't think I understand why this is necessary. Couldn't the apprt just sniff the URL value?

tristan957:
> Would `open_uri` be a better name than `open_url`?

Is this something you've considered? Maybe I'm blind.

tristan957:
> Did you see https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.OpenURI.html? Seems like we should use that and then fallback to xdg-open.

TIL that xdg-open uses the portal already, so I wonder if there is any benefit to using the portal directly vs just calling xdg-open. Maybe lower overhead?

jcollie:
> TIL that xdg-open uses the portal already, so I wonder if there is any benefit to using the portal directly vs just calling xdg-open. Maybe lower overhead?

Yeah, you could avoid the thread spawn and fork/exec in Ghostty, but that's pretty minimal since I doubt that it's going to be called thousands of times per second. The benefit of calling `xdg-open` is making the logic of making the magic of that happen "not our problem".

src/os/open.zig
jcollie: This error message is a little more generic now (it just uses argv[0]) since I merged the macOS, Windows, and Linux logic so getting that specific would be more complex.

src/apprt/action.zig
jcollie: It's basically because macOS' `open` has a `-t` flag that gives it a hint that it's opening the file in a `text` context so presumably it'll open in an editor rather than something else (like opening HTML files in a browser).

jcollie:
I've updated the PR to rebase on top of #7657 so there's no special "open on linux" function anymore. This still needs a macOS implementation of the `open_url` action before merging.

pluiedev:
Actually, I just realized that `xdg-open` isn't just a wrapper around the D-Bus API — instead it's a bash script that uses a bunch of heuristics to figure out what the user's desktop environment is, and then uses the DE-specific utilities when possible (`gio open` on GNOME, `kde-open` on KDE, etc.). In fact, it seems to only use the portal when it's running inside Flatpak. I guess then it would still be worth keeping it in case the user doesn't have a working portal (which seems to be concerningly common for a lot of our users)...?

tristan957:
> Actually, I just realized that `xdg-open` isn't just a wrapper around the D-Bus API — instead it's a bash script that uses a bunch of heuristics to figure out what the user's desktop environment is, and then uses the DE-specific utilities when possible (`gio open` on GNOME, `kde-open` on KDE, etc.). In fact, it seems to only use the portal when it's running inside Flatpak. I guess then it would still be worth keeping it in case the user doesn't have a working portal (which seems to be concerningly common for a lot of our users)...?

So then, I think my position will continue to be prioritize the portal with an xdg-open fallback (future work). Thanks for your research.

tristan957:
Hm, I guess I have a question. Why does libghostty do the physical opening. Why isn't it left to the apprt? Meaning the macOS apprt would run `open` for instance.

jcollie:
> Hm, I guess I have a question. Why does libghostty do the physical opening. Why isn't it left to the apprt? Meaning the macOS apprt would run `open` for instance.

The `open` function is in `libghostty` because there's a lot of functionality shared across all the apprts (it's also used by the "open config". If an apprt wanted to do something other than launch a CLI command (some sort of DBus call for GTK, native Swift API call for macOS, some Win32 call for Windows) that can be handled in the `open_url` apprt handler (GTK: https://github.com/ghostty-org/ghostty/blob/00f824d0efb441bbfe2057e7caeadfcec2ea1d48/src/apprt/gtk/App.zig#L1722-L1730)

src/os/open.zig
```diff
@@ -2,14 +2,10 @@ const std = @import("std");
 const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 
-const log = std.log.scoped(.@"os-open");
+const apprt = @import("../apprt.zig");
+const CircBuf = @import("../datastruct/circ_buf.zig").CircBuf;
```
mitchellh: On a future note, your PRs have a pattern of containing unrelated changes such as this. I'm going to revert all the circular buffer stuff. It makes sense but conflating the purpose and goal of a PR makes review harder and makes things like reverting uglier if we ever to. I'd appreciate keeping things tightly scoped.

mitchellh:
Okay, I made it so that if the apprt doesn't support this we fallback to our old implementation. This will let us merge this without worrying about macOS. I'll make a follow up issue for that. Once CI passes we're good.

mitchellh (approved):

PR #5992, opened 2025-02-25
Fix elvish sudo integration and update documentation

Yappaholic:
Elvish integration is broken when running `sudo`, because the function `sudo-with-terminfo` uses `command` command, which is not implemented in Elvish. Changing it to `external` command should fix possible error when bypassing aliases, functions and builtins, like `command` does in Bash.
Discussion about this issue: #5979 

Also I updated documentation about Elvish integration to provide fix when using `use ghostty-integration` outside of Ghostty with `rc.elv`.


mitchellh:
I don't use elvish but googling backs this up and the docs improvements are appreciated.

mitchellh (approved):

PR #5995, opened 2025-02-26
fix clipboard spam when dragging mouse

### LINKED ISSUE #4800: `copy-on-select = clipboard` causes "Copied to clipboard" toast spam

unknown:
Fixes https://github.com/ghostty-org/ghostty/issues/4800 and succeeds https://github.com/ghostty-org/ghostty/pull/5010.

As outlined in https://github.com/ghostty-org/ghostty/pull/5010#issuecomment-2611217928, changes mouse events to call `terminal.screen.select` and only call `setSelection` at the end of a mouse drag. This ensures that the `select_all` action still writes the selection to the clipboard.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2892,6 +2892,24 @@ pub fn mouseButtonCallback(
         }
     }
 
+    // The selection clipboard is only updated when the left button is released
+    // when copy on select is enabled.
+    if (button == .left and
+        action == .release and
+        self.config.copy_on_select != .false)
+    {
+        self.renderer_state.mutex.lock();
+        defer self.renderer_state.mutex.unlock();
+        const prev_ = self.io.terminal.screen.selection;
+        if (prev_) |prev| {
+            try self.setSelection(terminal.Selection.init(
+                prev.start(),
+                prev.end(),
+                false,
+            ));
+        }
+    }
+
     // Report mouse events if enabled
     {
         self.renderer_state.mutex.lock();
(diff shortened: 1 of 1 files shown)


pluiedev:
Removing the os/linux tag as I somehow *just* realized that this is purely a core issue

JeffreyVdb:
Screenshot_20250326_095638

I'd really love to have this merged in. This happens when I copy "ls -l"

albertonoys:
I'd really love this. Here's another example of my clipboard manager in Ubuntu, copying `thisisjustatest`:

mitchellh:
Superseded by #7704, thank you!

PR #5996, opened 2025-02-26
gtk: switch dbus operations in src/apprt/gtk/App.zig to zig-gobject

jcollie:


tristan957 (approved):
Now that we depend on libadwaita, I think we can just remove this code. We just subscribe to the style manager, but any removal should probably happen after this translation is merged. What do you think?

jcollie:
> Now that we depend on libadwaita, I think we can just remove this code. We just subscribe to the style manager, but any removal should probably happen after this translation is merged. What do you think?

Yeah, I think we'll tackle that in a new PR.

Issue #3020, opened 2024-12-19
Linux: Split resize area too large


Sirius902:
## Issue

The area in which you can click and drag to resize splits is too large, it is even possible to resize the splits without the cursor showing that it may be resized. With the horizontal split configuration shown below, the offending area seems to be larger on the edge bordering the right split. In the video, you can see that I can select starting from the third character from the left on the split border but not any closer to the border.

I can only reproduce this on my Linux box, not on macOS.

https://github.com/user-attachments/assets/14452dfd-8e98-4b8e-8d93-eb54879b3eb4

## Expectation

I would expect the following regardless of the OS in use, currently these expectations seem to be met on macOS but not with my Linux setup:
* It is possible to select characters bordering splits using the mouse (difficult right now due to this issue).
* Clicking and dragging will not resize splits unless the cursor is also changed to reflect this.
* The area you can click and drag to resize splits is appropriately sized to the split border.

## Configuration

Running on NixOS with GNOME.

ghostty built from flake with commit fa3896be03398cd8d718a1342af94767a8c219bf.

akhilerm:
Facing the same issue. This means that sometime it is difficult to select characters and text that start at the left of the split.

carragom:
Having the same problem, reported as duplicate in #5630. Bit sad to hear this is acceptable.

daviewales:
It is strange to me that this is a 'GTK issue', when other GTK apps on my system don't exhibit this behaviour. (Unless it's a GTK3 vs GTK4 issue? Or a Libadwaita issue?) I have noted in a related discussion that Nemo for example has a split divider between the left sidebar and the directory view, and it works as expected. In particular, the drag surface in Nemo and the drag cursor indicator are consistent, unlike Gostty. The drag surface in Nemo is also much narrower then in Ghostty.
See: https://github.com/ghostty-org/ghostty/discussions/5655#discussioncomment-12113416

daviewales:
**Confirming that this _is_ indeed a Gtk 4 issue**:

The following minimal Gtk 3 app _does not_ have the issue:

``` python
# https://stackoverflow.com/a/69773849

import gi

gi.require_version("Gtk", "3.0")
from gi.repository import Gtk

class App(Gtk.Window):
    """The window."""

    def __init__(self):
        super().__init__()
        self.connect("delete-event", Gtk.main_quit)

        self.set_default_size(200, 200)

        # The paned window
        self.paned = Gtk.Paned()
[...]

PR #6000, opened 2025-02-26
gtk: work around oversized drag handle for GtkPaned

daviewales:
Improves #3020.

Based on recommendation from upstream Gtk issue:
https://gitlab.gnome.org/GNOME/gtk/-/issues/4484#note_2362002

Without this, it's not possible to select the first character on the right-hand side of a split.

### DIFF
--- a/src/apprt/gtk/style.css
+++ b/src/apprt/gtk/style.css
@@ -48,4 +48,18 @@ window.ssd.no-border-radius {
 .terminal-window .notebook paned > separator {
   background-color: rgba(250, 250, 250, 1);
   background-clip: content-box;
+
+  /* This works around the oversized drag area for the right side of GtkPaned.
+   *
+   * Upstream Gtk issue:
+   * https://gitlab.gnome.org/GNOME/gtk/-/issues/4484#note_2362002
+   *
+   * Ghostty issue:
+   * https://github.com/ghostty-org/ghostty/issues/3020
+   *
+   * Without this, it's not possible to select the first character on the
+   * right-hand side of a split.
+   */
+  margin: 0;
+  padding: 0;
 }


daviewales:
This is a hopefully noncontroversial subset of #6002. They share commit 33e1930, which works around a Gtk bug which causes the pane drag handle to be oversized. This improves the issue, as it was previously not possible to select the first character in a right-hand split pane. With this PR it is possible if you start around halfway through the character. It's not a complete fix, as the drag handle still covers the left half of the first character in the right pane.

This PR is attempting to get something similar into Gtk upstream:
https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8238

#6002 adds an extra commit to enable a wide handle for the pane separator, which functionally resolves the issue completely, but has some style issues.

My thinking is that this PR can hopefully be merged without too much trouble, and then #6002 can take a bit longer to figure out the best complete solution.

tristan957:
Thanks for working with upstream!

src/apprt/gtk/style.css
```diff
@@ -41,6 +41,22 @@ window.ssd.no-border-radius {
   border-radius: 0 0;
 }
 
+paned > separator {
```
jcollie, line 44: This should be combined with the `paned > separator` CSS section later in the file.

jcollie: Probably need to run this through `prettier` as well.

jcollie (changes requested):

pluiedev (approved):

src/apprt/gtk/style.css
daviewales: Done.

jcollie (approved):

PR #6002, opened 2025-02-26
gtk: Use wide handle for GtkPaned

daviewales:
Improves #3020.

Ensures that the drag handle for the pane separator never overlaps pane content.

See recommendation in upstream Gtk issue:
https://gitlab.gnome.org/GNOME/gtk/-/issues/4484#note_2362002

**Note**: This is a more experimental / invasive alternative to #6000, as it changes the appearance and width of the split pane separator.

However, it is also more functional, as it is no longer possible to accidentally drag the pane handle when you intended to select text.

This should perhaps be a setting, but I'm not sure how to do that.

On reflection, a nicer approach would just be to include a small `6px` margin to the right of the split separator. (This is what WezTerm appears to do.) This maintains the nice appearance of a single split divider, and the usability of keeping the content away from the drag handle.

### DIFF
--- a/src/apprt/gtk/style.css
+++ b/src/apprt/gtk/style.css
@@ -41,6 +41,22 @@ window.ssd.no-border-radius {
   border-radius: 0 0;
 }
 
+paned > separator {
+  /* This works around the oversized drag area for the right side of GtkPaned.
+   *
+   * Upstream Gtk issue:
+   * https://gitlab.gnome.org/GNOME/gtk/-/issues/4484#note_2362002
+   *
+   * Ghostty issue:
+   * https://github.com/ghostty-org/ghostty/issues/3020
+   *
+   * Without this, it's not possible to select the first character on the
+   * right-hand side of a split.
+   */
+    margin: 0;
+    padding: 0;
+}
+
 .transparent {
   background-color: transparent;
 }
--- a/src/apprt/gtk/Split.zig
+++ b/src/apprt/gtk/Split.zig
@@ -111,6 +111,12 @@ pub fn init(
     // Keep a long-lived reference, which we unref in destroy.
     _ = c.g_object_ref(paned);
 
+    // Ensure that the drag handle for split panes never overlaps
+    // pane content. (#3020)
+    // See recommendation in upstream Gtk issue:
+    // https://gitlab.gnome.org/GNOME/gtk/-/issues/4484#note_2362002
+    c.gtk_paned_set_wide_handle(@ptrCast(paned), @intFromBool(true));
+
     // Update all of our containers to point to the right place.
     // The split has to point to where the sibling pointed to because
     // we're inheriting its parent. The sibling points to its location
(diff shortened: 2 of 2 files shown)


pluiedev:
Could you show the visual effect of this? Neither the GTK docs nor the linked issues seem to actually show how it looks like...

jcollie:
I can't say that I love the way the wide handles look, at least by default. If we accept this, we're going to need to do some more CSS work so that the "separator" is still 1px wide with padding on either side that's the appropriate background color, transparency etc.

src/apprt/gtk/style.css
```diff
@@ -41,6 +41,22 @@ window.ssd.no-border-radius {
   border-radius: 0 0;
 }
 
+paned > separator {
```
jcollie, line 44: This shoud be part of the `paned > separator` section a little further down. We also need to tweak things to make wide tabs look a little more like they used to.

daviewales:
Here's a screenshot of how it looks with the wide handle:

image

I agree 100% with @jcollie that it needs some tweaking.

You can see in this image that WezTerm (behind) adds just the right amount of space around the drag handle to prevent it overlapping the text content. Compare to Ghostty in the foreground:

image

Another screenshot comparing the padding around the handle for Ghostty (foreground) and WezTerm (background), without the wide handle:
image

src/apprt/gtk/style.css
daviewales: Do you mean, move the block to be next to the `.terminal-window .notebook paned > separator` block? Or move the `margin` and `padding` settings within that block?

An argument for keeping them separate is that Gtk are looking at making this the default for all GtkPaned handles. If I move it into the existing block, then it will only apply to the separator for the terminal splits, not to all Paned separators. However, as this improves a general Gtk bug, does it make sense to apply to all separators, to prevent this issue affecting any other case where GtkPaned might be used?

jcollie:
@daviewales now that #6000 is merged, I think that we should make using wide separators a config option that users can opt into. Until either GTK upstream figures things out or we get a CSS wizard to fix up the look of the wide separators we can let users choose their poison.

Adding a config option isn't _too_ difficult. Let us know if you want to tackle that or would like one of the core maintainers to take over that part.

mitchellh:
> I can't say that I love the way the wide handles look, at least by default. 

I agree, I'm not a fan of this. 

I'm not even sure if this should be configurable. I propose we close this and discuss this further before committing to any solution (or rather, risk committing to multiple solutions). For example, the padding option proposed here is reasonable but it'd suck to introduce two options eventually (one for this and one for padding). 

On macOS, we just accept that an invisible portion of the surrounding area of a divider is prioritized for dragging the divider. Other terminals with splits on macOS at least seem to behave similarly.

daviewales:
I agree. We should be aiming to [eliminate any _necessary configuration_](https://ghostty.org/docs/config). My preference is to figure out how to get the functionality of the wide handle (no drag handle overlap with content) with the appearance of the original pane separator (a single 1 px line). In practice, this will likely look like a vertical line with padding on either side.

I'm happy to close this and move the conversation to discussions until someone figures out the best way to achieve this.

daviewales:
Continued discussion here: https://github.com/ghostty-org/ghostty/discussions/6059

PR #6004, opened 2025-02-26
gtk: add localization support, take 3

pluiedev:
This is my third (!) attempt at implementing localization support. By leveraging GTK builder to do most of the `gettext` calls, I can avoid the whole mess about missing symbols on non-glibc platforms.

Added some documentation too for contributors and translators, just for good measure.

Supersedes #5214, resolves the GTK half of #2357


mitchellh (commented):
I'm supportive of this but just noting I want to review this. I'm less experienced with this so want to better understand some things that are perhaps trivial or obvious:

1. Is it typical for `xgettext` to run during binary build time or should we start considering this as a step we run prior to packaging up our source tarballs for packagers? i.e. is this a dependency on packaging or is this a dependency on source building?

Making it all one step for day to day dev makes total sense, just thinking about the final release time packaging here. I know it's typical for source tarballs to contain some preprocessing complete but unsure if this falls in that category.

2. I know it's typical for `po` files to go at the root, but we may want to make a note this is only for GTK currently. I need to do research on what the recommended approach for macOS is (we don't need to solve that in this PR but I want to make sure we're structured right for it)

pluiedev:
1. xgettext is usually run by developers as a standalone tool outside of the build system. Its sole purpose is to generate/update the PO template, and once the template is there only `msgfmt` is required for the build process (which does .po -> .mo)

2. On macOS there's two ways to go about it: either do it the "legacy" way of using `.strings` files that can be added to Xcode, or the "new" (Xcode 15+) way of using what's known as "string catalogs" `.xcstring`s, which are essentially JSON files with a fancy editing interface inside Xcode.

What I've gathered is that while string catalogs are definitely the recommended translation mechanism from Apple moving forward and is a lot more convenient to use for the macOS apprt (`Text`s in SwiftUI transparently use them), they're kind of a pain for thirdparty integrations. The gettext stack has native support for legacy `.strings` files and so we can unify translations for both apprts, while `.xcstrings` is completely foreign to it (or rather, any program that's not Xcode). I think it would be up to you and other macOS devs to decide which format we should use on the macOS side.

mitchellh:
> xgettext is usually run by developers as a standalone tool outside of the build system, therefore it's neither a build-time nor a runtime dependency.

Right, but we're only committing the po files to the repo, right?

So if I'm understanding properly, we should plan to run this prior to generating the source tarball so packagers do not need to do this.

pluiedev:
Yeah, my plan is to always update the PO template *whenever* strings are added, changed or removed in PRs, enforced via CI. That way the template is always accurate to the source files (which is already desirable since they include source locations) and translators always have an up-to-date pictures of what strings need to be translated. Packagers don't need to do anything.

jcollie:
I don't do any i18n/l10n, but I've never been comfortable having the `.pot` file checked into source. We at least should have a CI job that fails if the strings change and the template is out of date.

pluiedev:
That's what I've been planning to do :p

.github/workflows/test.yml
```diff
@@ -593,6 +594,38 @@ jobs:
...
+      - uses: cachix/install-nix-action@v30
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - uses: cachix/cachix-action@v15
+        with:
+          name: ghostty
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+          skipPush: true
+          useDaemon: false # sometimes fails on short jobs
+      - name: typos check
```
jcollie, line 621: this should be updated

tristan957:
> xgettext is usually run by developers as a standalone tool outside of the build system

It is better when it is part of the build system, coming from a Meson world.

pluiedev:
Well it is part of the build system (I've added a new top-level step to run it via `zig run update-translations`), just not part of the normal build process, because altering source files in the Zig build system without explicit user intervention is frowned upon.

tristan957:
Cool, cool, cool.

pluiedev:
@kenvandine Looks like the snap package needs to be updated with the dependency on gettext. I am completely clueless about snaps, however, so I'd really appreciate some assistance on this front :sweat_smile:

src/apprt/gtk/i18n.zig
```diff
@@ -0,0 +1,37 @@
+//! I18n support for the GTK frontend based on gettext/libintl
+//!
+//! This is normally built into the C standard library for the *vast* majority
+//! of users who use glibc, but for musl users we fall back to the `gettext-tiny`
+//! stub implementation which provides all of the necessary interfaces.
+//! Musl users who do want to use localization should know what they need to do.
```
mitchellh, line 6: I don't think Ghostty even builds for musl right now anyways but this is a good thought regardless to not give us any more debt to work on there. (I think the debt is mainly we have other deps that don't play nice with musl)

mitchellh (commented):
Looking good. I want to run this locally later today just to make sure I really understand the full build path here, and that may result in some other code comments but they're likely superficial.

src/apprt/gtk/i18n.zig
pluiedev: Yeah honestly like... I don't even know how people who use musl even manage to coax GTK & Friends to work on it, but if they do manage to do it then I *presume* they have enough know-how to get localization working

kenvandine (changes requested):
For the snap you need to add gettext to build-packages. Something like the following will work:

```
diff --git a/snap/snapcraft.yaml b/snap/snapcraft.yaml
index 49d452ef..9822619d 100644
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -75,6 +75,7 @@ parts:
       - libadwaita-1-dev
       - libxml2-utils
       - git
+      - gettext
       - patchelf
     override-build: |
       craftctl set version=$(git describe --abbrev=8)
```

src/apprt/gtk/App.zig
```diff
@@ -268,6 +269,8 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         std.posix.exit(1);
     };
 
+    try i18n.init(core_app.alloc);
```
mitchellh, line 272: A comment here that explains why this has to happen here in ordering and what exactly this is doing would be helpful.

mitchellh (changes requested):
One very easy change requested. The rest looks good.

I made one commit to update the build files. The only real change there is I think that top-level steps should only be created in `build.zig` and our helper structs should create unattached `*std.Build.Step` that can be depended on instead. 

I also did some minor style modifications to fit an 80 character width, which despite having a large screen I still find helpful for splits.

src/apprt/gtk/App.zig
pluiedev: I kinda just put it in an arbitrary place - *ideally* you want to initialize it before any localized messages can appear, and since we don't have any localized log messages you can put it anywhere really

mitchellh: Yeah I don't particularly care, I think its right where it is, but a comment explaining what this does and why here is helpful to future maintenance.

kenvandine:
> Got it. However, I wonder if if'd work OOTB - I see that LC_ALL is forcibly set to C.UTF-8 and there's some forum threads that suggest more work would be needed for it to work. (e.g. https://forum.snapcraft.io/t/the-gettext-launch-launcher-fix-gettext-based-internationalization-in-the-snap-runtime/9111, https://forum.snapcraft.io/t/lack-of-compiled-locales-breaks-gettext-based-localisation/3758) Is this still accurate?

Your current branch fails to build the snap because it can't find msgfmt which is provided by the gettext package. We need it for build time. For strictly confined snaps following our best practices, you get gettext automatically at build time, but ghostty is a classic snap that is a bit more complex.

pluiedev:
How does one add it as a build-time dependency? I'm again utterly clueless about Snaps so I'd really appreciate some more pointers. The CI looks green though...

kenvandine:
> Quote reply
Look at the inline patch I had included in my review.  You should be able to follow that, it's just adding one line in the snapcraft.yaml

kenvandine:
> I've already added that though. Squash commits are a bit hard to follow :p

Ah, great!  Then should be good from a snap perspective. Thanks!

pluiedev:
> Ah, great! Then should be good from a snap perspective. Thanks!

Can you rereview the PR to approve it? GitHub isn't allowing me to merge it right now

tristan957 (approved):

kenvandine (approved):
Snap changes look good, thanks!

po/README_TRANSLATORS.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Creating new translation files
+
+You can use the `msginit` tool to create new translation files.
+
+Run the command below, optionally replacing `$LANG` with the name of a locale
+that is _different_ to your system locale, or if the `LANG` environmental
+variable is not set.
+
+> [!WARNING] > **Make sure your selected locale uses the UTF-8 encoding, as it is the sole
```
trag1c, line 99: Looks like a missing newline here, so this doesn't render correctly right now:
```suggestion
> [!WARNING]
> **Make sure your selected locale uses the UTF-8 encoding, as it is the sole
```

po/README_TRANSLATORS.md
```diff
@@ -0,0 +1,148 @@
+# Localizing Ghostty: The Translators' Guide
+
+First of all, thanks for helping us localize Ghostty!
+
+To begin contributing, please make sure that you have installed the `gettext`
+package on your system, which should be available under the name `gettext`
+for most Linux and macOS package managers.
```
trag1c, line 7: Subjective, but calling `gettext` a package didn't feel right to me:
```suggestion
To begin contributing, please make sure that you have `gettext` installed on
your system, which should be available under the name `gettext` for most Linux
and macOS package managers.
```
Feel free to resolve this comment if you disagree.

po/README_CONTRIBUTORS.md
```diff
@@ -0,0 +1,58 @@
...
+    // Translators: This is the name of the button that opens the about dialog.
+    title: _("About Ghostty");
+```
+
+By default identical strings are collapsed together into one translatable entry.
+To avoid this, assign a _context_ to the string:
+
+```
+    label: C_("menu action", "Copy");
+```
```
trag1c, line 31: Unsure why these are indented, feels like an unintended clash with the fenced code block syntax? I'd also throw in some Zig highlighting:
`````suggestion
```zig
// Translators: This is the name of the button that opens the about dialog.
title: _("About Ghostty");
```

By default identical strings are collapsed together into one translatable entry.
To avoid this, assign a _context_ to the string:

```zig
label: C_("menu action", "Copy");
```
`````

**Review by trag1c — CHANGES_REQUESTED** [review]

File: po/README_CONTRIBUTORS.md (hunk 6 above)
Comment (pluiedev) [reply po/README_CONTRIBUTORS.md:31]: That's blueprint syntax and not Zig :p

File: po/README_TRANSLATORS.md (hunk 5 above)
Comment (pluiedev) [reply po/README_TRANSLATORS.md:7]: It's to differentiate gettext the tool and gettext the thing you can install. Maybe could be rephrased better

File: po/README_CONTRIBUTORS.md (hunk 6 above)
Comment (trag1c) [reply po/README_CONTRIBUTORS.md:31]: My bad! I just saw line 16 say "in Zig code" without reading prior text and jumped to concluding this is Zig. Not sure if my indentation note is also irrelevant now.

Comment (00-kat) [reply po/README_CONTRIBUTORS.md:31]: Rust works pretty well.
A screenshot of the rendered markdown using Rust syntax highlighting.

File: po/README_TRANSLATORS.md [hunk 7]
```diff
@@ -0,0 +1,148 @@
...
+
+On Windows `gettext` is available through the `scoop` and `winget` package
```
00-kat, line 9: Perhaps a link to the home page itself for people who don't fancy using Windows package managers?  https://mlocati.github.io/articles/gettext-iconv-windows.html is the one both Scoop and WinGet have.

po/README_TRANSLATORS.md
```diff
@@ -0,0 +1,148 @@
...
+```console
+$ msginit -i po/com.mitchellh.ghostty.pot -l $LANG -o "po/$LANG.po"
+```
+
+`msginit` may prompt you for other information such as your email address,
+which should be filled in accordingly. You can then add your translations
+within the newly created translation file.
+
+Afterwards, you need to update the list of known locales within Ghostty's
+build system. To do so, open `src/build/GhosttyI18n.zig` and find the list
```
00-kat, line 114: Perhaps an example diff here and/or a comment at https://github.com/ghostty-org/ghostty/blob/02e6cc11eb5cbd1ecf45db6c7be35e71060ca774/src/build/GhosttyI18n.zig#L9 would be helpful, or at least while the list is small enough to overlook.

po/README_TRANSLATORS.md
```diff
@@ -0,0 +1,148 @@
...
+  any background knowledge of the English source text. To _localize_ is to
+  transfer a concept between languages, not to translate each word at face value.
+
+- Be consistent with stylistic and tonal choices. Consult the translations
+  made by previous translators, and try to emulate them. Do not overwrite
+  someone else's hard work without justification.
+
+- Make Ghostty fit in with surrounding applications. Follow existing
+  translations for terms and concepts if possible, even if they may be suboptimal.
+  Follow the writing styles proscribed by the human interface guidelines of
```
00-kat, line 145:
```suggestion
  Follow the writing styles prescribed by the human interface guidelines of
```
Slightly different meaning, and the original has a negative connotation.

po/README_TRANSLATORS.md
```diff
@@ -0,0 +1,148 @@
...
+  each platform Ghostty is available for, including the [GNOME Human Interface Guidelines](https://developer.gnome.org/hig/guidelines/writing-style.html)
+  on Linux, and [Apple's Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/writing)
```
00-kat, line 147: > on Linux

What about non-GNOME-centric apprts?  Are all translations currently apprt-specific?  What about ones such as “You're running a debug build of Ghostty! Performance will be degraded”[^1]?

[^1]: This string looks to be defined in a GTK-specific location according to the translation template, but I believe this is the same one macOS uses?

po/com.mitchellh.ghostty.pot
```diff
@@ -0,0 +1,204 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR Mitchell Hashimoto
+# This file is distributed under the same license as the com.mitchellh.ghostty package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+#, fuzzy
```
00-kat, line 6: Should these `#,` comments be documented too?  (I assume they're either only found in template files[^1] or unimportant?)

[^1]: `zh_CN.UTF-8.po` doesn't have it, but I know nothing about gettext so I want to make sure.

po/README_CONTRIBUTORS.md
```diff
@@ -0,0 +1,58 @@
...
+Provide context to translators by adding a `// Translators: ` comment on the
+line above the string in question, such as:
+
+```
+    // Translators: This is the name of the button that opens the about dialog.
+    title: _("About Ghostty");
+```
+
+By default identical strings are collapsed together into one translatable entry.
+To avoid this, assign a _context_ to the string:
```
00-kat, line 27: Do these context strings follow the same syntax as outlined over at https://github.com/ghostty-org/ghostty/blob/02e6cc11eb5cbd1ecf45db6c7be35e71060ca774/po/README_TRANSLATORS.md#L62?  Or is there any extra line that is included (other than the link to the source code) denoting the context?

src/apprt/gtk/i18n.zig
00-kat, line 6: > should know what they need to do.

You detail what they need to do—install `libintl`—in one of the markdown files, right?  Why does this comment have to be so vague?

po/README_CONTRIBUTORS.md
00-kat, line 1: A note about the existence of these two README files should *probably* be mentioned in the [`CONTRIBUTING.md`].

[`CONTRIBUTING.md`]: https://github.com/ghostty-org/ghostty/blob/ee8ae196ee0fad3824c7ae9eac947f2128b4ae4f/CONTRIBUTING.md#developer-guide

po/com.mitchellh.ghostty.pot
```diff
@@ -0,0 +1,204 @@
+# SOME DESCRIPTIVE TITLE.
```
00-kat, line 1: I assume this template is directly copied when one attempts to create a new translation file, but if not a comment such as `# THIS FILE IS AUTO-GENERATED FROM THE GHOSTTY SOURCE CODE, DO NOT EDIT.` would be helpful.

00-kat (changes requested):
Just some small™ things about the documentation (the Zig looks fine to me).

po/com.mitchellh.ghostty.pot
pluiedev: Those are "message flags" and aren't really that important, since they just tell you whether a message uses fuzzy matching, is a C format string, etc

po/README_CONTRIBUTORS.md
pluiedev: There's an extra `msgctxt` line: https://www.gnu.org/software/gettext/manual/gettext.html#The-Format-of-PO-Files

00-kat: Should that perhaps be added to the example?

src/apprt/gtk/i18n.zig
pluiedev: Precisely because it's already been brought up there :p

po/README_TRANSLATORS.md
pluiedev: All translations are apprt-specific and are completely optional from a core perspective

po/README_TRANSLATORS.md
```diff
@@ -0,0 +1,148 @@
...
+There is NO WARRANTY, to the extent permitted by law.
+Written by Ulrich Drepper.
+```
+
+With this, you're ready to localize.
+
+## Locating translation files
+
+All translation files lie in the `po/` directory, including the main _template_
+file called `com.mitchellh.ghostty.pot`. **You may not edit this file.** The
```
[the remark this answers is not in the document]
00-kat: I suggested `must` as that is what you used in most other bolded imperatives.  `should` also works, though I assume you mean to say “it's fine as is” by your remark.

po/com.mitchellh.ghostty.pot
pluiedev: I'm really not sure how to fix this given that xgettext doesn't offer a command line parameter to change it, but I'll figure something out

po/README_CONTRIBUTORS.md
pluiedev: Probably

po/README_TRANSLATORS.md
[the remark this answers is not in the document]
Uzaaft: My five cents: `must` is fine. Or just straight up say: Dont edit this file.

po/com.mitchellh.ghostty.pot
pluiedev: I give up, it's literally impossible lol. Just leave it be

00-kat (approved):

trag1c (approved):

mitchellh (approved):

PR #6007, opened 2025-02-26
apprt/gtk: subscribe to AdwStyleManager::dark for ColorScheme

tristan957:


tristan957:
I couldn't figure out how to get `zig-gir` variant to compile. Got a compilation error that `notify` wasn't a signal on adw.StyleManager

tristan957:
Oops I get it

jcollie:
> Oops I get it. `notify` is on `gobject.Object`. But still, the app has a `c.GtkApplication` pointer, so I'm not sure it is possible to convert that to an `adw.Application` pointer.

Yeah until we're fully switched over you have to manually cast `c.GtkApplication` to `gtk.Application`, then you can use `app.as(gobject.Object)`.

tristan957:
Good question. Just if it changes. Seems like I need to do the equivalent of the dbus code. One sec.

tristan957:
> > Oops I get it. `notify` is on `gobject.Object`. But still, the app has a `c.GtkApplication` pointer, so I'm not sure it is possible to convert that to an `adw.Application` pointer.
> 
> Yeah until we're fully switched over you have to manually cast `c.GtkApplication` to `gtk.Application`, then you can use `app.as(gobject.Object)`.

Would you like me to use zig-gir code in this PR?

jcollie:
> Would you like me to use zig-gir code in this PR?

I think that we probably should.

src/apprt/gtk/App.zig
```diff
@@ -1256,9 +1257,22 @@ pub fn run(self: *App) !void {
...
     } else log.debug("cgroup isolation disabled config={}", .{self.config.@"linux-cgroup"});
 
-    // Setup our D-Bus connection for listening to settings changes,
-    // and asynchronously request the initial color scheme
-    self.initDbus();
+    // Setup color scheme notifications
+    const adw_app: *adw.Application = @ptrCast(@alignCast(self.app));
+    const style_manager: *adw.StyleManager = adw_app.getStyleManager();
+    _ = gobject.Object.signals.notify.connect(
+        style_manager.as(gobject.Object),
```
jcollie, line 1264: If you skip the `.as` here, you should be able to avoid the cast in the `adwNotifyDark` as well.

src/apprt/gtk/App.zig
```diff
@@ -1578,163 +1541,16 @@ fn gtkWindowIsActive(
...
-
-        if (value.isOfType(variant_type) == 0) {
-            log.err("unexpected value type: {s}", .{value.getTypeString()});
-            return;
-        }
-    }
-
-    const color_scheme: apprt.ColorScheme = if (value.getUint32() == 1)
-        .dark
+    const style_manager = gobject.ext.cast(adw.StyleManager, object).?;
```
jcollie, line 1549: This cast can be avoided I think by not casting when setting up the signal.

jcollie (changes requested):

jcollie (approved):

pluiedev (approved):

PR #6013, opened 2025-02-27
gtk: use language-neutral arrows for resize overlays

pluiedev:
The meaning of "c" and "r" can be somewhat cryptic to non-native English speakers as it may not be immediately obvious that "c" stands for "columns", and "r" stands for "rows". I propose replacing them with left-right and up-down double-headed arrows that convey the same meaning, but in a truly language-neutral manner.

Related to #2357

### DIFF
--- a/src/apprt/gtk/ResizeOverlay.zig
+++ b/src/apprt/gtk/ResizeOverlay.zig
@@ -114,7 +114,7 @@ fn gtkUpdate(ud: ?*anyopaque) callconv(.C) c_int {
     var buf: [32]u8 = undefined;
     const text = std.fmt.bufPrintZ(
         &buf,
-        "{d}c ⨯ {d}r",
+        "{d} ↔ ⨯ {d} ↕",
         .{
             grid_size.columns,
             grid_size.rows,


jcollie (approved):
↔ and ↕ are supported in both Cantarell and the new Adwaita fonts so any issues displaying these are on the user for choosing something other than the Gnome defaults.

pluiedev:
> but that's extra moments spent thinking about it when we can just include the units and be unambiguous about it.

This exactly. From an admittedly non-native perspective it always takes me a moment to register what "c" and "r" even mean, and just numbers would make that even worse 😅

tristan957:
I typically expect GNOME designers to know what they are doing :shrug:. GNOME Console also has no units

pluiedev:
Who knows? Maybe other designers would think of this as a better idea. We were already deviating from the norm per se, and nobody seems to have complained thus far

tristan957:
Internally I was complaining :laughing:

tristan957:
I do think this is an improvement over the previous version due to the mentioned language barrier.

pantheraleo-7:
and for macOS?

00-kat:
Two things:
1. These arrows have actually been removed in #6040.
2. Don't bump this PR, [open a new Discussion] instead.

[open a new Discussion]: https://github.com/ghostty-org/ghostty/discussions/new?category=feature-requests-ideas

PR #6024, opened 2025-02-27
Fix Terminal Inspector option turns inactive if toggled in the Quick Terminal

### LINKED ISSUE #2475: macOS: Terminal Inspector option turns inactive if toggled in the Quick Terminal
If I open the Terminal Inspector in the Quick Terminal, the Terminal Inspector option in the `View` menu becomes inactive.
I think this happens because the the focus will be set to the Inspector, if I click in the shell's area, or toggle the Quick Terminal on/off, the option will be active again, and the Inspector can be closed.
But I tried the same in a regular Ghostty window,  and there the Inspector option didn't turn inactive.

**Steps to reproduce:**

* Toggle Quick Terminal (`View > Quick Terminal`)
* Toggle the Terminal Inspector (`View > Terminal Inspector`)

**Ghostty version:**

Built locally from the `c2cbc214d5f6` commit.

CrazyMillerGitHub:
Fixed: [2475](https://github.com/ghostty-org/ghostty/issues/2475)

The problem actually existed because of the responder chain, as previously pointed out in the report (thanks to @mitchellh).

When we first click on Toggle Terminal Inspector:
* the responder chain goes to _toggleTerminalInspector_ (_SurfaceView_AppKit_ implementation). 
When we click the second toggleTerminalInspector:
* it tries to find the next responder, but the one available is _TerminalController_. (if we remove this method from there, the bug will reproduce even without quick mode)

**Problem**:
TerminalController not available during quick terminal, so there's no way to toggle inspector
**Solution**:
We add toggleTerminalInspector to the _QuickTerminalController_: selector, as we did with other similar methods.

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -459,6 +459,11 @@ class QuickTerminalController: BaseTerminalController {
         ghostty.toggleFullscreen(surface: surface)
     }
 
+    @IBAction func toggleTerminalInspector(_ sender: Any?) {
+        guard let surface = focusedSurface?.surface else { return }
+        ghostty.toggleTerminalInspector(surface: surface)
+    }
+
     // MARK: Notifications
 
     @objc private func applicationWillTerminate(_ notification: Notification) {


mitchellh (approved):
Approved, thanks. 

I wonder if we can actually just remove the one in `SurfaceView` now... but can do that later.

PR #6032, opened 2025-02-28
GTK: do not check for terminal area focus when setting window title

### LINKED ISSUE #5940: Custom title is not restored when tab overview is closed via creating a new tab
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5875

<sup>Originally posted by **gzusseat** February 19, 2025</sup>
Creating a new tab from the overview sets the title to Ghostty even though i have title = "  ".  If I click on an unfocused tab it disappears again. bug?</div>

b6k-dev:
Fixes https://github.com/ghostty-org/ghostty/issues/5940
The mentioned problem occurs because when creating a new tab through the tab overview we do not have focus on the terminal area widget, it does not matter if we have a custom title set or not.
I think it is just safe to remove this check from the code. I've tested the change and I don't really see a valid use case in which we would not want to set the window title even if we don't have focus on the terminal area.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -653,6 +653,10 @@ pub fn focusCurrentTab(self: *Window) void {
     const surface = tab.focus_child orelse return;
     const gl_area = @as(*c.GtkWidget, @ptrCast(surface.gl_area));
     _ = c.gtk_widget_grab_focus(gl_area);
+
+    if (surface.getTitle()) |title| {
+        self.setTitle(title);
+    }
 }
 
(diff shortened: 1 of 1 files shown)


pluiedev (changes requested):
Actually, I think a *far* easier fix would be to also update the title  in `Window.focusCurrentTab`, which is called by `AdwTabView::page-attached`, docs for which explicitly mention that that is the signal to listen for for title updates

jcollie:
Please rebase this to eliminate the "patch/revert patch" pair.

pluiedev (approved):

PR #6033, opened 2025-02-28
Add note to configuration that some settings might require a restart

geekgonecrazy:
This adds a note in the default config to note that may require a restart.  Also adds a note with the window-padding-x and window-padding-y to indicate a restart will be required to update existing windows.

I ran into this while updating one of the existing values in the default config file. `window-padding-x` The value defaulted to 2 so figured it was safe to assume and just uncomment it and try reloading the config.  

Doing that doesn't work only restarting will make it take effect for the main window ( or of course more tricky opening new windows and killing off the old one )

https://github.com/ghostty-org/ghostty/discussions/6022

### DIFF
--- a/src/config/config-template
+++ b/src/config/config-template
@@ -14,6 +14,12 @@
 #
 # Additionally, each config option is also explained in detail
 # on Ghostty's website, at https://ghostty.org/docs/config.
+#
+# Ghostty can reload the configuration while running by using the menu
+# options or the bound key (default: Command + Shift + comma on macOS and
+# Control + Shift + comma on other platforms). Not all config options can be
+# reloaded while running; some only apply to new windows and others may require
+# a full restart to take effect.
 
 # Config syntax crash course
 # ==========================


src/config/Config.zig
```diff
@@ -1086,7 +1086,7 @@ keybind: Keybinds = .{},
 /// will appear in the logs.
 ///
 /// Changing this configuration at runtime will only affect new terminals, i.e.
-/// new windows, tabs, etc.
+/// new windows, tabs, etc. A restart is necessary for existing windows.
```
mitchellh, line 1089: Doesn't the previous sentence imply this? You don't need to restart Ghostty, you only need to recreate the windows.

src/config/config-template
```diff
@@ -12,6 +12,8 @@
 # Run `ghostty +show-config --default --docs` to view a list of
 # all available config options and their default values.
 #
+# Some options may require a restart of ghostty in order to take effect
```
jcollie, line 15: Needs a `.` at the end.

jcollie (changes requested):

src/config/config-template
pluiedev: And proper capitalization of "Ghostty"

mitchellh:
I reworded it myself, thanks.

mitchellh (approved):

src/config/config-template
```diff
@@ -14,6 +14,12 @@
 #
 # Additionally, each config option is also explained in detail
 # on Ghostty's website, at https://ghostty.org/docs/config.
+#
+# Ghostty can reload the configuration while running by using the menu
+# options or the bound key (default: Super + Shift + comma. Super is
```
pluiedev, line 19: Isn't it Ctrl on Linux, not Super?

mitchellh: Oh, I overrode it. yes.

pluiedev (approved):

PR #6034, opened 2025-02-28
apprt initial_size is sent whenever the grid size changes

mitchellh:
As noted in the comments, this is so that apprt's can always know what the default size of a window would be so they can utilize this for "return to default size" actions.

The initial size shouldn't be treated as a "resize" event and was already documented as such. Prior to this commit the docs already noted that the initial size may be sent multiple times but only the first time during initialization should be used as a resize.

Therefore, this shouldn't impact prior behavior. I've verified this with the apprts.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -44,6 +44,13 @@ const log = std.log.scoped(.surface);
 // The renderer implementation to use.
 const Renderer = renderer.Renderer;
 
+/// Minimum window size in cells. This is used to prevent the window from
+/// being resized to a size that is too small to be useful. These defaults
+/// are chosen to match the default size of Mac's Terminal.app, but is
+/// otherwise somewhat arbitrary.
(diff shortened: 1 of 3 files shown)


mitchellh:
This has an interesting side effect in GTK that if you've never resized the window, this will change the window size. 😄 I'm looking into that, it shouldn't have behaved that way from the get-go but we should fix that before merging ths.

mitchellh:
For more background on "why": this lets apprts always know what the "default size" would be (if any). This information otherwise was unavailable making it impossible to implement the already accepted #1328 without reinventing the wheel quite a bit. 

I don't know of other use cases yet but the whole point of apprt actions is to inform apprts of information _in case_ its useful, so I think in the interest of utility, this is still useful beyond the use case above.

jcollie (approved):

pluiedev (approved):

Issue #6035, opened 2025-02-28
GTK: add support for "return to default size"


jcollie:
See #5974 for inspiration from the macOS apprt.

mitchellh:
#5974 finishes this for macOS and will close the associated issue, but this was done purely in apprt. 

I want to add a bit of scope to this GTK issue: let's also write a keybind action and apprt action for this that appts can listen for. I might do this in a separate PR ASAP but want to ensure that is done too. 😄

jcollie:
Well, this isn't _impossible_ but it's very very ugly (and probably breaks things in ways that I didn't test). The only way that I got this to "work" was to set the default size on the widget, then hide the window and quickly re-show the window. That resized the window, but it also _repositioned_ the window. The hide/show was also noticeable, likely due to animations that GTK uses.

So this is something that GTK users will just have to do without I think. Unless they are also using the macOS version I don't think they'll notice as this does not seem to be a common UI on any Linux app that I can think of.

PR #6038, opened 2025-02-28
Introduce `reset_window_size` keybinding and apprt action

mitchellh:
Related to #6035

This implements the keybind/action portion of #5974 so that this can have a binding and so that other apprts can respond to this and implement it this way.


include/ghostty.h
```diff
@@ -580,6 +580,7 @@ typedef enum {
   GHOSTTY_ACTION_TOGGLE_SPLIT_ZOOM,
   GHOSTTY_ACTION_PRESENT_TERMINAL,
   GHOSTTY_ACTION_SIZE_LIMIT,
+  GHOSTTY_ACTION_RESET_WINDOW_SIZE,
```
mitchellh, line 583: Note that our keybind and action do not match the macOS menu item because macOS has very specific terminology it uses "Return to Default Size" within a very specific menu item (Window) that I don't think translates well verbatim to a general cross platform term. 

So, I use the macOS terminology/location for macOS, but choose `reset_window_size` as a more descriptive term for the core actions.

jcollie (approved):

PR #6040, opened 2025-03-01
apprt/gtk: remove non-ascii characters from resize overlay

tristan957:
It is possible that fonts people are using don't contain these characters as evidenced by users in the Discord.


tristan957:
I haven't seen anyone complain about the unicode x, so if we still want that I'm down, but I struggle to see a real point honestly.

00-kat:
> I haven't seen anyone complain about the unicode x

If I remember correctly, I've seen two on Discord and one on GitHub.  It ends up showing as a replacement character.  I couldn't find any of them when I tried looking recently though.

00-kat (approved):

jcollie (approved):
Well I don't like the fact that we lose units of any kind. (With deference to the GNOME HIG people, _they're wrong_).

PR #6052, opened 2025-03-02
core!: modify scroll behavior

rockorager:
Modify the scroll behavior to better match other terminals, as well as provide a
better overall experience. Before this PR, ghostty would scale non-precision
scroll events dependent on the screen size. This is in line with kitty, but no
other terminal. Ghostty also was the only terminal to send *more than one* wheel
event.

```
 # 50% Screen height
| terminal   | arrows keys sent| wheels events sent|
|------------|-----------------|-------------------|
| alacritty  |        3        |          1        |
| foot       |        3        |          1        |
| xterm      |        5        |          1        |
| kitty      |        3        |          1        |
| ghostty    |        2        |          2        |

 # 100% Screen height
| terminal   | arrows keys sent| wheels events sent|
|------------|-----------------|-------------------|
| alacritty  |        3        |          1        |
| foot       |        3        |          1        |
| xterm      |        5        |          1        |
| kitty      |        5        |          1        |
| ghostty    |        3        |          3        |
```

This PR modifies Ghostty to behave like foot and alacritty.

For an improved user experience, we only use the configured multiplier for
non-precision scrolls. This multiplier now *only applies* to viewport scrolling
and alternate scroll mode. The default value has been updated to 3.0.

GTK also now supports precision scrolling.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -304,6 +304,9 @@ cgroup_path: ?[]const u8 = null,
 /// Our context menu.
 context_menu: Menu(Surface, "context_menu", false),
 
+/// True when we have a precision scroll in progress
+precision_scroll: bool = false,
+
 /// The state of the key event while we're doing IM composition.
 /// See gtkKeyPressed for detailed descriptions.
 pub const IMKeyEvent = enum {
@@ -418,10 +421,7 @@ pub fn init(self: *Surface, app: *App, opts: Options) !void {
     c.gtk_widget_add_controller(@ptrCast(@alignCast(overlay)), ec_motion);
 
     // Scroll events
-    const ec_scroll = c.gtk_event_controller_scroll_new(
-        c.GTK_EVENT_CONTROLLER_SCROLL_BOTH_AXES |
-            c.GTK_EVENT_CONTROLLER_SCROLL_DISCRETE,
-    );
+    const ec_scroll = c.gtk_event_controller_scroll_new(c.GTK_EVENT_CONTROLLER_SCROLL_BOTH_AXES);
     errdefer c.g_object_unref(ec_scroll);
     c.gtk_widget_add_controller(@ptrCast(overlay), ec_scroll);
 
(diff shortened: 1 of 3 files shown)


rockorager:
Note: I don't have a mac and was not able to test this behavior there. This *shouldn't* have changed the default behavior on mac, but it *will* change the behavior for users who have set a scroll multiplier.

src/apprt/gtk/Surface.zig
```diff
@@ -1533,15 +1551,17 @@ fn gtkMouseScroll(
     const scaled = self.scaledCoordinates(x, y);
 
     // GTK doesn't support any of the scroll mods.
-    const scroll_mods: input.ScrollMods = .{};
+    const scroll_mods: input.ScrollMods = .{ .precision = self.precision_scroll };
+    // Multiply precision scrolls by 10 to get a better response from touchpad scrolling
+    const multiplier: f64 = if (self.precision_scroll) 10 else 1;
```
rockorager, line 1556: I used 10, but it does feel a little high to me. I'm not sure what other people think for scroll behavior here.

src/config/Config.zig
```diff
@@ -609,10 +609,8 @@ palette: Palette = .{},
 /// than 0.01 or greater than 10,000 will be clamped to the nearest valid
 /// value.
 ///
-/// A value of "1" (default) scrolls the default amount. A value of "2" scrolls
-/// double the default amount. A value of "0.5" scrolls half the default amount.
-/// Et cetera.
-@"mouse-scroll-multiplier": f64 = 1.0,
+/// A value of "3" (default) scrolls 3 lines per tick.
+@"mouse-scroll-multiplier": f64 = 3.0,
```
rockorager, line 613: I think we could change this to a usize, or even a u8. Even the original comment says this only applies to wheel mice, so changing the meaning to be "lines per tick" makes sense to me - and this should only ever be an integer.

jcollie: Since the default is now `3` I think that changing to an integer make sense, you're still able to slow it down if you want to.

jcollie (approved):

src/config/Config.zig
mitchellh: I think the impact of this multiplier is meant to affect precision scrolling as well, which comes in at fractional deltas. So I think this should be float. We can continue this discussion though no need to block the PR on it. I know this PR removes that behavior but I'm not sure if we want that to stick or not, I recall during the beta some people wanting to change the momentum scrolling on macOS to do more or less.

mitchellh (approved):
Feels good on macOS too.

I think modifying the scroll behavior plus introducing the GTK precision scrolling in one PR was risky but it paid off ;)

rockorager:
> I think modifying the scroll behavior plus introducing the GTK precision
> scrolling in one PR was risky but it paid off ;)

I needed it for testing so I just went ahead and did it :). I agree it probably
should have been a second PR...

> > @@ -609,10 +609,8 @@ palette: Palette = .{},
>  /// than 0.01 or greater than 10,000 will be clamped to the nearest valid
>  /// value.
>  ///
> -/// A value of "1" (default) scrolls the default amount. A value of "2" scrolls
> -/// double the default amount. A value of "0.5" scrolls half the default amount.
> -/// Et cetera.
> -@"mouse-scroll-multiplier": f64 = 1.0,
> +/// A value of "3" (default) scrolls 3 lines per tick.
> +@"mouse-scroll-multiplier": f64 = 3.0,
> 
> I think the impact of this multiplier is meant to affect precision scrolling
> as well, which comes in at fractional deltas. So I think this should be float.
> We can continue this discussion though no need to block the PR on it. I know
> this PR removes that behavior but I'm not sure if we want that to stick or
> not, I recall during the beta some people wanting to change the momentum
> scrolling on macOS to do more or less.

You are correct - it *was* used for precision scrolling as well. I think adding
a new configuration option as a precision scroll multiplier would be good.

remorses:
Why set the mouse scroll multiplier to 3? All other terminals use a value of 1

PR #6054, opened 2025-03-02
gtk: implement sensitive content reveal on paste confirmation

### LINKED ISSUE #4947: Unsafe paste confirmation should not reveal contents in UI by default in secure input mode
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4873

<sup>Originally posted by **cryptocode** January  9, 2025</sup>
In the image below (where I've cut off the password) I pasted a password with special characters for an ssh session, while in secure input "padlock mode" 

The macOS secure input and padlock is a great feature that doesn't echo the password, so it's a bit of a bummer that the paste warning shows the password in clear text.

I feel like the warning should be suppressed, or at least not show the pasted value when in padlock mode.

image

</div>

### Solution

We should send information along with the paste confirmation request of whether the paste contents are considered sensitive or part of secure input. If so, our UIs should hide the contents by default.

b6k-dev:
Fixes https://github.com/ghostty-org/ghostty/issues/4947 for gtk
This PR implements the senstive content hiding when displaying the paste confirmation dialog in secure input mode.

Following changes are implemented:
- in the blueprint for each dialog add a show/hide button that is not visible by default, and a Revealer that is revealed by default
- save the `secure_input` action value for each surface in the GTK apprt
- pass the value when initializing the paste confirmation dialog
- in the dialog code, alter the visibility of the content and reveal/hide buttons based on secure input flag value

Demo:

https://github.com/user-attachments/assets/c91cbd3d-ed3b-464d-b4cf-e51fe7aa23b7

I feel like this is already a nearly full implementation, but I'm leaving this as a draft for now, since i need to look into blueprints for Adwaita 1.2, and verify if it behaves properly when the dialog is in not-sensitive input mode and in OSC52 mode.

### DIFF
--- a/src/apprt/gtk/ClipboardConfirmationWindow.zig
+++ b/src/apprt/gtk/ClipboardConfirmationWindow.zig
@@ -27,8 +27,8 @@ core_surface: *CoreSurface,
 pending_req: apprt.ClipboardRequest,
 text_view: *gtk.TextView,
 text_view_scroll: *gtk.ScrolledWindow,
-reveal_button_widget: *gtk.Widget,
-hide_button_widget: *gtk.Widget,
+reveal_button: *gtk.Button,
+hide_button: *gtk.Button,
 
 pub fn create(
     app: *App,
@@ -101,8 +101,8 @@ fn init(
         .pending_req = request,
         .text_view = text_view,
         .text_view_scroll = text_view_scroll,
-        .reveal_button_widget = gobject.ext.cast(gtk.Widget, reveal_button).?,
-        .hide_button_widget = gobject.ext.cast(gtk.Widget, hide_button).?,
+        .reveal_button = reveal_button,
+        .hide_button = hide_button,
     };
 
     const buffer = gtk.TextBuffer.new(null);
@@ -111,10 +111,10 @@ fn init(
     text_view.setBuffer(buffer);
 
     if (is_secure_input) {
-        gobject.ext.as(gtk.Widget, text_view_scroll).setSensitive(@intFromBool(false));
-        gobject.ext.as(gtk.Widget, self.text_view).addCssClass("blurred");
+        text_view_scroll.as(gtk.Widget).setSensitive(@intFromBool(false));
+        self.text_view.as(gtk.Widget).addCssClass("blurred");
 
-        self.reveal_button_widget.setVisible(@intFromBool(true));
+        self.reveal_button.as(gtk.Widget).setVisible(@intFromBool(true));
 
         _ = gtk.Button.signals.clicked.connect(
             reveal_button,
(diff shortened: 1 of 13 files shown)


src/apprt/gtk/ui/1.5/ccw-osc-52-read.blp
```diff
@@ -14,18 +14,46 @@ Adw.AlertDialog clipboard_confirmation_window {
...
+        icon-name: "dialog-warning-symbolic";
+      }
+    }
+
+    Button hide_button {
+      visible: false;
+      label: _("Hide Potentially Sensitive Content");
+    }
+
+    Revealer content_revealer {
```
jcollie, line 39: I would prefer an overlay that blocks/blurs the content rather than a revealer. I dislike how the window size jumps around when you reveal/hide the content.

src/apprt/gtk/Surface.zig
```diff
@@ -1203,12 +1207,7 @@ fn gtkClipboardRead(
         error.UnauthorizedPaste,
         => {
             // Create a dialog and ask the user if they want to paste anyway.
-            ClipboardConfirmationWindow.create(
-                self.app,
-                str,
-                &self.core_surface,
-                req.state,
-            ) catch |window_err| {
+            ClipboardConfirmationWindow.create(self.app, str, &self.core_surface, req.state, self.is_secure_input) catch |window_err| {
```
jcollie, line 1210: Make sure to keep the trailing `,` so that `zig fmt` wraps this line.

src/apprt/gtk/Surface.zig
```diff
@@ -2205,12 +2204,7 @@ fn doPaste(self: *Surface, data: [:0]const u8) void {
         error.UnsafePaste,
         error.UnauthorizedPaste,
         => {
-            ClipboardConfirmationWindow.create(
-                self.app,
-                data,
-                &self.core_surface,
-                .paste,
-            ) catch |window_err| {
+            ClipboardConfirmationWindow.create(self.app, data, &self.core_surface, .paste, self.is_secure_input) catch |window_err| {
```
jcollie, line 2207: Make sure to keep the trailing `,` so that `zig fmt` wraps this line.

src/apprt/gtk/ClipboardConfirmationWindow.zig
```diff
@@ -25,12 +25,16 @@ dialog: *DialogType,
 data: [:0]u8,
 core_surface: *CoreSurface,
 pending_req: apprt.ClipboardRequest,
+content_revealer: *gtk.Revealer,
+reveal_button_widget: *gtk.Widget,
+hide_button_widget: *gtk.Widget,
```
jcollie, line 30: The buttons should saved as `*gtk.Button` and cast as needed later.

jcollie (changes requested):
As noted in the start of the PR, need to get this working for Adwaita 1.2.

src/apprt/gtk/ui/1.5/ccw-osc-52-read.blp
maciekbartczak: Switched to the blur approach, I agree it looks much better when the window does not jump around.

https://github.com/user-attachments/assets/40d6d1cd-bcd4-4892-aa1e-030b28226ae4

src/apprt/gtk/ClipboardConfirmationWindow.zig
```diff
@@ -88,15 +99,39 @@ fn init(
...
-    const text_view = builder.getObject(gtk.TextView, "text_view").?;
-
     const buffer = gtk.TextBuffer.new(null);
     errdefer buffer.unref();
     buffer.insertAtCursor(copy.ptr, @intCast(copy.len));
     text_view.setBuffer(buffer);
 
+    if (is_secure_input) {
+        gobject.ext.as(gtk.Widget, text_view_scroll).setSensitive(@intFromBool(false));
+        gobject.ext.as(gtk.Widget, self.text_view).addCssClass("blurred");
```
pluiedev, line 115: You can write as in postfix position: `foo.as(gtk.Bar)`

pluiedev (changes requested):
From a UI design perspective I'm not sure if a background blur is idiomatic for GTK4 apps, although I'm not opposed to it. A separate row button is definitely not it, though... maybe an overlaid button in the corner could be used to reenable the overlay? I'm not sure why anyone would do this anyway.

tristan957:
The blur is sweet

jcollie:
I like the blur as well, but I agree that a separate button isn't the right UI. Generally clicking on the blurred panel is what I see for UI.

jcollie:
> How about something like this?

Looks good!

> I'm not really sure what to think about the hide button, it feels out of place in there, maybe it is actually not needed at all?

Yeah, I'd just get rid of the hide button, and let clicking on the text view itself hide the content again.

tristan957:
What about a button similar to a password entry. Just and :eye: button overlaid in the top right corner (or top left corner) in LTR languages?

pluiedev:
> Just an :eye: button overlaid in the top right corner 

This

maciekbartczak:
Ok, that looks perfect to me, but I'm open to any suggestions :smile: 

https://github.com/user-attachments/assets/3c63765d-edad-4d94-900f-295e730d9282

jcollie:
> Ok, that looks perfect to me, but I'm open to any suggestions 😄

Looks good!

src/apprt/gtk/ui/1.5/ccw-paste.blp
```diff
@@ -10,22 +10,62 @@ Adw.AlertDialog clipboard_confirmation_window {
...
+    Button reveal_button {
+      visible: false;
+      halign: end;
+      valign: start;
+      margin-end: 12;
+      margin-top: 12;
+
+      styles [
+        "destructive-action",
+      ]
```
tristan957, line 47: I would not use suggested or destructive action styles for the overlay. Just the icon in regular colors will be fine. The dialog already has suggested and destructive coloring. Seems a little overwhelming

maciekbartczak:
Ok, that should be the final version. I've Added the implemenation for Adwaita 1.2, addressed the other issues and resolved merge conflicts.

jcollie (approved):

pluiedev (approved):

jcollie:
Thanks @maciekbartczak for putting up with all our nitpicks, great work!

PR #6057, opened 2025-03-02
input: legacy encoding never encodes text for command mods on macOS

### LINKED ISSUE #5929: macOS: Cmd+H should not send "h" if its the last visible window

mitchellh:
Fixes #5929
Replaces #5984

On macOS, native applications typically never encode any text for key events that use the command key. This is because the command key is used for key equivalents and "commands" and should not be used for text input.

This can be verified with apps like TextEdit but also terminals like Terminal.app officially but also iTerm2 unofficially. Anything such as `Cmd+b` or `Cmd+Shift+b` will not produce any text input.

Cross-platform terminals generally don't follow this, for example Kitty performs CSI-u encoding and Alacritty and WezTerm encode the text as-is (i.e. `Cmd+b` will produce `b`).

On Linux, the super key (command-equivalent) does produce text input. For example, `Super+b` will produce `b` in Gnome Console, Foot, and all the cross-platform terminals mentioned above.

In the interest of matching the behavior of native macOS applications, we should not encode text for command key events on macOS. We continue to encode text for the super key on non-macOS platforms. This matches the behaviors appropriately on each platform.

### DIFF
--- a/src/input/KeyEncoder.zig
+++ b/src/input/KeyEncoder.zig
@@ -35,6 +35,7 @@ pub fn encode(
     self: *const KeyEncoder,
     buf: []u8,
 ) ![]const u8 {
+    // log.warn("KEYENCODER self={}", .{self.j});
     if (self.kitty_flags.int() != 0) return try self.kitty(buf);
     return try self.legacy(buf);
 }
@@ -411,6 +412,19 @@ fn legacy(
     // we need to prefix the utf8 sequence with an esc.
     if (try self.legacyAltPrefix(binding_mods, all_mods)) |byte| {
         return try std.fmt.bufPrint(buf, "\x1B{c}", .{byte});
+    }
+
+    // If we are on macOS, command+keys do not encode text. It isn't
+    // typical for command+keys on macOS to ever encode text. They
+    // don't in native text inputs (i.e. TextEdit) and they also don't
+    // in other native terminals (Terminal.app officially but also
+    // iTerm2).
+    //
+    // For Linux, we continue to encode text because it is typical.
+    // For example on Gnome Console Super+b will encode a "b" character
+    // with legacy encoding.
+    if ((comptime builtin.os.tag == .macos) and all_mods.super) {
+        return "";
     }
 
     return try copyToBuf(buf, utf8);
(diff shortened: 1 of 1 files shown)


src/input/KeyEncoder.zig
```diff
@@ -413,6 +414,19 @@ fn legacy(
...
+    // don't in native text inputs (i.e. TextEdit) and they also don't
+    // in other native terminals (Terminal.app officially but also
+    // iTerm2).
+    //
+    // For Linux, we continue to encode text because it is typical.
+    // For example on Gnome Console Super+b will encode a "b" character
+    // with legacy encoding.
+    if ((comptime builtin.os.tag == .macos) and all_mods.super) {
+        return "";
+    }
```
mitchellh, line 428: I will note that it is possible that this check is too late. We may want to omit ctrl sequence and other encodings as well when the super key is pressed on macOS. But the tests only cover this and this was the least-impacting change I could do with a very specific real world bug to verify it resolves. 

I'd be open to a broader change with a bit more research.

rockorager:
Just so I understand... kkp encoding will still encode the associated text field when super + "b" is pressed? I am thinking we *shouldn't* do that and instead report it as no produced text...what do you think?

mitchellh:
> Just so I understand... kkp encoding will still encode the associated text field when super + "b" is pressed? I am thinking we shouldn't do that and instead report it as no produced text...what do you think?

I can confirm it does not produce text, and the behavior matches Kitty.

rockorager (approved):

PR #6069, opened 2025-03-03
gtk: use updated blueprint compiler for needed new features

jcollie:
This should be reverted once NixOS stable has a new enough version. Note that this won't yet help those building on non-Nix systems but it should help unblock development.

### DIFF
--- a/nix/blueprint-compiler.nix
+++ b/nix/blueprint-compiler.nix
@@ -1,0 +1,83 @@
+{
+  dbus,
+  fetchFromGitLab,
+  gobject-introspection,
+  lib,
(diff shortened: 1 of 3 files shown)


mitchellh:
I'd prefer to just wait for `nixpkgs-unstable` to have what we need and overlay that rather than maintain a derivation fork ourselves. Let's stick with 0.14 for now, and I think 0.16 was already in PR yeah?

jcollie:
OK, looks like the upstream fix might happen more quickly than I thought:

https://github.com/NixOS/nixpkgs/pull/386862

PR #6080, opened 2025-03-04
Make `equalize_splits` action only affect current window

### LINKED ISSUE #6064: macOS: `equalize_splits` action applies to all windows
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6060

<sup>Originally posted by **taylrfnt** March  2, 2025</sup>
Porting from Discord...

---
The [documentation for the `equalize_splits` action](https://ghostty.org/docs/config/keybind/reference#equalize_splits) says that this action would affect the current window (quote below, emphasis mine):
> Equalize all splits in the **current** window

However, on macOS, the `equalize_splits` option seems to equalize splits in ALL surfaces, as shown in the recording below.
equalize_split_repro

---
All the other split actions (e.g. `resize_split`) work on the currently focused window, only the `equalize_split` action seems to apply to all windows/surfaces.  My only thought is perhaps this is due to the macOS version running as a single app instance, but I could not rationalize this while `resize_split` works on for a given focused surface.

--- 
This was reproduced on `tip`, and can be reproduced with the default settings & the steps below:
1. Open Ghostty on macOS.
2. Create a new window (<kbd>cmd</kbd>+<kbd>n</kbd>).
3. Open splits in both windows.
4. Resize the splits such that they are not equalized.
5. Use the `equalize_splits` action (<kbd>cmd</kbd>+<kbd>ctrl</kbd>+<kbd>=</kbd>) and observe all splits are equalized, not just the currently focused window.</div>

liby:
Fixes #6064 

https://github.com/user-attachments/assets/bbf393be-de98-41eb-aaad-3a185705ed4c

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -85,6 +85,12 @@ class TerminalController: BaseTerminalController {
             selector: #selector(onFrameDidChange),
             name: NSView.frameDidChangeNotification,
             object: nil)
+        center.addObserver(
+            self,
+            selector: #selector(onEqualizeSplits),
+            name: Ghostty.Notification.didEqualizeSplits,
+            object: nil
+        )
     }
 
     required init?(coder: NSCoder) {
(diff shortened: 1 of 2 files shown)


mitchellh (approved):
Perfect thanks.

PR #6090, opened 2025-03-05
gtk: implement quick terminal slide & autohide

pluiedev:
Yet another protocol that as far as I know only KWin implements. Oh well, might as well let KDE users such as myself enjoy it OOTB


jcollie (approved):
Didn't test but the code looks sane.

PR #6601, opened 2025-03-06
translations: add German translation

rpfaeffle:
This PR adds a translation for German `de_DE`.

Additionally it excludes all `*.po` files from the typos CI action.

Some comments on the decisions I made (open to discuss them):

- I choosed to use `du` instead of `Sie` as this seems appropriate
  to me.
- I added `Window` (`Fenster` in German) to all split commands
  as it appears more naturally to me.

### DIFF
--- a/po/de_DE.UTF-8.po
+++ b/po/de_DE.UTF-8.po
@@ -24,7 +24,7 @@ msgstr "Terminal-Titel bearbeiten"
 
 #: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:6
 msgid "Leave blank to restore the default title."
-msgstr "Leer lassen, um den Standardtitel wiederherzustellen"
+msgstr "Leer lassen, um den Standardtitel wiederherzustellen."
 
 #: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:9
 #: src/apprt/gtk/ui/1.5/ccw-paste.blp:10
@@ -209,4 +209,4 @@ msgstr "Ghostty-Entwickler"
 
 #: src/apprt/gtk/Surface.zig:1236
 msgid "Copied to clipboard"
-msgstr "In die Zwischenablage kopiert."
+msgstr "In die Zwischenablage kopiert"
--- a/po/de_DE.UTF-8.po
+++ b/po/de_DE.UTF-8.po
@@ -197,7 +197,7 @@ msgstr "Offene Tabs einblenden"
 msgid ""
 "⚠️ You're running a debug build of Ghostty! Performance will be degraded."
 msgstr ""
-"  Du verwendest einen Debug Build von Ghostty! Die Leistung wird reduziert sein."
+"⚠️ Du verwendest einen Debug Build von Ghostty! Die Leistung wird reduziert sein."
 
 #: src/apprt/gtk/Window.zig:660
 msgid "Reloaded the configuration"
--- a/po/de_DE.UTF-8.po
+++ b/po/de_DE.UTF-8.po
@@ -1,0 +1,212 @@
+# German translations for com.mitchellh.ghostty package
+# German translation for com.mitchellh.ghostty.
+# Copyright (C) 2025 Mitchell Hashimoto
+# This file is distributed under the same license as the com.mitchellh.ghostty package.
+# Robin Pfäffle <r@rpfaeffle.com>, 2025.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: com.mitchellh.ghostty\n"
+"Report-Msgid-Bugs-To: m@mitchellh.com\n"
+"POT-Creation-Date: 2025-02-28 22:12+0100\n"
+"PO-Revision-Date: 2025-03-06 14:57+0100\n"
+"Last-Translator: Robin <r@rpfaeffle.com>\n"
+"Language-Team: German <translation-team-de@lists.sourceforge.net>\n"
+"Language: de\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:5
+msgid "Change Terminal Title"
+msgstr "Terminal-Titel bearbeiten"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:6
+msgid "Leave blank to restore the default title."
+msgstr "Leer lassen, um den Standardtitel wiederherzustellen"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:9
+#: src/apprt/gtk/ui/1.5/ccw-paste.blp:10
+msgid "Cancel"
+msgstr "Abbrechen"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:10
+msgid "OK"
+msgstr "OK"
+
+#: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:6
+#: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:6
+msgid "Copy"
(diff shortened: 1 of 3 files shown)


mitchellh:
I think we should block any translations without an associated maintainer. Thoughts @pluiedev? The expectation would be that they'd be part of a team that CODEOWNERS this file along with the template file so they're notified of any changes as well. 

If so, would you volunteer @rpfaeffle?

rpfaeffle:
I would be happy to volunteer (at least for `de_DE`).

jcollie:
> I think we should block any translations without an associated maintainer.

There really should be _multiple_ maintainers so that more than one person reviews any translations before they are committed.

pluiedev:
I agree that for a major language like German the maintainer count should be at least N = 2. For minor languages N = 1 could be fine, but we'll see.

It should be noted that new language submissions, IMO, should imply that the submitter agrees to be responsible for its maintainership for the long term. If someone doesn't want to take up that responsibility then they shouldn't contribute: no translations > crappy, outdated, unmaintained translations. I should probably add that to the translator's guide.

mitchellh:
> There really should be multiple maintainers so that more than one person reviews any translations before they are committed.

I agree but in the interest of bootstrapping, I'd also be comfortable for major languages that are relatively easily verifiable to fallback to verifying with something like Google Translate or even an LLM. I know they're not perfectly accurate (and sometime wildly so) but as a baseline "well, it's not a racial slur" check it should work well enough.

I agree I'd prefer N >= 2 though.

po/de_DE.UTF-8.po
```diff
@@ -0,0 +1,212 @@
...
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:5
+msgid "Change Terminal Title"
+msgstr "Terminal-Titel bearbeiten"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:6
+msgid "Leave blank to restore the default title."
+msgstr "Leer lassen, um den Standardtitel wiederherzustellen"
```
pluiedev, line 27:
```suggestion
msgstr "Leer lassen, um den Standardtitel wiederherzustellen."
```

pluiedev (changes requested):

pluiedev (changes requested):
Also please also commit the addition to src/build/GhosttyI18n.zig that makes the build system aware of the new locale

po/de_DE.UTF-8.po
```diff
@@ -0,0 +1,212 @@
...
+
+#: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:48
+#: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:60
+msgid "Split Left"
+msgstr "Fenter nach links teilen"
+
+#: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:53
+#: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:65
+msgid "Split Right"
+msgstr "Fenster nach rechts teilen"
```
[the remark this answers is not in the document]
rpfaeffle: I thought the same. For the context menu it would definitely be too long in my opinion as well. I agree with you that "Fenster teilen > Nach rechts" would be the best option .

cactysman:
I've been using Ghostty at work every day since it initially released and I've done translation work for other projects in the past. I'd be down to add to the list of locale maintainers / reviewers.

rpfaeffle:
It seems like the CI assumes some German words are typos. Should we ignore the whole file, as it could lead to false negatives if we would mark "Titel" as a correct spelling for instance?

Maybe it would be a good idea to ignore all the `.po` files, as most of them will probably lead to false positives.

pluiedev:
Looks like upstream recommends setting this for PO files:

```toml
[type.po]
extend-glob = ["*.po"]
check-file = false
```

pluiedev:
#5741 was just merged and this PR should rebase on top of that

rpfaeffle:
Rebased onto `main`.

mitchellh:
I apologize but my recent PR messed this up again, no new translations from me though. Rebase please 😄

rpfaeffle:
No worries :smile:. I rebased onto `main`.

po/de_DE.UTF-8.po
```diff
@@ -0,0 +1,253 @@
...
+
+#: src/apprt/gtk/CloseDialog.zig:47
+#, fuzzy
+msgid "Close"
+msgstr "Schließen"
+
+#: src/apprt/gtk/CloseDialog.zig:87
+#, fuzzy
+msgid "Quit Ghostty?"
+msgstr "Über Ghostty?"
```
pluiedev, line 219: These lines are automatically "guessed" by gettext (see the fuzzy flag) and haven't actually been translated

po/de_DE.UTF-8.po
```diff
@@ -0,0 +1,253 @@
...
+msgstr "Fenster schließen?"
+
+#: src/apprt/gtk/CloseDialog.zig:89
+#, fuzzy
+msgid "Close Tab?"
+msgstr "Tab schließen?"
+
+#: src/apprt/gtk/CloseDialog.zig:90
+msgid "Close Split?"
+msgstr "Geteiltes Fenster schließen?"
```
pluiedev, line 233: It's not exactly "divided window" and more like "part of a window" I think. Does "Fensterteil" make sense?

pluiedev (changes requested):

po/de_DE.UTF-8.po
rpfaeffle: I don't really think "Fensterteil" makes a lot of sense in this context. I checked again what `Terminal.app` is using and they are using _"Geteiltes Fenster schließen"_ as well.

pluiedev: Hmm.. fair enough then

cactysman: > I don't really think "Fensterteil" makes a lot of sense in this context. I checked again what `Terminal.app` is using and they are using _"Geteiltes Fenster schließen"_ as well.

I'm still kinda new to the Appleverse (2 years), but do you guys think it would be feasible to kinda stick to the general lingo of a platform (which would then in some way infer that Ghostty shipped differing translations per platform) to add to the overall user experience (so that on MacOS Ghostty would pretty much use the same kind of terminology as the rest of the system) or should Ghostty just find its own "dialect" in translations and use translations that "feel good"?

There's cases of German translations (other languages probably also have that) e.g. in Microsoft tools (Visual Studio, Edge) or some Linux distributions that feel "overtranslated", dusty and super formal like "Kopfzeilen" instead of "Headers" or "Zwischenspeicher" instead of "Cache" in HTTP requests, and that just usually makes things sound weird and the average developer would probably rather stick to either more "informal" and laid back translations or just to using the English term you should be familiar with if you're already working with, to stick to the example above, HTTP stuff.

pluiedev: I think we should stick to OS terminology by default unless the default doesn't make sense or is specific to only that OS (for example, using "taskbar" or "dock" for the generic name of a list of open apps). As a rule of thumb the user should be familiar with these terms without much further thought.

pluiedev: So apparently Konsole uses "Ansicht teilen" for splitting and just "Terminal schließen" for closing splits. I chose to avoid "Close Terminal" in the original text since "terminal" could either just refer to one part of the terminal, one terminal window or the entire app, but it looks like that's what Konsole chose... TBH I wouldn't worry too much about it since they also call tabs "Unterfenster" which is just *weird*

pluiedev (approved):
I think this is good to merge aside from the issue of maintainership.

@rpfaeffle and @cactysman, if I understand correctly, both of you have volunteered to take up maintainership for the German translation, right? If so then all we need is some bureaucracy to form a translator team and add you two to the `CODEOWNERS` file. (I'll have to do the same with `zh_CN`.), before or after this PR is merged.

mitchellh:
I've added the teams. I'll update CODEOWNERS after merge.

PR #6602, opened 2025-03-06
font(freetype): constrain emoji to cell width

rockorager:
When scaling emoji (with freetype), we would unilaterally scale the bitmap to fit within the `cell_height`. For narrow fonts, this would result in a horizontal overflow:

image

Modify the glyph rendering such that we scale to fit within the cell width. After doing so, the above image looks like:

image

The emoji glyph is noticeably smaller because we have constrained the height further than before, but fits perfectly within two cells. I am using Victor Mono as my font, which is pretty narrow. The effect would be even more pronounced on something like Iosevka.

### DIFF
--- a/src/font/face/freetype.zig
+++ b/src/font/face/freetype.zig
@@ -390,13 +390,28 @@ pub const Face = struct {
         // and copy the atlas.
         const bitmap_original = bitmap_converted orelse bitmap_ft;
         const bitmap_resized: ?freetype.c.struct_FT_Bitmap_ = resized: {
-            const max = metrics.cell_height;
-            const bm = bitmap_original;
-            if (bm.rows <= max) break :resized null;
+            const original_width = bitmap_original.width;
+            const original_height = bitmap_original.rows;
+            var result = bitmap_original;
+            // TODO: We are limiting this to only emoji. We can rework this after a future
+            // improvement (promised by Qwerasd) which implements more flexible resizing rules. For
+            // now, this will suffice
+            if (self.isColorGlyph(glyph_index) and opts.cell_width != null) {
+                const cell_width = opts.cell_width orelse unreachable;
+                // If we have a cell_width, we constrain the glyph to fit within the cell
+                result.width = metrics.cell_width * @as(u32, cell_width);
+                result.rows = (result.width * original_height) / original_width;
+            } else {
+                // If we don't have a cell_width, we scale to fill vertically
+                result.rows = metrics.cell_height;
+                result.width = (metrics.cell_height * original_width) / original_height;
+            }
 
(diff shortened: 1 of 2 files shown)


qwerasd205:
Have you tested this with glyphs that are *supposed* to overflow the cell width? For example, probably some combining enclosing characters, certain arrows, glyphs from fonts that are intended to be adjoining so overlap past the edge some amount-- I'd be much more comfortable narrowly scoping this change to only apply to emojis.

rockorager:
I haven't, but I agree about limiting it to emoji only. Would something like this be ok?

```diff
diff --git a/src/font/face/freetype.zig b/src/font/face/freetype.zig
index 10eb72e54beb..38b4cc6df234 100644
--- a/src/font/face/freetype.zig
+++ b/src/font/face/freetype.zig
@@ -393,7 +393,8 @@ pub const Face = struct {
             const original_width = bitmap_original.width;
             const original_height = bitmap_original.rows;
             var result = bitmap_original;
-            if (opts.cell_width) |cell_width| {
+            if (self.isColorGlyph(glyph_index) and opts.cell_width != null) {
+                const cell_width = opts.cell_width orelse unreachable;
                 // If we have a cell_width, we constrain the glyph to fit within the cell
                 result.width = metrics.cell_width * @as(u32, cell_width);
                 result.rows = (result.width * original_height) / original_width;
```

qwerasd205:
I believe this actually is a regression, I removed code that intentionally constrained all colored glyphs as a part of #3016 -- restoring that behavior is not an *ideal* solution though, since there can easily be scenarios where even colored glyphs should not be constrained. I have a WIP branch which will greatly improve situations like this generally, allowing us to specify flexible constraint/resizing rules for different classes of glyphs- but I won't be able to finish that up for a while. I guess applying this for now is *probably* fine.

As a side note, constraining the size here instead of in the shader (like it was prior to #3016) is actually preferable in my mind, since that way we don't waste atlas space for no reason.

rockorager:
> I have a WIP branch which will greatly improve situations like this generally, allowing us to specify flexible constraint/resizing rules for different classes of glyphs- but I won't be able to finish that up for a while. I guess applying this for now is probably fine.

I could add a TODO?

qwerasd205:
Yeah, adding a TODO is probably a good idea. Otherwise this does seem like a good change to have for now so I'll approve

qwerasd205 (approved):
LGTM

mitchellh:
Going to merge this since it's been approved for awhile. I didn't review it, just trusting y'all. 😄

cespare:
This introduces a rendering regression on Linux. See #6781.

PR #6614, opened 2025-03-07
gtk: build gtk4-layer-shell ourselves

pluiedev:
As of now `gtk4-layer-shell` is unavailable on recent, stable releases of many distros (Debian 12, Ubuntu 24.04, openSUSE Leap & Tumbleweed, etc.) and outdated on many others (Nixpkgs 24.11/unstable, Fedora 41, etc.) This is inconvenient for our users and severely limits where the quick terminal can be used. As a result we then build gtk4-layer-shell ourselves by default unless `--system` or `-fsys=gtk4-layer-shell` are specified. This also allows me to add an idiomatic Zig API on top of the library and avoiding adding even more raw C code in the GTK apprt.

Since we now build gtk4-layer-shell it should be theoretically available on all Linux systems we target. As such, the `-Dgtk-layer-shell` build option has been removed. This is somewhat of an experimental change as I don't know if gtk4-layer-shell works perfectly across all distros, and we can always add the option back if need be.

### DIFF
--- a/pkg/gtk4-layer-shell/build.zig
+++ b/pkg/gtk4-layer-shell/build.zig
@@ -1,0 +1,106 @@
+const std = @import("std");
+
+// TODO: Import this from build.zig.zon when possible
+const version: std.SemanticVersion = .{ .major = 1, .minor = 1, .patch = 0 };
+
+pub fn build(b: *std.Build) !void {
+    const target = b.standardTargetOptions(.{});
+    const optimize = b.standardOptimizeOption(.{});
+
+    const upstream = b.dependency("gtk4_layer_shell", .{});
+    const wayland_protocols = b.dependency("wayland_protocols", .{});
+
+    // Zig API
+    const module = b.addModule("gtk4-layer-shell", .{
+        .root_source_file = b.path("src/main.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+    module.addIncludePath(upstream.path("include"));
+    // Needs the gtk.h header
+    module.linkSystemLibrary("gtk4", dynamic_link_opts);
+
+    // Shared library
+    const lib = b.addSharedLibrary(.{
+        .name = "gtk4-layer-shell",
+        .target = target,
+        .optimize = optimize,
+    });
+    lib.linkLibC();
+    lib.addIncludePath(upstream.path("include"));
+    lib.addIncludePath(upstream.path("src"));
+
+    // GTK
+    lib.linkSystemLibrary2("gtk4", dynamic_link_opts);
+
+    // Wayland headers and source files
+    {
+        const protocols = [_]struct { []const u8, std.Build.LazyPath }{
+            .{
+                "wlr-layer-shell-unstable-v1",
+                upstream.path("protocol/wlr-layer-shell-unstable-v1.xml"),
+            },
+            .{
+                "xdg-shell",
+                wayland_protocols.path("stable/xdg-shell/xdg-shell.xml"),
+            },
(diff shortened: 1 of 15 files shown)


pkg/gtk4-layer-shell/build.zig.zon
```diff
@@ -0,0 +1,15 @@
+.{
+    .name = "gtk4-layer-shell",
+    .version = "0.0.0",
```
mitchellh, line 3: We try to keep this matched up to the dependency we build, so this should also be 1.1.0.

src/build/SharedDeps.zig
```diff
@@ -261,6 +261,10 @@ pub fn add(
         try static_libs.append(breakpad_dep.artifact("breakpad").getEmittedBin());
     }
 
+    // Layer shell is only included if we're doing a GTK Wayland build,
+    // but we should offer the option here
+    _ = b.systemIntegrationOption("gtk4-layer-shell", .{});
```
mitchellh, line 266: This should go in `src/build/Config.zig` and not here (see there for a comment on why)

pkg/gtk4-layer-shell/build.zig
```diff
@@ -0,0 +1,105 @@
+const std = @import("std");
+
+const version: std.SemanticVersion = .{ .major = 1, .minor = 1, .patch = 0 };
```
mitchellh, line 3: Should add a TODO that once we can import our `build.zig.zon` we should replace this with the version field from there.

mitchellh (changes requested):
Minor changes requested, looks good otherwise ignoring CI failures.

pkg/gtk4-layer-shell/build.zig.zon
```diff
@@ -0,0 +1,15 @@
...
+    .paths = .{""},
+    .dependencies = .{
+        .gtk4_layer_shell = .{
+            .url = "https://github.com/wmww/gtk4-layer-shell/archive/refs/tags/v1.1.0.tar.gz",
+            .hash = "12203eff4829ad8afdd828eb323d48e5ba8dbb44d224e9e314d4ab1533c2bec20f4b",
+        },
+        .wayland_protocols = .{
+            .url = "https://deps.files.ghostty.org/wayland-protocols-258d8f88f2c8c25a830c6316f87d23ce1a0f12d9.tar.gz",
+            .hash = "12201a57c6ce0001aa034fa80fba3e1cd2253c560a45748f4f4dd21ff23b491cddef",
+        },
```
jcollie, line 13: Since there's probably not an automatic way to do this, we should add some comments here reminding people to sync this dep with the same dep in the main build.

pluiedev: Hmm... The cursed thing is that the repo itself vendors *some* protocols (i.e. the wlr-layer-shell protocol itself) but for the core/XDG protocols it relies on the wayland-protocols repo, probably for the higher stability guarantees. I think the main benefit of syncing would be dependency deduplication, but I don't know if Zig currently does that...

jcollie: If there's an explicit reason to keep them out of sync that should be documented in a comment. A zealous maintainer might update them without realizing. Otherwise I feel that they should be kept in sync.

pluiedev: Yeah, I'm also more inclined to just update it whenever it breaks (e.g. when upstream relies on protocols that are newer than what are in our copy of wayland-protocols)

mitchellh (approved):

PR #6622, opened 2025-03-08
core: refactor RepeatablePath into separate files and add Path

jcollie:
Slims down `Config.zig` and makes some of the code reusable in Path.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2639,11 +2639,10 @@ pub fn loadCliArgs(self: *Config, alloc_gpa: Allocator) !void {
         }
     }
 
-    // Config files loaded from the CLI args are relative to pwd
-    if (self.@"config-file".value.items.len > 0) {
-        var buf: [std.fs.max_path_bytes]u8 = undefined;
-        try self.expandPaths(try std.fs.cwd().realpath(".", &buf));
-    }
+    // Any paths referenced from the CLI are relative to the current working
+    // directory.
+    var buf: [std.fs.max_path_bytes]u8 = undefined;
+    try self.expandPaths(try std.fs.cwd().realpath(".", &buf));
 }
 
 /// Load and parse the config files that were added in the "config-file" key.
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -35,6 +35,8 @@ const ErrorList = @import("ErrorList.zig");
 const MetricModifier = fontpkg.Metrics.Modifier;
 const help_strings = @import("help_strings");
 const RepeatableStringMap = @import("RepeatableStringMap.zig");
+pub const Path = @import("path.zig").Path;
+pub const RepeatablePath = @import("path.zig").RepeatablePath;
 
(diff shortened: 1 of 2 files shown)


src/config/Config.zig
```diff
@@ -2828,12 +2829,15 @@ fn expandPaths(self: *Config, base: []const u8) !void {
...
-            );
+        switch (field.type) {
+            RepeatablePath, NonRepeatablePath => {
+                try @field(self, field.name).expand(
+                    arena_alloc,
+                    base,
+                    &self._diagnostics,
+                );
+            },
+            inline else => {},
```
pluiedev, line 2840: Does it need to be inline here? We don't do anything with the type anyway

src/config/NonRepeatablePath.zig
```diff
@@ -0,0 +1,180 @@
...
+/// Deep copy of the struct. Required by Config.
+pub fn clone(self: *const NonRepeatablePath, alloc: Allocator) Allocator.Error!NonRepeatablePath {
+    if (self.value) |value| {
+        return .{
+            .value = try value.clone(alloc),
+        };
+    }
+    return .{
+        .value = null,
+    };
```
pluiedev, line 49: Simplify

```suggestion
    return .{
        .value = if (self.value) |v| v.clone(alloc) else null,
    };
```

pluiedev (changes requested):
NonRepeatablePath is kind of a funny name for an optional path. Is there anything stopping us from using ?Path directly?

jcollie:
> NonRepeatablePath is kind of a funny name for an optional path. Is there anything stopping us from using ?Path directly?

`Path` is a normal `union`, I don't think that the rest of the config has the machinery for dealing with non-packed unions.

src/config/NonRepeatablePath.zig
jcollie: Fixed.

mitchellh (changes requested):
Since they're all path-related, I suggest we put all three into `src/config/path.zig` instead.

jcollie:
OK, switched to just exposing `Path` as a config type and merged everything into `src/config/path.zig`.

src/config/path.zig
```diff
@@ -0,0 +1,470 @@
...
+
+pub const ParseError = error{ValueRequired} || Allocator.Error;
+
+/// Path is like a string that represents a path value. The difference is that
+/// when loading the configuration the value for this will be automatically
+/// expanded relative to the path of the config file (or the home directory).
+pub const Path = union(enum) {
+    /// Path is unset.
+    null,
+
```
pluiedev, line 21: Why is `null` a variant? I'm pretty sure our config mechanism supports just setting a field to `?Path`

pluiedev (changes requested):

src/config/path.zig
jcollie: Fixed.

mitchellh (approved):

pluiedev (approved):

PR #6624, opened 2025-03-08
pkg(gtk4-layer-shell): Enable using system-installed headers for dynamic linking

AnthonyZhOon:
I noticed we weren't doing system-integration against the pkgconfig for gtk4-layer-shell. This behaviour differed from how we handled system integration for existing deps in `pkg/` (oniguruma, fontconfig).

Refactored `pkg/gtk4-layer-shell/build.zig` referencing `pkg/oniguruma/build.zig` to use pkgconfig names in system integration. Previously we used to libname `libgtk4-layer-shell.so` (`gtk4-layer-shell`) instead of pkgconfig name `gtk4-layer-shell-0.pc` which meant system integration still relied on fetching the C-headers via `zig fetch` instead of system C-headers. 

I've tested this with a `--system` build where the relevant `.zig-cache/p/<hash of gtk4-layer-shell>` is stubbed to an empty directory and `pkgconfig(gtk4-layer-shell-0)` is installed instead on fedora linux.

### DIFF
--- a/pkg/gtk4-layer-shell/build.zig
+++ b/pkg/gtk4-layer-shell/build.zig
@@ -3,12 +3,14 @@ const std = @import("std");
 // TODO: Import this from build.zig.zon when possible
 const version: std.SemanticVersion = .{ .major = 1, .minor = 1, .patch = 0 };
 
+const dynamic_link_opts: std.Build.Module.LinkSystemLibraryOptions = .{
+    .preferred_link_mode = .dynamic,
+    .search_strategy = .mode_first,
+};
+
 pub fn build(b: *std.Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
-
-    const upstream = b.dependency("gtk4_layer_shell", .{});
-    const wayland_protocols = b.dependency("wayland_protocols", .{});
 
     // Zig API
     const module = b.addModule("gtk4-layer-shell", .{
@@ -16,10 +18,25 @@ pub fn build(b: *std.Build) !void {
         .target = target,
         .optimize = optimize,
     });
-    module.addIncludePath(upstream.path("include"));
     // Needs the gtk.h header
     module.linkSystemLibrary("gtk4", dynamic_link_opts);
 
+    if (b.systemIntegrationOption("gtk4-layer-shell", .{})) {
+        module.linkSystemLibrary("gtk4-layer-shell-0", dynamic_link_opts);
+    } else {
+        _ = try buildLib(b, module, .{
+            .target = target,
+            .optimize = optimize,
+        });
+    }
+}
+
+fn buildLib(b: *std.Build, module: *std.Build.Module, options: anytype) !*std.Build.Step.Compile {
+    const target = options.target;
+    const optimize = options.optimize;
+
+    const upstream = b.dependency("gtk4_layer_shell", .{});
+    const wayland_protocols = b.dependency("wayland_protocols", .{});
     // Shared library
     const lib = b.addSharedLibrary(.{
         .name = "gtk4-layer-shell",
@@ -29,6 +46,7 @@ pub fn build(b: *std.Build) !void {
     lib.linkLibC();
     lib.addIncludePath(upstream.path("include"));
     lib.addIncludePath(upstream.path("src"));
+    module.addIncludePath(upstream.path("include"));
 
     // GTK
     lib.linkSystemLibrary2("gtk4", dynamic_link_opts);
@@ -76,6 +94,16 @@ pub fn build(b: *std.Build) !void {
         .{ .include_extensions = &.{".h"} },
     );
 
+    // Certain files relating to session lock were removed as we don't use them
+    const srcs: []const []const u8 = &.{
+        "gtk4-layer-shell.c",
+        "layer-surface.c",
+        "libwayland-shim.c",
+        "registry.c",
+        "stolen-from-libwayland.c",
+        "stubbed-surface.c",
+        "xdg-surface-server.c",
+    };
     lib.addCSourceFiles(.{
         .root = upstream.path("src"),
         .files = srcs,
@@ -87,20 +115,5 @@ pub fn build(b: *std.Build) !void {
     });
 
     b.installArtifact(lib);
+    return lib;
 }
-
-// Certain files relating to session lock were removed as we don't use them
-const srcs: []const []const u8 = &.{
-    "gtk4-layer-shell.c",
-    "layer-surface.c",
-    "libwayland-shim.c",
-    "registry.c",
-    "stolen-from-libwayland.c",
-    "stubbed-surface.c",
-    "xdg-surface-server.c",
-};
-
-const dynamic_link_opts: std.Build.Module.LinkSystemLibraryOptions = .{
-    .preferred_link_mode = .dynamic,
-    .search_strategy = .mode_first,
-};
--- a/src/build/SharedDeps.zig
+++ b/src/build/SharedDeps.zig
@@ -506,7 +506,7 @@ pub fn add(
                     // IMPORTANT: gtk4-layer-shell must be linked BEFORE
                     // wayland-client, as it relies on shimming libwayland's APIs.
                     if (b.systemIntegrationOption("gtk4-layer-shell", .{})) {
-                        step.linkSystemLibrary2("gtk4-layer-shell", dynamic_link_opts);
+                        step.linkSystemLibrary2("gtk4-layer-shell-0", dynamic_link_opts);
                     } else {
                         // gtk4-layer-shell *must* be dynamically linked,
                         // so we don't add it as a static library


pluiedev:
Why? You can just do -fsys=gtk4-layer-shell project-wide to use the system library and avoid compiling yourself. How would the system integration option propagate from the main project to the gtk4-layer-shell lib?

AnthonyZhOon:
> Why? You can just do -fsys=gtk4-layer-shell project-wide to use the system library and avoid compiling yourself. 
This won't mean compiling gtk4-layer-shell in the system-integration, only using the system-installed headers instead of the ones downloaded via the zig build system.

Our existing system-integrated libraries already prefer using pkgconfig names in the b.systemIntegration(<name>) instead of shared-object names. By specifying a pkgconfig name the build system can use it to find system headers and resolve the linking tree. 
```
> pkgconf --cflags --shared --libs gtk4-layer-shell-0
-I/usr/include/gtk4-layer-shell -I/usr/include/gtk-4.0 -I/usr/include/pango-1.0 -I/usr/include/fribidi -I/usr/include/harfbuzz -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/webp -I/usr/include/cairo -I/usr/include/libxml2 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/pixman-1 -I/usr/include/graphene-1.0 -I/usr/lib64/graphene-1.0/include -mfpmath=sse -msse -msse2 -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -DWITH_GZFILEOP -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/sysprof-6 -pthread -lgtk4-layer-shell -lpangocairo-1.0 -lpango-1.0 -lgdk_pixbuf-2.0 -lcairo-gobject -lcairo -lvulkan -lgraphene-1.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0
``` 

Otherwise, I want to minimise the bundled files (instead of system packaged) needed to build ghostty. It seems the package reviewers for fedora linux at https://bugzilla.redhat.com/show_bug.cgi?id=2334903 prefer it. 
In packaging for official channels in Fedora Linux, I have to audit the bundled dependencies and track down the licenses for all source files that are bundled into a `ghostty.src.rpm` which can perform a standalone offline build which specifies system-packages installed before the offline phase and contains bundled archives of other downloads. Minimising the size of this `.src.rpm` is preferred for packaging.
```sh
$ zig build --fetch --global-cache-dir $(pwd)/cache-thing
// Remove contents of the downloaded gtk4-layer-shell (including header files)
$ rm -r $(pwd)/cache-thing/p/12203eff4829ad8afdd828eb323d48e5ba8dbb44d224e9e314d4ab1533c2bec20f4b/*
$ zig build --system --global-cache-dir $(pwd)/cache-thing/p --help
// gtk4-layer-shell is system-integrated
$ zig build --system --global-cache-dir $(pwd)/cache-thing/p
```
On main this process errors as we rely on downloading the headers to dynamic link against instead of using the system-installed headers + system-installed compiled shared-lib.
```
install
└─ install ghostty
   └─ zig build-exe ghostty Debug native 2 errors
pkg/gtk4-layer-shell/src/main.zig:1:11: error: C import failed
const c = @cImport({
          ^~~~~~~~
referenced by:
    ShellLayer: pkg/gtk4-layer-shell/src/main.zig:7:18
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
/home/fedora/dev/ghostty/.zig-cache/o/01084885f82067596db10f997758cb59/cimport.h:1:10: error: 'gtk4-layer-shell.h' file not found
#include <gtk4-layer-shell.h>
```
> How would the system integration option propagate from the main project to the gtk4-layer-shell lib?
The sub-dependency is passed the Build object when declared as a dependency, `b.dependency(module: "gtk4_layer_shell")` in ghostty's `build.zig` passes the Build into the `build.zig.zon` dep with the module name and includes the module as a build-step in ghostty's build. This also lets it inherit ghostty's build and system-integration options, though technically it can override stuff there.

I think we've also kept our `pkg/` as able to be compiled standalone and taken as dependencies for other projects, perhaps in the future we'll be able to split them as separate zig-packages.

Not sure if it's a real problem, but we could potentially have mismatches between our downloaded headers and the system shared-object.

pluiedev:
On Nix at least the pkg-config name is definitely just `gtk4-layer-shell`. Maybe it's different on Fedora, but since our sole official build environment is Nix we should depend on whatever its naming conventions are.

AnthonyZhOon:
I don't personally run Nix, but I don't see any patch that would change the pkgconf file name from `gtk4-layer-shell-0.pc` as built from source at `https://github.com/wmww/gtk4-layer-shell`. 
Other distros seem to keep the name as is https://debian.pkgs.org/sid/debian-main-amd64/libgtk4-layer-shell-dev_1.0.4-1_amd64.deb.html , https://pkgs.org/search/?q=gtk4-layer-shell. And I wouldn't expect any distro to alter the pkgconf name from what upstream source provides.
Note that the zig build system can be configured to prefer pkgconf names, but will also fallback to searching the `/usr/lib*/` paths from `/usr/lib/lib<name provided>` and provide a plain `-l<name>` flag to match which is what I think we rely on until this PR. 

I think the build failure on main when the downloaded headers are removed are a sign that we aren't using system-headers. Note that I have the `-devel` package installed providing headers + pkgconf metadata.
```
$ zig build --fetch --global-cache-dir $(pwd)/cache-thing
// Remove contents of the downloaded gtk4-layer-shell (including header files)
$ rm -r $(pwd)/cache-thing/p/12203eff4829ad8afdd828eb323d48e5ba8dbb44d224e9e314d4ab1533c2bec20f4b/*
$ zig build --system --global-cache-dir $(pwd)/cache-thing/p --help
// gtk4-layer-shell is system-integrated
$ zig build --system --global-cache-dir $(pwd)/cache-thing/p
```

pluiedev:
I still think you misunderstand the point of this package. The point is that downstream packagers should NOT compile gtk4-layer-shell at all and should instead rely on their system package. When you do `-fsys=gtk4-layer-shell` or alternatively `--system` this is exactly what happens. Aside from the handwritten Zig bindings nothing is compiled here.

On some distros like Debian 12 and Ubuntu 24.04 however gtk4-layer-shell is not available and there is NO "system header" to speak of, and then we must vendor it ourselves. This is not the case for Fedora. For all intents and purposes the C shared library compiled here might as well not exist. 

If the dispute here is purely the difference in the name of the pkg-config file or the shared library then I think it's fine to make minor changes, but otherwise I entirely fail to understand the changes here.

AnthonyZhOon:
Oh, we only need to change to pkgconfig on the system-integration and the `pkg/gtk4-layer-shell/` can be untouched.

AnthonyZhOon:
Hmm, only changing to the pkgconfig name in `src/build/SharedDeps.zig` still attempts to build `pkg/gtk4-layer-shell`

pluiedev:
That sounds far more reasonable. Then perhaps Nix is the odd one out and we should try to follow what most distros do.

pluiedev:
Are you sure that system integration options are automatically enabled by pkg-config? For me at least (maybe this is yet another Nix oddity) the integration options always show up as disabled even when the right libraries are available in the environment.

AnthonyZhOon:
System integrations are only enabled when explicitly `-fsys=<name>` or using a `zig build --system` otherwise we default to disabled
```
> zig build --help
> zig build --fetch --global-cache-dir $(pwd)/cache-thing
> zig build --system cache-thing/p
```

pluiedev:
Yeah exactly, so I'm not sure what's wrong here. If `-fsys=gtk4-layer-shell` is set and it's still compiling gtk4-layer-shell then something is wrong

mitchellh:
This looks right to me.

@pluiedev The pkg-config name is `gtk4-layer-shell-0` (https://github.com/wmww/gtk4-layer-shell/blob/cf05a2ba3a00070d22dcd146d6f894507bcdafc9/src/meson.build#L36). Note that we've had issues in the past with Nix being weird here, but also I think pkg-config falls back to searching for lib names on some path if the pc file is not found and inferring stuff from there. This has happened before. When in doubt, we rely on the source package and it points to the `-0` suffix.

It's also just generally good form and expected that we add system integration to the `pkg/` libraries and I missed this. I don't see it applying to this one but some of the major reasons we do it for the rest is:

1. If you ever have a transitive dependency (e.g. `ghostty -> freetype`, but also `ghostty -> harfbuzz -> freetype`) then `-fsys` applies properly to the full tree. Without a change like this, the nested dep would build from source. Doesn't apply yet (ever?) to `gtk4-layer-shell` but would be silly to get shot in the foot down the road over this.

2. Let's you run `zig build` commands in the `pkg/` dir against system linking. Again, less important for this, more important when we have lots of logic we have tests for. But, no reason to not do it here.

AnthonyZhOon:
Oops, need to expose `gtk.h` header in the module even when not building it

pluiedev:
I'm honestly thinking of just externing symbols manually and replacing all types from GTK with anyopaques, given we don't even want to deal with raw C types in GTK code right now

jcollie (approved):

pluiedev (approved):

PR #6628, opened 2025-03-08
macOS: Set LANGUAGE env var based on macOS preferred language list

mitchellh:
Sets the LANGUAGE environment variable based on the preferred languages as reported by NSLocale. 

macOS has a concept of preferred languages separate from the system locale. The set of preferred languages is a list in priority order of what translations the user prefers. A user can have, for example, "fr_FR" as their locale but "en" as their preferred language. This would mean that they want to use French units, date formats, etc. but they prefer English translations.

gettext uses the LANGUAGE environment variable to override only translations and a priority order can be specified by separating the languages with colons. For example, "en:fr" would mean that English translations are preferred but if they are not available then French translations should be used.

To further complicate things, Apple reports the languages in BCP-47 format which is not compatible with gettext's POSIX locale format so we have to canonicalize them. To canonicalize the languages we use an internal function from libintl. This isn't normally available but since we compile from source on macOS we can use it. This isn't necessary for other platforms.

This logic is only run if the user didn't explicitly request a specific locale, so it should really only affect macOS app launches. From the CLI the environment will have a locale unless the user really explicitly clears it out.

### DIFF
--- a/src/os/locale.zig
+++ b/src/os/locale.zig
@@ -91,19 +91,113 @@ fn setLangFromCocoa() void {
     const z_lang = std.mem.sliceTo(c_lang, 0);
     const z_country = std.mem.sliceTo(c_country, 0);
 
-    // Format them into a buffer
-    var buf: [128]u8 = undefined;
-    const env_value = std.fmt.bufPrintZ(&buf, "{s}_{s}.UTF-8", .{ z_lang, z_country }) catch |err| {
-        log.warn("error setting locale from system. err={}", .{err});
-        return;
-    };
-    log.info("detected system locale={s}", .{env_value});
+    // Format our locale as "<lang>_<country>.UTF-8" and set it as LANG.
+    {
+        var buf: [128]u8 = undefined;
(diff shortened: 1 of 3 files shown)


src/os/locale.zig
```diff
@@ -91,19 +91,113 @@ fn setLangFromCocoa() void {
...
+        if (fbs.pos > 0) {
+            _ = writer.writeByte(':') catch
+                return error.NoSpaceLeft;
+        }
 
-    // Set it onto our environment
-    if (internal_os.setenv("LANG", env_value) < 0) {
-        log.warn("error setting locale env var", .{});
-        return;
+        // Apple languages are in BCPS-47 format, and we need to
```
pluiedev, line 178:
```suggestion
        // Apple languages are in BCP-47 format, and we need to
```

pluiedev (changes requested):

clason:
Sorry for posting here, but I believe the implementation here is incorrect: On my system, I get `LANGUAGE=en_US.UTF-8:de_AT.UTF-8` (where the _first_ language is the preferred system language). But according to https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html#The-LANGUAGE-variable, _later_ languages in this list have priority, which means I get (unwanted) translations.

PR #6629, opened 2025-03-08
gtk: implement quick-terminal-size

### LINKED ISSUE #2384: macOS: Customize initial size for quick terminal

pluiedev:
Fixes #2384 on GTK

I'm not exactly sure how to deal with centered quick terminals so I opted to make them similar to either top/bottom or left/right quick terminals based on the monitor's orientation (portrait/landscape). This may not be the right approach, so I'd like to hear more thoughts about this.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -82,6 +82,7 @@ pub const DerivedConfig = struct {
     gtk_toolbar_style: configpkg.Config.GtkToolbarStyle,
 
     quick_terminal_position: configpkg.Config.QuickTerminalPosition,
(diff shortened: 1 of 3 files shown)


jcollie (approved):
Can't test but looks sane to me.

mitchellh:
Note in the future I need to review the config changes. I’ll review post merge but may make changes after the fact. 

Apprt owners aren’t allowed to approve config changes that are unqualified. I should make a new code owner for it probably.

pluiedev:
Yeah I was also a bit iffy about merging it so quickly since the behavior of quick-terminal-position=center is entirely undefined in the original proposal and I had to pull it out of my backside so to speak

PR #6630, opened 2025-03-08
terminal: remove redundant assertIntegrity from clearPrompt

jparise:
clearCells() always asserts its page's integrity after finishing its work (via a `defer`). We don't need to re-assert the page's integrity immediately thereafter.


jparise:
https://github.com/ghostty-org/ghostty/blob/e03e98e106e96d9ceff5a9781b147376eaed0491/src/terminal/page.zig#L1066-L1072

mitchellh (approved):

PR #6634, opened 2025-03-09
gtk: convert Surface to zig-gobject

jcollie:
Marking this is as draft because I want to test this further before saying that it's ready, but putting it out there for feedback.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -7,6 +7,8 @@ const std = @import("std");
 
 const adw = @import("adw");
 const gtk = @import("gtk");
+const gdk = @import("gdk");
+const glib = @import("glib");
 const gio = @import("gio");
 const gobject = @import("gobject");
 
@@ -34,7 +36,6 @@ const URLWidget = @import("URLWidget.zig");
 const CloseDialog = @import("CloseDialog.zig");
 const inspectorpkg = @import("inspector.zig");
 const gtk_key = @import("key.zig");
-const c = @import("c.zig").c;
 const Builder = @import("Builder.zig");
 const adwaita = @import("adwaita.zig");
 
@@ -85,10 +86,10 @@ pub const Container = union(enum) {
 
         /// Returns the GTK widget to add to the paned for the given
         /// element
-        pub fn widget(self: Elem) *c.GtkWidget {
+        pub fn widget(self: Elem) *gtk.Widget {
             return switch (self) {
                 .surface => |s| s.primaryWidget(),
-                .split => |s| @ptrCast(@alignCast(s.paned)),
+                .split => |s| s.paned.as(gtk.Widget),
             };
         }
 
@@ -239,10 +240,10 @@ container: Container = .{ .none = {} },
 app: *App,
 
 /// The overlay, this is the primary widget
-overlay: *c.GtkOverlay,
+overlay: *gtk.Overlay,
 
 /// Our GTK area
-gl_area: *c.GtkGLArea,
+gl_area: *gtk.GLArea,
 
 /// If non-null this is the widget on the overlay that shows the URL.
 url_widget: ?URLWidget = null,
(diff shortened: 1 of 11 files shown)


src/apprt/gtk/Surface.zig
```diff
@@ -1871,15 +1965,18 @@ fn gtkInputPreeditStart(
...
+    self: *Surface,
 ) callconv(.C) void {
-    const self = userdataSelf(ud.?);
-
     // Get our pre-edit string that we'll use to show the user.
-    var buf: [*c]u8 = undefined;
-    _ = c.gtk_im_context_get_preedit_string(ctx, &buf, null, null);
-    defer c.g_free(buf);
+    var buf: [*:0]u8 = undefined;
+    var attrs: *pango.AttrList = undefined;
```
pluiedev, line 1973: Honestly I would just use an anyopaque here since we don't use this anyway. Avoids pulling in all of Pango

jcollie: Unfortunately this looks like a possible bug in the GIR that is used to generate the gobject bindings. Looking at some of the underlying source it looks like you _should_ be able to set those to `null` without problems but given that we're actually using `gtk.IMMulticontext` I'm unable to 100% track that down to be sure. And given that `pango.AttrList` is using gobject's reference couting scheme I'm not comfortable saying that we can just `glib.free` that memory without consequences if we coerce an `*anyopaque` into satisfying the compiler.

https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkimcontext.c#L362

Plus "pulling in all of Pango" is a bit of a stretch as Pango is a hard requirement of GTK. The only change really is that we have to link to the dynamic library now. There shouldn't be any changes to memory size or anything like that.

jcollie:
OK, have been running this "in production" for a couple of days so this is ready for more review.

tristan957 (approved):
Looks good pending pango comment

src/apprt/gtk/Surface.zig
jcollie: OK, I really didn't like the idea of using an anyopaque and just freeing the memory. So I took the harder route and built a custom GObject binding (still using zig-gobject) that allowed me to patch the GIR and allow `null` to be passed so that no memory would even be allocated in the first place.

This also allowed me to bring in fixes for bugs that I found while trying to convert part of `src/apprt/gtk/key.zig`. And doubtless many others as well.

https://github.com/ianprime0509/zig-gobject/issues/96

You can see how I built the custom bindings here:

https://github.com/jcollie/ghostty-gobject

I'd be willing to move this to the Ghostty GH organization so that it'll be more maintainable long-term.

tristan957: Can you upstream your work?

jcollie: Probably wouldn't be too difficult to get it into `zig-gobject`. The ultimate prize would be to see if we could upstream this all the way to GTK.

I think that we still need to build our own bindings as `zig-gobject` doesn't make very frequent releases and the repo has already moved on to 0.14. Not sure how long we'll be on 0.13 although momentum seems to be building.

mitchellh: I'll just note that I don't love the idea of bringing in another dependency to maintain. But I don't have the full context here. I'll have to take a closer look when I have time.

jcollie: Well, it's not _technically_ a new dependency. Ghostty doesn't actually consume `zig-gobject` directly. Instead `zig-gobject` is used to generate code that let's us use GTK with a nicer interface. Before, we were using the code generated by upstream. But that hasn't been regenerated since last October and there have been many fixes to the code generation since then (including several that directly affect us). All that I've done is generate a new set of bindings with the latest fixes (plus one tweak specifically for Ghostty). If/when `zig-gobject` generates a new set of bindings that suits Ghostty's needs it's easy enough to switch just by updating the URL.

jcollie: If you were specifically referring to linking to Pango, that's totally unnecessary with the updated bindings.

jcollie: With the update gobject bindings that were brought in with the 0.14 update this is resolved.

jcollie:
Updated for Zig 0.14 and even used a couple of decl literals!

Issue #5744, opened 2025-02-13
Zig 0.14 migration


pluiedev:
Zig 0.14 is dropping in a few days as of time of writing (~~February 13th,~~ March 3rd, 2025), and we should be prepared for it. List of dependencies that are 0.14 ready:

- [x] `libxev`: https://github.com/mitchellh/libxev/pull/153
- [x] `mach_glfw`: https://github.com/mitchellh/mach-glfw/pull/1
- [x] `libvaxis`: https://github.com/rockorager/libvaxis/pull/114
- [ ] `z2d`: https://github.com/vancluever/z2d/issues/86
- [x] `zig-objc`: https://github.com/mitchellh/zig-objc/pull/16
- [x] `zig-js`: mitchellh/zig-js#6
- [x] `zig-wayland`: Updated on main as of https://codeberg.org/ifreund/zig-wayland/commit/91dcb18041e92fb73d0c09c6b7416c24d5295923
- [x] `zf`: Already works™
- [x] `zig-gobject`: Updated on main as of https://github.com/ianprime0509/zig-gobject/commit/25fffb01af86f4a43f068917fa65545a9791b32d
- [x] `ziglyph`" Works, but we should migrate to zg later #5694 

## Other issues (non-blocking)

We should make dedicated issues for some of these once we have 0.14 support merged.

- [ ] Update to zg since ziglyph is abandoned #5694 
- [X] [Update zig2nix to 0.14](https://github.com/jcollie/zig2nix/commit/03696f84837c0432bafc33f752095724b7d48bc8)
- [x] https://github.com/ziglang/zig/issues/23219
- [ ] Move gobject build process into our build system

mitchellh:
> mach_glfw: [Upstream abandoned](https://github.com/emidoots/mach-glfw) — looks like we need to port to 0.14 ourselves

Note we already have this forked. https://github.com/mitchellh/mach-glfw and we use that fork, so it shouldn't be too much work I hope.

mitchellh:
> zig-objc: There's a community PR: https://github.com/mitchellh/zig-objc/pull/9

This PR looks pretty much good to go if it still passes once 0.14 releases. I'm the maintainer of this so that looks good.

mitchellh:
> zig-js

Our wasm builds are broken _anyways_, so this isn't a blocker. cc @gabydd on this one.

gabydd:
[...]

PR #6649, opened 2025-03-10
Update to Zig 0.14

selfisekai:
Bug: #5744.


mitchellh:
I don't really want a draft PR sitting around for a long time (we have enough as-is for me to clean up). I'll link to your branch in the issue.

PR #6658, opened 2025-03-10
termio, flatpak: implement process watcher with xev

alaviss:
This allows `termio.Exec` to track processes spawned via `FlatpakHostCommand`, finally allowing Ghostty to function as a Flatpak.

Alongside this is a few bug fixes:

* Don't add ghostty to PATH when running in flatpak mode since it's unreachable.
* Correctly handle exit status returned by Flatpak. Previously this was not processed and contains extra status bits.
* Use correct type for PID returned by Flatpak.

### DIFF
--- a/src/os/flatpak.zig
+++ b/src/os/flatpak.zig
@@ -3,6 +3,7 @@ const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
 const builtin = @import("builtin");
 const posix = std.posix;
+const xev = @import("../global.zig").xev;
 
 const log = std.log.scoped(.flatpak);
 
@@ -71,16 +72,26 @@ pub const FlatpakHostCommand = struct {
 
         /// Process started with the given pid on the host.
         started: struct {
-            pid: c_int,
+            pid: u32,
+            loop_xev: ?*xev.Loop,
+            completion: ?*Completion,
             subscription: c.guint,
             loop: *c.GMainLoop,
         },
 
         /// Process exited
         exited: struct {
-            pid: c_int,
+            pid: u32,
             status: u8,
         },
+    };
+
+    pub const Completion = struct {
+        callback: *const fn (ud: ?*anyopaque, l: *xev.Loop, c: *Completion, r: WaitError!u8) void = noopCallback,
+        c_xev: xev.Completion = .{},
+        userdata: ?*anyopaque = null,
+        timer: ?xev.Timer = null,
+        result: ?WaitError!u8 = null,
     };
 
     /// Errors that are possible from us.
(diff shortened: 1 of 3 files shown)


src/os/flatpak.zig
```diff
@@ -135,6 +148,77 @@ pub const FlatpakHostCommand = struct {
         }
     }
 
+    /// Wait for the process to end asynchronously via libxev. This
+    /// can only be called ONCE.
+    pub fn wait_xev(
```
pluiedev, line 153: Functions in Zig are camelCased

```suggestion
    pub fn waitXev(
```

src/os/flatpak.zig
```diff
@@ -71,18 +72,28 @@ pub const FlatpakHostCommand = struct {
 
         /// Process started with the given pid on the host.
         started: struct {
-            pid: c_int,
+            pid: u32,
```
pluiedev, line 75: Is the PID necessarily a u32? pid_t is always c_int AFAICT

pluiedev (changes requested):
I don't really know anything about Flatpak so here are some superficial nitpicks

src/os/flatpak.zig
alaviss: Yea but Flatpak's `HostCommand` interface outputs `u32`, so I made sure that's followed here.

mitchellh (changes requested):
There's some stuff I want to clean up, so hold off on merging, but this generally looks good to me. 

I think it may be time we add a flatpak build CI job back again. Even though it's still not an officially supported artifact type, I've given it the green stamp to move forward through contributions and want to help ensure we don't regress.

src/os/flatpak.zig
```diff
@@ -135,6 +148,77 @@ pub const FlatpakHostCommand = struct {
         }
     }
 
+    /// Wait for the process to end asynchronously via libxev. This
+    /// can only be called ONCE.
```
jparise, line 152: Could this be an assert-able / log-able precondition?

yorickpeterse:
I tested these changes for my [Flatpak build](https://github.com/yorickpeterse/ghostty-flatpak) and can confirm that Ghostty now starts correctly:

image

tristan957:
Have the Flatpak contributors read https://blogs.gnome.org/chergert/2023/12/14/prompt/? Does Ghostty run into any of the problems described there?

alaviss:
> Does Ghostty run into any of the problems described there?

It doesn't seem to run into any of those with my test programs.

src/os/flatpak.zig
alaviss: I can add an assertion to make sure that no `Completion` was attached to the current `FlatpakHostCommand` if that's wanted.

That wouldn't catch `Completion` object reuse bugs, though.

src/termio/Exec.zig
```diff
@@ -610,6 +641,7 @@ pub const ThreadData = struct {
     // The preallocation size for the write request pool. This should be big
     // enough to satisfy most write requests. It must be a power of 2.
     const WRITE_REQ_PREALLOC = std.math.pow(usize, 2, 5);
+    const FlatpakCompletion = if (Subprocess.FlatpakHostCommand != void) Subprocess.FlatpakHostCommand.Completion else struct {};
```
alaviss, line 644: Not sure if an empty struct is the best bet here, but I don't know if there's anything that can initialize void or struct.

src/os/flatpak.zig
pluiedev: Maybe [`std.once`](https://ziglang.org/documentation/master/std/#std.once.once) could help? Although that might not make sense depending on what "once" means here — is it once for the lifespan of the entire program, or something else?

pluiedev (changes requested):

src/os/flatpak.zig
mitchellh: libxev doesn't verify this at all either (mostly makes it up to the caller for performance reasons). I don't think we need to block merging on this.

mitchellh:
I pushed up some stylistic tweaks, simplified some conditional nesting, and added a CI job to verify `-Dflatpak` builds.

PR #6661, opened 2025-03-10
core, gtk: implement host resources dir for Flatpak

alaviss:
Introduces host resources directory as a new concept: A directory containing application resources that can only be accessed from the host operating system. This is significant for sandboxed application runtimes like Flatpak where shells spawned on the host should have access to application resources to enable integrations.

Alongside this, apprt is now allowed to override the resources lookup logic.

### DIFF
--- a/src/apprt/gtk/flatpak.zig
+++ b/src/apprt/gtk/flatpak.zig
@@ -1,0 +1,29 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const build_config = @import("../../build_config.zig");
+const internal_os = @import("../../os/main.zig");
+const glib = @import("glib");
+
+pub fn resourcesDir(alloc: Allocator) !internal_os.ResourcesDir {
+    if (comptime build_config.flatpak) {
+        // Only consult Flatpak runtime data for host case.
+        if (internal_os.isFlatpak()) {
+            var result: internal_os.ResourcesDir = .{
+                .app_path = try alloc.dupe(u8, "/app/share/ghostty"),
+            };
+            errdefer alloc.free(result.app_path.?);
+
+            const keyfile = glib.KeyFile.new();
+            defer keyfile.unref();
+
+            if (keyfile.loadFromFile("/.flatpak-info", .{}, null) == 0) return result;
+            const app_dir = std.mem.span(keyfile.getString("Instance", "app-path", null)) orelse return result;
+            defer glib.free(app_dir.ptr);
+
+            result.host_path = try std.fs.path.join(alloc, &[_][]const u8{ app_dir, "share", "ghostty" });
+            return result;
+        }
+    }
+
+    return try internal_os.resourcesDir(alloc);
+}
--- a/src/global.zig
+++ b/src/global.zig
@@ -9,6 +9,7 @@ const harfbuzz = @import("harfbuzz");
 const oni = @import("oniguruma");
 const crash = @import("crash/main.zig");
 const renderer = @import("renderer.zig");
+const apprt = @import("apprt.zig");
 
 /// We export the xev backend we want to use so that the rest of
 /// Ghostty can import this once and have access to the proper
@@ -35,7 +36,7 @@ pub const GlobalState = struct {
 
     /// The app resources directory, equivalent to zig-out/share when we build
     /// from source. This is null if we can't detect it.
-    resources_dir: ?[]const u8,
+    resources_dir: internal_os.ResourcesDir,
 
     /// Where logging should go
     pub const Logging = union(enum) {
@@ -62,7 +63,7 @@ pub const GlobalState = struct {
             .action = null,
             .logging = .{ .stderr = {} },
             .rlimits = .{},
-            .resources_dir = null,
+            .resources_dir = .{},
         };
         errdefer self.deinit();
 
@@ -170,11 +171,14 @@ pub const GlobalState = struct {
 
         // Find our resources directory once for the app so every launch
         // hereafter can use this cached value.
-        self.resources_dir = try internal_os.resourcesDir(self.alloc);
-        errdefer if (self.resources_dir) |dir| self.alloc.free(dir);
+        self.resources_dir = rd: {
+            if (@hasDecl(apprt.runtime, "resourcesDir")) break :rd try apprt.runtime.resourcesDir(self.alloc);
+            break :rd try internal_os.resourcesDir(self.alloc);
+        };
+        errdefer self.resources_dir.deinit(self.alloc);
 
         // Setup i18n
-        if (self.resources_dir) |v| internal_os.i18n.init(v) catch |err| {
+        if (self.resources_dir.app()) |v| internal_os.i18n.init(v) catch |err| {
             std.log.warn("failed to init i18n, translations will not be available err={}", .{err});
         };
     }
@@ -182,7 +186,7 @@ pub const GlobalState = struct {
     /// Cleans up the global state. This doesn't _need_ to be called but
     /// doing so in dev modes will check for memory leaks.
     pub fn deinit(self: *GlobalState) void {
-        if (self.resources_dir) |dir| self.alloc.free(dir);
+        self.resources_dir.deinit(self.alloc);
 
         // Flush our crash logs
         crash.deinit();
--- a/src/global.zig
+++ b/src/global.zig
@@ -171,10 +171,7 @@ pub const GlobalState = struct {
 
         // Find our resources directory once for the app so every launch
         // hereafter can use this cached value.
-        self.resources_dir = rd: {
-            if (@hasDecl(apprt.runtime, "resourcesDir")) break :rd try apprt.runtime.resourcesDir(self.alloc);
-            break :rd try internal_os.resourcesDir(self.alloc);
-        };
+        self.resources_dir = try apprt.runtime.resourcesDir(self.alloc);
         errdefer self.resources_dir.deinit(self.alloc);
 
         // Setup i18n
--- a/src/os/resourcesdir.zig
+++ b/src/os/resourcesdir.zig
@@ -2,13 +2,41 @@ const std = @import("std");
 const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 
+pub const ResourcesDir = struct {
+    app_path: ?[]const u8 = null,
+    host_path: ?[]const u8 = null,
+
+    /// Free resources held. Requires the same allocator as when resourcesDir()
+    /// is called.
+    pub fn deinit(self: *ResourcesDir, alloc: std.mem.Allocator) void {
+        if (self.app_path) |p| alloc.free(p);
+        if (self.host_path) |p| alloc.free(p);
+    }
+
+    /// Get the directory to the bundled resources directory accessible
+    /// by the application.
+    pub fn app(self: *ResourcesDir) ?[]const u8 {
+        return self.app_path;
+    }
+
+    /// Get the directory to the bundled resources directory accessible
+    /// by the host environment (i.e. for sandboxed applications). The
+    /// returned directory might not be accessible from the application
+    /// itself.
+    ///
+    /// In non-sandboxed environment, this should be the same as app().
+    pub fn host(self: *ResourcesDir) ?[]const u8 {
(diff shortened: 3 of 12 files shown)


src/os/resourcesdir.zig
```diff
@@ -1,6 +1,9 @@
 const std = @import("std");
 const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
+const build_config = @import("../build_config.zig");
+const internal_os = @import("main.zig");
+const glib = @import("glib");
```
jcollie, line 6: The Zig GObject bindings shouldn't be used outside of `src/gtk`. _Maybe_ we could get away with it inside the `comptime` block lower down, but this seems more like something that should perhaps be handled by each individual apprt, which requires a larger discussion.

jcollie (changes requested):

src/os/resourcesdir.zig
pluiedev: I agree that this should be moved inside apprt-specific code.

mitchellh: This code ends up being used by libghostty so I agree that we shouldn't have a hard requirement on glib for libghostty if we can help it. 

That being said, libghostty does expect to find a resources dir without any apprt loaded, so it can't be GTK-specific either (any app can presumably use libghostty without GTK). 

I suggest we allow apprt's to export a function with custom resourcesDir logic that we run first if it exists. This has to happen globally very early in libghostty's initialization so I suggest perhaps: `apprt.runtime.resourcesDir` with the same signature as this one.

src/os/flatpak.zig
```diff
@@ -3,6 +3,7 @@ const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
```
jcollie, line 3: This appears to be a duplicate commit from #6658 . Should rebase so that this PR only includes the resources dir change.

alaviss: I can't do much until #6658 is merged.

src/os/resourcesdir.zig
alaviss: A key file is really just an ini file, so an alternative would be to use a zig-based parser if you think that'd work better.

jcollie: Adding a INI parser seems like more work than is necessary. Adding `apprt.runtime.resourcesDir` in the way that Mitchell suggests shouldn't be very difficult depending on your experience with Zig. Let us know if you'd like some more pointers.

alaviss: I found an issue with the current approach: The resource path returned is accessible by applications running on the host (e.g. spawned shells), but Ghostty itself cannot access them and must use `/app`.

A solution I'm having in mind is to add a new function to `os.flatpak` to host-ify flatpak paths, exclusively used for `termio.Exec.Subprocess` environment setup, but I'm not sure if this is the way to roll.

alaviss:
Changes in v2:

- New `FlatpakInfo` struct that provides methods for querying information about the current Flatpak.
- Only resolve host paths for `termio.Exec` instead of everything. This fixes issues with themes as Ghostty cannot access the host paths directly and must use `/app` paths.

src/os/flatpak.zig
```diff
@@ -501,3 +501,61 @@ pub const FlatpakHostCommand = struct {
...
+
+/// A struct to help retrive information about the running Flatpak.
+///
+/// Requires GLib to be available and linked.
+pub const FlatpakInfo = struct {
+    const Self = @This();
+    const c = @cImport({
+        @cInclude("glib.h");
+    });
+
```
jcollie, line 513: New code that uses the raw C GTK interfaces is a non-starter, _expecially_ outside `src/gtk`. If you're not sure how to implement the interface that Mitchell described let us know and we can help.

src/os/flatpak.zig
```diff
@@ -501,3 +501,61 @@ pub const FlatpakHostCommand = struct {
...
+    pub fn init() !Self {
+        const keyfile = c.g_key_file_new();
+        errdefer c.g_key_file_unref(keyfile);
+
+        if (c.g_key_file_load_from_file(keyfile, "/.flatpak-info", 0, null) == 0) return error.FileNotFound;
+
+        return .{
+            .keyfile = keyfile.?,
+            .app_dir = null,
+        };
```
jcollie, line 526: There doesn't seem to be much point in keeping the keyfile around for the lifetime of the program. It would seem to me to be better to just read the key that we need, cache that, and then unref the keyfile.

src/termio/Exec.zig
```diff
@@ -756,8 +757,25 @@ const Subprocess = struct {
...
+            var info = FlatpakInfo.init() catch break :rd dir;
+            defer info.deinit();
+
+            break :rd info.resolveAppToHost(dir, &host_resources_dir_buf) catch |err| switch (err) {
+                // Resources directory is not app-relative, ignore
+                error.FlatpakNotAppPath => dir,
+                else => return err,
+            };
+        };
+
```
jcollie, line 776: I don't think that we should be adding more apprt-specific behavior here. This is already a very complex function. We should probably figure out a way to have this behavior encapsulated inside the apprt as well.

src/termio/Exec.zig
```diff
@@ -725,6 +725,7 @@ const Subprocess = struct {
     /// If we build with flatpak support then we have to keep track of
     /// a potential execution on the host.
     const FlatpakHostCommand = if (build_config.flatpak) internal_os.FlatpakHostCommand else void;
+    const FlatpakInfo = if (build_config.flatpak) internal_os.FlatpakInfo else void;
```
jcollie, line 728: As, indicated by CI, this won't compile on non-flatpak.

src/os/flatpak.zig
jcollie: CI never got a chance to try macOS builds because of other errors, but this is not going to compile on macOS.

jcollie (changes requested):

src/os/flatpak.zig
alaviss: > If you're not sure how to implement the interface that Mitchell described let us know and we can help.

I'd love to have some guidance on this.

What has been bugging me is that Flatpak comes with the distinction of host vs. app paths and it is not clear to me how to model this well.

Right now my solution is basically just more logic into termio, which isn't really great.

alaviss:
Changes in v3:

- Dropped `FlatpakInfo` as it is not very useful outside of the host resources directory case.
- `GlobalState` will now resolve `*resources_dir` with `apprt.runtime` implementation if available.
- Added `host_resources_dir` to `GlobalState`. I'm not entirely sure whether this is the best way, but it's certainly the cleanest to me.
- Removed additional logic in `termio.Exec`.

alaviss:
Changes in v4:

- Rebased to `main` and drop #6658 from commit chain.

src/apprt/gtk/flatpak.zig
```diff
@@ -0,0 +1,23 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const build_config = @import("../../build_config.zig");
+const internal_os = @import("../../os/main.zig");
+const glib = @import("glib");
+
+pub fn resourcesDir(alloc: Allocator, hostAccessible: bool) !?[]const u8 {
```
pluiedev, line 7: Non-snake_case variables can be confused for functions

```suggestion
pub fn resourcesDir(alloc: Allocator, host_accessible: bool) !?[]const u8 {
```

src/global.zig
```diff
@@ -170,9 +176,18 @@ pub const GlobalState = struct {
 
         // Find our resources directory once for the app so every launch
         // hereafter can use this cached value.
-        self.resources_dir = try internal_os.resourcesDir(self.alloc);
+        self.resources_dir = rd: {
+            if (@hasDecl(apprt.runtime, "resourcesDir")) break :rd try apprt.runtime.resourcesDir(self.alloc, false);
```
pluiedev, line 180: We try to avoid `@hasDecl` in new apprt-related code, although I'm not exactly sure what's a better alternative here

pluiedev (changes requested):

mitchellh (changes requested):
Its better, but I still don't love this approach...

Calling `resourcesDir` twice once for guest once for host is not great, especially since 99% of users (today, but a significant number) are not running in a sandboxed mode. Its just waste.

A perhaps better approach would be to return a struct instead so we can get all the info in one call. But stepping back its not clear to me if we ever need both, would it be more useful to redocument `resourcesDir` as always being host accessible?

alaviss:
> But stepping back its not clear to me if we ever need both

For platforms like Flatpak the host-accessible directory is usually non-accessible in the sandbox, but the sandbox path is not accessible outside.

We need sandbox-accessible paths for things like reading bundled themes, and only termio needs the host-accessible one since that's where host processes are launched. So on Flatpak you'd want both to implement all the features.

alaviss:
Changes in v5:

- Rebased to `main`
- New `ResourcesDir` struct that abstracts over host/app paths selection, removing the need to allocate two paths for non-sandboxed cases, as well as reducing the `resourcesDir` call count to 1.

src/global.zig
```diff
@@ -170,19 +171,22 @@ pub const GlobalState = struct {
 
         // Find our resources directory once for the app so every launch
         // hereafter can use this cached value.
-        self.resources_dir = try internal_os.resourcesDir(self.alloc);
-        errdefer if (self.resources_dir) |dir| self.alloc.free(dir);
+        self.resources_dir = rd: {
+            if (@hasDecl(apprt.runtime, "resourcesDir")) break :rd try apprt.runtime.resourcesDir(self.alloc);
```
pluiedev, line 175: I really don't like the use of `@hasDecl` here... Perhaps we should just force every apprt to define `resourcesDir`, and for apprts that don't do any sort of special handling, they could just declare `resourcesDir` to be an alias of `internal_os.resourcesDir`:

```zig
pub const resourcesDir = internal_os.resourcesDir;
```

pluiedev (changes requested):
Logic looks sound, just one small nitpick

src/global.zig
alaviss: Yea, resources should probably be managed by apprts. Pushed the change.

pluiedev (approved):
Much better. Thanks!

pluiedev:
IMO there isn't much that's blocking this PR right now and I'd like to get this merged for the Flatpak build to become more usable, but as this is a core change I would have to hear from someone else first. @mitchellh @jcollie you both have previously requested changes - any thoughts at the moment?

alaviss:
Rebased to solve merge conflicts.

jcollie (approved):
I don't have time to re-review right now but I'll trust @pluiedev 's judgement.

tristan957 (approved):
Looks pretty solid. I'll also leave it to Leah.

src/os/resourcesdir.zig
```diff
@@ -2,13 +2,41 @@ const std = @import("std");
 const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 
+pub const ResourcesDir = struct {
+    app_path: ?[]const u8 = null,
+    host_path: ?[]const u8 = null,
+
+    /// Free resources held. Requires the same allocator as when resourcesDir()
+    /// is called.
+    pub fn deinit(self: *ResourcesDir, alloc: std.mem.Allocator) void {
```
mitchellh: Can do just `Allocator`, we import it already.

src/os/resourcesdir.zig
```diff
@@ -2,13 +2,41 @@ const std = @import("std");
 const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 
+pub const ResourcesDir = struct {
+    app_path: ?[]const u8 = null,
+    host_path: ?[]const u8 = null,
```
mitchellh, line 8: We should put a comment above these that note callers should prefer to use the accessor functions rather than directly accessing these fields.

mitchellh (approved):
There's a couple reallyyyyy small things that would be good and quick to address.

I'm going to approve this so that we can merge once they're addressed and I don't block anyone.

mitchellh:
Actually I just quickly addressed them myself. CI running now, auto merge on!

PR #6668, opened 2025-03-11
pkg(highway): Enable system-integration

AnthonyZhOon:
Move bridge.cpp to being a module cSourceFile instead of part of the staticlib Dynamic link against highway with pkg-config name 'libhwy.pc' when system-integrated

Some concerns:
- Running `ldd zig-out/bin/ghostty` after building with `zig build -Doptimize=ReleaseFast` does not show `libhwy.so` but is included in debug optimize.
  *  The staticlib variant also compiles without source files in release mode but not in debug due to missing symbols `hwy::Abort()`. 
  * Would look into how highway works more to understand this, it looks to extensively use templates and template-specialisations which results in little shared object code?

- We configure a few optimization related flags (vectorization, constants), ran a quick DOOM-fire and cats and it seems okay
- There's a few macro defines for predefined macros (__DATE__, __TIMESTAMP__, __TIME__) that are overrided in the static lib. The comment seems to imply that they are for reproducible binaries so I think it shouldn't change runtime behaviour.

### DIFF
--- a/nix/devShell.nix
+++ b/nix/devShell.nix
@@ -40,6 +40,7 @@
   hicolor-icon-theme,
   harfbuzz,
   libpng,
+  libhwy,
   libxkbcommon,
   libX11,
   libXcursor,
@@ -141,6 +142,7 @@ in
         zlib
 
         glslang
+        libhwy
         spirv-cross
 
         libxkbcommon
--- a/pkg/highway/build.zig
+++ b/pkg/highway/build.zig
@@ -4,28 +4,15 @@ pub fn build(b: *std.Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
-    const upstream = b.dependency("highway", .{});
-
     const module = b.addModule("highway", .{
         .root_source_file = b.path("main.zig"),
         .target = target,
         .optimize = optimize,
     });
-
-    const lib = b.addStaticLibrary(.{
-        .name = "highway",
-        .target = target,
-        .optimize = optimize,
-    });
-    lib.linkLibCpp();
-    lib.addIncludePath(upstream.path(""));
-    module.addIncludePath(upstream.path(""));
-
-    if (target.result.os.tag.isDarwin()) {
-        const apple_sdk = @import("apple_sdk");
-        try apple_sdk.addPaths(b, lib.root_module);
-        try apple_sdk.addPaths(b, module);
-    }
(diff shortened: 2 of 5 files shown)


AnthonyZhOon:
Checking highway's source, it simply appears that much of the implementation is in the header files and template instantiations. It should be fine that `libhwy.so` is entirely cleaned out by link-time optimizations.

mitchellh:
Please rebase for all the latest build changes, shouldn't be much work. Thank you!

AnthonyZhOon:
I did chose to add an import module in the `SharedDeps.zig` similar to how other libs that expose a zig module over C files now that `bridge.cpp` is not compiled into the static lib. 
Although it seems the `extern "c" fn hwy_supported_targets() i64` is not used in ghostty

mitchellh:
I'm going to merge this for now and I defaulted the system integration to true (for packaging). I am suspicious that this will work in practice because I've found each highway version changes the API in ways that its usually pretty tied but we'll see!

mitchellh:
I'm not at the computer but we need to add highway as a package to our Debian 12 test.

AnthonyZhOon:
Updated the Debian 12 build env

mitchellh:
Okay yeah, this is what I expected. The dependency on Debian 12 is too old and the API isn't stable. I'm going to default the system integration to false and packagers who really care can enable it.

AnthonyZhOon:
Merged the Nix conflicts and built successfully with it

mitchellh (changes requested):
I think the merge with main was incorrect, it undoes some changes around the Nix files. I can take a look later but can't merge until then.

hqnna (changes requested):
Please rebase your branch so it includes the `po/nb_NO.UTF-8.po` file, thanks.

AnthonyZhOon:
I've taken the patch of the package and rebased it onto main but left out the `nix/` changes @mitchellh  made

hqnna (approved):
The changes look much cleaner now, and don't effect irrelevant files.

AnthonyZhOon:
Looking at the nix config, I've added libhwy to the devshell but not to the package.nix as we've decided not to have highway as a default system integration with `zig build --system`  since glslang and spirv-cross integration are similarly not in the nix package.

mitchellh:
Sorry for the long back and forth. I appreciate the iteration on this, but until a packager requests this specifically I'm going to merge forward with not merging this. This isn't something anyone has needed yet and since Highway is mostly header-only I'm not sure the complexity in implementation is worth it. Thank you though.

PR #6678, opened 2025-03-12
Fix flatpak packaging to a working state

alaviss:
This should make testing Flatpak builds a lot easier.

To build, enter `flatpak/` directory and run:

    flatpak-builder --repo=repo builddir com.mitchellh.ghostty.yml

alternatively, using org.flatpak.Builder flatpak:

    flatpak run -p org.flatpak.Builder \
      --repo=repo \
      builddir \
      com.mitchellh.ghostty.yml

The resulting flatpak can be installed using

    flatpak install ./repo com.mitchellh.ghostty

Credit of AppStream metadata goes to @yorickpeterse.


flatpak/com.mitchellh.ghostty.yml
```diff
@@ -0,0 +1,113 @@
...
+            | .browser_download_url
+          version-query: .tag_name
+
+  - name: ghostty
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/ziglang
+    build-commands:
+      - zig build
+        -Doptimize=ReleaseFast
```
yorickpeterse, line 96: Should this perhaps be `-Doptimize=Debug`, since this is used for testing purposes?

alaviss: Ideally I hope that one day this can go straight to Flathub like how OBS does it.

We can expand and have a `.Devel` variant for debugging then, similar to many GNOME apps.

yorickpeterse: @alaviss Keeping the manifest in the Ghostty repository would be nice, though this raises some questions:

- I had to use a bunch of [extra `-fno-sys` flags](https://github.com/yorickpeterse/ghostty-flatpak/blob/2757079ea59656ca8a11874d41245c90166ea54b/com.mitchellh.ghostty.yml#L50-L56) to get things to build. I'm not sure if this is still necessary, or how this applies to test builds, but I don't think Flatpak has a way of dynamically handling that?
- For the Flathub build, emitting docs makes no sense so I have that disabled, but I see it enabled here (or at least Pandoc is being installed). Is that strictly necessary?
- If the data lives in the Ghostty repository, then an actual maintainer of Ghostty would have to also maintain the Flathub package, instead of it being maintained out of tree. This isn't necessarily a problem, but it does require a decision by @mitchellh on the matter

flatpak/com.mitchellh.ghostty.yml
```diff
@@ -0,0 +1,113 @@
+app-id: com.mitchellh.ghostty
```
yorickpeterse, line 1: I had to use the following in my manifest:

```yaml
build-options:
  no-debuginfo: true
```

IIRC this was because otherwise it fails to build, though I don't remember the specifics. Is this no longer necessary?

alaviss: Yea it seemed fine when I built it.

AnthonyZhOon: Previously, the build always stripped debug in ReleaseFast, a -Dstrip=false can now be configured to make debuginfo work on release builds

flatpak/com.mitchellh.ghostty.yml
alaviss: > * I had to use a bunch of [extra `-fno-sys` flags](https://github.com/yorickpeterse/ghostty-flatpak/blob/2757079ea59656ca8a11874d41245c90166ea54b/com.mitchellh.ghostty.yml#L50-L56) to get things to build. I'm not sure if this is still necessary, or how this applies to test builds, but I don't think Flatpak has a way of dynamically handling that?

I just started with none of them, which ended up with the set of dependencies you see in the manifest. I believe the other ones are not necessary.

> * For the Flathub build, emitting docs makes no sense so I have that disabled, but I see it enabled here (or at least Pandoc is being installed). Is that strictly necessary?

I was looking at some macOS code that performs environment setup:

https://github.com/ghostty-org/ghostty/blob/95daca616db5c24d7bb37fd5a3ac2f8762bb4ead/src/termio/Exec.zig#L798-L835

We can do the same with Flatpak, which would result in ghostty man pages and certain integrations (i.e. fish) being accessible. But I can disable docs if that's not in scope.

> * If the data lives in the Ghostty repository, then an actual maintainer of Ghostty would have to also maintain the Flathub package, instead of it being maintained out of tree.

I think for regression CI to work then at least a working manifest should be in tree.

pluiedev: Honestly for me if the Snap is in-tree then it makes sense for the Flatpak to be included as well. We just need a concrete team of maintainers that would keep it working.

flatpak/com.mitchellh.ghostty.yml
alaviss: > a -Dstrip=false can now be configured to make debuginfo work on release builds

Thanks, I just added it to keep debuginfo handy for Ghostty too.

flatpak/com.mitchellh.ghostty.yml
yorickpeterse: I'd be happy to help out maintaining the build/packaging process, but for the actual Zig integration code we'd need others to help out as that's outside my area of expertise. With that said, if the Flatpak build process becomes part of CI that should be easier to deal with.

dist/linux/com.mitchellh.ghostty.metainfo.xml
```diff
@@ -0,0 +1,28 @@
...
+      feature-rich, and native. While there are many excellent terminal
+      emulators available, they all force you to choose between speed,
+      features, or native UIs. Ghostty provides all three.
+    </p>
+  </description>
+
+  <releases>
+    <release version="1.0.1" date="2024-12-31">
+        <url type="details">https://ghostty.org/docs/install/release-notes/1-0-1</url>
+    </release>
```
tristan957, line 26: Should we add all releases?

flatpak/com.mitchellh.ghostty.yml
```diff
@@ -0,0 +1,113 @@
...
+  - '*.la'
+  - '*.a'
+  - '*.so'
+
+modules:
+  - name: bzip2-redirect
+    buildsystem: simple
+    build-commands:
+      - install -dm755 /app/lib
+      - echo "INPUT(libbz2.so)" > /app/lib/libbzip2.so
```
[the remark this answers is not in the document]
tristan957: Instead of the linker script solution, can we just add a PR to check for the library name here?

https://github.com/ghostty-org/ghostty/blob/95daca616db5c24d7bb37fd5a3ac2f8762bb4ead/src/build/SharedDeps.zig#L117

flatpak/com.mitchellh.ghostty.yml
```diff
@@ -0,0 +1,114 @@
...
+  - name: gtk4-layer-shell
+    buildsystem: meson
+    sources:
+      - type: git
+        url: https://github.com/wmww/gtk4-layer-shell.git
+        tag: v1.1.0
+        commit: 93550245220cdc514be4701b517acd374a86acc2
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)$
```
tristan957, line 60: @pluiedev @jcollie how can we best make sure that these are matching the build.zig.zon or where ever else we track these? A CI job? Generating the manifest? CI job sounds better than generating, but that's just me opinion.

flatpak/com.mitchellh.ghostty.yml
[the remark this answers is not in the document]
alaviss: That's not feasible I believe. Many distributions (incl. homebrew) provides a pkg-config for `bzip2`, but none for `bz2`, so it is a good idea to keep it as is.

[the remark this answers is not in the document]
tristan957: The Zig build system doesn't allow you to pass the equivalent of `-lbz2`?

flatpak/com.mitchellh.ghostty.yml
tristan957: > We can expand and have a .Devel variant for debugging then, similar to many GNOME apps.

I think this is a good idea.

flatpak/com.mitchellh.ghostty.yml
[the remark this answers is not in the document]
alaviss: It doesn't look like alternatives (e.g. try `bz2` if `bzip2` not available) are possible to me at least.

flatpak/com.mitchellh.ghostty.yml
alaviss: Added `.Devel` variant.

flatpak/com.mitchellh.ghostty.Devel.yml
```diff
@@ -0,0 +1,63 @@
+app-id: com.mitchellh.ghostty.Devel
+runtime: org.gnome.Platform
+runtime-version: "47" # maybe use gnome-nightly master?
```
tristan957, line 3: I'm down for this pending what @jcollie and @pluiedev think

jcollie: Hmm, if developing Gnome/GTK were the primary object sure, but since this is about developing Ghostty we should probably stick to a released version of Gnome. 48 is very close to release so if there was like a 48-rc channel I'd be OK with that.

pluiedev: I agree. 48-rc/beta/etc if possible, 47 otherwise

yorickpeterse: If the goal is to reuse this manifest for Flathub, using RCs might be a problem as one wouldn't want a stable release of the Flatpak package to depend on RC/beta releases.

It might be better to just depend on a stable release and bump it every 3-6 months or something along those lines.

alaviss: This one is a development manifest (`.Devel`), I don't think it will ever make it Flathub.

The regular one will still use stable.

I guess with the current consensus I'll remove the comment since GNOME doesn't publish rc runtimes.

yorickpeterse: Oops, I completely missed the `.Devel` bit. Sorry for the noise! :man_facepalming:

dist/linux/com.mitchellh.ghostty.metainfo.xml
AnthonyZhOon: There was a PR for the appstream metainfo file with some discussion didn't merge  #5125 https://github.com/ghostty-org/ghostty/commit/1c70ba4b9252c0671428ccb07ab8ec4222fe475c

alaviss: Added more data based on the linked PR.

Some data was like `<categories>` or `<keywords>` are removed as they will be [pulled from `.desktop` file](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#categories-and-keywords)

dist/linux/com.mitchellh.ghostty.metainfo.xml
alaviss, line 1: Here's how Ghostty shows up in GNOME Software with this metainfo:

image

image

AnthonyZhOon: I believe we want to direct "Report an Issue" to our github discussions at https://github.com/ghostty-org/ghostty/discussions, looks good otherwise.

alaviss: Done

flatpak/com.mitchellh.ghostty.yml
jcollie: Can we use plain `https` downloads rather than git checkouts? If so it would be fairly easy to keep those in sync when we update the Nix deps. See `build.zig.zon.txt`

jcollie: hmm yeah, i think that the plan is to make blueprint compiler optional for packagers by precompiling all of the blueprints to .ui files. then unless you're really hardcore you wouldn't need blueprint compiler at all

alaviss: gtk4-layer-shell is now using archive.

I'd prefer that we get the installation bug fixed so I can just use the bundled version, but I haven't figured out how to make it work.

flatpak/dependencies.yml
tulilirockz, line 1: You probably want to put these on the main flatpak, flathub reviewers seem to hate having a separate file for this

alaviss: I'm not fond of duplicating the dependencies definition across two Flatpaks.

I think we will cross that bridge when we get there.

alaviss:
Rebased to Zig 0.14.0.

Due to https://github.com/ghostty-org/ghostty/discussions/6730, `flatpak-builder` will not be able to build Ghostty at the moment.

alaviss:
Rebased to `main`, flatpak build is now fully functional.

dist/linux/com.mitchellh.ghostty.metainfo.xml
```diff
@@ -0,0 +1,59 @@
...
+
+  <!--
+    TODO: Generate manifest location data.
+
+    Ref: https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#manifest-location
+
+  <custom>
+    <value key="flathub::manifest">https://github.com/ghostty-org/ghostty/blob/<hash>/flatpak/com.mitchellh.ghostty.yml</value>
+  </custom>
+  -->
```
tulilirockz, line 51: This isnt enforced by the flathub review process. At least I havent ran into this at all when publishing flatpak manifests to Flathub

alaviss: It sounds like a requirement if you're publishing straight from here to Flathub, though. At least I see that Firefox and OBS are following it.

mitchellh:
I'm trying to verify this so we can get it merged but I'm getting build errors locally:

```
Cache miss, checking out last cache hit
========================================================================
Building module ghostty in /home/mitchellh/code/go/src/github.com/ghostty-org/ghostty/flatpak/.flatpak-builder/build/ghostty-2
========================================================================
Updated Git hooks.
Git LFS initialized.
Note: switching to '4182b7fa42f27cf14a71dbdb54cfd82c5c6e3447'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 4182b7f switch zigimg ref to merged commit
Updated Git hooks.
Git LFS initialized.
Note: switching to '4a002763419a34d61dcbb1f415821b83b9bf8ddc'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 4a00276 Merge pull request 'fix removed b.host on v0.14.0-dev.2568+42dac40b3' (#25) from e0328eric/zg:master into master
Updated Git hooks.
Git LFS initialized.
fatal: unable to read tree (0ce4eca3560d5553b13263d6b6bb72e146dd43d0)
Error: module ghostty: Child process exited with code 128
```

Any ideas?

alaviss:
> fatal: unable to read tree (0ce4eca3560d5553b13263d6b6bb72e146dd43d0)

Looks like it's zigimg again.

They force pushed their main branch so now the commit referenced is unreachable from default branch.

Maybe I'll have to fix the cache gen script to prefer downloading archives instead...

alaviss:
Rebased and fixed the issue with zigimg. The flatpak should build correctly now.

alaviss:
Rebased to `main` and updated package cache.

alaviss:
Rebased to `main` and dropped `zigimg` workaround, given that the fork's author [committed to stop force pushing](https://github.com/ghostty-org/ghostty/issues/6734#issuecomment-2762961408).

yorickpeterse:
Is there anything that's holding back this PR (or reviewing of it)? :smiley:

mitchellh:
Okay hurray the build works for me. I want to add some CI checks next and we should be good. CI may be annoying though lets see!

tulilirockz:
Flathub recently moved to github workflows for CI, maybe there would be something useful there?

mitchellh:
Okay, a couple important CI failures and I'm not quite sure what they are yet:

- For some reason `check-zig-cache.sh` works on my local system (with Nix) but not in CI. 
- Seeing a strange failure about lazy dependencies here: https://github.com/ghostty-org/ghostty/actions/runs/14601245977/job/40959653302

If anyone has ideas, that's blocking this, because we need CI to make sure this doesn't bitrot.

alaviss:
> For some reason `check-zig-cache.sh` works on my local system (with Nix) but not in CI.

It's correct to fail, the lock file in `main` changed between the last rebase and now.
You'd have to merge `main` into this and update the deps.

mitchellh:
Thanks, CI looks good! I'm going to polish up the runs and dependencies a bit and we should b good

mitchellh (approved):
Thanks for sticking with this PR. With the Flatpak build being verified in CI now this should help keep everything much more stable. 😄

PR #6688, opened 2025-03-13
feat: add appstream metadata for linux packages

tulilirockz:
This is a standard file for metadata used by the snap store and flathub for getting metadata for your application. I wanted to submit this upstream as per flathub policies we need this metadata file on the upstream repository and not on the flathub repo

This is currently only missing screenshots, everything else should be fine, I believe.


00-kat:
#6678?

tulilirockz:
:OOOOO I didnt see that PR! This is obsolete then! :P Awesome!

PR #6701, opened 2025-03-13
global: use DebugAllocator/SmpAllocator instead of GeneralPurposeAllocator/libc allocator

jcollie:
Zig 0.14 introduces the SmpAllocator which claims performance on par with or potentially better than the glibc allocator. The GeneralPurposeAllocator has also been renamed to DebugAllocator. This PR updates the global allocator to use DebugAllocator or SmpAllocator depending on if runtime safety is enabled.


mitchellh:
I need many, many benchmarks before I can even consider merging this.

mitchellh:
A separate PR to rename GPA to the debug allocator would be an easy merge though.

mitchellh:
Closing this until further information is acquired.

PR #6706, opened 2025-03-14
deps: Default gtk4-layer-shell system integration to true

### LINKED ISSUE #6632: gtk4-layer-shell shared lib needs to be installed when not linking external library
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6631

<sup>Originally posted by **vaskark** March  8, 2025</sup>
Hi. I build ghostty on Ubuntu 24.10 and usually delete .zig-cache (along with .cache/zig) after a successful build. Ghostty usually runs just fine. But the past few days I've been getting this error:

`ghostty: error while loading shared libraries: libgtk4-layer-shell.so: cannot open shared object file: No such file or directory`

I suppose I just need to keep .zig-cache. But is there any reason why this has changed all the sudden?

Thanks.</div>

AnthonyZhOon:
Closes #6632 
When compiling the dynamic lib and linking, the rpath resolves to the compile cache location instead of the install location for the lib. This resulted in loading the dylib failing when the compile cache was removed or the install location is changed.

Based on https://github.com/ziglang/zig/issues/5827 , we specify the rpath to search relative to the executable.
Previous state:
- dynamic lib was not installed to the output directory -> now appears in `$OUT/lib/libgtk4-layer-shell.so`
- rpath only included the compile cache location, not the install location. 
Before:
```sh
$ patchelf --print-rpath zig-out/bin/ghostty
/home/anthony/dev/ghostty-pure/.zig-cache/o/0254fb4753185c5429180337a720248d
$ ldd zig-out/bin/ghostty
        ...
        libgtk4-layer-shell.so => /home/anthony/dev/ghostty-pure/.zig-cache/o/0254fb4753185c5429180337a720248d/libgtk4-layer-shell.so (0x00007f7975468000)
        ...
$  ldd zig-out/bin/ghostty
        ...
        libgtk4-layer-shell.so => not found
        ...
```

After:

```sh
$ zig build
$ patchelf --print-rpath zig-out/bin/ghostty
/home/anthony/dev/ghostty/.zig-cache/o/f45360ddde653cb3bc70966c326dd96d:$ORIGIN/../lib/
$ ldd zig-out/bin/ghostty
        ...
        libgtk4-layer-shell.so => /home/anthony/dev/ghostty/.zig-cache/o/f45360ddde653cb3bc70966c326dd96d/libgtk4-layer-shell.so (0x00007fbf81ad8000)
        ...
 $ rm -r .zig-cache/
 $ ldd zig-out/bin/ghostty
        ...
        libgtk4-layer-shell.so => /home/anthony/dev/ghostty/zig-out/bin/../lib/libgtk4-layer-shell.so (0x00007f60dc087000)
        ...
```

### DIFF
--- a/src/build/SharedDeps.zig
+++ b/src/build/SharedDeps.zig
@@ -652,14 +652,13 @@ fn addGTK(
             // IMPORTANT: gtk4-layer-shell must be linked BEFORE
             // wayland-client, as it relies on shimming libwayland's APIs.
             if (b.systemIntegrationOption("gtk4-layer-shell", .{})) {
-                step.linkSystemLibrary2(
-                    "gtk4-layer-shell-0",
-                    dynamic_link_opts,
-                );
+                step.linkSystemLibrary2("gtk4-layer-shell-0", dynamic_link_opts);
             } else {
                 // gtk4-layer-shell *must* be dynamically linked,
                 // so we don't add it as a static library
-                step.linkLibrary(gtk4_layer_shell.artifact("gtk4-layer-shell"));
+                const shared_lib = gtk4_layer_shell.artifact("gtk4-layer-shell");
+                b.installArtifact(shared_lib);
+                step.linkLibrary(shared_lib);
             }
         }
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -81,9 +81,6 @@ parts:
       craftctl set version=$(cat VERSION)
       $CRAFT_PART_SRC/../../zig/src/zig build -Dpatch-rpath=\$ORIGIN/../usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR:/snap/core24/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR -Doptimize=ReleaseFast -Dcpu=baseline -fno-sys=gtk4-layer-shell
       cp -rp zig-out/* $CRAFT_PART_INSTALL/
-      # Install libgtk4-layer-shell.so
-      mkdir -p $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
-      cp .zig-cache/*/*/libgtk4-layer-shell.so $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/
       sed -i 's|Icon=com.mitchellh.ghostty|Icon=${SNAP}/share/icons/hicolor/512x512/apps/com.mitchellh.ghostty.png|g' $CRAFT_PART_INSTALL/share/applications/com.mitchellh.ghostty.desktop
 
   libs:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -76,9 +76,10 @@ parts:
       - git
       - patchelf
       - gettext
+    # TODO: Remove -fno-sys=gtk4-layer-shell when we upgrade to a version that packages it Ubuntu 24.10+
     override-build: |
       craftctl set version=$(cat VERSION)
-      $CRAFT_PART_SRC/../../zig/src/zig build -Dpatch-rpath=\$ORIGIN/../usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR:/snap/core24/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR -Doptimize=ReleaseFast -Dcpu=baseline
+      $CRAFT_PART_SRC/../../zig/src/zig build -Dpatch-rpath=\$ORIGIN/../usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR:/snap/core24/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR -Doptimize=ReleaseFast -Dcpu=baseline -fno-sys=gtk4-layer-shell
       cp -rp zig-out/* $CRAFT_PART_INSTALL/
       # Install libgtk4-layer-shell.so
       mkdir -p $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
--- a/src/build/Config.zig
+++ b/src/build/Config.zig
@@ -361,7 +361,6 @@ pub fn init(b: *std.Build) !Config {
             "libpng",
             "zlib",
             "oniguruma",
-            "gtk4-layer-shell",
         }) |dep| {
             _ = b.systemIntegrationOption(
                 dep,
@@ -386,6 +385,15 @@ pub fn init(b: *std.Build) !Config {
             "glfw3",
         }) |dep| {
             _ = b.systemIntegrationOption(dep, .{ .default = false });
+        }
+
+        // These are dynamic libraries we default to true, preferring
+        // to use system packages over building and installing libs
+        // as they require additional ldconfig of library paths or
+        // patching the rpath of the program to discover the dynamic library
+        // at runtime
+        for (&[_][]const u8{"gtk4-layer-shell"}) |dep| {
+            _ = b.systemIntegrationOption(dep, .{ .default = true });
         }
     }
 


AnthonyZhOon:
Dissatified by:
- Potentially conflicts with our `patchelf --set-rpath` step in https://github.com/ghostty-org/ghostty/blob/73c7943fff38f679a9a434457b5089bc5722411d/src/build/GhosttyExe.zig#L36 . I should setup Nix and test if this can be converted to an addRPathSpecial() as well. 
- This mixes adding install steps into the deps.add() function
- We don't default to preferring a system-installed gtk4-layer-shell dylib.

AnthonyZhOon:
Open issues in Zig related to RPath handling in the build system https://github.com/ziglang/zig/issues/17373

src/build/SharedDeps.zig
```diff
@@ -510,7 +510,12 @@ pub fn add(
                     } else {
                         // gtk4-layer-shell *must* be dynamically linked,
                         // so we don't add it as a static library
-                        step.linkLibrary(gtk4_layer_shell.artifact("gtk4-layer-shell"));
+                        const sharedLib = gtk4_layer_shell.artifact("gtk4-layer-shell");
+                        const artifact: *std.Build.Step.InstallArtifact = b.addInstallArtifact(sharedLib, .{});
+                        b.getInstallStep().dependOn(&artifact.step);
+                        // Lookup dynamic libs from installed location
+                        step.root_module.addRPathSpecial("$ORIGIN/../lib/");
```
mitchellh, line 517: I think unconditionally doing this is probably a little scary, but I'm not fully sure. Qualifying this as a debug-only thing would make me feel better but I know that part of the issue is people that are building from source into a prefix too... I'm just not sure. Is there any prior art we can look at?

jwbowen:
FWIW, I applied this on top of 73c7943f on origin/main and can confirm that after removing `.zig-cache` it points to the system library:

```
[jbowen@bender ~]$ ldd /usr/local/bin/ghostty | grep libgtk4-layer
        libgtk4-layer-shell.so => /usr/local/bin/../lib/libgtk4-layer-shell.so (0x0000731136135000)
[jbowen@bender ~]$
```

I don't feel qualified to comment on any other aspects of it, but for now I'm going to carry the patch around to survive reboots.

src/build/SharedDeps.zig
AnthonyZhOon: Right, this is bad default behaviour as it can override the library path for other libs (rpath has higher priority than `lib/` dirs). The alternative is ld-config[^3] or `LD_LIBRARY_PATH`[^1]. In fact, Fedora explicitly bans rpath from packages[^fedora].
This dependency shouldn't be considered system-integration, as we're not self-contained with it disabled.

I believe appropriate way to install into a prefix would be adding the prefix's `lib/` to the `LD_LIBRARY_PATH` variable,  I found you can hardcode the installed lib's absolute path [^2] as a debug-only option and avoid potentially overriding other libs but breaking when installed to a prefix where the lib is not in a default `lib/` dir or on the `LD_LIBRARY_PATH`.

We should make systemLink() the default and add `gtk4-layer-shell-devel` to build depedencies, most usescases should be unaffected and an FAQ page can be made to describe configuring library paths, the error produced from shared lib not found is searchable.

[^1]: The man page ld.so which describes loading dynamic libs https://www.man7.org/linux/man-pages/man8/ld.so.8.html
[^2]: When resolving shared object dependencies, the dynamic linker
       first inspects each dependency string to see if it contains a
       slash (this can occur if a shared object pathname containing
       slashes was specified at link time).  If a slash is found, then
       the dependency string is interpreted as a (relative or absolute)
       pathname, and the shared object is loaded using that pathname.[^1]
[^3]: ldconfig man page https://man7.org/linux/man-pages/man8/ldconfig.8.html
[^fedora]: Beware of rpath https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath

pluiedev (approved):
Makes sense pending the linker path comment

jwbowen:
The patch no longer cleanly applies after cfea2ea12.

I have the library in `$PREFIX/lib/` after doing an install with this patch a few days ago, so I'm able to "survive" the loss of `$BUILDDIR/.zig-cache/`, but I won't be getting new copies of  the library installed after building.

00-kat:
> I have the library in `$PREFIX/lib/` after doing an install with this patch a few days ago, so I'm able to "survive" the loss of `$BUILDDIR/.zig-cache/`, but I won't be getting new copies of the library installed after building.

@jwbowen Couldn't you stick to an older commit (i.e.: not upgrade for a bit) until this PR is merged?

jwbowen:
> > I have the library in `$PREFIX/lib/` after doing an install with this patch a few days ago, so I'm able to "survive" the loss of `$BUILDDIR/.zig-cache/`, but I won't be getting new copies of the library installed after building.
> 
> @jwbowen Couldn't you stick to an older commit (i.e.: not upgrade for a bit) until this PR is merged?

I certainly _could_, but I really like riding the bleeding edge where I can and am willing to go out of my way to make it work. I'm not concerned with practicality in this case :)

I should have been clearer in my last message that I wasn't _complaining_ about the patch, I just wanted to give a heads up/FYI.

AnthonyZhOon:
Should I leave the install step config at https://github.com/ghostty-org/ghostty/pull/6706/files#diff-3c27f71e7b0ef00072a7d994ae2af66c62b15fc7301fb43aad52bcd96a64da24R653-R654 or is there a better place to organise it?

At this point, an absolute rpath link is made in Debug mode but we default to system integration for `gtk4-layer-shell` so a source compile will give a system package not found error.
```
> ~/zig/0.14.0/files/zig build
install
└─ install ghostty
   └─ zig build-exe ghostty Debug native failure
error: error: unable to find dynamic system library 'gtk4-layer-shell-0' using strategy 'mode_first'. searched paths:
  /usr/local/lib64/libgtk4-layer-shell-0.so
  /usr/local/lib/libgtk4-layer-shell-0.so
  /lib64/libgtk4-layer-shell-0.so
  /lib/libgtk4-layer-shell-0.so
  /usr/lib64/libgtk4-layer-shell-0.so
  /usr/lib/libgtk4-layer-shell-0.so
  /usr/local/lib64/libgtk4-layer-shell-0.a
  /usr/local/lib/libgtk4-layer-shell-0.a
  /lib64/libgtk4-layer-shell-0.a
  /lib/libgtk4-layer-shell-0.a
  /usr/lib64/libgtk4-layer-shell-0.a
  /usr/lib/libgtk4-layer-shell-0.a
  ```
Otherwise I think this is in a working state where docs for building on Debian 12 (missing gtk4-layer-shell) need `-fno-sys=gtk4-layer-shell` and should configure custom prefix `lib/` paths via `ldconfig`.

File: src/build/SharedDeps.zig [hunk 2]
```diff
@@ -644,14 +644,23 @@ fn addGTK(
...
-                    dynamic_link_opts,
-                );
+                step.linkSystemLibrary2("gtk4-layer-shell-0", dynamic_link_opts);
             } else {
                 // gtk4-layer-shell *must* be dynamically linked,
                 // so we don't add it as a static library
-                step.linkLibrary(gtk4_layer_shell.artifact("gtk4-layer-shell"));
+                const sharedLib = gtk4_layer_shell.artifact("gtk4-layer-shell");
+                const artifact: *std.Build.Step.InstallArtifact = b.addInstallArtifact(sharedLib, .{});
+                b.getInstallStep().dependOn(&artifact.step);
```
mitchellh, line 653: This shouldn't be necessary, since the `step` below will ensure it, no?

src/build/SharedDeps.zig
```diff
@@ -644,14 +644,23 @@ fn addGTK(
...
+                step.linkLibrary(sharedLib);
+                if (self.config.optimize == .Debug) {
+                    // Lookup dynamic libs from installed location
+                    const installPath = b.getInstallPath(artifact.dest_dir.?, "");
+                    if (!std.fs.path.isAbsolute(installPath)) {
+                        // Using a relative path as an rpath is almost useless
+                        return error.UndesirableRelativePath;
+                    }
+                    step.root_module.addRPathSpecial(b.getInstallPath(artifact.dest_dir.?, ""));
```
mitchellh, line 662: This can use `installPath`.

Also, our variables should be underscored `install_path`

mitchellh (changes requested):
Thanks for plugging away at this. A couple points of feedback.

I think this is looking _fine_. I still want to consider the release packaging implications. I think this will fix debug builds just fine but release builds seem like there's going to be issues and I wonder if we need to reintroduce a way to wholesale disable gtk4-layer-shell from some builds... (don't bother in this PR)

src/build/SharedDeps.zig
AnthonyZhOon: Without adding the dependency on the install step, `zig build -fno-sys=gtk4-layer-shell` doesn't install the shared lib into `zig-out/lib`. It's documented for `b.addInstallArtifact()` that it does not add itself to the top-level dependencies, this is done by `b.installArtifact()` but it doesn't return the step which I use for the install path.

src/build/SharedDeps.zig
AnthonyZhOon: Fixed

AnthonyZhOon:
Checking in on this, adding the RPath config for an installed library only in debug mode for development seems to be no improvement over relying on Zig's default behaviour of placing the RPath of the build cache's version of the library. 
It seems simpler to just default system integration for the majority of cases and correctly fix the missing installed library so users unable to system integrate can handle library path configuration on their own.

I think Zig defaults this RPath config to cache objects to make build script executables work, though forgot the source.

AnthonyZhOon:
Defaulting this to true should have implications on our snap build, flatpak build and nix develop to have gtk4-layer-shell system installed

pluiedev (approved):
Diff looks clean, thanks! If I understands correctly this would close #6632, right?

mitchellh:
I'm down to give this a try. Its a good time before release to merge this, put our ear on the ground and see if anything pops up.

PR #6708, opened 2025-03-14
Replace mach-glfw with pkg/glfw

### LINKED ISSUE #6702: Vendor Zig GLFW bindings
The GLFW bindings we depend on [are no longer maintained](https://github.com/thedeadtellnotales/mach-glfw). Even before being unmaintained, we've been pointing to a fork I maintain for a very long time. The fork also points to another contributors fork for another dependency. It's all a mess and we need to wrangle it in.

- [ ] Move mach-glfw into `vendor/`
- [ ] Move glfw (Zig C build) into `vendor/`
- [ ] Migrate Xcode/Vulkan/Wayland headers to our blob storage at least, if not remove them entirely by depending on system headers since we don't need cross-compiling support

mitchellh:
Closes #6702

This removes our mach-glfw dependency and replaces it with an in-tree pkg/glfw that includes both the source for compiling glfw as well as the Zig bindings. This matches the pattern from our other packages.

This is based on the upstream mach-glfw work and therefore includes the original license and copyright information.

The reasoning is stated in the issue but to summarize for the commit:

  - mach-glfw is no longer maintained, so we have to take ownership
  - mach-glfw depended on some large blobs of header files to enable cross-compilation but this isn't something we actually care about, so we can (and do) drop the blobs
  - mach-glfw blobs were hosted on mach hosts. given mach-glfw is unmaintained, we can't rely on this hosting
  - mach-glfw relied on a "glfw" package which was owned by another person to be Zig 0.14 compatible, but we no longer need to rely on this
  - mach-glfw builds were outdated based on latest Zig practices


mitchellh:
Most of the diff are vendored wayland headers. It'd be nicer to use Wayland-scanner to generate them as part of the build but GLFW just isn't important enough for us to do this now. If someone wanted to follow up with it I'd be fine with it.

PR #6715, opened 2025-03-14
feat: source local .envrc if it exists

Uzaaft:


00-kat (approved):

.envrc
```diff
@@ -4,3 +4,5 @@ if has nix; then
   watch_file nix/{devShell,package,wraptest}.nix
   use flake
 fi
+
+[ -f .envrc.local ] && source_env .envrc.local
```
pluiedev, line 8: envrc files are bash scripts so we're free to use a bashism here
```suggestion
[[ -f .envrc.local ]] && source_env .envrc.local
```

pluiedev (approved):

.envrc
Uzaaft: > envrc files are bash scripts so we're free to use a bashism here

Is that guaranteed? I have seen people add in `/bin/sh` to the top of .envrc

pluiedev: They shouldn't :)

image

Uzaaft: https://direnv.net/ as a reference. Lets use bash fuckery.

00-kat: Oh nice in that case double brackets should be fine then.  Not like it changes anything.
<sup>It was mostly my fault for [getting Uzaaft to change that](https://github.com/ghostty-org/ghostty/compare/baeb5c8ffda353924bdd2587e1816e25c2bf855c..53847a42bfc06e265822941afaad2dd888aa4191)... lol.</sup>

pluiedev (approved):

.envrc
```diff
@@ -4,3 +4,6 @@ if has nix; then
   watch_file nix/{devShell,package,wraptest}.nix
   use flake
 fi
+
+# Source an `.envrc.local` if it exists. You can add your custom changes there.
+[[ -f .envrc.local ]] && source_env .envrc.local
```
jparise, line 9: There is [source_env_if_exists](https://direnv.net/man/direnv-stdlib.1.html#codesourceenvifexists-ltfilenamegtcode) for this case, too.

PR #6717, opened 2025-03-14
fix: Use builtin `source_env_if_exists` for sourcing `.envrc.local`

Uzaaft:


pluiedev (approved):
Might as well add `.envrc.local` to `.gitignore` while you're at it

pluiedev (approved):

PR #6726, opened 2025-03-14
build: mark most dependencies as lazy

### LINKED ISSUE #6703: Convert a bunch of dependencies to lazy
We should mark a lot more of our dependencies to "lazy." There's no reason for Linux to download macOS dependencies and vice versa. This should be a pretty easy task, but just need to make sure that our CI remains green.

mitchellh:
Closes #6703 

Lazy dependencies are only fetched if the build script would actually reach a usage of that dependency at runtime (when the `lazyDependency` function is called). This can save a lot of network traffic, disk uage, and time because we don't have to fetch and build dependencies that we don't actually need.

Prior to this commit, Ghostty fetched almost everything for all platforms and configurations all the time. This commit reverses that to fetching almost nothing until it's actually needed.

There are very little downsides to doing this[1]. One downside is `zig build --fetch` doesn't fetch lazy dependencies, but we don't rely on this command for packaging and suggest using our custom shell script that downloads a cached list of URLs (`build.zig.zon.txt`).

This commit doesn't cover 100% of dependencies, since some provide no benefit to make lazy while the complexity to make them lazy is higher (in code style typically).

Conversely, some simple dependencies are marked lazy even if they're almost always needed if they don't introduce any real complexity to the code, because there is very little downside to do so.

[1]: https://ziggit.dev/t/lazy-dependencies-best-dependencies/5509/5

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -10,30 +10,36 @@
             // mitchellh/libxev
             .url = "https://github.com/mitchellh/libxev/archive/3df9337a9e84450a58a2c4af434ec1a036f7b494.tar.gz",
             .hash = "libxev-0.0.0-86vtc-ziEgDbLP0vihUn1MhsxNKY4GJEga6BEr7oyHpz",
+            .lazy = true,
         },
         .vaxis = .{
             // rockorager/libvaxis
             .url = "git+https://github.com/rockorager/libvaxis#1e24e0dfb509e974e1c8713bcd119d0ae032a8c7",
             .hash = "vaxis-0.1.0-BWNV_MHyCAARemSCSwwc3sA1etNgv7ge0BCIXspX6CZv",
+            .lazy = true,
         },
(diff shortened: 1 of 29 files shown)


mitchellh:
Note there is a build issue for GLFW on macOS with a clean build but this affects main as well so it shouldn't delay this PR. I'm trying to fix this now (on main):

```
install
└─ install ghostty
   └─ zig build-exe ghostty Debug aarch64-macos.13.0 2 errors
pkg/macos/foundation/c.zig:1:15: error: C import failed: FileNotFound
pub const c = @cImport({
              ^~~~~~~~
pkg/macos/text/c.zig:1:15: error: C import failed: FileNotFound
pub const c = @cImport({
              ^~~~~~~~
error: the following command failed with 2 compilation errors:
```

tristan957 (approved):

PR #6728, opened 2025-03-14
update zon2nix

jcollie:
Upstream is now mostly pure Zig and the build.zig.zon.* files are generated directly by zon2nix. The JSON file is no longer used as an intermediate file but is retained for downstream packager usage.


alaviss:
Do you mind keeping the JSON file? It's very helpful for automatically generating lists of required artifacts for the build without requiring new tooling to be built. I'm using it to generate one for Flatpak in #6678 at the moment.

jcollie:
> Do you mind keeping the JSON file? It's very helpful for automatically generating lists of required artifacts for the build without requiring new tooling to be built. I'm using it to generate one for Flatpak in #6678 at the moment.

Done, except that the file name is changed.

mitchellh (approved):

PR #6743, opened 2025-03-15
add zh-Hans translations for macOS app

Aaron-212:
macOS side of #2357

This pull request includes several changes to various `.swift` files Main Menu's `.xib` to add localization support.


mitchellh:
I'm going to close this for now. We're not going to tackle macOS localization quite yet. Maybe for 1.3 but this may get pushed a bit. Thank you for starting though!

PR #6750, opened 2025-03-16
scroll: translate non-precision to precision

rockorager:
Some wheel mice are capable of reporting fractional wheel ticks. These
mice don't necessarily report a corresponding precision scroll start
event, at least in Wayland + GTK. We can treat all discrete (ie
non-precision) events as the number of wheel ticks - for wheel mice,
yoff will be "1.0" per tick, while precision wheel mice may report
fractional values. This unifies handling of scroll events by normalizing
all events to "pixels to scroll".

We now report `mouse-scroll-multiplier` wheel or arrow events per wheel
tick (or per accumulated cell height). This means that applications
which subscribe to mouse button events will receive (by default) three
wheel events per wheel tick. For precision scrolls, they will receive
one wheel tick per line of scroll. In my opinion, this provides the best
user experience while also allowing customization of how much a
wheel tick should scroll

Reference: https://github.com/ghostty-org/ghostty/discussions/6677

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2325,13 +2325,6 @@ const ScrollAmount = struct {
     pub fn magnitude(self: ScrollAmount) usize {
         return @abs(self.delta);
     }
-
-    pub fn multiplied(self: ScrollAmount, multiplier: f64) ScrollAmount {
-        const delta_f64: f64 = @floatFromInt(self.delta);
-        const delta_adjusted: f64 = delta_f64 * multiplier;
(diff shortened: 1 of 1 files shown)


jcollie (approved):
Compiles and I can't tell a difference between before & after, but then I don't have a precision scroll wheel. But the theory looks sound.

mitchellh:
Feels fine over here, same boat as @jcollie 😄

mitchellh:
Thanks!

PR #6759, opened 2025-03-16
feat: add Ukrainian translations and locale support

danulqua:
Added the Ukrainian `uk_UA` translations.


src/os/i18n.zig
```diff
@@ -26,6 +26,7 @@ const log = std.log.scoped(.i18n);
 pub const locales = [_][:0]const u8{
     "de_DE.UTF-8",
     "zh_CN.UTF-8",
+    "uk_UA.UTF-8",
```
jcollie, line 29: I'd recommend keeping these sorted according to their ASCII ordering.

pluiedev: Actually the order does matter. See the doc comment above

pluiedev (commented):
I'm not sure why I got tagged for review (I know pretty much nothing about Ukrainian) but at a glance nothing seems particularly wrong. The strings may be a bit too long, though — have you tested your translations within Ghostty and see how much space the strings take up? (It's especially crucial for context menus where one ideally shouldn't have the menu covering half of the screen.)

danulqua:
@pluiedev I'm on macOS, and I heard that translations are not yet available here, so I can't test it. However, I was referring to other terminal apps and to the Ukrainian language in macOS in general - it's okay that some strings may be longer than in English. I also tried to keep them the same length in cases where multiline strings are being used as you can see.

mitchellh:
Do you accept the responsibility to be on the `uk_UA` localization team and get pinged for future string changes?

gordonbondon:
I've been working on Ukrainian translation as well. But @danulqua beat me to it : )

If you need more than one person to maintain Ukrainian translation I would be glad to help.

danulqua:
@varyform

> Ukrainian does not have capitalization like here: "Відкрити Конфігурацію".

Translation have capitalization since original text is also capitalized.

> "Перезавантажено конфігурацію" (which I would also change to "Конфігурацію перезавантажено").

I'll update this one, thanks.

pluiedev:
image

Just FYI, GNOME Console doesn't do title case either and we should follow what it does

danulqua:
@pluiedev I checked the macOS Terminal app in Ukrainian, and indeed, it also isn't using title case. I just thought that we should preserve the case from the original text. Now I will keep that in mind.

I pushed a fix that removes title casing.

mitchellh (approved):
Thanks for addressing feedback. Updated the code owners file and fixed some merge issues.

po/uk_UA.UTF-8.po
```diff
@@ -0,0 +1,268 @@
...
+msgid "Cancel"
+msgstr "Відмінити"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:10
+msgid "OK"
+msgstr "ОК"
+
+#: src/apprt/gtk/ui/1.5/config-errors-dialog.blp:5
+msgid "Configuration Errors"
+msgstr ""
```
mitchellh, line 40: This needs a translation

po/uk_UA.UTF-8.po
```diff
@@ -0,0 +1,268 @@
...
+
+#: src/apprt/gtk/ui/1.5/config-errors-dialog.blp:6
+msgid ""
+"One or more configuration errors were found. Please review the errors below, "
+"and either reload your configuration or ignore these errors."
+msgstr ""
```
mitchellh, line 46: This needs a translation

po/uk_UA.UTF-8.po
```diff
@@ -0,0 +1,268 @@
...
+
+#: src/apprt/gtk/ui/1.5/config-errors-dialog.blp:9
+msgid "Ignore"
+msgstr ""
```
mitchellh, line 50: This needs a translation

po/uk_UA.UTF-8.po
```diff
@@ -0,0 +1,268 @@
...
+msgid ""
+"Pasting this text into the terminal may be dangerous as it looks like some "
+"commands may be executed."
+msgstr ""
+"Вставка цього тексту в термінал може бути небезпечною, оскільки виглядає "
+"так, ніби деякі команди можуть бути виконані."
+
+#: src/apprt/gtk/inspector.zig:144
+msgid "Ghostty: Terminal Inspector"
+msgstr ""
```
mitchellh, line 205: This needs a translation

mitchellh (changes requested):
I found some blank translations

mitchellh (approved):

PR #6762, opened 2025-03-16
gtk: autogenerate GTK resource file from blueprints

### LINKED ISSUE #6760: Generate blueprint-compiler artifacts as part of source tarball generation
We should run the blueprint-compiler as part of our source tarball generation and remove it from the list of requirements to build from released sources. It's hard to ensure consistent toolchain versions like this for packagers and just idiomatically unnecessary.

Source tarballs generally should only depend on a working compiler and set of libraries to link to. Preprocessor steps aren't usually part of source tarballs.

This has been pretty standard Unixisms for decades but as a concrete example I found recently: Gnu gettext runs yacc prior to source ball generation so you only need a C compiler.

jcollie:
Fixes #6760

Eliminate the need for `blueprint-compiler` during build by generating all GTK resource definitions (`.ui` files) from Blueprint files (`.blp` files) during development and checking them into the source repository.

This reduces the number of build dependences needed by people not actively developing the GTK apprt, as well as work around limitations in distros that do not have new enough versions of `blueprint-compiler` packaged (I'm looking at you Debian 12).

A script (accessed by `nix run .#compile-blueprints`) is provided to conveniently recompile all of the blueprints in the repository. A CI job is also added to ensure that the files stay in sync.


mitchellh:
These should not be checked into the source tree. I’ll look to see if I can massage this otherwise though. I’m working on eliminating our other build steps too from tarballs. 

Source tree builds will still need the dep.

mitchellh (changes requested):
Going to mark this as "requesting changes" just so it doesn't get merged.

As noted in the previous comment, there's some changes I want to make here. The improvements made in subsequent commits are good but still not completely in the direction I'm imagining. 

I expect I'll use some of this work but step 1 I want to shore up our source tarball generation into the Zig build system. I also want to rely less on `nix run` and more on `zig build` (which can of course use Nix dependencies). 

It also _looks like_ there are some unrelated changes mixed into this PR? I'm not sure, its hard to tell. If so, please pull those out into separate PRs.

I don't expect to merge this.

mitchellh:
Closing in favor of #6822. There seems to be some useful things in here, so if you want to bring back those parts on top of the latest changes, feel free!

PR #6776, opened 2025-03-17
Update Debian 12 Dockerfile

jcollie:
1. Automatically detect the required Zig version rather than using a hardcoded value.
2. Run `ghostty +version` after the build as a sanity check.


mitchellh (approved):
Looks good pending CI pass!

mitchellh:
Oh, we can remove the ZIG_VERSION from the CI Yaml then

PR #6779, opened 2025-03-17
apprt/gtk: any preedit change should note a composing state

### LINKED ISSUE #6772: linux: fcitx5-hangful "한글" input does not work
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6615

<sup>Originally posted by **aca** March  7, 2025</sup>
It seems korean input is broken in nightly.
Creating this issue in the hope that Korean users can gather, organize, and resolve related issues.

WIP
- https://github.com/ghostty-org/ghostty/pull/5728
- https://github.com/ghostty-org/ghostty/issues/1663

Fixed
- https://github.com/ghostty-org/ghostty/pull/1665
- https://github.com/ghostty-org/ghostty/pull/1659
- https://github.com/ghostty-org/ghostty/pull/1651
- https://github.com/ghostty-org/ghostty/pull/1594
- https://github.com/ghostty-org/ghostty/pull/856

cc @exploitJ @leokang94
</div>

### Reproduction

1. Install fcitx5-hangul
2. Launch Ghostty
3. Switch to hangul ime
4. Type "dkssud" 
5. See "한글". Press space.
6. See" 한".

Environment I reproduced this:

- NixOS
- Wayland
- Fcitx

I didn't test X11.

mitchellh:
Fixes #6772

When typing Korean with the fcitx5-hangful input method, moving between graphemes does not trigger a preedit end/start cycle and instead just clears the preexisting preedit and reuses the started state.

Every other input method we've tested up until now doesn't do this. We need to mark composing set to "false" in "commit" because some input methods on the contrary fail to ever call END.

What is the point of start/end events if they are just ignored depending on the whim of the input method? Nothing. That's what. Its all a mess that GTK should be protecting us from but it doesn't and now its the app developer's problem. I'm frustrated because I feel like the point of an app framework is to mask this kind of complexity from the app developer and I'm playing whack-a-mole with input methods.

Well, here's another whack. Let's see if it works.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -1969,6 +1969,14 @@ fn gtkInputPreeditChanged(
     ctx: *gtk.IMMulticontext,
     self: *Surface,
 ) callconv(.C) void {
+    // Any preedit change should mark that we're composing. Its possible this
+    // is false using fcitx5-hangul and typing "dkssud<space>" ("안녕"). The
+    // second "s" results in a "commit" for "안" which sets composing to false,
+    // but then immediately sends a preedit change for the next symbol. With
+    // composing set to false we won't commit this text. Therefore, we must
+    // ensure it is set here.
+    self.im_composing = true;
+
     // Get our pre-edit string that we'll use to show the user.
     var buf: [*:0]u8 = undefined;
     ctx.as(gtk.IMContext).getPreeditString(&buf, null, null);


jcollie (approved):
No way for me to test, but it can't get any worse. Right? ...Right?

mitchellh:
> No way for me to test, but it can't get any worse. Right? ...Right?

There will be more bugs.

image

On a more serious note, I don't think we can reasonably prevent regressions here until we seriously look into VM-based automating test (i.e. in the style of `nixosTest`, but it could be anything).

PR #6783, opened 2025-03-17
gtk: add "remember choice" toggle for clipboard confirmation dialog

pluiedev:
Implements #6763 (not backporting to the 1.2 version since I'm way too lazy)

image

### DIFF
--- a/po/com.mitchellh.ghostty.pot
+++ b/po/com.mitchellh.ghostty.pot
@@ -1,5 +1,5 @@
 # SOME DESCRIPTIVE TITLE.
-# Copyright (C) YEAR Mitchell Hashimoto, Ghostty contributors
+# Copyright (C) YEAR "Mitchell Hashimoto, Ghostty contributors"
 # This file is distributed under the same license as the com.mitchellh.ghostty package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 #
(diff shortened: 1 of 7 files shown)


jcollie (approved):

jcollie (changes requested):
I dont't think that we need to make this work on Debian 12, but it should at least compile ;)

tristan957:
Should the text say something about the choice being remembered for the session?

tristan957:
I also left a comment at https://github.com/ghostty-org/ghostty/discussions/6763#discussioncomment-12531198

pluiedev:
The problem is that "session" is a very nebulous term, even more so in other languages (trying to express it in Chinese actually almost drove me insane). I think just implying that it only has an effect for this split, and that it will be reset after reloading the config, has pretty much the same effect.

pluiedev:
Implementing it per surface is very easy and I think could serve as an initial implementation. Per process would be a lot more complicated but perhaps still doable — out of scope for this PR though IMO

tristan957:
I guess I'm confused about what session even means. Is it per ghostty instance or per surface?

jcollie:
> I guess I'm confused about what session even means. Is it per ghostty instance or per surface? Referring to the original discussion

The current PR is per-surface instance correct? So 5 tabs with 2 splits each you could potentially be asked 10 times. So anytime you create a new surface you'd be asked again. Also, I believe that if you reload your config you'd reset all that to default so you'd be asked again, even for surfaces where you'd already been asked.

Mainly we just need to be clear with the documentation. Maybe there's a way to add a little "?" button that would explain in more detail than would fit in the confirmation dialog.

mitchellh (approved):

mitchellh:
I'm fine with this. I think per surface is completely reasonable for this action. I think copy could improve but its good to me.

PR #6787, opened 2025-03-18
gtk: convert App to zig-gobject

jcollie:


pluiedev (approved):
Looks good, CI failures notwithstanding (looks like they're just spurious network issues though)

PR #6796, opened 2025-03-18
Update Adwaita version check for Box unref

gpanders:
As of Adwaita 1.5.0, the GTK Box is not being properly unref'd when the parent window is closed. Update the conditional to account for this.

Also add a couple of missing unref()s in errdefers.

This fixes an issue where Ghostty would not properly quit after closing the last surface.

Resolves: https://github.com/ghostty-org/ghostty/discussions/3807

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -370,6 +370,7 @@ pub fn init(self: *Surface, app: *App, opts: Options) !void {
 
     // Create an overlay so we can layer the GL area with other widgets.
     const overlay = gtk.Overlay.new();
+    errdefer overlay.unref();
     const overlay_widget = overlay.as(gtk.Widget);
     overlay.setChild(gl_area_widget);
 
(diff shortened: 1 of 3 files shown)


jcollie (approved):
LGTM

cespare:
@gpanders FYI while this works for me, it does cause this new error message upon exiting:

```
(process:1114673): Gtk-CRITICAL **: 10:35:53.641: GtkBox 0x1dc57260 has a parent AdwBin 0x1dcf5460 during dispose. Parents hold a reference, so this should not happen.
Did you call g_object_unref() instead of gtk_widget_unparent()?
```

gpanders:
> @gpanders FYI while this works for me, it does cause this new error message upon exiting:
> 
> ```
> (process:1114673): Gtk-CRITICAL **: 10:35:53.641: GtkBox 0x1dc57260 has a parent AdwBin 0x1dcf5460 during dispose. Parents hold a reference, so this should not happen.
> Did you call g_object_unref() instead of gtk_widget_unparent()?
> ```

Yes that's already mentioned [in the comment](https://github.com/ghostty-org/ghostty/blob/0f2f0ab69f65941ab85c0563abdc5583d1983527/src/apprt/gtk/TabView.zig#L214-L215).

cespare:
> Yes that's already mentioned [in the comment](https://github.com/ghostty-org/ghostty/blob/0f2f0ab69f65941ab85c0563abdc5583d1983527/src/apprt/gtk/TabView.zig#L214-L215) and was the case before this PR.

Ah, sorry for the noise.

I noticed that the error message didn't appear before this PR, but I guess that's just because the shutdown wasn't happening at all.

PR #6800, opened 2025-03-18
`zig build dist` and `distcheck` for source tarballs

mitchellh:
This moves the source tarball creation process into the Zig build system and follows the autotools-standard naming conventions of `dist` and `distcheck`.

This doesn't change any of our build process otherwise. This is the foundation for #6760 along with other source tarball tasks I have planned (i.e. gobject bindings too). 

The `dist` target creates a source tarball in the `PREFIX/dist` directory. The tarball is named `ghostty-VERSION.tar.gz` as expected by standard source tarball conventions.

The `distcheck` target does the same as `dist`, but also takes the resulting tarball, extracts it, and runs tests on the extracted source to verify the source tarball works as expected. Distcheck currently only runs `zig build test` but in the future we can add additional checks to run. 

This commit also updates CI:

  1. Tagged releases now use the new `zig build distcheck` command.
  2. Tip releases now use the new `zig build dist` command.
  3. A new test build tests that source tarball generation works on every commit.

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -77,8 +77,17 @@ jobs:
     needs: [setup]
     env:
       GHOSTTY_VERSION: ${{ needs.setup.outputs.version }}
+      ZIG_LOCAL_CACHE_DIR: /zig/local-cache
+      ZIG_GLOBAL_CACHE_DIR: /zig/global-cache
     steps:
       - uses: actions/checkout@v4
+
+      - name: Setup Cache
+        uses: namespacelabs/nscloud-cache-action@v1.2.0
+        with:
+          path: |
+            /nix
+            /zig
 
       - uses: cachix/install-nix-action@v30
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/release-tag.yml
```diff
@@ -108,8 +117,6 @@ jobs:
           path: |-
             ghostty-${{ env.GHOSTTY_VERSION }}.tar.gz
             ghostty-${{ env.GHOSTTY_VERSION }}.tar.gz.minisig
-            ghostty-source.tar.gz
-            ghostty-source.tar.gz.minisig
```
mitchellh, line 112: These were from the 1.0 release and were generally a mistake to make, the `ghostty-VERSION` builds are the recommended and only documented ones at this point. I kept the old ones around for a time for build tooling but everything should be updated by now.

jcollie (approved):

pluiedev (approved):

PR #6801, opened 2025-03-18
CI: Add checks for blueprint compiler / Nix refactors

jcollie:
1. Refactored Nix devshell/package to make it easier to keep LD_LIBRARY_PATH & buildInputs in sync (plus make it easier to re-use in other Nix environment).
2. Added a CI job to ensure that Blueprints are formatted correctly and that they will compile using `blueprint-compiler` 0.16.0.
3. Reformatted all Blueprints with `blueprint-compiler format`.

### DIFF
--- a/nix/build-support/check-blueprints.sh
+++ b/nix/build-support/check-blueprints.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -o nounset -o pipefail -o errexit
+
+find src -name \*.blp -exec blueprint-compiler format {} \+
+find src -name \*.blp -exec blueprint-compiler compile --output=/dev/null {} \;
--- a/src/apprt/gtk/ui/1.2/config-errors-dialog.blp
+++ b/src/apprt/gtk/ui/1.2/config-errors-dialog.blp
@@ -2,27 +2,27 @@ using Gtk 4.0;
 using Adw 1;
 
 Adw.MessageDialog config_errors_dialog {
-    heading: _("Configuration Errors");
-    body: _("One or more configuration errors were found. Please review the errors below, and either reload your configuration or ignore these errors.");
+  heading: _("Configuration Errors");
+  body: _("One or more configuration errors were found. Please review the errors below, and either reload your configuration or ignore these errors.");
 
-    responses [
-        ignore: _("Ignore"),
-        reload: _("Reload Configuration") suggested,
-    ]
+  responses [
+    ignore: _("Ignore"),
+    reload: _("Reload Configuration") suggested,
+  ]
 
-    extra-child: ScrolledWindow {
-        min-content-width: 500;
-        min-content-height: 100;
-        
-        TextView {
-            editable: false;
-            cursor-visible: false;
-            top-margin: 8;
-            bottom-margin: 8;
-            left-margin: 8;
-            right-margin: 8;
+  extra-child: ScrolledWindow {
+    min-content-width: 500;
+    min-content-height: 100;
 
-            buffer: TextBuffer error_message { };
-        }
-    };
+    TextView {
+      editable: false;
+      cursor-visible: false;
+      top-margin: 8;
+      bottom-margin: 8;
+      left-margin: 8;
+      right-margin: 8;
+
+      buffer: TextBuffer error_message {};
+    }
+  };
 }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,7 @@ jobs:
       - alejandra
       - typos
       - translations
+      - blueprint-compiler
       - test-pkg-linux
       - test-debian-12
     steps:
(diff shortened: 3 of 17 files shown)


nix/build-support/ld_library_path.nix
```diff
@@ -0,0 +1,10 @@
+{
+  pkgs,
+  lib,
+  stdenv,
+  enableX11 ? true,
+  enableWayland ? true,
+}:
+lib.makeLibraryPath (import ./build_inputs.nix {
```
mitchellh, line 8: We're mixing `thisCase` and `this-case` and `this_case` all within the `nix/` folder. I understand `thisCase` because it matches Nix function style. We generally elsewhere in the project use hyphens.

nix/build-support/check-blueprints.nix
```diff
@@ -0,0 +1,29 @@
...
+    text = ''
+      export LD_LIBRARY_PATH="${ld_library_path}";
+      export GI_TYPELIB_PATH="${gi_typelib_path}"
+
+      find src -name \*.blp -exec blueprint-compiler format {} \+
+      find src -name \*.blp -exec blueprint-compiler compile --output=/dev/null {} \;
+    '';
+  };
+in {
+  type = "app";
```
mitchellh, line 27: Why does this have to be a whole Nix thing? Can't we just run the below?

```
find src -name \*.blp -exec blueprint-compiler format {} \+
find src -name \*.blp -exec blueprint-compiler compile --output=/dev/null {} \;
```

Should this be a simple shell script in `build-support` that others could run without Nix? (Although, the Nix devShell will give you the proper env vars). Our official build environment is Nix and I like using Nix for what it's good for, but I try to be pragmatic in balancing that between forcing it.

mitchellh (changes requested):
A couple easy questions, the rest looks good and this is definitely a good thing to do.

nix/build-support/ld_library_path.nix
pluiedev: Nix files are generally in kebab-case

nix/build-support/build_inputs.nix
```diff
@@ -0,0 +1,43 @@
...
+  pkgs,
+  lib,
+  stdenv,
+  enableX11 ? stdenv.hostPlatform.isLinux,
+  enableWayland ? stdenv.hostPlaform.isLinux,
+}:
+[
+  pkgs.libGL
+]
+++ (lib.optionals stdenv.hostPlatform.isLinux [
```
pluiedev, line 11: The parens are actually not required here

nix/build-support/gi_typelib_path.nix
```diff
@@ -0,0 +1,17 @@
+{
+  pkgs,
+  lib,
+  stdenv,
+}:
+lib.makeSearchPath "lib/girepository-1.0" (map (pkg: lib.getOutput "lib" pkg) (lib.optionals stdenv.hostPlatform.isLinux [
```
pluiedev, line 6: Eta-reduction
```suggestion
lib.makeSearchPath "lib/girepository-1.0" (map (lib.getOutput "lib") (lib.optionals stdenv.hostPlatform.isLinux [
```

src/apprt/gtk/ui/1.2/config-errors-dialog.blp
```diff
@@ -2,27 +2,27 @@ using Gtk 4.0;
 using Adw 1;
 
 Adw.MessageDialog config_errors_dialog {
-    heading: _("Configuration Errors");
-    body: _("One or more configuration errors were found. Please review the errors below, and either reload your configuration or ignore these errors.");
+  heading: _("Configuration Errors");
```
pluiedev, line 5: My editor somehow always reformats this with 4 spaces (I tried setting the indent style, sadly it didn't work.) Is there a way to specify this so that it remains consistent?

pluiedev (changes requested):

nix/build-support/check-blueprints.nix
pluiedev: I agree that we should really just have a shell script

nix/build-support/gi_typelib_path.nix
jcollie: Fixed.

nix/build-support/build_inputs.nix
jcollie: Fixed.

nix/build-support/check-blueprints.nix
jcollie: Other than I like Nix? No. Fixed.

src/apprt/gtk/ui/1.2/config-errors-dialog.blp
jcollie: Other than running `blueprint-compiler format` to reformat I don't know.

nix/build-support/check-blueprints.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -o nounset -o pipefail -o errexit
+
+find src -name \*.blp -exec blueprint-compiler format {} \+
+find src -name \*.blp -exec blueprint-compiler compile --output=/dev/null {} \;
```
mitchellh, line 6: For future note, I'm also totally fine just putting these directly in the GHA YAML files. I tend to only put scripts into `build-support` if a human can realistically run them regularly. In this case, I'm completely fine either way.

mitchellh (approved):

pluiedev (approved):

PR #6812, opened 2025-03-19
Tweak Norwegian Bokmål translation

cristeahub:
This tweak is minor and fixes some grammatical errors.

I did not change `kopier` to `kopiér`, even though that is a common way to write the word. The language council of Norway suggest writing the word as it is written here, without the accent, but it does read weird and allows for misunderstanding. If this was my project I would have added the accent. Let me know if you want the accent added.

What do you think regarding the accents @Uzaaft ?

### DIFF
--- a/po/nb_NO.UTF-8.po
+++ b/po/nb_NO.UTF-8.po
@@ -3,14 +3,15 @@
 # This file is distributed under the same license as the com.mitchellh.ghostty package.
 # Hanna Rose <hanna@hanna.lol>, 2025.
 # Uzair Aftab <uzaaft@outlook.com>, 2025.
+# Christoffer Tønnessen <christoffer@cto.gg>, 2025.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2025-03-19 08:28-0700\n"
-"PO-Revision-Date: 2025-03-14 20:41-0400\n"
-"Last-Translator: Hanna Rose <hanna@hanna.lol>\n"
+"PO-Revision-Date: 2025-03-19 09:52+0100\n"
+"Last-Translator: Christoffer Tønnessen <christoffer@cto.gg>\n"
 "Language-Team: Norwegian Bokmal <l10n-no@lister.huftis.org>\n"
 "Language: nb\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


po/nb_NO.UTF-8.po
```diff
@@ -191,21 +192,21 @@ msgstr "Hovedmeny"
...
 msgid "View Open Tabs"
-msgstr "Se åpen faner"
+msgstr "Se åpne faner"
 
 #: src/apprt/gtk/Window.zig:265
 msgid ""
 "⚠️ You're running a debug build of Ghostty! Performance will be degraded."
 msgstr ""
-"⚠️ Du kjører en debug-bygg av Ghostty. Debug-bygg har redusert ytelse."
+"⚠️ Du kjører et debug-bygg av Ghostty. Debug-bygg har redusert ytelse."
```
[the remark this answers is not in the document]
cristeahub: We currently use the proposed wording internally at my work, i.e. "bygg" for "build", and I have seen it having widespread usage in talks and events. If you feel "build" is better we can change to it, but it would still be neuter, and we would keep the hyphen.

So we would end up with this
```suggestion
"⚠️ Du kjører et debug-build av Ghostty. Debug-build har redusert ytelse."
```

We could even go crazy and ignore the hyphen and write it as one word, but I think that would cause more confusion than not, since build and debug is not that common used in our language.

Uzaaft:
I feel like the accent is just a bit archaic and most tech people wouldn't use it, and probably prone to be outdated at some points, so I am a happy to just keep what we have atm.  @cristeahub (Also what happened to your discord user 😢 )

pluiedev:
There's a merge conflict (new lines were added that need translating). Could you also translate those while you're here? Thanks :D

cristeahub:
Yes, will fix

po/nb_NO.UTF-8.po
```diff
@@ -37,17 +38,19 @@ msgstr "OK"
...
 msgid "Configuration Errors"
-msgstr ""
+msgstr "Konfigurasjonsfeil"
 
 #: src/apprt/gtk/ui/1.5/config-errors-dialog.blp:6
 msgid ""
 "One or more configuration errors were found. Please review the errors below, "
 "and either reload your configuration or ignore these errors."
 msgstr ""
+"Én eller flere konfigurasjonsfeil ble funnet. Vennligst gjennomgå feilene under, "
```
cristeahub, line 48: Just since I'm quite the language nerd and mentioned accents. This word means the number 1, while `en` is an article. So here it means "one or more", instead of "a or more".

Uzaaft: You god damn nerd😂But you are right.

Uzaaft (approved):
This looks much better to me :D Appreciate your language nerdiness :P

PR #6814, opened 2025-03-19
os: Add extra sentinel for GHOSTTY_RESOURCES_DIR

AnthonyZhOon:
Discover resourcesdir with `terminfo/g/ghostty`
as well as existing `terminfo/x/xterm-ghostty`.
This allows either terminfo file to be installed,
notably ncurses only provides a `terminfo/g/ghostty`.

It was brought up that the `ncurses-term` package on Fedora 42 at <https://packages.fedoraproject.org/pkgs/ncurses/ncurses-term/fedora-42.html> now provides a `/g/ghostty` terminfo entry which conflicts with installing the similarly named file from ghostty's build process. 
However a build with `zig build -Demit-terminfo=false` won't work as the `x/xterm-ghostty` terminfo entry is used as a sentinel to discover the `GHOSTTY_RESOURCES_DIR` used as a reference path for finding locale and theme files.

```
src/Surface.zig|546 col 43| .resources_dir = global_state.resources_dir,
src/cli/list_themes.zig|112 col 22| if (global_state.resources_dir == null)
src/global.zig|38 col 5| resources_dir: ?[]const u8,
src/global.zig|173 col 14| self.resources_dir = try internal_os.resourcesDir(self.alloc);
src/global.zig|174 col 27| errdefer if (self.resources_dir) |dir| self.alloc.free(dir);
src/global.zig|177 col 18| if (self.resources_dir) |v| internal_os.i18n.init(v) catch |err| {
src/global.zig|185 col 18| if (self.resources_dir) |dir| self.alloc.free(dir);
```

We also have some comments that intend to change how the terminfo database is discovered.
https://github.com/ghostty-org/ghostty/blob/main/src/termio/Exec.zig#L776C1-L781C60

### DIFF
--- a/src/os/resourcesdir.zig
+++ b/src/os/resourcesdir.zig
@@ -29,10 +29,10 @@ pub fn resourcesDir(alloc: std.mem.Allocator) !?[]const u8 {
 
     // This is the sentinel value we look for in the path to know
     // we've found the resources directory.
-    const sentinel = switch (comptime builtin.target.os.tag) {
-        .windows => "terminfo/ghostty.terminfo",
-        .macos => "terminfo/78/xterm-ghostty",
-        else => "terminfo/x/xterm-ghostty",
+    const sentinels = switch (comptime builtin.target.os.tag) {
+        .windows => .{"terminfo/ghostty.terminfo"},
+        .macos => .{"terminfo/78/xterm-ghostty"},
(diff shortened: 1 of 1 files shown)


00-kat (approved):

mitchellh (approved):
Thanks

PR #6820, opened 2025-03-19
Regenerate translations to fix CI

mitchellh:
I don't see any actual changes here, just reordering. It's using the Nix environment so I'm not sure why this happened but it seemed to stem from the Norwegian work originally. Fixing it back.


mitchellh:
Merged so CI passes, but we do have some missing translations I'll make an issue.

PR #6822, opened 2025-03-19
build: distribute gresource c/h with source tarball

### LINKED ISSUE #6760: Generate blueprint-compiler artifacts as part of source tarball generation

mitchellh:
Closes #6760 
Supersedes #6762

This introduces the concept of a "dist resource" (specifically a `GhosttyDist.Resource` type). This is a resource that may be present in dist tarballs but not in the source tree. If the resource is present and we're not in a Git checkout, then we use it directly instead of generating it. This is a generic concept we can apply to any preprocessing we want to do that we don't want users/packagers to do.

This is used for the first time in this commit for the gresource c/h files, which depend on a variety of external tools (blueprint-compiler, glib-compile-resources, etc.) that we do not want to require downstream users/packagers to have and we also do not want to worry about them having the right versions.

This also adds a check for `distcheck` to ensure our distribution contains all the expected files.

Note @jcollie that there may be elements of 6762 you'll want to bring back after this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -239,7 +239,17 @@ jobs:
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
       - name: Build and Check Source Tarball
-        run: nix develop -c zig build distcheck
+        run: |
+          rm -rf zig-out/dist
+          nix develop -c zig build distcheck
+          cp zig-out/dist/*.tar.gz ghostty-source.tar.gz
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: source-tarball
+          path: |-
+            ghostty-source.tar.gz
 
   build-macos:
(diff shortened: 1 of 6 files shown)


mitchellh:
I'm looking into updating our CI to validate this a bit more for us (e.g. change our Debian 12 builds to use the source ball).

mitchellh:
The Debian 12 build now validates the source tarball and doesn't have a blueprint compiler installed at all.

jcollie (approved):

PR #6824, opened 2025-03-19
font/freetype: disable SVG glyphs, simplify color check

qwerasd205:
We don't currently support rendering SVG glyphs so they should be ignored when loading. Additionally, the check for whether a glyph is colored has been simplified by just checking the pixel mode of the rendered bitmap.

This commit also fixes a bug caused by calling the color check inside of `renderGlyph`, which caused the bitmap to be freed creating a chance for memory corruption and garbled glyphs. This bug was introduced by 40c1140 (#6602) and discussed in #6781.

### DIFF
--- a/src/font/face/freetype.zig
+++ b/src/font/face/freetype.zig
@@ -270,25 +270,21 @@ pub const Face = struct {
 
     /// Returns true if the given glyph ID is colorized.
     pub fn isColorGlyph(self: *const Face, glyph_id: u32) bool {
-        // sbix table is always true for now
-        if (self.face.hasSBIX()) return true;
-
-        // CBDT/CBLC tables always imply colorized glyphs.
-        // These are used by Noto.
-        if (self.face.hasSfntTable(freetype.Tag.init("CBDT"))) return true;
-        if (self.face.hasSfntTable(freetype.Tag.init("CBLC"))) return true;
-
-        // Otherwise, load the glyph and see what format it is in.
+        // Load the glyph and see what pixel mode it renders with.
+        // All modes other than BGRA are non-color.
+        // If the glyph fails to load, just return false.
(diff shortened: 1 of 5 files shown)


mitchellh (approved):
The color detection stuff is a bit spooky. It makes sense but I wonder why my past self didn't do this haha. Approving we can find regressions in test.

qwerasd205:
Fixed the test failure -- this whole PR is fairly spooky so we definitely do need to watch out for regressions.

src/font/face/freetype.zig
```diff
@@ -270,25 +270,21 @@ pub const Face = struct {
...
             .color = self.face.hasColor(),
+            // NO_SVG set to true because we don't currently support rendering
+            // SVG glyphs under FreeType, since that requires bundling another
+            // dependency to handle rendering the SVG.
+            .no_svg = true,
         }) catch return false;
 
-        // If the glyph is SVG we assume colorized
         const glyph = self.face.handle.*.glyph;
-        if (glyph.*.format == freetype.c.FT_GLYPH_FORMAT_SVG) return true;
```
mitchellh, line 289: I realized that on distributed Ghostty, we dynamically link to freetype so SVG support may actually work. I'm not sure what this means for us but want to mention it here since you're more aware.

mitchellh (approved):
Looks good and I approve. I made one note about SVG support.

qwerasd205:
As far as I know there is no way to statically configure FreeType to support SVG, see https://freetype.org/freetype2/docs/reference/ft2-svg_fonts.html - the programmer has to provide hooks when initializing FreeType within their program.

mitchellh:
Ah I see. Well, very good then let's give it a go! Thanks

PR #6827, opened 2025-03-19
gtk: require blueprint-compiler 0.16 for building

jcollie:
Changes:

1. Require `blueprint-compiler` 0.16.0 (or newer) for building from a git checkout. With #6822 distributions that can't meet that requirement can use generated source tarballs to build.
2. Remove all `.ui` files as they are unnecessary.
3. Simplify the `Builder` interface since raw `.ui` files are no longer used.
4. Removed build-time check of raw `.ui` files.


jcollie:
This needs updates to the snap packaging. I decided to do that as a separate PR (#6831) instead of tacking it all onto this one.

pluiedev (approved):

mitchellh:
I think all our dependent PRs are merged. Is this safe to merge?

jcollie:
Yeah this is good to go.

PR #6831, opened 2025-03-20
snap: build from source tarball

jcollie:
Eliminates the need for `blueprint-compiler`, which isn't quite new enough on the snap builder.

### DIFF
--- a/src/build/GhosttyDist.zig
+++ b/src/build/GhosttyDist.zig
@@ -36,6 +36,17 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyDist {
         "--format=tgz",
     });
 
+    // embed the Ghostty version in the tarball
+    {
+        const version = b.addWriteFiles().add("VERSION", b.fmt("{}", .{cfg.version}));
+        // --add-file uses the most recent --prefix to determine the path
+        // in the archive to copy the file (the directory only).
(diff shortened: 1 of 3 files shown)


src/build/GhosttyDist.zig
```diff
@@ -36,6 +36,15 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyDist {
         "--format=tgz",
     });
 
+    // embed the Ghostty version in the tarball
+    {
+        const version = b.addWriteFiles().add(".version", b.fmt("{}", .{cfg.version}));
```
mitchellh, line 41: I really like this idea, but just want to bikeshed on the filename. Is there any other examples of projects doing this? I'll take a look too.

mitchellh (commented):
One comment, definitely want @kenvandine's thoughts on this. I like it!

src/build/GhosttyDist.zig
tristan957: In most projects I am a part of, the filename is `VERSION`. If you search `files('VERSION')` on GitHub, you'll get a pretty fair amount of search results back.

That would find meson build files most likely.

jcollie: Changed filename to VERSION

pluiedev (approved):

kenvandine (approved):
This looks fine to me from a snap POV

mitchellh (approved):

PR #6834, opened 2025-03-20
apprt/embedded: utf8 encoding buffer lifetime must extend beyond call

### LINKED ISSUE #6821: macOS: `Ctrl+C` doesn't work on Intel Macs since Zig 0.14
#6790 

I don't have an Intel Mac to debug this anymore, but I suspect there's a miscomputation or C ABI issue somewhere. It's been a very long time since we've experienced this. Reproduction _should_ be simple (but I can't verify since I don't have an Intel Mac):

1. Launch Ghostty
2. Ctrl+C
3. Expect ctrl+C, get `;5u`

This issue doesn't exist on any M-series processors.

mitchellh:
Fixes #6821

UTF8 translation using KeymapDarwin requires a buffer and the buffer was stack allocated in the coreKeyEvent call and returned from the function. We need the buffer to live longer than this.

Long term, we're removing KeymapDarwin (there is a whole TODO comment in there about how to do it), but this fixes a real problem today.

### DIFF
--- a/src/build/Config.zig
+++ b/src/build/Config.zig
@@ -72,7 +72,7 @@ pub fn init(b: *std.Build) !Config {
         if (result.result.os.tag == .macos and
             builtin.target.os.tag.isDarwin())
         {
-            result = genericMacOSTarget(b, null);
+            result = genericMacOSTarget(b, result.query.cpu_arch);
         }
 
         // If we have no minimum OS version, we set the default based on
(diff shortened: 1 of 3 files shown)


src/build/Config.zig
```diff
@@ -72,7 +72,7 @@ pub fn init(b: *std.Build) !Config {
         if (result.result.os.tag == .macos and
             builtin.target.os.tag.isDarwin())
         {
-            result = genericMacOSTarget(b, null);
+            result = genericMacOSTarget(b, result.query.cpu_arch);
```
mitchellh, line 75: This is an unrelated change that allows us to cross-compile on macOS.

PR #6842, opened 2025-03-20
i18n: allow localization of config diagnostics

jcollie:


mitchellh (commented):
This is making me wonder if we should address `sprintf`-style translations first and make these proper rather than doing substrings.

src/cli/args.zig
```diff
@@ -200,7 +203,7 @@ fn formatInvalidValue(
     var buf = std.ArrayList(u8).init(arena_alloc);
     errdefer buf.deinit();
     const writer = buf.writer();
-    try writer.print("invalid value \"{?s}\"", .{value});
+    try writer.print("{s} \"{?s}\"", .{ i18n._("invalid value"), value });
```
pluiedev, line 206: This is a bad idea for multiple reasons:
- Not all languages use English/Latin alphabet quotes
- Not all languages place the string as a postfix
- Not all languages even have "invalid value" and the string as two separate, distinct parts

And so on and so forth. As Mitchell stated we should explore sprintf-style printing first, or implement another mechanism for substituting variables

src/cli/args.zig
```diff
@@ -212,7 +215,7 @@ fn formatValues(comptime T: type, key: []const u8, writer: anytype) std.mem.Allo
         if (std.mem.eql(u8, key, f.name)) {
             switch (@typeInfo(f.type)) {
                 .@"enum" => |e| {
-                    try writer.print(", valid values are: ", .{});
+                    try writer.writeAll(std.mem.span(i18n._(", valid values are: ")));
```
pluiedev, line 218: This entire message needs to be rewritten so that it flows as one entire sentence

src/config/Config.zig
```diff
@@ -2732,8 +2733,8 @@ pub fn loadRecursiveFiles(self: *Config, alloc_gpa: Allocator) !void {
                 const diag: cli.Diagnostic = .{
                     .message = try std.fmt.allocPrintZ(
                         arena_alloc,
-                        "error opening config-file {s}: {}",
-                        .{ path, err },
+                        "{s} config-file {s}: {}",
```
pluiedev, line 2736: Not everyone understands what `config-file` means, especially when we implement GUI preferences and users won't be able to make the connection between the name and the config setting

src/config/Config.zig
```diff
@@ -2751,8 +2752,12 @@ pub fn loadRecursiveFiles(self: *Config, alloc_gpa: Allocator) !void {
...
                     .message = try std.fmt.allocPrintZ(
                         arena_alloc,
-                        "config-file {s}: not reading because file type is {s}",
-                        .{ path, @tagName(kind) },
+                        "config-file {s}: {s} {s}",
+                        .{
+                            path,
+                            i18n._("not reading because file type is"),
+                            @tagName(kind),
+                        },
```
pluiedev, line 2760: Translate the entire sentence

src/config/Config.zig
```diff
@@ -3167,8 +3172,8 @@ pub fn parseManuallyHook(
                 .location = try cli.Location.fromIter(iter, alloc),
                 .message = try std.fmt.allocPrintZ(
                     alloc,
-                    "missing command after {s}",
-                    .{arg},
+                    "{s} {s}",
+                    .{ i18n._("missing command after"), arg },
```
pluiedev, line 3176: Translate the entire sentence

pluiedev (changes requested):

jcollie:
OK, closing this. It's the wrong approach, plus the lack of `gettext` on Windows means we can't use it outside of a GTK build yet.

PR #6866, opened 2025-03-22
i18n: Add Japanese translations

lonsagisawa:
This pull request adds Japanese(ja_JP.UTF-8) translations.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -160,6 +160,7 @@
 /po/com.mitchellh.ghostty.pot @ghostty-org/localization
 /po/de_DE.UTF-8.po @ghostty-org/de_DE
 /po/id_ID.UTF-8.po @ghostty-org/id_ID
+/po/ja_JP.UTF-8.po @ghostty-org/ja_JP
 /po/nb_NO.UTF-8.po @ghostty-org/nb_NO
 /po/pl_PL.UTF-8.po @ghostty-org/pl_PL
 /po/uk_UA.UTF-8.po @ghostty-org/uk_UA
--- a/po/ja_JP.UTF-8.po
+++ b/po/ja_JP.UTF-8.po
@@ -1,0 +1,268 @@
+# Japanese translations for com.mitchellh.ghostty package
+# com.mitchellh.ghostty パッケージに対する英訳.
+# Copyright (C) 2025 Mitchell Hashimoto
+# This file is distributed under the same license as the com.mitchellh.ghostty package.
+# Lon Sagisawa <lon@sagisawa.me>, 2025.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: com.mitchellh.ghostty\n"
+"Report-Msgid-Bugs-To: m@mitchellh.com\n"
+"POT-Creation-Date: 2025-03-19 08:28-0700\n"
+"PO-Revision-Date: 2025-03-21 00:08+0900\n"
+"Last-Translator: Lon Sagisawa <lon@sagisawa.me>\n"
+"Language-Team: Japanese\n"
+"Language: ja\n"
+"MIME-Version: 1.0\n"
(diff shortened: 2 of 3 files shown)


CODEOWNERS
```diff
@@ -159,6 +159,7 @@
 /po/README_TRANSLATORS.md @ghostty-org/localization
 /po/com.mitchellh.ghostty.pot @ghostty-org/localization
 /po/de_DE.UTF-8.po @ghostty-org/de_DE
+/po/ja_JP.UTF-8.po @ghostty-org/ja_JP
```
00-kat, line 162: I don't think the team exists yet, so I'm not sure if this actually works; I'll let someone with access to the org's teams comment on this though.

pluiedev: We (or rather Mitchell) will add it when the PR is merged

lonsagisawa: Pasted from Catalan PR so I have no idea how it works.
https://github.com/ghostty-org/ghostty/pull/6841/commits/84d3554e2e461316ffb6bd151410be78b76137f3#diff-fcf14c4b7b34fe7a11916195871ae66a59be87a395f28db73e345ebdc828085b
Should I remove this or leave it as is since it will added later anyway?

00-kat: Leave it probably.

po/ja_JP.UTF-8.po
```diff
@@ -0,0 +1,268 @@
...
+"⚠️ You're running a debug build of Ghostty! Performance will be degraded."
+msgstr "⚠️ Ghosttyのデバッグビルドを実行しています! パフォーマンスが低下しています。"
+
+#: src/apprt/gtk/Window.zig:725
+msgid "Reloaded the configuration"
+msgstr "設定を再読み込みしました"
+
+#: src/apprt/gtk/Window.zig:941
+msgid "Ghostty Developers"
+msgstr "Ghostty 開発者"
```
pluiedev, line 268: Just curious - in all other strings there's no space between "Ghostty" and other text, so why is there a space here?

lonsagisawa: I have been inconsistent here, it looks like to put space around latin character in other project(e.g. xfce4-terminal, ptyxis) so made it consistently spaced.

pluiedev: Yeah, that's the convention I went with in zh_CN as well. Sucks that GTK/Pango doesn't do special spacing rules between CJK and Latin characters but it is what it is

po/ja_JP.UTF-8.po
```diff
@@ -0,0 +1,268 @@
...
+
+#: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:30
+#: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:42
+msgid "Split"
+msgstr "分割"
+
+#: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:33
+#: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:45
+msgid "Change Title…"
+msgstr "タイトルを変更…"
```
pluiedev, line 89: Do Japanese ellipses typically use three dots or six dots? (It's the latter in zh_CN, at least `……`)

lonsagisawa: Ellipses are usually used in even numbers in literature, but for indication of new window is three dots.

firefox_3YmeOScTn2

pluiedev: Good to know 👍

mitchellh:
@lonsagisawa I assume you volunteer to maintain these? Thank you!

pluiedev (changes requested):
Need at least one other person to review these, but otherwise looks good

mitchellh (approved):

PR #6871, opened 2025-03-22
shell-integration: switch to $GHOSTTY_SHELL_FEATURES

jparise:
This change consolidates all three opt-out shell integration environment variables into a single opt-in $GHOSTTY_SHELL_FEATURES variable. Its value is a comma-delimited list of the enabled shell feature names (e.g. "cursor,title").

$GHOSTTY_SHELL_FEATURES is set at runtime and automatically added to the shell environment. Its value is based on the shell-integration-features configuration option.

$GHOSTTY_SHELL_FEATURES is only set when at least one shell feature is enabled. It won't be set when 'shell-integration-features = false'.

$GHOSTTY_SHELL_FEATURES lists only the enabled shell feature names. We could have alternatively gone in the opposite direction and listed the disabled features, letting the scripts assume each feature is on by default like we did before, but I think this explicit approach is a little safer and easier to reason about / debug.

It also doesn't support the "no-" negation prefix used by the config system (e.g. "cursor,no-title"). This simplifies the implementation requirements of our (multiple) shell integration scripts, and because $GHOSTTY_SHELL_FEATURES is derived from shell-integration-features, the user-facing configuration interface retains that expressiveness.

$GHOSTTY_SHELL_FEATURES is intended to primarily be an internal concern: an interface between the runtime and our shell integration scripts. It could be used by people with particular use cases who want to manually source those scripts, but that isn't the intended audience.

... and because the previous $GHOSTTY_SHELL_INTEGRATION_NO_* variables were also meant to be an internal concern, this change does not include backwards compatibility support for those names.

One last advantage of a using a single $GHOSTTY_SHELL_FEATURES variable is that it can be easily forwarded to e.g. ssh sessions or other shell environments.

See: #5070

### DIFF
--- a/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
+++ b/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
@@ -49,10 +49,9 @@ status --is-interactive || ghostty_exit
 function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
     functions -e __ghostty_setup
 
-    set --local cursor string match -q "*cursor*" "$GHOSTTY_SHELL_FEATURES"
-    set --local sudo string match -q "*sudo*" "$GHOSTTY_SHELL_FEATURES"
+    set --local features (string split , $GHOSTTY_SHELL_FEATURES)
 
-    if $cursor
+    if contains cursor $features
         # Change the cursor to a beam on prompt.
         function __ghostty_set_cursor_beam --on-event fish_prompt -d "Set cursor shape"
             echo -en "\e[5 q"
@@ -64,7 +63,7 @@ function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
 
     # When using sudo shell integration feature, ensure $TERMINFO is set
     # and `sudo` is not already a function or alias
-    if $sudo and test -n "$TERMINFO"; and test "file" = (type -t sudo 2> /dev/null; or echo "x")
+    if contains sudo $features and test -n "$TERMINFO"; and test "file" = (type -t sudo 2> /dev/null; or echo "x")
         # Wrap `sudo` command to ensure Ghostty terminfo is preserved
         function sudo -d "Wrap sudo to preserve terminfo"
             set --function sudo_has_sudoedit_flags "no"
@@ -121,7 +120,7 @@ function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
     set --global fish_handle_reflow 1
 
     # Initial calls for first prompt
-    if $cursor
+    if contains cursor $features
         __ghostty_set_cursor_beam
     end
     __ghostty_mark_prompt_start
--- a/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
+++ b/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
@@ -49,10 +49,10 @@ status --is-interactive || ghostty_exit
 function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
     functions -e __ghostty_setup
 
-    # Check if we are setting cursors
-    set --local no_cursor "$GHOSTTY_SHELL_INTEGRATION_NO_CURSOR"
+    set --local cursor string match -q "*cursor*" "$GHOSTTY_SHELL_FEATURES"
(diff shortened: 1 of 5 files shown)


src/termio/shell_integration.zig
```diff
@@ -150,9 +150,18 @@ pub fn setupFeatures(
     env: *EnvMap,
     features: config.ShellIntegrationFeatures,
 ) !void {
-    if (!features.cursor) try env.put("GHOSTTY_SHELL_INTEGRATION_NO_CURSOR", "1");
-    if (!features.sudo) try env.put("GHOSTTY_SHELL_INTEGRATION_NO_SUDO", "1");
-    if (!features.title) try env.put("GHOSTTY_SHELL_INTEGRATION_NO_TITLE", "1");
+    var enabled = try std.BoundedArray(u8, 256).init(0);
```
jparise, line 153: I think we could also compute an precise array capacity at comptime.

src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
```diff
@@ -49,10 +49,10 @@ status --is-interactive || ghostty_exit
 function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
     functions -e __ghostty_setup
 
-    # Check if we are setting cursors
-    set --local no_cursor "$GHOSTTY_SHELL_INTEGRATION_NO_CURSOR"
+    set --local cursor string match -q "*cursor*" "$GHOSTTY_SHELL_FEATURES"
+    set --local sudo string match -q "*sudo*" "$GHOSTTY_SHELL_FEATURES"
```
pluiedev, line 53: For Fish at least I don't think this is the right approach. Fish has builtin list types and string operation primitives that guarantee the env var must be in a comma-delimited string. A combination of `string split` and `contains` makes much more sense IMO.

pluiedev (changes requested):

src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
jparise: That's a good idea. Thanks for the suggestion!

src/termio/shell_integration.zig
```diff
@@ -144,15 +144,29 @@ test "force shell" {
...
+        break :capacity n;
+    };
+    var buffer = try std.BoundedArray(u8, capacity).init(0);
+
+    inline for (fields) |field| {
+        if (@field(features, field.name)) {
+            if (buffer.len > 0) try buffer.append(',');
+            try buffer.appendSlice(field.name);
+        }
+    }
```
mitchellh, line 165: I think it'd be a good idea to extract this into a function that we can unit test. (e.g.. set all features, set none, etc. etc.)

pluiedev (approved):
Looks good on my end, but should we have any sort of backwards compatibility? I get that this is supposed to be a private API but I wonder if people are doing something funny with these env vars

00-kat:
> I wonder if people are doing something funny with these env vars

I believe at least one person set them in a remote's shell config after copying the shell integration over.  Though I highly doubt there's much fishy (pun intended…) business going on.

mitchellh:
> ... and because the previous $GHOSTTY_SHELL_INTEGRATION_NO_* variables were also meant to be an internal concern, this change does not include backwards compatibility support for those names.

I agree with this, so I would say there's no backwards compatibility concerns. We document how to manually source the configs but that's still in the context of running it through Ghostty (just disabling auto _injection_). So this is good to go.

mitchellh (changes requested):
Just making sure I put a request changes in here for the unit test coverage.

jparise:
> > ... and because the previous $GHOSTTY_SHELL_INTEGRATION_NO_* variables were also meant to be an internal concern, this change does not include backwards compatibility support for those names.
> 
> I agree with this, so I would say there's no backwards compatibility concerns. We document how to manually source the configs but that's still in the context of running it through Ghostty (just disabling auto _injection_). So this is good to go.

The other relevant context is that we now always set these feature variables as of #5048. Prior to that, when shell integration itself was disabled, we wouldn't provide these feature variables, leaving that to the user for the manual sourcing case.

src/termio/shell_integration.zig
jparise: @mitchellh just following up in case you didn't see my question

mitchellh (approved):
Sorry for the delay!

Issue #6872, opened 2025-03-22
GTK: Splits appear completely blank when using the new OpenGL renderer


pluiedev:
See #5855.

This is quite high priority since GTK removed the "old" OpenGL renderer starting from 4.17, and users using Debian 13 (Trixie) simply cannot get Ghostty rendering correctly no matter what, regardless of hardware.

Repro steps: Run `GSK_RENDERER=ngl ghostty --gtk-gsk-renderer=default`, then create a new split in any direction.

Video:

https://github.com/user-attachments/assets/aa983838-742e-4057-a009-e87e38fbf664

mitchellh:
More notes:

- Creating a new tab and closing it also triggers the bug
- Creating multiple new tabs works fine until a tab is closed
- Creating a new window and closing it works fine and _does not_ trigger the bug
- With default configs (`--config-default-files=false`), things work just in a really weird way
  * Padding around the viewport when the tab is closed
  * Active split works but inactive split is blank (not transparent)
  * It looks like background transparency causes major issues

jcollie:
I know in the past I've had issues messing with GTK widgets from anything but an idle handler (mainly the resize overlay). Not sure how east/hard it would be to do all the GL stuff from an idle handler. Probably kill performance/smoothness at best.

mitchellh:
> I know in the past I've had issues messing with GTK widgets from anything but an idle handler (mainly the resize overlay). Not sure how east/hard it would be to do all the GL stuff from an idle handler. Probably kill performance/smoothness at best.

Yeah this would be a bad idea for performance. I wrote some code to limit our resize events and it didn't fix anything so I don't think that's an issue anyway, probably a red herring.

mcdoogs:
Nice, this fixed a similar issue that I was having. For me, splitting once worked fine, but subsequent splits would break everything except for the first split. It looks like if I set transparency, then I got the behavior of all splits going full transparent instead. Both issues are fixed in main now.

https://github.com/user-attachments/assets/a11d9f4c-a0c9-4308-a4de-94809f75b8b5

00-kat:
Yep, that issue was what initially got the repro for this #5855.

ardasevinc:
when is 1.1.3 coming? i had this issue and downgrading didn't help so i'm currently using the nightly appimage (i'm on arch linux).

00-kat:
Probably a day or two.

kjkent:
[...]

PR #6875, opened 2025-03-22
gtk: switch to vulkan as the default renderer

jcollie:
Potential fix for #6872. Probably will need to change the defaults for older GTK versions, but this works on my GTK 4.16 system. I can't test on anything newer or older right now.


jcollie:
Also, I think that this is kind of a hacky workaround... Someone with some real GL/GTK chops will need to investigate why Ghostty is interacting so poorly with GTK depending on the renderer chosen.

mitchellh:
Superseded by #6877

Issue #5491, opened 2025-01-31
GTK: Tabs leak CPU memory when closed


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4366

<div type='discussions-op-text'>

<sup>Originally posted by **rhodes-b** January  1, 2025</sup>
This is easy to reproduce, open quite a few tabs (I did 20 new ones in this case) and see that memory is not being recovered when they are closed

@00-kat confirmed this in discord

![image](https://github.com/user-attachments/assets/abc5b334-c725-409f-89d5-acdfe6fe50d4)

These were my stats

`202MB` with one tab
`275MB` 20 tabs
close 20 tabs
`235MB` with one tab

It may be MiB im not sure what btop uses by default</div>

rhodes-b:
valgrind never worked for me but I could get heaptrack to work

[heaptrack.ghostty.zst.txt](https://github.com/user-attachments/files/18627007/heaptrack.ghostty.zst.txt)
(remove .txt extension github would let me upload otherwise)

00-kat:
I seem to be having a leak with OS windows and splits too (of about the same amount as I do with tabs), so I think this issue may not be only related to tabs.

Lking03x:
Hi, I have a similar issue on linux/ubuntu the terminal processes stay active after `CTRL + D`. If you run it from another terminal you can see this message upon closing.
<details>
  <summary>Output: CTRL + D</summary>
<code>
info: ghostty version=1.1.0
info: ghostty build optimize=ReleaseFast
info: runtime=apprt.Runtime.gtk
info: font_backend=font.main.Backend.fontconfig_freetype
info: dependency harfbuzz=8.3.0
info: dependency fontconfig=21500
info: renderer=renderer.OpenGL
info: libxev backend=main.Backend.io_uring
info(os): setlocale from env result=en_US.UTF-8
info(gtk): GTK version build=4.14.2 runtime=4.14.2
info: reading configuration file path={redacted}/.config/ghostty/config
info(config): default shell source=env value=/bin/bash
warning(gtk): setting GDK_DEBUG=gl-disable-gles,gl-no-fractional,vulkan-disable
warning(gtk): setting GDK_DISABLE=
warning(gtk): setting GSK_RENDERER=opengl
info(gtk): libadwaita version build=1.5.0 runtime=1.5.0
info(grid): loaded OpenGL 4.6
info(font_shared_grid_set): font regular: {redacted}
[...]

PR #6877, opened 2025-03-22
apprt/gtk: make GL context current in unrealize

### LINKED ISSUE #6872: GTK: Splits appear completely blank when using the new OpenGL renderer

mitchellh:
Fixes #6872

This commit explicitly acquires the GL context in the `unrealize` signal handler of the GTK Surface prior to cleaning up GPU resources.

A GLArea only guarantees that the associated GdkContext is current for the `render` signal (see the docs[1]). This is why our OpenGL renderer "defers" various operations such as resize, font grid changing, etc. (see the `deferred_`-prefix fields in `renderer/OpenGL.zig`). However, we missed a spot.

The `gtk-widget::unrealize` signal is emitted when the widget is destroyed, and it is the last chance we have to clean up our GPU resources. But it is not guaranteed that the GL context is current at that point, and we weren't making it current. On the NGL GTK renderer, this was freeing GPU resources we didn't own.

As best I can understand, the old GL renderer only ever used a handful of GL resources that were early in the ID space, so by coincidence we were probably freeing nothing and everything was fine. But with the new NGL renderer uses a LOT more GL resources (make a few splits and the ID space is already in the thousands, from GTK!), so we were freeing real resources that we didn't own, which caused rendering issues. :)

I suspect the above also resulted in VRAM memory leaks (which would be RAM memory leaks for unified memory GPUs). This potentially relates to #5491.

The fix is to explicitly make the GL context current in the `unrealize` handler.

[1]: https://docs.gtk.org/gtk4/method.GLArea.make_current.html

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -247,19 +247,6 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
         _ = internal_os.setenv("GDK_DISABLE", value[0 .. value.len - 1 :0]);
     }
 
-    if (gtk_version.runtimeAtLeast(4, 14, 0)) {
-        switch (config.@"gtk-gsk-renderer") {
-            .default => {},
-            else => |renderer| {
-                // Force the GSK renderer to a specific value. After GTK 4.14 the
-                // `ngl` renderer is used by default which causes artifacts when
-                // used with Ghostty so it should be avoided.
-                log.warn("setting GSK_RENDERER={s}", .{@tagName(renderer)});
-                _ = internal_os.setenv("GSK_RENDERER", @tagName(renderer));
-            },
-        }
-    }
-
     adw.init();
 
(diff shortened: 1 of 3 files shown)


jcollie (approved):
works on my system

pluiedev (approved):
Works on my system but god what a silly bug

mitchellh:
I've pushed a commit to also remove `gtk-gsk-renderer`. Users can override this manually by setting the standard env var if they need to, but Ghostty now works with all known renderers.

PR #6884, opened 2025-03-23
gtk: backport #6877 to 1.1.2

jcollie:


jcollie:
This'll need a new branch based on 52a5069decb6d698df347e4b46d5f26ebe639ebb to work properly.

mitchellh:
I prefer to cherry pick to backport, and I think we should keep the GSK renderer stuff in the backport since we should only port bug fixes and not improvements. We can change the default though to `default` since that'll fix issues in later versions.

mitchellh:
Will cherry pick this: https://github.com/ghostty-org/ghostty/commit/36f841ee80efe7f89d427b1482f34835599fcad5

mitchellh:
(To be clear, I cherry-pick and apply necessary backport changes on the cherry pick, since we won't have gobject here)

PR #6885, opened 2025-03-23
fix canonicalizing some zh locales

### LINKED ISSUE #6870: macOS: Some zh locales are incorrectly canonicalized from BCP to POSIX
See #6865 and my associated comment.

We should special case `zh-Hans-CN` and `zh-Hant-TW` to `zh_CN` and `zh_TW` respectively.

Aaron-212:
resolve #6870

### DIFF
--- a/src/os/i18n.zig
+++ b/src/os/i18n.zig
@@ -108,12 +108,15 @@ pub fn canonicalizeLocale(
     // Buffer must be 16 or at least as long as the locale and null term
     if (buf.len < @max(16, locale.len + 1)) return error.NoSpaceLeft;
 
+    // Fix zh locales for macOS
+    const updated = fixZHLocale(locale);
+
     // Copy our locale into the buffer since it modifies in place.
     // This must be null-terminated.
-    @memcpy(buf[0..locale.len], locale);
-    buf[locale.len] = 0;
+    @memcpy(buf[0..updated.len], updated);
+    buf[updated.len] = 0;
 
-    _libintl_locale_name_canonicalize(buf[0..locale.len :0]);
+    _libintl_locale_name_canonicalize(buf[0..updated.len :0]);
 
     // Convert the null-terminated result buffer into a slice. We
     // need to search for the null terminator and slice it back.
@@ -121,6 +124,24 @@ pub fn canonicalizeLocale(
     // null.
     const slice = std.mem.sliceTo(buf, 0);
     return buf[0..slice.len :0];
+}
+
+fn fixZHLocale(locale: []const u8) []const u8 {
+    if (locale.len == 10 and std.mem.eql(u8, locale[0..2], "zh")) {
+        const script = locale[3..7];
+        const region = locale[8..10];
+
+        if (std.mem.eql(u8, script, "Hans")) {
+            if (std.mem.eql(u8, region, "SG")) return "zh-SG";
+            return "zh-CN";
+        } else if (std.mem.eql(u8, script, "Hant")) {
+            if (std.mem.eql(u8, region, "HK")) return "zh-HK";
+            if (std.mem.eql(u8, region, "MO")) return "zh-MO";
+            return "zh-TW";
+        }
+    }
+
+    return locale;
 }
 
 /// This can be called at any point a compile-time-known locale is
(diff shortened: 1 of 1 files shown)


src/os/i18n.zig
```diff
@@ -108,12 +108,15 @@ pub fn canonicalizeLocale(
     // Buffer must be 16 or at least as long as the locale and null term
     if (buf.len < @max(16, locale.len + 1)) return error.NoSpaceLeft;
 
+    // Fix zh locales for macOS
+    const updated = fixZHLocale(locale);
```
00-kat, line 112: Pretty sure the Zig convention is to not full-caps acronyms.
>  Acronyms, initialisms, proper nouns, or any other word that has capitalization rules in written English are subject to naming conventions just like any other word. Even acronyms that are only 2 letters long are subject to these conventions. 
```suggestion
    const updated = fixZhLocale(locale);
```

src/os/i18n.zig
```diff
@@ -123,6 +126,24 @@ pub fn canonicalizeLocale(
     return buf[0..slice.len :0];
 }
 
+fn fixZHLocale(locale: []const u8) []const u8 {
+    if (locale.len == 10 and std.mem.eql(u8, locale[0..2], "zh")) {
+        const script = locale[3..7];
+        const region = locale[8..10];
+
+        if (std.mem.eql(u8, script, "Hans")) {
+            if (std.mem.eql(u8, region, "SG")) return "zh-SG";
```
00-kat, line 135: I don't really know this code or anything, but based on the description of #6870 I had thought these were meant to use underscores?

00-kat (changes requested):

pluiedev (changes requested):

00-kat (approved):

src/os/i18n.zig
```diff
@@ -123,6 +126,24 @@ pub fn canonicalizeLocale(
...
+            if (std.mem.eql(u8, region, "SG")) return "zh_SG";
+            return "zh_CN";
+        } else if (std.mem.eql(u8, script, "Hant")) {
+            if (std.mem.eql(u8, region, "HK")) return "zh_HK";
+            if (std.mem.eql(u8, region, "MO")) return "zh_MO";
+            return "zh_TW";
+        }
+    }
+
+    return locale;
```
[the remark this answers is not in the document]
Aaron-212: IMO it's okay because we're only dealing with `zh-Hans-CC` and `zh-Hant-CC`. Anything that doesn't match will just return the input locale. We can move on to more robust solutions later.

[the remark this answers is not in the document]
pluiedev: It's just in general quite hard to read and isn't that much "shorter" than the alternative. Why write unmaintainable code when the maintainable version is about as complex?
```zig
var it = std.mem.splitScalar(u8, locale, '-');
const name = it.next() orelse return;
if (!std.mem.eql(u8, name, "zh")) return;

const script = it.next() orelse return;
const region = it.next() orelse return;

if (std.mem.eql(u8, script, "Hans")) {
	if (std.mem.eql(u8, script, "SG")) return "zh_SG";
	return "zh_CN";
} else if (std.mem.eql(u8, script, "Hant")) {
	if (std.mem.eql(u8, script, "MO")) return "zh_MO";
	if (std.mem.eql(u8, script, "HK")) return "zh_HK";
	return "zh_TW";
}
```

src/os/i18n.zig
```diff
@@ -123,6 +126,26 @@ pub fn canonicalizeLocale(
     return buf[0..slice.len :0];
 }
 
+fn fixZhLocale(locale: []const u8) []const u8 {
```
pluiedev, line 129: Add a doc comment explaining why (e.g. the libintl's internal canonicalize function returns erroneous results, etc)

src/os/i18n.zig
```diff
@@ -108,12 +108,15 @@ pub fn canonicalizeLocale(
     // Buffer must be 16 or at least as long as the locale and null term
     if (buf.len < @max(16, locale.len + 1)) return error.NoSpaceLeft;
 
+    // Fix zh locales for macOS
+    const updated = fixZhLocale(locale);
```
pluiedev, line 112: Is there even a point of canonicalizing via libintl for "fixed" zh locales? If not then we can possibly have something like this: (the return type would need to be changed to `?[:0]const u8`)
```suggestion
    if (fixZhLocale(locale)) |fixed| {
        return fixed;
    } else {
        // normal logic
    }
```

pluiedev (changes requested):

pluiedev (approved):

mitchellh (changes requested):
I'm going to make a couple small changes but thanks for tackling this.

mitchellh (approved):

PR #6915, opened 2025-03-26
termio, flatpak: support spawning terminals in cwd

alaviss:
Implements path access testing for Flatpak via test spawning. This is required since Flatpak reserves certain paths from being accessible regardless of permissions.

Opened mostly as an RFC, this approach seems to work without any noticeable performance impact.

Ref: https://docs.flatpak.org/en/latest/sandbox-permissions.html#reserved-paths

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -745,7 +745,7 @@ const Subprocess = struct {
     });
 
     arena: std.heap.ArenaAllocator,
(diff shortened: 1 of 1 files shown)


mitchellh:
Not sure why CI isn't running here... I want to run CI so let me pull this down and try to kick it.

mitchellh (approved):
Looks good, will merge assuming CI passes.

PR #6920, opened 2025-03-26
snap: Build with gtk4-layer-shell and dependencies

kenvandine:
This vendors gtk4-layer-shell and dependencies into the snap to enable support in ghostty

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -64,24 +64,309 @@ parts:
     prime:
       - -*
 
+  buildenv:
+    plugin: nil
+    build-environment: &buildenv
+      - ACLOCAL_PATH: $CRAFT_STAGE/usr/share/aclocal
+      - XDG_DATA_DIRS: $CRAFT_STAGE/usr/share:/usr/share
+      - LD_LIBRARY_PATH: "$CRAFT_STAGE/usr/lib:\
+          $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR\
+          ${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
+      - PKG_CONFIG_PATH:
+          "$CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/pkgconfig:\
+          $CRAFT_STAGE/usr/lib/pkgconfig:\
+          $CRAFT_STAGE/usr/share/pkgconfig\
+          ${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}"
+      - GI_TYPELIB_PATH: "$CRAFT_STAGE/usr/lib/girepository-1.0:\
+          $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/girepository-1.0\
+          ${GI_TYPELIB_PATH:+:$GI_TYPELIB_PATH}"
+
+  ninja:
+    plugin: nil
+    source: https://github.com/ninja-build/ninja.git
+    source-tag: "v1.12.1"
+    source-depth: 1
+    override-build: |
+      rm -rf build
+      rm -f ninja
+      rm -f ninja_bootstrap
+      sed -i 's_^#!/usr/bin/env python$_#!/usr/bin/env python3_g' configure.py
+      ./configure.py --bootstrap
+      mv ninja ninja_bootstrap
+      rm -rf build
+      ./ninja_bootstrap
+      rm -f ninja_bootstrap
+      mkdir -p $CRAFT_PART_INSTALL/usr/bin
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -64,17 +64,222 @@ parts:
...
+    build-packages:
+      - python3
+    prime:
+      - -*
+
+  meson-deps:
+    after: [ ninja ]
+    plugin: nil
+    source: https://github.com/mesonbuild/meson.git
+    source-tag: '1.6.0'
```
jnsgruk, line 110: Same question, noting that `meson` 1.5 is in `noble-updates`?

snap/snapcraft.yaml
```diff
@@ -64,17 +64,222 @@ parts:
...
+      - zlib1g-dev
+      - liblzo2-dev
+      - libxml2-dev
+    prime:
+      - usr/lib/*/libcairo*.so.*
+
+  gtk4:
+    after: [meson-deps, cairo]
+    source: https://gitlab.gnome.org/GNOME/gtk.git
+    source-tag: '4.18.2'
```
jnsgruk, line 178: This is available on the `candidate` channel of the gnome-sdk snap: https://github.com/ubuntu/gnome-sdk/commit/50b8d157f3bf721b53ae186ade7ebcf1a3bcaf04

jnsgruk:
Actual changes look sound technically, my only question is whether or not it's worth waiting for the `latest/candidate` revision of the `gnome-46-2404` snap to be released, which contains much of the hand-build stuff you need?

kenvandine:
> Actual changes look sound technically, my only question is whether or not it's worth waiting for the `latest/candidate` revision of the `gnome-46-2404` snap to be released, which contains much of the hand-build stuff you need?
> 
> (EDIT: Ah, perhaps you can't use that with a `classic` snap...)

Correct, we can't leverage the content snap for classic

snap/snapcraft.yaml
```diff
@@ -64,17 +64,222 @@ parts:
...
+                          $CRAFT_STAGE/usr/share/pkgconfig\
+                          ${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}"
+      - GI_TYPELIB_PATH: "$CRAFT_STAGE/usr/lib/girepository-1.0:\
+                          $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/girepository-1.0\
+                          ${GI_TYPELIB_PATH:+:$GI_TYPELIB_PATH}"
+
+  ninja:
+    plugin: nil
+    source: https://github.com/ninja-build/ninja.git
+    source-tag: 'v1.12.1'
```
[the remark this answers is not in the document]
kenvandine: No, because the way snapcraft needs "meson-deps" internally

snap/snapcraft.yaml
kenvandine: it's possible that meson in noble-updates might work today, but it could stop working pretty quickly as gtk updates, so this makes it trivial to build app the components that gtk4 needs just by bumping tags as needed.

AnthonyZhOon:
I believe I've fixed this with #6706 , we zig build wasn't properly outputting gtk4-layer-shell.so to zig-out/lib/ which I fix.

jnsgruk:
@kenvandine do we need anything else here?

mkasberg:
I was under the impression snap builds from main are currently broken, and would hypothetically be fixed by either this or by #6706. (The build completes successfully, but the produced snap fails to start with a missing dependency on gtk4-layer-shell.)

I can reproduce this as follows:

1. Download [source-tarball](https://github.com/ghostty-org/ghostty/actions/runs/15334753512/artifacts/3224630313) from a recent build of main (as in the [test workflow](https://github.com/ghostty-org/ghostty/blob/9b45638c1557cce0e2670462215a120bd48db337/.github/workflows/test.yml#L376-L383), for `blueprint-compiler`, see #6831.)
2. Build the snap.
   
       unzip source-tarball.zip
       tar -xzf ghostty-source.tar.gz
       cd ghostty-1.1.4-main+d94bcda
       snapcraft
3. Install.
   
       sudo snap install --dangerous --classic ghostty_1.1.4-main+d94bcda_amd64.snap
4. Error:
   
       $ ghostty --version
       /usr/lib/x86_64-linux-gnu/gvfs/libgvfscommon.so: undefined symbol: g_variant_builder_init_static
       Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so
       /snap/ghostty/x1/bin/ghostty: error while loading shared libraries: libgtk4-layer-shell.so: cannot open shared object file: No such file or directory

I'm on Ubuntu 25.04, though I don't think that matters since this is a snap using core24.

Am I doing something wrong here?

mkasberg:
Just to close the loop here, this has been fixed by https://github.com/ghostty-org/ghostty/pull/7623

PR #6922, opened 2025-03-26
fix: add ( and ) to escape characters when dropping files in gtk

Iandenh:
Currently when dropping files in gtk with file names that include ( or ) will generate a bash error. 

With this change ( and ) will be escaped.

### DIFF
--- a/src/os/shell.zig
+++ b/src/os/shell.zig
@@ -23,6 +23,8 @@ pub fn ShellEscapeWriter(comptime T: type) type {
                     '?',
                     ' ',
                     '|',
+                    '(',
+                    ')',
                     => &[_]u8{ '\\', byte },
                     else => &[_]u8{byte},
                 };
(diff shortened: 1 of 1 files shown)


jcollie (approved):

src/os/shell.zig
```diff
@@ -93,3 +95,12 @@ test "shell escape 6" {
...
     try testing.expectEqualStrings("a\\\"c", fmt.getWritten());
 }
+
+test "shell escape 7" {
+    var buf: [128]u8 = undefined;
+    var fmt = std.io.fixedBufferStream(&buf);
+    var shell: ShellEscapeWriter(@TypeOf(fmt).Writer) = .{ .child_writer = fmt.writer() };
+    const writer = shell.writer();
+    try writer.writeAll("a(1)");
+    try testing.expectEqualStrings("a\(1\)", fmt.getWritten());
```
jcollie, line 105: Whoops! Need to double up those backslashes!

PR #6923, opened 2025-03-26
Revert "fix: add ( and ) to escape characters when dropping files in gtk"

jcollie:
Reverts ghostty-org/ghostty#6922

Enabled auto-merge too soon!

PR #6924, opened 2025-03-26
gtk: use up-to-date maximized & fullscreen state in syncAppearance

pluiedev:
DerivedConfig's maximize and fullscreen should only ever be used during window creation and nowhere else.


jcollie (changes requested):
LGTM but #6923 needs to get merged and then this needs rebased on top.

jcollie (approved):

PR #6940, opened 2025-03-28
Add Spanish translations (419 = Latin America) 

MiguelElGallo:
Hi!
I have followed the instructions and added translations for Spanish, Country Latin America 419 according to [this](https://localizely.com/locale-code/es-419/) and other online sources. 
In the provided links there was no references to 419, but I created it via the suggested command:
`msginit -i po/com.mitchellh.ghostty.pot -l es_419.UTF-8 -o "po/es_419.UTF-8.po" `
Glad to be able to contribute to this excellent product!
Cheers!


MiguelElGallo:
@00-kat Please could you take a look...

MiguelElGallo:
Also found in https://unstats.un.org/unsd/methodology/m49/ 
That 419 is Latin America and Caribean
image

MiguelElGallo:
Adde new commit with _BO (Bolivia) in case 419 does not work in some supported platforms.

00-kat (commented):
> @00-kat could you take a look, please :)

I know nothing about Spanish but I don't spot any punctuation issues or anything.

And as I previously mentioned, I know approximately nothing about locale codes so I can't confirm if 419 is definitely Latin America.

> Also found in https://unstats.un.org/unsd/methodology/m49/

The source is definitely reputable so I doubt it's wrong though.  I just don't know if gettext would like it.

MiguelElGallo:
I can also commit to maintenance of the translation.

Uzaaft:
just waiting on another speaker of Spanish reviewing for QA and we should be ready to go. 

> I can also commit to maintenance of the translation.

Awesome! 🚀

piedrahitac (approved):
Native spanish speaker here... This looks good!

pluiedev:
I'm not sure if es_419 would work... it has to match whatever locale code is set in the user's LANGUAGE environment variable, and to my knowledge they all look like es_MX, es_BO, es_AR, etc. We might need either some special logic to map those locales to one translation, or to duplicate them entirely, which is *potentially* beneficial for allowing more regionalized expressions but I don't know how much more maintenance burden that would bring.

MiguelElGallo:
Yeah, not sure either if 419 will work, but it is [something.](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.4) (See point 4)
This URL is also interesting: https://translate.fedoraproject.org/languages/es_419/. This means there are some Linux that support 419. 
I think would be good to have this merged.

MiguelElGallo:
Yeah, not sure either if 419 will work, but it is [something.](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.4) (See point 4)
This URL is also interesting: https://translate.fedoraproject.org/languages/es_419/. This means there are some linux that support 419. 
I think would be good to have this merged.

jacostag:
Spanish speaker here... This looks good!

mitchellh (approved):
Just because I'm uncertain about the 419, I'm going to remove that so we can move forward with merging. If we can get clarity on other projects and what they do about that I'd be happy to look at another PR.

00-kat:
xref #8460.

PR #6969, opened 2025-03-31
snap: Cleanup more environment variable set by launcher script

kenvandine:
This unsets a few more environment variables that were set by the launcher script which could potentially effect other apps.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -2374,6 +2374,9 @@ pub fn defaultTermioEnv(self: *Surface) !std.process.EnvMap {
         env.remove("GDK_PIXBUF_MODULEDIR");
         env.remove("GDK_PIXBUF_MODULE_FILE");
         env.remove("GTK_PATH");
+        env.remove("XDG_CONFIG_HOME");
+        env.remove("XDG_CACHE_HOME");
+        env.remove("XDG_DATA_HOME");
     }
 
     if (self.container.window()) |window| {


src/apprt/gtk/Surface.zig
```diff
@@ -2359,6 +2359,9 @@ pub fn defaultTermioEnv(self: *Surface) !std.process.EnvMap {
         env.remove("GDK_PIXBUF_MODULEDIR");
         env.remove("GDK_PIXBUF_MODULE_FILE");
         env.remove("GTK_PATH");
+        env.remove("XDG_CONFIG_HOME");
+        env.remove("XDG_CACHE_HOME");
+        env.remove("XDG_DATA_HOME");
```
tristan957, line 2364: I'm not sure this is safe. User or system could set this via systemd and wouldn't the launcher script inherit those paths?

tristan957 (commented):
Your git history is messed up

src/apprt/gtk/Surface.zig
kenvandine: __

> I'm not sure this is safe. User or system could set this via systemd and wouldn't the launcher script inherit those paths?

I think this should be fine. If a user is setting them in their own environment, they will get set when their shell starts. It's just environment variables that are set log enough to get ghostty running, before their shell starts.

kenvandine:
> Your git history is messed up

Yeah, I already fixed it.

src/apprt/gtk/Surface.zig
tristan957: I don't set these variables in my shell init script. I set them in systemd.

mitchellh: I'm also not sure about this, because the user may not be launching a shell, they can launch any `command` that could rely on these env vars, no?

mitchellh: This is the blocker for this PR, any feedback?

mitchellh:
I'm going to close this for now since the feedback wasn't addressed. Feel free to reopen or respond to my comment if we get more info.

kenvandine:
> I'm going to close this for now since the feedback wasn't addressed. Feel free to reopen or respond to my comment if we get more info.

Thanks, I'm not sure it was needed so if it comes up again I'll investigate then.

PR #6978, opened 2025-04-02
gtk(x11): fix blur regions when using >200% scaling

pluiedev:
See #6957

We were not considering GTK's internal scale factor that converts between "surface coordinates" and actual device coordinates, and that worked fine until the scale factor reached 2x (200%).

Since the code is now dependent on the scale factor (which could change at any given moment), we also listen to scale factor changes and then unconditionally call `winproto.syncAppearance`. Even though it's somewhat overkill, I don't expect people to change their scale factor dramatically all the time anyway...

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -281,6 +281,15 @@ pub fn init(self: *Window, app: *App) !void {
             .detail = "is-active",
         },
     );
+    _ = gobject.Object.signals.notify.connect(
+        self.window,
+        *Window,
+        gtkWindowUpdateScaleFactor,
+        self,
+        .{
+            .detail = "scale-factor",
+        },
+    );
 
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/Window.zig
```diff
@@ -784,6 +793,24 @@ fn gtkWindowNotifyIsActive(
     }
 }
 
+fn gtkWindowUpdateScaleFactor(
+    _: *adw.ApplicationWindow,
+    _: *gobject.ParamSpec,
+    self: *Window,
+) callconv(.C) void {
```
jcollie, line 800: nitpick: Zig 0.14 is switching this to lowercase `callconv(.c)`.

jcollie (approved):
one nitpick, and one typo to fix

jcollie (approved):

Issue #4323, opened 2025-01-01
Kitty images aren't scrolled with the text


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4315

<div type='discussions-op-text'>

<sup>Originally posted by **sergei-grechanik** January  1, 2025</sup>
Images placed using classic placement commands do not follow the text when it's scrolled with CSI sequences. To reproduce, run this script with an image as the argument:
```shell
#!/bin/sh

IMG="$(realpath $1)"

printf '\033[100S' # Scroll up by 100
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

printf '========= Image ============\n'
printf '\033_Gi=1,t=f,q=1,f=100,a=T,r=10,c=20;%s\033\\' "$(printf '%s' "$IMG" | base64 -w0)"
printf '<- image\n'
printf '1\n'
printf '2\n'
printf '3\n'

sleep 1

printf '\033[4T' # Scroll down by 4
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

sleep 5
```

`ghostty_bug_scroll.sh images/icons/icon_256.png` when running from kitty (the expected behavior):
![image](https://github.com/user-attachments/assets/b3e95f08-0ba5-4ae9-82ab-8bd80e14d7c9)

When running from ghostty:
![image](https://github.com/user-attachments/assets/baaee95d-b29d-4bc7-a4fc-59ad520715fe)
</div>

henriquevmac:
Hey, I have been looking at this bug and working on it. I was wondering if I could be assigned to it?

PR #6979, opened 2025-04-02
Fix ghostty-org#4323: Ensure Kitty images scroll with text

henriquevmac:
Previously, Kitty images did not scroll as shown in #4323 when the terminal was scrolled using ANSI commands (\033[xT for scrolling x down and \033[xS for scrolling x up).

This fix updates the insertLines and deleteLines methods in src/terminal/Terminal.zig to iterate over all kitty_images placements, updating their positions or deleting them if their top-left corner (the Pin position) scrolls off-screen.

Added four unit tests (two per method):
-  One verifies that image positions update correctly when scrolled up or down.

- The other ensures images are deleted when they scroll out of view.

I ran some tests as shown in this [video](https://youtu.be/R3NrFEN3Ao0)

The three scripts run where this:
- test_script:
```
#!/bin/sh

IMG="$(realpath $1)"

printf '\033[100S' # Scroll up by 100
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

printf '========= Image ============\n'
printf '\033_Gi=1,t=f,q=1,f=100,a=T,r=10,c=20;%s\033\\' "$(printf '%s' "$IMG" | base64 -w0)"
printf '<- image\n'
printf '1\n'
printf '2\n'
printf '3\n'

sleep 1

printf '\033[4T' # Scroll down by 4
sleep 1
printf '\033[4S' # Scroll up by 4
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

sleep 5
```

- test_script_out_bottom:
```
#!/bin/sh

IMG="$(realpath $1)"

printf '\033[100S' # Scroll up by 100
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

printf '========= Image ============\n'
printf '\033_Gi=1,t=f,q=1,f=100,a=T,r=10,c=20;%s\033\\' "$(printf '%s' "$IMG" | base64 -w0)"
printf '<- image\n'
printf '1\n'
printf '2\n'
printf '3\n'

sleep 1

printf '\033[100T' # Scroll down by 100
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

sleep 5
```

- test_script_out_top:

```
#!/bin/sh

IMG="$(realpath $1)"

printf '\033[100S' # Scroll up by 100
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

printf '========= Image ============\n'
printf '\033_Gi=1,t=f,q=1,f=100,a=T,r=10,c=20;%s\033\\' "$(printf '%s' "$IMG" | base64 -w0)"
printf '<- image\n'
printf '1\n'
printf '2\n'
printf '3\n'

sleep 1

printf '\033[4S' # Scroll up by 4
printf '\033[1d' # Move cursor to row 0
printf '\033[1G' # Move cursor to column 0

sleep 5
```

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -1451,9 +1451,6 @@ pub fn insertLines(self: *Terminal, count: usize) void {
         self.screen.cursor.x < self.scrolling_region.left or
         self.screen.cursor.x > self.scrolling_region.right) return;
 
-    // Scrolling dirties the images because it updates their placements pins.
-    self.screen.kitty_images.dirty = true;
-
     // At the end we need to return the cursor to the row it started on.
     const start_y = self.screen.cursor.y;
     defer {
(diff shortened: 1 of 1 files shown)


qwerasd205 (changes requested):
I feel like this should probably account for all tracked pins, not just the ones used for image placements, since we also use tracked pins for, for example, selection endpoints - anything that needs to follow content uses a tracked pin.

Additionally, your pin adjustment logic doesn't account for the left/right scroll margin, as I noted before -- it should, and tests should be added for that as well.

(Side note: You'll find that all the existing tracked pin adjustment logic lives in `PageList`, and operations in `Screen` rely on the `PageList` operations behaving appropriately - so doing this all the way in `Terminal` is unfortunate, but it needs a refactor to avoid; as my note above `insertLines` explains, this operation really should belong to `Screen` not `Terminal`, but it's fine for now to do it here until that gets refactored.)

qwerasd205:
Hi, sorry for the delay in getting back to you. Yes, iterate over all tracked pins (for examples search for `pin_keys` in `PageList.zig`), and account for the left/right scroll margins in the same way as is done when deciding how to move the cell contents in the `insertLines`/`deleteLines` functions.

mitchellh:
Closing due to a delay in addressing feedback. It's no problem, I understand things come up and this is volunteer open source. I'd still like to see this resolved but we can do so in a future PR.

PR #6987, opened 2025-04-03
docs: use Command instead of super for macOS

simono:
Command is the name Apple uses for this key and that's printed on the keyboard 😉

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -826,7 +826,7 @@ env: RepeatableStringMap = .{},
 link: RepeatableLink = .{},
 
 /// Enable URL matching. URLs are matched on hover with control (Linux) or
-/// super (macOS) pressed and open using the default system application for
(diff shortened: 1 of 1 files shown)


src/config/Config.zig
```diff
@@ -826,7 +826,7 @@ env: RepeatableStringMap = .{},
 link: RepeatableLink = .{},
 
 /// Enable URL matching. URLs are matched on hover with control (Linux) or
-/// super (macOS) pressed and open using the default system application for
+/// Command (macOS) pressed and open using the default system application for
```
00-kat, line 829: Why is it upper case?  That looks odd.
```suggestion
/// command (macOS) pressed and open using the default system application for
```

00-kat (changes requested):

00-kat (approved):

mitchellh (approved):

PR #6998, opened 2025-04-04
macos: left mouse click while not focused doesn't encode to pty

### LINKED ISSUE #2595: macOS: clicking unfocused window should not encode VT mouse event
See #2593 for detailed discussion including comparisons to other terminals.

Steps to reproduce:

1. Open Ghostty and launch `htop` (so we have click targets)
2. Focus another macOS application while having the Ghostty window visible in the background
3. Click a process in Ghostty

What you'll see is that the row is highlighted. Ghostty comes to focus _and_ encodes the pty event so htop highlights the row. 

**What should happen:** If the surface is unfocused or the app is unfocused, a click should bring it (the surface _and_ window) to focus but _not_ generate a pty event.

mitchellh:
Fixes #2595

This fixes an issue where a left mouse click on a terminal while not focused would subsequently be encoded to the pty as a mouse event. This is atypical for macOS applications in general and wasn't something we wanted to do.

We do, however, want to ensure our terminal gains focus when clicked without focus. Specifically, a split. This matches iTerm2 behavior and is rather nice. We had this behavior before but our logic to make this work before caused the issue this commit is fixing.

I also tested this with command+click which is a common macOS shortcut to emit a mouse event without raising the focus of the target window. In this case, we will properly focus the split but will not encode the mouse event to the pty. I think we actually do a _better job_ here tha iTerm2 (but, subjective) because we do encode the pty event properly if the split is focused whereas iTerm2 never does.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -201,7 +201,14 @@ extension Ghostty {
             self.eventMonitor = NSEvent.addLocalMonitorForEvents(
                 matching: [
                     // We need keyUp because command+key events don't trigger keyUp.
-                    .keyUp
+                    .keyUp,
+
+                    // We need leftMouseDown to determine if we should focus ourselves
+                    // when the app/window isn't in focus. We do this instead of
+                    // "acceptsFirstMouse" because that forces us to also handle the
+                    // event and encode the event to the pty which we want to avoid.
+                    // (Issue 2595)
+                    .leftMouseDown,
                 ]
             ) { [weak self] event in self?.localEventHandler(event) }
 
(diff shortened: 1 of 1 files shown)


edrozenberg:
Right click on an unfocused Ghostty window (or Quick Term) causes "empty" box cursors in all splits, workaround is to focus on some other window then focus back to the Ghostty window. Believe this issue was caused by this change.

https://github.com/user-attachments/assets/05de6b6c-13fe-4b15-9779-9c6c8fcca96f

edrozenberg:
Another bad side-effect - a right click followed by launching tmux, for ex `tmux new -s test` - moving the mouse does a selection, it appears as if the left mouse button is being pressed and held down.

mitchellh:
I'm unable to reproduce either of these before or after this change. Please open a discussion and we can debug further.

edrozenberg:
Will do thanks! I'm seeing the same even with an older version `6f1b22ac` b-`9944`, so it's some other issue, my bad.

PR #7020, opened 2025-04-07
Fix macOS shortcut binding for `close_window` action

### LINKED ISSUE #7003: macOS: `super+w=close_window` isn't working
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6996

<sup>Originally posted by **Tavish9** April  4, 2025</sup>
```
keybind = super+w=close_window
keybind = super+k>w=close_surface
```

According to [doc](https://ghostty.org/docs/config/keybind/reference#close_window), the whole window should disappear when i type `super+w`. However, only one surface(one typing area) is closed, which behaves the same as `close_surface`.

Why?
</div>

### Reproduction

```
keybind = super+w=close_window
```

1. Open Ghostty
2. Pretty `super+w`
3. Nothing happens.

When fixing this we should also verify that it properly closes the full window and not just a single surface either.

liby:
Fixes #7003

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -91,6 +91,12 @@ class TerminalController: BaseTerminalController {
             name: Ghostty.Notification.didEqualizeSplits,
             object: nil
         )
+        center.addObserver(
+            self,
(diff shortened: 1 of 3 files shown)


mitchellh (approved):
Oh that's super funny we just never implemented this action. Looks good!

PR #7021, opened 2025-04-07
shell-integration: Fix condition for sudo

faho:
A missing ";" meant the check for $TERMINFO was never executed.


jparise (approved):
Thank you!

PR #7025, opened 2025-04-07
Use built-in functions for elvish

hqnna:
Currently the elvish shell integration uses the `hostname` command, this may not exist on all systems and is somewhat redundant to rely on when elvish has an [`platform:hostname`](https://elv.sh/ref/platform.html#platform:hostname) function that can be used and will work across platform regardless of command availability. On top of this instead of directly calling the `pwd` function we can simply use the built-in [`$pwd`](https://elv.sh/ref/builtin.html#$pwd) variable that elvish gives us. This should prevent the shell integration from breaking due to external function availability.

### DIFF
--- a/src/shell-integration/elvish/lib/ghostty-integration.elv
+++ b/src/shell-integration/elvish/lib/ghostty-integration.elv
@@ -75,7 +75,8 @@
   }
 
   fn report-pwd {
-    printf "\e]7;file://%s%s\a" (hostname) (pwd)
+    use platform
+    printf "\e]7;file://%s%s\a" (platform:hostname) ($pwd)
   }
 
   fn sudo-with-terminfo {|@args|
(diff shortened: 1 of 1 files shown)


src/shell-integration/elvish/lib/ghostty-integration.elv
```diff
@@ -75,7 +75,8 @@
   }
 
   fn report-pwd {
-    printf "\e]7;file://%s%s\a" (hostname) (pwd)
+    use platform
+    printf "\e]7;file://%s%s\a" platform:hostname $pwd
```
jparise, line 79: Separately, OSC 7 with the `file://` scheme expects the URL to be percent-encoded. If there isn't a convenient way to do that in elvish, we should switch this to the `kitty-shell-cwd://` scheme, which does not use percent encoding.

jparise (approved):
Nice! I double-checked the elvish reference, and these changes look good. I imagine this will yield a small per-command performance improvement, too.

src/shell-integration/elvish/lib/ghostty-integration.elv
jparise: #7033 makes this improvement

PR #7029, opened 2025-04-08
pwd: fix hostname resolution on macos

reykjalin:
## Description

Yet another edge case in #2484 

When macOS's "Private WiFi address" feature is enabled it'll change the hostname to a mac address. Mac addresses look like URIs with a hostname and port component, e.g. `12:34:56:78:90:12` where `:12` looks like port `12`. However, mac addresses use hex numbers and as such can also contain letters `a` through `f`. So, a mac address like `ab:cd:ef:ab:cd:ef` is valid, but will not be parsed as a URI, because `:ef` is not a valid port.

This commit attempts to fix that by checking if the hostname is a valid mac address when `std.Uri.parse()` fails and constructing a new `std.Uri` struct using that information.

It's not perfect, but is equally compliant with the URI spec as `std.Uri` currently is. Meaning not at all compliant 😅 

### Unit tests

> [!IMPORTANT]
> I don't know if these tests are run in CI or if they're picked up by `zig build test`. I get an unrelated crash that mentions `minidump` and an invalid OSC command when I try to run `zig build test` on my mac.

1. Make sure `zig test src/os/hostname.zig` is passing.

#### Setup - Enable the "Private WiFi address" setting

> [!IMPORTANT]
> You must be connected to WiFi to be able to test this.

1. Open your mac's "System Settings".
2. Go to Network &rarr; Wi-Fi &rarr; Details.

image

3. Set the "Private Wi-Fi address" setting to `Rotating`.

image

> [!IMPORTANT]
> Now you wait. The private Wi-Fi address will eventually rotate to a mac address that ends with a non-digit, e.g. `0a`, `ff`, `e2`, etc. You'll notice this when your shell integration stops working, e.g. you open a new tab in Ghostty and the shell is in your home directory instead of whichever directory you had open in your previous tab.

#### Testing the changes

1. Open Ghostty.
3. `cd` to any directory that isn't the default (usually `$HOME`) directory, e.g. `cd Documents`.
4. Open a new tab (<kbd>Cmd+T</kbd>) or split (<kbd>Cmd+D</kbd>).
5. Assuming the setup steps have been followed you should:
    * On `main`:  land in `$HOME` in the new tab or split.
    * On this branch: land in the same working directory as the original tab or split.

### DIFF
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -199,6 +199,40 @@ test parseUrl {
     try std.testing.expectEqualStrings("ab:cd:ef:ab:cd:ef", uri.host.?.percent_encoded);
     try std.testing.expectEqualStrings("/home/test/", uri.path.percent_encoded);
     try std.testing.expect(uri.port == null);
+
+    // 3. Hostnames that are mac addresses with no path.
+
+    // Numerical mac addresses.
+
+    uri = try parseUrl("file://12:34:56:78:90:12");
+
+    try std.testing.expectEqualStrings("file", uri.scheme);
+    try std.testing.expectEqualStrings("12:34:56:78:90", uri.host.?.percent_encoded);
+    try std.testing.expectEqualStrings("", uri.path.percent_encoded);
+    try std.testing.expect(uri.port == 12);
+
+    uri = try parseUrl("kitty-shell-cwd://12:34:56:78:90:12");
+
+    try std.testing.expectEqualStrings("kitty-shell-cwd", uri.scheme);
+    try std.testing.expectEqualStrings("12:34:56:78:90", uri.host.?.percent_encoded);
+    try std.testing.expectEqualStrings("", uri.path.percent_encoded);
+    try std.testing.expect(uri.port == 12);
+
+    // Alphabetical mac addresses.
+
+    uri = try parseUrl("file://ab:cd:ef:ab:cd:ef");
+
+    try std.testing.expectEqualStrings("file", uri.scheme);
+    try std.testing.expectEqualStrings("ab:cd:ef:ab:cd:ef", uri.host.?.percent_encoded);
+    try std.testing.expectEqualStrings("", uri.path.percent_encoded);
+    try std.testing.expect(uri.port == null);
+
+    uri = try parseUrl("kitty-shell-cwd://ab:cd:ef:ab:cd:ef");
+
+    try std.testing.expectEqualStrings("kitty-shell-cwd", uri.scheme);
+    try std.testing.expectEqualStrings("ab:cd:ef:ab:cd:ef", uri.host.?.percent_encoded);
+    try std.testing.expectEqualStrings("", uri.path.percent_encoded);
+    try std.testing.expect(uri.port == null);
 }
 
 test "parseUrl succeeds even if path component is missing" {
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -19,9 +19,7 @@ fn isValidMacAddress(mac_address: []const u8) bool {
         return false;
     }
 
-    for (0..mac_address.len) |i| {
-        const c = mac_address[i];
-
+    for (mac_address, 0..) |c, i| {
         if ((i + 1) % 3 == 0) {
             if (c != ':') {
                 return false;
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const posix = std.posix;
 
 pub const HostnameParsingError = error{
@@ -40,6 +41,10 @@ fn isValidMacAddress(mac_address: []const u8) bool {
 /// correctly.
 pub fn parseUrl(url: []const u8) !std.Uri {
     return std.Uri.parse(url) catch |e| {
+        // The mac-address-as-hostname issue is specific to macOS so we just return an error if we
+        // hit it on other platforms.
+        comptime if (builtin.os.tag != .macos) return e;
+
         // It's possible this is a mac address on macOS where the last 2 characters in the
         // address are non-digits, e.g. 'ff', and thus an invalid port.
         //
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -48,7 +48,7 @@ pub fn parseUrl(url: []const u8) UrlParsingError!std.Uri {
     return std.Uri.parse(url) catch |e| {
         // The mac-address-as-hostname issue is specific to macOS so we just return an error if we
         // hit it on other platforms.
-        comptime if (builtin.os.tag != .macos) return e;
+        if (comptime builtin.os.tag != .macos) return e;
 
         // It's possible this is a mac address on macOS where the last 2 characters in the
         // address are non-digits, e.g. 'ff', and thus an invalid port.
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -12,6 +12,8 @@ pub const UrlParsingError = std.Uri.ParseError || error{
     NoSchemeProvided,
 };
 
+const mac_address_length = 17;
+
 fn isUriPathSeparator(c: u8) bool {
     return switch (c) {
         '?', '#' => true,
@@ -65,7 +67,6 @@ pub fn parseUrl(url: []const u8) UrlParsingError!std.Uri {
         // The first '/' after the scheme marks the end of the hostname. If the first '/'
         // following the end of the scheme is not at the right position this is not a
         // valid mac address.
-        const mac_address_length = 17;
         if (url_without_scheme.len != mac_address_length and
             std.mem.indexOfScalarPos(u8, url_without_scheme, 0, '/') != mac_address_length)
         {
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -55,20 +55,21 @@ pub fn parseUrl(url: []const u8) !std.Uri {
         // The first '/' after the scheme marks the end of the hostname. If the first '/'
         // following the end of the scheme is not at the right position this is not a
         // valid mac address.
-        if (url_without_scheme.len != 17 and
-            std.mem.indexOfScalarPos(u8, url_without_scheme, 0, '/') != 17)
+        const mac_address_length = 17;
+        if (url_without_scheme.len != mac_address_length and
+            std.mem.indexOfScalarPos(u8, url_without_scheme, 0, '/') != mac_address_length)
         {
             return error.HostnameIsNotMacAddress;
         }
 
         // At this point we may have a mac address as the hostname.
-        const mac_address = url_without_scheme[0..17];
+        const mac_address = url_without_scheme[0..mac_address_length];
 
         if (!isValidMacAddress(mac_address)) {
             return error.HostnameIsNotMacAddress;
         }
 
-        var uri_path_end_idx: usize = 17;
+        var uri_path_end_idx: usize = mac_address_length;
         while (uri_path_end_idx < url_without_scheme.len and
             !isUriPathSeparator(url_without_scheme[uri_path_end_idx]))
         {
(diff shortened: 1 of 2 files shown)


jparise (commented):
Two quick thoughts:

1. This is probably getting complicated enough to warrant a separate utility function (with tests), perhaps outside of the stream handler.
2. I'm starting to feel like we should follow Kitty's example and treat the `file://` and `kitty-report-cwd://` schemes differently: only `file://` URLs get the encoding. (See #7033)

jcollie:
It seems to me that this would be better solved on the macOS level. Either by using `scutil` to force a "real" hostname or modifying the shell integration to inject a better hostname into the OSC 7 escape.

reykjalin:
> It seems to me that this would be better solved on the macOS level. Either by using scutil to force a "real" hostname or modifying the shell integration to inject a better hostname into the OSC 7 escape.

I think the main problem is that [posix.gethostname()](https://ziglang.org/documentation/0.14.0/std/#std.posix.gethostname) - the hostname we compare to whatever the shell integration passes to us - returns the mac address version of the hostname.

I suppose we _could_ run `scutil --get LocalHostName` on macOS instead of calling `posix.gethostname()` and make sure the shell integration is getting a similar value for us to compare 🤔 
But it also feels really bad to have to run an external executable on macOS to get this working right.

And either way I think we should treat mac addresses as valid hostnames, even if they're not compliant with the [general linux hostname spec](https://www.man7.org/linux/man-pages/man7/hostname.7.html) of `[a-z0-9]` and a hyphen (`-`) due to using colons (`:`). This feature is enabled by default in macOS, and it isn't that big of a lift to support _specifically only valid mac addresses_, even if it does complicate this particular part of the codebase somewhat.

reykjalin:
Finally found got to a place where I can reproduce it and recorded a video just to demonstrate:


https://github.com/user-attachments/assets/a19537d5-1429-4f8b-9195-c6bacbd494d1

reykjalin:
And now to demonstrate that the current set of changes work:

https://github.com/user-attachments/assets/26cf4bdb-b3d0-4e06-94ef-3298c602d26f

reykjalin:
I still need to clean up the code here more before this is ready. I'm thinking of creating some sort of utility function in `src/os/hostname.zig` that will extract a hostname from a file url string passed to it. We can then add some tests for that function so we can be sure we're doing this right.

It's way too late here to do that now though, so I'll pick this up again when I have the time, hopefully in the next week or so.

reykjalin:
@jparise I think this is ready for a proper review now :)

src/os/hostname.zig
```diff
@@ -6,6 +6,91 @@ pub const HostnameParsingError = error{
...
+            };
+
+            return error.UnsupportedScheme;
+        };
+
+        // The first '/' after the scheme marks the end of the hostname. If the first '/'
+        // following the end of the scheme is not at the right position this is not a
+        // valid mac address.
+        if (std.mem.indexOfScalarPos(u8, url_without_scheme, 0, '/') != 17 and
+            url_without_scheme.len != 17)
```
jparise, line 65: Could `url_without_scheme.len != 17` be evaluated first? Seems like a cheap condition to check to see if we fall into the MAC address case.

The `17`s also seem like a magic number we could consider naming as a constant.

src/os/hostname.zig
```diff
@@ -6,6 +6,91 @@ pub const HostnameParsingError = error{
...
+    }
+
+    return true;
+}
+
+/// Parses the provided url to a `std.Uri` struct. This is very specific to getting hostname and
+/// path information for Ghostty's PWD reporting functionality. Takes into account that on macOS
+/// the url passed to this function might have a mac address as its hostname and parses it
+/// correctly.
+pub fn parseUrl(url: []const u8) !std.Uri {
```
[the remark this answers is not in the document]
reykjalin: Agh, yeah I was thinking about this the entire time I was working on this, but forgot to make an explicit error set 🤦 
I'll add it, nice catch!

src/os/hostname.zig
```diff
@@ -6,6 +6,91 @@ pub const HostnameParsingError = error{
...
+        //
+        // Example: file://12:34:56:78:90:12/path/to/file
+        if (e != error.InvalidPort) return e;
+
+        const scheme, const url_without_scheme = url: {
+            if (std.mem.startsWith(u8, url, "file://")) break :url .{ "file", url[7..] };
+            if (std.mem.startsWith(u8, url, "kitty-shell-cwd://")) break :url .{
+                "kitty-shell-cwd",
+                url[18..],
+            };
```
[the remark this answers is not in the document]
reykjalin: No, I think we can remove this. It's also checked in `stream_handler` so I don't think we need to worry about this here 👍

src/os/hostname.zig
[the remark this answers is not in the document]
reykjalin: Fixed in d6b99a642e72bec576ce2f97f87247c0e1735730

src/os/hostname.zig
```diff
@@ -6,6 +6,91 @@ pub const HostnameParsingError = error{
...
+}
+
+fn isValidMacAddress(mac_address: []const u8) bool {
+    // A valid mac address has 6 two-character components with 5 colons, e.g. 12:34:56:ab:cd:ef.
+    if (mac_address.len != 17) {
+        return false;
+    }
+
+    for (0..mac_address.len) |i| {
+        const c = mac_address[i];
```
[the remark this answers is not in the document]
reykjalin: fixed in 8e689b03747fdfc129de55149a4e2188dd63fe24

src/os/hostname.zig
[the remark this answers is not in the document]
reykjalin: Fixed in 534bdea9130521b2886a6919e5cbc8ced849dbb9

src/os/hostname.zig
```diff
@@ -6,6 +6,91 @@ pub const HostnameParsingError = error{
...
+                return false;
+            }
+        }
+    }
+
+    return true;
+}
+
+/// Parses the provided url to a `std.Uri` struct. This is very specific to getting hostname and
+/// path information for Ghostty's PWD reporting functionality. Takes into account that on macOS
```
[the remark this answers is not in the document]
reykjalin: Hmmm, there's an argument to be made that _in the future_ (or maybe even now?) a computer's hostname could be manually set to a mac address, so this might be useful as a generic handler 🤔 

I don't have any strong opinions here though. A `comptime` check seems good here at this point, so I added one in 5ab56467b4954143266d0e67d2cbb6cf27af3352.

src/os/hostname.zig
reykjalin: Oh, nice catch! Fixed the if-statement order in 550321375fe80573a9acbe9712440338efed478d and renamed the magic constant in 5fa0fb0a1c7c9a314ce51214d833ca9c078a8970 🙌

src/os/hostname.zig
```diff
@@ -1,11 +1,106 @@
...
+        const url_without_scheme_start = std.mem.indexOf(u8, url, "://") orelse {
+            return error.NoSchemeProvided;
+        };
+        const scheme = url[0..url_without_scheme_start];
+        const url_without_scheme = url[url_without_scheme_start + 3 ..];
+
+        // The first '/' after the scheme marks the end of the hostname. If the first '/'
+        // following the end of the scheme is not at the right position this is not a
+        // valid mac address.
+        const mac_address_length = 17;
```
jparise, line 71: This could be moved to file scope and used within `isValidMacAddress`. I suppose we could also move the MAC-specific bits to their own `mac.zig`, but I don't know if that's particularly useful yet given this is the only place we use those routines.

src/os/hostname.zig
```diff
@@ -1,11 +1,106 @@
...
     NoSpaceLeft,
 };
 
+pub const UrlParsingError = error{
+    HostnameIsNotMacAddress,
+    InvalidFormat,
+    InvalidPort,
+    NoSchemeProvided,
+    UnexpectedCharacter,
+};
```
jparise, line 16:
```suggestion
pub const UrlParseError = std.Uri.ParseError || error{
    HostnameIsNotMacAddress,
    NoSchemeProvided,
};
```

src/os/hostname.zig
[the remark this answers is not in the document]
jparise: Yeah, similarly no strong feelings. I think the `comptime` check makes sense if this is macOS-specific, but if it's safe/reasonable for all platforms, we don't need it because the extra work is only in the error path.

jparise (commented):
Looks good to me but deferring to others who have more regular maintainership over these files.

src/os/hostname.zig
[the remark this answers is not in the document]
reykjalin: let's keep it as a `comptime` check until there's a need for this check on other platforms :)

src/os/hostname.zig
reykjalin: oh, good point! Moved in 6ed94b00346a402d85ac83a9585de17ced93f351 🎉 

> I suppose we could also move the MAC-specific bits to their own `mac.zig`, but I don't know if that's particularly useful yet given this is the only place we use those routines.

Yeah I deliberately didn't export those functions since they're specific to this particular use in just this file. I'm happy to make them a module but I don't think there's a benefit to it at this point 🤷‍♂️

src/os/hostname.zig
reykjalin: I didn't know this was possible, thanks! Fixed in 73e5f7e5d6d0ce3bfe27a1d791585764cb17c536

reykjalin:
Sorry for the long response time here. I've rebased the branch on `main` and addressed the previous review, so this should be ready for another review 🎉

src/os/hostname.zig
```diff
@@ -1,11 +1,104 @@
...
+        return false;
+    }
+
+    for (mac_address, 0..) |c, i| {
+        if ((i + 1) % 3 == 0) {
+            if (c != ':') {
+                return false;
+            }
+        } else {
+            if (!std.mem.containsAtLeastScalar(u8, "0123456789ABCDEFabcdef", 1, c)) {
```
jparise: `else if` ?

jparise (approved):

src/os/hostname.zig
reykjalin: omg yeah, nice catch. fixed in 31e386afa6e4874be266ea083c773ffac9e6168e

jparise (approved):

src/os/hostname.zig
```diff
@@ -1,11 +1,102 @@
...
+
+/// Parses the provided url to a `std.Uri` struct. This is very specific to getting hostname and
+/// path information for Ghostty's PWD reporting functionality. Takes into account that on macOS
+/// the url passed to this function might have a mac address as its hostname and parses it
+/// correctly.
+pub fn parseUrl(url: []const u8) UrlParsingError!std.Uri {
+    return std.Uri.parse(url) catch |e| {
+        // The mac-address-as-hostname issue is specific to macOS so we just return an error if we
+        // hit it on other platforms.
+        if (comptime builtin.os.tag != .macos) return e;
```
reykjalin, line 51: @jparise I changed this from `comptime if (…` to `if (comptime …` to fix the build failure on non macOS systems, but I'm not familiar enough with Zig's comptime to know if it makes sense to even keep the `comptime` check in here? Should we just remove it, or does it still make sense to keep it? 🤔

jparise: In its current state, I don't think we need to gate this on macOS. It's done in a generic way, although it would be good to get some real-world confirmation that this all makes sense on other platforms (e.g. no false positives/negatives).

reykjalin: I did a very simple test run in a Fedora VM and this seemed to be working correctly. It was the same thing I did in https://github.com/ghostty-org/ghostty/pull/7029#issuecomment-2844073532 to confirm this was was working correctly. Super simple test so I'm not sure if we want something more thorough.

Anything I can do here to help with testing this on other platforms?

jparise: @jcollie and @pluiedev, do you have any opinions on how this will play on Linux?

pluiedev: Absolutely no clue

PR #7044, opened 2025-04-10
config: allow commands to specify whether they shell expand or not

### LINKED ISSUE #7032: Allow command to be exec'd directly (without shell-wrapping), make the default for `-e`
Discussion: #2633

The original issue as stated was that the `-e` command is not correctly parsed. `-e` comes from a shell or manual exec call which by definition already gives us a nice array of strings. I agree in this scenario there is no reason to run this through `/bin/sh`. 

Going further into implementation details, `-e` is implemented by being syntactic sugar over `initial-command` (and some other configs that are irrelevant for this issue and documented for that config). 

I think that running a command through `/bin/sh -c` by default for `command/initial-command` _in the config file_ results in the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). But for `-e`, the opposite is true. 

## Proposal

What I'd like to propose is some syntax for `command/initial-command` to NOT run it through `/bin/sh -c` and for `-e` to default to this behavior.

The syntax I've proposed is to prefix the command with either `direct:` or `shell:` to indicate whether a command is executed directly via `exec()` or wrapped with `/bin/sh -c`, respectively. The prefix form matches other modifiers that we have in our config, mostly in keybindings but used anywhere we have a tagged union internally.

For the `direct:` case, we will NOT do any shell expansion. We accept the args space-separated as-is. That means the following would NOT work: `! nvim "file with spaces"` (you'd have to drop the `!`). But, since this is an opt-in, the user is shooting themselves in the foot.  **To be clear, this would work with `-e` because it has a different default, so the original issue would be resolved.**

For the _untagged_ case, I'm going to leave it defined as Ghostty choosing based on what it thinks is best. This gives us the most flexibility in the future to make a decision. Initially, I think it's fair to run everything through `shell` since that's the most backwards compatible way.

## Examples

```
command = nvim "a b c"
command = shell: nvim "a b c"
```

These both evaluate to `/bin/sh -c 'nvim "a b c"'` so it works fine.

```
command = direct: nvim a b c
```

This evaluates to `nvim a b c` so it works fine and avoids an extra process.

```
command = direct: nvim "a b c"
```

This evaluates to `nvim '"a" b 'c"` so it's probably not what you expect. But it's an opt-in behavior and documented as such. It's doing what it says on the tin.

```
$ ghostty -e nvim "a b c"
```

Evaluates `nvim "a b c"` (argv len is 2) so it works correctly (this is the major bug fixed in this issue). 
command =

mitchellh:
Fixes #7032 

This introduces a syntax for `command` and `initial-command` that allows the user to specify whether it should be run via `/bin/sh -c` or not. The syntax is a prefix `direct:` or `shell:` prior to the command, with no prefix implying a default behavior as documented.

Previously, we unconditionally ran commands via `/bin/sh -c`, primarily to avoid having to do any shell expansion ourselves. We also leaned on it as a crutch for PATH-expansion but this is an easy problem compared to shell expansion.

For the principle of least surprise, this worked well for configurations specified via the config file, and is still the default. However, these configurations are also set via the `-e` special flag to the CLI, and it is very much not the principle of least surprise to have the command run via `/bin/sh -c` in that scenario since a shell has already expanded all the arguments and given them to us in a nice separated format. But we had no way to toggle this behavior.

This commit introduces the ability to do this, and changes the defaults so that `-e` doesn't shell expand. Further, we also do PATH lookups ourselves for the non-shell expanded case because thats easy (using execvpe style extensions but implemented as part of the Zig stdlib). We don't do path expansion (e.g. `~/`) because thats a shell expansion.

So to be clear, there are no two polar opposite behavioes here with clear semantics:

  1. Direct commands are passed to `execvpe` directly, space separated. This will not handle quoted strings, environment variables, path expansion (e.g. `~/`), command expansion (e.g. `$()`), etc.

  2. Shell commands are passed to `/bin/sh -c` and will be shell expanded as per the shell's rules. This will handle everything that `sh` supports.

In doing this work, I also stumbled upon a variety of smaller improvements that could be made:

  - A number of allocations have been removed from the startup path that only existed to add a null terminator to various strings. We now have null terminators from the beginning since we are almost always on a system that's going to need it anyways.

  - For bash shell integration, we no longer wrap the new bash command in a shell since we've formed a full parsed command line.

  - The process of creating the command to execute by termio is now unit tested, so we can test the various complex cases particularly on macOS of wrapping commands in the login command.

  - `xdg-terminal-exec` on Linux uses the `direct:` method by default since it is also assumed to be executed via a shell environment.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -22,7 +22,6 @@ const inputpkg = @import("../input.zig");
 const terminal = @import("../terminal/main.zig");
 const internal_os = @import("../os/main.zig");
 const cli = @import("../cli.zig");
-const Command = @import("../Command.zig");
 
 const conditional = @import("conditional.zig");
 const Conditional = conditional.Conditional;
@@ -34,6 +33,7 @@ const KeyValue = @import("key.zig").Value;
 const ErrorList = @import("ErrorList.zig");
 const MetricModifier = fontpkg.Metrics.Modifier;
 const help_strings = @import("help_strings");
+pub const Command = @import("command.zig").Command;
 const RepeatableStringMap = @import("RepeatableStringMap.zig");
 pub const Path = @import("path.zig").Path;
 pub const RepeatablePath = @import("path.zig").RepeatablePath;
@@ -691,8 +691,17 @@ palette: Palette = .{},
 ///   * `passwd` entry (user information)
 ///
 /// This can contain additional arguments to run the command with. If additional
-/// arguments are provided, the command will be executed using `/bin/sh -c`.
-/// Ghostty does not do any shell command parsing.
+/// arguments are provided, the command will be executed using `/bin/sh -c`
+/// to offload shell argument expansion.
+///
+/// To avoid shell expansion altogether, prefix the command with `direct:`,
+/// e.g. `direct:nvim foo`. This will avoid the roundtrip to `/bin/sh` but will
+/// also not support any shell parsing such as arguments with spaces, filepaths
+/// with `~`, globs, etc.
+///
+/// You can also explicitly prefix the command with `shell:` to always
+/// wrap the command in a shell. This can be used to ensure our heuristics
+/// to choose the right mode are not used in case they are wrong.
 ///
 /// This command will be used for all new terminal surfaces, i.e. new windows,
 /// tabs, etc. If you want to run a command only for the first terminal surface
@@ -702,7 +711,7 @@ palette: Palette = .{},
 /// arguments. For example, `ghostty -e fish --with --custom --args`.
 /// This flag sets the `initial-command` configuration, see that for more
 /// information.
-command: ?[]const u8 = null,
+command: ?Command = null,
 
 /// This is the same as "command", but only applies to the first terminal
 /// surface created when Ghostty starts. Subsequent terminal surfaces will use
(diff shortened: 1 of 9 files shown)


src/config/Config.zig
```diff
@@ -691,8 +691,17 @@ palette: Palette = .{},
 ///   * `passwd` entry (user information)
 ///
 /// This can contain additional arguments to run the command with. If additional
-/// arguments are provided, the command will be executed using `/bin/sh -c`.
-/// Ghostty does not do any shell command parsing.
+/// arguments are provided, the command will be executed using `/bin/sh -c`
+/// to offload shell argument expansion.
+///
+/// To avoid shell expansion altogether, prefix the command with `direct:`,
+/// e.g. `!nvim foo`. This will avoid the roundtrip to `/bin/sh` but will also
```
jparise, line 698: Should this example use `!` or `direct:`?

00-kat: Probably `direct:`; `!` was what was initially being discussed.

src/config/command.zig
```diff
@@ -0,0 +1,291 @@
...
+
+    pub fn formatEntry(self: Self, formatter: anytype) !void {
+        switch (self) {
+            .shell => |v| try formatter.formatEntry([]const u8, v),
+
+            .direct => |v| {
+                var buf: [4096]u8 = undefined;
+                var fbs = std.io.fixedBufferStream(&buf);
+                const writer = fbs.writer();
+                writer.writeAll("!") catch return error.OutOfMemory;
```
jparise, line 176: `direct:` here?

mitchellh: Good catch, also added unit tests for this.

PR #7052, opened 2025-04-10
ci: add a script and workflow for requesting i18n review

trag1c:
This helps get around the `CODEOWNERS` file requiring that contributors need to have write access to the repo in order to be requested for review. The mechanism is (for any new PR):
1. List the changed files
2. Fetch and parse the CODEOWNERS file to get a path:owner mapping
3. See if any changed file matches the CODEOWNERS mapping
4. Fetch team members for affected localization teams
5. Request review from individual accounts


mitchellh:
#7053 because Git.

PR #6837, opened 2025-03-20
Add sv_SE Swedish locale

designbyadrian:
Po files added for Swedish.

No amendments for the macOS version due to localisation on macOS being in progress, according to Ghostty devs.

PR #7068, opened 2025-04-12
i18n: add swedish translations

majjejjam:
I had issues translating "Split" as a noun. The only other application I found that uses splits and has translations in Swedish is GNOME Builder, and they use words that don't work as well as a verb (i.e. "Split \<Direction\>"). I added a comment about this in the `po` file.


majjejjam:
Whoops, I completely missed #6837 :(

PR #7075, opened 2025-04-13
Fix quick terminal fullscreen toggle in front of fullscreen window

### LINKED ISSUE #7070: macOS: quick terminal fullscreen only works once in front of Ghostty fullscreen window
```
macos-non-native-fullscreen = false
keybind = grave_accent=toggle_quick_terminal
```

1. Open Ghostty
2. Fullscreen the normal Ghostty window with `cmd+enter` (native fullscreen)
3. Open quick terminal with "`"
4. Press `cmd+enter` repeatedly. 

Notice the quick terminal will fullscreen once (enter and exit) but will not be able to do so again.

liby:
Fixes #7070 

The root issue occurs when the quick terminal has already entered fullscreen state and the main window is also in fullscreen mode. When attempting to toggle fullscreen again, the state detection fails due to interference from the main window's fullscreen state.

In the original code, `toggleFullscreen(mode: .nonNative)` attempts to toggle fullscreen based on the current state, but the detection is confused by the main window's fullscreen state.

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -33,6 +33,9 @@ class QuickTerminalController: BaseTerminalController {
     /// The configuration derived from the Ghostty config so we don't need to rely on references.
     private var derivedConfig: DerivedConfig
 
+    /// Storage for the window frame before entering fullscreen mode
+    private var savedWindowFrame: NSRect? = nil
+
     init(_ ghostty: Ghostty.App,
          position: QuickTerminalPosition = .top,
          baseConfig base: Ghostty.SurfaceConfiguration? = nil,
@@ -485,9 +488,16 @@ class QuickTerminalController: BaseTerminalController {
     @objc private func onToggleFullscreen(notification: SwiftUI.Notification) {
         guard let target = notification.object as? Ghostty.SurfaceView else { return }
         guard target == self.focusedSurface else { return }
+        guard let window = self.window else { return }
+        guard let screen = window.screen ?? NSScreen.main else { return }
 
-        // We ignore the requested mode and always use non-native for the quick terminal
-        toggleFullscreen(mode: .nonNative)
+        if let originalFrame = savedWindowFrame {
+            window.setFrame(originalFrame, display: true)
+            savedWindowFrame = nil
+        } else {
+            savedWindowFrame = window.frame
+            window.setFrame(screen.frame, display: true)
+        }
     }
 
     @objc private func ghosttyConfigDidChange(_ notification: Notification) {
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -485,9 +488,17 @@ class QuickTerminalController: BaseTerminalController {
     @objc private func onToggleFullscreen(notification: SwiftUI.Notification) {
         guard let target = notification.object as? Ghostty.SurfaceView else { return }
         guard target == self.focusedSurface else { return }
+        guard let window = self.window else { return }
+        guard let screen = window.screen ?? NSScreen.main else { return }
 
-        // We ignore the requested mode and always use non-native for the quick terminal
-        toggleFullscreen(mode: .nonNative)
+        if savedWindowFrame != nil {
+            let originalFrame = savedWindowFrame!
```
jparise, line 495: `if let savedWindowFrame` will let you drop the force unwrap on the next line.

jparise (approved):

mitchellh:
Thanks! I want to understand this a bit better to see the root cause.

mitchellh (changes requested):
I think the root cause here isn't correct. 

I did some investigation and the root cause seems to be mucking around with `autoHideMenuBar` while on a space with an existing fullscreen window. I suspect it's because our accounting around our acquire/release is wrong because AppKit is injecting style masks on its own for native fullscreen.

Your fix works because it avoids all the fullscreen machinery but I'd like to keep all fullscreen logic centralized in that single place so I think it'd be more robust to detect this scenario. I'm not sure how yet but I'm looking into it...

mitchellh:
From logs:

```
-[NSApplication _setPresentationOptions:instance:flags:0x40e] Both NSApplicationPresentationHideMenuBar and NSApplicationPresentationAutoHideMenuBar were specified; only one is allowed
```

One of these is from macOS and one is from us. We should be able to detect this.

liby:
I just pushed a fix for the fullscreen issue. Rather than using a custom fullscreen implementation, I focused on the root cause you identified - presentation options conflict when toggling on spaces with existing fullscreen windows.

The implementation now:
- Checks for existing presentation options before applying our own
- Only applies menu hiding when not already managed by macOS
- Skips menu bar management entirely when on a space with a fullscreen window

This should avoid the conflict while keeping all the fullscreen logic in the central implementation. Thanks for pointing me in the right direction!

mitchellh:
Thanks, but I took a different approach in #7091. The approach you ended up taking here isn't safe because we have multiple parts of the code that could nest presentation options which is why we must not inspect or touch it directly. The issue is that macOS is modifying it causing this issue...

PR #7083, opened 2025-04-14
gtk: implement global shortcuts

pluiedev:
It's been a lot of D-Bus related pain and suffering, but here it is.

I'm not sure about how well this is integrated inside App, but I'm fairly
proud of the standalone logic.

### DIFF
--- a/src/apprt/gtk/GlobalShortcuts.zig
+++ b/src/apprt/gtk/GlobalShortcuts.zig
@@ -1,0 +1,419 @@
+const GlobalShortcuts = @This();
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+const gio = @import("gio");
+const glib = @import("glib");
+const gobject = @import("gobject");
+
+const App = @import("App.zig");
+const configpkg = @import("../../config.zig");
+const Binding = @import("../../input.zig").Binding;
+const key = @import("key.zig");
+
+const log = std.log.scoped(.global_shortcuts);
+const Token = [16]u8;
+
+app: *App,
+arena: std.heap.ArenaAllocator,
+dbus: *gio.DBusConnection,
+
+/// A mapping from a unique ID to an action.
+/// Currently the unique ID is simply the serialized representation of the
+/// trigger that was used for the action as triggers are unique in the keymap,
+/// but this may change in the future.
+map: std.StringArrayHashMapUnmanaged(Binding.Action) = .{},
+
+/// The handle of the current global shortcuts portal session,
+/// as a D-Bus object path.
+handle: ?[:0]const u8 = null,
+
+/// The D-Bus signal subscription for the response signal on requests.
+/// The ID is guaranteed to be non-zero, so we can use 0 to indicate null.
+response_subscription: c_uint = 0,
+
+/// The D-Bus signal subscription for the keybind activate signal.
+/// The ID is guaranteed to be non-zero, so we can use 0 to indicate null.
+activate_subscription: c_uint = 0,
+
+pub fn init(alloc: Allocator, gio_app: *gio.Application) ?GlobalShortcuts {
+    const dbus = gio_app.getDbusConnection() orelse return null;
+
+    return .{
+        // To be initialized later
+        .app = undefined,
+        .arena = .init(alloc),
+        .dbus = dbus,
+    };
+}
+
+pub fn deinit(self: *GlobalShortcuts) void {
+    self.close();
+    self.arena.deinit();
+}
+
+fn close(self: *GlobalShortcuts) void {
+    if (self.response_subscription != 0) {
+        self.dbus.signalUnsubscribe(self.response_subscription);
+        self.response_subscription = 0;
+    }
+
+    if (self.activate_subscription != 0) {
+        self.dbus.signalUnsubscribe(self.activate_subscription);
+        self.activate_subscription = 0;
+    }
+
+    if (self.handle) |handle| {
+        // Close existing session
+        self.dbus.call(
+            "org.freedesktop.portal.Desktop",
+            handle,
+            "org.freedesktop.portal.Session",
+            "Close",
+            null,
+            null,
+            .{},
+            -1,
+            null,
+            null,
+            null,
+        );
+        self.handle = null;
+    }
+}
+
+pub fn refreshSession(self: *GlobalShortcuts, app: *App) !void {
+    // Ensure we have a valid reference to the app
+    // (it was left uninitialized in `init`)
+    self.app = app;
+
+    // Close any existing sessions
+    self.close();
+
+    // Update map
+    var trigger_buf: [256]u8 = undefined;
+
+    self.map.clearRetainingCapacity();
+    var it = self.app.config.keybind.set.bindings.iterator();
+
+    while (it.next()) |entry| {
+        const leaf = switch (entry.value_ptr.*) {
+            // Global shortcuts can't have leaders
+            .leader => continue,
+            .leaf => |leaf| leaf,
+        };
+        if (!leaf.flags.global) continue;
+
+        const trigger = try key.xdgShortcutFromTrigger(
+            &trigger_buf,
+            entry.key_ptr.*,
+        ) orelse continue;
+
+        try self.map.put(
+            self.arena.allocator(),
+            try self.arena.allocator().dupeZ(u8, trigger),
+            leaf.action,
+        );
+    }
+
+    try self.request(.create_session);
+}
+
+fn shortcutActivated(
+    _: *gio.DBusConnection,
+    _: ?[*:0]const u8,
+    _: [*:0]const u8,
+    _: [*:0]const u8,
+    _: [*:0]const u8,
+    params: *glib.Variant,
+    ud: ?*anyopaque,
+) callconv(.c) void {
+    const self: *GlobalShortcuts = @ptrCast(@alignCast(ud));
+
+    // 2nd value in the tuple is the activated shortcut ID
+    // See https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html#org-freedesktop-portal-globalshortcuts-activated
+    var shortcut_id: [*:0]const u8 = undefined;
+    params.getChild(1, "&s", &shortcut_id);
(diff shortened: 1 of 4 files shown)


src/apprt/gtk/GlobalShortcuts.zig
```diff
@@ -0,0 +1,391 @@
...
+        }
+    };
+
+    var request_token_buf: Token = undefined;
+    const request_token = generateToken(&request_token_buf);
+
+    const payload = method.makePayload(self, request_token) orelse return;
+    const request_path = try self.getRequestPath(request_token);
+
+    _ = self.dbus.signalSubscribe(
```
tristan957, line 328: Is it a leak if we don't unsubscribe at some point?

pluiedev: I'm honestly not sure, it's *recommended* to unsubscribe when you're done with it but I think not even libportal is truly strict about it

tristan957:
I wonder what the diff would look like if we used libportal

pluiedev:
Well... libportal doesn't implement this yet, and it doesn't seem like anyone else is interested. Would make my day a whole lot easier if it were the case to be honest

src/apprt/gtk/GlobalShortcuts.zig
pluiedev: Ah, looks like their idiomatic pattern is to store all signal IDs in an array and then removing them in one batch. Honestly that's pretty smart - I might steal that

tristan957: Yeah, I confirmed this is a memory leak thanks to Georges from Igalia!

https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gdbusconnection.c?ref_type=heads#L3916-3922

src/apprt/gtk/key.zig
```diff
@@ -20,10 +20,40 @@ pub fn accelFromTrigger(buf: []u8, trigger: input.Binding.Trigger) !?[:0]const u
...
+
+    // We need to make the string null terminated.
+    try writer.writeByte(0);
+    const slice = buf_stream.getWritten();
+    return slice[0 .. slice.len - 1 :0];
+}
+
+/// Returns a XDG-compliant shortcuts string from a trigger.
+/// Spec: https://specifications.freedesktop.org/shortcuts-spec/latest/
+pub fn xdgShortcutFromTrigger(buf: []u8, trigger: input.Binding.Trigger) !?[:0]const u8 {
```
tristan957, line 33: should we write tests for this function?

pluiedev: Oh yeah that would be a good idea. Will do it tomorrow as it's quite late here

src/apprt/gtk/GlobalShortcuts.zig
```diff
@@ -0,0 +1,412 @@
...
+
+app: *App,
+arena: std.heap.ArenaAllocator,
+dbus: *gio.DBusConnection,
+
+/// A mapping from a unique ID to an action.
+/// Currently the unique ID is simply the serialized representation of the
+/// trigger that was used for the action as triggers are unique in the keymap,
+/// but this may change in the future.
+map: std.StringHashMapUnmanaged(Binding.Action) = .{},
```
mitchellh, line 26: Should this be an ArrayHashMap to avoid the tombstone issue for very long running Ghostty instances? https://github.com/ziglang/zig/issues/17851

src/apprt/gtk/GlobalShortcuts.zig
```diff
@@ -0,0 +1,412 @@
...
+    // Ensure we have a valid reference to the app
+    // (it was left uninitialized in `init`)
+    self.app = app;
+    self.deinit();
+
+    // Update map
+    var trigger_buf: [256]u8 = undefined;
+
+    self.map.clearRetainingCapacity();
+    var it = self.app.config.keybind.set.bindings.iterator();
```
mitchellh, line 92: Note to self: create better API for this, I don't think the apprt's should be doing _this level_ of introspection. Its totally fine for this PR, but I'll get around to a better API soon, because its in the hot path of me implementing global shortcuts on macOS that don't require special permissions (that they do today).

mitchellh (changes requested):
Excellent work. That fact it's so encapsulated is really well done. I agree with @tristan957 on the test request for the one function, then I just hit the "red" button to consider the `ArrayHashMap` change.

src/apprt/gtk/GlobalShortcuts.zig
mitchellh: Yeah. In this case it'd be pretty pathological because you only clear and re-add when settings are reloaded so a user would have to reload _a lot_ but I've been bit enough by this bug that I instinctively think "why not"

mitchellh (approved):

jcollie (approved):
LGTM

jcollie (approved):
LGTM

pluiedev:
Gonna delay merging this until GNOME 48 support hits NixOS unstable/25.05 ([PR](https://github.com/NixOS/nixpkgs/pull/386514), [tracker](https://nixpk.gs/pr-tracker.html?pr=386514)) since I wanna see how well this works on GNOME first

jcollie:
Trying out now with Gnome 48 and I'm getting the following crash:
```
debug(gtk): windowing protocol=wayland
debug(gtk): cgroup isolation disabled config=config.Config.LinuxCgroup.single-instance
Segmentation fault at address 0x7f3994500008
/nix/store/hmmxm8rfbmdvcd45xi7h42zsyi4crbrb-zig-0.14.0/lib/zig/std/heap/arena_allocator.zig:185:77: 0x314ee6d in alloc (ghostty)
            const cur_alloc_buf = @as([*]u8, @ptrCast(cur_node))[0..cur_node.data];
                                                                            ^
/nix/store/hmmxm8rfbmdvcd45xi7h42zsyi4crbrb-zig-0.14.0/lib/zig/std/mem/Allocator.zig:129:26: 0x315a1da in allocBytesWithAlignment__anon_15860 (ghostty)
    return a.vtable.alloc(a.ptr, len, alignment, ret_addr);
                         ^
/nix/store/hmmxm8rfbmdvcd45xi7h42zsyi4crbrb-zig-0.14.0/lib/zig/std/mem/Allocator.zig:260:40: 0x314a412 in allocWithSizeAndAlignment__anon_13957 (ghostty)
    return self.allocBytesWithAlignment(alignment, byte_count, return_address);
                                       ^
/nix/store/hmmxm8rfbmdvcd45xi7h42zsyi4crbrb-zig-0.14.0/lib/zig/std/mem/Allocator.zig:254:75: 0x3130b58 in alloc__anon_11105 (ghostty)
    const ptr: [*]align(a) T = @ptrCast(try self.allocWithSizeAndAlignment(@sizeOf(T), a, n, return_address));
                                                                          ^
/nix/store/hmmxm8rfbmdvcd45xi7h42zsyi4crbrb-zig-0.14.0/lib/zig/std/mem.zig:3222:36: 0x33ad9fb in joinMaybeZ (ghostty)
    const buf = try allocator.alloc(u8, total_len);
                                   ^
/nix/store/hmmxm8rfbmdvcd45xi7h42zsyi4crbrb-zig-0.14.0/lib/zig/std/mem.zig:3208:31: 0x32fda62 in joinZ (ghostty)
    const out = try joinMaybeZ(allocator, separator, slices, true);
                              ^
/home/jeff/dev/ghostty/src/apprt/gtk/GlobalShortcuts.zig:400:42: 0x32fd8b1 in getRequestPath (ghostty)
    const object_path = try std.mem.joinZ(self.arena.allocator(), "/", &.{
                                         ^
/home/jeff/dev/ghostty/src/apprt/gtk/GlobalShortcuts.zig:348:49: 0x32fdd6a in request__anon_158218 (ghostty)
    const request_path = try self.getRequestPath(request_token);
                                                ^
/home/jeff/dev/ghostty/src/apprt/gtk/GlobalShortcuts.zig:114:21: 0x32fea12 in refreshSession (ghostty)
    try self.request(.create_session);
                    ^
/home/jeff/dev/ghostty/src/apprt/gtk/App.zig:1022:33: 0x32ffea7 in syncConfigChanges (ghostty)
        shortcuts.refreshSession(self) catch |err| {
                                ^
/home/jeff/dev/ghostty/src/apprt/gtk/App.zig:1317:27: 0x3322b9c in run (ghostty)
    self.syncConfigChanges(null) catch |err| {
                          ^
/home/jeff/dev/ghostty/src/main_ghostty.zig:114:24: 0x3323a88 in main (ghostty)
    try app_runtime.run();
                       ^
/nix/store/hmmxm8rfbmdvcd45xi7h42zsyi4crbrb-zig-0.14.0/lib/zig/std/start.zig:656:37: 0x33245b7 in main (ghostty)
            const result = root.main() catch |err| {
                                    ^
???:?:?: 0x7f39a7e2a47d in ??? (libc.so.6)
Unwind information for `libc.so.6:0x7f39a7e2a47d` was not available, trace may be incomplete
```

pluiedev:
Hmmm.. I'm running into the same error. Lemme see here

jcollie:
This seems to fix it:
```
diff --git a/src/apprt/gtk/GlobalShortcuts.zig b/src/apprt/gtk/GlobalShortcuts.zig
index 09d487686..c4ea67df5 100644
--- a/src/apprt/gtk/GlobalShortcuts.zig
+++ b/src/apprt/gtk/GlobalShortcuts.zig
@@ -83,7 +83,6 @@ pub fn refreshSession(self: *GlobalShortcuts, app: *App) !void {
     // Ensure we have a valid reference to the app
     // (it was left uninitialized in `init`)
     self.app = app;
-    self.deinit();

     // Update map
     var trigger_buf: [256]u8 = undefined;
```

pluiedev:
Works in my VM testing on GNOME 48 and Plasma 6.3, NixOS unstable (25.11).

@jcollie Can you see if it works on your end?

jcollie:
> Works in my VM testing on both GNOME 48 and Plasma 6.3, NixOS unstable (25.11).
> 
> @jcollie Can you see if it works on your end?

YESSS!!!! working for me now!

pluiedev:
Let's do this

PR #7085, opened 2025-04-14
i18: fix minor Norwegian grammar issues

cryptocode:
Changes the translation of 'clipboard' to definite singular form, and removes a misplaced verb.


Uzaaft (approved):
LGTM :shipit:

PR #7087, opened 2025-04-14
gtk: implement bell

pluiedev:
This PR implements a more lightweight alternative to #5326 that contains features that I personally think Just Make Sense for the bell.

No configs, no GStreamer stuff, just sane defaults to get us started.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1848,6 +1848,22 @@ keybind: Keybinds = .{},
 /// open terminals.
 @"custom-shader-animation": CustomShaderAnimation = .true,
 
+/// The list of enabled features that are activated after encountering
+/// a bell character.
+///
+/// Valid values are:
+///
+///  * `system` (default)
+///
+///    Instructs the system to notify the user using built-in system functions.
+///    This could result in an audiovisual effect, a notification, or something
+///    else entirely. Changing these effects require altering system settings:
+///    for instance under the "Sound > Alert Sound" setting in GNOME,
+///    or the "Accessibility > System Bell" settings in KDE Plasma.
+///
+/// Currently only implemented on Linux.
+@"bell-features": BellFeatures = .{},
+
 /// Control the in-app notifications that Ghostty shows.
 ///
 /// On Linux (GTK), in-app notifications show up as toasts. Toasts appear
@@ -5667,6 +5683,11 @@ pub const GtkToolbarStyle = enum {
 /// See app-notifications
 pub const AppNotifications = packed struct {
     @"clipboard-copy": bool = true,
+};
+
+/// See bell-features
+pub const BellFeatures = packed struct {
+    system: bool = false,
 };
 
 /// See mouse-shift-capture
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -601,6 +601,7 @@ typedef enum {
   GHOSTTY_ACTION_RELOAD_CONFIG,
   GHOSTTY_ACTION_CONFIG_CHANGE,
   GHOSTTY_ACTION_CLOSE_WINDOW,
+  GHOSTTY_ACTION_RING_BELL,
 } ghostty_action_tag_e;
 
 typedef union {
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -932,6 +932,16 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
         .present_surface => try self.presentSurface(),
 
         .password_input => |v| try self.passwordInput(v),
+
+        .ring_bell => {
+            _ = self.rt_app.performAction(
+                .{ .surface = self },
+                .ring_bell,
+                {},
+            ) catch |err| {
+                log.warn("apprt failed to ring bell={}", .{err});
+            };
+        },
     }
 }
 
--- a/src/apprt/action.zig
+++ b/src/apprt/action.zig
@@ -244,6 +244,8 @@ pub const Action = union(Key) {
     /// Closes the currently focused window.
     close_window,
 
+    ring_bell,
+
     /// Sync with: ghostty_action_tag_e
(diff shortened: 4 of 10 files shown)


pluiedev:
The system bell isn't exactly working on NixOS + KDE at the moment, since GTK 4.16 doesn't yet support the *new* system bell protocol which KWin implements. You'd have to ask a GNOME user what it exactly does ;)

As for the tab indicator thing:
image

tristan957:
The UI change looks good.

tristan957:
Should you add `Co-authored-by: jcollie...`?

pluiedev:
As I've already said in https://github.com/ghostty-org/ghostty/discussions/2710#discussioncomment-12017099, I personally feel like offering too many options falls in the same trap as other mega-customizable settings, in that the implementation costs don't outweigh the benefits of having that many degrees of freedom. Besides, this is intentionally designed to allow for more bell settings in the future, as you could just add if statements before the labelled blocks. I just think that this is a sensible default that works for most people.

About GStreamer... the fact of the matter is that we have to add more bindings, add deps to Nix (and Flatpaks, Snaps, etc.), embed our own audio files, and so on... I don't think that's worth it, not for me personally. Then again, this could be part of a future addition. I'd just like to see *some* kind of bell feature implemented inside Ghostty at the moment, which is evidently important to quite a few people.

tristan957:
> I think that the lack of configuration, even just the ability to enable/disable the bell, is not a good choice.

I agree with this. Adding a `bell` and `bell-features` config options would be good, which I think @jcollie had in his original PR.

Adding the custom sound could come in a separate PR based on jcollie's work.

tristan957:
> I personally feel like offering too many options falls in the same trap as other mega-customizable settings, in that the implementation costs don't outweigh the benefits of having that many degrees of freedom

This ship has sailed.

pluiedev:
Fine - I think enabling/disabling the system bell function makes sense, but I'm very much against implementing GStreamer stuff in this PR. It simply rose too many questions in the last PR and impeded the ball from actually rolling

jcollie:
My latest patch does not embed any audio files. I agree that my initial patches that did that were a mistake. If someone wants to use a custom audio file they would need to supply it themselves. 

Setting attention on the tab is unobtrusive enough to leave always on I think.

In Gnome, the system beep is semi-configurable from a list of options. You can see what they sound like by clicking on the options.

image

tristan957:
> Fine - I think enabling/disabling the system bell function makes sense (the tab thing is such a small visual fluorish I'm not really sure if we should make it configurable either)

Agree completely here.

> I'm very much against implementing GStreamer stuff in this PR. It simply rose too many questions in the last PR and impeded the ball from actually rolling

Yes, let's land the basics, and then @jcollie can rebase on it. I am +1 on custom sounds :smile:

pluiedev:
Added `bell-features` and also added @jcollie to a bunch of these commits as a co-author 'cos he's been part of this since even the first prototype, and this feels more right :p

mitchellh (changes requested):
One very simple change request.

The rest looks very good. Thanks for paring down the PR, this is very easy and much more palatable to merge. 😄 The config types/options may need to be tuned some more before a release but it's certainly good enough for a tip build for now.

Aside:

> I personally feel like offering too many options falls in the same trap as other mega-customizable settings, in that the implementation costs don't outweigh the benefits of having that many degrees of freedom.

I understand this opinion, but that isn't this project. Ghostty's config philosophy is "zero-config defaults" (e.g. 90+% of people don't need to touch a single thing) with extreme flexibility otherwise, I've said this publicly before that I want Ghostty to be one of the most customizable terminals out there _for people who want it_. So, we'll have a lot of config options over time (we already do).

src/config/Config.zig
```diff
@@ -5669,6 +5685,11 @@ pub const AppNotifications = packed struct {
     @"clipboard-copy": bool = true,
 };
 
+/// See bell-features
+pub const BellFeatures = packed struct {
+    system: bool = true,
```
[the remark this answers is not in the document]
pluiedev: It can do more than just audio beeps - on KDE at least supposedly it also does visual flashes. You're right that `system` is *kind of* a bad name though.

mitchellh (approved):

tristan957 (approved):
Thanks to you and @jcollie for getting the first iteration of this done!

tristan957:
Maybe a force push would kick github runners to try again?

jcollie (approved):

mitchellh:
I'm willing to risk it. LOL. CI on main.

src/config/Config.zig
```diff
@@ -1848,6 +1848,22 @@ keybind: Keybinds = .{},
 /// open terminals.
 @"custom-shader-animation": CustomShaderAnimation = .true,
 
+/// The list of enabled features that are activated after encountering
+/// a bell character.
+///
+/// Valid values are:
+///
+///  * `system` (default)
```
ncfavier, line 1856: It says "default" here but the value defaults to `false`.

ncfavier:
This doesn't seem to set the X11 urgency hint, which is the only thing I care about the bell doing. Would it be possible to support that?

jcollie:
> This doesn't seem to set the X11 urgency hint, which is the only thing I care about the bell doing. Would it be possible to support that?

It's possible, but please open a new feature discussion to discuss it.

00-kat:
In my opinion, create a new Discussion instead of bumping the old one.

ncfavier:
Opened https://github.com/ghostty-org/ghostty/discussions/7124

src/config/Config.zig
00-kat: Oh you had commented this before.  I randomly noticed this a few minutes ago too and created a pull requested → #7148.

PR #7091, opened 2025-04-14
macOS: fix quick terminal fullscreen issues

### LINKED ISSUE #7070: macOS: quick terminal fullscreen only works once in front of Ghostty fullscreen window

mitchellh:
Supersedes #7075 
Fixes #7070 

This fixes a few separate fullscreen issues with the quick terminal:

1. If we're on a fullscreen space, we can't reliably set the `autoHideMenuBar` presentation option because macOS itself is managing it. The fix is to use private APIs to detect we're on a fullscreen space and avoid this.

2. If our quick terminal is fullscreen when we move spaces, we must exit and re-enter fullscreen because the frame may change (e.g. due to menubar changes).

3. If we aren't the frontmost app, we must avoid hiding the menu because it has no effect and our fullscreen frame would be wrong.


mitchellh:
Btw, I know 100% that there are still bugs lurking in here. There's no way this is perfect, but it does improve the situation.

peanball:
thank you! this resolves https://github.com/ghostty-org/ghostty/discussions/6612 for me.

PR #7093, opened 2025-04-14
terminal: clear correct row on index operation in certain edge cases

### LINKED ISSUE #7066: vim status line leaks into scrollback
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4802

<sup>Originally posted by **dragonJACson** January  7, 2025</sup>
# Status line leaked into scrollback buffer

## Issue description
The vim status line ("-- INSERT --") incorrectly appears within the terminal scrollback buffer instead of staying fixed at the bottom of the terminal window. (The line starts with 21888)

图片

## Environment

Issue reproducible on both environments, Ghostty itself runs on macOS 15.2
- Remote: Debian 10 with Vim (VIM - Vi IMproved 8.1 (2018 May 18, compiled Sep 27 2023 19:47:00)
- Local: macOS 15.2 with Vim (VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Nov 9 2024 23:15:55) 
- Vim configuration: Default settings, no customization
- Terminfo: `xterm-256color`
- Steps to reproduce
  - `seq 1 100000 > numbers.txt`
  - `vim numbers.txt`
  - Enter insert mode and keep scrolling
- Ghostty version: 
```
ghostty --version
Ghostty 1.0.2-main+a2445359

Version
  - version: 1.0.2-main+a2445359
  - channel: tip
Build Config
  - Zig version: 0.13.0
  - renderer   : renderer.Metal
```
- Ghostty config: 
```
font-family = Sarasa Term Slab SC
font-size = 16
theme = OneHalfDark

window-save-state = always
scrollback-limit = 10000000
```

Asciinema recording: 
[demo2.cast.zip](https://github.com/user-attachments/files/18342302/demo2.cast.zip)
</div>

### Reproduction

The asciinema (reattached on this issue below) reproduces this:

```
$ zig build -Dapp-runtime=glfw run -- --window-width=186 --window-height=43 --font-size=8
$ asciinema play demo2.cast
```

[demo2.cast.zip](https://github.com/user-attachments/files/19719975/demo2.cast.zip)

mitchellh:
Fixes #7066

This fixes an issue where under certain conditions (expanded below), we would not clear the correct row, leading to the screen having duplicate data.

This was triggered by a page state of the following:

```
      +----------+ = PAGE 0
  ... :          :
 4305 |1ABCD00000|
 4306 |2EFGH00000|
      :^         : = PIN 0
     +-------------+ ACTIVE
 4307 |3IJKL00000| | 0
      +----------+ :
      +----------+ : = PAGE 1
    0 |          | | 1
    1 |          | | 2
      +----------+ :
     +-------------+
```

Namely, the cursor had to NOT be on the last row of the first page, but somewhere on the first page. Then, when an `index` (LF) operation was performed the result would look like this:

```
      +----------+ = PAGE 0
  ... :          :
 4305 |1ABCD00000|
 4306 |2EFGH00000|
     +-------------+ ACTIVE
 4307 |3IJKL00000| | 0
      :^         : : = PIN 0
      +----------+ :
      +----------+ : = PAGE 1
    0 |3IJKL00000| | 1
    1 |          | | 2
      +----------+ :
     +-------------+
```

The `3IJKL` line was duplicated. What was happening here is that we performed the index operation correctly but failed to clear the cursor line as expected.

This is because we were always clearing the first row in the page instead of the row of the cursor.

Test added.

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -924,8 +924,8 @@ fn cursorScrollAboveRotate(self: *Screen) !void {
     fastmem.rotateOnceR(Row, cur_rows[self.cursor.page_pin.y..cur_page.size.rows]);
     self.clearCells(
         cur_page,
-        &cur_rows[0],
-        cur_page.getCells(&cur_rows[0]),
+        &cur_rows[self.cursor.page_pin.y],
+        cur_page.getCells(&cur_rows[self.cursor.page_pin.y]),
     );
 
     // Set all the rows we rotated and cleared dirty
@@ -1254,6 +1254,17 @@ pub fn clearCells(
         page.pauseIntegrityChecks(false);
         page.assertIntegrity();
         self.assertIntegrity();
+    }
+
+    if (comptime std.debug.runtime_safety) {
+        // Our row and cells should be within the page.
+        const page_rows = page.rows.ptr(page.memory.ptr);
+        assert(@intFromPtr(row) >= @intFromPtr(&page_rows[0]));
+        assert(@intFromPtr(row) <= @intFromPtr(&page_rows[page.size.rows - 1]));
+
+        const row_cells = page.getCells(row);
+        assert(@intFromPtr(&cells[0]) >= @intFromPtr(&row_cells[0]));
+        assert(@intFromPtr(&cells[cells.len - 1]) <= @intFromPtr(&row_cells[row_cells.len - 1]));
     }
 
     // If this row has graphemes, then we need go through a slow path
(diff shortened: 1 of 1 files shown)


src/terminal/Screen.zig
```diff
@@ -925,7 +925,7 @@ fn cursorScrollAboveRotate(self: *Screen) !void {
     self.clearCells(
         cur_page,
         &cur_rows[0],
```
qwerasd205, line 927: This needs to be changed too, no?

mitchellh: Good catch. I also added some assertions to validate this more.

PR #7100, opened 2025-04-15
font: use regular JB Mono with dedicated Symbol font

mitchellh:
Fixes #7069

I'm not sure if this will have follow-on negative effects. In my day to day usage everything looks fine but its hard to really know. I think our best chance is to merge this and see what feedback we get. I know for sure positively that it does two things:

  1. Fixes the small symbols noted in #7069.
  2. Makes our binary size smaller since we no longer have unnecessary symbols embedded in our stylized fonts.


qwerasd205:
This has negative effects on symbols that need to be stretched to certain sizes for alignment purposes. I'll check later to remember which ones specifically, but it's for that reason that I held off on making this change until I implemented the nerd font patcher sizing rules in to our code

mitchellh:
> I'll check later to remember which ones specifically, but it's for that reason that I held off on making this change until I implemented the nerd font patcher sizing rules in to our code

Okay, that's good enough reason for me to drop this for now. I'll add this context to the original issue and we can look back at it with a better approach.

PR #7101, opened 2025-04-15
macOS: Implement basic bell features (no sound)

### LINKED ISSUE #7099: macOS: implement bell support
#7087 implemented support for the BELL character on GTK (and added infrastructure to the core). #5326 will add support for custom audio for GTK. Bell support was originally discussed in #2710 .

mitchellh:
Fixes #7099

This adds basic bell features to macOS to conceptually match the GTK implementation. When a bell is triggered, macOS will do the following:

  1. Bounce the dock icon once, if the app isn't already in focus.
  2. Add a bell emoji (🔔) to the title of the surface that triggered the bell. This emoji will be removed after the surface is focused or a keyboard event if the surface is already focused. This behavior matches iTerm2.

Note that neither of these respect the `system` `bell-features` config because they're both unobtrusive (the dock icon bounces only once, the title change is silent and similar to GTK tab attention) and unrelated to system settings.

This doesn't add an icon badge because macOS's dockTitle.badgeLabel API wasn't doing anything for me and I wasn't able to fully figure out why...

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -186,6 +186,12 @@ class AppDelegate: NSObject,
             name: .ghosttyConfigDidChange,
             object: nil
         )
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(ghosttyBellDidRing(_:)),
+            name: .ghosttyBellDidRing,
+            object: nil
+        )
 
         // Configure user notifications
         let actions = [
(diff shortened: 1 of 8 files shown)


mitchellh:
## Demo 

https://github.com/user-attachments/assets/cfdb5daa-3346-46c4-9ded-db731c17ef88

mitchellh:
cc @jcollie @pluiedev

jcollie (approved):
Can't really comment on the Swift code but LGTM

ftpd:
@mitchellh thanks for implementing this, but let me still emphasize on a dock badge.

Imagine the use case: I'm running a `./really-long-command` and I want to be notified when it finishes. So, I'm running `./really-long-commmand;  tput bel` and it should be fine - sure, but what happens if it finished while I was AFK?

The icon badge is an ultimate solution - when I'm back I know that my terminal emulator needs an attention. Bouncing icon is fine when assumed I'm always looking it it.

Can it be reopened or should I create a new discussion/issue about this?

jcollie:
> Can it be reopened or should I create a new discussion/issue about this?

Please start a new discussion focused on the dock icon badge. At the minimum someone needs to investigate the proper API that needs to be used.

PR #7118, opened 2025-04-17
macOS: Add dock badge notification for bell events

### LINKED ISSUE #7108: macOS: badge on dock icon for bell
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7104

<sup>Originally posted by **ftpd** April 15, 2025</sup>
In https://github.com/ghostty-org/ghostty/pull/7101 the new way of notification was introduced - on bell the dock icon will bounce one time. I would like to ask for adding a badge to the dock icon as well.

Use case: I'm running a `./really-long-command` and I want to be notified when it finishes. So, I'm running `./really-long-commmand; tput bel` and it should be fine - sure, but what happens if it finished while I was AFK?

The icon badge is an ultimate solution - when I'm back I know that my terminal emulator needs an attention. Bouncing icon is fine when assumed I'm always looking on the screen.</div>

## Proposal

We should do this. To keep things simple, I think the initial implementation can add a badge that is removed when Ghostty becomes active.

A more complex implementation in the future would be to keep track of how many active surfaces have the bell active and use that number in the badge, decrementing it as necessary when the bells are deactivated.

liby:
Resolves #7108

This PR adds visual notification badges to the Ghostty dock icon when bell events are triggered while the application is in the background. This complements the existing dock bounce notification, making it easier for users to notice when a terminal needs attention.

https://github.com/user-attachments/assets/b54c881f-fea8-4085-8614-432d9e5847b9

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -199,6 +199,7 @@ class AppDelegate: NSObject,
         ]
 
         let center = UNUserNotificationCenter.current()
+
         center.setNotificationCategories([
             UNNotificationCategory(
                 identifier: Ghostty.userNotificationCategory,
@@ -230,6 +231,9 @@ class AppDelegate: NSObject,
     func applicationDidBecomeActive(_ notification: Notification) {
         // If we're back manually then clear the hidden state because macOS handles it.
         self.hiddenState = nil
+
+        // Clear the dock badge when the app becomes active
+        self.setDockBadge(nil)
 
         // First launch stuff
         if (!applicationHasBecomeActive) {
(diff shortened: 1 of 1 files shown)


macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -509,8 +520,15 @@ class AppDelegate: NSObject,
     }
 
     @objc private func ghosttyBellDidRing(_ notification: Notification) {
-        // Bounce the dock icon if we're not focused.
-        NSApp.requestUserAttention(.informationalRequest)
+        // Only request attention and set badge when the app is not active
+        if !NSApp.isActive {
+            // Bounce the dock icon if we're not focused.
+            NSApp.requestUserAttention(.informationalRequest)
```
mitchellh, line 526: We don't need to guard this one because it automatically does nothing if the app is active. We should just let macOS do it.

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -572,6 +572,14 @@ extension Ghostty {
             bell = true
         }
 
+        @objc private func ghosttyBellDidReset() {
+            // Reset bell state
+            bell = false
+            // Clear dock badge
+            NSApp.dockTile.badgeLabel = nil
+            NSApp.dockTile.display()
```
mitchellh, line 580: I don't think we should be modifying App state directly from the NSView here. It might be easier to just unconditionally reset `NSApp.dockTile` when the app becomes active in AppDelegate for now.

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -199,6 +199,13 @@ class AppDelegate: NSObject,
         ]
 
         let center = UNUserNotificationCenter.current()
+        // Request notification authorization, including badge permission
+        center.requestAuthorization(options: [.alert, .sound, .badge]) { (granted, error) in
```
mitchellh, line 203: Does this trigger a consent notification? If so, we should delay this until the first bell, I think.

mitchellh (changes requested):

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -509,8 +520,15 @@ class AppDelegate: NSObject,
...
     @objc private func ghosttyBellDidRing(_ notification: Notification) {
-        // Bounce the dock icon if we're not focused.
-        NSApp.requestUserAttention(.informationalRequest)
+        // Only request attention and set badge when the app is not active
+        if !NSApp.isActive {
+            // Bounce the dock icon if we're not focused.
+            NSApp.requestUserAttention(.informationalRequest)
+
+            // Set badge label to match Slack's notification style (medium-sized white dot)
+            NSApp.dockTile.badgeLabel = "•"
```
AbeEstrada, line 529: Could a Zero-width space work?

liby: Just feel that it looks better this way, similar to the style of Slack.
image

mitchellh (approved):
Well done. Thank you!

PR #7119, opened 2025-04-17
vim: fix syntax highlight on scratch buffer

phanen:
Sometimes we need detect ghostty filetype on scratch buffer.
In this case `set iskeyword` in ftplugin is not loaded.

Screenshot: https://github.com/ibhagwan/fzf-lua/pull/1913#issuecomment-2813289819.

e.g.
```sh
nvim +'0r ~/.config/ghostty/config' +'se syntax=ghostty'
```

### DIFF
--- a/src/config/vim.zig
+++ b/src/config/vim.zig
@@ -88,6 +88,7 @@ fn writeSyntax(writer: anytype) !void {
         \\let s:cpo_save = &cpo
         \\set cpo&vim
         \\
+        \\syn iskeyword @,48-57,-
         \\syn keyword ghosttyConfigKeyword
     );
 


src/config/vim.zig
```diff
@@ -88,6 +88,7 @@ fn writeSyntax(writer: anytype) !void {
         \\let s:cpo_save = &cpo
         \\set cpo&vim
         \\
+        \\syn iskeyword @,-
```
gpanders, line 91: Some config keywords have digits, so this should include digits as well.

```suggestion
        \\syn iskeyword @,48-57,-
```

gpanders (approved):

mitchellh (approved):
Thank you

PR #7126, opened 2025-04-18
apprt/gtk: refactor action callbacks to reduce code duplication

tristan957:
It was getting very monotonous reading the code.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -811,14 +811,17 @@ fn gtkWindowUpdateScaleFactor(
     };
 }
 
-// Note: we MUST NOT use the GtkButton parameter because gtkActionNewTab
-// sends an undefined value.
-fn gtkTabNewClick(_: *gtk.Button, self: *Window) callconv(.c) void {
+/// Perform a binding action on the window's action surface.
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/Window.zig
```diff
@@ -1007,128 +1010,87 @@ fn gtkActionNewWindow(
     _: ?*glib.Variant,
     self: *Window,
 ) callconv(.C) void {
```
jcollie, line 1012: Since we're messing around in this area Zig 0.14 is deprecating `callconv(.C)` in favor of `callconv(.c)`.

jcollie (approved):
One nit, but maybe it'd just be better to mechanically go through the code and switch `.C` -> `.c`.

tristan957:
> One nit, but maybe it'd just be better to mechanically go through the code and switch `.C` -> `.c`.

I think we should do this for the whole codebase at once if that's ok with you.

edrozenberg:
At `src/apprt/gtk/Window.zig` line 1061, should be `up` not `right` (probably a copy/paste error).

tristan957:
Fix is https://github.com/ghostty-org/ghostty/pull/7152

PR #7127, opened 2025-04-18
apprt/gtk: add menu to new tab button to create splits

tristan957:
image

Closes: https://github.com/ghostty-org/ghostty/discussions/6828

PR #7134, opened 2025-04-18
ci: deduplicate reviewers and reduce requests

trag1c:
The current way seems to lead to a mess when a lot of `.po` files are changed (see #7127). The code uses one request per person because GitHub just disregarded other users in testing, but it seems to work correctly now… Also I made a mistake of not deduplicating members who are in multiple teams, so that's fixed as well.

### DIFF
--- a/po/zh_CN.UTF-8.po
+++ b/po/zh_CN.UTF-8.po
@@ -28,7 +28,7 @@ msgstr "留空以重置至默认标题。"
 #: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:9
 #: src/apprt/gtk/ui/1.5/ccw-paste.blp:10 src/apprt/gtk/CloseDialog.zig:44
 msgid "Cancel"
-msgstr "取消"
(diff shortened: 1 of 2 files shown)


po/zh_CN.UTF-8.po
```diff
@@ -28,7 +28,7 @@ msgstr "留空以重置至默认标题。"
 #: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:9
 #: src/apprt/gtk/ui/1.5/ccw-paste.blp:10 src/apprt/gtk/CloseDialog.zig:44
 msgid "Cancel"
-msgstr "取消"
+msgstr "foo"
```
Aaron-212, line 31: Why change this line?

00-kat: Testing in production :trollface: (note the commit messages).

Aaron-212: 🤣🤣🤣

PR #7139, opened 2025-04-19
macOS: use KeyboardShortcut rather than homegrown KeyEquivalent

mitchellh:
This replaces the use of our custom `Ghostty.KeyEquivalent` with the SwiftUI `KeyboardShortcut` type. This is a more standard way to represent keyboard shortcuts and lets us more tightly integrate with SwiftUI/AppKit when necessary over our custom type.

This PR should have no user impact. This is just some cleanup for future work.

Note that not all Ghostty triggers can be represented as KeyboardShortcut values because macOS itself does not support binding keys such as function keys (e.g. F1-F12) to KeyboardShortcuts.

This isn't an issue since all input also passes through a lower level libghostty API which can handle all key events, we just can't show these keyboard shortcuts on things like the menu bar. This was already true before this commit.

### DIFF
--- a/macos/Sources/Helpers/KeyboardShortcut+Extension.swift
+++ b/macos/Sources/Helpers/KeyboardShortcut+Extension.swift
@@ -1,0 +1,45 @@
+import SwiftUI
+
+extension KeyboardShortcut: @retroactive CustomStringConvertible {
+    public var description: String {
+        var result = ""
+
+        if modifiers.contains(.command) {
+            result.append("⌘")
+        }
+        if modifiers.contains(.control) {
+            result.append("⌃")
+        }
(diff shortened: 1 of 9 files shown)


macos/Sources/Helpers/KeyboardShortcut+Extension.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+        if modifiers.contains(.option) {
+            result.append("⌥")
+        }
+        if modifiers.contains(.shift) {
+            result.append("⇧")
+        }
+
+        let keyString: String
+        switch key {
```
jparise, line 21: You could also use the `let keyString = switch key { ... }` syntax here.

Issue #2446, opened 2024-10-16
gtk: flickering on resize


plichard:
https://github.com/user-attachments/assets/c1dbeeec-6688-4976-8240-81140be315e0

**ghostty version**: 0.1.0-main+9d341a9f

Built from source with `zig build -p $HOME/.local -Doptimize=ReleaseFast`.

```
# Needed to install these packages:
$ dnf info gtk4-devel libadwaita-devel libXtst-devel
Last metadata expiration check: 0:58:20 ago on mer 16 oct 2024 12:01:46.
Installed Packages
Name         : gtk4-devel
Version      : 4.14.5
Release      : 1.fc40
Architecture : x86_64
Size         : 23 M
Source       : gtk4-4.14.5-1.fc40.src.rpm
Repository   : @System
From repo    : updates
Summary      : Development files for GTK
URL          : https://www.gtk.org
License      : LGPL-2.0-or-later
Description  : This package contains the libraries and header files that are needed
             : for writing applications with version 4 of the GTK widget toolkit.

Name         : libXtst-devel
Version      : 1.2.5
Release      : 1.fc40
Architecture : x86_64
Size         : 12 k
Source       : libXtst-1.2.5-1.fc40.src.rpm
Repository   : @System
From repo    : updates
Summary      : X.Org X11 libXtst development package
URL          : http://www.x.org
License      : MIT-open-group AND HPND-sell-variant AND X11 AND HPND-doc AND HPND-doc-sell
Description  : X.Org X11 libXtst development package

Name         : libadwaita-devel
Version      : 1.5.4
Release      : 1.fc40
Architecture : x86_64
Size         : 2.1 M
Source       : libadwaita-1.5.4-1.fc40.src.rpm
Repository   : @System
From repo    : updates
Summary      : Development files for libadwaita
URL          : https://gitlab.gnome.org/GNOME/libadwaita
License      : LGPL-2.1-or-later AND MIT
Description  : Development files for libadwaita.
```

Operating System: Fedora Linux 40
KDE Plasma Version: 6.2.0
KDE Frameworks Version: 6.7.0
Qt Version: 6.7.2
Kernel Version: 6.10.12-200.fc40.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 24 × AMD Ryzen 9 5900X 12-Core Processor
Memory: 62,7 GiB of RAM
Graphics Processor: AMD Radeon RX 6600
Manufacturer: Micro-Star International Co., Ltd.
[...]

PR #7144, opened 2025-04-20
Linux: stop flickering and stretching on resize

adlr:
A few commits to address the concerns of flickering and content stretching on https://github.com/ghostty-org/ghostty/issues/2446

First, tell OpenGL the true size it has at the time drawFrame is called, so that it's not using an older size and thus stretching content.

Second, revert a patch that freed up memory from unused cell capacity in the OpenGL renderer, and replace it with a new patch that frees up the memory. The problem with the earlier patch was it was freeing the cells fully, then OpenGL would display content (which was blank!) and then rendering would complete and cells would be set again. This was leading to occasional blank frames during render. The new change keeps memory down, as was the goal with the earlier patch, but keeps the old cells around a bit longer in case they are needed to paint.

This is my first change on this codebase and in the Zig language; any advice is welcome!

NOTE: I am hoping CI will test that this doesn't break macOS. I don't have a mac handy to test.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1635,6 +1635,9 @@ fn resize(self: *Surface, size: rendererpkg.ScreenSize) !void {
     self.size.screen = size;
     self.balancePaddingIfNeeded();
 
+    // Update renderer with latest size
+    self.renderer.setTrueSize(size);
+
     // Recalculate our grid size. Because Ghostty supports fluid resizing,
     // its possible the grid doesn't change at all even if the screen size changes.
     // We have to update the IO thread no matter what because we send
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -1286,6 +1286,8 @@ pub fn updateFrame(
     }
 }
 
+pub fn setTrueSize(_: *Metal, _: renderer.ScreenSize) void {}
+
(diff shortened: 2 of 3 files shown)


mitchellh (commented):
Thanks for taking this on, I appreciate it. Let me take a closer look and see if any of what you did makes sense anywhere else, but the fundamentals of it are sound!

jcollie:
I tried it out locally and it definitely seemed to help. Most notably when I had helix open to a blank scratch buffer. I can't really comment on the code as I have little experience with the OpenGL code.

mitchellh:
Thank you! I modified and made a new PR #7155. I gave you coauthor credit. The real change I made was around `setTrueSize`. Instead of that, I changed it to query the viewport dimensions on every frame draw. GTK guarantees the GL context viewport is updated by the time draw frame is called.

adlr:
> Thank you! I modified and made a new PR #7155. I gave you coauthor credit. The real change I made was around `setTrueSize`. Instead of that, I changed it to query the viewport dimensions on every frame draw. GTK guarantees the GL context viewport is updated by the time draw frame is called.

Thanks so much! I didn't know that was possible and it looks like a cleaner solution. Really appreciate the quick response here and I'm glad this is in.

PR #7153, opened 2025-04-21
macOS: Command Palette

mitchellh:
This introduces a command palette (inspired by @pluiedev's work in #5681, but not using it as a base) for macOS.

The command palette is available in the `View` menu and also bindable via `toggle_command_palette`, default binding is `cmd+shift+p` to match VSCode.

The commands in the command palette must map to a _bindable_ action, though they may not have an associated keybinding. This means that any new binding actions we add in the future can be represented here and also makes it easy in the future to add configuration to add new custom entries to the command palette. For this initial PR, the available commands are hardcoded (`src/input/commands.zig`). 

I've noticed in other programs (VSCode, Zed), the command palette contains pretty much _all available actions_ even if they're basically useless in the context of a command palette. For example, Zed has the "toggle command palette" action in the command palette and it... does nothing (it probably should hide the palette). I followed @pluiedev's lead and made this subjective in this PR but I wonder if we should actually force all binding actions to be available.

There are various other improvements I'd like to make but omitted from this PR for the sake of limiting scope:

  * Instead of an entry with no matches doing nothing, we can allow users to manually input _any_ configurable binding.
  * Localization, since macOS doesn't have any yet. But for Linux when we port this we probably have to change our strings extraction.

## Demo

https://github.com/user-attachments/assets/a2155cfb-d86b-4c1a-82b5-74ba927e4d69


src/config/Config.zig
```diff
@@ -4866,6 +4866,13 @@ pub const Keybinds = struct {
                 .{ .jump_to_prompt = 1 },
             );
 
+            // Toggle command palette, matches VSCode
+            try self.set.put(
+                alloc,
+                .{ .key = .{ .translated = .p }, .mods = .{ .super = true, .shift = true } },
```
pluiedev, line 4872: Super+Shift+P sounds *very* wrong to me. AFAICT GNOME Builder uses Ctrl+Return, and VS Code on Linux uses Ctrl+Shift+P

pluiedev (changes requested):

src/config/Config.zig
mitchellh: This only applies to macOS. The binding is unset on Linux. 

When we support GTK we can use our helper we already use all over to swap Ctrl for Super. Mac doesn’t use control for most shortcuts because it has the “command” (super) key. You can see that across most shared keybinds and popular programs like Chrome, Firefox, and even in your example VSCode.

src/input/command.zig
```diff
@@ -0,0 +1,407 @@
...
+    // to consider whether that new binding should have commands or not.
+    const result: []const Command = switch (action) {
+        // Note: the use of `comptime` prefix on the return values
+        // ensures that the data returned is all in the binary and
+        // and not pointing to the stack.
+
+        .reset => comptime &.{.{
+            .action = .reset,
+            .title = "Reset Terminal",
+            .description = "Reset the terminal to a clean state.",
```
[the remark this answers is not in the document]
mitchellh: Indeed, I noted the localization todo in the PR description

src/config/Config.zig
pluiedev: I mean I feel like we can do "correctly" (e.g. using `ctrlOrSuper`) in this PR, and just make the action be a no-op on GTK for now

pluiedev (approved):
The Zig code looks fine, I don't really understand what's happening in Swiftland but I'd leave the task of reviewing that to someone else

src/input/command.zig
```diff
@@ -0,0 +1,407 @@
...
+            .description = "Toggle the fullscreen state of the current window.",
+        }},
+
+        .toggle_window_decorations => comptime if (!builtin.os.tag.isDarwin()) &.{.{
+            .action = .toggle_window_decorations,
+            .title = "Toggle Window Decorations",
+            .description = "Toggle the window decorations.",
+        }} else &.{},
+
+        .toggle_secure_input => comptime if (builtin.os.tag.isDarwin()) &.{.{
```
pluiedev, line 348: Actually, secure input is already implemented on Linux (#6054). For the future IMO we should leave the responsibility of filtering out unimplemented features to the apprt, since core shouldn't really be aware of what's implemented/relevant and what's not

mitchellh: Agree on the filtering. 

But secure input is not implemented in Linux that’s a different feature. Secure input is a macOS only system call that prevents all event taps so keyloggers aren’t possible. Ghostty automatically enables secure input on macOS when it detects a password prompt or users can do it manually.

mitchellh:
Okay let's send this and iterate on tip.

PR #7168, opened 2025-04-23
Added bold-color option

roberthawdon:
As discussed in https://github.com/ghostty-org/ghostty/discussions/3134

To allow for the option to render bold text in a different colour for better visibility as an extension of `bold-is-bright`.

This is a feature that is available in other terminals.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -69,6 +69,10 @@ pub const compatibility = std.StaticStringMap(
     // this behavior. This applies to selection too.
     .{ "cursor-invert-fg-bg", compatCursorInvertFgBg },
     .{ "selection-invert-fg-bg", compatSelectionInvertFgBg },
+
+    // Ghostty 1.2 merged `bold-is-bright` into the new `bold-color`
+    // by setting the value to "bright".
+    .{ "bold-is-bright", compatBoldIsBright },
 });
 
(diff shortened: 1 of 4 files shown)


src/config/Config.zig
```diff
@@ -2352,6 +2352,9 @@ keybind: Keybinds = .{},
 /// If `true`, the bold text will use the bright color palette.
 @"bold-is-bright": bool = false,
 
+/// Set color to be used for foreground while rendering bold text.
+@"bold-color": ?Color = null,
```
mitchellh: I think we should unify `bold-color` and `bold-is-bright` since they're mutually exclusive. Maybe a custom value here that parses a tagged union.

mitchellh (changes requested):
This has to be rebased against the new renderer and I have one request on the config. The rest is acceptable.

mitchellh:
I did the rebase. I want to take a closer look at the config option though.

mitchellh:
I fixed an issue where `bold-is-bright` took priority over `bold-color`. I don't know if that was intentional but I think it makes sense for `bold-color` to always "win". Last thing I'm looking into is unifying the configs. The rest looks good.

mitchellh (approved):
Pending CI.

PR #7175, opened 2025-04-23
macOS: Command palette has no selection by default, selection wraps

mitchellh:
#7173

(1) The command palette no longer has any selection by default. If and when we introduce most recently used commands, defaulting to that would make sense. A selection only appears when the arrow keys are used or the user starts typing.

(2) The selection with arrow keys now wraps, so if you press "down" on the last option, it will wrap to the first option, and if you press "up" on the first option, it will wrap to the last option. This matches both VSCode and Zed.

### DIFF
--- a/macos/Sources/Features/Command Palette/CommandPalette.swift
+++ b/macos/Sources/Features/Command Palette/CommandPalette.swift
@@ -21,8 +21,8 @@ struct CommandPaletteView: View {
     var backgroundColor: Color = Color(nsColor: .windowBackgroundColor)
     var options: [CommandOption]
     @State private var query = ""
-    @State private var selectedIndex: UInt = 0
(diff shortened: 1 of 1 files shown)


gmile:
Selecting command palette's items using up/down arrows doesn't work for me at latest available tip build, e.g.: https://github.com/ghostty-org/ghostty/commits/185a4ea10.

https://github.com/user-attachments/assets/f7f1e280-85f7-4977-bf29-37f4c2ef3a52

This is the first time I am trying the tip build. Should the arrows be enabled in the config somehow? 

Using macOS. Came to this issue after searching for "palette arrow" string in the repo.

mitchellh:
@gmile Please don't necro issues and open a new discussion. Fwiw, arrows are working for me, so more information will be required. Locking this now.

PR #7177, opened 2025-04-23
macOS: add descriptions for PgUp, PgDown, End, and Home keys

### LINKED ISSUE #7174: macOS: some binding keys (e.g. arrows) don't show in command palette properly
See:

Image

freepicheep:
Resolves #7174. The PgUp, PgDown, End, and Home keys were not being assigned a description in the `KeyboardShortcut` extension.

### DIFF
--- a/macos/Sources/Helpers/KeyboardShortcut+Extension.swift
+++ b/macos/Sources/Helpers/KeyboardShortcut+Extension.swift
@@ -28,6 +28,10 @@ extension KeyboardShortcut: @retroactive CustomStringConvertible {
         case .downArrow: keyString = "↓"
         case .leftArrow: keyString = "←"
         case .rightArrow: keyString = "→"
+        case .pageUp: keyString = "PgUp"
+        case .pageDown: keyString = "PgDown"
+        case .end: keyString = "End"
+        case .home: keyString = "Home"
         default:
             keyString = String(key.character.uppercased())
         }


freepicheep:
Here are some commands with the new shortcut descriptions.
updated_commands

mitchellh (approved):
Thanks, the capitalization feels weird but I'm willing to accept this immediately just to fix our unknown character issue. If there's examples of idiomatic ways to do this (maybe this is it), that'd help!

freepicheep:
I agree. Zed doesn't use the unicode symbols, so Ghostty already departs from them there. Raycast gives each key in the shortcut sequence it's own separate background. They use capitalization, full spelling, and spaces for these keys. I know PRs aren't the place for design discussion, so I'll leave it at that.

Raycast UI
CleanShot 2025-04-23 at 16 23 43@2x

mitchellh:
> They use capitalization, full spelling, and spaces for these keys. I know PRs aren't the place for design discussion, so I'll leave it at that.

I think for these sorts of keys that's a good tradeoff, they're relatively rare. If you want to PR that following this feel free, otherwise I might change that eventually. I'm going to keep the auto merge on because I'd like to get the ??? symbols out ASAP. 😄

freepicheep:
Sounds good. Thanks for the feedback. I'll try to work on it tomorrow sometime.

PR #7179, opened 2025-04-24
feat: beautify macOS command palette

Aaron-212:
Resolve 1. of #7173
image

Changes made:
1. Change shortcut from `String` to `[String]` so its easier to iterate over it.
2. Overlay background color on top of an `ultraThinMaterial` for better aesthetic.
3. Reorganize and beautify the spacings and paddings.
4. Unhide the scrollbar.
5. Reorder the modifier keys to Control, Option, Shift and then Command. <https://leancrew.com/all-this/2017/11/modifier-key-order/>
6. Style shortcut keys to resemble macOS menu bar items, using corresponding symbols and fixed-width for alignment.

### DIFF
--- a/macos/Sources/Features/Command Palette/CommandPalette.swift
+++ b/macos/Sources/Features/Command Palette/CommandPalette.swift
@@ -4,7 +4,7 @@ struct CommandOption: Identifiable, Hashable {
     let id = UUID()
     let title: String
     let description: String?
-    let shortcut: String?
+    let symbols: [String]?
     let action: () -> Void
 
     static func == (lhs: CommandOption, rhs: CommandOption) -> Bool {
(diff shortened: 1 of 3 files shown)


mitchellh:
Thanks, this looks good. The only thing I'm unsure of here is whether the background should match your terminal appearance. I think having it be themed similar to your color theme for your terminal is a nice feature since terminals unlike most native apps are very much a personal aesthetic for people.

00-kat:
(Cross-referencing #7176.)

Aaron-212:
image

macos/Sources/Features/Command Palette/CommandPalette.swift
```diff
@@ -44,6 +44,12 @@ struct CommandPaletteView: View {
     }
 
     var body: some View {
+        @State var scheme: ColorScheme = if OSColor(backgroundColor).isLightColor {
```
mitchellh: I'm pretty sure this doesn't need to be `@State` or `var` (can just be a plain `let`). I tested that locally and it worked fine is there something I'm missing?

mitchellh (approved):
Yes, this looks _fantastic_. Thank you.

PR #7185, opened 2025-04-24
scroll: round up fractional mouse scroll ticks

jdrouhard:
Scrolling with a mouse on macos doesn't work very well when doing small, single tick scrolls. macos attempts to mimic precision scrolling by changing the magnitude of the scroll deltas based on scrolling speed. Slow scrolls only send deltas with a magnitude of 0.1, which isn't enough to send a single scroll event with the default scroll multiplier of 3. Changing the scroll multiplier to 10 as a workaround (so even single small scroll ticks are enough to register a scroll event) cause scrolling to be way too fast if the scroll speed is ramped up.

This commit causes the yoffset delta to be rounded out to at least a magnitude of 1 in the appropriate direction. Single scroll ticks now register as a single vertical cell scroll event, but as scroll speed is ramped up, the true delta reported to the surface is used again. Setting a scroll multiplier of 1 with the changes here makes mouse scrolling feel just as good as trackpad precision scrolling.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2766,8 +2766,21 @@ pub fn scrollCallback(
         // that a wheel tick of 1 results in single scroll event.
         const yoff_adjusted: f64 = if (scroll_mods.precision)
             yoff
-        else
-            yoff * cell_size * self.config.mouse_scroll_multiplier;
+        else yoff_adjusted: {
+            // Round out the yoff to an absolute minimum of 1. macos tries to
+            // simulate precision scrolling with non precision events by
+            // ramping up the magnitude of the offsets as it detects faster
+            // scrolling. Single click (very slow) scrolls are reported with a
+            // magnitude of 0.1 which would normally require a few clicks
+            // before we register an actual scroll event (depending on cell
+            // height and the mouse_scroll_multiplier setting).
+            const yoff_max: f64 = if (yoff > 0)
+                @max(yoff, 1)
+            else
+                @min(yoff, -1);
+
+            break :yoff_adjusted yoff_max * cell_size * self.config.mouse_scroll_multiplier;
+        };
 
         // Add our previously saved pending amount to the offset to get the
         // new offset value. The signs of the pending and yoff should match


rockorager:
On Linux, we get a magnitude of 1.0 for a generic wheel click regardless of speed. If I understand this correctly, macOS does speed scaling at the OS level for wheel ticks? As in - a wheel tick magnitude depends on the momentum of the scroll?

jdrouhard:
> On Linux, we get a magnitude of 1.0 for a generic wheel click regardless of speed. If I understand this correctly, macOS does speed scaling at the OS level for wheel ticks? As in - a wheel tick magnitude depends on the momentum of the scroll?

Yeah, that's basically exactly what happens. macOS "pretends" to do precision scrolling by ramping up the offsets when you scroll "fast" (ticks / time unit). The smallest unit (one tick / time unit) is always -0.1 or 0.1 but it quickly ramps up depending on the mouse scroll speed setting in macos' mouse settings panel.

I was pretty sure Linux always reports a magnitude of 1.0, so this change shouldn't affect the way Linux scrolling works at all. It just makes scrolling on macOS behave much closer to intuitive expectations and how other apps scroll within the OS.

rockorager:
I can confirm this doesn't change behavior of scrolling on linux. I think we should have @mitchellh confirm the behavior on macos before merging though. If he signs off on how it feels I am good on this one.

jdrouhard:
Ping @rockorager @mitchellh

jdrouhard:
Ping @mitchellh - would you be able to test / merge this when you get some time? Thanks!

mitchellh (approved):
Was able to test this with both a trackpad (precision) and mouse (not) and it seems good to me. Doesn't affect precision as expected and makes the mouse feel better. Sorry for the delay in review, took awhile to just get the setup to test this. Will merge once CI goes.

yorickvP:
This PR changes the behavior of fractional scrolling mice (like the mx master) on Linux by quite a lot, since it rounds up fractional scrolling events. It might be worthwhile to gate the clamping behavior on `builtin.os.tag == .macos`

PR #7191, opened 2025-04-25
feat: implement toggleMaximize for macOS

### LINKED ISSUE #7030: macOS: Handle `toggle_maximize` action
Looks like we forgot to consider the toggle maximize action in the macOS apprt (the case for it is simply missing, even among the fallthrough cases: https://github.com/ghostty-org/ghostty/blob/17ba0252e831457418da1b5c26ce16bdf83f0e5c/macos/Sources/Ghostty/Ghostty.App.swift#L541-L553)... I wonder if we could implement some sort of validation to make sure we do handle all known enum cases.

I'm not sure if you *could* maximize a window on macOS à la Linux or Windows, but even then we should explicitly mark it as an unsupported but *known* action.

Aaron-212:
Resolve #7030

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -807,12 +807,6 @@ class AppDelegate: NSObject,
         setSecureInput(.toggle)
     }
 
-    @IBAction func toggleMaximize(_ sender: Any) {
-        if NSApp.isActive, let keyWindow = NSApp.keyWindow {
-            keyWindow.zoom(self)
-        }
-    }
-
     @IBAction func toggleQuickTerminal(_ sender: Any) {
         if quickController == nil {
             quickController = QuickTerminalController(
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -807,6 +807,12 @@ class AppDelegate: NSObject,
         setSecureInput(.toggle)
     }
 
+    @IBAction func toggleMaximize(_ sender: Any) {
+        if NSApp.isActive, let keyWindow = NSApp.keyWindow {
+            keyWindow.zoom(self)
+        }
+    }
+
     @IBAction func toggleQuickTerminal(_ sender: Any) {
         if quickController == nil {
             quickController = QuickTerminalController(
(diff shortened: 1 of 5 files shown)


macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -807,6 +807,12 @@ class AppDelegate: NSObject,
         setSecureInput(.toggle)
     }
 
+    @IBAction func toggleMaximize(_ sender: Any) {
```
mitchellh, line 810: I think toggleMaximize should only affect terminal windows so putting this on `BaseTerminalController` and responding via first responder or via a notification seems like a better approach.

Today this doesn't matter much because all of our windows are terminal windows but that won't be true forever and if something like the command palette becomes a window one day (not tied to a specific terminal window) then it becomes a bigger issue and I'd rather not build debt for something simple here.

mitchellh (changes requested):
Thanks for taking this on, one suggestion...

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -548,6 +553,11 @@ class BaseTerminalController: NSWindowController,
         window.performClose(sender)
     }
 
+    @IBAction func toggleMaximize(_ sender: Any) {
+        guard let window = window else { return }
```
mitchellh, line 557: There needs to be a guard here that the object is within the `surfaceTree` otherwise toggle maximize will maximize all open terminal windows. 😄 

There should be at least one other action that has this check, search for `contains`.

macos/Sources/Ghostty/Package.swift
```diff
@@ -257,6 +257,9 @@ extension Notification.Name {
     /// Ring the bell
     static let ghosttyBellDidRing = Notification.Name("com.mitchellh.ghostty.ghosttyBellDidRing")
     static let ghosttyCommandPaletteDidToggle = Notification.Name("com.mitchellh.ghostty.commandPaletteDidToggle")
+
+    /// Toggle maximize of current window
+    static let ghosttyToggleMaximize = Notification.Name("com.mitchellh.ghostty.toggleMaximize")
```
mitchellh, line 262: I try to follow AppKit's notification naming convention of roughly a noun phrase followed by a past participle adjective. i.e. `[Subject] [Event (past participle)]`.

So, we should do `ghosttyFullscreenDidToggle`. 

There's an inconsistency here because the notifications were never fully converted to this style (we should do that in a future PR) but for new stuff we should follow it.

mitchellh (changes requested):
Thanks for the changes, much closer.

mitchellh (approved):

PR #7195, opened 2025-04-25
Add config color palette C bindings

freepicheep:
C bindings to expose the color palette to Swift for macOS. Returns the full 256 colors from the current color scheme. After fetching the palette with `ghostty_config_get`, you can access the desired color by its index in the list.

### Usage

Here is one way to get the palette in Swift.

```swift
import GhosttyKit

private(set) var config: ghostty_config_t? = nil {
    didSet {
        // Free the old value whenever we change
        guard let old = oldValue else { return }
        ghostty_config_free(old)
    }
}

var paletteColors: [Color] {
    var paletteList: ghostty_config_palette_s = .init()
    let key = "palette"
    
    if (!ghostty_config_get(config, &paletteList, key, UInt(key.count))) {
        return []
    }
    
    var colors: [Color] = []
    let mirror = Mirror(reflecting: paletteList.colors)
    
    for (_, element) in mirror.children {
        if let color = element as? ghostty_config_color_s {
            colors.append(Color(
                red: Double(color.r) / 255,
                green: Double(color.g) / 255,
                blue: Double(color.b) / 255
            ))
        }
    }
    
    print("Palette Colors: ", colors)
    return colors
}
```
Result (GruvboxDarkHard theme)
CleanShot 2025-04-25 at 14 21 39

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3930,6 +3930,24 @@ pub const Palette = struct {
     /// The actual value that is updated as we parse.
     value: terminal.color.Palette = terminal.color.default,
 
+    /// ghostty_config_palette_s
+    pub const C = extern struct {
+        colors: [265]Color.C,
+    };
+
+    pub fn cval(self: Self) Palette.C {
+        var result: Palette.C = undefined;
+        for (self.value, 0..) |color, i| {
+            result.colors[i] = Color.C{
+                .r = color.r,
+                .g = color.g,
+                .b = color.b,
+            };
+        }
+
+        return result;
+    }
+
     pub fn parseCLI(
         self: *Self,
         input: ?[]const u8,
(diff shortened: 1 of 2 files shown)


src/config/Config.zig
```diff
@@ -3930,6 +3930,24 @@ pub const Palette = struct {
...
 
+    /// ghostty_config_palette_s
+    pub const C = extern struct {
+        colors: [265]Color.C,
+    };
+
+    pub fn cval(self: Self) Palette.C {
+        var result: Palette.C = undefined;
+        for (self.value, 0..) |color, i| {
+            result.colors[i] = Color.C{
```
mitchellh, line 3941: Style: can do `.{` here since the result type is well known.

If you want to get really fancy you don't need the block in the `for` either since this is a single expression, but we mix that up all over the codebase.

mitchellh (approved):
A style note here that is unimportant for the merge. Thank you, this is great!

src/config/Config.zig
freepicheep: Thank you! I'm new to Zig, so I appreciate the tutoring.

PR #7201, opened 2025-04-26
macos: fix focus loss when toggling non-native fullscreen

### LINKED ISSUE #6999: macOS: focus is moved to first split when toggling non-native fullscreen
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6997

<sup>Originally posted by **frangio** April  4, 2025</sup>
This appears to have been fixed before in https://github.com/ghostty-org/ghostty/pull/405 but I'm seeing it in 1.1.3 with non-native fullscreen in macOS Sequoia 15.3.2.</div>

### Reproduction

```
```

1. Start Ghostty
2. `Cmd+D` to create right split, focus is on right
3. `Cmd+Enter` to fullscreen (non-native)

Observe: focus is on the left split

This only happens with non-native fullscreen

liby:
Fixes #6999

https://github.com/user-attachments/assets/c3426eba-8e01-4cb6-9626-efd24da3291b

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -279,11 +279,11 @@ class BaseTerminalController: NSWindowController,
 
     func titleDidChange(to: String) {
         guard let window else { return }
-        
+
         // Set the main window title
         window.title = to
     }
-    
+
     func pwdDidChange(to: URL?) {
         guard let window else { return }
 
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Thanks for investigating this.

This feels like a blunt solution. Maybe it's the right one, but I want to understand the nature of the problem more before coming to that conclusion. The first question I have around this issue is why does the responder change in the first place? And subsequently, why doesn't macOS restore it?

If for whatever reason we must manually do this, does the delegate truly need this or can we do it on animate in/out ourselves?

liby:
>  The first question I have around this issue is why does the responder change in the first place? And subsequently, why doesn't macOS restore it?

I discovered through breakpoint debugging: when removing the `.titled` style mask in non-native fullscreen mode, the focused surface is reset to the first split. 

Currently solution preserves focus by:

1. Storing the focused surface in `SavedState` during initialization
2. Using this saved reference in `fullscreenDidChange()`
3. Ensuring we call `delegate?.fullscreenDidChange()` before clearing `savedState`

Does this approach seem reasonable? I'm open to better alternatives if you have other ideas for solving this focus preservation issue.

mitchellh:
Thanks for the iteration. I spent a bit more time on this and think that #7209 is a less invasive fix, and gets a bit closer to the root cause.

PR #7211, opened 2025-04-27
z2d 0.6.0 -> 0.6.1 for security fix

rhodes-b:
version 0.6.0 of z2d has a security vulnerability and the author suggests upgrading (randomly saw this in the zig discord)
https://github.com/vancluever/z2d/security/advisories/GHSA-mm4c-p35v-7hx3

I don't know enough about how ghostty uses to it to say if it could be affected

also updates that Context no longer returns an error type this says it was in the 0.6.0 changes which ghostty was already pinned to (it seems like that the tag might have been updated or something im not sure) you can see it listed as a change in 0.6.0 below
https://github.com/vancluever/z2d/blob/main/CHANGELOG.md

 I'm not sure if any other changes are required I ran a `zig build` and `zig build test` and things seem to be fine on my end

~~I'm not sure I updated the zon nix stuff correctly I just ran the command in jcollie readme
`nix run github:jcollie/zon2nix#zon2nix -- --nix=build.zig.zon.nix build.zig.zon`~~ looks like I should have just ran the script

ps: sorry for the direct ping mitchell (I thought this might be important that you look at it)

### DIFF
--- a/src/font/sprite/canvas.zig
+++ b/src/font/sprite/canvas.zig
@@ -149,8 +149,8 @@ pub const Canvas = struct {
     }
 
     /// Acquires a z2d drawing context, caller MUST deinit context.
-    pub fn getContext(self: *Canvas) Allocator.Error!z2d.Context {
-        return try z2d.Context.init(self.alloc, &self.sfc);
+    pub fn getContext(self: *Canvas) z2d.Context {
+        return z2d.Context.init(self.alloc, &self.sfc);
     }
 
     /// Draw and fill a single pixel
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -20,8 +20,8 @@
         },
         .z2d = .{
             // vancluever/z2d
-            .url = "https://github.com/vancluever/z2d/archive/1e89605a624940c310c7a1d81b46a7c5c05919e3.tar.gz",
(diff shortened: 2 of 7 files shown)


src/font/sprite/canvas.zig
```diff
@@ -149,8 +149,8 @@ pub const Canvas = struct {
     }
 
     /// Acquires a z2d drawing context, caller MUST deinit context.
-    pub fn getContext(self: *Canvas) Allocator.Error!z2d.Context {
-        return try z2d.Context.init(self.alloc, &self.sfc);
+    pub fn getContext(self: *Canvas) z2d.Context {
+        return z2d.Context.init(self.alloc, &self.sfc);
```
mitchellh, line 153: This can be `.init(` now with Zig 0.14 but I'm not going to block the merge on that. 😄

mitchellh (approved):

mitchellh:
Thanks! This was never part of a released version of Ghostty (1.1.3 is still on v0.4.x which was unaffected). Additionally, I'm not even sure we could trigger this issue since all of our draw calls are hardcoded and we have bounds checking on the canvas. In any case, there's no need for a security advisory from our side for either reason. But getting it updated in tip is good.

PR #7233, opened 2025-04-30
Fix removed GDK_DEBUG  `gl-no-fractional`

morgan9e:
`gl-no-fractional` is removed from GTK v4.17.5

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -190,7 +190,6 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
             // For the remainder of "why" see the 4.14 comment below.
             gdk_disable.@"gles-api" = true;
             gdk_disable.vulkan = true;
-            gdk_debug.@"gl-no-fractional" = true;
             break :environment;
         }
         if (gtk_version.runtimeAtLeast(4, 14, 0)) {
@@ -201,8 +200,11 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
             //
             // Upstream issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/6589
             gdk_debug.@"gl-disable-gles" = true;
-            gdk_debug.@"gl-no-fractional" = true;
             gdk_debug.@"vulkan-disable" = true;
+
+            if (gtk_version.runtimeUntil(4, 17, 5)) {
+                gdk_debug.@"gl-no-fractional" = true;
+            }
             break :environment;
         }
         // Versions prior to 4.14 are a bit of an unknown for Ghostty. It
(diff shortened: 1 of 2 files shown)


morgan9e:
> First of all, why is this necessary? gl-no-fractional has not done anything for quite a while but all it does is adding a log message at startup. Unless GTK 4.17.5+ can't recover from having an invalid debug flag (which isn't true) this shouldn't need to be here.
> 
> There are still some users that are using GTK versions that are quite a bit older (Debian 12 ships GTK 4.8, for example) and might still benefit from this.

I just added upper bound check, nothing had changed.

And okay, I'll check what's minimum version for that flag and add it.

pluiedev:
If the only intention here is to suppress the unknown debug flag message then this PR makes a lot more sense - thanks. Still, it was definitely removed at a point a lot earlier than 4.17.5; if we were to really care about this we should have due diligence and set the upper bound to whichever version first removed it.

morgan9e:
Thanks. I double checked, its 4.17.5.

https://gitlab.gnome.org/GNOME/gtk/-/blob/main/NEWS#L442

src/apprt/gtk/App.zig
```diff
@@ -201,8 +200,11 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
             //
             // Upstream issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/6589
             gdk_debug.@"gl-disable-gles" = true;
-            gdk_debug.@"gl-no-fractional" = true;
             gdk_debug.@"vulkan-disable" = true;
+
+            if (gtk_version.runtimeUntil(4, 17, 5)) {
```
mitchellh, line 205: I know it didn't exist before, but we should take this opportunity to add a comment about why we set this at all. The history starts here: https://github.com/ghostty-org/ghostty/pull/4255

I'm not sure how to word the explanation because I myself don't fully grasp the nature of fractional scaling with GTK yet. I just know we don't handle it well at the moment and it causes blurriness. I think the blurriness returned after they removed this and we need to still address this. I'm not sure.

mitchellh (commented):
One comment. I won't approve/deny here, I'm fine merging without my comment being addressed, but did want to note it. I think removing the noisy log message is good because it does lead people to feel they should report a bug.

src/apprt/gtk/App.zig
morgan9e: I think it was quick and dirty fix, there was upstream issue about it. ([GNOME/gtk#6514](https://gitlab.gnome.org/GNOME/gtk/-/issues/6514))

NGL has little issue on fractional scaling which can cause blurry fonts, and `gl-no-fractional` just reverts it and do integer scaling. it makes fonts crisp, but its legacy solution.

morgan9e: There is upstream issue about it. ([GNOME/gtk#6514](https://gitlab.gnome.org/GNOME/gtk/-/issues/6514)), so.. best solution was to rely on upstream fix but I thought quick and dirty fix as beneficial.

NGL has little issue on fractional scaling which can cause blurry fonts, and gl-no-fractional just reverts it and do integer scaling. it makes fonts crisp, but its legacy solution.

PR #7237, opened 2025-04-30
macos: add float on top feature for terminal windows

hettiger:
Closes #5188

**See:** 

- https://github.com/ghostty-org/ghostty/discussions/5188#discussioncomment-12607781
- https://github.com/ghostty-org/ghostty/discussions/5188#discussioncomment-12984898


mitchellh (approved):
Thank you! Very clean. I made a couple changes and fixed one issue I found:

* Removed the custom notifications and use `windowDidBecomeKey`, guarding on object being a TerminalWindow
* Don't unset the menu state on resign key, because we actually resign key when we LOOK AT the menu, so this was causing a funny behavior where you'd click "Window" and it'd be checked then shortly later be unchecked. 
* Implement `NSMenuItemValidation` so that the menu items are only active on terminal windows (importantly: not quick terminals!)
* Move the userdefaults string out to a constant

hettiger:
Really happy to see this merged @mitchellh Thank you! 🎉

PR #7238, opened 2025-04-30
font/freetype: introduce mutexes to ensure thread safety of Library and Face

qwerasd205:
tl;dr: FT_New_Face and FT_Done_Face require the Library to be locked for thread safety, and FT_Load_Glyph and FT_Render_Glyph and friends need the face to be locked for thread safety, since we're sharing faces across threads.

For details see comments and FreeType docs @
- https://freetype.org/freetype2/docs/reference/ft2-library_setup.html#ft_library
- https://freetype.org/freetype2/docs/reference/ft2-face_creation.html#ft_face

---

This might resolve the issue discussed in #7016 but I wasn't able to reliably reproduce it in a way I could debug, so someone who was experiencing it should probably test this PR.

Additionally I can still semi-reliably produce a crash with the GTK apprt by setting an `all:` keybind to adjust font sizes and changing the font size rapidly with many surfaces open with emojis on them. Unfortunately I can't really tell what the root cause is because the debug info is broken in QEMU.

However, I do think this is a good idea for us to be thread safe with this stuff even if it isn't related to that problem.

### DIFF
--- a/src/font/Collection.zig
+++ b/src/font/Collection.zig
@@ -714,15 +714,18 @@ test "add full" {
         ) });
     }
 
-    try testing.expectError(error.CollectionFull, c.add(
-        alloc,
-        .regular,
-        .{ .loaded = try Face.init(
-            lib,
-            testFont,
-            .{ .size = .{ .points = 12 } },
-        ) },
-    ));
+    var face = try Face.init(
+        lib,
+        testFont,
+        .{ .size = .{ .points = 12 } },
+    );
+    // We have to deinit it manually since the
+    // collection doesn't do it if adding fails.
+    defer face.deinit();
+    try testing.expectError(
+        error.CollectionFull,
+        c.add(alloc, .regular, .{ .loaded = face }),
+    );
 }
 
 test "add deferred without loading options" {
--- a/src/font/Collection.zig
+++ b/src/font/Collection.zig
@@ -78,8 +78,8 @@ pub const AddError = Allocator.Error || error{
 /// next in priority if others exist already, i.e. it'll be the _last_ to be
 /// searched for a glyph in that list.
 ///
-/// The collection takes ownership of the face. The face will be deallocated
-/// when the collection is deallocated.
+/// If no error is encountered then the collection takes ownership of the face,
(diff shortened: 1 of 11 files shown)


qwerasd205:
Oh... a test failure... I'll look in to it later tonight if I have time-- it seems like it's a face that isn't properly deinited, which didn't used to create a test failure because there was nothing allocated for it with a Zig allocator, but because of the added mutex it's now sensitive to that.

src/font/Collection.zig
```diff
@@ -93,6 +93,13 @@ pub fn add(
 ) AddError!Index {
     const list = self.faces.getPtr(style);
 
+    // If we error here, we're responsible for
+    // deiniting the face that we were passed.
+    errdefer {
+        var f = face;
+        f.deinit();
+    }
```
mitchellh, line 101: This block is suspicious. We do take ownership of the face but usually in APIs like this I've found this only applies _when it succeeds_. If it doesn't succeed I think it's fair to let the caller determine what they want to do with the face (it isn't clear to me that they definitely want it reinitialized).

mitchellh (changes requested):
Of course the tests need to be fixed up but I noted one thing I'm suspicious about.

mitchellh:
And of course thanks for chasing this down. ❤️

src/font/Collection.zig
qwerasd205: Yeah I added this because of a failing test (`Collection.test.add full`), but it could just be changed to handle it at any call site if you think that'd be better

mitchellh: I think that would be more functionally correct, though perhaps noisier or more annoying to implement in tests. unsure. But I think we should do it. I'm happy to do that cleanup myself if you don't get to it.

qwerasd205:
`build-snap` test failures seem unrelated to this PR, everything else passes now.

NicoElbers:
From my quick 5 minute tryout, at least 2 cases that used to very consistently cause a crash on my machine no longer seem to crash with this branch. I'll stay on this branch until it's merged and report if I find anything still crashing. 

Thank you for your work :)

mitchellh (approved):

PR #7241, opened 2025-05-01
apprt/gtk: ensure configuration is loaded on startup

alaviss:
Restores the app configuration code removed in
https://github.com/ghostty-org/ghostty/pull/6792.

The was unnoticed due to `colorSchemeEvent` triggering a configuration reload if `window-theme` deviates from the default (i.e. dark mode is used).

Fixes https://github.com/ghostty-org/ghostty/discussions/7206

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -1291,6 +1291,13 @@ pub fn run(self: *App) !void {
     // Setup our actions
     self.initActions();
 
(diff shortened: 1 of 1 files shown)


tristan957 (approved):

mitchellh (approved):
Funny enough, if I'm reading through the codepath right, we also wouldn't be setting up menu accelerators properly without this initial call? Anyways, this looks good. Thanks.

PR #7246, opened 2025-05-01
Binding for toggling window float on top (macOS only)

mitchellh:
This adds a keybinding and apprt action for #7237.


mitchellh:
I also took the liberty in this PR to fix all our `: void` usage in the files I touched.

PR #7248, opened 2025-05-01
Implement dark/light theme filtering in theme preview

### LINKED ISSUE #7235: Create a flag to filter dark and light themes in `+list-themes`
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7223

<sup>Originally posted by **SamuelGadiel** April 28, 2025</sup>
The command `ghostty +list-themes` lists all the themes currently available.
It shows both dark and light themes, but most of the time we want to see only one of them.
It would be really helpful to have an option to filter what themes are going to be shown by the `+list-themes` command.

We could have a flag to pass into the command

```sh
ghostty +list-themes                   # Shows both dark and light themes
ghostty +list-themes --dark            # Shows only the dark themes
ghostty +list-themes --light           # Shows only the light themes
ghostty +list-themes --dark --light    # Shows both dark and light themes
```

And maybe an option inside the viewer that cycles between light, dark and both.
Something like pressing "ctrl+b" to change the brightness filter
Or a clickable button for those who uses a mouse in the terminal (not required, but it's inclusive for them)

We could take the background color, compute its luminance and determine if it is a dark or light color based on a threshold
Generally if the luminance is above 0.5 it's a light color, and below 0.5 it's a dark color.

The WCAG standard recommends a lower threshold (≈ 0.179) to ensure a minimum contrast ratio of 4.5:1 for text.
But 0.5 usually works well, since we just want to classify the themes in two groups, and the ratio was already considered when creating the theme.

What do you guys think about it?</div>

b6k-dev:
This pr closes https://github.com/ghostty-org/ghostty/issues/7235

Demo:

https://github.com/user-attachments/assets/88366ac7-fcdb-49e9-bed7-8deb0eebeb9e

### DIFF
--- a/src/cli/list_themes.zig
+++ b/src/cli/list_themes.zig
@@ -24,6 +24,12 @@ pub const Options = struct {
     /// If true, force a plain list of themes.
     plain: bool = false,
 
+    /// If true, print only the dark themes.
+    dark: bool = false,
+
+    /// If true, print only the light themes.
+    light: bool = false,
+
     pub fn deinit(self: Options) void {
         _ = self;
     }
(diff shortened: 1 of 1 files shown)


src/cli/list_themes.zig
```diff
@@ -24,6 +24,12 @@ pub const Options = struct {
     /// If true, force a plain list of themes.
     plain: bool = false,
 
+    /// If true, print only the dark themes.
+    dark: bool = false,
+
+    /// If true, print only the light themes.
+    light: bool = false,
```
mitchellh, line 31: We should've had this discussion in the discussion, but perhaps we should change to a `--color` flag that takes an enum value `all/light/dark` right now so we can enforce only one. 

Then the implementation can be a nice switch.

mitchellh (changes requested):
Thanks, one note.

src/cli/list_themes.zig
m4k-dev: Yeah, it definitely makes more sense to have just one flag.

mitchellh (approved):

PR #7253, opened 2025-05-02
bash: explicitly request a login shell

jparise:
Prior to #7044, on macOS, our shell integrated command line would be executed under `exec -l`, which caused bash to be started as a login shell. Now that we're using direct command execution, add `--login` to our bash command's arguments on macOS to get that same behavior.

### DIFF
--- a/src/termio/shell_integration.zig
+++ b/src/termio/shell_integration.zig
@@ -239,7 +239,7 @@ fn setupBash(
     resource_dir: []const u8,
     env: *EnvMap,
 ) !?config.Command {
-    var args = try std.ArrayList([:0]const u8).initCapacity(alloc, 2);
+    var args = try std.ArrayList([:0]const u8).initCapacity(alloc, 3);
     defer args.deinit();
 
     // Iterator that yields each argument in the original command line.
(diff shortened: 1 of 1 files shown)


jparise:
I don't love the way the unit tests look after this change (with the `isDarwin()` conditions). If there's a nice pattern to use there, I'm all for it.

jparise:
Alternatively, we could change the bash shell integration to return a `.shell` command, which would match the non-shell-integrated path and more specifically match the prior behavior. That would be a lot simpler of a code change because it won't spread the `--login` logic across two places, and I don't think the `.direct` command approach brings us another other advantages in this case.

src/termio/shell_integration.zig
```diff
@@ -239,20 +239,27 @@ fn setupBash(
...
-    // Start accumulating arguments with the executable and `--posix` mode flag.
+    // Start accumulating arguments with the executable and initial flags.
     if (iter.next()) |exe| {
         try args.append(try alloc.dupeZ(u8, exe));
     } else return null;
     try args.append("--posix");
 
+    // On macOS, we request a login shell to match that platform's norms.
+    // We need to do this explicitly because we've also set the --posix
+    // flag, which suppresses some of the default bash startup sequence.
```
jparise, line 258: I'm still a little unsure about this comment (reason).

mitchellh (approved):
For now let's go with this. If this ends up being not the right approach then this is a simple enough commit to redo. Thank you!

danielwe:
(With apologies for the necromancy) A discord user pointed out that bash on macOS doesn't get `-` prepended to `$0` as one usually expects for login shells. Other shells are not affected, and neither is bash without shell integration. This is a consequence of choosing this PR over #7254.

More generally, it's unclear to me how the custom-built command line for bash addresses all the concerns that led to the `login -l bash -c "exec -l <command>"` wrapper in the first place, as described in the big explainer comment: https://github.com/ghostty-org/ghostty/blob/4a173052fb181580b1ea013555508b8f37c0407e/src/termio/Exec.zig#L1462-L1482

Is this worth reconsidering, for consistent behavior across shells and configurations?

jparise:
> Is this worth reconsidering, for consistent behavior across shells and configurations?

Yes, I'll spend some time soon revisiting this.

PR #7254, opened 2025-05-02
bash: use a shell command for shell integration

jparise:
Prior to #7044, on macOS, on macOS, our shell-integrated command line would be executed under `exec -l`, which causes bash to be started as a login shell. This matches the macOS platform norms.

The change to direct command execution meant that we'd skip that path, and bash would start as a normal interactive (non-login) shell on macOS.

Restore the previous behavior by returning a shell command from our bash integration routine. This is consistent with the code path used when shell integration is disabled.

We could alternatively add `--login` to our shell-integrated bash command line for the macOS build. That would let us continue using the direct command strategy, but it would also result in having coordinated, conditional behavior in two places (shell integration? macOS?).

### DIFF
--- a/src/termio/shell_integration.zig
+++ b/src/termio/shell_integration.zig
@@ -326,9 +326,8 @@ fn setupBash(
     );
     try env.put("ENV", integ_dir);
 
-    // Since we built up a command line, we don't need to wrap it in
-    // ANOTHER shell anymore and can do a direct command.
-    return .{ .direct = try args.toOwnedSlice() };
+    // Join the accumulated arguments to form the final command string.
+    return .{ .shell = try std.mem.joinZ(alloc, " ", args.items) };
 }
 
 test "bash" {
@@ -342,9 +341,7 @@ test "bash" {
 
     const command = try setupBash(alloc, .{ .shell = "bash" }, ".", &env);
 
-    try testing.expectEqual(2, command.?.direct.len);
-    try testing.expectEqualStrings("bash", command.?.direct[0]);
-    try testing.expectEqualStrings("--posix", command.?.direct[1]);
+    try testing.expectEqualStrings("bash --posix", command.?.shell);
     try testing.expectEqualStrings("./shell-integration/bash/ghostty.bash", env.get("ENV").?);
     try testing.expectEqualStrings("1", env.get("GHOSTTY_BASH_INJECT").?);
 }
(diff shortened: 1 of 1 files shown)


mitchellh:
> We could alternatively add --login to our shell-integrated bash command line for the macOS build. That would let us continue using the direct command strategy, but it would also result in having coordinated, conditional behavior in two places (shell integration? macOS?).

I think this is the better option (your other PR). 

I think the overhead of starting a process just to exec another process just to get a login shell is pretty silly. And it is purely a macOS-ism, and this PR affects other platforms negatively (would result in wrapping on Linux).

jparise:
> > We could alternatively add --login to our shell-integrated bash command line for the macOS build. That would let us continue using the direct command strategy, but it would also result in having coordinated, conditional behavior in two places (shell integration? macOS?).
> 
> I think this is the better option (your other PR).
> 
> I think the overhead of starting a process just to exec another process just to get a login shell is pretty silly. And it is purely a macOS-ism, and this PR affects other platforms negatively (would result in wrapping on Linux).

Yes, that makes sense. Do you think we should (separately) attempt to start the (default) shell using `.direct` instead? The main reason I came back to the approach here is to match what we're doing in the non-shell-integration case.

jparise:
> > Do you think we should (separately) attempt to start the (default) shell using .direct instead? The main reason I came back to the approach here is to match what we're doing in the non-shell-integration case.
> 
> Yes, I did the easy thing originally and it had problems so I backed out of it in my PR. But I do want to ultimately have more cases where don't do the wrapping.

In that case, we could also address issue this at that level, where it's the initial (bash) shell command that gets the `--login` treatment, and then the shell integration code would inherit that argument from the initial command without an specialization.

jparise:
> > Do you think we should (separately) attempt to start the (default) shell using .direct instead? The main reason I came back to the approach here is to match what we're doing in the non-shell-integration case.
> 
> Yes, I did the easy thing originally and it had problems so I backed out of it in my PR. But I do want to ultimately have more cases where don't do the wrapping.

I think I found the source of that issue (or at least one of them 😉): #7262

mitchellh:
I just checked fish and zsh and they don't seem to have the same problem (they both report being login shells to me on macOS) so I think special-casing bash with the `--login` flag is fine.

jparise:
> I just checked fish and zsh and they don't seem to have the same problem (they both report being login shells to me on macOS) so I think special-casing bash with the `--login` flag is fine.

That's right, because the bash shell integration case is the only one that alters the shell command. The others all use the initial command as given, which is currently always a `.shell` `Command`.

PR #7262, opened 2025-05-03
termio: exec direct commands as login sessions

jparise:
When executing a `.direct` command under macOS, we pass the arguments directly to login. We want this to be a login session so we don't want to use login's `-l` flag (which suppress the login session behavior).

We continue to use `login -l` for `.shell` commands because we wrap them in a call to `exec -l`, which handles the login session itself.

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -1575,13 +1575,15 @@ fn execCommand(
         // Awesome.
         try args.append("/usr/bin/login");
         if (hush) try args.append("-q");
-        try args.append("-flp");
+        if (command == .shell) try args.append("-l");
+        try args.append("-fp");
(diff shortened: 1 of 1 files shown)


jparise:
Hmm, I forgot that this might cause a problem with the initial directory (being forced to $HOME). I'll check that and close this if it's a dead end. _sigh_

jparise:
> Hmm, I forgot that this might cause a problem with the initial directory (being forced to $HOME). I'll check that and close this if it's a dead end. _sigh_

Yeah, there's unfortunately no way around that, so we can't make this work with direct args, either.

I think we'll need to either require all commands to opt-in to being a login shell theirselves or _also_ use `exec -l` in the direct path to get a login shell.

PR #7265, opened 2025-05-03
Add "Scroll to Selection" command

FnControlOption:
This is mostly intended for implementing the "Jump to Selection" (<kbd>⌘</kbd> <kbd>J</kbd>) command on macOS (but as per the TODO, the pin still needs to be centered)

Demo: (Selected text is "10")

<table><tr>
<td>Screenshot 2025-05-03 at 9 21 55 AM</td>
<td>Screenshot 2025-05-03 at 9 21 58 AM</td>
</tr></table>

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4119,6 +4119,14 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
             }, .unlocked);
         },
 
+        .scroll_to_selection => {
+            self.renderer_state.mutex.lock();
+            defer self.renderer_state.mutex.unlock();
+            const sel = self.io.terminal.screen.selection orelse return false;
+            const tl = sel.topLeft(&self.io.terminal.screen);
+            self.io.terminal.screen.scroll(.{ .pin = tl });
+        },
+
         .scroll_page_up => {
             const rows: isize = @intCast(self.size.grid().rows);
             self.io.queueMessage(.{
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -279,6 +279,7 @@ pub const Action = union(enum) {
     /// Scroll the screen varying amounts.
     scroll_to_top,
     scroll_to_bottom,
+    scroll_to_selection,
     scroll_page_up,
     scroll_page_down,
     scroll_page_fractional: f32,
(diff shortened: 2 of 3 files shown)


src/Surface.zig
```diff
@@ -4119,6 +4119,17 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
             }, .unlocked);
         },
 
+        .scroll_to_selection => {
+            const sel = self.io.terminal.screen.selection orelse return false;
+            const pin = sel.start();
+            // TODO: If the selection is outside the active area and smaller
+            // than the viewport, we should move the pin backwards a little so
+            // that the selection is vertically centered.
```
mitchellh, line 4127: The TODO I think is unnecessary, we can consider this a future improvement if we want.

src/Surface.zig
```diff
@@ -4119,6 +4119,17 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
             }, .unlocked);
         },
 
+        .scroll_to_selection => {
+            const sel = self.io.terminal.screen.selection orelse return false;
```
mitchellh, line 4123: This is unsafe, we need the lock.

src/config/Config.zig
```diff
@@ -4754,6 +4754,11 @@ pub const Keybinds = struct {
                 .{ .key = .{ .translated = .end }, .mods = .{ .super = true } },
                 .{ .scroll_to_bottom = {} },
             );
+            try self.set.put(
+                alloc,
+                .{ .key = .{ .translated = .j }, .mods = .{ .super = true } },
+                .{ .scroll_to_selection = {} },
+            );
```
mitchellh, line 4761: I don't think this needs a default binding. If you want one, we should discuss it. Why did you choose this?

mitchellh (changes requested):
I love this. Thank you. With the command palette as demonstrated this is extra useful. There are some changes I plan on making myself, but please read the comments.

mitchellh:
I made some changes. I realized that sending a pin across the message queue is wholly unsafe because by the time it is processed it may be invalid (point to dangling memory). So we need to do the scroll immediately within the confines of the lock.

mitchellh (approved):

src/config/Config.zig
FnControlOption: > Why did you choose this?

It’s the binding that Terminal.app uses for Edit -> Find -> Jump to Selection, but admittedly I actually never knew it existed until recently. I was experimenting with adding a Find bar to Ghostty and noticed that Jump to Selection was a thing

PR #7271, opened 2025-05-05
macOS: don't crash when opening non-native fullsreen at launch

dzhlobo:
When we launch ghostty with `fullscreen = true` and `macos-non-native-fullscreen = true` there's no window to check CGSSpaces for.

Fixes #7114


mitchellh:
Thanks for this. I went a different direction with #7277.

I think longer term I wonder if the issue is more a race condition. I'm still investigating that route.

dzhlobo:
@mitchellh, thank you ❤️

PR #7273, opened 2025-05-06
fix: ensure consistent mouse event format in terminfo

gnodet:
The terminal code in Surface.zig correctly sends '\x1b[M' for mouse events,
which is properly defined in the terminfo entry as '\E[M' for the kmous
capability. This commit ensures that the mouse event format is consistent
between the implementation in Surface.zig and the terminfo definition.

This maintains compatibility with applications that rely on the correct
mouse event format when interacting with the terminal.

See https://github.com/ghostty-org/ghostty/blob/6e11d947e7ae6f37567faddb863b48107cdf278b/src/Surface.zig#L2696-L2698

### DIFF
--- a/src/terminfo/ghostty.zig
+++ b/src/terminfo/ghostty.zig
@@ -379,7 +379,7 @@ pub const ghostty: Source = .{
         .{ .name = "khome", .value = .{ .string = "\\EOH" } },
         .{ .name = "kich1", .value = .{ .string = "\\E[2~" } },
         .{ .name = "kind", .value = .{ .string = "\\E[1;2B" } },
-        .{ .name = "kmous", .value = .{ .string = "\\E[<" } },
+        .{ .name = "kmous", .value = .{ .string = "\\E[M" } },
         .{ .name = "knp", .value = .{ .string = "\\E[6~" } },
         .{ .name = "kpp", .value = .{ .string = "\\E[5~" } },
         .{ .name = "kri", .value = .{ .string = "\\E[1;2A" } },


rockorager:
What is the motivation or problem this is solving?

The only terminfo files I see that define `kmous` this way are console based or do not otherwise support SGR mouse reports. There are 25 in total by my count in [terminfo.src](https://invisible-island.net/ncurses/terminfo.src.html) (search for "kmous=\E[M")

The reason for `\E[<` is that ghostty's `XM` entry (which enables mouse mode) turns on SGR reports, which have the prefix `\E[<` The `\E[M` prefix is *only* for X10 reports, which are considered legacy.

gnodet:
> What is the motivation or problem this is solving?
> 
> The only terminfo files I see that define `kmous` this way are console based or do not otherwise support SGR mouse reports. There are 25 in total by my count in [terminfo.src](https://invisible-island.net/ncurses/terminfo.src.html) (search for "kmous=\E[M")
> 
> The reason for `\E[<` is that ghostty's `XM` entry (which enables mouse mode) turns on SGR reports, which have the prefix `\E[<` The `\E[M` prefix is _only_ for X10 reports, which are considered legacy.

When I enable mouse tracking by using "\E[

> What is the motivation or problem this is solving?
> 
> The only terminfo files I see that define `kmous` this way are console based or do not otherwise support SGR mouse reports. There are 25 in total by my count in [terminfo.src](https://invisible-island.net/ncurses/terminfo.src.html) (search for "kmous=\E[M")
> 
> The reason for `\E[<` is that ghostty's `XM` entry (which enables mouse mode) turns on SGR reports, which have the prefix `\E[<` The `\E[M` prefix is _only_ for X10 reports, which are considered legacy.

The latest ncurses do not support `xm` and `XM`, and it's supposed to be the capability used to toggle SGR codes.
See https://gist.github.com/cstamas/e4bec7b9c1bff6326c5dc2bac709bcb9

rockorager:
> The latest ncurses do not support `xm` and `XM`

`XM` is not an officially supported capability, so you need to include `-x`. I'm on arch linux and the `xterm` terminfo includes it:

```
~  infocmp -x xterm | rg XM
        XM=\E[?1006;1000%?%p1%{1}%=%th%el%;, XR=\E[>0q,
```

rockorager:
> I'm the [JLine](https://github.com/jline/jline3) author and found that JLine's mouse support does not work well with ghostty. Admittedly, it does not support SGR mouse events...

I'm curious if you have this issue on other terminals, eg kitty, alacritty, foot, xterm even. Each of these has their XM capability set to turn on SGR mode and has a `kmous=\E[<` entry.

I am extremely hesitant to approve this PR given that this is the first sign of this issue and that none of these other popular, older projects have made this change.

gnodet:
> > I'm the [JLine](https://github.com/jline/jline3) author and found that JLine's mouse support does not work well with ghostty. Admittedly, it does not support SGR mouse events...
> 
> I'm curious if you have this issue on other terminals, eg kitty, alacritty, foot, xterm even. Each of these has their XM capability set to turn on SGR mode and has a `kmous=\E[<` entry.
> 
> I am extremely hesitant to approve this PR given that this is the first sign of this issue and that none of these other popular, older projects have made this change.

I'll add support for parsing SGR and other mouse event formats in JLine.

PR #7277, opened 2025-05-06
macOS: handle scenario cgWindowId is nil

mitchellh:
Fixes #7114
Supercedes #7271

This fixes a crash that could occur with non-native fullscreen and `fullscreen = true` set at once.

The "windowNumber" can be `<= 0` if the window "doesn't have a window device." I don't fully know all the scenarios this is true but it is true when the window is not visible, at least.


mitchellh:
I think a better solution here is to wait for the window device to become available, but this should fix a bug anyways so I want to merge this and build on top of it.

PR #7298, opened 2025-05-08
Invert cursor

alienman5k:
Merge from main branch


src/config/Config.zig
```diff
@@ -888,7 +892,7 @@ title: ?[:0]const u8 = null,
 /// The class name must follow the requirements defined [in the GTK
 /// documentation](https://docs.gtk.org/gio/type_func.Application.id_is_valid.html).
 ///
-/// The default is `com.mitchellh.ghostty`.
+/// The default is `com.ghostty.terminal`.
```
00-kat, line 895: What is this change for, it's both incorrect and completely unrelated to cursors.

00-kat (changes requested):

Issue #7316, opened 2025-05-09
Clarify building from source tarballs vs. git in website and PACKAGING.md


mitchellh:
For Ghostty 1.2.x (current tip), we now require newer dependencies to build from _git_ vs. building from _source_. Our source tarballs contain generated artifacts that let us build with older dependencies. Namely, we require libadw 1.5+ to build from git and libadw 1.2+ (at the time of authoring this issue) to build from source. 

Note that libadwaita is used as an example here but isn't the only dependency. Building from Git also depends on `git` being available to generate the version, our source tarball has the version embedded so it doesn't require `git`.

And in the future we'll probably continue to add more pre-processing into our dist step.

This different isn't obvious and we need to update this anyways for the eventual 1.2 release.

mitchellh:
This is done.

PR #7317, opened 2025-05-09
update PACKAGING.md to be explicit about source vs. git

mitchellh:
Related to #7316

I'm not going to close the issue yet because I want to also update the website, but this should help clarify things more. I noticed the README already makes a distinction, thankfully.


mitchellh:
cc @pluiedev

pluiedev (approved):

PR #7320, opened 2025-05-09
Use W3C key code specification for keybindings (Breaking Change)

### LINKED ISSUE #7309: Ctrl+A, Ctrl+B, Ctrl+C etc. should always be handled as if the current KB layout were US English, at least for non-Latin layouts
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6724

<sup>Originally posted by **sirburpalot** March 14, 2025</sup>
The current behavior does not match the behavior of any other popular terminal app (Konsole, gnome-terminal, Kitty, iTerm2).

Steps to reproduce:

1. Open a shell in Ghostty.
2. Switch the keyboard layout to Russian, Greek or anything else that doesn't use the Latin alphabet.
3. Press Ctrl-C or Ctrl-D. Instead of the expected behavior, you'll see a bunch of escape sequences appear in the command line.</div>

### LINKED ISSUE #7310: macOS: German Swiss layout `ctrl+^` doesn't produce correct value
Discussed in https://github.com/ghostty-org/ghostty/discussions/5780

1. Set keyboard layout to "Swiss German"
2. Press `ctrl+=` on a US physical keyboard
3. Expect to see `=` on a shell prompt, but don't.

Also look at `kitty +kitten show_key` and notice the output doesn't match at all. 

Expected: `=`

Actual: `\x1e`

### LINKED ISSUE #7314: macOS: Apostrophe key and norwegian key 'æ' seem conflated
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6789

<sup>Originally posted by **torarvid** March 18, 2025</sup>
As shown in the video below, when I press apostrophe `'` and the `æ` key on my keyboard (with norwegian layout — on macOS if that matters), the dev-tools recognize it as `apostrophe` in both cases, even though they print correctly.

However, when I press `cmd+æ`, I get the same behavior as if I press `cmd+'`, resulting in the quick_terminal opening. In the video, I first press `'`, `æ`, `'`. `æ` (so you see the chars in the terminal window). Then while holding `cmd`, I press `'` and `'` again (**expected** showing+hiding the quick terminal), then `æ` and `æ` again (**unexpected** showing+hiding the quick terminal).

https://github.com/user-attachments/assets/4781aece-e0c5-4fc7-a056-607c50358770

</div>

### LINKED ISSUE #7315: macOS: Shortcuts are broken on "Dvorak-Qwerty ⌘" layout
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5466

<sup>Originally posted by **Viknet** January 31, 2025</sup>
"Dvorak-Qwerty ⌘" transmorphic layout uses the Dvorak layout for unmodified keystrokes, but unlike the “normal” Dvorak layout, it actually retains the QWERTY layout for keys when pressed in combination with the ⌘ modifier.
For example: if i press key "v" on a keyboard, letter "k" will print out. But if I hold ⌘ and press key "v", it should activate "⌘+v" shortcut and do paste from clipboard.

After recent update to version 1.1.0 (build 9283, commit [facda0c3](https://github.com/ghostty-org/ghostty/commits/facda0c3)) shortcuts with ⌘ modifier behave like it's regular Dvorak layout.
Now when I hold ⌘ and press key "v" it activates "⌘+k" shortcut.</div>

mitchellh:
Fixes #7315
Fixes #7314 
Fixes #7310 
Fixes #7309

**This PR has breaking changes, noted later.**

This changes our internal key events to match the [W3C key event code specification](https://www.w3.org/TR/uievents-code/). With these changes, apprts are expected to produce key codes within that spec and everything follows from there. This gives us a standard cross-platform keycode behavior.

Previously, due to our history, we based our key codes on GLFW and Linux more generally (since Ghostty was on Linux first). This served us surprisingly well through today, but has been riddled with various issues (many fixed, but note how many bugs this PR is fixing). From a user experience perspective, it also caused confusion about how Ghostty's keybinds interact with keyboard layouts which led to various complicated features like `physical:` (which is removed in this PR).

## Overview of Changes

* `physical:` is now gone. Physical keys are now specified by the W3C key codes. Example: `ctrl+KeyA` will always match the "a" key on a US physical layout (the name `KeyA` lining up with US keyboards is mandated by the spec, not us). Note when we say "physical" here we mean the keycode sent by the OS or GUI framework; these can often be overridden using programs to remap keys at the "hardware" level but software layouts do not do this.

* All single codepoint characters match the character produced by the keyboard layout (i.e. are layout-dependent). So `ctrl+c` matches the physical "c" key on a US standard keyboard with a US layout, but matches the "i" key on a Dvorak layout. This also works for international characters. Codepoints are case-insensitive and match via Unicode case folding (this is how both Windows and macOS treat keyboard shortcuts). 

* W3C key names replace all previous non-character names and always match _physical keys_. For example, `ctrl+grave_accent` (Ghostty 1.1x) is now `ctrl+backquote` to match the W3C spec. This PR maintains backwards compatibility so the old values are aliased to the new values. 

* W3C key names can be both snake case and match the spec exactly. e.g. `KeyA` and `key_a` are both valid. This is an aesthetic choice, because I think the capitalization is really ugly. We have tests to verify this mapping so its officially supported.

* Because we support W3C keys, we now support significantly more media keys such as `context_menu`, `browser_home` and so many more. These work as long as your OS and keyboard can produce the valid key codes.

## Breaking Changes

* Key names have changed, e.g. `grave_accent` is now `backquote`. The new names are all W3C standard. There are backwards compatible aliases, so **old configurations will still work.** But the `+list-keybinds` CLI will always output the new versions.

* `physical:` is gone, all W3C names are always physical and all code points are always logical. There are backwards compatible aliases for unambiguous keys, so **most old configurations will still work.** If there is ambiguity, old configurations will error, but I don't expect this to be common at all.

* Physical keybinds always take priority over logical keybinds. This was _inconsistent_ before. Physical keybinds are relatively rare so I don't think this will impact people but just noting this is now an explicit, tested, documented behavior.


mitchellh:
I think this is good to go. I'd appreciate any further testing here from anyone, I probably won't merge on a Saturday when I don't have a lot of time to follow up on regressions. But expect this to merge soon.

mitchellh:
Note: a couple more changes coming in today:

1. Updated docs on case folding and layouts
2. `ctrl++` doesn't parse. Weird but it should.

PR #7338, opened 2025-05-12
macos: do not send UTF-8 PUA codepoints to key events

### LINKED ISSUE #7337: macOS tip regression: arrow keys don't work in certain KKP scenarios
From #7320

1. Open nvim
2. Try to use any arrow keys

I noticed the Kitty show key kitten works fine so it must have to do with the modes Neovim sets.

mitchellh:
Fixes #7337

AppKit encodes functional keys as PUA codepoints. We don't want to send that down as valid text encoding for a key event because KKP uses that in particular to change the encoding with associated text.

I think there may be a more specific solution to this by only doing this within the KKP encoding part of KeyEncoder but that was filled with edge cases and I didn't want to risk breaking anything else.

### DIFF
--- a/macos/Sources/Ghostty/NSEvent+Extension.swift
+++ b/macos/Sources/Ghostty/NSEvent+Extension.swift
@@ -56,13 +56,20 @@ extension NSEvent {
         // If we have no characters associated with this event we do nothing.
         guard let characters else { return nil }
 
-        // If we have a single control character, then we return the characters
-        // without control pressed. We do this because we handle control character
-        // encoding directly within Ghostty's KeyEncoder.
         if characters.count == 1,
-            let scalar = characters.unicodeScalars.first,
-            scalar.value < 0x20 {
-            return self.characters(byApplyingModifiers: modifierFlags.subtracting(.control))
+           let scalar = characters.unicodeScalars.first {
+            // If we have a single control character, then we return the characters
+            // without control pressed. We do this because we handle control character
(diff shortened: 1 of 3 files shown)


src/terminal/kitty/key.zig
```diff
@@ -83,6 +83,15 @@ pub const Flags = packed struct(u5) {
     report_all: bool = false,
     report_associated: bool = false,
 
+    /// Sets all modes on.
+    pub const @"true": Flags = .{
```
mitchellh, line 87: I didn't end up using this in this PR (I tried for a bit) but this is useful to have so I kept it around.

macos/Sources/Ghostty/NSEvent+Extension.swift
```diff
@@ -56,13 +56,20 @@ extension NSEvent {
...
+            // If we have a single control character, then we return the characters
+            // without control pressed. We do this because we handle control character
+            // encoding directly within Ghostty's KeyEncoder.
+            if scalar.value < 0x20 {
+                return self.characters(byApplyingModifiers: modifierFlags.subtracting(.control))
+            }
+
+            // If we have a single value in the PUA, then its a function key and
+            // we don't want to send PUA ranges down to Ghostty.
+            if scalar.value >= 0xF700 && scalar.value <= 0xF8FF {
```
mitchellh, line 70: This isn't the full PUA range, but this is the range that Apple uses for functional keys.

PR #7355, opened 2025-05-14
url: restrict file paths regex to one slash

ownia:
Fixes https://github.com/ghostty-org/ghostty/discussions/7351.

This restricts the valid path prefixes to prevent false matches caused by literal dot.


mitchellh (approved):
Doesn't change any tests (only adds new tests). Regex change looks good. Approved. Thank you!

PR #7359, opened 2025-05-14
bench: add `--mode=gen-osc` to generate synthetic OSC sequences

mitchellh:
cc @qwerasd205 

This commit adds a few new mode flags to the `bench-stream` program to generator synthetic OSC sequences. The new modes are `gen-osc`, `gen-osc-valid`, and `gen-osc-invalid`. The `gen-osc` mode generates equal parts valid and invalid OSC sequences, while the suffixed variants are for generating only valid or invalid sequences, respectively.

This commit also fixes our build system to actually be able to build the benchmarks. It turns out we were just rebuilding the main Ghostty binary for `-Demit-bench`. And, our benchmarks didn't run under Zig 0.14, which is now fixed.

An important new design I'm working towards in this commit is to split out synthetic data generation to a dedicated package in `src/bench/synth` although I'm tempted to move it to `src/synth` since it may be useful outside of benchmarks.

The synth package is a work-in-progress, but it contains a hint of what's to come. I ultimately want to able to generate all kinds of synthetic data with a lot of knobs to control dimensionality (e.g. in the case of OSC sequences: valid/invalid, length, operation types, etc.).

### DIFF
--- a/src/bench/stream.zig
+++ b/src/bench/stream.zig
@@ -15,6 +15,7 @@ const ArenaAllocator = std.heap.ArenaAllocator;
 const ziglyph = @import("ziglyph");
 const cli = @import("../cli.zig");
 const terminal = @import("../terminal/main.zig");
+const synth = @import("synth/main.zig");
 
 const Args = struct {
     mode: Mode = .noop,
@@ -70,6 +71,14 @@ const Mode = enum {
 
     // Generate an infinite stream of arbitrary random bytes.
     @"gen-rand",
+
+    // Generate an infinite stream of OSC requests. These will be mixed
+    // with valid and invalid OSC requests by default, but the
+    // `-valid` and `-invalid`-suffixed variants can be used to get only
+    // a specific type of OSC request.
+    @"gen-osc",
+    @"gen-osc-valid",
+    @"gen-osc-invalid",
 };
 
 pub const std_options: std.Options = .{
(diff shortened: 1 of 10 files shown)


src/bench/stream.zig
```diff
@@ -100,6 +109,9 @@ pub fn main() !void {
         .@"gen-ascii" => try genAscii(writer, seed),
         .@"gen-utf8" => try genUtf8(writer, seed),
         .@"gen-rand" => try genRand(writer, seed),
+        .@"gen-osc" => try genOsc(writer, seed, 0.5),
```
mitchellh, line 112: Note `0.5` is a pretty unrealistic workload, but it gives us a ballpark of "average" performance if this were fed into an OSC parser. I think realistically OSC sequences are at least 90% valid (but likely closer to 99%). If we were benchmarking, I think it'd be cleaner to separate out 100% valid and 100% invalid sources rather than use this mixed version.

mitchellh:
Going to merge this to mainly just fix our benchmarks. I want to work on the synthetic data library more though.

qwerasd205:
I have some code locally which randomly walks the VT100 state machine to generate random output-- I can clean it up a bit and contribute it to the synthetic data generation if you'd like

mitchellh:
I’d love that. I’ve got a much cleaned up interface locally I’ll PR next week probably so hold off until then!

PR #7367, opened 2025-05-15
fix(renderer): Don't force images to grid/cell sizes.

qwerasd205:
Also fixed Metal not interpolating scaled images.

See commit message for more details.

|`main` (Metal)|`main` (OpenGL)|Kitty|fixed (Metal)|fixed (OpenGL)|
|-|-|-|-|-|
|image|image|image|image|image|

<sup>(Zoom in for details.)</sup>

> [!NOTE]
> This comparison reveals a separate problem we have with image scaling, which Kitty gets right but we don't -- the interpolation is too dark because of gamma error, we need to be interpolating in linear space but instead we're interpolating in gamma compressed space. I'll try to figure out the best way to resolve this.


qwerasd205:
Haha oops forgot to run tests and typo check locally before PRing

mitchellh (approved):

PR #7395, opened 2025-05-19
nix: don't strip if we are building a debug version of the package

jcollie:


qwerasd205:
The condition is backwards, ***dont***Strip should be *true* if we *are* building `Debug`

jcollie:
> The condition is backwards, _**dont**_Strip should be _true_ if we _are_ building `Debug`

Doh! Fixed. Also added some CI checks to make sure that debug symbols actually make it into the package.

mitchellh (approved):

PR #7409, opened 2025-05-21
synthetic package

mitchellh:
This introduces a new package `src/synthetic` for generating synthetic data, currently primarily for benchmarking but other use cases can emerge.

The synthetic package exports a runtime-dispatched type `Generator` that can generate data of various types. To start, we have a bytes, utf8, and OSC generator. The goal of each generator is to expose knobs to tune the probabilities of various outcomes. For example, the UTF-8 generator has a knob to tune the probability of generating 1, 2, 3, or 4-byte UTF-8 sequences.

Ultimately, the goal is to be able to collect probability data empirically that we can then use for benchmarks so we can optimize various parts of the codebase on real-world data shape distributions.


mitchellh:
The test failures look like some temporary Nix installation issue... 

Separately, none of this code is currently user-facing.

qwerasd205 (approved):
Structure looks pretty good, no notes really- seems like a good foundation to build up a robust set of generators.

PR #7413, opened 2025-05-21
ci: switch to Determinate Nix installer

mitchellh:
In testing, it's slower but generally more reliable (the installer, not Nix). I've emailed Determinate with my concerns about speed and I hope they can address those. We are still using upstream Nix (not Determinate Nix), this just changes our installer.


mitchellh:
Eh. I still want to do this but this doesn't fix our immediate CI issues, so I'm going to delay this to later...

PR #7430, opened 2025-05-23
add cut/copy/paste keys

Mic92:
The origin of these keys are old sun keyboards.
They are getting picked up by the custom (progammable) keyboard scene (see https://github.com/manna-harbour/miryoku for a popular layout). Support in ghosty is quite handy because it allows to bind copy/paste in a way that doesn't overlap with ctrl-c/ctrl-v, which can have special bindings in some terminal applications.


00-kat (approved):

jcollie:
I'm not an expert in this area, but I don't see anything obviously wrong with this. Except that I don't have a keyboard with cut/copy/paste keys on it.

Mic92:
Support for them is also fairly widespread. It works in qt/gtk, browsers, electron...

mitchellh (approved):
This is correctly done. Thank you.

PR #7433, opened 2025-05-24
linux: add dbus and systemd activation services

jcollie:


src/apprt/gtk/App.zig
```diff
@@ -397,10 +397,12 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
...
     // routine so we just call it, but only if the config allows it (this allows
     // for launching Ghostty in the "background" without immediately opening
-    // a window)
+    // a window). An initial window will not be immediately created if we were
+    // launched by D-Bus activation or systemd.  D-Bus activation will send it's
+    // own `activate` or `new-window` signal later.
     //
     // https://gitlab.gnome.org/GNOME/glib/-/blob/bd2ccc2f69ecfd78ca3f34ab59e42e2b462bad65/gio/gapplication.c#L2302
-    if (config.@"initial-window")
+    if (config.@"initial-window" and !internal_os.launchedByDBusActivation() and !internal_os.launchedBySystemd())
```
alaviss, line 405: This feels like an overcomplication, and also make testing Ghostty server mode more difficult.

Shouldn't there be a flag or command that only setup the application without spawning windows, then use it within the systemd/dbus service instead?

rhodes-b: this is that command 
`ghostty --initial-window=false --quit-after-last-window-closed` is how its currently done

I dont know much about dbus part but you can certainly pass that command into systemd service file

alaviss: Sounds like the perfect case to have flags/commands that are not controlled by the config, but I don't think there are any precedents for those.

alaviss:
As a general note, on Linux an user may change the application id by setting the `class` configuration variable. This meant that the D-Bus name of Ghostty would then differs from the well-known name specified by the desktop file and D-Bus service, which would break the feature.

jcollie:
> As a general note, on Linux an user may change the application id by setting the `class` configuration variable. This meant that the D-Bus name of Ghostty would then differs from the well-known name specified by the desktop file and D-Bus service, which would break the feature.

TBF, the user can do _a lot_ of things to break stuff. Probably the only way to deal with this is to add some documentation to the `class` config that tells users what might happen.

mitchellh:
I don't have time to review in depth but I'm also starting to wonder if we should just have a `launched-from` configuration to have the flag and then derive default behaviors from that. I agree the logic here is getting a little messy. But moreso, the performance cost to detect this stuff is getting a little nasty too.

nix/package.nix
```diff
@@ -117,6 +118,11 @@ in
 
       mkdir -p "$out/nix-support"
 
+      sed -i -e "s@^Exec=.*ghostty@Exec=$out/bin/ghostty@" $out/share/applications/com.mitchellh.ghostty.desktop
+      sed -i -e "s@^TryExec=.*ghostty@TryExec=$out/bin/ghostty@" $out/share/applications/com.mitchellh.ghostty.desktop
+      sed -i -e "s@^Exec=.*ghostty@Exec=$out/bin/ghostty@" $out/share/dbus-1/services/com.mitchellh.ghostty.service
+      sed -i -e "s@^ExecStart=.*ghostty@ExecStart=$out/bin/ghostty@" $out/lib/systemd/user/com.mitchellh.ghostty.service
+
```
pluiedev, line 125: IIRC it's recommended that service files don't refer to the binary via an absolute path, since after a rebuild it's possible that the cached service could point to a GC'd and nonexistent store path

pluiedev (changes requested):

nix/package.nix
jcollie: For "normal" service files on a non-Nix system yes, but `nix/package.nix` is modifying the services files in the _installed NixOS package_. Just looking at my system with `grep ExecStart -- /etc/systemd/system/*` it's actually the other way around. It's vastly more common for a Nix service file to have an absolute path to the executable in `/nix/store` rather than relying on the executable being in the PATH.

The service files that are copied to `zig-out/share` rely on the binary being in the path.

Also, if the binary referred to by a service is GC'd, the service file itself would be GC'd and Nix would tell systemd to reload all of the service definitions.

alaviss:
> you can't specify a CLI flag if you're launching Ghostty from the dock or even just double-clicking the app.

I think this can be solved by adding a wrapper shell script that adds the CLI flag and have the `.app` bundle launches it instead of ghostty binary directly.

jcollie:
> > you can't specify a CLI flag if you're launching Ghostty from the dock or even just double-clicking the app.
> 
> I think this can be solved by adding a wrapper shell script that adds the CLI flag and have the `.app` bundle launches it instead of ghostty binary directly.

That does not strike me as a very robust solution. I think what might help is breaking some of this logic out into a separate function rather than a giant boolean expression in a if statement. I don't think the performance of the detection matters a whole lot since it's done once at startup, and if that information is needed outside of startup it should be done once at startup and saved in a global variable for future use.

alaviss:
I personally am not concerned about performance here, but code complexity and the use of heuristics. It would be more robust to have the execution site tell us directly what kind of behavior do they want than trying to guess what it is.

As a plus, if there's a switch, then it's trivial to perform manual testing of these behaviors.

pluiedev:
Adding a manual flag would be nice I think, although I don't think the heuristics should entirely go away - for instance, someone can write a systemd service on their own without ever specifying via CLI that Ghostty should consider itself launched via systemd. This obviously wouldn't be an issue for the *official* systemd service of course, but it is something that could happen (especially when one uses a window manager that expects more user intervention and customization than normal desktop environments).

Instead, we should probably respect a `launched-from` flag of some sort, and if one isn't present we attempt to find out using heuristics just how the app was launched. That way we get the best of both worlds.

I do think the heuristics need a LOT more links pointing to official D-Bus/systemd documentation as to the veracity and reliability of these environment variables, though... ;)

pluiedev (changes requested):

src/os/dbus.zig
```diff
@@ -0,0 +1,21 @@
...
+
+        // On Linux, D-Bus activation sets `DBUS_STARTER_ADDRESS` and
+        // `DBUS_STARTER_BUS_TYPE`. If these environment variables are present
+        // (no matter the value) we were launched by D-Bus activation.
+        .linux => std.posix.getenv("DBUS_STARTER_ADDRESS") != null and std.posix.getenv("DBUS_STARTER_BUS_TYPE") != null,
+
+        // No other system supports D-Bus so always return false.
+        else => false,
+    };
+}
```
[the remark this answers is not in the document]
mitchellh: I'm fine either way. A lot of OS-specific logic is in `src/os` (that's the point of the package), but also because it tends to be used by libghostty (e.g. `launchedFromDesktop` is). But if something isn't used by libghostty yet I'm fine keeping it apprt-specific for now and extracting later.

src/os/systemd.zig
```diff
@@ -0,0 +1,64 @@
...
+            // If `INVOCATION_ID` and `JOURNAL_STREAM` are present, check to make sure
+            // that our parent process is actually `systemd`, not some other terminal
+            // emulator that doesn't clean up those environment variables.
+
+            const ppid = std.os.linux.getppid();
+
+            // If the parent PID is 1 we'll assume that it's `systemd` as other init systems
+            // are unlikely.
+
+            if (ppid == 1) break :linux true;
```
mitchellh, line 32: This seems like quite an assumption. This is true _today_ because most distros use systemd, but historically any init system would launch children from pid 1 no?

mitchellh (changes requested):
> The main problem I think would be macOS - AFAIK you can't specify a CLI flag if you're launching Ghostty from the dock or even just double-clicking the app.

The main entry point is different on macOS and we already use it to specify where it was launched from: https://github.com/ghostty-org/ghostty/blob/1ff91625981ef92add92624c01b6b9095c365ed1/macos/Sources/App/macOS/main.swift This could be modified easily.

I think what @pluiedev said is probably best: introduce the config option, and change the config defaults to auto-detect. This already only happens once (per config reload which is rare) and we get caching this way. 

That approach would get rid of the performance issue, make testing easier because we can simulate different launch environments, centralize detection in one place, we can probably more easily unit test as well instead of littering it throughout the codebase, and it lets users have an escape hatch to manually override our behavior if we detect incorrectly. 

I looked at where we call these functions and in most cases we easily have a `config` available. The one place we don't is the locale loading for macOS but we can add a special case there I think because we have to load the locale before configuration is loaded necessarily for error messages at least. 

I can try to extract that myself in a separate PR so we don't confuse it with the dbus/systemd stuff in here.

src/build/GhosttyResources.zig
```diff
@@ -196,11 +196,27 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyResources {
     if (cfg.target.result.os.tag == .linux) {
         // https://developer.gnome.org/documentation/guidelines/maintainer/integrating.html
 
-        // Desktop file so that we have an icon and other metadata
-        try steps.append(&b.addInstallFile(
-            b.path("dist/linux/app.desktop"),
-            "share/applications/com.mitchellh.ghostty.desktop",
-        ).step);
+        if (b.option(bool, "dbus-activation", "Enable DBus activation and systemd user service.") orelse false) {
```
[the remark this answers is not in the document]
pluiedev: Then it would error out I think. I still don't understand why we don't just unconditionally include them though - `Exec` still serves as a backup for systems that don't support D-Bus activation: [(source)](https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html)

> Applications should still include Exec= lines in their desktop files for compatibility with implementations that do not understand the DBusActivatable key.

src/os/systemd.zig
mitchellh: Nevermind, I see the guards before this that verify systemd via env vars.

src/os/systemd.zig
```diff
@@ -0,0 +1,64 @@
...
+                log.err("unable to format comm path for pid {d}", .{ppid});
+                break :linux false;
+            };
+            const comm_file = std.fs.openFileAbsolute(comm_path, .{ .mode = .read_only }) catch {
+                log.err("unable to open '{s}' for reading", .{comm_path});
+                break :linux false;
+            };
+            defer comm_file.close();
+
+            var comm_data_buf: [std.fs.MAX_NAME_BYTES]u8 = undefined;
```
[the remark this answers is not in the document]
mitchellh: I've got a cleanup for this in my branch, but we can hardcode this to 16 because even if it is ever increased, if the result is longer than 16 it can't possibly be "systemd" so its fine.

mitchellh:
All reviews, esp @jcollie, see #7503. You'll have to rebase your changes onto that, it just focuses on the `launched-from` flag.

src/apprt/gtk/App.zig
jcollie: Latest patch is now rebased on top of #7503

src/build/GhosttyResources.zig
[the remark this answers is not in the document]
jcollie: Latest version removed the build option and just always installs the DBus-enabled config files.

src/os/systemd.zig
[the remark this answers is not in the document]
jcollie: Fixed by #7503

jcollie:
Patch has been rebased on top of #7503

tristan957:
Do the title of the PR and commit need to change? Also, will there be any accompanying documentation about this? I'm a pretty experienced Linux person, but I don't have any idea what I could do with this.

dist/linux/app.desktop
```diff
@@ -1,8 +1,10 @@
 [Desktop Entry]
+Version=1.0
 Name=Ghostty
 Type=Application
 Comment=A terminal emulator
-Exec=ghostty
+TryExec=ghostty
+Exec=ghostty %F
```
alaviss: Why `%F`? Does Ghostty takes input files?

Also, can you add `--launched-from` for this? Heuristics are useless behind some layers like Flatpak, so it's better to avoid them if possible.

dist/linux/dbus.service
```diff
@@ -0,0 +1,4 @@
+[D-BUS Service]
+Name=com.mitchellh.ghostty
+SystemdService=com.mitchellh.ghostty.service
+Exec=ghostty
```
alaviss: Similar notes about `--launched-from`

dist/linux/systemd.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Ghostty
+
+[Service]
+Type=dbus
+BusName=com.mitchellh.ghostty
+ExecStart=ghostty
```
alaviss: If you want the service to be able to be enabled on startup, you should add an `[Install]` directive, ideally with an `Alias` to the actual bus name.

dist/linux/dbus.service
```diff
@@ -0,0 +1,4 @@
+[D-BUS Service]
+Name=com.mitchellh.ghostty
+SystemdService=com.mitchellh.ghostty.service
```
alaviss, line 3: You might want to name this `dbus-com.mitchellh.ghostty.service`, and use `Alias` within the systemd file to decide how this would be enabled. This is [recommended by the D-Bus specification](https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-starting-services-systemd)

src/apprt/gtk/App.zig
```diff
@@ -400,11 +400,15 @@ pub fn init(core_app: *CoreApp, opts: Options) !App {
...
+    // a window). An initial window will not be immediately created if we were
+    // launched by D-Bus activation or systemd.  D-Bus activation will send it's
+    // own `activate` or `new-window` signal later.
     //
     // https://gitlab.gnome.org/GNOME/glib/-/blob/bd2ccc2f69ecfd78ca3f34ab59e42e2b462bad65/gio/gapplication.c#L2302
-    if (config.@"initial-window")
-        gio_app.activate();
+    if (config.@"initial-window") switch (config.@"launched-from".?) {
+        .dbus, .systemd => {},
+        else => gio_app.activate(),
```
mitchellh: Let's avoid `else` here and use an exhaustive enum. In case we add any new cases in the future I _want_ the compiler to tell me all the cases where I should think a bit about what I want the logic to be.

mitchellh (changes requested):
One simple change requested. Looking good!

dist/linux/app.desktop
```diff
@@ -1,8 +1,10 @@
 [Desktop Entry]
+Version=1.0
 Name=Ghostty
 Type=Application
 Comment=A terminal emulator
-Exec=ghostty
+TryExec=ghostty
```
alaviss, line 6: According to the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html), this key is used to detect whether the application is installed so as to either show or autostart it.

For Ghostty the only way this file made it to the user system is if it's installed, so this key doesn't do much.

alaviss (commented):
Some notes about the desktop and service files

dist/linux/app.desktop
jcollie: The only way the `.desktop` file makes sense is if it's installed somewhere so presumably the user has installed the Ghostty binary somewhere too.

dist/linux/dbus.service
jcollie: Done.

dist/linux/systemd.service
jcollie: I tried this out but it broke launching Ghostty. I've found that the name of the service is _very_ sensitive and keeping it as `com.mitchellh.ghostty` is working just fine. If you can elaborate on what functionality would be gained by adding a `dbus-` prefix I might look at it again and try and figure out all the quirks.

dist/linux/dbus.service
jcollie: See above.

dist/linux/app.desktop
jcollie: Done.

src/apprt/gtk/App.zig
jcollie: Fixed.

dist/linux/app.desktop
alaviss: That's what I'm saying, this key does nothing useful and can be removed.

dist/linux/systemd.service
alaviss: > On the well-known system bus, it is a common practice to set `SystemdService` to `dbus-`, followed by the well-known bus name, followed by `.service`, then register that name as an alias for the real systemd service. This allows D-Bus activation of a service to be enabled or disabled independently of whether the service is started by systemd during boot.

From [the D-Bus specification](https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-starting-services-systemd).

dist/linux/app.desktop
```diff
@@ -16,6 +18,7 @@ X-TerminalArgTitle=--title=
 X-TerminalArgAppId=--class=
 X-TerminalArgDir=--working-directory=
 X-TerminalArgHold=--wait-after-command
+DBusActivatable=true
 
 [Desktop Action new-window]
 Name=New Window
```
alaviss, line 24: You might want to pass `--launched-from` for `new-window` action, too.

jcollie:
> Also, will there be any accompanying documentation about this? I'm a pretty experienced Linux person, but I don't have any idea what I could do with this.

It should "just work" if all the files get installed in the right place, as I would hope that they would when Ghostty is installed via a package. Not sure what we should document as I don't think that we need to document how `.desktop` and
the other `.service` files work - we don't document what users should do with the current `.desktop` file as that's more of a packager concern.

As far as the title of the PR is concerned I think removing the word `option` might be a good idea.

alaviss:
As noted in https://github.com/ghostty-org/ghostty/pull/7433#issuecomment-2908049788, this might break when `class` is configured to a non-default value.

I think `class` documentation should be updated to reflect this change.

dist/linux/systemd.service
jcollie: Yes, I saw that in the docs, tried it, and launching broke so there must be some nuance that I missed. I'm still not sure _why_ we need that as the presence of systemd pretty much guarantees the presence of DBus. Are there systems that use systemd and DBus but _do not_ use DBus activation?

dist/linux/app.desktop
jcollie: Done.

jcollie:
Added some documentation to the `class` config that explains how changing it might affect launching.

dist/linux/systemd.service
alaviss: You'd have to `enable` the service from systemd side for it to be picked up (i.e. `systemctl --user enable ghostty.service`), which would then install the alias names.

But if I understand you correctly, then D-Bus activation just breaks when the linked `SystemdService` could not be found? I'm not sure if it should be used in this case then.

At the very least we would have to patch that out for Flatpak, since user service files are not exported.

alaviss: Yea I just tested the flatpak, it wouldn't launch with `SystemdService` in the D-Bus service.

I think either we patch this out with a `sed '/SystemdService=/d'` for Flatpak or just not have them be linked at all, but I can do the patching in a follow up, since `.Devel` builds would have to be fixed as the desktop name doesn't match the bus name.

mitchellh (approved):

mitchellh:
@pluiedev's changes were resolved so I'm going to merge this. Thanks 😄

PR #7434, opened 2025-05-24
PageList.zig: Allow a Pin's x to be a column or negative to allow full line select when mouse is dragging left of the terminal

adlr:
I looked at a few ways to solve this problem. The core idea is that PageList.zig:Pin needs to be able to represent being off the left side of the terminal. Historically that's been impossible as the x coordinate is a u16 and an x value of 0 in the selection end means the character at column 0 is included in the selection.

Alternatives considered:
- Change Pin.x to signed int. A value in the 40,000 range was required to fit into the x value somewhere in the code base, so that would mean i32 could be a decent type to choose. However, this led to many casts when converting to/from or comparing to u16.
- Shift the meaning of Pin.x by 1, so that 0 means off the left side, 1 means the first column is included, etc. This results in a lot of "+ 1" being added to the code, which seems error prone.

The approach taken was to change x to a tagged union, either a column which is u16 as before, with (mostly) all the same behavior as before, or it's a void type indicating it's negative, meaning off to the left.

The vast majority of the change is mechanically changing code to handle this type change without having a behavior change, other than the intended one.

Fixes https://github.com/ghostty-org/ghostty/discussions/5058

Feedback welcome! Thank you to the amazing devs here!

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3583,7 +3583,7 @@ pub fn cursorPosCallback(
 
         // Convert to points
         const screen = &self.renderer_state.terminal.screen;
-        const pin = screen.pages.pin(.{
+        var pin = screen.pages.pin(.{
             .viewport = .{
                 .x = pos_vp.x,
                 .y = pos_vp.y,
@@ -3592,6 +3592,9 @@ pub fn cursorPosCallback(
             if (comptime std.debug.runtime_safety) unreachable;
             return;
         };
+        if (pos.x <= @as(f32, @floatFromInt(self.size.padding.left))) {
+            pin.x = .neg;
+        }
 
         // Handle dragging depending on click count
         switch (self.mouse.left_click_count) {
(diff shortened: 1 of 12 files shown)


qwerasd205:
This doesn't feel like the right solution to me. You're making changes to a data structure that is used pervasively throughout the codebase - including several places you missed, and in performance-critical areas of the code - in order to solve a problem that originates only in `Surface.zig`.

A much better solution would be to modify the drag logic in `Surface.zig` to take an `apprt.CursorPos` (or, really, I would introduce a new struct which provides the position in fractional cells) and add the necessary logic to set the selection end at the end of the previous row if the position is off the left of the screen. Making this change would also allow the selection logic to be generally improved by considering whether the cursor is on the left or right half of a cell- which coincidentally would remove the need for that special case logic since it would just work as expected with clamped coordinates if this was done.

adlr:
Thanks for the reviews, folks. I felt like I was picking between not great
options. Next weekend I'll take another look with your advice and hopefully
find something a bit more targeted.

qwerasd205:
I've reworked the selection logic in more or less the way I recommended above, which fixes the relevant issue and generally cleans it up. I'll be adding unit tests and then making a PR later tonight after I have dinner.

Thank you for taking the initiative to fix this, even if your approach wasn't ideal-- it brought it to my attention and made me think of the right way to solve it, so it was valuable in that way.

I'm gonna go ahead and close this PR now, thank you!

Issue #7386, opened 2025-05-18
vt: support buffer switching `CSI ? 47 h/l`


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7377

<div type='discussions-op-text'>

<sup>Originally posted by **TheCommieAxolotl** May 17, 2025</sup>
Add support for xterm's [`CSI ? 47 h` and  `CSI ? 47 l`](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Control-Bytes_-Characters_-and-Sequences) (`Ps = 4 7`).</div>

PR #7443, opened 2025-05-25
Add support for buffer switching with CSI ? 47 h/l

jborkowski:
This PR adds support for handling the escape sequences CSI ? 47 h/l, which are related to alternate screen buffer switching. This is in response to the issue reported in [ghostty-org/ghostty#7386](https://github.com/ghostty-org/ghostty/issues/7386).

I'm unsure where to add tests for this change. Would it make sense to add visual tests for this behavior, or is there a preferred approach or location for testing such functionality? I tested it locally using the following commands:
```bash
echo -e "\e[?47h"
echo "Printed on the alt screen"
echo -e "\e[?47l"
```

### DIFF
--- a/src/terminal/modes.zig
+++ b/src/terminal/modes.zig
@@ -206,6 +206,7 @@ const entries: []const ModeEntry = &.{
     .{ .name = "cursor_visible", .value = 25, .default = true },
     .{ .name = "enable_mode_3", .value = 40 },
     .{ .name = "reverse_wrap", .value = 45 },
+    .{ .name = "alt_screen_legacy", .value = 47 },
     .{ .name = "keypad_keys", .value = 66 },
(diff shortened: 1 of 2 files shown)


mitchellh (commented):
The code here looks great, right. I need to just verify that the alternate screen behavior (just clear on enter) is correct for this.

mitchellh:
Based on the above code and verified with the shell script below, mode 47 _does not_ clear on enter. It simply switches to alternate screen and clears nothing, moves no cursors, etc. Likewise on exit.

```sh
#!/usr/bin/env bash

function csi() {
  echo -ne "\033[$1"
}

csi "?47h"
echo hello
read
csi "?47l"
read
csi "?47h"
echo world
read
csi "?47l"
```

PR #7444, opened 2025-05-26
Rework mouse selection logic

qwerasd205:
This PR fixes the problem discussed in #5058 and #7434 by reworking the selection logic in a way that better handles edge cases as well as being generally cleaner.

This rework does change how selection behaves slightly, especially rectangular selection, but the new behavior of rectangular selection is more in line with other terminals I tested (Terminal.app, Kitty).

There are some TODO comments for adding unit tests- I ran out of steam tonight, but if this PR is still open tomorrow I'll go ahead and add them.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3676,11 +3676,29 @@ fn dragLeftClickTriple(
 fn dragLeftClickSingle(
     self: *Surface,
     drag_pin: terminal.Pin,
-    xpos: f64,
+    drag_x: f64,
 ) !void {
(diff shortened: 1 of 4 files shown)


qwerasd205:
Actually, I'm refactoring those tests I just added to use a helper function so that there's not a ton of repeated code and it's easier to understand them.

mitchellh (approved):
Awesome, thanks. Surface unit tests are excellent. The other TODOs I'm willing to defer.

PR #7451, opened 2025-05-27
Miscellaneous `TODO`s

qwerasd205:
See commit messages for details.

If some of the commits in this PR have problems with them I'm perfectly fine with the others being cherry-picked out while the problems are addressed.

The biggest/broadest reaching changes in this PR come from converting a lot of code to use decl literals where possible, so there are a lot of files where only a handful of lines are modified very slightly.

### DIFF
--- a/pkg/fontconfig/pattern.zig
+++ b/pkg/fontconfig/pattern.zig
@@ -44,7 +44,7 @@ pub const Pattern = opaque {
             &val,
         ))).toError();
 
-        return Value.init(&val);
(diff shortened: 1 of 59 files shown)


pluiedev (approved):
ACK GTK diffs

mitchellh (approved):
Fantastic. Simply out of curiosity, would love a micro bench before/after on the `rotateToZero` circ buf change. (I can do this myself later).

PR #7461, opened 2025-05-28
code style: use `@splat` where possible

qwerasd205:
As of Zig 0.14.0, `@splat` can be used for array types, which eliminates a lot of redundant syntax and makes things generally cleaner.

I've explicitly avoided applying this change in the renderer files for now since it would just create rebasing conflicts in my renderer rework branch which I'll be PR-ing pretty soon.


jcollie (approved):

mitchellh (approved):
Didn't realize this, love this.

PR #7468, opened 2025-05-29
GTK: add action to show the GTK inspector

jcollie:
The default keybinds for showing the GTK inspector (`ctrl+shift+i` and `ctrl+shift+d`) don't work reliably in Ghostty due to the way Ghostty handles input. You can show the GTK inspector by setting the environment variable `GTK_DEBUG` to `interactive` before starting Ghostty but that's not always convenient.

This adds a keybind action that will show the GTK inspector. Due to API limitations toggling the GTK inspector using the keybind action is impractical because GTK does not provide a convenient API to determine if the GTK inspector is already showing. Thus we limit ourselves to strictly showing the GTK inspector. To close the GTK inspector the user must click the close button on the GTK inspector window. If the GTK inspector window is already visible but is hidden, calling the keybind action will not bring the GTK inspector window to the front.


tristan957 (commented):
Conceptually it seems weird to have this exposed in libghostty. Very toolkit specific

jcollie:
> Conceptually it seems weird to have this exposed in libghostty. Very toolkit specific

True, but AFAIK there isn't a way to do a keybind action that isn't in libghostty so there isn't really a way to do that and have the keybind configurable from the Ghostty config. I think that even using GTK's "normal" keyboard shortcuts would run into the same problem as the standard keyboard shortcuts have.

mitchellh:
> AFAIK there's no way to change the keyboard shortcuts that GTK uses for the GTK inspector (aside from what I've done here).

Okay, that's a very reasonable argument to expose this.

> Being able to pass ctrl+shift+i or ctrl+shift+d through to the GTK UI for processing might be possible

This should happen automatically if in our key handler we don't mark that it was handled (in GTK). We already have all the machinery for this, but I think Ghostty core reports `ignore` as a handled key potentially... I'm not sure I'd have to take a look. 

Note the default we chose for the inspector is also because it's the default that web browsers use for the inspector so I'd like to keep it. They don't use GTK so its not an issue but... just noting.

tristan957:
FWIW, I think Ghostty inspector should continue to live on the keybind. I, as a ghostty dev, will just unbind it for myself.

> I wonder if a better solution is to make the ignore binding action work here, if it doesn't already

I tried this suggestion. The inspector still didn't open, but the keybind is no longer encoded into the shell

mitchellh (approved):

PR #7474, opened 2025-05-29
input: "ignore" binding action are still be processed by the OS/GUI

mitchellh:
Related to #7468

This changes the behavior of "ignore". Previously, Ghostty would consider "ignore" actions consumed but do nothing. They were like a black hole. Now, Ghostty returns `ignored` which lets the apprt forward the event to the OS/GUI.

This enables keys that would otherwise be pty-encoded to be processed later, such as for GTK to show the GTK inspector.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2069,12 +2069,18 @@ fn maybeHandleBinding(
         break :performed try self.performBindingAction(action);
     };
 
-    // If we performed an action and it was a closing action,
-    // our "self" pointer is not safe to use anymore so we need to
-    // just exit immediately.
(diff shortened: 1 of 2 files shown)


mitchellh:
@tristan957 Can you try this branch with `ctrl+shift+i=ignore` and see if the GTK inspector shows up?

tristan957:
Indeed it does work.

tristan957 (approved):

pluiedev (approved):

mitchellh:
Cool. I don’t think this should eliminate #7468 but I think in general this is a good core behavior.

PR #7477, opened 2025-05-30
gtk(wayland): customize keyboard interactivity for quick terminal

### LINKED ISSUE #7476: GTK: configurable Wayland keyboard interactivity mode for quick terminal
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7475

<sup>Originally posted by **alex-huff** May 30, 2025</sup>
On some wayland compositors, for example sway, spawning a panel with `GTK_LAYER_SHELL_KEYBOARD_MODE_ON_DEMAND` doesn't automatically focus the panel. This means the user has to manually focus the panel using a mouse which is not very quick. Maybe there could be an option that allows the user to use `GTK_LAYER_SHELL_KEYBOARD_MODE_EXCLUSIVE` for the keyboard mode?</div>

<hr>

**Maintainer's note:** See https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:keyboard_interactivity for more information.

pluiedev:
Fixes #7476

### DIFF
--- a/src/apprt/gtk/winproto/wayland.zig
+++ b/src/apprt/gtk/winproto/wayland.zig
@@ -6,8 +6,8 @@ const build_options = @import("build_options");
 const gdk = @import("gdk");
 const gdk_wayland = @import("gdk_wayland");
 const gobject = @import("gobject");
-const gtk4_layer_shell = @import("gtk4-layer-shell");
 const gtk = @import("gtk");
+const layer_shell = @import("gtk4-layer-shell");
 const wayland = @import("wayland");
 
 const Config = @import("../../../config.zig").Config;
@@ -98,7 +98,7 @@ pub const App = struct {
     }
 
     pub fn supportsQuickTerminal(_: App) bool {
-        if (!gtk4_layer_shell.isSupported()) {
+        if (!layer_shell.isSupported()) {
             log.warn("your compositor does not support the wlr-layer-shell protocol; disabling quick terminal", .{});
             return false;
         }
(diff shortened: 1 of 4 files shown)


src/apprt/gtk/winproto/wayland.zig
```diff
@@ -356,53 +356,66 @@ pub const Window = struct {
 
     fn syncQuickTerminal(self: *Window) !void {
         const window = self.apprt_window.window.as(gtk.Window);
-        const position = self.apprt_window.config.quick_terminal_position;
+        const config = self.apprt_window.config;
```
mitchellh, line 359: Not sure if you want a `&` here to make it a pointer or if you actually want to copy the full config structure.

mitchellh (approved):
Looks good to me. It's a bit of an odd configuration to expose since I'm not sure I see a use case for setting the other values but I'm fine with it, can always reassess later. More configurability is better than less to keep folks happy, especially if its minimal impact like this.

I made one comment to look at but no need for it to block this merge.

pluiedev:
AFAICT on demand mode has some *strange* behavior on wlroots-based compositors like Sway, and so for them exclusive would work a bit better. It's been a while since I've used it, though.

FWIW in my experience both Niri and KWin exhibit what is "correct" in my experience - `none` prevents the QT from intercepting key events, `on-demand` makes it only do so when focused, and `exclusive` makes it steal inputs at all times. However if you read the actual spec you'll see a lot of things are just kind of handwaved away, and that's why I want users to be able to experiment with what feels better on their system.

PR #7482, opened 2025-05-30
gtk: request user attention on bell

pluiedev:
I'm not sure if this should be enabled by default like the tab animation, but on KDE at least this is unintrusive enough for me to always enable by default. Alacritty appears to agree with me as well.

Fixes #7124

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -2447,6 +2447,11 @@ pub fn ringBell(self: *Surface) !void {
         // Need attention if we're not the currently selected tab
         if (page.getSelected() == 0) page.setNeedsAttention(@intFromBool(true));
     }
+
+    // Request user attention
+    window.winproto.setUrgent(true) catch |err| {
+        log.err("failed to request user attention={}", .{err});
+    };
 }
 
 /// Handle a stream that is in an error state.
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -814,11 +814,15 @@ fn gtkWindowNotifyIsActive(
     _: *gobject.ParamSpec,
     self: *Window,
 ) callconv(.c) void {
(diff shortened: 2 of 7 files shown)


jcollie (approved):
LGTM

mitchellh (approved):
LGTM. 

On macOS we do this and there has been a request to make this configurable. Now that both platforms support it seems like we should follow this up at some point with a new packed structure entry on the bell features to turn this off.

mitchellh:
On macOS, the dock icon simply bounces (once, not continuously).

pluiedev:
On KDE the icon on the taskbar simply turns orange (indicating that focus has been requested). I suspect GNOME is being too lax with XDG activation requests - normally you're supposed to block activation requests from the app to itself since that would count as focus-stealing, but I guess GNOME just lets that slip past...

tristan957:
I'll think about bringing it up with the GNOME Shell folks.

jcollie:
> I'm betting GNOME 48 is probably the same behavior.

On Gnome 48 I get a notification, the window doesn't move to the top until I click the notification.

pbsds:
On gnome the behavior is dependent on what `/org/gnome/desktop/wm/preferences/focus-new-windows` is set to in dconf. It is usually `smart` or `strict`. Gnome is transitioning from one to the other as default, with the goal of in the future denying applications from grabbing focus.

I have yet to test much with Gnome 48, but on 47 the focus grabbing behavior is prone to only be visual. It is also buggy and prone to only move he visual focus but not which window receives the keyboard inputs, risking the user trying to type into the terminal but instead writing in a background chat application.

You should test both modes

pluiedev:
Ouch, that doesn't seem ideal. KDE has already switched over to being "strict" by default a while ago and now I worry that GNOME users with old configs might get a bad default here... a toggle for this feature is definitely needed then

PR #7483, opened 2025-05-30
font: rework coretext discovery sorting

qwerasd205:
This should make the sorting more robust to fonts with questionable metadata or atypical style names.

I was originally just going to change the scoring slightly to account for fonts whose regular italic style is named "Regular Italic" - which previously resulted in the Bold Italic or Thin Italic style being chosen instead because they're shorter names, but I decided to do some better inspection of the metadata and looser style name matching while I was changing code here anyway.

Also adds a unit test to verify the sorting works correctly, though a more comprehensive set of tests may be desirable in the future.

(Feel free to make any changes to the PR you feel necessary before merging while I'm gone over the weekend.)

### DIFF
--- a/src/font/discovery.zig
+++ b/src/font/discovery.zig
@@ -4,6 +4,7 @@ const Allocator = std.mem.Allocator;
 const assert = std.debug.assert;
 const fontconfig = @import("fontconfig");
 const macos = @import("macos");
+const opentype = @import("opentype.zig");
 const options = @import("main.zig").options;
 const Collection = @import("main.zig").Collection;
 const DeferredFace = @import("main.zig").DeferredFace;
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Changing these can be scary but we should be able to just rely on reports from the community and follow up with unit tests as we find them.

mitchellh:
CI issues

qwerasd205:
Ah oops, was my assumption bad that SF Pro would be available on any mac?

mitchellh:
> Ah oops, was my assumption bad that SF Pro would be available on any mac?

Hm, I guess so. Unsure what fonts are installed on CI Macs (we don't control the image).

PR #7486, opened 2025-05-30
macos: quick terminal can equalize splits

mitchellh:
Fixes #7480

I regret to inform the haters that an AI agent did this while I was out getting a coffee. I only had to follow up once to ask it rename the method to be more idiomatic. I absolutely didn't need AI to do this, but I was leaving the office _anyways_ and figured why not. Got back and it was done. What a time to be alive.


gianpaj:
Who committed the code then? Who has the license? (trolling) :P 

Nice. I've started doing the same with smaller tasks, using @claude Github action/app and letting it open a PR for me to review on my phone.

inqw-qtxr (approved):

PR #7487, opened 2025-05-30
gtk: clean up per-surface cgroup on close

### LINKED ISSUE #6766: linux: per-surface control group isn't removed on close
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6669

<sup>Originally posted by **heftig** March 11, 2025</sup>
The surface control groups created are named `<12 hex digits>.scope` but aren't actually systemd scopes. I've found this confusing because I wanted to use systemd configuration and tools to configure these scopes.

In other terminals supporting scopes (like GNOME Terminal, Ptyxis and tmux), `systemctl --user whoami` outputs a unique scope for each surface, while for Ghostty it always outputs the same scope that Ghostty was started in.

In addition, Ghostty is leaking these cgroups, as one is created for each new surface and then not removed with the surface. This is visible in `systemd-cgls --all`.

Ghostty 1.1.2-arch, GNOME, Arch Linux.</div>

### Reproduction

1. Launch Ghostty
2. Create a bunch of tabs (any number will do, let's say 10). Close them all.
3. Run `systemd-cgls --all`

Notice that there are a bunch of empty control groups.

mitchellh:
Fixes #6766

This ensures that during surface deinit the cgroup is removed. By the time the surface is deinitialized, the subprocess should already be dead so the cgroup can be safely removed. If the cgroup cannot be removed for any reason we log a warning.

I'm still investigating whether we also need to remove the transient cgroup we create per app but that's a lot less noisy since app startup and shutdown is a lot more rare.

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -746,7 +746,21 @@ pub fn deinit(self: *Surface) void {
     self.core_surface.deinit();
     self.core_surface = undefined;
 
-    if (self.cgroup_path) |path| self.app.core_app.alloc.free(path);
+    // Remove the cgroup if we have one. We do this after deiniting the core
+    // surface to ensure all processes have exited.
+    if (self.cgroup_path) |path| {
+        internal_os.cgroup.remove(path) catch |err| {
+            // We don't want this to be fatal in any way so we just log
+            // and continue. A dangling empty cgroup is not a big deal
+            // and this should be rare.
+            log.warn(
+                "failed to remove cgroup for surface path={s} err={}",
+                .{ path, err },
+            );
+        };
+
+        self.app.core_app.alloc.free(path);
+    }
 
     // Free all our GTK stuff
     //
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/Surface.zig
```diff
@@ -746,7 +746,21 @@ pub fn deinit(self: *Surface) void {
     self.core_surface.deinit();
     self.core_surface = undefined;
 
-    if (self.cgroup_path) |path| self.app.core_app.alloc.free(path);
+    // Remove the cgroup if we have one. We do this after deiniting the core
+    // surface to ensure all processes have exited.
+    if (self.cgroup_path) |path| {
+        internal_os.cgroup.remove(path) catch |err| {
+            // We don't want this to be fatal in anyway so we just log
```
jcollie, line 753: `anyway` => `any way`

jcollie (approved):
Trivial comment typo, otherwise looks good to me.

PR #7502, opened 2025-06-02
Enable reset zoom button when `macos-titlebar-style = tabs` and only one tab

danielwe:
This is the conversion of #7496 into a PR. The problem was discussed in #3288: When `macos-titlebar-style = tabs`, the reset zoom button doesn't appear with only a single tab, i.e., no tab bar.

This fix was simply to remove an unnecessary branch, as the titlebar style doesn't make any difference for the required logic. If the tab bar is visible, you should unconditionally hide the titlebar's button and rely on the per-tab buttons. If the tab bar is not visible, you should hide or show the titlebar button depending on `surfaceIsZoomed`.

One wrinkle was that the main toolbar title started showing on top of the tabs. Turns out it was previously not showing only because it would always be pushed into the hidden overflow menu. With the reset zoom button present (either hidden or visible, but always there), this no longer happens reliably, presumably due to some cascading change in the overflow calculations. In any case, it seems brittle to rely on this way of concealing the main title, so I added a few lines of code to properly hide it when the tab bar is visible.

Screenshots:
Screenshot 2025-06-01 at 15 42 37
Screenshot 2025-06-01 at 15 42 51
Screenshot 2025-06-01 at 15 42 58

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalWindow.swift
+++ b/macos/Sources/Features/Terminal/TerminalWindow.swift
@@ -153,7 +153,7 @@ class TerminalWindow: NSWindow {
         // This is required because the removeTitlebarAccessoryViewController hook does not
         // catch the creation of a new window by "tearing off" a tab from a tabbed window.
         if let tabGroup = self.tabGroup, tabGroup.windows.count < 2 {
-            hideCustomTabBarViews()
+            resetCustomTabBarViews()
         }
 
         super.becomeKey()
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/Terminal/TerminalWindow.swift
```diff
@@ -560,6 +549,11 @@ class TerminalWindow: NSWindow {
 
         // Hide the window drag handle.
         windowDragHandle?.isHidden = true
+
+        // Reenable the main toolbar title
+        if let toolbar = toolbar as? TerminalToolbar {
+            toolbar.titleIsHidden = false
+        }
     }
```
danielwe, line 557: Perhaps this function should be renamed to something like `resetCustomTabBarViews` now that it's responsible for both hiding some elements and unhiding others? (The current name is `hideCustomTabBarViews`, see 3 lines up.)

mitchellh: I agree with this. Thank you.

mitchellh (approved):
Excellent work. I ran this locally and it is working great. Love to see it. Thank you!

PR #7503, opened 2025-06-02
config: add launched-from to specify launch source

mitchellh:
Related to #7433

This extracts our "launched from desktop" logic into a config option. The default value is detection using the same logic as before, but now this can be overridden by the user.

This also adds the systemd and dbus activation sources from #7433.

There are a number of reasons why we decided to do this:

  1. It automatically gets us caching since the configuration is only loaded once (per reload, a rare occurrence).

  2. It allows us to override the logic when testing. Previously, we had to do more complex environment faking to get the same behavior.

  3. It forces exhaustive switches in any desktop handling code, which will make it easier to ensure valid behaviors if we introduce new launch sources (as we are in #7433).

  4. It lowers code complexity since callsites don't need to have N `launchedFromX()` checks and can use a single value.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2428,6 +2428,23 @@ term: []const u8 = "xterm-ghostty",
 /// running. Defaults to an empty string if not set.
 @"enquiry-response": []const u8 = "",
 
+/// The mechanism used to launch Ghostty. This should generally not be
+/// set by users, see the warning below.
+///
+/// WARNING: This is a low-level configuration that is not intended to be
+/// modified by users. All the values will be automatically detected as they
+/// are needed by Ghostty. This is only here in case our detection logic is
+/// incorrect for your environment or for developers who want to test
+/// Ghostty's behavior in different, forced environments.
+///
+/// This is set using the standard `no-[value]`, `[value]` syntax separated
+/// by commas. Example: "no-desktop,systemd". Specific details about the
+/// available values are documented on LaunchProperties in the code. Since
+/// this isn't intended to be modified by users, the documentation is
+/// lighter than the other configurations and users are expected to
+/// refer to the code for details.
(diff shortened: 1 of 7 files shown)


src/os/locale.zig
```diff
@@ -108,11 +108,8 @@ fn setLangFromCocoa() void {
     }
 
     // Get our preferred languages and set that to the LANGUAGE
-    // env var in case our language differs from our locale. We only
-    // do this when the app is launched from the desktop because then
-    // we're in an app bundle and we are expected to read from our
-    // Bundle's preferred languages.
-    if (internal_os.launchedFromDesktop()) language: {
```
mitchellh, line 115: I could be wrong, but I spent some time dwelling on this and I don't think we needed this guard. If we launch from a non-desktop source, one of the prior language setting cases should match first. If they don't, we should probably still try this.

jcollie (approved):
LGTM

src/config/Config.zig
```diff
@@ -3111,6 +3128,11 @@ pub fn finalize(self: *Config) !void {
 
     const alloc = self._arena.?.allocator();
 
+    // Ensure our launch source is properly set.
+    if (self.@"launched-from" == null) {
+        self.@"launched-from" = .detect();
+    }
```
pluiedev, line 3134: I dunno how I feel about this... Maybe we should add a `_launched_from` variable that is always non-null? I just don't like seeing `.?` everywhere I guess

pluiedev (approved):
Looks great

src/config/Config.zig
mitchellh: Yeah I felt the same way and thought the same thing, but was worried that it'd be too easy to not know which one to check as a consumer (there being 2 sources of seeming truth). The `?` is optimized away in unsafe builds and I think the safety it affords in safe builds is fine.

PR #7504, opened 2025-06-02
macOS: Add option to hide window buttons

danielwe:
Conversion of #7497 to a PR. This implements a feature requested in #7331: an option to hide the default window buttons on macOS for a cleaner aesthetic.

~~Builds on #7502 as it requires the same change to avoid the main toolbar title showing on top of the tab bar.~~ EDIT: rebased on main now that #7502 was merged.

I aligned the scope of the new option with `macos-titlebar-style`, since they both customize titlebar elements. This means it has the same edge case quirks: For example, if you change the setting, reload the config, and then open a new tab, the appearance of the current window will depend on which tab is in the foreground. I did it this way because `macos-titlebar-style` provided an easy template for which derived configs and functions to modify. Let me know if you want me to try adjusting this so that a change in the setting also takes effect for current windows/tabs, which I _think_ should be possible.

Screenshots:
* `macos-titlebar-style = transparent` (default)
  Screenshot 2025-06-01 at 18 04 56
  Screenshot 2025-06-01 at 18 07 24
* `macos-titlebar-style = tabs`
  Screenshot 2025-06-01 at 17 56 35
  Screenshot 2025-06-01 at 17 56 48

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -377,6 +377,14 @@ class TerminalController: BaseTerminalController {
         shouldCascadeWindows = false
     }
 
+   fileprivate func hideWindowButtons() {
+        guard let window else { return }
+
+        window.standardWindowButton(.closeButton)?.isHidden = true
+        window.standardWindowButton(.miniaturizeButton)?.isHidden = true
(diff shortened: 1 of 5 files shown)


mitchellh (approved):
Looks great. Thanks.

PR #7508, opened 2025-06-03
macos: simplify some ServiceProvider code

jparise:
First, remove the always-inlined openTerminalFromPasteboard code and combine it with openTerminal. Now that we're doing a bit of work inside openTerminal, there's little better to having an intermediate, inlined function.

Second, combine some type-casting operations (saving a .map() call).

Lastly, adjust some variable names because a generic `objs` or `urls` was a little ambiguous now that we're all in one function scope.


mitchellh (approved):

mitchellh:
Thanks, this part of the macOS code is/was indeed messy.

PR #7510, opened 2025-06-03
gtk/GlobalShortcuts: don't request session with no shortcuts

alaviss:
There aren't any reasons to pay the D-Bus tax if you don't use global shortcuts.

Fixes https://github.com/ghostty-org/ghostty/discussions/7509


mitchellh (commented):
Seems right to me but want to double check with @pluiedev

pluiedev (approved):
Makes sense

PR #7515, opened 2025-06-03
gtk/TabView: do not closeTab within close-page signal handler

alaviss:
`TabView` assumes to be the sole owner of all `Tab`s within a Window. As such, it could close the managed `Window` once all tabs are removed from its widget.

However, during `AdwTabView::close-page` signal triggered by libadwaita, the `Tab` to be closed will gain an another reference for the duration of the signal, breaking `TabView.closeTab` (called via `Tab.closeWithConfirmation`) assumptions that having no tabs meant they are all destroyed.

This commit solves the issue by scheduling `Tab.closeWithConfirmation` to be run after `AdwTabView::close-page` signal has finished processing.

Fixes #7426

### DIFF
--- a/src/apprt/gtk/TabView.zig
+++ b/src/apprt/gtk/TabView.zig
@@ -7,6 +7,7 @@ const std = @import("std");
 const gtk = @import("gtk");
 const adw = @import("adw");
 const gobject = @import("gobject");
+const glib = @import("glib");
 
 const Window = @import("Window.zig");
 const Tab = @import("Tab.zig");
@@ -243,7 +244,14 @@ fn adwClosePage(
     const child = page.getChild().as(gobject.Object);
     const tab: *Tab = @ptrCast(@alignCast(child.getData(Tab.GHOSTTY_TAB) orelse return 0));
     self.tab_view.closePageFinish(page, @intFromBool(self.forcing_close));
-    if (!self.forcing_close) tab.closeWithConfirmation();
+    if (!self.forcing_close) {
+        // We cannot trigger a close directly in here as the page will stay
+        // alive until this handler returns, breaking the assumption where
+        // no pages means they are all destroyed.
+        //
+        // Schedule the close request to happen in the next event cycle.
+        _ = glib.idleAddOnce(glibIdleOnceCloseTab, tab);
+    }
 
     return 1;
 }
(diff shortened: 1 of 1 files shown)


pluiedev (approved):
This feels a bit like a hack, but I'm fine with it for now

tristan957 (commented):
I don't like this solution. Can we not just refactor tab.closeWithConfirmation() to return a boolean on whether or not to close the tab? Then we could return either `GDK_EVENT_STOP` or `GDK_EVENT_CONTINUE` based on the return value?

alaviss:
That wouldn't work either. The main issue here is Ghostty using `TabView` as a mean to do bookkeeping on tabs to manage Window state.

However, `AdwTabView` weren't really made with that in mind, the `close-page` signal stipulates that the emitter holds a reference, so all closed tabs will have an extra reference outside of `AdwTabView` jurisdiction, even if briefly. (p/s: we technically violates this in `closeTab` by calling `close_page` without holding a reference first)

I'd say that it's probably better to flip the current ownership relationship, that is: `Tab`s holding a reference to `Window`s instead of `Window` somehow owning the `Tab`s, since `Tab` uses `Window` as a container to display things within and that container isn't very useful after all tabs are gone (hence why `TabView` tries to close it in the first place).

I went with this quick fix since it's the least disruptive, but the use of `AdwTabView` might have to be revisited.

tristan957:
Will you have time to submit a follow up to clean everything up? Sounds like you have a good grasp of everything.

alaviss:
> Will you have time to submit a follow up to clean everything up?

I don't think so, unfortunately.

I'd want to run this idea through others who know this area, since I'm not that confident in my understanding of it.

mitchellh:
This is a simple and well commented hack so I'm fine with this. I agree that I don't want this to stick around. I also agree that the ownership structure in GTK is a bit weird. I've tried a couple times to fix it and I did make it _better_ but I never fully resolved it.

GTK is, for better and worse, a reference counted system. So the order in which we drop to zero references matters. Ideally, we want to be triggering our free callbacks for GTK objects in the right order so we don't get this kind of dangling issue.

I've mostly figured this out for macOS first because it's much easier to manage with automatic reference counting vs GTK. But I think my learning there can be conceptually ported to GTK over time. GTK already conceptually matches a lot of it (on purpose, for this same reason) but it's just not super clean.

The conceptual idea is that the end leaf GTK Surface owns the core Ghostty `Surface` (I think that's already true). When GTK Surface is deinit, so is the Ghostty one. But the GTK surface needs to be unattachable so we can say... move it from a split to a tab or a tab to another window and so on. This _should_ be true today.

Above that, its `Window -> Tab -> Splits`. A window destroy should trigger all its children and so on. I don't think we should assume a window exists when a tab closes. On macOS, we checked if a tab is attached to a window on close (I don't know when it can't be but we check anyways) and if not we assume we're definitely closing and don't show confirmation. We probably want to do something similar here.

Anyways, a bit of an idle brain dump but I think we need to do a more thoughtful audit of our ownership and ref counting on the GTK side.

PR #7521, opened 2025-06-04
gtk: make requesting attention configurable

### LINKED ISSUE #7520: Bell: make requesting attention configurable
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7519

<sup>Originally posted by **ahmedelgabri** June  4, 2025</sup>
I'm on the `@tip` version, and I started to notice that the icon started to bounce on the dock, so I went to check if I can disable this, because it's very distracting. But unfortunately, it says that there is currently no way to change it as the feature is unobtrusive. I know that the easy choice for me now it to move off `@tip` but I'm worried if this lands like this on the stable release, can we have a way to disable it?

https://github.com/ghostty-org/ghostty/blob/108fab11a52f09388728483e7b64c3f6a451f687/src/config/Config.zig#L1978C1-L1982C28</div>

pluiedev:
Partially fixes #7520

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1963,7 +1963,7 @@ keybind: Keybinds = .{},
 ///
 ///  * `system`
 ///
-///    Instructs the system to notify the user using built-in system functions.
+///    Instruct the system to notify the user using built-in system functions.
 ///    This could result in an audiovisual effect, a notification, or something
 ///    else entirely. Changing these effects require altering system settings:
 ///    for instance under the "Sound > Alert Sound" setting in GNOME,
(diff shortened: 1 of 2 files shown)


jcollie (approved):

src/config/Config.zig
```diff
@@ -1973,15 +1973,24 @@ keybind: Keybinds = .{},
...
-/// surface will contain a bell emoji (🔔) until the terminal is focused
-/// or a key is pressed. These are not currently configurable since they're
-/// considered unobtrusive.
+///    Request the user's attention when Ghostty is unfocused, until it has
+///    received focus again.
 ///
-/// By default, no bell features are enabled.
+///    On macOS, this will bounce the app icon in the dock once. Additionally,
+///    the title of the window with the alerted terminal surface will contain
+///    a bell emoji (🔔) until the terminal is focused or a key is pressed.
```
mitchellh: I think we should separate the title from the bounce, since the title is fairly unobtrusive and in the feedback the bounce has been the bigger issue.  Does GTK do any title stuff yet? I wonder if we should just introduce `title` in this packed structure now.

pluiedev: I think we only change the title for the zoomed-in state and not bell *yet*. Either way it's not really ideal that we're changing the title since we can put [arbitrary icons](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/property.TabPage.indicator-icon.html) on the tab

mitchellh: I think "title" here is not literally a GTK title and may in fact be tab icons, I think this is more of a user-facing or understood "title" which is just that general "title area." So I think it should still be separated.

tristan957 (approved):

mitchellh (approved):
#7526 to track the macOS implementation.

PR #7522, opened 2025-06-04
core: document keybind actions better

pluiedev:
The current documentation for actions are very sparse and would leave someone (even including contributors) as to what exactly they do. On top of that there are many stylistic and grammatical problems that are simply no longer in line with our current standards, and certainly not on par with our configuration options reference.

Hence, I've taken it upon myself to add, clarify, supplement, edit and even rewrite the documentation for most of these actions, in a wider effort of trying to offer better, clearer documentation for our users.

### DIFF
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -222,114 +222,191 @@ pub fn lessThan(_: void, lhs: Binding, rhs: Binding) bool {
 
 /// The set of actions that a keybinding can take.
 pub const Action = union(enum) {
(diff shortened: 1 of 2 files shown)


mitchellh (approved):
Thanks, I'm good with this as-is, we can continue to improve.

mitchellh:
Let's see what our AI overlords say.

mitchellh:
AI GODS ARE PLEASED.

PR #7525, opened 2025-06-05
fix: exit non-native fullscreen on close

frangio:
Fixes https://github.com/ghostty-org/ghostty/discussions/7159.


mitchellh:
Thanks, this looks good to me. I made one small change where I added the `deinit` along with some comments. I know the parent deinit would clean this up (I think?) but this just makes it extra safe in case we change how we handle notifications in the parent.

PR #7528, opened 2025-06-05
macos: fix hasWindowButtons logic (#7504 follow-up)

danielwe:
Took another look through #7504 after the merge and realized that the logic behind the `hasWindowButtons` property wasn't quite sound. It would return `false` if either _at least one_ button were missing in the `standardWindowButton(.theButton) == nil` sense, or if _all_ buttons were hidden in the `isHidden` sense.

With this PR, the logic is rectified: `false` if _all_ buttons are missing or hidden in any sense, otherwise `true`.

In practice, I suppose Ghostty won't ever instantiate a `TerminalWindow` where `standardWindowButton(.theButton) == nil`, but might as well get it right and sleep better at night.

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalWindow.swift
+++ b/macos/Sources/Features/Terminal/TerminalWindow.swift
@@ -45,15 +45,14 @@ class TerminalWindow: NSWindow {
         },
     ]
 
+    // false if all three traffic lights are missing/hidden, otherwise true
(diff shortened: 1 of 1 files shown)


mitchellh:
I'll take it. Thanks. I noticed that too and felt fine either way haha

mitchellh (approved):

Issue #5932, opened 2025-02-22
OSC 133: Support semantic prompt regions


jparise:
Our support for [semantic prompts](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md) (OSC 133) is currently [row-based](https://github.com/mitchellh/ghostty/blob/726ac36612a575ea71350d1eda020a733e2359f6/src/terminal/page.zig#L1916-L1918). It should be region-based, allowing semantic prompts to span multiple rows and allowing a row to contain multiple semantic prompts.

Row-based support is too limiting and has led to problems like #3003, where we needed to change the way we emit prompt markers to work around the implicit "last prompt marker on a row wins" behavior of the current implementation.

Region-based support would also let us reliably support more advanced prompt behavior, such as styling different prompt regions, collapsing prompts and their output, and implementing more advanced mouse-click-based movements. Here are [some examples from DomTerm](https://per.bothner.com/blog/2019/shell-integration-proposal/):

#### Prompt region styling

![Image](https://github.com/user-attachments/assets/e7cebe74-01eb-4eeb-ba87-2c92ea5bc2ee)

#### Command output collapsing

![Image](https://github.com/user-attachments/assets/26859261-3330-4962-8a49-e0a5a69e8344)

#### Prompt-aware selection

![Image](https://github.com/user-attachments/assets/54865973-b2e5-49f0-b948-6cb7dcc5be61)

## Notes

- Wezterm has a [`SemanticZone`](https://github.com/wezterm/wezterm/blob/a2fc1c212316d1887f1b217577c08a18519db81b/term/src/lib.rs#L118-L125) type that is created by semantic prompt markers. This could be a nice generalization to adopt.
- We only support a minimal subset of the [OSC 133 commands](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md). We should most/all of those commands and their behaviors.
- Similar to how we manage hyperlinks, we should optimize for sparse page-based prompt storage.
- We could add an Inspector mode to visualize semantic prompt regions.
- #1966 exists to specifically track the click-based movement part of the specification. This would be a nice-to-have, but it's not important as the other fundamentals described here.

mitchellh:
I think we should separate the shell integration side of this from the core side. I believe the actionable part of this issue is to enhance the core to understand semantic prompt regions, is that correct?

jparise:
[...]

PR #7529, opened 2025-06-06
Add macOS accessibility support for terminal content

steipete:
This is a vibe-coded basic implementation of the macOS accessibility support for terminal content.

I'm building an app that manages Claude Code AI status and while it's easy to get the screent text out with Terminal or iTer, I currently use OCR on Ghostty to get the text... which is not reliable and wasteful.

- Add `ghostty_surface_viewport_text` C API to extract visible terminal text (unsure about this, maybe better to just use the text buffer?)
- Implement `NSAccessibility` protocol in `SurfaceView_AppKit.swift`
- Add throttled accessibility notifications on content changes

This allows VoiceOver and other assistive technologies to read terminal content.

I'm pretty good at Swift, zig is new to me. Happy to work more on this, but first wanna see if this is a good approach & sth you consider adding.

Before:
GsuCfLXXwAAvSre

After:
Screenshot 2025-06-06 at 02 22 32

# AI notes incoming:
# Overview

The implementation adds basic accessibility support to the Ghostty terminal emulator, allowing assistive technologies like VoiceOver to read the terminal content. This is achieved by:

1. Adding a C API function to extract the visible terminal viewport text
2. Implementing NSAccessibility protocol methods in the macOS SurfaceView

### 1. C API Addition (src/apprt/embedded.zig)

Added `ghostty_surface_viewport_text` function that:
- Extracts the visible text content from the terminal viewport
- Handles proper UTF-8 encoding
- Trims trailing spaces from lines
- Returns properly formatted text with newlines between rows

### 2. Header File Update (include/ghostty.h)

Added the function declaration:
```c
uintptr_t ghostty_surface_viewport_text(ghostty_surface_t, char*, uintptr_t);
```
### 3. macOS Accessibility Implementation (macos/Sources/Ghostty/SurfaceView_AppKit.swift)
Implemented the following NSAccessibility protocol methods:
- `isAccessibilityElement()` - Returns true to indicate this is an accessible element
- `accessibilityRole()` - Returns `.textArea` role
- `accessibilityValue()` - Returns the current viewport text using the C API
- `accessibilityLabel()` - Returns "Terminal" as the label
- `isAccessibilityFocused()` - Returns the current focus state
- `accessibilitySelectedText()` - Returns selected text if any
- `accessibilitySelectedTextRange()` - Returns the range of selected text
- `accessibilityNumberOfCharacters()` - Returns the character count
- `accessibilityString(for:)` - Returns text for a specific range
- `accessibilityLine(for:)` - Returns line number for a character index
- `accessibilityRange(forLine:)` - Returns the range for a specific line
- `accessibilityPerformPress()` - Focuses the terminal when activated
### 4. Automatic Updates
Added throttled accessibility notifications when the terminal content changes:
- Updates are throttled to maximum 2 times per second to avoid overwhelming the accessibility system
- Notifications are sent when the terminal layer is redrawn
## Usage
With these changes, macOS accessibility tools can now:
1. Read the terminal content using VoiceOver
2. Navigate through the text line by line
3. Access selected text
4. Be notified when content changes
## Building
To build Ghostty with these changes:
1. Install Zig (required for building the core library)
2. Run `zig build` to build the XCFramework
3. Open `macos/Ghostty.xcodeproj` in Xcode
4. Build and run the project
## Testing
To test the accessibility features:
1. Enable VoiceOver (Cmd+F5)
2. Navigate to the Ghostty terminal window
3. Use VoiceOver commands to read the terminal content
4. The Accessibility Inspector should now show the terminal text content


mitchellh (changes requested):
Thanks. There is definitely a lot here that is not right at all. I still want to review this to see if there is anything we can use here, but may end up just closing it.

I think the idea of exposing the terminal contents as simply an unformatted dumb string to start for AX is an interesting idea. Doing AX right will require a lot more work and making it aware of the cursor, lines, text input status, etc. But, starting out by just saying "this is a label with this static text" is _interesting_ (which is more or less what this PR seems to be doing).

steipete:
Michael, thanks for taking a look!

Should I give this a go and expose text lines instead? I was looking for a minimal solution to start out, and happy to spend more time on a more complex implementation. I do open source for fun to pass the time after my exit, and having an opportunity to play with Zig is exciting!

This implementation would unblock my immediate needs and it works on my machine (tm). I could also take a stab at writing tests next?

jparise:
This also a good reminder (to me) to get back to my #5932 branch, which will give us an improved view of which text regions represent prompts, inputs, and output. I can see that being useful to higher-level use cases like the one described here.

steipete:
We probably want both; a way to get either the visible text or the full scrollback buffer out, and a way to access indivudal lines of text via a screenreader. I'm also motivated to look into AppleScript, but step by step.

mitchellh:
Superseded by #7601. Thanks for the inspiration. Mine is a rewrite, no vibing.

PR #7535, opened 2025-06-06
macOS: Undo/Redo for changes to windows, tabs, and splits

mitchellh:
This PR implements the ability to undo/redo new and closed windows, tabs, and splits.

## Demo

https://github.com/user-attachments/assets/98601810-71b8-4adb-bfa4-bdfaa2526dc6

### Undo Timeout

Running terminal sessions _remain running_ for a configurable period of time after close, during which time they're undoable. This is similar to "email unsend" (since email in the traditional sense can't be unsent, clients simply hold onto it for a period of time before sending).

This behavior is not unique to Ghostty. The first and only place I've seen it is in iTerm2. And iTerm2 behaves similarly, although details of our behavior and our implementation vary greatly. 

The configurable period of time is done via the `undo-timeout` configuration. The default value is 5 seconds. This feels reasonable to be and is grounded in being the default for iTerm2 as well, so it's probably a safe choice.

Undo can be disabled by setting `undo-timeout = 0`. 

### Future

The actions that can be potentially undone/redone can be easily expanded in the future. Some thoughts on things that make sense to me:

- Any sort of split resizing, including equalization
- Moving tabs or splits

#### What about Linux?

I'd love to support this on Linux. I don't think any other terminal on Linux has this kind of feature (definitely might be wrong, but I've never seen it and I've looked at a lot of terminal emulators 😄 ). But there's some work to be done to get there.

## TODO for the Draft PR

This is still a draft. There are some items remaining (list will update as I go):

- [x] Undoing a closed window is sometimes buggy still and I'm not sure why, I have to dig into this.
- [x] New window should be undoable
- [x] New tab should be undoable
- [x] Close All Windows should be undoable
- [x] I think I have to get rid of TerminalManager. Undone windows won't be in TerminalManager's list of controllers and I think that's going to break a lot of things.
- [x] I haven't tested this with QuickTerminal at all. I expect bugs there but I want undo to work with splits there.
- [x] Close window with the red traffic light button doesn't trigger undo
- [x] Closing window with multiple tabs undoes them as separate windows


leonjza:
Cool feature, thank you! Maybe a bit tongue in cheek, but curious if you'd consider making the key bind to restore cmd+shift+t rather than cmd+z, mimicking the behavior of browsers and allowing muscle memory to be transferred?

mitchellh:
> Cool feature, thank you! Maybe a bit tongue in cheek, but curious if you'd consider making the key bind to restore cmd+shift+t rather than cmd+z, mimicking the behavior of browsers and allowing muscle memory to be transferred?

Good idea. Done.

mitchellh:
Alright, all known TODOs addressed. I think I ended up fixing a handful of quick terminal issues here not related to this PR at all. But the good news is undo works great with the quick terminal too now. 😄 

I'll keep letting this sit a bit longer and consider what else I may have missed, but this is looking close.

mitchellh:
Flipping this to ready for review. I think any other changes here will be minor.

PR #7541, opened 2025-06-07
gtk/CommandPalette: prevent leaks on initialization

alaviss:
* Deallocate the builder after use
* Don't hold a reference to `Command` after appending to `GListStore`


pluiedev (approved):
Makes sense though I thought the builder instance held the ownership of all of its objects. Maybe things are different when you actually place them in a dialog? Who knows

alaviss:
> Makes sense though I thought the builder instance held the ownership of all of its objects.

It does, but they are also held by their parents in the hierarchy. Since the code already keep a ref of the top-level widget, that's all you need.

PR #7547, opened 2025-06-08
macos: add id to SplitTreeView to detect tree structural changes

### LINKED ISSUE #7546: macOS regression: closing leftmost split does the wrong thing (UI issue?)
This is a regression on tip, it doesn't affect 1.1.3.

Reproduce:

1. Launch Ghostty
2. Split right (command-d) twice, so you have 3 vertical columns
3. Focus leftmost split, enter "a" so its clear what it is
4. Close with command-w

See that the wrong split appears to close.

I think what's happening is that the correct split is being closed but we have a SwiftUI issue or something being manifested here. Note I don't think the bug is in SwiftUI actually, but our window is not redrawing or something.

mitchellh:
Fixes #7546

The comments explain what was going on here.

### DIFF
--- a/macos/Sources/Features/Splits/SplitTree.swift
+++ b/macos/Sources/Features/Splits/SplitTree.swift
@@ -1116,3 +1116,148 @@ extension SplitTree: Collection {
         return i + 1
     }
 }
+
+// MARK: Structural Identity
+
+extension SplitTree.Node {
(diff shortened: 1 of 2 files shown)


mitchellh:
This isn't quite right, I'm looking into it.

mitchellh:
Okay, a much better approach.

PR #7553, opened 2025-06-09
feat: implement mode 1048 for saving/restoring cursor position

AjStraight619:
Implements mode 1048 for saving/restoring cursor position.

This is the same as mode 1049 but only saves cursor position without touching the alternate screen.

**save/restore cursor position:**
- `ESC[?1048h` - save cursor position  
- `ESC[?1048l` - restore cursor position

**Quick test:**
```bash
printf '\e[5;10H[SAVED HERE]'
printf '\e[?1048h'                # save position
printf '\e[15;1HMoved somewhere else...'
printf '\e[?1048l'                # restore 
printf ' RESTORED!'           # should appear next to [SAVED HERE]
```

Fixes #7473


mitchellh (approved):
Added some comments to clarify what this is, but looks good thanks

PR #7576, opened 2025-06-11
macOS: Add support for quick terminal sizing configuration

### LINKED ISSUE #2384: macOS: Customize initial size for quick terminal

freepicheep:
Resolves #2384 

Since support for customizing the initial size for the quick terminal has already been added for linux, I just had to expose the configuration to Swift with the C bindings. 

Through the new `QuickTerminalSize` struct the user's custom configuration is used to set the height and width of the quick terminal. It supports pixels or percentage. It also supports a custom width for the quick terminal through the configuration. 

---

EDIT to Include Demos

**Adjusting Height**

adjusting height

**Adjusting Height and Width**

adjusting height and width


src/config/Config.zig
```diff
@@ -6103,6 +6103,35 @@ pub const QuickTerminalSize = struct {
         height: u32,
     };
 
+    /// C API structure for QuickTerminalSize
+    pub const C = extern struct {
+        primary_type: u8, // 0 = none, 1 = percentage, 2 = pixels
```
pluiedev: Use an enum with an underlying u8 representation - then you can just use `@intFromEnum`

src/config/Config.zig
```diff
@@ -6103,6 +6103,35 @@ pub const QuickTerminalSize = struct {
...
+
+    pub fn cval(self: QuickTerminalSize) C {
+        return .{
+            .primary_type = if (self.primary) |p| switch (p) {
+                .percentage => 1,
+                .pixels => 2,
+            } else 0,
+            .primary_value = if (self.primary) |p| switch (p) {
+                .percentage => |v| v,
+                .pixels => |v| @floatFromInt(v),
```
pluiedev: `@floatFromInt` is theoretically susceptible to precision loss, even though at typical values this shouldn't happen. My suggestion is to be honest with reality and model each value as a tagged union:

```zig
const Size = extern union {
  tag: Tag,
  value: u32, // @bitCast an f32 to store it here
  
  const Tag = extern enum(u8) { none, percentage, pixels };
};
```

In C this will look like:

```c
typedef enum {
  GHOSTTY_QUICK_TERMINAL_SIZE_NONE,
  GHOSTTY_QUICK_TERMINAL_SIZE_PERCENTAGE,
  GHOSTTY_QUICK_TERMINAL_SIZE_PIXELS,
} ghostty_quick_terminal_size_e;

typedef union {
  ghostty_quick_terminal_size_e tag;
  uint32_t value;
} ghostty_quick_terminal_size_u;
```

pluiedev (changes requested):
Only going to comment on the Zig and C parts as I don't know enough about Swift

src/config/Config.zig
freepicheep: Thanks for the feedback! I wasn't able to get it working with a union, but I used a struct with the same concept.

src/config/Config.zig
freepicheep: Thanks!

freepicheep:
I broke some things with manually resizing the quick terminal window. I am working on fixing that.

freepicheep:
Here are some examples of the configurable and more adjustable quick terminal. I decided to expand the quick terminal on both sides when the width (for terminal from top, bottom, center) or height (from left, right, or center) is adjusted. 

## 75% 75% and Adjustments from Top

resizable-from-top

## 75% 75% and Adjustments from Right

resizable-from-right

## 75% 75% and Adjustments from Center

resizable-from-center

## No Config

no-config

src/config/Config.zig
```diff
@@ -6234,6 +6234,44 @@ pub const QuickTerminalSize = struct {
...
+
+    pub const CSize = extern struct {
+        type: Type,
+        value: u32,
+
+        pub const Type = enum(u8) { none, percentage, pixels };
+
+        fn none() CSize {
+            return .{ .type = .none, .value = 0 };
+        }
```
pluiedev: Use decl literals

```suggestion
        pub const none: CSize = .{ .type = .none, .value = 0 };
```

src/config/Config.zig
```diff
@@ -6234,6 +6234,44 @@ pub const QuickTerminalSize = struct {
...
+    pub fn cval(self: QuickTerminalSize) C {
+        return .{
+            .primary = if (self.primary) |p| switch (p) {
+                .percentage => |v| CSize.percentage(v),
+                .pixels => |v| CSize.pixels(v),
+            } else CSize.none(),
+            .secondary = if (self.secondary) |s| switch (s) {
+                .percentage => |v| CSize.percentage(v),
+                .pixels => |v| CSize.pixels(v),
+            } else CSize.none(),
```
pluiedev: Use decl literals

```suggestion
            .primary = if (self.primary) |p| switch (p) {
                .percentage => |v| .percentage(v),
                .pixels => |v| .pixels(v),
            } else .none,
            .secondary = if (self.secondary) |s| switch (s) {
                .percentage => |v| .percentage(v),
                .pixels => |v| .pixels(v),
            } else .none,
```

pluiedev (changes requested):

freepicheep (commented):
I believe the requested changes have been implemented. Thanks for your feedback!

include/ghostty.h
```diff
@@ -445,6 +445,23 @@ typedef struct {
...
+typedef enum {
+  GHOSTTY_QUICK_TERMINAL_SIZE_NONE,
+  GHOSTTY_QUICK_TERMINAL_SIZE_PERCENTAGE,
+  GHOSTTY_QUICK_TERMINAL_SIZE_PIXELS,
+} ghostty_quick_terminal_size_e;
+
+typedef struct {
+  ghostty_quick_terminal_size_e type;
+  uint32_t value;
+} ghostty_quick_terminal_size_u;
```
mitchellh: This should be suffixed `_s` since its a struct, not a union.

macos/Sources/Features/QuickTerminal/QuickTerminalSize.swift
```diff
@@ -0,0 +1,84 @@
+import GhosttyKit
+
+struct QuickTerminalSize {
```
mitchellh: We should put this in the `Ghostty` folder because its a Swift structure wrapper around a libghostty type.

mitchellh (changes requested):
I need to review the Swift in more detail but here's some quick feedback

mitchellh:
I pushed a rebase, no other changes.

freepicheep:
Thanks for your feedback and for rebasing it. I just implemented your suggestions and pushed the update.

freepicheep:
Is there anything I can do to move this PR forward?

luweiCN:
> I really need this feature. Currently, I’m using Aerospace to open Ghostty with a single hotkey. Before switching to Ghostty, I had been using iTerm2 for a long time, mainly because of its Hotkey Window feature, which allowed me to summon a terminal from any display and any desktop space instantly. I usually ran Neovim inside that window.
> 
> After switching to Ghostty, I was pleasantly surprised to find that it also has a similar feature — Quick Terminal. However, I encountered one limitation that prevents it from fully replacing my previous workflow: the Quick Terminal does not remember its window size.
> 
> Currently, while the Quick Terminal size can be adjusted, it always resets to the default size after I hide and reopen it. Even restarting Ghostty does not preserve the size.
> 
> Suggestions I think having a configurable default size could be nice, but it’s not essential. What’s more important is that the Quick Terminal should remember its last size — even after being hidden, and ideally even after restarting Ghostty.
> 
> I noticed there has been some consideration about which direction the Quick Terminal should appear from, and how the animation and resize behavior work depending on that direction. In my opinion, these are less important. The core need for Quick Terminal users is to quickly summon a terminal from any display and any desktop space using a global hotkey — without interfering with the rest of the workspace. A semi-transparent background, for example, allows you to see the web page or messages underneath.
> 
> Animations and directional controls can be nice to have, but if they make implementation more complex, I’d prefer focusing on the core functionality — remembering window size and being able to summon it from anywhere with a global hotkey.
> 
> I’m not familiar with Zig, so I can’t contribute code to this feature, but I hope my feedback helps clarify the main use case. Really looking forward to seeing this PR merged.

Sorry my mistake, I just found nightly version quick terminal can remember the window size

pluiedev (approved):
Ack Zig & C changes

mitchellh:
I think I remember this having some bugs which is why I didn’t merge after my original feedback was addressed. I’ll have to double check.

freepicheep:
> I think I remember this having some bugs which is why I didn’t merge after my original feedback was addressed. I’ll have to double check.

Let me know if you find any. This is a feature I'm really looking forward to having and would like to see it ship with 1.2.

mitchellh:
Thanks again! I reviewed this, made some changes, and pushed them as #8402 because I couldn't modify this PR. Original authorship was maintained. I'll merge that once CI passes.

PR #7601, opened 2025-06-15
macOS: Basic Read-Only Accessibility Integration

mitchellh:
This integrates with macOS accessibility APIs to expose Ghostty terminal structure and content.

This is a very, very bare implementation and the terminal contents currently reported are the _full screen and scrollback_ which is way too much for realistic human accessibility use. The target use case for this PR is to enable automated tooling (namely, AI screen readers). However, this is all groundwork we'll need to iterate and improve the accessibility work anyways.

To make this work, I also replatformed some of our hacky C APIs onto a more robust `ghostty_surface_read_text` API that can now read arbitrary ranges of the screen into C strings for consumers to use. This will be useful in more places going forward (hint hint).

## Before

Accessibility tooling can't read anything, Ghostty has no attributes, no contents, just shows up as a square.

CleanShot 2025-06-15 at 14 06 55@2x

## After

A lot of metadata, including the screen contents as text.

CleanShot 2025-06-15 at 14 07 25@2x

Also, split hierarchies are navigable:

https://github.com/user-attachments/assets/a7b2ffb7-dbeb-41b2-8705-9c3200812c4d

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -355,6 +355,27 @@ typedef struct {
   double tl_px_y;
   uint32_t offset_start;
   uint32_t offset_len;
+  const char* text;
+  uintptr_t text_len;
+} ghostty_text_s;
+
+typedef enum {
+  GHOSTTY_POINT_ACTIVE,
+  GHOSTTY_POINT_VIEWPORT,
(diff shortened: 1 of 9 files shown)


steipete:
Love to see this being picked up, thanks!

PR #7606, opened 2025-06-16
feat: add FreeBSD support

charlesrocket:
- [x] Waiting for mitchellh/libxev#167
- [x] Translations
- [x] x11
- [x] Wayland
- [ ] CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -269,6 +269,29 @@ jobs:
           path: |-
             ghostty-source.tar.gz
 
+  build-freebsd:
+    runs-on: ubuntu-latest
+    needs: test
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: test
+        uses: vmactions/freebsd-vm@v1
+        with:
+          usesh: true
+          prepare: |
+            sed 's/quarterly/latest/g' /etc/pkg/FreeBSD.conf
+            pkg update -f
+            pkg upgrade -y
+            pkg install -y wget
+
+          run: |
+            # switch to pkg
+            wget https://ziglang.org/builds/zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
+            tar -xf zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
+            zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42/zig build -Dapp-runtime=none test
+
   build-macos:
     runs-on: namespace-profile-ghostty-macos-sequoia
     needs: test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -280,16 +280,13 @@ jobs:
       - run: sudo systemctl start ssh
 
       - name: test
-        uses: vmactions/freebsd-vm@v1
+        uses: acj/freebsd-firecracker-action@v0.5.0
         with:
-          usesh: true
-          prepare: |
+          run: |
             sed 's/quarterly/latest/g' /etc/pkg/FreeBSD.conf
             pkg update -f
             pkg upgrade -y
             pkg install -y wget
-
-          run: |
             # switch to pkg
             wget https://ziglang.org/builds/zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
             tar -xf zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -269,29 +269,6 @@ jobs:
           path: |-
             ghostty-source.tar.gz
 
-  build-freebsd:
-    runs-on: namespace-profile-ghostty-md-systemd
-    needs: test
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      # Needed to work on Namespace
-      - run: sudo systemctl start ssh
-
-      - name: test
-        uses: acj/freebsd-firecracker-action@v0.5.0
-        with:
-          run: |
-            sed 's/quarterly/latest/g' /etc/pkg/FreeBSD.conf
-            pkg update -f
-            pkg upgrade -y
-            pkg install -y wget
-            # switch to pkg
-            wget https://ziglang.org/builds/zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
-            tar -xf zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
-            zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42/zig build -Dapp-runtime=none test
-
   build-macos:
     runs-on: namespace-profile-ghostty-macos-sequoia
     needs: test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -270,11 +270,14 @@ jobs:
             ghostty-source.tar.gz
 
   build-freebsd:
-    runs-on: ubuntu-latest
+    runs-on: namespace-profile-ghostty-md-systemd
     needs: test
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+
+      # Needed to work on Namespace
+      - run: sudo systemctl start ssh
 
       - name: test
         uses: vmactions/freebsd-vm@v1
--- a/build.zig
+++ b/build.zig
@@ -46,7 +46,7 @@ pub fn build(b: *std.Build) !void {
     if (config.app_runtime != .none) {
         exe.install();
         resources.install();
-        if (builtin.target.os.tag != .freebsd) i18n.install();
+        i18n.install();
     }
 
     // Libghostty
--- a/build.zig
+++ b/build.zig
@@ -46,7 +46,7 @@ pub fn build(b: *std.Build) !void {
     if (config.app_runtime != .none) {
         exe.install();
         resources.install();
-        i18n.install();
+        if (builtin.target.os.tag != .freebsd) i18n.install();
     }
 
     // Libghostty
--- a/pkg/fontconfig/build.zig
+++ b/pkg/fontconfig/build.zig
@@ -164,11 +164,23 @@ fn buildLib(b: *std.Build, module: *std.Build.Module, options: anytype) !*std.Bu
             "-DHAVE_SYS_STATVFS_H",
 
             "-DFC_CACHEDIR=\"/var/cache/fontconfig\"",
-            "-DFC_TEMPLATEDIR=\"/usr/share/fontconfig/conf.avail\"",
-            "-DFONTCONFIG_PATH=\"/etc/fonts\"",
-            "-DCONFIGDIR=\"/usr/local/fontconfig/conf.d\"",
             "-DFC_DEFAULT_FONTS=\"<dir>/usr/share/fonts</dir><dir>/usr/local/share/fonts</dir>\"",
         });
+
+        if (target.result.os.tag == .freebsd) {
+            try flags.appendSlice(&.{
+                "-DFC_TEMPLATEDIR=\"/usr/local/etc/fonts/conf.avail\"",
+                "-DFONTCONFIG_PATH=\"/usr/local/etc/fonts\"",
+                "-DCONFIGDIR=\"/usr/local/etc/fonts/conf.d\"",
+            });
+        } else {
+            try flags.appendSlice(&.{
+                "-DFC_TEMPLATEDIR=\"/usr/share/fontconfig/conf.avail\"",
+                "-DFONTCONFIG_PATH=\"/etc/fonts\"",
+                "-DCONFIGDIR=\"/usr/local/fontconfig/conf.d\"",
+            });
+        }
+
         if (target.result.os.tag == .linux) {
             try flags.appendSlice(&.{
                 "-DHAVE_SYS_STATFS_H",
--- a/src/build/GhosttyI18n.zig
+++ b/src/build/GhosttyI18n.zig
@@ -1,6 +1,7 @@
 const GhosttyI18n = @This();
 
 const std = @import("std");
+const builtin = @import("builtin");
 const Config = @import("Config.zig");
 const gresource = @import("../apprt/gtk/gresource.zig");
 const internal_os = @import("../os/main.zig");
@@ -21,6 +22,11 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyI18n {
     defer steps.deinit();
 
     inline for (internal_os.i18n.locales) |locale| {
+        const target_locale = comptime if (builtin.target.os.tag == .freebsd)
+            std.mem.trimRight(u8, locale, ".UTF-8")
+        else
+            locale;
+
         const msgfmt = b.addSystemCommand(&.{ "msgfmt", "-o", "-" });
         msgfmt.addFileArg(b.path("po/" ++ locale ++ ".po"));
 
@@ -28,7 +34,7 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyI18n {
             msgfmt.captureStdOut(),
             std.fmt.comptimePrint(
                 "share/locale/{s}/LC_MESSAGES/{s}.mo",
-                .{ locale, domain },
+                .{ target_locale, domain },
             ),
         ).step);
     }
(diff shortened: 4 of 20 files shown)


charlesrocket:
`libxev` needs to be replaced after the upstream merge

build.zig
```diff
@@ -46,7 +46,7 @@ pub fn build(b: *std.Build) !void {
     if (config.app_runtime != .none) {
         exe.install();
         resources.install();
-        i18n.install();
+        if (builtin.target.os.tag != .freebsd) i18n.install();
```
pluiedev: Why is this here? I thought FreeBSD supports gettext just fine.

src/config/Config.zig
```diff
@@ -1823,7 +1823,7 @@ keybind: Keybinds = .{},
 /// Automatically hide the quick terminal when focus shifts to another window.
 /// Set it to false for the quick terminal to remain open even when it loses focus.
 ///
-/// Defaults to true on macOS and on false on Linux. This is because global
+/// Defaults to true on macOS and on false on Linux/BSD. This is because global
 /// shortcuts on Linux require system configuration and are considerably less
```
pluiedev, line 1827: You forgot to edit this one, too. I wonder if we should just replace all instances of "Linux" to "the GTK app (runtime)"... pending discussion

src/config/Config.zig
```diff
@@ -2381,7 +2381,10 @@ keybind: Keybinds = .{},
 ///   * `single-instance` - Enable cgroups only for Ghostty instances launched
 ///     as single-instance applications (see gtk-single-instance).
 ///
-@"linux-cgroup": LinuxCgroup = .@"single-instance",
+@"linux-cgroup": LinuxCgroup = if (builtin.os.tag == .linux)
+    .@"single-instance"
+else
+    .never,
```
pluiedev, line 2410: Why do we need a specific switch for Linux even though the option already indicates that it's Linux-specific? Maybe we should rename the option then?

src/os/i18n.zig
```diff
@@ -69,22 +69,24 @@ pub const InitError = error{
...
-    var buf: [std.fs.max_path_bytes]u8 = undefined;
-    const path = std.fmt.bufPrintZ(&buf, "{s}/locale", .{share_dir}) catch
-        return error.OutOfMemory;
-
-    // Bind our bundle ID to the given locale path
-    log.debug("binding domain={s} path={s}", .{ build_config.bundle_id, path });
-    _ = bindtextdomain(build_config.bundle_id, path.ptr) orelse
-        return error.OutOfMemory;
+    switch (builtin.os.tag) {
+        .windows, .freebsd => return,
```
pluiedev: Again, why is it unsupported on FreeBSD?

pluiedev (changes requested):
Also in general I think that instead of Linux/BSD we should just unify on "the GTK app", even though GTK can technically also be built for macOS and Windows. It's a bit of a mess, since FreeBSD behaves like Linux in userland and more like macOS in kernel land, so I really am not sure how best we should explain this.

build.zig
charlesrocket: Cannot get it working yet

src/config/Config.zig
charlesrocket: It should be just `cgroups` since one can enable it on BSD as well, so I just gated it for now. Config change such as this should probably be done in its own PR, but I can push it if it is okay.

src/os/i18n.zig
charlesrocket: It is off until I figure out a proper way to enable it.

src/config/Config.zig
charlesrocket: Another option is to replace Linux with Unix/Unix-like.

pluiedev: The thing is that macOS is also technically a Unix system (even more so than Linux), but on the user-facing level the BSDs are far more similar to Linux. It's a tricky nomenclature problem for sure

pluiedev:
I'm going to convert this PR to draft for now since some parts of it are still unfinished (i.e. localization support) and pending on the libxev PR. It's just not productive right now IMO to assess this PR as it currently stands. Once the libxev PR is merged and this PR has reached (near-)feature parity with Linux, then we can again mark it as ready for review.

charlesrocket:
@pluiedev fixed local issues, builds are passing now with i18n. Oh, theres error on the release that I need to look into.

.github/workflows/test.yml
```diff
@@ -269,6 +269,29 @@ jobs:
...
+            sed 's/quarterly/latest/g' /etc/pkg/FreeBSD.conf
+            pkg update -f
+            pkg upgrade -y
+            pkg install -y wget
+
+          run: |
+            # switch to pkg
+            wget https://ziglang.org/builds/zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
+            tar -xf zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42.tar.xz
+            zig-x86_64-freebsd-0.15.0-dev.777+6810ffa42/zig build -Dapp-runtime=none test
```
tristan957: Hmm, is it not possible to use 0.14.0 or 0.14.1? Seems like this could cause headaches later.

charlesrocket: Fresh zig is already in the latest branch so it will be in quarterly soon.  This is just a temp workaround.

00-kat:
> Also in general I think that instead of Linux/BSD we should just unify on "the GTK app"

Aside: the repo currently has `os/macos` and `os/linux` labels; would FreeBSD get its own label or should there be a label for “GTK app” instead?

charlesrocket:
@00-kat I'd say it should be both BSD and GTK (`gui/gtk`). There is probably no need to have individual labels for BSDs (at least not yet).

charlesrocket:
Translations are working as expected now.

src/build/GhosttyI18n.zig
```diff
@@ -21,14 +22,19 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyI18n {
     defer steps.deinit();
 
     inline for (internal_os.i18n.locales) |locale| {
+        const target_locale = comptime if (builtin.target.os.tag == .freebsd)
```
mitchellh, line 28: Can you add a comment to indicate why this is necessary? I'm not very familiar with BSD so the first question is why it differs here.

mitchellh (commented):
Looks very good. Here is the TODO on my end:

- [x] Review the xev changes and fix as necessary
- [ ] Migrate CI to Namespace and verify it works there
- [ ] Consider adding `prefer` to the non-dynamic xev to avoid the comptime checks here

src/build/GhosttyI18n.zig
charlesrocket: Done

mitchellh:
Note: `vmactions/freebsd-vm` doesn't work on Namespace yet. I've opened a support ticket with them to get this looked at (cc @hugosantos). I didn't block libxev on it because its low traffic, but I want to wait for that to merge this because its so much more expensive to run Ghostty's CI.

charlesrocket:
@mitchellh I thought [boxybsd](https://boxybsd.com/) could be used, but I doubt I can get GH runner working. There is also https://github.com/acj/freebsd-firecracker-action.

mitchellh:
Namespace fixed the issue, we can run the FreeBSD VM action on Namespace now, see libxev: https://github.com/mitchellh/libxev/pull/168

mitchellh:
Moving this to ready for review and will make some tweaks to get this merged. It may not be perfect yet but it at least gets Ghostty working to some extent (similar to Windows).

mitchellh:
@charlesrocket We only test unit tests on FreeBSD right now in CI. It would be better to test both the GLFW and GTK builds as well. Can you work on a future PR to address that?

mitchellh (approved):
Excellent work @charlesrocket. Thank you very much! 😄 

FreeBSD will remain a lower tier platform for now since none of the maintainers have access to actually test on it, but we at least have CI ensuring we don't break it and an active community member to help support it.

We can talk about more maintenance going forward in Discord. Thanks 😄

mitchellh:
Hm, looks like the FreeBSD CI job failed... kind of a weird failure. I'll rerun it and see if that works.

charlesrocket:
@mitchellh definitely, finishing the freebsd port and will switch to GLFW build error.

mitchellh:
@charlesrocket The firecracker VM approach didn't work for probably unrelated CI reasons. I'm sorry I haven't had time to look into this yet. I'd be fine merging this WITHOUT CI for now but we have to both be okay with the reality that FreeBSD support _will probably bitrot_ until we have CI support because we don't have anyone staying on top of it.

If you're okay with that in the interim we can remove the freebsd job and try a subsequent PR to add it.

charlesrocket:
@mitchellh I am okay with missing CI for now

src/build/GhosttyI18n.zig
```diff
@@ -21,14 +22,22 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyI18n {
     defer steps.deinit();
 
     inline for (internal_os.i18n.locales) |locale| {
+        // There is no encoding suffix in the LC_MESSAGES path on FreeBSD,
```
pluiedev, line 25: Not a blocker, but I'm curious how programs like vim gets away with it on FBSD. Are there truly no programs out there that contain the encoding suffix?

src/config/Config.zig
pluiedev, line 1827: Note to self: go through all the configs mentioning "Linux" and think about whether it applies to BSDs as well

src/config/Config.zig
```diff
@@ -2812,7 +2815,7 @@ pub fn loadCliArgs(self: *Config, alloc_gpa: Allocator) !void {
     //     styling, etc. based on the command.
     //
     // See: https://github.com/Vladimir-csp/xdg-terminal-exec
-    if (comptime builtin.os.tag == .linux) {
+    if ((comptime builtin.os.tag == .linux) or (comptime builtin.os.tag == .freebsd)) {
```
pluiedev, line 2841: We should probably add a helper function inside `std/os` that determines whether the current system is based on FreeDesktop (XDG) standards, since I can totally see us needing to expand this check a lot later on.

pluiedev (approved):
I still have my reservations about some of these changes (a lot of them look a bit too tacked-on with the `builtin.target.os.tag` check), but I don't feel like blocking this further.

In the future I think there needs to be a cleanup PR to unify the terminology a little bit: we should use "GTK app" instead of "Linux" in a lot of cases within the codebase, unless something is truly specific to Linux (nothing immediately comes to mind yet). This IMO is the least intrusive option since although the GTK app can technically be built for macOS, we never distribute it in that form and we generally assume it to be run in a FreeDesktop-based environment (Linux, FreeBSD, etc.). We also should probably update the website ahead of Ghostty 1.2 to reflect this nomenclature change.

ghost:
i mean if the want for more portability beyond freebsd is interesting i have quiet a few friends who use openbsd which i could ask to help with knowledge about the inner workings of it

mitchellh:
@pluiedev 

> I still have my reservations about some of these changes (a lot of them look a bit too tacked-on with the builtin.target.os.tag check), but I don't feel like blocking this further.

I don't think there was much. There was zero cases added to the GTK app (just an xev check unrelated to BSD), and almost the entirety of OS tag checks are in the build system and `src/os` package, both of which are inherently extremely OS/target dependent. I think it's pretty clean, considering.

> In the future I think there needs to be a cleanup PR to unify the terminology a little bit: we should use "GTK app" instead of "Linux" in a lot of cases within the codebase, unless something is truly specific to Linux (nothing immediately comes to mind yet).

This is probably right. I've avoided "GTK" in case we ever actually dropped GTK as the default for Linux (and now BSD). But that's probably a stretch. Using the apprt as the description is probably more appropriate. I think `macOS` is in line with that because while macOS uses the libghostty apprt technically, its really a dedicated macOS app there consuming that.

mitchellh (approved):
Approved, given the caveats noted earlier, which I'll repeat here.

We never got CI working. As such, we've all agreed that the FreeBSD support _may bitrot_. We don't have any active maintainers running FreeBSD (or any BSD for that matter). We can't guarantee that PRs the community or ourselves make will continue to work on FreeBSD. As such, we should strive to get FreeBSD in CI.

Until then, I don't want to block improvements on our portability, so I'm willing to approve and merge with that disclaimer out of the way.

And I want to thank @charlesrocket again for the excellent effort and work to get us here!

pkg/fontconfig/build.zig
```diff
@@ -164,11 +164,23 @@ fn buildLib(b: *std.Build, module: *std.Build.Module, options: anytype) !*std.Bu
...
             "-DFC_CACHEDIR=\"/var/cache/fontconfig\"",
-            "-DFC_TEMPLATEDIR=\"/usr/share/fontconfig/conf.avail\"",
-            "-DFONTCONFIG_PATH=\"/etc/fonts\"",
-            "-DCONFIGDIR=\"/usr/local/fontconfig/conf.d\"",
             "-DFC_DEFAULT_FONTS=\"<dir>/usr/share/fonts</dir><dir>/usr/local/share/fonts</dir>\"",
         });
+
+        if (target.result.os.tag == .freebsd) {
+            try flags.appendSlice(&.{
+                "-DFC_TEMPLATEDIR=\"/usr/local/etc/fonts/conf.avail\"",
```
arrowd, line 172: The correct way to obtain the `/usr/local` part on FreeBSD is to query `sysctl user.localbase`. Hardcoding it would work for 99% cases, though.

src/build/GhosttyResources.zig
```diff
@@ -16,6 +17,12 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyResources {
 
     // Terminfo
     terminfo: {
+        const os_tag = cfg.target.result.os.tag;
+        const terminfo_share_dir = if (os_tag == .freebsd)
```
arrowd, line 21: I don't get this part. We do have `/usr/local/share/terminfo`, but not `site-terminfo`.

Skimming through FreeBSD's `man terminfo` there seem to be no way to add terminfo data at all. So I'd rather revert these changes to simplify ghostty code and then do a full research how 3rd party terminfos should be installed.

src/os/resourcesdir.zig
```diff
@@ -54,11 +55,16 @@ pub fn resourcesDir(alloc: std.mem.Allocator) !?[]const u8 {
...
-        // On all platforms, we look for a /usr/share style path. This
+        // On all platforms (except BSD), we look for a /usr/share style path. This
         // is valid even on Mac since there is nothing that requires
         // Ghostty to be in an app bundle.
         inline for (sentinels) |sentinel| {
-            if (try maybeDir(&dir_buf, dir, "share", sentinel)) |v| {
+            if (try maybeDir(
+                &dir_buf,
+                dir,
+                if (builtin.target.os.tag == .freebsd) "local/share" else "share",
```
arrowd, line 65: This seems wrong. If this path is relative to the installation prefix, then just "share" is correct.

It is the PREFIX value that is different between FreeBSD and Linux (/usr/local vs. /usr), but not the layout inside.

pluiedev:
@arrowd Could you perhaps open a new discussion/PR for the comments you've made here? Leaving comments on a merged PR could cause them to be overlooked.

src/build/GhosttyResources.zig
charlesrocket:
```
/usr/local/share/site-terminfo
/usr/local/share/site-terminfo/f
/usr/local/share/site-terminfo/f/foot
/usr/local/share/site-terminfo/f/foot-direct
```
```
cat: /usr/local/share/terminfo: No such file or directory
```

14.3-n271432-8c9ce319fef7

arrowd: Hum, not sure where these files come from on your system. Here's mine:
```
pkg which /usr/local/share/terminfo/f/foot-direct 
/usr/local/share/terminfo/f/foot-direct was installed by package terminfo-db-20231209
```
Looking at `misc/terminfo-db`, it always installs into `DATADIR`, which is `${LOCALBASE}/share/terminfo`.

If `pkg which` doesn't tell you which package these files come from, then you've somehow created them on your own.

charlesrocket:
```
/usr/local/share/site-terminfo/f/foot-direct was installed by package foot-1.22.3
```

arrowd: Aha, so `x11/foot` does moving to avoid conflicting with terminfo-db: https://github.com/freebsd/freebsd-ports/blob/7120f467f63539443aa5b6c6939dfbcad8da8102/x11/foot/Makefile#L54-L62

We should do the same in our port, but upstream should not bother with our stuff, so my initial proposition to revert the `site-terminfo` change stays.

louwers:
I can recommend CirrusCI for FreeBSD builds. https://cirrus-ci.org/

Nicely integrates with GitHub.

PR #7619, opened 2025-06-17
build: use `xcrun --sdk <sdk> metal` for metal paths

mitchellh:
This wasn't working before but it just requires a restart of the machine for the changes to take effect. The namespace runners have this prebuilt so this should work now.

The other workaround was flaky for unknown reasons so I'd prefer to go back to this.

### DIFF
--- a/src/build/MetallibStep.zig
+++ b/src/build/MetallibStep.zig
@@ -22,10 +22,11 @@ step: *Step,
 output: LazyPath,
 
 pub fn create(b: *std.Build, opts: Options) ?*MetallibStep {
-    switch (opts.target.result.os.tag) {
-        .macos, .ios => {},
-        else => return null, // Only macOS and iOS are supported.
-    }
+    const sdk = switch (opts.target.result.os.tag) {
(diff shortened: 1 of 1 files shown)


jacobsandlund:
Just commenting that this `requires a restart` was important for solving this issue:

```
/usr/bin/xcrun -sdk macosx metal -o /Users/jacobsandlund/dev/ghostty/.zig-cache/o/79821378fc3dd7f0a099c7ccb4725d65/Ghostty.ir -c /Users/jacobsandlund/dev/ghostty/src/renderer/shaders/shaders.metal -mmacos-version-min=13.0.0

error: error: cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain
```

i also followed this workaround for getting the metal toolchain, which i'm now not sure if i actually needed to do, since the restart was the final step to getting this working: https://flutter-review.googlesource.com/c/recipes/+/66520 (see also the issue: https://github.com/flutter/flutter/issues/170301)

PR #7620, opened 2025-06-17
The Big Renderer Rework

qwerasd205:
It's here, the long-foretold and long-procrastinated renderer rework! Hopefully this makes it easier to adapt and modify the renderer in the future and ensures feature parity between Metal and OpenGL. Despite having been a lot of work to write initially, with the abstraction layer in place I feel like working on the renderer will be a much more pleasant experience going forward.

## Key points
- CPU-side renderer logic is now mostly unified via a generic `Renderer`.
- A graphics API abstraction layer over OpenGL and Metal has been introduced.
- Minimum OpenGL version bumped to `4.3`, so can no longer be run on macOS; I used the nix VM stuff for my testing during development. (Edit by @mitchellh: Note for readers that Ghostty still works on macOS, but the OpenGL backend doesn't, only the Metal one)
- The OpenGL backend now supports linear blending! Woohoo! The default `alpha-blending` has been updated to `linear-corrected` since it's essentially a strict improvement over `native`. The default on macOS is still `native` though to match other mac apps in appearance, since macOS users are more sensitive to text appearance.
- Custom shaders can now be hot reloaded.
- The background color is once again drawn by us, so custom shaders can interact with it properly. In general, custom shaders should be a little more robust.

## The abstraction layer
The general hierarchy of the abstraction layer is as such:
```
 [ GraphicsAPI ] - Responsible for configuring the runtime surface
    |     |        and providing render `Target`s that draw to it,
    |     |        as well as `Frame`s and `Pipeline`s.
    |     V
    | [ Target ] - Represents an abstract target for rendering, which
    |              could be a surface directly but is also used as an
    |              abstraction for off-screen frame buffers.
    V
 [ Frame ] - Represents the context for drawing a given frame,
    |        provides `RenderPass`es for issuing draw commands
    |        to, and reports the frame health when complete.
    V
 [ RenderPass ] - Represents a render pass in a frame, consisting of
   :              one or more `Step`s applied to the same target(s),
 [ Step ] - - - - each describing the input buffers and textures and
   :              the vertex/fragment functions and geometry to use.
   :_ _ _ _ _ _ _ _ _ _/
   v
 [ Pipeline ] - Describes a vertex and fragment function to be used
                for a `Step`; the `GraphicsAPI` is responsible for
                these and they should be constructed and cached
                ahead of time.

 [ Buffer ] - An abstraction over a GPU buffer.

 [ Texture ] - An abstraction over a GPU texture.
```
More specific documentation can be found on the relevant structures.

## Miscellany
Renderers (which effectively just means the generic renderer) are now expected to only touch GPU resources in `init`, certain lifecycle functions such as the `displayRealized`/`displayUnrealized` callbacks from GTK-- and `drawFrame`; and are also expected to be thread-safe. This allows the renderer thread to build the CPU-side buffers (`updateFrame`) even if we can only *draw* from the app thread.

Because of this change, we can draw synchronously from the main thread on macOS when necessary to always have a frame of the correct size during a resize animation. This was necessary to allow the background to be drawn by our GPU code (instead of setting a background color on the layer) while still avoiding holes during resize.

The OpenGL backend now theoretically has access to multi-buffering, but it's disabled (by setting the buffer count to 1) because it synchronously waits for frames to complete anyway which means that the extra buffers were just a waste of memory.

## Validation
To validate that there are no significant or obvious problems, I exercised both backends with a variety of configurations, and visually inspected the results. Everything looks to be in order.

The images are available in a gist here: https://gist.github.com/qwerasd205/c1bd3e4c694d888e41612e53c0560179

## Memory
Here's a comparison of memory usage for ReleaseFast builds on macOS, between `main` and this branch.
Memory figures given are values from Activity Monitor measuring windows of the same size, with two tabs with 3 splits each.

||Before|After|
|-:|-|-|
|**Memory**|247.9 MB|224.2 MB|
|**Real Memory**|174.4 MB|172.5 MB|

Happily, the rework has slightly *reduced* the memory footprint- likely due to removing the overhead of `CAMetalLayer`. (The footprint could be reduced much further if we got rid of multi-buffering and satisfied ourselves with blocking for each frame, but that's a discussion for another day.)

If someone could do a similar comparison for Linux, that'd be much appreciated!

## Notes / future work
- There are a couple structures that *can* be unified using the abstraction layer, but I haven't gotten around to unifying yet. Specifically, in `renderer/(opengl|metal)/`, there's `cell.zig` and `image.zig`, both of which are substantially identical between the two backends. `shaders.zig` may also be a candidate for unification, but that might be *overly* DRY.
- ~~I did not double-check the documentation for config options, which may mention whether certain options can be hot-reloaded; if it does then that will need to be updated.~~ Fixed: be5908f
- The `fragCoord` for custom shaders originates at the top left for Metal, but *bottom* left for OpenGL; fixing this will be a bit annoying, since the screen texture is likewise vertically flipped between the two. Some shaders rely on the fragcoord for things like falling particles, so this does need to be fixed.
- `Target` should be improved to support multiple types of targets right now it only represents a framebuffer or iosurface, but it should also be able to represent a texture; right now a kind of messy tagged union is used so that steps can accept both.
- Custom shader cursor uniforms (#6912) and terminal background images (#4226, #5233) should be much more straightforward to implement on top of this rework, and I plan to make follow-up PRs for them once this is merged.
- I *do* want to do a rework of the pipelines themselves, since the way we're rendering stuff is a bit messy currently, but this is already a huge enough PR as it is- so for now the renderer still uses the same rendering passes that Metal did before.
- We should probably add a system requirements section to the README where we can note the minimum required OpenGL version of `4.3`, any even slightly modern Linux system will support this, but it would be good to document it somewhere user-facing anyway.

# TODO BEFORE MERGE
- [x] Have multiple people test this on both macOS and linux.
- [ ] ~~Have someone with a better dev setup on linux check for memory leaks and other problems.~~ (Skipped, will merge and let tip users figure this out, someone should *specifically* look for memory leaks before the next versioned release though.)
- [x] Address any code review feedback.


Chiramisudo:
Very pleased at the TODO list. Wish I could help out with the Linux testing, but it's a bit above my pay grade at this point. Great work though. This was a massive effort. That much is obvious! 👏🏽

chardskarth:
I keep getting back to this PR to see if this is already merged 🤣. 

I'm just too excited to have this awesome cursor animation along with other shaders. 👏

src/crash/sentry.zig
```diff
@@ -81,6 +81,13 @@ pub fn init(gpa: Allocator) !void {
 fn initThread(gpa: Allocator) !void {
     if (comptime !build_options.sentry) return;
 
+    // Right now, on Darwin, `std.Thread.setName` can only name the current
+    // thread, and we have no way to get the current thread from within it,
+    // so instead we use this code to name the thread instead.
+    if (builtin.os.tag.isDarwin()) {
+        internal_os.macos.pthread_setname_np(&"sentry-init".*);
```
mitchellh, line 88: We may want to pull this out into a helper in `src/os` so we can do something like `internal_os.thread.setName("sentry-init")` and let it handle the cross-platform stuff.

This way we can use `Thread.setName` if/when it fixes itself, or alternate options on other platforms more readily.

src/renderer/Thread.zig
```diff
@@ -307,6 +324,16 @@ fn stopDrawTimer(self: *Thread) void {
...
 fn drainMailbox(self: *Thread) !void {
+    // There's probably a more elegant way to do this...
+    //
+    // This is effectively an @autoreleasepool{} block, which we need in
+    // order to ensure that autoreleased objects are properly released.
+    const pool = if (builtin.os.tag.isDarwin())
+        @import("objc").AutoreleasePool.init()
+    else
+        void;
+    defer if (builtin.os.tag.isDarwin()) pool.deinit();
```
mitchellh, line 335: Doing this for every possible mailbox message feels bad, but I can see why you went this route. I suppose we have no messages that are so expensive that this becomes expensive. It's probably something to keep an eye on though. I'm going to think about this one a bit more as I review the rest of the PR.

src/renderer/Thread.zig
```diff
@@ -425,7 +452,7 @@ fn drainMailbox(self: *Thread) !void {
                 self.renderer.markDirty();
             },
 
-            .resize => |v| try self.renderer.setScreenSize(v),
+            .resize => {}, //|v| try self.renderer.setScreenSize(v),
```
mitchellh: If this isn't necessary anymore we can remove the message entirely.

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+    const shaderpkg = GraphicsAPI.shaders;
+
+    const cellpkg = GraphicsAPI.cellpkg;
+    const imagepkg = GraphicsAPI.imagepkg;
+    const Image = imagepkg.Image;
+    const ImageMap = imagepkg.ImageMap;
+
+    const Shaders = shaderpkg.Shaders;
+
+    const ImagePlacementList = std.ArrayListUnmanaged(imagepkg.Placement);
```
mitchellh: It's probably all fixed now given that tests pass, but there have periodically been Zig instabilities with capturing these consts for the type in a comptime function outside the structure. 

To be safe, I'd feel better if we moved them all into the structure below.

src/renderer/Metal.zig
```diff
@@ -581,2768 +114,236 @@ pub fn init(alloc: Allocator, options: renderer.Options) !Metal {
...
-    }
-    self.image_placements.deinit(self.alloc);
-
-    self.deinitShaders();
+    self.queue.release();
+    self.device.release();
 
-    self.* = undefined;
+    // NOTE: We don't release the layer here because that should be taken
+    //       care of automatically when the hosting view is destroyed.
```
mitchellh: Thanks for writing this comment. It makes sense that the hosting view would retain/release its `layer`, but I'm surprised that it doesn't `retain` when we set the property so that we can release our reference. I just want to double check your note here is correct or if you have more details about it.

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+        /// Swap chain which maintains multiple copies of the state needed to
+        /// render a frame, so that we can start building the next frame while
+        /// the previous frame is still being processed on the GPU.
+        const SwapChain = struct {
+            // The count of buffers we use for double/triple buffering.
+            // If this is one then we don't do any double+ buffering at all.
+            // This is comptime because there isn't a good reason to change
+            // this at runtime and there is a lot of complexity to support it.
+            const buf_count = count: {
+                if (@hasDecl(GraphicsAPI, "swap_chain_count")) {
```
mitchellh: I'm doing my best to move away from conditional `hasDecl` where possible because it makes the "comptime interface" more mysterious and implicit. It does make sense in some cases where the behavior is super rare, but I'd lean towards avoiding by default.

In this case, I think we should make `swap_chain_count` a required decl on GraphicsAPI to ensure our renderer authors think about it.

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+            frames: [buf_count]FrameState,
+            /// Index of the most recently used frame state struct.
+            frame_index: std.math.IntFittingRange(0, buf_count) = 0,
+            /// Semaphore that we wait on to make sure we have an available
+            /// frame state struct so we can start working on a new frame.
+            frame_sema: std.Thread.Semaphore = .{ .permits = buf_count },
+
+            /// Set to true when deinited, if you try to deinit a defunct
+            /// swap chain it will just be ignored, to prevent double-free.
+            defunct: bool = false,
```
mitchellh, line 224: Is this just for runtime safety? I tried spying how this could ever happen under normal scenarios and don't see it.

It's a very lightweight thing for a lot of safety so I'm happy to keep it, just want clarification.

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+            pub fn init(api: GraphicsAPI, custom_shaders: bool) !FrameState {
+                // Uniform buffer contains exactly 1 uniform struct. The
+                // uniform data will be undefined so this must be set before
+                // a frame is drawn.
+                var uniforms = try UniformBuffer.init(api.uniformBufferOptions(), 1);
+                errdefer uniforms.deinit();
+
+                // Create the buffers for our vertex data. The preallocation size
+                // is likely too small but our first frame update will resize it.
+                var cells = try CellTextBuffer.init(api.fgBufferOptions(), 10 * 10);
```
mitchellh: If this is likely too small and we're going to resize anyways, is there any reason to allocate 100 elements to begin with? Why not just do `1`. _Really_, we should avoid the allocation completely since it just forces a double-allocation on every new path anyways but the cost of that is probably ignorable.

Not a big deal, but just seems funny to even put 100 here.

src/renderer/Metal.zig
```diff
@@ -581,2768 +114,236 @@ pub fn init(alloc: Allocator, options: renderer.Options) !Metal {
...
-    encoder.msgSend(
-        void,
-        objc.sel("setFragmentBuffer:offset:atIndex:"),
-        .{ frame.uniforms.buffer.value, @as(c_ulong, 0), @as(c_ulong, 2) },
-    );
-
-    encoder.msgSend(
-        void,
-        objc.sel("drawIndexedPrimitives:indexCount:indexType:indexBuffer:indexBufferOffset:instanceCount:"),
+    return Texture.init(
```
mitchellh: This should be `return try Texture.init(` I think. Same with OpenGL.

src/renderer/metal/Texture.zig
```diff
@@ -0,0 +1,196 @@
...
+/// Bytes per pixel for this texture.
+bpp: usize,
+
+/// Initialize a texture
+pub fn init(
+    opts: Options,
+    width: usize,
+    height: usize,
+    data: ?[]const u8,
+) !Self {
```
mitchellh: Pretty sure this can have an explicit error set. Eyeballed it down and it looks clean.

src/renderer/OpenGL.zig
```diff
@@ -1,450 +1,171 @@
...
+/// Alpha blending mode
+blending: configpkg.Config.AlphaBlending,
 
-    var gl_state = try GLState.init(alloc, options.config, grid);
-    errdefer gl_state.deinit();
+/// The most recently presented target, in case we need to present it again.
+last_target: ?Target = null,
 
-    return OpenGL{
+pub fn init(alloc: Allocator, opts: rendererpkg.Options) !OpenGL {
```
mitchellh: I think this can trivially have a defined error set too. No need to dive all the way into `pkg/opengl`, can just set it here.

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+            var font_shaper = try font.Shaper.init(alloc, .{
+                .features = options.config.font_features.items,
+            });
+            errdefer font_shaper.deinit();
+
+            // Initialize all the data that requires a critical font section.
+            const font_critical: struct {
+                metrics: font.Metrics,
+            } = font_critical: {
+                const grid = options.font_grid;
```
mitchellh: I know this is copied from old code so its not your fault, but we should put an explicit type here because we want to be sure we're always getting a pointer and never a copy (if we ever change the options types).

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+        }
+
+        /// Update the frame data.
+        pub fn updateFrame(
+            self: *Self,
+            surface: *apprt.Surface,
+            state: *renderer.State,
+            cursor_blink_visible: bool,
+        ) !void {
+            _ = surface;
```
mitchellh: Now that this is all unified we can probably remove this paramater all the way up.

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+        pub fn drawFrame(
+            self: *Self,
+            sync: bool,
+        ) !void {
+            // We hold a the draw mutex to prevent changes to any
+            // data we access while we're in the middle of drawing.
+            self.draw_mutex.lock();
+            defer self.draw_mutex.unlock();
+
+            // There's probably a more elegant way to do this...
```
mitchellh: We may want to introduce a `drawFrameStart()/drawFrameEnd()` `GraphicsAPI` function so that we can defer stuff like this to the implementation rather than in generic.

src/renderer/generic.zig
```diff
@@ -0,0 +1,2873 @@
...
+                size_changed or
+                self.cells_rebuilt or
+                self.hasAnimations() or
+                sync;
+
+            if (!needs_redraw) {
+                // We still need to present the last target again, because the
+                // apprt may be swapping buffers and display an outdated frame
+                // if we don't draw something new.
+                try self.api.repeat();
```
mitchellh: As an API name, I don't love this, its unclear what needs to be repeated. I looked at both implementations and the comments and wonder if `presentLastFrame` or something like that would be better here.

mitchellh (changes requested):
Alright, first big review being submitted here. Its not difficult feedback, so _very well done_ 👏 .

I'm going to continue to take another pass at this, but overall I'm happy with what this looks like, the shape of things. I can see room for improvement but its mostly due to the legacy you adopted and whats possible now with the new abstractions, so let's defer to later and smaller PRs.

src/renderer/Thread.zig
qwerasd205: Oh yeah, I completely forget about this actually, I commented it out temporarily because I wasn't sure we needed it... and I guess we don't! Noted.

src/renderer/Metal.zig
qwerasd205: I was surprised by this too! But I was getting a double-free when closing the window, and I checked the memory graph in xcode to make sure that this wasn't a leak if we didn't do it and it didn't seem to be a leak afaict.

src/renderer/generic.zig
qwerasd205: In the unrealize handler we  the swapchain, if the unrealize handler runs and then the graphics API is deinited, that would double-free

src/renderer/generic.zig
qwerasd205: I believe I copied this from the existing code, I agree that we may as well start with 1 like with the 1x1 textures

src/renderer/Thread.zig
qwerasd205: Ah right, we do need this to get the padding, but the other aspects (grid size and screen size) are determined based on the copied pagelist and provided by the renderer (respectively) - so this just needs to be uncommented. Or do you think it should be replaced with something that directly provides just the padding?

src/crash/sentry.zig
qwerasd205: Well, the thing is that macOS's `pthread_setname_np` refers to the thread it runs on, Linux and other platforms require a reference to the thread to set the name of it, and `std.Thread` has no function for acquiring the current thread. Right now all other platforms get their thread names set from the thread that launched the thread, not the launched thread itself. Unifying it would require writing cross-platform code for getting the current thread which seems like a big pain to do and to maintain; it should be a part of `std.Thread`.

mitchellh: Right I understand that. But I think putting it all in one place would let us unify this comment to one place and make that value call if we want to do the complex thing or not later. We can always remove it. Ultimately will defer to you, this isn’t terrible it just felt repetitive.

src/renderer/Thread.zig
qwerasd205: Resolved by re-enabling this handling- we can switch it to only send the padding in the future.

src/renderer/Metal.zig
qwerasd205: Resolved by manually retaining and then releasing the layer to avoid the un-zig-ish implicit autorelease behavior.

src/renderer/generic.zig
qwerasd205: Added clarification to this doc comment about that.

src/renderer/generic.zig
qwerasd205: Reduced size to 1, also reduced the initial atlas texture sizes to 1 since they'll be resized anyway

src/renderer/Thread.zig
qwerasd205: Because I placed an AutoreleasePool in `drawFrame` and that's pretty much the only spot we actually need it (since it's one of the only spots allowed to interact with GPU resources) it may be the case that this can be safely removed. Tomorrow night I'll look in to removing it and check for any memory leaks.

On the other hand, having this here does just feel like a very solid safety net, making sure any code that might interact with objc objects is inside an autorelease block, so idk.

mitchellh (approved):
I don't see any missing functionality and as I said on Discord, I don't think I or anyone else is going to be able to review this for correctness completely, so I think we should just ship it into the tip channel and shake any issues out on main. 

I did trace the execution all the way through all the changes and I didn't spot anything. It all looks good!

Great job. 👏

mitchellh (approved):

mitchellh:
@qwerasd205 Merge when you're ready. 😄

src/crash/sentry.zig
qwerasd205: I think I'll probably leave it and fix it up at some point in the future so that this can be merged.

src/renderer/Thread.zig
qwerasd205: I'll actually look in to this later, and for now I'll merge it like it is.

PR #7634, opened 2025-06-20
macOS: App Intents

mitchellh:
This PR integrates Ghostty on macOS with the [App Intents](https://developer.apple.com/documentation/appintents) system. The focus of this initial work was on enabling [Apple Shortcuts](https://support.apple.com/guide/shortcuts/welcome/ios) on macOS, but App Intents are the same underlying system that powers a number of other Apple features such as Spotlight, Siri, Widgets, and more. We don't do much with these latter ones yet, though.

Additionally, this PR begins to refactor and untangle some of our libghostty API calls from macOS views. Presently, macOS views and view controllers directly call into the libghostty API and own libghostty data models. This tight coupling is kind of nasty because it tends to also couple libghostty API calls to the main GUI thread when they don't really have to be (they just have to not be concurrently accessed). This becomes an issue because App Intents run on background threads. This PR starts to extract out some of this business logic into standalone classes, but we still force all execution onto the main thread during the transition.

**Version requirement:** Most of the shortcuts will work on macOS 13, but there are some that require macOS 14, and some functionality will require macOS 26. We gracefully degrade in all scenarios (the capabilities that are unavailable just don't show up on older systems).

> [!IMPORTANT]
>
> This bumps our build requirements on macOS to Xcode 26 and the macOS 26 SDK. **You can still build on macOS 15,** but you must be using the Xcode 26 beta. This includes a README update about that. 

## Why?

Apple Shortcuts is an extremely powerful scripting tool on Apple platforms. It comes with a number of built-in capabilities such as moving windows, taking screenshots, fetching secrets, and more. By integrating with Apple Shortcuts, it allows Ghostty to become scriptable to a certain extent while also being able to take advantage of this large ecosystem.

It is a huge downside that Shortcuts is Apple-only and I still would like to make Ghostty scriptable to some extent on Linux and other platforms as well. This work doesn't preclude that goal, but gives us an answer to a large subset of users (macOS users), which is great.

Beyond this, no terminals integrate with Apple Shortcuts except the built-in Terminal. And even then, the built-in Terminal only exposes two actions (run script and run script over SSH). I think there's a lot that can be done by exposing more functionality and I'm excited to see what people do with this.

Finally, I think Shortcuts is possibly a way we can do some GUI testing on macOS. That remains to be explored but it seems promising. 

## Capability

The initial set of Shortcut actions is shown in the screenshot below:

CleanShot 2025-06-20 at 12 19 47@2x

These can be combined with the built-in shortcuts to do some pretty interesting things.

### Future

There are more capabilities I'd like to expose, but they require changing core parts of Ghostty that I didn't want to mix into this PR.

## Security

Scripting Ghostty can be considered a security risk, since it allows arbitrary command execution, reading terminal output, etc. Therefore, Ghostty will ask for permission prior to allowing any Shortcut to remote control it:

image

This can be directly overridden using the new `macos-shortcuts` configuration, which defaults to `ask` but can also be set to `deny` or `allow` with self-explanatory behaviors.

### DIFF
--- a/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
+++ b/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
@@ -55,6 +55,11 @@ struct TerminalEntity: AppEntity {
         Self.defaultQuery.all.first { $0.uuid == self.id }
     }
 
+    @MainActor
+    var surfaceModel: Ghostty.Surface? {
+        surfaceView?.surfaceModel
+    }
+
     static var defaultQuery = TerminalQuery()
 
     init(_ view: Ghostty.SurfaceView) {
--- a/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
+++ b/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
@@ -11,6 +11,9 @@ struct TerminalEntity: AppEntity {
     @Property(title: "Working Directory")
     var workingDirectory: String?
 
+    @Property(title: "Kind")
+    var kind: Kind
+
     @MainActor
     @DeferredProperty(title: "Full Contents")
     @available(macOS 26.0, *)
@@ -67,6 +70,27 @@ struct TerminalEntity: AppEntity {
         self.title = view.title
         self.workingDirectory = view.pwd
         self.screenshot = view.screenshot()
+        
+        // Determine the kind based on the window controller type
+        if view.window?.windowController is QuickTerminalController {
+            self.kind = .quick
+        } else {
+            self.kind = .normal
+        }
+    }
+}
+
+extension TerminalEntity {
+    enum Kind: String, AppEnum {
+        case normal
+        case quick
+        
+        static var typeDisplayRepresentation = TypeDisplayRepresentation(name: "Terminal Kind")
+        
+        static var caseDisplayRepresentations: [Self: DisplayRepresentation] = [
+            .normal: .init(title: "Normal"),
+            .quick: .init(title: "Quick")
+        ]
     }
 }
 
@@ -101,7 +125,8 @@ struct TerminalQuery: EntityStringQuery, EnumerableEntityQuery {
 
     @MainActor
     var all: [Ghostty.SurfaceView] {
-        // Find all of our terminal windows (includes quick terminal)
+        // Find all of our terminal windows. This will include the quick terminal
+        // but only if it was previously opened.
         let controllers = NSApp.windows.compactMap {
             $0.windowController as? BaseTerminalController
         }
--- a/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
+++ b/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
@@ -130,10 +130,10 @@ struct TerminalQuery: EntityStringQuery, EnumerableEntityQuery {
         let controllers = NSApp.windows.compactMap {
             $0.windowController as? BaseTerminalController
         }
-
+        
         // Get all our surfaces
-        return controllers.reduce([]) { result, c in
-            result + (c.surfaceTree.root?.leaves() ?? [])
+        return controllers.flatMap {
+            $0.surfaceTree.root?.leaves() ?? []
         }
     }
 }
--- a/macos/Sources/Features/App Intents/GhosttyIntentError.swift
+++ b/macos/Sources/Features/App Intents/GhosttyIntentError.swift
@@ -1,11 +1,13 @@
 enum GhosttyIntentError: Error, CustomLocalizedStringResourceConvertible {
     case appUnavailable
     case surfaceNotFound
+    case permissionDenied
 
     var localizedStringResource: LocalizedStringResource {
         switch self {
         case .appUnavailable: return "The Ghostty app isn't properly initialized."
         case .surfaceNotFound: return "The terminal no longer exists."
+        case .permissionDenied: return "Ghostty doesn't allow Shortcuts."
         }
     }
 }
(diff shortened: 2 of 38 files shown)


macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @MainActor
+    func entities(matching string: String) async throws -> [TerminalEntity] {
+        return all.filter {
+            $0.title.localizedCaseInsensitiveContains(string)
+        }.map {
+            TerminalEntity($0)
+        }
+    }
```
jparise, line 114: I think both of these cases could be slightly improved using either a `lazy` collection (`all.lazy.filter.map`) or `compactMap` to avoid extra iterations.

macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
```diff
@@ -0,0 +1,139 @@
...
+        // Find all of our terminal windows. This will include the quick terminal
+        // but only if it was previously opened.
+        let controllers = NSApp.windows.compactMap {
+            $0.windowController as? BaseTerminalController
+        }
+
+        // Get all our surfaces
+        return controllers.reduce([]) { result, c in
+            result + (c.surfaceTree.root?.leaves() ?? [])
+        }
```
jparise:
```suggestion
        return controllers.flatMap { 
            $0.surfaceTree.root?.leaves() ?? []
        }
```

macos/Sources/Features/App Intents/NewTerminalIntent.swift
```diff
@@ -0,0 +1,163 @@
...
+        if let url = workingDirectory?.fileURL {
+            let dir = url.hasDirectoryPath ? url : url.deletingLastPathComponent()
+            config.workingDirectory = dir.path(percentEncoded: false)
+        }
+
+        // Parse environment variables from KEY=VALUE format
+        for envVar in env {
+            if let separatorIndex = envVar.firstIndex(of: "=") {
+                let key = String(envVar[..<separatorIndex])
+                let value = String(envVar[envVar.index(after: separatorIndex)...])
```
jparise, line 75: Alternatively:
```suggestion
                let key = String(envVar.prefix(upTo: separatorIndex))
                let value = String(envVar.suffix(from: envVar.index(after: separatorIndex)))
```

Just FYI in case folks find the range-based syntax less approachable. (My suggestion would be even nicer if there was a `suffix(after:)`, which I guess we could provide if we did more of these operations.)

macos/Sources/Features/App Intents/GhosttyIntentError.swift
```diff
@@ -0,0 +1,13 @@
+enum GhosttyIntentError: Error, CustomLocalizedStringResourceConvertible {
+    case appUnavailable
+    case surfaceNotFound
+    case permissionDenied
+
+    var localizedStringResource: LocalizedStringResource {
+        switch self {
+        case .appUnavailable: return "The Ghostty app isn't properly initialized."
+        case .surfaceNotFound: return "The terminal no longer exists."
+        case .permissionDenied: return "Ghostty doesn't allow Shortcuts."
```
jparise: I _think_ the `return` keywords are optional in this context.

macos/Sources/Helpers/Extensions/Array+Extension.swift
```diff
@@ -21,3 +21,28 @@ extension Array {
...
+        // Handle empty array
+        if isEmpty {
+            return try body([])
+        }
+
+        // Recursive helper to process strings
+        func helper(index: Int, accumulated: [UnsafePointer<Int8>?], body: ([UnsafePointer<Int8>?]) throws -> T) rethrows -> T {
+            if index == count {
+                return try body(accumulated)
+            } else {
```
jparise: Could remove the `else` (or switch to a `guard`) to avoid nesting the next branch given they both `return`.

jparise (approved):
Looks great!

`IntentDescription` has an optional [`categoryName`](https://developer.apple.com/documentation/appintents/intentdescription/categoryname) argument that the Shortcuts UI will use to categorize (+ group) intents. There's also [`searchKeywords`](https://developer.apple.com/documentation/appintents/intentdescription/searchkeywords), which I haven't used before.

Not sure if you looked at these but might be useful given how many we offer.

macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
mitchellh: There's never going to be so many terminals that this will ever matter. Even if there were 1,000 terminals it'd be fast, so I don't think we need to worry about it...

macos/Sources/Features/App Intents/GhosttyIntentError.swift
mitchellh: They are good catch. I did find out (via `Ghostty.Input`) that if the switch becomes large enough that this implicit return doesn't work anymore though lol.

macos/Sources/Helpers/Extensions/Array+Extension.swift
mitchellh: Thanks, this was AI slop lol

mitchellh:
> IntentDescription has an optional [categoryName](https://developer.apple.com/documentation/appintents/intentdescription/categoryname) argument that the Shortcuts UI will use to categorize (+ group) intents. There's also [searchKeywords](https://developer.apple.com/documentation/appintents/intentdescription/searchkeywords), which I haven't used before.

Both sound good but not sure what I'd put as the category. Search keywords we can also continue to think of.

mitchellh:
Based on various feedback, I modified the `ask` behavior to ask once and remember forever by default. This is more in line with how Apple permission dialogs work. Additionally, many people suggested that Shortcuts itself is responsible for protecting users from malicious scripts so I shouldn't bear this upon myself. I think asking once is a fair line between these two.

PR #7645, opened 2025-06-22
Improves the click-to-move cursor functionality when command is too long

liby:
Partially fixes #2582
## Background

The current click-to-move cursor implementation uses up/down arrow keys when navigating between lines.
This causes issues in some shells:

1. In fish shell, up/down arrows trigger history search (up-or-search/down-or-search), preventing proper
cursor movement
2. For long wrapped commands, the vertical navigation often fails to position the cursor correctly
3. The user experiences either no movement or unintended command history navigation

## Solution

This change modifies the cursor movement strategy to use only left/right arrow keys:

- Calculate the linear character distance between current and target positions
- Move the cursor using only left/right arrows, treating long input as a continuous stream
- This avoids shell-specific key bindings that interfere with cursor positioning

## Testing

- Long commands that naturally wrap to multiple lines (now working correctly)
- Commands with line continuations using `\` (has positioning issues - see limitations)
- Multiple shells: zsh, bash, fish

https://github.com/user-attachments/assets/1d823cba-690e-4908-9ef3-58f36c187612

## Limitations

This approach has a known limitation with line continuation commands (using `\`). For example:

```bash
❯ echo "long command" \
    continuation line
```

The positioning may be incorrect because the linear distance calculation (`y * cols + x`) assumes all lines
are full width, but:
- Lines with `\` may end before reaching the terminal edge
- Continuation lines often have indentation

I'm open to feedback on this approach and happy to explore alternative solutions.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3350,38 +3350,35 @@ fn clickMoveCursor(self: *Surface, to: terminal.Pin) !void {
     // move the cursor so we can fast path out of here.
     if (!t.flags.shell_redraws_prompt) return;
 
-    // Get our path
+    // Get the current cursor position and target position
     const from = t.screen.cursor.page_pin.*;
-    const path = t.screen.promptPath(from, to);
-    log.debug("click-to-move-cursor from={} to={} path={}", .{ from, to, path });
+    log.debug("click-to-move-cursor from={} to={}", .{ from, to });
 
     // If we aren't moving at all, fast path out of here.
-    if (path.x == 0 and path.y == 0) return;
+    if (from.x == to.x and from.y == to.y) return;
 
-    // Convert our path to arrow key inputs. Yes, that is how this works.
-    // Yes, that is pretty sad. Yes, this could backfire in various ways.
-    // But its the best we can do.
+    // First, calculate the total character distance.
+    // We need to consider the terminal width for proper wrapping.
+    const cols = t.cols;
+    const from_pos = from.y * cols + from.x;
+    const to_pos = to.y * cols + to.x;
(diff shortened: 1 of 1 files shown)


pfgithub:
The way the builtin macos terminal seems to handle line continuations is by seeing that the previous line didn't wrap when printing and then using the up arrow in that case.

liby:
> The way the builtin macos terminal seems to handle line continuations is by seeing that the previous line didn't wrap when printing and then using the up arrow in that case.

I was curious about how the macOS Terminal handles this, so I just tested it myself. Interestingly, it seems like the macOS Terminal also has issues with line continuations using `\` - so it looks like this is a fundamentally challenging problem rather than just a limitation of our current approach.

https://github.com/user-attachments/assets/121b3e1c-6e34-49f1-8d0c-bc5d5ac442df

The core issue is that terminals operate at the visual/character level, while line continuations are a shell syntax concept. The terminal doesn't have a reliable way to distinguish between:

1. Natural line wrapping
2. Manual line continuations with `\` (where the line lengths are irregular due to the \ placement and continuation prompt indentation)

If there are specific scenarios where this approach fails badly, we could consider adding configuration options or further refinements, but for now this seems like a solid improvement over the previous vertical navigation approach.

mitchellh:
Thanks for the PR, the change is simple enough, but I think I want to see a more detailed analysis of how other terminals handle this in a variety of cases. After that, we should pair it with a set of unit tests that recreate that terminal state and ensure we get the expected directional movements back (this is how the current path functions are done in `Screen.zig). I don't think we should YOLO this in Surface.zig.

PR #7648, opened 2025-06-22
Custom Shader Cursor Uniforms

qwerasd205:
Supersedes #6912, implements #6901

Also included in this PR is a fix/cleanup of the custom shader uniform handling, moved the CPU-side custom shader uniforms struct to the main renderer struct instead of having it be per-frame, moved the layout struct to `shadertoy.zig` since it has the `std140` layout for both backends.

Also, I added the current/previous cursor colors to the uniforms, since I figured they'd be useful to have and it's a trivial addition.

### Future Work
- This extension to the shadertoy uniforms needs to be documented somewhere so it's discoverable by users.
- The flipped Y axis on Metal still needs to be fully addressed instead of just being patched over like it is right now.

### DIFF
--- a/src/font/sprite/cursor.zig
+++ b/src/font/sprite/cursor.zig
@@ -50,7 +50,11 @@ pub fn renderGlyph(
     const region = try canvas.writeAtlas(alloc, atlas);
 
     return font.Glyph{
-        .width = width,
+        // HACK: Set the width for the bar cursor to just the thickness,
+        //       this is just for the benefit of the custom shader cursor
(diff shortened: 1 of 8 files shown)


mitchellh (approved):
Looks good. 

For docs, I think we should add it to the `custom-shader` config and exhaustively list the uniforms we provide (including the Shadertoy ones) just so its all there for people.

mitchellh (approved):

GeneralRealYC:
how to use it?

00-kat:
PRs are not for support; see #6901.

PR #7650, opened 2025-06-22
i18n: Add Irish (ga_IE) translation

aindriu80:
This PR introduces support for Irish (Gaeilge), the first official language of Ireland and an EU working language.

The translation file was initialized using the standard gettext tooling:

` msginit -i po/com.mitchellh.ghostty.pot -l $LANG -o "po/$LANG.po"`

The locale code `ga_IE` follows ISO standards for Gaeilge as spoken in Ireland.

I'm happy to volunteer as the ongoing maintainer of the Irish translation and will keep it up to date as Ghostty evolves.

Go raibh maith agat!

### DIFF
--- a/po/ga_IE.UTF-8.po
+++ b/po/ga_IE.UTF-8.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2025-04-23 16:58+0800\n"
-"PO-Revision-Date: 2025-06-24 12:42+0100\n"
+"PO-Revision-Date: 2025-06-29 21:15+0100\n"
 "Last-Translator: Aindriú Mac Giolla Eoin <aindriu80@gmail.com>\n"
 "Language-Team: Irish <gaeilge-gnulinux@lists.sourceforge.net>\n"
 "Language: ga\n"
@@ -16,7 +16,7 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=3; plural=n==1 ? 0 : n==2 ? 1 : 2;\n"
-"X-Generator: Poedit 3.4.2\n"
+"X-Generator: Poedit 3.4.4\n"
 
 #: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:5
 msgid "Change Terminal Title"
@@ -212,7 +212,7 @@ msgid ""
 "commands may be executed."
 msgstr ""
 "D’fhéadfadh sé a bheith contúirteach an téacs seo a ghreamú isteach sa "
-"chríochfort mar is cosúil go bhféadfaí roinnt orduithe a fhorghníomhú."
+"teirminéal, toisc go d'fhéadfadh roinnt orduithe a fhorghníomhú."
 
 #: src/apprt/gtk/Window.zig:208
 msgid "Main Menu"
@@ -234,7 +234,7 @@ msgstr ""
 
 #: src/apprt/gtk/Window.zig:765
 msgid "Reloaded the configuration"
-msgstr "Athluchtaigh an chumraíocht"
+msgstr "Tá an chumraíocht athlódáilte"
 
 #: src/apprt/gtk/Window.zig:1005
 msgid "Ghostty Developers"
--- a/po/ga_IE.UTF-8.po
+++ b/po/ga_IE.UTF-8.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2025-04-23 16:58+0800\n"
-"PO-Revision-Date: 2025-06-23 09:00+0100\n"
+"PO-Revision-Date: 2025-06-24 09:21+0100\n"
 "Last-Translator: Aindriú Mac Giolla Eoin <aindriu80@gmail.com>\n"
 "Language-Team: Irish <gaeilge-gnulinux@lists.sourceforge.net>\n"
 "Language: ga\n"
@@ -156,7 +156,7 @@ msgstr "Oscail cumraíocht"
 
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:85
 msgid "Command Palette"
-msgstr "Pailéad ordú"
+msgstr "Pailéad ordaithe"
 
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:90
 msgid "Terminal Inspector"
--- a/po/ga_IE.UTF-8.po
+++ b/po/ga_IE.UTF-8.po
@@ -1,0 +1,286 @@
+# Irish translations for com.mitchellh.ghostty package.
+# Copyright (C) 2025 Mitchell Hashimoto, Ghostty contributors
+# This file is distributed under the same license as the com.mitchellh.ghostty package.
+# Aindriú Mac Giolla Eoin <aindriu80@gmail.com>, 2025.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: com.mitchellh.ghostty\n"
+"Report-Msgid-Bugs-To: m@mitchellh.com\n"
+"POT-Creation-Date: 2025-04-23 16:58+0800\n"
+"PO-Revision-Date: 2025-06-22 19:43+0100\n"
+"Last-Translator: Aindriú Mac Giolla Eoin <aindriu80@gmail.com>\n"
+"Language-Team: Irish <gaeilge-gnulinux@lists.sourceforge.net>\n"
+"Language: ga\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=n==1 ? 0 : n==2 ? 1 : 2;\n"
+"X-Generator: Poedit 3.4.2\n"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:5
+msgid "Change Terminal Title"
+msgstr "Athraigh Teideal Teirminéil"
+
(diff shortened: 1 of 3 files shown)


po/ga_IE.UTF-8.po
```diff
@@ -0,0 +1,286 @@
...
+"Language: ga\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=n==1 ? 0 : n==2 ? 1 : 2;\n"
+"X-Generator: Poedit 3.4.2\n"
+
+#: src/apprt/gtk/ui/1.5/prompt-title-dialog.blp:5
+msgid "Change Terminal Title"
+msgstr "Athraigh Teideal Teirminéil"
```
00-kat: I don't speak Irish, but does Irish feature title-case?  Most other translations seem to not use title-case (it's a peculiarity of English), so I want to ask if Irish does use title-case too or if you were simply copying the capitalisation of the English source.

00-kat (changes requested):
Please update this struct too! https://github.com/ghostty-org/ghostty/blob/0721955dde45bbb53c0af25eea503b825935bb1b/src/os/i18n.zig#L7-L48

And optionally also this file: https://github.com/ghostty-org/ghostty/blob/3e79c4b7eaf7c875001fd028df5188d10a6246ee/CODEOWNERS#L158-L175

po/ga_IE.UTF-8.po
aindriu80: Yes you are absolutely right,  I had initially followed the English title casing to keep things visually consistent, but in fact, sentence case is the natural and standard approach in Irish for UI text, so I went ahead and updated all strings to use sentence case accordingly. Go raibh maith agat!

po/ga_IE.UTF-8.po
```diff
@@ -124,12 +124,12 @@ msgstr "Cluaisín"
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:30
 #: src/apprt/gtk/Window.zig:255
 msgid "New Tab"
-msgstr "Cluaisín Nua"
+msgstr "Cluaisín nua"
```
OshDubh: The irish for tab in the context of computing is [táb](https://www.tearma.ie/id/3582990/).

po/ga_IE.UTF-8.po
```diff
@@ -152,15 +152,15 @@ msgstr "Cumraíocht"
 #: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:92
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:95
 msgid "Open Configuration"
-msgstr "Oscail Cumraíocht"
+msgstr "Oscail cumraíocht"
 
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:85
 msgid "Command Palette"
-msgstr "Pailéad Ordú"
+msgstr "Pailéad ordú"
```
OshDubh: I think "Pailéad ordaithe" would be a better fit, since "Command palette" is a "Palette (of) Commands"

OshDubh (commented):
Great quality translation, so cool to see it translated into Irish.
Just two comments, ach maith an obair! An-dheas.

po/ga_IE.UTF-8.po
aindriu80: I double-checked and I think the word used in this case is cluaisín, see picture here of Firefox - open new tab | oscail cluaisín nua.  

tab

I looked at the official dictionary.  Táb refers to 

 '3 noun COMP, PUBL on screen, keyboard táb' 

and not

'2 noun part of card, page, folder, etc cluaisín masc4' 



####  **"Táb"**:

* Is a **loanword**, increasingly common in computing contexts.
* It's short, mirrors English, and might be familiar to more technical users.
* However, it's **not standardised** in major resources like [tearma.ie](https://www.tearma.ie/) or [foclóir.ie](https://www.focloir.ie/en/dictionary/ei/tab).

####  **"Cluaisín"**:

* Is the **official term** according to [Téarma.ie](https://www.tearma.ie/Search.aspx?term=tab):
* It has been adopted in official localisations, e.g., Firefox, LibreOffice, and GNOME.
* It's more consistent with other UI terms that avoid English loanwords.

 "Cluaisín"** — it’s the standard Irish term and more widely accepted in FOSS localisation efforts.

po/ga_IE.UTF-8.po
aindriu80: Pailéad ordaithe looks like a better suggestion!

#### **"Pailéad ordaithe"**

* Uses the **genitive plural form** *ordaithe*, which can be translated as *"of commands"*
* Grammatically sound, and might be more idiomatic

#### **"Pailéad ordú"**

* More literal translation: *"command palette"*
* **"Ordú"** is singular (command), so it reads as *"palette of a command"* rather than *"of commands"*
* Might feel slightly off to fluent speakers

"Pailéad ordaithe" is more grammatically accurate and sounds more natural.

00-kat:
@OshDubh by the way, since you appear to be knowledgeable in Irish too, would you also be willing to help maintain these translations in the future?  (I'm asking now since Mitchell is almost certainly going to ask when he gets around to this PR.)

po/ga_IE.UTF-8.po
OshDubh: The tearma.ie "táb" I linked above (and focloir's definition) backs up that is the correct term in this context. The etymology for words that describe browser / terminal "tabs" comes directly from the tab key on the keyboard, since that was the key historically used to navigate between them. 

"Cluaisín" is a term that describes physical tabs like clothes labels or notebook dividers, literally "ears" or "things that stick out". It's only possible to use loanwords, since the word to describe this does not exist in Irish originally. Words like "táb", "cúrsóir" and other computing terms are part of the effort to keep Irish modern and able to describe technological words that can't be accurately described using existing words. 

It's possible that Firefox's translation also stems from the understandable assumption that they mean the same thing, or maybe to avoid borrowing, but it's not quite accurate.

OshDubh:
Sure, does that mean I get notified when there are new strings to be translated? 

I'm sure aindriu has it covered but I'm happy to be a backup haha. And yes I'm a native speaker 🙂

00-kat:
> Sure, does that mean I get notified when there are new strings to be translated?

Yeah.

po/ga_IE.UTF-8.po
aindriu80: I replaced the 6 strings with táb

00-kat (approved):
(I don't actually speak Irish so I can't review any of the actual translations, which is why I wasn't going to add a review, especially since it won't block the merge anyway.)

aindriu80:
> (I don't actually speak Irish so I can't review any of the actual translations, which is why I wasn't going to add a review, especially since it won't block the merge anyway.)

oh no problem, I wasn't too sure on the Pull Request process, so I pressed the review button there... I think all is ok my end, maybe @OshDubh wants to take another look

00-kat:
> I think all is ok my end

Yep, the only thing left is for Mitchell to see it (I think he's a bit busy lately so that may take a while).

aindriu80:
> > I think all is ok my end
> 
> Yep, the only thing left is for Mitchell to see it (I think he's a bit busy lately so that may take a while).

no rush, that's great! thanks very much :)

po/ga_IE.UTF-8.po
```diff
@@ -0,0 +1,286 @@
...
+
+#: src/apprt/gtk/Window.zig:319
+msgid ""
+"⚠️ You're running a debug build of Ghostty! Performance will be degraded."
+msgstr ""
+"⚠️ Tá leagan dífhabhtaithe de Ghostty á rith agat! Laghdófar an fheidhmíocht."
+
+#: src/apprt/gtk/Window.zig:765
+msgid "Reloaded the configuration"
+msgstr "Athluchtaigh an chumraíocht"
```
OshDubh: You use "athlódáil" elsewhere, perhaps this should be something like "Tá an chumraíocht athlódáilte."? (The configuration was reloaded")

I'm not quite sure how messages like these should be phrased, since using it like "Athlódáladh an chumraíocht" sounds awkward to me, but this could also be a valid translation.

po/ga_IE.UTF-8.po
```diff
@@ -0,0 +1,286 @@
...
+msgid "Warning: Potentially Unsafe Paste"
+msgstr "Rabhadh: Greamaigh a d'fhéadfadh a bheith neamhshábháilte"
+
+#: src/apprt/gtk/ui/1.5/ccw-paste.blp:7
+msgid ""
+"Pasting this text into the terminal may be dangerous as it looks like some "
+"commands may be executed."
+msgstr ""
+"D’fhéadfadh sé a bheith contúirteach an téacs seo a ghreamú isteach sa "
+"chríochfort mar is cosúil go bhféadfaí roinnt orduithe a fhorghníomhú."
```
OshDubh: Chríochfort is a terminal in the sense of public transport, for trains or airplanes. "teirminéal" is a loanword but I don't believe that "críochfort" is suitable here.

Additionally, "is cosúil" is a little awkward to me, since it's generally used for comparison. Maybe a more semantic translation would be one that highlights the possibility of dangerous characters being in the paste, with something like
```txt
D’fhéadfadh sé a bheith contúirteach an téacs seo a ghreamú isteach sa teirminéal, toisc go d'fhéadfadh roinnt orduithe a fhorghníomhú.
```

OshDubh (changes requested):
Yes, a fantastic translation overall, I just have two more comments. Caighdeánach isn't the dialect I speak natively, so its possible I'm wrong with my comments or there is maybe something I've missed, but it all looks great to me!

po/ga_IE.UTF-8.po
aindriu80: I updated to your suggested: Tá an chumraíocht athlódáilte.  Yes, I think verbal noun sounds more natural than a direct imperative

po/ga_IE.UTF-8.po
aindriu80: "Teirminéal" is an accepted loanword in Irish computing for terminal emulators and is much better than chríochfort, I made your suggested change which also removes the "is cosúil" which is used for similes or comparisons (upon checking).  The  “toisc go d’fhéadfadh…” flows better and sounds more natural!

OshDubh (approved):

Issue #7647, opened 2025-06-22
macOS: directly executed scripts should run via stdin


mitchellh:
From #7465 

Ghostty currently executes scripts dropped onto it or via the services menu directly as if `command` has been set. The problem with this is we don't run any login/profile scripts (zshrc, zprofile, etc.). 

On macOS, this seems unexpected: both iTerm2 and Terminal.app _do_ load these scripts. This works by starting your normal default command and sending `<script>; exit<newline>` via the pty initially. 

Gnome Console appears to do what Ghostty does today and runs commands directly. My profiles aren't loaded. 

Therefore, I'm marking this as macOS only: if a script is executed via dropping onto the icon, the services menu, or an App Intent, we should run it like Terminal/iTerm2 do.

This will require some core work since we don't have a way currently to specify some stdin text to send.

PR #7652, opened 2025-06-22
`input` configuration to pass input as stdin on startup

mitchellh:
This adds a new configuration `input` that allows passing either raw text or file contents as stdin when starting the terminal. 

The input is sent byte-for-byte to the terminal, so control characters such as `\n` will be interpreted by the shell and can be used to run programs in the context of the loaded shell. After the input is sent, it switches to `stdin` as usual.

Example: `ghostty --input="hello, world\n"` will start the your default shell, run `echo hello, world`, and then show the prompt

The `--input` configuration can be repeated to send multiple data concatenated directly to together. Values can be also be prefixed with `file:` to send a file contents, e.g. `file:data.txt`. File contents are limited to 10MB to prevent excessive memory usage and avoid malicious use. Beyond that, users should write their own script...

This is all in pursuit of #7647, but we're not there yet.

### DIFF
--- a/src/cli/args.zig
+++ b/src/cli/args.zig
@@ -414,7 +414,7 @@ pub fn parseIntoField(
     return error.InvalidField;
 }
 
-fn parseTaggedUnion(comptime T: type, alloc: Allocator, v: []const u8) !T {
+pub fn parseTaggedUnion(comptime T: type, alloc: Allocator, v: []const u8) !T {
     const info = @typeInfo(T).@"union";
     assert(@typeInfo(info.tag_type.?) == .@"enum");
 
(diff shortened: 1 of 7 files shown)


mitchellh:
Bikeshed note: I didn't name this `stdin` for a couple reasons.

1. It isn't replacing stdin. After the input is sent, stdin is still connected to the pty. I didn't want anyone confused that they'd be replacing stdin if they did something like `--stdin=file:/some/file`. 

2. `stdin` isn't super user friendly. There are a lot of terminal users who aren't very tech savvy. To be fair, they're probably not Ghostty users [yet]. But I wanted to consider that a bit when designing this. If someone was looking how to pass input to a script, this would show up easily in search.

3. In Apple Shortcuts, macOS's built-in terminal calls this same feature "Input". I couldn't find a similar feature on any Linux terminal. I suspect this just isn't that interesting on Linux because they don't do things like #7647

PR #7668, opened 2025-06-23
cli: +edit-config command to open the config file in $EDITOR

mitchellh:
This adds a new CLI `ghostty +edit-config`. This will open the config file in the user's specified `$EDITOR`. If Ghostty has never been configured, this will also create the initial config file with some default templated contents (the same as that which we introduced back in Ghostty 1.0.1 or something -- not new behavior here).

This is useful on its own because it will find the correct configuration path to open. If users are terminal users anyway (not a big stretch since this is a terminal app), this will allow them to easily edit config right away.

This is also forward looking: I want to replace our "Open Config" action to open a Ghostty window executing this command so that users can edit their config in a terminal editor. This has been heavily requested since forever (short of a full GUI settings editor, which is not ready yet). I don't do this in this PR but plan to in a future PR.

Even further forward looking: when we have an API, we can have `edit-config` auto-reload the config on exit. This isn't possible today but this is where we'd put that.

### DIFF
--- a/src/cli/edit_config.zig
+++ b/src/cli/edit_config.zig
@@ -1,0 +1,159 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const assert = std.debug.assert;
+const args = @import("args.zig");
+const Allocator = std.mem.Allocator;
+const Action = @import("action.zig").Action;
+const configpkg = @import("../config.zig");
+const internal_os = @import("../os/main.zig");
+const Config = configpkg.Config;
+
+pub const Options = struct {
+    pub fn deinit(self: Options) void {
+        _ = self;
+    }
+
+    /// Enables `-h` and `--help` to work.
+    pub fn help(self: Options) !void {
+        _ = self;
+        return Action.help_error;
+    }
+};
+
+/// The `edit-config` command opens the Ghostty configuration file in the
+/// editor specified by the `$VISUAL` or `$EDITOR` environment variables.
+///
+/// IMPORTANT: This command will not reload the configuration after
+/// editing. You will need to manually reload the configuration using the
+/// application menu, configured keybind, or by restarting Ghostty. We
+/// plan to auto-reload in the future, but Ghostty isn't capable of
+/// this yet.
+///
+/// The filepath opened is the default user-specific configuration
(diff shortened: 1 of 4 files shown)


src/cli/edit_config.zig
```diff
@@ -0,0 +1,136 @@
...
+    const path = try Config.preferredDefaultFilePath(alloc);
+    defer alloc.free(path);
+
+    // We don't currently support Windows because we use the exec syscall.
+    if (comptime builtin.os.tag == .windows) {
+        try stderr.print(
+            \\ The `ghostty +edit-config` command is not supported on Windows.
+            \\ Please edit the configuration file manually at the following path:
+            \\
+            \\ {s}
```
tristan957: Would it be useful to print an OSC8 `file://` URI here?

src/cli/edit_config.zig
```diff
@@ -0,0 +1,136 @@
...
+        editorZ,
+        &.{ editorZ, pathZ },
+        std.c.environ,
+    );
+
+    // If we reached this point then exec failed.
+    try stderr.print(
+        \\ Failed to execute the editor. Error code={}.
+        \\
+        \\ This is usually due to the editor path not existing, invalid
```
tristan957:
```suggestion
        \\ This is usually due to the editor executable not existing, invalid
```

tristan957: Maybe a better word, but who knows

danielwe:
Note that the proper posix variable for this is actually `VISUAL`, while `EDITOR` is supposed to be a simple line editor. See https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference. The distinction in capabilities is hardly relevant in 2025, but lots of people (like me) have their editor configured through `VISUAL` rather than `EDITOR`, so functionality like this should probably support both, with `VISUAL` taking precedence if both are set.

(EDIT: double-checked my shell config and I actually set both of them to `nvim`, so I'll be fine either way 👍 )

mitchellh:
Wow, I've never even heard of VISUAL. That's crazy. Yeah, happy to check in that order.

PR #7671, opened 2025-06-23
Miscellaneous Renderer Cleanup

qwerasd205:
This PR is a collection of small improvements, cleanup, and fixes to the renderer, as a follow-up to #7620.

### Summary of changes
- The nearly identical-between-backends `cell.zig` and `image.zig` are now both unified.
- Shader pipeline preparation is now even DRYer in preparation for future changes that will involve adding new pipelines, where I don't want to be slowed down by having to add them in multiple places to get things working. There is now a single source of truth at the top of `shaders.zig` for what core pipelines are available and how they're configured.
- Global background color drawn in a separate step before individual cell background colors, this is required to be able to draw kitty images between these layers. Can't use the clear color for this because that would require color space conversions on the CPU-side which we don't yet have utilities for.
- Fixed a bug with the kitty image z-index logic where not having foreground images made the background images act as foreground images.
- Moved the custom shader uniform buffer to the frame state so we don't create a new one every frame like we were before.
- Fixed color glyphs under OpenGL being channel-swapped because the texture format was RGBA when the data was BGRA (#7670).


qwerasd205:
Oops forgot to update test code when I renamed that enum value haha

src/datastruct/array_list_pool.zig
```diff
@@ -0,0 +1,44 @@
...
+const Allocator = std.mem.Allocator;
+
+/// A pool of ArrayLists with methods for bulk operations.
+pub fn ArrayListPool(comptime T: type) type {
+    return struct {
+        const Self = ArrayListPool(T);
+        const ArrayListT = std.ArrayListUnmanaged(T);
+
+        // An array containing the lists that belong to this pool.
+        lists: []ArrayListT = &[_]ArrayListT{},
```
mitchellh: I think the only thing weirding me out about this structure is this default value. This would put this on the stack, no? If you copied this value this would be a big footgun and it feels not very Zig like. But I might just be tired and not thinking right about this.

src/datastruct/array_list_pool.zig
```diff
@@ -0,0 +1,44 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+/// A pool of ArrayLists with methods for bulk operations.
+pub fn ArrayListPool(comptime T: type) type {
```
mitchellh: To bikeshed, "Pool" doesn't feel like the right word, since pool as its used in other places in our codebase and more generally usually means a thing you can pull out of and put back (and, possibly, typically for performance). In this case, it's more of an `ArrayListCollection` or `ArrayListGroup` or something.

mitchellh: Separately, bikeshed aside, we can keep it specific like this, but I think we could make this generic over the type if we make a signature like this **pseudocode**:

```
pub fn do(comptime decl: []const u8, comptime args: anytype) void {
  for (self.list) |v| @call(.auto, @field(v, decl), args);
}
```

We could even reflect the return value and return an [allocated] slice of it if we wanted... It's kind of nasty but this structure is so simple I wonder if it'd be worth it eventually. Not sure if you have other plans here.

mitchellh (changes requested):
Mostly looks good, but a few things hear. I appreciate the follow ups, this is great work. 😄

src/datastruct/array_list_pool.zig
mitchellh: I should note, we absolutely do not need to do this generic thing for this PR. So this can be resolved, I'm just talking out loud about it. The other comments are more important.

src/datastruct/array_list_pool.zig
qwerasd205: It has a length of 0, nothing to copy, and it's allocated on the heap when actually used. That said, I'm not sure why it has a default value so we may be able to chuck it.

src/datastruct/array_list_pool.zig
qwerasd205: Yeah fair, pool is a bad name since that sounds like a re-use structure. To be clear this is an existing data structure that I just extracted verbatim out of the cell.zig file. I'm fine with renaming it `ArrayListCollection`

src/datastruct/array_list_pool.zig
mitchellh: Yeah, true. Just felt weird. I'm fine with it then.

mitchellh (approved):
Going to make sure this has an approval so you can merge it. Once the feedback is addressed I feel good about it. Thanks! I don't think I have the GH setting on that pushes after approval require re-approval, so hopefully this works.

src/datastruct/array_list_pool.zig
qwerasd205: Removed the default value, which was just there so it could be default initialized empty as `.{}` and instead did `.{ .lists = &.{} }` at the place where that's necessary.

src/datastruct/array_list_pool.zig
qwerasd205: Renamed to `ArrayListCollection`.

PR #7676, opened 2025-06-24
flatpak: remove references to systemd unit

alaviss:
Flatpak currently does not export systemd user units. As such, remove references to it from D-Bus services to prevent D-Bus daemon from trying to start a non-existent service.

Additionally, make sure that the D-Bus service name is correct for debug builds.

Follow up to #7433


alaviss:
Note that after #7433, this fix is required for Flatpak builds to run when launched from launcher.

mitchellh (approved):
Appreciate the follow up. Thanks.

PR #7677, opened 2025-06-24
font: faster glyph hashing

Rexicon226:
There are two main improvements being made here. First, we move away from using autohash and instead use a one-shot strategy similar to the Style hashing. Since the GlyphKey includes the Metrics struct, which contains quite a few fields, autohash was performing expensive and unnecessary repeated updates.

The second improvement is actually just, not hashing Metrics. By ignoring the Metrics field, we can fit the rest of the GlyphKey into a 64-bit packed struct and just return that as the hash! It ends up being unique for each GlyphKey in renderGlyph, and is nearly a zero-cost operation.

This ends up boosting the performance (on my machine at least), from around 560fps to 590fps on the DOOM-fire benchmark.

### DIFF
--- a/src/font/SharedGrid.zig
+++ b/src/font/SharedGrid.zig
@@ -40,7 +40,7 @@ const log = std.log.scoped(.font_shared_grid);
 codepoints: std.AutoHashMapUnmanaged(CodepointKey, ?Collection.Index) = .{},
 
 /// Cache for glyph renders into the atlas.
-glyphs: std.AutoHashMapUnmanaged(GlyphKey, Render) = .{},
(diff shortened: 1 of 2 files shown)


qwerasd205:
Since the grid metrics are always the same, this implies we should fully remove them from the `renderGlyph` `Options` struct entirely, if we don't have any places where we don't pass in the metrics that are already owned by the grid.

I'll look in to that in a follow-up to this, this looks good to me as an incremental improvement.

qwerasd205 (approved):

mitchellh (approved):
Approve pending CI.

mitchellh:
Thanks @Rexicon226. Appreciate you doing any performance work over here. 🙏

PR #7679, opened 2025-06-25
dbus and systemd activation - take 2

jcollie:
This replaces #7433. The improvements are:

1) Install the systemd user service in the proper directory depending on if it's a 'user' install or a 'system' install. This is controlled either by using the `--system` build flag (as most packages will) or by the `-Dsystem-package` flag.

2) Add the absolute path to the `ghostty` binary in the application file, the DBus service, and the systemd user service. This is done so that they do not depend on `ghostty` being in the `PATH` of whatever is launching Ghostty. That `PATH` is not necessarily the same as the `PATH` in a user shell (especially for DBus activation and systemd user services).

3) Adjust the DBus bus name that is expected by the system depending on the optimization level that Ghostty is compiled with.


jcollie:
@alaviss this will probably affect the flatpak

dist/linux/app.desktop
```diff
@@ -16,7 +18,8 @@ X-TerminalArgTitle=--title=
 X-TerminalArgAppId=--class=
 X-TerminalArgDir=--working-directory=
 X-TerminalArgHold=--wait-after-command
+DBusActivatable=true
 
 [Desktop Action new-window]
 Name=New Window
-Exec=ghostty
+Exec=ghostty --launched-from=desktop
```
alaviss: This one should use absolute path as well.

src/build/GhosttyResources.zig
```diff
@@ -224,16 +224,76 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyResources {
...
+                        switch (cfg.optimize) {
+                            .Debug, .ReleaseSafe => "-debug",
+                            .ReleaseFast, .ReleaseSmall => "",
+                        },
+                    },
+                ),
+            ),
+        );
+        // systemd user service
+        if (!cfg.flatpak)
```
alaviss: It might be better to add new systemd configuration option and use that to control this (and the associated dbus service patching). That should allow non-systemd distros like Alpine to package Ghostty without patching as well.

alaviss (commented):
Just a few notes. This looks good from Flatpak perspective to me.

dist/linux/app.desktop
jcollie: Ah, good catch! Fixed.

src/build/GhosttyResources.zig
```diff
@@ -224,16 +224,76 @@ pub fn init(b: *std.Build, cfg: *const Config) !GhosttyResources {
...
+                formatService(
+                    b,
+                    cfg,
+                    b.path("dist/linux/systemd.service"),
+                    b.fmt(
+                        "{s}/systemd/user/com.mitchellh.ghostty{s}.service",
+                        .{
+                            if (cfg.system_package) "lib" else "share",
+                            switch (cfg.optimize) {
+                                .Debug, .ReleaseSafe => "-debug",
```
tristan957: Is there any precedent for the `-debug`? Should we follow Flatpak here and use `.Debug`?

src/build/GhosttyResources.zig
jcollie: I sort of had that in the initial PR. Would Alpine even support DBus activation, which is the only way the systemd service would get started? I've never used it outside of a container so I don't know what sort of environment we're dealing with there.

tristan957 (commented):
I'm not convinced that using the full path is right.

I decided to check out my system to look for precedent:

```
$ rg -F -I -N --no-heading Exec= /usr/share/applications/ | grep -E 'Exec="?/' | wc -l
44

$ rg -F -I -N --no-heading Exec= /usr/share/applications/  | wc -l
166
```

~25% use the full path. I'm not as knowledgeable as you in this area, so will defer.

src/build/GhosttyResources.zig
jcollie: Yes, it's based on the DBus name Ghostty uses when compiled in debug mode so we don't have much of a choice.

alaviss: We could also just patch the used app-id, it's coded in only one place. But I have no issues with `-debug` suffix.

If we decide to patch that, I'd recommend `.Devel` or something similar, because `.Debug` is used for debug information package generated by flatpak.

src/build/GhosttyResources.zig
alaviss: You can run a desktop environment, which would include dbus-daemon/dbus-broker, both of which supports D-Bus activation.

An another distro of interest is Gentoo, where the no-systemd option is still supported.

dist/linux/app.desktop
```diff
@@ -1,13 +1,15 @@
 [Desktop Entry]
-Name=Ghostty
+Version=1.0
+Name=Ghostty@@NAME@@
```
pluiedev: The `@@` style substitutions are kinda cursed IMO. We should follow how most GTK/GNOME apps do it and just have separate desktop/metainfo/etc files for the release build and the debug build (cf. Ptyxis). It's also a bit misleading because usually by convention these "template" files that need substitution are marked with an `.in` suffix so that readers know that some variables are left unexpanded.

mitchellh: We need some kind of templating anyways. The `@@` doesn't bother me, but we should suffix this with `.in`.

src/build/Config.zig
```diff
@@ -87,7 +88,11 @@ pub fn init(b: *std.Build) !Config {
     // This is set to true when we're building a system package. For now
     // this is trivially detected using the "system_package_mode" bool
     // but we may want to make this more sophisticated in the future.
-    const system_package: bool = b.graph.system_package_mode;
+    const system_package = b.option(
```
mitchellh: For now I'd say nix the new option and continue to use `b.graph.system_package_mode`. That's what it's technically there for. 

If we need something to more carefully control the templating of the dist files we should consider whether we want to do something like this or a dedicated flag like you used to have, but I want to defer that to later.

src/build/GhosttyResources.zig
```diff
@@ -303,3 +363,36 @@ pub fn install(self: *const GhosttyResources) void {
     const b = self.steps[0].owner;
     for (self.steps) |step| b.getInstallStep().dependOn(step);
 }
+
+pub fn formatService(b: *std.Build, cfg: *const Config, src: std.Build.LazyPath, dest: []const u8) *std.Build.Step {
+    var cmd = b.addSystemCommand(&.{"sed"});
```
mitchellh: Be curious if this works on Mac (GNU vs BSD sed). I'd prefer to just read the file and do the string replacement myself in Zig, honestly, they're small files and the replacement text is easy.

I'm also curious if the Zig build system has something we can use like the C header formatting to do this for us.

mitchellh (changes requested):
Overall fine, there's a couple details I'm not super happy with. Happy to dig in myself but I won't be able to get back to this for probably... a day or two.

dist/linux/app.desktop
jcollie: I chose the "@@" because it's "traditional". It's a long established standard basically because "@@" is unlikely to appear in normal texts. Our needs for templating are very simple. If our needs ever get more complex we can look into any of the various templating engines that are available rather than trying to roll our own more sophisticated one.

jcollie: The "templated" files have been renamed with a `.in` suffix.

src/build/Config.zig
jcollie: Removed.

src/build/GhosttyResources.zig
jcollie: macOS isn't much of a problem (the templated files are never generated on anything but Linux) but FreeBSD (or other BSDs) might be since they'll need the desktop files as well. 

I just wrote a stupidly simple Zig program to do the templating so the GNU vs BSD sec issue is moot now.

src/build/GhosttyResources.zig
jcollie: Hmmm. Let's get this merged and then we can look into supporting no-systemd distros. Would be good if we had some real users that could help testing though.

mitchellh:
> I'm not convinced that using the full path is right.

I didn't do a full research on this yesterday to know the right answer, but I did find this which notes that systemd units don't do PATH lookup unless its specified in `Environment` (or a few other keys). I also found a stack overflow that says that systemd has a default PATH set to things like `/usr/bin` but it's limited. 

https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html

If we can remove the absolute path later I'll take it. For now I think this gets things working, but haven't verified yet.

mitchellh:
Okay I redid the templating part to use cmake format which lets us take advantage of `b.addConfigHeader` to produce our files without having to have a custom binary. A bit of a hack, but cmake formatting isn't going away anytime soon so this is pretty safe to do and I'd rather lean on Zig stdlib.

mitchellh:
Oh crap, Zig puts a header at the top that it was generated by the build system and it breaks everything. LOL.

mitchellh:
Okay may have to bring back our dedicated binary (rip). But I want to do it in the style of the other refactors I made so let me come back to this this afternoon. Ball is in my court.

tristan957:
> Okay I redid the templating part to use cmake format which lets us take advantage of `b.addConfigHeader` to produce our files without having to have a custom binary. A bit of a hack, but cmake formatting isn't going away anytime soon so this is pretty safe to do and I'd rather lean on Zig stdlib.

Let no good idea go unpunished! :'(

mitchellh:
Okay I kept the config header usage and just use `tail` to strip the first line. The usage of tail here is POSIX-standard only so this should work fine on BSD and Linux systems. It's something we can/should replace eventually but for all our target environments and platforms its pretty safe.

mitchellh (approved):

mitchellh:
Did auto-merge merge with failing tests? Computers aren't on my side with this one. Following up now.

PR #7683, opened 2025-06-25
config: make split/tab navigation keybinds performable

### LINKED ISSUE #7680: Make split/tab navigation keybinds performable by default
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7662

Currently our keybinds for navigating between splits and tabs (`next_tab`, `prev_tab`, etc.) are not performable by default, which makes Ghostty "swallow" keybinds like <kbd>Ctrl</kbd><kbd>PgUp</kbd> even when they're not consumed as an app-wide shortcut. They should be changed so that in case an action was not performed, the key input would then be passed into terminal programs.

pluiedev:
Fixes #7680

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4870,25 +4870,29 @@ pub const Keybinds = struct {
                 .{ .key = .{ .unicode = 'w' }, .mods = .{ .ctrl = true, .shift = true } },
                 .{ .close_tab = {} },
             );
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Just double checked and it looks like macOS is performable but also had some quirks here. I'll verify those and make an issue.

PR #7685, opened 2025-06-25
Add support for adjusting thickness of underline cursor

Aaron-212:
Resolve #7651

### DIFF
--- a/src/font/sprite.zig
+++ b/src/font/sprite.zig
@@ -32,6 +32,7 @@ pub const Sprite = enum(u32) {
     cursor_rect,
     cursor_hollow_rect,
     cursor_bar,
+    cursor_underline,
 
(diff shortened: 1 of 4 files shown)


qwerasd205 (changes requested):
A similar hack to the one for the bar cursor (in the returned `font.Glyph`) has to be implemented for proper animated cursor support.

Also, more worryingly, this unconditionally draws at the bottom of the cell, when it should be placed underneath the baseline by some amount, like with the actual underline.

It's possible my initial assessment of how to solve this wasn't correct, I'll have to think about this and may end up implementing the fix myself. Thank you for taking the time to do this, regardless.

qwerasd205:
This is superseded by the sprite face rework I decided to do in #7732 - thank you anyway for doing this!

Issue #7158, opened 2025-04-21
Ability to configure custom command palette entries


mitchellh:
We should add a configuration that can configure custom command palette entries. Since command palette entries must be keybinding actions we can reuse the keybind syntax for this:

```
command-palette-entry = title:Neovim, action:text:"nvim\n"
```

This syntax is just our already-existing struct syntax for config options.

pluiedev:
No longer specific to macOS following #7167

mitchellh:
Back to being macOS specific because GTK implemented this.

PR #7688, opened 2025-06-25
config: add `command-palette-entry` config option

pluiedev:
Implements #7158 for GTK

### DIFF
--- a/src/apprt/gtk/CommandPalette.zig
+++ b/src/apprt/gtk/CommandPalette.zig
@@ -104,7 +104,7 @@ pub fn updateConfig(self: *CommandPalette, config: *const configpkg.Config) !voi
     _ = self.arena.reset(.retain_capacity);
 
     // TODO: Allow user-configured palette entries
-    for (inputpkg.command.defaults) |command| {
+    for (config.@"command-palette-entry".value.items) |command| {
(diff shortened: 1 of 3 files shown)


qwerasd205:
Typo in test causing failure; capitalization of `Bobr` vs `bobr`.

Also, side note, by saying "Closes #7158 ..." in the PR description, merging this would close that issue. Idk if there's a way to circumvent that but what I do in such cases where it's an incomplete resolution of an issue is use different phrasing like, for example, "GTK implementation for #7158".

mitchellh:
I fixed the tests (there were a handful, the failures were just blocking other failures), which also yielded one bug (fixed), and I also updated the comment to not close the issue. I'll enable auto-merge now!

mitchellh (approved):

src/apprt/gtk/CommandPalette.zig
```diff
@@ -105,12 +104,17 @@ pub fn updateConfig(self: *CommandPalette, config: *const configpkg.Config) !voi
     _ = self.arena.reset(.retain_capacity);
 
     // TODO: Allow user-configured palette entries
```
tristan957: Does this todo need to be removed?

PR #7692, opened 2025-06-26
terminal/Screen: account for rectangle selection in clone

qwerasd205:
Fixes an issue where rectangle selections would appear visually wrong if their start or end were out of the viewport area, because when cloning them the restored pins were defaulting to the start and end of the row instead of the appropriate column.

This issue is shown in discussion #7687.

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -402,8 +402,8 @@ pub fn clonePool(
         };
 
         const start_pin = pin_remap.get(ordered.tl) orelse start: {
-            // No start means it is outside the cloned area. We change it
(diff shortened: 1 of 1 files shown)


mitchellh:
This looks good but it seems like something that is trivially testable and we probably should. As usual, I'm happy to do it when I have time.

qwerasd205:
I'll add tests later today.

qwerasd205:
Good thing you prompted me to add tests, because in doing so I found a bug that's been hiding in this area of the code for a while haha -- it doesn't affect any current use cases but I'm fixing it regardless.

PR #7709, opened 2025-06-28
terminal: fix unexpected line wrapping in tests

jparise:
These tests write specific lines into a 10-column-wide test screen. The "prompt3$ input3\n" writes exceed that column limit, and some of their characters wrap onto the following line.

These tests' current assertions aren't sensitive to that overflow, but I spotted the problem while doing some related work, and I thought it worth making these corrections to avoid any future surprises.


src/terminal/Screen.zig
```diff
@@ -7999,7 +7999,7 @@ test "Screen: selectOutput" {
         } }, s.pages.pointFromPin(.active, sel.start()).?);
         try testing.expectEqual(point.Point{ .active = .{
             .x = 9,
-            .y = 12,
+            .y = 11,
```
jparise, line 8002: This change confirms that we're only writing the 11 expected lines.

mitchellh (approved):

PR #7710, opened 2025-06-28
Equalize splits based on children oriented in the same direction

ibash:
This changes equalization so it only counts children oriented in the same direction.

This makes splits a bit more aesthetic, and replicates how split equalization works in neovim.

### This is how splits look before this change:

Notice how the left pane gets squeezed.

Before equalization
Screenshot 2025-06-28 at 8 47 40 AM

After equalization
Screenshot 2025-06-28 at 8 47 49 AM

---

### And here's how the equalization works after this change:

Before equalization
Screenshot 2025-06-28 at 8 48 18 AM

After equalization
Screenshot 2025-06-28 at 8 48 24 AM

For many splits it looks much more aesthetic:
Screenshot 2025-06-28 at 8 48 28 AM
Screenshot 2025-06-28 at 8 48 35 AM

### DIFF
--- a/macos/Sources/Features/Splits/SplitTree.swift
+++ b/macos/Sources/Features/Splits/SplitTree.swift
@@ -610,14 +610,18 @@ extension SplitTree.Node {
             return (self, 1)
             
         case .split(let split):
-            // Recursively equalize children
-            let (leftNode, leftWeight) = split.left.equalizeWithWeight()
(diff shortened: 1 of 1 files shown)


mitchellh:
Thanks, this behaves a lot more like other applications with splits so this feels good to me. We should plan on migrating this work to GTK too.

PR #7712, opened 2025-06-28
gtk(wayland): prevent gtk4-layer-shell crash on old versions

pluiedev:
Supersedes #7154

In gtk4-layer-shell versions < 1.0.4, the app could crash upon opening a quick terminal window on certain compositors that implement the `xdg_wm_dialog_v1` protocol. The exact reason is a bit complicated, but is nicely summarized in the upstream issue (wmww/gtk4-layer-shell#50).

The circumstances that could cause this crash to occur should gradually diminish as distros update to newer gtk4-layer-shell versions, but this is known to crash on Fedora 41 and Hyprland, which could be a sizable chunk of our userbase given that this would also occur on GNOME/Mutter and KDE/KWin. The diff should be minimal enough that this can be removed or reverted once this band-aid fix is no longer necessary.

### DIFF
--- a/src/apprt/gtk/winproto/wayland.zig
+++ b/src/apprt/gtk/winproto/wayland.zig
@@ -37,6 +37,19 @@ pub const App = struct {
         default_deco_mode: ?org.KdeKwinServerDecorationManager.Mode = null,
 
         xdg_activation: ?*xdg.ActivationV1 = null,
+
+        /// Whether the xdg_wm_dialog_v1 protocol is present.
+        ///
+        /// If it is present, gtk4-layer-shell < 1.0.4 may crash when the user
+        /// creates a quick terminal, and we need to ensure this fails
+        /// gracefully if this situation occurs.
+        ///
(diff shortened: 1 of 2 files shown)


AnthonyZhOon (approved):

src/apprt/gtk/winproto/wayland.zig
```diff
@@ -95,11 +108,20 @@ pub const App = struct {
...
             return false;
         }
+
+        if (self.context.xdg_wm_dialog_present and layer_shell.getLibraryVersion().order(.{
+            .major = 1,
+            .minor = 0,
+            .patch = 4,
+        }) == .lt) {
+            log.warn("the version of gtk4-layer-shell installed on your system is too old (must be 1.0.4 or newer); disabling quick terminal", .{});
+        }
```
mitchellh, line 124: I’m reviewing this on mobile but should there be a return false here?

pluiedev: Oh true

mitchellh (approved):

PR #7717, opened 2025-06-28
config: more general purpose backwards compatibility handlers

### LINKED ISSUE #7706: config: accept gtk-tabs-location=hidden for window-show-tab-bar=never
From #5590 

> Rebased and replaced gtk-tabs-location=hidden with window-show-tab-bar=never. This is obviously a breaking change but we can add a migration mechanism in a different PR

We should accept `gtk-tabs-location=hidden` still, but internally should map it to the new config and we should log a warning. 

We have a mechanism for renaming fields but not for removing single options. I think I want to take a look at this myself.

mitchellh:
Fixes #7706

We previously had a very specific backwards compatibility handler for handling renamed fields. We always knew that wouldn't scale but I wanted to wait for a real case. Well, #7706 is a real case, so here we are.

This commit makes our backwards compatibility handler more general purpose, and makes a special-case handler for renamed fields built on top of this same general purpose system. The new system lets us do a lot more with regards to backwards compatibility.

To start, this addresses #7706 by allowing us to handle a removed single enum value of a still-existing field.

In the future, I think this may continue to get _more_ general purpose by moving the handlers from functions to structs so we can have more metadata like descriptions and so on that we may use to generate docs or other help strings.

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -2,6 +2,8 @@ const diags = @import("cli/diagnostics.zig");
 
 pub const args = @import("cli/args.zig");
 pub const Action = @import("cli/action.zig").Action;
+pub const CompatibilityHandler = args.CompatibilityHandler;
+pub const compatibilityRenamed = args.compatibilityRenamed;
 pub const DiagnosticList = diags.DiagnosticList;
 pub const Diagnostic = diags.Diagnostic;
 pub const Location = diags.Location;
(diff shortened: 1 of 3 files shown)


pluiedev (approved):
Just reviewing this on mobile but this looks super clean

mitchellh:
Thanks!

PR #7721, opened 2025-06-29
Introduce action for copying into clipboard

tt:
This introduces an action for copying the path of a written screen/selection file into the clipboard.

Pasting the path into the terminal doesn't work well if you have a program still running and opening the file outside the terminal (on macOS in TextEdit by default) isn't always a great experience.

Allowing to copy the file path into the clipboard seems like a minor and hopefully uncontroversial addition. 😅

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -5034,6 +5034,12 @@ pub const Keybinds = struct {
 
         try self.set.put(
             alloc,
+            .{ .key = .{ .unicode = 'j' }, .mods = .{ .shift = true, .ctrl = true, .super = true } },
+            .{ .write_screen_file = .copy },
+        );
+
+        try self.set.put(
+            alloc,
             .{ .key = .{ .unicode = 'j' }, .mods = inputpkg.ctrlOrSuper(.{ .shift = true }) },
             .{ .write_screen_file = .paste },
         );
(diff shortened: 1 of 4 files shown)


00-kat (approved):

src/config/Config.zig
```diff
@@ -5032,6 +5032,12 @@ pub const Keybinds = struct {
             .{ .reset_font_size = {} },
         );
 
+        try self.set.put(
+            alloc,
+            .{ .key = .{ .unicode = 'j' }, .mods = .{ .shift = true, .ctrl = true, .super = true } },
```
tt, line 5037: I chose this combination for parity with how taking a screenshot works on macOS. [Holding the <kbd>Control</kbd> key](
https://support.apple.com/en-us/102646#:~:text=Or%20press%20Control%20in%20addition%20to%20the%20other%20keys%20you%20press%20to%20take%20the%20screenshot.) picks the clipboard.

mitchellh (approved):

src/config/Config.zig
mitchellh: I'm not sure if this needs a default keybinding but I see it also lines up with our other write screen back file actions so lets give it a go.

PR #7725, opened 2025-06-29
termio: indicate support for OSC 52 in primary DA report

j4james:
This is an extension agreed upon by modern terminals to indicate that
they support copying to the clipboard with XTerm's OSC 52 sequence. It
is only reported when writing to the clipboard is actually allowed.

Ref: #7590


j4james:
This extension was discussed in #7590, but I'm not sure if anyone officially endorsed the idea, so feel free to close this PR if you don't want to support it.

mitchellh:
I'm happy to follow suit with the named terminals. If this causes unknown problems we can adjust.

PR #7731, opened 2025-06-29
update libxev to workaround the io_uring regression in Linux 6.15.4

### LINKED ISSUE #7724: Linux Kernel 6.15.4 makes io_uring backend unusably slow
**For users looking for a workaround, see here: https://github.com/ghostty-org/ghostty/discussions/7720**

The title says it all. With Linux Kernel 6.15.4 Ghostty is unusably slow. There isn't any special config needed. It just happens out of the box. Downgrading to 6.15.3 fixes this. A workaround is noted in #7720. 

We haven't fully diagnosed the issue yet so this issue isn't directly actionable, against our usual standards. But this is so egregious and confirmed to exist that I'm opening the issue now and we can figure out resolutions here

mitchellh:
Fixes #7724

Background at the end of the commit message. The fix in libxev is described in the PR and commit we pin to here, but basically we swap read for poll for eventfd/timerfd. libxev commit: https://github.com/mitchellh/libxev/commit/75a10d0fb374e8eb84948dcfc68d865e755e59c2

From Jens Axboe on X:

> This will fix it: https://pastebin.com/n7JSZWpW which makes me suspicious
> that it's an S_IFREG check somewhere else, as anon inodes are now listed as
> regular files. Has potentially pretty broad implications...

> I think I can already answer why that breaks things - io_uring checks if
> this is a regular file, and if it is, it doesn't do short reads. Short
> reads on regular files (or a bdev) will cause application issues, as
> basically nobody expects them.

> Now we have what acts like a char dev, but where io_uring will retry IO
> because the application asked for more data than what was delivered. This
> will cause the weird slowdowns as data isn't delivered as soon as it's
> available.

## Backporting to 1.1.3

It's a bit risky, but we can back port the libxev update to 1.1.3 (and release a 1.1.4) since we don't plan on a 1.2 release for a couple months. Realistically, we've been running the latest libxev on tip (before this commit) for awhile and there haven't been issues so I think it's safe enough.

I think we should let this merge settle for a few days at least before we do that back port though.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -8,8 +8,8 @@
 
         .libxev = .{
             // mitchellh/libxev
-            .url = "https://github.com/mitchellh/libxev/archive/9bc52324d4f0c036a3b244e992680a9fb217bbd3.tar.gz",
-            .hash = "libxev-0.0.0-86vtc5b1EgB7vFmt9Tk7ySteR5AeEHW7xcR6gK9dMUD3",
+            .url = "https://github.com/mitchellh/libxev/archive/75a10d0fb374e8eb84948dcfc68d865e755e59c2.tar.gz",
+            .hash = "libxev-0.0.0-86vtcyMBEwA-UpcjfOICyI2GYG8o6MiRbinS1_8g1_rw",
             .lazy = true,
         },
         .vaxis = .{
--- a/build.zig.zon.json
+++ b/build.zig.zon.json
@@ -64,10 +64,10 @@
     "url": "https://deps.files.ghostty.org/libpng-1220aa013f0c83da3fb64ea6d327f9173fa008d10e28bc9349eac3463457723b1c66.tar.gz",
     "hash": "sha256-/syVtGzwXo4/yKQUdQ4LparQDYnp/fF16U/wQcrxoDo="
   },
-  "libxev-0.0.0-86vtc5b1EgB7vFmt9Tk7ySteR5AeEHW7xcR6gK9dMUD3": {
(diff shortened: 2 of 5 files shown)


emirkmo:
A backport would be highly appreciated for packaging reasons.

Else a new release is even better :)

Since neither this nor the configuration workaround are available on any release.

darix:
do I understand this correctly ... if distros backport the patch in the pastebin then the issue would also be fixed?

jcollie:
How quickly do we expect distros to update to a fixed 6.15.5 or backport a fix? I'm not sure that 6.15.4 is going to be around long enough to really merit a release.

mitchellh:
> How quickly do we expect distros to update to a fixed 6.15.5 or backport a fix? I'm not sure that 6.15.4 is going to be around long enough to really merit a release.

Yes, I'd like to know this as well.

All things considered, the libxev workaround doesn't really introduce any new complexity or cost, so I'd like to go ahead and merge this _anyways_ and continue to maintain this behavior going forward. The question really is if we back port or not.

jerome-benoit:
I do not think introducing inconsistency in API usage in a piece of code, because of a regression targeted to a specific "dependency" version, is a wise idea in general. A simple runtime check is enough without compromising the consistency of API consumption.

pluiedev:
We could always just revert this change after 6.15.5 has reached general availability in all impacted distros, which still isn't the case for Fedora at least.

mitchellh:
We probably will revert it but note it’s not really inconsistency here. Epoll always did this. This has been a very safe change so far.

PR #7733, opened 2025-06-29
terminal: introduce testWriteSemanticString

jparise:
This test-only function wraps testWriteString with semantic prompt marking. This replaces the manual, row-based semantic_prompt field manipulation we were doing in all of our prompt-related test setups.

This function's heuristics are a little complex because it wraps testWriteString as a "black box"; we don't benefit from that function's own line-based logic to know which rows need to be updated with the semantic prompt flag. We need to infer them externally instead.

I considered adding an options argument to testWriteString that would allow passing e.g. a semantic_prompt prompt. Given that it's called from 200+ places, that would involve a lot of unrelated changes, but it remains an "option" (ha!) if there's value there for other cases.

I also have plans that move us from row-based to cell-based semantic tracking, where the current semantic type is tracked by the cursor. In that implementation, testWriteString can update the written cells directly, and testWriteSemanticString just helps manage the cursor's state. Introducing testWriteSemanticString here and now therefore helps bridge us to that world while maintaining test consistency.

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -3068,6 +3068,29 @@ pub fn testWriteString(self: *Screen, text: []const u8) !void {
     }
 }
 
+/// Write text that's marked as a semantic prompt.
+fn testWriteSemanticString(self: *Screen, text: []const u8, semantic_prompt: Row.SemanticPrompt) !void {
+    // Determine the first row using the cursor position. If we know that our
+    // first write is going to start on the next line because of a pending
+    // wrap, we'll proactively start there.
+    const start_y = if (self.cursor.pending_wrap) self.cursor.y + 1 else self.cursor.y;
+
+    try self.testWriteString(text);
+
+    // Determine the last row that we actually wrote by inspecting the cursor's
+    // position. If we're in the first column, we haven't actually written any
+    // characters to it, so we end at the preceding row instead.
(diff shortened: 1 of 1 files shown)


mitchellh:
Given this is so tightly scoped I'm fine merging this for now. Thanks for making small PRs.

> I considered adding an options argument to testWriteString that would allow passing e.g. a semantic_prompt prompt. Given that it's called from 200+ places, that would involve a lot of unrelated changes, but it remains an "option" (ha!) if there's value there for other cases.

I'm pretty unhappy with `testWriteString` as it is. It's one of those test functions that started _at the beginning_ and just iterated from there. Semantic prompts as an example its just way too inflexible. I think longer term I do want to replace `testWriteString` with a full-featured `writeString`-type function that takes an options structure or something to do more things.

mitchellh (approved):

jparise:
> I'm pretty unhappy with `testWriteString` as it is. It's one of those test functions that started _at the beginning_ and just iterated from there. Semantic prompts as an example its just way too inflexible. I think longer term I do want to replace `testWriteString` with a full-featured `writeString`-type function that takes an options structure or something to do more things.

I agree, and I'll keep this in mind as a work in related areas. I think there's a balance to be struck between how much a (test) function like this relies on `Screen`- and `Cursor`-based state versus how much is explicitly parameterized, and it's hit that limit in its current form.

PR #7738, opened 2025-06-30
macos: don't overwrite the .fullScreen styleMask option in reapplyHiddenStyle

danielwe:
Fixes #7625 (which apparently didn't catch anyone's attention, but I'd be very surprised if no one were able to reproduce it if they tried; also, re discord chat)

### DIFF
--- a/macos/Sources/Features/Terminal/Window Styles/HiddenTitlebarTerminalWindow.swift
+++ b/macos/Sources/Features/Terminal/Window Styles/HiddenTitlebarTerminalWindow.swift
@@ -19,20 +19,27 @@ class HiddenTitlebarTerminalWindow: TerminalWindow {
         NotificationCenter.default.removeObserver(self)
     }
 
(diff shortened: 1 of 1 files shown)


qwerasd205:
This looks like a nice well-scoped fix

macos/Sources/Features/Terminal/Window Styles/HiddenTitlebarTerminalWindow.swift
```diff
@@ -19,20 +19,27 @@ class HiddenTitlebarTerminalWindow: TerminalWindow {
         NotificationCenter.default.removeObserver(self)
     }
 
+    private let hiddenStyleMask: NSWindow.StyleMask = [
```
jparise: Maybe make this `static`?

danielwe: done!

mitchellh (approved):

PR #7751, opened 2025-07-01
core: handle SIGUSR2 as a signal to reload the config

### LINKED ISSUE #7747: SIGUSR2 to reload config
We should react to the `SIGUSR2` signal and reload Ghostty configuration.

Practically speaking, [Omarchy](https://omarchy.org) wants to utilize Ghostty more and this is blocking them, because they have a centralized "change theme" command that changes a theme across your entire system and Ghostty doesn't have a way to react to config changes outside the process. 

## Background

It is a common pattern across software of varying categories on Unix-y systems to reload configuration in reaction to a signal. The signal in use varies a bit between SIGHUP, SIGUSR1, SIGUSR2. In general though, the USR signals are usually more common. 

Most commonly is infrastructure software: web servers, proxies, databases, etc. For example, nginx [reloads on SIGHUP](https://docs.nginx.com/nginx/admin-guide/basic-functionality/runtime-control/), 

For non-infrastructure software, [Waybar uses SIGUSR2](https://github.com/Alexays/Waybar/wiki/FAQ#how-can-I-reload-the-configuration-without-restarting-waybar). 

This is useful because it lets software external to the system trigger a reload without resorting to a full API system.

### Security

From a security standpoint, signals can only be sent to processes owned by the same user or from a privileged user. Additionally, config files already have their own filesystem permissions. So this doesn't change the threat model or surface area of access at all. 

## Alternatives

There are alternatives to this, but I don't think its mutually exclusive:

* A full API. This is planned at some point, but has a significantly higher technical complexity and security burden.

* Filesystem watching for the config. This is also something I want to do eventually, but sometimes it's easier to just send a `kill -SIGUSR2` from something like a script rather than assume/hope that filesystem watching is happening. Additionally, filesystem watching may not be fully desirable for a user and I plan on making it configurable.

The signal approach is extremely simple and can be implemented quickly, and as I said is not mutually exclusive to the other approaches.

jcollie:
This PR adds the infrastructure to handle POSIX signals. The only signal currently handled is SIGUSR2, which will reload the config.

SIGPIPE is set up to be ignored in the global state and should not be changed as macOS relies on SIGPIPE being ignored for proper operation.

Note that any signal not configured by Ghostty will generally result in termination of the process.

Fixes #7747

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -20,6 +20,7 @@ const font = @import("font/main.zig");
 const internal_os = @import("os/main.zig");
 const macos = @import("macos");
 const objc = @import("objc");
+const Signals = @import("Signals.zig");
 
 const log = std.log.scoped(.app);
 
@@ -74,6 +75,9 @@ config_conditional_state: configpkg.ConditionalState,
 /// if they are the first surface.
 first: bool = true,
 
+/// Structure for managing POSIX signal handlers.
+signals: Signals = .init,
+
 pub const CreateError = Allocator.Error || font.SharedGridSet.InitError;
 
 /// Create a new app instance. This returns a stable pointer to the app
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Unfortunately not the right approach because it's technically legal to have multiple `App` via the C API. No one does it, but it is technically legal and I want to keep it that way from an API perspective. I have something locally im working on so I'll push that up separately.

PR #7752, opened 2025-07-01
Fix abnormal exit detection on macOS

mitchellh:
I made an oopsie with #7705 and omitted the check entirely on macOS when the original logic only omitted the exit code check.


src/Surface.zig
```diff
@@ -1002,10 +1002,11 @@ fn childExited(self: *Surface, info: apprt.surface.Message.ChildExited) void {
     if (info.runtime_ms <= self.config.abnormal_command_exit_runtime_ms) runtime: {
         // On macOS, our exit code detection doesn't work, possibly
         // because of our `login` wrapper. More investigation required.
-        if (comptime builtin.target.os.tag.isDarwin()) break :runtime;
+        if (comptime !builtin.target.os.tag.isDarwin()) {
+            // If the exit code is 0 then we it was a good exit.
```
gpanders:
```suggestion
            // If the exit code is 0 then it was a good exit.
```

PR #7754, opened 2025-07-01
linux: add some docs on launching/managing Ghostty

jcollie:
The addition of D-Bus activation and systemd user services in #7679 has made the "default" launching of Ghostty a little more complicated. This PR attempts to document that process works and commands for managing/controlling that process.

### DIFF
--- a/README-LINUX.md
+++ b/README-LINUX.md
@@ -73,7 +73,7 @@ log in by running this command:
 systemctl enable --user com.mitchell.ghostty.service
 ```
 
-By doing so any startup delay will be "hidden" by the time that your system
+By doing so, any startup delay will be "hidden" by the time that your system
 needs to finish setting up everything after logging in.
 
 ## Stopping Ghostty
--- a/README-LINUX.md
+++ b/README-LINUX.md
@@ -71,7 +71,7 @@ be running.
 
 ### If Ghostty is Already Running
 
-If Ghostty is already running, the broker will send a D-Bus signal directy to
+If Ghostty is already running, the broker will send a D-Bus signal directly to
 Ghostty asking it to create a new window.
 
 ```mermaid
--- a/README-LINUX.md
+++ b/README-LINUX.md
@@ -2,10 +2,14 @@
 
 ## Launching
 
-If installed by a system package or by compiling from source and installing into
-`$HOME/.local` (with `zig build -p $HOME/.local -Doptimize=ReleaseFast`) several
-configuration files will be installed which allow Ghostty to be launched from
-most GUI application launchers (like Gnome Shell or KDE Plasma).
+If properly installed, Ghostty will take advantage of D-Bus activation and
+`systemd` user services to provide the best experience possible.
+
+If installed by a system package or by compiling from source and installing
+into `$HOME/.local` (with `zig build -p $HOME/.local -Doptimize=ReleaseFast`)
+several configuration files will be installed which allow Ghostty to be launched
+from most GUI application launchers (like Gnome Shell or KDE Plasma) using D-Bus
+activation and a `systemd` user service.
 
 These are the files that control the launching of Ghostty if installed as
 a user:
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
I haven't looked, but will this not fit somewhere in the README? Ultimately this should probably be on the website instead (for 1.2).

I think a second question I wanted to look into: does Ghostty effectively require systemd now? I want to think more about this if so, because if it does we should probably introduce a build flag to have a non-systemd build. I think that's maybe what you were doing in your original PR but I didn't understand it. We don't need to do that in this PR, its just a question and maybe should be clarified here.

jcollie:
> I haven't looked, but will this not fit somewhere in the README? Ultimately this should probably be on the website instead (for 1.2).

I felt that the README was more build & install related things, but it could go there (at the risk of getting very long and the content getting lost). Yes, ultimately this should be on the website once 1.2 is released. Right now I wanted to get this information written down _somewhere_ so that it can be referred to by users and helpers.

> I think a second question I wanted to look into: does Ghostty effectively require systemd now? I want to think more about this if so, because if it does we should probably introduce a build flag to have a non-systemd build. I think that's maybe what you were doing in your original PR but I didn't understand it. We don't need to do that in this PR, its just a question and maybe should be clarified here.

It's my understanding that the `systemd` specific-stuff added into the `.desktop` and `.service` files will be ignored if `systemd` is not used on the system so we do not have a hard dependency on `systemd` at this point. Non-systemd Linux systems are a bit rare but if there's an easy-to-use Alpine (for example) VM that would be a good test.

Maybe some of the FreeBSD users could try the default `.desktop` and `.service` files on their systems to see what happens. Right now FreeBSD systems don't install any of those at all.

README-LINUX.md
```diff
@@ -0,0 +1,148 @@
...
+## Creating New Windows
+
+With D-Bus activation and `systemd` user services in use, simply launching
+`ghostty` from the CLI (or through a global keybind) will not create a new
+Ghostty terminal window. The following command can be used to send a D-Bus
+signal to Ghostty to create a new window:
+
+```sh
+gdbus call --session --dest com.mitchellh.ghostty --object-path /com/mitchellh/ghostty --method org.gtk.Actions.Activate new-window [] []
+```
```
tristan957, line 156: This seems like something that the Ghostty CLI could handle. Is that a wrong assumption?

README-LINUX.md
```diff
@@ -0,0 +1,148 @@
...
+only incurred once per login.
+
+The Ghostty `systemd` user service can be configured to start up as soon as you
+log in by running this command:
+
+```sh
+systemctl enable --user com.mitchell.ghostty.service
+```
+
+By doing so any startup delay will be "hidden" by the time that your system
```
tristan957:
```suggestion
By doing so, any startup delay will be "hidden" by the time that your system
```

README-LINUX.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+If installed as a system package, the files will be found in the following
+locations:
+
+```
+$PREFIX/share/applications/com.mitchellh.ghostty-debug.desktop
+$PREFIX/share/dbus-1/services/com.mitchellh.ghostty-debug.service
+$PREFiX/lib/systemd/user/com.mitchellh.ghostty-debug.service
+```
```
tristan957: Seems repeated.

tristan957 (commented):
I agree with the website comment.

README-LINUX.md
tristan957: I tested this on latest master, and didn't find this to be the case.

jcollie: How did you install? The service files need to be in the right locations in `~/.local` for this to work.

README-LINUX.md
jcollie: Yeah, but I thought it would be best to be explicit.

README-LINUX.md
jcollie: > This seems like something that the Ghostty CLI could handle. Is that a wrong assumption?

It wouldn't be hard to add a `+gtk-new-window` CLI action that sends the D-Bus signal, but that might be too GTK specific for a CLI action.  Doing something deeper would require rethinking the GTK app init code and would take some deeper thinking.

tristan957: > How did you install? The service files need to be in the right locations in `~/.local` for this to work

That's what I did.

jcollie: Hmm, I'd check the files to see if they are correct. Also, have you logged in/out recently, or run `systemctl daemon-reload --user`? Maybe `systemd` hasn't picked up the new files.

mitchellh:
I've setup the tip website (https://tip.ghostty.org). I suggest we put this information there. I'm open to ideas where it should go. But I think it'd work best there, and then it will become part of 1.2.0 when released. We can have this discussion on the website repo, just PR into the `tip` branch.

PR #7757, opened 2025-07-01
linux: add install target to systemd user service

jcollie:
This will allow users to enable Ghostty startup on login. Users will need to explicitly enable startup on login via this command:

```sh
systemctl enable --user com.mitchellh.ghostty.service
```

### DIFF
--- a/dist/linux/systemd.service.in
+++ b/dist/linux/systemd.service.in
@@ -1,7 +1,11 @@
 [Unit]
 Description=@NAME@
+After=graphical-session.target
 
 [Service]
 Type=dbus
 BusName=@APPID@
 ExecStart=@GHOSTTY@ --launched-from=systemd
+
+[Install]
+WantedBy=graphical-session.target


mitchellh:
One thing users have requested is a way to have Ghostty startup with no windows on startup so that subsequent activations are really fast. This probably isn't the best place to discuss this, but how do you think that fits in here? Having an optional systemd unit that we pre-install would be really cool.

jcollie:
> One thing users have requested is a way to have Ghostty startup with no windows on startup so that subsequent activations are really fast. This probably isn't the best place to discuss this, but how do you think that fits in here? Having an optional systemd unit that we pre-install would be really cool.

Yes, this is exactly that. They have to manually run the listed command, but after that Ghostty will start up at login but not show a window. Sorry if I didn't properly explain.

tristan957 (approved):

mitchellh (approved):

PR #7759, opened 2025-07-01
reload configuration on SIGUSR2

mitchellh:
This is done at the apprt-level for a couple reasons.

  (1) For libghostty, we don't have a way to know what the embedding
      application is doing, so its risky to create signal handlers that
      might overwrite the application's signal handlers.

  (2) It's extremely messy to deal with signals and multi-threading.
      Apprts have framework access that handles this for us.

For GTK, we use g_unix_signal_add.

For macOS, we use `DispatchSource.makeSignalSource`. This is an awkward API but made for this purpose.

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -373,6 +373,13 @@ pub fn init(self: *App, core_app: *CoreApp, opts: Options) !void {
         .{},
     );
 
+    // Setup a listener for SIGUSR2 to reload the configuration.
+    _ = glib.unixSignalAdd(
+        std.posix.SIG.USR2,
+        sigusr2,
+        self,
+    );
+
     // We don't use g_application_run, we want to manually control the
     // loop so we have to do the same things the run function does:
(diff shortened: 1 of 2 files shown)


tristan957 (approved):

src/apprt/gtk/App.zig
```diff
@@ -1508,6 +1515,22 @@ pub fn quitNow(self: *App) void {
     self.running = false;
 }
 
+// SIGUSR2 signal handler via g_unix_signal_add
+fn sigusr2(ud: ?*anyopaque) callconv(.c) c_int {
+    const self: *App = @ptrCast(@alignCast(ud orelse
+        return @intFromBool(glib.SOURCE_CONTINUE)));
```
mitchellh, line 1521: Things are arguably extremely fucked if this triggers so we should maybe remove the source but I don't think it really matters either way.

jcollie: I agree, if this fails it doesn't matter because things are way messed up.

jcollie (approved):
LGTM

PR #7773, opened 2025-07-03
Do not resolve the symbolic link for the initial working directory

basil:
Fixes #7769

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -847,6 +847,15 @@ const Subprocess = struct {
         else
             null;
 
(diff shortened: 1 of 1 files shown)


00-kat (approved):

jparise (approved):
This looks good to me. I also confirmed that `PWD` isn't one of the environment variables set by `login(1)` on macOS, so doing this in the common path is the right place.

mitchellh:
Thanks for the research into this. I've pushed up just a bit more detail in the comment but otherwise this looks good to me. Thank you.

PR #7790, opened 2025-07-03
Add linux kernel information to +version

Secrus:
Adds basic kernel information to `ghostty +version`. 
`ghostty +version` on my machine now prints:
```
Ghostty 1.1.4-kernel-version-info+cdbc78bce

Version
  - version: 1.1.4-kernel-version-info+cdbc78bce
  - channel: tip
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.Debug
  - app runtime   : apprt.Runtime.gtk
  - font engine   : font.main.Backend.fontconfig_freetype
  - libxev        : io_uring
  - kernel version: 6.15.4-200.fc42.x86_64
  - desktop env   : other
  - GTK version   :
    build         : 4.18.5
    runtime       : 4.18.5
  - libadwaita    : enabled
    build         : 1.7.4
    runtime       : 1.7.4
  - libX11        : enabled
  - libwayland    : enabled

```

PS This is my first time writing prod code in Zig, so any suggestions and guidelines are welcome

### DIFF
--- a/src/cli/version.zig
+++ b/src/cli/version.zig
@@ -15,8 +15,6 @@ pub const Options = struct {};
 /// The `version` command is used to display information about Ghostty. Recognized as
 /// either `+version` or `--version`.
 pub fn run(alloc: Allocator) !u8 {
-    _ = alloc;
-
     const stdout = std.io.getStdOut().writer();
     const tty = std.io.getStdOut().isTty();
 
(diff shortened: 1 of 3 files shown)


src/cli/version.zig
```diff
@@ -34,32 +32,37 @@ pub fn run(alloc: Allocator) !u8 {
...
-        try stdout.print("    runtime    : {}\n", .{gtk_version.getRuntimeVersion()});
-        try stdout.print("  - libadwaita : enabled\n", .{});
-        try stdout.print("    build      : {}\n", .{adw_version.comptime_version});
-        try stdout.print("    runtime    : {}\n", .{adw_version.getRuntimeVersion()});
+        if (comptime builtin.os.tag == .linux) {
+            const kernel_info = internal_os.getKernelInfo(alloc);
+            defer if (kernel_info) |k| alloc.free(k);
+            try stdout.print("  - kernel version: {s}\n", .{kernel_info orelse "Kernel information unavailable"});
+        }
+        try stdout.print("  - desktop env   : {s}\n", .{@tagName(internal_os.desktopEnvironment())});
```
jcollie, line 47: It might be nice to move the kernel version outside the GTK block in case there's ever a non-GTK linux runtime, but that's minor and can be fixed up later if that ever happens. Or if we extend the kernel version to support macOS/FreeBSD (if it even matters there).

jcollie (approved):
LGTM

src/cli/version.zig
Secrus: Yeah, I assumed that if that ever happens, then it will be moved along with that change, since that would most likely change the API that this is using.

PR #7808, opened 2025-07-04
bash: stop using PS0 for the 'cursor' feature

### LINKED ISSUE #7802: bash shell integration can cause garbage characters to be printed
I reliably reproduced this while messing with multiline bash aliases:

```bash
$ alias garbage='echo start
> echo end'

$ garbage
start
�\��dend
```

```bash
$ alias normal='(echo start
> echo end)'

$ normal
start
end
```

Adding `shell-integration = none` seemed to make the garbage alias behave normal.

_Originally posted by @wong-justin in https://github.com/ghostty-org/ghostty/discussions/3755#discussioncomment-12268943_

jparise:
Our use of PS0 (which bash runs before command execution) was causing raw command sequences to be printed between multiple commands in a sequence.

    $ alias garbage='echo start
    > echo end'

    $ garbage
    start
    �\���dend

I wasn't able to definitely track down all of the reasons for why this only happens in the command sequence case, but I suspect it's related to the way that __ghostty_preexec runs from within the bash DEBUG trap (by way of bash-preexec).

This problem occurs when PS0 is set to _any_ string (even "") inside of __ghostty_preexec, which also rules out most/any Ghostty-specific code. PS1 and PS2 appear to be safe to (re)set in this context.

Fortunately, we can avoid using PS0 entirely by instead printing the cursor reset escape sequence directly from __ghostty_precmd because it also runs just before command execution.

Fixes #7802

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -106,7 +106,6 @@ _ghostty_last_reported_cwd=""
 function __ghostty_precmd() {
     local ret="$?"
     if test "$_ghostty_executing" != "0"; then
-      _GHOSTTY_SAVE_PS0="$PS0"
       _GHOSTTY_SAVE_PS1="$PS1"
       _GHOSTTY_SAVE_PS2="$PS2"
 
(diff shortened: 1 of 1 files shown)


mitchellh (approved):

jparise:
This fixed the linked issue, but I now see cases where the blinking bar insertion cursor persists into other contexts, such as when resuming a suspended vim session. I don't have a solution to that yet, so I may consider reverting this change because a leaking cursor shape is probably a "worse" outcome than unprintable characters in command sequences.

mitchellh:
Thanks. I'll defer to you whether you think we should revert. If so, just open a PR with the revert and I'll approve it.

jparise:
> Thanks. I'll defer to you whether you think we should revert. If so, just open a PR with the revert and I'll approve it.

#7839 is an alternate approach that should address both issues.

PR #7809, opened 2025-07-04
Better Glyph Constraint Logic

qwerasd205:
This is a big'un.

- **Glyph constraint logic is now done fully on the CPU** at the rasterization stage, so it only needs to be done once per glyph instead of every frame. This also lets us eliminate padding between glyphs on the atlas because we're doing nearest-neighbor sampling instead of interpolating-- which ever so slightly increases our packing efficiency.
- **Special constraints for nerd font glyphs** are applied based roughly on the constraints they use in their patcher. It's a simplification of what they do, the largest difference being that they scale groups of glyphs based on a shared bounding box so that they maintain relative size to one another, but that would require loading all glyphs on the group and I'd want to do that on font load TBH and at that point I'd basically be re-implementing the nerd fonts patcher in Zig to patch fonts at load time which is way beyond the scope I want to have. (Fixes #7069)
- These constraints allow for **perfectly sized and centered emojis**, this is very nice.
- **Changed the default embedded fonts** from 4 copies (regular, italic, bold, bold italic) of a patched (and outdated) JetBrains Mono to a single JetBrains Mono variable font and a single Nerd Fonts Symbols Only font. This cuts the weight of those down from 9MB to 3MB!
- **FreeType's `renderGlyph` is significantly reworked**, and the new code is, IMO, much cleaner- although there are probably some edge case behavior differences I've introduced.

> [!NOTE]
> One breaking change I definitely introduced is changing the `monochrome` freetype load flag config from its previous completely backwards meaning to instead the correct one (I also changed the default, so this won't affect any user who hasn't touched it, but users who set the `monochrome` flag will find their fonts quite crispy after this change because they will have no anti-aliasing anymore)

### Future work

Following this change I want to get to work on automatic font size matching (a la CSS [`font-size-adjust`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust)). I set the stage for that quite some time ago so it shouldn't be too much work and it will be a big benefit for users who regularly use multiple writing systems and so have multiple fonts for them that aren't necessarily size-compatible.


anmolw:
On the subject of freetype load flags, could you have a look at force-autohint too if possible? It seems to be negated for some reason (the config value defaults to true, but then it gets negated while being passed).

mitchellh (approved):

mitchellh (changes requested):
Actually stand by, I thought I fixed the flatpak issue :( but I didn't

mitchellh (approved):

mpettibone:
From @mitchellh [post](https://hachyderm.io/@mitchellh/114802860241636535), just to confirm, should icons be clipped at the top?

IMG_1319


It appears the _circle_, _beaker_, and capital _B_ icons look clipped at top.

mitchellh:
I can post the file I used a bit later. Away from computer at the moment.

qwerasd205:
I need to do a tiny follow-up for a couple things anyway, I think I see the problem that's causing the clipping so I'll fix that too

GustavoWidman:
this had a bunch of unintended side effects, why was this not tested further?

image

ghostty on left, terminal.app on the right.

mitchellh:
@GustavoWidman I hid your comment as abuse. Its unnecessary. You're just as capable of testing this, where were you? Why didn't you help? Why didn't you fix it? We're all volunteers here, same as what you could be. If you have an issue, feel free to open a discussion. But no one here is obligated to do any more or less than you. And you've done almost nothing except complain (so far, but you can be the change you want to see in the world!).

This was good work, it improved many things. Doing complicated work often has regressions. We're addressing the regressions. We're doing our best. But at least we're doing something.

PR #7810, opened 2025-07-04
fix: draw cursor on top of text for most cursor styles

### LINKED ISSUE #7739: Renderer: Most cursor styles should be drawn on top of text
We used to handle this but regressed some time ago- cursors such as the hollow block, bar, and probably the underline cursor should be drawn _on top_ of all of the text, rather than behind, to avoid them being partially obscured. The rect cursor still needs to be drawn behind though.

This can probably be handled easily enough by adding a second dedicated cursor list in the cell contents ArrayListCollection that lives at the end rather than the start, and deciding which one to put the cursor in depending on its style.

Eventually we'll have a dedicated shader for drawing the cursor which can be drawn before or after the text is drawn, depending on the cursor style, but for now the above solution should be fine.

dal-liu:
This addresses #7739.

## Changes
* Added second cursor list to the renderer as suggested
* Modified `Contents.setCursor` to accept a cursor style argument in order to conditionally render the cursor
  * Not sure if this the optimal workaround so suggestions would be appreciated

## Screenshots
Block:
Screenshot 2025-07-04 at 7 10 19 PM

Hollow block:
Screenshot 2025-07-04 at 7 10 06 PM

### DIFF
--- a/src/renderer/cell.zig
+++ b/src/renderer/cell.zig
@@ -145,15 +145,14 @@ pub const Contents = struct {
         self.fg_rows.lists[0].clearRetainingCapacity();
         self.fg_rows.lists[self.size.rows + 1].clearRetainingCapacity();
 
-        if (v) |cell| {
-            if (cursor_style) |style| {
-                switch (style) {
-                    // Block cursors should be drawn first
-                    .block => self.fg_rows.lists[0].appendAssumeCapacity(cell),
-                    // Other cursor styles should be drawn last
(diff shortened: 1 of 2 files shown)


qwerasd205:
Nice, thanks for this. From a brief glance this looks good (nothing jumps out at me as wrong) but I don't have time to do a proper review right now, if @mitchellh thinks it looks good too (and all tests pass) I'm good with this being merged.

mitchellh (approved):
Thanks! Looks good to me.

PR #7816, opened 2025-07-05
chore(ci): pin GitHub Actions to specific SHAs

brenoepics:
Follow-up on #7076

SHAs were generated using [pinact](https://github.com/suzuki-shunsuke/pinact).

By the way, all repository workflows don’t declare permissions, so they use the defaults, which are usually [too permissive](https://docs.zizmor.sh/audits/#excessive-permissions),  I’d suggest using per-workflow/job permissions instead, since most (if not all) jobs don’t need full access. If that’s added, it should go in a separate issue/PR so we can review the minimum needed per job.

Refs:
https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token 

You can check everything with this SARIF file:

[ghostty-sarif.json](https://github.com/user-attachments/files/21081630/ghostty-sarif.json)
read it at https://microsoft.github.io/sarif-web-component/

### DIFF
--- a/.github/pinact.yml
+++ b/.github/pinact.yml
@@ -1,0 +1,4 @@
+version: 3
+ignore_actions:
+  - name: "DeterminateSystems/nix-installer-action"
+    ref: "main"
--- a/.github/workflows/clean-artifacts.yml
+++ b/.github/workflows/clean-artifacts.yml
@@ -10,7 +10,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Remove old artifacts
-        uses: c-hive/gha-remove-artifacts@v1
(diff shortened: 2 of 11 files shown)


mitchellh:
This looks good. I'm verifying all the SHAs are what they say they are manually for this first run.

I also added a CI check to verify that we don't inadvertently push an un-pinned action.

> By the way, all repository workflows don’t declare permissions, so they use the defaults, which are usually [too permissive](https://docs.zizmor.sh/audits/#excessive-permissions), I’d suggest using per-workflow/job permissions instead, since most (if not all) jobs don’t need full access. If that’s added, it should go in a separate issue/PR so we can review the minimum needed per job.

That seems like a good follow up to me.

mitchellh (approved):
Looks good pending CI.

PR #7818, opened 2025-07-05
Add action to set font size

### LINKED ISSUE #7795: Add `set_font_size` keybind action
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7756

<sup>Originally posted by **felipesere** July  1, 2025</sup>
At the moment Ghostty users can increase/decrease the font size by some amount using `increase_font_size` and `decrease_font_size`.
I was wondering if you'd take a PR to set it to a specific value, e.g 14, 18, or 20pt?
My use case is that I switch between my laptop screen and my external screen frequently, so in Alacritty I have a script that modifies the config (and it gets automatically reload) and I have an extra large one for presentations.
I can use increase/decrese but its inconsistent.

I'd be happy to give this a try if the PR would be valuable.</div>

## Proposal

Add a `set_font_size:<size>` keybind action that sets the font to a specific size.

jcollie:
Fixes #7795 

Adds the `set_font_size` keybind action.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -6531,8 +6531,9 @@ pub const RepeatableCommand = struct {
         try list.parseCLI(alloc, "title:Foo,action:ignore");
         try list.parseCLI(alloc, "title:Bar,description:bobr,action:text:ale bydle");
         try list.parseCLI(alloc, "title:Quux,description:boo,action:increase_font_size:2.5");
+        try list.parseCLI(alloc, "title:Baz,description:Raspberry Pie,action:set_font_size:3.14");
 
-        try testing.expectEqual(@as(usize, 3), list.value.items.len);
+        try testing.expectEqual(@as(usize, 4), list.value.items.len);
 
         try testing.expectEqual(inputpkg.Binding.Action.ignore, list.value.items[0].action);
         try testing.expectEqualStrings("Foo", list.value.items[0].title);
@@ -6548,6 +6549,13 @@ pub const RepeatableCommand = struct {
         );
         try testing.expectEqualStrings("Quux", list.value.items[2].title);
         try testing.expectEqualStrings("boo", list.value.items[2].description);
+
+        try testing.expectEqual(
+            inputpkg.Binding.Action{ .set_font_size = 3.14 },
+            list.value.items[3].action,
+        );
+        try testing.expectEqualStrings("Baz", list.value.items[3].title);
+        try testing.expectEqualStrings("Raspberry Pie", list.value.items[3].description);
 
         try list.parseCLI(alloc, "");
         try testing.expectEqual(@as(usize, 0), list.value.items.len);
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -296,6 +296,12 @@ pub const Action = union(enum) {
     /// Reset the font size to the original configured size.
     reset_font_size,
 
+    /// Set the font size to the specified size in points (pt).
+    ///
+    /// For example, `set_font_size:14.5` will set the font size
+    /// to 14.5 points.
+    set_font_size: f32,
+
     /// Clear the screen and all scrollback.
     clear_screen,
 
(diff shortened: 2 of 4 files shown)


mitchellh:
I’m not sure if this is how I want to deprecate keybind actions. Since we don’t have an agreed upon design for that can you split the scope on this PR. We can determine the deprecation later.

the issue in question just wanted to add the new binding. 

I’m not saying I’d reject this outright but limiting PR scope will result in PRs being merged MUCH faster

jcollie:
Removed the commit that deprecates the font size related keybind actions.

src/config/Config.zig
```diff
@@ -5150,23 +5150,23 @@ pub const Keybinds = struct {
         try self.set.put(
             alloc,
             .{ .key = .{ .physical = .equal }, .mods = inputpkg.ctrlOrSuper(.{}) },
-            .{ .increase_font_size = 1 },
```
mitchellh: We'll have to update the macOS app to look for these to sync to the menu. Does GTK have to do the same? Or, we can just keep them the old `increase/decrease/reset` values and everything will keep working. I'd prefer that for now.

src/input/Binding.zig
```diff
@@ -890,43 +982,50 @@ pub const Action = union(enum) {
...
+        if (action.len == 0) return error.InvalidFormat;
 
         const actionInfo = @typeInfo(Action).@"union";
         inline for (actionInfo.fields) |field| {
             if (std.mem.eql(u8, action, field.name)) {
                 // If the field type is void we expect no value
                 switch (field.type) {
                     void => {
-                        if (colonIdx != null) return Error.InvalidFormat;
+                        if (param_) |_| return error.InvalidFormat;
```
mitchellh: I think it's more idiomatic to do a `!= null` if you're throwing away the value anyways.

src/input/Binding.zig
```diff
@@ -890,43 +982,50 @@ pub const Action = union(enum) {
...
     /// action name and value is the action parameter. The parameter
     /// is optional depending on the action.
-    pub fn parse(input: []const u8) !Action {
+    pub fn parse(input: []const u8) ParseActionErrors!Action {
         // Split our action by colon. A colon may not exist for some
         // actions so it is optional. The part preceding the colon is the
         // action name.
         const colonIdx = std.mem.indexOf(u8, input, ":");
         const action = input[0..(colonIdx orelse input.len)];
+        const param_ = if (colonIdx) |idx| input[idx + 1 ..] else null;
```
mitchellh: I think this might change the performance in a subtle way. I think I did the previous approach because we can avoid the branch on `colonIdx` being set if we don't need to care. I'll have to double check.

mitchellh (changes requested):
Some specific notes, but overall I'm not sure if I like this complexity in the keybinding parser. I think it might be neater to just keep this all as separate commands, and make `set_font_size` take an absolute value only. Deltas should use `increase`, `decrease`, etc. 

I haven't seen an argument for unifying them from a practical point of view, and the code complexity here not only in the parsing but in the overall behavior (cloning, formatting, etc.) seems pretty nasty. 

Let's do simple.

jcollie:
Simplified, all the PR does now is add a simple `set_font_size` keybind action.

src/input/command.zig
```diff
@@ -161,6 +162,19 @@ fn actionCommands(action: Action.Key) []const Command {
...
+                .set_font_size = switch (builtin.os.tag) {
+                    // On macOS we default a little bigger since this tends to look better. This
+                    // is purely subjective but this is easy to modify.
+                    .macos => 13,
+                    else => 12,
+                },
+            },
+            .title = "Set Font Size",
+            .description = "Reset the font size to the default.",
+        }},
```
mitchellh: We already have a reset font size command so we don't need this. I'll remove it.

mitchellh (approved):
Thanks!

PR #7819, opened 2025-07-06
refactor nerd font codegen script

trag1c:
Mainly added type annotations, cleaned up weird AI slop and used more idiomatic stuff.
This requires Python 3.12 to run (I can downgrade it if need be).

### DIFF
--- a/src/font/nerd_font_codegen.py
+++ b/src/font/nerd_font_codegen.py
@@ -1,102 +1,122 @@
 """
-This file is mostly vibe coded because I don't like Python. It extracts the
-patch sets from the nerd fonts font patcher file in order to extract scaling
-rules and attributes for different codepoint ranges which it then codegens
-in to a Zig file with a function that switches over codepoints and returns
-the attributes and scaling rules.
+This file extracts the patch sets from the nerd fonts font patcher file in order to
+extract scaling rules and attributes for different codepoint ranges which it then
+codegens in to a Zig file with a function that switches over codepoints and returns the
+attributes and scaling rules.
 
-This does include an `eval` call! This is spooky, but we trust
-the nerd fonts code to be safe and not malicious or anything.
+This does include an `eval` call! This is spooky, but we trust the nerd fonts code to
+be safe and not malicious or anything.
 """
 
 import ast
 import math
-from pathlib import Path
 from collections import defaultdict
+from contextlib import suppress
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Literal, TypedDict, cast
+
+type PatchSetAttributes = dict[Literal["default"] | int, PatchSetAttributeEntry]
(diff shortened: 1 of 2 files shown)


src/font/nerd_font_codegen.py
```diff
@@ -225,23 +236,16 @@ def generate_zig_switch_arms(patch_set):
...
-    with open(path, "r", encoding="utf-8") as f:
-        source = f.read()
+    project_root = Path(__file__).resolve().parents[2]
 
+    patcher_path = project_root / "vendor" / "nerd-fonts" / "font-patcher.py"
+    source = patcher_path.read_text(encoding="utf-8")
     patch_set = extract_patch_set_values(source)
 
-    out_path = Path(__file__).resolve().parent / "nerd_font_attributes.zig"
+    out_path = project_root / "src" / "font" / "nerd_font_attributes.zig"
```
jcollie: Integrating this into the build system will be easier if this just outputs to stdout.

src/font/nerd_font_codegen.py
```diff
@@ -225,23 +236,16 @@ def generate_zig_switch_arms(patch_set):
...
-        / "vendor"
-        / "nerd-fonts"
-        / "font-patcher.py"
-    )
-    with open(path, "r", encoding="utf-8") as f:
-        source = f.read()
+    project_root = Path(__file__).resolve().parents[2]
 
+    patcher_path = project_root / "vendor" / "nerd-fonts" / "font-patcher.py"
+    source = patcher_path.read_text(encoding="utf-8")
```
jcollie: Integrating this into the build system will be easier if it reads the source from stdin.

jcollie (approved):
I didn't look closely at the logic, as long as the output is identical to what was in @qwerasd205 's original commit I think we're good. A couple of notes about integrating this into the build system.

jcollie:
Also, Python 3.12 should be fine as that's the current default in Nix.

src/font/nerd_font_codegen.py
```diff
@@ -225,23 +236,9 @@ def generate_zig_switch_arms(patch_set):
...
-        source = f.read()
-
+    source = sys.stdin.read()
     patch_set = extract_patch_set_values(source)
-
-    out_path = Path(__file__).resolve().parent / "nerd_font_attributes.zig"
-
-    with open(out_path, "w", encoding="utf-8") as f:
-        f.write("""//! This is a generate file, produced by nerd_font_codegen.py
+    print("""//! This is a generated file, produced by nerd_font_codegen.py
```
00-kat, line 241: Random style suggestion:
```suggestion
    print("""\
//! This is a generated file, produced by nerd_font_codegen.py
```

00-kat (approved):

mitchellh (approved):
Looks like it generates the same data, happy with it. Thank you!

PR #7826, opened 2025-07-06
gtk: don't allow focusing on tab overview or menu buttons

jcollie:
This prevents GTK from focusing on those buttons, which would prevent focus from returning to the terminal after those buttons had been used.

### DIFF
--- a/src/apprt/gtk/Window.zig
+++ b/src/apprt/gtk/Window.zig
@@ -214,6 +214,7 @@ pub fn init(self: *Window, app: *App) !void {
     {
         const btn = gtk.MenuButton.new();
         btn.as(gtk.Widget).setTooltipText(i18n._("Main Menu"));
+        btn.as(gtk.Widget).setCanFocus(0);
         btn.setIconName("open-menu-symbolic");
         btn.setPopover(self.titlebar_menu.asWidget());
         _ = gobject.Object.signals.notify.connect(
@@ -253,6 +254,7 @@ pub fn init(self: *Window, app: *App) !void {
             },
         };
 
+        btn.setCanFocus(0);
         btn.setFocusOnClick(0);
         self.headerbar.packEnd(btn);
     }


pluiedev:
Hmm... how would keyboard-only users open the menu then? (Tab overview at least has a bindable shortcut)

jcollie:
> Hmm... how would keyboard-only users open the menu then? (Tab overview at least has a bindable shortcut)

No idea, but it's incredibly annoying to me to use the menu and not having my terminal window be focused again after doing something with the menu. ISTR that we don't get a reliable indication that the menu has closed (unlike the context menu, which we do get a signal).

How does one open the menu with the keyboard? Not something I ever do so I have no idea how one would do that intentionally.

mitchellh (approved):
We used to do this at one point, I'm guessing it regressed during one of the refactors. I agree it's incredibly annoying and something I run into too. I think the solution if this ends up being an issue for a not-tiny amount of people (I doubt it, though) is to make keybind actions to open the menu and/or worst case configurability but this is kind of a weird thing to configure.

Fwiw on macOS we also do stuff like this. It doesn't affect a11y or anything because a11y tools ignore this (on macOS). For GTK, Ghostty isn't accessible anyways (the terminal screen isn't readable) so its not really an issue right now.

tristan957:
Logging my disagreement with this PR. Ptyxis is designed by the GNOME Design Team and this is not the behavior at all. Would love to see this reverted.

I think this problem can be solved with a new action: `focus_surface` that can be bound to a keybind for people who disagree with the original behavior.

mitchellh:
I want to wait and see what the community reaction is on this. I've accidentally focused the buttons in the header way too many times. I disagree with Ptyxis if they do this, but want to see if this actually bites more people. I guess my question would be: do _you_ actually use the keyboard to navigate and use those buttons? (regardless of what other apps do)

tristan957:
This pattern is not coherent with the rest of my system (GNOME 48), and sometimes I do hit the menu button. The surface grabbing focus after clicking the menu button is confusing to me, personally. I went ahead and checked GNOME Console, and it behaves the same as Ptyxis. I had to go all the way back to GNOME Terminal to get the same behavior as this PR.

mitchellh:
I'm going to wait to see if others complain about this behavior. I don't really care about what other Gnome apps do in this case, only what users feel is right.

PR #7827, opened 2025-07-06
build: generate nerd font attributes during the build

jcollie:
Rather than committing a generated Zig file, build it on the fly during the build.


mitchellh (changes requested):
Since the patcher contains an eval, this isn’t something I want to do. Sorry. Additonally, This isn’t a file that will change often so I’d rather just keep it checked in for now. Finally, we’d probably want to add this to the dist anyways so source tarballs don’t need python.

kenvandine (approved):
The snap change looks good

PR #7828, opened 2025-07-06
Add termio to terminal CODEOWNERS group

gpanders:
The termio directory contains the implementation of many terminal features that those in the terminal reviewer group may want to be notified about.

---

An example of a terminal related PR that only changed files in the `termio` directory (and thus didn't assign @ghostty-org/terminal): https://github.com/ghostty-org/ghostty/pull/7725

I included the entire `termio` directory. If others in the @ghostty-org/terminal group think this is too broad we can narrow it down.


mitchellh (approved):
It might be too broad but lets start here.

PR #7831, opened 2025-07-06
Add link-previews option

gpanders:
Implement the `link-previews` option from https://github.com/ghostty-org/ghostty/discussions/7727.

The feature request isn't accepted yet, but I had a bit of free time to write this up and it was pretty quick, so I figure there's no harm in proposing an implementation.

Demo of `link-previews = osc8`:

https://github.com/user-attachments/assets/17a72ab2-c727-4a85-9edd-9b6e7da05d98

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1046,6 +1046,14 @@ link: RepeatableLink = .{},
 /// `link`). If you want to customize URL matching, use `link` and disable this.
 @"link-url": bool = true,
 
+/// Show link previews for a matched URL.
+///
+/// When true, link previews are shown for all matched URLs. When false, link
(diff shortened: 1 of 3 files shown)


jcollie (approved):
LGTM

src/config/Config.zig
```diff
@@ -4326,6 +4334,12 @@ pub const WindowSubtitle = enum {
     @"working-directory",
 };
 
+pub const LinkPreviews = enum {
```
mitchellh, line 4337: I foresee this one day maybe being a packed struct to enable/disable difference cases, but for now this is fine!

mitchellh (approved):
I'm good with it!

PR #7833, opened 2025-07-06
keybind: add copy_title_to_clipboard action

### LINKED ISSUE #7829: Keybind action to copy terminal title
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7825

<sup>Originally posted by **sj26** July  5, 2025</sup>
I'm not sure if this is a good idea or not, but I found myself today wanting to copy the title of a tab (into the macOS pasteboard):

image

Would a "copy title" option here be useful? I'd probably have a go at implementing if so.</div>

## Proposal

Let's introduce a `copy_title` keybind action that copies the current terminal title. This doesn't even need apprt integration (beyond invoking copy clipboard actions that already exist) because the title is known to the core.

The documentation can be that this will specifically copy the current terminal title _as set by terminal programs_ (i.e. not manually). If a terminal title is NOT set, then this should do NOTHING (do not clear the clipboard).

jcollie:
Fixes #7829

This will copy the terminal title to the clipboard. If the terminal title is not set or the title is empty the action has no effect.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4445,6 +4445,7 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
 
         .copy_title_to_clipboard => {
             const title = self.rt_surface.getTitle() orelse return false;
+            if (title.len == 0) return false;
 
             self.rt_surface.setClipboardString(title, .standard, false) catch |err| {
                 log.err("error copying title to clipboard err={}", .{err});
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4443,6 +4443,17 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
             return false;
         },
 
+        .copy_title => {
+            const title = self.rt_surface.getTitle() orelse return false;
+
+            self.rt_surface.setClipboardString(title, .standard, false) catch |err| {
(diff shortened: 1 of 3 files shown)


src/input/Binding.zig
```diff
@@ -1005,6 +1009,7 @@ pub const Action = union(enum) {
             .reset,
             .copy_to_clipboard,
             .copy_url_to_clipboard,
+            .copy_title,
```
jparise: For consistency?

```suggestion
            .copy_title_to_clipboard,
```

jcollie: Makes sense. Fixed.

src/Surface.zig
```diff
@@ -4443,6 +4443,17 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
             return false;
         },
 
+        .copy_title_to_clipboard => {
+            const title = self.rt_surface.getTitle() orelse return false;
+
```
mitchellh, line 4449: We should ignore empty titles too (length zero). I don't want to clobber the clipboard for an empty title. That was noted in the original issue.

mitchellh (changes requested):

src/Surface.zig
jcollie: Fixed.

mitchellh (approved):

PR #7835, opened 2025-07-06
linux/kde: add KDE shortcut to desktop file

### LINKED ISSUE #7673: GTK: Add default launch keybind in KDE
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7659

<sup>Originally posted by **andrewrk** June 23, 2025</sup>
image

vs 

image

I suspect that enabling Ctrl+Alt+T as a default shortcut (or copying whatever konsole does with respect to kde) will make it trivial to swap in ghostty for konsole without the user needing to set up a shortcut.</div>

<hr>

Maintainer's note: It seems that we just need to add `X-KDE-Shortcuts=Ctrl+Alt+T` to the desktop file. I'm not sure if doing this would conflict with the existing Konsole keybind, though.

jcollie:
Fixes #7673

This adds `Ctrl+Alt+T` as a KDE shortcut to the desktop file. If Konsole is installed (or any other prorgam that has the same shortcut) the user will need to go into the KDE system settings and manually reassign the `Ctrl+Alt+T` shortcut to Ghostty.

If Ghostty is the only terminal installed that claims that shortcut KDE _should_ automatically enable the shortcut (but YMMV).

Non-KDE systems will ignore this setting and if the user desires a global shortcut to open a Ghostty window it will need to be accomplished in other ways.

### DIFF
--- a/dist/linux/app.desktop.in
+++ b/dist/linux/app.desktop.in
@@ -19,6 +19,7 @@ X-TerminalArgAppId=--class=
 X-TerminalArgDir=--working-directory=
 X-TerminalArgHold=--wait-after-command
 DBusActivatable=true
+X-KDE-Shortcuts=Ctrl+Alt+T
 
 [Desktop Action new-window]
 Name=New Window


pluiedev (approved):
I think just *offering* a one-click (mostly) solution to do this would already be an improvement to the status quo, especially considering this is just a one-line fix

jcollie:
> I think just _offering_ a one-click (mostly) solution to do this would already be an improvement to the status quo, especially considering this is just a one-line fix

Yeah, if/when this is accepted I plan on adding docs to the tip website on how to enable this in KDE.

Issue #7649, opened 2025-06-22
Use native GUI to show child process exited with `wait-after-command`


mitchellh:
Today, when `wait-after-command` is set, Ghostty injects that the process has exited directly into the terminal screen, as shown below. This is true on both macOS and in GTK. 

![Image](https://github.com/user-attachments/assets/0026a571-3260-4af9-84fb-a98a6956a469)

Gnome Console uses this native banner instead (below). Every terminal I've tried on macOS does what Ghostty does, surprisingly!

<img width="693" alt="Image" src="https://github.com/user-attachments/assets/a7f0d0fc-9d07-4856-ac76-d82676d007ab" />

I think this is a lot more pleasant. And using a native widget to highlight information like this is directly in line with Ghostty's philosophy. We should do it, too. This issue is to track that. We should do this on both macOS and GTK. 

The exact styling doesn't need to match Gnome Console, and it doesn't need to match across platforms, but it should probably be somewhat similar. We can bikeshed that styling here since it's a detail. There's more work that needs to be done to plumb this information around in core, too.

For styling, I do like the bottom-banner. For copy, I think we should continue to use what we have: "Command exited. Press any key to close the terminal."

mitchellh:
As a note, with #7705 merged, this became a lot easier. All our logic for handling child exit scenarios is now in the apprt thread (Surface.zig). Next steps to implement this would be to implement an apprt action to handle the various cases, and using the performed return value to determine whether to fallback to our existing approach.

mitchellh:
This is done on GTK, so labeling this macOS now.

PR #7836, opened 2025-07-06
core/gtk: add apprt action to show native GUI warning when child exits

jcollie:
Addresses #7649 for the core and GTK. macOS support will need to be added later.

This adds an apprt action to show a native GUI warning of some kind when the child process of a terminal exits.

Also adds a basic GTK implementation of this. In GTK it overlays an Adwaita banner at the bottom of the window (similar to the banner that shows up in at the top of windows in debug builds).
Screenshot From 2025-07-08 14-49-59
Screenshot From 2025-07-08 14-50-14

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1011,19 +1011,21 @@ fn childExited(self: *Surface, info: apprt.surface.Message.ChildExited) void {
 
         log.warn("abnormal process exit detected, showing error message", .{});
 
-        // Update our terminal to note the abnormal exit. In the future we
-        // may want the apprt to handle this to show some native GUI element.
-        self.childExitedAbnormally(info) catch |err| {
-            log.err("error handling abnormal child exit err={}", .{err});
-            return;
-        };
-
-        _ = self.rt_app.performAction(
+        // Try and show a GUI message. If it returns true, don't do anything else.
+        if (self.rt_app.performAction(
             .{ .surface = self },
             .show_child_exited,
             info,
-        ) catch |err| {
+        ) catch |err| gui: {
             log.err("error trying to show native child exited GUI err={}", .{err});
+            break :gui false;
+        }) return;
+
+        // If a native GUI notification was not showm. update our terminal to
+        // note the abnormal exit.
+        self.childExitedAbnormally(info) catch |err| {
+            log.err("error handling abnormal child exit err={}", .{err});
+            return;
         };
 
         return;
@@ -1036,6 +1038,18 @@ fn childExited(self: *Surface, info: apprt.surface.Message.ChildExited) void {
     // surface then they will see this message and know the process has
     // completed.
     terminal: {
+        // First try and show a native GUI message.
+        if (self.rt_app.performAction(
+            .{ .surface = self },
+            .show_child_exited,
+            info,
+        ) catch |err| gui: {
+            log.err("error trying to show native child exited GUI err={}", .{err});
+            break :gui false;
+        }) break :terminal;
+
+        // If the native GUI can't be shown, display a text message in the
+        // terminal.
         self.renderer_state.mutex.lock();
         defer self.renderer_state.mutex.unlock();
         const t: *terminal.Terminal = self.renderer_state.terminal;
@@ -1051,14 +1065,6 @@ fn childExited(self: *Surface, info: apprt.surface.Message.ChildExited) void {
         t.modes.set(.disable_keyboard, false);
         t.screen.kitty_keyboard.set(.set, .{});
     }
-
-    _ = self.rt_app.performAction(
-        .{ .surface = self },
-        .show_child_exited,
-        info,
-    ) catch |err| {
-        log.err("error trying to show native child exited GUI err={}", .{err});
-    };
 
     // Waiting after command we stop here. The terminal is updated, our
     // state is updated, and now its up to the user to decide what to do.
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1018,6 +1018,14 @@ fn childExited(self: *Surface, info: apprt.surface.Message.ChildExited) void {
             return;
         };
 
+        _ = self.rt_app.performAction(
+            .{ .surface = self },
+            .show_child_exited,
+            info,
+        ) catch |err| {
+            log.err("error trying to show native child exited GUI err={}", .{err});
+        };
+
         return;
     }
 
@@ -1043,6 +1051,14 @@ fn childExited(self: *Surface, info: apprt.surface.Message.ChildExited) void {
         t.modes.set(.disable_keyboard, false);
         t.screen.kitty_keyboard.set(.set, .{});
     }
+
+    _ = self.rt_app.performAction(
+        .{ .surface = self },
+        .show_child_exited,
+        info,
+    ) catch |err| {
+        log.err("error trying to show native child exited GUI err={}", .{err});
+    };
 
     // Waiting after command we stop here. The terminal is updated, our
     // state is updated, and now its up to the user to decide what to do.
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -2513,7 +2513,7 @@ pub fn showChildExited(self: *Surface, info: apprt.surface.Message.ChildExited) 
     warning_box.as(gtk.Widget).setValign(.end);
 
     const warning_text = if (info.exit_code == 0)
-        i18n._("⚠️ Process exited normally. Press any key to close the terminal.")
+        i18n._("✔️ Process exited normally. Press any key to close the terminal.")
     else
         i18n._("⚠️ Process exited abnormally. Press any key to close the terminal.");
 
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -2522,8 +2522,6 @@ pub fn showChildExited(self: *Surface, info: apprt.surface.Message.ChildExited) 
 
     const banner_widget = banner.as(gtk.Widget);
 
-    banner_widget.addCssClass("child_exited");
-
     if (info.exit_code == 0)
         banner_widget.addCssClass("child_exited_normally")
     else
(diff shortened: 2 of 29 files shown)


src/Surface.zig
```diff
@@ -1018,6 +1018,14 @@ fn childExited(self: *Surface, info: apprt.surface.Message.ChildExited) void {
             return;
         };
 
+        _ = self.rt_app.performAction(
```
mitchellh: I think if the apprt handles this (returns true) then we should not call `childExitedAbnormally` and modify the terminal contents. This more closely matches the Gnome Terminal behavior which I like. Its just duplicate information for a well-behaved apprt.

mitchellh (changes requested):
One code change.

I think stylistically I have to think about this. 

One UX question I have is what does this do to the terminal window if there is text that takes up the full screen? Does the overlay cover it so it's unreadable or does it shrink the terminal area or how is that handled? I'm not able to test this right now.

jcollie:
> One code change.
> 
> I think stylistically I have to think about this.
> 
> One UX question I have is what does this do to the terminal window if there is text that takes up the full screen? Does the overlay cover it so it's unreadable or does it shrink the terminal area or how is that handled? I'm not able to test this right now.

Right now the banner will cover up anything at the bottom of the screen. I'm looking at making the banner transparent. Adding a button to dismiss the banner would be easy enough too.

mitchellh:
I'd be curious what Gnome Terminal does. I think transparency won't solve things enough because it may be useful to copy text back there. And I'm not sure about dismissing either because it's unclear what state the window is in at that state. This requires some more thought (and is the main reason I haven't pursued it yet). You may want to open a discussion so we can do design discussion in something other than a PR, and hold this PR until then.

jcollie:
Got transparency working... Tested selecting text that's behind the window and the overlay has no effect.

image
image

mitchellh:
Oh okay, the overlay doesn't impact selection, that's maybe workable... I'll have to try it!

src/Surface.zig
jcollie: OK, this is implemented. Will update the screenshots in the first message.

jcollie:
One note, not sure how this affects restoring from a saved state. GTK doesn't do that so I have no way to test. There may need to be some more work that needs to be done to make sure that the appropriate banner is shown again when a surface is restored.

src/Surface.zig

src/apprt/gtk/Surface.zig
```diff
@@ -2502,3 +2502,35 @@ fn gtkStreamError(media_file: *gtk.MediaFile, _: *gobject.ParamSpec, _: ?*anyopa
 fn gtkStreamEnded(media_file: *gtk.MediaFile, _: *gobject.ParamSpec, _: ?*anyopaque) callconv(.c) void {
     media_file.unref();
 }
+
+pub fn showChildExited(self: *Surface, info: apprt.surface.Message.ChildExited) (error{})!bool {
```
pluiedev: Honestly `(error{})!bool` is really funny looking. Couldn't it just be `error{}!bool`?

src/apprt/gtk/Surface.zig
```diff
@@ -2502,3 +2502,35 @@ fn gtkStreamError(media_file: *gtk.MediaFile, _: *gobject.ParamSpec, _: ?*anyopa
...
+    if (!adw_version.supportsBanner()) return false;
+
+    const warning_box = gtk.Box.new(.vertical, 0);
+
+    warning_box.as(gtk.Widget).setHalign(.fill);
+    warning_box.as(gtk.Widget).setValign(.end);
+
+    const warning_text = if (info.exit_code == 0)
+        i18n._("⚠️ Process exited normally. Press any key to close the terminal.")
```
pluiedev: It's a bit strange that the normal path still uses the `⚠️` emoji... Maybe ✔️ could work better?

src/apprt/gtk/Surface.zig
```diff
@@ -2502,3 +2502,35 @@ fn gtkStreamError(media_file: *gtk.MediaFile, _: *gobject.ParamSpec, _: ?*anyopa
...
+    else
+        i18n._("⚠️ Process exited abnormally. Press any key to close the terminal.");
+
+    const banner = adw.Banner.new(warning_text);
+    banner.setRevealed(1);
+
+    const banner_widget = banner.as(gtk.Widget);
+
+    banner_widget.addCssClass("child_exited");
```
pluiedev: Is this CSS class ever used anywhere?

tristan957:
I don't love the transparency. It really hurts readability. I love the coloring to signify success, but green/red is not accessible for certain variations of color blindness. Does libadwaita abstract this issue away? I'm actually not aware of color blindness being something that is stored in the settings portal.

I don't understand why the banner is overlaid.

tristan957:
What happens if your mouse hovers a URL while this banner is visible?

src/apprt/gtk/Surface.zig
```diff
@@ -2502,3 +2502,35 @@ fn gtkStreamError(media_file: *gtk.MediaFile, _: *gobject.ParamSpec, _: ?*anyopa
...
+
+    if (info.exit_code == 0)
+        banner_widget.addCssClass("child_exited_normally")
+    else
+        banner_widget.addCssClass("child_exited_abnormally");
```
pluiedev: Honestly I probably would just do the dumb route and have two copies of the banner as blueprints, and just hardcode the style and the label text

pluiedev (changes requested):

jcollie:
> I don't love the transparency. It really hurts readability. I love the coloring to signify success, but green/red is not accessible for certain variations of color blindness. Does libadwaita abstract this issue away? I'm actually not aware of color blindness being something that is stored in the settings portal.

It's not reflected in the screenshots but one of the last commits changed the verbiage to include whether the exit was normal/abnormal, specifically because of color blindness and screen reader concerns.

This is custom CSS adding the color and transparency.

> I don't understand why the banner is overlaid.

1) It's easy because we already have an overlay.
2) I didn't like the idea of adding a widget to the bottom of the terminal as that would change the height of the window/tab/split or the terminal. That would be kinda janky I think as it might cause a reflow, esp with multiple splits. Tiled window manager behavior would be odd I think.

src/apprt/gtk/Surface.zig
jcollie: Not at the moment.

src/apprt/gtk/Surface.zig
jcollie: Not sure, but I've had problems trying to put an explicit error set into a function signature before.

src/apprt/gtk/Surface.zig
jcollie: Hmmm maybe, I can see arguments for both.

pluiedev:
I personally think we should make the success case have the default background and keep the red background for the error case. That creates a nice contrast that shouldn't be problematic for colorblind users.

jcollie:
> What happens if your mouse hovers a URL while this banner is visible?

The link preview appears above the banner.

src/apprt/gtk/Surface.zig
jcollie: As far as the text goes, I'd prefer to keep that in Zig. Once the I18N system can handle it I'd like to be able to include more information, like runtime, non-zero exit code, killed by signal, etc.

src/apprt/gtk/Surface.zig
jcollie: Switched to ✔️

src/apprt/gtk/Surface.zig
jcollie: Yeah that works just fine, not sure why I ran into problems before.

tristan957:
I think that we should just drop the emojis. Using color looks much better.

src/apprt/gtk/Surface.zig
jcollie: Removed.

jcollie:
FYI, this is what it looks like when running multiple splits:
image

pluiedev:
I wonder if we could make use of the banner's [optional button](https://developer.gnome.org/hig/patterns/feedback/banners.html) as a way to more intuitively indicate that the split can be safely closed — perhaps something like `Process exited abnormally [ Close ]`

jcollie:
> I wonder if we could make use of the banner's [optional button](https://developer.gnome.org/hig/patterns/feedback/banners.html) as a way to more intuitively indicate that the split can be safely closed — perhaps something like `Process exited abnormally [ Close ]`

I mentioned that above, but I was thinking more of as "dismiss banner" button. Closing the surface shouldn't be hard to do...

jcollie:
Here's what the close button looks like:

image
image

pluiedev:
I really like how that turned out — still probably would advocate for using the default color for the success case, though. The text also shouldn't end with any punctuations per the [GNOME HIG](https://developer.gnome.org/hig/guidelines/writing-style.html#heading-style).

Also, maybe it's just me as someone who speaks English as a second language, but the difference between "normally" and "abnormally" is kind of small at a glance. Maybe "Command succeeded" vs "Command failed (with code xyz)" would offer a better contrast? (Process also IMO is kind of an opaque word, especially since we use "command" everywhere in the config options like in `command` or `wait-after-command`.) This could also give way to "Command terminated by signal" and other cases we might want to indicate.

Unrelated: this gave me an idea that maybe we should allow users to run ephemeral tasks in a new split/tab/window, which can be closed by a handy button like this 🤔

jcollie:
> Also, maybe it's just me as someone who speaks English as a second language, but the difference between "normally" and "abnormally" is kind of small at a glance. Maybe "Command succeeded" vs "Command failed (with code xyz)" would offer a better contrast? (Process also IMO is kind of an opaque word, especially since we use "command" everywhere in the config options like in `command` or `wait-after-command`.) This could also give way to "Command terminated by signal" and other cases we might want to indicate.

I'd _like_ to do "Command failed (code xyz)" and similar but then we can't translate that text can we? At least not without developing our own templating system since Zig formatters require a comptime string.

pluiedev:
Yeah that's slightly annoying, but we can leave that as a future improvement. Right now we should focus on the code=0 vs everything else distinction

mitchellh:
I'm good with the look of this.

mitchellh (approved):
Approve pending @pluiedev

pluiedev (approved):

pluiedev:
Looks like there's a merge conflict... :p

jcollie:
> Looks like there's a merge conflict... :p

Yup. Fixed. Waiting for CI to be green.

PR #7840, opened 2025-07-06
Fallback Font Size Adjustment

qwerasd205:
This PR changes `font.Collection` to automagically adjust the sizes of added fonts so that their metrics (specifically their ex height, or their ideograph character width if they have one) match the primary font. This is like [`font-size-adjust`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust) from CSS.

This is a big win for users who use mixed writing systems and rely heavily on fallback fonts. For example, in #7774 it's pointed out that CJK characters are not very well harmonized with existing Latin glyphs, well:
|Before (`main`)|After (this PR)|
|-|-|
|image|image|

This also improves our handling of the horizontal alignment of fallback glyphs. It's not an ideal solution; it only works for glyphs narrower than the cell width because it messes with ligatures if we include glyphs wider than the cell width; and most things would look better if the center were proportionally remapped based on the ratio from the glyph advance to the cell width, but that messes with glyphs designed to align vertically so it can't be done, instead the original advance width is centered in the cell width.

> [!NOTE]
> I can add a couple tests for the size adjust behavior (one for ex height, one for ic width) if you'd like.

### Future work
When working on this I was very bothered by the fact that our fallback fonts are not selected deterministically. That is, if a fallback with glyphs A and B is selected as the fallback for B because no other font has it, then it will automatically be used as the fallback for A as well even if other fonts would otherwise rank higher. This is a big problem if you have something like UniFont installed, since the moment that's loaded no other font will be loaded afterwards and you get very suboptimal selections for your glyphs.

So I'll have a think on how to go about making our fallbacks deterministic without killing performance (it's not a trivial problem unfortunately) and that will probably be my next area of work.

### DIFF
--- a/src/font/Collection.zig
+++ b/src/font/Collection.zig
@@ -69,10 +69,14 @@ pub fn deinit(self: *Collection, alloc: Allocator) void {
     if (self.load_options) |*v| v.deinit(alloc);
 }
 
-pub const AddError = Allocator.Error || error{
-    CollectionFull,
-    DeferredLoadingUnavailable,
-};
+pub const AddError =
+    Allocator.Error ||
+    AdjustSizeError ||
+    error{
+        CollectionFull,
+        DeferredLoadingUnavailable,
+        SetSizeFailed,
+    };
 
 /// Add a face to the collection for the given style. This face will be added
 /// next in priority if others exist already, i.e. it'll be the _last_ to be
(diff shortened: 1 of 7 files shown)


mitchellh (approved):
I'll let you merge this one when you're ready in case there are any adjustments you wanted. Looks great!

qwerasd205:
Thanks, I went to add a test and the test found a bug haha, got that fixed, cleaning it up and will update the PR then merge.

ken-wizhodl:
this PR literaly ruined Chinese charaters, please revert

00-kat:
Many disagree: https://github.com/ghostty-org/ghostty/discussions/8422#discussioncomment-14235659. See also: #7929 (the title is misleading but it does discuss this issue).

PR #7843, opened 2025-07-07
macos: open URLs with NSWorkspace APIs instead of `open`

### LINKED ISSUE #5256: macOS: Use system API to open links

mitchellh:
Fixes #5256

This updates the macOS apprt to implement the `OPEN_URL` apprt action to use the NSWorkspace APIs instead of the `open` command line utility.

As part of this, we removed the `ghostty_config_open` libghostty API and instead introduced a new `ghostty_config_open_path` API that returns the path to open, and then we use the `NSWorkspace` APIs to open it (same function as the `OPEN_URL` action).

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -351,6 +351,11 @@ typedef struct {
 } ghostty_diagnostic_s;
 
 typedef struct {
+  const char* ptr;
+  uintptr_t len;
+} ghostty_string_s;
+
+typedef struct {
   double tl_px_x;
   double tl_px_y;
   uint32_t offset_start;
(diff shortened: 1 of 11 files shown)


mitchellh:
Added a GTK change so that GTK's open config ALSO goes through its `openURL` mechanism.

jcollie (approved):
GTK changes LGTM.

PR #7848, opened 2025-07-07
font/coretext: fix horizontal bearing calculation

qwerasd205:
This was subtly wrong in a way that was most obvious when text switched from regular to bold, where it would seem to wiggle since the bearings of each letter would shift by a pixel in either direction. This affected applications like fzf which uses bold to dynamically highlight the line you have selected.


qwerasd205:
Unclear what's causing that build failure in CI...

jcollie:
> Unclear what's causing that build failure in CI...

Heisenbug... re-running the test seems to have fixed it.

mitchellh (approved):

PR #7850, opened 2025-07-07
Nerd Font Icon Height Constraint

qwerasd205:
Nerd font icons were ***WAY*** too big depending on your font setup, this is because we were always using the full cell height when the nerd font patcher instead uses an "icon height" for most things. The patcher calculates the icon height as two thirds of the font's cap height and one third of the line height, but I've chosen to instead use 1.2 times the cap height for more consistent results across fonts-- if the user wants their icons bigger, they can use the `adjust-icon-height` metric modifier (and they can also use it to make them smaller if they want that for some reason).

I also adjusted the attributes to user horizontal cover + vertical fit for `^` stretch modes (proportional scaling but scale up), which makes it so that it never exceeds the cell size, since first it covers horizontally and then scales down to fit vertically if necessary; previously, if there were a particularly wide glyph that was scaled with cover/cover it would exceed the available width and overflow in to neighboring cells which wasn't good.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -396,6 +396,18 @@ pub const compatibility = std.StaticStringMap(
 /// Thickness in pixels or percentage adjustment of box drawing characters.
 /// See the notes about adjustments in `adjust-cell-width`.
 @"adjust-box-thickness": ?MetricModifier = null,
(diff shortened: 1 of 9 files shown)


mitchellh (approved):
Pending CI

PR #7851, opened 2025-07-07
build: improve caching of macOS builds

mitchellh:
This improves caching of builds to speed up `zig build` and `zig build run` on macOS by avoiding unnecessary rebuilds of the Ghostty xcframework.

I had bigger ambitions to avoid the `xcodebuild` as well but I kept running into issues where the Zig build system wasn't getting the hit/miss right. This is surely my problem with setting the proper inputs, but I couldn't figure it out so just want to get this in to start.


mitchellh:
Let me work on this more.

PR #7859, opened 2025-07-08
core: document which release added config entries

jcollie:
For each config entry, add a comment specifying in which release it was first added. In some cases I note when certain aspects of each config entry were modified.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -126,8 +126,6 @@ pub const compatibility = std.StaticStringMap(
 ///
 /// Changing this configuration at runtime will only affect new terminals, i.e.
 /// new windows, tabs, etc.
-///
-/// Available since: 1.0.0
 @"font-family": RepeatableString = .{},
 @"font-family-bold": RepeatableString = .{},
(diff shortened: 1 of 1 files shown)


jcollie:
Did my best searching through old PRs and `git blame` but it's possible I missed something.

src/config/Config.zig
```diff
@@ -126,6 +126,8 @@ pub const compatibility = std.StaticStringMap(
 ///
 /// Changing this configuration at runtime will only affect new terminals, i.e.
 /// new windows, tabs, etc.
+///
+/// Available since: 1.0.0
```
mitchellh: I think we can remove the 1.0.0 ones and have that be implied. That'll help clean up a lot of the noise in the docs and here.

mitchellh (changes requested):
One request that should be a simple `sed` but thanks for doing this!

src/config/Config.zig
jcollie: Done.

mitchellh (approved):

PR #7861, opened 2025-07-08
Hopefully Last Nerd Fonts Tweak

qwerasd205:
With these two changes (see commits for details) our scaling should be identical to the nerd font patcher in all the ways that matter and even slightly better in some others.

I'm really hoping this is the last change I have to make to the nerd font scaling...


mitchellh (approved):

src/font/face.zig
```diff
@@ -216,6 +233,22 @@ pub const RenderOptions = struct {
...
+            //
+            // But we *only* do this if the constraint width is 2, since
+            // otherwise it would make them way too small when sized for
+            // a single cell.
+            const is_icon_width = self.height == .icon and @min(self.max_constraint_width, constraint_width) > 1;
+            const orig_avail_width = available_width;
+            if (is_icon_width) {
+                const cell_height: f64 = @floatFromInt(metrics.cell_height);
+                const ratio = available_height / cell_height;
+                available_width *= ratio;
```
danielwe, line 249: Some post-merge review: We discovered in #7905 that this logic leads to _smaller_ icons when `adjust-cell-height` is positive, because `cell_height` goes up while `available_height` stays fixed (linked to cap height); thus, `ratio` goes down. That doesn't seem right.

I think `ratio` should just be a constant here, and from some quick experimenting, I think `0.75` is a good value for getting just enough margin to neighboring text, say, a folder icon next to a folder name. Happy to make the PR if that sounds OK. (I might also propose a different heuristic for `icon_height`, but that's a separate discussion.)

PR #7864, opened 2025-07-09
gtk: open non-file URLs using the XDG Desktop Portal

### LINKED ISSUE #5991: GTK: use portals to open URLs instead of xdg-open

_Originally posted by @tristan957 in https://github.com/ghostty-org/ghostty/issues/5988#issuecomment-2682937459_

jcollie:
Use the XDG Desktop Portal OpenURI DBus method to open non-file URLs. File URLs use a separate DBus method and are not handled here. If the DBus method call fails or the URL is a file URL, use the cross-platform fallback to open the URL.

### DIFF
--- a/src/apprt/gtk/portal.zig
+++ b/src/apprt/gtk/portal.zig
@@ -1,0 +1,52 @@
+const std = @import("std");
+
+const gio = @import("gio");
+
+const Allocator = std.mem.Allocator;
+
+pub const OpenURI = @import("portal/OpenURI.zig");
+
+/// Generate a token suitable for use in requests to the XDG Desktop Portal
+pub fn generateToken() usize {
+    return std.crypto.random.int(usize);
+}
+
+/// Get the XDG portal request path for the current Ghostty instance.
+///
+/// If this sounds like nonsense, see `request` for an explanation as to
+/// why we need to do this.
+pub fn getRequestPath(alloc: Allocator, dbus: *gio.DBusConnection, token: usize) (Allocator.Error || error{NoDBusUniqueName})![:0]const u8 {
+    // See https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html
+    // for the syntax XDG portals expect.
+
+    const token_string = try std.fmt.allocPrint(alloc, "{x:0>16}", .{token});
+    defer alloc.free(token_string);
+
+    const unique_name = try alloc.dupe(u8, std.mem.span(dbus.getUniqueName() orelse {
+        return error.NoDBusUniqueName;
+    })[1..]);
+    defer alloc.free(unique_name);
+
+    // Sanitize the unique name by replacing every `.` with `_`.
+    // In effect, this will turn a unique name like `:1.192` into `1_192`.
+    std.mem.replaceScalar(u8, unique_name, '.', '_');
+
+    const object_path = try std.mem.joinZ(
+        alloc,
+        "/",
+        &.{
+            "/org/freedesktop/portal/desktop/request",
+            unique_name, // Remove leading `:`
+            token_string,
+        },
+    );
+
+    return object_path;
+}
+
+/// Try and parse the token out of a request path.
+pub fn parseRequestPath(request_path: []const u8) ?usize {
+    const index = std.mem.lastIndexOfScalar(u8, request_path, '/') orelse return null;
+    const token = request_path[index + 1 ..];
+    return std.fmt.parseUnsigned(usize, token, 16) catch return null;
+}
--- a/src/apprt/gtk/portal.zig
+++ b/src/apprt/gtk/portal.zig
@@ -15,20 +15,23 @@ pub fn generateToken() usize {
 ///
 /// If this sounds like nonsense, see `request` for an explanation as to
 /// why we need to do this.
-pub fn getRequestPath(alloc: Allocator, dbus: *gio.DBusConnection, token: usize) (Allocator.Error || error{NoDBusUniqueName})![:0]const u8 {
+pub fn getRequestPath(alloc: Allocator, dbus: *gio.DBusConnection, token: usize) (Allocator.Error || std.fmt.BufPrintError || error{NoDBusUniqueName})![:0]const u8 {
     // See https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html
     // for the syntax XDG portals expect.
 
-    const token_string = try std.fmt.allocPrint(alloc, "{x:0>16}", .{token});
-    defer alloc.free(token_string);
+    var token_buf: [16]u8 = undefined;
+    const token_string = try std.fmt.bufPrint(&token_buf, "{x:0>16}", .{token});
 
-    const unique_name = try alloc.dupe(u8, std.mem.span(dbus.getUniqueName() orelse {
-        return error.NoDBusUniqueName;
-    })[1..]);
+    // Get the unique name from D-Bus and strip the leading `:`
+    const unique_name = try alloc.dupe(u8, std.mem.span(
+        dbus.getUniqueName() orelse {
+            return error.NoDBusUniqueName;
+        },
+    )[1..]);
     defer alloc.free(unique_name);
 
-    // Sanitize the unique name by replacing every `.` with `_`.
-    // In effect, this will turn a unique name like `:1.192` into `1_192`.
+    // Sanitize the unique name by replacing every `.` with `_`. In effect, this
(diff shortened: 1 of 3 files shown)


jcollie:
Not 100% happy with this yet but I figured I'd put it out there and get some feedback. Mainly I'm not sure if I have all the error conditions handled properly and am not leaking memory.

One note: I did not use `libportal` because it's API that corresponds to the DBus method used _requires_ a window handle. We do not have a window handle at that point in the code. It's not impossible to get, but would require changing a lot of upstream code. By using a raw DBus call we can avoid the necessity for a window handle.

There's also opportunity for sharing some code with the GTK global shortcuts code. I just copied what I needed for now rather than refactoring that part of the code.

mitchellh (commented):
I'll wait for others before I review, especially those more familiar with the mechanisms involved here. From a pure code perspective this is pretty clean though.

jcollie:
OK, I've restructured this a whole lot so I'm much more sanguine that I'm not leaking any memory or will have any use-after-free problems. There may still be some edge cases around error handling that need to be taken care of but I think that this should be pretty robust.

Yes, it's quite a bit more complicated than just executing `xdg-open` but that's part of what comes with dealing with D-Bus and the XDG Portal (I agree with Leah, the Portal's API is execrable). It's also likely that we need to do it this way to allow URLs to be opened in flatpak builds.

src/apprt/gtk/portal/OpenURI.zig
```diff
@@ -0,0 +1,400 @@
...
+pub fn start(self: *OpenURI, value: apprt.action.OpenUrl) (Allocator.Error || Errors)!void {
+    const alloc = self.app.core_app.alloc;
+
+    const token = portal.generateToken();
+
+    self.mutex.lock();
+    defer self.mutex.unlock();
+
+    try self.entries.putNoClobber(alloc, token, .{
+        .start = std.time.Instant.now() catch @panic("can't get the current time!"),
```
mitchellh: I don't know how this possibly fails, but crashing a program for a failed URL open doesn't seem worth it. I'd prefer to see this turn into an error return (with proper cleanup if necessary).

src/apprt/gtk/portal/OpenURI.zig
```diff
@@ -0,0 +1,400 @@
...
+        .token = token,
+        .kind = value.kind,
+        .uri = try alloc.dupeZ(u8, value.url),
+    });
+
+    const entry = self.entries.getPtr(token) orelse {
+        return error.PutFailed;
```
mitchellh: I think we probably need some `errdefer` in here to handle cleanup in these cases.

src/apprt/gtk/portal/OpenURI.zig
```diff
@@ -0,0 +1,400 @@
...
+            {
+                const option = glib.VariantType.new("{sv}");
+                defer glib.free(option);
+
+                builder.open(option);
+                defer builder.close();
+
+                builder.add("s", "handle_token");
+
+                const token = try std.fmt.allocPrintZ(alloc, "{x:0<16}", .{entry.token});
```
mitchellh, line 217: Is token always fixed size (or a predictable max)? Can we get away with a bufAlloc here?

src/apprt/gtk/portal/OpenURI.zig
```diff
@@ -0,0 +1,400 @@
...
+
+                builder.add("s", "ask");
+
+                const ask = glib.Variant.newBoolean(@intFromBool(false));
+                builder.add("v", ask);
+            }
+        }
+
+        break :payload builder.end();
+    };
```
mitchellh, line 250: Do we need to add an errdefer after this to clean up the payload? (I know its impossible RIGHT NOW for this to trigger, but it'd make things easier one day)

src/apprt/gtk/portal/OpenURI.zig
```diff
@@ -0,0 +1,400 @@
...
+        log.warn("cleanup called with null userdata", .{});
+        return @intFromBool(glib.SOURCE_REMOVE);
+    }));
+
+    self.mutex.lock();
+    defer self.mutex.unlock();
+
+    self.cleanup_timer = null;
+
+    const now = std.time.Instant.now() catch @panic("unable to get time!");
```
mitchellh: Same thing with the panic here. I'd log a warning that there may be dangling entries but no need to crash.

src/apprt/gtk/portal/OpenURI.zig
```diff
@@ -0,0 +1,400 @@
...
+app: *App,
+
+/// Connection to the D-Bus session bus that we'll use for all of our messaging.
+dbus: *gio.DBusConnection,
+
+/// Mutex to protect modification of the entries map or the cleanup timer.
+mutex: std.Thread.Mutex = .{},
+
+/// Map to store data about any in-flight calls to the portal.
+entries: std.AutoHashMapUnmanaged(usize, Entry) = .empty,
```
mitchellh: We should use an `ArrayHashMap` to avoid the tombstone issue.

src/apprt/gtk/portal.zig
```diff
@@ -0,0 +1,52 @@
...
+            unique_name, // Remove leading `:`
+            token_string,
+        },
+    );
+
+    return object_path;
+}
+
+/// Try and parse the token out of a request path.
+pub fn parseRequestPath(request_path: []const u8) ?usize {
```
mitchellh, line 51: This deserves unit tests

mitchellh (changes requested):
All looking good, some changes requested, questions asked.

src/apprt/gtk/portal/OpenURI.zig
jcollie: The only way `std.time.Instant.now()` fails if the system it's being run on lacks timers, which is AFAIK impossible on the sorts of systems that Ghostty runs on. So maybe just `unreachable` then.

src/apprt/gtk/portal/OpenURI.zig
jcollie: Token can be any size we choose as long as it's reasonably certain that it's unique.

src/apprt/gtk/portal/OpenURI.zig
jcollie: I've added an `errdefer` that will clean that up.

src/apprt/gtk/portal/OpenURI.zig
jcollie: The token is now formatted using a static buffer.

src/apprt/gtk/portal/OpenURI.zig
jcollie: It took more than just a single `errdefer` but the error handling/cleanup is much better now.

src/apprt/gtk/portal/OpenURI.zig
jcollie: Panic is removed. If we can't get the time opening the URL via the portal will fail.

src/apprt/gtk/portal/OpenURI.zig
jcollie: Done.

src/apprt/gtk/portal/OpenURI.zig
jcollie: Done.

src/apprt/gtk/portal.zig
jcollie: Done.

src/apprt/gtk/portal/OpenURI.zig
```diff
@@ -0,0 +1,438 @@
...
+        .{},
+        responseReceived,
+        self,
+        null,
+    );
+}
+
+// Unsubscribe to the D-Bus signal that contains the result of the method call.
+// This will prevent a response from being processed multiple times. This must
+// be called when the mutex is locked.
```
tristan957: Does this need to be a `///` comment? Should we add an assert that the lock is taken?

jcollie: Done and done.

tristan957:
I'll let Leah speak as other portal person before approval :D

mitchellh:
I'm going to close this and let you port it to `-ng` @jcollie. I still think @pluiedev should review this since she has more experience in the area, too.

jcollie:
> I'm going to close this and let you port it to `-ng` @jcollie. I still think @pluiedev should review this since she has more experience in the area, too.

At this point, let's punt it to 1.3. This probably only really makes a difference for flatpak anyway.

PR #7871, opened 2025-07-09
Changed behaviour of bold-color

roberthawdon:
Continuation of discussion in https://github.com/ghostty-org/ghostty/discussions/3134

This changes the behaviour of the new bold-color option to only affect the default foreground text when set to a static colour. By using a static colour, the remaining colours are rendered as bright.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2825,8 +2825,8 @@ else
 /// This can be set to a specific color, using the same format as
 /// `background` or `foreground` (e.g. `#RRGGBB` but other formats
 /// are also supported; see the aforementioned documentation). If a
-/// specific color is set, this color will always be used for all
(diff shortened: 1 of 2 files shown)


mitchellh (approved):
I think this is a good compromise since all the requests were for a specific color to do this. I think if it comes up in the future we can introduce alternate syntax to set ALL colors to this.

ftpd:
@mitchellh thanks for approving this.

Any predictions when the new version will be released? I can build from the tip, sure - I'm just curious.

rhodes-b:
> @mitchellh thanks for approving this.
> 
> Any predictions when the new version will be released? I can build from the tip, sure - I'm just curious.

Aug/Sept is the current plan for 1.2

PR #7876, opened 2025-07-09
Improve CoreText "Quantization"

qwerasd205:
The old math didn't allow fractional pixels on the left and bottom, and stretched glyphs vertically since the height was always rounded up. At very small font sizes this looked good, but at medium and even large sizes this just made things inconsistent and janky.

These new calculations are practically pixel-identical to whatever CoreText is doing in 99% of cases, and the remaining cases seem to be some sort of auto-hinting since it's internal features of the glyph getting repositioned.

Over all, I still prefer this to CoreText's quantize option, but if this causes further issues we should probably just revert the whole thing and go ahead and add an extra pixel of padding to the bottom and left...

### DIFF
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -361,48 +361,75 @@ pub const Face = struct {
             opts.constraint_width,
         );
 
-        // We manually quantize the position and size of the glyph to whole
-        // pixel boundaries. Since macOS doesn't do font hinting this helps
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Thanks for following all this up. Overall, I'd rather have exact control over our own font rendering than rely on some undocumented opaque boolean. So, I hope this works well for us long term!

PR #7880, opened 2025-07-09
fish: prefer 'command -q' to check for commands

jparise:
This is a fish built-in 'command' option that's the more idiomatic way to check for the availability of a command.

https://fishshell.com/docs/current/cmds/command.html


jcollie (approved):

mitchellh (approved):
Approved BUT I noticed this: instead of looking for `Ghostty` on the path we should be using `$GHOSTTY_BIN_DIR/ghostty` which is set by termio to point to the running binary. This will fix up stuff like dev builds too.

jparise:
> Approved BUT I noticed this: instead of looking for `Ghostty` on the path we should be using `$GHOSTTY_BIN_DIR/ghostty` which is set by termio to point to the running binary. This will fix up stuff like dev builds too.

Yes, that makes sense for `ghostty`. I'll follow-up with that across all of the scripts.

PR #7881, opened 2025-07-09
build: fix compilation failures when using older SDKs with macOS 26 APIs

liby:
## Background

When using an SDK older than macOS Tahoe 26 (e.g., the macOS 15 SDK), the compiler cannot find types that are only defined in the macOS Tahoe 26 SDK. This results in a fatal compile-time error, not just a runtime availability warning:

```
/path/to/file.swift:XX:XX: error: Cannot find type 'IntentModes' in scope
```

This prevents the project from being built on any system that doesn't have the very latest macOS SDK installed.

## Solution

Implement compile-time SDK version checking using the Objective-C bridging header. This separates new APIs from the core codebase, allowing them to be compiled only when the required SDK is present.

1.  **Add SDK version macro** in `ghostty-bridging-header.h`:
    *   Define a new flag, `SUPPORTS_MACOS_26_FEATURES`, only when the build environment's `__MAC_OS_X_VERSION_MAX_ALLOWED` corresponds to the macOS 26.0 SDK or newer.

2.  **Wrap macOS 26 APIs** with `#if SUPPORTS_MACOS_26_FEATURES`:
    *   All App Intents' `supportedModes` properties, which use the `IntentModes` type.
    *   The `syncAppearanceTahoe` method call in `TransparentTitlebarTerminalWindow`, which is part of the new feature set.

## Why some runtime checks remain unchanged

Three instances of `#available(macOS 26.0, *)` were intentionally left unchanged:

1.  *TerminalController.swift* - ["TerminalTabsTitlebarTahoe" string selection](https://github.com/liby/ghostty/blob/main/macos/Sources/Features/Terminal/TerminalController.swift#L26-L30)
2.  *TerminalWindow.swift* - [`topPadding` calculation](https://github.com/liby/ghostty/blob/main/macos/Sources/Features/Terminal/Window%20Styles/TerminalWindow.swift#L453-L457)

These don't reference macOS 26-specific APIs, only string literals and numeric values. The runtime checks here are appropriate since they determine behavior differences between OS versions at runtime, not compile-time API availability.

### DIFF
--- a/macos/Sources/App/macOS/ghostty-bridging-header.h
+++ b/macos/Sources/App/macOS/ghostty-bridging-header.h
@@ -1,3 +1,12 @@
 // C imports here are exposed to Swift.
 
 #import "VibrantLayer.h"
+
+// SDK version check for macOS 26 features
+#import <AvailabilityMacros.h>
+
+// Check if SDK version is macOS 26 or higher
+// macOS 26.0 would be 260000 in the version macro
+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 260000
+    #define SUPPORTS_MACOS_26_FEATURES 1
+#endif
--- a/macos/Sources/Features/App Intents/CloseTerminalIntent.swift
+++ b/macos/Sources/Features/App Intents/CloseTerminalIntent.swift
@@ -12,8 +12,10 @@ struct CloseTerminalIntent: AppIntent {
     )
     var terminal: TerminalEntity
 
+    #if SUPPORTS_MACOS_26_FEATURES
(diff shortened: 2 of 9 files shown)


mitchellh:
I'm not sure the complexity of gating everything with C macro checks and Swift macro checks is a path I want to go down. I don't think there is an issue requiring our tip builds to require the beta SDK. What argument is there against this pov?

liby:
You're right, the complexity might not be worth it. My only thought was this allows developers not yet on macOS 26 to build the project. But it's not a strong preference.

PR #7887, opened 2025-07-09
shell-integration: use $GHOSTTY_BIN_DIR/ghostty

jparise:
Locate our ghostty binary using $GHOSTTY_BIN_DIR rather than searching the PATH.


jparise:
I'll update the Elvish implementation separately. It has a few other issues so I'll fix them all up together.

mitchellh (approved):

PR #7889, opened 2025-07-09
Add missing Dutch translations, fix grammar, improve flow

mtak:
- Add missing translations for Dutch
- Fix minor grammar mistakes
- Update entries to flow more naturally and in line with other Dutch apps

For Dutch reviewers: "standaardtitel" is a compound word, (two nouns combined), so it should be written without a space. Same for "terminalinspecteur", "debugversie" and "Ghostty-ontwikkelaars" (with a dash for legibility).


po/nl_NL.UTF-8.po
```diff
@@ -118,7 +119,7 @@ msgstr "Splitsen"
 #: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:33
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:45
 msgid "Change Title…"
-msgstr "Wijzig titel…"
+msgstr "Titel wijzigen…"
```
nwehg: This is not really necessary, it's in line with the other sentences you created. You can also create new sentences that are in this style. For example: "Onthoud keuze voor deze splitsing" instead of "Keuze onthouden voor deze splitsing"

po/nl_NL.UTF-8.po
```diff
@@ -160,11 +161,11 @@ msgstr "Open configuratie"
 
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:85
 msgid "Command Palette"
-msgstr ""
+msgstr "Commandopalet"
```
nwehg: I'm not sure what this is. @00-kat do you know what the command palette does? There may be a better name for it.

00-kat: Lets you run keybinding actions from a palette à la that found in VSCode or Zed:

A screenshot of the GTK command palette.

If you have a tip build you can find it with <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>p</kbd> on Linux (I'm not sure what the default keybinding is on macOS).

po/nl_NL.UTF-8.po
```diff
@@ -87,7 +88,7 @@ msgstr "Splits naar rechts"
 
 #: src/apprt/gtk/ui/1.5/command-palette.blp:16
 msgid "Execute a command…"
-msgstr ""
+msgstr "Voer een commando uit..."
```
kjvdven: I would use `…` instead of `...`

po/nl_NL.UTF-8.po
nwehg: I can't really come up with a better name. I think we can just keep it this way unless someone has a better suggestion.

kjvdven: I think "Opdrachtpalet" is the most common.

po/nl_NL.UTF-8.po
kjvdven: I agree, it looks like all other translations are "gebiedende wijs" (imperative).

po/nl_NL.UTF-8.po
nwehg: You're right, that's probably best

po/nl_NL.UTF-8.po
mtak: Agreed.

po/nl_NL.UTF-8.po
mtak: That was my first thought as well, and if this was Microsoft Office, I'd agree. However, this is a tool for technical people, which will resonate more with the word "commando". Elsewhere in the translation, the word "commando" is also used in similar contexts, so "opdracht" would also break consistency.

po/nl_NL.UTF-8.po
mtak: Missed that one. My awesome terminal with support for ligatures hid that from me :)

trag1c:
Hi @mtak, do you volunteer to join the `nl_NL` localization team to be pinged about future string changes?

mtak:
> Hi @mtak, do you volunteer to join the `nl_NL` localization team to be pinged about future string changes?

Thank you, and yes, I'd be interested. Besides `CONTRIBUTING.md` and `po/README_*` is there anything else I should know?

trag1c:
I don't think so, the files you mentioned should be enough.

po/nl_NL.UTF-8.po
kjvdven: The Dutch Wikipedia calls it "opdracht", but I also found articles where they mention "commando". So it's a preference in the end. For me it's even as long as we stick to it and don't mix them.

https://nl.wikipedia.org/wiki/Command-line-interface

nwehg: To me, opdrachtpalet sounds more natural, so I would go for that but let me know what you think.

mtak: I say we should keep consistency, which would mean the following would need to be updated as well:

```
#: src/apprt/gtk/ui/1.5/ccw-paste.blp:7 src/apprt/gtk/ui/1.2/ccw-paste.blp:7
msgid ""
"Pasting this text into the terminal may be dangerous as it looks like some "
"commands may be executed."
msgstr ""
"Het plakken van deze tekst in de terminal is mogelijk gevaarlijk, omdat het "
"lijkt op een commando dat uitgevoerd kan worden."
```

```
#: src/apprt/gtk/ui/1.5/command-palette.blp:16
msgid "Execute a command…"
msgstr "Voer een commando uit…"
```

Using "opdracht" in these contexts seems odd to me, it's not how tech people around me talk.

nwehg: You're right. There is a difference between terminal commands and commands in the command palette though. That may be a reason to justify the difference. Let me know what you think. If you don't think that's a good reason, I agree in keeping commandopalet.

mtak: That makes sense yes, and could certainly justify the difference. I've updated the PR.

po/nl_NL.UTF-8.po
```diff
@@ -208,12 +209,12 @@ msgstr "Toestaan"
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-read.blp:81
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-write.blp:77
 msgid "Remember choice for this split"
-msgstr ""
+msgstr "Keuze onthouden voor deze splitsing"
```
nwehg: This needs to be changed to imperative still:
"Onthoud keuze voor deze splitsing"

nwehg (commented):
Only one thing needs to be changed to imperative, rest looks good to me.

po/nl_NL.UTF-8.po
mtak: Done.

nwehg (approved):

nwehg:
I think this can be merged when @kjvdven approves too?

kjvdven (approved):

trag1c (approved):

PR #7891, opened 2025-07-09
Modernize our benchmarks

mitchellh:
This PR modernizes our benchmarks. This PR focuses on the benchmark _framework_ and not the benchmarks themselves. That will come in later PRs.

We now produce two binaries with `-Demit-bench`: `ghostty-bench` and `ghostty-gen`. The former is our benchmark tool. The latter is our synthetic data generation tool. The benchmarking CLI usually takes in data from the synthetic generator but we want to do that offline because synthetic data generation can be slow and CPU intensive and mess up our benchmarks.

Our previous benchmark-specific binaries (like `ghostty-bench-codepoint-width`) are all gone. This is all executed as subcommands in the format similar to Ghostty users: `ghostty-bench +codepoint-width --other --args`.

Previously, synthetic data generation was a mess and all unified with `ghostty-bench-stream` which is just nasty. A dedicated CLI now gets us args like `ghostty-gen +osc --p-valid=0.5`. Neat! 

## Signposts and Xcode/Instruments on macOS

The benchmark framework now automatically emits [signposts](https://developer.apple.com/documentation/os/recording-performance-data) around the code that is under test. This is surfaced in Instruments as a region that you can visualize and zoom in on so you can omit any of the other overhead.

Additionally, I've integrated benchmarks with libghostty and our Xcode project so you can just right click a benchmark to open it in Instruments.

These are macOS-specific niceties but the core benchmarking tool is platform-agnostic. 

## Generalized CLI Actions

The `src/cli/action.zig` file was generalized so that it can be shared amongst our three action-ized binaries. The Ghostty-specific actions are now in `src/cli/ghostty.zig`. As an added bonus, our action parsing is now fully unit tested. 

I don't like mixing refactors in with other tasks in PRs but in this case this one was done to enable not one but two other consumers in the same PR, so I think it fits. 

## TODO

Some things I want to do before merge.

- [ ] Add flags to `ghostty-bench` to configure once mode vs duration mode


mitchellh:
Going to merge this to prevent other merge conflicts, will address my TODO in follow-ups.

PR #7896, opened 2025-07-10
cli/gtk: add +new-window action

jcollie:
This will (on GTK) use a D-Bus method call to tell a running Ghostty instance to open a new window. If D-Bus activation is configured properly, Ghostty does not need to be running first.

This could be extended to other platforms, e.g. AppleScript on macOS.

When Ghostty develops a native API, that could be used instead to create a new window.


mitchellh:
I'm probably going to merge #7891 first because stuff like this conflicts pretty hard with it (since I made the `+action` CLI parser generic and moved the Ghostty CLI to `src/cli/ghostty.zig`). Just a heads up that you'll have to content with some conflicts if that gets merged first. Its not messy, but Git gets confused.

jcollie:
Rebased with the new generic actions.

tristan957 (commented):
I don't do any custom class shenanigans, but overall the code looks good!

src/cli/new_window.zig
```diff
@@ -0,0 +1,270 @@
...
+        return 1;
+    }
+
+    var arena = ArenaAllocator.init(alloc_gpa);
+    defer arena.deinit();
+    const alloc = arena.allocator();
+
+    // Use a D-Bus method call to open a new window on GTK.
+    // See: https://wiki.gnome.org/Projects/GLib/GApplication/DBusAPI
+    if (comptime build_config.app_runtime == .gtk) {
```
tristan957: Maybe this could move to a new file within a gtk subfolder?

`cli/gtk/new_window.zig`

jcollie: Done, except I moved it to `src/cli/new_window/gtk.zig`.

jcollie:
Note that I'm working on an apprt-action style setup for the IPC instead of the `@hasDecl` style.

src/cli/new_window.zig
```diff
@@ -0,0 +1,186 @@
...
+
+pub const Options = struct {
+    /// This is set by the CLI parser for deinit.
+    _arena: ?ArenaAllocator = null,
+
+    /// If `true`, open up a new window in a release instance of Ghostty.
+    release: bool = false,
+
+    /// If `true`, open up a new window in a debug instance of Ghostty.
+    debug: bool = false,
```
mitchellh: I think we should remove `--release` and `--debug` and have `class` default to the class matching the build config. That way a debug-cli will open a new window in a debug build by default, release to release, and if someone is scripting this in some other way they can always specify a custom class.

src/cli/new_window.zig
```diff
@@ -0,0 +1,186 @@
...
+
+    /// If set, open up a new window in a custom instance of Ghostty. Takes
+    /// precedence over `--debug`.
+    class: ?[:0]const u8 = null,
+
+    /// Set to `true` if `-e` was found on the command line.
+    _command: bool = false,
```
mitchellh: Can remove this and make `_arguments` an optional type to detect the same thing.

src/cli/new_window.zig
```diff
@@ -0,0 +1,186 @@
...
+    if (count > 1) {
+        try stderr.print("The --release, --debug, and --class flags are mutually exclusive, only one may be specified at a time.\n", .{});
+        return 1;
+    }
+
+    var arena = ArenaAllocator.init(alloc_gpa);
+    defer arena.deinit();
+    const alloc = arena.allocator();
+
+    if (@hasDecl(apprt.IPC, "openNewWindow")) {
```
mitchellh: This is probably fine for now, but history has shown me that the `hasDecl` approach isn't a great design. It's particularly hard to do for libghostty which tends to implement all decls but needs the consumer to declare support.

I'd prefer to consider a `performAction`-style approach, perhaps we can declare a `apprt/ipc.zig` that has an `apprt.ipc.Action` tagged union, and follow the same `performAction` pattern (maybe its `apprt.performIpcAction`) that returns true/false whether it was handled.

src/apprt/gtk/Surface.zig
```diff
@@ -2358,6 +2358,13 @@ pub fn defaultTermioEnv(self: *Surface) !std.process.EnvMap {
         try window.winproto.addSubprocessEnv(&env);
     }
 
+    class: {
+        const gio_app = self.app.app.as(gio.Application);
+        const class = std.mem.span(gio_app.getApplicationId() orelse break :class);
+
+        try env.put("GHOSTTY_CLASS", class);
```
mitchellh: I'd eliminate this for the design I proposed with the CLI flags.

mitchellh (changes requested):
This is very clean, very well done. I have some design change requests but this is heading in a great direction.

src/cli/new_window.zig
jcollie: Yep, was already working on it, but I had some chores to take care of before I finished that up.

src/cli/new_window.zig
jcollie: Done.

src/cli/new_window.zig
jcollie: Done.

src/cli/new_window.zig
jcollie: Done. There's work to be done on the C API side for macOS though.

src/apprt/gtk/Surface.zig
jcollie: Done.

src/apprt.zig
```diff
@@ -49,6 +51,7 @@ pub const runtime = switch (build_config.artifact) {
 
 pub const App = runtime.App;
 pub const Surface = runtime.Surface;
+pub const IPC = runtime.IPC;
```
mitchellh: I don't see a benefit in making this a separate namespace. This can be a static function on `App` since IPC is a property of applications, I think.

I'm thinking: `fn performIpc(Allocator, Target, Key, Value) bool`

mitchellh (changes requested):
Good progress!

mitchellh (approved):
Awesome. This is really clean. I'd be happy to clean up anything else we find in incremental PRs. Thanks.

ncfavier:
This should be the default action, right? I was watching this PR because of [this discussion](https://github.com/ghostty-org/ghostty/discussions/7708#discussioncomment-13733167), but this does not solve the issue that running `ghostty` by itself from a hotkey daemon does nothing (I am still not sure why that makes a difference).

jcollie:
> This should be the default action, right? I was watching this PR because of [this discussion](https://github.com/ghostty-org/ghostty/discussions/7708#discussioncomment-13733167), but this does not solve the issue that running `ghostty` by itself from a hotkey daemon does nothing (I am still not sure why that makes a difference).

Please do not use PRs as a support forum. Open a new discussion and reference this PR.

PR #7897, opened 2025-07-10
Fix custom shader cursor uniforms not set for non-block cursors (#7893)

### LINKED ISSUE #7893: Custom shader cursor uniforms missing for non-block cursors
Discussion #7865

This is a pretty simple bug; we grab size and position information for the cursor from the `cells` struct, previously there was only one place the cursor could be in it but now there are two since it can either be in the front list or the back list depending on whether it should render in front of or behind text.

The solution should probably be to add a helper function to the cells struct that returns the current cursor if there is one (by seeing if either cursor list has a glyph in it and grabbing it from there), and use that to retrieve the cursor glyph when computing the custom shader uniforms.

ClearAspect:
Fixes #7893

Previously, custom shader cursor uniforms were only updated when the cursor glyph was in the front (block) cursor list. This caused non-block cursors (such as bar, underline, hollow block, and lock) to be missing from custom shader effects.

This commit adds a helper to the cell contents struct to retrieve the current cursor glyph from either the front or back cursor lists, and updates the renderer to use this helper when setting custom shader uniforms. As a result, custom shaders now receive correct cursor information for all supported cursor styles.

### DIFF
--- a/src/renderer/cell.zig
+++ b/src/renderer/cell.zig
@@ -361,14 +361,17 @@ test Contents {
     };
     c.setCursor(cursor_cell, .block);
     try testing.expectEqual(cursor_cell, c.fg_rows.lists[0].items[0]);
+    try testing.expectEqual(cursor_cell, c.getCursorGlyph().?);
 
     // And remove it.
     c.setCursor(null, null);
     try testing.expectEqual(0, c.fg_rows.lists[0].items.len);
+    try testing.expect(c.getCursorGlyph() == null);
 
     // Add a hollow cursor.
     c.setCursor(cursor_cell, .block_hollow);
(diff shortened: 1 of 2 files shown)


qwerasd205 (changes requested):
The fix looks great! But if you could add a check for this helper's behavior to the unit test that'd be awesome.

ClearAspect:
> The fix looks great! But if you could add a check for this helper's behavior to the unit test that'd be awesome.

Shoot, didn't realize. Are you expecting a whole new test or an addon to the existing test for the Contents struct?

qwerasd205:
You can just add it to the existing one

qwerasd205 (approved):
Looks great! Thanks for this ^-^

PR #7904, opened 2025-07-10
ci: add shellcheck linting for shell scripts

liby:
Resolves #7882

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,7 @@ jobs:
       - prettier
       - alejandra
       - typos
+      - shellcheck
       - translations
       - blueprint-compiler
       - test-pkg-linux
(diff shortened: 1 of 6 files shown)


jcollie (approved):
LGTM

.github/workflows/test.yml
```diff
@@ -778,6 +779,40 @@ jobs:
...
+          useDaemon: false # sometimes fails on short jobs
+      - name: shellcheck
+        run: |
+          nix develop -c shellcheck \
+            --check-sourced \
+            --color=always \
+            --severity=warning \
+            --shell=bash \
+            --external-sources \
+            $(find . -name "*.sh" -type f ! -path "./zig-out/*" ! -path "./macos/build/*" ! -path "./.git/*" | sort)
```
jparise: We also want to lint `*.bash` files (e.g. `src/shell-integration/bash/ghostty.bash`).

jcollie: What about the generated files, like the bash completion file?

liby: Updated to include `*.bash` files in shellcheck. The find command now checks both `.sh` and `.bash` files. Generated files in `./zig-out/*` are excluded for now.

jparise (approved):
Looks good to me. Let's start with this and expand it to other files later.

mitchellh (approved):

.github/workflows/test.yml
```diff
@@ -778,6 +779,40 @@ jobs:
...
+          useDaemon: false # sometimes fails on short jobs
+      - name: shellcheck
+        run: |
+          nix develop -c shellcheck \
+            --check-sourced \
+            --color=always \
+            --severity=warning \
+            --shell=bash \
+            --external-sources \
+            $(find . \( -name "*.sh" -o -name "*.bash" \) -type f ! -path "./zig-out/*" ! -path "./macos/build/*" ! -path "./.git/*" | sort)
```
mitchellh, line 814: Given the nastiness of this (and it will only grow), I think I'd feel better if we pulled this out into a shell script But I don't know where, won't block the merge on it.

mitchellh (approved):

PR #7917, opened 2025-07-11
Decouple icon constraint width from `adjust-cell-{height,width}`

danielwe:
This fixes #7905 by calculating the icon width constraint from the bare face metrics before the `adjust-cell-{width,height}` modifiers are applied. The result is unchanged when `adjust-cell-{width,height}` is not set. See also https://github.com/ghostty-org/ghostty/pull/7861#discussion_r2198694769

---

Side note: As part of understanding how the constrain code works, I ended up rewriting it in a way that I think is clearer and easier to reason along with, centering the three relevant bounding boxes (glyph, scale group, constraints) and their relationships. I didn't want to be presumptious and include it in this PR, but you can see the diff here: https://github.com/danielwe/ghostty/commit/78407b212b9347c48b30d9a2ebcedfef45ce839f. As far as I can tell, the results are identical, so it's only a clarity/readability refactor. Let me know if I should PR it.

### DIFF
--- a/src/font/Metrics.zig
+++ b/src/font/Metrics.zig
@@ -38,6 +38,9 @@ cursor_height: u32,
 /// The constraint height for nerd fonts icons.
 icon_height: u32,
 
+/// The constraint width for nerd fonts icons.
+icon_width: u32,
+
 /// Minimum acceptable values for some fields to prevent modifiers
 /// from being able to, for example, cause 0-thickness underlines.
 const Minimums = struct {
@@ -50,6 +53,7 @@ const Minimums = struct {
     const cursor_thickness = 1;
     const cursor_height = 1;
     const icon_height = 1;
+    const icon_width = 1;
 };
 
 /// Metrics extracted from a font face, based on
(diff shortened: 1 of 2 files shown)


mitchellh:
Going to leave this to @qwerasd205

src/font/Metrics.zig
```diff
@@ -188,6 +192,19 @@ pub fn calc(face: FaceMetrics) Metrics {
...
+    // icon_width / (2 * cell_width) == icon_height / cell_height
+    // This only applies when the constraint width is two cells;
+    // icons constrained to a single cell can take up the full
+    // cell width.
+    //
+    // It's important that the icon_width is calculated here using
+    // the cell_width and cell_height derived from the face, and
+    // not in Constraint.constrain, which is called after the
+    // adjust-cell-{width,height} modifiers have been applied.
+    const icon_width = 2 * cell_width * icon_height / cell_height;
```
danielwe: The PR as-is preserves the heuristic of using the same margin-to-content ratio for the width and height. However, I've been playing a lot with this today, and I think setting `icon_width = c * cell_width` for some fixed `c` works even better across different primary fonts with different cell aspect ratios. My suggestion would be `c = 4.0 / 3.0` (slightly larger could also work, up to about 1.5).

This decoupling would also let us increase the height a bit for tall icons. Incidentally, I'd also suggest `icon_height = (4.0 / 3.0) * cap_height`.

LMK if I should push these changes, or if you want to discuss over some screenshots on Discord.

danielwe: NVM, I'm deep in a curiosity-driven hole of studying `font_patcher` and other possible scaling heuristics across narrow and wide fonts. Not sure what will come of it if anything. So just disregard the previous comment and let's keep this PR narrowly focused on the `adjust-cell-*` bugfix.

danielwe:
Superseded by #7953, closing

PR #7924, opened 2025-07-12
nix: fix flake input

ncfavier:
Setting `zig/flake-compat` to follow `""` (the current flake, ghostty) is incorrect and introduces an unnecessary dependency cycle. This causes problems in my NixOS configuration, because I collect flake inputs recursively and add them to my system closure to work around https://github.com/NixOS/nix/issues/3995#issuecomment-1537108310.

The other change to `flake.lock` was done automatically by my version of Nix (Lix 2.91.3).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -116,8 +116,8 @@
       },
       "original": {
         "owner": "jcollie",
-        "ref": "56c159be489cc6c0e73c3930bd908ddc6fe89613",
         "repo": "zon2nix",
+        "rev": "56c159be489cc6c0e73c3930bd908ddc6fe89613",
         "type": "github"
       }
     }
--- a/flake.lock
+++ b/flake.lock
@@ -73,7 +73,9 @@
     },
     "zig": {
       "inputs": {
-        "flake-compat": [],
+        "flake-compat": [
+          "flake-compat"
+        ],
         "flake-utils": [
           "flake-utils"
         ],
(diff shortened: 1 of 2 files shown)


mitchellh:
It appears the flake.lock change is causing some issues since our CI verifies that doesn't change. That's probably not a worthwhile thing to test since running differing Nix versions should be allowed/expected. We'll have to fix that though before we can merge this.

mitchellh:
I think we need to fix the `flake.lock` to match upstream Nix. I don't really have a dog in the fight of Lix vs. Nix, but my Nix version changes `flake.lock` on Nix invocations and it'll be annoying to flip flop between this. We need to make _some decision_ and I think its fair to stick to what we're currently doing (I think every maintainer atm uses upstream Nix).

Two reasons this matters:

1. Avoid dirty `flake.lock` whenever we run a Nix command locally
2. Our CI currently verifies running Nix doesn't change this. 

I'd prefer if it didn't matter. I don't care at all what fork of Nix people use. But I want to avoid that noise.

ncfavier:
Oh that's totally fair, I assumed it was a version thing but it seems like a Lix bug then. Reverted.

ncfavier:
Oh that's totally fair, I assumed it was a version thing but this is actually [a known Lix bug](https://git.lix.systems/lix-project/lix/issues/520). Reverted that change.

ncfavier:
Okay, we can make both Lix and Nix happy by changing `ref` to `rev` in `flake.nix` as well. Hopefully that's more reasonable.

PR #7932, opened 2025-07-14
Set title as argv[0] for commands specified with `-e`

rhodes-b:
If a user passes a command with `-e` we set the title of the window as `argv[0]` after `-e`

for example `ghostty -e nvim ~/directory` becomes `nvim`

Mitchell seemed fine with this approach and it sounded easy enough to do so I went ahead and raised a PR without an issue being made before hand
https://github.com/ghostty-org/ghostty/discussions/7870#discussioncomment-13745947

note this does not stop tuis like `yazi` to set the title with escape sequences it just simply sets the title for those that don't so it doesn't end up as a ghost emoji or nothing (if you don't have a font with that emoji on your system)

~~Note i didn't add a check if `direct[0]` exists as the cli parser should (and does) reject the command if you don't pass anything after `-e`
I can add that check for better defensive practice if you want~~

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -679,14 +679,16 @@ pub fn init(
         switch (initial_command) {
             // If a user specifies a command with `-e` it is appropriate to set the title
             // as argv[0]
-            .direct => {
-                _ = try rt_app.performAction(
-                    .{ .surface = self },
-                    .set_title,
-                    .{ .title = initial_command.direct[0] },
-                );
+            .direct => |cmd| {
+                if (cmd.len != 0) {
+                    _ = try rt_app.performAction(
+                        .{ .surface = self },
+                        .set_title,
+                        .{ .title = cmd[0] },
+                    );
+                }
             },
-            // we won't set the title in the case the shell expands the command
+            // We won't set the title in the case the shell expands the command
             // as that should typically be used to launch a shell which should
             // set its own titles
             .shell => {},
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -675,6 +675,22 @@ pub fn init(
             .set_title,
             .{ .title = title },
         );
+    } else if (config.@"initial-command") |initial_command| {
+        switch (initial_command) {
+            // If a user specifies a command with `-e` it is appropriate to set the title
+            // as argv[0]
+            .direct => {
+                _ = try rt_app.performAction(
+                    .{ .surface = self },
+                    .set_title,
+                    .{ .title = initial_command.direct[0] },
+                );
+            },
+            // we won't set the title in the case the shell expands the command
+            // as that should typically be used to launch a shell which should
+            // set its own titles
+            .shell => {},
+        }
     }
 
     // We are no longer the first surface


src/Surface.zig
```diff
@@ -675,6 +675,22 @@ pub fn init(
...
+    } else if (config.@"initial-command") |initial_command| {
+        switch (initial_command) {
+            // If a user specifies a command with `-e` it is appropriate to set the title
+            // as argv[0]
+            .direct => {
+                _ = try rt_app.performAction(
+                    .{ .surface = self },
+                    .set_title,
+                    .{ .title = initial_command.direct[0] },
+                );
```
jcollie: The value of `direct` should be captured by the switch statement rather than using `initial_command.direct` to access the value.

jcollie (changes requested):
One small change, otherwise LGTM.

jcollie:
> I can add that check for better defensive practice if you want

Yes, best to be defensive.

jcollie (approved):
LGTM

00-kat (approved):

rhodes-b:
don't know why github wont let you upload patch files I don't think it should matter as again command after initial command should be a shell and set its own title but I should also check that app.first is set?
`0001-check-app.first-as-well-to-ensure-title-from-e-would.patch`
```
From de9c1565f8ac490939877d1b0f4c68236b3d1fe7 Mon Sep 17 00:00:00 2001
From: rhodes-b <59537185+rhodes-b@users.noreply.github.com>
Date: Sun, 13 Jul 2025 22:57:01 -0500
Subject: [PATCH] check app.first as well to ensure title from `-e` would only
 be set once

---
 src/Surface.zig | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

index d1ca54a8a..da7bdb670 100644
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -675,7 +675,7 @@ pub fn init(
             .set_title,
             .{ .title = title },
         );
-    } else if (config.@"initial-command") |initial_command| {
+    } else if (config.@"initial-command") |initial_command| if (app.first) {
         switch (initial_command) {
             // If a user specifies a command with `-e` it is appropriate to set the title
             // as argv[0]
@@ -693,7 +693,7 @@ pub fn init(
             // set its own titles
             .shell => {},
         }
-    }
+    };

     // We are no longer the first surface
     app.first = false;
--
2.49.0
```

I can push that up as well (or feel free to do it yourself)

jcollie:
> but I should also check that app.first is set? 

IMHO, no. If we create a surface with a custom command using the API (whenever that happens) we'd want that surface's title to start out as `argv[0]` too.

src/Surface.zig
```diff
@@ -675,6 +675,24 @@ pub fn init(
             .set_title,
             .{ .title = title },
         );
+    } else if (config.@"initial-command") |initial_command| {
```
mitchellh, line 678: This is super safe and makes sense, but my proposal was to use this logic for _any_ `initial-command` _or_ `command` set (not just the first surface).

rhodes-b: To make sure I understand I should just always set the title using
https://github.com/ghostty-org/ghostty/blob/main/src/Surface.zig#L532-L536

if it was supplied by the user? I say user since command if not in the config will be determined from `/passwd` as an example. if we keep only setting it for direct that is easy to do since the default command ghostty comes up with shouldn't be direct. Otherwise there is currently no mechanism to detect how `command` was found / generated
If we do want to do that I have another branch with those changes
https://github.com/rhodes-b/ghostty/commit/6101fbed9ea62a635710b73d0be1e2f26600b73b
its not finished (need to update tests, and maybe clean up some of the copy past in `config/command.zig`)
but serves as a proof of concept if that was more what you were looking for

I think keeping that it only sets the title on `direct` commands make sense because of what I said above, and that non direct commands should really be a shell but I can change it to just apply the title in either scenario

and then just make this an `else` statement since `config.title` should take precedence?

Otherwise I don't think is is limited to the first surface beyond that its only `initial-command` currently
this would be a patch for the easier approach of only doing direct commands
```
From 8ba3c506717257219f3f73e46a86413add180cfe Mon Sep 17 00:00:00 2001
From: rhodes-b <59537185+rhodes-b@users.noreply.github.com>
Date: Mon, 14 Jul 2025 23:29:57 -0500
Subject: [PATCH] also include direct normal  arg

---
 src/Surface.zig | 14 +++++++-------
 1 file changed, 7 insertions(+), 7 deletions(-)

index d1ca54a8a..7060497a3 100644
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -675,16 +675,16 @@ pub fn init(
             .set_title,
             .{ .title = title },
         );
-    } else if (config.@"initial-command") |initial_command| {
-        switch (initial_command) {
-            // If a user specifies a command with `-e` it is appropriate to set the title
-            // as argv[0]
-            .direct => |cmd| {
-                if (cmd.len != 0) {
+    } else if (command) |cmd| {
+        switch (cmd) {
+            // If a user specifies a command it is appropriate to set the title as argv[0]
+            // we know in the case of a direct command it has been supplied by the user
+            .direct => |cmd_str| {
+                if (cmd_str.len != 0) {
                     _ = try rt_app.performAction(
                         .{ .surface = self },
                         .set_title,
-                        .{ .title = cmd[0] },
+                        .{ .title = cmd_str[0] },
                     );
                 }
             },
--
2.49.0
```

rmunn:
> don't know why github wont let you upload patch files ...

According to https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files (scroll to the bottom) this is a known issue affecting Linux. If there's a .patch file that would be really tedious to copy and paste, maybe see if you can upload it via a Windows machine?

ambroisie:
Friendly ping @rhodes-b :-)

mitchellh:
Closing for #9121

PR #7944, opened 2025-07-14
fix: rename tab title popup focus

willswire:
This fixes issue #7940 

Screenshot 2025-07-14 at 17 20 03

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -480,7 +480,10 @@ extension Ghostty {
             // Add buttons
             alert.addButton(withTitle: "OK")
             alert.addButton(withTitle: "Cancel")
-
+            
+            // Make the text field the first responder so it gets focus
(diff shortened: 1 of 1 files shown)


macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -876,7 +879,7 @@ extension Ghostty {
             var x = event.scrollingDeltaX
             var y = event.scrollingDeltaY
             let precision = event.hasPreciseScrollingDeltas
-            
+
```
jcollie: While we appreciate the cleanup of the whitespace, it should at least be confined to a separate commit to make review easier.

willswire: No problem! Just had a little IDE format on save action happen here… I’ll get this cleaned up prior to taking the PR out of draft status

mitchellh (approved):

mitchellh:
Thank you! Just waiting on CI.

PR #7947, opened 2025-07-14
windows: get +ssh-cache building on Windows

jcollie:
There are still problems linking due to `gettext`. No idea if this actually _works_ on Windows. File locking had to be disabled on Windows because of a bug in the Zig std library. Adding all of the explicit error sets happened due to disabling file locking. Fixing permissions had to be disabled on Windows as the Windows file system does not support permissions in the way that POSIX systems like macOS and Linux do.


jcollie:
@jasonrayne

PR #7952, opened 2025-07-15
osc: once again, fix up osd color operations

### LINKED ISSUE #7951: OSC color reporting still not right
The main problem (as far as `msedit` is concerned) is `OSC 4`. Consider a multiple color query like this:
```
printf '\e]4;0;?;1;?\e\\'; read
```
That is expected to return two `OSC` responses like this:
```
^[]4;0;rgb:1d1d/1f1f/2121^[\^[]4;1;rgb:cccc/6666/6666^[\
```
But Ghostty combines them into a single `OSC` response like this:
```
^[]4;0;rgb:1d1d/1f1f/2121;1;rgb:cccc/6666/6666^[\
```

And although this is not an issue for `msedit`, it's worth noting that the `OSC 10/11/12` queries are also not handled correctly. Here's an example that doesn't work in Ghostty:
```
printf '\e]10;?;?\e\\'
```
That is expected to return two `OSC` responses like this:
```
^[]10;rgb:ffff/ffff/ffff^[\^[]11;rgb:2828/2c2c/3434^[\
```
But Ghostty just returns one:
```
^[]10;rgb:ffff/ffff/ffff^[\
```

_Originally posted by @j4james in https://github.com/ghostty-org/ghostty/discussions/7874#discussioncomment-13762215_

jcollie:
Fixes #7951

PR #7429 did a lot to improve handling of various OSC escape codes that get/set/reset colors. However, it didn't get it _quite_ right. This PR fixes OSC color handling to hopefully get things closer to perfect.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -863,7 +863,7 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
             }, .unlocked);
         },
 
-        .color_change => |change| {
+        .color_change => |change| color_change: {
             // Notify our apprt, but don't send a mode 2031 DSR report
             // because VT sequences were used to change the color.
             _ = try self.rt_app.performAction(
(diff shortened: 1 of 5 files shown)


mitchellh:
I've been hesitant to review this PR mostly because of the magnitude of change here and uncertainty in how it impacts programs because I don't regularly see programs with color operations, I think. I think it'd be helpful to have a bash test script that we can run in various terminals to check this. I don't expect you to do that yourself necessarily but I think I need some more confidence here.

jcollie:
This was triggered by Microsoft Edit (which also triggered the previous PR). The linked issue has more detail, but `printf '\e]4;0;?;1;?\e\\'; read` should result in _two_ OSC sequences, as seen in the Kitty and xterm screenshots. Ghostty combines them into one.

Screenshot From 2025-09-05 14-47-11
image
Screenshot From 2025-09-05 14-47-08

mitchellh:
Right I understand the issue its addressing and the nature of the issue. Where I'm lacking confidence is the behavior of old cases since `stream_handler` is not tested in this case.

mitchellh:
Replaced by #8590

PR #7953, opened 2025-07-15
Add per-font size adjustment, don't adjust nf symbol font or emoji font; use non-Mono symbols nerd font

danielwe:
This adds functionality for choosing different normalization metrics for each fallback font. It's not exposed as a config option, but could be in the future, which would probably go a long way towards addressing concerns like #7929.

The currently available reference metrics are, in priority order: `ic_width, ex_height, cap_height, line_height, em_size`. The default value is `ic_width`.

By priority order, I mean that if the chosen metric is not defined in the fallback font, we move to the next metric in the list---we don't normalize by an estimated metric from the fallback font (however, we're happy to use an estimated metric from the primary font, that's how `ic_width` normalization between CJK and Latin fonts work). This extends the pattern that was used between `ic_width` and `ex_height` in the existing hardcoded rule. `line_height` is always defined, so the buck stops there.

What motivated me to implement this was the fact that, with the existing hardcoded rule, the embedded symbols-only Nerd Font was always scaled up by a factor of 1.2, which turned out to be an important reason why it's been difficult to make icon scaling work to everyone's satisfaction. Accordingly, the symbols-only font is the first to take advantage of the new functionality. If this PR is merged, #7917 is no longer needed. (To limit the scope of this PR, it only includes the minimal changes to let icon scaling take advantage of this functionality. I may submit a follow-up PR with some further icon scaling improvement enabled by this.)


danielwe:
While adding tests, I realized that this implementation doesn't play well with things like `Collection.setSize`. Hold on while I take care of that.

EDIT: Converting this to a draft while I sleep on the design (I've got everything working with tests added and passing, it's just a matter of getting invariants and separation of concerns straight). My sincere apologies, I know WIP PRs are not the way things are done here. Was simply too eager to click submit this time.

danielwe:
Ready for review. This turned into a bigger diff than expected because, especially with an eye to a future config option, I think the scale reference should live in a translation layer mapping `Collection` options to `Face` options, such that it can persist while the face itself is swapped out from a `DeferredFace` to a `Face` (there are also other reasons it would be an awkward fit down in the `Face` struct itself). This means it belongs in the face's `Entry`---though not if the entry is an alias.

Hence I decided to split up `Entry`, which used to hold entry aliasing and  *loaded/*deferred face states all in a single union. Now, the owned vs. aliased state and the *loaded vs. *deferred face states are in two separate unions (`EntryOrAlias`, `AnyFace`, names up for discussion of course), and there's a struct in the middle (`Entry`) that holds the face union along with its scale reference (and potentially other "face modifiers" in the future).

Perhaps this sounds like a bit too much wrapping paper, but it actually works out very nicely and adds minimal boilerplate once you get the separation of concerns right. As a bonus, it's now impossible to create nested aliases.

The semantics remain as in the OP.

The main changes are in `Collection.zig`, with some auxiliary changes to `Metrics.zig`. Changes in other files are just syntax updates (`.{ .loaded = myface }` to `.init(.{ .loaded = myface })` for initializing an entry).

Below are screenshots showing the difference this makes for icon scaling. As mentioned in the OP, these are the minimal changes to take advantage of the symbols font no longer being loaded at a scaled-up size. While the most obvious change is that the icons are a bit larger overall, the important difference is that their relative sizes are more harmonious. For example, the nix snowflake got a lot bigger and looks much better next to the toml logo, which only got a little bit bigger. This is because, now that the symbol font is no longer oversized to begin with, there is much less need for scaling icons back down, especially by width, so most can stay at their original design size and maintain their relative proportions.

**Before**

Screenshot 2025-07-17 at 17 31 11

**After**

Screenshot 2025-07-17 at 17 31 14

I have a few more icon scaling/alignment improvements in the pipeline if this gets merged, basically replicating the nerd font patcher exactly if we want that. If we get to that point, I'll open a discussion before submitting a PR.

qwerasd205:
From a quick glance I think I like this change, but I won't have time to give it a proper review for a couple days most likely.

danielwe:
Just discovered from investigating #8054 that some fonts specify huge ascent and descent, adding up to a line height of almost 3 em. The current heuristics (both before and after this PR) result in super tiny glyphs when such a font is loaded as a fallback next to a "normal" font where the line height is in the neighborhood of 1 em, due to the capping of the line height ratio at 1.2. My takeaway from this is that line height is an unreliable metric that should not be used in these calculations at all. We could cap the ratio of point sizes at 1.2 instead. How does that sound?

src/font/SharedGridSet.zig
```diff
@@ -324,7 +327,7 @@ fn collection(
             _ = try c.add(
                 self.alloc,
                 .regular,
-                .{ .fallback_deferred = face },
+                .init(.{ .fallback_deferred = face }),
```
danielwe: We probably want to opt out of scaling for the emoji font too? (Not gonna push the change right now as I know you're in the middle of reviewing.)

```suggestion
                .initWithScaleReference(.{ .fallback_deferred = face }, .em_size),
```

Ditto for the non-Darwin branch below

qwerasd205:
Did some cleanup I wanna add before merging this, but it ended up being more of a change than expected. I'll finish it tomorrow most likely.

qwerasd205:
I did some cleanup on your code (mostly adjusting naming and comments) and then ended up generally cleaning up the interface and code in `Collection` a bit.

I moved the size adjustment logic back outside of `Entry` because there's less abstraction and conceptual overhead just doing it in the collection rather than having to pass a bunch of stuff in to the entry's method.

I did go ahead and remove the line height limit since I couldn't make a convincing case where it caused actual problems in my brief testing. I also removed the size adjustment from the emoji fallback(s) and the specified font(s) for the collection so we don't get any weird size changes going on if the bold/italic face(s) have slightly different metrics to the regular face.

The way I changed the size adjustment logic (including the naming) is setting it up so that more complex adjustments would be easy enough to implement in the future. Rather than directly matching a metric, when I add user config I want users to be able to match metrics pretty flexibly (something like `a * x = b * y` where `a` and `b` are metrics on the primary and specified face and `x` and `y` are scalars).

Included in my changes are some miscellaneous fixes, such as allowing fractional pixel sizes for fonts (we were rounding to the nearest pixel before) which is what caused the large discrepancy in the size adjustment tests-- after fixing that the values are much closer, as they should be.

I still want to clean up a lot of other stuff in `Collection` but this is a state I feel comfortable merging these changes in, I'll make the additional changes in a separate PR.

With this PR, I'm definitely a lot happier with the nerd font sizing (though I might fiddle with it a little more) -- a big aspect of that turns out to be just using the non-Mono symbols font, rather than any of these other changes. (But these other changes *are* still good to have! We need them for future user config options.)

danielwe:
Looks great, thanks for all the improvements! I also had some concerns about consistency across multiple specified fonts and especially synthesized styles, so thanks for straightening out all that.

My only question is whether the line height should be in the fallback path at all? Might be safer to just use no scaling as the catch-all fallback.

> a big aspect of that turns out to be just using the non-Mono symbols font, rather than any of these other changes.

I agree that the non-mono symbols font is much nicer; there are a lot of compromises in relative icon sizes in the mono one. However, an important condition for the non-mono version to work well is that it's loaded at the same point size as the primary. If you load it oversized, most icons have to be scaled down to the constraints, leading to a lot of compromises, while if you load it at the nominal size, most icons pass through the constraints unchanged except wrt. alignment (unless your primary font is very narrow and tall). So I'd say the magic is in the sum of these two changes.

> happier with the nerd font sizing (though I might fiddle with it a little more)

I have a branch with nerd font scaling tweaks that's ready to go once rebased. It implements a few details from font_patcher that are not implemented here, mostly regarding single vs. double-width constraints and group vs. individual alignment. What would you like me to do with this proposal? Should I submit a PR, start a discussion, or just send you a link to the branch to look at? Or should I wait and let you fiddle until you're done, and then see if I still have changes I'd like to propose?

mitchellh (approved):
Excellent work to you both! I scanned through it looking for any obvious confusion and it all looks straightforward to me, though I'm going to trust on the basic logic of size adjustment.

PR #7960, opened 2025-07-16
shell-integration.fish: fix ssh-term integration bug, don't use $ssh_terminfo

HuaDeity:
This PR fixes a bug in the fish shell's `ssh-term` integration script.

The issue occurs because fish shell variables (defined with `set`) cannot properly store the terminfo output from `infocmp -0 -x xterm-ghostty 2>/dev/null`. To resolve this, we now pipe the output directly to the SSH remote server.

I'm uncertain whether we should maintain the `$ssh_terminfo` variable or remove it entirely.

### DIFF
--- a/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
+++ b/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
@@ -131,7 +131,7 @@ function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
                         set -l ssh_cpath_dir
                         set -l ssh_cpath
 
-                        set ssh_terminfo (infocmp -0 -x xterm-ghostty 2>/dev/null)
+                        set ssh_terminfo "$(infocmp -0 -x xterm-ghostty 2>/dev/null)"
 
                         if test -n "$ssh_terminfo"
                             echo "Setting up xterm-ghostty terminfo on $ssh_hostname..." >&2
@@ -139,7 +139,7 @@ function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
                             set ssh_cpath_dir (mktemp -d "/tmp/ghostty-ssh-$ssh_user.XXXXXX" 2>/dev/null; or echo "/tmp/ghostty-ssh-$ssh_user."(random))
                             set ssh_cpath "$ssh_cpath_dir/socket"
 
-                            if infocmp -0 -x xterm-ghostty 2>/dev/null | command ssh $ssh_opts -o ControlMaster=yes -o ControlPath="$ssh_cpath" -o ControlPersist=60s $argv '
+                            if echo "$ssh_terminfo" 2>/dev/null | command ssh $ssh_opts -o ControlMaster=yes -o ControlPath="$ssh_cpath" -o ControlPersist=60s $argv '
                                 infocmp xterm-ghostty >/dev/null 2>&1 && exit 0
                                 command -v tic >/dev/null 2>&1 || exit 1
                                 mkdir -p ~/.terminfo 2>/dev/null && tic -x - 2>/dev/null && exit 0
--- a/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
+++ b/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
@@ -139,7 +139,7 @@ function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
                             set ssh_cpath_dir (mktemp -d "/tmp/ghostty-ssh-$ssh_user.XXXXXX" 2>/dev/null; or echo "/tmp/ghostty-ssh-$ssh_user."(random))
                             set ssh_cpath "$ssh_cpath_dir/socket"
 
-                            if echo "$ssh_terminfo" | command ssh $ssh_opts -o ControlMaster=yes -o ControlPath="$ssh_cpath" -o ControlPersist=60s $argv '
+                            if infocmp -0 -x xterm-ghostty 2>/dev/null | command ssh $ssh_opts -o ControlMaster=yes -o ControlPath="$ssh_cpath" -o ControlPersist=60s $argv '
                                 infocmp xterm-ghostty >/dev/null 2>&1 && exit 0
                                 command -v tic >/dev/null 2>&1 || exit 1
                                 mkdir -p ~/.terminfo 2>/dev/null && tic -x - 2>/dev/null && exit 0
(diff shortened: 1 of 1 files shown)


jparise:
> The issue occurs because fish shell variables (defined with `set`) cannot properly store the terminfo output from `infocmp -0 -x xterm-ghostty 2>/dev/null`.

Could you help me better understand this? This works fine for me on macOS with fish 4.0.2:

```fish
set ssh_terminfo (infocmp -0 -x xterm-ghostty 2>/dev/null)
echo $ssh_terminfo
```

... so maybe there's something system-specific about `infocmp`'s (or `fish`'s) behavior?

HuaDeity:
Thanks for taking a look! You're right, the issue is subtle and was tricky to debug.

The problem is that fish shell variables can't store null bytes (`\0`), which are part of the `infocmp -0` output. This is misleading because using `echo` on the variable appears to work, but the null byte is silently dropped. The zsh script works because zsh variables can store them correctly.

When the fish script tried to save the `infocmp` output to a variable, the data was silently corrupted. Piping the command's output directly to `ssh` (as this PR does) bypasses the broken variable assignment and sends the correct data, fixing the issue.

#### Proof of the Limitation
This limitation is rooted in how fish handles C-style null-terminated strings and is acknowledged in the [official documentation](https://fishshell.com/docs/current/cmds/string.html#cmd-string-join0). 

> Because Unix uses NUL as the string terminator, passing the output of string join0 as an argument to a command (via a [command substitution](https://fishshell.com/docs/current/language.html#expand-command-substitution)) won’t actually work. Fish will pass the correct bytes along, but the command won’t be able to tell where the argument ends. This is a limitation of Unix’ argument passing.

The most definitive proof is this simple comparative test:
##### Fish Shell Test (Fails to store null byte):
```fish
# Try to store a string containing a null byte
> set my_var (printf 'A\0B')

# The variable is truncated at the null byte
> echo $my_var
A

# The length is incorrect
> string length $my_var
1
```

##### Zsh Shell Test (Correctly stores null byte):
```zsh
# Store the same string
% my_var=$(printf 'A\0B')

# The length is correct, proving the null byte is stored
% print "Length: ${#my_var}"
Length: 3

# Hexdump confirms the null byte (00) is present
% print -rn -- "$my_var" | hexdump -C
00000000  41 00 42                                          |A.B|
00000003
```

This test clearly shows why the data must be piped directly in fish.

Hope this detailed explanation clarifies the reasoning behind the change!

HuaDeity:
**The bug causes the following issue:**

When using the fish script, the remote `~/.terminfo` directory is created successfully but remains empty. This process works correctly in zsh.

The remote `tic` command failed to compile the terminfo file because it received corrupted data.

HuaDeity:
Besides the handling of the null character (`\0`), there are other important differences in how escape sequences are processed by the `echo` command in **fish** and the `print` command in **zsh**.

A significant example is the `\E` escape sequence. According to its documentation, **zsh**'s `print` command correctly interprets `\E` as a synonym for `\e` (the standard escape character).

> print [ -abcDilmnNoOpPrsSz ] [ -u n ] [ -f format ] [ -C cols ]
      [ -v name ] [ -xX tabstop ] [ -R [ -en ]] [ arg ... ]
With the ‘-f’ option the arguments are printed as described by printf. With no flags or with the flag ‘-’, the arguments are printed on the standard output as described by echo, with the following differences: the escape sequence ‘\M-x’ (or ‘\Mx’) metafies the character x (sets the highest bit), ‘\C-x’ (or ‘\Cx’) produces a control character (‘\C-@’ and ‘\C-?’ give the characters NULL and delete), a character code in octal is represented by ‘\NNN’ (instead of ‘\0NNN’), and ‘\E’ is a synonym for ‘\e’. Finally, if not in an escape sequence, ‘\’ escapes the following character and is not printed.

However, **fish shell does not handle `\E` in the same way**. Attempts to process this sequence using `echo -e` or `printf '%b'` will fail, as these commands do not recognize `\E` as a valid escape sequence in `fish`.

Given the many differences and edge cases between built-in commands in different shells, **it is advisable to avoid storing raw strings with shell-specific escape sequences in variables**. Relying on `echo` or `print` to interpret such variables often leads to non-portable scripts and can require significant time to debug compatibility issues.

CleanShot 2025-07-17 at 11 10 10@2x

00-kat:
I'm not sure if that would actually be problematic since `tic` interprets the `\E` just fine and I think passing a literal escape character to `tic` works fine too.

HuaDeity:
Although I'm not sure of the exact cause, the facts are clear: the script to install `terminfo` fails in **fish shell** but succeeds in **zsh**.

The problem is fixed by piping the output from `infocmp` directly over SSH.

jparise:
> The problem is that fish shell variables can't store null bytes (`\0`), which are part of the `infocmp -0` output.

I'm reading through the other details, but I wanted to note that the `-0` option doesn't enable NUL bytes in the output (unlike `find -0`, for example). From `infocmp(1)`:

```
 -0   causes the fields to be printed on one line, without wrapping.
```

jparise:
> I'm uncertain whether we should maintain the `$ssh_terminfo` variable or remove it entirely.

I think we should drop it and change the condition to something like:

```fish
if infocmp xterm-ghostty >/dev/null
```

jparise:
@HuaDeity I haven't been able to reproduce the problem you're seeing. Your proposed change should be fine (although it adds an extra call to `infocmp`), but I think it's important to understood the root cause of what you're seeing so we can better maintain this going forward.

Can you share some more details about your environment (OS, fish, and ncurses version)?

HuaDeity:
Hi, @jparise, below is my environment:

**Environment Details:**

I'm experiencing the issue that persists across multiple remote servers, despite their varying environments.

**Local Environment:**
- OS: macOS 26 Developer Beta 3
- Fish: 4.0.2 (installed via Homebrew)
- Ncurses: 6.0.20150808 (system version)

**Remote Environment (example):**
- OS: Ubuntu 20.04.6 LTS
- Fish: 4.0.2 (installed via nixpkgs)
- Ncurses: 6.2.20200212 (system version)

HuaDeity:
Ok, I found the real issue: it’s caused by Fish shell’s [command substitution](https://fishshell.com/docs/current/language.html#command-substitution).

Without double quotes, Fish treats each line of the command output as a separate argument to the outer command. The infocmp output has two lines—the first is a comment:
``` bash
#	Reconstructed via infocmp from file: /Applications/Ghostty.app/Contents/Resources/terminfo/78/xterm-ghostty
```
The second line is the one we actually want. Without quotes, both lines are split by spaces and merged into a single line, which results a comment only.

jparise (approved):

jparise:
> Ok, I found the real issue: it’s caused by Fish shell’s [command substitution](https://fishshell.com/docs/current/language.html#command-substitution).
> 
> Without double quotes, Fish treats each line of the command output as a separate argument to the outer command. The infocmp output has two lines—the first is a comment:
> 
> ```shell
> #	Reconstructed via infocmp from file: /Applications/Ghostty.app/Contents/Resources/terminfo/78/xterm-ghostty
> ```
> 
> The second line is the one we actually want. Without quotes, both lines are split by spaces and merged into a single line, which results a comment only.

Ah, this makes sense. Thanks for tracking this down!

PR #7968, opened 2025-07-17
gtk-ng: port ConfigErrorsDialog

mitchellh:
This ports the config errors dialog from `apprt/gtk` to `gtk-ng`. 

The major change here is that we now use proper template bindings for the content. To do this, a `ghostty.Config` is now wrapped in a GObject `GhosttyConfig` to make it safe to pass around (ref count) and to provide helpful properties like the diagnostics buffer we bind to.

As a minor change, I stripped the `Ghostty` prefix from our GObject classes in Zig code. For templates its all still there as is the norm.

This retains the exact same version requirements and layout as the existing one.

### DIFF
--- a/src/apprt/gtk-ng.zig
+++ b/src/apprt/gtk-ng.zig
@@ -6,9 +6,9 @@ pub const Surface = @import("gtk-ng/Surface.zig");
 pub const resourcesDir = internal_os.resourcesDir;
 
 // The exported API, custom for the apprt.
-pub const GhosttyApplication = @import("gtk-ng/class/application.zig").GhosttyApplication;
-pub const GhosttyWindow = @import("gtk-ng/class/window.zig").GhosttyWindow;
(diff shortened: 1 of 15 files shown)


mitchellh:
The snap errors are because my branch name is too long: 

```
Cannot pack snap: error: cannot validate snap "ghostty": invalid snap version "1.1.4-gtk-ng-config-error+b2fe9fd": cannot be longer than 32 characters (got: 33)
```

lol. The snap builds the `gtk` apprt so this should be fine.

mitchellh:
It's some major scope creep but I added a `valgrind.supp` in this PR and verified this PR passes Valgrind as noted in that file. I think that's a big step to making sure as we iterate we keep Valgrind passing.

mitchellh:
Going to send this in the name of forward progress on the refactor. I'll react to any reviews still and can follow up with PRs. I'm still interested in making sure things make sense _structurally_ so I'll pay attention!

PR #7972, opened 2025-07-17
Surface: use rect selection state when setting selection on release

vancluever:
Looks like 52354b8 missed noting the outgoing screen selection state's rectangle flag when setting the selection on mouse release, this was causing the selection that was actually set to be
standard/wrap-selected. This corrects that by just shipping said flag when calling `setSelection`.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3341,7 +3341,7 @@ pub fn mouseButtonCallback(
                 try self.setSelection(terminal.Selection.init(
                     prev.start(),
                     prev.end(),
-                    false,
+                    prev.rectangle,
                 ));
             }
         }


mitchellh (approved):

vancluever:
Thanks @mitchellh! Happy to throw in a test too if we have an easy way to hit that code.

edrozenberg:
Great fix! Appears to resolve the issues w/ copy paste of terminal content into for ex. a text editor. Pasted content would often be wrapped badly in an unusable way due to trailing spaces that were included in the copied content, requiring lots of manual cleanup.

I'd seen the issue weeks ago but I've been slower to report these "small issues" since they understandably don't get looked at while more important big items and refactors get worked on. I'll need to get to the point where I can submit PR's myself, not there atm.

vancluever:
@edrozenberg  glad it fixed things for you!

PR #7975, opened 2025-07-17
core/gtk: add support for displaying a progress bar with OSC 9;4

jcollie:
Ghostty has had support for a while (since PR #3124) for parsing progress reports but never did anything with them. This PR adds the core infrastructure and an implementation for GTK.

On GTK, the progress bar will show up as a thin bar along the top of the terminal. Under normal circumstances it will use whatever you have set as your accent color. If the progam sending the progress report indicates an error, it will change to a reddish color.

### DIFF
--- a/src/apprt/gtk/ProgressBar.zig
+++ b/src/apprt/gtk/ProgressBar.zig
@@ -50,7 +50,6 @@ pub fn handleProgressReport(self: *ProgressBar, value: terminal.osc.Command.Prog
 
         // Set the progress bar to a fixed value if one was provided, otherwise pulse.
         // Remove the `error` CSS class so that the progress bar shows as normal.
-        // TODO: how to deal with accessibility?
         .set => {
             progress_bar.as(gtk.Widget).removeCssClass("error");
             if (value.progress) |progress| {
@@ -62,7 +61,6 @@ pub fn handleProgressReport(self: *ProgressBar, value: terminal.osc.Command.Prog
 
         // Set the progress bar to a fixed value if one was provided, otherwise pulse.
         // Set the `error` CSS class so that the progress bar shows as an error color.
-        // TODO: how to deal with accessibility?
         .@"error" => {
             progress_bar.as(gtk.Widget).addCssClass("error");
             if (value.progress) |progress| {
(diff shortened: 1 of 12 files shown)


jcollie:
Here's a quick bash one-liner to test the progress bar.

```
counter=1; while [ $counter -le 25 ]; do printf "\x1b]9;4;2;${counter}\x07"; ((counter++)); sleep 0.1; done; while [ $counter -le 50 ]; do printf "\x1b]9;4;1;${counter}\x07"; ((counter++)); sleep 0.1; done; while [ $counter -le 75 ]; do printf "\x1b]9;4;3\x07"; ((counter++)); sleep 0.1; done; while [ $counter -le 100 ]; do printf "\x1b]9;4;1;${counter}\x07"; ((counter++)); sleep 0.1; done; printf "\x1b]9;4;0\x07"
```

src/terminal/osc.zig
```diff
@@ -206,12 +203,29 @@ pub const Command = union(enum) {
...
+            set,
+            @"error",
+            indeterminate,
+            pause,
+        };
+
+        state: State,
+        progress: ?u8 = null,
+
+        pub const C = extern struct {
```
mitchellh, line 220: Try to comment these of what they're synced with so its easier to find the mappings.

src/apprt/gtk/ProgressBar.zig
```diff
@@ -0,0 +1,169 @@
...
+    const progress_bar = self.addProgressBar();
+    self.startProgressBarTimer();
+
+    switch (value.state) {
+        // already handled above
+        .remove => unreachable,
+
+        // Set the progress bar to a fixed value if one was provided, otherwise pulse.
+        // Remove the `error` CSS class so that the progress bar shows as normal.
+        // TODO: how to deal with accessibility?
```
mitchellh: I'd just remove these. a11y isn't really a concern on the GTK side at this time (our answer is "use macOS"), and a11y is so bad in general in GTK w/ Ghostty that we have much bigger problems when we get there.

mitchellh (changes requested):
A couple small requests. Overall this looks great though.

src/terminal/osc.zig
jcollie: Done.

src/apprt/gtk/ProgressBar.zig
jcollie: Done.

mitchellh (approved):

PR #7981, opened 2025-07-18
apprt/gtk-ng: extract common methods into a mixin

mitchellh:
Every GObject class we're ever going to make has the same handful of methods. This adds *just enough* noise to be annoying. This commit extracts the common ones so far into a central mixin. Since Zig is removing `usingnamespace` and has no other mixin mechanism, we must forward the decls, but this is still cleaner imo than what we did before.

I suspect longer term we can probably abstract more of the `zig-gobject` boilerplate into a higher level abstraction but I'm not confident doing that yet across the 4 classes we have so far.

Credit to @pluiedev for pointing this out.

### DIFF
--- a/src/apprt/gtk-ng/class.zig
+++ b/src/apprt/gtk-ng/class.zig
@@ -24,3 +24,45 @@ pub fn unrefLater(obj: anytype) void {
         }
     }).callback, obj.as(gobject.Object));
 }
+
+/// Common methods for all GObject classes we create.
+pub fn Common(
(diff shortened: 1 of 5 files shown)


tristan957 (approved):

pluiedev (approved):

jcollie (approved):

mitchellh:
It passed the only tests that run gtk-ng tests, so forcing the merge. Thanks y'all.

PR #7992, opened 2025-07-19
gtk: update zig-gobject to get improved gobject accessors

jcollie:


mitchellh:
Beautiful. I added a commit that cleaned up `gtk-ng` with these changes. I made it a separate commit so you can take a look if it interests you.

PR #7994, opened 2025-07-19
apprt/gtk-ng: implement `Surface.close`

mitchellh:
A small, simple change. This implements the `Surface.close` apprt required function. After this PR, a process exiting within the terminal will close the window properly (unless `wait-after-command` is set of course!).

This doesn't yet show close confirmation. I'm working on some dialog refactors on the side to see if we can simplify our Adw 1.2 vs. 1.5 dialogs and didn't want to include it here.

Close now works by the `GhosttySurface` class emitting the `close-request` signal (similar to a `gtk.Window`) and the parent container is responsible for closing it. This will let us reuse the surface within different contexts: tabs, splits, etc.

This also remove the unused `shouldClose`, `setShouldClose` apprt APIs which are a holdover from the glfw days!

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -132,18 +132,6 @@ pub fn destroy(self: *App) void {
 /// events. This should be called by the application runtime on every loop
 /// tick.
 pub fn tick(self: *App, rt_app: *apprt.App) !void {
-    // If any surfaces are closing, destroy them
-    var i: usize = 0;
-    while (i < self.surfaces.items.len) {
-        const surface = self.surfaces.items[i];
-        if (surface.shouldClose()) {
-            surface.close(false);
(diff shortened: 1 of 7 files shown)


pluiedev (approved):
Oh that's really smart. I like that. I assume then once we have tabs and split trees we would propagate the close signal upwards once all the children inside the container are gone? So for instance when the sole surface inside a tab signals to the tab to be closed, the tab also signals to the window to be closed, etc

mitchellh:
Exactly, yeah. That's how the macOS app was refactored to work (recently) and it works really well. It also isn't unheard of for GTK since `gtk.Window` also does a `close-request` based flow.

PR #7995, opened 2025-07-19
apprt/gtk-ng: abstract our alert vs msg dialog into a superclass

mitchellh:
This introduces a new `GhosttyDialog` class that either inherits from  `adw.MessageDialog` or `adw.AlertDialog`, depending on the version of libadwaita we compile against. This is the same logic we used previously.

This lets us have a single libadw 1.2 blueprint file for all dialogs and  we just do the right thing at compile time!


src/apprt/gtk-ng/class/dialog.zig
```diff
@@ -0,0 +1,69 @@
...
+    pub const getGObjectType = gobject.ext.defineClass(Self, .{
+        .name = "GhosttyDialog",
+        .classInit = &Class.init,
+        .parent_class = &Class.parent,
+    });
+
+    pub fn present(self: *Self, parent: ?*gtk.Widget) void {
+        switch (Parent) {
+            adw.AlertDialog => self.as(adw.Dialog).present(parent),
+            adw.MessageDialog => self.as(gtk.Window).present(),
```
pluiedev: Note that this currently doesn't that as expected on Adw 1.2. How parenting works on 1.2 is that since all MessageDialogs are also windows, their parents must also be windows and be set via `gtk.Window.setTransientFor` (normally `adw.MessageDialog.new` already does this for you).

I think the easiest thing to do here is to change the parent type to `?*gtk.Window` and then setting the transient parent window immediately before presenting. It's a bit ugly but it should work

pluiedev (changes requested):

src/apprt/gtk-ng/class/dialog.zig
mitchellh: This code is copy and pasted from the old `gtk` one, and I just gripped the codebase and I don't see any `setTransientFor` calls anywhere either. Is it because we're a superclass are you saying that `new` won't be called? Because otherwise the blueprint files are the same too as before (we inherit from this class). Or is the existing `gtk` apprt just broken also?

I tested this code with forcing `MessageDialog` on 1.5 and it works just fine. Or are you saying specifically on 1.2 this happens but MessageDialog later fixed it? (Note: explicitly using `MessageDialog`, not Alert).

pluiedev: In the MessageDialog's constructor the `transient-for` property is explicitly set (in Zig code the equivalent would be `gobject.ext.newInstance(MessageDialog, .{ .@"transient-for" = parent })`), so when you subclass it that property won't be set. GTK is one of those strange OOP systems where the `new` method and the `init` method aren't the same thing, so subclasses don't recursively call the constructors of parent classes when properties are involved like that.

To be more exact, the `new` method should always call the `newInstance` helper with a list of properties to set, which allocates the right amount of memory, calls the `init` methods recursively, then sets the properties. The `new` method won't ever be called by the subclasses (making it a *final* method funnily enough), only the `init` method would.

Not setting `transient-for` will technically *work* but on e.g. Wayland the dialog window would be its own toplevel window instead of being a dialog window as it should, which could greatly confuse the compositor (e.g. not presenting the dialog along with the main window, not centering the dialog properly, etc.) We should be setting the parent explicitly whenever we can.

mitchellh: I understand that, but is this also broken on the previous apprt, cause it doesn't do that? I copy and pasted it.

Also, I assume you don't need to set this if there is no parent (e.g. the config errors dialog) because there might not even be an ApplicationWindow created. Reading the docs and the GTK source code this seems to be totally safe and fine.

pluiedev: https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/CloseDialog.zig#L38

The old apprt has an extra parameter passed to `adw.MessageDialog.new`

Also yeah it's totally okay to not set it or set it to null, I'm just saying that in case it does makes sense to set a parent (e.g. command palette) then we do need to specify it ourselves

mitchellh: > https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/CloseDialog.zig#L38
> The old apprt has an extra parameter passed to adw.MessageDialog.new

Got it. Yeah this is the close dialog. I was copying from the config errors dialog because that's all we have at the moment. I'll be sure to port this over when I do the close one. 😄

mitchellh: I decided to handle this in this PR. Pushed it!

PR #8008, opened 2025-07-21
core/vt: initialize OSC command to a known value

jcollie:
Fixes: #8007

If we don't initialize the command value when we create an OSC parser, there may be situations where the parser gets reset and it appears that the command contains a `kitty_color_protocol` value (due to the value Zig uses to initialize `undefined` values with). The program will then crash when the ArrayList inside the `kitty_color_protocol` is attempted to be deinitialized since the ArrayList was never properly initialized.

By always initializing the command to a known value, we avoid this situation.

### DIFF
--- a/src/terminal/osc.zig
+++ b/src/terminal/osc.zig
@@ -16,6 +16,11 @@ const kitty = @import("kitty.zig");
 const log = std.log.scoped(.osc);
 
 pub const Command = union(enum) {
+    /// The default value for the Command. We can't leave it as uninitialized
+    /// or we risk crashes due to calling deinit on allocations that were never
+    /// actually allocated.
+    invalid,
+
     /// Set the window title of the terminal
     ///
     /// If title mode 0  is set text is expect to be hex encoded (i.e. utf-8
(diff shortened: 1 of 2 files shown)


src/terminal/osc.zig
```diff
@@ -3365,3 +3369,19 @@ test "OSC: kitty color protocol no key" {
     try testing.expect(cmd == .kitty_color_protocol);
     try testing.expectEqual(0, cmd.kitty_color_protocol.list.items.len);
 }
+
+// Note that this test doesn't really do much, as the parser seems to
+// be initialized differently when being run with `zig build test` vs a
+// normal build. In a normal build `p.command` appears as if it contained a
```
gpanders, line 3375: Hm this is surprising and a bit unfortunate. Does this mean it's not possible to write unit tests that cover use of undefined values?

jcollie: Not that I can tell, at least not when it depends specifically on the undefined value. Which I guess makes sense because in a ReleaseFast version you'll just get whatever randomly happened to be in memory at that location.

mitchellh: We should be able to eventually run our test exe under Valgrind and it'll catch undefined access.

mitchellh: I've confirmed that Valgrind can in fact find all of this: #8304

mitchellh:
Replaced by #8307

PR #8010, opened 2025-07-21
apprt/gtk-ng: hovered url overlay

mitchellh:
I was going to do more overlays but this introduced enough new complexity that I want to PR this on its own.

Valgrind clean!

### DIFF
--- a/src/apprt/gtk-ng/class/surface.zig
+++ b/src/apprt/gtk-ng/class/surface.zig
@@ -94,6 +94,23 @@ pub const Surface = extern struct {
             );
         };
 
+        pub const @"mouse-hover-url" = struct {
+            pub const name = "mouse-hover-url";
+            pub const get = impl.get;
+            pub const set = impl.set;
+            const impl = gobject.ext.defineProperty(
+                name,
+                Self,
+                ?[:0]const u8,
+                .{
+                    .nick = "Mouse Hover URL",
+                    .blurb = "The URL the mouse is currently hovering over (if any).",
+                    .default = null,
+                    .accessor = C.privateStringFieldAccessor("mouse_hover_url"),
+                },
+            );
+        };
+
         pub const pwd = struct {
             pub const name = "pwd";
             pub const get = impl.get;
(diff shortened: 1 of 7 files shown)


jcollie (approved):
LGTM

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -823,6 +858,41 @@ pub const Surface = extern struct {
             null,
             .{ .detail = "mouse-shape" },
         );
+
+        // Some other initialization steps
+        self.initUrlOverlay();
+    }
+
+    fn initUrlOverlay(self: *Self) void {
```
tristan957, line 866: Is it possible to add this to the Blueprint file or is that something that would come in a PR down the line?

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -863,6 +933,10 @@ pub const Surface = extern struct {
             priv.core_surface = null;
         }
 
+        if (priv.mouse_hover_url) |v| {
+            glib.free(@constCast(@ptrCast(v)));
+            priv.mouse_hover_url = null;
+        }
```
tristan957, line 939: Is there no zig equivalent to `g_clear_pointer()`?

tristan957 (approved):

src/apprt/gtk-ng/class/surface.zig
jcollie: `glib.clearPointer` is the equivalent.

mitchellh: No, I plan on writing a helper eventually.

src/apprt/gtk-ng/class/surface.zig
mitchellh: It doesn't seem possible to initialize overlays in a blueprint file. If we can, I'll do it!

mitchellh: Oh also, its not possible to initialize event controllers in a blueprint/builder file. So, regardless of overlays, thats an issue. And again, if I'm wrong, I'd love to fix it. I couldn't find any docs on it, I read the source, and I just tried various thing, but none worked.

src/apprt/gtk-ng/class/surface.zig
mitchellh: > `glib.clearPointer` is the equivalent.

I don't think this will work Zig slices.

mitchellh:
Btw, I think these overlays can be their own dedicated gobjects. I'm looking into that.

PR #8022, opened 2025-07-22
Fix off-by-one error & adjust overline pos in cell height mod

danielwe:
I noticed that there was an off-by-one error in cell height adjustment when the number of pixels to add/subtract is odd. The metrics measured from the top would be shifted by one less than they should, so, for example, the underline position would move one pixel closer to the baseline than it had been (or one pixel further away if subtracting).

Also noticed that the overline position was missing here, so added that.


mitchellh:
Looks good to me but will wait for @qwerasd205 to take a look.

mitchellh (approved):
Confirmed with Qwer in discord.

PR #8023, opened 2025-07-22
terminal/osc: allow empty payloads

pluiedev:
Fixes #8018

### DIFF
--- a/src/apprt/gtk/Surface.zig
+++ b/src/apprt/gtk/Surface.zig
@@ -1259,9 +1259,13 @@ pub fn setClipboardString(
         // We only toast if we are copying to the standard clipboard.
         if (clipboard_type == .standard and
             self.app.config.@"app-notifications".@"clipboard-copy")
-        {
-            if (self.container.window()) |window|
-                window.sendToast(i18n._("Copied to clipboard"));
+        toast: {
+            const window = self.container.window() orelse break :toast;
+
+            if (val.len > 0)
+                window.sendToast(i18n._("Copied to clipboard"))
+            else
+                window.sendToast(i18n._("Cleared clipboard"));
         }
         return;
     }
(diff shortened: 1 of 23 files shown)


src/terminal/osc.zig
```diff
@@ -1928,6 +1936,21 @@ test "OSC: get/set clipboard with allocator" {
...
+    var p: Parser = .{ .alloc = testing.allocator };
+    defer p.deinit();
+
+    const input = "52;;";
+    for (input) |ch| p.next(ch);
+
+    const cmd = p.end(null).?;
+    try testing.expect(cmd == .clipboard_contents);
+    try testing.expect(cmd.clipboard_contents.kind == 'c');
+    try testing.expect(std.mem.eql(u8, "", cmd.clipboard_contents.data));
```
jcollie, line 1951: `expectEqualStrings` would be better here IMHO.

mitchellh (approved):
Approve pending CI.

src/apprt/gtk/Surface.zig
```diff
@@ -1256,8 +1256,10 @@ pub fn setClipboardString(
         const clipboard = getClipboard(self.gl_area.as(gtk.Widget), clipboard_type) orelse return;
         clipboard.setText(val);
 
-        // We only toast if we are copying to the standard clipboard.
+        // We only toast if we are copying to the standard clipboard
+        // and that the new content is non-empty
```
jcollie: I think that it would be good to have "Clipboard Cleared" toast as a possibility, although I can see some people wanting to turn that off independent of the "Clipboard Copied" toast.

jcollie (approved):
Couple of nits, but won't block.

src/terminal/osc.zig
pluiedev: I literally just copied this from the previous one :bobr:

src/apprt/gtk/Surface.zig
mitchellh: Yeah, I can also see the argument that we should toast when its empty too just with "clipboard set" because it was set.... to empty.

pluiedev: Hmmmm... fair enough

jcollie (approved):
LGTM

PR #8026, opened 2025-07-22
Work around strange SwiftUI behavior in "older" macOSen. (might fix #7690)

krono:
The Quick Terminal would not appear anymore, as somewhere in the framework the Quick Terminal Window's geometry gets corrupted when the window is added to the UI.

This works around by caching the windows geometry and reusing it afterwards

This might fix #7690

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -111,12 +111,24 @@ class QuickTerminalController: BaseTerminalController {
         // Setup our initial size based on our configured position
         position.setLoaded(window)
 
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Thanks, this looks good. I'm going to pull it down and do some testing so I'm going to slap this with a "request changes" to block merge. But it looks good.

krono:
Do you need any more input here?

mitchellh (approved):
Didn't test on Ventura, but verified this doesn't do anything weird on Sequoia and Tahoe.

krono:
🙏

rmacklin:
Verified the quick terminal is now fixed on macOS Sonoma! Thanks!

PR #8029, opened 2025-07-22
build-system: Replace deprecated usages of addStaticLibrary with addLibrary

jaysonsantos:
Hi there, this is just a low-hanging fruit and it also prepares the way for the future 0.15, which removes addStaticLibrary.
Please, let me know what to do on the `// TODO` comments.

### DIFF
--- a/pkg/breakpad/build.zig
+++ b/pkg/breakpad/build.zig
@@ -4,10 +4,13 @@ pub fn build(b: *std.Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
-    const lib = b.addStaticLibrary(.{
+    const lib = b.addLibrary(.{
         .name = "breakpad",
-        .target = target,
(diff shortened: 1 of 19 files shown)


mitchellh:
Thanks! As it stands, I think we can simply remove the `TODOs` you have and keep them separate modules. I think that makes it the same as the previous behavior. I simply don't know enough about the build system changes to know if we should start sharing that.

jaysonsantos:
Thanks for the answer! I sent a new amended commit removing the TODOs

jcollie (approved):
LGTM

jaysonsantos:
The pipeline had the following error:
```
Starting instance
Error: Instance is not running
Incompatible base detected: instance config is empty.
Recommended resolution: Clean incompatible instance and retry the requested operation.
```
is that base that it is talking about, ghostty's base branch or something with snap?

mitchellh:
The snap failure is known flappy. Thanks!

PR #8030, opened 2025-07-22
apprt/gtk-ng: clipboard support

mitchellh:
This ports over read/write clipboard to gtk-ng.

This was a surprisingly massive amount of work! The clipboard confirmation dialog is non-trivial: it supports multiple read/write types, blurring, remember choice, and spans multiple Adw versions. I was able to port all of the functionality into a single `CloseConfirmationDialog` class and make use of a good amount of Blueprint binds to simplify some stuff.


src/apprt/gtk-ng/class/clipboard_confirmation_dialog.zig
```diff
@@ -92,9 +136,18 @@ pub const ClipboardConfirmationDialog = extern struct {
         /// The clipboard contents being read/written.
         clipboard_contents: ?*gtk.TextBuffer = null,
 
+        /// Whether the contents should be blurred.
+        blur: bool = false,
+
         /// Whether the user can remember the choice.
         can_remember: bool = false,
 
+        /// Template bindings
```
tristan957: Should probably just be a `//` comment.

src/apprt/gtk-ng/class/clipboard_confirmation_dialog.zig
```diff
@@ -0,0 +1,398 @@
...
+    pub const ref = C.ref;
+    pub const unref = C.unref;
+    const private = C.private;
+
+    pub const Class = extern struct {
+        parent_class: Parent.Class,
+        var parent: *Parent.Class = undefined;
+        pub const Instance = Self;
+
+        fn init(class: *Class) callconv(.C) void {
```
tristan957, line 351: If you call [`gtk.WidgetClass.setCssName()`](https://docs.gtk.org/gtk4/class_method.Widget.set_css_name.html), you can use selectors in CSS like:

```css
clipboard-confirmation-dialog {
  // blah
}
```

Could be an improvement for later. Looked like you were using CSS classes in places where you probably wanted a name.

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -2169,7 +2169,7 @@ const Clipboard = struct {
         if (remember) switch (req.*) {
             .osc_52_read => surface.config.clipboard_read = .deny,
             .osc_52_write => surface.config.clipboard_write = .deny,
-            .paste => @panic("paste should not be rememberable"),
```
tristan957: Goat typo

src/apprt/gtk-ng/ui/1.0/clipboard-confirmation-dialog.blp
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      [overlay]
+      Button hide_button {
+        visible: false;
+        halign: end;
+        valign: start;
+        margin-end: 12;
+        margin-top: 12;
+      }
```
jcollie, line 79: Seems to be missing an image and adding the "opaque" style.

src/apprt/gtk-ng/ui/1.4/clipboard-confirmation-dialog.blp
```diff
@@ -0,0 +1,84 @@
...
+      }
+
+      [overlay]
+      Button hide_button {
+        visible: false;
+        halign: end;
+        valign: start;
+        margin-end: 12;
+        margin-top: 12;
+      }
```
jcollie, line 79: Missing the Image and the "opaque" style.

jcollie (changes requested):
Couple of changes needed to the blueprints…

src/apprt/gtk-ng/class/clipboard_confirmation_dialog.zig
jcollie: One would think that you'd be able to set the CSS name from the blueprint but Google is not turning up anything obvious.

mitchellh: Yeah this is a good idea. I was aware of this but didn't try it. Let me try it now.

mitchellh: This just isn't working for me... it won't style. I'm calling it directly in our class init. Worth looking into later.

PR #8031, opened 2025-07-22
build-system: Replace deprecated usages of root_source_file on addTests

jaysonsantos:
Yet another low-hanging fruit

### DIFF
--- a/pkg/wuffs/build.zig
+++ b/pkg/wuffs/build.zig
@@ -12,9 +12,12 @@ pub fn build(b: *std.Build) !void {
     });
 
     const unit_tests = b.addTest(.{
-        .root_source_file = b.path("src/main.zig"),
-        .target = target,
-        .optimize = optimize,
+        .name = "test",
+        .root_module = b.createModule(.{
+            .root_source_file = b.path("src/main.zig"),
+            .target = target,
+            .optimize = optimize,
+        }),
     });
     unit_tests.linkLibC();
 
--- a/pkg/cimgui/build.zig
+++ b/pkg/cimgui/build.zig
@@ -108,9 +108,11 @@ pub fn build(b: *std.Build) !void {
 
     const test_exe = b.addTest(.{
         .name = "test",
-        .root_source_file = b.path("main.zig"),
-        .target = target,
-        .optimize = optimize,
+        .root_module = b.createModule(.{
+            .root_source_file = b.path("main.zig"),
+            .target = target,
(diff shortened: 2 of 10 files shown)


pkg/wuffs/build.zig
```diff
@@ -12,9 +12,11 @@ pub fn build(b: *std.Build) !void {
     });
 
     const unit_tests = b.addTest(.{
-        .root_source_file = b.path("src/main.zig"),
-        .target = target,
-        .optimize = optimize,
+        .root_module = b.createModule(.{
+            .root_source_file = b.path("src/main.zig"),
+            .target = target,
+            .optimize = optimize,
```
jcollie, line 19: ~~To be consistent with the others, perhaps `.target` and `.optimize` shouldn't be moved to the `createModule` call and kept with the `addTest` call?~~

pkg/wuffs/build.zig
```diff
@@ -12,9 +12,11 @@ pub fn build(b: *std.Build) !void {
     });
 
     const unit_tests = b.addTest(.{
```
jcollie, line 14: Should have `.name = "test",` for consistency with the others.

jcollie (changes requested):
Couple of small things, otherwise LGTM.

pkg/wuffs/build.zig
jcollie: Whoops! Looks like it's the opposite. `.target` and `.optimize` need to be moved to `.createModule`.

pkg/wuffs/build.zig
jaysonsantos: done

mitchellh:
Looks like some legit CI failures here.

jaysonsantos:
Yes, sorry for the many commits.
zig build run (and zig build -Dapp-runtime=none test) on macos did not catch those and my linux box is off right now :D
i will do some cdd (ci driven development here) :D

pkg/wuffs/build.zig
jcollie: Looks like some of the `createModule` calls still need to be fixed up.

jcollie (approved):
LGTM, although it might be good to squash these commits together.

jaysonsantos:
@jcollie i just ran the tests on macos and linux then squashed the commits

mitchellh (approved):
Thanks! All this stuff will make our 0.15 migration that much easier...

PR #8038, opened 2025-07-23
New icon for Linux GTK app that aligns with the Gnome HIG

mitchellh:
This adds a new icon for the GTK-based application that adheres (mostly) to the Gnome Human Interface Guidelines (HIG). The icon is designed to fit in better with other Gnome applications. 

While there isn't a single standard "native" style amongst Linux applications, I believe this better fits the general Linux desktop ecosystem over our macOS icon.

The icon itself is undeniably Ghostty. The core design language is the same and I don't think ayone will mistake it for anything else. I wanted to keep the brand the same, but making it fit in better aligns with Ghostty's goal of being "platform native".

As of this PR, you can't use the macOS icon on Linux without modifying the source. We may provide a compile-time option to swap icons in the future (unfortunately Linux desktop applications require hardcoding an icon path in the desktop files, so making it runtime selectable is... messy!)


appsforartists:
are the sources to this icon published somewhere?

i was really tempted to replace the icon at 48dp and below with just the ghost (from the tahoe icon assets) to make the icon more legible at small panel sizes (and honestly still am).  

but while digging through this, i realized that you do have progressive detail disclosure at various sizes (the 16dp icon is laid out to make the ghost bigger than the 32dp icon).  i think the problem is that there is no `48x48@2` definition, so plasma is choosing a lower fidelity icon and upscaling it, which makes the icon feel illegible.

PR #8039, opened 2025-07-23
apprt/gtk-ng: surface drag and drop

mitchellh:


mitchellh:
Stand by, this has memory leaks.

mitchellh:
Fixed the leaks that were our fault. There are two more leaks that I think are in GTK (unsure if on purpose or not). There's a chance we're forgetting something but I can't see it. I ported the leak fixes to the existing gtk apprt.

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -1837,6 +1935,52 @@ const Clipboard = struct {
         );
     }
 
+    /// Paste explicit test directly into the surface, regardless of the
```
jcollie: `test` -> `text`

jcollie (approved):
Small typo, but LGTM

PR #8050, opened 2025-07-24
macos: add more Tahoe menu item icons from SF Symbols

ownia:
This is a follow up to https://github.com/ghostty-org/ghostty/pull/7594
Screenshot 2025-07-24 at 23 10 20
Screenshot 2025-07-24 at 23 11 01

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -471,6 +471,7 @@ class AppDelegate: NSObject,
         self.menuSplitUp?.setImageIfDesired(systemSymbolName: "rectangle.tophalf.inset.filled")
         self.menuSplitDown?.setImageIfDesired(systemSymbolName: "rectangle.bottomhalf.inset.filled")
         self.menuClose?.setImageIfDesired(systemSymbolName: "xmark")
+        self.menuPasteSelection?.setImageIfDesired(systemSymbolName: "doc.on.clipboard.fill")
         self.menuIncreaseFontSize?.setImageIfDesired(systemSymbolName: "textformat.size.larger")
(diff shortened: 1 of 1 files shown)


Aaron-212:
This design might not fully align with Apple’s Human Interface Guidelines.
From <https://developer.apple.com/design/human-interface-guidelines/the-menu-bar>
> **Use a single icon to introduce a group of similar items.** Instead of adding individual icons for each action, or reusing the same icon for all of them, establish a common theme with the symbol for the first item and rely on the menu item text to keep the remaining items distinct.

ownia:
> This design might not fully align with Apple’s Human Interface Guidelines. From https://developer.apple.com/design/human-interface-guidelines/the-menu-bar
> 
> > **Use a single icon to introduce a group of similar items.** Instead of adding individual icons for each action, or reusing the same icon for all of them, establish a common theme with the symbol for the first item and rely on the menu item text to keep the remaining items distinct.

Thanks for pointing this out! I'm wondering how to determine which menu items belong to a group of similar items. For example, should the prefix-based grouping 'New', 'Split' and 'Close' be used in the 'File' menu to categorize similar items.

mitchellh:
Indeed, it seems unclear. For these though I was copying macOS official apps on macOS 26 beta. For example, Xcode doesn't appear to put icons next to these, so I'm not sure we should either. I haven't looked in recent betas though.

ownia:
I just checked Xcode 26 beta 4 and I think it is reasonable to keep the current 'Close *' option as a single icon, but the 'Paste Selection' menu item needs to add an additional icon.

mitchellh:
If it matches Xcode 26 then I'm good with it, can you fix up the PR? Thanks

ownia:
> If it matches Xcode 26 then I'm good with it, can you fix up the PR? Thanks

done

PR #8057, opened 2025-07-25
gtk-ng: add debug warning banner

jcollie:


src/apprt/gtk-ng/class/window.zig
```diff
@@ -23,6 +25,29 @@ pub const Window = extern struct {
...
+    pub const properties = struct {
+        pub const debug = struct {
+            pub const name = "debug";
+            const impl = gobject.ext.defineProperty(
+                name,
+                Self,
+                bool,
+                .{
+                    .nick = "Debug",
+                    .blurb = "True if we were compiled with debug optimizations.",
```
00-kat: Aside: “debug optimizations” reads oddly to me because debug mode is the lack of optimizations as far as I know?

mitchellh (changes requested):
Welcome to NG :) Instead of making the window multi version I recommend making a GhosttyDebugBanner widget and making THAT multi version and then having a single window. I’m not sure if we can keep window single version but I think that’s been a better pattern so far (see the child exited as an example)

src/apprt/gtk-ng/class/window.zig

jcollie:
> Welcome to NG :) Instead of making the window multi version I recommend making a GhosttyDebugBanner widget and making THAT multi version and then having a single window. I’m not sure if we can keep window single version but I think that’s been a better pattern so far (see the child exited as an example)

Ah, yeah much better idea.

src/apprt/gtk-ng/class/debug_warning.zig
```diff
@@ -0,0 +1,80 @@
...
+    pub const Parent = if (adw_version.supportsBanner()) adw.Bin else gtk.Box;
+    pub const getGObjectType = gobject.ext.defineClass(Self, .{
+        .name = "GhosttyDebugWarning",
+        .instanceInit = &init,
+        .classInit = &Class.init,
+        .parent_class = &Class.parent,
+    });
+
+    pub const properties = struct {
+        pub const debug = struct {
```
mitchellh: You actually don't need this at all! This widget can assume it's always visible, and we can set the `visible` binding in window.blp! 😄

src/apprt/gtk-ng/ui/1.5/window.blp
```diff
@@ -2,10 +2,21 @@ using Gtk 4.0;
...
   default-width: 800;
   default-height: 600;
 
-  content: $GhosttySurface surface {
-    close-request => $surface_close_request();
+  content: Gtk.Box {
+    orientation: vertical;
+    spacing: 0;
+
+    $GhosttyDebugWarning {}
```
mitchellh: put `visible: bind template.whatever` here instead!

mitchellh (changes requested):
Awesome, much better. One more simplification.

src/apprt/gtk-ng/class/debug_warning.zig
jcollie: Done.

src/apprt/gtk-ng/ui/1.5/window.blp
jcollie: Done.

mitchellh (approved):

PR #8063, opened 2025-07-25
ci: move to bleedging edge sequoia builds

mitchellh:
This will bring in Xcode 26 Beta 4 which I believe fixes all the known issue we were dealing with keeping us on beta 1.


mitchellh:
Hm, this is straight up _crashing_ xcodebuild in CI with a C++ exception, but everything works locally even with a clean cache.

PR #8064, opened 2025-07-25
macos: in new_window action, activate App

aljoscha:
> [!NOTE]  
> The change might have been intentional, and so I lack context. I mention two ways to fix below, this PR implements the first possible fix.

This change makes sure that the new window is focused and visible.

When commit 33d128bcff2ee529359a844bde50c3aa9dfed460 removed the TerminalManager class and moved its functionality into TerminalController, it accidentally removed app activation for windows triggered by global keybinds.

How the bug works:

   1. Menu actions (like File → New Window) call AppDelegate.newWindow() which: 2. Calls TerminalController.newWindow() 3. AND explicitly calls NSApp.activate(ignoringOtherApps: true) in the AppDelegate
   4. Global keybind actions trigger ghosttyNewWindow() notification handler which:
      5. Only calls TerminalController.newWindow()
      6. Does NOT call NSApp.activate(ignoringOtherApps: true)
   7. While TerminalController.newWindow() does call NSApp.activate(ignoringOtherApps: true) internally, this call happens before the async dispatch that shows the window, so the activation occurs but the window isn't focused when it's actually shown.
   8. In the old TerminalManager.newWindow(), the activation happened immediately before the async dispatch, ensuring proper timing for window focus.

To see the bug in action:
- run recent Ghostty `main`
- set up a global keybind for `new_window`
- focus some other window
- trigger keybind
- notice that Ghostty doesn't come to the foreground, but when manually switching to Ghostty you will see that the new window _was_ created

The fix would be to either move the NSApp.activate() call back into TerminalController.newWindow(), as it was for TerminalManager, or add the activation call to the notification handlers in AppDelegate.

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -214,6 +214,10 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
             }
         }
 
+        // All new_window actions force our app to be active, so that the new
+        // window is focused and visible.
+        NSApp.activate(ignoringOtherApps: true)
+
         // We're dispatching this async because otherwise the lastCascadePoint doesn't
         // take effect. Our best theory is there is some next-event-loop-tick logic
         // that Cocoa is doing that we need to be after.


danielwe:
This is great! The new window not coming to the front has been annoying me a lot.

> or add the activation call to the notification handlers in AppDelegate.

To me, this seems like the better solution for two reasons:
* It avoids activating twice in the `@IBAction` handler
* It looks like `TerminalController.newWindow` is sometimes called in contexts where there shouldn't be an `activate` call, like in `AppDelegate.application`

So I think the call should be added at the end of `AppDelegate.ghosttyNewWindow`, and possibly also `AppDelegate.ghosttyNewTab`, rather than in `TerminalController.newWindow`.

aljoscha:
> So I think the call should be added at the end of AppDelegate.ghosttyNewWindow, and possibly also AppDelegate.ghosttyNewTab, rather than in TerminalController.newWindow.

I can push that change if there's interest in merging this 👌

danielwe:
Really hope so! This should be a super quick and easy review for anyone who has merge rights. And regarding your comment in the OP, I highly doubt this was intentional. Almost certainly just an oversight during a big refactor.

danielwe:
While we wait for someone to review this: Do you know if there's a way to make it so that only the new/primary window pops to the front when activating? On my patched Ghostty where I add the AppDelegate variant of this fix, my global new window keybind brings all Ghostty windows to the front, not just the newly created one, and that's kind of annoying.

aljoscha:
Nah, I don't know how to do that. For me it brings the new window and one previous window to the front. 🤷‍♂️

danielwe:
Ah, figured it out. You need to make the activate call async, presumably to ensure that it happens after the new window is fully set up, as that also involves some async calls.

So instead of
```swift
        NSApp.activate(ignoringOtherApps: true)
```
do
```swift
        DispatchQueue.main.async {
            NSApp.activate(ignoringOtherApps: true)
        }
```

mitchellh (approved):

aljoscha:
🙇‍♂️

PR #8065, opened 2025-07-25
apprt/gtk-ng: hook up bell ringing

mitchellh:
For now, this just emits a signal that an embedding widget will react to to do something like flashing the window. 

I think we should implement the audio bell in the actual surface view but I don't currently have audio drivers hooked up in my Linux VM and I'm away from my desktop PC. :)

cc @jcollie if you're interested, pretty tightly scoped.


jcollie (approved):
LGTM, gets things started.

PR #8069, opened 2025-07-25
gtk-ng: add ipc infrastructure and connect +new-window ipcs

jcollie:


src/apprt/gtk-ng/global.zig
```diff
@@ -0,0 +1,15 @@
+//! global static definitions
+
+const builtin = @import("builtin");
+
+/// GTK application ID
+pub const application_id = switch (builtin.mode) {
```
mitchellh: Let's just put these constants either directly `gtk-ng.zig` or `gtk-ng/App.zig`. I think we should _really_ use `build_config.zig` for this instead because this can be overridden via build configs no? But I'm willing to TODO that.

src/apprt/gtk-ng/ipc.zig
```diff
@@ -0,0 +1,3 @@
+//! functions for handling IPC
+
+pub const newWindow = @import("ipc/new_window.zig").newWindow;
```
mitchellh: Given our `App.zig` is just a thin layer now, I think we can move this directly into App.zig. This feels like a lot of ceremony to forward a function declaration a couple times.

src/apprt/gtk-ng/class/application.zig
```diff
@@ -751,6 +755,40 @@ pub const Application = extern struct {
         );
     }
 
+    fn startupGtkActions(self: *Self) void {
```
mitchellh: My branch is going to conflict with this probably... ill PR mine shortly.

src/apprt/gtk-ng/class/application/GtkApplicationAction.zig
```diff
@@ -0,0 +1,57 @@
+//! GTK action handlers for Application
+const GtkApplicationAction = @This();
```
mitchellh: I don't think we should put these in a separate file, take a look at my `gtk-ng-win` branch to see where I put them.

mitchellh (changes requested):

src/apprt/gtk-ng/global.zig
jcollie: Hmm, I initially had these in `build_config.zig` but I figured that they were just too GTK specific. `apprt/gtk-ng/App.zig` seems like a decent spot though.

src/apprt/gtk-ng/class/application/GtkApplicationAction.zig
jcollie: OK, moved to `application.zig`

src/apprt/gtk-ng/ipc.zig
jcollie: Done.

src/apprt/gtk-ng/class/application.zig
jcollie: Yeah, but I've synced it with your branch so conflicts should be minimal.

mitchellh:
Okay merged. Please rebase!

jcollie:
> Okay merged. Please rebase!

Rebase done!

mitchellh (approved):

PR #8071, opened 2025-07-25
apprt/gtk-ng: window headerbar, maximize, fullscreen

mitchellh:
This ports over a basic headerbar, maximize, and fullscreen. The headerbar only has the main menu button for now since we have no tabbing or splits. The main menu has the full main menu that mainline GTK has but most of it is disabled since we don't implement the actions yet. 

I didn't use anything from your branch @tristan957 so I didn't add coauthor but I want to note that @tristan957 worked on this as well and I suspect there's overlap.


mitchellh:
Got a bit more in there: window close confirmations, copy/paste/reset/clear actions, and copy action doesn't activate unless there is a selection.

PR #8075, opened 2025-07-26
Fix i3 window border disappearing after fullscreen toggle

jebaum:
Bug: When toggling a Ghostty window between fullscreen and windowed mode in the i3 window manager, or between tabbed mode and default tiling mode, window borders would permanently stop being drawn around the terminal window.

I bisected to find that the issue was first introduced in aa2dbe2.

The below explanation + the code in this PR was 100% generated by claude code, since I don't know the first thing about x11 or terminal development, and have never written any Zig before. I verified this code change builds and solves the issue for me which I hope is more helpful than just a bug report. If this code is actually garbage and the issue should be fixed a different way it wouldn't shock me and this PR can just be closed in favor of one from someone who knows what they're doing.

Anyway this is the explanation that the llm generated:

Root cause was that syncAppearance() was updating X11 properties on every call during window transitions, even when values hadn't changed. These redundant property updates interfered with i3's border management.

The fix adds caching to syncBlur() and syncDecorations() to only update X11 properties when values actually change, eliminating unnecessary property changes during fullscreen transitions.

### DIFF
--- a/src/apprt/gtk/winproto/x11.zig
+++ b/src/apprt/gtk/winproto/x11.zig
@@ -173,7 +173,9 @@ pub const Window = struct {
 
     blur_region: Region = .{},
 
-    // Cache last applied values to avoid redundant X11 property updates
+    // Cache last applied values to avoid redundant X11 property updates.
+    // Redundant property updates seem to cause some visual glitches
+    // with some window managers: https://github.com/ghostty-org/ghostty/pull/8075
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Sorry for the delay here. I think your experience of this working plus this code being sensible works for me. I did some manual cleanups but otherwise we're good.

Issue #7930, opened 2025-07-13
Add TUI mechanism to filter light/dark themes to `+list-themes`


mitchellh:
#7927 

Our `+list-themes` CLI already has arguments for `--color` to filter, but it'd be nice to have a way to filter them once the TUI is already running, too. I don't have a strong opinion right now on what this might look like, open to ideas cc @jcollie

jcollie:
I think a simple keybind would work well, maybe 'f' and/or 'F' for "filter"? Would toggle between showing all themes, light themes and dark themes.

mitchellh:
I'm fine with that at least for now, it'd show up in `?` anyways.

Sheeplet1:
Hi, I've also implemented this in PR #8082 prior and don't mean to step on any toes. Happy to let maintainers decide which approach works best. :)

jcollie:
Fixed in #8082

PR #8079, opened 2025-07-26
cli: add color theme filter to +list-themes TUI

ddogfoodd:
This PR tries to implement #7930
In the TUI, key `f` toggles between all/dark/light theme filters.

However I would really appreciate some further guidance, since there are still some shortcomings I am yet unsure about.

- [ ] This PR modifies the `Preview` struct rather than `Options.colors` which `ghostty +list-themes --color=light` does. This results in the CLI taking priority over the TUI toggle, which I personally think is fair. Is that okay, though?
- [x] Searching for a theme currently only works when the filter is set to all themes. It doesn't work when filtering dark/light. I think I can figure out the issue here, just need a bit more time and testing. If you see it, before I do, let me know.
- [ ] This PR copied and modified the work of commit 0b5160e9f01b6d939a1c213478cbcbec2982e3aa `shouldIncludeTheme()` and introduces `shouldIncludeFiltered()`. I am almost certain that there is a more proper way to do that, but which I don't know yet. Please let me know if you know how to do that in a cleaner way (see both functions near the end of file).

### DIFF
--- a/src/cli/list_themes.zig
+++ b/src/cli/list_themes.zig
@@ -220,6 +220,11 @@ const Preview = struct {
         search,
         save,
     },
+    color_theme_filter: enum {
+        all,
+        dark,
+        light,
+    },
     color_scheme: vaxis.Color.Scheme,
     text_input: vaxis.widgets.TextInput,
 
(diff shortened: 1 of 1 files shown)


ddogfoodd:
Closing in favor of #8082 which fixes the two remaining pain points and is overall a cleaner implementation.

PR #8080, opened 2025-07-26
apprt/gtk-ng: fix scrolling on surface

mitchellh:
We forgot to specify the scroll controller's `flags` property.


src/apprt/gtk-ng/ui/1.2/surface.blp
```diff
@@ -103,6 +103,7 @@ template $GhosttySurface: Adw.Bin {
     scroll => $scroll();
     scroll-begin => $scroll_begin();
     scroll-end => $scroll_end();
+    flags: both_axes;
```
mitchellh, line 106: This is the actual change. The other was just a nesting style thing.

PR #8083, opened 2025-07-27
gtk-ng: port the command palette

jcollie:


src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+        if (comptime !Self.isSupported()) return null;
+
+        const self = gobject.ext.newInstance(Self, .{});
+
+        const app = Application.default();
+        const priv = self.private();
+
+        // The config that we will use to begin with. It should already have
+        // had it's reference count increased, and we take ownership of it.
+        priv.config = config;
```
mitchellh: I found its cleaner to make sure these are properties with proper setter/getters and setup initialization in `init` rather than `new`. That's how we do config in other classes.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+
+        // Note that we _do not_ take a reference to the window as this can
+        // cause a circular reference count, which means that the window/command
+        // palette pair will never be destroyed`.
+        priv.window = window;
```
mitchellh: I'm curious to look into this... feels like during `dispose` of `Window` we should release the reference?

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+
+        // Initialize an arena to make cleaning up our allocations easier.
+        priv.arena = .init(app.core().alloc);
```
mitchellh: Definitely should do this in `init` instead.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+        // Ensure that we always have the same config as the window.
+        _ = gobject.Object.bindProperty(
+            window.as(gobject.Object),
+            "config",
+            self.as(gobject.Object),
+            "config",
+            .{},
+        );
+
+        self.updateConfig();
```
mitchellh: Also definitely an `init` thing.  I think this whole `new` logic can go in `init`.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+
+    fn init(self: *Self, _: *Class) callconv(.c) void {
+        gtk.Widget.initTemplate(self.as(gtk.Widget));
+    }
+
+    fn dispose(self: *Self) callconv(.c) void {
+        const priv = self.private();
+
+        priv.source.removeAll();
+        priv.arena.deinit();
```
mitchellh: This should be done in `finalize`. If I understand correctly, `dispose` in the gobject model should only be for removing references to things.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+
+    fn rowActivated(_: *gtk.ListView, pos: c_uint, self: *CommandPalette) callconv(.c) void {
+        self.activated(pos);
+    }
+
+    //---------------------------------------------------------------
+
+    pub fn toggle(self: *CommandPalette) void {
+        const priv = self.private();
+        const window = priv.window orelse {
```
mitchellh: Hm, just thinking out loud here, I wonder if we should associate the command palette with a _surface_ instead. That's how it works on macOS and this lets the command palette remain opening when switching tabs. I'm not sure if that'd translate well to GTK so this is more of an open thought.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+            const accel = try key.accelFromTrigger(&buf, trigger) orelse break :action null;
+            break :action try alloc.dupeZ(u8, accel);
+        };
+
+        priv.cmd_c = .{
+            .title = cmd.title.ptr,
+            .description = cmd.description.ptr,
+            .action = if (action) |v| v.ptr else "",
+            .action_key = try std.fmt.allocPrintZ(alloc, "{}", .{cmd.action}),
+        };
```
mitchellh, line 521: Same thoughts here around `new` vs `init`. And I wonder if we should make the C type a boxed type so we can just pass it in as a normal GTK property.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,426 @@
...
+
+        // Close before running the action in order to avoid being replaced by
+        // another dialog (such as the change title dialog). If that occurs then
+        // the command palette dialog won't be counted as having closed properly
+        // and cannot receive focus when reopened.
+        _ = priv.dialog.close();
+
+        const action = cmd.getAction() orelse return;
+
+        window.performBindingAction(action);
```
mitchellh: Feels like this should be a signal instead and allow embedders to do what they want when an action is chosen.

mitchellh (changes requested):

src/apprt/gtk-ng/class/command_palette.zig
jcollie: I found that would leak a reference to the `config` because if you pass the config to `gobject.ext.newInstance` it would increment the references of `config`. Let me try again as I was doing a lot of this late at night.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: I tried that, but then Ghostty wouldn't quit - the dispose virtual calls were never called because there was always at least one reference.

src/apprt/gtk-ng/class/command_palette.zig
mitchellh: You'd have to write a custom direct `Value` setter/getter.

src/apprt/gtk-ng/class/command_palette.zig
mitchellh: I wonder if we can make the property a WeakRef type?

src/apprt/gtk-ng/class/command_palette.zig
jcollie: That would mean a lot more command palette objects laying around. Right now we have one per window and it lasts the lifetime of the window. Doing it per surface means a lot more long-lived objects. Plus the dialog is modal to the window so you can't really be switching tabs/splits while the palette is open.

mitchellh: On macOS we dispose of the entire object whenever it's closed so you only have as many active as are open. I think we should do this here too (regardless of our target). 

> Plus the dialog is modal to the window so you can't really be switching tabs/splits while the palette is open.

Ah, yeah this would be a limitation. I wonder if longer term we should change it to an overlay item.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Yeah a WeakRef seems to work well.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: May need to update this with the custom private ref-counted accessor as discussed on Discord but I've moved all the logic to `.init`.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Done.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Actually not needed anymore with signals and binding properties.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Done.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: That seems like we're over-complicating things, as this is `gtk-ng` talking to `gtk-ng` so there's no libghostty embedding going on.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: I've changed the command palette so that it's dynamically allocated as needed, and then disposed of. Attaching it to a surface would be possible in the future, but right now there is no code for getting the window that a surface is attached to (at least not that I can find).

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Rebased with the new custom getter/setter for ref counted objects.

src/apprt/gtk-ng/class/command_palette.zig
pluiedev: Well it can't be a boxed type because the gio.ListModel interface is annoying and requires its items to implement gobject.Object 😅 I left that comment there for that exact reason

src/apprt/gtk-ng/class/command_palette.zig
pluiedev: FWIW adw.Dialog can only ever be attached to a window (the present method technically takes any widget but the dialog will just attach itself to the window that contains said widget), so it seems like a fundamental limitation unless we switch to use overlays. But then, I'm not sure if that's a good idea since we do have a lot of information to display and restricting ourselves to only the area of the split might only complicate matters

mitchellh: On macOS the palette always goes over the window, not the split. It’s just per tab and not a modal on the window.

mitchellh:
FYI I'm holding on this for a second until I get tabbing in because that's going to move a bunch of stuff around and I don't want to also have to move this at the same time.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: I've redone the Command object to be more fully gobject-ized.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Well, I've changed my mind on this and made much more use of signals to communicate between various objects.

jcollie:
I've significantly reworked the command palette. I also added some suppressions to valgrind to clean up some possible leaks in fontconfig. I'm pretty sure they are fontconfig's problem but I'm new enough at valgrind to be unsure. I noticed other possible leaks when messing around with various actions (notably reset_font_size) but I was even less sure about adding suppressions for those.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,652 @@
...
+        // if there were errors.
+        defer self.unref();
+
+        // Use priv.model and not priv.source here to use the list of *visible* results
+        const object = priv.model.as(gio.ListModel).getObject(pos) orelse return;
+        defer object.unref();
+
+        const cmd = gobject.ext.cast(Command, object) orelse return;
+
+        const activation = CommandActivation.new(cmd.getAction() orelse return);
```
mitchellh: We can get rid of `CommandActivation` completely and use a `*const Binding.Action` directly.

In my PR yesterday for global shortcuts, I made binding action a boxed type. It doesn't do a full clone, but we don't need it to because signal emission is synchronous and never copies the value. Connected handlers can copy it if they need to.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -0,0 +1,652 @@
...
+                    .blurb = "The configuration that this command palette is using.",
+                    .accessor = C.privateObjFieldAccessor("config"),
+                },
+            );
+        };
+    };
+
+    pub const signals = struct {
+        /// Emitted when a command from the command palette is activated.
+        pub const activate = struct {
```
mitchellh: Nit: I understand why `activate` because it matches some other nomenclature for widgets used by the palette, but I think `trigger` would be better here to match Ghostty naming and our global shortcuts.

mitchellh (changes requested):
Looking really good! A couple requests here.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Done.

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Done.

mitchellh (approved):
The suppressions look good to me. All the fontconfig stuff is definitely good.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -50,7 +50,7 @@ pub const CommandPalette = extern struct {
         /// action contains pointers to allocated data so if a receiver of this
         /// signal needs to keep the action around it will need to clone the
         /// action or there may be use-after-free errors.
-        pub const tigger = struct {
```
jcollie: DOH!

PR #8085, opened 2025-07-28
feat: add option to disable the "Reloaded the configuration" notification

dy0gu:
These changes were discussed in #8048.

The option defaults to `true` for backwards compatibility.

**NOTE:** I was feeling a bit lazy and didn't want to setup Nix (always a pain on Windows), so I decided to take this chance to try the Copilot agent for the first time, using just the GitHub web UI. I hope that's okay given the small scope of these changes!

### DIFF
--- a/src/apprt/gtk-ng/class/window.zig
+++ b/src/apprt/gtk-ng/class/window.zig
@@ -306,7 +306,16 @@ pub const Window = extern struct {
         _: *gobject.ParamSpec,
         self: *Self,
     ) callconv(.c) void {
-        self.addToast(i18n._("Reloaded the configuration"));
+        // We only toast if configured to
+        const priv = self.private();
+        const config_obj = priv.config orelse {
+            self.syncAppearance();
+            return;
+        };
+        const config = config_obj.get();
+        if (config.@"app-notifications".@"config-reload") {
+            self.addToast(i18n._("Reloaded the configuration"));
+        }
         self.syncAppearance();
     }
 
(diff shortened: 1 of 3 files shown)


pluiedev:
Uhm, where's the code gone

dy0gu:
WTH, the force push from the bot reset everything back.

I'll add it again in a sec.

src/apprt/gtk-ng/class/window.zig
```diff
@@ -306,7 +306,16 @@ pub const Window = extern struct {
...
+        // We only toast if configured to
+        const priv = self.private();
+        const config_obj = priv.config orelse {
+            self.syncAppearance();
+            return;
+        };
+        const config = config_obj.get();
+        if (config.@"app-notifications".@"config-reload") {
+            self.addToast(i18n._("Reloaded the configuration"));
+        }
```
dy0gu, line 318: I'm not sure how `syncAppearance()` should be called in this case so to preserve the initial behavior I'm calling it whether we toast or don't toast and even if there is no `priv.config`.

dy0gu:
Yo what the bot doing, it reverted my changes with the force push again ;(

dy0gu:
Sorry about that, should be ready for review now.

mitchellh (approved):
Looks good pending CI. Thanks.

mitchellh:
Replaced by #8366 which retains your commits but just adds some style changes. Can't push back up to your tree for some reason.

PR #8093, opened 2025-07-28
macos: fix quick terminal fullscreen crash bug

peanball:
Fullscreen on quick terminal was failing with a crash, when it tried
to save the state of a non-existent toolbar and its accessory view
controllers.

See #7980

### DIFF
--- a/macos/Sources/Helpers/Fullscreen.swift
+++ b/macos/Sources/Helpers/Fullscreen.swift
@@ -407,8 +407,14 @@ class NonNativeFullscreen: FullscreenBase, FullscreenStyle {
             self.styleMask = window.styleMask
             self.toolbar = window.toolbar
             self.toolbarStyle = window.toolbarStyle
-            self.titlebarAccessoryViewControllers = window.titlebarAccessoryViewControllers
             self.dock = window.screen?.hasDock ?? false
+            
+            self.titlebarAccessoryViewControllers = if (window.hasTitleBar) {
+                // Accessing titlebarAccessoryViewControllers without a titlebar triggers a crash.
+                window.titlebarAccessoryViewControllers
+            } else {
+                []
+            }
 
             if let cgWindowId = window.cgWindowId {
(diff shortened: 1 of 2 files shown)


mitchellh (commented):
Huh, I thought I did something like this already! Maybe it got pulled out at some point... I'll take a look and test this. Thanks.

macos/Sources/Helpers/Fullscreen.swift
```diff
@@ -407,7 +407,13 @@ class NonNativeFullscreen: FullscreenBase, FullscreenStyle {
             self.styleMask = window.styleMask
             self.toolbar = window.toolbar
             self.toolbarStyle = window.toolbarStyle
-            self.titlebarAccessoryViewControllers = window.titlebarAccessoryViewControllers
+            
+            if (window.hasTitleBar) {
+                /* QuickTerminal has no title bar and does not support window.titlebarAccessoryViewControllers */
```
freepicheep: This isn't a big deal, but to my knowledge, all comments in swift files in the ghostty codebase use `//` and not `/* */`. It's nice to be consistent.

peanball: Yep, I'll adjust it. The comment should also be in the if block, as that's the part that requires explanation.

peanball: Adjusted and moved the comment to the relevant part, where we _don't_ save the `titlebarAccessoryViewControllers` if there is no titlebar.

peanball:
@mitchellh do you think we could include this in the scope for 1.2.0?

mitchellh:
@peanball Yes absolutely will check this before 1.2

mitchellh (approved):
Thanks! Verified the fix. Found a bug with fullscreen + hide/show that I also fixed while testing this. Added as a second commit.

peanball:
Awesome, thanks! Looking forward to the next `tip` 😁

PR #8094, opened 2025-07-28
Keybind to toggle between latest used tab.

buraksirma:
My workflow heavily depends on that keybinding, I feel like my code is quite hacky but it solves the problem I have. I use it on both linux and macos so I'm sure it works for both sides. I'm also sure that it works with mouse actions as well and movements / closure of tabs. If someone leads me I can try to improve the code quality.

It solves: #1844

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -507,6 +507,7 @@ typedef enum {
   GHOSTTY_GOTO_TAB_PREVIOUS = -1,
   GHOSTTY_GOTO_TAB_NEXT = -2,
   GHOSTTY_GOTO_TAB_LAST = -3,
+  GHOSTTY_GOTO_TAB_LAST_USED = -4,
 } ghostty_action_goto_tab_e;
 
 // apprt.action.Fullscreen
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -56,6 +56,9 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
     /// This will be set to the initial frame of the window from the xib on load.
     private var initialFrame: NSRect? = nil
 
+    /// This is the last used tab window for tab switching
+    private weak var lastUsedTabWindow: NSWindow?
+
     init(_ ghostty: Ghostty.App,
          withBaseConfig base: Ghostty.SurfaceConfiguration? = nil,
          withSurfaceTree tree: SplitTree<Ghostty.SurfaceView>? = nil,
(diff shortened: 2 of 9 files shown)


phush0:
I am also for this PR, as it is very handy and needed function (left over from TMUX)

edit: working as expected under linux

jcollie:
I'm not going to comment much on the code, except to say that too many debug log statements were left in the PR. But you may want to redo this one and patch the `gtk-ng` app runtime instead. We are close to completely removing the old `gtk` app runtime so I doubt that anyone is going to review/accept this patch as-is.

pluiedev:
It also doesn't address the use case of cycling across *all* tabs in most-recently-used order as some have requested in the original thread. While this could work as a stop-gap patch for some users I'm not sure if we would want to accept this as-is.

phush0:
there should be option for LRU order as I don't want it, I want to jum trough two tabs only not all of them lile I can in TMUX and kitty

pluiedev:
Well this is why that discussion remains a *discussion* - we need to iron out exactly what the feature should include and how users should use it. As this currently stands there's no way to implement MRU order without undoing what this PR already does. 

Again, it's fine if this is a personal patch, but we do have certain standards for PRs that aim to be merged into Ghostty itself.

buraksirma:
It's a personal patch. It was holding me back from trying Ghostty daily, so I implemented it and posted it here. Anyone experiencing the same issue can apply this patch and utilize it. 

If there is an intention to implement this feature in Ghostty, which involves moving back and forth with the last active tab, I can also implement it for `gtk-ng`. Traversing in the recent tabs is not a feature I'm looking for; it's not a desired feature for me, which is why I didn't try to do anything on that part, and there was no decision for what to do at all.

pluiedev:
If it's a personal patch you should maintain it in your fork, which other people can already see anyway. A PR communicates an intent to merge changes upstream, and this is currently not up to our standards. Lots of other people have posted their own patches and forks inside discussions and that's great — but we cannot accept this as an upstream change that us maintainers have to deal with later on. As documented in our CONTRIBUTING.md, we do reserve the right to deny any PRs that we do not deem sufficiently suitable.

mitchellh:
Before anyone gets the wrong idea, I want to note I appreciate the PR. It's helpful to see even personal patches, so thank you for your time. We'll assess if this makes sense in its current form to include, but as noted we may have to defer back to the discussion if it isn't in the form we expect. Thanks. ❤️

paranoidPhantom:
I personally would love this feature. It seems to just make sense to have "LRU" as the default CTRL + TAB behavior 🤔.

freddong:
image
I think it's a common requirement, warp has this.

PR #8105, opened 2025-07-30
apprt/gtk-ng: desktop notifications, open config, open url, present terminal

mitchellh:
Implements a number of minor apprt actions. More or less directly ported. Some notes added for future improvements given the new architecture.


mitchellh:
Spawned off some desktop notifications and Valgrind is clean here.

PR #8114, opened 2025-07-31
macOS: Add macos-dock-drop-folder-behavior (new tab or window) configuration option

kgn:
This PR adds a new configuration option `macos-dock-drop-folder-behavior` that controls whether folders dropped onto the Ghostty dock icon open in a new tab (default) or a new window.

### Configuration Option Added
- **Option name**: `macos-dock-drop-folder-behavior`
- **Valid values**: 
  - `tab` (default) - Opens folders in a new tab in the main window
  - `window` - Opens folders in a new window
- **Platform**: macOS only

### Files Modified

1. **`src/config/Config.zig`**
   - Added `MacOSDockDropFolderBehavior` enum with `tab` and `window` values
   - Added configuration field with default value of `.tab`
   - Added documentation explaining the option

2. **`macos/Sources/Ghostty/Package.swift`**
   - Added `MacOSDockDropFolderBehavior` enum to match the Zig enum

3. **`macos/Sources/Ghostty/Ghostty.Config.swift`**
   - Added `macosDockDropFolderBehavior` computed property to access the configuration value from Swift

4. **`macos/Sources/App/macOS/AppDelegate.swift`**
   - Modified `application(_:openFile:)` method to check the configuration
   - When a folder is dropped on the dock icon, it now respects the user's preference

## Usage

Add to your Ghostty configuration file:
```
macos-dock-drop-folder-behavior = window
```

## Motivation

This feature is useful for users (like me!) who prefer window-based workflows over tab-based workflows when opening folders via drag and drop on macOS.

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -405,14 +405,6 @@ class AppDelegate: NSObject,
             // When opening a directory, check the configuration to decide
             // whether to open in a new tab or new window.
             config.workingDirectory = filename
-            
-            let behavior = ghostty.config.macosDockDropFolderBehavior
-            if behavior == .window {
-                _ = TerminalController.newWindow(ghostty, withBaseConfig: config)
-            } else {
-                // Default to tab behavior
-                _ = TerminalController.newTab(ghostty, withBaseConfig: config)
-            }
         } else {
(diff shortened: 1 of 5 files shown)


mitchellh (approved):
Thanks, looks good. I modified the configuration to apply to both folders/files because it appears iTerm does that as well (not sure if its configurable, but the same behavior for both).

PR #8115, opened 2025-07-31
apprt/gtk-ng: initial size apprt action (window-width/height)

mitchellh:
Simple port. I might add size limits if I get to it, but finished this and it works so opened it up.


jcollie (approved):
LGTM

mitchellh:
This has a bug where every time you reload the config it resizes the window. I'm going to fix this in a follow up PR when I add size limits, too.

PR #8122, opened 2025-08-02
fix: correct the cursor Y position value exposed to shader uniforms

ClearAspect:
Fix for discussion #8113

The cursor Y position value exposed to the shader uniforms was incorrectly calculated. As per the doc in cell_text.v.glsl: In order to get the top left of the glyph, we compute an offset based on the bearings. The Y bearing is the distance from the bottom of the cell to the top of the glyph, so we subtract it from the cell height to get the y offset.

This calculation was mistakenly left out of the original code.

This will ensure that the custom shaders using
iCurrentCursor/iPreviousCursor get the correct Y coordinate representing the top-left corner of the cursor rectangle, matching the documented uniform behavior

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -2233,15 +2233,13 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                 );
 
                 pixel_x += @floatFromInt(cursor.bearings[0]);
-                pixel_y += @floatFromInt(cursor.bearings[1]);
+                // Convert the Y coordinate from bottom-to-top to top-to-bottom.
+                // Otherwise we end up with glyphs like underline at the top of the cell.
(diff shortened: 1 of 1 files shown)


qwerasd205:
Alright, so, as I suspected, this breaks the positioning under Metal lol. Don't worry, I'll fix it so it's correct on both platforms and push the changes to this PR.

qwerasd205:
Fixed under both platforms now, tested with custom cursor glyph that has different distances from each edge of the cell.
Metal:
image
OpenGL:
image

ClearAspect:
> Alright, so, as I suspected, this breaks the positioning under Metal lol. Don't worry, I'll fix it so it's correct on both platforms and push the changes to this PR.

Ah, thank you. I'm not exceeding familiar with the whole pipeline and when I got around to building and testing my fix I must have only done it on my Linux machine.

mitchellh (approved):

PR #8123, opened 2025-08-02
apprt/gtk-ng: winproto behaviors (quick terminal, csd/ssd, blur, etc.)

mitchellh:
This ports over the winproto behaviors to gtk-ng. The core winproto logic is unchanged except for trivial typing changes. The interaction with winproto is a bit different in ng due to the class separation of logic between surfaces and windows, but functionally the same.

Ran against Valgrind and all looks good.

### DIFF
--- a/src/apprt/gtk-ng/class/application.zig
+++ b/src/apprt/gtk-ng/class/application.zig
@@ -1727,20 +1727,20 @@ const Action = struct {
         // Find a quick terminal window.
         const list = gtk.Window.listToplevels();
         defer list.free();
-        const elem_: ?*glib.List = list.findCustom(null, struct {
-            fn callback(data: ?*const anyopaque, _: ?*const anyopaque) callconv(.c) c_int {
-                const ptr = data orelse return 1;
-                const gtk_window: *gtk.Window = @ptrCast(@alignCast(@constCast(ptr)));
-                const window = gobject.ext.cast(
+        if (ext.listFind(gtk.Window, list, struct {
+            fn find(gtk_win: *gtk.Window) bool {
+                const win = gobject.ext.cast(
                     Window,
-                    gtk_window,
-                ) orelse return 1;
-                if (window.isQuickTerminal()) return 0;
-                return 1;
+                    gtk_win,
+                ) orelse return false;
+                return win.isQuickTerminal();
             }
-        }.callback);
-        const elem = elem_ orelse return null;
-        return @ptrCast(@alignCast(elem.f_data));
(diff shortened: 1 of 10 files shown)


src/apprt/gtk-ng/class/application.zig
```diff
@@ -1694,6 +1701,48 @@ const Action = struct {
...
+                const window = gobject.ext.cast(
+                    Window,
+                    gtk_window,
+                ) orelse return 1;
+                if (window.isQuickTerminal()) return 0;
+                return 1;
+            }
+        }.callback);
+        const elem = elem_ orelse return null;
+        return @ptrCast(@alignCast(elem.f_data));
```
pluiedev: This pattern is getting common enough for me to wonder if we should introduce a helper for it...

src/apprt/gtk-ng/class/window.zig
```diff
@@ -466,12 +522,28 @@ pub const Window = extern struct {
...
+        const priv = self.private();
+        const csd_enabled = priv.winproto.clientSideDecorationEnabled();
+        self.as(gtk.Window).setDecorated(@intFromBool(csd_enabled));
+
+        // Fix any artifacting that may occur in window corners. The .ssd CSS
+        // class is defined in the GtkWindow documentation:
+        // https://docs.gtk.org/gtk4/class.Window.html#css-nodes. A definition
+        // for .ssd is provided by GTK and Adwaita.
+        self.toggleCssClass("csd", csd_enabled);
+        self.toggleCssClass("ssd", !csd_enabled);
```
pluiedev, line 545: This should also cover `solid-csd` (see https://github.com/ghostty-org/ghostty/discussions/5660#discussioncomment-13598149)

src/apprt/gtk-ng/class/application.zig
mitchellh: Definitely. The ptrCast/alignCast/constCast is also nasty so I'll pull this out into a helper. We can't use the `gobject.ext` helpers here either because the any opaque type doesn't work with them (comp time error, I think their checks are too strict). So yeah, I'll pull this out.

src/apprt/gtk-ng/class/window.zig
mitchellh: Sounds good. I've been trying not to fix bugs in the `-ng` side that legacy doesn't have but this seems small enough...

mitchellh: I actually reverted my change here because it was too dumb and caused a border to exist on my setup. I made #8127 to track.

PR #8129, opened 2025-08-02
gtk-ng: port options for ringing bell (system, audio, attention)

jcollie:


src/apprt/gtk-ng/class/surface.zig
```diff
@@ -525,7 +525,76 @@ pub const Surface = extern struct {
 
     /// Ring the bell.
     pub fn ringBell(self: *Self) void {
-        // TODO: Audio feature
+        const window = ext.getAncestor(Window, self.as(gtk.Widget)) orelse {
```
mitchellh: I wanted to do all this in the `Window` class by listening to the surface `ring-bell` signal. We should not be assuming any ancestor tree at all within `Surface`. The one place we do right now has a comment noting we should not do that (termio subprocess env).

mitchellh (changes requested):

src/apprt/gtk-ng/class/surface.zig
jcollie: Fixed. One note is that I did not have the `Window` listen directly to the signal from the `Surface` because we need to know which `Tab` the surface is attached to. That's basically the function of the `bell` signal on the `Tab`, basically it just forwards the signal but with the `Tab` attached.

jcollie:
Also added an implementation of the `title` bell feature.

src/apprt/gtk-ng/class/tab.zig
```diff
@@ -92,6 +92,21 @@ pub const Tab = extern struct {
     };
 
     pub const signals = struct {
+        /// The bell was rung by one of the child surfaces.
+        ///
+        /// The tab doesn't handle any functionality so it is up to the
+        /// embedding widget to react to this.
+        pub const bell = struct {
```
mitchellh: We don't need to do this. See how we attach signals to Surface from Window.

mitchellh (changes requested):
I wonder if the surface should handle the audio... I guess so... 

I also wonder if we should separate properties for titles a bit. 

Let me think about this. May just take this PR and go with it.

src/apprt/gtk-ng/class/tab.zig
jcollie: If you read my previous note this was done specifically because we need the tab that the surface is attached to. If the window attaches directly to the surface we have to look that information up elsewhere.

jcollie:
> I wonder if the surface should handle the audio... I guess so...

For the custom audio it really doesn't matter as it doesn't need any information that makes it necessary to have it one place or the other. So the reason I put it in the surface is because that's where it was and there wasn't a specific reason to move it.

> I also wonder if we should separate properties for titles a bit.

Not sure what needs separating.

src/apprt/gtk-ng/class/tab.zig
mitchellh: We should be able to bind to the "active surface". If I didn't hook that up yet then let me work on that. That'll come with splits but I thought I wired that property though already.

jcollie: It's not a question of where/when the signal is bound, it's when the signal is emitted. Routing the signal though the tab means we can get what tab the surface is attached to without searching through all the objects.

mitchellh: I don't understand. I'll take a look when I context load on this more.

jcollie: I've rewritten this to use actions, which are much cleaner (and should work with the new splits too).

jcollie:
Alright, did a big rewrite to use actions instead of signals.

jcollie:
Rebased on main with the new method for adding flair to the title.

src/apprt/gtk-ng/ext.zig
```diff
@@ -50,3 +51,73 @@ pub fn getAncestor(comptime T: type, widget: *gtk.Widget) ?*T {
     // We can assert the unwrap because getAncestor above
     return gobject.ext.cast(T, ancestor).?;
 }
+
+/// Check that an action name is valid.
+///
+/// Reimplementation of g_action_name_is_valid so that it can be
```
tristan957:
```suggestion
/// Reimplementation of g_action_name_is_valid() so that it can be
```

src/apprt/gtk-ng/ext.zig
```diff
@@ -50,3 +51,73 @@ pub fn getAncestor(comptime T: type, widget: *gtk.Widget) ?*T {
...
+/// use at comptime.
```
tristan957:
```suggestion
/// used at comptime.
```

tristan957 (approved):

mitchellh:
See #8221

PR #8137, opened 2025-08-04
macOS: properly handle buffer allocation in ZH locale canonicalization

Aaron-212:
Fix a bug introduced by #6885 

`fixZhLocale` returns locale string literal without copying it into the buffer, causing the `LANGUAGE` environment variable to be set incorrectly.
```log
$ LANG="" zig build run
...
debug(os_locale): setting LANGUAGE from preferred languages value=�����.UTF-8:�����.UTF-8:en_US.UTF-8:ja_CN.UTF-8
...
```

### DIFF
--- a/src/os/i18n.zig
+++ b/src/os/i18n.zig
@@ -133,7 +133,12 @@ pub fn canonicalizeLocale(
     locale: []const u8,
 ) error{NoSpaceLeft}![:0]const u8 {
     // Fix zh locales for macOS
-    if (fixZhLocale(locale)) |fixed| return fixed;
+    if (fixZhLocale(locale)) |fixed| {
+        if (buf.len < fixed.len + 1) return error.NoSpaceLeft;
+        @memcpy(buf[0..fixed.len], fixed);
+        buf[fixed.len] = 0;
+        return buf[0..fixed.len :0];
+    }
 
     // Buffer must be 16 or at least as long as the locale and null term
     if (buf.len < @max(16, locale.len + 1)) return error.NoSpaceLeft;


mitchellh:
The buffer is optional. We should be returning constant, static memory here. I don't understand how this fixes this issue. I believe you that you're seeing weird corruption but unsure how it's happening and want to understand that more. We shouldn't need to copy to the buffer.

Aaron-212:
It is caused by `fixZhLocale()`, this is the log after this patch:
```log
$ LANG="" zig build run
...
debug(os_locale): setting LANGUAGE from preferred languages value=zh_CN.UTF-8:zh_TW.UTF-8:en_US.UTF-8:ja_CN.UTF-8
...
```

And here is Claude Code's explanation:
`preferredLanguageFromCocoa()` uses a `FixedBufferStream` that expects `canonicalizeLocale()` to write results into the provided buffer. When `fixZhLocale()` returns static strings, it breaks this contract.

1. `canonicalizeLocale()` is called with `fbs.buffer[fbs.pos..]`
2. `fixZhLocale()` returns `"zh_CN"` (static memory, NOT in the buffer)
3. `fbs.seekBy(canon.len)` advances the buffer position by 5 bytes
4. But those 5 bytes in the buffer contain garbage because they were never written
5. `.UTF-8` gets appended after the garbage → `�����.UTF-8`

mitchellh (changes requested):
Ah okay, I see where we expect it to be written to the buffer. In that case, I think a few changes:

1. Keep `fixZhLocale` how it is, it _always_ returns static memory and that's a good contract.
2. Modify `canonicalizeLocale` to just add the result to the buffer.

mitchellh (approved):

Zxxxxxl:
Hi there 👋
I'm using Ghostty on macOS, and my system locale is already set to zh_CN.UTF-8. However, I can't get Chinese characters to display or input properly in the terminal.
I've tried checking the locale settings with locale command (the output shows LANG=zh_CN.UTF-8 and other related variables are correctly set), but the issue still persists.
Could you please give me some guidance on how to configure Ghostty to support Chinese? Any help would be greatly appreciated! Thanks a lot 😊

PR #8142, opened 2025-08-04
apprt/gtk-ng: global shortcuts

mitchellh:
Ports global shortcuts. 

This is mostly a direct logic copy. The primary difference is I converted `GlobalShortcuts` to a `GObject` which has a config and dbus property and emits a trigger signal. Importantly, it's no longer tied or dependent on the `gio.Application` in any way. The config and dbus connection are updated as normal properties.

Verified with Valgrind we're clean. Found one memory leak I ported back to legacy.


pluiedev (approved):

jcollie (approved):
LGTM

PR #8144, opened 2025-08-04
apprt/gtk-ng: surface context menu

mitchellh:
Port with changes:

* Utilizes the Surface blueprint for defining the `PopoverMenu`
* We can't attach it directly to the Overlay using blueprints because an overlay can only have a single child property and you can't see other children via Blueprint. To overcome this, use a `Box`
* Utilizing a `menu` signal the window can listen to to refresh its action map instead of digging into ancestor hierarchy.


pluiedev (approved):
:bober:

PR #8146, opened 2025-08-04
apprt/gtk-ng: action accelerators, clean up explicit error sets

mitchellh:
Not a lot here, ported the action accelerators which was small.

Besides that, cleaned up a bunch of explicit error sets which allowed us to remove some forwarded errors because they're so unlikely, and could unify others.


jcollie (approved):
LGTM

mitchellh:
Added an unrelated commit to fix some background opacity issues. Can see the commit message on it.

PR #8147, opened 2025-08-05
linux: interpret 'dark' window-theme as 'prefer_dark'

willnorris:
Linux allows applications to express four color schemes that related to dark mode: force-light, prefer-light, prefer-dark, force-dark. The system then has three schemes: prefer-light, default, prefer-dark. The schemes and their interaction is pretty well explored in the blog post at https://blogs.gnome.org/alicem/2021/10/04/dark-style-preference/.

This commit changes Ghostty's 'dark' window-theme to mean 'prefer-dark' instead of 'force-dark'. For most users, this should have no impact. GNOME (and likely other desktops) do not actually set the 'prefer-light' system scheme, so the result for both 'dark' and 'prefer-dark' is that Ghostty will use the configured dark theme. However, this change will allow a user to manually set 'prefer-light' as the system scheme to override the (now) 'prefer-dark' setting in Ghostty.

This does mean that there is now no window-theme which maps to 'force-dark' as the application scheme. Again, that should be fine for most users since common desktops never set 'prefer-light' anyway. But if force-dark is still desired, users can achieve that simply by setting a single dark theme, rather than separate light and dark themes.

Discussion: https://github.com/ghostty-org/ghostty/discussions/8134

### DIFF
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -334,7 +334,7 @@ pub fn init(self: *App, core_app: *CoreApp, opts: Options) !void {
                     .prefer_dark;
             },
             .system => .prefer_light,
-            .dark => .force_dark,
+            .dark => .prefer_dark,
             .light => .force_light,
         },
     );
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1591,6 +1591,7 @@ keybind: Keybinds = .{},
 ///   * `system` - Use the system theme.
 ///   * `light` - Use the light theme regardless of system theme.
 ///   * `dark` - Use the dark theme regardless of system theme.
+///     On Linux, prefer the dark theme, but still respect the system theme.
 ///   * `ghostty` - Use the background and foreground colors specified in the
 ///     Ghostty configuration. This is only supported on Linux builds.
 ///
--- a/src/apprt/gtk-ng/class/application.zig
+++ b/src/apprt/gtk-ng/class/application.zig
@@ -1006,7 +1006,7 @@ pub const Application = extern struct {
                     .prefer_dark;
             },
             .system => .prefer_light,
-            .dark => .force_dark,
+            .dark => .prefer_dark,
             .light => .force_light,
         });
 


00-kat (approved):

jcollie:
Meh, this seems like it'd just be better to go whole hog and add _all_ the options. macOS should interpret the prefer and force variants as just forcing a particular theme.

https://github.com/jcollie/ghostty/commit/9e7627418d763dd4c6f3dde2d62377cff9285e1e

pluiedev (changes requested):
If this change were to remain as is code-wise we definitely need to add a LOT more documentation explaining this. I'll try to come up with something but right now I'm not a fan

src/config/Config.zig
```diff
@@ -1591,6 +1591,7 @@ keybind: Keybinds = .{},
 ///   * `system` - Use the system theme.
 ///   * `light` - Use the light theme regardless of system theme.
 ///   * `dark` - Use the dark theme regardless of system theme.
+///     On Linux, prefer the dark theme, but still respect the system theme.
```
mitchellh, line 1594: I recommend changing this to a more detailed callback later in the docs here (probably a full paragraph) that carefully explains what this maps to, links to the gnome post you linked, etc.

src/apprt/gtk/App.zig
```diff
@@ -334,7 +334,7 @@ pub fn init(self: *App, core_app: *CoreApp, opts: Options) !void {
                     .prefer_dark;
             },
             .system => .prefer_light,
-            .dark => .force_dark,
+            .dark => .prefer_dark,
             .light => .force_light,
```
mitchellh, line 338: We need a comment here on why we don't `prefer_light`. I read the discussion so I get it, but future maintainers that pop into this code need more context.

mitchellh (changes requested):
I read the discussion and this looks good to me. Requesting a couple docs changes.

I'm surprised Gnome somehow managed to mindfuck light/dark mode (actually no, I'm not lol).

src/apprt/gtk/App.zig
willnorris: Actually, the more I think about it, the easier this is to reason about if this were to change to `prefer_light` as well.  Then there is a consistent story between both light and dark...

- If what you want is to always force a light or dark theme, you don't really need to use `window-theme` at all.  Just set the the `theme` option to only the singular preferred theme.
- The `dark` and `light` window-theme options are then both preferences for when the system doesn't specify a color scheme (with `light` being the same as `system`, since desktops seem to default to light anyway).  
  - So if the system setting is `prefer-dark`, you'll always get dark.  If the system setting is `prefer-light` (which it typically isn't in practice), you'll always get light.  If the system setting is `default`, you'll get whatever you specify here in `window-theme`.

Does that seem like a reasonable approach?

tristan957: Your explanation makes sense to me! I agree with Leah that we definitely want documentation around theming, so an accompanying PR to the website would be awesome.

willnorris: hmm... I think that might result in a behavior difference between Linux and macOS, since on macOS these do behave as a "force" light and dark I think, overriding the system setting?  I guess the changes as written get the right behavior, but are just harder to explain.

Maybe explicit `prefer-dark` and `prefer-light` options are a lot easier to understand, even if they only do anything on Linux.  For macOS, they would just be the same as `dark` and `light`.  (I don't think I'd go as far as adding `force-dark` and `force-light` options like @jcollie has linked above... that's truly redundant with the existing `dark` and `light`)

mitchellh: I think doing what is most expected by users is the right thing here. On macOS, we're able to override the theme on a per-app (actually per _widget/view_) basis so what we do is actually force only the Ghostty app to light/dark. 

I don't think users think in "prefer/force", I think they think in "light vs dark" mode. I think macOS/Windows got this right and Gnome got this wrong. So I think we should do what is expected and map accordingly.

pluiedev: Yeah I agree, GNOME is freaking weird. Again, my opinion would be to just preserve the status quo and treat "default" as light and dark as dark, since that's what the end user understands anyway. You have to mess with GSettings to even know that prefer-light is a thing. Then, if you really need to make Ghostty behave differently then you explicitly specify `window-theme` to be `light` or `dark`.

mitchellh:
I apologize, but I'm going to close this because there just isn't consensus on the right path forward yet. The. change as you've shown is fairly trivial, but it's unclear semantically what we really want. I'm sorry!

willnorris:
Sorry... I thought the main remaining task was to figure out how to explain it?  I've had this in my backlog to come back and take another stab at the documentation if you're still open to it.

tristan957:
@willnorris do you think you could start a discussion? This PR became too focused on trying to understand the problem.

A new discussion would help clear things up, I think!

tristan957:
Actually, let's continue discussing in the original discussion linked in the PR description

PR #8154, opened 2025-08-05
macOS: Add debug info to "About Ghostty"

udf2457:
I should start this PR with an honest clarification:

_I am not familiar with macOS, C or Zig development_.

What is being proposed is proposed with the assistance of Claude agentic development.  I should clearly state that like many I am cynical of these sorts of PRs, but I decided to post it here for three reasons:

1. This is a tight change, which does not propose largescale refactoring or introduction of major new features, it builds upon existing code.

2. I have had a quick look through the `git diff` and to my non-macOS dev eyes, it looks clean, reasonable and easy to follow.

3. I have run `zig build` on a macOS Sequoia 15.5.  `ghostty` compiles without error (*IMPORTANT*: See note [below](#important-note)) and I can open the binary and confirmed the new functionality works.

I understand this is a controversial PR so if you want to delete it, no hard feelings.  I just ask that you do it silently without bashing me about the use of Claude etc.

Ghostty1

Ghostty2

# Important note

Finally, regarding the **IMPORTANT** note:

As i said, `zig build` works with this PR on Sequoia 15.5, it worked because (in order to fix build errors) Claude also proposed to remove:

```diff
-    @available(macOS 26.0, *)
-    static var supportedModes: IntentModes = [.background, .foreground]
```

From the following files:

1. [macos/Sources/Features/App Intents/CloseTerminalIntent.swift](https://github.com/ghostty-org/ghostty/blob/70ec59d5663eb4d62df0431c1f71d7c64a6a3630/macos/Sources/Features/App%20Intents/CloseTerminalIntent.swift#L15-L16)
2. [macos/Sources/Features/App Intents/CommandPaletteIntent.swift](https://github.com/ghostty-org/ghostty/blob/70ec59d5663eb4d62df0431c1f71d7c64a6a3630/macos/Sources/Features/App%20Intents/CommandPaletteIntent.swift#L22-L23)
3. [macos/Sources/Features/App Intents/GetTerminalDetailsIntent.swift](https://github.com/ghostty-org/ghostty/blob/70ec59d5663eb4d62df0431c1f71d7c64a6a3630/macos/Sources/Features/App%20Intents/GetTerminalDetailsIntent.swift#L20-L21)
5. [macos/Sources/Features/App Intents/InputIntent.swift](https://github.com/ghostty-org/ghostty/blob/70ec59d5663eb4d62df0431c1f71d7c64a6a3630/macos/Sources/Features/App%20Intents/InputIntent.swift#L27-L28)
6. [macos/Sources/Features/App Intents/KeybindIntent.swift](https://github.com/ghostty-org/ghostty/blob/70ec59d5663eb4d62df0431c1f71d7c64a6a3630/macos/Sources/Features/App%20Intents/KeybindIntent.swift#L19-L20)
7. [macos/Sources/Features/App Intents/NewTerminalIntent.swift](https://github.com/ghostty-org/ghostty/blob/70ec59d5663eb4d62df0431c1f71d7c64a6a3630/macos/Sources/Features/App%20Intents/NewTerminalIntent.swift#L46-L47)
8. [macos/Sources/Features/App Intents/QuickTerminalIntent.swift](https://github.com/ghostty-org/ghostty/blob/70ec59d5663eb4d62df0431c1f71d7c64a6a3630/macos/Sources/Features/App%20Intents/QuickTerminalIntent.swift#L8-L9)

For, example:

```diff
diff --git a/macos/Sources/Features/App Intents/QuickTerminalIntent.swift b/macos/Sources/Features/App Intents/QuickTerminalIntent.swift
index 2e6c9850c..1936d3e0c 100644
--- a/macos/Sources/Features/App Intents/QuickTerminalIntent.swift
+++ b/macos/Sources/Features/App Intents/QuickTerminalIntent.swift
@@ -5,8 +5,6 @@ struct QuickTerminalIntent: AppIntent {
     static var title: LocalizedStringResource = "Open the Quick Terminal"
     static var description = IntentDescription("Open the Quick Terminal. If it is already open, then do nothing.")

-    @available(macOS 26.0, *)
-    static var supportedModes: IntentModes = .background

     @MainActor
     func perform() async throws -> some IntentResult & ReturnsValue<[TerminalEntity]> {
```

I decided ***NOT*** to include these changes in this PR.  Because I thought that you may well be looking to build against the upcoming macOS Tahoe. 

So if you try to build this PR on a 15.5 machine it will fail due to the above, as IntentModes API is only available in the Tahoe SDK, not in Seqoia SDK which is what my machine has.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -347,6 +347,23 @@ typedef struct {
 } ghostty_info_s;
 
 typedef struct {
+  const char* version;
+  uintptr_t version_len;
+  const char* zig_version;
+  uintptr_t zig_version_len;
+  const char* build_mode;
+  uintptr_t build_mode_len;
+  const char* app_runtime;
+  uintptr_t app_runtime_len;
+  const char* font_backend;
+  uintptr_t font_backend_len;
+  const char* renderer;
+  uintptr_t renderer_len;
+  const char* release_channel;
+  uintptr_t release_channel_len;
+} ghostty_build_info_s;
+
+typedef struct {
   const char* message;
 } ghostty_diagnostic_s;
 
@@ -859,6 +876,7 @@ typedef enum {
 int ghostty_init(uintptr_t, char**);
 void ghostty_cli_try_action(void);
 ghostty_info_s ghostty_info(void);
+ghostty_build_info_s ghostty_build_info(void);
 const char* ghostty_translate(const char*);
 void ghostty_string_free(ghostty_string_s);
 
--- a/macos/Sources/App/macOS/ghostty-bridging-header.h
+++ b/macos/Sources/App/macOS/ghostty-bridging-header.h
@@ -1,3 +1,4 @@
 // C imports here are exposed to Swift.
 
 #import "VibrantLayer.h"
+#import "ghostty.h"
--- a/macos/Sources/Features/About/AboutView.swift
+++ b/macos/Sources/Features/About/AboutView.swift
@@ -1,7 +1,11 @@
 import SwiftUI
+#if os(macOS)
+import AppKit
+#endif
 
 struct AboutView: View {
     @Environment(\.openURL) var openURL
+    @State private var showBuildDetails = false
 
     private let githubURL = URL(string: "https://github.com/ghostty-org/ghostty")
     private let docsURL = URL(string: "https://ghostty.org/docs")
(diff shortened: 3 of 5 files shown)


mitchellh:
Thanks. This information isn't particularly helpful for the macOS app because it doesn't have the same diversity of build configurations that Linux has. I think long term a "debug bundle" that people can attach to issues is useful, but would probably contain a lot more than just this. Thanks for taking a stab at this though.

udf2457:
Thanks @mitchellh .  Appreciate your reasoning.

Could I ask you, therefore, to consider amending the wording of the `Ghostty Version` of your `Issue Discussion` form, because this PR originated from the fact I was told I should not just provide the information available in the "About ghostty" box but actually `ghostty +version` because "it was different".

From what you are saying, therefore, "About ghostty" in its present form is indeed sufficient for macOS and macOS users do not need to use `ghostty +version` when starting issue discussions (unless they are using a non-standard version, but then in that event they are probably familiar with `ghostty +version` anyway).

Point being that for mac users, going to "about" is a very natural practice.

PR #8158, opened 2025-08-06
build: allow disabling i18n

pluiedev:
GNU gettext simply is a PITA on certain platforms (i.e. Windows, musl Linux, etc.) and currently it's not possible to cleanly remove i18n from the build process, making building Ghostty on the aforementioned platforms difficult. By providing users with a way to opt-out of the i18n mechanisms (or opt-in, on platforms where i18n is disabled by default) we can make sure that people at least have *some* way of building Ghostty before i18n mechanisms can be integrated neatly.


jcollie (approved):

mitchellh:
We should add a CI build to exercise this option.

PR #8164, opened 2025-08-06
font: disable discretionary ligatures by default

### LINKED ISSUE #5372: Font quirk: Broken “ます” ligature with BIZ UDGothic when discretionary ligatures are enabled
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5203

<sup>Originally posted by **masisz** January 18, 2025</sup>
Hello,

When I input the Japanese “ます” in ghostty, it is displayed as the symbol “〼”. Is there a way to fix this by changing the settings?

The issue occurred in an environment using the ~~JetBrains Mono~~[^1] font on macOS.

This is my first time submitting to an issue tracker, so if there’s any missing information or anything unclear in my description, please let me know, and I’ll provide additional details.

I’m looking forward to seeing this become an excellent terminal application.

Sincerely,</div>

[^1]: Editor's note: it is not actually using JetBrains Mono as it does not have any CJK characters. BIZ UDGothic is used instead.

pluiedev:
Closes #5372

Discretionary ligatures (denoted by the OpenType feature tag `dlig`) are sometimes used by programming fonts (e.g. Iosevka) to provide more "complex" and uncommon ligatures that may be useful in a programming context. Unfortunately, this has some nasty side effects with certain Japanese fallback fonts (#5372) due to perhaps a misaligned understanding of the OpenType spec[^spec].

The spec details that `dlig` ligatures should only be used to contract sequences of glyphs together into one glyph, and that it should be used only for "special effect", **at the user's preference** (emphasis mine). Indeed, it also suggests that:

> UI suggestion: This feature should be off by default.

All of this, combined with the fact that historical, nowadays unused and even unintelligible Kanji ligatures are explicitly included as examples of discretionary ligatures, shows that in the Japanese context at least that the "level of discretion" is significantly higher than what is found in programming fonts, where it is more understood to be "opinionated and uncommon", rather than "obsolete and unreadable".

Furthermore, it appears that a lot of common programming fonts don't even make use of the `dlig` feature —  JetBrains Mono, FiraCode and MonoLisa lack a `dlig` feature altogether, while Inconsolata seems to only use it for ligatures that are more commonly found in `liga` or `calt`, such as the `->` ligature. To a lot of people, then, this change would literally alter nothing.

Therefore, it's my opinion that we should disable `dlig` by default. It's arguably not being used correctly in the programming font space (or at least not in a way that's coherent with other fonts), and it only provides a marginal benefit while potentially rendering entire sentences in Japanese (and possibly other languages) unreadable out of the box.

If someone upgrades to tip or 1.2 and then asks "why aren't the ligatures working anymore", then at least they can always just turn on `dlig` by themselves.

[^spec]: https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-dlig

### DIFF
--- a/src/font/shaper/coretext.zig
+++ b/src/font/shaper/coretext.zig
@@ -1833,7 +1833,10 @@ fn testShaperWithFont(alloc: Allocator, font_req: TestFont) !TestShaper {
     grid_ptr.* = try .init(alloc, .{ .collection = c });
     errdefer grid_ptr.*.deinit(alloc);
 
-    var shaper = try Shaper.init(alloc, .{});
+    var shaper = try Shaper.init(alloc, .{
+        // Some of our tests rely on dlig being enabled by default
+        .features = &.{"dlig"},
+    });
     errdefer shaper.deinit();
 
     return TestShaper{
--- a/src/font/shaper/feature.zig
+++ b/src/font/shaper/feature.zig
@@ -287,7 +287,6 @@ pub const FeatureList = struct {
 /// These features are hardcoded to always be on by default. Users
 /// can turn them off by setting the features to "-liga" for example.
 pub const default_features = [_]Feature{
-    .{ .tag = "dlig".*, .value = 1 },
     .{ .tag = "liga".*, .value = 1 },
 };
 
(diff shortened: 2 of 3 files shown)


jcollie (approved):

00-kat (approved):

mitchellh (approved):
I think this is reasonable, let's send it and see if anyone complains (validly) on the other side. I've also considered adding default font features to our `quirks.zig` just as a zero-default for fonts where certain font features are preferred in most cases (even optional ones). We can cross that bridge another day.

PR #8165, opened 2025-08-06
apprt/gtk-ng: split tree foundations

mitchellh:
This begins to bring back splits to `gtk-ng`. As of this PR, **splits still don't work**, but the architectural underpinnings for them are all present. Namely, our tab view now embeds a `GhosttySplitTree` widget which has a full split tree present, and all the signals and active surface properties and so on are hooked up to the tree.

In theory, once we hook up split creation, close, etc. everything should _just work_.

But, this PR is already very large and I don't want to make it even larger, so I'm opening this PR to add the foundations for this while keeping `gtk-ng` in the state it more or less is on main right now.

The meat of this is in a pure Zig data structure `SplitTree` added to `src/datastruct`. This is a Ziggified port of our macOS implementation (but, much better if I do say so myself). Being in pure Zig lets us write unit tests easily, control allocations tightly, get our safety checks, etc. There is coverage in this PR. 

## Other Bugs Fixed

- Boxed accessors use the proper `g_boxed_copy/free` functions. Didn't really cause any issues because this is the first PR where we actually use custom implementations for that.

- `Surface` properly emits a notify event for focus change

### DIFF
--- a/src/apprt/gtk-ng/class/split_tree.zig
+++ b/src/apprt/gtk-ng/class/split_tree.zig
@@ -1,0 +1,157 @@
+const std = @import("std");
+const build_config = @import("../../../build_config.zig");
+const assert = std.debug.assert;
+const adw = @import("adw");
+const gio = @import("gio");
+const glib = @import("glib");
+const gobject = @import("gobject");
+const gtk = @import("gtk");
+
+const i18n = @import("../../../os/main.zig").i18n;
+const apprt = @import("../../../apprt.zig");
+const input = @import("../../../input.zig");
+const CoreSurface = @import("../../../Surface.zig");
+const gtk_version = @import("../gtk_version.zig");
+const adw_version = @import("../adw_version.zig");
+const gresource = @import("../build/gresource.zig");
+const Common = @import("../class.zig").Common;
+const Config = @import("config.zig").Config;
+const Application = @import("application.zig").Application;
+const CloseConfirmationDialog = @import("close_confirmation_dialog.zig").CloseConfirmationDialog;
+const Surface = @import("surface.zig").Surface;
+
+const log = std.log.scoped(.gtk_ghostty_split_tree);
+
+pub const SplitTree = extern struct {
+    const Self = @This();
+    parent_instance: Parent,
+    pub const Parent = adw.Bin;
+    pub const getGObjectType = gobject.ext.defineClass(Self, .{
+        .name = "GhosttySplitTree",
+        .instanceInit = &init,
+        .classInit = &Class.init,
+        .parent_class = &Class.parent,
+        .private = .{ .Type = Private, .offset = &Private.offset },
+    });
+
+    pub const properties = struct {
+        pub const @"is-empty" = struct {
+            pub const name = "is-empty";
+            const impl = gobject.ext.defineProperty(
+                name,
+                Self,
+                bool,
+                .{
+                    .nick = "Tree Is Empty",
+                    .blurb = "True when the tree has no surfaces.",
+                    .default = false,
+                    .accessor = gobject.ext.typedAccessor(
+                        Self,
+                        bool,
+                        .{
+                            .getter = getIsEmpty,
+                        },
+                    ),
+                },
+            );
+        };
+    };
+
+    const Private = struct {
+        /// The tree datastructure containing all of our surface views.
+        tree: Surface.Tree,
+
+        pub var offset: c_int = 0;
+    };
+
+    fn init(self: *Self, _: *Class) callconv(.c) void {
+        gtk.Widget.initTemplate(self.as(gtk.Widget));
+
+        // Start with an empty split tree.
+        const priv = self.private();
+        priv.tree = .empty;
+    }
+
+    //---------------------------------------------------------------
+    // Properties
+
+    pub fn getIsEmpty(self: *Self) bool {
+        const priv = self.private();
+        return priv.tree.isEmpty();
+    }
+
+    //---------------------------------------------------------------
+    // Virtual methods
+
+    fn dispose(self: *Self) callconv(.c) void {
+        gtk.Widget.disposeTemplate(
+            self.as(gtk.Widget),
+            getGObjectType(),
+        );
+
+        gobject.Object.virtual_methods.dispose.call(
+            Class.parent,
+            self.as(Parent),
+        );
+    }
+
+    fn finalize(self: *Self) callconv(.c) void {
+        const priv = self.private();
+        priv.tree.deinit();
+        priv.tree = .empty;
+
+        gobject.Object.virtual_methods.finalize.call(
+            Class.parent,
+            self.as(Parent),
+        );
+    }
+
+    //---------------------------------------------------------------
+    // Signal handlers
+
+    //---------------------------------------------------------------
+    // Class
+
+    const C = Common(Self, Private);
+    pub const as = C.as;
+    pub const ref = C.ref;
+    pub const unref = C.unref;
+    const private = C.private;
+
+    pub const Class = extern struct {
+        parent_class: Parent.Class,
+        var parent: *Parent.Class = undefined;
+        pub const Instance = Self;
+
+        fn init(class: *Class) callconv(.c) void {
+            gobject.ext.ensureType(Surface);
+            gtk.Widget.Class.setTemplateFromResource(
+                class.as(gtk.Widget.Class),
+                comptime gresource.blueprint(.{
+                    .major = 1,
+                    .minor = 5,
+                    .name = "split-tree",
+                }),
+            );
+
+            // Properties
+            gobject.ext.registerProperties(class, &.{
+                properties.@"is-empty".impl,
+            });
+
+            // Bindings
+
+            // Template Callbacks
+
+            // Signals
+
+            // Virtual methods
+            gobject.Object.virtual_methods.dispose.implement(class, &dispose);
+            gobject.Object.virtual_methods.finalize.implement(class, &finalize);
+        }
+
+        pub const as = C.Class.as;
+        pub const bindTemplateChildPrivate = C.Class.bindTemplateChildPrivate;
+        pub const bindTemplateCallback = C.Class.bindTemplateCallback;
+    };
+};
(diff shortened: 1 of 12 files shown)


src/apprt/gtk-ng/class/split_tree.zig
```diff
@@ -0,0 +1,290 @@
...
+        pub const @"tree-will-change" = struct {
+            pub const name = "tree-will-change";
+            pub const connect = impl.connect;
+            const impl = gobject.ext.defineSignal(
+                name,
+                Self,
+                &.{?*const Surface.Tree},
+                void,
+            );
+        };
```
tristan957: In GTK, there is precedent for [`changed` signals](https://docs.gtk.org/gtk4/index.html?q=%3A%3Achanged). Our signal could pass both the old and the new values. Would that suffice? Feels a bit cleaner in that way, but you made this PR, so you would have the best context.

src/apprt/gtk-ng/class/split_tree.zig
```diff
@@ -0,0 +1,290 @@
...
+                    .accessor = gobject.ext.typedAccessor(
+                        Self,
+                        bool,
+                        .{
+                            .getter = getIsEmpty,
+                        },
+                    ),
+                },
+            );
+        };
```
tristan957: Instead of `is-empty`, GObject APIs seem to be more designed around a property like `n-splits`:

- https://docs.gtk.org/gtk4/index.html?q=n-items
- https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/?q=n-pages

src/apprt/gtk-ng/class/tab.zig
```diff
@@ -153,29 +172,99 @@ pub const Tab = extern struct {
...
+            priv.split_tree.setTree(&tree);
+
+            // Hacky because we need a non-const result.
+            var mut = tree;
+            mut.deinit();
+        } else |_| {
+            // TODO: We should make our "no surfaces" state more aesthetically
+            // pleasing and show something like an "Oops, something went wrong"
+            // message. For now, this is incredibly unlikely.
+            @panic("oom");
```
tristan957, line 190: An [`AdwStatusPage`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.StatusPage.html) may be good here.

Make a tab page a `GtkStack` with one page being the status page. The other page being the split tree widget.

src/apprt/gtk-ng/class/window.zig
```diff
@@ -637,6 +656,102 @@ pub const Window = extern struct {
...
+                self,
+                .{},
+            );
+            _ = Surface.signals.@"toggle-command-palette".connect(
+                surface,
+                *Self,
+                surfaceToggleCommandPalette,
+                self,
+                .{},
+            );
```
tristan957, line 716: Why is this per surface and not on the window or application? I know the code was just moved.

src/apprt/gtk-ng/ui/1.5/split-tree.blp
```diff
@@ -0,0 +1,28 @@
+using Gtk 4.0;
+using Adw 1;
+
+template $GhosttySplitTree: Adw.Bin {
+  notify::tree => $notify_tree();
+
+  Box {
```
tristan957: I think it makes more sense for `GhosttySplitTree` to inherit from `GtkWidget`, and use a box layout manager:

```c
gtk_widget_class_set_layout_manager_type(klass, GTK_TYPE_BOX_LAYOUT);
```

It'll remove two widgets I think from the tree, but definitely at least 1. Was there a reason you went with subclassing `AdwBin`?

src/apprt/gtk-ng/ui/1.5/split-tree.blp
```diff
@@ -0,0 +1,28 @@
...
+    // ever happen at the time of writing this comment. A surface-less split
+    // tree always closes its parent.
+    Label {
+      visible: bind template.is-empty;
+      // Purposely not localized currently because this shouldn't really
+      // ever appear. When we have a situation it does appear, we may want
+      // to change the styling and text so I don't want to burden localizers
+      // to handle this yet.
+      label: "No surfaces.";
+    }
```
tristan957: An improvement for later: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.StatusPage.html

src/apprt/gtk-ng/class/split_tree.zig
tristan957: Having access to `n-splits` could enable some more UI design later about exposing number of surfaces in a tab for instance.

src/apprt/gtk-ng/class/window.zig
jcollie: It's a way for the surface to tell the window/application to do something without storing our own reference, or doing gross things like searching through every window to see which one is the right one.

tristan957: > It's a way for the surface to tell the window/application to do something without storing our own reference

Why does the surface need to tell the window/application what to do? The window/application can just do it.

tristan957: Is this for OSC commands or something?

jcollie: Because sometimes different parts of the stack have access to private information that's not readily accessible elsewhere, especially state that's implicit in the GTK widget tree. Basically the idea is that the surface has no (direct) knowledge of what window it's attached to. So it just emits a signal when something needs to happen on the window. The window then listens to all of the signals from all of the surfaces embedded within it.

Case in point, to toggle the command palette from a keybind (which happens in the surface) we need a window to attach the command palette dialog to. Since the surface doesn't know what window it's attached to, it emits a signal. The window gets that signal and can then present the dialog.

It's a little unusual, but that's the way that Mitchell wants things structured right now and it seems to work pretty well at the moment.

jcollie: > Is this for OSC commands or something?

And keybinds. Input processing is handled by the surface.

mitchellh: We need to know what surface is the target for the palette too. GTK currently doesn’t stick to this but macOS for example sticks to it heavily (you can focus other surfaces while the palette is up!)

tristan957: > Case in point, to toggle the command palette from a keybind (which happens in the surface) we need a window to attach the command palette dialog to. Since the surface doesn't know what window it's attached to, it emits a signal. The window gets that signal and can then present the dialog.

Why not just activate an action? [`gtk_widget_activate_action()`](https://docs.gtk.org/gtk4/method.Widget.activate_action.html). Seems like a waste to connect to every surface's signal. The surface can even send a reference to itself as a part of the action.

mitchellh: I think window actions for some of this makes sense.

jcollie: Actions have interesting implications for IPC as well because GTK "automagically" has machinery to trigger actions from DBus, which is how the `+new-window` CLI is implemented.

tristan957: > We need to know what surface is the target for the palette too.

I believe you can define what arguments the receiver of an action takes. At the very least it takes a GValue or something. Been a minute since I did custom actions.

jcollie: Yup, we do that already for `app.new-window-command` and `app.present-surface`:

https://github.com/ghostty-org/ghostty/blob/1d042f1e71f1533f92831c50610efa7a1fc1856e/src/apprt/gtk-ng/class/application.zig#L1105-L1144

src/apprt/gtk-ng/ui/1.5/split-tree.blp
pluiedev: AdwBin is designed to be a sane base class that provides most of the defaults you'd expect from an Adw widget — GtkWidget can do the same thing but with a *lot* more boilerplate

tristan957: What boilerplate do you think is necessary? The GtkBoxLayoutManager is supposed to handle all the boilerplate from my understanding. I have successfully used it without any boilerplate when thinking about subclassing GtkBox for instance.

src/apprt/gtk-ng/class/split_tree.zig
mitchellh: I saw that, but calculating `n` vs `empty` is computationally very different. In all reality for a split tree its probably not a big difference, but I think wasting CPU cycles for no reason doesn't make sense if we can help it. I think having both eventually makes sense.

src/apprt/gtk-ng/class/split_tree.zig
mitchellh: I saw that, but it seems to be past tense (DID change vs WILL change) so I didn't want to use the `changed` naming if someone more familiar expected that. I suppose as long as we defer freeing the old value until after then it'd be fine though... so let me make the change and give it a shot.

mitchellh: Note on macOS the difference of the "will change" notifications is that listeners can halt the change. I don't need that functionality here, but just noting.

src/apprt/gtk-ng/class/window.zig
mitchellh: A good note for the future to cleanup.

src/apprt/gtk-ng/ui/1.5/split-tree.blp
mitchellh: I think a box in this case might make sense, because it has multiple potential children.

I prefer a `adw.Bin` when I have exactly one child, because `set_child` is a much nicer API to clear prior children than wrangling the multiple children approach that a GTK widget has. And it seems like it was made for exactly that: https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-bin.c

In this case, it would get rid of 1 widget (I think only one, the directly nested `Box`).

src/apprt/gtk-ng/class/split_tree.zig
mitchellh: It appears the better precedent here is `has-*` properties across GTK widgets. I'll change to that.

src/apprt/gtk-ng/class/tab.zig
mitchellh: I like that. Let's do that in a future PR, its a nice simple change.

PR #8166, opened 2025-08-06
gtk-ng: add "title bar styles"

jcollie:
This PR adds a "tabs" title bar style similar to the macOS title bar style. When `gtk-titlebar-style=tabs` the title bar and the tab bar will be merged together.

The config entry for controlling this is kept separate from macOS as macOS has more styles defined that don't map to a GTK title bar style and it's likely that users that use both macOS and GTK would want different settings for each platform.

Screenshot From 2025-08-06 16-38-28

### DIFF
--- a/src/apprt/gtk-ng/class/window.zig
+++ b/src/apprt/gtk-ng/class/window.zig
@@ -906,9 +906,7 @@ pub const Window = extern struct {
             .tabs => {
                 // *Conditionally* disable the tab bar when maximized, the titlebar
                 // style is tabs, and gtk-titlebar-hide-when-maximized is set.
-                if (self.isMaximized() and config.@"gtk-titlebar-hide-when-maximized") {
-                    return false;
-                }
+                if (self.isMaximized() and config.@"gtk-titlebar-hide-when-maximized") return false;
 
                 // If the titlebar style is tabs the tab bar must always be visible.
                 return true;
--- a/src/apprt/gtk-ng/class/window.zig
+++ b/src/apprt/gtk-ng/class/window.zig
@@ -11,6 +11,7 @@ const gtk = @import("gtk");
 const i18n = @import("../../../os/main.zig").i18n;
 const apprt = @import("../../../apprt.zig");
 const configpkg = @import("../../../config.zig");
+const TitlebarStyle = configpkg.Config.GtkTitlebarStyle;
 const input = @import("../../../input.zig");
 const CoreSurface = @import("../../../Surface.zig");
 const ext = @import("../ext.zig");
@@ -92,6 +93,25 @@ pub const Window = extern struct {
                             }
                         }.getter,
                     }),
+                },
+            );
+        };
+
+        pub const @"titlebar-style" = struct {
+            pub const name = "titlebar-style";
+            const impl = gobject.ext.defineProperty(
+                name,
+                Self,
+                TitlebarStyle,
+                .{
+                    .default = .native,
+                    .accessor = gobject.ext.typedAccessor(
+                        Self,
+                        TitlebarStyle,
+                        .{
+                            .getter = Self.getTitlebarStyle,
+                        },
+                    ),
                 },
             );
         };
@@ -548,6 +568,7 @@ pub const Window = extern struct {
             "tabs-visible",
             "tabs-wide",
             "toolbar-style",
+            "titlebar-style",
         }) |key| {
             self.as(gobject.Object).notifyByPspec(
                 @field(properties, key).impl.param_spec,
@@ -814,6 +835,14 @@ pub const Window = extern struct {
         return false;
     }
 
+    fn isFullscreen(self: *Window) bool {
+        return self.as(gtk.Window).isFullscreen() != 0;
+    }
+
+    fn isMaximized(self: *Window) bool {
+        return self.as(gtk.Window).isMaximized() != 0;
+    }
+
     fn getHeaderbarVisible(self: *Self) bool {
         const priv = self.private();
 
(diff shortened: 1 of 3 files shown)


tristan957:
can we unify `macos-titlebar-style` and `gtk-titlebar-style`?

src/apprt/gtk-ng/class/window.zig
```diff
@@ -101,6 +101,40 @@ pub const Window = extern struct {
...
+                bool,
+                .{
+                    .nick = "Titlebar style is tabs",
+                    .blurb = "True if the titlebar style is tabs",
+                    .default = true,
+                    .accessor = gobject.ext.typedAccessor(Self, bool, .{
+                        .getter = Self.getTitlebarStyleTabs,
+                    }),
+                },
+            );
```
tristan957: Can't we use a GEnum here? That's what I would do in C. I'm not sure what zig-gobject pattern in though.

pluiedev: This is mainly for the Blueprint, but I think it makes a lot more sense if the Blueprint calls a closure instead (`bind template.is-titlebar-style-tabs()`)

jcollie:
> can we unify `macos-titlebar-style` and `gtk-titlebar-style`?

IMHO, no. The default for `macos-titlebar-style` is `transparent`, which is basically equivalent to `window-theme=ghostty` on GTK. Then there is the macOS style "hidden", which is kinda equivalent to `gtk-titlebar=false` (but there's going to be interactions with `window-decoration` and client side/server side decorations).

We'd need to rethink how all of these interact to give people the effect that they are looking for.

src/apprt/gtk-ng/class/window.zig
jcollie: The purpose of these properties was for binding to properties in the window blueprint. I've not found any evidence that you can do comparisons to enums in the blueprint (even boolean logic seems out of scope for what blueprints were designed for).

mitchellh (changes requested):
Just a red review so I can review before merge. Thanks

tristan957:
> > can we unify `macos-titlebar-style` and `gtk-titlebar-style`?
> 
> IMHO, no. The default for `macos-titlebar-style` is `transparent`, which is basically equivalent to `window-theme=ghostty` on GTK. Then there is the macOS style "hidden", which is kinda equivalent to `gtk-titlebar=false` (but there's going to be interactions with `window-decoration` and client side/server side decorations).
> 
> We'd need to rethink how all of these interact to give people the effect that they are looking for.

Maybe something to rethink for 1.2 and just slide on this.

src/apprt/gtk-ng/class/window.zig
tristan957: If using a blueprint bind leads us to writing less than optimal code (my opinion), I would not use the blueprint bind. I don't think we are saving any lines of code with this.

mitchellh: Agree for now. I think the proper way to do this is via an exposed function as @pluiedev said. I never got the exact function signature (on the Zig side) figured out for that so I just crashed out. I have to spelunk the source to figure that out. But doing something like what @pluiedev said would be ideal.

src/apprt/gtk-ng/class/window.zig
```diff
@@ -733,20 +769,27 @@ pub const Window = extern struct {
...
+        // *Conditionally* disable the header bar when maximized, and
+        // gtk-titlebar-hide-when-maximized is set
         if (self.as(gtk.Window).isMaximized() != 0 and
             config.@"gtk-titlebar-hide-when-maximized")
         {
             return false;
         }
 
+        // If the titlebar style is tabs never show the titlebar.
+        if (config.@"gtk-titlebar-style" == .tabs) return false;
```
mitchellh: I prefer a switch here (even if most prongs are `{}`) so the compiler can yell at us when we introduce new options so we can think about it.

src/apprt/gtk-ng/class/window.zig
```diff
@@ -733,20 +769,27 @@ pub const Window = extern struct {
...
+
         return config.@"gtk-titlebar";
     }
 
     fn getTabsAutohide(self: *Self) bool {
         const priv = self.private();
         const config = if (priv.config) |v| v.get() else return true;
+
+        // If the titlebar style is tabs we cannot autohide.
+        if (config.@"gtk-titlebar-style" == .tabs) return false;
```
mitchellh: Same here, switch.

src/apprt/gtk-ng/class/window.zig
```diff
@@ -763,6 +806,19 @@ pub const Window = extern struct {
     fn getTabsVisible(self: *Self) bool {
         const priv = self.private();
         const config = if (priv.config) |v| v.get() else return true;
+
+        // *Conditionally* disable the tab bar when maximized, the titlebar
+        // style is tabs, and gtk-titlebar-hide-when-maximized is set.
+        if (self.as(gtk.Window).isMaximized() != 0 and
+            config.@"gtk-titlebar-style" == .tabs and
```
mitchellh: Same here, switch around the other parts.

mitchellh (changes requested):
Okay, got a real review in. Clean!

mitchellh:
> can we unify macos-titlebar-style and gtk-titlebar-style?

I agree with jcollie: no... for now at least. I think GUI systems across platforms are distinct enough that unifying _styling_ configuration is going to be a challenge in addition to I think probably just not being desirable. I think the configuration quirks per platform are too vast (there'd be too many "asterisks" on each option) and stylistically it doesn't make sense to support every style on every platform because they just culturally are _different_ in style.

src/apprt/gtk-ng/class/window.zig
```diff
@@ -101,6 +102,27 @@ pub const Window = extern struct {
...
         };
 
+        pub const @"titlebar-style" = struct {
+            pub const name = "titlebar-style";
+            const impl = gobject.ext.defineProperty(
+                name,
+                Self,
+                TitlebarStyle,
+                .{
+                    .nick = "Titlebar style",
```
tristan957: nicks are usually titlecased

tristan957: aside: in GTK applications, we can use NULL values for nick and blurb

tristan957 (approved):

src/apprt/gtk-ng/class/window.zig
jcollie: Done.

src/apprt/gtk-ng/class/window.zig
jcollie: Done.

src/apprt/gtk-ng/class/window.zig
jcollie: Done.

src/apprt/gtk-ng/class/window.zig
jcollie: Figured out how to set up a closure so have switched to that.

src/apprt/gtk-ng/class/window.zig
jcollie: Fixed. I kinda like putting them there as a kind of self-documentation. GTK doesn't seem to use it anywhere, not even the inspector.

tristan957: I'm not even sure what they are used for at all to be honest

mitchellh: Oh, didn't realize these were optional. Yeah fuck these let's delete all of them. lol.

tristan957: I'm not even sure what they are used for at all to be honest

jcollie: Nuked.

src/apprt/gtk-ng/class/window.zig
```diff
@@ -825,46 +854,72 @@ pub const Window = extern struct {
...
-        return switch (config.@"window-show-tab-bar") {
-            .always, .auto => true,
-            .never => false,
-        };
+
+        switch (config.@"gtk-titlebar-style") {
+            .tabs => {
+                // *Conditionally* disable the tab bar when maximized, the titlebar
+                // style is tabs, and gtk-titlebar-hide-when-maximized is set.
+                if (self.isMaximized() and config.@"gtk-titlebar-hide-when-maximized") {
```
mitchellh: Can be simplified to a single expression rather than an `if` block.

mitchellh (changes requested):
One tiny request

src/apprt/gtk-ng/class/window.zig

mitchellh (approved):
Verified Valgrind on my end

PR #8180, opened 2025-08-08
fix: capture screenshot for app intents views as NSImage

peanball:
SwiftUI's ImageRenderer must not be called outside the main thread.

The `@MainActor` annotation is only relevant for our own code, not
for calls from frameworks. The machinations around Shortcuts end up
calling the displayRepresentation method outside the main thread.

By capturing the screenshot as NSImage, all data is retained and can
be processed outside the main thread.

Thread in Discord: https://discord.com/channels/1005603569187160125/1403384231694172372


mitchellh (approved):

mitchellh:
Thanks!

PR #8182, opened 2025-08-08
gtk-ng: don't use signals to toggle command palette

jcollie:


jcollie:
For some reason I'm getting memory leaks when using the command palette that others have been unable to reproduce so there's maybe something with my particular system that is messed up. So I'm not super confident that I haven't introduced a memory leak with this patch.

src/apprt/gtk-ng/class/command_palette.zig
```diff
@@ -216,6 +217,10 @@ pub const CommandPalette = extern struct {
     /// Helper function to send a signal containing the action that should be
     /// performed.
     fn activated(self: *CommandPalette, pos: c_uint) void {
+        // add a reference to keep ourselves around until we're done
+        _ = self.ref();
+        defer self.unref();
```
mitchellh: We should be able to extract the `cmd` from the list model before calling `close` (but also before signal still) so we don't have to do this, because our `cmd` object is still referenced.

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -560,13 +547,7 @@ pub const Surface = extern struct {
...
 
     pub fn toggleCommandPalette(self: *Self) bool {
-        signals.@"toggle-command-palette".impl.emit(
-            self,
-            null,
-            .{},
-            null,
-        );
-        return true;
+        return self.as(gtk.Widget).activateAction("win.toggle-command-palette", null) != 0;
```
mitchellh, line 555: Add a comment here that in the future we'll pass the target surface as a param. No need yet.

mitchellh (changes requested):
One small change requested, and I want to run it through Valgrind after, but looks good!

src/apprt/gtk-ng/class/command_palette.zig
jcollie: Fixed.

src/apprt/gtk-ng/class/surface.zig
jcollie: Fixed.

tristan957 (approved):

mitchellh (changes requested):
I pulled this down, showed the palette, and closed the window, and I got tons of Zig leaks.

1. Open Ghostty
2. Open palette
3. Close palette with `esc`
4. Close window (exit or menubar)

jcollie:
> I pulled this down, showed the palette, and closed the window, and I got tons of Zig leaks.

AHA I'M NOT CRAZY!!!

I've spent several hours trying to debug that leak and I can't figure out why that happens.

mitchellh:
Yeah, it doesn't happen on main for me. So its something in this branch (or a heisenbug). I don't have time to dig in yet.

jcollie:
> Yeah, it doesn't happen on main for me. So its something in this branch (or a heisenbug). I don't have time to dig in yet.

What if you open the command palette, and then close it with the "X"?

pluiedev (approved):

pluiedev:
Just the code by itself looks good (can't test Valgrind stuff yet)

jcollie:
> Just the code by itself looks good (can't test Valgrind stuff yet)

Yeah, I'm still tracking down leaks when you close the command palette with "X". It's present in main too AFAICT.

jcollie:
And I think that I've tracked down the memory leaks!

src/apprt/gtk-ng/weak_ref.zig
```diff
@@ -10,6 +10,8 @@ pub fn WeakRef(comptime T: type) type {
 
         ref: gobject.WeakRef = std.mem.zeroes(gobject.WeakRef),
 
+        pub const init: Self = .{};
```
mitchellh: Is this idiomatic? I know the new `empty` const idioms but I haven't seen `init` as a const used. Just curious... I'm fine naming this `.empty` though.

mitchellh (commented):
Going to run it later today to test leaks then we should be good.

src/apprt/gtk-ng/weak_ref.zig
jcollie: Switched to .empty

mitchellh (approved):

PR #8184, opened 2025-08-08
apprt/gtk-ng: add proper setters for surface properties

mitchellh:
This also fixes a bug where we were setting custom cursors on the wrong gtk widget, this showed up most terribly with `mouse-hide-while-typing` where the mouse would never reappear.


mitchellh:
I still want to do a cleanup where our accessors always call `notify` but that's a bit beyond the bugs I was fixing here.

PR #8187, opened 2025-08-08
fix: ensure copy menu appears regardless of selection direction

wallentx:
The copy menu on macOS was only appearing when selecting text from end to start, but not when selecting from start to end. This was due to the offset calculation using raw selection pins instead of forward-ordered pins.

- Use forward-ordered selection for offset calculations in dumpTextLocked()
- Apply same fix to selectionInfo() for consistency
- Remove unused variables after refactoring

Fixes the issue where right-click copy menu wouldn't appear when selecting text by starting from the beginning and dragging to the end.

[ghosttyfix.webm](https://github.com/user-attachments/assets/fe6e6c6b-7823-4115-bd08-3395a6c2f0ee)

I opened a discussion here https://github.com/ghostty-org/ghostty/discussions/8188


src/Surface.zig
```diff
@@ -1524,28 +1524,19 @@ fn recomputeInitialSize(
 /// Represents text read from the terminal and some metadata about it
 /// that is often useful to apprts.
 pub const Text = struct {
-    /// The text that was read from the terminal.
```
mitchellh, line 1527: Please don't remove doc comments on fields. I know they're redundant sometimes but things like LSP use this on a per-field basis.

src/Surface.zig
```diff
@@ -1524,28 +1524,19 @@ fn recomputeInitialSize(
...
-
-    /// The viewport information about this text, if it is visible in
-    /// the viewport.
-    ///
-    /// NOTE(mitchellh): This will only be non-null currently if the entirety
-    /// of the selection is contained within the viewport. We don't have a
-    /// use case currently for partial bounds but we should support this
-    /// eventually.
-    viewport: ?Viewport = null,
+    offset_start: u32,
```
mitchellh, line 1533: Need doc comments.

mitchellh (changes requested):
In general looks good, but you removed a lot of comments. Please add them back unless there was a good reason to remove them. Thanks for chasing this down.

wallentx:
@mitchellh Thanks, I've went back and cleaned up some stuff and made sure comments remained, or were updated to be applicable. I built and tested the app locally to ensure that it all still worked as intended.

mitchellh:
Replaced by #8288. There was some weird logic in here I couldn't explain, and the core premise of the topleft/bottomright being wrong was incorrect.

PR #8200, opened 2025-08-10
core: add a UUID to every core surface

jcollie:
- Adds a 128bit UUID to every core surface.
- Expose that UUID as an environment variable.
- Add a function to the core app to search for surfaces by UUID.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -52,6 +52,11 @@ const Renderer = rendererpkg.Renderer;
 const min_window_width_cells: u32 = 10;
 const min_window_height_cells: u32 = 4;
 
+/// Unique ID used to identify this surface for IPC purposes. It is
+/// exposed to the commands running in surfaces as the environment variable
+/// GHOSTTY_SURFACE.
+id: u64,
+
 /// Allocator
 alloc: Allocator,
 
@@ -532,6 +537,7 @@ pub fn init(
     errdefer io_thread.deinit();
 
     self.* = .{
+        .id = std.crypto.random.int(u64),
         .alloc = alloc,
         .app = app,
         .rt_app = rt_app,
@@ -577,6 +583,12 @@ pub fn init(
                 std.process.EnvMap.init(alloc);
         };
         errdefer env.deinit();
+
+        var buf: [18]u8 = undefined;
+        try env.put(
+            "GHOSTTY_SURFACE",
+            std.fmt.bufPrint(&buf, "0x{x:0>16}", .{self.id}) catch unreachable,
+        );
 
         // Initialize our IO backend
         var io_exec = try termio.Exec.init(alloc, .{
--- a/src/App.zig
+++ b/src/App.zig
@@ -510,6 +510,16 @@ fn hasSurface(self: *const App, surface: *const Surface) bool {
     return false;
 }
 
+/// Search for a surface by a 64 bit unique ID.
+pub fn findSurfaceByID(self: *const App, id: u64) ?*Surface {
+    for (self.surfaces.items) |v| {
+        const surface: *Surface = v.core();
+        if (surface.id == id) return surface;
+    }
+
+    return null;
+}
+
 fn hasRtSurface(self: *const App, surface: *apprt.Surface) bool {
     for (self.surfaces.items) |v| {
         if (v == surface) return true;


flatpak/zig-packages.json
```diff
@@ -131,6 +131,12 @@
     "dest": "vendor/p/N-V-__8AAHffAgDU0YQmynL8K35WzkcnMUmBVQHQ0jlcKpjH",
     "sha256": "ffc668a310e77607d393f3c18b32715f223da1eac4c4d6e0579a11df8e6b59cf"
   },
+  {
+    "type": "archive",
+    "url": "https://github.com/r4gus/uuid-zig/archive/refs/tags/0.3.2.tar.gz",
```
mitchellh: We definitely don't need a dependency for this. The implementation of v4 UUIDs is so small we should just copy this into our source tree (and give proper attribution).

https://github.com/r4gus/uuid-zig/blob/master/src/v4.zig

(I know we need the serialization stuff too).

src/Surface.zig
```diff
@@ -51,6 +52,8 @@ const Renderer = rendererpkg.Renderer;
 const min_window_width_cells: u32 = 10;
 const min_window_height_cells: u32 = 4;
 
+id: u128,
```
mitchellh: Let's comment to explain this.

mitchellh (changes requested):
Just a change requested on the dependency.

Other than that, I'm fine with this as it is. I think we'll take a much harder look at surface IDs in the future when we also make libghostty aware of hierarchy (windows vs tabs vs splits) but for now this is fine.

flatpak/zig-packages.json
jcollie: Vendored the library.

src/Surface.zig

mitchellh (changes requested):
I don't want to vendor the UUID. I just want to make our own UUID type that copy and pastes what we need. I can take this over.

jcollie:
I went ahead and implemented our own UUID type.

pluiedev:
One downside to an UUID is that you can't fit it inside a glib.Variant by itself (the type system only supports 64-bit integers at most), which means that when you want to pass the surface ID through our GTK actions, you either have to stuff it inside a string (ew, bad) or split it up into two u64s. Neither option sounds enticing, really...

pluiedev:
Plus: u128s are also ill-defined at best in C land (just search "__int128" and the horrors therein), so *every time* you want to refer to a surface via libghostty you have to specify two u64s (or even worse, a string). Why not just use one single u64 in this case?

mitchellh:
Yeah we may be able to get away with an incrementing u64. Surface creation can grab a lock to increment and you’d have to have an unfathomable amount of surfaces to ever overflow. 

I’ve been meaning to look at that as an option.

jcollie:
> Yeah we may be able to get away with an incrementing u64. Surface creation can grab a lock to increment and you’d have to have an unfathomable amount of surfaces to ever overflow.

Incrementing a u64 would def be my _least_ preferred option. So many network protocols have had security issues due to easily predictable identifiers that I just automatically reject that as an option. Plus the need for a mutex.

If we aren't going for "real" UUIDs then `std.crypto.random.int(u64)` is the best way to go IMHO.

mitchellh:
Yeah sorry saying incrementing was definitely wrong. I mean something closer to Snowflake.

jcollie:
If UUIDs are overkill because of their size, Snowflake IDs are overkill because of their complexity. Ghostty's not really a distributed system so all of that work constructing that ID is kinda pointless.

mitchellh:
It’s roughly the same complexity as a UUID, some bits for time, some bits incremented, some bits random. Just 64 instead of 128 bits.

jcollie:
Switched to a plain 'ol random u64.

src/Surface.zig
```diff
@@ -547,6 +553,12 @@ pub fn init(
         };
         errdefer env.deinit();
 
+        var buf: [16]u8 = undefined;
+        try env.put(
+            "GHOSTTY_SURFACE",
+            std.fmt.bufPrint(&buf, "{x:0>16}", .{self.id}) catch unreachable,
```
pluiedev: Hmm... thinking about this, for scripting purposes I think it *might* be actually easier to output a decimal integer instead, mostly because the GVariant format syntax that users would use to call GTK actions requires some extra finagling if we just output an unprefixed hex number. Not a blocker though

pluiedev (approved):

src/Surface.zig
jcollie: Added a `0x` prefix. I like keeping it in hexadecimal because then it's a fixed-length string. Pretty much anything should be able to deal with that, including the `gdbus` CLI.

mitchellh:
Going to wait on this until we need it.

PR #8202, opened 2025-08-10
apprt/gtk-ng: create/close split functionality

mitchellh:
This adds on to our existing foundations from #8165 and adds the ability to create and close splits. We're still missing split navigation, resizing via keybindings, etc. And there are a number of known issues (listed below). But this is a strict improvement from where we're at and includes a number of important bug fixes to our split tree.

The only nasty thing in this PR is that I learned that GTK _did not like_ rebuilding our split widget tree on every data model change. I don't know enough about how all the re-parenting plus size allocation interactions work together. As a compromise, this PR adds a listener, waits for our surface tree to "settle" by having all surfaces have no parents, then schedules a single rebuild after that. This works well, but results in some noticeable flashing for a frame or so. I think we can improve this later, it works completely well enough.

Importantly, all of this is Valgrind clean. I long suspected our splits on legacy are NOT free of leaks, but never proved it, so this makes me happy.

## Demo

https://github.com/user-attachments/assets/e231d89f-581e-486b-ade0-1d7e6795262e

## Known Issues

I may fix this in this PR, I may follow up.

- #8208
- [x] Focus doesn't go to the right place after closing a split (#8210)
- [x] Divider with a transparent background is transparent
- [x] Close split doesn't show any close confirmation dialog
- Missing features:
  * [x] Equalize splits (#8211)
  * [ ] Resize splits keybind (manual mouse action works fine)
  * [x] Go to split keybind (#8210)

### DIFF
--- a/src/apprt/gtk-ng/class/split_tree.zig
+++ b/src/apprt/gtk-ng/class/split_tree.zig
@@ -17,6 +17,7 @@ const adw_version = @import("../adw_version.zig");
 const ext = @import("../ext.zig");
 const gresource = @import("../build/gresource.zig");
 const Common = @import("../class.zig").Common;
+const WeakRef = @import("../weak_ref.zig").WeakRef;
 const Config = @import("config.zig").Config;
 const Application = @import("application.zig").Application;
 const CloseConfirmationDialog = @import("close_confirmation_dialog.zig").CloseConfirmationDialog;
@@ -118,6 +119,10 @@ pub const SplitTree = extern struct {
         // Template bindings
         tree_bin: *adw.Bin,
 
+        /// Last focused surface in the tree. We need this to handle various
+        /// tree change states.
+        last_focused: WeakRef(Surface) = .{},
+
         /// The source that we use to rebuild the tree. This is also
         /// used to debounce updates.
         rebuild_source: ?c_uint = null,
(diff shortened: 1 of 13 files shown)


src/apprt/gtk-ng/class/split_tree.zig
```diff
@@ -93,16 +117,237 @@ pub const SplitTree = extern struct {
...
+        //
+        // For action names:
+        // https://docs.gtk.org/gio/type_func.Action.name_is_valid.html
+        const actions = .{
+            // All of these will eventually take a target surface parameter.
+            // For now all our targets originate from the focused surface.
+            .{ "new-left", actionNewLeft, null },
+            .{ "new-right", actionNewRight, null },
+            .{ "new-up", actionNewUp, null },
+            .{ "new-down", actionNewDown, null },
```
mitchellh, line 165: A future TODO is to change this to take a direction as a parameter, but I'm not super familiar with how to do that AND bind it in a blueprint file so I'm going to defer this in the interest of getting things done (and doing things the way we always have in the mean time).

pluiedev:
Ouch that looks like a *lot* of flashing... definitely needs to be ironed out before we make -ng the default

mitchellh:
> Ouch that looks like a lot of flashing... definitely needs to be ironed out before we make -ng the default

As a heavy user of splits, it's not that bad. It has to be fixed, but I don't think it needs to block making `-ng` the default, because the correctness and maintainability of the whole new apprt outweighs this minor UX issue. The reason its not that bad is because splits aren't usually created/destroyed that often. 

So you only see the flash momentarily, then you work for a long time without issue. The demo here is sort of a worst case and unrealistic scenario. :smile: 

Its also a lot better in ReleaseFast builds. Like I said, it still needs to be improved, but not a fire.

I have some ideas to improve it.

mitchellh:
#8208 so we don't forget the flicker.

PR #8204, opened 2025-08-10
font/freetype: convert encoding of font names

jcollie:
Freetype encodes some font names internally in formats other than UTF-8. This only affects debug logs but it was annoying me so I fixed it. There may be other encodings that might need to be dealt with but I took care of the one that I ran across.


qwerasd205 (approved):
LGTM

mitchellh (approved):
It’d be nice to have a font that we could embed to test this but it’s simple enough to also just accept

qwerasd205:
I think a test could be added using the embedded jb mono variable font, since that's where the issue was first noticed so I think it does have the UTF-16 name

jcollie:
Added a test.

PR #8206, opened 2025-08-11
Fix up font raster position + other small fixes

qwerasd205:
I've cleaned up the code we use for scaling and positioning glyphs for raster, under both CoreText and FreeType. Before we had some imprecision, and under CoreText we were sometimes stretching glyphs in unseemly ways. These changes make it so that our constraints can position and size glyphs *exactly* and we don't have any chopped-off row/column issues for CoreText. With this, PowerLine Extra symbols now always align *perfectly* with the cell height:
||Before|After|
|-:|-|-|
|**CoreText**|image|image|
|**FreeType**|image|image|

The other changes are mainly just cleanup stuff, though one of the changes makes it so that we do once again properly apply constraints to symbols from the dingbats block (it was a regression, noted in #7955, that we stopped doing that).

### Future work
This has been a problem since we introduced the custom constraints, but I noticed it while preparing the before/after images: the left-edge PLE symbols (meant to connect to a full block on the right) expand out to the *right*, so if they're followed immediately by another character than they actually get squished and don't match the right-edge symbols:

image

I have a WIP change to move constraint logic to the shapers, and at that point we can maybe do something to allow the constraint to grow in to whitespace on the left side instead of on the right side.

### DIFF
--- a/src/font/face.zig
+++ b/src/font/face.zig
@@ -222,6 +222,16 @@ pub const RenderOptions = struct {
             y: f64,
         };
 
+        /// Returns true if the constraint does anything. If it doesn't,
+        /// because it neither sizes nor positions the glyph, then this
+        /// returns false.
+        pub inline fn doesAnything(self: Constraint) bool {
+            return self.size_horizontal != .none or
+                self.align_horizontal != .none or
+                self.size_vertical != .none or
+                self.align_vertical != .none;
+        }
+
         /// Apply this constraint to the provided glyph
         /// size, given the available width and height.
         pub fn constrain(
--- a/src/font/Metrics.zig
+++ b/src/font/Metrics.zig
@@ -38,6 +38,10 @@ cursor_height: u32,
 /// The constraint height for nerd fonts icons.
 icon_height: u32,
 
+/// Original cell width in pixels. This is used to keep
+/// glyphs centered if the cell width is adjusted wider.
+original_cell_width: ?u32 = null,
+
 /// Minimum acceptable values for some fields to prevent modifiers
 /// from being able to, for example, cause 0-thickness underlines.
 const Minimums = struct {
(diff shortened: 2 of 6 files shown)


src/font/face.zig
```diff
@@ -222,6 +222,16 @@ pub const RenderOptions = struct {
             y: f64,
         };
 
+        /// Returns true if the constraint does anything. If it doesn't,
+        /// because it neither sizes nor positions the glyph, then this
+        /// returns false.
+        pub inline fn doesAnything(self: Constraint) bool {
```
mitchellh, line 228: An idea... `Size` and `Align` can both fit into a `u2` (separately), so all 4 of these fields could fit into a `packed struct` as a `u8`, then you could do a `!= 0` comparison here.

It would probably save space on our structure too, not sure how much its used.

mitchellh (approved):
Looks good to me. Assuming the various math bits are right. I made one suggestion I had while passing through but no need to do it to merge.

src/font/face.zig
qwerasd205: Yeah, I considered making the constraint a packed struct as a whole, but it has a nullable field in it. I guess just condensing those specific fields would work though (and would make a smaller comparison than comparing the whole struct to something). I wasn't super concerned about performance since rasterizing glyphs isn't high throughput since it's all cached, but I wouldn't object to this change.

qwerasd205:
As for the math being correct- I stared at my terminal under both CoreText and FreeType, zooming in and out, selecting different things, and it all looks fine, at least with JetBrains Mono, the Symbols Only Nerd Font, and the emoji fonts (Apple Color Emoji for CoreText, Noto Color Emoji for FreeType).

Also took lots of screenshots and zoomed in on the pixels, looked for discrepancies or other problems. There *could* be some issue(s) remaining but whatever they are they'd be very minor-- for CoreText this puts the rasterization of glyphs back to how they were before the first time I disabled `shouldSubpixelQuantizeFonts`, except now we have better alignment and scaling for constrained glyphs; and for FreeType this improves that alignment and scaling for constrained glyphs and I'm *pretty* sure it doesn't mess up the hinting for non-constrained glyphs (I tried to be careful about that).

PR #8212, opened 2025-08-12
gtk-ng: port the terminal inspector

jcollie:
This is a (relatively) straightforward port of the terminal inspector from the old GTK application runtime. It's split into three widgets. At the lowest level is a widget designed for showing a generic Dear ImGui application. Above that is a widget that embeds the ImGui widget and plumbs it into the core Inspector. At the top is a custom Window widget that embeds the Inspector widget.

And then there's all the plumbing necessary to hook everything into the rest of Ghostty.

In theory this design _should_ allow showing the Inspector in a split or a tab in the future, not just in a separate window. It should also make it easier to display _other_ Dear ImGui applications if they are ever needed.

### DIFF
--- a/src/apprt/gtk-ng/App.zig
+++ b/src/apprt/gtk-ng/App.zig
@@ -99,7 +99,6 @@ pub fn performIpc(
 }
 
 /// Redraw the inspector for the given surface.
-pub fn redrawInspector(self: *App, surface: *Surface) void {
-    _ = self;
-    _ = surface;
+pub fn redrawInspector(_: *App, surface: *Surface) void {
+    surface.redrawInspector();
 }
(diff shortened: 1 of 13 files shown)


jcollie:
`valgrind` seemed happy, at least for a short test.

mitchellh:
I've gotta refamiliarize myself with how cimgui and the inspector all hooked together but the high level architecture sounds right.

mitchellh:
I made some changes to `ImguiWidget` to make it use signals rather than callbacks. It also should support the widget being unrealized and realized multiple times (something that will happen if we ever support moving the widget e.g. between a window and a split).

I'm looking into the inspector itself now.

mitchellh (changes requested):
Just putting the red review to block merging. I'm slowly making my way through here, getting close! 😄

mitchellh:
Okay Valgrind is happy on this and we're ready to go.

mitchellh (approved):
Okay, I'm happy with this. 

* The dangling `ref` call so we don't "close immediately" felt weird to me, and it was. I got rid of that and everything is working well (required reworking some stuff of course).
* used a more gobject-style API for callbacks (signals)
* centralized all inspector control to one API (`controlInspector`) vs four (show, hide, toggle, control). Less chance to mess things up, easier to understand.

PR #8216, opened 2025-08-12
gtk-ng: get window-subtitle working

jcollie:


tristan957 (approved):

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -1270,6 +1290,28 @@ pub const Surface = extern struct {
     //---------------------------------------------------------------
     // Properties
 
+    /// Get the subtitle. If the `window-subtitle` config is set to
+    /// `working-directory` this will be the `pwd`, otherwise it is `null`.
+    fn propGetSubTitle(self: *Self, value: *gobject.Value) void {
```
mitchellh, line 1295: I've been trying to idiomatically use `getSubtitleValue` formats for the value-based accessors. Also subtitle probably should not have the `T` caps.

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -216,6 +216,26 @@ pub const Surface = extern struct {
             );
         };
 
+        /// The subtitle for this surface. If `window-subtitle` is set to
+        /// `working-directory` this will be the working directory, otherwise it
+        /// will be null.
+        pub const subtitle = struct {
```
mitchellh, line 222: I think the surface shouldn't be aware of this at all. It seems like this is a window/tab concern.

mitchellh (changes requested):
I think there's some question of where things go. 

I'm going to be getting to reworking title stuff here in a sec cause I think I have to do it as part of split zooming (because we add the magnifier emoji).

mitchellh:
Replaced by #8218

PR #8221, opened 2025-08-13
apprt/gtk-ng: bell 

mitchellh:
Supersedes #8129

This is a rewrite but I did take pieces of #8129. I dropped the new feature that was mixed into the PR because I'm trying not to introduce new features in `-ng` right now. Feel free to PR that separately @jcollie. I also dropped some of the action group validation stuff which admittedly would be nice, so also happy to add that.

A big change I made here is we don't need to expose `bell-features` from surface, because we can use the relevant config that we have access to. I passed the config as a closure parameter so it recomputes when config changes, too. 

I also fixed a bug I found where we'd lose computed titles on non-focused tabs because `active-surface` would start returning null (since none are focused there). We now fallback to the active surface being the _last focused_ surface if no focused surface exists, which matches the behavior we also have on macOS.

### DIFF
--- a/src/apprt/gtk-ng/class.zig
+++ b/src/apprt/gtk-ng/class.zig
@@ -53,6 +53,23 @@ pub fn Common(
             }
         }).private else {};
 
+        /// A helper that creates a property that reads and writes a
+        /// private field with only shallow copies. This is good for primitives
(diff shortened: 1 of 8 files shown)


mitchellh:
Added @jcollie as co-author.

PR #8223, opened 2025-08-13
apprt/gtk-ng: prompt surface title

mitchellh:
Straightforward port. A hell of a lot cleaner with `-ng`.


mitchellh:
Valgrind is happy on both the set and unset case.

PR #8224, opened 2025-08-13
gtk-ng: show on-screen keyboard on LMB release

### LINKED ISSUE #7987: GTK: Support on-screen keyboards
### Discussed in https://github.com/ghostty-org/ghostty/discussions/7958

<sup>Originally posted by **houssemko** July 16, 2025</sup>
### Issue Description

Ghostty currently doesn't prompt for the on-screen keyboard when one is available, unlike VTE-based terminals like GNOME Console or Ptyxis. 

This appears to have affected VTE as well just a few years ago: https://gitlab.gnome.org/GNOME/vte/-/issues/2630. Looks like all we need to do to fix this issue is to call `Gtk.IMContext.activate_osk` when the user clicks on the terminal pane, but this is logic that we need to implement ourselves. Maybe we can also make an apprt action for this.

pluiedev:
This aligns with VTE behavior when the on-screen keyboard is enabled in GNOME's accessibility settings.

Also adds a corresponding binding action since why not

Closes #7987


mitchellh (approved):
Approved pending CI.

PR #8228, opened 2025-08-13
gtk-ng: add a helper for creating GTK actions

jcollie:
- Reduces boilerplate.
- Adds type safety.
- Adds comptime checks for action and group names which
  otherwise could cause panics at runtime.


mitchellh (changes requested):
Love it, but since there is so much action related stuff, maybe we should start to namespace a bit, and we can clean up func names that way. How about `ext.action.add` `addTomap`, etc.

mitchellh:
It can still be all in one file (ext.zig), though we may want to consider pulling out `ext/action.zig` and stuff too.

jcollie:
OK, rebased on main to pick up the `new-split` parameterization, and moved the helpers into a namespace.

mitchellh:
conflict again please rebase

jcollie:
Rebased again to pick up the inspector and close changes.

pluiedev (commented):
I feel like the API surface is a bit too broad and confusing - `addToMap` and `add` could totally be merged together, and then the assertion failure when Self doesn't inherit from ActionMap would be apparent in the call site and not in the helper function.

I also wonder if we should conceptually model actions as a tagged union instead, and use comptime to iterate over each variant instead of using a list. Would probably also allow us to avoid hardcoding action strings everywhere and use enum literals instead

jcollie:
> I feel like the API surface is a bit too broad and confusing - `addToMap` and `add` could totally be merged together, and then the assertion failure when Self doesn't inherit from ActionMap would be apparent in the call site and not in the helper function.

You can't really merge them because `addToMap` is a helper for `add` and `addAsGroup`. Even though it's marked as `pub` it's not really meant to be called by anything else.

The reason that they can't be merged is that in `addToMap` `self` and `map` are not necessarily the same object so there's no inheritance to be checked. `addAsGroup` depends on that.

> I also wonder if we should conceptually model actions as a tagged union instead, and use comptime to iterate over each variant instead of using a list. Would probably also allow us to avoid hardcoding action strings everywhere and use enum literals instead

That seems overly complicated, as we still have to use strings in the GTK apis.

jcollie:
> I also wonder if we should conceptually model actions as a tagged union instead, and use comptime to iterate over each variant instead of using a list. Would probably also allow us to avoid hardcoding action strings everywhere and use enum literals instead

Also, using comptime is complicated by the fact that you can't construct a Variant except by calling extern C functions. That's why I had to add the `.init` function to `ext.actions.Action` rather than doing the comptime checks in `addToMap`.

mitchellh (approved):
I think this is a good improvement. I see places we can do some other cleanup but happy with the direction.

PR #8230, opened 2025-08-14
macOS Custom Icon + Persistence

NicholasMata:
This PR aims to improve custom icons on macOS in the following ways. (I based this PR on the discussion #3631) 

### Currently
- Current Icon customizations are not persistent *(when closing the application the icon in dock reverts back to official icon)*
- There is no officially supported way to change icon to be something completely custom. 

### After this PR
- Current icon customizations are persistent (closing the application no longer reverts back to official icon)
- Ghostty config `macos-icon` has a new option `custom` which by default looks for icon `~/.config/ghostty/Ghostty.icns`. It has an accompanying new configuration `macos-custom-icon` which allows for a different path to be specified, it does support more than just `.icns` as well.

Both changes are based on the thread with @sfsam in https://github.com/ghostty-org/ghostty/discussions/3631#discussioncomment-12180647

Feedback is always welcome, if I have not done something up to par please let me know and I will do my best to correct it.

NOTE: I did notice some newlines with indents which seems to be against convention in those files so I removed the whitespace if this is not preferred I can revert.

---

P.S. Thanks for all the work you put into making an awesome terminal!

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -837,6 +837,13 @@ class AppDelegate: NSObject,
         case .xray:
             self.appIcon = NSImage(named: "XrayImage")!
 
+        case .custom:
+            if let userIcon = NSImage(contentsOfFile: config.macosCustomIcon) {
+                self.appIcon = userIcon
+            } else {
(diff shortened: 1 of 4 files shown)


mitchellh:
This looks good to me. Approved pending CI. Thank you.

mitchellh (approved):

PR #8233, opened 2025-08-14
apprt/gtk-ng: clean up close handling of all types

mitchellh:
This cleans up our close handling of all types (surfaces, tabs, windows). Surfaces no longer emit their scope; their scope is always just the surface itself. For tab and window scope we use widget actions.

This makes `close_tab` work properly (previously broken).

### DIFF
--- a/src/apprt/gtk-ng/Surface.zig
+++ b/src/apprt/gtk-ng/Surface.zig
@@ -32,7 +32,8 @@ pub fn rtApp(self: *Self) *ApprtApp {
 }
 
 pub fn close(self: *Self, process_active: bool) void {
-    self.surface.close(.{ .surface = process_active });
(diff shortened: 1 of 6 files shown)


src/apprt/gtk-ng/Surface.zig
```diff
@@ -32,7 +32,8 @@ pub fn rtApp(self: *Self) *ApprtApp {
 }
 
 pub fn close(self: *Self, process_active: bool) void {
-    self.surface.close(.{ .surface = process_active });
+    _ = process_active;
```
mitchellh, line 35: We should look to remove this from the core libghostty API. `core.needsConfirmQuit` handles this for us. But didn't want to blow up scope and wanted to do a bit more investigation.

PR #8237, opened 2025-08-14
gtk-ng: use virtual methods to draw the inspector

jcollie:
Insead of signals between the ImGui widget and the Inspector widget, make the Inspector widget a subclass of the ImGui widget and use virtual methods to handle setup and rendering of the Inspector.

### DIFF
--- a/src/apprt/gtk-ng/class.zig
+++ b/src/apprt/gtk-ng/class.zig
@@ -53,6 +53,18 @@ pub fn Common(
             }
         }).private else {};
 
+        /// Get the class structure for the object.
+        ///
+        /// This seems ugly and unsafe to me but this is what GObject is doing
+        /// under the hood.
+        ///
+        /// https://gitlab.gnome.org/GNOME/glib/-/blob/main/gobject/gtype.h?ref_type=heads#L555-571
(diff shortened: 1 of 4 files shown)


mitchellh:
I'm going out of town for a bit. This looks good on the surface but just want to double check that this is safe and what gobject is really doing under the hood.

jcollie:
> just want to double check that this is safe and what gobject is really doing under the hood

Yeah, seems too much like magic to me.

src/apprt/gtk-ng/class.zig
```diff
@@ -53,6 +53,18 @@ pub fn Common(
             }
         }).private else {};
 
+        /// Get the class structure for the object.
+        ///
+        /// This seems ugly and unsafe to me but this is what GObject is doing
+        /// under the hood.
+        ///
+        /// https://gitlab.gnome.org/GNOME/glib/-/blob/main/gobject/gtype.h?ref_type=heads#L555-571
+        /// https://gitlab.gnome.org/GNOME/glib/-/blob/main/gobject/gtype.h?ref_type=heads#L2673
```
tristan957: Good to use permalinks!

jcollie: They hide those pretty good!

jcollie:
Add some comments with some preliminary investigations into GObject memory layout.

PR #8238, opened 2025-08-15
ci: switch to debian 13

jcollie:


.github/workflows/test.yml
```diff
@@ -957,34 +957,30 @@ jobs:
...
+  test-debian-13:
+    name: Test build on Debian 13
     runs-on: namespace-profile-ghostty-sm
-    needs: [test, build-dist]
+    needs: [test]
+    env:
+      ZIG_LOCAL_CACHE_DIR: /zig/local-cache
+      ZIG_GLOBAL_CACHE_DIR: /zig/global-cache
     steps:
+      - name: Checkout code
```
mitchellh: This should use the dist tarball and not our Git source. We don't want to commit to Debian 13 from Git source, we just want to make sure it packages fine (and packages use dist).

mitchellh: This should use the dist tarball and not our Git source. We don't want to commit to Debian 13 from Git source, we just want to make sure it packages fine (and packages use dist).

mitchellh (changes requested):

.github/workflows/test.yml
jcollie: Fixed.

jcollie: Fixed.

PR #8249, opened 2025-08-15
fix UAF in grow

matklad:
Grow needs to allocate and might fail midway. It tries to handle this using "undo" pattern, and restoring old state on error. But this is exactly what steps into UAF, as, on error, both errdefer and defer are run, and the old data is freed.

Instead, use a more robust "reservation" pattern, where we first fallibly resrve all the resources we need, without applying any changes, and than do the actual change once we are sure that cannot fail.

### DIFF
--- a/src/font/Atlas.zig
+++ b/src/font/Atlas.zig
@@ -287,27 +287,30 @@ pub fn grow(self: *Atlas, alloc: Allocator, size_new: u32) Allocator.Error!void 
     assert(size_new >= self.size);
     if (size_new == self.size) return;
 
+    // We reserve space ahead of time for the new node, so that we
+    // won't have to handle any errors after allocating our new data.
     try self.nodes.ensureUnusedCapacity(alloc, 1);
-    const data_new = try alloc.alloc(u8, size_new * size_new * self.format.depth());
-    errdefer comptime unreachable; // End resource reservation phase.
 
-    // Preserve our old values so we can copy the old data
+    const data_new = try alloc.alloc(
+        u8,
+        size_new * size_new * self.format.depth(),
+    );
+
+    // Function is infallible from this point.
+    errdefer comptime unreachable;
+
+    // Keep track of our old data so that we can copy it.
     const data_old = self.data;
     const size_old = self.size;
 
+    // Update our data and size to our new ones.
     self.data = data_new;
+    self.size = size_new;
+
+    // Free the old data once we're done with it.
     defer alloc.free(data_old);
 
-    // Add our new rectangle for our added righthand space.
-    self.nodes.appendAssumeCapacity(alloc, .{
-        .x = size_old - 1,
-        .y = 1,
-        .width = size_new - size_old,
-    });
-
-    // If our allocation and rectangle add succeeded, we can go ahead
-    // and persist our new size and copy over the old data.
-    self.size = size_new;
+    // Zero the new data out and copy the old data over.
     @memset(self.data, 0);
     self.set(.{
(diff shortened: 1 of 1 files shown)


qwerasd205:
Nice catch. Would like some explanatory comments in the code here though. Also the code change makes the contents of the comment above the append inaccurate, since the possibility of failing was moved up.

src/font/Atlas.zig
```diff
@@ -291,18 +291,17 @@ pub fn grow(self: *Atlas, alloc: Allocator, size_new: u32) Allocator.Error!void
     const data_old = self.data;
     const size_old = self.size;
 
-    // Allocate our new data
-    self.data = try alloc.alloc(u8, size_new * size_new * self.format.depth());
+    try self.nodes.ensureUnusedCapacity(1);
```
qwerasd205: You forgot to pass the allocator here-- test fails because this code can't even compile.

src/font/Atlas.zig
```diff
@@ -291,18 +291,17 @@ pub fn grow(self: *Atlas, alloc: Allocator, size_new: u32) Allocator.Error!void
...
-    errdefer {
-        alloc.free(self.data);
-        self.data = data_old;
-    }
 
     // Add our new rectangle for our added righthand space. We do this
     // right away since its the only operation that can fail and we want
     // to make error cleanup easier.
-    try self.nodes.append(alloc, .{
+    self.nodes.appendAssumeCapacity(alloc, .{
```
qwerasd205: `appendAssumeCapacity` doesn't take an allocator. Please make sure your code actually compiles before making a PR.

qwerasd205 (changes requested):
There are multiple errors in this code that make it not even compile.

src/font/Atlas.zig
matklad: I am not motivated enough to setup a local dev environment for testing, and I am fine with PR being closed for that reason :)

matklad: See also https://matklad.github.io/2022/10/24/actions-permissions.html though, given that you use custom runners, this might not be applicate due to security implications.

matklad:
Adjusted comments by mostly removing them. resource reservation phase and `errdefer comptime unreachable` are a specific code pattern, so I think it should be known by name, rather than be explained every time in the code comments. For the curious reader, a full explanation is available in the commit message.

src/font/Atlas.zig
qwerasd205: I can do the finishing work on this, you can leave the PR open and I'll commit the rest of the necessary changes to it before merging. Thanks for finding the UAF.

Side note: Since you don't even have a local dev environment set up (which would literally just mean cloning the repo and having zig 0.14.1 installed btw) I'm guessing you didn't find this while exploring the code yourself? Are you working on an analyzer to find this pattern that leads to UAF or something?

qwerasd205: (The workflows were manually approved to run, by the way- I don't think there's any issue with the permissions we have set up.)

matklad: > I'm guessing you didn't find this while exploring the code yourself

No, I just have a hobby of grepping zig projects for `errdefer`, as the code around is aways buggy: 

https://github.com/ziglang/zig/pull/22131

>The workflows were manually approved to run,

Yeah, I mean that if workflow didn't require manual approvals, I'd be able to iterate here quicker.

mitchellh (approved):

qwerasd205 (approved):

PR #8254, opened 2025-08-15
Configurable right click behavior

### LINKED ISSUE #4404: Introduce option to right click to copy

lujobi:
This MR addresses  #4404 following the approach suggested [here](https://github.com/ghostty-org/ghostty/issues/4404#issuecomment-2708410143). Implementing the behavior known e.g. from Putty or Windows Terminal. 

The following configuration values for `right-click-action` are provided: 
* `context-menu` - Show the context menu.
* `paste` - Paste the contents of the clipboard.
* `copy` - Copy the selected text to the clipboard.
* `copy-and-paste` - Copy the selected text to the clipboard, paste if nothing is selected.
*  `ignore` - Do nothing, ignore the right-click.

I followed #5935 for getting an idea on where to start. I hope this to be a temporary solution until "bindable mouse bindings" are introduced.

This is my first time writing Zig code, so I am happy to incorporate any feedback. 

Thank you all very much for your work!

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1833,6 +1833,32 @@ fn clipboardWrite(self: *const Surface, data: []const u8, loc: apprt.Clipboard) 
     };
 }
 
+fn copySelectionToClipboards(
+    self: *Surface,
+    sel: terminal.Selection,
+    clipboards: []const apprt.Clipboard,
+) void {
+    const buf = self.io.terminal.screen.selectionString(self.alloc, .{
+        .sel = sel,
+        .trim = self.config.clipboard_trim_trailing_spaces,
+    }) catch |err| {
+        log.err("error reading selection string err={}", .{err});
+        return;
+    };
+    defer self.alloc.free(buf);
+
+    for (clipboards) |clipboard| self.rt_surface.setClipboardString(
+        buf,
+        clipboard,
+        false,
+    ) catch |err| {
+        log.err(
+            "error setting clipboard string clipboard={} err={}",
+            .{ clipboard, err },
+        );
+    };
+}
+
 /// Set the selection contents.
 ///
 /// This must be called with the renderer mutex held.
@@ -1850,33 +1876,13 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
     const sel = sel_ orelse return;
     if (prev_) |prev| if (sel.eql(prev)) return;
 
-    const buf = self.io.terminal.screen.selectionString(self.alloc, .{
-        .sel = sel,
-        .trim = self.config.clipboard_trim_trailing_spaces,
-    }) catch |err| {
-        log.err("error reading selection string err={}", .{err});
-        return;
-    };
-    defer self.alloc.free(buf);
-
-    // Set the clipboard. This is not super DRY but it is clear what
-    // we're doing for each setting without being clever.
     switch (self.config.copy_on_select) {
         .false => unreachable, // handled above with an early exit
 
         // Both standard and selection clipboards are set.
         .clipboard => {
             const clipboards: []const apprt.Clipboard = &.{ .standard, .selection };
-            for (clipboards) |clipboard| self.rt_surface.setClipboardString(
-                buf,
-                clipboard,
-                false,
-            ) catch |err| {
-                log.err(
-                    "error setting clipboard string clipboard={} err={}",
-                    .{ clipboard, err },
-                );
-            };
+            copySelectionToClipboards(self, sel, clipboards);
         },
 
         // The selection clipboard is set if supported, otherwise the standard.
@@ -1885,17 +1891,7 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
                 .selection
             else
                 .standard;
-
-            self.rt_surface.setClipboardString(
-                buf,
-                clipboard,
-                false,
-            ) catch |err| {
-                log.err(
-                    "error setting clipboard string clipboard={} err={}",
-                    .{ clipboard, err },
-                );
-            };
+            copySelectionToClipboards(self, sel, &.{clipboard});
         },
     }
 }
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -247,6 +247,7 @@ const DerivedConfig = struct {
     clipboard_paste_protection: bool,
     clipboard_paste_bracketed_safe: bool,
     copy_on_select: configpkg.CopyOnSelect,
+    right_click_action: configpkg.RightClickAction,
     confirm_close_surface: configpkg.ConfirmCloseSurface,
     cursor_click_to_move: bool,
     desktop_notifications: bool,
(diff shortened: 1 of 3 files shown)


src/config/Config.zig
```diff
@@ -1886,6 +1886,20 @@ keybind: Keybinds = .{},
...
 
+/// The action to take when the user right-clicks on the terminal surface.
+///
+/// Valid values:
+///   * `context-menu` - Show the context menu.
+///   * `paste` - Paste the contents of the clipboard.
+///   * `copy` - Copy the selected text to the clipboard.
+///   * `copy-and-paste` - Depending on the current selection:
+///         * Some text selected: Copy the selected text to the clipboard.
+///         * No selection: Paste the contents of the clipboard.
```
pluiedev: This is more accurately copy ***or*** paste

```suggestion
///   * `copy-or-paste` - If there is a selection, copy the selected text to the clipboard; otherwise, paste the contents of the clipboard.
```

pluiedev (changes requested):
One small suggestion to the docs

src/config/Config.zig
```diff
@@ -1886,6 +1886,18 @@ keybind: Keybinds = .{},
     else => .false,
 },
 
+/// The action to take when the user right-clicks on the terminal surface.
+///
+/// Valid values:
+///   * `context-menu` - Show the context menu.
+///   * `paste` - Paste the contents of the clipboard.
+///   * `copy` - Copy the selected text to the clipboard.
+///   * `copy-or-paste` - If there is a selection, copy the selected text to the clipboard; otherwise, paste the contents of the clipboard.
```
00-kat: This should probably be hard-wrapped like everything else.
```suggestion
///   * `copy-or-paste` - If there is a selection, copy the selected text to the
///      clipboard; otherwise, paste the contents of the clipboard.
```

src/Surface.zig
```diff
@@ -3582,18 +3580,45 @@ pub fn mouseButtonCallback(
...
+                const sel = screen.selectWord(pin) orelse break :sel;
+                try self.setSelection(sel);
+                try self.queueRender();
+                return false;
+            },
         }
 
-        const sel = screen.selectWord(pin) orelse break :sel;
-        try self.setSelection(sel);
+        try self.setSelection(null);
```
00-kat, line 3620: This makes `right-click-action = ignore` clear the selection, which isn't the same as ignoring right click.

00-kat (changes requested):

src/Surface.zig
lujobi: Good point, thank you! I added an early return to the switch statement.

src/Surface.zig
```diff
@@ -3582,18 +3580,47 @@ pub fn mouseButtonCallback(
             break :pin pin;
         };
 
-        // If we already have a selection and the selection contains
-        // where we clicked then we don't want to modify the selection.
-        if (self.io.terminal.screen.selection) |prev_sel| {
-            if (prev_sel.contains(screen, pin)) break :sel;
+        switch (self.config.right_click_action) {
+            .ignore => {
+                return true;
```
00-kat, line 3586:
```suggestion
                // Return early to skip clearing the selection.
                try self.queueRender();
                return true;
```

00-kat (changes requested):

00-kat (approved):

lujobi:
@pluiedev would you mind taking another look the requested changes? Thank you!

src/Surface.zig
```diff
@@ -1850,33 +1878,13 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
...
-                buf,
-                clipboard,
-                false,
-            ) catch |err| {
-                log.err(
-                    "error setting clipboard string clipboard={} err={}",
-                    .{ clipboard, err },
-                );
-            };
+            copySelectionToClipboards(self, sel, clipboards);
```
pluiedev: Call methods with method call syntax (and also inline the clipboards too)
```suggestion
            self.copySelectionToClipboards(sel, &.{ .standard, .selection });
```

src/Surface.zig
```diff
@@ -1885,17 +1893,7 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
...
-                buf,
-                clipboard,
-                false,
-            ) catch |err| {
-                log.err(
-                    "error setting clipboard string clipboard={} err={}",
-                    .{ clipboard, err },
-                );
-            };
+            copySelectionToClipboards(self, sel, &.{clipboard});
```
pluiedev:
```suggestion
            self.copySelectionToClipboards(sel, &.{clipboard});
```

src/Surface.zig
```diff
@@ -3582,18 +3580,49 @@ pub fn mouseButtonCallback(
...
-            if (prev_sel.contains(screen, pin)) break :sel;
+        switch (self.config.right_click_action) {
+            .ignore => {
+                // Return early to skip clearing the selection.
+                try self.queueRender();
+                return true;
+            },
+            .copy => {
+                if (self.io.terminal.screen.selection) |sel| {
+                    copySelectionToClipboards(self, sel, &.{.standard});
```
pluiedev:
```suggestion
                    self.copySelectionToClipboards(sel, &.{.standard});
```

src/Surface.zig
```diff
@@ -3582,18 +3580,49 @@ pub fn mouseButtonCallback(
...
+                }
+            },
+            .@"copy-or-paste" => {
+                if (self.io.terminal.screen.selection) |sel| {
+                    copySelectionToClipboards(self, sel, &.{.standard});
```
pluiedev:
```suggestion
                    self.copySelectionToClipboards(sel, &.{.standard});
```

src/Surface.zig
```diff
@@ -3582,18 +3580,49 @@ pub fn mouseButtonCallback(
...
+                } else {
+                    try self.startClipboardRequest(.standard, .{ .paste = {} });
```
pluiedev: A tagged union variant with type `void` can be more simply written as an enum literal
```suggestion
                    try self.startClipboardRequest(.standard, .paste);
```

src/Surface.zig
```diff
@@ -3582,18 +3580,49 @@ pub fn mouseButtonCallback(
...
+                }
+            },
+            .paste => {
+                try self.startClipboardRequest(.standard, .{ .paste = {} });
```
pluiedev:
```suggestion
                try self.startClipboardRequest(.standard, .paste);
```

pluiedev (changes requested):
The rest looks good, just some code nitpicks

pluiedev (approved):
Code looks good pending CI

mitchellh (approved):
Thanks, looks great.

anaisbetts:
I spent an evening vibe coding QuickEdit only to discover that it's already here, omg thank you!

PR #8277, opened 2025-08-19
Fix PageList Reflow OOM Conditions

### LINKED ISSUE #8009: Crash when hyperlinks cause an OutOfMemory error
Here's a transcript of running `dotnet build` on a medium-size project. gunzip the attached file, then `cat` it to the terminal. It will consistently crash a debug build of Ghostty (tried on version 1.1.4-main+015efcf9e, and also just now on 1.1.4-main+a88689ca7 as well). It does not seem to crash a release-optimized build, interestingly. It also does not crash other terminals (though I've only tried on Kitty so far, if I find a terminal besides Ghostty that this *does* crash I will mention it).

[dotnet-build-languageforge-lexbox.txt.gz](https://github.com/user-attachments/files/21339674/dotnet-build-languageforge-lexbox.txt.gz)

_Originally posted by @rmunn in https://github.com/ghostty-org/ghostty/discussions/7977#discussioncomment-13828475_

qwerasd205:
Previously, when encountering an OOM when copying graphemes, hyperlinks, or styles to a new page during reflow, the attempted resolution was to copy the current row in to a new page and continue on- which works in 99% of cases, but isn't sound, since it's possible for a single row to exceed the capacity on any of these.

This led to rare but real crashes like #8009.

I've added tests that produce all of the failure conditions, and resolved them by changing the strategy from making a new page to increasing the capacity of the current one.

There should probably be some level of abstraction added around this, since multiple places in the code now do this sort of thing- attempt to add some managed memory to a page, adjusting their capacity upwards as necessary. But for now, I kept it all inline here.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -7442,10 +7442,12 @@ test "PageList resize reflow exceeds grapheme memory forcing capacity increase" 
             rac.row,
             rac.cell,
             &@as(
-                [@divFloor(
-                    pagepkg.grapheme_bytes_default - 1,
-                    @sizeOf(u21),
-                )]u21,
+                [
+                    @divFloor(
+                        pagepkg.grapheme_bytes_default - 1,
+                        @sizeOf(u21),
+                    )
+                ]u21,
                 @splat('a'),
             ),
         );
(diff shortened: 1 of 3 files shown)


mitchellh (approved):
Always hard to review this stuff but nothing screams out as suspicious to me, and I feel good there is a test backing this.

PR #8289, opened 2025-08-19
AI tooling must be disclosed for contributions

mitchellh:
I think, at this stage of AI, it is a common courtesy to disclose this. 

In a perfect world, AI assistance would produce equal or higher quality work than any human. That isn't the world we live in today, and in many cases it's generating slop. I say this despite being a fan of and using them successfully myself (with heavy supervision)! I think the major issue is **inexperienced human drivers of AI** that aren't able to **adequately review their generated code.** As a result, they're pull requesting code that I'm sure they would be ashamed of if they knew how bad it was.

The disclosure is to help maintainers assess how much attention to give a PR. While we aren't obligated to in any way, I try to assist inexperienced contributors and coach them to the finish line, because getting a PR accepted is an achievement to be proud of. But if it's just an AI on the other side, I don't need to put in this effort, and it's rude to trick me into doing so.

**I'm a fan of AI assistance and use AI tooling myself.** But, we need to be responsible about what we're using it for and respectful to the humans on the other side that may have to review or maintain this code. 

(In the spirit of this PR... none of this PR was AI generated. lol.)


yawaramin (approved):
Broadly agree with this; maybe a pull request template can be added to make it explicit: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

Another advantage of a PR template is it can have a checklist for other stuff, eg [Developer Certificate of Origin](https://developercertificate.org/).

jcollie (approved):

amadeus:
gotta put a stop 2️⃣ da slop

tobi:
I think this is an opportunity for ai tools, too. GitHub should publish a standard for a special AI byline that all ai tools can write to. Anytime you use an ai tool, it adds itself to a .git staging file and the next commit lists all the involved tools appended to the message similar to co authors. 

Then GitHub lists and links to the tools. This way the tools get some exposure, but maintainers also can require this standard to be followed. Everyone wins. And the tools can stop spamming the co-authors byline which is the current viral loop which is a bit obnoxious.

stuaxo:
It seems sensible: if AI tools were used, the contributors should have reviewed, revised and understand what they are contributing.

zanellig:
LGTM

silva96:
Does this apply to tab completions? Most of the tab completions are just finishing what the user wanted to write anyways ...

chrishas35:
> Does this apply to tab completions? Most of the tab completions are just finishing what the user wanted to write anyways ...

this is covered in the doc change.

PR #8291, opened 2025-08-20
gtk-ng: use a custom fuzzy search to search the command palette

jcollie:
Create a custom GTK filter using Zf and use that to search the command palette instead of a strict substring search.


mitchellh:
I'm going to want to review this and run it through Valgrind (I'm sure you did too, but as a double-check).

jcollie:
Switching this to a draft. Something with this PR is triggering a crash in GTK.

PR #8292, opened 2025-08-20
snap: remove workaround for build failures

jcollie:


mitchellh:
We can actually remove this now. The base image is updated.

jcollie:
> We can actually remove this now. The base image is updated.

Ah, I was wondering if that was the case.

PR #8302, opened 2025-08-20
build: add run-valgrind and test-valgrind steps

mitchellh:
This adds two explicit `zig build` steps: `run-valgrind` and `test-valgrind` to run the Ghostty exe or tests under Valgrind, respectively.

This simplifies the manual Valgrind calls in a few ways:

1. It automatically sets the CPU to baseline, which is a frequent and requirement for Valgrind on newer CPUs, and generally safe.

2. It sets up the rather complicated set of flags to call Valgrind with, importantly setting up our suppressions.

3. It enables pairing it with the typical and comfortable workflow of specifying extra args (with `--`) or flags like `-Dtest-filter` for tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,37 +142,18 @@ Zig cannot detect by itself.
 
 ### On Linux
 
-On Linux the recommended tool to check for memory leaks is Valgrind. We supply
-a file containing suppressions for false positives and known leaks in 3rd party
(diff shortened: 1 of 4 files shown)


mitchellh:
Tests don't currently pass Valgrind at all so we won't run them in CI. 

I removed the Nix script and updated CONTRIBUTING.md.

cc @jcollie

jcollie (approved):

PR #8307, opened 2025-08-20
terminal: fix undefined memory access in OSC parser

mitchellh:
Fixes #8007

Verified with `test-valgrind -Dtest-filter="OSC"` which had cond access errors before, and none after this. Basically a copy of #8008.

### DIFF
--- a/src/terminal/osc.zig
+++ b/src/terminal/osc.zig
@@ -16,6 +16,10 @@ const kitty = @import("kitty.zig");
 const log = std.log.scoped(.osc);
 
 pub const Command = union(enum) {
+    /// This generally shouldn't ever be set except as an initial zero value.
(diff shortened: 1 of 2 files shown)


jcollie (changes requested):
This is missing the unit test from #8008

mitchellh:
> This is missing the unit test from #8008

It doesn't need it anymore, it triggers in more than 10 unit tests via Valgrind if you revert this.

mitchellh:
Also, your unit test in #8008 wasn't guaranteed. For example, it never crashed on my machines (Mac or Linux). Relying on undefined memory for a crash isn't reliable. Valgrind is.

jcollie (approved):

PR #8311, opened 2025-08-20
flatpak: manually install Zig 0.14.1

jcollie:
The SDK published on Flathub updated to Zig 0.15.1 which broke the Flathub build in CI. So let's install it ourselves so that we can control the version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,6 +38,7 @@ jobs:
       - test-debian-13
       - valgrind
       - zig-fmt
+      - flatpak
     steps:
       - id: status
         name: Determine status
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -1013,32 +1013,31 @@ jobs:
...
-  #   steps:
-  #     - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-  #     - uses: flatpak/flatpak-github-actions/flatpak-builder@10a3c29f0162516f0f68006be14c92f34bd4fa6c # v6.5
-  #       with:
-  #         bundle: com.mitchellh.ghostty
-  #         manifest-path: flatpak/com.mitchellh.ghostty.yml
-  #         cache-key: flatpak-builder-${{ github.sha }}
-  #         arch: ${{ matrix.variant.arch }}
-  #         verbose: true
+  flatpak:
```
mitchellh, line 1017: We have to add this back to the required tests section at the top.

jcollie: Done.

mitchellh (approved):

PR #8322, opened 2025-08-21
gtk-ng: properly skip Zig test

jcollie:


mitchellh (approved):
Effectively what we were already doing but making it show up more obviously in the logs is good.

PR #8324, opened 2025-08-21
gtk-ng: more complete GTK startup/shutdown in test

jcollie:


mitchellh (approved):
Why? If it works fine but why is this needed?

jcollie:
> Why? If it works fine but why is this needed?

Mainly to make the test more accurately reflect a real GTK environment. Probably doesn't make a difference here but I want a template to be able to test other GTK things where it might.

PR #8330, opened 2025-08-21
gtk-ng: allow XKB remaps for non-writing-system keys

pluiedev:
Compromise solution to #7356

XKB is naughty. It's really really naughty. I don't understand why we didn't just kill XKB with hammers during the Wayland migration and change it for something much better. I don't understand why we're content with what amounts to an OS-level software key remapper that completely jumbles information about original physical key codes in order to fake keyboard layouts, and not just let users who really want to remap keys use some sort of evdev or udev-based mapper program.

In a sane system like macOS, the "c" key is always the "c" key, but it's understood to produce the Unicode character "ц" when using a Russian layout. XKB defies sanity, and just pretends that your "c" key is actually a "ц" key instead, and so when you ask for the keybind "Ctrl+C" it just shrugs in apathy (#7309). And so, we took matters into our own hands and interpreted hardware keycodes ourselves.

But then, a *lot* of people have the ingrained muscle memory of swapping Escape with Caps Lock so that it is easier to hit. We respect that. In a sane system, they would use a remapper that actually makes the system think you've hit the Escape key when in reality you've hit the Caps Lock key, so in all intents and purposes to the OS and any app developer, these two just have their wires swapped. But not on Linux. Somehow this and the aforementioned case should be treated by the same key transform algorithm, which is completely diabolical.

As a result, we have to settle for a compromise that truly satisfies neither party — by allowing XKB remaps for keys that don't really change depending on the layout.

The Linux input stack besets all hopes and aspirations.

### DIFF
--- a/src/apprt/gtk-ng/class/surface.zig
+++ b/src/apprt/gtk-ng/class/surface.zig
@@ -838,7 +838,7 @@ pub const Surface = extern struct {
                 // such as single quote on a US international keyboard layout.
                 if (priv.im_composing) return true;
 
-                // If we were composing and now we're not it means that we committed
+                // If we were composing and now we're not, it means that we committed
                 // the text. We also don't want to encode a key event for this.
                 // Example: enable Japanese input method, press "konn" and then
                 // press enter. The final enter should not be encoded and "konn"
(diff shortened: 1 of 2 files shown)


src/apprt/gtk-ng/class/surface.zig
```diff
@@ -838,7 +838,7 @@ pub const Surface = extern struct {
                 // such as single quote on a US international keyboard layout.
                 if (priv.im_composing) return true;
 
-                // If we were composing and now we're not it means that we committed
+                // If we were composing and now we're not, it means that we committed
```
mitchellh, line 841: Excellent discovery.

src/input/key.zig
```diff
@@ -589,6 +589,84 @@ pub const Key = enum(c_int) {
...
+    /// physical key, leading to very annoying bugs like #7309 where the
+    /// physical key `XKB_KEY_c` gets remapped into `XKB_KEY_Cyrillic_tse`,
+    /// which causes all of our physical key handling to completely break down.
+    /// _Very naughty._
+    ///
+    /// As a compromise, given that writing system keys (§3.1.1) comprise the
+    /// majority of keys that "change meaning [...] based on the current locale
+    /// and keyboard layout", we allow all other keys to be remapped by default
+    /// since they should be fairly harmless. We might consider making this
+    /// configurable, but for now this should at least placate most people.
```
mitchellh, line 610: Okay, so we went with the approach of being explicit about writing keys. That works for me given the language of the w3c spec.

mitchellh (approved):
I'm good with this.

weskoerber:
You're awesome! I tested this and it works great for my use case (`caps:swapescape` in sway).

<details><summary>Show <code>ghostty +version</code> output</summary>


```
❯ ghostty +version
Ghostty 1.1.4-pluie/jj-rslmomxuxwpl+534aa508d

Version
  - version: 1.1.4-pluie/jj-rslmomxuxwpl+534aa508d
  - channel: tip
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.gtk-ng
  - font engine   : font.main.Backend.fontconfig_freetype
  - libxev        : io_uring
```


</details>

PR #8331, opened 2025-08-21
build: build ReleaseFast and baseline CPU by default for system packages

mitchellh:
This makes `--system` imply `-Doptimize=ReleaseFast` and `-Dcpu=baseline`. These can still be overridden by the user by setting the same flags explicitly.

The `--system` flag is used to build system packages and its a safe assumption that in those cases, Ghostty is being built for general use amongst many users, so the default should be optimized for performance and compatibility.

This was always documented in our PACKAGING.md, but is now automatic. The PACKAGING.md file has been updated to reflect this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -300,7 +300,11 @@ jobs:
       # GhosttyKit is the framework that is built from Zig for our native
       # Mac app to access.
       - name: Build GhosttyKit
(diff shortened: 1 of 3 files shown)


mitchellh:
macOS is having some problems. I have to look into this further.

mitchellh:
Closing this for now, don't have time to chase this issue down and its not that important.

PR #8339, opened 2025-08-21
Nix VM integration tests

jcollie:
This adds a couple of Nix-based VM integration tests:

- Does `ghostty +version` run successfully?
- Can we create a new terminal window? (This is detected by setting the background to a color that doesn't appear
  normally on the desktop and seeing if we can detect that color in a screenshot).

Obviously more can be done but I thought that these would be a couple of good first steps.

The whole test suite can be run with `nix flake check`. Individual tests can be run with a command like this:

```
nix run .#check.x86_64-linux.<test name>.driver
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ zig-cache/
 .zig-cache/
 zig-out/
 /result*
+/.nixos-test-history
 example/*.wasm
 test/ghostty
 test/cases/**/*.actual.png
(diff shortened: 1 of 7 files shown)


jcollie:
The introduction of `home-manager` is likely to be controversial. However I find it the best way to declaratively manage user-specific settings. There are ways to do it in "normal" Nix configuration or imperatively in the test script but I prefer using home-manager.

mitchellh (changes requested):
I had a WIP branch with this that I unfortunately lost (looks like I never pushed it). One thing I remember doing was documenting how to start the VMs with SSH, VNC, etc. so that you can interactively debug these. I forgot the incantation but I'd love to make sure that's in here, that's the only sane way to work on these. :)

PR #8345, opened 2025-08-22
i18n: update bg_BG translations

reo101:
Part of https://github.com/ghostty-org/ghostty/issues/8344

### DIFF
--- a/po/bg_BG.UTF-8.po
+++ b/po/bg_BG.UTF-8.po
@@ -2,14 +2,15 @@
 # Copyright (C) 2025 Mitchell Hashimoto
 # This file is distributed under the same license as the com.mitchellh.ghostty package.
 # Damyan Bogoev <damyan.bogoev@gmail.com>, 2025.
+# reo101 <pavel.atanasov2001@gmail.com>, 2025.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2025-07-22 17:18+0000\n"
-"PO-Revision-Date: 2025-05-19 11:34+0300\n"
-"Last-Translator: Damyan Bogoev <damyan.bogoev@gmail.com>\n"
+"PO-Revision-Date: 2025-08-22 14:52+0300\n"
+"Last-Translator: reo101 <pavel.atanasov2001@gmail.com>\n"
 "Language-Team: Bulgarian <dict@ludost.net>\n"
 "Language: bg\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


po/bg_BG.UTF-8.po
```diff
@@ -208,12 +209,12 @@ msgstr "Позволи"
...
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-write.blp:77
 msgid "Remember choice for this split"
-msgstr ""
+msgstr "Запомни избора за това разделяне"
 
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-read.blp:82
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-write.blp:78
 msgid "Reload configuration to show this prompt again"
-msgstr ""
+msgstr "Презареди конфигурацията, за да покажеш това съобщение отново"
```
phush0: probably this should be worded backward "За да покажеш това съобщение отново, презареди конфигурацията"

po/bg_BG.UTF-8.po
```diff
@@ -278,15 +279,15 @@ msgstr "Копирано в клипборда"
 
 #: src/apprt/gtk/Surface.zig:1268
 msgid "Cleared clipboard"
-msgstr ""
+msgstr "Клипбордът бе изчистен"
```
damyanbogoev: I think 'Клипбордът е изчистен' (The clipboard is cleared) is a bit more suitable here, as it indicates a current state rather than a past action.

damyanbogoev (approved):

po/bg_BG.UTF-8.po
phush0: Totally agree

reo101: I think it indicates the immediate state of change that has just happenned, hence why the original text isn't the general statement "The cliboard is cleared" (which does not hint at when the action has happenned)

po/bg_BG.UTF-8.po
reo101: Both sound adequate, why do you prefer the reversed?

phush0: because technical literature is using passive voice

po/bg_BG.UTF-8.po
phush0: it is the state of the cliboard and again it is passive voice and shuld be as @damyanbogoev proposed

po/bg_BG.UTF-8.po
reo101: Fair, will implement

po/bg_BG.UTF-8.po
reo101: Ditto, will implement

reo101:
On the note of passive voice, should we also change strings like `Командата завърши (не)успешно` to `Командата бе завършена (не)успешно`? Sounds a bit out of place, but I cannot seem to think of a better localization for "finished"...

phush0:
If it is passive voice it shoud be "Командата е успешна/неуспешна" , because it goes to adjectiv in passive voice. Other wise it will be equal to "Command completed successfully" , which is not the case literal it is "Командата успя", but because it is passive voice it should be what I wrote. It is up to you to change this one as I don't mind it.

edit: probably change "бе" with 'е',  again up to you as I don't mind this one

trag1c:
Hi @phush0, do you volunteer to join the `bg_BG` localization team to be pinged on future string changes? :)

phush0:
Sure no problem

reo101:
> If it is passive voice it shoud be "Командата е успешна/неуспешна" , because it goes to adjectiv in passive voice. Other wise it will be equal to "Command completed successfully" , which is not the case literal it is "Командата успя", but because it is passive voice it should be what I wrote. It is up to you to change this one as I don't mind it.
> 
> edit: probably change "бе" with 'е', again up to you as I don't mind this one

Thanks for the extensive answer! Малко ми куца техническата граматика, тъй като не съм правил подобни преводи преди :smile:

phush0:
> > If it is passive voice it shoud be "Командата е успешна/неуспешна" , because it goes to adjectiv in passive voice. Other wise it will be equal to "Command completed successfully" , which is not the case literal it is "Командата успя", but because it is passive voice it should be what I wrote. It is up to you to change this one as I don't mind it.
> > edit: probably change "бе" with 'е', again up to you as I don't mind this one
> 
> Thanks for the extensive answer! Малко ми куца техническата граматика, тъй като не съм правил подобни преводи преди 😄

човек се учи докато е жив :)

phush0:
LGTM

trag1c (approved):

PR #8349, opened 2025-08-22
i18n: Updating Irish translation for Ghostty 1.2

aindriu80:
This Pull Request updates  5 translation strings for Ghostty 1.2 as per #8344

### DIFF
--- a/po/ga_IE.UTF-8.po
+++ b/po/ga_IE.UTF-8.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2025-07-22 17:18+0000\n"
-"PO-Revision-Date: 2025-08-22 14:52+0100\n"
+"PO-Revision-Date: 2025-08-26 15:46+0100\n"
 "Last-Translator: Aindriú Mac Giolla Eoin <aindriu80@gmail.com>\n"
 "Language-Team: Irish <gaeilge-gnulinux@lists.sourceforge.net>\n"
 "Language: ga\n"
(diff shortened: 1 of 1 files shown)


po/ga_IE.UTF-8.po
```diff
@@ -209,12 +209,12 @@ msgstr "Ceadaigh"
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-read.blp:81
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-write.blp:77
 msgid "Remember choice for this split"
-msgstr ""
+msgstr "Cuimhnigh ar an rogha don scoilt seo"
```
OshDubh: "cuimhnigh" is to remember in your mind, to recall, and doesn't map exactly to how "remember" is used in software in Englush

I don't know where this dialogue appears exactly, but depending on context;

If it's about saving a setting, `"Sábháil an rogha don scoilt seo"` might be more appropriate. 

If it's about keeping a choice for a current split, something like `"coiméad an rogha don scoilt seo"` perchance.

po/ga_IE.UTF-8.po
```diff
@@ -280,15 +280,15 @@ msgstr "Cóipeáilte chuig an ghearrthaisce"
 
 #: src/apprt/gtk/Surface.zig:1268
 msgid "Cleared clipboard"
-msgstr ""
+msgstr "Glanadh an ghearrthóg"
```
OshDubh: this more just refers to the act of clearing the clipboard, "cleaning of the clipboard" directly. 

`"Gearrthóg glanta"` would be a good way to inform a user that the clipboard has been cleared. 

You use "gearrthaisce" in other places to refer to the clipboard, and I would continue to use that word over "gearrthóg" here.

OshDubh (changes requested):
Nice work, a few changes.

po/ga_IE.UTF-8.po
aindriu80: I will make this change

po/ga_IE.UTF-8.po
aindriu80: Yes, I see 'gearrthaisce' used elsewhere, and there is only this one 'ghearrthóg'. I will change it o 'Gearrthaisce glanta'

OshDubh (approved):

trag1c (approved):

PR #8358, opened 2025-08-22
terminal: fix use-after-free in exec

recursiveforte:
This was only an issue on Linux, as MacOS' command is reallocated and rewritten. We hit this using embedded Ghostty w/o a login shell :p

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -1689,3 +1689,35 @@ test "execCommand: direct command, error passwd" {
     try testing.expectEqualStrings(result[0], "foo");
     try testing.expectEqualStrings(result[1], "bar baz");
 }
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Thank you. We should be able to unit test this, since `execCommand` has unit tests. That'd be good to do to prevent this in the future. I can get to it if you don't, just putting a red block on this so it isn't merged.

mitchellh:
Looks like we have some tests, but it's using static memory which is why this never triggered valgrind. We should be able to add another test that uses allocated memory, free it, and then access it.

recursiveforte:
Added that test!

mitchellh (approved):

PR #8359, opened 2025-08-22
terminal: fix build with -Di18n=false

recursiveforte:
canonicalizeLocale should return a null-terminated string, and didn't previously.

Compiler output:
```
src/os/i18n.zig:139:45: error: expected type 'error{NoSpaceLeft}![:0]const u8', found '[]const u8'
    if (comptime !build_config.i18n) return locale;
                                            ^~~~~~
src/os/i18n.zig:139:45: note: destination pointer requires '0' sentinel
src/os/i18n.zig:138:21: note: function return type declared here
) error{NoSpaceLeft}![:0]const u8 {
  ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
```

### DIFF
--- a/src/os/i18n.zig
+++ b/src/os/i18n.zig
@@ -136,7 +136,12 @@ pub fn canonicalizeLocale(
     buf: []u8,
     locale: []const u8,
 ) error{NoSpaceLeft}![:0]const u8 {
-    if (comptime !build_config.i18n) return locale;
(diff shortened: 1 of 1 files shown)


mitchellh:
Thanks, unfortunately this build option isn't currently in CI, so regressions like this will likely happen. Noted that in #8360.

PR #8363, opened 2025-08-23
Close other tabs feature on Mac.

jamylak:
Supporting command line, file menu and keybindings. Default mac shortcut of `super + alt + o` (other)

Not able to test on Linux so excluding `close_other_tabs` from `gtk` for now


mitchellh (changes requested):
I'm not opposed to this feature, I've seen some programs with it. I'm curious what macOS applications implement this so I can compare. Thank you.

trag1c:
> I'm curious what macOS applications implement this so I can compare.

Terminal.app has <kbd>cmd+opt+w</kbd>, not aware of others with keybinds. TextEdit and Safari definitely have right click options though.

mitchellh:
> Terminal.app has cmd+opt+w, not aware of others with keybinds. TextEdit and Safari definitely have right click options though.

I think for now the simplest thing would be to not have a default binding. I'll do that for now.

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -1023,6 +1028,32 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
         }
     }
 
+    @IBAction func closeOtherTabs(_ sender: Any?) {
```
mitchellh, line 1081: I still want to review this function. This implementation is a bit odd to me and it's unclear this behaves well with the undo system. 

It seems AI written, honestly, but I'm sorry if its not 😝

mitchellh (changes requested):
I made some changes, there's still one part I want to look into.

macos/Sources/Features/Terminal/TerminalController.swift
mitchellh: As a note: undo does work but it focuses the wrong tab on restore.

pluiedev:
Hmm.. I don't know if we want to introduce a new binding action for this. It needlessly broadens the implementation surface to apprt authors and results in largely duplicate code with the existing `close_tab` action. In this sense it's very similar to the `close_all_windows` action that we've already deprecated - it's specifically designed to implement an idiom that only exists on macOS and nowhere else.

IMO it would be much easier if we accept an optional target parameter in the existing `close_tab` action - so that it would be more like `close_tab:others`. Then once we assign IDs to tabs and surfaces for #2353 purposes, we can naturally extend this to accept IDs as well.

mitchellh:
> It needlessly broadens the implementation surface to apprt authors and results in largely duplicate code with the existing close_tab action.

At the moment the code is very much non-duplicated on macOS (I'm working on it now). In particular, the method by which we check for confirmation and undo/redo implementation means almost no code is shared. So I'm not sure about this argument. 

It increases the apprt surface area but also lets apprts implement the simpler stuff first.

> IMO it would be much easier if we accept an optional target parameter in the existing close_tab action - so that it would be more like close_tab:others. Then once we assign IDs to tabs and surfaces for https://github.com/ghostty-org/ghostty/discussions/2353 purposes, we can naturally extend this to accept IDs as well.

I'm also fine with this, and agree. I recommend we back into this with a deprecation. We don't currently have a deprecation system (such as that in our config system) for bindings, but we'll have to get there anyways and I don't mind using it when we do.

macos/Sources/Features/Terminal/TerminalController.swift
mitchellh: Cleaned it up and made undo/redo work properly.

mitchellh (approved):
Okay, I cleaned it up and made undo/redo work properly on macOS. 

I also fully support deprecating this binding into a parameterized `close_tab` in the future.

mitchellh:
I removed the menu item. I'm not finding this is typical on macOS applications.

I also note another bug, which exists regardless of this PR: when you "close other tabs" from the native tab context menu (built-in to macOS), undo doesn't refocus the tab properly. I'm not sure if this can be easily fixed but I'm going to make an issue to track it anyways.

macos/Sources/Features/Terminal/TerminalController.swift
jamylak: > I still want to review this function. This implementation is a bit odd to me and it's unclear this behaves well with the undo system.
> 
> It seems AI written, honestly, but I'm sorry if its not 😝

haha you are right about the swift `closeOtherTabs` part being AI written. Sorry about that. I should have spent some extra time to review the AI stuff but my excuse was that it was late on Saturday night, then some worries started to come into my head the next day but was out with family. Thanks for cleaning up that part I will try and learn some more swift from your adjustments

PR #8372, opened 2025-08-24
port to Zig 0.15

### LINKED ISSUE #8361: Zig 0.15 migration
[Zig 0.15](https://ziglang.org/download/0.15.1/release-notes.html) has been released — though due to its *massive* breaking changes regarding the standard library (defaulting to unmanaged lists, Writergate, etc.) we're staying with 0.14.1 for the 1.2 release (and any patch updates thereof).

PR over at #8372

As usual, we need to update our dependencies:

- [X] `libxev` - Updated to 0.15 with https://github.com/mitchellh/libxev/pull/180
- [x] `libvaxis` - Main branch has been updated to Zig 0.15 to a state where *our* use cases work 
- [X] `z2d` - Has a 0.15 branch: https://github.com/vancluever/z2d/tree/zig-0.15.0
- [ ] `zig-objc` - https://github.com/mitchellh/zig-objc/pull/26
- [x] `zig-js` - https://github.com/mitchellh/zig-js/pull/7
- [X] `ziglyph` - Patched branch over at https://codeberg.org/pluiedev/ziglyph/src/branch/zig-0.15
- [X] `zig-wayland` - Updated to 0.15 as of [0.5.0-dev](https://codeberg.org/ifreund/zig-wayland/commit/1b5c038ec10da20ed3a15b0b2a6db1c21383e8ea)
- [ ] `zf` - https://github.com/natecraddock/zf/pull/81
- [x] `zig-gobject` - https://github.com/ianprime0509/zig-gobject/pull/119

## Other issues (non-blocking)

- [X] Update zig2nix to 0.15 (#8370)

pluiedev:
Or: "How I learned to stopped worrying and love `std.Io.Writer.Allocating`"

Progress:
- [X] Build system compiles
- [X] libghostty compiles
- [x] Test suite compiles
- [x] Test suite all green
- [x] Benchmarks compile
- [ ] Benchmarks run and perform well against 0.14
- [x] All CLI commands compile
- [x] GTK-NG + OpenGL compiles
- [x] macOS app + Metal compiles
  - Blocked by zig-objc
  - I don't have a Mac ;)
- [x] Packaging manifests/metadata/instructions updated
- [x] CI all green
  - FreeBSD builds disabled for now since Ports has yet to update to 0.15

Closes #8361


src/apprt/gtk/App.zig
```diff
@@ -101,7 +101,7 @@ transient_cgroup_base: ?[]const u8 = null,
 css_provider: *gtk.CssProvider,
 
 /// Providers for loading custom stylesheets defined by user
-custom_css_providers: std.ArrayListUnmanaged(*gtk.CssProvider) = .{},
+custom_css_providers: std.ArrayList(*gtk.CssProvider) = .{},
```
jcollie: `.empty` for consistency

pluiedev: It's part of the automated ArrayListUnmanaged -> ArrayList migration that I did through a literal mass search-and-replace operation lol, I'll fix this later

src/config/Config.zig
```diff
@@ -3143,7 +3143,7 @@ _conditional_set: std.EnumSet(conditional.Key) = .{},
 /// The steps we can use to reload the configuration after it has been loaded
 /// without reopening the files. This is used in very specific cases such
 /// as loadTheme which has more details on why.
-_replay_steps: std.ArrayListUnmanaged(Replay.Step) = .{},
+_replay_steps: std.ArrayList(Replay.Step) = .{},
```
jcollie, line 3242: Another one. Muahahaha!

src/config/Config.zig
```diff
@@ -5124,7 +5141,7 @@ pub const RepeatableString = struct {
     const Self = @This();
 
     // Allocator for the list is the arena for the parent config.
-    list: std.ArrayListUnmanaged([:0]const u8) = .{},
+    list: std.ArrayList([:0]const u8) = .{},
```
jcollie, line 5308: Three! Three! Three places `.empty` should be used! Ah ah ah ah!

jcollie:
Upstream on zig-gobject is working on porting to Zig 0.15: https://github.com/ianprime0509/zig-gobject/pull/119

pluiedev:
image

IT LIVES

pluiedev:
Cc @ghostty-org/font @danielwe would appreciate some help regarding this crash:

```
thread 216789 panic: access of union field 'adjustment' while field 'scale' is active
/home/leah/coding/ghostty/src/font/Collection.zig:229:39: 0x2d0905f in getFaceFromEntry (ghostty)
                    entry.scale_factor.adjustment,
                                      ^
/home/leah/coding/ghostty/src/font/Collection.zig:197:37: 0x2d0bdd6 in getFace (ghostty)
    return try self.getFaceFromEntry(try self.getEntry(index));
                                    ^
/home/leah/coding/ghostty/src/font/Collection.zig:594:42: 0x2d0a8f8 in scaleFactor (ghostty)
        const primary_face = self.getFace(.{ .idx = 0 }) catch return 1.0;
                                         ^
/home/leah/coding/ghostty/src/font/Collection.zig:129:42: 0x2d1102f in add (ghostty)
    const scale_factor = self.scaleFactor(
                                         ^
/home/leah/coding/ghostty/src/font/SharedGridSet.zig:260:18: 0x2d0e747 in collection (ghostty)
    _ = try c.add(
                 ^
/home/leah/coding/ghostty/src/font/SharedGridSet.zig:133:36: 0x2d12667 in ref (ghostty)
        var c = try self.collection(&key, font_size, config);
                                   ^
/home/leah/coding/ghostty/src/Surface.zig:451:69: 0x304435c in init (ghostty)
    const font_grid_key, const font_grid = try app.font_grid_set.ref(
                                                                    ^
/home/leah/coding/ghostty/src/apprt/gtk/class/surface.zig:2643:21: 0x2fce4f0 in initSurface (ghostty)
        surface.init(
                    ^
/home/leah/coding/ghostty/src/apprt/gtk/class/surface.zig:2591:25: 0x2fcda5c in glareaResize (ghostty)
        self.initSurface() catch |err| {
```

Looks like `entry.scale_factor` here got initialized with undefined memory (more specifically, the *tag* got initialized just fine, but the *value* was undefined), and that's causing Zig to freak out. This only occurs in debug mode, curiously enough, so I suspect we're not initializing some fields where we should be, but I can't pinpoint where it is since I'm not that familiar with the font stack. Any help would be greatly appreciated 🙏🏼 

EDIT: Looks like this is impacting macOS tests too.

qwerasd205:
That error looks like a deferred face is being added to a collection without an explicit size adjustment, so it's defaulting to a scale factor instead -- we should probably actually just properly handle that case, use the scale for the deferred face if that's how it's specified.

qwerasd205:
Just adding an `if` before computing (at the location of the error there), like this:
```zig
if (entry.scale_factor == .adjustment)
    entry.scale_factor = .{
        .scale = self.scaleFactor(
            face.getMetrics(),
            entry.scale_factor.adjustment,
        ),
    };
```
should work (though maybe a switch would be nicer)

qwerasd205:
Looking closer at that trace, actually, is it trying to scale the primary face against itself maybe?

qwerasd205:
Ok, really funny behavior here-- when you have
```zig
some_union = .{
    .tag = fn_call(),
};
```
the tag is mutated on some_union before the fn_call is invoked. That's what's happening here. This behavior changed between 0.14.1 and 0.15.x. Here's a minimal repro:
```zig
const std = @import("std");

const T = union(enum) {
    a: u32,
    b: [4]u8,
};

fn doit(x: T) [4]u8 {
    std.debug.print("{}\n", .{x});
    return "erm?".*;
}

pub fn main() void {
    var x: T = .{
        .a = @bitCast([4]u8{'w', 'h', 'a', 't'}),
    };
    std.debug.print("{}\n", .{x});
    x = .{
        .b = doit(x),
    };
}
```

Found the relevant Zig issue: https://github.com/ziglang/zig/issues/24627

pluiedev:
what the hell

qwerasd205:
So the solution is that we need to calculate the scale factor first and then set it on the entry, smth like
```zig
if (entry.scale_factor == .adjustment) {
    const scale =  self.scaleFactor(
        face.getMetrics(),
        entry.scale_factor.adjustment,
    );
    entry.scale_factor = .{
        .scale = scale,
    };
}
```

pluiedev:
image

CI green :tada:

charlesrocket (commented):
Not seeing anything that would break on FBSD from the first read, though I haven't tested new deps. For FBSD CI, we could switch to binary pulls for now. I was thinking of pushing something like

```sh
ZIG_VERSION_LCL=0.15.1
cd /tmp
wget https://ziglang.org/download/$ZIG_VERSION_LCL/zig-x86_64-freebsd-$ZIG_VERSION_LCL.tar.xz
tar -xf zig-x86_64-freebsd-$ZIG_VERSION_LCL.tar.xz
alias zig=/tmp/zig-x86_64-freebsd-$ZIG_VERSION_LCL/zig
```
```git
+  #lang/zig \

mitchellh:
#9004

PR #8376, opened 2025-08-24
gtk-ng: fix setting/unsetting of urgency

jcollie:
- Don't set urgency on windows that are the topmost window.
- Turn off urgency on windows that become the topmost window.

Fixes #8373

### DIFF
--- a/src/apprt/gtk-ng/class/window.zig
+++ b/src/apprt/gtk-ng/class/window.zig
@@ -987,6 +987,22 @@ pub const Window = extern struct {
         };
     }
 
+    fn propIsActive(
+        _: *gtk.Window,
+        _: *gobject.ParamSpec,
+        self: *Self,
+    ) callconv(.c) void {
+        // Don't change urgency if we're not the active window.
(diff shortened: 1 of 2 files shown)


zquestz:
Appreciate the quick fix. 😀

src/apprt/gtk-ng/class/window.zig
```diff
@@ -1152,6 +1168,14 @@ pub const Window = extern struct {
             return;
         }
 
+        _ = gobject.Object.signals.notify.connect(
+            self.as(gtk.Window),
+            *Self,
+            propIsActive,
+            self,
+            .{ .detail = "is-active" },
+        );
```
mitchellh: We can register this in the blueprint file, can't we? `winproto` initializes to `noop` if I recall so even if the is-active gets triggered before realize it should do nothing.

mitchellh (changes requested):

src/apprt/gtk-ng/class/window.zig

mitchellh (approved):

PR #8380, opened 2025-08-25
Implement "close other tabs" on gtk-ng and parameterize `close_tab`

jcollie:
Follow-up to #8363

- Implements "close other tabs" on gtk-ng
- Removes the `close_other_tabs` keybinding and apprt action and replaces it with a parameterised `close_tab` keybinding and apprt action.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -680,6 +680,12 @@ typedef struct {
   uintptr_t len;
 } ghostty_action_open_url_s;
 
+// apprt.action.CloseTabMode
+typedef enum {
+  GHOSTTY_ACTION_CLOSE_TAB_MODE_THIS,
+  GHOSTTY_ACTION_CLOSE_TAB_MODE_OTHER,
+} ghostty_action_close_tab_mode_e;
+
 // apprt.surface.Message.ChildExited
 typedef struct {
   uint32_t exit_code;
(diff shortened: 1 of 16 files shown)


mitchellh (changes requested):
This doesn't resolve the issue that this is an unnecessary breaking change. Bindings with `x=close_tab` would have to change to `x=close_tab:this`, if I'm reading this correctly.

Please PR the two major changes separately. I think the scope of this is too large. The changes to `gtk-ng` to implement `close_other_tabs` look great. If those were isolated I can merge immediately.

pluiedev:
Is there support for optional parameters for the action parser at the moment? My idea was to make `close_tab` equivalent to `close_tab:this` so that we don't break backwards compatibility

mitchellh:
> My idea was to make close_tab equivalent to close_tab:this so that we don't break backwards compatibility

That'd be fine for now. I think longer term I want to do a backwards compatibility system similar to config for bindings, but this would suffice for this specific case. 

I still think this PR should be split though. PRs that focus on one thing always make review and merge must faster.

mitchellh:
FYI I can't remember if binding actions support custom parse logic either. So there might be some other work to do there, but if so we should follow a model identical to config parsing.

jcollie:
Replacing with #8389.

jcollie:
> Is there support for optional parameters for the action parser at the moment? My idea was to make `close_tab` equivalent to `close_tab:this` so that we don't break backwards compatibility

Yes, this is easily possible. The keybind action parsed already supported it, I just didn't think to look.

PR #8389, opened 2025-08-25
parameterize close_tab

jcollie:
- Add mode (`this`/`other`) parameter to `close_tab` keybind/apprt action.
- Keybinds will default to `this` if not specified, eliminating backward compatibility issues (`keybind=x=close_tab` === `keybind=x=close_tab:this`).
- Remove `close_other_tabs` keybind and apprt action.

Replaces #8380

### DIFF
--- a/src/apprt/gtk-ng/class/tab.zig
+++ b/src/apprt/gtk-ng/class/tab.zig
@@ -352,12 +352,8 @@ pub const Tab = extern struct {
         // Delegate to our parent to handle this, since this will emit
         // a close-page signal that the parent can intercept.
         switch (mode) {
-            .this => {
-                tab_view.closePage(page);
-            },
-            .other => {
-                log.warn("close-tab:other is not implemented", .{});
-            },
+            .this => tab_view.closePage(page),
+            .other => log.warn("close-tab:other is not implemented", .{}),
         }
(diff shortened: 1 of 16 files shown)


src/apprt/gtk-ng/class/tab.zig
```diff
@@ -314,18 +317,48 @@ pub const Tab = extern struct {
...
+            log.warn("invalid mode provided to tab.close-tab: {s}", .{str.?});
+            return;
+        };
+
         // Delegate to our parent to handle this, since this will emit
         // a close-page signal that the parent can intercept.
-        tab_view.closePage(page);
+        switch (mode) {
+            .this => {
+                tab_view.closePage(page);
```
mitchellh: Style: get rid of the block here and below, just make it a single expression

mitchellh (commented):
Looks good. I just want to verify macOS continues to work then this is good to go.

src/apprt/gtk-ng/class/tab.zig
jcollie: Fixed.

mitchellh (approved):

mitchellh:
Verified macOS is good.

PR #8390, opened 2025-08-25
apprt/gtk-ng: show error widget if GLArea fails to initialize

mitchellh:
If GTK can't acquire an OpenGL context, this shows a message. Previously, we would only log a warning which was difficult to find. The GUI previously was the default GTK view which showed "Failed to acquire EGL display" which was equally confusing.

This is a draft. There are TODOs (listed below).

## TODO

- [x] Disable context menu in error state
- [x] Use property to bind to unhealthy state instead of directly setting stack child
- [x] Create a web page and put that in the error description
- [x] Set non-transparent background in error state
- [x] Bug where closing the window isn't exiting Ghostty

### DIFF
--- a/src/apprt/gtk-ng/class/application.zig
+++ b/src/apprt/gtk-ng/class/application.zig
@@ -116,6 +116,11 @@ pub const Application = extern struct {
         /// and initialization was successful.
         transient_cgroup_base: ?[]const u8 = null,
 
+        /// This is set to true so long as we request a window exactly
+        /// once. This prevents quitting the app before we've shown one
+        /// window.
+        requested_window: bool = false,
+
         /// This is set to false internally when the event loop
         /// should exit and the application should quit. This must
         /// only be set by the main loop thread.
@@ -461,7 +466,13 @@ pub const Application = extern struct {
                 // If the quit timer has expired, quit.
                 if (priv.quit_timer == .expired) break :q true;
 
-                // There's no quit timer running, or it hasn't expired, don't quit.
+                // If we have no windows attached to our app, also quit.
+                if (priv.requested_window and @as(
+                    ?*glib.List,
+                    self.as(gtk.Application).getWindows(),
+                ) == null) break :q true;
+
+                // No quit conditions met
                 break :q false;
             };
 
@@ -1858,6 +1869,13 @@ const Action = struct {
         self: *Application,
         parent: ?*CoreSurface,
     ) !void {
+        // Note that we've requested a window at least once. This is used
+        // to trigger quit on no windows. Note I'm not sure if this is REALLY
+        // necessary, but I don't want to risk a bug where on a slow machine
+        // or something we quit immediately after starting up because there
+        // was a delay in the event loop before we created a Window.
+        self.private().requested_window = true;
+
         const win = Window.new(self);
         initAndShowWindow(self, win, parent);
     }
--- a/src/apprt/gtk-ng/class/surface.zig
+++ b/src/apprt/gtk-ng/class/surface.zig
@@ -105,6 +105,24 @@ pub const Surface = extern struct {
             );
         };
 
+        pub const @"error" = struct {
+            pub const name = "error";
+            const impl = gobject.ext.defineProperty(
+                name,
+                Self,
+                bool,
+                .{
+                    .default = false,
+                    .accessor = gobject.ext.privateFieldAccessor(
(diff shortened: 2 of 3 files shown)


src/apprt/gtk-ng/ui/1.2/surface.blp
```diff
@@ -20,18 +20,38 @@ template $GhosttySurface: Adw.Bin {
...
+          };
+        }
+
+        StackPage {
+          name: "error";
+
+          child: Adw.StatusPage {
+            icon-name: "computer-fail-symbolic";
+            title: _("Uh Oh");
+            description: _("The terminal has crashed.");
```
mitchellh: This copy is very much a placeholder. See the PR description for what I want to do here.

mitchellh:
Unsure currently why the `LinkButton` is not clickable. But pushed up changes addressing some of the TODOs.

mitchellh:
Okay this is very clean and ready for review, but have to change the copy still.

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -2303,21 +2369,23 @@ pub const Surface = extern struct {
...
 
+        // Make the GL area current so we can detect any OpenGL errors. If
+        // we have errors here we can't render and we switch to the error
+        // state.
+        const priv = self.private();
+        priv.gl_area.makeCurrent();
+        if (priv.gl_area.getError()) |err| {
+            log.warn("failed to make GL context current: {s}", .{err.f_message orelse "(no message)"});
+            log.warn("this error is usually due to a driver or gtk bug", .{});
+            log.warn("this is a common cause of this issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/4950", .{});
```
jcollie: This should probably be the same link as in the error page, as that's not the only bug that can cause the issue.

jcollie (approved):
LGTM, pending website and copy updates.

src/apprt/gtk-ng/ui/1.2/surface.blp
```diff
@@ -8,146 +8,172 @@ template $GhosttySurface: Adw.Bin {
...
-      halign: start;
-      valign: end;
-      label: bind template.mouse-hover-url;
-
-      EventControllerMotion url_ec_motion {
-        enter => $url_mouse_enter();
-        leave => $url_mouse_leave();
-      }
+        child: LinkButton {
+          label: "https://ghostty.org/docs/help/gtk-opengl-context";
```
jcollie, line 168: Maybe instead of using the URL as the label, maybe something like "Get more information…" or something like that?

mitchellh: I wanted to use the label just in case they had a broken mouse or input event like I coincidentally did at one point. They can at least type it out. Since this is a pretty bad situation already it’s possible things are worse. Thoughts?

jcollie: Yeah, that's valid. Maybe some text that says "For more information, click here: <linkbutton>" with the URL as the label of the button.

mitchellh: I was also trying to avoid more translations :) But let me play around with it.

jcollie: in for a penny, in for a pound…

mitchellh: I played with a few things but I'm pretty happy with the current look and feel. I think its pretty obvious in multiple ways thi leads to a help page.

mitchellh:
Sorry @trag1c this introduces two new translation strings. They're extremely unimportant though (rare error state).

src/apprt/gtk-ng/class/application.zig
```diff
@@ -461,7 +466,13 @@ pub const Application = extern struct {
                 // If the quit timer has expired, quit.
                 if (priv.quit_timer == .expired) break :q true;
 
-                // There's no quit timer running, or it hasn't expired, don't quit.
+                // If we have no windows attached to our app, also quit.
+                if (priv.requested_window and @as(
+                    ?*glib.List,
+                    self.as(gtk.Application).getWindows(),
```
mitchellh, line 472: Here is another GIR issue btw @jcollie if you're interested in fixing this upstream. :)

PR #8393, opened 2025-08-25
cli: show colors in +list-colors if possible

### LINKED ISSUE #8386: Improve `+list-colors` to show actual colors visually
### Discussed in https://github.com/ghostty-org/ghostty/discussions/3064

<sup>Originally posted by **anund** December 21, 2024</sup>
Following on from the work in https://github.com/ghostty-org/ghostty/pull/2052 it would be nice to add color and additional formatting to `+list-colors`. At the very least printing the color needs output a colored result.

Here's an attempt:
2024-12-22-181102_1159x249_scrot

- the name sorted output has groupings of colors that print more nicely down a column than across a row
- there are a lot of grey colors, they can't really be grouped to not break across from one column to the other
- terminal width controls the number of columns, which controls when a color group (color1, color2) might be split across columns 
- the ascii `■` prints across 2 columns with the font in the screenshot, overlapping the following space. Is there something better? 
- no idea how this looks on windows
- the longest color name is `light goldenrod yellow` leading to the space between names and `=`
- a `--plain` flag and tty checking are in place following the way `+list-keybinds` works

Suggestions for formatting?</div>

jcollie:
Fixes #8386

This is a fairly simple implementaion, there's no interactivity or searching. It will adapt the number of columns to the available width of the display though.

Will fallback to a plain text dump if there's no tty or the `--plain` argument is specified on the CLI.

image

### DIFF
--- a/src/cli/list_colors.zig
+++ b/src/cli/list_colors.zig
@@ -1,12 +1,19 @@
 const std = @import("std");
+const builtin = @import("builtin");
+const Allocator = std.mem.Allocator;
 const Action = @import("ghostty.zig").Action;
 const args = @import("args.zig");
 const x11_color = @import("../terminal/main.zig").x11_color;
+const vaxis = @import("vaxis");
+const tui = @import("tui.zig");
 
 pub const Options = struct {
     pub fn deinit(self: Options) void {
         _ = self;
     }
+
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
I'm good with this, looks great to me!

PR #8410, opened 2025-08-26
osc: parse OSC 9;6 gui macros

jcollie:


src/terminal/osc.zig
```diff
@@ -1600,6 +1624,12 @@ pub const Parser = struct {
             .hyperlink_uri => self.endHyperlink(),
             .string => self.endString(),
             .conemu_sleep_value => self.endConEmuSleepValue(),
+            .conemu_guimacro => {
```
mitchellh, line 1627: Looks like we might be missing this for other ConEmu incomplete values. Do you want to add it for those too? I don't think its all of them.

jcollie: Yeah, I was going to follow up with a PR or two that fixed up the existing ones (as well as add tests).

mitchellh: Yeah would normally prefer that but it's so simple. But I'll merge this now. Its good!

mitchellh (approved):

PR #8412, opened 2025-08-26
config: add entry for scroll-to-bottom

### LINKED ISSUE #8408: Add `scroll-to-bottom` configuration to control scrolling behavior
From #5573 

Add a new `scroll-to-bottom` configuration that is a `packed struct` of bools that contains `keystroke`, `output` to mimic these Ptyxis configurations:

Image

The separate behavior can be PR'd separately.

jcollie:
Related #8408

(EDIT @mitchellh: Removed "Fixes" to avoid closing)

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -272,6 +272,7 @@ const DerivedConfig = struct {
     title_report: bool,
     links: []Link,
     link_previews: configpkg.LinkPreviews,
+    scroll_to_bottom_on_keystroke: bool,
 
     const Link = struct {
         regex: oni.Regex,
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
This should have the paired implementation. If you want to split them up that's fine but at least include the one that is implemented. We don't usually accept bare configs on their own.

jcollie:
Added the implementation for `keystroke`.

src/Surface.zig
```diff
@@ -272,6 +272,7 @@ const DerivedConfig = struct {
     title_report: bool,
     links: []Link,
     link_previews: configpkg.LinkPreviews,
+    scroll_to_bottom_on_keystroke: bool,
```
mitchellh: We usually just store the full config struct. I think that's what we should do here, since we'll need it for future values.

mitchellh (changes requested):
One small change, rest looks good!

src/Surface.zig

mitchellh (approved):

PR #8423, opened 2025-08-27
macos: use visible frame for quick terminal sizing calculation

mitchellh:
Fixes #8418

This fixes issues where left/right positions would be cut off from the menu bar. And makes it so that size 100%,100% doesn't overflow into the non-visible space of the edge of the screen.

I didn't just copy and paste, I tested each of these code paths.


krono:
Thanks!

PR #8427, opened 2025-08-27
apprt/gtk-ng: "cancel" should be default close response for dialog

mitchellh:
Fixes #8424

This is the response that will be sent when "escape" is pressed.

This also fixes a null ptr deref that was possible when this fix wasn't in.


src/apprt/gtk-ng/ui/1.2/close-confirmation-dialog.blp
```diff
@@ -7,4 +7,6 @@ template $GhosttyCloseConfirmationDialog: $GhosttyDialog {
     cancel: _("Cancel"),
     close: _("Close") destructive,
   ]
+
+  close-response: "cancel";
```
mitchellh, line 11: This is the real fix.

jcollie (approved):
LGTM

PR #8442, opened 2025-08-28
macOS: Always require confirmation when executing a script via `open` (Finder, etc.)

mitchellh:
Right now, passing a file path to Ghostty will always execute it unconditionally. This has various risks associated with it. I think we can mitigate a lot of risks in the future by inspecting what is being executed, but to be safe now we should always ask for confirmation.

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -394,18 +394,29 @@ class AppDelegate: NSObject,
         // Ghostty will validate as well but we can avoid creating an entirely new
         // surface by doing our own validation here. We can also show a useful error
         // this way.
-
+        
         var isDirectory = ObjCBool(true)
         guard FileManager.default.fileExists(atPath: filename, isDirectory: &isDirectory) else { return false }
-
+        
+        // Set to true if confirmation is required before starting up the
+        // new terminal.
+        var requiresConfirm: Bool = false
+        
         // Initialize the surface config which will be used to create the tab or window for the opened file.
         var config = Ghostty.SurfaceConfiguration()
-
+        
         if (isDirectory.boolValue) {
             // When opening a directory, check the configuration to decide
             // whether to open in a new tab or new window.
             config.workingDirectory = filename
         } else {
+            // Unconditionally require confirmation in the file execution case.
+            // In the future I have ideas about making this more fine-grained if
+            // we can not inherit of unsandboxed state. For now, we need to confirm
+            // because there is a sandbox escape possible if a sandboxed application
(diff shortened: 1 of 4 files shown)


src/apprt/embedded.zig
```diff
@@ -447,6 +447,9 @@ pub const Surface = struct {
 
         /// Input to send to the command after it is started.
         initial_input: ?[*:0]const u8 = null,
+
+        // Wait after the command exits
```
tristan957:
```suggestion
        /// Wait after the command exits
```

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -394,35 +394,69 @@ class AppDelegate: NSObject,
         // Ghostty will validate as well but we can avoid creating an entirely new
         // surface by doing our own validation here. We can also show a useful error
         // this way.
-
+
```
tristan957, line 397: Seems like there is a bit of trailing space?

mitchellh: Yeah we don't auto format Swift yet. We're working on it. Until then, there's a LOT of this.

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -394,35 +394,69 @@ class AppDelegate: NSObject,
...
         
+        if requiresConfirm {
+            // Confirmation required. We use an app-wide NSAlert for now. In the future we
+            // may want to show this as a sheet on the focused window (especially if we're
+            // opening a tab). I'm not sure.
+            let alert = NSAlert()
+            alert.messageText = "Allow Ghostty to execute \"\(filename)\"?"
+            alert.addButton(withTitle: "Allow")
+            alert.addButton(withTitle: "Cancel")
+            alert.alertStyle = .warning
```
mitchellh, line 445: I'd really like to use an `accessoryView` here to add a bunch of details but accessory views don't work with SwiftUI NSHostingViews and that's annoying. Submitted to Apple as FB19958056.

sampl3x:
Can the warning be disabled? I run script .sh files from Alfred and now i get this annoying popup with v1.20

00-kat:
Open a [new discussion].

[new discussion]: https://github.com/ghostty-org/ghostty/discussions/new?category=feature-requests-ideas

ManuelSchneid3r:
@sampl3x's usecase is totally valid. Please add means to relax this.

sampl3x:
I switched away from Ghostty because of this and other things

trag1c:
> @sampl3x's usecase is totally valid. Please add means to relax this.

As noted by Kat, open a new discussion. Locking as it looks like you two can't read.

PR #8445, opened 2025-08-28
docs: reorganize and separate out HACKING.md for technical instructions

pluiedev:
Also did some copy-editing and removed some things that haven't been present in the codebase for a while (I thought we nuked conformance stuff in 1.1 already...?)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,4 +137,3 @@ pull request will be accepted with a high degree of certainty.
 > **Pull requests are NOT a place to discuss feature design.** Please do
 > not open a WIP pull request to discuss a feature. Instead, use a discussion
 > and link to your branch.
-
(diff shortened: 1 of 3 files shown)


jcollie (commented):
Maybe instead of a new file, this should be merged into `CONTRRIBUTING.md`?

pluiedev:
There's a lot of overlap for sure... Personally I think CONTRIBUTING.md is already *way* too large and I think we can move the technical parts of it over to HACKING.md, and then keep CONTRIBUTING.md as the document that more describes the *process* than the exact technical details.

mitchellh:
> There's a lot of overlap for sure... Personally I think CONTRIBUTING.md is already _way_ too large and I think we can move the technical parts of it over to HACKING.md, and then keep CONTRIBUTING.md as the document that more describes the _process_ than the exact technical details.

Haven't reviewed yet, but I agree with this assessment. 

CONTRIBUTING should be as small as possible to hopefully encourage as many contributors as possible to read it and follow the rules. It should be focused primarily on guidelines to contribution and not really _how to dev_ itself.

mitchellh (approved):

PR #8448, opened 2025-08-28
gtk-ng: deprecate gtk-single-instance=desktop

jcollie:
This fixes the problem of running a bare `ghostty` at the CLI (or from any of the numerous launchers used by tiling window managers) creating a new instance of Ghostty rather than sending a signal to an existing Ghostty instance to open up a new window.

This happened because when `gtk-single-instance=desktop` and you ran `ghostty` from the CLI, it was treated as `gtk-single-instance=false` which would create a new Ghostty instance.

The `desktop` option was introduced before we had the `launched-from` option. Now that `launched-from` is available, the `desktop` option is no longer necessary (and sometimes harmful).

`ghostty +new-window` will still be slightly faster at getting a new window since it does less initialization before it sends the signal to the existing process but this will eliminate _a lot_ of confusion from users.

### DIFF
--- a/src/apprt/gtk-ng/class/application.zig
+++ b/src/apprt/gtk-ng/class/application.zig
@@ -223,10 +223,6 @@ pub const Application = extern struct {
         const single_instance = switch (config.@"gtk-single-instance") {
             .true => true,
             .false => false,
-            .desktop => switch (config.@"launched-from".?) {
-                .desktop, .systemd, .dbus => true,
-                .cli => false,
-            },
         };
 
         // Setup the flags for our application.
--- a/src/apprt/gtk/App.zig
+++ b/src/apprt/gtk/App.zig
@@ -280,10 +280,6 @@ pub fn init(self: *App, core_app: *CoreApp, opts: Options) !void {
     const single_instance = switch (config.@"gtk-single-instance") {
         .true => true,
         .false => false,
-        .desktop => switch (config.@"launched-from".?) {
-            .desktop, .systemd, .dbus => true,
-            .cli => false,
-        },
     };
 
     // Setup the flags for our application.
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2925,14 +2925,14 @@ else
 ///
 /// If `false`, each new ghostty process will launch a separate application.
 ///
-/// The default value is `desktop` which will default to `true` if Ghostty
-/// detects that it was launched from the `.desktop` file such as an app
-/// launcher (like Gnome Shell)  or by D-Bus activation. If Ghostty is launched
-/// from the command line, it will default to `false`.
+/// The pre-1.2 option `desktop` has been deprecated. If encountered it will be
+/// treated as `true`.
+///
+/// The default value is `true`.
 ///
 /// Note that debug builds of Ghostty have a separate single-instance ID
 /// so you can test single instance without conflicting with release builds.
-@"gtk-single-instance": GtkSingleInstance = .desktop,
+@"gtk-single-instance": GtkSingleInstance = .default,
 
 /// When enabled, the full GTK titlebar is displayed instead of your window
 /// manager's simple titlebar. The behavior of this option will vary with your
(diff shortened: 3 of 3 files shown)


pluiedev (approved):
Makes a lot of sense once I wrapped my head around it

mitchellh:
I’ll have to wrap my own head around it. But when running Ghostty from the CLI this will result in a new instance right?

jcollie:
> I’ll have to wrap my own head around it. But when running Ghostty from the CLI this will result in a new instance right?

Before, running `ghostty` from the CLI would _always_ result in a new instance because `gtk-single-instance=desktop` implied `gtk-single-instance=false` _on the CLI only_.

Now with this PR, running `ghostty` from the CLI will create a new instance _if there is not one already running_. If there _is_ already an instance, a signal will be sent to running instance to create a new window and the CLI command will exit.

mitchellh:
This unfortunately is not desired. It’s how Ghostty worked in beta for a long time and it produced a flurry of issues, namely because it’s extremely common for CLI users to pass custom config in and reusing the existing instance will ignore all that. 

Additionally, it’s also often desired for CLI invocations to block until Ghostty quits again, eg if a script opens the terminal to review a file or something before continuing. 

One of my thoughts was maybe to flip the desktop logic on its head and only create a new instance if we can detect we were executed from a terminal itself (eg check for things like the TERM PROGRAM env var and emit detection sequences).

jcollie:
> This unfortunately is not desired. It’s how Ghostty worked in beta for a long time and it produced a flurry of issues, namely because it’s extremely common for CLI users to pass custom config in and reusing the existing instance will ignore all that.
>
> Additionally, it’s also often desired for CLI invocations to block until Ghostty quits again, eg if a script opens the terminal to review a file or something before continuing.
> 
> One of my thoughts was maybe to flip the desktop logic on its head and only create a new instance if we can detect we were executed from a terminal itself (eg check for things like the TERM PROGRAM env var and emit detection sequences).

I think that it would be better:

- User executes just `ghostty` - no extra config on the CLI - re-use an existing instance (unless `gtk-single-instance=false` in the config).
- User executes `ghostty --blahblahblah` - create a new instance (overriding any previous settings of `gtk-single-instance`)

The less auto-detect magic the better.

Once we can pass config options to the existing instance this can be revisited.

mitchellh:
> Once we can pass config options to the existing instance this can be revisited.

I'm happy if you want to reinvestigate this but my conclusion was that it is not possible. You can probably find my GH comment somewhere. Some of the configs people set are _app level_ and you can't have some windows act some way and others act other way for _app level_ configs (window and surface are of course fine). A good example are app-level keybinds.

I can't overstate how common this issue was before I added desktop detection. 

> I think that it would be better

I agree in less magic. The `-e` command already forces `gtk-single-instance=false` which helps a lot. And `-e` is begrudgingly magical and sets a half dozen or more configs because it's an ecosystem norm we have to bend ourselves to (lots of scripts parameterize `$TERMINAL -e foo`, including things like Omarchy!)

This doesn't solve the issue that ghostty would return immediately and I think the best behavior for 1.2 would be to try to mimic the 1.1 behavior here as much as possible, this just doesn't feel worth having any breaks for (which is why I'm asking for `ghostty` to just work for desktops too to begin with). 

The annoyance around this issue is already high, but it was generally worth it because the mechanism by which we have single instance now and the systemd units does provide a lot of quality of life. 

My suggestion would be: 

  - if there are CLI flags, always new instance
  - if `TERM_PROGRAM` is set, always new instance

We can walk back (or forward) on these changes as necessary, but I think it would keep us with 1.1.x behavior 99% of the time. And I'm pretty sure desktop environments aren't setting `TERM_PROGRAM` for their launcher configs which is the predominant use case for the single instance mode, as named by the prior `desktop` config. 😄 

I think adding back a `gtk-single-instance=detect` configuration and documenting this is worth it.

mitchellh:
Also I'm not sure if we do it right now, but Ghostty should emit logs to stdout/stderr if it's running in `launched-from=cli` mode. That was also a common point of issue. We can do that in a separate PR though.

(The logs can be in addition to journald if you want, or instead of, I don't think it makes a big difference)

jcollie:
> Also I'm not sure if we do it right now, but Ghostty should emit logs to stdout/stderr if it's running in `launched-from=cli` mode. That was also a common point of issue. We can do that in a separate PR though.
>
> (The logs can be in addition to journald if you want, or instead of, I don't think it makes a big difference)

Ghostty always sends logs to stdout/stderr. The only way that they end up in the journal is when Ghostty is launched by a  systemd unit (or launched by something that was launched by a systemd unit) because systemd captures stdout/stderr and redirects it to the journal.

We certainly _can_ send logs directly to journald, but we don't.

mitchellh (changes requested):
Just marking the changes requested pending the last feedback.

jcollie:
Closing this, going to take a different tack.

PR #8459, opened 2025-08-29
gtk-ng: ensure CSS works on both 4.14 and 4.16+

jcollie:
Ghostty 1.2 needs to support GTK 4.14 because that's the version that ships with Ubuntu 24.04.

This PR ensures that any GTK 4.16 CSS features are not used in any static CSS and that the runtime CSS loading handles both 4.14 and 4.16+ appropriately.

### DIFF
--- a/src/apprt/gtk-ng/class/application.zig
+++ b/src/apprt/gtk-ng/class/application.zig
@@ -732,27 +732,24 @@ pub const Application = extern struct {
         }
     }
 
-    fn loadRuntimeCss(
-        self: *Self,
-    ) Allocator.Error!void {
+    fn loadRuntimeCss(self: *Self) Allocator.Error!void {
         const alloc = self.allocator();
 
-        var buf: std.ArrayListUnmanaged(u8) = .empty;
+        const config = self.private().config.get();
+
+        var buf: std.ArrayListUnmanaged(u8) = try .initCapacity(alloc, 2048);
         defer buf.deinit(alloc);
 
         const writer = buf.writer(alloc);
 
-        const config = self.private().config.get();
-        const window_theme = config.@"window-theme";
         const unfocused_fill: CoreConfig.Color = config.@"unfocused-split-fill" orelse config.background;
-        const headerbar_background = config.@"window-titlebar-background" orelse config.background;
-        const headerbar_foreground = config.@"window-titlebar-foreground" orelse config.foreground;
 
         try writer.print(
             \\widget.unfocused-split {{
             \\ opacity: {d:.2};
             \\ background-color: rgb({d},{d},{d});
             \\}}
+            \\
         , .{
             1.0 - config.@"unfocused-split-opacity",
             unfocused_fill.r,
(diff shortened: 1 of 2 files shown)


src/apprt/gtk-ng/class/application.zig
```diff
@@ -778,9 +776,129 @@ pub const Application = extern struct {
...
+
+        // Clears any previously loaded CSS from this provider
+        loadCssProviderFromData(
+            self.private().css_provider,
+            data,
+        );
+    }
+
+    /// Load runtime CSS for older than GTK 4.16
+    fn loadRuntimeCss414(
```
tristan957, line 801:
```suggestion
    fn loadRuntimeCss4_14(
```

src/apprt/gtk-ng/class/application.zig
```diff
@@ -778,9 +776,129 @@ pub const Application = extern struct {
...
+                headerbar_background.r,
+                headerbar_background.g,
+                headerbar_background.b,
+            }),
+            else => {},
+        }
+    }
+
+    /// Load runtime for GTK 4.16 and newer
+    fn loadRuntimeCss416(
```
tristan957, line 837:
```suggestion
    fn loadRuntimeCss4_16(
```

src/apprt/gtk-ng/css/style.css
```diff
@@ -83,27 +83,21 @@ label.resize-overlay {
...
 .child-exited.normal revealer widget {
   background-color: rgba(38, 162, 105, 0.5);
-  /* after GTK 4.16 is a requirement, switch to the following:
-  /* background-color: color-mix(in srgb, var(--success-bg-color), transparent 50%); */
 }
 
 .child-exited.abnormal revealer widget {
   background-color: rgba(192, 28, 40, 0.5);
-  /* after GTK 4.16 is a requirement, switch to the following:
-  /* background-color: color-mix(in srgb, var(--error-bg-color), transparent 50%); */
```
tristan957: Why remove the comments? They seem useful.

src/apprt/gtk-ng/class/application.zig
jcollie: Zig functions don't generally use `_` AFAICS so I left these as-is.

src/apprt/gtk-ng/class/application.zig
jcollie: Zig functions don't generally use `_` AFAICS so I left these as-is.

src/apprt/gtk-ng/css/style.css
jcollie: I restored the comments.

src/apprt/gtk-ng/class/application.zig
```diff
@@ -778,9 +776,129 @@ pub const Application = extern struct {
...
                 \\}}
+                \\
             , .{ .font_family = font_family });
         }
 
+        try loadRuntimeCss414(config, &writer);
+        try loadRuntimeCss416(config, &writer);
+
+        // ensure that we have a sentinel
+        try writer.writeByte('\x00');
```
mitchellh: You can just do `writeByte(0)`

mitchellh (approved):
There's a small style thing, but the rest looks good to me. I'll leave it up to you what you want to do.

src/apprt/gtk-ng/class/application.zig
jcollie: Done.

PR #8462, opened 2025-08-30
feat: add `selection-clear-on-copy` configuration option

twfksh:
Addresses issue: Add selection-clear-on-copy configuration #8407

Added configuration option `selection-clear-on-copy` that matches with the `selection-clear-on-typing` option. 
And `copy-on-select` is ignored when `selection-clear-on-copy` is true regardless of whether `copy-on-select` is set to true or clipboard. 
Also `.copy_to_clipboard` binding action was refactored to use `copySelectionToClipboards` for consistent behavior.

> Consulted with Copilot (Claude Sonnet 4) to understand the control flow of copy operations and help write the docs. Solution was authored and implemented by me.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -258,6 +258,7 @@ const DerivedConfig = struct {
     mouse_shift_capture: configpkg.MouseShiftCapture,
     macos_non_native_fullscreen: configpkg.NonNativeFullscreen,
     macos_option_as_alt: ?configpkg.OptionAsAlt,
+    selection_clear_on_copy: bool,
     selection_clear_on_typing: bool,
     vt_kam_allowed: bool,
     wait_after_command: bool,
@@ -327,6 +328,7 @@ const DerivedConfig = struct {
             .mouse_shift_capture = config.@"mouse-shift-capture",
             .macos_non_native_fullscreen = config.@"macos-non-native-fullscreen",
             .macos_option_as_alt = config.@"macos-option-as-alt",
+            .selection_clear_on_copy = config.@"selection-clear-on-copy",
             .selection_clear_on_typing = config.@"selection-clear-on-typing",
             .vt_kam_allowed = config.@"vt-kam-allowed",
             .wait_after_command = config.@"wait-after-command",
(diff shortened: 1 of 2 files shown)


src/Surface.zig
```diff
@@ -1848,6 +1860,10 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
     // If copy on select is false then exit early.
     if (self.config.copy_on_select == .false) return;
 
+    // If copy on select not false and selection clear on copy is true
+    // then copy on select is ignored.
+    if (self.config.selection_clear_on_copy) return;
```
mitchellh: I think this is the only suspicious part to me. In my issue, I think I said that `selection-clear-on-copy` should not apply to `copy-on-select` and only apply to bindings. That is, we should not clear the section when copy on select is active because I think its a bit jarring to lose that immediately.

But `copy-on-select` should _still work_.

mitchellh (changes requested):
Thanks for the AI disclosure. This looks pretty neat, I didn't test it, but one comment.

src/Surface.zig
twfksh: Thanks for the follow up. My bad, I overlooked the keybinding part. I'll update it so that both `copy-on-select` and `selection-clear-on-copy` work as intended.

twfksh:
Now `selection-clear-on-copy` is performed only via keybindings and `copy-on-select` remains as is. 

*Apologies for any confusion* this is my first time contributing to a opensource software. I reset my last two commit on my feature-branch and did a force-push which I realized isn't a good practice. Trying to undo it I deleted the branch and after synchronized my fork with the upstream then re-commit the changes and pushed it back to the feature-branch, Which was probably unnecessary in this case.

src/Surface.zig
twfksh: I’ve updated the implementation so that selection-clear-on-copy only applies via keybindings, while copy-on-select behaves as expected. Please let me know if anything else needs adjustment!

mitchellh (approved):
Did some manual cleanups but looks great, thank you!

PR #8466, opened 2025-08-30
ci: add freebsd tests

jcollie:


jcollie:
@charlesrocket @arrowd

.github/workflows/test.yml
```diff
@@ -1015,3 +1016,57 @@ jobs:
...
+    needs: test
+    runs-on: namespace-profile-mitchellh-sm-systemd
+    strategy:
+      matrix:
+        release:
+          - "14.3"
+          # - "15.0" # disable until fixed: https://github.com/vmactions/freebsd-vm/issues/108
+    steps:
+      - name: Checkout Ghostty
+        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
```
arrowd, line 1031: Why not use the tag name there?

.github/workflows/test.yml
```diff
@@ -1015,3 +1016,57 @@ jobs:
...
+
+      - name: Start SSH
+        run: |
+          sudo systemctl start ssh
+
+      - name: Set up FreeBSD VM
+        uses: vmactions/freebsd-vm@05856381fab64eeee9b038a0818f6cec649ca17a # v1.2.3
```
arrowd, line 1038: Why not use the tag name there?

.github/workflows/test.yml
```diff
@@ -1015,3 +1016,57 @@ jobs:
...
+        with:
+          release: ${{ matrix.release }}
+          copyback: false
+          usesh: true
+          prepare: |
+            pkg install -y \
```
arrowd, line 1044: You might want to switch to the `latest` package set before installing deps like it is done in https://github.com/PackageKit/PackageKit/blob/c7508778a1ead262c425a305bd02a92ee2ab2cc8/.cirrus.yml#L7-L8

.github/workflows/test.yml
mitchellh: We hardcode to hashes throughout so that the upgrades of every individual commit are controlled by us, so we aren't subject to the various GHA attacks that have happened multiple times before.

src/benchmark/Benchmark.zig
```diff
@@ -131,6 +131,13 @@ pub const VTable = struct {
 };
 
 test Benchmark {
+    // This test fails on FreeBSD so skip:
+    //
+    // /home/runner/work/ghostty/ghostty/src/benchmark/Benchmark.zig:165:5: 0x3cd2de1 in decltest.Benchmark (ghostty-test)
+    //     try testing.expect(result.duration > 0);
```
mitchellh, line 137: This isn't unique to FreeBSD, this happens because this test is timing dependent, which is naughty. We should fix that (not in this PR).

mitchellh (approved):

mitchellh:
Thanks for taking this on. This looks good to me.

.github/workflows/test.yml
charlesrocket: @arrowd not all projects run immutable tags, and with dependabot there is not much of a difference

PR #8472, opened 2025-08-31
feat: added faint-opacity option

pamburus:
This pull request adds the `--faint-opacity` option, as discussed in #7637.

The default value of the option is also changed from `0.68` to `0.5` for greater consistency with other popular terminal emulators.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3040,7 +3040,7 @@ else
 /// or greater than 1 will be clamped to the nearest valid value.
 ///
 /// Available since Ghostty 1.2.0.
-@"faint-opacity": f64 = 0.6,
+@"faint-opacity": f64 = 0.5,
 
 /// This will be used to set the `TERM` environment variable.
 /// HACK: We set this with an `xterm` prefix because vim uses that to enable key
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4024,6 +4024,8 @@ pub fn finalize(self: *Config) !void {
     if (self.@"auto-update-channel" == null) {
         self.@"auto-update-channel" = build_config.release_channel;
     }
+
+    self.@"faint-opacity" = std.math.clamp(self.@"faint-opacity", 0.0, 1.0);
 }
 
 /// Callback for src/cli/args.zig to allow us to handle special cases
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3035,6 +3035,13 @@ else
 /// Available since Ghostty 1.2.0.
 @"bold-color": ?BoldColor = null,
 
+/// The opacity level (opposite of transparency) of the faint text. A value of
+/// 1 is fully opaque and a value of 0 is fully transparent. A value less than 0
+/// or greater than 1 will be clamped to the nearest valid value.
+///
+/// Available since Ghostty 1.2.0.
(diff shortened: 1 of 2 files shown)


00-kat (approved):

00-kat:
xref #7637.

src/renderer/generic.zig
```diff
@@ -584,6 +585,7 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     .background = config.background.toTerminalRGB(),
                     .foreground = config.foreground.toTerminalRGB(),
                     .bold_color = config.@"bold-color",
+                    .faint_opacity = @intFromFloat(@max(0.0, @min(1.0, (config.@"faint-opacity"))) * 255 + 0.5),
```
pluiedev:
```suggestion
                    .faint_opacity = @ceil(std.math.clamp(config.@"faint-opacity", 0.0, 1.0) * 255),
```

pluiedev (changes requested):

src/renderer/generic.zig
pamburus: Changed. Though `@ceil` returns `f64`, so `@intFromFloat` is still needed.

pluiedev (approved):

src/config/Config.zig
```diff
@@ -3035,6 +3035,13 @@ else
 /// Available since Ghostty 1.2.0.
 @"bold-color": ?BoldColor = null,
 
+/// The opacity level (opposite of transparency) of the faint text. A value of
+/// 1 is fully opaque and a value of 0 is fully transparent. A value less than 0
+/// or greater than 1 will be clamped to the nearest valid value.
+///
+/// Available since Ghostty 1.2.0.
+@"faint-opacity": f64 = 0.6,
```
mitchellh: This doesn't match the existing default, which is closer to 0.68. Should we split the middle and do something like 0.65 or 0.7? Or did you purposely choose 0.6?

src/renderer/generic.zig
```diff
@@ -584,6 +585,7 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     .background = config.background.toTerminalRGB(),
                     .foreground = config.foreground.toTerminalRGB(),
                     .bold_color = config.@"bold-color",
+                    .faint_opacity = @intFromFloat(@ceil(std.math.clamp(config.@"faint-opacity", 0.0, 1.0) * 255)),
```
mitchellh: Instead of clamping here, we should clamp in `finalize` in `Config` and assume correct.

mitchellh (changes requested):

src/config/Config.zig
pamburus: No, I just rounded it roughly to the lowest multiple of 0.1.

If you would prefer the default not to change, I can use the closest value to the existing one.

However, based on my observations, the default value in many other popular terminal emulators is 0.5.
To maintain consistency with other terminal emulators, I would prefer to set this value as the default.
If you are okay with changing the default but are unsure about the exact value, I can research the defaults of other terminal emulators more thoroughly.

src/renderer/generic.zig
pamburus: Changed.

mitchellh (approved):

PR #8479, opened 2025-09-01
i18n: update ja_JP translations

lonsagisawa:
Update Japanese translation for Ghostty 1.2. Part of #8344

### DIFF
--- a/po/ja_JP.UTF-8.po
+++ b/po/ja_JP.UTF-8.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2025-07-22 17:18+0000\n"
-"PO-Revision-Date: 2025-03-21 00:08+0900\n"
+"PO-Revision-Date: 2025-09-01 14:43+0900\n"
 "Last-Translator: Lon Sagisawa <lon@sagisawa.me>\n"
 "Language-Team: Japanese\n"
 "Language: ja\n"
@@ -88,7 +88,7 @@ msgstr "右に分割"
 
 #: src/apprt/gtk/ui/1.5/command-palette.blp:16
 msgid "Execute a command…"
-msgstr ""
+msgstr "コマンドを実行…"
 
 #: src/apprt/gtk/ui/1.0/menu-surface-context_menu.blp:6
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:6
@@ -161,7 +161,7 @@ msgstr "設定ファイルを開く"
 
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:85
 msgid "Command Palette"
-msgstr ""
+msgstr "コマンドパレット"
 
 #: src/apprt/gtk/ui/1.0/menu-window-titlebar_menu.blp:90
 msgid "Terminal Inspector"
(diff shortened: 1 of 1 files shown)


kawarimidoll:
Nice work 👍 
Good Japanese translation!

00-kat:
@kawarimidoll: do you mean that you have reviewed the translations and have not spotted any errors? I'm a little unsure as to whether you have reviewed them as you added a comment rather than a review, so I wanted to ask to be certain.

po/ja_JP.UTF-8.po
```diff
@@ -209,12 +209,12 @@ msgstr "許可"
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-read.blp:81
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-write.blp:77
 msgid "Remember choice for this split"
-msgstr ""
+msgstr "この分割ウィンドウに対して設定を記憶する"
```
tnagatomi, line 212:
```suggestion
msgstr "この分割に対して設定を記憶する"
```

I think "split" in this message refers to splitting a tab rather than a window, so "分割ウィンドウ" (split window) seems somewhat inaccurate.

I've suggested a more literal translation.

po/ja_JP.UTF-8.po
```diff
@@ -209,12 +209,12 @@ msgstr "許可"
...
 
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-read.blp:82
 #: src/apprt/gtk/ui/1.5/ccw-osc-52-write.blp:78
 msgid "Reload configuration to show this prompt again"
-msgstr ""
+msgstr "設定を再読み込みしてプロンプトを再び表示"
```
tnagatomi:
```suggestion
msgstr "このプロンプトを再び表示するには設定を再読み込みしてください"
```

I think this message is an instruction to the user rather than a description of what it does.

tnagatomi (commented):
Looking good! I've made some suggestions.

po/ja_JP.UTF-8.po
kawarimidoll: Then "Close Split?" on L257-258 should be "分割を閉じる" for standardization?

tnagatomi: Ah yeah, and along with that, we'd also need to update the "分割ウィンドウ内のすべてのプロセスが終了します。" message if we adopt my suggestion:
https://github.com/ghostty-org/ghostty/blob/ea878f9b2fe96f73a33fa9e1ae42c569c6df4a79/po/ja_JP.UTF-8.po#L272-L274

Those are the only places where the phrase "分割ウィンドウ" appears.

kawarimidoll:
@00-kat 
> @kawarimidoll: do you mean that you have reviewed the translations and have not spotted any errors? I'm a little unsure as to whether you have reviewed them as you added a comment rather than a review, so I wanted to ask to be certain.

This pr looked good to me at that time, but  @tnagatomi 's suggestion makes sense.
I will leave re-review work to him.

po/ja_JP.UTF-8.po
lonsagisawa: Makes sense, thanks!

po/ja_JP.UTF-8.po
lonsagisawa: I have same consistency concern as @kawarimidoll but I have originally translated "split" meaning of split area to "分割ウィンドウ" so it may be my bad at first :p

tnagatomi: So you'd like to leave "分割ウィンドウ" to keep consistency for now? I'm fine with that 👍

lonsagisawa: I'll keep this translation for consistency and clarity.

tnagatomi (approved):

00-kat (approved):

kawarimidoll (approved):

00-kat:
@tnagatomi and @kawarimidoll: would you be interested in joining the locale team for Japanese to help out with future translations? There's currently only one member which is a bit problematic, so it would be great if you were able to do so! (No harsh feelings if you don't want to, of course.)

tnagatomi:
@00-kat Sure, I'd be glad to help!

kawarimidoll:
@00-kat Yes, I’d be happy to help with translation!

PR #8484, opened 2025-09-02
gtk-ng/wayland: allow more quick terminal configs

pluiedev:
We should at least be on par with what Kitty's panels could do in terms of customization

### DIFF
--- a/src/apprt/gtk-ng/winproto/wayland.zig
+++ b/src/apprt/gtk-ng/winproto/wayland.zig
@@ -114,11 +114,13 @@ pub const App = struct {
             return false;
         }
 
-        if (self.context.xdg_wm_dialog_present and layer_shell.getLibraryVersion().order(.{
-            .major = 1,
-            .minor = 0,
-            .patch = 4,
-        }) == .lt) {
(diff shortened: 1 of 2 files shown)


jcollie (approved):

mitchellh (changes requested):
macOS has a similar config and people have asked us to expose it but the naming, set of layers, and semantics are COMPLETELY different. This feels like it should be a platform-prefixed config because I don’t think there’s anyway to make this reliably cross platform, eg gtk- or Linux or even Wayland.

pluiedev:
`gtk-quick-terminal-layer` etc. then?

mitchellh:
I think that works, since Wayland can run beyond Linux. I'd be fine with introducing our first `wayland-`-prefixed config here too since all this terminology and behavior is totally unique to that and would likely change if we ever supported X11 (impossible?) or some future technology (hope not). :P

pluiedev:
I think Kitty repurposes the namespace as the X11 class too...? Not sure. I'm open to making them `gtk-wl-` prefixed or something like that though

mitchellh (approved):

PR #8488, opened 2025-09-02
fix(renderer): kitty images should all be processed

qwerasd205:
When processing kitty images in a loop in a few places we were returning under certain conditions where we should instead have just continued the loop. This caused serious problems for kitty images, especially for apps that used multiple images on screen at once.

... I have no clue how I originally wrote this code and didn't see such a trivial mistake, I think I was sleep deprived or something.

Should fix #8471


mitchellh (approved):

mitchellh:
> ... I have no clue how I originally wrote this code and didn't see such a trivial mistake, I think I was sleep deprived or something.

Easy for stuff like this to fall into the cracks with large diffs. It was a big body of work. All good. 😄

PR #8491, opened 2025-09-02
fix(font): detect and reject improper advance for icwidth

qwerasd205:
Fixes #8481

Explained in code comments, basically the NF patcher can produce fonts that have CJK characters with 1-cell advances, which screws up fallback font scaling; fixed by not counting the ic width metric if the width of the glyph is greater than the advance width.

> [!NOTE]
> As follow-on work to this it may be worth setting limits for scaling, so you can't have one font scaled like twice as large as the primary font, since that's almost always going to indicate something is very wrong.

### DIFF
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -806,14 +806,31 @@ pub const Face = struct {
         const ic_width: ?f64 = ic_width: {
             const glyph = self.glyphIndex('水') orelse break :ic_width null;
 
-            var advances: [1]macos.graphics.Size = undefined;
-            _ = ct_font.getAdvancesForGlyphs(
+            const advance = ct_font.getAdvancesForGlyphs(
                 .horizontal,
                 &.{@intCast(glyph)},
-                &advances,
+                null,
             );
 
-            break :ic_width advances[0].width;
+            const bounds = ct_font.getBoundingRectsForGlyphs(
+                .horizontal,
+                &.{@intCast(glyph)},
+                null,
+            );
+
+            // If the advance of the glyph is less than the width of the actual
+            // glyph then we just treat it as invalid since it's probably wrong
+            // and using it for size normalization will instead make the font
+            // way too big.
+            //
+            // This can sometimes happen if there's a CJK font that has been
+            // patched with the nerd fonts patcher and it butchers the advance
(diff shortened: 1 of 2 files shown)


src/font/face/coretext.zig
```diff
@@ -806,14 +806,31 @@ pub const Face = struct {
...
+            // If the advance of the glyph is less than the width of the actual
+            // glyph then we just treat it as invalid since it's probably wrong
+            // and using it for size normalization will instead make the font
+            // way too big.
+            //
+            // This can sometimes happen if there's a CJK font that has been
+            // patched with the nerd fonts patcher and it butchers the advance
+            // values so the advance ends up half the width of the actual glyph.
+            if (bounds.size.width > advance) {
+                break :ic_width null;
```
mitchellh, line 840: I suggest adding a `log.warn` or `log.info` right here so that if this is happening maybe someone can see it in logs. Also since it's rare.

src/font/face/freetype.zig
```diff
@@ -1007,7 +1007,21 @@ pub const Face = struct {
...
+            // If the advance of the glyph is less than the width of the actual
+            // glyph then we just treat it as invalid since it's probably wrong
+            // and using it for size normalization will instead make the font
+            // way too big.
+            //
+            // This can sometimes happen if there's a CJK font that has been
+            // patched with the nerd fonts patcher and it butchers the advance
+            // values so the advance ends up half the width of the actual glyph.
+            if (ft_glyph.*.metrics.width > ft_glyph.*.advance.x) {
+                break :ic_width null;
```
mitchellh, line 1031: Same here.

mitchellh (changes requested):
Thanks for looking into this. I requested one thing if it makes sense.

src/font/face/coretext.zig
qwerasd205: Yeah that makes sense, will do.

qwerasd205: Alright, warning added, looks like this:
```
warning(font_face): (getMetrics) Width of glyph '水' for font "D2CodingLigature Nerd Font Mono" is greater than its advance (21.449999999999996 > 13), discarding ic_width metric.
```

mitchellh (approved):

PR #8492, opened 2025-09-02
macOS: firstRect should return full rect width/height

mitchellh:
Fixes #2473

This commit changes `ghostty_surface_ime_point` to return a full rect with the width/height calculated for the preedit.

The `firstRect` function, which calls `ghostty_surface_ime_point` was previously setting the width/height to zero. macOS didn't like this. We then changed it to just hardcode it to width/height of one cell. This worked but made it so the IME cursor didn't follow the preedit.

The result is shown in the video below. Notice the dictation icon follows the text properly:

https://github.com/user-attachments/assets/81be8c63-9f0a-49b7-ac30-2db930beb238


mitchellh:
hm, I realized from textedit that the icon should be UNDER the text, so im looking into that now.

PR #8500, opened 2025-09-03
apprt/gtk-ng: must quit scenarios should quit immediately

### LINKED ISSUE #8495: gtk-ng regression: closing window can cause an infinite loop creating new OpenGL contexts
Reproduction:

1. Open `btop`
2. Close window with "X"
3. Click "Confirm" close
4. Watch logs/memory spin out of control

mitchellh:
Fixes #8495

We were incorrectly calling graceful quit under must quit scenarios. This would do things like confirm quit by inspecting for running processes. However, must quit scenarios (namely when all windows are destroyed) should quit immediately without any checks because the dispose process takes more event loop ticks to fully finish.

### DIFF
--- a/src/apprt/gtk-ng/class/application.zig
+++ b/src/apprt/gtk-ng/class/application.zig
@@ -476,7 +476,14 @@ pub const Application = extern struct {
                 break :q false;
             };
 
-            if (must_quit) self.quit();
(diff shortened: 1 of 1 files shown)


jcollie (approved):
LGTM

PR #8510, opened 2025-09-04
Constrain dingbats + add some more symbol-like blocks to isSymbol

qwerasd205:
This is a very minimal change to fix a regression from when the constraints were reworked (moved off the GPU and on to the CPU), where we weren't constraining dingbats anymore. This fixes that and also adds several other Unicode blocks to also treat as symbols.

In the future (post-1.2) I'll make a more comprehensive set of custom constraints that differ depending on a hand-rolled list of character attributes.

|Before|After|
|-|-|
|image|image|

You'll notice that with spaces in between (so each glyph can be up to 2 cells wide), there's very little change with a font with wider cells, there would be more of a change for people with narrower fonts. The real star of the show is when they're all shoved together and need to be constrained to 1 cell; it's utter chaos without constraints, but with constraints it's nice and orderly (some glyphs get too small to read well, but there's nothing we can really do about that, users should really have a font like Iosevka Fixed installed on their system if they want a good source for terminal-compatible symbol glyphs)

### DIFF
--- a/src/renderer/cell.zig
+++ b/src/renderer/cell.zig
@@ -237,13 +237,27 @@ pub fn isCovering(cp: u21) bool {
 
 /// Returns true of the codepoint is a "symbol-like" character, which
 /// for now we define as anything in a private use area and anything
-/// in the "dingbats" unicode block.
+/// in several unicode blocks:
+/// - Dingbats
+/// - Emoticons
+/// - Miscellaneous Symbols
+/// - Enclosed Alphanumerics
+/// - Enclosed Alphanumeric Supplement
+/// - Miscellaneous Symbols and Pictographs
+/// - Transport and Map Symbols
 ///
 /// In the future it may be prudent to expand this to encompass more
 /// symbol-like characters, and/or exclude some PUA sections.
 pub fn isSymbol(cp: u21) bool {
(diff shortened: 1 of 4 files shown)


src/renderer/cell.zig
```diff
@@ -237,13 +237,27 @@ pub fn isCovering(cp: u21) bool {
...
+/// - Miscellaneous Symbols
+/// - Enclosed Alphanumerics
+/// - Enclosed Alphanumeric Supplement
+/// - Miscellaneous Symbols and Pictographs
+/// - Transport and Map Symbols
 ///
 /// In the future it may be prudent to expand this to encompass more
 /// symbol-like characters, and/or exclude some PUA sections.
 pub fn isSymbol(cp: u21) bool {
+    // TODO: This should probably become a codegen'd LUT
```
mitchellh, line 252: Oh yeah. This is super nasty. I think we should make a "task" type issue to track this right now. It might be worth making a LUT based on our current unicode stuff before we even get the ziglyph replacement figured out.

mitchellh (approved):
Approve but note my comment, I think we should make the LUT task asap.

src/renderer/cell.zig
qwerasd205: #8523

PR #8511, opened 2025-09-04
gtk-ng: deprecate detection of launch source

jcollie:
Detecting the launch source frequently failed because various launchers fail to sanitize the environment variables that Ghostty used to detect the launch source. For example, if your desktop environment was launched by `systemd`, but your desktop environment did not sanitize the `INVOCATION_ID` or the `JOURNAL_STREAM` environment variables, Ghostty would assume that it had been launched by `systemd` and behave as such.

This led to complaints about Ghostty not creating new windows when users expected that it would.

To remedy this, Ghostty no longer does any detection of the launch source. If your launch source is something other than the CLI, it must be explicitly speciflied on the CLI. All of Ghostty's default desktop and service files do this. Users or packagers that create custom desktop or service files will need to take this into account.

On GTK, the `desktop` setting for `gtk-single-instance` is replaced with `detect`. `detect` behaves as `gtk-single-instance=true` if one of the following conditions is true:

1. If no CLI arguments have been set.
2. If `--launched-from` has been set to `desktop`, `dbus`, or `systemd`.

Otherwise `detect` behaves as `gtk-single-instance=false`.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -73,6 +73,10 @@ pub const compatibility = std.StaticStringMap(
     // Ghostty 1.2 merged `bold-is-bright` into the new `bold-color`
     // by setting the value to "bright".
     .{ "bold-is-bright", compatBoldIsBright },
+
+    // Ghostty 1.2 removed the "desktop" option and renamed it to "detect".
+    // The semantics also changed slightly but this is the correct mapping.
+    .{ "gtk-single-instance", compatGtkSingleInstance },
 });
 
 /// The font families to use.
@@ -2975,16 +2979,23 @@ else
 ///
 /// If `false`, each new ghostty process will launch a separate application.
 ///
-/// If `detect`, Ghostty will act as if it was `true` if one of the following
-/// conditions is true:
+/// If `detect`, Ghostty will assume true (single instance) unless one of
+/// the following scenarios is found:
 ///
-/// 1. If no CLI arguments have been set.
-/// 2. If `--launched-from` has been set to `desktop`, `dbus`, or `systemd`.
+/// 1. TERM_PROGRAM environment variable is a non-empty value. In this
+/// case, we assume Ghostty is being launched from a graphical terminal
+/// session and you want a dedicated instance.
 ///
-/// Otherwise, Ghostty will act as if it was `false`.
+/// 2. Any CLI arguments exist. In this case, we assume you are passing
+/// custom Ghostty configuration. Single instance mode inherits the
+/// configuration from when it was launched, so we must disable single
+/// instance to load the new configuration.
 ///
-/// The pre-1.2 option `desktop` has been deprecated. If encountered it will be
-/// treated as `detect`.
+/// If either of these scenarios is producing a false positive, you can
+/// set this configuration explicitly to the behavior you want.
+///
+/// The pre-1.2 option `desktop` has been deprecated. Please replace
+/// this with `detect`.
 ///
 /// The default value is `detect`.
 ///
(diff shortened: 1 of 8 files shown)


pluiedev (approved):

mitchellh (changes requested):
Just putting red so I can look myself. Thanks for looking into this

src/config/Config.zig
```diff
@@ -3488,7 +3494,23 @@ pub fn loadCliArgs(self: *Config, alloc_gpa: Allocator) !void {
         .windows => {},
 
         // Fast-path if we are Linux and have no args.
-        .linux, .freebsd => if (std.os.argv.len <= 1) return,
+        .linux, .freebsd => {
+            if (std.os.argv.len <= 1) {
```
mitchellh: This probably isn't the right place for this, and we should put detection in `finalize`. We can technically call any number of `load` calls after this which may set `gtk-single-instance` back to `detect` (e.g. from a file, but the API doesn't dictate the order of load calls either so future libghostty consumers can do anything).

src/config/Config.zig
```diff
@@ -2975,14 +2975,22 @@ else
...
+///
+/// The pre-1.2 option `desktop` has been deprecated. If encountered it will be
+/// treated as `detect`.
+///
+/// The default value is `detect`.
 ///
 /// Note that debug builds of Ghostty have a separate single-instance ID
 /// so you can test single instance without conflicting with release builds.
-@"gtk-single-instance": GtkSingleInstance = .desktop,
+@"gtk-single-instance": GtkSingleInstance = .default,
```
mitchellh, line 3004: I'm not sure how I feel about using default decls everywhere. It's making it harder to read the default configs because it now requires a jump to definition on every single field versus having `Config` be the source of truth before. 

No need to address this in PR, just noticing it more recently as I've fixed other bugs.

src/config/Config.zig
```diff
@@ -3488,7 +3494,23 @@ pub fn loadCliArgs(self: *Config, alloc_gpa: Allocator) !void {
...
 
         // Fast-path if we are Linux and have no args.
-        .linux, .freebsd => if (std.os.argv.len <= 1) return,
+        .linux, .freebsd => {
+            if (std.os.argv.len <= 1) {
+                if (self.@"gtk-single-instance" == .detect) {
+                    const arena_alloc = self._arena.?.allocator();
+                    // Add an artificial replay step so that replaying the
+                    // inputs doesn't undo this change.
+                    try self._replay_steps.append(
```
mitchellh: Doing it in finalize would avoid this nastiness too, since replayed configs are automatically finalized again. Modifying replay steps in general is very spooky.

mitchellh (changes requested):

mitchellh:
Haven't run this yet to verify the behavior, just reviewed the code at a high level.

src/config/Config.zig
```diff
@@ -4493,6 +4525,32 @@ fn equalField(comptime T: type, old: T, new: T) bool {
     }
 }
 
+/// This runs a heuristic to determine if we are likely running
+/// Ghostty in a CLI environment. We need this to change some behaviors.
+/// We should keep the set of behaviors that depend on this as small
+/// as possible because magic sucks, but each place is well documented.
+fn probableCliEnvironment() bool {
```
mitchellh, line 4532: I put this here instead of `src/os` because its _only_ used by config defaults and I don't want this bleeding into other places.

mitchellh:
I think I figured something out, details in the commit:

```
apprt/gtk-ng: clean up our single instance, new window interactions

This removes `launched-from` entirely and moves our `gtk-single-instance`
detection logic to assume true unless we detect CLI instead of assume
false unless we detect desktop/dbus/systemd.

The "assume true" scenario for single instance is desirable because
detecting a CLI instance is much more reliable.

Removing `launched-from` fixes an issue where we had a
difficult-to-understand relationship between `launched-from`,
`gtk-single-instance`, and `initial-window`. Now, only
`gtk-single-instance` has some hueristic logic. And `initial-window`
ALWAYS sends a GTK activation signal regardless of single instance or
not.

As a result, we need to be explicit in our systemd, dbus, desktop files
about what we want Ghostty to do, but everything works as you'd mostly
expect.

Now, if you put plain old `ghostty` in your terminal, you get a new
Ghostty instance. If you put it anywhere else, you get a GTK single
instance activation call (either creates a first instance or opens a new
window in the existing instance). Works for launchers and so on.
```

mitchellh:
I need to do more testing tomorrow, but reviews welcome. Initial testing on my Arch+Hyprland system is that everything works as expected. But more in depth testing required.

jcollie:
Changes LGTM. Going to stick this in my daily driver build and see how it works.

mitchellh (approved):
I think this resolves everything... its working on my system and @jcollie for the past day.

danielwe:
Since this PR, specifically the last commit, the behavior with `open` on macOS changed as follows:

* `open -a Ghostty.app` launches in the home directory as usual
* `open -a Ghostty.app --args --keybind=global:ctrl+cmd+t=new_window` (or any other cmdline arg) launches in the root directory

Previously, both versions would launch in the home directory.

Is this the expected behavior on macOS going forward?

PR #8512, opened 2025-09-04
Fix font handling for bitmap and non-sfnt fonts

### LINKED ISSUE #2991: Graceful handling of font files that fail to load (invalid font tables, bad format, etc.)
We need to handle bad fonts (corrupted, or unsupported due to missing tables) better. Currently they cause rendering issues when they appear as fallbacks and they cause straight up crashes when set as the primary font. We need to handle this gracefully, and the font discovery mechanism(s) need to be updated to exclude bitmap fonts.

### LINKED ISSUE #8483: FreeType: Don't assume all fonts have SFNT tables
Escalating #8073. Please no rotten tomatoes ❤️ I've been asked to convert this to a proper issue  

FreeType supports old-school BDF and PCF bitmap fonts and will include such fonts in font discovery, so any time such a font is installed on the system, Ghostty may load it (whether configured or as a fallback). However, since these are not SFNT file formats, Ghostty will throw a `CopyTableError` when trying to get the font's metrics, and possibly crash or corrupt the terminal state when printing codepoints that map to this font.

The [`getMetrics`](https://github.com/ghostty-org/ghostty/blob/5ef6412823f789d530dcba54705a533b2e6ce0a7/src/font/face/freetype.zig#L791-L1032) function already contains the necessary fallback calculations to get all the required metrics from a font without using tables, so what's needed is just a refactor such that these fallbacks are used when SFNT tables don't exist (currently, the fallbacks are only used for missing quantities; the existence of the tables themselves is a hardcoded assumption).

To reproduce:

* Use Linux or another platform where Ghostty uses FreeType (i.e., not macOS).
* Install the `efont-unicode` collection
  1. Don't use a package manager, as some of them will convert the fonts to the OTB format, masking the issue
  2. Download http://openlab.jp/efont/dist/unicode-bdf/efont-unicode-bdf-0.4.2.tar.bz2
  3. Extract and copy the resulting folder into `~/.local/share/fonts/`
  4. Run `fc-cache -fv`
* Install a modern CJK-supporting font such as Noto Sans Mono CJK (feel free to use a package manager for this)
* Launch Ghostty with the following config:
  ```ini
  font-family = "JetBrains Mono"
  font-family = "Noto Sans Mono CJK"
  ```
* Run `ghostty +show-face --cp=0x6f22`
  - This should print `U+6F22 « 漢 » found in face “Noto Sans CJK JP”.` (This is just a sanity check.)
* Now, launch Ghostty with the following config:
  ```ini
  font-family = "JetBrains Mono"
  font-family = "Biwidth"
  ```
* Again, run `ghostty +show-face --cp=0x6f22`
  - On tip as of this writing, this will print `Unable to get face: error.CopyTableError`; also, `cat cjk_file.txt` will print blank spaces, and `nvim cjk_file.txt` may corrupt the terminal state
  - On some older versions, Ghostty may crash

---

I started fiddling with the refactor here: https://github.com/ghostty-org/ghostty/compare/main...danielwe:ghostty:bitmap_metrics. Anyone feel free to use it as a starting point. Just beware that it's far from ready. There's a bunch of needlessly duplicated code.

More importantly, one thing I learned is that there seems to be a bug in FreeType where it segfaults during resource cleanup if a BDF font is loaded. This could also be a bug in Ghostty's use of HarfBuzz's refcounting (the `self.face.ref()` stuff), but I wasn't able to fix the problem by messing around with that. Anyway, it's not a huge deal, as the segfault only happens as Ghostty is quitting, and BDF fonts are rare these days. But no segfault would obviously be nicer.

qwerasd205:
Fixes #8483, fixes #2991

With this change, `font.face.getMetrics` is now infallible, and real bitmap fonts are properly handled and can be configured as the primary font (or used as fallbacks), as long as the backend (FreeType or CoreText) knows how to interpret them, since we now fall back on the backend for any metrics we can't extract from sfnt tables (which means we don't need any to be present in the first place).

Also, doing this uncovered a double-free issue in our FreeType `renderGlyph` code, which thankfully wasn't too hard to track down and fix.

> [!NOTE]
> We should vendor a true bitmap font in each of the native formats supported by each backend and add tests for the metrics being computed right and the glyphs being rendered correctly. Idk if we wanna block this PR on that or not.

### DIFF
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -648,68 +648,64 @@ pub const Face = struct {
         const px_per_unit: f64 = px_per_em / units_per_em;
 
         const ascent: f64, const descent: f64, const line_gap: f64 = vertical_metrics: {
-            if (hhea_) |hhea| {
-                const hhea_ascent: f64 = @floatFromInt(hhea.ascender);
-                const hhea_descent: f64 = @floatFromInt(hhea.descender);
-                const hhea_line_gap: f64 = @floatFromInt(hhea.lineGap);
-
-                if (os2_) |os2| {
-                    const os2_ascent: f64 = @floatFromInt(os2.sTypoAscender);
-                    const os2_descent: f64 = @floatFromInt(os2.sTypoDescender);
-                    const os2_line_gap: f64 = @floatFromInt(os2.sTypoLineGap);
-
-                    // If the font says to use typo metrics, trust it.
-                    if (os2.fsSelection.use_typo_metrics) break :vertical_metrics .{
-                        os2_ascent * px_per_unit,
-                        os2_descent * px_per_unit,
-                        os2_line_gap * px_per_unit,
-                    };
-
-                    // Otherwise we prefer the height metrics from 'hhea' if they
-                    // are available, or else OS/2 sTypo* metrics, and if all else
-                    // fails then we use OS/2 usWin* metrics.
-                    //
-                    // This is not "standard" behavior, but it's our best bet to
-                    // account for fonts being... just weird. It's pretty much what
-                    // FreeType does to get its generic ascent and descent metrics.
-
-                    if (hhea.ascender != 0 or hhea.descender != 0) break :vertical_metrics .{
-                        hhea_ascent * px_per_unit,
-                        hhea_descent * px_per_unit,
-                        hhea_line_gap * px_per_unit,
-                    };
-
-                    if (os2_ascent != 0 or os2_descent != 0) break :vertical_metrics .{
-                        os2_ascent * px_per_unit,
-                        os2_descent * px_per_unit,
-                        os2_line_gap * px_per_unit,
-                    };
-
-                    const win_ascent: f64 = @floatFromInt(os2.usWinAscent);
-                    const win_descent: f64 = @floatFromInt(os2.usWinDescent);
-                    break :vertical_metrics .{
-                        win_ascent * px_per_unit,
-                        // usWinDescent is *positive* -> down unlike sTypoDescender
-                        // and hhea.Descender, so we flip its sign to fix this.
-                        -win_descent * px_per_unit,
-                        0.0,
-                    };
-                }
-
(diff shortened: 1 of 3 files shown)


src/font/face/coretext.zig
```diff
@@ -637,96 +627,114 @@ pub const Face = struct {
...
-                };
-
-                // Otherwise we prefer the height metrics from 'hhea' if they
-                // are available, or else OS/2 sTypo* metrics, and if all else
-                // fails then we use OS/2 usWin* metrics.
-                //
-                // This is not "standard" behavior, but it's our best bet to
-                // account for fonts being... just weird. It's pretty much what
-                // FreeType does to get its generic ascent and descent metrics.
+            if (hhea_) |hhea| {
```
mitchellh: I'd flip this into a `const hhea = hhea_ orelse` block. I think you'd lower a lot of nesting that way since it seems the else case is a lot shorter than the if case.

src/font/face/freetype.zig
```diff
@@ -812,92 +814,130 @@ pub const Face = struct {
...
-                // If the font says to use typo metrics, trust it.
-                // (The USE_TYPO_METRICS bit is bit 7)
-                if (os2.fsSelection & (1 << 7) != 0) {
-                    break :vertical_metrics .{
-                        os2_ascent * px_per_unit,
-                        os2_descent * px_per_unit,
-                        os2_line_gap * px_per_unit,
-                    };
-                }
+            if (hhea_) |hhea| {
```
mitchellh: Same feedback here about the orelse

mitchellh (changes requested):

mitchellh:
> We should vendor a true bitmap font in each of the native formats supported by each backend and add tests for the metrics being computed right and the glyphs being rendered correctly. Idk if we wanna block this PR on that or not.

I'd be satisfied with making a task issue right now to track this and not blocking this PR.

The rest looks great, some small stylistic comments. But appreciate this.

mitchellh (approved):

PR #8504, opened 2025-09-03
config: make default copy_to_clipboard binds performable

rockorager:
Make the default keybind for copy_to_clipboard performable. This allows
TUIs to receive events when keybinds aren't used, for example cmd+c on
macos for copy, or ctrl+shift+c elsewhere.

Disclosure: This commit was made with the assistance of AI (ampcode.com)

PR #8516, opened 2025-09-04
feat(search): Implement terminal search functionality

cdelgehier:
- Add SearchManager with threading support to Surface.zig
- Create GTK SearchBar widget with navigation controls and result display
- Implement macOS SearchController and SearchView with SwiftUI/Combine
- Add search actions (toggle_search, search_next, search_previous) to input bindings
- Integrate apprt action system for UI communication between core and frontends

Issue #189 for search scrollbar functionality

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -141,6 +141,12 @@ focused: bool = true,
 /// Used to determine whether to continuously scroll.
 selection_scroll_active: bool = false,
 
+/// Manages text search functionality for this terminal surface.
+/// Each terminal window/tab has its own search state so users can
+/// search for different things in different terminals at the same time.
+/// This handles finding text in the terminal history and highlighting results.
+search_manager: SearchManager,
+
 /// The effect of an input event. This can be used by callers to take
 /// the appropriate action after an input event. For example, key
 /// input can be forwarded to the OS for further processing if it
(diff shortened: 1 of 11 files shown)


rhodes-b:
is this supposed to be remotely functional on the gtk-ng side? It seems to be missing quite a bit to work properly.

I didn't get it working since im not really familiar with gtk-ng / how the actions need to be passed around but there was some basic things I fixed (although its not in a working state and I dont think its much worth wasted time since im likely doing it wrong)

https://github.com/ghostty-org/ghostty/commit/a40664f27da013e056c5e9eb373f8b570b68d41c

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+    // a complete search interface that users can interact with.
+    fn init(self: *Self) callconv(.C) void {
+        // Get our private data storage
+        const priv = self.private();
+        
+        // Set up the main container as a horizontal box.
+        // This arranges all search controls in a row from left to right.
+        self.parent_instance.setOrientation(gtk.Orientation.horizontal);
+        self.parent_instance.setSpacing(6); // Add small gaps between widgets
+        self.parent_instance.addCssClass("search-bar"); // For styling with CSS
```
pluiedev, line 123: We usually use blueprints to declaratively construct UIs instead of using code like this. See `src/apprt/gtk-ng/ui/` for some examples.

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+
+        // Create the search text input field.
+        // This is where users type what they want to find in terminal history.
+        priv.search_entry = gtk.Entry.new();
+        priv.search_entry.setPlaceholderText("Search in terminal...");
```
pluiedev, line 128: Needs to be localized (use `_("...")` in blueprints or `i18n._` in Zig code)

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+        gobject.ext.impl_helpers.emitSignal(
+            self,
+            signals.search_text_changed.name,
+            .{search_text},
+        );
+    }
+
+    // Called when user presses keys while focused on the search input.
+    // This handles special keys like Enter (next result) and Escape (close search).
+    fn onKeyPressed(
```
pluiedev, line 251: You can use a `gtk.SearchEntry` instead which has special signals specifically for this. (See command palette)

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+                gobject.ext.impl_helpers.emitSignal(
+                    self,
+                    signals.search_closed.name,
+                    .{},
+                );
+                return true; // We handled this key press
+            },
+            
+            // F3 key moves to next result, Shift+F3 moves to previous
+            gtk.KEY_F3 => {
```
pluiedev, line 285: Is there any precedent that F3 should be used for this?

src/apprt/gtk-ng/Surface.zig
```diff
@@ -101,3 +101,18 @@ pub fn defaultTermioEnv(self: *Self) !std.process.EnvMap {
...
+pub fn actionToggleSearchMode(self: *Self, action: *const apprt.action.ToggleSearchMode) !void {
+    self.surface.toggleSearchMode(action.visible);
+}
+
+/// Handle the update_search_results action from the core terminal.
+/// This updates the search bar display to show current search results.
+/// The search bar will show "X of Y matches" or "No results" based on the data.
+pub fn actionUpdateSearchResults(self: *Self, action: *const apprt.action.UpdateSearchResults) !void {
+    self.surface.updateSearchResults(action.current, action.total);
+}
```
pluiedev, line 118: Don't needlessly broaden the code apprt interface - send apprt actions instead

src/Surface.zig
```diff
@@ -232,6 +238,273 @@ pub const Keyboard = struct {
...
+    search_active: bool = false,
+
+    /// Initialize a new search manager. This sets up all the internal
+    /// structures needed for searching, but doesn't start any searches.
+    /// Users need to call startSearch() when they want to find something.
+    pub fn init(_: Allocator) SearchManager {
+        return SearchManager{
+            // All fields start with their default values defined above
+        };
+    }
```
pluiedev, line 297: If this isn't necessary then just default-init with `.{}`

src/Surface.zig
```diff
@@ -232,6 +238,273 @@ pub const Keyboard = struct {
...
+        self.search_needle = try alloc.dupe(u8, needle);
+
+        // Mark that we're starting a search
+        self.search_in_progress = true;
+
+        // Create or get thread pool for background search work
+        if (self.thread_pool == null) {
+            self.thread_pool = try alloc.create(std.Thread.Pool);
+            self.thread_pool.?.* = std.Thread.Pool{};
+            try self.thread_pool.?.init(.{ .allocator = alloc, .n_jobs = 1 });
```
pluiedev, line 360: Style (avoid `.?` if you can, avoid explicit types)
```suggestion
            const pool = try alloc.create(std.Thread.Pool);
            pool.* = .{};
            try pool.init(.{ .allocator = alloc, .n_jobs = 1 });
            self.thread_pool = pool;
```

src/Surface.zig
```diff
@@ -232,6 +238,273 @@ pub const Keyboard = struct {
...
+    /// of searching through terminal history for matches.
+    fn runSearch(job: *SearchJob) void {
+        // Set up search operation using the existing terminal search code
+        var search = terminal.PageListSearch.init(
+            job.allocator,
+            job.page_list,
+            job.needle,
+        ) catch |err| {
+            // If we can't create the search, mark as complete and return
+            std.log.warn("failed to initialize search: {}", .{err});
```
pluiedev, line 448: Don't use `std.log` - use scoped loggers like
```zig
const log = std.log.scoped(.search);
log.warn("...", .{...});
```

src/Surface.zig
```diff
@@ -232,6 +238,273 @@ pub const Keyboard = struct {
...
+            job.search_manager.search_in_progress = false;
+            job.cleanup();
+            return;
+        };
+        defer search.deinit(job.allocator);
+
+        // Collect all search results
+        var temp_results = std.ArrayListUnmanaged(terminal.Selection){};
```
pluiedev, line 456: Default-init of container types is deprecated
```suggestion
        var temp_results: std.ArrayListUnmanaged(terminal.Selection) = .empty;
```

src/Surface.zig
```diff
@@ -533,6 +806,11 @@ pub fn init(
         // Our conditional state is initialized to the app state. This
         // lets us get the most likely correct color theme and so on.
         .config_conditional_state = app.config_conditional_state,
+
+        // Initialize the search manager for this terminal surface.
+        // This sets up all the structures needed for searching through
+        // terminal history, but doesn't start any searches yet.
+        .search_manager = SearchManager.init(alloc),
```
pluiedev, line 813:
```suggestion
        .search_manager = .init(alloc),
```

test_search_config.conf
```diff
@@ -0,0 +1,21 @@
+# Test configuration for Ghostty with search functionality
```
pluiedev, line 1: I don't think you meant to commit this

src/apprt/gtk-ng/class/surface.zig
```diff
@@ -3107,6 +3114,95 @@ const Clipboard = struct {
...
+/// When disabled (visible=false), hides the search bar and returns focus to terminal.
+/// This is called when users press the search toggle keybinding (like Ctrl+F).
+pub fn toggleSearchMode(self: *Self, visible: bool) void {
+    const priv = self.private();
+    
+    if (visible) {
+        // Show search mode - create search bar if it doesn't exist yet
+        if (priv.search_bar == null) {
+            // Create a new search bar widget for this surface
+            priv.search_bar = SearchBar.new();
```
pluiedev, line 3128: You shouldn't really need to do this — just always have the search bar ready in the widget tree and hidden by default

src/input/Binding.zig
```diff
@@ -740,6 +740,27 @@ pub const Action = union(enum) {
...
+    /// When search mode is active, users can type what they want to find.
+    /// Press this action again to close the search bar and return to normal mode.
+    toggle_search,
+
+    /// Go to the next search result found in the terminal history.
+    /// This moves the highlight from the current result to the next one.
+    /// If you are at the last result, it goes back to the first result.
+    /// Does nothing if no search is active or no results were found.
+    /// This is like pressing F3 or Enter in most search interfaces.
+    search_next,
```
pluiedev, line 755: A single `goto_search` action with a `previous` or `next` parameter would likely work better (and would allow us to add `first`/`last`/etc. variants should they become necessary)

src/input/Binding.zig
```diff
@@ -740,6 +740,27 @@ pub const Action = union(enum) {
     ///
     crash: CrashThread,
 
+    /// Start or stop the search mode in the terminal.
+    /// This opens a search bar where users can type text to find in their
+    /// terminal history. It works like pressing Ctrl+F in web browsers.
+    /// When search mode is active, users can type what they want to find.
+    /// Press this action again to close the search bar and return to normal mode.
+    toggle_search,
```
pluiedev, line 748: Maybe change this to be more like `inspector` (`toggle`/`show`/`hide`)

pluiedev (changes requested):
Just a first pass — I'm going to try this for myself. Big if true

Also this will likely not make it in time for 1.2 as we've already entered into a partial feature freeze where no new major UI features would be added. Would love to see this for 1.3 though

pluiedev:
May I ask: is this written with AI assistance? Right now this PR doesn't compile at all on GTK-NG and there's even plenty of missing symbols. If this PR is indeed written with AI assistance, please indicate it as such in the PR description. Same if it hasn't been rigorously tested yet with GTK.

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+    // Get access to the private data for this widget instance.
+    // This is a GTK/GObject pattern for accessing widget-specific data.
+    fn private(self: *Self) *Private {
+        return gobject.ext.impl_helpers.getPrivate(self, Private, Private.offset);
+    }
+
+    // Initialize a new search bar instance.
+    // This sets up all the GTK widgets and connects them together into
+    // a complete search interface that users can interact with.
+    fn init(self: *Self) callconv(.C) void {
```
pluiedev, line 115: This has the wrong signature. It should take both a `*Self` and a `*Class`

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+        results_label: *gtk.Label = undefined,
+
+        // Button to close the search bar.
+        // Shows an X icon and hides the search interface when clicked.
+        close_button: *gtk.Button = undefined,
+    };
+
+    // Get access to the private data for this widget instance.
+    // This is a GTK/GObject pattern for accessing widget-specific data.
+    fn private(self: *Self) *Private {
```
pluiedev, line 108: Not necessary if you use the common helper (grep for `const C = Common(Self`)

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+
+        // Initially disable navigation buttons since there are no results yet.
+        // These will be enabled when a search finds matches.
+        priv.prev_button.setSensitive(false);
+        priv.next_button.setSensitive(false);
+    }
+
+    // Called when user clicks the "previous result" button.
+    // This tells the connected terminal surface to move to the previous search match.
+    fn onPrevClicked(self: *Self, _: *gtk.Button) callconv(.C) void {
```
pluiedev, line 187: Starting from Zig 0.14 it's `callconv(.c)` (lowercase)

src/apprt/gtk-ng/class/search_bar.zig
```diff
@@ -0,0 +1,380 @@
...
+    };
+
+    // C-compatible helper functions for GObject integration
+    const C = struct {
+        // Helper to access private fields from property getters/setters
+        fn privateObjFieldAccessor(comptime field: []const u8) type {
+            return gobject.ext.impl_helpers.PrivateObjFieldAccessor(Self, Private, field);
+        }
+    };
+};
```
pluiedev, line 380: What is this? We never do this anywhere within `gtk-ng`.

pluiedev (changes requested):
After trying this for myself I'm somewhat disappointed in the code quality here. After adding apprt actions you didn't even modify `src/input/command.zig`, `src/apprt/gtk-ng/class/application.zig`, etc. to reflect the new apprt actions. The new methods are put in completely wrong positions, and certain symbols like `SearchBar.new` flat out don't exist.

I've already spent half an hour trying to make this *somehow* compile on GTK-NG but there's just problem after problem after problem. Maybe this works great in macOS but the GTK side is severely undercooked. IMO it's better to leave code out than to present something that doesn't even begin to work. If you can't test the GTK side then don't worry — us GTK maintainers would be able to implement this in another PR.

src/Surface.zig
```diff
@@ -232,6 +238,273 @@ pub const Keyboard = struct {
...
+    /// The terminal page list to search through.
+    /// This contains all the terminal history we need to scan.
+    page_list: *terminal.PageList,
+
+    /// Main function that runs the search in a background thread.
+    /// This is called by the thread pool and does all the actual work
+    /// of searching through terminal history for matches.
+    fn runSearch(job: *SearchJob) void {
+        // Set up search operation using the existing terminal search code
+        var search = terminal.PageListSearch.init(
```
mitchellh, line 442: PageListSearch is absolutely not thread safe. You can't call this from a thread while a PageList is actively in use. This is noted in the docs for it, and is the major TODO:

```
/// Searches for a term in a PageList structure.
///
/// At the time of writing, this does not support searching a pagelist
/// simultaneously as its being used by another thread. This will be resolved
/// in the future.
```

mitchellh (changes requested):
I haven't reviewed the rest yet, my primary question was "oh wow how did they solve the data race issue?" And it appears you haven't. You can't access a `PageList` from another thread while it's in use, and you can't block search on the main thread because it kills IO performance. You opted for path A, but didn't address races.

This is a major architectural issue (the major one if you will blocking all of search). 

Please let me know if I misunderstood that.

mitchellh:
As noted previously, we also [require AI disclosure](https://github.com/ghostty-org/ghostty/blob/c3e7857a2c5031e8274a411c84695351e8e3fffe/CONTRIBUTING.md#ai-assistance-notice) on this project. We welcome AI contributions but need to know if they were used. I'm seeing a lot of quite frankly _very suspicious_ patterns that make no sense in this PR, and I'd like to understand why you felt this was the right approach, as a human.

Just so I'm clear, we very much welcome AI assisted commits. A recent PR I wrote #8508 was AI, I use it daily. And others like #8504 and #8462 were also AI disclosed. All merged. 

But we expect that any AI-assisted work can be properly explained and justified in full by the submitter, or disclosed that they don't understand it so we can better understand our requirements as maintainers.

mitchellh:
Two weeks has passed without a response. There are now conflicts in addition to the major issues with this PR. I'm going to close this.

Issue #8523, opened 2025-09-04
Convert `cell.isSymbol` to use a LUT


qwerasd205:
Follow-on to #8510

This lives in `renderer/cell.zig` when really it should be in `font`, but that's a refactor for another day. Right now we just need to do this TODO since it's in a hot path and is way too much work with all these checks in a row when it could be a LUT.
https://github.com/ghostty-org/ghostty/blob/3ac2da99f4b360f96cae34020fc4bab933b1d260/src/renderer/cell.zig#L251-L261

This should probably use the existing unicode LUT generation stuff we have, though it's for a binary property so maybe worth trimming down. Should validate this change with benchmarks.

mitchellh:
I'm going to throw this into the 1.2 milestone just to _try_ and see if it makes sense.

PR #8526, opened 2025-09-04
ai: add `gh-issue` command to help diagnose GitHub issues

mitchellh:
This enables agents (namely Amp) to use `/gh-issue <number/url>` to  begin diagnosing a GitHub issue, explaining the problem, and suggesting a plan of action. This action explicitly prompts the AI to not write code.

**You can run this manually too,** for testing or curiosity or for pasting into another LLM. Execute it like any other script: `.agents/command/gh-issue <issue>`

I've used this manually for months with good results, so now I'm formalizing it in the repo for other contributors.

Example diagnosing #8523:

https://ampcode.com/threads/T-3e26e8cc-83d1-4e3c-9b5e-02d9111909a7

**I'm going to be highly selective about integrating repository-level commands.** I think guiding AI contributors in the right direction is going to result in less AI slop being sent to us. But I want to only add commands that maintainers use and also can vouch for. The best way to vouch is to share something like an Amp thread link that shows it working on real data. 

Ironically, no AI was used to write this PR. I did consult Claude Chat for help on some Nu syntax, but verified it manually with the Nu manual.

### DIFF
--- a/.agents/commands/gh-issue
+++ b/.agents/commands/gh-issue
@@ -1,0 +1,64 @@
+#!/usr/bin/env nu
+
+# A command to generate an agent prompt to diagnose and formulate 
+# a plan for resolving a GitHub issue.
+#
+# IMPORTANT: This command is prompted to NOT write any code and to ONLY
+# produce a plan. You should still be vigilant when running this but that
+# is the expected behavior.
+# 
+# The `<issue>` parameter can be either an issue number or a full GitHub
+# issue URL.
+def main [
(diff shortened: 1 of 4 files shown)


mitchellh:
Updated the HACKING/CONTRIBUTOR files.

mitchellh:
Some auto-formatter removed the trailing `!` in the contributing guide and I'm way too lazy to figure out what I fucked up in my Vim config for that to happen. lol.

PR #8528, opened 2025-09-05
renderer: add LUT-based implementation of isSymbol

### LINKED ISSUE #8523: Convert `cell.isSymbol` to use a LUT

jcollie:
The LUT-based lookup gives a ~20%-30% speedup over the "naive" isSymbol implementation.

Screenshot From 2025-09-04 22-45-10
Screenshot From 2025-09-04 22-41-54

Fixes #8523


jcollie:
The benchmarks show two "tables" benchmarks. "tables1" is the original LUT implementation used for looking up unicode data. "tables2" is an alternate LUT implementation that I had hoped would do even better but turned out to have no benefits over the existing implementation.

src/unicode/lut.zig
```diff
@@ -142,6 +142,32 @@ pub fn Tables(comptime Elem: type) type {
             return self.stage3[self.stage2[self.stage1[high] + low]];
         }
 
+        pub inline fn getInline(self: *const Self, cp: u21) Elem {
```
mitchellh: future note: There's probably little downside to marking `get` as inline and removing this. We don't need to do this in this PR because I don't want to cause more churn waiting on benchmarks. But this should be as simple as testing terminal throughput (we already have a benchmark).

src/unicode/symbols.zig
```diff
@@ -0,0 +1,93 @@
...
+    // std.log.warn("stage1={} stage2={} stage3={}", .{
+    //     t.stage1.len,
+    //     t.stage2.len,
+    //     t.stage3.len,
+    // });
+}
+
+// This is not very fast in debug modes, so its commented by default.
+// IMPORTANT: UNCOMMENT THIS WHENEVER MAKING CHANGES.
+test "unicode symbols1: tables match ziglyph" {
```
mitchellh: Do we want to comment this out? Usually these slow tests we don't need to run every time. At the very least we should disable this under Valgrind.

src/unicode/lut.zig
```diff
@@ -142,6 +142,32 @@ pub fn Tables(comptime Elem: type) type {
...
+            const high = cp >> 8;
+            const low = cp & 0xFF;
+            return self.stage3[self.stage2[self.stage1[high] + low]];
+        }
+
+        pub fn getBool(self: *const Self, cp: u21) bool {
```
mitchellh: Maybe blind, but actually are any of these used? Should we delete this?

mitchellh (changes requested):
Thanks for doing this work, some questions.

src/unicode/lut.zig
jcollie: I didn't include the results when I tested, but when I initially was doing some testing `inline` seemed to slow things down. I didn't test again after I fixed the problem with the lookups being optimized out.

src/unicode/symbols.zig
jcollie: I left the test enabled because it's not actually that slow, unlike the unicode properties one.

src/unicode/lut.zig
jcollie: Yeah I thought that I had deleted these in the 2nd commit but I guess not.

src/unicode/symbols.zig
mitchellh: We should still disable during Valgrind, because Valgrind is at least 100,000x slower (not an exaggeration) and I don't think there's a practical Valgrind test here. There are other examples of how to do this.

src/unicode/lut.zig
jcollie: Re-tested and `inline` gets a percent or so speedup.

src/unicode/symbols.zig
jcollie: Done.

src/unicode/lut.zig
jcollie: Removed.

mitchellh (approved):

src/unicode/lut.zig
mitchellh: My guess is release mode generally inlines this anyways.

PR #8529, opened 2025-09-05
devshell: add poop

jcollie:


mitchellh:
This might have to be scoped to Linux only. I thought it doesn’t build for macOS. But I’ll check tmrw and CI should also.

jcollie:
> This might have to be scoped to Linux only. I thought it doesn’t build for macOS. But I’ll check tmrw and CI should also.

Whoops yep. Fixed.

PR #8534, opened 2025-09-05
apprt/gtk: don't use Stacked for surface error status page

### LINKED ISSUE #8533: gtk: repeated splits hide surfaces
### Reproduction
1. Split right or down a few times in a row
2. At some point, new surfaces stop appearing (but they definitely are getting created)
3. Close any surface
4. All surfaces should disappear (see note below)

**Note:** Whether (4) happens seems to depend on window size. On my machine, the default sized window doesn't cause this, but when I make it larger or maximize it, all surfaces go poof (they still exist but are invisible). For smaller windows, the invisible surfaces seem to be placed underneath(?), as they start showing up when I close all the visible ones.

**Debug build logs:** https://gist.github.com/trag1c/735924511e842b0b9fd5a3c212b9bfdc

mitchellh:
Fixes #8533

Replace the usage of `Stacked` for error pages with programmatically swapping the child of the `adw.Bin`.

I regret to say I don't know the root cause of this. I only know that the usage of `Stacked` plus `Gtk.Paned` and the way we programmatically  change the paned position and stack child during initialization causes major issues.

This change isn't without its warts, too, and you can see them heavily commented in the diff. 

(1) We have to workaround a GTK template double-free bug that is well known  to us: if you bind a template child that is also the direct child of the  template class, GTK does a double free on dispose. We workaround this by removing our child in dispose. Valgrind verifies the fix.

(2) We have to workaround an issue where setting an `Adw.Bin` child during a glarea realize causes some kind of critical GTK error that results in a hard crash. We delay changing our bin child to an idle tick.

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -496,6 +496,9 @@ pub const Surface = extern struct {
         /// if this is true, then it means the terminal is non-functional.
         @"error": bool = false,
 
+        /// The source that handles setting our child property.
+        idle_rechild: ?c_uint = null,
+
         /// A weak reference to an inspector window.
         inspector: ?*InspectorWindow = null,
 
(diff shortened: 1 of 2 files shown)


mitchellh:
Note I tested both the error and non-error case. The error case I found an interesting memory leak but I'm going to move that to another PR because the fix isn't clear, it's unrelated to this, and it's only hit if our gl area fails.

jcollie (approved):
LGTM

PR #8535, opened 2025-09-05
apprt/gtk: set the title on the window immediately if set

### LINKED ISSUE #5934: `title` config should be respected and set immediately on new windows, tabs

mitchellh:
Fixes #5934

This was never confirmed to be a real issue on GTK, but it is theoretically possible and good hygiene in general. Typically, we'd get the title through a binding which comes from a bindinggroup which comes from the active surface in the active tab. All of this takes multiple event loop ticks to settle, if you will.

This commit changes it so that if an explicit, static title is set, we set that title on startup before the window is mapped. The syncing still happens later, but at least the window will have a title from the initialization.

### DIFF
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -306,6 +306,13 @@ pub const Window = extern struct {
             const config = config_obj.get();
             if (config.maximize) self.as(gtk.Window).maximize();
             if (config.fullscreen) self.as(gtk.Window).fullscreen();
+
+            // If we have an explicit title set, we set that immediately
+            // so that any applications inspecting the window states see
+            // an immediate title set when the window appears, rather than
+            // waiting possibly a few event loop ticks for it to sync from
+            // the surface.
+            if (config.title) |v| self.as(gtk.Window).setTitle(v);
         }
 
         // We always sync our appearance at the end because loading our


jcollie (approved):
LGTM. This should make a number of tiling wm ricers happy.

00-kat:
> This was never confirmed to be a real issue on GTK

For what it's worth, I'm currently running a commit that **doesn't** include this PR (d2ac29c), and I can definitely see that the title is initially `Ghostty` when running `hyprctl activewindow` inside a `ghostty --config-default-files=false --title=foo` window:
```yaml
class: com.mitchellh.ghostty
title: foo
initialClass: com.mitchellh.ghostty
initialTitle: Ghostty
```

mitchellh:
@00-kat can you confirm this PR fixes it?

00-kat:
Doesn't look like it has; I see the following output on e4c3a56 (current latest commit):
```yaml
class: com.mitchellh.ghostty
title: foo
initialClass: com.mitchellh.ghostty
initialTitle: Ghostty
```

mitchellh:
Very strange. I’ll have to try to reproduce this myself because I’m not sure how we can set the title any earlier.

tlj:
> Very strange. I’ll have to try to reproduce this myself because I’m not sure how we can set the title any earlier.

@mitchellh, sorry for commenting on a closed issue. It seems like both issues related to this are closed with resolutions that didn't fix the issue.

. It seems like the problem stems from the _tab_ somehow setting the title before the actual _window_ (?). 

I experimented with changing a section of [tab.zig L391](https://github.com/ghostty-org/ghostty/blob/f2c7f4ec0f7882a30abcd3496c9044a424083447/src/apprt/gtk/class/tab.zig#L391) and that had the desired effect: `niri msg event-stream` reports my given --title as the only title it sets, and my niri window-rule also picks it up.

I am not a zig developer and I don't think the way I fixed it is the correct way, because I feel like something further up should have set the title already, but I'm not able to figure out what. Maybe you will know intuitively. 

Here's the change I did from L391 - just fetching the title from config and using it as the last fallback. Let me know if you want me to start a PR for this or if you have a better way already. 

```zig
const config_title: ?[*:0]const u8 = blk: {
            if (config_) |cfg_ptr| {
                const cfg = cfg_ptr.get();
                if (cfg.title) |t| break :blk t;
            }
            break :blk null;
        };

        const plain = plain: {
            const default = "Ghostty";
            const plain = override_ orelse
                terminal_ orelse
                config_title orelse
                break :plain default;
            break :plain std.mem.span(plain);
        };
```

jcollie:
@tlj please do not post on old PRs like this. Start a new discussion and reference old PRs, issues, or discussions if needed.

mitchellh:
Fwiw @tlj that's super helpful, I'd accept a PR for something this specific and we can jam on it there.

PR #8536, opened 2025-09-05
Micro-optimize GlyphKey Context

j-c-m:
Use fast hash function on key for better distribution.

Direct compare glyph in eql to avoid Packed.from() if not neccessary.

16% -> 6.4% reduction during profiling runs.

### DIFF
--- a/src/font/SharedGrid.zig
+++ b/src/font/SharedGrid.zig
@@ -332,11 +332,12 @@ const GlyphKey = struct {
 
     const Context = struct {
         pub fn hash(_: Context, key: GlyphKey) u64 {
-            return @bitCast(Packed.from(key));
+            const packed_key = Packed.from(key);
+            return std.hash.CityHash64.hash(std.mem.asBytes(&packed_key));
         }
 
         pub fn eql(_: Context, a: GlyphKey, b: GlyphKey) bool {
-            return Packed.from(a) == Packed.from(b);
+            return a.glyph == b.glyph and Packed.from(a) == Packed.from(b);
         }
     };
 
--- a/src/font/SharedGrid.zig
+++ b/src/font/SharedGrid.zig
@@ -332,11 +332,15 @@ const GlyphKey = struct {
 
     const Context = struct {
         pub fn hash(_: Context, key: GlyphKey) u64 {
+            // Packed is a u64 but std.hash.int improves uniformity and
+            // avoids collisions in our hashmap.
             const packed_key = Packed.from(key);
-            return std.hash.CityHash64.hash(std.mem.asBytes(&packed_key));
+            return std.hash.int(@as(u64, @bitCast(packed_key)));
         }
 
         pub fn eql(_: Context, a: GlyphKey, b: GlyphKey) bool {
+            // Packed checks glyphs but in most cases the glyphs are NOT
+            // equal so the first check leads to increased throughput.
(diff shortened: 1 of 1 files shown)


src/font/SharedGrid.zig
```diff
@@ -332,11 +332,12 @@ const GlyphKey = struct {
 
     const Context = struct {
         pub fn hash(_: Context, key: GlyphKey) u64 {
-            return @bitCast(Packed.from(key));
+            const packed_key = Packed.from(key);
+            return std.hash.CityHash64.hash(std.mem.asBytes(&packed_key));
```
mitchellh: How is hashing the value improving throughput? This seems to be doing strictly more work.

src/font/SharedGrid.zig
```diff
@@ -332,11 +332,12 @@ const GlyphKey = struct {
...
         }
 
         pub fn eql(_: Context, a: GlyphKey, b: GlyphKey) bool {
-            return Packed.from(a) == Packed.from(b);
+            return a.glyph == b.glyph and Packed.from(a) == Packed.from(b);
```
mitchellh, line 344: This makes sense. It's slower in the equal case but faster in the unequal case which is more common, presumably?

mitchellh (changes requested):
Couple questions. I appreciate this work though.

src/font/SharedGrid.zig
qwerasd205: I would like to see the benchmark change of this specifically. It might improve uniformity, which will reduce clustering, but in that case we should probably use `std.hash.int`, rather than cityhash, since that's specifically what it's designed for.

mitchellh: > It might improve uniformity, which will reduce clustering,

Ah duh, forgot for a second this was going into a hash table and tunnel visioned on the equality part of it.

src/font/SharedGrid.zig
j-c-m: yes, the overwhelmingly common case is the glyph's are different and we avoid packing both. It could be further unrolled like I did with the Style eql, but that didn't have such a massive and fast u32 cmp common case.

src/font/SharedGrid.zig
j-c-m:
```
zig run benchmark_glyph_hash.zig -Doptimize=ReleaseFast
Hash Function Benchmarks (30000 inserts + 90000 lookups):
[SUGGESTED] std.hash.int:              7123083 ns
[PR]        std.hash.CityHash64:       8934916 ns
[CURRENT]   @bitCast(Packed.from(key)):2733519375 ns
```

Yes it does look like std.hash.int would be better, I just overlooked this as I am new to Zig.

https://gist.github.com/j-c-m/074081c3133841563a3921116c492845

(Mostly AI written benchmark)

j-c-m:
The change to std.hash.int has improved the performance/efficiency further. Profiling the same task from 16% to 4%

src/font/SharedGrid.zig
mitchellh: Thanks, thanks for the disclosure. I reviewed the benchmark and it isn't ideal but I think its perfectly adequate for the task at hand, and the changes are so minimal and human understandable I don't need it to be perfect.

mitchellh (approved):

mitchellh:
I added comments so future work in this area can recognize what's happening here.

PR #8538, opened 2025-09-05
fix(terminal): move cursor left if VS15 makes cell narrow

qwerasd205:
Without this change, a phantom space appears after any character with default emoji presentation that is converted to text with VS15. The only other terminal I know of that respects variation selectors is Kitty, and it walks the cursor back, which feels like the best choice, since that way the behavior is observable (no way to know if the terminal supports variation selectors otherwise without hard-coding that info per term) and "dumb" programs like `cat` will output things correctly, and not gain a phantom space after any VS15'd emoji.

> [!NOTE]
> ### Tests should be added for this behavior, including edge cases like with cursor pending wrap


mitchellh:
I added tests. The pending wrap state was indeed broken.

mitchellh (approved):

PR #8539, opened 2025-09-05
gtk: why codegen when you can embed

pluiedev:
When you do `glib-compile-resources` with `--generate-header` or`--generate-source`, it generates really long and stinky C source files littered with lengthy escape codes, macros and compiler directives, just because C doesn't (yet!) have a standard way to just embed a file as a literal bunch of bytes.

However, we do have access to that in Zig. So why not just do the same thing but better?

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -1673,19 +1673,20 @@ pub const Application = extern struct {
 
         fn init(class: *Class) callconv(.c) void {
             // Register our compiled resources exactly once.
-            {
-                const c = @cImport({
-                    // generated header files
-                    @cInclude("ghostty_resources.h");
+            const gresource = @embedFile("gresource");
+            const bytes = glib.Bytes.newStatic(gresource, gresource.len);
+
+            var err: ?*glib.Error = null;
+            if (gio.Resource.newFromData(bytes, &err)) |resource| {
+                gio.resourcesRegister(resource);
+            } else {
+                // If we fail to load resources then things will
+                // probably look really bad but it shouldn't stop our
+                // app from loading.
+                log.warn("unable to load resources={?s} ({})", .{
+                    err.?.f_message,
+                    err.?.f_code,
                 });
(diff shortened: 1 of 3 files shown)


mitchellh (commented):
Since this is part of the startup path, I’d be curious how much slower it is to load them via bytes like this versus compiling the C files which presumably stores them in some ready format? (I’m not actually sure what they do)

If this affects startup time it’s probably not worth it.

pluiedev:
The C file is essentially a big byte array plus some shims to make it work with the gio.Resource API. More specifically they're kind of using a hack (gio.StaticResource) so that you don't have to go through glib.Bytes, but given this is only ever called once I don't think this will change anything

pluiedev:
More specifically it looks like this:
```c
#include <gio/gio.h>

#if defined (__ELF__) && ( __GNUC__ > 2 || (__GNUC__ == 2 && __GNUC_MINOR__ >= 6))
# define SECTION __attribute__ ((section (".gresource.ghostty"), aligned (sizeof(void *) > 8 ? sizeof(void *) : 8)))
#else
# define SECTION
#endif

static const SECTION union { const guint8 data[1805701]; const double alignment; void * const ptr;}  ghostty_resource_data = {
  "\107\126\141\162\151\141\156\164\000\000\000\000\000\000\000\000"
/* A LOT OF RAW STRING LITERALS */
};

static GStaticResource static_resource = { ghostty_resource_data.data, sizeof (ghostty_resource_data.data) - 1 /* nul terminator */, NULL, NULL, NULL };

G_MODULE_EXPORT
GResource *ghostty_get_resource (void);
GResource *ghostty_get_resource (void)
{
  return g_static_resource_get_resource (&static_resource);
}
```

mitchellh:
I’d like to measure this before merging to be sure. 

It may be a hack but it’s not our hack, it’s an official hack even so I don’t mind it. Startup time matters.

mitchellh:
Going to close this, I think there isn't any practical reason to do this at the moment.

PR #8542, opened 2025-09-06
macOS: opt out of probable_cli

danielwe:
Suggested fix for the quirk reported in https://github.com/ghostty-org/ghostty/pull/8511#issuecomment-3261123413, i.e., that the working directory of `open -a Ghostty.app` and `open -a Ghostty.app --args <something>` are different.

The `probable_cli` detection seems poorly suited for macOS because the only CLI-like way to launch there is with `open -a`, which invokes the launch system and starts a detached process with no working directory to inherit. It seems more akin to a desktop/systemd launch. So, until it's possible to directly invoke a `ghostty` executable on the command line in macOS, my suggestion is to have this function always return false there.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4535,6 +4535,11 @@ fn probableCliEnvironment() bool {
         // its not a real supported target and GTK via WSL2 assuming
         // single instance is probably fine.
         .windows => return false,
+        // On macOS there is no direct CLI launch, only `open -a`,
+        // which detaches the process and doesn't provide a working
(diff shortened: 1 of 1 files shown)


mitchellh:
Ghostty does support CLI launching by running the `ghostty` binary in the app bundle currently. 

You're right in identifying this as a problem though. The better solution is to check `launchedByDesktop` for macOS and invert it because unlike Linux, the desktop detection for macOS is pretty much 100% solid and infallible due to launch services.

mitchellh:
Replaced by #8544

PR #8544, opened 2025-09-06
macOS: probable cli should return false on macOS for desktop launch

mitchellh:
Fixes #8542

The comment explains why this is needed.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4535,6 +4535,13 @@ fn probableCliEnvironment() bool {
         // its not a real supported target and GTK via WSL2 assuming
         // single instance is probably fine.
         .windows => return false,
+
+        // On macOS, we don't want to detect `open` calls as CLI envs.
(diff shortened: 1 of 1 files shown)


mitchellh:
Importantly, this lets directly executing the CLI work properly still.

danielwe:
If desktop launch detection is 100 % reliable, should this just always return `!internal_os.launchedFromDesktop()`, rather than falling through to the envvar/argv heuristics when `launchedFromDesktop()` is false?

mitchellh:
> If desktop launch detection is 100 % reliable, should this just always return !internal_os.launchedFromDesktop(), rather than falling through to the envvar/argv heuristics when launchedFromDesktop() is false?

I thought that too and I don't know. I was worried about situations where another program might be sub processing Ghostty and sanitizing the environment. I don't know if that's realistic. I'd rather wait for some known cases on either side...

PR #8550, opened 2025-09-06
fix(font): Ensure face is fully contained within cell

danielwe:
The alignment of the baseline with a pixel boundary would, half of the time, make the face protrude slightly below the cell. This fixes that such that the face is always fully contained within the cell, at the cost of sometimes making the cell a pixel taller.

I'm actually not sure whether this or the existing behavior is preferable, but figured a small PR implementing the change is the easiest and most precise way of asking.

Pros of the existing implementation:
* Minimizes the line height padding added by pixel rounding
* However much the face protrudes below the cell, it will always have an equally large or larger margin on top, so there's no actual overlap; it's just that the true line boundary doesn't perfectly align with a pixel boundary

Cons of the existing implementation, addressed by this PR:
* Glyphs with descenders will sometimes protrude below the cursor and highlighting, since the latter are sized and aligned to the cell, not the face
* Even though the font outlines from two lines will never overlap, they may touch the same pixels, so after rendering, the top pixel row of an ascender may overlap with the bottom pixel row of a descender from the line above

Curious to hear what people think.

### DIFF
--- a/src/font/Collection.zig
+++ b/src/font/Collection.zig
@@ -1203,23 +1203,26 @@ test "metrics" {
 
     try comparison.expectApproxEqual(font.Metrics{
         .cell_width = 8,
-        // The cell height is 17 px because the calculation is
+        // The cell height is 18 px because the calculation is
         //
         //  ascender - descender + gap
         //
         // which, for inconsolata is
         //
-        //  859 - -190 + 0
+        //  859 - (-190) + 0
         //
         // font units, at 1000 units per em that works out to 1.049 em,
         // and 1em should be the point size * dpi scale, so 12 * (96/72)
-        // which is 16, and 16 * 1.049 = 16.784, which finally is rounded
-        // to 17.
+        // which is 16, and 16 * 1.049 = 16.784.
         //
-        // The icon height is (2 * cap_height + face_height) / 3
-        // = (2 * 623 + 1049) / 3 = 765, and 16 * 0.765 = 12.24.
-        .cell_height = 17,
-        .cell_baseline = 3,
+        // The unrounded baseline is at
+        // -descender + (gap / 2) = -(-190) + 0 = 0.19 em. In pixels,
+        // that's 16 * 0.19 = 3.04, which rounds up to 4, so the
+        // bottom padding to align the baseline with a pixel boundary
+        // is 4 - 3.04 = 0.96. Hence, the top of the face is at
+        // 16.784 + 0.96 = 17.744, which finally is rounded up to 18.
+        .cell_height = 18,
+        .cell_baseline = 4,
         .underline_position = 17,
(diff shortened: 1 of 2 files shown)


danielwe:
Actually, maybe a better approach is to only do this if the face's line gap is less than a pixel. Otherwise, there's no issue.

danielwe:
Pushed an update such that the correction is only applied when actually needed to keep ascenders and descenders inside the cell.

mitchellh:
Other PRs were merged first, can you rebase this please. 😄 Thank you!

src/font/Collection.zig
```diff
@@ -1203,55 +1203,58 @@ test "metrics" {
...
         .overline_position = 0,
         .overline_thickness = 2,
         .box_thickness = 2,
-        .cursor_height = 34,
+        .cursor_height = 35,
         .icon_height = 24.48,
         .face_width = 16.0,
         .face_height = 33.568,
-        .face_y = -0.08,
+        .face_y = 0.92,
```
mitchellh, line 1257: Sorry @danielwe can you review this change from a rebase? This seems suspicious that this changed so wildly.

mitchellh (changes requested):

src/font/Collection.zig
danielwe: Nope, that's correct---it's exactly the change this PR implements: adding an extra pixel row on the bottom if needed to ensure that `cell_baseline >= -face.descent`, or equivalently, `face_y + half_line_gap >= 0`. This font has `half_line_gap == 0`, so this simplifies to `face_y >= 0`. Since it was negative before, we should expect it to increase by 1 with this PR, which is what we see here.

That said, as explained in the OP, I'm kind of 50/50 on whether this is the correct thing to do. What it does is ensure that well-behaved glyphs stay within the terminal's grid cell, so they never draw on the same pixel row as well-behaved glyphs on the next line, and never protrude below selection/cursor highlighting. (By well-behaved here I mean that the outlines don't overshoot the font's ascent and descent metrics.) The price is that some fonts at some sizes end up with a cell height of `@ceil(line_height) + 1` instead of just `@ceil(line_height)`.

danielwe: And in case it was unclear earlier: The reason `@ceil(line_height) + 1` is sometimes needed to avoid overshoot is that we round the baseline to a pixel boundary. If instead we simply centered the face vertically within the cell, `@ceil(line_height)` would always be sufficient.

danielwe:
Debating closing vs. rebasing this, but I'll add a screenshot showing the problem(?) it solves first. This PR would add another row of pixels at the bottom of the cell for this particular font, font size, and DPI, avoiding the protruding descenders. 
Screenshot 2025-10-30 at 12 36 39

PR #8551, opened 2025-09-06
ci: do not run release tip workflow if corresponding commit released

### LINKED ISSUE #8549: ci: "release tip" should not overwrite already-released commits
Currently, the "release tip" job will overwrite previously released commits if run again. This breaks downstream package managers that are trying to hardcode checksums because the data can change underneath them expectedly.

**We should perform a check early on with "release tip" and bail if the expected artifact already exists.**

Perfectly reproducible builds would be better but that's a much larger, difficult task (maybe impossible with Xcode). So, let's do the simple thing here.

mitchellh:
Fixes #8549

This also brings the release tip workflow more in line with the release tag workflow by using a setup job to create outputs that are reused by the other jobs.

This PR was almost fully written by AI (Amp) because being a YAML engineer fucking sucks. I understand GHA and the changes look good to me, but it's hard to tell until the job is run, AI or not. Full prompt/context here:

https://ampcode.com/threads/T-e2d431ad-8be8-46d2-aaa3-9fae71f9ff31

### DIFF
--- a/.github/scripts/ghostty-tip
+++ b/.github/scripts/ghostty-tip
@@ -1,0 +1,19 @@
+#!/usr/bin/env nu
+
+# Check if a given commit SHA has a corresponding tip release.
+#
+# This does not validate that the commit SHA is valid for the
+# Ghostty repository, only that a tip release exists for it.
+def main [
+  commit: string, # The full length commit SHA 
+] {
+  let url = $"https://tip.files.ghostty.org/($commit)/ghostty-macos-universal.zip"
+
+  try {
+    http head $url
+    exit 0
+  } catch {
+    print -e $"The SHA ($commit) does not have a corresponding tip release."
+    exit 1
+  }
+}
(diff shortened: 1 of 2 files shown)


mitchellh:
Manually triggered a release job on this branch to verify it: https://github.com/ghostty-org/ghostty/actions/runs/17519807423/job/49762147137

mitchellh:
This didn't work and I'm not quite sure why yet. The setup job on run 2 triggered the "already exists" path properly but the subsequent jobs are all still running. Manually reviewed everything in the diff and looks right to me... probably some dumb GHA detail. Looking into it.

mitchellh:
Okay, pushed some changes:

- Apparently outputs can't contain hyphens, so changed to underscores
- Multiline if conditionals are a disaster: https://github.com/orgs/community/discussions/25641 so switched to non-expressions and hoping for the best.

I found both of these but guided the AI to fix it.

mitchellh:
We've done it. 

CleanShot 2025-09-06 at 20 33 21@2x

PR #8552, opened 2025-09-06
fix(font): No NF icon smoothing

danielwe:
This disables CoreText font smoothing for glyphs with nontrivial constraints, which, for the most part, means Nerd Font icons (also emojis, but they are bitmaps, so they were never smoothed anyway).

This is the right thing to do because icons aren't like other glyphs, and font smoothing may make their details less crisp (see, for example, the Nix snowflake and the TeX logo). Perhaps even more importantly, font smoothing interferes badly with icon scaling because it adds a 1-pixel border to the glyph bounding box, causing the scaling algorithm to perceive the icon as larger than it is and therefore scale it to be smaller than it should be. The magnitude of this effect depends on font size and monitor DPI, and is greater on lower-DPI monitors using smaller font sizes.

I also disabled synthetic bold and italic for icons (except italic in CoreText, as it applies per face, and I didn't see a way to opt out per glyph). This is a more hypothetical concern, but it's possible to set the primary font to a patched nerd font that lacks bold and/or italic, in which case you wouldn't want the synthesizing to apply to the icon glyphs.

Lastly, a fly-by bugfix: quantizing bitmap glyphs should be done after glyph centering, as the latter is generally fractional and would negate the quantizing.

### DIFF
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -382,16 +382,6 @@ pub const Face = struct {
         var width = glyph_size.width;
         var height = glyph_size.height;
 
-        // If this is a bitmap glyph, it will always render as full pixels,
-        // not fractional pixels, so we need to quantize its position and
-        // size accordingly to align to full pixels so we get good results.
-        if (sbix) {
-            width = cell_width - @round(cell_width - width - x) - @round(x);
-            height = cell_height - @round(cell_height - height - y) - @round(y);
-            x = @round(x);
-            y = @round(y);
-        }
-
         // If the cell width was adjusted wider, we re-center all glyphs
         // in the new width, so that they aren't weirdly off to the left.
         if (metrics.original_cell_width) |original| recenter: {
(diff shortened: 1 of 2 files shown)


qwerasd205:
I don't know that this is the correct solution. I think maybe we'd be better off just calculating the constraints on the original bounding box and then adding 1px of padding to the raster surface if thickening is enabled.

I do agree a lot of things look better without thickening, but I don't know whether we want to make that judgement call for users or not. I'll have to give it some thought.

danielwe:
Adding the pixel padding after applying the constraints certainly works to fix the scaling issue. I guess if users were using a patched font in a less fancy terminal, thickening would apply equally to icons and other glyphs, so I definitely see the case for not being too clever with this. Let me know what you decide, and I'll push whatever changes are needed.

Do you agree with the synthetic bold/italic opt-out, or would you rather just give the user what they asked for there as well?

danielwe:
Alternative branch ready to go here: https://github.com/ghostty-org/ghostty/compare/main...danielwe:ghostty:constraints_before_thickening. No opting out of anything, only reordering operations to avoid the impact on scaling and alignment.

danielwe:
Maybe let's look at some screenshots. Here's a file listing with a selection of more or less detailed icons. Ghostty is running with config `font-thicken = true` and nothing else on a 1920x1080 monitor. Both versions include some extra icon scaling patches that I'll submit later, however, the only difference is that one of them includes this PR (no thickening for icons) and the other includes the alternative branch I linked above (thickening for all glyphs, but scaling issues fixed).

Screenshot 2025-09-06 at 22 29 10Screenshot 2025-09-06 at 22 26 57

danielwe:
How about we just go for the other version without any special-case opt-out, at least for now? That's a plain bugfix and should be uncontroversial. The argument against trying to be too clever is growing on me.

I'll submit that branch as a new PR and close this one.

PR #8555, opened 2025-09-07
Position-independent font shaper caching

j-c-m:
Use relative cluster positioning to allow identical texts runs in different row positions to share the same cache entry.

I am opening this PR clean w/o the cache size change. There could be some benefit to a larger 256->512 shaper cache, but this still performs amazingly well and I don't know the full memory impacts of moving the cache size up.

https://github.com/ghostty-org/ghostty/discussions/8547#discussioncomment-14329590

### DIFF
--- a/src/font/shaper/run.zig
+++ b/src/font/shaper/run.zig
@@ -23,7 +23,9 @@ pub const TextRun = struct {
     /// to share the same cache entry, improving cache efficiency.
     hash: u64,
 
-    /// The offset in the row where this run started
+    /// The offset in the row where this run started. This is added to the
+    /// X position of the final shaped cells to get the absolute position
+    /// in the row where they belong.
     offset: u16,
 
     /// The total number of cells produced by this run.
--- a/src/font/shaper/run.zig
+++ b/src/font/shaper/run.zig
@@ -17,6 +17,10 @@ pub const TextRun = struct {
     /// lower the chance of hash collisions if they become a problem. If
     /// there are hash collisions, it would result in rendering issues but
     /// the core data would be correct.
+    ///
+    /// The hash is position-independent within the row by using relative
+    /// cluster positions. This allows identical runs in different positions
+    /// to share the same cache entry, improving cache efficiency.
     hash: u64,
 
     /// The offset in the row where this run started
@@ -77,7 +81,11 @@ pub const RunIterator = struct {
         // Go through cell by cell and accumulate while we build our run.
         var j: usize = self.i;
         while (j < max) : (j += 1) {
-            const cluster = j;
+            // Use relative cluster positions (offset from run start) to make
+            // the shaping cache position-independent. This ensures that runs
+            // with identical content but different starting positions in the
+            // row produce the same hash, enabling cache reuse.
+            const cluster = j - self.i;
             const cell = &cells[j];
 
             // If we have a selection and we're at a boundary point, then
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -2528,9 +2528,11 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                                 break :cache cells;
                             };
 
+                        const cells = shaper_cells.?;
+
                         // Advance our index until we reach or pass
                         // our current x position in the shaper cells.
-                        while (shaper_cells.?[shaper_cells_i].x + run.offset < x) {
+                        while (run.offset + cells[shaper_cells_i].x < x) {
                             shaper_cells_i += 1;
                         }
                     }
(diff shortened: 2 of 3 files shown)


mitchellh (commented):
Can you quantify the improvement?

If @qwerasd205 is comfortable with this, I'm good getting this in right away.

src/font/shaper/run.zig
```diff
@@ -263,11 +271,7 @@ pub const RunIterator = struct {
             }
 
             // Add all the codepoints for our grapheme
-            try self.addCodepoint(
-                &hasher,
-                if (cell.codepoint() == 0) ' ' else cell.codepoint(),
-                @intCast(cluster),
-            );
+            try self.addCodepoint(&hasher, if (cell.codepoint() == 0) ' ' else cell.codepoint(), @intCast(cluster));
```
qwerasd205: Why change the formatting here?

src/renderer/generic.zig
```diff
@@ -2530,7 +2530,7 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
 
                         // Advance our index until we reach or pass
                         // our current x position in the shaper cells.
-                        while (shaper_cells.?[shaper_cells_i].x < x) {
+                        while (shaper_cells.?[shaper_cells_i].x < x - run.offset) {
```
qwerasd205: This should be rephrased as <...>`.x + run.offset < x`, to avoid an underflow if x is ever less than run.offset.

src/renderer/generic.zig
```diff
@@ -2769,13 +2769,13 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                         // If we encounter a shaper cell to the left of the current
                         // cell then we have some problems. This logic relies on x
                         // position monotonically increasing.
-                        assert(cells[shaper_cells_i].x >= x);
+                        assert(cells[shaper_cells_i].x >= x - run.offset);
```
qwerasd205: Same here, switch from a subtraction on the right to an addition on the left.

src/renderer/generic.zig
```diff
@@ -2769,13 +2769,13 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
...
 
                         // NOTE: An assumption is made here that a single cell will never
                         // be present in more than one shaper run. If that assumption is
                         // violated, this logic breaks.
 
-                        while (shaper_cells_i < cells.len and cells[shaper_cells_i].x == x) : ({
+                        while (shaper_cells_i < cells.len and cells[shaper_cells_i].x == x - run.offset) : ({
```
qwerasd205: And here.

qwerasd205 (changes requested):
Some style notes, but the core idea of the change feels safe to me. Before this is merged though, I would like to test it out myself and double check for anywhere this change might affect that might have been overlooked. I should have time to do that later today hopefully.

src/font/shaper/run.zig
j-c-m: zig fmt, don't really like the change but didnt want to get held up in the check.

jcollie: Add a trailing comma and `zig fmt` will format it the way it was before.

j-c-m:
Cleaned up the underflow issue, will have to remind my brain going forward zig is C.

src/font/shaper/run.zig
j-c-m: @jcollie thank you! I couldn't figure out why that magically changed on me.

j-c-m:
> Can you quantify the improvement?
> 

Loosely..

For the changes here, one of our worst case stress tests, cmatrix, running in a 105x35 window dropped on my M2 from ~14% CPU to ~8% and from coretext.shaper being at the top of the profiler to background noise.

In more realistic cases, scrolling syntax highlighted code in vim, just much more effective and efficient with cache hits. I don't have anything created here so it was a manual situation of me just scrolling up and down 10 pages watching the cache hits log & profiler.

For a larger cache, nothing concrete there, the 256/8 cache still seems very effective as thats what I've been stress testing here. I think going forward its something to look at combined with slightly more intelligent run splitting to promote cache hits (but that has to balance with efficient use of actual shaping when needed).

This is my fun test -

image

qwerasd205:
Unfortunately my day was longer than I anticipated, but I'll give this a proper review tomorrow, excited to merge this though since it looks like a pure perf win with a very simple change.

qwerasd205:
I did a benchmark. I used the existing (commented out) `rebuildCells` timing code and did 3 things:
1. `DOOM-fire-zig`
2. `cmatrix -arbu3`
3. `cmatrix -arbu0`

Left is main, right is with this PR's change, vertical axis is µs
image
As you can see, in the worst performing sections, on my hardware with the window size I was testing with, this change cuts `rebuildCells` times down to one *sixth* of what they were! Awesome!

I have a couple of comment changes and a style change (`run.offset + cell.x` rather than `cell.x + run.offset`, it just feels more logical to me that way) that I'm gonna add to this PR but then I'll approve.

qwerasd205 (approved):

j-c-m:
looks great! and we can't move run.offset any further left. 🤣

PR #8556, opened 2025-09-07
ci: cleanups

mitchellh:
* Only update appcast (trigger macOS updates) on `main`-branch triggers
* Echo the release URLs for download as part of the job
* Remove the `release-pr` workflow. We can now use `release-tip` manually dispatched on a branch because it won't update the appcast.


mitchellh:
Did these ones manually (no AI) because the YAML was straightforward and mostly copy/paste from other parts.

PR #8557, opened 2025-09-07
fix triggering the snap workflow

jcollie:


jcollie:
Well, fixed one problem, now there's another. Looks like the github token needs write permissions of some kind.

image

mitchellh:
I already pushed this to main a few minutes ago.

PR #8559, opened 2025-09-07
ci/snap: also specify the run id

jcollie:


.github/workflows/snap.yml
```diff
@@ -22,9 +26,11 @@ jobs:
       ZIG_GLOBAL_CACHE_DIR: /zig/global-cache
     steps:
       - name: Download Source Tarball Artifacts
-        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
+        uses: actions/download-artifact@v11
```
mitchellh: We should change this to a hash pin before merging

.github/workflows/test.yml
```diff
@@ -281,9 +281,9 @@ jobs:
 
       - name: Trigger Snap workflow
         run: |
-          gh workflow run snap.yml --ref ${{ github.ref_name || 'main' }} --field source-tarball-id=${{ needs.build-dist.outputs.artifact-id }}
+          gh workflow run snap.yml --ref ${{ github.ref_name || 'main' }} --field source-run-id=${{ github.event.workflow_run.id }} --field source-artifact-id=${{ needs.build-dist.outputs.artifact-id }}
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
```
mitchellh: Yeah this should work for this. The secret one has more privileges we use for releases

mitchellh (changes requested):

.github/workflows/snap.yml
jcollie: Yeah, fixed. I was initially looking at using a different action but forgot to put the hash back.

jcollie:
OK, I think that I've taken this as far as I can, looks like the github tokens that I have access to don't allow me to manually trigger a workflow.

.github/workflows/test.yml
```diff
@@ -281,9 +281,9 @@ jobs:
 
       - name: Trigger Snap workflow
         run: |
-          gh workflow run snap.yml --ref ${{ github.ref_name || 'main' }} --field source-tarball-id=${{ needs.build-dist.outputs.artifact-id }}
+          gh workflow run snap.yml --ref ${{ github.ref_name || 'main' }} --field source-run-id=${{ github.run_id }} --field source-artifact-id=${{ needs.build-dist.outputs.artifact-id }}
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_TOKEN }}
```
mitchellh, line 286: This isn't necessary. The other token works just fine. I'll bring the changes in manually!

PR #8575, opened 2025-09-09
fonts: fix CJK fallback font loading

recursiveforte:
Clearing font charset restrictions by copying a font descriptor's attributes and re-creating it sometimes doesn't work; with '.PingFangUITextSC-Regular' (derived from '.AppleSystemUIFontMonospaced' in `CoreText.discoverCodepoint(...)`), `face.hasCodepoint(cp, p_mode)` will erroneously return false after copying, causing the font to be rejected.

We should only be clearing charset restrictions if we applied them first (ie in `CoreText.discover(...)`); doing so lets us avoid this bug, but I'm not sure how fragile the charset restriction clearing is for other scenarios.

### DIFF
--- a/src/font/discovery.zig
+++ b/src/font/discovery.zig
@@ -8,6 +8,8 @@ const opentype = @import("opentype.zig");
 const options = @import("main.zig").options;
 const Collection = @import("main.zig").Collection;
 const DeferredFace = @import("main.zig").DeferredFace;
+const Face = @import("main.zig").Face;
+const Library = @import("main.zig").Library;
 const Variation = @import("main.zig").face.Variation;
 
 const log = std.log.scoped(.discovery);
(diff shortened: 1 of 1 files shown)


mitchellh:
If these are fonts built-in to macOS, then it'd be good to add a unit test here that demonstrates this issue.

recursiveforte:
Added those tests and realized this is probably because we're not supposed to be accessing system UI fonts by name.

Output from running a test where we try to look up a fallback font for a CJK codepoint & try to clear charset restrictions:
```
[I] ➜  ghostty git:(upstream-3) ✗ zig build test -Dtest-filter="discoverFallback"
test
└─ run ghostty-test 59/60 passed, 1 failed
error: 'font.discovery.test.coretext discoverFallback CJK codepoint' failed: 2025-09-10 13:53:31.792 ghostty-test[51978:12386835] CoreText note: Client requested name ".PingFangUITextSC-Regular", it will get TimesNewRomanPSMT rather than the intended font. All system UI font access should be through proper APIs such as CTFontCreateUIFontForLanguage() or +[NSFont systemFontOfSize:].
2025-09-10 13:53:31.792 ghostty-test[51978:12386835] CoreText note: Set a breakpoint on CTFontLogSystemFontNameRequest to debug.
/opt/homebrew/Cellar/zig/0.14.1/lib/zig/std/testing.zig:580:14: 0x10575e86f in expect (ghostty-test)
    if (!ok) return error.TestUnexpectedResult;
             ^
/Users/cheru/Code/ghostty/src/font/discovery.zig:1133:5: 0x10577445f in test.coretext discoverFallback CJK codepoint (ghostty-test)
    try testing.expect(face.hasCodepoint(0x4E2D, null));
    ^
error: while executing test 'os.TempDir.test_0', the following test command failed:
/Users/cheru/Code/ghostty/.zig-cache/o/d8ac3a9ccf171ecb22ca38e4e30bc420/ghostty-test --seed=0x9fe58469 --cache-dir=/Users/cheru/Code/ghostty/.zig-cache --listen=-
Build Summary: 49/51 steps succeeded; 1 failed; 59/60 tests passed; 1 failed
test transitive failure
└─ run ghostty-test 59/60 passed, 1 failed
error: the following build command failed with exit code 1:
/Users/cheru/Code/ghostty/.zig-cache/o/526a4ea34f8d95eb90ea2f3856ac80e8/build /opt/homebrew/Cellar/zig/0.14.1/bin/zig /opt/homebrew/Cellar/zig/0.14.1/lib/zig /Users/cheru/Code/ghostty /Users/cheru/Code/ghostty/.zig-cache /Users/cheru/.cache/zig --seed 0x9fe58469 -Z70fd58e5d3dc4a11 test -Dtest-filter=discoverFallback
```

pluiedev:
I don't know about the code quality here, but it would be great if someone on the macOS and/or font team could review this ASAP. #8712 has been impacting a lot of people since the 1.2 release and I'd love to see this land in 1.2.1.

src/font/discovery.zig
```diff
@@ -8,6 +8,8 @@ const opentype = @import("opentype.zig");
 const options = @import("main.zig").options;
 const Collection = @import("main.zig").Collection;
 const DeferredFace = @import("main.zig").DeferredFace;
+const Face = @import("main.zig").Face;
```
danielwe, line 11: Seems like this line isn't needed.
```suggestion
```

danielwe (commented):
I'm not on any team, but thought I could try contributing a pair of eyes in case it's helpful.

I'm a bit unclear on what problem this is fixing. The terminal seems to behave identically with or without this patch when pasting `中` to trigger a CJK fallback font lookup. The logs are identical, and the loaded font is PingFang SC in both cases. This is on macOS 26, with one manually installed CJK font in addition to the ones provided by the system.

That said, I can trigger the same test failure as quoted above by keeping the tests and removing the other changes.

> realized this is probably because we're not supposed to be accessing system UI fonts by name

@recursiveforte if I understand you correctly, you're saying that this patch is actually a workaround for erroneously specifying unsupported system font identifiers like `.AppleSystemUIFontMonospaced` as the font family name. Does Ghostty ever do that on its own, or would this workaround only be needed if a user puts `font-family = .AppleSystemUIFontMonospaced` in their config?

mitchellh:
Going to close this. It's unclear to me if this is still required and what problem this is fixing, as @danielwe noted. If I'm mistaken, please respond. Thank you.

danielwe:
It now looks to me like this was solving the same problem that #9649 solved. Since #9649 didn't include tests, one option would be to only poach the tests from here.

PR #8580, opened 2025-09-10
fix(font): Apply glyph constraints before thickening and centering before quantizing

danielwe:
In CoreText, when thickening (font smoothing) is enabled or Ghostty is synthesizing a bold face, the glyph bounding box is padded to make sure the thicker glyph can fit. Currently, this happens before applying constraints (scaling and alignment), which makes the size and position of constrained glyphs dependent on font size, font thickening strength, and display DPI.

With this PR, constraints are applied before any other adjustments, and padding is applied directly to the rasterization canvas without modifying any metrics.

For consistency, I also moved constraint application above emboldening in the FreeType code, although under that API, the two operations are orthogonal as far as I can tell.

Secondly, this PR moves glyph centering above bitmap quantization, as centering is generally fractional and will therefore undo the quantizing if done after.

Supersedes #8552.

### DIFF
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -378,16 +378,6 @@ pub const Face = struct {
         var width = glyph_size.width;
         var height = glyph_size.height;
 
-        // If this is a bitmap glyph, it will always render as full pixels,
-        // not fractional pixels, so we need to quantize its position and
-        // size accordingly to align to full pixels so we get good results.
-        if (sbix) {
-            width = cell_width - @round(cell_width - width - x) - @round(x);
-            height = cell_height - @round(cell_height - height - y) - @round(y);
-            x = @round(x);
-            y = @round(y);
-        }
-
         // We center all glyphs within the pixel-rounded and adjusted
         // cell width if it's larger than the face width, so that they
         // aren't weirdly off to the left.
@@ -398,6 +388,16 @@ pub const Face = struct {
         if ((opts.constraint.align_horizontal == .none) and (metrics.face_width < cell_width)) {
             // We add half the difference to re-center.
             x += (cell_width - metrics.face_width) / 2;
+        }
+
+        // If this is a bitmap glyph, it will always render as full pixels,
+        // not fractional pixels, so we need to quantize its position and
+        // size accordingly to align to full pixels so we get good results.
+        if (sbix) {
+            width = cell_width - @round(cell_width - width - x) - @round(x);
+            height = cell_height - @round(cell_height - height - y) - @round(y);
+            x = @round(x);
+            y = @round(y);
         }
 
         // Our whole-pixel bearings for the final glyph.
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -319,17 +319,6 @@ pub const Face = struct {
             rect.origin.y -= line_width / 2;
         };
 
-        // We make an assumption that font smoothing ("thicken")
-        // adds no more than 1 extra pixel to any edge. We don't
-        // add extra size if it's a sbix color font though, since
-        // bitmaps aren't affected by smoothing.
-        if (opts.thicken and !sbix) {
-            rect.size.width += 2.0;
-            rect.size.height += 2.0;
-            rect.origin.x -= 1.0;
-            rect.origin.y -= 1.0;
-        }
-
         // If our rect is smaller than a quarter pixel in either axis
         // then it has no outlines or they're too small to render.
         //
@@ -400,10 +389,16 @@ pub const Face = struct {
             y = @round(y);
         }
 
+        // We make an assumption that font smoothing ("thicken")
+        // adds no more than 1 extra pixel to any edge. We don't
+        // add extra size if it's a sbix color font though, since
+        // bitmaps aren't affected by smoothing.
+        const canvas_padding: u32 = if (opts.thicken and !sbix) 1 else 0;
+
         // Our whole-pixel bearings for the final glyph.
         // The fractional portion will be included in the rasterized position.
-        const px_x: i32 = @intFromFloat(@floor(x));
-        const px_y: i32 = @intFromFloat(@floor(y));
+        const px_x = @as(i32, @intFromFloat(@floor(x))) - @as(i32, @intCast(canvas_padding));
+        const px_y = @as(i32, @intFromFloat(@floor(y))) - @as(i32, @intCast(canvas_padding));
 
         // We keep track of the fractional part of the pixel bearings, which
         // we will add as an offset when rasterizing to make sure we get the
@@ -413,9 +408,9 @@ pub const Face = struct {
 
         // Add the fractional pixel to the width and height and take
         // the ceiling to get a canvas size that will definitely fit
-        // our drawn glyph, including the fractional offset.
-        const px_width: u32 = @intFromFloat(@ceil(width + frac_x));
-        const px_height: u32 = @intFromFloat(@ceil(height + frac_y));
+        // our drawn glyph, including the fractional offset and font smoothing.
+        const px_width = @as(u32, @intFromFloat(@ceil(width + frac_x))) + (2 * canvas_padding);
+        const px_height = @as(u32, @intFromFloat(@ceil(height + frac_y))) + (2 * canvas_padding);
 
         // Settings that are specific to if we are rendering text or emoji.
         const color: struct {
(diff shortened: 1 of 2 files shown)


mitchellh:
Merged some of your other PRs first which caused a conflict, can you please rebase this one.

danielwe:
Rebased and ready. Made a tiny change to avoid an unnecessary merge conflict between this #8847.

src/font/face/coretext.zig
```diff
@@ -400,22 +383,32 @@ pub const Face = struct {
...
+        const frac_x = x - @floor(x - canvas_padding);
+        const frac_y = y - @floor(y - canvas_padding);
 
         // Add the fractional pixel to the width and height and take
         // the ceiling to get a canvas size that will definitely fit
         // our drawn glyph, including the fractional offset.
-        const px_width: u32 = @intFromFloat(@ceil(width + frac_x));
-        const px_height: u32 = @intFromFloat(@ceil(height + frac_y));
+        const px_width: u32 = @intFromFloat(@ceil(width + frac_x + canvas_padding));
+        const px_height: u32 = @intFromFloat(@ceil(height + frac_y + canvas_padding));
```
qwerasd205: This should be adding `2 * canvas_padding` since the width and height need the padding added to both edges.

qwerasd205 (commented):
Why did you move the code that bails out for glyphs that are too small after all the (relatively) expensive constraint calculations?

Also, as a side note, I'm pretty sure if we're gonna move the FreeType outline emboldening code at all it should be moved *before* we measure the glyph (additional safety checks may need to be added to make sure that there are outlines to embolden in the first place).

src/font/face/coretext.zig
danielwe: No, because it's also added to `frac_x` and `frac_y` right above, as it must be for correct positioning:
```zig
        const frac_x = x - @floor(x - canvas_padding);
        const frac_y = y - @floor(y - canvas_padding);
```

qwerasd205: Oh, huh, that's an odd way to go about it. I'm not sure I like the way the padding is implemented but I'll think about it.

danielwe:
> Why did you move the code that bails out for glyphs that are too small after all the (relatively) expensive constraint calculations?

The idea was (since I was reordering this flow anyway) to make this call based on the actual size of the glyph after it's possibly been scaled and emboldened. In any case, if the constraint is nontrivial it seems unlikely that there's going to be cause for bailing, i.e., we'd probably never actually go through the expensive calculations only to bail right after (except if we've ended up with a font size so small that we bail on every glyph, in which case the user probably has bigger things to worry about than this inefficiency).

But happy to reorder and do this on the unscaled and unpadded rect if that's what you prefer!

danielwe:
> I'm pretty sure if we're gonna move the FreeType outline emboldening code at all it should be moved _before_ we measure the glyph

No, the point of this PR is precisely that constraints must be applied to the bbox of the unboldened/unthickened glyph to work correctly.

With Coretext, this is important for everyday usage---you don't want the size of NF icons to depend on the amount of font thickening.

With FreeType, it's perhaps a bit more academic, but it's totally possible for a user to specify a font that does not have a bold variant, but is a Nerd Fonts patched font. In that case, we will be generating synthetically emboldened NF icons (which is kind of dumb, but it is what it is), and an emboldened glyph should not be downscaled to fit in the same bbox as the unboldened glyph (imagine that for letters, it would look like the bold is a smaller font size than non-bold).

qwerasd205:
Font thickening and emboldening are completely different things.

qwerasd205:
I see your argument about bold symbols having the same basic proportions as non-bold ones with synthetic bold, but that's not the effect we'd get if a font *did* have a bold with symbols that have constraints applied, and it feels odd to me to be inconsistent about that.

Realistically, the symbols I'm more concerned about aren't the ones that are always scaled, but the ones that are simply constrained to cells-- I don't want synthetic bold versions of those symbols to overhang the edge(s) of their cell(s).

src/font/face/coretext.zig
danielwe: Consider the following code a little bit later:

```zig
        // Translate our drawing context so that when we draw our
        // glyph the bottom/left edge is at the correct sub-pixel
        // position. The bottom/left edges are guaranteed to be at
        // exactly [0, 0] relative to this because when we call to
        // `drawGlyphs`, we pass the negated bearings.
        context.translateCTM(
            ctx,
            frac_x,
            frac_y,
        );

        // Scale the drawing context so that when we draw
        // our glyph it's stretched to the constrained size.
        context.scaleCTM(
            ctx,
            width / rect.size.width,
            height / rect.size.height,
        );
```

For constraints to be correctly transferred, `width` and `height` here must still refer to the actual bbox of the outlines post-constraint. They cannot include any of the padding we add to the canvas to make room for thickening and synthetic bold, as that would make the scale factors incorrect. Meanwhile, `frac_x` and `frac_y` must point to the origin of the same outlines on the rendering canvas, which means they must include both the whole-pixel padding and the fractional offset.

In short, I don't see a better way to do this while keeping it correct, do you?

danielwe:
> the ones that are simply constrained to cells

Which symbols are this? Normally, if a font specifies outlines that go outside the cells, they are just drawn outside the cells, right? There's probably some case I'm forgetting.

src/font/face/coretext.zig
qwerasd205: It feels cleaner to me to just add the padding value inside of the translate call (i.e. `frac_<x/y> + canvas_padding`) and add twice the padding to the px width and px height. It confuses things to fold the padding in to `frac_x` and `frac_y` variables because the purpose of those variables is to hold the *fractional* pixel position. Entangling the padding with the sub-pixel positioning is a needlessly confusing addition to what's already a fairly unintuitive section of code.

qwerasd205:
https://github.com/ghostty-org/ghostty/blob/a5aff0e347b0016e2735d4ec4b4cdca96b5438d1/src/renderer/generic.zig#L3091-L3097

src/font/face/coretext.zig
danielwe: Fair! I'll include that change in the next push.

danielwe: No, wait a second, the problem is that `canvas_padding` isn't necessarily an integer either, so we'd have to split that too into a rounded and fractional part. That seems more complicated. Maybe we could rename the variables to make the current flow more intuitive?

danielwe:
Thanks, I see.

~~As an aside, I now realize this won't actually ensure that the symbol is contained within a single cell, only that its outlines aren't wider than one cell. It could still be shifted. For example, Iosevka has a number of (non-NF) symbols that are centered within a 2x advance width (I discovered this yesterday when looking into #8929, though the codepoint discussed there, 25A0, is currently not considered a symbol by `isSymbol`).~~ **Never mind, this is both incorrect and irrelevant.**

~~Also, downscaling without alignment currently just scales down the bearings without any vertical correction, i.e., it pulls the symbol toward the lower left corner. It would probably be better to use the baseline or midline as anchor for vertical scaling, or maybe better, default to center-center alignment when the glyph is rescaled and no alignment is specified. This would also take care of the issue in the previous paragraph. I can submit a separate PR for that, or add to this.~~ #8990

src/font/face/coretext.zig
danielwe: Though if we move synthetic bold pre-constraint as discussed below, the padding is always integral, so this problem goes away.

danielwe:
In summary, let me know which changes you want me to push:

* Bail based on pre-constraint rect instead of post-constraint padded canvas?
* Measure and constrain post-synthetic bold instead of pre?
* ~~Re scaling without specified alignment: no change; vertically anchor rescaling at baseline/midline instead of bottom; or default to centered alignment, both vertically and horizontally? Do this in a separate PR or add to this?~~ #8990

danielwe:
Alright, I pushed a new version that makes a simpler and more minimal set of changes along the lines you indicated.
* In FreeType, we measure the possibly emboldened glyph, which is consistent with the rect padding done in CoreText. Thus, both backends apply constraints to the emboldened bbox.
* We keep the main change: only padding the canvas, not the rect, for font smoothing. Since this padding is always integer (no fractional contribution from emboldening), we keep it separate from `frac_{x,y}`.

qwerasd205:
This looks good to me now, I think, but I don't have the energy to do a proper review tonight. Hopefully in the morning I can.

danielwe:
Sounds great!

I'll note that #8990 and #8847 address higher-severity bugs/regressions, so maybe they should take priority if the time window for 1.2.1 is getting tight. That said, this one is probably the quickest and easiest to review.

qwerasd205 (approved):

PR #8585, opened 2025-09-11
config(default): Increase scrollback-limit to 100MB

jumploops:
## tl;dr 
  
- Single LLM response logs don't fit in the default scrollback limit
- Change default from 10MB -> 100MB
 
   
   
_Note: Following is AI generated (GPT-5)_

### Summary

This PR increases the default `scrollback-limit` from **10MB** to **100MB**.
The scrollback buffer determines how much terminal history is kept in memory before older lines are discarded.

### Motivation

* The previous default (16MB) often proved insufficient for modern workflows involving large build logs, long-running processes, and verbose server output.
* Increasing the buffer to 100MB provides a much more usable history window without imposing a significant memory burden on modern systems, where multi-gigabyte RAM is common.
* Because scrollback memory is allocated lazily, the higher default will not increase baseline memory consumption — users will only pay the cost when generating large output.

### Details

* **New default**: `scrollback-limit = 100_000_000`
* This change applies per terminal surface. The visible screen remains guaranteed to be allocated, and additional memory is only used as the scrollback grows.
* Users who need smaller or larger values can still override the setting in their configuration.

### Impact

* Improved usability for developers who need access to more command history.
* Minimal risk of excessive memory usage, given modern hardware capacity and the lazy allocation model.
* No changes to configuration format or compatibility.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1105,7 +1105,7 @@ input: RepeatableReadableIO = .{},
 /// This is a future planned feature.
 ///
 /// This can be changed at runtime but will only affect new terminal surfaces.
-@"scrollback-limit": usize = 10_000_000, // 10MB
+@"scrollback-limit": usize = 100_000_000, // 100MB
 
 /// Match a regular expression against the terminal text and associate clicking
 /// it with an action. This can be used to match URLs, file paths, etc. Actions


mitchellh:
Thank you. You're the first person to bring it up to the project that this may be too small a default. I'm not ready to change the default based not he anecdotal experience of one person of a thing that is already configurable. I use LLMs heavily and haven't ran into this issue yet, personally.

Please feel free to open a discussion and we can wait to gather feedback from the broader community.

jumploops:
Thanks for the response @mitchellh -- discussion opened here: https://github.com/ghostty-org/ghostty/discussions/8593

PR #8590, opened 2025-09-11
xterm color operations compatibility and rewrite

### LINKED ISSUE #7951: OSC color reporting still not right

mitchellh:
Replaces #7952
Fixes #7951 

This reimplements our color operation parsing completely to make it fully compatible (as far as I can tell) with xterm. Our previous implementation had numerous problems, I think because we kept addressing singular compatibility issues as they were experienced in the field rather than doing a proper thoughtful audit compared to the xterm implementation. This PR does that audit.

**Specifically, this updates/adds: OSC 4, 5, 10-19, 104, 105, 110-119.**

To ease maintenance, understanding, and testing, I've pulled color operation parsing out into a separate file and function that operates on the full buffered OSC command. This is similar to Kitty protocols previously. This hurts performance but that's acceptable to me for now while we get compatibility down and test coverage added. 

We can address more performance later if it becomes a bottleneck, but these color operations are pretty rare.

I've associated each test with a `printf` command you can run in xterm to compare.

## Xterm Divergence

We purposely diverge from xterm in some scenarios:

- Whitespace is allowed around x11 color names. Kitty allows this.
- Invalid index values for 104/105 are ignored. xterm typically halts processing. Kitty allows this.

## TODO

- [x] Update our parser to use the new color parsing functions
- [x] Update the stream handler to use the new types
- [x] Fix our stream handler to emit on response per query


mitchellh:
What a righteous mess this all is. 😄 I tried to capture most of the complexity in `osc/color.zig`. But we really need to probably abstract out the reporting out into a testable area too. That's too messy right now.

Our `osc/color.zig` I at least feel very good about, and I brought along support for every color operation documented by xterm. We don't support all that in the stream handlers since we don't support every color but at least it's all parsed properly. 

cc @jcollie

j-c-m:
The palette is still not resetting with printf "\e]104\e\\" (no semi-colon)

xterm, iterm2, alacritty, kitty all reset the palette without the semicolon.

mitchellh:
> The palette is still not resetting with printf "\e]104\e\" (no semi-colon)

Hm, this is unit tested as passing. It is parsing correctly. It has to be somewhere else. I'll take a look.

mitchellh:
Oh found it. Easy fix one sec.

PR #8602, opened 2025-09-12
feat: support macos fullscreen with window-decorations=none

danielfalbo:
**before**
on macos with `window-decoration = none` clicking `cmd + shift - f` (macos shortcut to go fullscreen) did nothing

https://github.com/user-attachments/assets/5d4c750c-322d-47ed-834c-2f1152e7573b

**after**
now it correctly goes to full screen, and back to normal (without window-decorations) when pressed from fullscreen

https://github.com/user-attachments/assets/4f8c2364-d5ef-4cff-9e8c-4d9633f6d801

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -62,6 +62,9 @@ class BaseTerminalController: NSWindowController,
 
     /// Fullscreen state management.
     private(set) var fullscreenStyle: FullscreenStyle?
+    
+    /// Tracks whether window decorations were enabled before entering fullscreen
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Thanks for your contribution. 

It appears there may be two conflated changes mixed in here: toggle window decorations AND the fullscreen fix. I suggest we split those out. Merging two changes is always more complicated than merging one.

On the fullscreen side, I think the fix should probably go into the fullscreen style implementations rather than a controller.

The window decoration toggle looks potentially fine on its own, just need to do a bit more testing on my part.

danielfalbo:
Thank you! Closing this, will open two new separate PRs

PR #8609, opened 2025-09-12
deps/ci: Use Ghostty-specific themes release

AnthonyZhOon:
Closes #8608 
iterm2_themes now provides terminal-specific releases, this reduces the size of the dependency from 55MB->53kB.
- Update the `.github/workflows/update-colorschemes.yml` to track the latest release and fetch the ghostty-specific theme archive.

### DIFF
--- a/.github/workflows/update-colorschemes.yml
+++ b/.github/workflows/update-colorschemes.yml
@@ -40,9 +40,11 @@ jobs:
       - name: Run zig fetch
         id: zig_fetch
         run: |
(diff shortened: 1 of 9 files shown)


mitchellh:
Hot diggity!

jcollie:
One thing that we may want to change is that upstream generates a release every Monday at 15:00 UTC. We currently run our process on Saturdays. Maybe every Monday at 17:00 UTC?

PR #8620, opened 2025-09-14
feat(macos): implement GHOSTTY_ACTION_TOGGLE_WINDOW_DECORATIONS

danielfalbo:
note: this will be useful to support going to full screen on macOS when `window-decorations = none`

see https://github.com/ghostty-org/ghostty/pull/8602

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -627,7 +627,7 @@ extension Ghostty {
             case GHOSTTY_ACTION_TOGGLE_TAB_OVERVIEW:
                 fallthrough
             case GHOSTTY_ACTION_TOGGLE_WINDOW_DECORATIONS:
-                fallthrough
+                toggleWindowDecorations(app, target: target)
             case GHOSTTY_ACTION_PRESENT_TERMINAL:
                 fallthrough
             case GHOSTTY_ACTION_SIZE_LIMIT:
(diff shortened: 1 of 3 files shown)


danielfalbo:
ping

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1249,6 +1249,26 @@ extension Ghostty {
...
+        ) {
+            switch (target.tag) {
+            case GHOSTTY_TARGET_APP:
+                Ghostty.logger.warning("toggle window decorations does nothing with an app target")
+                return
+
+            case GHOSTTY_TARGET_SURFACE:
+                guard let surface = target.target.surface else { return }
+                guard let surfaceView = self.surfaceView(from: surface) else { return }
+                guard let terminalController = surfaceView.window?.windowController as? BaseTerminalController else { return }
```
mitchellh: I'd prefer to use a notification.

mitchellh (changes requested):
Sorry for the late review, getting around to these now. I think I'd prefer a notification-based approach like we use for some other logic so that future terminal types could choose to implement this if they wanted without hardcoding `BaseTerminalController`.

danielfalbo:
Thank you @mitchellh , I was not familiar with the `NotificationCenter` pattern of Swift apps and Ghostty in particular. Does it look good now?

I'm also drafting the e2e fullscreen feature at https://github.com/ghostty-org/ghostty/pull/9182 so, while we will review that later in another PR, we can test these changing are working already.

mitchellh (changes requested):
Thanks for this and I do apologize for the late review. Unfortunately, this PR is incomplete and doesn't handle the complexity of macOS titlebar states properly. It isn't as simple as showing/hiding `titled`, since there are a number of other configurations at play here:

- Fullscreen state
- `macos-window-buttons` config
- I think this impacts the quick terminal but didn't verify, and it probably shouldn't
- If you remove `titled` from a window in a tab group, it drops out of the tab group
- I believe this corrupts our titlebar accessories but also didn't verify (but I think removing a title removes the accessories)

I'm going to close this since this requires some significant change. Note some of the above may not be fully accurate but I believe it warrants investigation and should be noted in the PR. Thank you.

danielfalbo:
Got it, thank you!

PR #8632, opened 2025-09-15
pin iTerm2 themes dependency

mitchellh:
There is an upstream bug where subsequent releases will nuke prior released packages, resulting in 404s. Not good!


kuntau:
Did this PR remove many themes? My day theme `dawnfox` missing after v1.2 update

mitchellh:
No, they’re all there. Read the release notes

00-kat:
A bit late, but TL;DR is they were renamed, check `ghostty +list-themes` (is it maybe `DayFox`?).

PR #8636, opened 2025-09-15
macos: when toggling visibility, hide individual windows not app

### LINKED ISSUE #8414: macOS: toggle_quick_terminal makes hidden window visible again
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6958

<sup>Originally posted by **candhub** March 29, 2025</sup>
Using the following keybinds:

keybind = global:f10=toggle_visibility
keybind = global:f9=toggle_quick_terminal

- Pressing F10 hides the Ghostty window (toggle_visibility)
- Pressing F9 to open the quick terminal causes the previously hidden Ghostty window to become visible again
- Pressing F9 again to close the quick terminal leaves the Ghostty window still visible

This issue only occurs when the window is hidden using toggle_visibility.
If the Ghostty window is minimized using macOS native window management (e.g., Command+M), this behavior does not occur.

environment
- macOS
- Ghostty 1.1.3</div>

mjdouglas:
Fixes #8414

Previously, toggling visibility on Mac was implemented via NSApp.hide(). This caused a bug where bringing up the quick terminal would unintentionally restore hidden windows. It appears making any window key and front will first unhide the app.

This diff fixes the issue by first manually removing the visible windows from the window list. Windows that are ordered out aren't restored automatically when a different window (the quick terminal) is made key and front.

Disclosure: I consulted GPT-5 Thinking on the trade offs between hiding the app, ordering out windows, and minaturizing windows.

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -107,7 +107,7 @@ class AppDelegate: NSObject,
     }
 
     /// Tracks the windows that we hid for toggleVisibility.
-    private var hiddenState: ToggleVisibilityState? = nil
+    private var windowStash: HiddenWindowStash = HiddenWindowStash()
 
     /// The observer for the app appearance.
     private var appearanceObserver: NSKeyValueObservation? = nil
@@ -294,8 +294,8 @@ class AppDelegate: NSObject,
     }
 
     func applicationDidBecomeActive(_ notification: Notification) {
-        // If we're back manually then clear the hidden state because macOS handles it.
-        self.hiddenState = nil
+        // Restore hidden windows, because MacOS won't do it for us
+        windowStash.restoreAll()
 
         // Clear the dock badge when the app becomes active
         self.setDockBadge(nil)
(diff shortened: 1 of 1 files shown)


macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -1039,27 +1039,22 @@ class AppDelegate: NSObject,
...
             // fullscreen. This is only relevant if Ghostty is active.
             guard let keyWindow = NSApp.keyWindow,
                   !keyWindow.styleMask.contains(.fullScreen) else { return }
 
-            // Keep track of our hidden state to restore properly
-            self.hiddenState = .init()
+            windowStash.hideAll()
+
+            // Deactivate the app, as we no longer have any visible windows
             NSApp.hide(nil)
```
mitchellh, line 1056: This still seems to use NSApp.hide?

macos/Sources/App/macOS/AppDelegate.swift
```diff
@@ -1096,14 +1091,18 @@ class AppDelegate: NSObject,
...
-        init() {
+        init() {}
+
+        func hasHiddenWindows() -> Bool { !hiddenWindows.isEmpty }
+
+        mutating func hideAll() {
             // We need to know the key window so that we can bring focus back to the
             // right window if it was hidden.
-            self.keyWindow = if let keyWindow = NSApp.keyWindow {
+            self.keyWindow = if let keyWindow = NSApp.keyWindow, !keyWindow.styleMask.contains(.nonactivatingPanel) {
```
mitchellh, line 1105: What's the reasoning behind ignoring `nonactivatingPanel`. What does Ghostty have that triggers this?

mitchellh (changes requested):
I haven't built or tested this yet, my review just based on the code.

macos/Sources/App/macOS/AppDelegate.swift
mjdouglas: The quick terminal is a non-activating panel.

macos/Sources/App/macOS/AppDelegate.swift
mjdouglas: I can see how that's confusing. Windows are hidden with `window.orderOut()` in `hideAll()`, but we still need `NSApp.hide(nil)` to deactivate the app and bring whatever app was running behind Ghostty in the stack to the foreground.

mitchellh:
The original issue was fixed another way. If there is something we missed please let me know!

PR #8638, opened 2025-09-15
Pass config to splits in NewTerminalConfig

calebhearth:
Config contains the command, working directory, and environment
variables intended to be passed to the new split, but it looks like we
forgot to include it as an argument in this branch.

Discussion: https://github.com/ghostty-org/ghostty/discussions/8637


mitchellh:
This looks right to me. Thanks!

PR #8701, opened 2025-09-17
Default config template be explicit that you do not copy the default values

rhodes-b:
This has been brought up against at least 2 times since the 1.2 release.

Ideally this should also have a big warning on the website somewhere.

### DIFF
--- a/src/config/config-template
+++ b/src/config/config-template
@@ -6,14 +6,14 @@
 #   {[path]s}
 #
 # The template does not set any default options, since Ghostty ships
-# with sensible defaults for all options. Users should only need to set
-# options that they want to change from the default.
+# with sensible defaults for all options.
+# Note that you should not paste the output of `ghostty +show-config
+# --default` into your config: some default options actually conflict with each other
+# when explicitly set in a configuration file. Instead, only set the
+# options you actually need.
 #
 # Run `ghostty +show-config --default --docs` to view a list of
 # all available config options and their default values.
-#
-# Do not copy the output of the defaults into your config, like stated above
-# users only need to set options to change the defaults.
 #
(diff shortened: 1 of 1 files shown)


src/config/config-template
```diff
@@ -12,6 +12,9 @@
 # Run `ghostty +show-config --default --docs` to view a list of
 # all available config options and their default values.
 #
+# Do not copy the output of the defaults into your config, like stated above
+# users only need to set options to change the defaults.
```
00-kat:
```suggestion
# Do not copy the output of the defaults into your config. As stated above,
# users only need to set options to change their values from the default.
```

rhodes-b: i didn't want to go with change their values wording because that doesn't apply to new keybinds for example

since there may not be a default for everything (toggle_tab_overview) You could say the default for that is none but it still doesn't make sense for random keybinds then like `ctrl+a=text:hi`

pluiedev: This feels a bit too forceful. I would instead edit the above paragraph to be
```diff
# The template does not set any default options, since Ghostty ships
-# with sensible defaults for all options. Users should only need to set
-# options that they want to change from the default.
+# with sensible defaults for all options.
+# 
+# Note that you should _not_ paste the output of `ghostty +show-config
+# --default` here; some default options actually conflict with each other
+# when explicitly set in a configuration file. Instead, **only set the
+# options you actually need**.
```

pluiedev (changes requested):

src/config/config-template
00-kat: (Inconsequential nit: that semicolon should be a colon.)

00-kat (approved):

pluiedev (approved):

Issue #8674, opened 2025-09-16
font: FreeType backend should respect fontconfig's aa/hinting options


qwerasd205:
Right now we have our own opinionated load flag defaults for FreeType which may conflict with what the user expects due to the implicit defaults under fontconfig or their explicitly adjusted font settings on their system.

We need to respect these flags unless explicitly overridden by the user's Ghostty config (`freetype-load-flags`):
- `antialias`
- `hintstyle`
- `hinting`
- `autohint`

We should also make sure we're querying these per-font not just the global settings, since it is possible with fontconfig to adjust these per-font.

### ref:
- https://fontconfig.pages.freedesktop.org/fontconfig/fontconfig-user.html
- https://fontconfig.pages.freedesktop.org/fontconfig/fontconfig-devel/

---

P.S. Do we even need `freetype-load-flags` if we respect the system configuration? Something to think about, since it feels like that config was really just added as a workaround for this issue.

mitchellh:
> P.S. Do we even need freetype-load-flags if we respect the system configuration? Something to think about, since it feels like that config was really just added as a workaround for this issue.

Probably not. But I don't see a downside in letting people toy around with these flags. It can help debug things in situations like this too.

mitchellh:
Depending on the magnitude of the change, I'm willing to accept this as a backport into 1.2.x.

WhyNotHugo:
As mentioned in https://github.com/ghostty-org/ghostty/discussions/8125, ghostty also completely ignores the font _families_ themselves too.

My local configuration sets Noto Sans Mono CJK SC as my default monospaced font and Twemoji as my default emoji font, but ghostty ignored both of these Fontconfig settings.

qwerasd205:
That is a separate question, ultimately. I think it probably *is* a good idea to use those defaults if they're configured, but that's a distinct change from these which apply to *how* we load fonts, not *which* fonts we load.

WhyNotHugo:
Shall I open a separate issue to track that instead? Or can we include that within scope for this issue?

I'm not really sure it makes any sense to respect the local antialias/hintstyle/hinting/autohint configuration but ignore the configured font-family itself.

danielwe:
I dug into Fontconfig and FreeType config in response to a Discord question, and found that
* Many (most?) distros set the default Fontconfig hintstyle to `hintslight` (to check, see if `/etc/fonts/conf.d/10-hinting-slight.conf` exists)
[...]

PR #8720, opened 2025-09-17
font: Measure ascii height and use to upper bound ic_width

danielwe:
Measure the ~actual height between the ascender and descender lines as the height of the~ overall bounding box of the font's ASCII characters[^1], and use this to upper bound the IC width estimate. This ensures that the scaling of fallback CJK fonts to a wide-aspect primary font doesn't result in oversized CJK glyphs.

Fixes #8709.

I'd appreciate feedback from Chinese speakers on the suitability of this metric across a variety of primary fonts.

Screenshots using an empty Ghostty config on macOS and the test file from #8651. The font loaded as fallback for CJK is PingFang SC.

**1.2.0**
Screenshot 2025-09-17 at 09 51 43

**This PR**
Screenshot 2025-09-17 at 09 51 25

[^1]: Note that this may be different from the difference between the nominal ascent - descent, as non-letter ASCII characters often exceed the ascender and descnder lines, and fonts often bake the line gap into the ascent/descent and set `line_gap` to zero, as per official recommendations like those from Google Fonts: https://simoncozens.github.io/gf-docs/metrics.html

### DIFF
--- a/src/font/Metrics.zig
+++ b/src/font/Metrics.zig
@@ -117,6 +117,16 @@ pub const FaceMetrics = struct {
     /// lowercase x glyph.
     ex_height: ?f64 = null,
 
+    /// The measured height of the bounding box containing all printable
+    /// ASCII characters. This can be different from ascent - descent for
+    /// two reasons: non-letter symbols like @ and $ often exceed the
+    /// the ascender and descender lines; and fonts often bake the line
+    /// gap into the ascent and descent metrics (as per, e.g., the Google
+    /// Fonts guidelines: https://simoncozens.github.io/gf-docs/metrics.html).
+    ///
+    /// Positive value in px
+    ascii_height: ?f64 = null,
+
     /// The width of the character "水" (CJK water ideograph, U+6C34),
     /// if present. This is used for font size adjustment, to normalize
     /// the width of CJK fonts mixed with latin fonts.
@@ -144,11 +154,20 @@ pub const FaceMetrics = struct {
         return 0.75 * self.capHeight();
     }
 
+    /// Convenience function for getting the ASCII height. If we
+    /// couldn't measure this, we use 1.5 * cap_height as our
+    /// estimator, based on measurements across programming fonts.
+    pub inline fn asciiHeight(self: FaceMetrics) f64 {
+        if (self.ascii_height) |value| if (value > 0) return value;
+        return 1.5 * self.capHeight();
+    }
+
     /// Convenience function for getting the ideograph width. If this is
-    /// not defined in the font, we estimate it as two cell widths.
+    /// not defined in the font, we estimate it as the minimum of the
+    /// ascii height and two cell widths.
     pub inline fn icWidth(self: FaceMetrics) f64 {
         if (self.ic_width) |value| if (value > 0) return value;
-        return 2 * self.cell_width;
+        return @min(self.asciiHeight(), 2 * self.cell_width);
     }
 
     /// Convenience function for getting the underline thickness. If
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -775,7 +775,10 @@ pub const Face = struct {
         // Cell width is calculated by calculating the widest width of the
         // visible ASCII characters. Usually 'M' is widest but we just take
         // whatever is widest.
-        const cell_width: f64 = cell_width: {
+        //
+        // ASCII height is calculated as the height of the overall bounding
+        // box of the same characters.
+        const cell_width: f64, const ascii_height: f64 = measurements: {
             // Build a comptime array of all the ASCII chars
             const unichars = comptime unichars: {
                 const len = 127 - 32;
@@ -803,7 +806,10 @@ pub const Face = struct {
                 max = @max(advances[i].width, max);
             }
 
-            break :cell_width max;
+            // Get the overall bounding rect for the glyphs
+            const rect = ct_font.getBoundingRectsForGlyphs(.horizontal, &glyphs, null);
+
+            break :measurements .{ max, rect.size.height };
         };
 
         // Measure "水" (CJK water ideograph, U+6C34) for our ic width.
@@ -864,6 +870,7 @@ pub const Face = struct {
 
             .cap_height = cap_height,
             .ex_height = ex_height,
+            .ascii_height = ascii_height,
             .ic_width = ic_width,
         };
     }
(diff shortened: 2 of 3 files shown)


danielwe:
~Hang on while I fire up a linux VM to fix the freetype test failure!~ Fixed

src/font/Metrics.zig
```diff
@@ -144,11 +155,19 @@ pub const FaceMetrics = struct {
         return 0.75 * self.capHeight();
     }
 
+    /// Convenience function for getting the j-height. If we couldn't
+    /// measure this, we estimate it as ascent - descent.
+    pub inline fn jayHeight(self: FaceMetrics) f64 {
+        if (self.jay_height) |value| if (value > 0) return value;
+        return self.ascent - self.descent;
```
qwerasd205: Might wanna change this estimator to something more evidence based. `ascent - descent` is very likely to be too large. I have some code that measures metrics from an actual corpus of fonts, which I verified the other multiplier-based estimations with to be more or less correct for a broad range of programming fonts. I'll go ahead and figure out the best multiplier to get the maximum vertical extent of ASCII from known quantities.

qwerasd205 (commented):
Right now this code is measuring the vertical extent of ASCII, which makes calling it "j-height" fairly inaccurate since in most fonts, characters like `@`, `$`, and `|`, as well as brackets, will extend above the latin ascender height and below the descender height. I feel it would be a lot clearer to rename it to "ascii height" instead since that reflects what is actually measured and is more obvious in its meaning to someone reading the code (since "j-height" is not a well-established metric).

danielwe:
> Right now this code is measuring the vertical extent of ASCII, which makes calling it "j-height" fairly inaccurate since in most fonts, characters like `@`, `$`, and `|`, as well as brackets, will extend above the latin ascender height and below the descender height. I feel it would be a lot clearer to rename it to "ascii height" instead

Good point, but my gut feeling is the better thing to do is to skip the non-letter ascii characters such that this actually is a measure of ascender to descender.

> (since "j-height" is not a well-established metric)

Yeah, definitely open to a better name. Maybe `latin_height` if we only consider the letters? I chose j-height because, even though not firmly established, it's the only name I've seen that unambiguously refers to this quantity (https://stackoverflow.com/a/30184114).

src/font/Metrics.zig
qwerasd205: So it seems like the best estimator for vertical extent of ASCII in programming fonts is `1.5 * cap_height`, it has the strongest clustering and fewest outliers compared to using something like line height as the basis.

If we instead want to change this to latin height instead of ASCII height then the best estimator for that is closer to `1.4 * cap_height`.

danielwe: Sounds good, I'll include the appropriate estimator and push an update after playing around a bit more to see whether ascii or latin height looks best (to my untrained eyes).

danielwe:
Most recent update:
* Renamed `jay_height` to `ascii_height`
  - To me this looks much better than the `latin_height` I proposed above. The latter makes CJK too small.
* Using `1.5 * cap_height` as estimator when ASCII measurement failed
  - However, if our cap height is only an estimate, we use `ascent - descent` as the ultimate fallback, rather than stacking estimators.
    * An alternative would be to use `lineHeight()` as the ultimate fallback. Let me know what you think.

qwerasd205:
Stacking estimators shouldn't be a problem in this case, since that would still yield a better estimate than line height (or line height without gap)

danielwe:
If you say so! My assumption was that the `cap_height = 0.75 * ascent` estimate is rather crude, but if you think it's still the best basis for estimating ASCII height, I'll go ahead and use that instead.

src/font/Metrics.zig
```diff
@@ -144,11 +154,22 @@ pub const FaceMetrics = struct {
         return 0.75 * self.capHeight();
     }
 
+    /// Convenience function for getting the ASCII height. If we
+    /// couldn't measure this, we use 1.5 * cap_height as our
+    /// estimator, based on measurements across programming fonts.
+    /// If our cap height is itself an estimate, we fall back on
+    /// ascent - descent.
```
qwerasd205: You forgot to update the comment here

danielwe:
Adding some screenshots with other primary fonts. The CJK fallback is the same in all cases: PingFang SC.

### Iosevka (narrow)

**1.2.0**
Screenshot 2025-09-17 at 12 39 36

**This PR**
Screenshot 2025-09-17 at 12 34 12

### Commit Mono (wide)

**1.2.0**
Screenshot 2025-09-17 at 12 40 10

**This PR**
Screenshot 2025-09-17 at 12 34 50

src/font/Metrics.zig
danielwe: Good catch!

qwerasd205 (approved):

danielwe:
Maybe good to hear some feedback from a handful of native speakers before merging? @pluiedev? Should we ask people in #8651 to take a look at these screenshots? (It's locked, so I can't post.)

(IDK whether you read Chinese or not, @qwerasd205, but in any case it's probably good to have N > 1.)

pluiedev:
I think this looks pretty dang good

mitchellh (approved):
I'd prefer to have some fonts with unit tests in here so we don't regress this when we make changes here, but willing to accept it on faith for now based on the other reviews. Looks good.

danielwe:
I was working on unit tests and realized that FreeType and CoreText provide quite different glyph measurements for the same font due to FreeType's hinting. This affects `cell_width`, `ascii_height`, `ic_width` if measurable, and the fallback measurements for `cap_height` and `ex_height`.

If I include `.no_hinting = true` in the FreeType load flags, the values come out the same. (Only adding the load flag when glyphs are loaded for measurement, of course; hinting is still enabled when rendering.)

Just want to confirm that we do want `cell_width` and `ascii_height` to be measured on hinted glyphs. If yes, I just need to customize the tests per platform. If no, I'll add the `.no_hinting` load flag as part of the unit test PR. @qwerasd205?

Edit: actually, if yes, we should probably add the same load flags as during rendering, so glyphs are hinted the same way for measurement and rendering.

qwerasd205:
Glyphs should probably be measured with the same hinting as is used when they're rendered, which right now depends on user config, and in the future will depend on fontconfig (#8674). This will be a bit annoying for the unit tests, but it's the most correct answer.

PR #8738, opened 2025-09-18
fix(font): Improve FreeType glyph measurements and add unit tests for face metrics

danielwe:
Follow-up to #8720 adding

* Two improvements to FreeType glyph measurements:
  - Ensuring that glyphs are measured with the same hinting as they are rendered, ref [#8720#issuecomment-3305408157](https://github.com/ghostty-org/ghostty/pull/8720#issuecomment-3305408157);
  - For outline glyphs, using the outline bbox instead of the built-in metrics, like `renderGlyph()`.
* Basic unit tests for face metrics and their estimators, using the narrowest and widest fonts from the resource directory, Cozette Vector and Geist Mono.

---

I also made one unrelated change to `freetype.zig`,  replacing `@alignCast(@ptrCast(...))` with `@ptrCast(@alignCast(...))` on line 173. Autoformatting has been making this change on every save for weeks, and reverting the hunk before each commit is getting old, so I hope it's OK that I use this PR to upstream this decree from the formatter.

### DIFF
--- a/src/font/Collection.zig
+++ b/src/font/Collection.zig
@@ -1489,11 +1489,10 @@ test "face metrics" {
             const actual = @field(metricsActual, field.name);
             // Unwrap optional fields
             const expectedValue, const actualValue = unwrap: switch (@typeInfo(field.type)) {
-                .optional => |Tinfo| {
-                    if (expected) |expectedValue| {
-                        const actualValue = actual orelse std.math.nan(Tinfo.child);
+                .optional => {
+                    if (expected) |expectedValue| if (actual) |actualValue| {
                         break :unwrap .{ expectedValue, actualValue };
-                    }
+                    };
                     // Null values can be compared directly
                     try std.testing.expectEqual(expected, actual);
                     continue;
--- a/src/font/Collection.zig
+++ b/src/font/Collection.zig
@@ -1369,3 +1369,125 @@ test "adjusted sizes" {
         );
     }
 }
+
+test "face metrics" {
+    const testing = std.testing;
+    const alloc = testing.allocator;
+    const narrowFont = font.embedded.cozette;
+    const wideFont = font.embedded.geist_mono;
+
+    var lib = try Library.init(alloc);
+    defer lib.deinit();
+
+    var c = init();
+    defer c.deinit(alloc);
+    const size: DesiredSize = .{ .points = 12, .xdpi = 96, .ydpi = 96 };
+    c.load_options = .{ .library = lib, .size = size };
+
+    const narrowIndex = try c.add(alloc, try .init(
+        lib,
+        narrowFont,
+        .{ .size = size },
+    ), .{
+        .style = .regular,
+        .fallback = false,
+        .size_adjustment = .none,
+    });
+    const wideIndex = try c.add(alloc, try .init(
+        lib,
+        wideFont,
+        .{ .size = size },
+    ), .{
+        .style = .regular,
+        .fallback = false,
+        .size_adjustment = .none,
+    });
+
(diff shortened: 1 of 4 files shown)


danielwe:
Pushed an update with a better design for sharing FreeType load flags between measurements and rendering. The previous CI failures seemed to be a matter of CI resource constraints, not an issue with the code.

src/font/Collection.zig
```diff
@@ -1369,3 +1369,125 @@ test "adjusted sizes" {
...
+                .coretext,
+                .coretext_harfbuzz,
+                .coretext_noshape,
+                => 7.3828125,
+                .web_canvas => unreachable,
+            },
+            .ascent = 12.3046875,
+            .descent = -3.6953125,
+            .line_gap = 0.0,
+            .underline_position = -1.2265625,
```
mitchellh: comparing floats like this is going to be super sketchy. We should use `std.math.approxEqRel`. It's going to be tedious to pull them all out but that will be significantly more robust.

mitchellh (changes requested):

src/font/Collection.zig
danielwe: I know, never compare floats for equality. I was actually surprised when the values came out exactly equal in CoreText and FreeType (modulo the ones affected by hinting), but given that they did I figured the calculations producing these values don't have a lot of potential for reassociations that would change the exact result (basically, they are all some variant of `value_in_integer_design_units * px_per_unit`, so to reassociate you'd probably have to split up the `px_per_unit` ratio), so I went with the quick and dirty.

Will fix.

danielwe: Pushed an update that addresses this.

danielwe (commented):
Should probably also factor this bboxery into a separate function rather than repeating the same conditional over and over. [Should have been a review comment on specific lines, see below.]

src/font/face/freetype.zig
```diff
@@ -976,23 +982,27 @@ pub const Face = struct {
...
+                        const ymin, const ymax = metrics: {
+                            if (glyph.*.format == freetype.c.FT_GLYPH_FORMAT_OUTLINE) {
+                                var bbox: freetype.c.FT_BBox = undefined;
+                                _ = freetype.c.FT_Outline_Get_BBox(&glyph.*.outline, &bbox);
+                                break :metrics .{ bbox.yMin, bbox.yMax };
+                            }
+                            break :metrics .{
+                                glyph.*.metrics.horiBearingY - glyph.*.metrics.height,
+                                glyph.*.metrics.horiBearingY,
+                            };
```
danielwe: Should probably also factor this bboxery into a separate function rather than repeating the same conditional over and over.

src/font/face/freetype.zig
```diff
@@ -392,42 +382,56 @@ pub const Face = struct {
             // SVG glyphs under FreeType, since that requires bundling another
             // dependency to handle rendering the SVG.
             .no_svg = true,
-        });
+        };
+    }
+
+    /// Get a rect that represents the position and size of the loaded glyph.
+    fn getGlyphSize(glyph: freetype.c.FT_GlyphSlot) font.face.RenderOptions.Constraint.GlyphSize {
```
danielwe: This looks a bit ridiculous. The definition of `GlyphSize` should probably be lifted out of these nested scopes if we want to use it here.

danielwe:
Anything else I should do to get this ready? Merging the tests isn't that urgent/consequential of course, but the bounding box measurements fix is necessary to make #8720 work as intended (the metrics-based bounding box can be quite a bit too large).

mitchellh (approved):

PR #8752, opened 2025-09-18
documentation: fix MacOSDockDropBehavior valid values

mavonarx:
The documentation shows that the enum values should be "new-window" and "new-tab", 
However, "new-window" currently fails, but "window" is still accepted.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2710,7 +2710,7 @@ keybind: Keybinds = .{},
 ///
 ///   * `new-tab` - Create a new tab in the current window, or open
 ///     a new window if none exist.
-///   * `new-window` - Create a new window unconditionally.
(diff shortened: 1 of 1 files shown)


src/config/Config.zig
```diff
@@ -7282,7 +7282,7 @@ pub const WindowNewTabPosition = enum {
 /// See macos-dock-drop-behavior
 pub const MacOSDockDropBehavior = enum {
     @"new-tab",
-    window,
+    @"new-window",
```
mitchellh: This is technically a breaking change, since we already shipped this. I recommend for now just changing the docs instead.

mitchellh (changes requested):

mitchellh (approved):

PR #8759, opened 2025-09-18
fix default increase_font_size binding config from physical to unicode

LiamGoodacre:
During d015efc87d59f873e2ae7a5df8fa432cd5db2bdc some of the bindings were converted from physical keys to unicode. The binding for font size increase via `=` was missed. Therefore there was an inconsistency between font size increase and font size decrease. Decreasing with `-` would be relative to the keyboard layout, whereas increase required the physical `=` key.

Dedicated discussion: https://github.com/ghostty-org/ghostty/discussions/8743
Original thread on: https://github.com/ghostty-org/ghostty/discussions/8729#discussioncomment-14439306

I've ran `zig build run` to sanity check the behaviour and all seems well.
Let me know if there's anything else I need to do.

---

A work-around in the mean time is:
```
keybind = ctrl+equal=unbind
keybind = ctrl+==increase_font_size:1
```

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -5620,7 +5620,7 @@ pub const Keybinds = struct {
         // set the expected keybind for the menu.
         try self.set.put(
             alloc,
-            .{ .key = .{ .physical = .equal }, .mods = inputpkg.ctrlOrSuper(.{}) },
+            .{ .key = .{ .unicode = '=' }, .mods = inputpkg.ctrlOrSuper(.{}) },
             .{ .increase_font_size = 1 },
         );
         try self.set.put(


pluiedev (approved):
Makes sense to me. Let's see what others think

mitchellh:
The comment above this implies we're doing this on purpose. I want to understand both sets of problems more before approving this.

chrismo:
I have a [related Discussion](https://github.com/ghostty-org/ghostty/discussions/9183) to this issue. Currently the `increase_font_size` keybind [[src](https://github.com/ghostty-org/ghostty/blob/97a5a59cc3edf86cb939ca563b3bb43afe43a79a/src/config/Config.zig#L5664-L5668)] is using unicode `+` and the `next_tab` keybind [[src](https://github.com/ghostty-org/ghostty/blob/97a5a59cc3edf86cb939ca563b3bb43afe43a79a/src/config/Config.zig#L6144C1-L6148C15)] is using physical_key `bracket_right` ... Dvorak maps `+` to `bracket_right` so these two keybindings are in conflict.

As a Dvorak/Mac user for ~15 years, I agree with [this comment](https://github.com/ghostty-org/ghostty/discussions/7540#discussioncomment-13403871) on a prior Dvorak discussion, that Ghostty should not be shipping physical keybinds for keys like `bracket_right`/`bracket_left`

dontwanttothink:
I use a Spanish (Latin America) keyboard on macOS. Every other application uses the actual + key for this keybind —not the 'physical' or US + key, which in my keyboard maps to "¿". The menu item for "Increase Font Size" also currently shows the wrong keybind for this reason. It shows ⌘+, which doesn't work unless the menu bar is focused.

dwt:
On a German layout, cmd - decreases the font size, but cmd + does not increase it again. This is broken on nightlies for at least a month now and I would really like a fix. The proposed workaround above also doesn't work on my system.

dontwanttothink:
@dwt, I think this might be related to your issue, as my problem was identical: https://github.com/helix-editor/helix/discussions/14624. I originally wrote about it in the comments to this PR but deleted it because I wasn’t sure about the relevance.

One big factor to my confusion with this keybind was that the menu item asserted Cmd + would work —it does not, unless the menu bar is focused. Given that d015efc mentions making bindings ‘match macOS menus,’ perhaps this PR corrects that.

dwt:
@dontwanttothink Indeed, your comment helps and gives me a workaround for now with this snippet in my Ghostty config:

```
# Workaround for https://github.com/ghostty-org/ghostty/pull/8759
# Enables cmd+plus to work on german keyboard layouts
keybind=cmd+BracketRight=increase_font_size:1
```

Sorry, for grabbing your short name on GitHub. The long form of mine is onstensible: DontWannaTell - but I also really like yours. :)

dwt:
Looking at the code this pull request is trying to change, it only talks about the `=` key. Does this by any chance cover the `+` key too, or is this going to be another pull request?

PickleNik:
got confused on Dvorak, Mac
`⌘-` works as expected decreasing font-size (`-` in dvorak is `'"` qwerty keyboard button)
`⌘+` switched me to the next split pane (`=+` in dvorak is `}]` qwerty keyboard button) 
i got confused at first thinking the defaults were cooked until i tried the other button and looked up the keybinds, works right upon switching to qwerty tho

found lazy/temp fix `keybind = cmd+bracket_right=increase_font_size:1` from [related discord thread](https://discord.com/channels/1005603569187160125/1417378780829450280) on the issue to work (at least on built-in macbook keeb)
but it nukes the keybind for `goto_split:next`(both buttons increase font size now basically)

some people in [related GH Discussion](https://github.com/ghostty-org/ghostty/discussions/9183) (also linked above) posted some longer config changes that may or may not fix the `goto_split:next` and also address potential issues with external keebs but I ain't reading all that and I don't feel like having more than 2-3 lines in my config. And it also didn't work for at least one user in the comment with a Spanish layout.
[EDIT: another related discussion #7540 also seems to provide a similar, more full config fix]

this PR seems to only address the single `=` vs `]` key and not physical vs logical mapping, which is what this pr seems to be about https://github.com/ghostty-org/ghostty/pull/9469 so linking it down here for reference

mitchellh:
Replaced by #9469 thank you.

PR #8765, opened 2025-09-18
build: explicitly suffix generated Zig source files

pluiedev:
Previously we're just feeding the compiler source files generated via capturing stdout, which all by default have the filename `stdout`. Under some particular yet uncertain circumstances in Zig 0.14 (only factor we've found so far is a large amount of cores/compilation shards) the compiler will actually crash on an unreachable code path that assumes all source files either end in .zig or .zon, causing crashes for packagers for distros like Nixpkgs and Gentoo.

Given this has been explicitly made illegal in Zig 0.15 (see ziglang/zig#24957) I don't really see why we shouldn't fix this for 1.2 too. We have to do this at some point no matter what anyways.


mitchellh (approved):

jcollie (approved):
LGTM

PR #8767, opened 2025-09-18
slightly improve logs

asibahi:
Hello

This is a small thing I noticed when building Ghostty. The logs for SetTitle and Pwd are unreadable due to them being shown as an array of integers, so I added a custom formatter for them to be shows as text.

### DIFF
--- a/src/apprt/action.zig
+++ b/src/apprt/action.zig
@@ -569,6 +569,15 @@ pub const SetTitle = struct {
             .title = self.title.ptr,
         };
     }
+
+    pub fn format(
+        value: @This(),
+        comptime _: []const u8,
+        _: std.fmt.FormatOptions,
+        writer: anytype,
+    ) !void {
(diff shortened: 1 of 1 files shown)


src/apprt/action.zig
```diff
@@ -603,6 +621,19 @@ pub const DesktopNotification = struct {
...
         };
     }
+
+    pub fn format(
+        value: @This(),
+        comptime _: []const u8,
+        _: std.fmt.FormatOptions,
+        writer: anytype,
+    ) !void {
+        try writer.print("{s}{{ title: {s}, body: }}", .{
```
mitchellh: I think this is missing an interpolation.

mitchellh (changes requested):
Thanks, I'm good with it, one note.

src/apprt/action.zig
asibahi: oops. fixed.

mitchellh (approved):

PR #8768, opened 2025-09-18
macos: implement bell-features=border on macOS

mitchellh:
https://github.com/user-attachments/assets/ac58ff8e-9ef6-4cb2-8e40-9f588aaaca4e


mitchellh:
I think the aesthetic here can be improved significantly. Namely would love to see a gradient inwards so it's not a harsh solid border. The color is questionable too. Open to suggestions.

PR #8771, opened 2025-09-19
Snap: Do not leak snap variables or snap paths into children

3v1n0:
Avoid leaking snap environment values and in particular the `$SNAP*` values to the children that we run from the terminal.

Do this programmatically instead of the launcher, since ghostty needs know the environment it runs in, while it must not leak the info to the children.

We've also another leak on snap, that creates a more visible problem (wrong matching of children applications) that is the apparmor security profile.

I've handled it at https://github.com/3v1n0/ghostty/commit/cc3b46f6001aa9539508a1e82ebb85a588c6c4f2 but that requires some love in order to fully decouple the snap option to the build, to avoid including it in non-snap builds, so an help would be appreciated there.

> This PR was contains code (in `filterSnapPaths`) that was improved by DeepSeek.

### DIFF
--- a/snap/local/launcher
+++ b/snap/local/launcher
@@ -61,10 +61,4 @@ fi
 
 [ "$needs_update" = true ] && echo "LAST_REVISION=$SNAP_REVISION" > "$SNAP_USER_DATA/.last_revision"
 
-# Unset all SNAP specific environment variables to keep them from leaking
-# into other snaps that might get executed from within the shell
-for var in $(printenv | grep SNAP_ | cut -d= -f1); do
-  unset $var
-done
-
 exec "$@"
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -80,7 +80,7 @@ parts:
     override-build: |
       craftctl set version=$(cat VERSION)
       $CRAFT_PART_SRC/../../zig/src/zig build \
-        -Dsnap=true \
+        -Dsnap \
         -Dpatch-rpath=\$ORIGIN/../usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR:/snap/core24/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR \
         -Doptimize=ReleaseFast \
         -Dcpu=baseline \
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -20,7 +20,7 @@ platforms:
 apps:
   ghostty:
     command: bin/ghostty
-    command-chain: [bin/launcher]
+    command-chain: [app/launcher]
     completer: share/bash-completion/completions/ghostty.bash
     desktop: share/applications/com.mitchellh.ghostty.desktop
     #refresh-mode: ignore-running # Store rejects this, needs fix in review-tools
@@ -35,7 +35,7 @@ parts:
     source: snap/local
     source-type: local
     organize:
-      launcher: bin/
+      launcher: app/
 
   zig:
     plugin: nil
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -79,7 +79,12 @@ parts:
     # TODO: Remove -fno-sys=gtk4-layer-shell when we upgrade to a version that packages it Ubuntu 24.10+
     override-build: |
       craftctl set version=$(cat VERSION)
-      $CRAFT_PART_SRC/../../zig/src/zig build -Dpatch-rpath=\$ORIGIN/../usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR:/snap/core24/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR -Doptimize=ReleaseFast -Dcpu=baseline -fno-sys=gtk4-layer-shell
+      $CRAFT_PART_SRC/../../zig/src/zig build \
+        -Dsnap=true \
+        -Dpatch-rpath=\$ORIGIN/../usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR:/snap/core24/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR \
+        -Doptimize=ReleaseFast \
+        -Dcpu=baseline \
+        -fno-sys=gtk4-layer-shell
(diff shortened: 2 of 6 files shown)


src/apprt/gtk/class/surface.zig
```diff
@@ -1227,19 +1228,10 @@ pub const Surface = extern struct {
...
-            env.remove("__EGL_VENDOR_LIBRARY_DIRS");
-            env.remove("LD_LIBRARY_PATH");
-            env.remove("LIBGL_DRIVERS_PATH");
-            env.remove("LIBVA_DRIVERS_PATH");
-            env.remove("VK_LAYER_PATH");
-            env.remove("XLOCALEDIR");
-            env.remove("GDK_PIXBUF_MODULEDIR");
-            env.remove("GDK_PIXBUF_MODULE_FILE");
-            env.remove("GTK_PATH");
+        if (build_config.snap) {
```
mitchellh, line 1231: Just to be safe, let's do `if (comptime build_config..)` here.

snap/snapcraft.yaml
```diff
@@ -35,7 +35,7 @@ parts:
     source: snap/local
     source-type: local
     organize:
-      launcher: bin/
+      launcher: app/
```
mitchellh, line 38: Can you explain why you changed this?

snap/local/launcher
```diff
@@ -61,10 +61,4 @@ fi
 
 [ "$needs_update" = true ] && echo "LAST_REVISION=$SNAP_REVISION" > "$SNAP_USER_DATA/.last_revision"
 
-# Unset all SNAP specific environment variables to keep them from leaking
-# into other snaps that might get executed from within the shell
-for var in $(printenv | grep SNAP_ | cut -d= -f1); do
-  unset $var
-done
```
mitchellh, line 68: > Do this programmatically instead of the launcher, since ghostty needs know the environment it runs in, while it must not leak the info to the children.

Can you explain this? Why doesn't doing it in the launcher before it starts work?

src/apprt/gtk/class/surface.zig
```diff
@@ -1253,6 +1245,94 @@ pub const Surface = extern struct {
         return env;
     }
 
+    /// Filter out environment variables that start with forbidden prefixes
+    fn filterSnapPaths(allocator: std.mem.Allocator, env_map: *std.process.EnvMap) !void {
```
mitchellh, line 1249: I'd feel better if we split this out into the pure/impure logical sections so we can unit test it.

mitchellh (changes requested):
Thanks for improving the snap functionality. I have some questions, some of which are simply due to my lack of understanding of Snap, others are more general code style.

snap/snapcraft.yaml
3v1n0: So, if you currently launch the ghostty from a snap and you do `which launcher` you'd get `/snap/ghostty/<version>/bin`, this is because the binary ends in a well know $PATH.

So instead of moving it to a a folder in $PATH, just move it somewhere else that is not tracked and then when you do `ghostty` the snap machinery will still call it directly

snap/local/launcher
3v1n0: We need to know these variables now in order to filter them out... So better to just do it inside ghostty instead of spreading the same logic in two places.

src/apprt/gtk/class/surface.zig
3v1n0: Mh, I'm not into zig so much to understand this well 😄

snap/snapcraft.yaml
mitchellh: Thanks, got it!

src/apprt/gtk/class/surface.zig
```diff
@@ -1227,19 +1228,10 @@ pub const Surface = extern struct {
 
         // Unset environment varies set by snaps if we're running in a snap.
         // This allows Ghostty to further launch additional snaps.
-        if (env.get("SNAP")) |_| {
-            env.remove("SNAP");
-            env.remove("DRIRC_CONFIGDIR");
```
mitchellh, line 1232: Sorry I missed this before, but is removing these other env vars no longer necessary? I believe these were all added because they caused some other issues but just checking.

src/apprt/gtk/class/surface.zig
mitchellh: No problem! I can take this over and clear this out and you can see my changes then. 😄

src/apprt/gtk/class/surface.zig
kenvandine: Yeah, @3v1n0 is filtering out all environment variables that could be leaking out of the ghostty snap more dynamically than my static list of env variables. This should be more future proof and thorough than what I had done.

3v1n0: Ah, as Ken said... We're filtering them by their contents, instead of by their name, so the host environment values that are good (as `XDG_CACHE_HOME`, if set) are re-used only if they point to a non-snap location, otherwise we just unset them.

And the idea here is to avoid keep a list updated.

src/apprt/gtk/class/surface.zig
3v1n0: Nice, I would appreciate it :)

Also if you could include https://github.com/3v1n0/ghostty/commit/cc3b46f6001aa9539508a1e82ebb85a588c6c4f2 (in the right place)... Ideally I would like to move it to a file that was compiled only in the snap case, and adding then a wrapper that is a no-op other than in snaps, but it was too much for me 🥵 😂

mitchellh:
Okay, I made some changes. I didn't do the testing thing I requested, I think it's fine on second look. But I made a couple other changes:

- Used a single arena allocator rather than dealing with allocations, cause all of it is temporary state
- Zig stylistic things
- Added CI job to ensure `-Dsnap` builds, that's faster than waiting for the full snap build and our snap build depends on that

mitchellh (approved):
Approve pending CI. Thank you!

src/apprt/gtk/class/surface.zig
mitchellh: Yeah this looks good, I can see a path to make this happen. Let me make an issue to track it but won't pollute this PR with it.

PR #8776, opened 2025-09-19
Optimize PNG images losslessly: 73 → 51 KiB (−30%)

maximal:


pluiedev (changes requested):
This doesn't work since we use z2d's PNG export function to generate a PNG which can be compared to the test data:
> [font_sprite] (err): !!! Test PNG data does not match reference, but pixels do match! Either z2d's PNG exporter changed or someone is trying to smuggle binary data in the test files!

If you want to fix this, you need to optimize z2d's PNG exporter (which is admittedly very crude and used only for testing purposes) instead of trying to fix the test images here.

maximal:
@pluiedev, ah, I see. Maybe I shouldn’t have touched `src/font/sprite/testdata` and commit only `test/cases`. Or, they also use `z2d`?

pluiedev:
I think those are fine to touch, yeah

mitchellh:
Aside, if there is a CLI tool we can use to do this, we can add this to CI to ensure all PNGs added are losslessly compressed as well as they can be (by just comparing file size, since I'm guessing its not byte-equivalent every time)

maximal:
@mitchellh, sure!

Basically, I use `optipng -o7 -zm1-9 image.png`, it bruteforces different filters, palettes, and algorithms in PNG data to make it smaller losslessly using classic zlib/gzip compression.

There is a relatively new tool `zopflipng` from Google, it can do smaller, but it uses a new compression algorithm (https://en.wikipedia.org/wiki/Zopfli), I didn’t test its compatibility with the old tools though.

PR #8779, opened 2025-09-19
gtk: add glib log writer function

jcollie:
This will funnel GLib/GObject/GTK logging messages through Zig's logging system rather than just dumping them directly to stderr.

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -42,6 +42,70 @@ const GlobalShortcuts = @import("global_shortcuts.zig").GlobalShortcuts;
 
 const log = std.log.scoped(.gtk_ghostty_application);
 
+/// Function used to funnel GLib/GObject/GTK log messages into Zig's logging
+/// system rather than just getting dumped directly to stderr.
+fn glibLogWriterFunction(
+    level: glib.LogLevelFlags,
+    fields: [*]const glib.LogField,
+    n_fields: usize,
+    _: ?*anyopaque,
+) callconv(.c) glib.LogWriterOutput {
+    const glib_log = std.log.scoped(.glib);
+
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/class/application.zig
```diff
@@ -42,6 +42,70 @@ const GlobalShortcuts = @import("global_shortcuts.zig").GlobalShortcuts;
...
+            domain_ = v;
+            continue;
+        }
+    }
+
+    const message = message_ orelse return .unhandled;
+    const domain = domain_ orelse "«unknown»";
+
+    if (level.level_error) {
+        glib_log.err("ERROR: {s}: {s}", .{ domain, message });
```
3v1n0, line 82: This is calling `g_error` eventually? As it would also imply a crash, no?

mitchellh, line 82: I'd avoid calling `err` on any Zig logger unless we're calling it directly. The `err` level is reserved for cases where the program will crash immediately after (it's unrecoverable). I'm not sure if this maps to the same concept for glib.

mitchellh (changes requested):

src/apprt/gtk/class/application.zig
3v1n0: Ok, good... So it's the same in GLib, then we're good.

mitchellh: If its the same then we can keep the `err` here, but then the question remains if glib still crashes the program after these or if we have to do it ourselves.

3v1n0: I didn't test for the zig bindings, but indeed caling `g_error` causes a `SIGABRT`

mitchellh (approved):

Issue #8681, opened 2025-09-16
config: default `shell-integration-features = no-cursor` if `cursor-style` is manually set


mitchellh:
This pitfall has bit us for far too long.

To make this change happen, I want to do it the right way: we should use comptime to create a bitset on `Config` that flags what configs were _manually set_ (via any source). Or, in other words, which configs are _not their default value_ (but may have been manually set to their default value!).

Based on that, in `finalize`, we should set `shell-integration-features` to disable the `cursor` feature if `cursor-style` or `cursor-style-blink` is manually set.

mitchellh:
This should go into 1.3 and not 1.2.x.

PR #8780, opened 2025-09-19
core: add bitset to Config to track what configs have been changed

jcollie:
Does not yet add the machinery to actually update the bitset.

Related to #8681


mitchellh:
I’m pretty sure this will be a feature of the general CLI args parsing and not our config structure directly. Similar to how we handle arenas or diagnostics

mitchellh:
Going to close this since its stale and not the direction I really want to go (prior notes noted).

PR #8782, opened 2025-09-19
renderer/opengl: minimum contrast for black sets proper color

mitchellh:
Fixes #8745

When rendering black for minimum contrast we were setting opacity to 0 making it invisible.


qwerasd205 (approved):
Wow, oops, my bad 😂

mitchellh:
Lol given the magnitude of the changes, these are minor. 😄 I missed it in review too!

PR #8786, opened 2025-09-19
xdg: treat empty env values as unset and fall back to home

dynamicy:
## Summary
Handle zero-length environment values in `src/os/xdg.zig`’s `dir` function.
If an env value is present but empty, treat it as unset and fall back to the
default home-based location. Apply the same rule to the Windows fallback branch
that uses `LOCALAPPDATA`.

## Changes
- Detect `env.len == 0` and free any owned allocation for that env value.
- Continue resolution as if the variable were not set, using the default
  home-based location.
- Apply the same logic to the Windows fallback (`LOCALAPPDATA`) branch so both
  the primary and fallback env vars obey the rule.
- Add/adjust tests that set XDG_* variables to an empty string and verify the
  default locations are used.

### DIFF
--- a/src/os/xdg.zig
+++ b/src/os/xdg.zig
@@ -74,32 +74,45 @@ fn dir(
         else => .{ posix.getenv(internal_opts.env), false },
         .windows => windows: {
             if (std.process.getEnvVarOwned(alloc, internal_opts.env)) |env| {
+                if (env.len == 0) {
+                    alloc.free(env);
+                } else {
+                    break :windows .{ env, true };
+                }
+            } else |err| switch (err) {
+                error.EnvironmentVariableNotFound => {},
+                else => return err,
+            };
+
+            if (std.process.getEnvVarOwned(alloc, internal_opts.windows_env)) |env| {
+                if (env.len == 0) {
+                    alloc.free(env);
+                    break :windows .{ null, false };
+                }
                 break :windows .{ env, true };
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
The core idea is fine. 

The code needs work. Empty if statements are weird. We can better use Zig's block semantics to minimize duplication. Random blocks for no reason at all.

Was this AI written?

dynamicy:
> The core idea is fine.
> 
> The code needs work. Empty if statements are weird. We can better use Zig's block semantics to minimize duplication. Random blocks for no reason at all.
> 
> Was this AI written?

Thanks for the feedback @mitchellh!

You're absolutely right about the code quality issues. I used AI assistance for the initial implementation, and it definitely showed in some awkward patterns like the empty if statements and overly nested logic.

I've completely refactored the code based on your suggestions.

The core logic remains the same, but the control flow is much cleaner and more readable. I really appreciate you pushing back on the code quality - it resulted in a much better implementation.

Would love your thoughts on the refactored approach!

mitchellh:
We require AI disclosure for this project. The code quality is still trash. I'm sorry. Rejected.

Just so I'm clear, we absolutely accept AI contributions. I myself use AI a lot. But we expect the drivers to be experts that can understand and clean up their own code. This doesn't appear to be the case here. I'm sorry.

PR #8796, opened 2025-09-19
macos: quick terminal stores the last closed size by screen

mitchellh:
Fixes #8713

This stores the last closed state of the quick terminal by screen
pointer. We use a weak mapping so if a screen is unplugged we'll clear
the memory. We will not remember the size if you unplug and replug in a
monitor.

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -21,13 +21,13 @@ class QuickTerminalController: BaseTerminalController {
     // The active space when the quick terminal was last shown.
     private var previousActiveSpace: CGSSpace? = nil
 
-    /// The window frame saved when the quick terminal's surface tree becomes empty.
+    /// The saved state when the quick terminal's surface tree becomes empty.
     ///
     /// This preserves the user's window size and position when all terminal surfaces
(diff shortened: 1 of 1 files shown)


edrozenberg:
This change causes the quick terminal to revert to a 400px height despite manual resize and despite any `quick-terminal-size` setting. With the following steps:

1) Show/hide quick terminal on Mac main/first desktop
2) Switch to another desktop for ex. a fullscreen one w/a Ghostty window
3) Show/hide quick terminal
4) Switch back to first desktop
5) Show/hide quick terminal
6) Quick terminal height reverted to 400px

MacOS 15.7 (24G222)

mitchellh:
I ran your exact steps @edrozenberg and my prior size was maintained. Perhaps key-ing by the weak ref to the screen isn't right and for some reason your system is freeing them. I wonder if there is a screen ID we can use instead...

edrozenberg:
> I ran your exact steps @edrozenberg and my prior size was maintained.

Thanks for checking! Was able to repro earlier but not anymore... Will submit a discussion if I see it again.

mitchellh:
Yeah I bet you keying on the `NSScreen` is not the way to go. I bet macOS regularly recycles those.

radiosilence:
Hmm, since the latest release the quick terminal is now:

1. Is stuck as being too wide for my 14" MBP display.
2. Is consistently too small on my main external display.

PR #8798, opened 2025-09-19
build: lower build.zig dependencies to improve caching

mitchellh:
This is stomping towards minimizing our build.zig dependencies so that it can be cached more often. Right now, touching almost any file in the project forces the build.zig to rebuild which is destroying my productivity.

The first set of work is to move all our completion and syntax file generation into a data generator exe, e.g. vim, zsh, fish, etc. This might seem like just shuffling bits but it results in a real tangible improvement: when you run `zig build test`, we no longer have to rebuild our `build.zig` when you for example... modify a CLI action. 

This is just a small improvement. Our build.zig depends on way too much stuff, so this PR will be draft while I continue to use commits to separate out scopes of change.


mitchellh:
This is now complete. Our `build.zig` dependencies now only depend on a handful of files transitively rather than hundreds. A very simple test:

```
$ zig build test -Dtest-filter="parse"
```

Then modify `Config.zig`. Previously this would rebuild the build binary. Now it doesn't. True for many more files.

mitchellh:
@pluiedev If you want to do the output flag for our build data binary, feel free in a follow up PR. I'll try to do so as well, but the wins in here especially for GTK dev should be huge.

PR #8801, opened 2025-09-20
feat: add liquid glass background effect support

Lustyn:
## Description

This PR implements the basic functionality for "Liquid Glass" style background support (#8155). I used OpenCode (Claude 4) in some capacity to write this as I'm not super familiar with AppKit / SwiftUI. A good chunk of this I still needed to write by hand since Claude doesn't understand the Glass APIs, but I'm not 100% if the implementation here makes the best decisions since the practices in Ghostty config and separation of the AppKit code and SwiftUI seemed inconsistent to me.

Some of the combinations of options obviously create entirely unreadable terminals, but I've found that regular glass and transparent with opacity to be fairly readable. We *don't* enable this feature by default since it would of course break existing users setups.

## Open Questions

- [x] How to determine the correct cornerRadius? For now this is eyeballed, I can't see any macOS public API or clearly documented constants.
- [x] Should boolean options be exposed for reasonable defaults?
- [x] Should the option need to be namespaced to macos-\*?

## Screenshots

0% Opacity, Regular
image
50% Opacity, Transparent
image
0% Opacity, Transparent
image

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -561,7 +561,7 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
             vsync: bool,
             colorspace: configpkg.Config.WindowColorspace,
             blending: configpkg.Config.AlphaBlending,
-            macos_background_style: configpkg.Config.MacBackgroundStyle,
+            background_blur: configpkg.Config.BackgroundBlur,
 
             pub fn init(
                 alloc_gpa: Allocator,
@@ -634,7 +634,7 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     .vsync = config.@"window-vsync",
                     .colorspace = config.@"window-colorspace",
                     .blending = config.@"alpha-blending",
-                    .macos_background_style = config.@"macos-background-style",
+                    .background_blur = config.@"background-blur",
                     .arena = arena,
                 };
             }
@@ -645,16 +645,6 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                 self.arena.deinit();
             }
         };
-
-        /// Determines if the terminal background should be disabled based on platform and config.
-        /// On macOS, when background effects are enabled (background style != default), the effect
-        /// layer handles the background rendering instead of the terminal renderer.
-        fn shouldDisableBackground(config: DerivedConfig) bool {
-            return switch (builtin.os.tag) {
-                .macos => config.macos_background_style != .default,
-                else => false,
-            };
-        }
 
         pub fn init(alloc: Allocator, options: renderer.Options) !Self {
             // Initialize our graphics API wrapper, this will prepare the
@@ -728,6 +718,9 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                         options.config.background.r,
                         options.config.background.g,
                         options.config.background.b,
+                        // Note that if we're on macOS with glass effects
+                        // we'll disable background opacity but we handle
+                        // that in updateFrame.
                         @intFromFloat(@round(options.config.background_opacity * 255.0)),
                     },
                     .bools = .{
@@ -1305,10 +1298,18 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     self.terminal_state.colors.background.r,
                     self.terminal_state.colors.background.g,
                     self.terminal_state.colors.background.b,
-                    if (shouldDisableBackground(self.config))
-                        0
-                    else
-                        @intFromFloat(@round(self.config.background_opacity * 255.0)),
+                    @intFromFloat(@round(self.config.background_opacity * 255.0)),
+                };
+
+                // If we're on macOS and have glass styles, we remove
+                // the background opacity because the glass effect handles
+                // it.
+                if (comptime builtin.os.tag == .macos) switch (self.config.background_blur) {
+                    .@"macos-glass-regular",
+                    .@"macos-glass-clear",
+                    => self.uniforms.bg_color[3] = 0,
+
+                    else => {},
                 };
             }
         }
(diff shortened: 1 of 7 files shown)


Aaron-212:
> How to determine the correct cornerRadius? For now this is eyeballed, I can't see any macOS public API or clearly documented constants.

https://www.reddit.com/r/MacOSBeta/comments/1lb2jt5/window_corner_radius_in_macos_tahoe_depends_on/
They are fixed. You can do a simple switch-case on the config `macos-titlebar-style`: 16pt for `native`, `transparent` and `hidden`, 20pt for `tabs`.

> Should the option need to be namespaced to macos-*?

IMHO the config should not be limited to Liquid Glass effect. A better config name would be `macos-background-style` with following options:
- `default`: uses regular blur
- `regular-glass` and `clear-glass`: Liquid Glass effect (implemented by this PR)
- other `NSVisualEffectView` styles

Lustyn:
Thanks for the find on the corner radius!

I think vibrancy / NSVisualEffect is a good call out for how this feature should be structured. On the blur option I'm a bit unsure of since in the current state this allows users to mix glass, traditional blur, and opacity to achieve a little more readability.

(But if we have reasons we think mixing them is a no-no, that's ok!)

Lustyn:
For now I've made standalone blur and the glass effects mutually exclusive, likewise I think the same would be true of NSVisualEffect if I understand correctly. My thoughts are that this will reduce confusion for people enabling the glass effect that already have a blur value set. A blur + glass merged mode or override might make more sense for users that actually intend on doing that?

insomnia-creator:
Any updates on this PR?

Lustyn:
Just waiting on a review, happy to make any changes the maintainers see fit.

tisba:
Question out of curiosity: Will this respect the system wide reduce transparency setting?

Aaron-212:
> Question out of curiosity: Will this respect the system wide reduce transparency setting?

Yes. `NSGlassEffectView` (and `NSVisualEffectView`) respects "Reduce Transparency" settings.

mitchellh (approved):
I like this. Really well done.

mitchellh (changes requested):
Actually sorry, there is an issue. This breaks `macos-titlebar-style = tabs`. This is probably due to recent fixes. If someone could take a look at that and fix that up I'll be happy to merge this. Thank you!

Lustyn:
image

I don't think I'm able to reproduce that. Looks fine on my end after your rebase. Is this only an issue on earlier macOS versions?

danielwe:
@Lustyn I think the issue is that the color/transparency of the titlebar isn't identical to the terminal surface, so you get a style break. IIUC the solution is just to make the `tabs` titlebar fully transparent when the glass effect is enabled, like you do for surface rendering. The same probably goes for the `macos-titlebar-style = transparent` variant.

`macos-background-style = default`
Screenshot 2025-10-17 at 11 04 54

`macos-background-style = regular-glass`
Screenshot 2025-10-17 at 11 05 49

Lustyn:
Aha! Thanks for the pointer, that completely slipped by me because of the debug indicator. I have the fix:
image

insomnia-creator:
Anything left before being merged? Really excited for this PR.

Tankenstein:
Seconded @insomnia-creator. I'm happy to fix conflicts myself if it helps to move this closer to release.

src/renderer/generic.zig
```diff
@@ -1301,7 +1316,10 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     critical.bg.r,
                     critical.bg.g,
                     critical.bg.b,
-                    @intFromFloat(@round(self.config.background_opacity * 255.0)),
+                    if (shouldDisableBackground(self.config))
+                        0
+                    else
+                        @intFromFloat(@round(self.config.background_opacity * 255.0)),
                 };
```
danielwe: (Disclaimer: I don't make any decisions, I'm just commenting as someone who's also interested in this.)
I think that, per https://github.com/ghostty-org/ghostty/pull/9260#pullrequestreview-3353915814, this approach is a non-starter. Background color and opacity must be the responsibility of the core renderer. Can the macOS app only do the liquid glass blurring effect without adding color and alpha?

Lustyn: If we split the rendering into two passes that is technically possible, but would actually render the liquid glass effect incorrectly. These effects are background aware (behind the application itself) and not just simple transparent overlays, so the concerns mentioned in #9260 would eliminate the possibility of this feature if we stick to that.

I think the point that this conflicts with some options like alpha blending is valid, however it explicitly only changes this behavior when liquid glass is enabled, and does not affect those consuming the renderer alone. That comment also mentions branching on OS being a bad idea, though its a bit confusing if that would actually apply here since this feature is inherently locked to macOS and it isn't feasible to support on other platforms.

mitchellh (approved):
This looks good and ready to go. Thank you for the diligence on this issue. I made one major configuration change but no core logic changes:

I unified the config into `background-blur` as special macOS-only enum values. If these are set, they imply `true` on non-macOS platforms.

mitchellh:
I should note that I suspect this may cause some edge cases with custom shaders and may impact alpha blending as noted above. I think highlighting this in the docs is the way to go just to give people the creative freedom to have this style if they want it if they don't care about those other features. It's default off and very much an optional thing so I think that's alright.

Lustyn:
The blur change makes sense, thank you for reviewing and solidifying the PR! Looking forward to using it in my Ghostty soon ❤️

8pxl:
not sure if this is a bug with implementation or just a quirk of liquid glass (either way would like to work towards some fix), but the appearance between focused and unfocused windows is extremely jarring.



image

PR #8806, opened 2025-09-20
minor: service should not quit after last window is closed

abenz1267:
Hello,

currently the ghostty service is pointless if you don't have `quit-after-last-window-closed = false` in your config. The problem: if you HAVE that AND run ghostty with `--class=my.class` ghostty will launch a new instance, which will be kept open as well after closing, so you'd need to run `--close-after-last-window-closed=true --class=my.class` which seems... wrong?

is there a reason the service should EVER be closed? Unless ofc you actually stop the systemd service.

if not this pr should be fine?

Regards

### DIFF
--- a/dist/linux/dbus.service.flatpak.in
+++ b/dist/linux/dbus.service.flatpak.in
@@ -1,3 +1,3 @@
 [D-BUS Service]
 Name=@APPID@
-Exec=@GHOSTTY@ --gtk-single-instance=true --initial-window=false
+Exec=@GHOSTTY@ --gtk-single-instance=true --quit-after-last-window-closed=false --initial-window=false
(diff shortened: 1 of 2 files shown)


jcollie:
The problem is that the service is triggered in more situations than just manually enabling the service to start when the user logs in. It's also triggered if the user just clicks on the desktop icon (and they have DBus activation enabled). In that case, I would think that the user could quite reasonably expect Ghostty to quit completely when there are no more windows. Most Linux apps behave that way (e.g. Firefox) though there are exceptions. If the user wants something different they need to manually add `quit-after-last-window-closed=false` to their config.

Then there's also the fact that CLI arguments take precedence over anything set in the default configuration file. So adding `--quit-after-last-window-closed=false` to the service means that the user could never change that without editing the service file.

PR #8810, opened 2025-09-20
unicode: isolate properties, tables, and ziglyph into separate files

mitchellh:
This makes it cleaner visually where the separation of concerns is. There is now the generic `Properties.zig`, and then the implementation-specific `props_<impl>.zig` files. Despite Zig's lazy analysis, I find this is much easier to understand as a human.

Doing this resulted in finding one part in `src/terminal` where we were still inadvertently using ziglyph directly instead of our LUTs! I switched this out.

After this PR, `src/terminal` as a standalone module no longer depends on `ziglyph` at all.[^1]

cc @jacobsandlund this is going to cause conflicts in your PR. I'm sorry about that. But it should make it cleaner to bring in the uucode work by adding a dedicated `props_uucode.zig` file!

[^1]: Why would I be talking about `src/terminal` as a standalone module? That's interesting.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -344,7 +344,7 @@ pub fn print(self: *Terminal, c: u21) !void {
             // VS15 makes it narrow.
             if (c == 0xFE0F or c == 0xFE0E) {
                 // This only applies to emoji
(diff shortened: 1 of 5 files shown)


mitchellh:
Oh whoops, forgot to delete props.zig. I'll follow this up.

PR #8813, opened 2025-09-21
GTK Fix unfocused-split-fill

rhodes-b:
Attempts a resolution for https://github.com/ghostty-org/ghostty/discussions/8572

This matches the behavior of the old GTK apprt where unfocused-split-fill /opacity doesn't apply when there is only one active surface.

### DIFF
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -710,13 +710,7 @@ pub const SplitTree = extern struct {
         // We never CLEAR our last_focused because the property is specifically
         // the last focused surface. We let the weakref clear itself when
         // the surface is destroyed.
-        if (!surface.getFocused()) {
-            // If we have more than 1 active surface and we aren't focused we want to apply unfocused-split-fill
-            if (self.getTree()) |tree| {
-                if (tree.nodes.len > 1) surface.setUnfocusedSplit();
-            }
-            return;
-        }
+        if (!surface.getFocused()) return;
         self.private().last_focused.set(surface);
 
         // Our active surface probably changed
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -399,12 +399,7 @@ pub const SplitTree = extern struct {
         const tree = self.getTree() orelse return null;
         var it = tree.iterator();
         while (it.next()) |entry| {
-            if (!entry.view.getFocused() and it.nodes.len > 1) {
-                entry.view.setUnfocused();
-            }
-            if (entry.view.getFocused()) {
-                return entry.handle;
-            }
+            if (entry.view.getFocused()) return entry.handle;
         }
 
         // If none are currently focused, the most previously focused
@@ -715,7 +710,13 @@ pub const SplitTree = extern struct {
         // We never CLEAR our last_focused because the property is specifically
         // the last focused surface. We let the weakref clear itself when
         // the surface is destroyed.
-        if (!surface.getFocused()) return;
+        if (!surface.getFocused()) {
+            // If we have more than 1 active surface and we aren't focused we want to apply unfocused-split-fill
+            if (self.getTree()) |tree| {
+                if (tree.nodes.len > 1) surface.setUnfocused();
+            }
+            return;
+        }
         self.private().last_focused.set(surface);
 
         // Our active surface probably changed
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -399,7 +399,12 @@ pub const SplitTree = extern struct {
         const tree = self.getTree() orelse return null;
         var it = tree.iterator();
         while (it.next()) |entry| {
-            if (entry.view.getFocused()) return entry.handle;
+            if (!entry.view.getFocused() and it.nodes.len > 1) {
+                entry.view.setUnfocused();
+            }
+            if (entry.view.getFocused()) {
+                return entry.handle;
+            }
         }
 
         // If none are currently focused, the most previously focused
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -172,9 +172,6 @@ pub const SplitTree = extern struct {
         /// close dialog.
         pending_close: ?Surface.Tree.Node.Handle,
 
-        /// True if the current split tree contains more than one surface
-        is_split: bool = false,
-
         pub var offset: c_int = 0;
     };
 
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -713,7 +713,7 @@ pub const SplitTree = extern struct {
         if (!surface.getFocused()) {
             // If we have more than 1 active surface and we aren't focused we want to apply unfocused-split-fill
             if (self.getTree()) |tree| {
-                if (tree.nodes.len > 1) surface.setUnfocusedFill();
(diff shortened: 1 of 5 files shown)


jcollie (changes requested):
Instead of adding/removing the `DrawingArea` programmatically this should be done mostly in the blueprint. See how the border bell is done for inspiration.

jcollie (changes requested):
Excellent, I think that we're getting closer! Instead though of explicitly setting an `unfocused_split` property on a surface, we should be setting a `n_siblings` (or some such named property) on the surface which is basically the number of splits - 1. Then the `closureShouldUnfocusedSplitBeShown` depends on the `n_siblings` and `focused`.

rhodes-b:
Would there be some other way to call `setNSiblings` for the surfaces outside of the tree?

the failing test makes me think im not doing it the right way (since that method does not exist on `TestView`)

when I first added it I was calling it inside of the `class/split_tree.zig` area not in the datastruct but I couldn't see how surfaces were getting deleted in the `class/split_tree.zig`

`closeConfirmationClose` only applies when the confirm close dialog popup happens? that is the only place I saw them getting removed

src/apprt/gtk/class/surface.zig
```diff
@@ -502,6 +522,11 @@ pub const Surface = extern struct {
         /// A weak reference to an inspector window.
         inspector: ?*InspectorWindow = null,
 
+        /// Number of siblings related to this surface. This is used for
+        /// the unfocused-split-* options which are only applied when their is
```
rhodes-b: `there` not `their`

jcollie:
> Would there be some other way to call `setNSiblings` for the surfaces outside of the tree?
> 
> the failing test makes me think im not doing it the right way (since that method does not exist on `TestView`)
> 
> when I first added it I was calling it inside of the `class/split_tree.zig` area not in the datastruct but I couldn't see how surfaces were getting deleted in the `class/split_tree.zig`
> 
> `closeConfirmationClose` only applies when the confirm close dialog popup happens? that is the only place I saw them getting removed

I think that it would be better to create a `n-surfaces` property on the GTK split_tree class, and then bind that property to a `n-surfaces` property on each surface. Then the condition becomes `n-surfaces > 1 and !focused`. The properties should be updated automatically when the property is updated on the split_tree class.

mitchellh:
Note we went down the `n-surfaces` path before. The data structure being used here is not amenable to efficient counting (a tree, you'd have to traverse it). If we can get away with an `is-split` property that would be much, much cheaper (just check if the root is a leaf node or not).

jcollie:
> If we can get away with an `is-split` property that would be much, much cheaper (just check if the root is a leaf node or not

Yeah, I think that would work, we just need to know if there's more than one surface.

rhodes-b:
> I think that it would be better to create a `n-surfaces` property on the GTK split_tree class, and then bind that property to a `n-surfaces` property on each surface. 

Is there example of something similar?
I can't figure out how to make a property in surface bound to a property in the split tree class

I assume you mean all the surfaces share whatever the value is in the split tree class is that way I don't have to iterate the tree and manually update the property on each of them.

I got something working in the split tree class in the `onRebuild` method but I still have to iterate all of the nodes to individually tell each one if the root node was a leaf or not. (I did swap to detecting if the root node is a `.leaf` or `.split` member so that seems to be working fine)

mitchellh:
FYI I haven't had a chance to look at this in detail but can help more next week if you don't get there yet. Thanks!

jcollie:
> > I think that it would be better to create a `n-surfaces` property on the GTK split_tree class, and then bind that property to a `n-surfaces` property on each surface.
> 
> Is there example of something similar? I can't figure out how to make a property in surface bound to a property in the split tree class
> 
> I assume you mean all the surfaces share whatever the value is in the split tree class is that way I don't have to iterate the tree and manually update the property on each of them.

Here's one spot where we manually bind some properties together. So basically I'd create an `is-split` property on the split tree class and on the surfaces. Then when a surface is created bind the split tree `is-split` property to the surface `is-split` property. Then when the split tree property is updates all of the surfaces get updated automatically.

https://github.com/ghostty-org/ghostty/blob/bc95317ba45a686dcda9721a0d5148a53378099e/src/apprt/gtk/class/window.zig#L399-L411

src/apprt/gtk/class/split_tree.zig
```diff
@@ -112,6 +112,25 @@ pub const SplitTree = extern struct {
...
+                name,
+                Self,
+                bool,
+                .{
+                    .default = false,
+                    .accessor = gobject.ext.typedAccessor(
+                        Self,
+                        bool,
+                        .{
+                            .setter = setIsSplit,
```
jcollie: I think that this would be better if there were only a getter, here `getIsSplit`.

src/apprt/gtk/class/split_tree.zig
```diff
@@ -153,6 +172,9 @@ pub const SplitTree = extern struct {
         /// close dialog.
         pending_close: ?Surface.Tree.Node.Handle,
 
+        /// True if the current split tree contains more than one surface
+        is_split: bool = false,
+
```
jcollie: With just a getter on `is-split`, this field would not be needed.

src/apprt/gtk/class/split_tree.zig
```diff
@@ -511,6 +541,12 @@ pub const SplitTree = extern struct {
         ));
     }
 
+    fn setIsSplit(self: *Self, v: bool) void {
+        const priv = self.private();
+        priv.is_split = v;
+        self.as(gobject.Object).notifyByPspec(properties.@"is-split".impl.param_spec);
+    }
+
```
jcollie, line 552: With a getter, this would not be necessary.

src/apprt/gtk/class/split_tree.zig
```diff
@@ -808,6 +844,16 @@ pub const SplitTree = extern struct {
             ));
         }
 
+        // Determine if tree has more than one surface
+        const root_handle: Surface.Tree.Node.Handle = .root;
+        const root = tree.nodes[root_handle.idx()];
+        const is_split: bool = switch (root) {
+            .leaf => false,
+            .split => true,
+        };
```
jcollie: This code would become the `getIsSplit` property getter.

src/apprt/gtk/class/split_tree.zig
```diff
@@ -808,6 +844,16 @@ pub const SplitTree = extern struct {
...
+
+        self.setIsSplit(is_split);
+
```
jcollie: This could just emit the property changed signal with the getter.

jcollie (changes requested):
Couple of small changes, then I think that this will be ready to go!

src/apprt/gtk/class/split_tree.zig
jcollie: Verified that this field is not needed and can be deleted.

jcollie (approved):
One final comment about deleting `is_split` from the split tree class. A quick test under valgrind looked clean as well. Excellent work, and thanks for bearing with me!

rhodes-b:
> One final comment about deleting `is_split` from the split tree class. A quick test under valgrind looked clean as well. Excellent work, and thanks for bearing with me!

Thanks for pointers in the right direction! 
I have done a little hacking on the old gtk and this was my first attempt at trying anything in the new design wasn't too bad to pick up

mitchellh (approved):
I didn't run this, but the logic looks good. I trust you and @jcollie otherwise that this works.

rwese:
Tested, fixed the issue for me.

PR #8814, opened 2025-09-21
unicode: delete props.zig and clean up symbols deps too

mitchellh:
Follow up to #8810

Same reasoning.


src/unicode/symbols_table.zig
```diff
@@ -0,0 +1,14 @@
+const lut = @import("lut.zig");
+
+/// The lookup tables for Ghostty.
+pub const table = table: {
+    // This is only available after running main() below as part of the Ghostty
+    // build.zig, but due to Zig's lazy analysis we can still reference it here.
```
jcollie: This comment should probably be fixed up since main() is no longer in this file.

PR #8815, opened 2025-09-21
logging: document GHOSTTY_LOG and make it more flexible

jcollie:


mitchellh (changes requested):
Just coming back to old PRs. I still have concerns about this. 

My main concern is that `log_level = debug` will make all log statements "active" (won't be compiled out with comptime). Logs are already a pretty impactful part of our performance and I write debug logs with the expectation they won't be in the final binary, such as including more-expensive-to-calculate parameters. So I feel really uncomfortable about that change. **Easy fix here would be to just revert the `std_options` stuff for now.**

Beyond that, I think this looks good.

src/build/mdgen/ghostty_5_header.md
```diff
@@ -93,3 +89,34 @@ will fall back to default values for erroneous keys.
...
+`stderr` is available. Depending on the platform and how Ghostty was launched,
+logs sent to `stderr` may be stored by the system and made available for later
+retrieval.
+
+On Linux if Ghostty is launched by the default `systemd` user service, you can use
+`journald` to see Ghostty's logs: `journalctl --user --unit app-com.mitchellh.ghostty.service`.
+
+On macOS logging to the macOS unified log is available and enabled by default.
+--Use the system `log` CLI to view Ghostty's logs: `sudo log stream level debug
+----predicate 'subsystem=="com.mitchellh.ghostty"'`.
```
mitchellh: Typo here.

mitchellh:
Fixed a typo and reverted the default log type, the rest looks good.

mitchellh (approved):

PR #8816, opened 2025-09-21
gtk: restore flatpak-aware resource directory support

alaviss:
This was not ported to gtk-ng before old runtime was removed, breaking shell integration on Flatpak.

This implementation is copied verbatim from old runtime.


pluiedev (approved):
Not sure how that slipped under the cracks

PR #8826, opened 2025-09-21
build: many more lazy dependencies, defer deps add unless needed

mitchellh:
This makes more dependencies lazy. This has a practical effect of reducing the number of dependencies that need to be downloaded when running certain zig build steps.

This is all limited because we're blocked on an upstream Zig issue: https://github.com/ziglang/zig/issues/21525 This prevents us from fully avoiding downloading many dependencies, but at least they're relatively small.

One major improvement here is the usage of `lazyImport` for `zig-wayland` that prevents downloading `zig_wayland` unconditionally on all platforms. On macOS, we don't download this at all anymore.

Another, weirder change is that all our transitive dependencies are now marked lazy (e.g. glslang's upstream source) even if the glslang build always requires it. This was necessary because without this, even if we simply referenced glslang in the root build.zig, it would force the source package to download unconditionally. This no longer happens.

cc @pluiedev Minor improvements here, doesn't change the long term plan, but improves things in the interim.


jcollie (approved):

mitchellh:
Will want to see a full green CI on this one just due to the sprawl of build impacts...

PR #8829, opened 2025-09-22
fix(font): Treat Powerline glyphs as normal characters for constraint width purposes

danielwe:
Powerline glyphs were treated as whitespace, giving the preceding cell a constraint width of 2 and cutting off icons in people's prompts and statuslines. It is however correct to not treat Powerline glyphs like other Nerd Font symbols; they should simply be treated as normal characters, just like their relatives in the block elements unicode block.

This resolves https://discord.com/channels/1005603569187160125/1417236683266592798 (never promoted to an issue, but real and easy to reproduce).

**Tip**
Screenshot 2025-09-21 at 16 57 58

**This PR**
Screenshot 2025-09-21 at 16 58 42

The constraint width logic was untested but contains some quite subtle interactions, so I wrote a suite of tests covering the cases I'm aware of.

While working on this code I also resolved a TODO comment to add all the box drawing/block element type characters to the set of codepoints excluded from the minimum contrast settings.

### DIFF
--- a/src/renderer/cell.zig
+++ b/src/renderer/cell.zig
@@ -352,7 +352,7 @@ fn isLegacyComputing(char: u21) bool {
 // Returns true if the codepoint is a part of the Powerline range.
 fn isPowerline(char: u21) bool {
     return switch (char) {
-        0xE0B0...0xE0C8, 0xE0CA, 0xE0CC...0xE0D2, 0xE0D4 => true,
+        0xE0B0...0xE0D7 => true,
         else => false,
     };
 }
--- a/src/renderer/cell.zig
+++ b/src/renderer/cell.zig
@@ -236,8 +236,8 @@ pub fn isCovering(cp: u21) bool {
 }
 
 /// Returns true of the codepoint is a "symbol-like" character, which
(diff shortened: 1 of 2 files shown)


danielwe:
Fixes #8787

GustavoWidman:
does this include "U+E0A0"? it's in the list of powerline glyphs but i don't see it included in the range you set. the **official** powerline glyph ranges are U+E0A0–U+E0A7 (Powerline Extra Symbols) and U+E0B0–U+E0C8 (Main Powerline glyphs). adding these should fix most issues related to powerline and glyphs being out-of-bounds

danielwe:
The E0Ax range consists of what I’d consider regular icons, like the branch symbol. I think it’s fine to treat these like any other nerd font icon. The ones that require special treatment are the block elements in E0Bx, E0Cx, and E0Dx.

(Note that special treatment here only means that they don’t affect the scaling of the _subsequent_ glyph, unlike other nerd font icons. The point of this PR is to make them _less_ special.)

mitchellh (approved):
Clean. Thank you.

PR #8847, opened 2025-09-22
fix(font): Final font patcher fixes

danielwe:
This is my final set of fixes to the font patcher/icon scaling code. It builds on #8563 and there's not much reason to pay attention here until that one has been reviewed (the unique changes in this PR only touch the two `nerd_font_*` files; the other 8 files in the diff are just #8563). However, I wanted to make sure the full set of changes/fixes I propose are out in the open, such that any substantial edits by maintainers (like in #7953) can take into account the full context.

I think this and the related patches should be considered fixes, not features, so I hope they can be considered for a 1.2.x release.

This PR fixes some bugs in the extraction of scale and alignment rules from the `font_patcher` script. Roughly in order of importance:

* Nerd fonts apply an offset to some codepoint ranges when extracting glyphs from their original font (e.g., Font Awesome) and placing them in a Nerd Font. Rules are specified in terms of the former codepoints, but must be applied to the latter. This offset was previously not taken into account, so rules were applied to the wrong glyphs, and some glyphs that should have rules didn't get any.
* Previously, the rules from every single patch set was included, but the embedded Symbols Only font doesn't contain all of them. Most importantly, there's a legacy patch set that only exists for historical reasons and is never used anymore, which was overwriting some other rules because of overlapping codepoint ranges. Also, the Symbols Only font contains no box drawing characters, so those rules should not be included. With this PR, irrelevant patch sets are filtered out.
* Some patch sets specify overlapping codepoint ranges, though in reality the original fonts don't actually cover the full ranges and the overlaps just imply that they're filling each other's gaps. During font patching, the presence/absence of a glyph at each codepoint in the original font takes care of the ambiguity. Since we don't have that information, we need to hardcode which patch set "wins" for each case (it's not always the latest set in the list). Luckily, there are only two cases.
* Many glyphs belong to scale groups that should be scaled and aligned as a unit. However, in `font_patcher`, the scale group is _not_ used for _horizontal_ alignment, _unless_ the entire scale group has a single advance width (remember, the original symbol fonts are not monospace). This PR implements this rule by only setting `relative_width` and `relative_x` if the group is monospace.

There are some additional tweaks to ensure that each codepoint actually gets the rule it's supposed to when it belongs to multiple scale groups or patch sets, and to avoid setting rules for codepoints that don't exist in the embedded font.


danielwe:
Fixes #8842

Specifically, the second bullet point above is the ticket. The scaling rules for box drawing characters needs to be excluded, since these codepoints are not present in the symbols only embedded font. The rules end up applying to glyphs that don't need them and that they weren't designed for.

Also fixes #8335

mitchellh:
@danielwe Since I merged your base branch, can you update to remove the commit and fix the conflicts? Thank you!

danielwe:
EDIT: This part was moved to #8990.

~~I added one more small change: Stretched glyphs work best if they're scaled and aligned to the grid, since they're typically meant to align across cells (this is the exception to the argument in #8563 that icons should be aligned to the face). This also lets us eliminate the band-aid of horizontal negative padding (overlap), which is another improvement over `font-patcher` that's only possible because we're applying constraints with knowledge of the pixel grid.~~

~~Here are screenshots showing the difference this makes for the progress bar elements from Nerd Fonts. Notice the overlap artifacts (which would be even more pronounced on a lower-DPI display) no longer appearing after this change.~~

Screenshot 2025-10-01 at 09 34 36
Screenshot 2025-10-01 at 09 36 10

danielwe:
Now that we have #8990, the change mentioned in the previous comment is a better fit there, so I moved it over, see https://github.com/ghostty-org/ghostty/pull/8990#issuecomment-3358834158.

I also realized the implementation of ungrouped horizontal alignment could be vastly simplified---all that's needed is to _not_ set `relative_width` and `relative_x`, with no further changes needed to either the codegen or face.zig. As a result, this PR now only touches nerd_font_{codegen,attributes} and nothing else.

PR #8850, opened 2025-09-23
Make PageList/Search concurrency-safe for background search (Part 2 of #189)

dave-fl:
Prepares the terminal core for multi-threaded search by making PageList mutation safe to run alongside a background PageListSearch.  Part 2 of #189.

- Introduce std.Thread.RwLock-backed WriteGuard/ReadGuard on PageList; writers take the re-entrant guard, which drains pending frees and bumps the snapshot epoch exactly once per structural mutation.
- Add per-node refcounts, serials, and a pending-free stack; retainNode/releaseNode keep nodes
alive across concurrent readers and defer frees to the IO thread.
- Prevent stale reuse during pruning by skipping nodes whose refcount is still held and quarantining them until released.
- Teach PageListSearch and SlidingWindow to retain/release nodes, hold the shared read guard only
while sampling, and rescan when the pagelist epoch changes so stale selections are dropped.
- Extend coverage with PageListSearch drops stale nodes after epoch bump and PageListSearch
survives concurrent mutation to assert the new concurrency behavior under prune/rebuild pressure.

### DIFF
--- a/src/benchmark/TerminalSearch.zig
+++ b/src/benchmark/TerminalSearch.zig
@@ -1,0 +1,224 @@
+//! Benchmark core terminal search throughput by scanning a synthetic
+//! scrollback with a configurable match density.
+const TerminalSearch = @This();
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const Benchmark = @import("Benchmark.zig");
+const options = @import("options.zig");
+const terminalpkg = @import("../terminal/main.zig");
+const Screen = terminalpkg.Screen;
+const PageListSearch = @import("../terminal/search.zig").PageListSearch;
+
+const log = std.log.scoped(.@"terminal-search-bench");
+
+opts: Options,
+alloc: Allocator,
+screen: Screen,
+needle: []const u8,
+generated_bytes: usize,
+bytes_scanned: usize = 0,
+iterations: usize = 0,
+
+row_buf: []u8,
+
+pub const Options = struct {
+    /// Active rows for the terminal.
+    rows: u16 = 24,
+
+    /// Active cols for the terminal.
+    cols: u16 = 80,
+
+    /// Optional benchmark duration; if null the benchmark runs exactly once.
+    @"duration-ns": ?u64 = null,
+
+    /// How many pages of scrollback to synthesize. Each page is roughly
+    /// `rows` rows.
+    @"history-pages": usize = 256,
+
+    /// Number of matches per page. Zero means no matches are inserted.
+    @"hits-per-page": usize = 1,
+
+    /// The needle to search for.
+    needle: []const u8 = "ghost",
+
+    /// Seed for deterministic data generation.
+    seed: u64 = 0x41c6_6e57_f00d_beef,
+
+    /// When true, emit throughput information after the benchmark runs.
+    report: bool = false,
+};
+
+pub fn create(alloc: Allocator, opts: Options) !*TerminalSearch {
+    if (opts.needle.len == 0) return error.EmptyNeedle;
+
+    const ptr = try alloc.create(TerminalSearch);
+    errdefer alloc.destroy(ptr);
+
+    var screen = try Screen.init(alloc, opts.cols, opts.rows, opts.@"history-pages" * opts.rows);
+    errdefer screen.deinit();
+
+    const buf = try alloc.alloc(u8, opts.cols + 1);
+    errdefer alloc.free(buf);
+
+    ptr.* = .{
+        .opts = opts,
+        .alloc = alloc,
+        .screen = screen,
+        .needle = opts.needle,
+        .generated_bytes = 0,
+        .row_buf = buf,
+    };
+
+    try ptr.populate();
+
+    return ptr;
+}
+
+pub fn destroy(self: *TerminalSearch, alloc: Allocator) void {
+    alloc.free(self.row_buf);
+    self.screen.deinit();
+    alloc.destroy(self);
+}
+
+pub fn benchmark(self: *TerminalSearch) Benchmark {
+    return .init(self, .{ .stepFn = step, .setupFn = setup });
+}
+
+fn setup(ptr: *anyopaque) Benchmark.Error!void {
+    const self: *TerminalSearch = @ptrCast(@alignCast(ptr));
+    self.bytes_scanned = 0;
+    self.iterations = 0;
+}
+
+fn step(ptr: *anyopaque) Benchmark.Error!void {
+    const self: *TerminalSearch = @ptrCast(@alignCast(ptr));
+
+    var search = PageListSearch.init(self.alloc, &self.screen.pages, self.needle) catch |err| {
+        log.err("init failed err={}", .{err});
+        return error.BenchmarkFailed;
+    };
+    defer search.deinit(self.alloc);
+
+    while (true) {
+        const result = search.next(self.alloc) catch |err| {
+            log.err("next failed err={}", .{err});
+            return error.BenchmarkFailed;
+        };
+        if (result == null) break;
+    }
+
+    self.bytes_scanned += self.generated_bytes;
+    self.iterations += 1;
+}
+
+fn populate(self: *TerminalSearch) !void {
+    var prng = std.Random.DefaultPrng.init(self.opts.seed);
+    const rand = prng.random();
+
+    const rows_per_page: usize = self.opts.rows;
+    const total_rows = self.opts.@"history-pages" * rows_per_page;
+
+    const cols = self.opts.cols;
+
+    const hits_per_page = self.opts.@"hits-per-page";
+    const needle = self.needle;
+
+    var rows_written: usize = 0;
+    var hits_in_current_page: usize = 0;
+    const hit_spacing = if (hits_per_page == 0)
+        0
+    else
+        @max(1, (rows_per_page + hits_per_page - 1) / hits_per_page);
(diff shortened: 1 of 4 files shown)


mitchellh (commented):
Oh awesome. This is definitely heading in the right direction and the code quality looks... normal at first glance (that's a good thing). 

One thing that I want to be really careful about as we go down this path is testing the no-search throughput to ensure we're not regressing that. To that end, do you have benchmark numbers of the terminal stream before and after this branch? We should be able to use our existing `terminal-stream` benchmark. 

It'd be good to test that with plain ASCII, Japanese or some other complex Unicode language, and some mix of escape sequences.

dave-fl:
Here are the numbers - Note that 5a29dd3ef5e2fdf47963ded980251f314a0c6b95 changed the build defaults for the benchmarks, I can test on `main` with `-Doptimize=ReleaseFast` if you would like.

```
zig-out/bin/ghostty-gen +ascii | head -c 67108864 > /tmp/ascii_64m.stream
zig-out/bin/ghostty-gen +osc --p-valid=0.7 | head -c 67108864 > /tmp/osc_64m.stream
zig-out/bin/ghostty-gen +utf8 | head -c 67108864 > /tmp/utf8_64m.stream

git log -1 --oneline
1a6af1704 (HEAD) gtk: restore flatpak-aware resource directory support (#8816)

hyperfine -w 2 'zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/ascii_64m.stream --terminal-rows=24 --terminal-cols=80'
Benchmark 1: zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/ascii_64m.stream --terminal-rows=24 --terminal-cols=80
  Time (mean ± σ):     400.9 ms ±   3.4 ms    [User: 385.1 ms, System: 14.6 ms]
  Range (min … max):   396.0 ms … 407.5 ms    10 runs

hyperfine -w 2 'zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/osc_64m.stream --terminal-rows=24 --terminal-cols=80'
Benchmark 1: zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/osc_64m.stream --terminal-rows=24 --terminal-cols=80
  Time (mean ± σ):      1.093 s ±  0.013 s    [User: 1.073 s, System: 0.018 s]
  Range (min … max):    1.080 s …  1.124 s    10 runs

hyperfine -w 2 'zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/utf8_64m.stream --terminal-rows=24 --terminal-cols=80'
Benchmark 1: zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/utf8_64m.stream --terminal-rows=24 --terminal-cols=80
  Time (mean ± σ):     607.8 ms ±   2.6 ms    [User: 541.5 ms, System: 62.1 ms]
  Range (min … max):   604.1 ms … 611.6 ms    10 runs

git log -1 --oneline
bd32b1ce1 (HEAD -> feature/search-concurrency, origin/feature/search-concurrency) PageList: write-guard RAII and epoch-safe search window

hyperfine -w 2 'zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/ascii_64m.stream --terminal-rows=24 --terminal-cols=80'
Benchmark 1: zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/ascii_64m.stream --terminal-rows=24 --terminal-cols=80
  Time (mean ± σ):     401.2 ms ±   1.7 ms    [User: 386.1 ms, System: 14.0 ms]
  Range (min … max):   398.3 ms … 403.9 ms    10 runs

hyperfine -w 2 'zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/osc_64m.stream --terminal-rows=24 --terminal-cols=80'
Benchmark 1: zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/osc_64m.stream --terminal-rows=24 --terminal-cols=80
  Time (mean ± σ):      1.090 s ±  0.006 s    [User: 1.072 s, System: 0.016 s]
  Range (min … max):    1.083 s …  1.102 s    10 runs

hyperfine -w 2 'zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/utf8_64m.stream --terminal-rows=24 --terminal-cols=80'
Benchmark 1: zig-out/bin/ghostty-bench +terminal-stream --data=/tmp/utf8_64m.stream --terminal-rows=24 --terminal-cols=80
  Time (mean ± σ):     615.5 ms ±   3.3 ms    [User: 547.6 ms, System: 63.4 ms]
  Range (min … max):   613.1 ms … 624.2 ms    10 runs
```

dave-fl:
@mitchellh Let me know if there is anything I can do - I got part 3/4 cooking.

dave-fl:
~~PR not ready yet — I'm investigating a possible race on tracked pins under load (clone iterating pins while search mutates).~~

dave-fl:
Added `trackedPinsGuard`so any caller that mutates `tracked_pins` (renderer clone or app-side search) runs through the same lock. Without it we can still hit the original race: the search controller tracks/untracks pins while the renderer copies the set, and we corrupt the map. This guard eliminates that interleaving; pins are tracked long enough to keep matches alive and the renderer safe.

Numbers did not change on benchmark results shown above.

dave-fl:
- I split PageList into safe reader/writer paths so the search worker can scan while the IO thread mutates. Every structural change bumps a snapshot_epoch; the worker snapshots it at start, and each batch carries that epoch. If the PageList epoch changes, we treat incoming batches as stale and restart cleanly.
- Matches stream as pointer‑free anchors: {node_serial, y/x, len, snapshot_epoch}. The serial is a monotonic ID per page node, so we can order reliably (forward = strictly after, backward = nearest strictly before) even across restarts. The epoch tells us if an anchor is still valid; the serial tells us where it belongs.
- For highlighting, we only materialize the active match as pins (under a read lock), keeping retarget work tiny. Renderer clones and writer updates iterate pins under a small pin lock.  Screen‑level guards keep the cursor in sync during scroll/grow.
- Live demo: typing streams results smoothly; forward/backward restarts land where you expect; and it stays stable under constant output.

https://github.com/user-attachments/assets/5062dae1-2da8-4e72-a4e5-d8bfa4b57a07

mitchellh:
@dave-fl FYI I'm on vacation currently but plan on reviewing this in earnest perhaps next week. Sorry for the delay, appreciate your efforts here. 😄

kingsleyh:
any ETA on search functionality? I'm very keen to be able to search in the terminal

mitchellh:
Just as an update here, I've been working through the changes here in a slightly different way and experimenting. I'm actively working on this and will give any credit to @dave-fl once I get somewhere with it (or just accept this PR as-is). Thanks!

dave-fl:
Hi @mitchellh welcome back.  While this waited on review, I prototyped an "anchor" variant of search. Here's the breakdown:


---

## Context

- **Original path**: Worker returns live Selections (pins) and keeps nodes alive via `retain/release`. Page reuse in `grow()` can be skipped when a node is retained by a reader.

- **Anchor path**: Worker returns pointerless anchors `{ node_serial, y, x, len }`. Sliding window stores only `node_serial` (no `*Node` pointers, no `retainNode()`). PageList drops refcounts/pending‑free; prune always reuses the head page.

## What changed

**src/terminal/search.zig:**
- `PageListSearch.next` returns `Message.Match` (anchor) instead of `Selection`.
- `SlidingWindow.Meta` stores `{ node_serial, cell_map }` instead of `{ list, node, cell_map }`.
- No `retainNode()` call in `append()`.
- "Find next node" uses `findNodeBySerial(...).next` instead of `meta.node.next`.

**src/terminal/PageList.zig:**
- Remove per‑node refcounts and pending‑free queue (~70 lines); `processPendingFree()` becomes no‑op.
- `grow()` always reuses pruned head page (no "skip reuse when retained" path).

## Trade-offs

**Anchors:**
- No cross-thread node ownership, no atomic refcounting, no deferred free queue.
- Predictable memory: always reuses pruned pages.
- Must re-lookup nodes from serial when materializing results (epoch validation ensures correctness).

**Refcounts:**
- Can hold `*Node` pointers across lock boundaries (more flexible).
- Nodes stay alive while retained (prune may skip reuse or allocate).
- Cost: atomic ops + deferred free queue + reuse gating logic.

---

Happy to stick with refcounts if we want that flexibility. Just wanted you aware the anchor approach exists and what it trades. No recommendation implied.

mitchellh:
Thanks thats an interesting idea. My immediate reaction is that I think we'll want the refcounting and such on Nodes _anyways_ because concurrent PageList will be used for more than just search. Getting this core right will be key to also enabling stuff like background archiving of scroll back to disk (or, the reverse, restoring it). _I think_. That's at least how I've always thought about stuff like this. 

Btw are you in Discord? Would be useful to run stuff by you but if not no big deal.

dave-fl:
Hey @mitchellh. Makes sense - good to understand your vision on this. I'm going on vacation this week but will ping you when I get back on Discord. Happy to discuss further then!

mitchellh:
See #9585 for a different direction. If the "big lock" approach works fine for performance then we can probably yeet all this PageList complexity completely. If it doesn't work fine, then I think the components I laid out there are still going to be necessary and can be augmented to be fine-grained-locking-aware.

mitchellh:
Going to close this. We've shipped search [on macOS] and the internals are shared. We avoided the concurrency. For now. If this ends up being useful I may pluck certain things from it. Thank you so much @dave-fl for looking into this, sorry we didn't use it.

PR #8853, opened 2025-09-23
GTK: Fix split-divider-color

rhodes-b:
The `loadRuntimeCss416` overrode a color option for the split divider color
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/application.zig#L959-L965

I moved the user config options until the other runtime css is loaded so they will always take priority

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -809,6 +809,9 @@ pub const Application = extern struct {
 
         const writer = buf.writer(alloc);
 
+        try loadRuntimeCss414(config, &writer);
+        try loadRuntimeCss416(config, &writer);
+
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/class/application.zig
```diff
@@ -809,6 +809,9 @@ pub const Application = extern struct {
 
         const writer = buf.writer(alloc);
 
+        try loadRuntimeCss414(config, &writer);
```
mitchellh, line 813: We should add a comment above this about the importance of its position, because its not immediately clear that it is.

mitchellh (changes requested):
Looks good, thanks, just one little ask.

mitchellh (approved):

PR #8858, opened 2025-09-23
ci: Add test for building with i18n on and off

### LINKED ISSUE #8360: Add a `-Di18n=false` build to CI
We aren't currently testing this build flag, so unsurprisingly it didn't even build: https://github.com/ghostty-org/ghostty/pull/8359 

We should add this to CI. We don't need to exhaustively test every permutation of build flag. Let's just add a simple single case.

AnthonyZhOon:
Closes #8360

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,7 @@ jobs:
       - test-simd
       - test-gtk
       - test-sentry-linux
+      - test-i18n
       - test-macos
       - pinact
       - prettier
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -701,6 +702,41 @@ jobs:
       - name: test
         run: nix develop -c zig build test --system ${{ steps.deps.outputs.deps }}
 
+  test-i18n:
+    strategy:
+      fail-fast: false
+      matrix:
+        i18n: ["true", "false"]
```
jcollie, line 709: I don't think that we need a separate test for `-Di18n=true` as that's the default and will be tested thoroughly elsewhere. Perhaps it would be best to add the matrix to the main `test` so that `-Di18n=false` gets tested early.

jcollie (changes requested):

.github/workflows/test.yml
mitchellh: I think he's just following how I do this for simd. For simd I chose to do both in case there's ever a scenario that we default off for our main tests and forget to re-enable it. Given these are happening in parallel on small profiles and are very much not the blocker on our CI, I'm fine with doing both.

PR #8871, opened 2025-09-23
chore: pin zig 0.14 in build.zig.zon

GauBen:
Hi!

I'm a full Zig noob but [Mitchell's recent post](https://mitchellh.com/writing/libghostty-is-coming) made me want to clone the repo and take a look at the tooling.

My first attempt at running examples though VSCode failed because the latest version of Zig is 0.15.1, but Ghostty requires Zig 0.14*. When configuring the extension to use a compatible version if Zig, it suggested pinning the version in a .zigversion file. I'm not familiar with the pattern, but if it can help someone else's onboarding, I figured I'd open a PR to suggest the change.

Cheers

*edit: I had a hard time figuring that out


mitchellh:
I'm not familiar with what tooling uses this. Our preferred method is via the official `build.zig.zon` file which always contains the version of Zig we require. It is specified as a "minimum" but given Zig breaks every minor release, I always recommend its considered an _exact_.

What tooling uses this?

GauBen:
At least the [Zig VS Code extension](https://github.com/ziglang/vscode-zig/blob/88ba172d62b5716ebf9a89652af27046e0febd54/src/zigSetup.ts#L226-L230). It also supports `minimum_zig_version` in `build.zig.zon`, but [none is set](https://github.com/ghostty-org/ghostty/blob/main/build.zig.zon), and that may install Zig 0.15 if the minimum is set to 0.14 (maybe?)

Edit: regarding build.zig.zon, the extension uses [caret semver ranges](https://github.com/ziglang/vscode-zig/blob/88ba172d62b5716ebf9a89652af27046e0febd54/src/zigSetup.ts#L68), so setting 0.14 should be safe

Edit': using build.zig.zon looks like this b0e85d900ec0d383d548bf3962e65ba1851c5260 I guess

mitchellh:
Did using the `build.zig.zon` work for you?

GauBen:
Yes! It automatically downloads the latest version of the 0.14.x branch

image

I'll update minimum_zig_version to be 0.14.1 so that all [build.zig.zon](https://github.com/ghostty-org/ghostty/blob/b0e85d900ec0d383d548bf3962e65ba1851c5260/example/zig-vt/build.zig.zon#L5) files use the same version

build.zig.zon
```diff
@@ -3,6 +3,7 @@
     .version = "1.2.1",
     .paths = .{""},
     .fingerprint = 0x64407a2a0b4147e5,
+    .minimum_zig_version = "0.14.0",
```
GauBen:
```suggestion
    .minimum_zig_version = "0.14.1",
```

Use the same version as [example/zig-vt/build.zig.zon](https://github.com/ghostty-org/ghostty/blob/b0e85d900ec0d383d548bf3962e65ba1851c5260/example/zig-vt/build.zig.zon#L5)

PR #8873, opened 2025-09-24
config: smarter parsing in autoParseStruct

jcollie:
Fixes #8849

Previously, the `parseAutoStruct` function that was used to parse generic structs for the config simply split the input value on commas without taking into account quoting or escapes. This led to problems because it was impossible to include a comma in the value of config entries that were parsed by `parseAutoStruct`. This is particularly problematic because `ghostty +show-config --default` would produce output like the following:

```
command-palette-entry = title:Focus Split: Next,description:Focus the next split, if any.,action:goto_split:next
```

Because the `description` contains a comma, Ghostty is unable to parse this correctly. The value would be split into four parts:

```
title:Focus Split: Next
description:Focus the next split
 if any.
action:goto_split:next
```

Instead of three parts:

```
title:Focus Split: Next
description:Focus the next split, if any.
action:goto_split:next
```

Because `parseAutoStruct` simply looked for commas to split on, no amount of quoting or escaping would allow that to be parsed correctly.

This is fixed by (1) introducing a parser that will split the input to `parseAutoStruct` into fields while taking into account quotes and escaping. And (2) changing the `ghostty +show-config` output to put the values in `command-palette-entry` into quotes so that Ghostty can parse it's own output.

`parseAutoStruct` will also now parse double quoted values as a Zig string literal. This makes it easier to embed control codes, whitespace, and commas in values.

### DIFF
--- a/src/cli/Splitter.zig
+++ b/src/cli/Splitter.zig
@@ -1,0 +1,424 @@
+//! Iterator to split a string into fields by commas, taking into account
+//! quotes and escapes.
+//!
+//! Supports the same escapes as in Zig literal strings.
+//!
+//! Quotes must begin and end with a double quote (`"`). It is an error to not
+//! end a quote that was begun. To include a double quote inside a quote (or to
+//! not have a double quote start a quoted section) escape it with a backslash.
+//!
+//! Single quotes (`'`) are not special, they do not begin a quoted block.
+//!
+//! Zig multiline string literals are NOT supported.
+//!
+//! Quotes and escapes are not stripped or decoded, that must be handled as a
+//! separate step!
+const Splitter = @This();
+
+pub const Error = error{
+    UnclosedQuote,
+    UnfinishedEscape,
+    IllegalEscape,
+};
+
+/// the string that we are splitting
+str: []const u8,
+/// how much of the string has been consumed so far
+index: usize,
+
+/// initialize a splitter with the given string
+pub fn init(str: []const u8) Splitter {
+    return .{
+        .str = str,
+        .index = 0,
+    };
+}
+
+/// return the next field, null if no more fields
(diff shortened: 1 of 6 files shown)


src/cli/Splitter.zig
```diff
@@ -0,0 +1,424 @@
...
+//! end a quote that was begun. To include a double quote inside a quote (or to
+//! not have a double quote start a quoted section) escape it with a backslash.
+//!
+//! Single quotes (`'`) are not special, they do not begin a quoted block.
+//!
+//! Zig multiline string literals are NOT supported.
+//!
+//! Quotes and escapes are not stripped or decoded, that must be handled as a
+//! separate step!
+const Splitter = @This();
```
mitchellh: I'd bikeshed this and name it something a bit more specific, `CommaSplitter` or something. "Splitter" feels like a real thing that could exist for configs that does anything.

src/input/command.zig
```diff
@@ -472,13 +472,18 @@ fn actionCommands(action: Action.Key) []const Command {
             .description = "Quit the application.",
         }},
 
+        .text => comptime &.{.{
+            .action = .{ .text = "👻" },
+            .title = "Ghostty",
+            .description = "Put a little Ghostty in your terminal.",
```
mitchellh, line 478: haha, okay.

mitchellh (changes requested):
One small request, just naming.

I'm a bit hesitant that Splitter has to be aware of valid escapes for Zig strings. I'll be honest when I thought of this I was thinking we should just take advantage of the Zig stdlib AST/Parser but this is undoubtedly going to be faster and Zig strings probably won't change too much. And it's well tested. So okay!

src/cli/Splitter.zig
jcollie: Done.

PR #8885, opened 2025-09-24
Add `.ghostty` extension to `config`

### LINKED ISSUE #8689: Allow the `.ghostty` extension to be used for the default configuration file
Discussion: #8438 

Motivations are well discussed in the linked discussion. Bikeshedding is complete.

Allow the `.ghostty` extension to be used for the default configuration file. We only need to add this to the default because non-defaults are explicitly specified with full filenames and we're not going to automatically add extensions for those.

Since we have some legacy to deal with, we need to come up with some arbitrary behavior when _both_ exist. In the case _both_ exist, the `.ghostty` version will take priority over the same-named non-extension version. We should probably log a warning, too.

CoderJoshDK:
Resolves #8689

For various reason, ghostty wants to have a unique file extension for the config files. The name was settled on `config.ghostty`. This will help with tooling. See #8438 (original discussion) for more details. 

This PR introduces the preferred default of `.ghostty` while still supporting the previous `config` file. If both files exist, a warning log is sent. 

The docs / website will need to be updated to reflect this change. 

> [!NOTE]
> Only tested on macOS 26.0.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2071,7 +2071,7 @@ keybind: Keybinds = .{},
 
 /// When this is true, the default configuration file paths will be loaded.
 /// The default configuration file paths are currently only the XDG
-/// config path ($XDG_CONFIG_HOME/ghostty/config).
+/// config path ($XDG_CONFIG_HOME/ghostty/config.ghostty).
 ///
 /// If this is false, the default configuration paths will not be loaded.
 /// This is targeted directly at using Ghostty from the CLI in a way
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3471,40 +3471,40 @@ fn writeConfigTemplate(path: []const u8) !void {
 pub fn loadDefaultFiles(self: *Config, alloc: Allocator) !void {
     // Load XDG first
     const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
-    const xdg_path = try defaultXdgPath(alloc);
     defer alloc.free(legacy_xdg_path);
+    const xdg_path = try defaultXdgPath(alloc);
     defer alloc.free(xdg_path);
-    var xdg_loaded = false;
-    {
+    const xdg_loaded: bool = xdg_loaded: {
         const legacy_xdg_action = self.loadOptionalFile(alloc, legacy_xdg_path);
         const xdg_action = self.loadOptionalFile(alloc, xdg_path);
         if (xdg_action != .not_found and legacy_xdg_action != .not_found) {
             log.warn("both config files `{s}` and `{s}` exist.", .{ legacy_xdg_path, xdg_path });
             log.warn("loading them both in that order", .{});
-            xdg_loaded = true;
-        } else if (xdg_action != .not_found or legacy_xdg_action != .not_found) {
-            xdg_loaded = true;
+            break :xdg_loaded true;
         }
-    }
+
+        break :xdg_loaded xdg_action != .not_found or
+            legacy_xdg_action != .not_found;
+    };
 
     // On macOS load the app support directory as well
     if (comptime builtin.os.tag == .macos) {
         const legacy_app_support_path = try legacyDefaultAppSupportPath(alloc);
+        defer alloc.free(legacy_app_support_path);
         const app_support_path = try preferredAppSupportPath(alloc);
         defer alloc.free(app_support_path);
-        defer alloc.free(legacy_app_support_path);
-        var app_support_loaded = false;
-        {
+        const app_support_loaded: bool = loaded: {
             const legacy_app_support_action = self.loadOptionalFile(alloc, legacy_app_support_path);
             const app_support_action = self.loadOptionalFile(alloc, app_support_path);
             if (app_support_action != .not_found and legacy_app_support_action != .not_found) {
                 log.warn("both config files `{s}` and `{s}` exist.", .{ legacy_app_support_path, app_support_path });
                 log.warn("loading them both in that order", .{});
-                app_support_loaded = true;
-            } else if (app_support_action != .not_found or legacy_app_support_action != .not_found) {
-                app_support_loaded = true;
+                break :loaded true;
             }
-        }
+
+            break :loaded app_support_action != .not_found or
+                legacy_app_support_action != .not_found;
+        };
 
         // If both files are not found, then we create a template file.
         // For macOS, we only create the template file in the app support
@@ -3543,18 +3543,25 @@ pub fn legacyDefaultXdgPath(alloc: Allocator) ![]const u8 {
 /// Preferred default path for the XDG home configuration file.
 /// Returned value must be freed by the caller.
 fn preferredXdgPath(alloc: Allocator) ![]const u8 {
+    // If the XDG path exists, use that.
     const xdg_path = try defaultXdgPath(alloc);
-    const xdg_file = openFile(xdg_path) catch {
-        const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
-        const legacy_xdg_file = openFile(legacy_xdg_path) catch {
-            alloc.free(legacy_xdg_path);
-            return xdg_path;
-        };
-        legacy_xdg_file.close();
+    if (openFile(xdg_path)) |f| {
+        f.close();
+        return xdg_path;
+    } else |_| {}
+
+    // Try the legacy path
+    errdefer alloc.free(xdg_path);
+    const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
+    if (openFile(legacy_xdg_path)) |f| {
+        f.close();
         alloc.free(xdg_path);
         return legacy_xdg_path;
-    };
-    xdg_file.close();
+    } else |_| {}
+
+    // Legacy path and XDG path both don't exist. Return the
+    // new one.
+    alloc.free(legacy_xdg_path);
     return xdg_path;
 }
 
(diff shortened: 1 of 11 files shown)


src/config/Config.zig
```diff
@@ -3397,19 +3397,88 @@ pub fn loadDefaultFiles(self: *Config, alloc: Allocator) !void {
...
+        if (!log_on_collision) {
+            return xdg_path;
+        }
+        const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
+        defer alloc.free(legacy_xdg_path);
+        if (openFile(legacy_xdg_path)) |legacy_f| {
+            legacy_f.close();
+            log.warn("Both `{s}` and `{s}` exist.", .{ xdg_path, legacy_xdg_path });
+            log.warn("Only loading `{s}`.", .{xdg_path});
+        } else |_| {}
```
jcollie: I can't say as I like the empty elses - looks very awkward. Can you try using `const f = openFile(..) catch {...}` syntax?

src/config/Config.zig
```diff
@@ -3397,19 +3397,88 @@ pub fn loadDefaultFiles(self: *Config, alloc: Allocator) !void {
...
 
+/// Preferred default path for the XDG home configuration file.
+/// Returned value must be freed by the caller.
+fn preferredXdgPath(alloc: Allocator, log_on_collision: bool) ![]const u8 {
+    const xdg_path = try defaultXdgPath(alloc);
+    if (openFile(xdg_path)) |f| {
+        f.close();
+        if (!log_on_collision) {
+            return xdg_path;
+        }
```
jcollie: This can be a single line without the `{}`

jcollie (changes requested):
Looks, good but the empty `else` seems awkward. I think using `catch` might make things flow better.

src/config/Config.zig
CoderJoshDK: I might be missing something, but I am not sure how that would make the flow any better? Actually, I am not sure how you would make that change? Since the logic is contained in the positive case and not the error (hence the empty `else`.

Idk if you have a better idea in mind, but 
```diff
diff --git a/src/config/Config.zig b/src/config/Config.zig
index 1f2117ade..273305102 100644
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3461,11 +3461,12 @@ fn preferredAppSupportPath(alloc: Allocator, log_on_collision: bool) ![]const u8

         const legacy_app_support_path = try legacyDefaultAppSupportPath(alloc);
         defer alloc.free(legacy_app_support_path);
-        if (openFile(legacy_app_support_path)) |legacy_f| {
+        _ = {
+            const legacy_f = openFile(legacy_app_support_path);
             legacy_f.close();
             log.warn("Both `{s}` and `{s}` exist.", .{ app_support_path, legacy_app_support_path });
             log.warn("Only loading `{s}`.", .{app_support_path});
-        } else |_| {}
+        } catch 0;
         return app_support_path;
     } else |_| {}
     const legacy_app_support_path = try legacyDefaultAppSupportPath(alloc);
```
Looks the same to me. I assume this is not what you mean. Just unsure how else to achieve this?

jcollie: This is probably the way I'd write it:
```
diff --git a/src/config/Config.zig b/src/config/Config.zig
index 1f2117ade..1abf11933 100644
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3417,25 +3417,31 @@ pub fn legacyDefaultXdgPath(alloc: Allocator) ![]const u8 {
 /// Returned value must be freed by the caller.
 fn preferredXdgPath(alloc: Allocator, log_on_collision: bool) ![]const u8 {
     const xdg_path = try defaultXdgPath(alloc);
-    if (openFile(xdg_path)) |f| {
+
+    {
+        const f = openFile(xdg_path) catch {
+            alloc.free(xdg_path);
+
+            const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
+            errdefer alloc.free(legacy_xdg_path);
+
+            const f = try openFile(legacy_xdg_path);
+            f.close();
+
+            return legacy_xdg_path;
+        };
         f.close();
-        if (!log_on_collision) return xdg_path;
+    }

-        const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
+    if (log_on_collision) log_on_collision: {
+        const legacy_xdg_path = legacyDefaultXdgPath(alloc) catch break :log_on_collision;
         defer alloc.free(legacy_xdg_path);
-        if (openFile(legacy_xdg_path)) |legacy_f| {
-            legacy_f.close();
-            log.warn("Both `{s}` and `{s}` exist.", .{ xdg_path, legacy_xdg_path });
-            log.warn("Only loading `{s}`.", .{xdg_path});
-        } else |_| {}
-        return xdg_path;
-    } else |_| {}
-    const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
-    if (openFile(legacy_xdg_path)) |f| {
+        const f = openFile(legacy_xdg_path) catch break :log_on_collision;
         f.close();
-        alloc.free(xdg_path);
-        return legacy_xdg_path;
-    } else |_| {}
+        log.warn("Both `{s}` and `{s}` exist.", .{ xdg_path, legacy_xdg_path });
+        log.warn("Only loading `{s}`.", .{xdg_path});
+    }
+
     return xdg_path;
 }
```

CoderJoshDK: This makes sense. Slight modification I had to make, we don't want to return an error if neither file exists, we instead want to return `xdg_path`. Meaning I did have to use an `if else`, but at least this time it isn't blank. Will have that commit here in a bit. 

```zig
fn preferredXdgPath(alloc: Allocator, log_on_collision: bool) ![]const u8 {
    const xdg_path = try defaultXdgPath(alloc);
    {
        const f = openFile(xdg_path) catch {
            const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
            if (openFile(legacy_xdg_path)) |f| {
                f.close();
                alloc.free(xdg_path);
                return legacy_xdg_path;
            } else |_| {
                alloc.free(legacy_xdg_path);
                return xdg_path;
            }
        };
        f.close();
    }
    if (log_on_collision) log_on_collision: {
        const legacy_xdg_path = legacyDefaultXdgPath(alloc) catch break :log_on_collision;
        defer alloc.free(legacy_xdg_path);
        const f = openFile(legacy_xdg_path) catch break :log_on_collision;
        f.close();
        log.warn("Both `{s}` and `{s}` exist.", .{ xdg_path, legacy_xdg_path });
        log.warn("Only loading `{s}`.", .{xdg_path});
    }

    return xdg_path;
}
```

src/config/Config.zig
```diff
@@ -3397,19 +3397,94 @@ pub fn loadDefaultFiles(self: *Config, alloc: Allocator) !void {
...
+                return legacy_app_support_path;
+            } else |_| {
+                alloc.free(legacy_app_support_path);
+                return app_support_path;
+            }
+        };
+        f.close();
+    }
+    if (log_on_collision) log_on_collision: {
+        const legacy_xdg_path = legacyDefaultAppSupportPath(alloc) catch break :log_on_collision;
```
jcollie: nitpick: `legacy_app_support_path`

jcollie (approved):
One nitpick, otherwise looks good.

src/config/edit.zig
```diff
@@ -4,6 +4,7 @@ const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
 const ArenaAllocator = std.heap.ArenaAllocator;
 const internal_os = @import("../os/main.zig");
+const Config = @import("./config.zig");
```
jcollie: Whoops, should be `Config.zig`

jcollie (changes requested):
CI error needs fixing.

src/config/edit.zig
CoderJoshDK: Somehow the build system and LSP is smart enough to figure it out. But yea, that is wrong! Resolved

jcollie (approved):

mitchellh (changes requested):
This looks mostly good! Thank you!

The only issue is that in the original issue I suggested loading BOTH if there is a collision and having `.ghostty` overwrite the non-extension one. For macOS that'd result in an order of: no-ext, ext, no-ext, ext for the XDG and App Support dirs.

I don't see any real downside to this (one sys call cost unfortunately per file) and I think it'd avoid confusion of why configs aren't being loaded based on a single log file.

CoderJoshDK:
Ah, makes sense. Sorry for misunderstanding. Should be simple enough to fix.

CoderJoshDK:
~Converted to draft because I noticed strange behavior in the logs of `+show-config`. Going to trace that down first before I open this back up.~ I was running the wrong binary. Everything is working as expected.

anund:
Likely missing a few places tooling is assuming the config in specific path. For example: https://github.com/ghostty-org/ghostty/blob/main/src/extra/vim.zig#L13 would no longer 'attach' to this new file name.

CoderJoshDK:
> Likely missing a few places tooling is assuming the config in specific path. For example: [`main`/src/extra/vim.zig#L13](https://github.com/ghostty-org/ghostty/blob/main/src/extra/vim.zig?rgh-link-date=2025-10-08T00%3A19%3A55Z#L13) would no longer 'attach' to this new file name.

Thanks. This was handled at some point, but lost during some other change to that file. The other tools seem to take the config path as some variable. So no change should be required for those. But I am not familiar enough with every single tool that extras handles to say with 100% certainty.

src/config/Config.zig
```diff
@@ -3464,31 +3464,60 @@ fn writeConfigTemplate(path: []const u8) !void {
...
+/// On macOS, `$HOME/Library/Application Support/$CFBundleIdentifier/`
 /// is also loaded.
+///
+/// The legacy `config` file (without extension) is first loaded,
+/// then `config.ghostty`.
 pub fn loadDefaultFiles(self: *Config, alloc: Allocator) !void {
     // Load XDG first
+    const legacy_xdg_path = try legacyDefaultXdgPath(alloc);
     const xdg_path = try defaultXdgPath(alloc);
+    defer alloc.free(legacy_xdg_path);
```
mitchellh, line 3475: This alloc should go above the `xdg_path`. If that fails then we don't free this and we leak.

mitchellh (approved):
Thanks! Finally reviewed this. I added a couple fixup commits. There were possible memory leaks in error scenarios and some style changes. Take a look at the commits if you're interested.

raymond-w-ko:
some of the conditionals are not correct, at least on Linux (WSL2 Ubuntu 24.04). as you can see in the log below, it finds the config below and then tries to create it from a non-existent template, leading to the file becoming zapped and blank, and needs to be restored from version control/backups.

i think line 3512 and 3518 in Config.zig need negation / the boolean is flipped. The comment above also doesn't make sense, talking about files not found, but the var names `*_loaded`

```log
info: ghostty version=1.2.1-main+a2b516868
info: runtime=.gtk
info: dependency harfbuzz=11.0.0
info(os_locale): setlocale from env result=C.UTF-8
info(gtk): GTK version build=4.14.5 runtime=4.14.5
info(gtk): libadwaita version build=1.5.0 runtime=1.5.0
info: reading configuration file path=/home/rko/.config/ghostty/config.ghostty
info(config): creating template config file: path=/home/rko/.config/ghostty/config.ghostty
info(config): loading config-file path=/home/rko/dot/config/ghostty/config.local.ghostty
info(config): default shell src=passwd value=/usr/bin/fish
info(config): default working directory src=passwd value=/home/rko
```

raymond-w-ko:
something like this works for me:

```patch
diff --git a/src/config/Config.zig b/src/config/Config.zig
index aabefcd8f..da7fced08 100644
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3509,13 +3509,13 @@ pub fn loadDefaultFiles(self: *Config, alloc: Allocator) !void {

         // If both files are not found, then we create a template file.
         // For macOS, we only create the template file in the app support
-        if (app_support_loaded and xdg_loaded) {
+        if (!app_support_loaded and !xdg_loaded) {
             writeConfigTemplate(app_support_path) catch |err| {
                 log.warn("error creating template config file err={}", .{err});
             };
         }
     } else {
-        if (xdg_loaded) {
+        if (!xdg_loaded) {
             writeConfigTemplate(xdg_path) catch |err| {
                 log.warn("error creating template config file err={}", .{err});
             };
```

mitchellh:
Thanks @raymond-w-ko good catch. That's fixed now.

PR #8886, opened 2025-09-24
fix: alloc free off by one

CoderJoshDK:
Fix provided by @jcollie 

The swift `open_config` action was triggering an allocation error `error(gpa): Allocation size 41 bytes does not match free size 40.`. 

> A string that was created as a `[:0]const u8` was cast to `[]const u8` and then freed. The sentinel is the off-by-one. 

@jcollie 

For full context, see https://discord.com/channels/1005603569187160125/1420367156071239820

Co-authored-by: Jeffrey C. Ollie <jcollie@dmacc.edu>

### DIFF
--- a/src/main_c.zig
+++ b/src/main_c.zig
@@ -63,20 +63,41 @@ const Info = extern struct {
 pub const String = extern struct {
     ptr: ?[*]const u8,
     len: usize,
-    cap: usize,
+    sentinel: bool,
 
     pub const empty: String = .{
         .ptr = null,
         .len = 0,
-        .cap = 0,
+        .sentinel = false,
     };
 
-    pub fn fromSlice(slice: []const u8, cap: usize) String {
+    pub fn fromSlice(slice: anytype) String {
         return .{
             .ptr = slice.ptr,
             .len = slice.len,
-            .cap = cap,
+            .sentinel = sentinel: {
+                const info = @typeInfo(@TypeOf(slice));
+                switch (info) {
+                    .pointer => |p| {
+                        if (p.size != .slice) @compileError("only slices supported");
+                        if (p.child != u8) @compileError("only u8 slices supported");
+                        const sentinel_ = p.sentinel();
+                        if (sentinel_) |sentinel| if (sentinel != 0) @compileError("only 0 is supported for sentinels");
+                        break :sentinel sentinel_ != null;
+                    },
+                    else => @compileError("only []const u8 and [:0]const u8"),
+                }
+            },
         };
+    }
+
+    pub fn deinit(self: *const String) void {
+        const ptr = self.ptr orelse return;
+        if (self.sentinel) {
+            state.alloc.free(ptr[0..self.len :0]);
+        } else {
+            state.alloc.free(ptr[0..self.len]);
+        }
     }
 };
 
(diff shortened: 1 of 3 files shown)


jcollie (approved):
This is a quick fix, there appears to be only one place that this code is used. Longer term if this API gets more widely used will probably want to enhance the `String` struct to handle both `[] const u8` and `[:0] const u8`.

00-kat (approved):

jcollie (approved):

mitchellh (changes requested):
I want to double check this is always right. I’m not 100% sure we always send sentinel slices through this so we might trade being too short with too long in some cases. I can’t remember. I just want to check

jcollie:
> I want to double check this is always right. I’m not 100% sure we always send sentinel slices through this so we might trade being too short with too long in some cases. I can’t remember. I just want to check

I can find it used only once in the code, to open the config on macOS and then free the path. If it's used somewhere else I can't find it.

mitchellh (changes requested):
Yeah these are technically not supposed to require being null-terminated, so forcing it is a mistake. Although you're both right that it's our only usage today. Still, it's confusing in the API today that we have both a `Len` and null-termination. The idea behind `ghostty_string_s` was to give us a string type that didn't require null termination. 

I think the easiest fix for this will be to modify the structure to have a `cap: usize` as well that gives us the allocated size of the array.

CoderJoshDK:
I am not 100% sure what you had in mind for the API here. After messing around for a bit, I settled on 
```zig
fromSlice(slice: []const u8, cap: usize)
```

This is because, as far as I could tell, there is no way to determine if a `[*]` slice contains a sentinel or not. And so the caller is responsible for informing the cap. 

This is a small PR. If you have something specific in mind that isn't what I did, feel free to either let me know or just close this PR // take over. 

This PR, as it is now, does fix the error.

jcollie:
Hope you don't mind, but I pushed a commit that uses some comptime magic to make the interface a little cleaner. It'll now accept both `[]const u8` and [:0]const u8` and just do the right thing™. Couldn't really test since I don't have a mac so would be good for someone to do that before merging.

CoderJoshDK:
Oh wow, that sure is some magic. Yea, I am way too new to zig to have ever figured that out on my own. Looks very nice (thank you for taking over).

Tested this branch on my Mac and everything works. But to be fair, this doesn't actually test the non sentinel case.

mitchellh (changes requested):
This looks acceptable, I think we should add tests cause it's fairly trivial and will guard against any dumbness we might introduce one day.

src/main_c.zig
```diff
@@ -155,3 +155,23 @@ pub export fn ghostty_translate(msgid: [*:0]const u8) [*:0]const u8 {
...
+
+    const alt_slice: []const u8 = &.{ 'h', 'e', 'l', 'l', 'o' };
+    const zig_string = String.fromSlice(alt_slice);
+    try testing.expect(@TypeOf(alt_slice) == []const u8);
+    try testing.expect(zig_string.sentinel == false);
+
+    try testing.expectEqualSlices(u8, slice, alt_slice);
+    try testing.expect(std.mem.eql(u8, slice, alt_slice));
+    try testing.expect(c_null_string.len == zig_string.len);
+}
```
jcollie, line 197: we should test explicitly deallocating the strings as well.

jcollie (approved):

mitchellh:
Thank you!

PR #8895, opened 2025-09-24
libghostty-vt C shared library boilerplate, custom allocators API

mitchellh:
This adds the boilerplate necessary for `libghostty-vt` the C library.

> [!IMPORTANT]
>
> This _does not expose almost any APIs_. The point of this PR is to setup our boilerplate, build system, docs system, etc. for the libghostty-vt C lib.

- Adds a `zig build lib-vt` target to _only_ build `libghostty-vt`
- Adds the beginning of `include/ghostty-vt.h` for the C API 
- Adds a full custom allocator interface to the C API mimicking Zig custom allocators
- Adds an example in `example/c-vt` that builds a pure C program and links to our shared library and calls functions
- Adds the `osc_parser_new/free` C APIs just as a proof of concept that things work
- Adds a basic Doxygen config so we have _something_ (I'm not at all committed to Doxygen, but want us to doc from the beginning)
- Updates CI to test building the shared library for macOS, Windows, and Linux (yes, it builds for Windows!)

**Note:** To use the `dep.artifact` function provided by Zig, we must install the artifact. But this means that every `zig build` now includes `libghostty-vt`. That... could be completely fine, but it's something to consider for packagers. 

## Bikeshed

We're at a pivotal point where we must define the general _style_ of our C API. 

This includes the very bike shed things such as capitalization styling, but also general API form. 

ABI compatibility will eventually be important.

I'm very much open and would love to receive feedback form more experience C programmers on what they feel would constitute a good API. I've consumed _many_ C APIs but I haven't provided many directly. 

cc @gpanders


Doxyfile
```diff
@@ -0,0 +1,28 @@
...
+INPUT                  = include/ghostty-vt.h
+INPUT_ENCODING         = UTF-8
+RECURSIVE              = NO
+
+#---------------------------------------------------------------------------
+# HTML Output
+#---------------------------------------------------------------------------
+
+GENERATE_HTML          = YES
+HTML_OUTPUT            = zig-out/share/ghostty/doc/libghostty
```
tristan957, line 14: What do you think about templating these output paths?

include/ghostty-vt.h
```diff
@@ -7,6 +7,127 @@
 extern "C" {
 #endif
 
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
```
tristan957: Sorting the headers alphabetically could be good practice.

example/c-vt/src/main.c
```diff
@@ -1,3 +1,9 @@
+#include <stddef.h>
+#include <ghostty-vt.h>
+
 int main() {
-  return 42;
+  GhosttyOscParser parser;
+  ghostty_vt_osc_new(NULL, &parser);
```
tristan957: Checking for an error here would be good.

tristan957 (commented):
This looks awesome. Maybe we want to hook up clang-format for formatting? I'm not sure it will be totally useful given we will only have headers, but clang-tidy could be useful too.

A few other thoughts related to packaging and consumption:

- What do you think about moving the headers into a ghostty subfolder instead of the prefix? `ghostty-vt.h` -> `ghostty/vt.h`?
- Should we version the include paths like GTK or GLib?
- Let's also produce a pkg-config file. It will make consuming this library easier

I wonder if there are good practices for keeping things like a zig enum and a C enum in lock-step so that no one ever forgets to update one side of `GhosttyVtResult` for instance.

include/ghostty-vt.h
```diff
@@ -0,0 +1,197 @@
...
+ * 
+ * Creates a new OSC (Operating System Command) parser using the provided
+ * allocator. The parser must be freed using ghostty_vt_osc_free() when
+ * no longer needed.
+ * 
+ * @param allocator Pointer to the allocator to use for memory management, or NULL to use the default allocator
+ * @param parser Pointer to store the created parser handle
+ * @return GHOSTTY_VT_SUCCESS on success, or an error code on failure
+ */
+GhosttyVtResult ghostty_vt_osc_new(const GhosttyVtAllocator* allocator, GhosttyOscParser* parser);
```
JakeWharton: Mixing pointer styles within the file. Above attaches them towards the parameter name
```suggestion
GhosttyVtResult ghostty_vt_osc_new(const GhosttyVtAllocator *allocator, GhosttyOscParser *parser);
```

JakeWharton: I'm no C person, but my understanding was that `const` on the header doesn't communicate anything meaningful to the caller since they're passing the pointer location by value.
```suggestion
GhosttyVtResult ghostty_vt_osc_new(GhosttyVtAllocator* allocator, GhosttyOscParser* parser);
```

mitchellh: That's my understanding as well, I think there's no real compiler protection here. But I wanted to self-document that the API will not be modifying this value.

Doxyfile
mitchellh: I think for this PR we leave it as-is. The Doxy stuff isn't hooked up to the build system at all so it isn't meant [yet] to go into any builds. We can cross the templating path later if we hook it into the build system.

mitchellh:
* Added pkg-config file
* Include is now `ghostty/vt.h`
* Example error checks

mitchellh:
* Types and functions removed the `vt` prefix.

mitchellh:
> I wonder if there are good practices for keeping things like a zig enum and a C enum in lock-step so that no one ever forgets to update one side of GhosttyVtResult for instance.

Yeah, this would be nice.

> `clang-format`

I want to bring this in, but we can always do so later. I want to make sure we get the styling right and clang-format has a lot of knobs.

PR #8907, opened 2025-09-25
scroll: add mouse-scroll-single-event config

ownia:
Introduce a new config: mouse-scroll-single-event. If it is true, when mouse reporting is enabled, emit a single mouse wheel event per scroll gesture regardless of computed y.delta. This aligns other mainstream linux terminals and prevents the multiplier from inflating the number of reported events.

Fixes: https://github.com/ghostty-org/ghostty/discussions/8875

|               | Ghostty                   | Alacritty        | Kitty            | Ghostty (mouse-scroll-single-event = true) |
|---------------|---------------------------|------------------|------------------|--------------------------------------------|
| tmux scroll   | 15 lines per tick (3 * 5) | 5 lines per tick | 5 lines per tick | 5 lines per tick                           |
| vim scroll    | 9 lines per tick (3 * 3)  | 3 lines per tick | 3 lines per tick | 3 lines per tick                           |
| normal scroll | 3 lines per tick          | 1 lines per tick | 5 lines per tick | 3 lines per tick                           |

tmux conf has:
```
set -g mouse on
```
vimrc has:
```
set mouse=a
set ttymouse=sgr
```

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -261,6 +261,7 @@ const DerivedConfig = struct {
     mouse_interval: u64,
     mouse_hide_while_typing: bool,
     mouse_scroll_multiplier: f64,
+    mouse_scroll_single_event: bool,
     mouse_shift_capture: configpkg.MouseShiftCapture,
     macos_non_native_fullscreen: configpkg.NonNativeFullscreen,
     macos_option_as_alt: ?configpkg.OptionAsAlt,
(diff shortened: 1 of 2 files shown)


mitchellh:
Instead of a new config, I'd rather just do the right thing here. I think we should discuss if we should always do this. cc @rockorager

ownia:
Hi @rockorager @mitchellh , do you have any updates/suggestions for this issue? I'd really like to keep the mouse events to a suitable value without affecting the global settings.

rockorager:
I think it makes sense to change. I personally like the current behavior because there's no current way to tell between a wheel event and a more precise touchpad scroll. Most GUIs are able to distinguish and handle these differently, TUIs can't though.

So, I am fine going back to the original behavior to maintain compatibility but we (the broader terminal world) need to come up with something to report touchpad events.

joshka-oai:
I built out a small tool in https://github.com/joshka-oai/crossterm/tree/joshka/mouse-terminal-diagnosis (`cargo run --example mouse-scroll`) to help diagnose real world mouse scrolling events for trackpad and mouse.

In my testing on ghostty I was seeing 9 events for a single scroll wheel in about 0.5-3ms. This really should be 3 not 9, as otherwise TUIs have to introduce a bunch of time dependent tracking of scroll events to divide this to make scrolling on trackpad and wheel mouse feel right compare to native.

This should be the default, not a config option.

> we (the broader terminal world) need to come up with something to report touchpad events.

+1 to adding a new mouse mode that has high fidelity mouse scroll events - it would be nice to be able to have scroll events feel as native as native scrolling across the board without having to do much at the tui layer. Obviously solving the now problem is more important in the short term though.

PR #8914, opened 2025-09-25
vim: use :setf to set the filetype

jparise:
This is nicer because it only sets the filetype if it hasn't already been set. :setf[iletype] has been available since vim version 6.

See: https://vimhelp.org/options.txt.html#%3Asetf


mitchellh:
@gpanders Post-merge but: you cool with this?

gpanders:
Yep, makes sense 👍

PR #8917, opened 2025-09-26
feat: added move hover to activate pane instead of clicking a pane

unclemate:
This feature helps to simply move the mouse over a pane, to highlight the pane, instead of clicking the pane.

Please review. 

thank you

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -28,6 +28,7 @@ const Config = @import("config.zig").Config;
 const ResizeOverlay = @import("resize_overlay.zig").ResizeOverlay;
 const ChildExited = @import("surface_child_exited.zig").SurfaceChildExited;
 const ClipboardConfirmationDialog = @import("clipboard_confirmation_dialog.zig").ClipboardConfirmationDialog;
+const SplitTree = @import("split_tree.zig").SplitTree;
 const TitleDialog = @import("surface_title_dialog.zig").SurfaceTitleDialog;
 const Window = @import("window.zig").Window;
 const WeakRef = @import("../weak_ref.zig").WeakRef;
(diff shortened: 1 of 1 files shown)


00-kat:
> This feature helps to simply move the mouse over a pane, to highlight the pane, instead of clicking the pane.

…isn't that what [`focus-follows-mouse`] does? In fact, unless this code was AI-generated or written by someone else, you already knew about that configuration option since it's mentioned in the code you added: https://github.com/ghostty-org/ghostty/blob/0bef22d99fcb62831e28b050a08fddb48b91d0dd/src/apprt/gtk/class/surface.zig#L2320-L2322

I didn't look at the code all that closely so I don't know what it *actually* does, but based off your PR and commit descriptions, it sounds like you're looking for [`focus-follows-mouse`].

[`focus-follows-mouse`]: https://ghostty.org/docs/config/reference#focus-follows-mouse

rhodes-b:
> …isn't that what [`focus-follows-mouse`](https://ghostty.org/docs/config/reference#focus-follows-mouse) does? In fact, unless this code was AI-generated or written by someone else, you already knew about that configuration option since it's mentioned in the code you added:

It is

> https://github.com/ghostty-org/ghostty/blob/0bef22d99fcb62831e28b050a08fddb48b91d0dd/src/apprt/gtk/class/surface.zig#L2320-L2322

Expand a few lines above where the diff is you can see this change is right under the original behavior for `focus-follows-mouse` (which works)
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/surface.zig#L2244-L2252

pluiedev (changes requested):
I too am extremely confused. The code for focus-follows-mouse is already there, and the only significant difference is that we're using the spatial tree to decide whether the split receiving the input is the same as the split that's located under the cursor...? Why is that necessary at all?

As Kat said, we do absolutely require AI disclosure for any contributions to Ghostty. We do accept AI-written PRs but we have to fully know to what extent LLMs are responsible for each PR. As things stand I'm expecting a few answers.

pluiedev:
https://github.com/user-attachments/assets/3e1caf93-797b-424c-aeeb-092fe512fe0e

From what I can tell this feature already works as intended. I don't know what the deal with this PR is about, but I'm inclined to reject this since it does not seem to fix any particular problem. If you can rationalize your changes then I'm happy to change my mind, but at the present it doesn't seem sensible to me. Sorry.

PR #8925, opened 2025-09-26
build: limit cpu affinity to 32 cpus on Linux

jcollie:
Related to #8924

Zig currenly has a bug where it crashes when compiling Ghostty on systems with more than 32 cpus (See the linked issue for the gory details). As a temporary hack, use `sched_setaffinity` on Linux systems to limit the compile to the first 32 cores. Note that this affects the build only. The resulting Ghostty executable is not limited in any way.

This is a more general fix than wrapping the Zig compiler with `taskset`. First of all, it requires no action from the user or packagers. Second, it will be easier for us to remove once the upstream Zig bug is fixed.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -8,6 +8,10 @@ comptime {
 }
 
 pub fn build(b: *std.Build) !void {
+    // Works around a Zig but still present in 0.15.1. Remove when fixed.
+    // https://github.com/ghostty-org/ghostty/issues/8924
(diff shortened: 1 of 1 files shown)


qubitnano:
`nix build .#ghostty` and `nix develop -c bash -lc "zig build"` both pass on my threadripper. Thanks!

mitchellh (approved):

Issue #8903, opened 2025-09-24
keybind in `+list-themes` to write out theme config file.


jcollie:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/8874

<div type='discussions-op-text'>

<sup>Originally posted by **jcollie** September 23, 2025</sup>
Adding a keybind to the `+list-themes` TUI that would write out a file that contained `theme = <current theme name>` to a file, maybe `~/.config/ghostty/auto/theme.ghostty`? Users would then add `config-file = ?auto/theme.ghostty` to their config to use the file. Users would need to reload their config for it to take effect, but eventually that could be automated as well.

Adding `config-file = ?auto/theme.ghostty` to the file that is created when there is no config file would be a possibility as well.</div>

greathongtu:
Hi I'm looking into this but have some questions:
1. Since config-file paths are resolved relative to the main config file. If we always write to `~/.config/ghostty/auto/theme.ghostty` but the main config lives at `~/Library/Application Support/com.mitchellh.ghostty/config `(macOS), then `config-file = ?auto/theme.ghostty` won’t resolve. 
2. About “Adding `config-file = ?auto/theme.ghostty` to the file that is created when there is no config file”: there’s a default template config. Do we insert it into template config?

jcollie:
1. If the main config is `~/Library/Application Support/com.mitchellh.ghostty/config` then the theme file should be written to `~/Library/Application Support/com.mitchellh.ghostty/auto/theme.ghostty`.
2. Yes, it should be added to [src/config/config-template](https://github.com/ghostty-org/ghostty/blob/main/src/config/config-template)

greathongtu:
Hi, I just created a PR: https://github.com/ghostty-org/ghostty/pull/8930. Hope you can give it a review. Thank you!

jcollie:
Fixed by PR #8930.

skarnl:
Not sure if I have done something wrong, but for future searches.

I couldn't make this work, installed Ghostty on MacOS via Homebrew, my config in `~/.config/ghostty/config` was working, but adding the `config-file = ?auto/theme.ghostty` to it didn't work.

Eventually I had to replace it with:

```
config-file = ~/Library/Application Support/com.mitchellh.ghostty/auto/theme.ghostty
```

I found that that was the file which got updated with the ENTER and then pressing 'w' on a theme from within the `ghostty +list-themes` preview panel

[...]

PR #8930, opened 2025-09-27
Adding keybind to the +list-themes TUI that would write out a file that contained themes

greathongtu:
This resolves issue: https://github.com/ghostty-org/ghostty/issues/8903, so we can press 'w' when in TUI preview save mode to write a auto theme config


rudywasfound (commented):
LGTM

src/cli/list_themes.zig
```diff
@@ -192,6 +192,31 @@ pub fn run(gpa_alloc: std.mem.Allocator) !u8 {
...
+
+fn writeAutoThemeFile(alloc: std.mem.Allocator, theme_name: []const u8) !void {
+    const auto_path = try resolveAutoThemePath(alloc);
+    defer alloc.free(auto_path);
+
+    if (std.fs.path.dirname(auto_path)) |dir| {
+        try std.fs.cwd().makePath(dir);
+    }
+
+    var f = try std.fs.createFileAbsolute(auto_path, .{ .truncate = true });
```
jcollie: Rather than directly writing the file out, I'd use the [atomicFile](https://ziglang.org/documentation/0.15.2/std/#std.fs.Dir.atomicFile) interface to ensure that the file is always kept consistent.

src/config/config-template
```diff
@@ -24,6 +24,13 @@
 # reloaded while running; some only apply to new windows and others may require
 # a full restart to take effect.
 
+# Auto theme include
+# ==================
+# This include makes it easy to pick a theme via `ghostty +list-themes`:
+# press Enter on a theme, then 'w' to write the auto theme file.
+# This path is relative to this config file.
+config-file = ?auto/theme.ghostty
```
mitchellh: This is the only part of this PR I don't love. Instead of doing this for people, I suggest we add a message to the CLI to let the user know what to do (and to reload their config). A lot of users don't use the config template and the "write an auto theme file" sounds more automatic than it is.

mitchellh (changes requested):

src/config/config-template
greathongtu: Hi Mitchell, thanks for the feedback! I’ve reverted the config-template change and kept the guidance in the +list-themes TUI, asking users to add `config-file = ?auto/theme.ghostty` and reload. Please let me know if there’s anything else you’d like adjusted.

mitchellh (approved):

PR #8946, opened 2025-09-29
Inline All The Things

qwerasd205:
I used the new CPU counter mode in Instruments.app to track down functions that had instruction delivery bottlenecks (indicating i-cache misses) and picked a bunch of trivial functions to mark as inline (plus a couple that are only used once or twice and which benefit from inlining).

The size of `macos-arm64/libghostty-fat.a` built with `zig build -Doptimize=ReleaseFast -Dxcframework-target=native` goes from `145,538,856` bytes on `main` to `145,595,952` on this branch, a negligible increase.

These changes resulted in some pretty sizable improvements in vtebench results on my machine (Apple M3 Max):
image

With this, the only vtebench test we're slower than Alacritty in (on my machine, at 130x51 window size) is `dense_cells` (which, IMO, is so artificial that optimizing for it might actually negatively impact real world performance).

I also did a pretty simple improvement to how we copy the screen in the renderer, gave it its own page pool for less memory churn. Further optimization in that area should be explored since in some scenarios it seems like as much as 35% of the time on the `io-reader` thread is spent waiting for the lock.

> [!NOTE]
> Before this is merged, someone really ought to test this on an x86 processor to see how the performance compares there, since this *is* tuning for my processor specifically, and I know that M chips have pretty big i-cache compared to some x86 processors which could impact the performance characteristics of these changes.

### DIFF
--- a/src/datastruct/intrusive_linked_list.zig
+++ b/src/datastruct/intrusive_linked_list.zig
@@ -23,7 +23,7 @@ pub fn DoublyLinkedList(comptime T: type) type {
         /// Arguments:
         ///     node: Pointer to a node in the list.
         ///     new_node: Pointer to the new node to insert.
-        pub fn insertAfter(list: *Self, node: *Node, new_node: *Node) void {
+        pub inline fn insertAfter(list: *Self, node: *Node, new_node: *Node) void {
             new_node.prev = node;
             if (node.next) |next_node| {
                 // Intermediate node.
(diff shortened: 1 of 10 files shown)


qwerasd205:
Oh, test failures. I totally neglected to test before PRing. I'll look in to that tomorrow.

Edit: They're very confusing errors actually-- is it some sort of miscompile? It's seemingly caused by `Parser.doAction` being `inline` and then `Parser.next` being called with a comptime-known `0x1B`...

mitchellh:
I can run the vtebench marks at home when I have my x86 machine.

mitchellh:
output

mitchellh:
Uploaded the results from x86_64. The scrolling benchmarks are way higher but I think its because my window was a lot bigger. I'm going to grab results from Alacritty too once I get the same window size and font.

mitchellh:
The results are obviously still trending good on x86_64 so will merge. I think we have some scrolling issues on x86_64 because Alacritty was in the ~120ms range (the rest we were close).

PR #8949, opened 2025-09-29
gtk: make Enter confirm "Change Terminal Title"

meldron:
Fixes https://github.com/ghostty-org/ghostty/discussions/8697 by making `OK` the suggested default and activating it by default.

Previously `OK` was `destructive` which imo is not a good approach for just setting a terminal title.


mitchellh:
I believe this was destructive on purpose to match Gnome guidelines. Cc @pluiedev who would probably remember more. 

For “change title” in particular though it definitely seems odd, because that’s an op that can be undone easily (it’s absolutely not destructive). It might’ve just been a mental holdover from our close dialogues.

mitchellh:
Going to proactively accept this. This is definitely not a destructive action so making it default is safe. Users can always revert by resetting it.

PR #8960, opened 2025-09-29
Small Renderer Changes

qwerasd205:
Two small changes that were mostly for code cleanliness but both happen to fix an existing problem.

The cursor recoloring change fixes #8353 
The background cell blending change fixes #7957

Had issues running my VM so someone else should test this PR on Linux (OpenGL) to make sure there's not a problem with it before it's merged.

### DIFF
--- a/src/renderer/cell.zig
+++ b/src/renderer/cell.zig
@@ -383,7 +383,7 @@ test Contents {
     // Add a block cursor.
     const cursor_cell: shaderpkg.CellText = .{
         .atlas = .grayscale,
-        .bools = .{ .is_cursor_glyph = true },
+        .bools = .{},
         .grid_pos = .{ 2, 3 },
         .color = .{ 0, 0, 0, 1 },
     };
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -1452,7 +1452,6 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     pass.step(.{
                         .pipeline = self.shaders.pipelines.bg_color,
                         .uniforms = frame.uniforms.buffer,
-                        .buffers = &.{ null, frame.cells_bg.buffer },
                         .draw = .{ .type = .triangle, .vertex_count = 3 },
                     });
                 }
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -700,8 +700,6 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     .screen_size = undefined,
                     .padding_extend = .{},
                     .min_contrast = options.config.min_contrast,
-                    .cursor_pos = .{ std.math.maxInt(u16), std.math.maxInt(u16) },
-                    .cursor_color = undefined,
                     .bg_color = .{
                         options.config.background.r,
                         options.config.background.g,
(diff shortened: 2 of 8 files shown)


mitchellh:
Works for me in OpenGL. ❤️

mitchellh:
Note I didn't test the exact cases this fixes in OpenGL, so I want to get test cases for that so that I can verify. But I ran this for editing another PR for a bit and didn't notice any issues in normal usage.

qwerasd205:
The test cases are verifying that `background-cells-opacity` doesn't blend the cell background with the main bg color (or the bg image), and that the cursor text color looks correct (including for wide characters) -- and actually, while typing this out I just realized that minimum contrast might be applied wrong for the cursor cell since I neglected to set the bg color to the cursor color in that situation

mitchellh:
> The test cases are verifying that background-cells-opacity doesn't blend the cell background with the main bg color (or the bg image), and that the cursor text color looks correct (including for wide characters)

These both look correct.

> and actually, while typing this out I just realized that minimum contrast might be applied wrong for the cursor cell since I neglected to set the bg color to the cursor color in that situation

Correct. This is broken on OpenGL.

qwerasd205:
It'll actually be a little annoying to fix the minimum contrast on the cursor since the minimum contrast uses the same cell bg color value used to draw the cell bgs- and implementing the block cursor by just setting the cell background  is undesirable since it would make the cursor height always the cell height and it would get in the way of implementing the first-party animated cursor...

I also don't want to sacrifice the parallelism of doing the minimum contrast calculation on the GPU... so I might just revert the cursor text color change and fix the problem it happened to solve separately. I'll have a think about it though and see if I can come up with an elegant solution

qwerasd205:
Ok there are two options I can think of:
1. Revert the change, keep cursor text recoloring on the GPU, fix the one bug to do with that
2. Do a minimum contrast calculation on the CPU *only* for cursor text, and exempt it from the GPU-side minimum contrast calculation

Neither is *super* desirable, but I like 2 a little better than 1 personally.

jcollie:
> Ok there are two options I can think of:
> 
>     1. Revert the change, keep cursor text recoloring on the GPU, fix the one bug to do with that
> 
>     2. Do a minimum contrast calculation on the CPU _only_ for cursor text, and exempt it from the GPU-side minimum contrast calculation
> 
> Neither is _super_ desirable, but I like 2 a little better than 1 personally.

Other than not breaking backwards compatibility is there a case for not dropping `minimum-contrast`? Is it actually useful as an accessibility solution for people with vision problems?

qwerasd205:
It's not just for people with vision problems. If a terminal application uses bad colors that don't have good contrast, you should still be able to see what the text is. Sometimes the bad contrast is just bad choices, more often it's because they're using palette colors that just don't happen to contrast well in the user's current theme.

mitchellh:
Subjectively, it is a fairly popular configuration, too. Whenever we break it we get quite a few reports (we broke it in 1.2.0).

qwerasd205:
Actually, we can't do the minimum contrast calculations CPU-side unless we want to duplicate the color math (including color space conversions)... I guess it'll have to be 1. (revert, fix the cursor text load options)

mitchellh:
What's the status of this PR? I know its conflicted but wasn't sure if I should put the effort in to rebase and look into this or not.

mitchellh:
Going to close this. There's obviously fixes here we can and should pick up, but no need to cause any undue stress on anyone by keeping it stale. This PR is linked to the linked issues so we can always look back on it to extract the proper fixes.

PR #8961, opened 2025-09-29
macOS: Focus Terminal App Intent

### LINKED ISSUE #8791: macOS: add focus terminal app intent
Discussion: #8657 

We should add an app intent that brings a terminal to focus. This should not bring the entire app to focus which is satisfied by the `Open App` built-in shortcut already.

himura467:
Closes #8791 
Discussion: #8657 

### Summary
This adds the FocusTerminalIntent App Intent and related function (focusSurface), allows external tools (such as Shortcuts/Siri) to programmatically move focus to a specific terminal window or tab.

### Verification
This functionality has been tested across following scenarios, confirming correct focus behavior for:

- Split Window
- Tab Group
- Quick Terminal

### Note
It is not supported to move focus to a split that is hidden by a zoomed split. The same applies to the CloseTerminalIntent.

### AI Disclosure
This pull request was made with assistance from Claude Code.
I reviewed all AI-generated code and wrote the final output manually.

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -75,6 +75,7 @@
 				"Features/App Intents/CommandPaletteIntent.swift",
 				"Features/App Intents/Entities/CommandEntity.swift",
 				"Features/App Intents/Entities/TerminalEntity.swift",
+				"Features/App Intents/FocusTerminalIntent.swift",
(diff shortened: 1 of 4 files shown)


mitchellh (approved):
Awesome. This looks good to me. 

Just noting it so it's somewhere: with this in place, we can now implement the `PRESENT_TERMINAL` action too. I don't think this is used anywhere on the macOS side yet but it is an apprt action we haven't implemented.

PR #8974, opened 2025-09-30
Revert "renderer: slightly optimize screen copy"

jcollie:
This reverts commit fcea09e413a55c677dca377f716aa9bc6465306b because it appears to be causing memory leaks.

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -95,9 +95,6 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
         /// Allocator that can be used
         alloc: std.mem.Allocator,
 
-        /// MemoryPool for PageList pages which we use when cloning the screen.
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Approved. Happy to see this commit come back once we fix the leaks.

qwerasd205 (approved):
Oh I know why the leak happens. I'm fine with this revert merging though cause idk when I'll get around to fixing it (essentially: pages of non-standard size use the child allocator of the page pool to init, which is the page allocator, and never get deinited since I assumed it would use the arena and removed the explicit deinit :p)

mitchellh:
Darn, I missed this in the 1.2.1 back port. That was my mistake. 1.2.1 has a horrible memory leak because of this. Going to cut a 1.2.2 for it.

PR #8989, opened 2025-10-01
Convert framegen to C, add compressed data to source tarball

mitchellh:
Zig 0.15 removed the ability to compress from the stdlib, which makes porting our framegen tool to Zig 0.15+ more work than it's worth. We already depend on and have the ability to build zlib, and Zig is a full blown C compiler, so let's just use C.

The framegen C program doesn't free any memory, because it is meant to exit quickly. It otherwise behaves pretty much the same as the old Zig codebase.

The build scripts were modified to build the C program and run it, but also to include the framedata in the generated source tarball so that downstream packagers don't have to do this (although they'll have all the deps anyways).

**AI disclosure:** The Zig to C conversion was done by Amp. I know C and reviewed the code. The build system improvements are partially done by Amp as well but I finished it all out. I've reviewed everything. Full thread: https://ampcode.com/threads/T-4e0a80af-3d2a-48d6-bf59-3e2f05eb7673


mitchellh:
There's a couple modifications I'm making to this still before merge. I don't want to put the Zig in the dist resources, I'm going to always gen that at build time.

SilasLock:
Just in case it's useful for making future Ghostty design decisions, I wanted to point out that compress was recontributed to the Zig stdlib quite recently:
https://github.com/ziglang/zig/pull/25301

PR #8990, opened 2025-10-01
fix(font): Fix positioning of scaled glyphs that don’t specify alignment

danielwe:
Follow-up to #8563, which broke scaling without alignment. This change recovers the behavior from before #8563, such that a scaled group is clamped to the constraint width and height if necessary, and otherwise, scaling does not shift the center of the group bounding box.

As a part of this change, horizontal alignment was rewritten to assume the face is flush with the left edge of the cell. The cell-to-face offset in the rendering code is then applied regardless of the value of `align_horizontal`. This both simplifies the code and improves consistency, as it ensures that the offset is the same for all non-bitmap glyphs (rounded in FreeType, not rounded in CoreText). It's the right thing to do following the align-to-face changes in #8563.

### DIFF
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -370,11 +370,7 @@ pub const Face = struct {
         // We center all glyphs within the pixel-rounded and adjusted
         // cell width if it's larger than the face width, so that they
         // aren't weirdly off to the left.
-        //
-        // We don't do this if the constraint has a horizontal alignment,
-        // since in that case the position was already calculated with the
-        // new cell width in mind.
-        if ((opts.constraint.align_horizontal == .none) and (metrics.face_width < cell_width)) {
+        if (metrics.face_width < cell_width) {
             // We add half the difference to re-center.
             x += (cell_width - metrics.face_width) / 2;
         }
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -338,14 +338,7 @@ pub const Face = struct {
         const cell_height: f64 = @floatFromInt(metrics.cell_height);
 
         // Next we apply any constraints to get the final size of the glyph.
-        var constraint = opts.constraint;
-
-        // We eliminate any negative vertical padding since these overlap
-        // values aren't needed  with how precisely we apply constraints,
-        // and they can lead to extra height that looks bad for things like
-        // powerline glyphs.
-        constraint.pad_top = @max(0.0, constraint.pad_top);
-        constraint.pad_bottom = @max(0.0, constraint.pad_bottom);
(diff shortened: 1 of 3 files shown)


danielwe:
I moved the alignment change for stretched glyphs from #8847 to here, where it's a more natural fit and eliminates the need for rebasing between these PRs. Note that #8847 is needed to take full advantage of this; until then, many codepoints have incorrect or missing constraints. In particular, the screenshots below are taken with #8847 included, and would look much worse otherwise (see #8335).

Copying over the comment from the other thread (with new, lower-DPI screenshots):

> I added one more small change: Stretched glyphs work best if they're scaled and aligned to the grid, since they're typically meant to align across cells (this is the exception to the argument in #8563 that icons should be aligned to the face). This also lets us eliminate the band-aid of horizontal negative padding (overlap), which is another improvement over `font-patcher` that's only possible because we're applying constraints with knowledge of the pixel grid.
>
> Here are screenshots showing the difference this makes for the progress bar elements from Nerd Fonts. Notice the overlap artifacts no longer appearing after this change.
>
> Screenshot 2025-10-01 at 19 47 59
> Screenshot 2025-10-01 at 19 46 55

danielwe:
Would also love to add some tests to avoid regressions here, but since they’d be comparing floats, It would be great to have #8979 merged first.

Otherwise, this is ready, after a deluge of force pushes and updates as I kept seeing better ways to do things—sorry about that.

mitchellh:
> Would also love to add some tests to avoid regressions here, but since they’d be comparing floats, It would be great to have https://github.com/ghostty-org/ghostty/pull/8979 merged first.

Done. 

I'll review this and if you get tests in first great. If not, lets add them after.

src/font/face/coretext.zig
```diff
@@ -370,11 +363,7 @@ pub const Face = struct {
         // We center all glyphs within the pixel-rounded and adjusted
         // cell width if it's larger than the face width, so that they
         // aren't weirdly off to the left.
-        //
-        // We don't do this if the constraint has a horizontal alignment,
-        // since in that case the position was already calculated with the
-        // new cell width in mind.
-        if ((opts.constraint.align_horizontal == .none) and (metrics.face_width < cell_width)) {
+        if (metrics.face_width < cell_width) {
```
danielwe, line 366: Something went awry here, this should also have included the `constraint.size != .stretch` condition, like in `freetype.zig`. Not sure what could have happened. I'll fix it in the tests PR, coming shortly.

mitchellh: Probably a funky rebase. I’ll look for the fix up PR thanks!

PR #8992, opened 2025-10-02
core: add 'command finished' notifications

### LINKED ISSUE #8991: Implement command finished notifications.
`notify-on-command-finish = never | unfocused | always`
`notify-on-command-finish-action = bell , no-notify`
`notify-on-command-finish-after = 5s`

From Mitchell:
> If we implement this as a generic apprt action (I'm sure you will) that takes the action as a parameter, then we can easily enhance this in the future to manual per-command subscriptions, too. For example, right clicking the terminal and saying "notify when the next command finishes". I wouldn't do THAT in the same PR, but as long as we build the apprt action right that should be easy later.

_Originally posted by @jcollie in https://github.com/ghostty-org/ghostty/discussions/3555#discussioncomment-14556266_

jcollie:
Fixes #8991

Uses OSC 133 esc sequences to keep track of how long commands take to execute. If the user chooses, commands that take longer than a user specified limit will trigger a notification. The user can choose between a bell notification or a desktop notification.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -733,6 +733,14 @@ typedef struct {
   int8_t progress;
 } ghostty_action_progress_report_s;
 
+// apprt.action.CommandFinished.C
+typedef struct {
+  // -1 if no exit code was reported, otherwise 0-255
+  int16_t exit_code;
+  // number of nanoseconds that command was running for
+  uint64_t duration;
+} ghostty_action_command_finished_s;
+
 // apprt.Action.Key
 typedef enum {
   GHOSTTY_ACTION_QUIT,
@@ -788,6 +796,7 @@ typedef enum {
   GHOSTTY_ACTION_SHOW_CHILD_EXITED,
   GHOSTTY_ACTION_PROGRESS_REPORT,
   GHOSTTY_ACTION_SHOW_ON_SCREEN_KEYBOARD,
+  GHOSTTY_ACTION_COMMAND_FINISHED,
 } ghostty_action_tag_e;
 
 typedef union {
@@ -819,6 +828,7 @@ typedef union {
   ghostty_action_close_tab_mode_e close_tab_mode;
   ghostty_surface_message_childexited_s child_exited;
   ghostty_action_progress_report_s progress_report;
+  ghostty_action_command_finished_s command_finished;
 } ghostty_action_u;
 
 typedef struct {
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -33,6 +33,7 @@ const font = @import("font/main.zig");
 const Command = @import("Command.zig");
 const terminal = @import("terminal/main.zig");
 const configpkg = @import("config.zig");
+const Duration = configpkg.Config.Duration;
 const input = @import("input.zig");
 const App = @import("App.zig");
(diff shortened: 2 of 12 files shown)


jcollie:
This'll need a macOS implementation since most of the functionality is implemented in the apprt.

include/ghostty.h
```diff
@@ -733,6 +733,14 @@ typedef struct {
   int8_t progress;
 } ghostty_action_progress_report_s;
 
+// apprt.action.CommandFinished.C
+typedef struct {
+  // -1 if no exit code was reported, otherwise 0-255
+  int16_t exit_code;
+  // number of nanoseconds that command was running for
+  uint64_t duration;
+} ghostty_action_command_finished_s;
```
mitchellh, line 742: Glad we aren't chostty this time lol.

src/apprt/surface.zig
```diff
@@ -96,6 +96,14 @@ pub const Message = union(enum) {
     /// Report the progress of an action using a GUI element
     progress_report: terminal.osc.Command.ProgressReport,
 
+    /// A command has started in the shell, start a timer.
+    start_command_timer,
```
mitchellh: I think we can just name this `start_command`. The timer part is an implementation detail of however the surface wants to handle this.

src/apprt/surface.zig
```diff
@@ -96,6 +96,14 @@ pub const Message = union(enum) {
...
+
+    /// A command has finished in the shell, stop the timer and send out
+    /// notifications as appropriate. The optional u8 is the exit code
+    /// of the command.
+    stop_command_timer: ?u8,
```
mitchellh: Same here, I think we can say `stop_command`.

src/Surface.zig
```diff
@@ -984,6 +998,36 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
             self.selection_scroll_active = active;
             try self.selectionScrollTick();
         },
+
+        .start_command_timer => {
+            self.command_timer = std.time.nanoTimestamp();
```
mitchellh: I prefer to use `Instant` here, which I think always uses the monotonic clock. Unless I'm mistaken.

mitchellh (changes requested):
Some fairly simple requests. The big picture design looks good.

include/ghostty.h
jcollie: 🤣🤯

src/apprt/surface.zig
jcollie: Done.

src/apprt/surface.zig
jcollie: Done.

src/Surface.zig

mitchellh (approved):

00-kat:
> This'll need a macOS implementation since most of the functionality is implemented in the apprt.

Did that ever happen? I don't see it in this PR, nor the source tree, nor can I find an issue tracking it.

jcollie:
> > This'll need a macOS implementation since most of the functionality is implemented in the apprt.
> 
> Did that ever happen? I don't see it in this PR, nor the source tree, nor can I find an issue tracking it.

Neither has happened.

PR #8998, opened 2025-10-02
build: Turn `GhosttyKit` into a Swift package for future Swift macro integration

bo2themax:
### Changes
- Renamed `GhosttyKit.xcframework` to `CGhosttyKit.xcframework`
- Added GhosttyKit Swift package which depends on `CGhosttyKit.xcframework` and exports `Carbon` and `AppKit/UIKit` for easy use
- Modified `zig build` steps to current workflow

### AI Disclosure
This pull request was made with assistance from GitHub Copilot. I reviewed all AI-generated Zig code and wrote the final output manually.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -148,9 +148,6 @@ pub fn build(b: *std.Build) !void {
         if (config.emit_xcframework) {
             xcframework.install();
 
-            // The xcframework build always installs resources because our
-            // macOS xcode project contains references to them.
-            resources.install();
             if (i18n) |v| v.install();
         }
 
--- a/include/module.modulemap
+++ b/include/module.modulemap
@@ -1,7 +1,7 @@
 // This makes Ghostty available to the XCode build for the macOS app.
 // We append "Kit" to it not to be cute, but because targets have to have
 // unique names and we use Ghostty for other things.
-module GhosttyKit {
+module CGhosttyKit {
     umbrella header "ghostty.h"
     export *
 }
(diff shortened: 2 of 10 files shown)


mitchellh:
Thank you. It's good to know how to do this but the `GhosttyKit` framework in its current form is not intended to be supported as a generally consumable package. Anyone who wants to do so is welcome to fork the project and maintain it on their own. Longer term, I'll likely make a supported Swift package for libghostty but I'm not ready to expose this particular code that way.

bo2themax:
> Thank you. It's good to know how to do this but the `GhosttyKit` framework in its current form is not intended to be supported as a generally consumable package. Anyone who wants to do so is welcome to fork the project and maintain it on their own. Longer term, I'll likely make a supported Swift package for libghostty but I'm not ready to expose this particular code that way.

This is actually a local package and not built for public use, since the xcframework is ignored. But ok

mitchellh:
Ah, I misunderstood that. What's the benefit of this?

bo2themax:
> Ah, I misunderstood that. What's the benefit of this?

I was implementing the settings ui and trying to use a Swift macro to generate UI-related code for configuration keys.  Since Swift macros require a Swift package, I thought I’d turn it into one to read supported keys and separate it into a new pull request to support this. 

However, it could also work without this by declaring a dummy project that wraps `GhosttyKit` or just using zig to generate swift files.  Anyway, it’s fine. I haven’t finished the settings UI and will open a WIP pull request later for everyone’s suggestions. Let’s see if this is actually necessary.

PR #9000, opened 2025-10-02
crash: remove minidump parser

mitchellh:
We never used it because our minidump files on Linux didn't contain meaningful information. With Zig's Writergate, let's drop this and rewrite it later, we can always resurrect it from the git history.

Rejoice @pluiedev


jcollie (approved):

pluiedev:
Nice. Happy 9k :tada:

PR #9001, opened 2025-10-02
gtk: fix duplicate signal handlers

jcollie:
Signal handlers are connected to surface objects in two spots - when a tab is added to a page and when the split tree changes. This resulted in duplicate signal handlers being added for each surface. This was most noticeable when copying the selection to the clipboard - you would see two "Copied to clipboard" toasts. Ensure that there is only one signal handler by removing any old ones before adding the new ones.

### DIFF
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -693,14 +693,23 @@ pub const Window = extern struct {
         self: *Self,
         tree: *const Surface.Tree,
     ) void {
-        // Ensure that all old signal handlers have been removed before adding
-        // them. Otherwise we get duplicate surface handlers.
-        self.disconnectSurfaceHandlers(tree);
-
         const priv = self.private();
         var it = tree.iterator();
         while (it.next()) |entry| {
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Question: This is a simple change, but would it be more robust to change `connectSurfaceHandlers` to `setSurfaceHandlers` and change `tree` to an optional. If it is null then we just disconnect. Then we can replace all usage everywhere with the one call. Thought?

jcollie:
> Question: This is a simple change, but would it be more robust to change `connectSurfaceHandlers` to `setSurfaceHandlers` and change `tree` to an optional. If it is null then we just disconnect. Then we can replace all usage everywhere with the one call. Thought?

I don't think that would work as the disconnect procedure requires a tree, and there's one case where we disconnect the signal handlers but don't reconnect any new ones.

jcollie:
Ok, I've updated the patch so that it only makes one pass through the surfaces rather than two.

mitchellh (approved):

PR #9004, opened 2025-10-02
Zig 0.15

mitchellh:
Replaces #8372 

Before merge I'm going to squash this and give @pluiedev coauthor, since I took a lot of her work. I just have to go through this myself to make sure I learn all the changes in Zig 0.15, but as I got things, I copy and pasted her work in. My work is probably less thorough and there are places we can convert deprecated things, probably, but this results in green CI.

## Benchmarks

It looks like there are some speed regressions in isolated places. I'm not sure if this is noise or not, I'm going to keep running some tests. If someone can check macOS that'd be helpful (my vtebench is down on macOS atm), cc @qwerasd205 if interested.

On an x86_64 system:

output


qwerasd205:
macOS vtebench results on an M3 Max (I realized that a lot of the benchmarks weren't running / running correctly because of the outdated system bash version, swapped the shebangs for a newer version and got them to run)
image

kenvandine (approved):
The snap related changes look good to me.

PR #9010, opened 2025-10-03
Add update-mirror Nu script to ease mirror updating for deps

mitchellh:
This adds a new script we can manually run that downloads all the files that need to be uploaded to the mirror and updates our build.zig.zon. The upload still happens manually [by me] but this simplifies the task greatly.


mitchellh:
I forgot to disclose that Amp wrote this but I reviewed and modified manually. I’ll share the thread when I’m back at a PC. Off on mobile now.

PR #9023, opened 2025-10-04
font: Add comprehensive constraint tests

danielwe:
As promised in #8990.

I opted for hardcoded metrics and bounding boxes rather than actually loading fonts and glyphs, both to avoid backend dependence and limit the focus to the constraint calculations themselves, and because I wanted to test a case that isn't exhibited by any of the fonts available in the repo.

This also fixes an error from #8990, probably due to a botched cherry-pick or rebase.

### DIFF
--- a/src/font/face.zig
+++ b/src/font/face.zig
@@ -507,3 +507,196 @@ test "Variation.Id: slnt should be 1936486004" {
     try testing.expectEqual(@as(u32, 1936486004), @as(u32, @bitCast(id)));
     try testing.expectEqualStrings("slnt", &(id.str()));
 }
+
+test "Constraints" {
+    const comparison = @import("../datastruct/comparison.zig");
+    const getConstraint = @import("nerd_font_attributes.zig").getConstraint;
+
+    // Hardcoded data matches metrics from CoreText at size 12 and DPI 96.
+
+    // Define grid metrics (matches font-family = JetBrains Mono)
+    const metrics: Metrics = .{
+        .cell_width = 10,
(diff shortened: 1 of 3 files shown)


danielwe:
Oh dang I didn't realize that #8979 and hence this PR would be punted to 1.3.0, otherwise I'd have submitted the pure fixup part of this in a separate PR. This means that with 1.2.1 on macOS, all non-sprite glyphs are slightly offset to the right (usually between 0 and 0.5 pixels, unless `adjust-cell-width` has been used), and bitmap glyphs are not quantized.

The possible observable artifacts of this are:

* Non-sprite graphics elements, such as non-geometric powerline glyphs, may not be perfectly aligned (note gap):\
  Screenshot 2025-10-06 at 18 36 10
* Pixelation artifacts/antialiasing may be asymmetric for characters where it could have been symmetric:\
  Screenshot 2025-10-06 at 18 39 36
* Bitmap glyphs like emojis may not look quite as sharp as designed.
* Cursor and selection highlighting may be _slightly_ asymmetric around glyph positions.

This will typically be more noticeable on lower-DPI screens and smaller font sizes, except for the alignment gap which does not depend on font size. Everything will be more noticeable with a positive `adjust-cell-width`.

mitchellh:
Oh did I forget to backport this? It had the right milestone. We can pull it onto a 1.2.2 well still get out in the next month.

danielwe:
Great! #8979 also needs the 1.2.2 milestone, then.

PR #9024, opened 2025-10-04
Expand `~` in `macos-custom-icon`

danielwe:
Since #8999, `macos-custom-icon` works when its a fully expanded absolute path like `/Users/username/dir/icon.icns`, but not when it's abbreviated as `~/dir/icon.icns`. Users were understandably surprised and confused by this. This PR adds tilde expansion using `NSString`s built-in property for this.

Also removed a line from the config docs that seemed erroneous. Given that the option has a functional default, it seems incorrect to say that it's required.

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.Config.swift
+++ b/macos/Sources/Ghostty/Ghostty.Config.swift
@@ -314,17 +314,14 @@ extension Ghostty {
 
         var macosCustomIcon: String {
             #if os(macOS)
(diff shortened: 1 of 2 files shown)


mitchellh:
I'm fine with this here for now but we should probably do this in `Config.zig` similar to how we handle other `Path` values. We may want to just switch this to a `Path` too but it doesn't currently support the C API so we'd have to make those changes first. 😄

mitchellh (approved):

PR #9028, opened 2025-10-04
Move paste encoding to the input package, test, optimize away one alloc

mitchellh:
This moves our paste logic to `src/input` in preparation for exposing this as part of libghostty-vt. This yields an immediate benefit of unit tests for paste encoding. 

Additionally, we were able to remove one allocation on every unbracketed paste path unless the input specifically contains a newline. Unlikely to be noticable, but nice.

NOTE: This also includes one change in behavior: we no longer encode `\r\n` and a single `\r`, but as a duplicate `\r\r`. This matches xterm behavior and I don't think will result in any issues since duplicate carriage returns should do nothing in well-behaved terminals.


mitchellh:
Pushed a change to expose paste encoding as part of `libghostty-vt`. cc @rockorager

PR #9033, opened 2025-10-05
fix:use builtin memmove

NikoMalik:


mitchellh (changes requested):
Before replacing the libc path we’d need benchmarks showing this doesn’t regress performance. We can replace the non libc path though.

jcollie:
I was curious, so I threw together this: https://codeberg.org/jcollie/bench-zig-memmove

If this is right, there's almost no difference in performance with maybe a tiny advantage to libc. Of course this applies only to my CPU (AMD Ryzen 7 5800X). Apple Silicon may produce different results.

Screenshot From 2025-10-05 10-05-22

mitchellh:
Thanks @jcollie the place we last saw a major difference was in vtebench and our terminal stream benchmarks. I’d like to see those too. Hopefully can get rid of our fastmem file completely one day :)

jcollie:
Interesting. Instead of large moves (which likely overlapped), I ran the benchmark moving at most 8 bytes at a time. So seems as if in this scenario Zig beats libc. Perhaps because Zig can inline `@memmove` whereas it can't with `libc`.

Screenshot From 2025-10-05 10-29-22

PR #9040, opened 2025-10-05
lib-vt: expose key encoding as a C API

mitchellh:
Example in `example/c-vt-key-encode`, but here is an abridged version:

```c
#include <assert.h>
#include <stddef.h>
#include <stdio.h>
#include <string.h>
#include <ghostty/vt.h>

int main() {
  GhosttyKeyEncoder encoder;
  GhosttyResult result = ghostty_key_encoder_new(NULL, &encoder);
  assert(result == GHOSTTY_SUCCESS);

  ghostty_key_encoder_setopt(encoder, GHOSTTY_KEY_ENCODER_OPT_KITTY_FLAGS, &(uint8_t){GHOSTTY_KITTY_KEY_ALL});

  GhosttyKeyEvent event;
  result = ghostty_key_event_new(NULL, &event);
  assert(result == GHOSTTY_SUCCESS);
  ghostty_key_event_set_action(event, GHOSTTY_KEY_ACTION_RELEASE);
  ghostty_key_event_set_key(event, GHOSTTY_KEY_CONTROL_LEFT);
  ghostty_key_event_set_mods(event, GHOSTTY_MODS_CTRL);

  char buf[128];
  size_t written = 0;
  result = ghostty_key_encoder_encode(encoder, event, buf, sizeof(buf), &written);
  assert(result == GHOSTTY_SUCCESS);

  ghostty_key_event_free(event);
  ghostty_key_encoder_free(encoder);
  return 0;
}
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,9 +13,16 @@ A file for [guiding coding agents](https://agents.md/).
 ## Directory Structure
 
 - Shared Zig core: `src/`
-- C API: `include/ghostty.h`
+- C API: `include`
 - macOS app: `macos/`
 - GTK (Linux and FreeBSD) app: `src/apprt/gtk`
+
+## libghostty-vt
+
+- Build: `zig build lib-vt`
+- Test: `zig build test-lib-vt`
+- Test filter: `zig build test-lib-vt -Dtest-filter=<test name>`
+- When working on libghostty-vt, do not build the full app.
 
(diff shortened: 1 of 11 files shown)


mitchellh:
As always, the exact _style_ of the C API is 100% up for debate still. Its more important to just get some of this logic out there.

PR #9059, opened 2025-10-06
osc: parse additional OSC 133 options

jcollie:
OSC 133;A can have:

- special_key
- click_events

OSC 133;C can have:

- cmdline
- cmdline_url

Notably, they are in use by `fish`. Not sure what other shells currently use these options.

Note that the options are only parsed. Nothing further is done with them at this point.


mitchellh (approved):

mitchellh:
Thanks for reorganizing the tests in a separate PR. Your PR splits have been much better recently. ❤️

PR #9082, opened 2025-10-08
Add the command that finished to the command finished notifications

jcollie:
Some shells use OSC 133;C sequences to report the actual command that is about to be executed to the terminal. We take advantage of that to include the command in command finished notifications.

At this time I know of only one shell that does this, fish.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -735,9 +735,22 @@ typedef struct {
 
 // apprt.action.CommandFinished.C
 typedef struct {
-  // -1 if no exit code was reported, otherwise 0-255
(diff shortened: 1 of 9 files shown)


mitchellh (changes requested):
I'm not sure what exactly I'm looking for yet, but I don't like the various memory management things happening here in Surface to accept this outright.

The OSC parsing looks fine, if you want to split that out I'd be happy to approve that separately until I have time to think more critically about the connection to apprts.

jcollie:
OK, gonna break this up into some smaller PRs.

PR #9089, opened 2025-10-08
introduce split-preserve-zoom config to maintain zoomed splits during navigation

### LINKED ISSUE #8458: Add `split-preserve-zoom` configuration
Discussion: #8447 

Add a new config option `split-preserve-zoom` that is a packed struct that starts with a single option `navigation` (default false). The semantics are that when `navigation` is set, zoomed splits will remain zoomed when split navigation (`goto_split`) is done. 

I'm using a packed structure because there are a variety of scenarios we unzoom splits and we may want to add more options later, unclear.

Example:

```
split-preserve-zoom = navigation
```

lorenries:
Closes #8458.

- Adds the split-preserve-zoom config option with a navigation flag.
- GTK and macOS runtimes now respect the flag so zoomed splits stay zoomed when using split navigation. I've tested this on macOS (video below), but have not tested on GTK.

This PR was written primarily with Codex CLI, using the [gh-issue](https://github.com/ghostty-org/ghostty/blob/main/.agents/commands/gh-issue) command.

Here is a short video of the debug build:

https://github.com/user-attachments/assets/3abea255-98e1-4a4f-9196-7c1b2663b9d2

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -985,6 +985,14 @@ palette: Palette = .{},
 /// Available since: 1.1.0
 @"split-divider-color": ?Color = null,
 
+/// Control when Ghostty preserves the zoomed state of a split. This is a packed
+/// struct so more options can be added in the future. The `navigation` option
+/// keeps the current split zoomed when split navigation (`goto_split`) changes
+/// the focused split.
+///
+/// Example: `split-preserve-zoom = navigation`
+@"split-preserve-zoom": SplitPreserveZoom = .{},
+
 /// The foreground and background color for search matches. This only applies
 /// to non-focused search matches, also known as candidate matches.
 ///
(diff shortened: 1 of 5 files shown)


mitchellh (approved):
Thanks! Sorry for the late review, but finally rebased this, reviewed it, and fixed a couple bugs. Overall was solid though.

src/config/Config.zig
```diff
@@ -985,6 +985,14 @@ palette: Palette = .{},
 /// Available since: 1.1.0
 @"split-divider-color": ?Color = null,
 
+/// Control when Ghostty preserves the zoomed state of a split. This is a packed
```
tristan957, line 988: Seems like we are missing an `Available since:` annotation. Also, the doc string seems to leak Zig internals: "This is a packed struct...".

mitchellh: Good point, let me follow that up real quick.

khangnm1340:
Thanks you so much for this feature.
Before, I had to add this workaround to replicate the feature. (where alt+l is `toggle zoom` and alt+backslash is `switch to the other pane`). But it only works best with 2 panes.
```
hyprctl dispatch sendshortcut alt,l,class:com.mitchellh.ghostty
sleep 0.010
hyprctl dispatch sendshortcut alt,backslash,class:com.mitchellh.ghostty
hyprctl dispatch sendshortcut alt,l,class:com.mitchellh.ghostty
```

PR #9101, opened 2025-10-09
Get app_version from build.zig.zon .version

EriksRemess:
Reads .version from build.zig.zon and passes it to Config.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -3,9 +3,10 @@ const assert = std.debug.assert;
 const builtin = @import("builtin");
 const buildpkg = @import("src/build/main.zig");
 const appVersion = @import("build.zig.zon").version;
+const minimumZigVersion = @import("build.zig.zon").minimum_zig_version;
 
 comptime {
-    buildpkg.requireZig("0.15.1");
+    buildpkg.requireZig(minimumZigVersion);
 }
 
 pub fn build(b: *std.Build) !void {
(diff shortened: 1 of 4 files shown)


build.zig
```diff
@@ -2,6 +2,7 @@ const std = @import("std");
 const assert = std.debug.assert;
 const builtin = @import("builtin");
 const buildpkg = @import("src/build/main.zig");
+const appVersion = @import("build.zig.zon").version;
 
 comptime {
     buildpkg.requireZig("0.15.1");
```
jcollie: I think that we should import the Zig version from `build.zig.zon` as well.

jcollie (approved):

build.zig
EriksRemess: added that in ea5ea5f98ed90e482e12da35c8374fdbf9a5b163

jcollie: Oof. I forgot about the CI jobs that get the Zig version by parsing it from the `build.zig` so you may want to back that out unless you want to fix those up.

EriksRemess:
updated test.yml and Dockerfile so they read from build.zig.zon too

build.zig
EriksRemess: fixed in 402c492d9467e0dfac1c28be3df6aec9fbac10dd

mitchellh (approved):
Thank you!

PR #9111, opened 2025-10-09
feat: Scroll bindings respect performable flag in alternate screen apps

### LINKED ISSUE #6765: `scroll_*` bindings should respect "performable" and do nothing in alt screen
### Discussed in https://github.com/ghostty-org/ghostty/discussions/6719

<sup>Originally posted by **sirburpalot** March 14, 2025</sup>
Key bindings for actions like scroll_page_up, scroll_to_top, etc., when marked with the "performable:" prefix, should be ignored in fullscreen applications like `nano` or `far2l`, because such applications do not typically utilize the scrollback buffer.

How to reproduce the bug:

1. Add `keybind = performable:shift+home=scroll_to_top` to Ghostty's config.
2. Launch `nano`.
3. Type a string of random characters and press Shift+Home. The Shift+Home shortcut will not be passed to `nano`.</div>

### Notes

I am constraining this issue specifically to alt screen, but technically scroll should respect performable in ANY scenario where there is no scrollback, so if we fix it that way that'd be best. ❤ 

Marked as a feature because performable is documented as not working for everything yet.

cmvanb:
Fixes #6765

When key bindings for scroll actions are marked as performable, they are not executed in alt screen apps.

Tested with nano and nvim on Arch Linux.

AI Disclaimer: I use a Github Copilot autocompletion plugin in Neovim.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4772,12 +4772,24 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
         },
 
         .scroll_to_top => {
+            {
+                self.renderer_state.mutex.lock();
+                defer self.renderer_state.mutex.unlock();
+                if (self.io.terminal.active_screen == .alternate) return false;
+            }
+
             self.io.queueMessage(.{
                 .scroll_viewport = .{ .top = {} },
             }, .unlocked);
         },
 
         .scroll_to_bottom => {
+            {
+                self.renderer_state.mutex.lock();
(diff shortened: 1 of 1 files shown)


jcollie:
Hmm, gotta quibble a bit with the description of what this does. Seems to me a more accurate description would be that the scrolling actions are _never_ performed on the alternate screen. The new behavior doesn't depend on the state of the `perfomable` keybind flag.

cmvanb:
edit: nevermind. let me rethink this.

PR #9121, opened 2025-10-10
Set title as argv[0] for commands specified with `-e`

ambroisie:
I want to see #7932 get merged, so applied the latest proposed patch.

Will close if the original PR gets some traction, as I do _not_ know Zig nor this project.


mitchellh (approved):
Thanks, this looks good. I made some style changes.

PR #9122, opened 2025-10-10
gtk4-layer-shell: version from build.zig.zon, reenable ubsan

EriksRemess:
added version read from build.zig.zon
didn't get any ubsan error locally with zig 0.15.1 (related to https://github.com/ghostty-org/ghostty/issues/5744#issuecomment-2719313984)


mitchellh:
I'm willing to just let this sit in tip and see if it causes problems. Thanks.

PR #9127, opened 2025-10-10
osc: do inplace decoding of cmdline passed in OSC 133;C

jcollie:


src/os/inplace.zig
```diff
@@ -0,0 +1,267 @@
+const std = @import("std");
+
+/// Do an in-place decode of a string that has been encoded in the same way
+/// that `bash`'s `printf %q` encodes a string. This is safe because a string
+/// can only get shorter after decoding. This destructively modifies the buffer
+/// given to it. If an error is returned the buffer may be in an unusable state.
+pub fn printf_q_decode(buf: [:0]u8) error{DecodeError}![:0]const u8 {
```
mitchellh: I think `inplace.zig` is not the right name for this file. Maybe its `string_encoding.zig` or something?

mitchellh: Oh also the underscore style isn't idiomatic to our codebase either. Should be `printfQDecode`

mitchellh (changes requested):
One bikeshed request. The rest looks good!

src/os/inplace.zig
jcollie: Fixed.

jcollie: Fixed.

Issue #8432, opened 2025-08-27
Add support for marking a terminal as read-only


mitchellh:
Discussion: #3575

Based on: https://learn.microsoft.com/en-us/windows/terminal/panes#marking-a-pane-as-read-only

This would result in the following behavior:

* No input is sent to the pty (mouse events, key encoding)
* Input can still be used at the terminal level to do things like make selections, copy/paste (keybinds), scroll, etc.
* Warn before quit always enabled in this state even if an active process is not running

To enable this we should add a new keybind action `toggle_readonly` that would apply to the given surface.

The actual behavior should be fully implemented in `src/Surface.zig` (apprt's still send input events).

We'll also need apprt changes so the readonly state is obvious. I leave that to the apprt designers.

mitchellh:
Done by #9130

PR #9130, opened 2025-10-10
feat: add readonly surface mode

matthew-hre:
Tried my hand at #8432

Currently lacking tests and some sort of visual indicator that the surface is locked.

Also, not entirely sure if I needed to touch `application.zig`.

Found what needed changing with help from Copilot (and added Docs w/ Copilot), but wrote most of the code myself.


include/ghostty.h
```diff
@@ -762,6 +762,7 @@ typedef enum {
   GHOSTTY_ACTION_RESIZE_SPLIT,
   GHOSTTY_ACTION_EQUALIZE_SPLITS,
   GHOSTTY_ACTION_TOGGLE_SPLIT_ZOOM,
+  GHOSTTY_ACTION_TOGGLE_READONLY,
```
mitchellh: Same feedback here as your other PR, we don't need an apprt action for this.

We should have an apprt action `GHOSTTY_ACTION_READONLY` or something that sends a boolean with it that notes the current state of the surface. 

This way GUIs can eventually (not this PR) show that the terminal is in a read-only state.

src/Surface.zig
```diff
@@ -818,6 +824,9 @@ pub fn deactivateInspector(self: *Surface) void {
 /// True if the surface requires confirmation to quit. This should be called
 /// by apprt to determine if the surface should confirm before quitting.
 pub fn needsConfirmQuit(self: *Surface) bool {
+    // If the surface is in read-only mode, always require confirmation
+    if (self.readonly) return true;
```
mitchellh, line 905: Interesting, why?

src/Surface.zig
```diff
@@ -3069,6 +3086,9 @@ fn mouseReport(
     mods: input.Mods,
     pos: apprt.CursorPos,
 ) !void {
+    // If the surface is in read-only mode, do not send mouse reports to the PTY
+    if (self.readonly) return;
```
mitchellh: Probably should change this to an assert and let callers verify this.

mitchellh (changes requested):
I think the core changes look good, I didn't verify you checked in all the right places. Just note the apprt change.

src/Surface.zig
matthew-hre: From the original issue:

> "Warn before quit always enabled in this state even if an active process is not running"

I believe Windows Terminal has the same behavior. Happy to remove it!

matthew-hre:
Removed the apprt action, merged in changes from main, and cleaned up the integration with my previous pr regarding mouse reporting. Since we're breaking out the apprt into another PR down the road, I'll open this guy up for review!

mitchellh (approved):
Thank you for getting this going. I made some changes to get it over the finish line:

- I moved our readonly blocking to all write requests to termio rather than scatter it across Surface.zig
- I added an apprt action so they can detect it
- I added macOS menu items for it
- I added a macOS overlay with a popover describing the feature while its active

PR #9134, opened 2025-10-11
core/gtk: use command line supplied by the shell in notifications

jcollie:
Some shells use OSC 133;C to tell us what command is being executed. Use that information to enhance our notifications that a command has finished.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -826,9 +826,20 @@ typedef struct {
 
 // apprt.action.CommandFinished.C
 typedef struct {
-  // -1 if no exit code was reported, otherwise 0-255
+  // The command line or len==0 if the command line
+  // is unknown.
+  struct {
+    const char *ptr;
+    uint64_t len;
+  } command_line;
+  // The exit code of the command. The exit code will be a number between 0
+  // and 255 or -1 if no exit code was provided. 0 indicates that the command
+  // was successful. Any number from 1 to 255 indicates an application specific
+  // error code.
   int16_t exit_code;
-  // number of nanoseconds that command was running for
+  // How long the command took in nanoseconds. Despite the duration being
+  // reported in nanoseconds the accuracy is probably only within a few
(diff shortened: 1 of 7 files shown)


include/ghostty.h
```diff
@@ -735,9 +735,22 @@ typedef struct {
 
 // apprt.action.CommandFinished.C
 typedef struct {
-  // -1 if no exit code was reported, otherwise 0-255
+  // The command line, as reported by OSC 133;C, or null if the command line
```
mitchellh: I don't care for this PR, but as a general statement, I'd say the _source_ of apprt notifications is usually left unspecified, because if alternate methods of supplying it come about we'll adopt those too.

In this case for example, I'd say for apprt authors there doesn't have to be any awareness these come from OSC 133.

src/apprt/action.zig
```diff
@@ -747,17 +747,49 @@ pub const CloseTabMode = enum(c_int) {
 };
 
 pub const CommandFinished = struct {
+    /// The command line, as reported by OSC 133;C, or null if the command line
+    /// is unknown.
+    cmdline: ?[:0]const u8,
```
mitchellh: Bikeshed, I'd prefer something less abbreviated here, `command_line` is totally fine.

mitchellh (changes requested):
Couple comments. Still working my way through it.

include/ghostty.h
jcollie: Removed mentions of the OSCs in the comments.

src/apprt/action.zig
jcollie: Renamed.

mitchellh (changes requested):
Can you pull out the parser updates from the rest?

jcollie:
> Can you pull out the parser updates from the rest?

#11046

jcollie:
Rebased after #11046 was merged.

PR #9139, opened 2025-10-11
GTK fix quick terminal autohide

rhodes-b:
This is pretty much a direct port of the previous GTK app. still inside of the `isActive` handler for a window

https://github.com/ghostty-org/ghostty/blob/7e429d73d6af65a397c6264b18ab60609ae8eefe/src/apprt/gtk/Window.zig#L822-L837

Fixes: https://github.com/ghostty-org/ghostty/discussions/9137


mitchellh:
Thanks!

PR #9142, opened 2025-10-11
font(fix): Extract and apply Nerd Font codepoint mapping table

danielwe:
Fixes #9076

**Before** 
Screenshot 2025-10-11 at 00 07 09

**After**
Screenshot 2025-10-11 at 00 07 31

These screenshots show the chevrons mentioned in https://github.com/ghostty-org/ghostty/discussions/7820#discussioncomment-14617170, which should be scaled as a group but were not until this PR.

The added code downloads each individual symbol font file from the Nerd Fonts github repo (making sure to get the version corresponding to the vendored `font-patcher.py`) and iterates over all of them to build the correct and complete codepoint mapping table. The table is saved to `nerd_font_codepoint_tables.py`, which `nerd_font_codegen.py` will reuse if possible instead of downloading the font files again.

I'm not going to utter any famous last words or anything, but... after this, I don't think the number of remaining issues with icon scaling/alignment is _large._


mitchellh:
This reminded me to mark the tables as generated, which I did here: https://github.com/ghostty-org/ghostty/commit/6ab416376a6ba529bca55203435f61cb9ac6afbf

danielwe:
Ah, good, I actually wasn't aware of that functionality. Should probably be applied to `src/font/nerd_font_attributes.zig` too? (In fact, the tables file is only introduced in this PR, so right now it's marked generated on main even though it doesn't exist there, but I suppose is unproblematic.)

mitchellh:
Ah whoops yep, I added that too now.

mitchellh:
Looks good to me. I'm going to tentatively mark this for 1.2.3... we'll see if there are any complaints. 😄

PR #9146, opened 2025-10-11
gtk: fix clicking on desktop notifications

jcollie:
Clicking on desktop notifications sent by Ghostty _should_ cause the window that sent the notification to come to the top. However, because the notification that was sent targeted the wrong surface (apprt surface vs core surface) and the window did not call `present()` on itself the window would never be brought to the surface, the correct tab would not be selected, etc.

Fixes #9145


pluiedev (approved):
intFromPtr considered evil

mitchellh:
Thanks, looks good. I'm sending this to 1.2.3, but I notice you milestoned it 1.3. Do you have concerns?

mitchellh:
Doesn't back port cleanly, moving to 1.3.

PR #9148, opened 2025-10-11
macOS: Improve titlebar tabs even further 

### LINKED ISSUE #1691: Improve titlebar tabs when using `background-opacity`
Here is [an example](https://github.com/mitchellh/ghostty/pull/1623#issuecomment-2052225654) of the current nastiness.

bo2themax:
### Background

While I was working on another PR, I noticed that if the path is too long, then the title won't appear. That’s because that toolbar item was overflowed into an invisible menu. Even if you change to a directory that has a shorter title, the title won't appear, unless you resize the window to a size that can fit the previous long title. Then the title gets updated and becomes shorter.

This issue exists on all platforms, since they are all using the toolbar item to implement the centred title.

[Screenshot](https://github.com/user-attachments/assets/36bc0221-0f3a-4804-85ef-6e8e1e38dd07)

### Changes

To fix this, I instead of using a custom view that will be added as a subview to TitleBarView. By doing this, we not only can fix the issues above, but also can achieve a smoother transition and gain more control. In this PR, I also added the resetZoom button to this view and a tooltip to view the full title.

To reduce the maintenance burden for different platforms, I changed `TitlebarTabsVenturaTerminalWindow` to subclass `TitlebarTabsTahoeTerminalWindow`, which inherits all those fixed I made, without losing its ability to adapt to old tab bar style. But **the change is relatively huge,** if you have any doubts please let me know. I tested all the scenarios in the last fix, it works pretty well so far, also fixed #1691.

Along with this, there are also some other minor improvements:
- Support building with Xcode 16.4 (with this I can test most of the old tab bar behaviour without using a virtual machine)
- Move major title update logics to the base `TerminalWindow`
- Match the titlebar style to the tab bar more closely
- Fix title alignment with custom font with `native/transparent` titlebar

image
image
image


mitchellh:
Any chance you can split this PR up into multiple? The diff is a bit hard to read on its own.

For example, the Xcode 16 support would be nice to just have as its own PR. Thanks!

> To reduce the maintenance burden for different platforms, I changed TitlebarTabsVenturaTerminalWindow to subclass TitlebarTabsTahoeTerminalWindow, which inherits all those fixed I made, without losing its ability to adapt to old tab bar style. But the change is relatively huge, if you have any doubts please let me know. I tested all the scenarios in the last fix, it works pretty well so far, also fixed https://github.com/ghostty-org/ghostty/issues/1691.

On this point, I think I'd prefer to go the code duplication route. The reason I went the direction I went was because there was too many scenarios before where fixing an issue on one platform regressed on another. It's really nice currently to see a change in the Tahoe window and know the blast radius is _only_ Tahoe. The older macOS versions were combined into one due to legacy reasons.

I'm willing to change my mind here, but I'd be curious to see both approaches.

mitchellh:
Also, I really appreciate you fixing these hairy issues. Very helpful. ❤️

bo2themax:
> Any chance you can split this PR up into multiple? The diff is a bit hard to read on its own.
> 
> For example, the Xcode 16 support would be nice to just have as its own PR. Thanks!
> 
> > To reduce the maintenance burden for different platforms, I changed TitlebarTabsVenturaTerminalWindow to subclass TitlebarTabsTahoeTerminalWindow, which inherits all those fixed I made, without losing its ability to adapt to old tab bar style. But the change is relatively huge, if you have any doubts please let me know. I tested all the scenarios in the last fix, it works pretty well so far, also fixed #1691.
> 
> On this point, I think I'd prefer to go the code duplication route. The reason I went the direction I went was because there was too many scenarios before where fixing an issue on one platform regressed on another. It's really nice currently to see a change in the Tahoe window and know the blast radius is _only_ Tahoe. The older macOS versions were combined into one due to legacy reasons.
> 
> I'm willing to change my mind here, but I'd be curious to see both approaches.

Absolutely, I'll do that tomorrow

bo2themax:
> On this point, I think I'd prefer to go the code duplication route. The reason I went the direction I went was because there was too many scenarios before where fixing an issue on one platform regressed on another. It's really nice currently to see a change in the Tahoe window and know the blast radius is _only_ Tahoe. The older macOS versions were combined into one due to legacy reasons.
> 
 
Makes perfect sense, given so many history os versions to consider. In retrospect, I think I can separate this into: 

- Xcode 16 support (this is clean and easy to review)
- Issue fix for #1691, rc is clear now
- New custom title bar for Tahoe only
- New custom title bar for all

mitchellh:
Closing this one since you're splitting it up. Thanks!

PR #9149, opened 2025-10-11
macOS: distinguish between Debug and Release(Stable/Tip)

bo2themax:
### Background

Been running Ghostty locally for a while now, and I use the Finder service a lot. It often confuses me which one is the official one, until I actually open it.

### Changes

- Use blueprint to distinguish from release app, if no custom icon specified
- Change BundleDisplayName to Ghostty[Debug]
- Enable Info.plist preprocessing for reading `$(INFOPLIST_KEY_CFBundleDisplayName)` for providing different services with different configurations
  > (Preprocessing was once reverted before](https://github.com/ghostty-org/ghostty/commit/6508fec), so I'm not sure whether this follows the 'rules' here, but for now, there are no links in the plist file, so I think it’s [safe](https://developer.apple.com/library/archive/technotes/tn2175/_index.html#:~:text=can%20pass%20the-,%2Dtraditional,-flag%20to%20the) to enable it


mitchellh:
I admit I don't love the icon but any different icon is fine and we can fine tune it later. The rest looks good. Small prettier failure in CI. Let me know when you're ready for review on this (its marked as draft).

00-kat:
It *could* be the [blueprint icon], though that can be used for the main app at runtime too (but I'm not sure if this icon can be approximated with [`macos-icon = custom-style`][macos-icon]).

[blueprint icon]: https://github.com/ghostty-org/ghostty/blob/v1.2.2/macos/Assets.xcassets/Alternate%20Icons/BlueprintImage.imageset/macOS-AppIcon-1024px.png
[macos-icon]: https://ghostty.org/docs/config/reference#macos-icon

bo2themax:
> It _could_ be the [blueprint icon](https://github.com/ghostty-org/ghostty/blob/v1.2.2/macos/Assets.xcassets/Alternate%20Icons/BlueprintImage.imageset/macOS-AppIcon-1024px.png), though that can be used for the main app at runtime too (but I'm not sure if this icon can be approximated with [`macos-icon = custom-style`](https://ghostty.org/docs/config/reference#macos-icon)).

Hmm, good idea! I didn't think of that, I'll mark this as draft again, and take a look at it.

bo2themax:
@mitchellh @00-kat Changed it by adding an overlay now. You can also fine-tune it afterwards, and replace the new content in `AppDelegate+DebugIcon.swift`

This way, the resource bundle will not be changed.

macos/Sources/App/macOS/AppDelegate+DebugIcon.swift
```diff
@@ -0,0 +1,33 @@
...
+#else
+        appIcon
+#endif
+    }
+}
+
+#if DEBUG
+// A simple overlay, written in base64 to avoid increasing bundle size
+// https://github.com/user-attachments/assets/5b26a773-3b7a-457b-af08-139bdbeef568
+private let overlayContent = """
```
mitchellh: I don't like blobs like this, even if they're innocent it makes it easier to sneak stuff in and harder to understand. 

Can't we just get away with using a custom icon we already bundle (e.g. blueprint?). For the rare case devs use that as their primary it'd be weird but I think in most cases it'd be fine.

mitchellh (changes requested):

macos/Sources/App/macOS/AppDelegate+DebugIcon.swift
bo2themax: Works for me👻

mitchellh (approved):

PR #9157, opened 2025-10-12
Fix fish shell cursor integration in fish vi mode

XrXr:
Previously, the fish shell integration interfered with fish's builtin vi mode cursor switching configurations such as `$fish_cursor_default` and `$fish_cursor_insert`.

```console
$ ghostty --config-default-files=false -e fish --no-config --init-command 'source "$GHOSTTY_RESOURCES_DIR"/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish; fish_vi_key_bindings'
```

The above command starts fish in vi mode with Ghostty shell integrations. Manually loading the integration is necessary due to `--no-config` blocking auto injection.

 1. At the prompt, fish is in insert mode, and the cursor is a blinking beam. However, press escape and then "i" to exit then re-enter insert mode, and the cursor will be a solid beam due to the `$fish_cursor_unknown` setting. Without the shell integration, insert mode always uses a solid beam cursor.

 2. A similar problem shows if we start fish with `fish_vi_key_bindings default`. The cursor ends up as a blinking beam in normal mode only due to the shell integration interfering. This glitch can also be reset away by entering then exiting insert mode.

 3. Also, `$fish_cursor_external` has no effect when used with shell integration. After `fish_vi_key_bindings`, set it to `line`, run cat(1), and shell integration will give you a blinking block, not the asked for line/beam.

I verified that this patch makes the shell integration stop interfering in three scenarios above, and it still changes the cursor when not using fish's vi mode.

Note that `$fish_cursor_*` variables can be set when fish isn't in vi mode, so they're not great signals for the shell integration hooks.

### DIFF
--- a/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
+++ b/src/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
@@ -54,10 +54,14 @@ function __ghostty_setup --on-event fish_prompt -d "Setup ghostty integration"
     if contains cursor $features
         # Change the cursor to a beam on prompt.
         function __ghostty_set_cursor_beam --on-event fish_prompt -d "Set cursor shape"
-            echo -en "\e[5 q"
+            if not functions -q fish_vi_cursor_handle
+                echo -en "\e[5 q"
+            end
         end
         function __ghostty_reset_cursor --on-event fish_preexec -d "Reset cursor shape"
-            echo -en "\e[0 q"
+            if not functions -q fish_vi_cursor_handle
+                echo -en "\e[0 q"
+            end
         end
     end
 


rhodes-b:
im a bit confused why this is required?

ghostty offers the ability to disable shell integration features which is what it looks like this is doing
`shell-integration-features = no-cursor`
will ensure that the vi cursor mode is not broken
https://ghostty.org/docs/config/reference#shell-integration-features

note that for point 3 that setting the cursor style within ghostty itself (or if the shell was doing it) gets overriden by shell-integration and that you should use `no-cursor` option in that case this is listed in the docs
https://ghostty.org/docs/config/reference#cursor-style

mitchellh:
> im a bit confused why this is required?

I've heard complaints about the shell integration and vi mode so I suspect this makes it so our `cursor` feature places nicely with that, e.g. setting the cursor in normal mode but not in vi mode. I'm not sure if I've got that right, but that is the general intention of our `cursor` shell integration feature.

XrXr:
It's not required of course, because it's possible to turn off shell integration as mentioned. But I think fixing this improves the zero config experience.

rhodes-b:
> I've heard complaints about the shell integration and vi mode so I suspect this makes it so our `cursor` feature places nicely with that, e.g. setting the cursor in normal mode but not in vi mode. I'm not sure if I've got that right, but that is the general intention of our `cursor` shell integration feature.

point 2 makes it seem like that the shell-integration feature is also disabled for normal mode. From the description to me this makes it so when using vi mode or set `$fish_cursor_external` that you dont need to add the no-cursor option in your config

jparise (commented):
This generally makes sense to me, but I'm curious if we should potentially be doing "smarter" than just checking for the `fish_vi_cursor_handle` function?

Here's the fish script that handles vi cursor shapes: https://github.com/fish-shell/fish-shell/blob/master/share/functions/fish_vi_cursor.fish

XrXr:
> I'm curious if we should potentially be doing "smarter" than just checking for the fish_vi_cursor_handle function?

IMO since fish has native configurations for the cursor, the shell integration should try to stay transparent to avoid introducing surprises. Checking the one function is simple and robust.

jparise:
> > I'm curious if we should potentially be doing "smarter" than just checking for the fish_vi_cursor_handle function?
> 
> IMO since fish has native configurations for the cursor, the shell integration should try to stay transparent to avoid introducing surprises. Checking the one function is simple and robust.

Right. I was hoping to find some other more direct feature flag to check whether `fish_vi_cursor` is active, rather than the presence of this implementation function, but I don't see anything better in [the script](https://github.com/fish-shell/fish-shell/blob/master/share/functions/fish_vi_key_bindings.fish).

I also see that this function is erased when the vi key bindings are removed: https://github.com/fish-shell/fish-shell/blob/5f0d83d2f2f3bf45c9cff45d756ff8d9130598d3/share/functions/fish_vi_key_bindings.fish#L433C27-L433C48

... so this looks like a good approach to addressing the problem. It also seems correct to do this within these event handler functions (as you've done) so that they can response to vi mode being enabled/disabled later. I think we could also listen to some events and do this more surgically, but this approach should be plenty fast and unobtrusive for what we're trying to accomplish.

jparise (approved):

PR #9162, opened 2025-10-12
macOS: Support building with Xcode 16

bo2themax:
With this you can test most of the old tab bar behaviour without using a virtual machine


mitchellh:
I think it'd be good to have CI test this so we don't regress on it, I think the Tahoe images have Xcode 16 installed too. I'll investigate this later but for now will merge.

PR #9168, opened 2025-10-12
macOS: fix title misalignment in tabs

bo2themax:
While I was testing #9166, noticed another edge case🤯. 

This appears both in Tahoe and Sequoia👇🏻

https://github.com/user-attachments/assets/9cecea35-1241-4f31-9c15-0f2a7a6f342a


macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
```diff
@@ -205,9 +205,16 @@ class TerminalWindow: NSWindow {
 
     /// Returns true if there is a tab bar visible on this window.
     var hasTabBar: Bool {
+        // TODO: use titlebarView to find it instead
```
bo2themax, line 208: I'll change this in #9166, gonna need more time on regression test

mitchellh (approved):

PR #9177, opened 2025-10-13
deps: update libvaxis

rockorager:
Update libvaxis. The latest commit of libvaxis includes `uucode` as the unicode
library. `uucode` has a much cleaner API and is actively developed by a Ghostty
maintainer (@jacobsandlund). This also has the advantage of removing the last
transitive dependency Ghostty has that is hosted on codeberg, which separates us
from the frequent outages.

Disclosures: I used AI to debug the import issue in `boo.zig`

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -15,8 +15,8 @@
         },
         .vaxis = .{
             // rockorager/libvaxis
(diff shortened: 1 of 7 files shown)


mitchellh (approved):
Updated the extra files. Thanks!

PR #9185, opened 2025-10-13
macOS: Add GhosttyUITests (Drafting to save some ci checks)

bo2themax:
### Background

~~I was trying to add a few UI test cases for `macOS-titlebar-style`[Already in this PR]~~. In order to do this, I need a way from `GhosttyKit` to load a temporary configuration without messing around with users'.

### Changes

- Add `ghostty_config_load_file` using the existing [`loadFile`](https://github.com/ghostty-org/ghostty/blob/dafb9e89a3470674e0bfd5d9b01edbf1d87f330a/src/config/Config.zig#L3399)
- Use `xcbeautify` to format test&build errors
  **Couldn't find a way to do this in `GhosttyXcodebuild`, if you have a better approach please let me know!**
- Add GhosttyUITests target and test cases for `GhosttyTitlebarTabsUITests`(#2349)  and `GhosttyThemeTests`(#9360)

### NOTE

Running UI tests on the runner could be **very** slow and I couldn't find a way to guarantee success, so I made these only runnable by manually testing in Xcode.

Better to squash this🤪

> > Some of the test cases could fail when testing all the cases together; a rerun would succeed.


bo2themax:
@mitchellh Added two steps to enable `automation without authentication` temporarily, I tested in my fork using GitHub runners, it should work now.
image

mitchellh:
Thanks! I'll take a look soon, btw @bo2themax are you on Discord or do you have an email? Can you send me one at `m@mitchellh.com` I've got a question for you 😄

bo2themax:
> Thanks! I'll take a look soon, btw @bo2themax are you on Discord or do you have an email? Can you send me one at `m@mitchellh.com` I've got a question for you 😄 

Yeah, I'll ~~dm you on discord~~ send you one

mitchellh (approved):
Thank you! Sorry for the delay. I think the framework for this makes sense. 

In order for this to be reliable, we need to find a way to run it in CI. Otherwise, I think it'll bit rot. I agree that it feels a bit time/money expensive to run this all the time but perhaps manual triggers would work well for this as well as every commit on main (feels worth it).

mitchellh:
I'm going to merge this anyways to get it in, but will create a follow-up issue to track running in CI.

bo2themax:
> I'm going to merge this anyways to get it in, but will create a follow-up issue to track running in CI.

Thanks for taking the time! I'm thinking maybe we can find a way to detect ui related changes, then we run the check based on that.

PR #9189, opened 2025-10-13
renderer: Include arrows block in constrained symbols

danielwe:
Fixes #8693 

**Before**
Screenshot 2025-10-13 at 14 00 28

**After**
Screenshot 2025-10-13 at 14 01 14

The effect is somewhat subtle with my combination of fonts. See #8693 for the more egregious examples that this fixes.


mitchellh (approved):
I'm not 100% sure if this is the right way long term to do this (just haven't studied it enough) but I think this is simple, obvious, and limited enough as a band-aid that I'm find with it.

mitchellh:
This modifies uucode, so we have to push this to 1.3.

PR #9192, opened 2025-10-13
fix make clean: change dir name from zig-cache to .zig-cache

peterguy:
`make clean` was not removing the `.zig-cache` directory. Instead it was removing the `zig-cache` directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ vendor/glad/include/glad/glad.h: vendor/glad/include/glad/gl.h
 
 clean:
 	rm -rf \
-		zig-out zig-cache \
+		zig-out .zig-cache \
 		macos/build \
 		macos/GhosttyKit.xcframework
 .PHONY: clean


pluiedev (approved):
I don't think anyone really uses the Makefile these days since the Zig build system can do pretty much everything we needed to do. `.zig-cache` used to be just `zig-cache` until 0.13 I believe, so this is a blast in the past for sure

peterguy:
Ah, ok. I was hunting for a way to clean out all build artifacts, especially the macos ones, and found `make clean`.

jcollie:
It's a not very intuitive I think, but `zig build uninstall` is _supposed to be_ the Zig alternative to `make clean`, except that it's [unimplemented](https://github.com/ziglang/zig/issues/14943) and will panic the build.

Plus I doubt that it would clean up `.zig-cache`. macOS artifact cleanup would be doubtful too.

jcollie (approved):
In the future, someone should make sure that this cleans up all the macOS build artifacts.

danielwe:
> In the future, someone should make sure that this cleans up all the macOS build artifacts.

Doing a full macOS build from scratch (both debug and release), I don't get any artifacts other than the ones already covered.

In the past, I've also gotten these, so I guess it wouldn't hurt to add them (maybe easiest to add to this PR?):
```
macos/Ghostty.xcodeproj/xcuserdata
macos/Ghostty.xcodeproj/project.xcworkspace/xcuserdata
```

PR #9193, opened 2025-10-14
termio: reimplement OSC 7 URI handling

jparise:
This reimplements the MAC address-aware URI parsing logic used by the
OSC 7 handler and adds an additional .raw_path option that returns the
full, unencoded path string (including query and fragment values), which
is needed for compliant kitty-shell-cwd:// handling.

Notably, this implementation takes an options-based approach that allows
these additional behaviors to be enabled at runtime. It also leverages
two std.Uri.parse guarantees:

1. Return slices point into the original text string.
2. .raw components don't require unescaping (.percent_encoded does).

The implementation is in a new 'os.uri' module because its now generic
enough to not be hostname-oriented.

We use os.uri.parseUri and its parsing options to reimplement our OSC 7
file-style URI handling. This has two advantages:

First, it fixes kitty-shell-cwd scheme handling. This scheme expects the
full, unencoded path string, whereas the file scheme expects normal URI
percent encoding. This was preventing paths containing "special" URI
characters (like "path?") from working correctly in our bash, zsh, and
elvish shell integrations, which report working directories using the
kitty-shell-cwd scheme. (fish uses file URIs, which work as expected.)

Second, we can greatly simplify our hostname and path string handling
because we can now rely on the "raw" std.Uri component form to always
provide the correct representation.

Lastly, this lets us remove the previous URI-related code from the
os.hostname module, restoring its focus to hostname-related functions.

See: #5289


src/termio/stream_handler.zig
```diff
@@ -1089,44 +1089,40 @@ pub const StreamHandler = struct {
             return;
         }
 
-        const uri: std.Uri = internal_os.hostname.parseUrl(url) catch |e| {
+        // Attempt to parse this file-style URI using options appropriate
+        // for this OSC 7 context (e.g. kitty-shell-cwd expects the full,
+        // unencoded path).
+        const uri: std.Uri = internal_os.uri.parseUri(url, .{
```
mitchellh: Bikeshed: just name this `parse` because its already in the `uri` namespace

src/termio/stream_handler.zig
```diff
@@ -1089,44 +1089,40 @@ pub const StreamHandler = struct {
...
+            .mac_address = (comptime builtin.os.tag != .macos),
```
mitchellh: Don't need parens here, pretty sure.

src/termio/stream_handler.zig
```diff
@@ -1089,44 +1089,40 @@ pub const StreamHandler = struct {
...
             },
         };
 
         // OSC 7 is a little sketchy because anyone can send any value from
         // any host (such an SSH session). The best practice terminals follow
         // is to valid the hostname to be local.
-        const host_valid = internal_os.hostname.isLocalHostname(
-            hostname_from_uri,
-        ) catch |err| switch (err) {
+        const host_valid = internal_os.hostname.isLocalHostname(host) catch |err| switch (err) {
```
mitchellh: Rename this to `isLocal` (I know you didn't write it, but while we're here...)

src/termio/stream_handler.zig
```diff
@@ -1135,43 +1131,16 @@ pub const StreamHandler = struct {
...
-            } else |_| {}
-
-            // Fall back to using it directly...
-            log.warn("failed to unescape OSC 7 path, using it directly path={s}", .{path});
-            break :path .{ path, false };
-        };
-        defer if (heap) self.alloc.free(path);
+        // We need the raw path, which might require unescaping. We try to
+        // avoid making any heap allocations by using the stack first.
+        var arena_alloc = std.heap.ArenaAllocator.init(self.alloc);
```
mitchellh: Use the more idiomatic `var arena_alloc: std.heap.ArenaAllocator = .init(...)` syntax

mitchellh (changes requested):
Looks really good, just a handful of small stylistic things.

mitchellh:
Unfortunately need to push to 1.3 because of Zig 0.15 changes.

PR #9204, opened 2025-10-14
synthetic: make bytes generation more flexible

jcollie:


mitchellh (changes requested):
I'm not a fan of this style (lots of underscores).

And I want to keep `Bytes` dumb and not be aware of things like "osc" or "kv". I explicitly want downstream to handle this. 

I think you're poking at some use case here, its not clear to me, but I don't think this API is the right direction.

jcollie:
Cleaned up the underscores and moved all of the alphabet generation downstream. I'm looking at improvements to the OSC parser, but I wanted to benchmark against a wider variety of OSC sequences as performance of the OSC parser varies depending on the OSC being parsed. While adding the generation of other OSC sequences I found it helpful to have this more flexible interface to the Bytes generator.

mitchellh (approved):

PR #9240, opened 2025-10-17
libvt C Api expose paste encoding

oxelf:
As the title says, I exported the paste encoding function using the same pattern as with the key encoder. The GhosttyPasteEncoder is just a wrapper around the paste.Options and the allocator to keep track of. 
Here is an example how it can be used: 
```c
 #include <ghostty/vt.h>
#include <stdbool.h>
#include <stdio.h>
#include <string.h>

int main() {
// Create a paste encoder
GhosttyPasteEncoder encoder;
if (ghostty_paste_encoder_new(NULL, &encoder) != GHOSTTY_SUCCESS) {
  printf("Failed to create paste encoder\n");
  return 1;
}

// Enable bracketed paste mode
ghostty_paste_encoder_set_bracketed(encoder, true);

// we could use this to find out the required size, or just use a large
// enough buffer
size_t required;
char simple_paste[] = "pasted content";
char encoded[128];

if (ghostty_paste_encoder_encode(simple_paste, strlen(simple_paste), encoder,
                                 encoded, sizeof(encoded),
                                 &required) != GHOSTTY_SUCCESS) {
  printf("Failed to encode paste data\n");
  return 1;
}

printf("Encoded paste data: ");
for (size_t i = 0; i < strlen(encoded); i++) {
  if (encoded[i] == 0x1b) {
    printf("\\x1b");
  } else {
    printf("%c", encoded[i]);
  }
}

 // free resources
 ghostty_paste_encoder_free(encoder);
} 
```
(Mitchellh said in the discord that he would be interested in some help and wants small PRs so I didnt open an issue just for this).

### DIFF
--- a/include/ghostty/vt/paste.h
+++ b/include/ghostty/vt/paste.h
@@ -9,14 +9,19 @@
 
 /** @defgroup paste Paste Utilities
  *
- * Utilities for validating paste data safety.
+ * Utilities for validating paste data safety and encoding data to be pasted.
  *
  * ## Basic Usage
  *
- * Use ghostty_paste_is_safe() to check if paste data contains potentially
+ *1. Use ghostty_paste_is_safe() to check if paste data contains potentially
  * dangerous sequences before sending it to the terminal.
+ * 2. Create a GhosttyPasteEncoder instance using ghostty_paste_encoder_new().
+ * 3. Configure the encoder using ghostty_paste_encoder_set_bracketed() to set
+ *   bracketed paste mode if desired.
+ * 4. Use ghostty_paste_encoder_encode() to encode the paste data into a buffer
+ * 5. Free the encoder using ghostty_paste_encoder_free() when done.
  *
- * ## Example
+ * ## Paste Safety Check Example
  *
  * @code{.c}
  * #include <stdio.h>
@@ -39,13 +44,13 @@
  * }
  * @endcode
  *
- * Use ghostty_paste_encode() to encode
+ * Use ghostty_paste_encoder_encode() with a GhosttyPasteEncoder to encode
  * the paste data to be sent to the terminal.
  * There are two modes:
  * bracketed paste mode, which wraps the data in \x1b[200~ ... \x1b[201~,
  * and non-bracketed mode, which replaces newlines with carriage returns.
  *
- * ## Example
+ * ## Encoding Paste Data Example
  *
  * @code{.c}
  * #include <ghostty/vt.h>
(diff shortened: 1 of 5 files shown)


src/terminal/c/paste.zig
```diff
@@ -1,11 +1,89 @@
...
 
+test "alloc paste encoder" {
+    const testing = std.testing;
+    var e: Encoder = undefined;
+    try testing.expectEqual(Result.success, new(
+        &lib_alloc.test_allocator,
+        &e,
+    ));
+    free(e);
+}
```
mitchellh: We should have a full test here of the full API encoding something from Zig. It's awkward but see the other tests. Just helps ensure we have functionality coverage since we only run tests from Zig.

include/ghostty/vt/paste.h
```diff
@@ -66,6 +138,84 @@ extern "C" {
...
+ * no longer needed.
+ *
+ * @param allocator Pointer to the allocator to use for memory management, or
+ * NULL to use the default allocator
+ * @param options Pointer to store the created paste encoder handle
+ * @return GHOSTTY_SUCCESS on success, or an error code on failure
+ *
+ * @ingroup paste
+ */
+GhosttyResult ghostty_paste_encoder_new(const GhosttyAllocator* allocator,
```
mitchellh, line 155: The main design thing I ran into here is whether we want an object like this or if we just want to do a single flat `ghostty_paste_encode` that takes an option structure. For ABI compatibility, the option structure may want to be a pointer with opaque `setopt`-style functions but then I struggle with whether that complexity should just back into what you did here anyways...

That's my main hesitation on this API.

mitchellh (changes requested):
Thank you! A couple points.

include/ghostty/vt/paste.h
oxelf: When taking an option struct directly, we would either need to make it C ABI compatible or do it with opaque pointers like in this implementation. The problem with the opaque pointers is that we have to allocate. When we have just an extern struct for the options, we could let users of the api decide if they want to alloc the options or use a local reference. Complexity wise, I dont think that the current way is really that complicated. It also sounds logical that you create a PasteEnocder and set the bracketed mode on the encoder.

pluiedev: If we want to be *really* complicated we could copy Vulkan and use essentially type-erased, stack-allocated linked lists for extensions, but that is honestly quite cursed and over the top for what we're doing here

mitchellh: Yeah the reason I'm even sweating these details is because the patterns we pick here are likely to extend to what we do with other APIs. Does it _really matter_ for the paste encoding API? No. But for being consistent with the terminal API? Probably. But we can always break ABI for now because we aren't even released.

oxelf: Just to compare a little bit: 
#### setopt pros:
- smaller API surface
#### setopt cons:
- weaker typing(can that even be considered typed?)
- user has to first look at the docs to find out what type to use per opt.

#### setters pros:
- Type safe
- People are used to setters
- More easy to document what each opt does

#### setters cons:
- A lot of ghostty_* functions.

Do you guys have something to add?

mitchellh:
Note this is just blocked on some test changes. I'm fine with the API itself.

oxelf:
Everything should be correct now. One thing I found quite weird is that in the build.zig: 
```zig
        // Normal tests always test our libghostty modules
        //test_step.dependOn(test_lib_vt_step);
```
https://github.com/ghostty-org/ghostty/blob/9dc2e5978f6ee69f4d784b61865589b502c4012d/build.zig#L297
why is this commented out when it says that normal tests should always test the libghostty modules?

PR #9245, opened 2025-10-17
gtk: Scrollbars

mitchellh:
Fixes #111 

This builds on the prior work and adds scrollbars to the GTK application. After this PR, both the macOS and GTK applications support scrollbars and #111 can be closed.

## TODO

- [x] Verify that with text coming out if we manually scroll to the bottom we bind to the bottom
- [x] Valgrind checks

### DIFF
--- a/src/apprt/gtk/build/gresource.zig
+++ b/src/apprt/gtk/build/gresource.zig
@@ -46,6 +46,7 @@ pub const blueprints: []const Blueprint = &.{
     .{ .major = 1, .minor = 5, .name = "split-tree" },
     .{ .major = 1, .minor = 5, .name = "split-tree-split" },
     .{ .major = 1, .minor = 2, .name = "surface" },
+    .{ .major = 1, .minor = 5, .name = "surface-scrolled-window" },
     .{ .major = 1, .minor = 5, .name = "surface-title-dialog" },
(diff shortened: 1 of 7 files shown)


jcollie (approved):
Is `scrollbar=never` going to remain a macOS-only setting?

With `counter=1; while true; do echo "${counter}"; ((counter++)); sleep 0.1; done` running, I verified that the screen will stick to the bottom if you scroll up and then all the way back down, both with the mouse moving the scrollbar and scrolling with the keyboard. Works when scrolling to the bottom on a keypress as well.

One annoyance, if there's a scrollbar you can't select/click on the rightmost columns of text.

Valgrind looks clean based upon my testing.

mitchellh:
> Is scrollbar=never going to remain a macOS-only setting?

oh whoops no I have to add that

mitchellh:
Okay, just pushed a change that respects the `scrollbar = never` config.

jcollie (approved):

mitchellh:
> One annoyance, if there's a scrollbar you can't select/click on the rightmost columns of text.

Yeah. I'm not sure of a good solution here, happy to discuss and figure that out though. You can at least grab left of it and drag to it.

mitchellh:
Valgrind looks good on both my systems with the latest changes.

PR #9253, opened 2025-10-17
font: Default to light hinting in FreeType

danielwe:
This changes the default FreeType load target to `FT_LOAD_TARGET_LIGHT`, giving the hinter a lighter touch in line with the default behavior in most other GTK apps, and adds a load flag such that the old hinting behavior can be restored via config.

As discussed in https://github.com/ghostty-org/ghostty/issues/8674#issuecomment-3417082534. However, this doesn't close that issue, as it still doesn't respect custom Fontconfig settings. It's just a stopgap solution bringing Ghostty's defaults more in line with the typical results of not customizing Fontconfig.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -476,6 +476,11 @@ pub const compatibility = std.StaticStringMap(
 ///
 ///   * `autohint` - Enable the freetype auto-hinter. Enabled by default.
 ///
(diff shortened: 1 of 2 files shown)


qwerasd205 (approved):
LGTM pending CI

mitchellh:
I’m going to milestone this to 1.3 since it’ll probably dramatically change the default look for some people

Issue #9248, opened 2025-10-17
macOS: Always visible scrollbar lacks the correct background


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9247

<div type='discussions-op-text'>

<sup>Originally posted by **iamEvanYT** October 17, 2025</sup>
### Issue Description

When using ghostty `tip` after @mitchellh's macOS scrollbar commit, you can see that the background does not use the correct background as everything else.

### Expected Behavior

It should show the same background as everything else.

### Actual Behavior

The scrollbar's background is transparent, and does not use the background color specified.

<img width="1126" height="782" alt="Screenshot 2025-10-17 at 5 34 20 PM" src="https://github.com/user-attachments/assets/60e05045-4edc-4956-837f-73fd4c2aeec1" />

### Reproduction Steps

1. Set "Scrollbar Behavior" to "Always" in macOS System Settings
1. Open Ghostty
2. Run `for run in {1..1000}; do echo "line"; done` to get the scrollbar to appear

### Ghostty Version

```text
Ghostty 1.3.0-main+a2e95ce7b

Version
  - version: 1.3.0-main+a2e95ce7b
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS Tahoe 26.0.1

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
background-opacity = 0.9
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

mitchellh:
This is specifically when the styling is the "legacy" scrollbar (`scrollerStyle == .legacy`). I haven't found a way to reliably style this and also preserve the regular scrollbar, yet.

PR #9260, opened 2025-10-18
macOS: Always handle background rendering in `TerminalWindow`

danielwe:
Taking a crack at fixing #9248. I figured the simplest and most robust solution would be if the macOS app always uses the `TerminalWindow` background on the Swift side to render the background, while all views including terminal surfaces have clear backgrounds. That way you avoid problems due to partially overlapping views with partial transparency, which is what #9248 is ultimately about. I think the implementation of #8801 would also be simpler once rebased on this.

I'd appreciate reviews of what I have here, so I won't mark this as draft. However, the PR is NOT ready to merge, as the Ventura titlebar tabs changes are not finished. I don't have an environment for testing those at the moment. If someone who does wants to take a crack at finishing this, please go ahead! Otherwise, I'll see what I can figure out within the next couple of days.

**To do**

- [ ] Make necessary changes to Ventura titlebar tabs

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -562,19 +562,16 @@ class QuickTerminalController: BaseTerminalController {
         // Some APIs such as window blur have no effect unless the window is visible.
         guard window.isVisible else { return }
 
-        // If we have window transparency then set it transparent. Otherwise set it opaque.
-        if (self.derivedConfig.backgroundOpacity < 1) {
-            window.isOpaque = false
+        let alpha = self.derivedConfig.backgroundOpacity.clamped(to: 0.001...1)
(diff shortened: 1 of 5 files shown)


danielwe:
Screenshot 2025-10-18 at 01 34 10

qwerasd205 (commented):
This is not the way. We briefly drew the background color as a part of the apprt view but switched back because doing so causes multiple issues which are unacceptable:
- Breaks `alpha-blending`, since the blending of text with the default background is done by the system compositor instead of being controlled by us.
- Breaks the behavior of custom shaders that care about the background color.
- Less important, but this also increases the complexity for people using Ghostty code to embed a terminal in another application.

As a side note, if were *were* to go forward with this solution (which we shouldn't), controlling this behavior by branching on the OS rather than making it runtime configurable by the apprt would not be a good idea.

danielwe:
I understand, thanks for explaining! Replaced by #9291.

PR #9271, opened 2025-10-19
config/os: add support for clickable tilde paths

gsLocker:
Enables clicking on tilde-based paths like ~/.config/file.txt. When clicked, the path is expanded to the user's home directory and opened.

Related to #1972

This addresses the maintainer feedback from PR #5813 by reusing the existing os.expandHome() function rather than duplicating tilde expansion logic. The implementation is cross-platform (Linux/macOS/FreeBSD) from the start.

Implementation:
- Modified URL regex to detect ~/path patterns (excludes bare ~ and ~user)
- Removed :,. from file path character class to avoid underlining punctuation
- Added negative lookbehind (?<!\.) to prevent trailing period matches
- Reuse os.expandHome() for tilde expansion in os/open.zig
- Added safety trimming for edge case punctuation handling
- Added test cases for positive and negative scenarios

The regex matches ~/path but not bare ~ or ~user/path (which would require different expansion logic). Trailing punctuation like colons and periods are excluded from matching to avoid visual clutter in common output like "Error at ~/.bashrc:42".

### DIFF
--- a/src/os/open.zig
+++ b/src/os/open.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 const apprt = @import("../apprt.zig");
+const internal_os = @import("main.zig");
 
 const log = std.log.scoped(.@"os-open");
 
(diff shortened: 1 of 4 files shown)


src/os/open.zig
```diff
@@ -20,21 +21,33 @@ pub fn open(
     kind: apprt.action.OpenUrl.Kind,
     url: []const u8,
 ) !void {
+    // Trim trailing punctuation that may have been captured by the regex
```
mitchellh: This is not the proper place for this. This is only a _fallback_ open helper. Each OS implements its own native opening (NSWorkspace, DBus, etc.). If this is necessary, it should be done in the URL matching itself.

mitchellh (changes requested):
I have to do a bit more review but one note.

src/os/open.zig
gsLocker: Thank you for the feedback. I moved this to the URL detection layer in Surface.zig.

mitchellh (changes requested):
Repeating the hostname expansion is also surely not right. Is this AI-assisted?

gsLocker:
> Repeating the hostname expansion is also surely not right. Is this AI-assisted?

@mitchellh yeah you were right, having the same expansion code copy-pasted 4 times was me being lazy and not checking my agent's work.

I consolidated it all into one `prepareUrl()` helper that handles both OSC8 and regex links properly (OSC8 stays untouched, regex gets trimmed)

one thing - we're still expanding the same URL multiple times when you hover then click. want me to add caching? didn't wanna overcomplicate if it's not needed

mitchellh:
Apologies for the delay. The rendering for Urls changed quite a bit with our render state change. I'm still interested in the core idea here. if you're still interested too please resubmit. Thank you.

PR #9273, opened 2025-10-19
fix: fish shell integration should not update the universal `fish_user_paths` variable

unlimitedsola:
`fish_add_path` by default updates the `fish_user_paths` universal variable which makes the modification persist across shell sessions.

The integration also tries to update the `fish_user_paths` when the desired path already appears in the `PATH` environment variable but not in `fish_user_paths`. Because `fish_user_paths` will always be inserted before the inherited `PATH` env. This makes the added path unintentionally has a higher priority.

This patch fixes  the above issues by adding `--global` and `--path` options to `fish_user_paths` which limits the modification scope and ensures that the path won't be added if it already exists in `PATH`.

Ref: https://fishshell.com/docs/current/cmds/fish_add_path.html


jparise (approved):
Thanks! This all makes sense to me, and I appreciate the detailed explanation.

PR #9274, opened 2025-10-19
macOS: shed responsible process using private API

### LINKED ISSUE #9263: macOS: use a command launch helper to shed responsible process bit for power, memory, etc. tracking
When Ghostty launches the shell, macOS views it as the "responsible process." All permissions (TCC like Documents/Downloads/etc.), metrics (power reporting), etc. go through Ghostty and are represented by Ghostty.

This has led to confusion on multiple fronts:

* People ask why Ghostty is using so much power, when its a running program in the shell.
* The above also hides issues when we're actually the source of power usage (a bug!)
* People have gotten scared when "Ghostty" asks for permission to access something like a camera. We don't use any of those APIs! But when a subprocess does it gets attributed to "Ghostty." No!

Afaik this is still true for every terminal, including Terminal.app itself. But it looks like there might be a way around it as shown in the links below. The workaround requires we use `posix_spawn`. But `posix_spawn` doesn't support two critical things we need to do in our "pre-exec" state: (1) `setsid` and (2) `ioctl(TIOCSCTTY)`.

So here is the critical idea: **create a command launch helper like `ghostty +_command`** that we can `posix_spawn` to and shed responsibility, then `fork/exec` from in there to the proper final command so we can call the functions that `posix_spawn` doesn't accept.

I don't know if it'll work, but conceptually I don't see why not! No one has ever tried it afaict.

Lots of resources on this:

- https://www.qt.io/blog/the-curious-case-of-the-responsible-process
- https://steipete.me/posts/2025/applescript-cli-macos-complete-guide
- https://issues.chromium.org/issues/40620172#comment7

Copied for context:

<details>
<summary>Qt</summary>
As of some of you might remember, Mac OS X Mountain Lion (10.8) added a new Privacy tab to the Security & Privacy pane of System Preferences. 

tcc

The feature allowed granular access control to features such as contacts or location, and would ask the user to confirm attempts by the application to access them before granting access.

Most of you probably got to know this part of the OS when macOS Mojave (10.14) and Catalina (10.15) greatly expanded the number of privacy settings, resulting in a few rounds of whack-a-mole after upgrading.

tcc2

Behind the scenes
The feature is backed by a subsystem called Transparency, Consent, and Control (TCC). The permissions are mediated by the tccd service, which is backed by a couple of permission databases; one for system wide permissions in /Library/Application Support/com.apple.TCC, and one for user level permissions in ~/Library/Application Support/com.apple.TCC.

If using the System Preferences UI is not your cup of tea you can use the [tccutil](https://ss64.com/osx/tccutil.html) binary to manage the permissions, limited to revoking them, either wholesale or per application bundle ID:

❯ tccutil reset Calendar <optional-bundle-id>

To peek behind the curtains one can observe the stream of log messages produced by the com.apple.TCC subsystem:

❯ log stream --debug --predicate 'subsystem == "com.apple.TCC"'

A simple example
Unfortunately there's no uniform API to interact with the TCC subsystem, to request or manage permissions, so as a simple test-case let's try to access the files for the user's calendars:

#include <sys/types.h>#include <dirent.h>#include <stdio.h>#include <limits.h>#include <stdlib.h>

int main()
{
    char *calendarPath = 0;
    asprintf(&calendarPath, "%s/Library/Calendars", getenv("HOME"));
    if (DIR *calendarsDir = opendir(calendarPath)) {
        printf("Calendar access granted\n");
        closedir(calendarsDir);
    } else {
        printf("Calendar access denied\n");
    }
    free(calendarPath);
    return 0;
}
With the corresponding CMakeLists.txt file:

cmake_minimum_required(VERSION 3.19)
project(MyApp LANGUAGES CXX)
add_executable(MyApp MACOSX_BUNDLE main.cpp)
set_target_properties(MyApp PROPERTIES MACOSX_BUNDLE_GUI_IDENTIFIER com.company.MyApp)
Building and then running this application from Finder should pop up the following dialog:

And after choosing any of the options the result will be visible in the System Preferences:

We can reset the choice to make another attempt, by issuing:

❯ tccutil reset Calendar com.company.MyApp

Now, you might be tempted to launch the application from the terminal, to see the printf output:

❯ ./MyApp.app/Contents/MacOS/MyApp

If you do, you'll end up with this dialog instead:

What just happened? Why is iTerm now requesting access?

The responsible process
As it turns out, permissions are inherited by child processes. And when a process is about to access some protected resource, the TCC subsystem figure's out which process is the responsible one, and uses that as basis for requesting and persisting the result.

We can observe this by looking at the log stream, but an easier way is to consult launchctl:

❯ sudo launchctl procinfo $(pidof MyApp) | grep responsible
responsible pid = 1471
responsible unique pid = 1471
responsible path = /Applications/iTerm.app/Contents/MacOS/iTerm2

In this case we've launched MyApp directly from within iTerm2, and hence TCC decides that iTerm2 is responsible for MyApp. Whereas if we opened MyApp from Finder or via open, TCC would attribute the calendar access to MyApp:

❯ sudo launchctl procinfo $(pidof MyApp) | grep responsible
responsible pid = 3425
responsible unique pid = 202085
responsible path = /Users/torarne/dev/tmp/tcc/MyApp/build/MyApp.app/Contents/MacOS/MyApp

In the case of an application embedding and launching helper executables this behavior of course makes sense, but it can be a bit surprising in cases such as launching apps from the terminal.

The rabbit hole that started it all
What led me down this path was an [observation](https://codereview.qt-project.org/c/qt-creator/qt-creator/+/385288) that Qt Creator was gaining more and more permission-related usage descriptions in its own Info.plist file (a way that applications can give more context to users about why the given permissions are needed, which in some cases is a requirement to gain that permission). I was surprised by this, as I wasn't expecting Qt Creator to need access to my contacts, calendars, or photos.

As it turned out, since Qt Creator was launching user applications when running and debugging, it was effectively becoming the responsible process for all these user applications. And if one of them required a permission that needed a corresponding usage description, then the only way to make the application work was to add the description to the responsible process; Qt Creator.

This didn't sit right with me, so I started poking at how things worked in Xcode. It turned out that Xcode did not have this issue, nor did launching the application from within lldb:

❯ lldb ./MyApp.app/Contents/MacOS/MyApp
(lldb) target create "./MyApp.app/Contents/MacOS/MyApp"
Current executable set to '/Users/torarne/dev/tmp/tcc/MyApp/build/MyApp.app/Contents/MacOS/MyApp' (x86_64).
(lldb) r
Process 4277 launched: '/Users/torarne/dev/tmp/tcc/MyApp/build/MyApp.app/Contents/MacOS/MyApp' (x86_64)

❯ sudo launchctl procinfo $(pidof MyApp) | grep responsible
responsible pid = 3143
responsible unique pid = 204085
responsible path = /Users/torarne/dev/tmp/tcc/MyApp/build/MyApp.app/Contents/MacOS/MyApp

 

Somehow lldb was circumventing the logic that was deciding which process was the responsible one.

Luckily LLDB is part of the open source LLVM project, so I was able to track it down to [this](https://github.com/llvm/llvm-project/commit/041c7b84a4b925476d1e21ed302786033bb6035f) change, with the magic formula:

int responsibility_spawnattrs_setdisclaim(posix_spawnattr_t attrs, int disclaim);

By setting a special flag on the spawn-attributes before spawning, the resulting process would break free from its parent process and become responsible for its own permissions. A quick search for the function name showed that LLDB wasn't the [only](https://bugzilla.mozilla.org/show_bug.cgi?id=1576733) [one](https://chromium.googlesource.com/chromium/src/+/lkgr/base/process/launch_mac.cc) using this trick.

A quick prototype later I had a [tool](https://codereview.qt-project.org/c/qt-creator/qt-creator/+/386978/11/src/tools/disclaim/disclaim.mm) that would let Qt Creator disclaim any responsibilities for the user applications it launches, allowing us to remove all of the permission usage-descriptions from Qt Creator:

❯ disclaim ./MyApp.app/Contents/MacOS/MyApp

Calendar access granted

This feature will be released in the upcoming Qt Creator 7 release. As noted in the change log it may result in some applications needing adjustments:

Applications launched from Qt Creator will no longer inherit hardened runtime entitlements from Qt Creator, which may result in the application needing to declare its own set of entitlements if the hardened runtime is enabled.
Let us know if this causes any unexpected issues.

Postscript
Testing other IDEs, I was seeing similar issues as Qt Creator had, so this seems like a common issue for this use-case:

❯ sudo launchctl procinfo $(pidof MyApp) | grep responsible
responsible pid = 5900
responsible unique pid = 204539
responsible path = /Applications/CLion.app/Contents/MacOS/clion

❯ sudo launchctl procinfo $(pidof MyApp) | grep responsible
responsible pid = 6215
responsible unique pid = 204854
responsible path = /Applications/Visual Studio Code.app/Contents/MacOS/Electron

I'm still not sure how Xcode does things when running an application without debugging it (i.e. not going via LLDB). From looking at the process tree, the launched process is not a child of  LLDB's debugserver, like it is when debugging, but instead is a direct child of launchd. I can achieve the same with skipping the POSIX_SPAWN_SETEXEC flag, but I'm not sure if that's really what's going on.

One other option would be to use open to launch the application, as that parents it under launchd, but this only works for application bundles and not for standalone executables.

Let me know in the comments if you have any tips or tricks 😊
</details>

<details>
<summary>Swift</summary>
Remember how permission dialogs blame the hosting app (e.g. Cursor when you build an MCP) instead of your app? There’s an undocumented API to fix that: responsibility_spawnattrs_setdisclaim. I discovered this gem from [Qt’s excellent blog post about the responsible process problem](https://www.qt.io/blog/the-curious-case-of-the-responsible-process).

// Bridge the private API
@_silgen_name("responsibility_spawnattrs_setdisclaim")
func responsibility_spawnattrs_setdisclaim(
    _ attr: UnsafeMutablePointer<posix_spawnattr_t?>,
    _ disclaim: Int32
) -> Int32

// Use it to launch a subprocess that owns its permissions
func launchWithOwnPermissions(path: String, arguments: [String]) throws {
    var attr: posix_spawnattr_t?
    posix_spawnattr_init(&attr)
    defer { posix_spawnattr_destroy(&attr) }
    
    // The magic happens here
    responsibility_spawnattrs_setdisclaim(&attr, 1)
    
    var pid: pid_t = 0
    let argv = ([path] + arguments).map { strdup($0) }
    defer { argv.forEach { free($0) } }
    
    let status = posix_spawn(&pid, path, nil, &attr, argv + [nil], environ)
    guard status == 0 else {
        throw POSIXError(POSIXError.Code(rawValue: status) ?? .ENODEV)
    }
}
</details>

<details>
<summary>Chromium</summary>
I have done more research and it appears that what I wrote in #0 is pretty accurate. The Quarantine.kext tracks the “responsible” process through the MAC policy framework, and it corresponds to the proc_t::p_responsible_pid field in the kernel structure. It appears that when a process is created from launchd, it is responsible for itself. Any child processes it creates are then its responsibility. Interestingly, this means any process launched from Terminal is Terminal’s responsibility.

When using posix_spawn, the responsibility_spawnattrs_setdisclaim() call sets a bit in a macpolicyinfo action that Quarantine.kext uses to reset the responsibility chain. The new child then becomes responsible for itself (and any child processes). There is also a responsibility_set_pid_responsible_for_pid but it appears to be a privileged operation.

The Quarantine.kext also forces certain processes (named by their code signing identifier) to disclaim. Notably they are "com.apple.WebKit.Plugin.64" and "com.apple.WebKit.Plugin.32". I’ve to put together a CL to use setdisclaim on PPAPI processes and Native Messaging hosts.
</details>

mitchellh:
Fixes #9263 

This changes macOS process launching to use the private API `responsibility_spawnattrs_setdisclaim` to make the spawned process the "responsible process." This ensures that child processes spawned from Ghostty are responsible for their own TCC (Transparency, Consent, and Control) permissions and resource accounting rather than being attributed to Ghostty itself.

Concretely, this means things like Downloads folder access, Camera access, etc. all are attributed to the child process (usually the shell) rather than Ghostty. This has been a common source of confusion/concern for Ghostty, since it is understandably alarming when Ghostty asks for camera access (we have no code that uses the camera) and it was actually a child process!

This also has security implications, related to GHSA-q9fg-cpmh-c78x. More details below.

## Implementation

The `responsibility_spawnattrs_setdisclaim` private API only works with `posix_spawn`. We can't use `posix_spawn` on its own because it is missing one critical function for a terminal: the ability to set the `TIOCSCTTY` ioctl on the pty fd. 

To workaround this, we use the standard `fork/exec` mechanism we always have to launch a trampoline command `ghostty +_macos-disclaim`. This then uses `posix_spawn` with two Apple-specific, private APIs:

- `POSIX_SPAWN_SETEXEC` flag to have `posix_spawn` behave like `exec` (replace the process)
- `responsibility_spawnattrs_setdisclaim` API to disclaim responsible process attribution

> [!NOTE]
>
> This PR also includes a change so that `_`-prefixed CLI actions are hidden from help. They're considered internal-only.

## Performance

This adds extra overhead on macOS to the process start codepath which likely slows down terminal start. I haven't noticed anything, but it is an extra process and an extra syscall. 

## Security 

GHSA-q9fg-cpmh-c78x fully revolved around launched programs inheriting Ghostty's TCC permissions. I suspect that with this change, we can now provide an option to avoid asking for permission entirely and may also be able to do other heuristics to simply trust launching the program completely without permission. This PR doesn't go this far, since I want to think about this further. cc @nmggithub

### DIFF
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -409,7 +409,10 @@ pub fn getData(self: Command, comptime DT: type) ?*DT {
 
 // Copied from Zig. This is a publicly exported function but there is no
 // way to get it from the std package.
-fn createNullDelimitedEnvMap(arena: mem.Allocator, env_map: *const EnvMap) ![:null]?[*:0]u8 {
+fn createNullDelimitedEnvMap(
+    arena: mem.Allocator,
+    env_map: *const EnvMap,
+) ![:null]?[*:0]u8 {
     const envp_count = env_map.count();
     const envp_buf = try arena.allocSentinel(?[*:0]u8, envp_count, null);
 
--- a/src/cli/ghostty.zig
+++ b/src/cli/ghostty.zig
@@ -19,6 +19,7 @@ const crash_report = @import("crash_report.zig");
 const show_face = @import("show_face.zig");
 const boo = @import("boo.zig");
 const new_window = @import("new_window.zig");
+const macos_disclaim = @import("macos_disclaim.zig");
 
 /// Special commands that can be invoked via CLI flags. These are all
 /// invoked by using `+<action>` as a CLI flag. The only exception is
@@ -69,6 +70,9 @@ pub const Action = enum {
     // Use IPC to tell the running Ghostty to open a new window.
     @"new-window",
 
+    // Internal helper for posix_spawn that performs pre-exec setup
+    @"_macos-disclaim",
+
     pub fn detectSpecialCase(arg: []const u8) ?SpecialCase(Action) {
         // If we see a "-e" and we haven't seen a command yet, then
         // we are done looking for commands. This special case enables
@@ -102,6 +106,9 @@ pub const Action = enum {
             // to find this action in the help strings and output that.
             help_error => err: {
                 inline for (@typeInfo(Action).@"enum".fields) |field| {
+                    // Skip internal commands (prefixed with underscore)
+                    if (field.name[0] == '_') continue;
+
                     // Future note: for now we just output the help text directly
                     // to stdout. In the future we can style this much prettier
                     // for all commands by just changing this one place.
@@ -147,6 +154,7 @@ pub const Action = enum {
             .@"show-face" => try show_face.run(alloc),
             .boo => try boo.run(alloc),
             .@"new-window" => try new_window.run(alloc),
+            .@"_macos-disclaim" => try macos_disclaim.run(alloc),
         };
     }
 
@@ -186,6 +194,7 @@ pub const Action = enum {
                 .@"show-face" => show_face.Options,
                 .boo => boo.Options,
                 .@"new-window" => new_window.Options,
+                .@"_macos-disclaim" => macos_disclaim.Options,
             };
         }
     }
--- a/src/cli/help.zig
+++ b/src/cli/help.zig
@@ -63,6 +63,8 @@ pub fn run(alloc: Allocator) !u8 {
     );
 
     inline for (@typeInfo(Action).@"enum".fields) |field| {
+        // Skip internal commands (prefixed with underscore)
+        if (field.name[0] == '_') continue;
         try stdout.print("  +{s}\n", .{field.name});
     }
 
--- a/src/cli/macos_disclaim.zig
+++ b/src/cli/macos_disclaim.zig
@@ -13,10 +13,24 @@ pub const Options = struct {
 };
 
 /// The `_macos-disclaim` command is an internal-only Ghostty command that
-/// is only available on macOS. It uses private posix_spawn APIs to
-/// make the child process the "responssible process" in macOS so it is
-/// in charge of its own TCC (permissions like Downloads folder access or
-/// camera) and resource accounting rather than Ghostty.
+/// is only available on macOS. It acts as a trampoline that uses private
+/// posix_spawn APIs to make the child process the "responsible process" in
+/// macOS so it is in charge of its own TCC (permissions like Downloads folder
+/// access or camera) and resource accounting rather than Ghostty.
+///
+/// ## Responsible Process Concept
+///
+/// In macOS, the "responsible process" is a system-level attribution mechanism
+/// used for TCC (Transparency, Consent, and Control) permissions and resource
+/// accounting (energy, memory, etc.). When a process spawns children, those
+/// children normally inherit the parent's responsible process designation. This
+/// means when programs launched from Ghostty request permissions or consume
+/// resources, macOS attributes those actions to Ghostty itself.
+///
+/// By using the private `responsibility_spawnattrs_setdisclaim` API, we make
+/// the spawned process responsible for itself rather than being attributed to
+/// Ghostty. This ensures that TCC prompts and resource accounting are correctly
+/// associated with the actual program being run, not the terminal emulator.
 pub fn run(alloc: Allocator) !u8 {
     // This helper is only for Apple systems. POSIX in general has posix_spawn
     // but we only use it on Apple platforms because it lets us shed our
@@ -33,7 +47,7 @@ pub fn run(alloc: Allocator) !u8 {
     _ = arg_iter.skip();
     _ = arg_iter.skip();
 
-    // Collect remaining args for exec
+    // Collect remaining args for exec.
     var args: std.ArrayList(?[*:0]const u8) = .empty;
     defer args.deinit(alloc);
     while (arg_iter.next()) |arg| try args.append(alloc, arg);
@@ -47,7 +61,9 @@ pub fn run(alloc: Allocator) !u8 {
     defer posix_spawn.spawn_attr.destroy(&attrs);
     {
         try posix_spawn.spawn_attr.setflags(&attrs, .{
-            // Act like exec(): replace this process.
+            // POSIX_SPAWN_SETEXEC is a macOS extension that makes posix_spawn
+            // behave like exec(), replacing the current process image with the
+            // spawned program.
             .setexec = true,
         });
 
@@ -57,6 +73,8 @@ pub fn run(alloc: Allocator) !u8 {
         try posix_spawn.spawn_attr.disclaim(&attrs, true);
     }
 
+    // On success, this call DOES NOT RETURN because POSIX_SPAWN_SETEXEC
+    // replaces this process image. On failure, we log and return 1.
     _ = posix_spawn.spawnp(
         std.mem.span(args.items[0].?),
         null,
@@ -71,6 +89,8 @@ pub fn run(alloc: Allocator) !u8 {
         return 1;
     };
 
-    // We set the exec flag so we can't reach this point.
+    // Unreachable because POSIX_SPAWN_SETEXEC replaces this process on success.
+    // If we reach here, either the spawn failed (handled above) or the platform
+    // didn't honor SETEXEC (which would be a serious bug).
     unreachable;
 }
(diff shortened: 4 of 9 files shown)


mitchellh:
Hm, this seems to cause some problems with the built-in shells in `/bin` which fail to access all TCC-protected directories and don't ask for further permissions. I'm not sure why this is yet and I'm looking into it.

mitchellh:
Okay this has some problems: Apple "platform binaries" have a policy of automatic deny for all access requests. They don't even ask for permission! Logs at the end here. So if you run a Homebrew-installed shell, it's fine. But if you use the Apple built-in shells, nothing works! Oh my god!

I have to think about this.

```
2025-10-19 14:11:49.661700-0700 0x92a29b   Default     0x11a98d2            656    0    tccd: [com.apple.TCC:access] -[TCCDAccessIdentity staticCode]: static code for: identifier /bin/zsh, type: 1: 0xa21138c00 at /bin/zsh
2025-10-19 14:11:49.664597-0700 0x92a29b   Info        0x11a98d2            656    0    tccd: [com.apple.TCC:access] -[TCCDAccessIdentity matchesCodeRequirement:]: SecStaticCodeCheckValidity() static code (0xa21138c00) from /bin/zsh : anchor apple; status: 0
2025-10-19 14:11:49.664619-0700 0x92a29b   Info        0x11a98d2            656    0    tccd: [com.apple.TCC:access] For /bin/zsh: matches platform requirements: Yes  2025-10-19 14:11:49.664637-0700 0x92a29b   Default     0x11a98d2            656    0    tccd: [com.apple.TCC:access] Platform binary prompting is 'Deny' because: is Platform Binary
```

mitchellh:
Abandoning this effort for reasons posted here: https://github.com/ghostty-org/ghostty/issues/9263#issuecomment-3419990811 There are some things in this PR I want to revive unrelated to how we launch, and I'll pull that out into its own PR.

PR #9276, opened 2025-10-19
os: add RFC 1123-compliant hostname.isValid

jparise:
std.net.isValidHostname is currently too generous. It considers strings like ".example.com", "exa..mple.com", and "-example.com" to be valid hostnames, which is incorrect according to RFC 1123 (the currently accepted standard).

Until the standard library function is improved, we can use this local implementation that does follow the RFC 1123 standard.

I asked Claude to perform an audit of the code based on its understand of the RFC. It suggested some additional test cases and considers the overall implementation to be robust (its words) and standards compliant.

Ref: https://www.rfc-editor.org/rfc/rfc1123

### DIFF
--- a/src/os/hostname.zig
+++ b/src/os/hostname.zig
@@ -11,6 +11,91 @@ pub const LocalHostnameValidationError = error{
     Unexpected,
 };
 
+/// Validates a hostname according to [RFC 1123](https://www.rfc-editor.org/rfc/rfc1123)
+///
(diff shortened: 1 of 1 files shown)


jparise:
I plan on using this in the `+ssh-config` code, but I still have a bit more work to do there, so I'm proposing this separately.

mitchellh (approved):

mitchellh:
Tagging this for 1.2.3 since the ssh-config fix if it's simple enough I plan on backporting.

PR #9278, opened 2025-10-19
termio: use a union to represent how a process is started

mitchellh:
This cleans up some of our termio exec code by unifying process launch state into a single union type. This makes it easier to distinguish between the current two mutually exclusive modes of launching a process: fork/exec and flatpak dbus commands. 

It also ensures everyplace we touch related to process launching is forced to address every case (exhaustive switch handling). I did find one resource cleanup bug based on this cleanup, which is also fixed here. This just improves memory slightly so it's not a big deal.

If we add future ways to launch processes, we can add a new union case. For example, I originally had a `posix_spawn` option while I was experimenting with that before abandoning it (see #9274).

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -102,24 +102,17 @@ pub fn threadEnter(
     errdefer self.subprocess.stop();
 
     // Watcher to detect subprocess exit
-    var process: ?xev.Process = process: {
+    var process: ?xev.Process = if (self.subprocess.process) |v| switch (v) {
+        .fork_exec => |cmd| try xev.Process.init(
+            cmd.pid orelse return error.ProcessNoPid,
+        ),
+
         // If we're executing via Flatpak then we can't do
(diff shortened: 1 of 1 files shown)


src/termio/Exec.zig
```diff
@@ -587,10 +582,18 @@ const Subprocess = struct {
     grid_size: renderer.GridSize,
     screen_size: renderer.ScreenSize,
     pty: ?Pty = null,
-    command: ?Command = null,
-    flatpak_command: ?FlatpakHostCommand = null,
+    process: ?Process = null,
     linux_cgroup: Command.LinuxCgroup = Command.linux_cgroup_default,
```
mitchellh, line 586: Considering making this part of the fork/exec state.

mitchellh: This actually gets set before the process so it has nothing to do with that so not going to do that.

PR #9282, opened 2025-10-20
feat: added mouse-reporting / toggle-mouse-reporting

### LINKED ISSUE #8430: Add `mouse-reporting` configuration to disable all mouse reports
From #5019

This is a feature that macOS Terminal.app. This has been requested multiple times and is simple enough that we should do it. This should disable all mouse reporting even if it's requested by the terminal application. It should default to true.

```
mouse-reporting = false
```

macOS terminal also allows this to be toggled, so we should also introduce a `toggle_mouse_reporting` keybind action.

matthew-hre:
Closes #8430

A few questions:
* Should I set a default keybind for `toggle-mouse-reporting`? The issue mentioned one, it's currently unset.
* Am I handling the `toggle-mouse-reporting` action properly in `performAction` (gtk) / `action` (macos)?

Copilot was used to understand the codebase, but code was authored manually.


src/apprt/action.zig
```diff
@@ -301,6 +301,10 @@ pub const Action = union(Key) {
     /// A command has finished,
     command_finished: CommandFinished,
 
+    /// Toggle mouse reporting on or off. When off, mouse events are not
+    /// reported to terminal applications even if they request it.
+    toggle_mouse_reporting,
```
mitchellh: We don't need an apprt action if apprt's done' need to react to this.

I think in the future (not this PR), it'd be useful to have an apprt action that maybe notifies the apprt that mouse reporting has changed, though.

The reason I say to not do that in this PR is I think that's a bigger discussion where we can decide how we can report config vs. terminal mouse reporting because it would be useful for GUIs to show that the mouse is currently being captured.

src/Surface.zig
```diff
@@ -3107,6 +3109,9 @@ fn mouseReport(
     mods: input.Mods,
     pos: apprt.CursorPos,
 ) !void {
+    // If mouse reporting is disabled, do nothing
+    if (!self.config.mouse_reporting) return;
```
mitchellh: I don't think this is good enough or correct.

I'd change this to an assert and change all the upstream callers to check for this state.

To make things cleaner, I'd probably add a `fn isMouseReporting` that checks both the config and the terminal state. 

This is required so that text selection and so on as you said continues to work.

mitchellh (changes requested):
Overall I like the idea for this. I think it'll be primarily useful via the command palette, especially. There are some simplifications and changes that need to happen though.

Thank you!

src/apprt/action.zig
matthew-hre: Great point, I'll cut out the apprt for this pr

src/Surface.zig
matthew-hre: Good call, I've broken out the `self.io.terminal.flags.mouse_event` check and the `self.config.mouse_reporting` check into an `isMouseReporting` function, and used that in place of some of the `self.io.terminal.flags.mouse_event != .none` checks done previously.

mitchellh (approved):
Well done. Thank you.

trollkarlen:
It does not mention a key binding to toggle this, it there one, if not how do I set one ?
Inte the terminal.app they use the FN key.

PR #9290, opened 2025-10-20
os: handle nil languageCode/countryCode in setLangFromCocoa

jaredgizersky:
Fixes a crash when NSLocale returns nil for languageCode or countryCode properties. This can happen when the app launches without locale environment variables set.

The crash occurs at `src/os/locale.zig:87-88` when trying to call `getProperty()` on a nil object. The fix adds a null check and falls back to `en_US.UTF-8` instead of dereferencing null.

## Testing
Tested by running with locale variables unset:
```bash
unset LC_ALL && ./zig-out/Ghostty.app/Contents/MacOS/ghostty
```

Before: segmentation fault  
After: launches successfully with fallback locale

### DIFF
--- a/src/os/locale.zig
+++ b/src/os/locale.zig
@@ -83,6 +83,13 @@ fn setLangFromCocoa() void {
     const lang = locale.getProperty(objc.Object, "languageCode");
     const country = locale.getProperty(objc.Object, "countryCode");
 
+    // If either property is nil, we can't construct a valid locale string.
+    // Return early and let the normal locale resolution continue.
+    if (lang.value == null or country.value == null) {
+        log.debug("languageCode or countryCode is nil", .{});
+        return;
+    }
+
     // Get our UTF8 string values
     const c_lang = lang.getProperty([*:0]const u8, "UTF8String");
     const c_country = country.getProperty([*:0]const u8, "UTF8String");
--- a/src/os/locale.zig
+++ b/src/os/locale.zig
@@ -83,10 +83,8 @@ fn setLangFromCocoa() void {
     const lang = locale.getProperty(objc.Object, "languageCode");
     const country = locale.getProperty(objc.Object, "countryCode");
 
-    // If either property is nil, we can't construct a valid locale string.
-    // Return early and let the normal locale resolution continue.
     if (lang.value == null or country.value == null) {
-        log.debug("languageCode or countryCode is nil", .{});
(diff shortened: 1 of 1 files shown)


src/os/locale.zig
```diff
@@ -83,6 +83,13 @@ fn setLangFromCocoa() void {
     const lang = locale.getProperty(objc.Object, "languageCode");
     const country = locale.getProperty(objc.Object, "countryCode");
 
+    // Check if either lang or country is nil - if so, fallback to en_US
+    if (lang.value == null or country.value == null) {
+        log.warn("languageCode or countryCode is nil, falling back to en_US.UTF-8", .{});
+        _ = internal_os.setenv("LANG", "en_US.UTF-8");
```
jparise: Given that this code path is already conditional/optional, I wonder if it might be better to not do anything (just bail or return an error).

https://github.com/ghostty-org/ghostty/blob/2696d50ca42003d648aa0172315b6f4d3b66443e/src/os/locale.zig#L20-L29

src/os/locale.zig
```diff
@@ -83,6 +83,13 @@ fn setLangFromCocoa() void {
     const lang = locale.getProperty(objc.Object, "languageCode");
     const country = locale.getProperty(objc.Object, "countryCode");
 
+    // If either property is nil, we can't construct a valid locale string.
+    // Return early and let the normal locale resolution continue.
+    if (lang.value == null or country.value == null) {
+        log.debug("languageCode or countryCode is nil", .{});
```
jparise: `log.warn` is probably the right level, given that this is an exceptional case.


```suggestion
        log.warn("languageCode or countryCode not found. Locale may be incorrect.", .{});
```

We could also use this pattern if we want to be even more specific:

```zig
const lang = locale.getProperty(objc.Object, "languageCode") orelse {
    log.warn("languageCode not found. Locale may be incorrect.", .{});
    return;
}
const country = locale.getProperty(objc.Object, "countryCode") orelse {
    log.warn("countryCode not found. Locale may be incorrect.", .{});
    return;
}
```

I don't have a strong preference but others might.

jcollie: Yes, I think that this would be preferable.

jaredgizersky: Good point, I updated it to match that pattern.

mitchellh (approved):

PR #9291, opened 2025-10-21
Add scrollbar margin to surfaces

### LINKED ISSUE #9248: macOS: Always visible scrollbar lacks the correct background

danielwe:
Alright, second attempt to fix #9248. This gives surfaces a right-hand margin that is distinct from padding, solely for keeping the terminal content out of the way of scrollbars. That way, the surface itself can fill the whole SurfaceScrollView and keep drawing the background behind the transparent scrollbar.

On the macOS app side, this means that `surfaceView` cannot be a child view of `documentView`, since the latter is automatically resized to avoid overlap with the scrollbar. Instead, `surfaceView` is now a direct subview of the `SurfaceScrollView`, stacked behind `scrollView`. This actually simplifies the scrollbar implementation quite a bit, since the origin of the `surfaceView` frame no longer has to change as we scroll, as it's relative to the fixed `SurfaceScrollView` rather than the moving `documentView`. That means we no longer have to listen and synchronize to bounds changes on the `contentView`. The only time we need to worry about the `surfaceView` frame is when there are layout changes.

This also provides #9255 for free. It would be _more_ work to condition the margin on whether the legacy scrollbar is visible or hidden.

Along the way, I fixed a bug where the layout recalculation wasn't taking into account the change of padding around the terminal grid when the surface is resized. This change must be reflected in the height of the `documentView` even if the number of rows haven't changed, otherwise the scrollbar will pop up out of nowhere when resizing an empty surface (see #9241).

Also committed some drive-by simplification of the padding math from #9241.

---

I hope this can be a step towards addressing the concern I voiced here: https://github.com/ghostty-org/ghostty/discussions/9254#discussioncomment-14716483. Basically, I think the terminal/surface should ignore the scrollbar margin for the alternate screen (or, more generally, when max scrollback is zero). With this PR, all that could be done in the core without the macOS app's involvement. But that's a future PR.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -36,6 +36,9 @@ class SurfaceScrollView: NSView {
         scrollView.usesPredominantAxisScrolling = true
         // hide default background to show blur effect properly
         scrollView.drawsBackground = false
+        // don't let the content view clip it's subviews, to enable the
+        // surface to draw the background behind non-overlay scrollers
+        scrollView.contentView.clipsToBounds = false
         
         // The document view is what the scrollview is actually going
         // to be directly scrolling. We set it up to a "blank" NSView
@@ -142,6 +145,11 @@ class SurfaceScrollView: NSView {
         
         // Fill entire bounds with scroll view
         scrollView.frame = bounds
+        surfaceView.frame.size = scrollView.bounds.size
+
+        // The width of the documentView itself doesn't really matter,
+        // as it doesn't clip subviews, but we might as well be honest
+        documentView.frame.size.width = scrollView.bounds.width
         
         // When our scrollview changes make sure our scroller and surface views are synchronized
         synchronizeScrollView()
@@ -154,10 +162,9 @@ class SurfaceScrollView: NSView {
         // Only update the pty if we have a valid (non-zero) content size. The content size
         // can be zero when this is added early to a view, or to an invisible hierarchy.
         // Practically, this happened in the quick terminal.
-        let width = surfaceContentWidth()
-        let height = surfaceView.frame.height
-        if width > 0 && height > 0 {
-            surfaceView.sizeDidChange(CGSize(width: width, height: height))
+        let size = surfaceView.frame.size
+        if size.width > 0 && size.height > 0 {
+            surfaceView.sizeDidChange(size, scrollerInset: scrollerWidth())
         }
     }
     
@@ -175,20 +182,13 @@ class SurfaceScrollView: NSView {
     /// so the renderer only needs to render what's currently on screen.
     private func synchronizeSurfaceView() {
         let visibleRect = scrollView.contentView.documentVisibleRect
-        surfaceView.frame = visibleRect
+        surfaceView.frame.origin = visibleRect.origin
     }
 
     /// Sizes the document view and scrolls the content view according to the scrollbar state
     private func synchronizeScrollView() {
-        // We adjust the document height first, as the content width may depend on it.
+        // Update the document height to give our scroller the correct proportions
         documentView.frame.size.height = documentHeight()
-
-        // Our width should be the content width to account for visible scrollers.
-        // We don't do horizontal scrolling in terminals. The surfaceView width is
-        // yoked to the document width (this is distinct from the content width
-        // passed to surfaceView.sizeDidChange, which is only updated on layout).
-        documentView.frame.size.width = scrollView.contentSize.width
-        surfaceView.frame.size.width = scrollView.contentSize.width
         
         // Only update our actual scroll position if we're not actively scrolling.
         if !isLiveScrolling {
@@ -310,22 +310,17 @@ class SurfaceScrollView: NSView {
     /// in advance, because legacy scrollbars change our contentSize and force reflow
     /// of our terminal which is not desirable.
     /// See: https://github.com/ghostty-org/ghostty/discussions/9254
-    private func surfaceContentWidth() -> CGFloat {
-        let contentWidth = scrollView.contentSize.width
+    private func scrollerWidth() -> CGFloat {
         if scrollView.hasVerticalScroller {
             let style =
                 scrollView.verticalScroller?.scrollerStyle
                 ?? NSScroller.preferredScrollerStyle
-            // We only subtract the scrollbar width if it's hidden or not present,
-            // otherwise its width is already accounted for in contentSize.
-            if style == .legacy && (scrollView.verticalScroller?.isHidden ?? true) {
-                let scrollerWidth =
-                    scrollView.verticalScroller?.frame.width
+            if style == .legacy {
+                return scrollView.verticalScroller?.frame.width
                     ?? NSScroller.scrollerWidth(for: .regular, scrollerStyle: .legacy)
-                return max(0, contentWidth - scrollerWidth)
             }
         }
-        return contentWidth
+        return 0
     }
 
     /// Calculate the appropriate document view height given a scrollbar state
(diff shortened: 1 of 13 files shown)


danielwe:
Screenshot 2025-10-21 at 00 33 54

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -403,20 +407,26 @@ extension Ghostty {
...
+            let xScale = scaledSize.width / size.width
+            setSurfaceSize(
+                width: UInt32(scaledSize.width),
+                height: UInt32(scaledSize.height),
+                scroller_width: UInt32(xScale * scrollerWidth),
+            )
         }
 
-        private func setSurfaceSize(width: UInt32, height: UInt32) {
+        private func setSurfaceSize(width: UInt32, height: UInt32, scroller_width: UInt32) {
```
jparise: nit: `scrollerWidth`

src/Surface.zig
```diff
@@ -2127,13 +2134,27 @@ fn resize(self: *Surface, size: rendererpkg.ScreenSize) !void {
...
+fn updatePadding(self: *Surface, scrollbar_width: ?u32) void {
+    const new_scrollbar_width = new: {
+        if (scrollbar_width) |width| {
+            self.scrollbar_width = width;
+            break :new true;
+        } else {
+            break :new false;
+        }
+    };
+    if (!new_scrollbar_width and !self.config.window_padding_balance) return;
```
jparise: I think you could check `window_padding_balance` in the else above and then avoid needing the `new_scrollbar_width` flag entirely.

mitchellh (changes requested):
I haven't read the full thing yet, but I think we don't need to introduce a dedicated scroller width awareness into core. I'd rather allow the apprt to document "safe areas" similar to macOS and let the core reuse that concept. This would cover more than just one side (right) and allow padding any side. 

Not exactly sure what that API looks like yet.

macos/Sources/Ghostty/SurfaceView_AppKit.swift
danielwe: Good catch!

src/Surface.zig
danielwe: Ah, of course!

danielwe:
> I think we don't need to introduce a dedicated scroller width awareness into core. I'd rather allow the apprt to document "safe areas" similar to macOS and let the core reuse that concept.

I considered a more general implementation along these lines, but the abstraction felt premature (YAGNI et cetera) so I went with the minimal solution. In any case, I would argue that it's desirable for the core to know that an unsafe area is scrolling specific in order to address the following:

> I think the terminal/surface should ignore the scrollbar margin for the alternate screen (or, more generally, when max scrollback is zero).

I find it surprisingly grating to have this superfluous margin in a text editor. The effect depends on `window-padding-color`, but it's not good either way:

Screenshot 2025-10-21 at 09 13 25
Screenshot 2025-10-21 at 09 26 46

Solving this problem either requires the core to know that an unsafe area can be ignored for terminal screens that don't have scrollback, or requires the apprt to know whether or not the active terminal screen has scrollback. I thought the former seemed cleaner.

danielwe:
> Solving this problem either requires the core to know that an unsafe area can be ignored for terminal screens that don't have scrollback, or requires the apprt to know whether or not the active terminal screen has scrollback. I thought the former seemed cleaner.

I take this back. Playing around I realize that _by far_ the simplest way of keeping this in sync is use a sentinel value in the scrollbar state, `scrollbar.total == 0`, to indicate that the current screen isn't scrollable at all, and let the apprt adjust its safe area accordingly.

EDIT: This can be done completely independently of this PR, so I submitted #9295 separately.

danielwe:
I factored out the pure bugfix part of this PR into #9296. Converting this to draft for now.

danielwe:
OK, I've pushed a revised approach. Here are some key points, probably repeating a few things from earlier, but that way you don't have to scroll(!) back and reread.

* The apprt can now tell the core to add edge insets (margins/unsafe areas) to any edge of the surface using a struct `SurfaceEdgeInsets`. In fact, `surface.sizeCallback` now requires this argument, but the default is all-zero, so apprts that don't use it can just pass `.{}`.
* The C API provides functions both with and without the inset argument, so you don't have to write out redundant zero structs when using this API without insets (e.g., the iOS/UIKit app).

The macOS app uses this API to make room for the legacy scroller overlaid on the surface. For this to work properly, `surfaceView` can _not_:

* Be a subview of `scrollView.{contentView,documentView}`, as it would be clipped or resized to not overlap with the scrollbar, defeating the purpose.
* Be a direct subview of `SurfaceScrollView`, as it would either be drawn over the entire `scrollView`, obscuring the scroller and blocking it from mouse events, or under the entire `scrollView`, which means you wouldn't be able to focus the surface by clicking on it, due to how `hitTest` works. Also it could end up behind some background styling views of the kind discussed in #9317.

The solution is for `surfaceView` to be a subview of `scrollView`, sandwiched below the `NSScroller`s but above `scrollView.contentView`. To this end I implemented a subclass of `NSScrollView` that overrides `addSubview` to maintain this invariant. This class also subsumes the subview filtering from #9317, as we can take care of that too in the overridden `addSubview`.

Moving `surfaceView` out of the `contentView/documentView` hierarchy has the side benefit that it now has a fixed frame and no longer requires any synchronization logic to compensate for the scrolling of its superviews.

Naming is hard. Looking forward to the bikeshedding.

Btw. this includes the fix from #9296 because it would conflict anyway. ~~If #9296 is merged first I'll rebase.~~ Done.

danielwe:
Pushed a rebase after the #9296 merge

macos/Sources/Ghostty/SurfaceScrollView.swift
```diff
@@ -293,3 +244,61 @@ class SurfaceScrollView: NSView {
...
+fileprivate class SurfaceScrollSubView: NSScrollView {
+    private var titlebarIsHidden: Bool = false {
+        didSet {
+            if titlebarIsHidden {
+                subviews = subviews.filter({ !$0.className.contains("NSScrollPocket") })
+            }
+        }
+    }
+    override func viewDidMoveToWindow() {
+        titlebarIsHidden = window is HiddenTitlebarTerminalWindow
```
bo2themax: Why is `super.viewDidMoveToWindow()` missing here?

danielwe:
Marking this as draft until #9446 is reviewed and merged, as that one does some cleanup that will ultimately simplify and improve this, but will require a rewrite.

PR #9295, opened 2025-10-21
macOS: Use sentinel scrollbar state to disable scroller margin

danielwe:
This adresses the issue of having a scrollbar margin (with the legacy scroller style on macOS) even when the terminal state is explicitly non-scrollable, such as in the alternate screen. I reported this in https://github.com/ghostty-org/ghostty/discussions/9254#discussioncomment-14716483 and more extensively in https://github.com/ghostty-org/ghostty/pull/9291#issuecomment-3427521889.

**Before**
Screenshot 2025-10-21 at 12 25 58

**After**
Screenshot 2025-10-21 at 12 20 30

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.Action.swift
+++ b/macos/Sources/Ghostty/Ghostty.Action.swift
@@ -108,6 +108,7 @@ extension Ghostty.Action {
         let total: UInt64
         let offset: UInt64
         let len: UInt64
+        var active: Bool { return total >= len }
         
         init(c: ghostty_action_scrollbar_s) {
             total = c.total
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -32,7 +32,7 @@ class SurfaceScrollView: NSView {
         scrollView = NSScrollView()
         scrollView.hasVerticalScroller = false
         scrollView.hasHorizontalScroller = false
-        scrollView.autohidesScrollers = false
+        scrollView.autohidesScrollers = !(surfaceView.scrollbar?.active ?? true)
         scrollView.usesPredominantAxisScrolling = true
         // hide default background to show blur effect properly
         scrollView.drawsBackground = false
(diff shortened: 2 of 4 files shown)


mitchellh:
I do question if this is the right solution, but beyond that I also wonder if this will cause reflow since resizing the terminal resizes BOTH screens and we have no way to only resize one screen currently (I'm open to the idea of that in general but its probably filled with a bunch of edge cases).

danielwe:
Yeah if you have a lot of wrapped text on the primary screen, you can sometimes see a split second reflow happening as you exit or stop (ctrl+z) neovim, but everything seems to work itself out correctly regardless of shell and shell integration. I guess the unhappy edge case would be if you have a complex, width-sensitive shell prompt but your shell/shell-integration doesn't support repainting it on reflow/resize---then the first prompt when exiting the alternate screen would be off.

Conversely you can sometimes notice the resize event as neovim is launched if you're paying close attention.

Agree these things are not ideal, but browsing the code it looks like it would be a pretty complex change to have separate primary and alternate screen sizes. But probably the better solution if it can be made to work.

danielwe:
> looks like it would be a pretty complex change to have separate primary and alternate screen sizes. But probably the better solution if it can be made to work.

But to bring back the discussion from #9291, I think this approach would require the core to be aware of the scrollbar margin, or more generally, to know which parts of the apprt insets/safe areas are scrolling related. The decision about whether to heed the margin would then happen in `Screen.init`, depending on whether `max_scrollback > 0`.

danielwe:
Marking this as draft until #9446 is reviewed and merged, as that one does some cleanup that will ultimately simplify and improve this, but will require a rewrite.

danielwe:
Rebased and ready for reivew (or thoughts about what a better approach could look like).

danielwe:
Here's a screen recording showing the split-second reflow when exiting the alternate screen and having lots of wrapped content on the primary screen. I don't mind it myself, but it's clearly noticeable.

https://github.com/user-attachments/assets/335f62cf-919a-4c08-9cc7-066576a05e6f

danielwe:
Just dropping some breadcrumbs: I've grown to prefer the solution proposed in #9610 over the one implemented here. But I'll leave this PR open while waiting for feedback on that.

mitchellh:
Let's go with #9610. Thank you!

PR #9296, opened 2025-10-21
macOS: Adjust documentView padding on layout changes

danielwe:
Fixes a bug described in #9291, where resizing an empty window causes the scrollbar to appear even though the window remains larger than the total content, because the relayouting fails to account for the change in padding around the cell grid.

To reproduce the issue:
1. Enable legacy scrollbars (System Settings -> Appearance -> Show scroll bars -> Always)
2. Open Ghostty
3. Vertically resize the window to make it smaller. The scrollbar will pop up, and as you drag the window edge, it will cycle between a maximum offset and zero depending on how far the resize is from an integer multiple of the cell height.

With this PR you'll still see the scrollbar flicker while resizing, but when you stop it will always disappear. Haven't figured out how to get rid of the flickering yet. I tried to condition various updates on the window not being in a live resize, but so far no luck.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -174,10 +174,19 @@ class SurfaceScrollView: NSView {
             }
         }
         
-        // Keep document width synchronized with content width
+        // Keep document width synchronized with content width, and
+        // recalculate the height of the document view to account for the
+        // change in padding around the cell grid due to the resize.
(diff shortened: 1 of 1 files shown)


bo2themax:
Looks good, and tested locally. Keeping the size of SurfaceView/DocumentView in sync is tricky if you resize it really fast. I couldn't think of any workaround either.

danielwe:
Rebased to silence the failing FreeBSD test

bo2themax (approved):

PR #9301, opened 2025-10-22
lib-vt: enable freestanding wasm builds

mitchellh:
This makes `libghostty-vt` build for freestanding wasm targets (aka a browser) and produce a `ghostty-vt.wasm` file. This exports the same C API that libghostty-vt does.

This commit specifically makes the changes necessary for the build to build properly and for us to run the build in CI. We don't yet actually try using it...


mitchellh:
image

I slopped some stuff together with AI and verified this does actually work (not pushing this slop example). There are definitely some DX issues that we need to iron out and I will look into that but the goal of this PR was to make it compile and ensure we keep it compiling via CI.

Issue #8387, opened 2025-08-25
Add `goto_window:next/previous` binding action


mitchellh:
From #3193

This would work similar to `next_tab`/`previous_tab`, namely it would cycle if it reaches the end. We use the more modern `goto_x:param` format which matches `goto_split`.

For semantics, the behavior should be deterministic, but otherwise the exact order is up to the apprt. For macOS, I think the simplest thing to do would get all the apps windows, find the index of the current window, and add one (modulo when necessary).

This should not be bound by default, this should not be in any menu items.

As always, happy to accept a single apprt to start.

brtmax:
Is this one still available? I would be happy to contribute! Would be my first contribution to this project as well

mitchellh:
Yes, feel free to pick it up. 😄

brtmax:
Hi, first of all since I haven't mentioned it before: Thanks a lot for all the great work on Ghostty so far, I have been using and enjoying it a lot since release!

I did a little more digging on this issue and have some follow-up questions: 

As far as I am aware, Ghostty does not currently maintain a global list of windows, and gtk does not return the windows returned through `gtk.Window.listToplevels()` in any order. To not have to introduce global state and to keep Window management as much on GTK's side as possible, I found that the method [g_object_set_data](https://docs.gtk.org/gobject/method.Object.set_data.html) could be used to assign some ID/numbering to a window on creation. That way the metadata would rest within the object, and when doing goto_window:{next,previous} all windows could be retrieved and their ID could be used to determine the next one. 

Since this feature involves some overall design related additions, I wanted to ask for some quick feedback on this before I start writing the code. Also, I am fairly new to gtk, so if there is a better and more obvious choice that I am unaware of, I apologize. 

Should the ordering of windows used for the cycling be based on the window creation order, or on the active history of windows? I looked at the discussion in #3193 and since both iTerm2 and wez were mentioned, I looked at how they order it but could not get a conclusive answer

Thanks again!

jcollie:
> Should the ordering of windows used for the cycling be based on the window creation order, or on the active history of windows? I looked at the discussion in [#3193](https://github.com/ghostty-org/ghostty/discussions/3193) and since both iTerm2 and wez were mentioned, I looked at how they order it but could not get a conclusive answer

Unfortunately, we're probably going to have people want both least recently used as well as strict creation order. Strict creation order will probably be easier to implement though.

[...]

PR #9326, opened 2025-10-23
Add goto_window:next/previous binding action

brtmax:
Related to #8387 

This implements cycling through windows (next/previous) by using the list gtk provides, as mentioned in the discussion in #8387 . I tried to follow the same pattern as 'goto_tab' and 'goto_split'. It presents the next target window and then transfers the focus to its active surface.

Sorry, this took some time. Also, this is a GTK implementation only, I currently don't have access to Mac hardware. Hope this is enough to start.

**AI Disclosure:** 
I used deepwiki.com to familiarize myself with the repository and look for patterns to reuse.


src/apprt/gtk/class/application.zig
```diff
@@ -1990,6 +1868,43 @@ const Action = struct {
...
+    ) bool {
+        const glist = gtk.Window.listToplevels();
+        defer glist.free();
+
+        const node = glist.findCustom(null, findActiveWindow);
+
+        const target_node = switch (direction) {
+            .next => node.f_next orelse glist,
+            .previous => node.f_prev orelse last: {
+                var current = glist;
```
jcollie, line 2006: Why not make use of the [first](https://docs.gtk.org/glib/type_func.List.first.html) and [last](https://docs.gtk.org/glib/type_func.List.last.html) list functions?

src/input/Binding.zig
```diff
@@ -884,6 +847,11 @@ pub const Action = union(enum) {
         right,
     };
 
+    pub const WindowDirection = enum {
```
jcollie: Is there a reason this is named differently here? Why not keep it `GotoWindow` as well?

src/input/command.zig
```diff
@@ -340,6 +339,19 @@ fn actionCommands(action: Action.Key) []const Command {
...
         },
 
+        .goto_window => comptime &.{
+            .{
+                .action = .{ .goto_window = .previous },
+                .title = "Focus Window: Previous",
+                .description = "Focus the previous window, if any.",
+            },
+            .{
+                .action = .{ .goto_window = .previous },
```
jcollie: Using the wrong action here.

src/apprt/action.zig
```diff
@@ -319,6 +316,7 @@ pub const Action = union(Key) {
         move_tab,
         goto_tab,
         goto_split,
+        goto_window,
```
jcollie, line 325: This needs to be added to the corresponding `enum` in `ghostty.h` as well.

src/Surface.zig
```diff
@@ -4975,6 +4819,17 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
             },
         ),
 
+        .goto_window => |direction| return try self.rt_app.performAction(
+            .{ .surface = self },
+            .goto_window,
+            switch (direction) {
+                inline else => |tag| @field(
```
jcollie: The inline else seems like overkill here since there are only two values and it's unlikely to ever change.

jcollie (changes requested):
Overall very good, but a few things need to be cleaned up. Thank you for the AI disclosure.

src/apprt/gtk/class/application.zig
brtmax: Thanks for all the feedback! I believe the Zig wrapper for this does not expose those two functions, so in order to use them I would have to import the C headers and then cast the wrapper type to the C pointer type. This is my first real Zig project, maybe I got it wrong. I thought the current way would maybe be cleaner. How should I proceed?

src/input/Binding.zig
brtmax: My bad, overlooked this from an earlier version. Changed it to GotoWindow now

src/input/command.zig
brtmax: Thanks, should be fixed now

src/apprt/action.zig
brtmax: Added

src/Surface.zig
brtmax: Replaced this with directly handling next/previous cases

bo2themax:
@brtmax there are some conflicts now

brtmax:
Sorry, I forgot the upstream changes

bo2themax:
> Sorry, I forgot the upstream changes

Nothing to be sorry about, it happens all the time. Thanks for your contribution, looks good to me, but let’s wait for the gtk or terminal team to approve this.

src/apprt/gtk/class/application.zig
pluiedev: They very much are present: https://ianjohnson.dev/zig-gobject/#glib2.List.last
Although I will be honest and say that the function signatures suck (in the C API, an empty list is represented as a `null` so all list-related APIs should be nullable), so it might be better to just do it in pure Zig instead.

src/apprt/gtk/class/application.zig
```diff
@@ -1990,6 +1992,43 @@ const Action = struct {
...
+                while (current.f_next) |next| {
+                    current = next;
+                }
+                break :last current;
+            },
+        };
+        const gtk_window: *gtk.Window = @ptrCast(@alignCast(target_node.f_data orelse return false));
+        gtk.Window.present(gtk_window);
+
+        const ghostty_window: *Window = @ptrCast(@alignCast(gtk_window));
```
pluiedev: Don't do this for GTK wrapper types. Use a type-safe alternative like `gobject.ext.as` when upcasting to a parent type (and it's comptime known that this is safe), and `gobject.ext.cast` when downcasting to a child type.

src/apprt/gtk/class/application.zig
```diff
@@ -1990,6 +1992,43 @@ const Action = struct {
...
+        var value = std.mem.zeroes(gobject.Value);
+        defer value.unset();
+        _ = value.init(gobject.ext.typeFor(?*Surface));
+        ghostty_window.as(gobject.Object).getProperty("active-surface", &value);
+
+        const surface: ?*Surface = @ptrCast(@alignCast(value.getObject()));
```
pluiedev: Ditto here

src/apprt/gtk/class/application.zig
```diff
@@ -1990,6 +1992,43 @@ const Action = struct {
...
+            },
+        };
+        const gtk_window: *gtk.Window = @ptrCast(@alignCast(target_node.f_data orelse return false));
+        gtk.Window.present(gtk_window);
+
+        const ghostty_window: *Window = @ptrCast(@alignCast(gtk_window));
+        var value = std.mem.zeroes(gobject.Value);
+        defer value.unset();
+        _ = value.init(gobject.ext.typeFor(?*Surface));
+        ghostty_window.as(gobject.Object).getProperty("active-surface", &value);
```
pluiedev: I think using `gobject.Object.get`, even though it's only supposed to be used in C, might make things a bit cleaner here.
```zig
var surface: ?*gobject.Object = null;
ghostty_window.as(gobject.Object).get("active-surface", &surface, @as(?*anyopaque, null));
```

src/Surface.zig
```diff
@@ -4975,6 +4975,15 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
             },
         ),
 
+        .goto_window => |direction| return try self.rt_app.performAction(
+            .{ .surface = self },
+            .goto_window,
+            switch (direction) {
+                .next => apprt.action.GotoWindow.next,
+                .previous => apprt.action.GotoWindow.previous,
```
pluiedev: Decl literals should just work here, no?
```suggestion
                .next => .next,
                .previous => .previous,
```

pluiedev (changes requested):

src/apprt/gtk/class/application.zig
brtmax: I tried to apply it, is this what you had in mind? Also thanks for the feedback as well

src/apprt/gtk/class/application.zig
brtmax: Added

src/Surface.zig
brtmax: Thanks, yes it works! I noticed that I got the ordering inconsistent in this part, so I committed it again with correct ordering.

src/apprt/gtk/class/application.zig
brtmax: Thanks for the hint, I think I understand it better now. I've kept the pure Zig version but added Null handling similar to the quit function by making the result of the findCustom optional and falling back to the first window when no active window is found, is this better or do you want me to change it?

src/apprt/gtk/class/application.zig
brtmax: Addressed above

mitchellh:
I did some manual cleanups and merged this manually. Thank you.

PR #9335, opened 2025-10-24
Add configurable word boundary characters for text selection

mauroporras:
## Summary

This PR adds a new `selection-word-chars` configuration option that allows users to customize which characters mark word boundaries during text selection operations (double-click, word selection, etc.).

## Motivation

This's been on my wishlist for a while. Inspired by #9069 which added semicolon as a hardcoded word boundary, this PR takes the concept further by making word boundaries fully configurable. Different workflows and use cases benefit from different boundary characters - SQL developers might want semicolons as boundaries, while others working with file paths or URLs might prefer different settings.

This approach is similar to zsh's `WORDCHARS` environment variable, giving users fine-grained control over text selection behavior.

## Changes

- **New config option**: `selection-word-chars` with default value `` ` \t'"│`|:;,()[]{}<>$` ``
- **Runtime UTF-8 parsing**: Boundary characters are parsed from UTF-8 string to u32 codepoints
- **Updated function signatures**: `selectWord()` and `selectWordBetween()` now accept boundary characters as parameters
- **All call sites updated**: Surface.zig, embedded.zig, and all test cases updated

## Usage

Users can now customize word boundaries in their config:

```ini
# Remove semicolon from boundaries (treat as part of words)
selection-word-chars = " \t'\"│`|:,()[]{}<>$"

# Remove periods for better URL selection
selection-word-chars = " \t'\"│`|:;,()[]{}<>$"
```

## Implementation Details

- Boundary characters are stored in `DerivedConfig` and passed through to selection functions
- UTF-8 parsing happens at runtime with graceful fallback for invalid input
- Null character (U+0000) is always included as a boundary automatically
- Multi-byte UTF-8 characters are fully supported

## AI Assistance Disclosure

With gratitude for the team and respect for the [Contributing Guidelines](https://github.com/ghostty-org/ghostty/blob/main/CONTRIBUTING.md), I want to disclose that this PR was written with AI assistance (Claude Code). I have reviewed all the code, and to the extent of my understanding, I'm prepared to answer any questions about the changes.

## Related

- Inspired by #9069


src/config/Config.zig
```diff
@@ -683,6 +683,31 @@ foreground: Color = .{ .r = 0xFF, .g = 0xFF, .b = 0xFF },
...
+///
+/// Default: ` \t'"│`|:;,()[]{}<>$`
+///
+/// To add or remove specific characters, you can set this to a custom value.
+/// For example, to treat semicolons as part of words:
+///
+///     selection-word-chars = " \t'\"│`|:,()[]{}<>$"
+///
+/// Available since: 1.2.0
+@"selection-word-chars": []const u8 = " \t'\"│`|:;,()[]{}<>$",
```
mitchellh: Instead of making this a `[]const u8`, I'd recommend making a new type here that automatically expands these into a list of codepoints. 

This way we don't need an arbitrary max, we can limit it by the allocator (or put a really high limit), and we can allocate, in general! 

It also limits the runtime cost when we actually do selection since the boundary characters are already built up.

mitchellh (changes requested):
I'm conceptually fine with this but I would use a slightly different approach, as noted in the comment.

src/config/Config.zig
mauroporras: @mitchellh, I made the change, thanks for your feedback.

Also, I ran `zig build run` and did a quick test with "some-hyphenated-words" and it worked.

src/config/Config.zig
```diff
@@ -5487,6 +5510,111 @@ pub const RepeatableString = struct {
...
+    };
+
+    /// The parsed codepoints. Always includes null (U+0000) at index 0.
+    codepoints: []const u32 = &default_codepoints,
+
+    pub fn parseCLI(self: *Self, alloc: Allocator, input: ?[]const u8) !void {
+        const value = input orelse return error.ValueRequired;
+
+        // Parse UTF-8 string into codepoints
+        var list = std.ArrayList(u32).init(alloc);
```
pluiedev: In Zig 0.15 collection types are unmanaged by default - you need to pass the allocator into every use of the list that may (de-)allocate memory

```suggestion
        var list: std.ArrayList(u32) = .empty;
```

src/config/Config.zig
```diff
@@ -5487,6 +5510,111 @@ pub const RepeatableString = struct {
...
+        ']', // right bracket
+        '{', // left brace
+        '}', // right brace
+        '<', // less than
+        '>', // greater than
+        '$', // dollar
+    };
+
+    /// The parsed codepoints. Always includes null (U+0000) at index 0.
+    codepoints: []const u32 = &default_codepoints,
```
pluiedev: Unicode codepoints are expressed as u21s in the Zig standard library, so we should do the same here and avoid the `@intCast` below

pluiedev (changes requested):
There's a *lot* of CI errors — have you tried running the tests yourself first? You have to run `zig fmt` to clean up the code, too.

mauroporras:
> There's a _lot_ of CI errors — have you tried running the tests yourself first? You have to run `zig fmt` to clean up the code, too.

@pluiedev, sorry about this. I was using Zig 0.15.1 since 0.15.2 is not available in Homebrew just yet, so I built 0.15.2 from source.
~~I converted the PR to draft while I address your comments. Thanks.~~
Done, thanks.

codemedic:
@mauroporras Any chance this can be revived, free of merge conflicts.

I would love to see this feature merged; I miss the ability to double-click on grep outputs to select `file-name.ext:123`

mauroporras:
@codemedic, done, thanks.

mitchellh (approved):
Thanks, the changes after review look good. I fixed one memory leak you put in here (we have to clone memory for DerivedConfig). And I added a note this is in 1.3 in the docs.

PR #9339, opened 2025-10-25
macos: use system beep for bell

duskwuff:
This seems pretty straightforward. I've tested it and it does what I'd expect it to do.

Fixes #9338


mitchellh (approved):
Since `system` is off by default, I'm good with this. I just don't want any audio by default.

tylerlaprade:
~Docs still say this is GTK-only. Does it actually work? I'm not hearing any sounds even with `bell-features = system`.~

EDIT: Sorry, just realized this is part of an unreleased version, ignore me. I'll leave my comment up in case it helps anyone confused like I was.

PR #9346, opened 2025-10-25
terminal: add ReadonlyStream that updates terminal state

mitchellh:
This adds a new stream handler implementation that updates terminal state in reaction to VT sequences, but doesn't perform any of the actions that would require responses (e.g. queries).

This is exposed in two ways: first, as a standalone `ReadonlyStream` and `ReadonlyHandler` type that contains all the implementation. Second, as  a convenience func on `Terminal` as `vtStream` and `vtHandler` which return their respective types preconfigured to update the calling terminal state.

This dramatically simplifies libghostty-vt usage from Zig (and will eventually be exposed to C, too) since a Terminal on its own is ready to go as a full VT parser and state machine without needing to build any custom types!

There's a second big bonus here which is that our `stream_readonly.zig` tests are true end-to-end tests for raw bytes to terminal state. This will let us test a wider variety of situations more broadly. To start, there are only a handful of tests implemented here. 

**AI disclosure:** Amp wrote basically this whole thing, but I reviewed it. https://ampcode.com/threads/T-3490efd2-1137-4112-96f6-4bf8a0141ff5


mitchellh:
I added a new example showing the usage.

PR #9353, opened 2025-10-26
gtk: Register toggle-maximize and toggle-fullscreen as SimpleActions

vietcgi:
This PR adds GTK SimpleAction registration for `toggle_maximize` and `toggle_fullscreen`, enabling
  these actions to be triggered via D-Bus and other programmatic interfaces.

  ## Background

  These actions already work perfectly fine through keybinds (thanks to the apprt callback system in
  `application.zig`), but they weren't registered as GTK actions, which meant external tools and D-Bus
  couldn't trigger them.

  ## What Changed

  Added GTK SimpleAction registration for both actions in `window.zig`:
  - Added `"toggle-maximize"` and `"toggle-fullscreen"` to the window actions array
  - Implemented handler functions that route to the existing `performBindingAction()` infrastructure

  This follows the same pattern as other window actions like `split-*`, `copy`, `paste`, etc.

  ## What This Enables

  ### D-Bus Activation

  Before this PR (v1.2.3):
  ```bash
  $ gdbus call --session --dest com.mitchellh.ghostty \
      --object-path /com/mitchellh/ghostty/window/1 \
      --method org.gtk.Actions.Activate "toggle-fullscreen" "[]" '@a{sv} {}'
  Error: Unknown action 'toggle-fullscreen'

  After:
  $ gdbus call --session --dest com.mitchellh.ghostty \
      --object-path /com/mitchellh/ghostty/window/1 \
      --method org.gtk.Actions.Activate "toggle-fullscreen" "[]" '@a{sv} {}'
  ()  # Success - window toggles fullscreen

  GTK Action Discovery

  The actions now show up when you query the GTK action registry - useful for automation tools, desktop
   environment integrations, accessibility tools, and any other external software.

  Note on Keybinds

  Keybinds like keybind = alt+enter=toggle_maximize already work in v1.2.3 through the apprt callback
  system. This PR doesn't change keybind functionality - it only adds the ability to trigger these
  actions programmatically through D-Bus and GTK's action system.

  Testing

  Tested on GNOME/Wayland with GTK 4.14.5 - keybinds still work, D-Bus activation now works, and
  actions are present in GTK registry.

**AI Assistance**: This fix was developed with assistance from Claude Code (Anthropic), including root cause investigation, code analysis, patch creation, automated testing, and comprehensive feature testing that discovered the second bug.

### DIFF
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -349,6 +349,7 @@ pub const Window = extern struct {
             .init("toggle-command-palette", actionToggleCommandPalette, null),
             .init("toggle-inspector", actionToggleInspector, null),
             .init("toggle-maximize", actionToggleMaximize, null),
+            .init("toggle-fullscreen", actionToggleFullscreen, null),
         };
 
         ext.actions.add(Self, self, &actions);
@@ -1836,6 +1837,14 @@ pub const Window = extern struct {
         self: *Window,
     ) callconv(.c) void {
         self.performBindingAction(.toggle_maximize);
+    }
+
+    fn actionToggleFullscreen(
+        _: *gio.SimpleAction,
+        _: ?*glib.Variant,
+        self: *Window,
+    ) callconv(.c) void {
+        self.performBindingAction(.toggle_fullscreen);
     }
 
     fn actionRingBell(
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -348,6 +348,7 @@ pub const Window = extern struct {
             // TODO: accept the surface that toggled the command palette
             .init("toggle-command-palette", actionToggleCommandPalette, null),
             .init("toggle-inspector", actionToggleInspector, null),
+            .init("toggle-maximize", actionToggleMaximize, null),
         };
 
         ext.actions.add(Self, self, &actions);
@@ -1618,6 +1619,7 @@ pub const Window = extern struct {
         self: *Self,
     ) callconv(.c) void {
         _ = surface;
+
         if (self.as(gtk.Window).isMaximized() != 0) {
             self.as(gtk.Window).unmaximize();
         } else {
@@ -1826,6 +1828,14 @@ pub const Window = extern struct {
         self: *Window,
     ) callconv(.c) void {
         self.performBindingAction(.clear_screen);
+    }
+
+    fn actionToggleMaximize(
+        _: *gio.SimpleAction,
+        _: ?*glib.Variant,
+        self: *Window,
+    ) callconv(.c) void {
+        self.performBindingAction(.toggle_maximize);
     }
 
     fn actionRingBell(


rhodes-b:
Toggle fullscreen is also not registered and works perfectly fine these are done with apprt callbacks not though the gobject system
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/application.zig#L722-L723
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/application.zig#L2465-L2470
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/surface.zig#L735-L742

I can't test maximize on my window manager but fullscreen action below works perfectly fine
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/application.zig#L2386-L2391
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/surface.zig#L726-L733

rhodes-b:
The point of my comment was not that you missed fullscreen its that `toggle_fullscreen` already worked before you made this PR

vietcgi:
Interesting.  It was not working for me.  Let me confirm it  @rhodes-b

rhodes-b:
Fullscreen definitely works (edit a better video showing im using the ghostty keybind not my window manager)

https://github.com/user-attachments/assets/ee5ce9fd-2ae5-4edc-9d0d-1f3d8beeeae1

vietcgi:
@rhodes-b  

 You're absolutely right.  I apologize for the confusion in my initial PR description. After thorough
  testing, I can confirm the keybinds already work perfectly in v1.2.3 through the apprt callback
  system.

  The issue I encountered was actually a window manager keybind conflict. GNOME was intercepting F11
  before Ghostty could see it (via org.gnome.Terminal.Legacy.Keybindings full-screen 'F11'). When I
  tested with an alternative keybind like Ctrl+Shift+F, it worked fine in v1.2.3.

  I've updated the PR description to clarify that this fix is specifically for D-Bus/programmatic 
  activation, not keybinds. The GTK SimpleAction registration enables:

  1. D-Bus activation via org.gtk.Actions.Activate
  2. Action discovery for external tools
  3. Programmatic triggering from automation scripts

  Here's the actual difference this PR makes:

  v1.2.3:
  - Keybinds work (via apprt callbacks)
  - D-Bus activation fails with "Unknown action 'toggle-fullscreen'"

  With this PR:
  -  Keybinds still work (unchanged)
  - D-Bus activation now works

mitchellh:
The motivation for this change isn't articulated well at all, although you're using a shitload of words to say nothing. You sound like an AI, maybe are one. In any case, please, like a human, explain why these need to be actions. Is GNOME trying to trigger these actions? Can you point me to where?

vietcgi:
Hey @mitchellh 
[test-pr-9353.sh](https://github.com/user-attachments/files/23154728/test-pr-9353.sh)

thanks for pointing out that I didn’t make the actual motivation clear earlier.

This just exposes toggle-maximize and toggle-fullscreen as GTK actions so they can be triggered over D-Bus. Keybindings already work. This just makes them accessible to external test tooling, using the same pattern we already use for things like toggle-command-palette and the split-* actions.

You can see the difference by running the test script:

./test-pr-9353.sh


On my end (Ubuntu 24.04):

With the patch: the actions register and gdbus can activate them.

Without it: D-Bus reports Unknown action.

No behavior change for normal users, it just wires the existing toggle logic to an exposed action entry point.

It’s literally just the two .init(...) lines in src/apprt/gtk/class/window.zig, and the callbacks forward straight to the same performBindingAction paths the keybindings already use. So it stays consistent with the rest of the window actions.

The use case is automated integration tests that need to resize or fullscreen the window via D-Bus, e.g.:

gdbus call ... --method org.gtk.Actions.Activate "toggle-maximize"

GNOME itself isn’t triggering these, this is strictly for test automation.

pluiedev:
I think this would be better implemented by extending our existing D-Bus-based IPC mechanism instead.

mitchellh:
Let's close this and move this back to discussion. The fix is simple. The motivation though is "personal test automation." In that case, maintaining your own fork of Ghostty seems great and it's unclear if there is broad appeal for something like this. OSS is about forks, so enjoy it. 

I'm not convinced yet that this needs to be shipped in main.

PR #9354, opened 2025-10-26
ci: update milestone action

mitchellh:
Fixes our failing CI, hopefully!


hustcer:
@mitchellh I think you are using the wrong version: https://github.com/ghostty-org/ghostty/actions/runs/18818896837/job/53691289639#step:3:1  V3's hash should be: dcd6c3742acc1846929c054251c64cccd555a00d  you are currently using v2.11

mitchellh:
Ah whoops! Yeah it looks like this fixed it! Thank you

PR #9382, opened 2025-10-28
Search feature for Ghostty (Mac only)

### LINKED ISSUE #189: GTK: Search scrollback

mattheworiordan:
Fixes https://github.com/ghostty-org/ghostty/issues/189

Not having search was just too painful for me to use Ghostty. Given https://github.com/ghostty-org/ghostty/issues/189 now has 1k+ upvotes, I decided to fix the problem for myself (or go back to iTerm2!). 

I hope this PR represents a good improvement on Ghostty for Mac users (sorry Linux users, I don't have any way to build and test at present).  

Here it is in action:

MO screenshot 2025-10-28 at 22 06 39

Note:

- AI assistance was used (Claude). I wrote a small PRD and developed in two phases. Phase 3 (not done) would be to add Linux support. I hope someone else can take my work and do that.


avarayr:
https://github.com/ghostty-org/ghostty/pull/8850#issuecomment-3344101885

pluiedev (changes requested):
This doesn't solve the fundamental issue behind #189, which is that the naïve search implementation is not thread-safe. I would rather prefer if more effort went into testing and improving #8850 instead of vibing half-complete solutions.

mattheworiordan:
@pluiedev I was not aware of #8850.

> which is that the naïve search implementation is not thread-safe

Out of interest, how does that play out then in practice?  Under what conditions is search not thread safe?  If it's to do live searches of a changing terminal, I could see that as being an issue, but my implementation focuses on what's in the buffer at the time of search instead, which I think brings functionality people sorely need (search the buffer) more than what is a nice to have (live searches). Perhaps I have missed something more key here?

mitchellh:
> Out of interest, how does that play out then in practice? Under what conditions is search not thread safe?

In this PR it's thread-safe because you're holding the big lock during search. We'd like to implement search in such a way that this lock doesn't need to be held, because it holds back IO handling and rendering. 

In practice, this is unlikely to cause major issues, since many terminals are not generating content that quickly. But for a general purpose feature we ship to millions of users, we need to address the edges. Plus, we tend to simply care about those details as a group of maintainers. 

However, search is commonly paired with a log tailing or some active process going while the search is active. We don't want to plummet IO performance (specifically) while search is active. And this would do that. In particular, this search implementation scales `O(N)` with scroll backbuffer size on my cursory look. That won't work for people who have unlimited scroll back and gigabytes of data back there (we've seen it). 

We need to think this through, and this isn't the right path forward for that.

Thank you for this PR. If it works for you, great. If it works for others, please pull this in and build your own binaries. That's what open source and forking is all about, freedom to modify your own software. 

I'm going to move forward with #8850, thank you.

mattheworiordan:
Thanks for the response.

> In practice, this is unlikely to cause major issues
Well yes, that was my belief.

> But for a general purpose feature we ship to millions of users, we need to address the edges
Sure, I feel you. My take was that that is exactly the point, it's an edge, and 1k people want this feature and this edge focus is sadly blocking progress.

> Plus, we tend to simply care about those details as a group of maintainers.
I hear your too. This PR came about because I was speaking with a colleague who said he doesn't use Ghostty because of features like this that are missing that are just showstoppers, so I kind of wondered why. This is your project and it's a great piece of work (thanks). But perfection is the enemy of good, and personally I believe a feature like this should just exist, even if it comes with tradeoffs. But this is not my project!

> search is commonly paired with a log tailing or some active process going while the search is active
I thought about that, and didn't think that was a real use case.

IMHO, if you want to search a log tail, I'd expect people to use command line tools for that which is way more efficient, like `tail | grep 'foo'`. I am not really sure what the use case is to have a live log tail and UI search running concurrently (how would it even work). I can see it being done, but I think it's valid for the search to be a one time operation (that is blocking momentarily) and needs to be execute again if the buffer changes.

> In particular, this search implementation scales O(N) with scroll backbuffer size on my cursory look. That won't work for people who have unlimited scroll back and gigabytes of data back there (we've seen it).

Agreed, that's a problem.

Look, thanks for the feedback. Most of what you said is 100% accurate. One thing I would consider though is a potential workaround that could make this land if a shorter-term solution to the other issue is needed - I could simply limit how far back the search goes. That gives us a bounded amount of work in the search operation, which I think is quite reasonable from a UX perspective. I appreciate you’d probably rather just proceed with issue #8850, but I thought I’d float it anyway—it’s easy to implement and may give us an interim solution. Good luck with that issue and I appreciate the response.

mitchellh:
> I am not really sure what the use case is to have a live log tail and UI search running concurrently (how would it even work). 

Since I'm not a common search user, I did interview a number of users on their general search usage and a very frequent one was for example having a live running local dev server (e.g. `next dev`) for example. Others are via SSH.

> I can see it being done, but I think it's valid for the search to be a one time operation (that is blocking momentarily) and needs to be execute again if the buffer changes.

I think for a lot of people this would be a fine tradeoff, but as the official app distribution, we need to solve for the general case as much as possible. We have the time to do that (search isn't promised until 1.3, which is in March), so we might as well do it right from the beginning. We have laid out the plan in #189 already for how to do this, and #8550 is marching along that path.

> One thing I would consider though is a potential workaround that could make this land if a shorter-term solution to the other issue is needed - I could simply limit how far back the search goes.

We have a good workaround in place: people who like this can pull in this PR and build from source. They can add size limiting if they'd like! I still appreciate the work here and this can provide some insight to future contributors or maintainers on how to implement other parts of this.

mattheworiordan:
> I did interview a number of users on their general search usage and a very frequent one was for example having a live running local dev server

Oh, OK. That's good to know, and clearly does make my solution incompatible with that need then. 👍 

> I still appreciate the work here and this can provide some insight to future contributors or maintainers on how to implement other parts of this.

Thanks, sorry it didn't get into `main`, and thanks again for taking the time to respond so quickly!

PR #9388, opened 2025-10-29
macOS: keep app badge sync with surface bells

bo2themax:
Keep track of surfaces that have rung the bell to fix #8487.

https://github.com/user-attachments/assets/6c2c6b41-8219-437b-8fa0-8b815fc66fd2

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -83,7 +83,15 @@ extension Ghostty {
         @Published private(set) var backgroundColor: Color? = nil
 
         /// True when the bell is active. This is set inactive on focus or event.
-        @Published private(set) var bell: Bool = false
+        @Published private(set) var bell: Bool = false {
+            didSet {
+                if bell {
+                    Self.ghosttyApp?.belledSurfaceIDs.insert(id)
+                } else {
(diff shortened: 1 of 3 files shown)


macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -83,7 +83,15 @@ extension Ghostty {
         @Published private(set) var backgroundColor: Color? = nil
 
         /// True when the bell is active. This is set inactive on focus or event.
-        @Published private(set) var bell: Bool = false
+        @Published private(set) var bell: Bool = false {
+            didSet {
```
mitchellh, line 87: Instead of this and tightly coupling it to the app, I think it'd make sense instead to use a notification that anyone can listen to. 

Probably AppDelegate on its own would have to listen to it.

mitchellh (changes requested):
Thanks, I'd like to propose a slightly different approach, see my inline comment.

We also have to consider closing surfaces. We may want to add an apprt action that is triggered from core when an Surface closes, actually.

bo2themax:
> Thanks, I'd like to propose a slightly different approach, see my inline comment.
> 
> We also have to consider closing surfaces. We may want to add an apprt action that is triggered from core when an Surface closes, actually.

Make sense, i didn't think about closing surfaces before, closing this for now.

PR #9400, opened 2025-10-29
macOS: make text editor in clipboard confirmation non focusable

bo2themax:
With its being `focusable`(default), the first responder became the text editor instead of the paste button.

This fixes the issue where one can't confirm with the keyboard.

This doesn't affect its selection.


mitchellh (approved):

unsignedint:
Thank you @bo2themax for fixing this 👍

PR #9402, opened 2025-10-29
terminal: HTML formatting

### LINKED ISSUE #9395: Add an `html` formatting type to the core formatter package
This would emit any terminal contents as HTML. See `formatter.zig`. We already have plain and VT at the time of writing this issue. Make sure to write tests.

mitchellh:
This adds HTML formatting capabilities to the formatter package. HTML is emitted as inline styles. For palette indexes, direct RGB is emitted if we have access to a palette; otherwise, we fall back to CSS variables.

This isn't exposed to end users yet, but will enable copy as html features. This is available in libghostty. 

Fixes #9395

**AI disclosure:** I used AI (Amp) to help me write tests, but the implementation was done manually. I reviewed everything.

### DIFF
--- a/src/lib_vt.zig
+++ b/src/lib_vt.zig
@@ -25,6 +25,7 @@ pub const osc = terminal.osc;
 pub const point = terminal.point;
 pub const color = terminal.color;
 pub const device_status = terminal.device_status;
+pub const formatter = terminal.formatter;
 pub const kitty = terminal.kitty;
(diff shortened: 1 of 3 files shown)


mitchellh:
Ah, one todo here is we need to preserve whitespace better automatically. This is shrinking to one line. Of course!

mitchellh:
CleanShot 2025-10-29 at 20 54 19@2x

PR #9406, opened 2025-10-30
Don't encode option as alt in modify other keys 2

danielwe:
There have been frequent reports of key encoding issues in vim and tmux with version 1.2.3 on macOS: #9340, #9361, #9401, https://discord.com/channels/1005603569187160125/1432413679806320772.

I think I found the culprit: the option modifier is always passed as alt to the core, regardless of `macos-option-as-alt`. Since #9289, this means that a key event where option was used (as option) for translation is encoded as if it also has the alt modifier.

For example, consider the many European keyboard layouts where option+8 sends `[`. If `macos-option-as-alt = true`, Ghostty correctly intercepts the option and encodes option+8 as alt+8 instead (that is, `^[[27;3;56~`). But if `macos-option-as-alt = false`, Ghostty first allows option to be used for translation, obtaining `[`, and then encodes the key event as alt+[ (that is, `^[[27;3;91~`), rather than just `[`.

Tweaking the test case from #9289, here's a quick way to see this: set `macos-option-as-alt = left`, run
```
printf '\033[>4;2m'
cat
```
choose a European keyboard layout (e.g., Norwegian), and hit both left-option+8 and right-option+8. The former inserts `^[[27;3;56~` in all well-behaved terminals. The latter inserts `[` in other terminals, but `^[[27;3;91~` in Ghostty.

Basically, while modify other keys 2 does require encoding consumed modifiers, the option key is not one of the supported modifiers, and should not be included (as alt or anything else) when `macos-option-as-alt = false`.

This PR removes alts that were actually options when using modify other keys 2.

### DIFF
--- a/src/input/key_encode.zig
+++ b/src/input/key_encode.zig
@@ -412,8 +412,20 @@ fn legacy(
         if (it.nextCodepoint() != null) break :modify_other;
 
         // The mods we encode for this are just the binding mods (shift, ctrl,
-        // super, alt).
-        const mods = event.mods.binding();
+        // super, alt unless it is actually option).
+        const mods = mods: {
+            var mods_binding = event.mods.binding();
+            if (comptime builtin.target.os.tag.isDarwin()) alt: {
+                switch (opts.macos_option_as_alt) {
+                    .false => {},
+                    .true => break :alt,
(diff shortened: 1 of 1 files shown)


mitchellh:
Thank you!

PR #9415, opened 2025-10-30
terminal: add osc8 tag handling for HTML formatter

matthew-hre:
The HTML page formatter can now track hyperlink state so <a> tags open and close when the OSC 8 data changes. Also added a new `writeHtmlEscaped` helper to keep generated markup safe.

Originally written with Copilot, revised by hand.

### DIFF
--- a/src/terminal/formatter.zig
+++ b/src/terminal/formatter.zig
@@ -1283,13 +1283,13 @@ pub const PageFormatter = struct {
                             const uri = link.uri.offset.ptr(self.page.memory)[0..link.uri.len];
 
                             try writer.writeAll("<a href=\"");
-                            try self.writeHtmlEscaped(writer, uri);
+                            for (uri) |byte| try self.writeCodepoint(writer, byte);
                             try writer.writeAll("\">");
 
                             if (self.point_map) |*map| {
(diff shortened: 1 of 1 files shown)


src/terminal/formatter.zig
```diff
@@ -1269,6 +1354,26 @@ pub const PageFormatter = struct {
         }
     }
 
+    /// Write a string with HTML escaping. Used for escaping href attributes
+    /// and other HTML attribute values.
+    fn writeHtmlEscaped(
```
mitchellh: We should already have this helper somewhere, `writeCodepoint` I think. Should be reused.

mitchellh (changes requested):
Thanks, I think there's some stylistic stuff to improve. I only noted one major issue though.

src/terminal/formatter.zig
matthew-hre: Oops! Should be good now!

mitchellh (approved):
Okay, thanks for starting on this. I cleaned it up, rebased it, and it's good to go.

Note there was a lot of slop here. The general shape of it was right (which AI is good at) but some of the details were way off, including multiple bugs. Just as a warning in the future to be careful about this.

PR #9418, opened 2025-10-30
Copy formatted text to clipboard with plain, make it configurable 

### LINKED ISSUE #9397: Parameterize `copy_to_clipboard` binding to accept formats
Now that we have a formatter that supports emitting in multiple formats, we should expose this to users. At the time of writing this, we support only `plain` and `vt` but have open issues for HTML (#9395) and RTF (#9396). 

Syntax: `copy_to_clipboard:<format>`, e.g. `copy_to_clipboard:html`. For backwards compatibility, no format should always be `plain`.

We should expose this in the command palette too.

mitchellh:
Fixes #9397 

This makes `copy_to_clipboard` take an optional parameter with the format to copy. **The default has changed to `mixed`,** which will set multiple content types on the clipboard allowing the OS or target application to choose what they prefer. In this case, we set both `text/plain` and `text/html`. 

This only includes the macOS implementation. The GTK side still needs to be done, but is likely trivial to do.

https://github.com/user-attachments/assets/b1b2f5cd-d59a-496e-bb77-86a60571ed7f


macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -364,23 +366,48 @@ extension Ghostty {
...
+                // Set data for each type
+                for item in contentArray {
+                    guard let type = NSPasteboard.PasteboardType(mimeType: item.mime) else { continue }
+                    pasteboard.setString(item.data, forType: type)
+                }
                 return
             }
 
+            // For confirmation, use the text/plain content if it exists
+            guard let textPlainContent = contentArray.first(where: { $0.mime == "text/plain" }) else {
```
bo2themax, line 407: Why not join all the text/plain contents?

mitchellh: I think the clipboard allows only one anyways, right? 

We currently guarantee there is only one.  But since we set them in an array we may want to take the last one actually...

PR #9423, opened 2025-10-31
macos: Fix documentView padding calculations

### LINKED ISSUE #9420: macOS: mouse events in first row on screen (first ever, after scrolling is fine) not working
#9409 

I think something in the NSScrollView is blocking it. I've verified this is at the macOS layer, not the core layer.

Reproduce:

1. Open Ghostty
2. Double click the top row to try to highlight something.

If you create one line of scroll back everything works fine.

danielwe:
Fixes #9420 

The problem was ultimately that the padding calculations assumed that the total vertical padding is always less than one cell height (just doing `fmod(contentHeight, cellHeight)` instead of the more careful `contentHeight - scrollbar.len * cellHeight`). This is not at all true, and as a result, the calculated document height was often a cell height short of what it should be.

For similar reasons, we shouldn't rely on `fmod`-based padding update calculations during relayouting. This PR takes the proper approach of saving and reusing the current scrollbar state to calculate the correct document height on every layout. As a bonus, this removes the flickering scrollbar during resize that I complained about in #9296.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -19,6 +19,7 @@ class SurfaceScrollView: NSView {
     private var observers: [NSObjectProtocol] = []
     private var cancellables: Set<AnyCancellable> = []
     private var isLiveScrolling = false
+    private var currentScrollbar: Ghostty.Action.Scrollbar? = nil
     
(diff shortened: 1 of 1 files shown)


bo2themax (approved):

bo2themax:
Looks great! I did one time when I resized the window 'very' quickly, the scrollbar stuck to the window, but I couldn't find a way to reproduce it again after trying for a long time.

PR #9432, opened 2025-10-31
Fix horizontal glyph spacing on normal-DPI monitors

charlesnicholson:
This PR partially addresses #4504 with a one-liner, all feedback is very welcome.

### AI disclaimer
I used Claude Code to help navigate the various layers of the rendering stack, to instrument a ton of intermediate now-deleted log statements, and ultimately to identify and fix this bug. I directed it conversationally, gave it experiments to run, audited all of its work, threw most of it out, and finally landed on this extremely small and simple change that fixes the issue _for me_ but I think for a lot of other cases as well. The fix ended up being small, so hopefully it's easy to review and discuss.

Details:

#  Fix horizontal glyph spacing on non-Retina displays

On external monitors (1.0x scale, 72 DPI), text had excessive horizontal spacing between glyphs. The issue was less noticeable on Retina displays (2.0x scale, 144 DPI) due to higher pixel density, but the proportional spacing error was identical.

## Before
Background is ghostty 1.2.3 from homebrew, foregreound is iTerm2:
Screenshot 2025-10-31 at 3 04 47 PM

## After
Background is ghostty tip + this change, foreground is iTerm2:
Screenshot 2025-10-31 at 3 00 42 PM

(my iTerm2 has some custom thickening / brightening you can see; that's not a part of this change)

## Root Cause

The metrics calculation in Metrics.zig used `@ceil()` to round the cell width from CoreText's glyph measurements, which created a mismatch between cell width (which determines glyph positions) and the actual glyph advances.

At 72 DPI with a 12pt font:
- CoreText returns glyph advance: 7.224609375 pixels
- Cell width was ceiled to: 8 pixels
- Gap per character: 0.78 pixels (~10.8% error)

At 144 DPI with the same font:
- CoreText returns glyph advance: 14.44921875 pixels (exactly 2x)
- Cell width was ceiled to: 15 pixels
- Gap per character: 0.55 pixels (~3.8% error)

The error at high DPI is much better than at low DPI, since the absolute error is always no more than 1px.

##  Fix

Changed `@ceil(face_width)` to `@round(face_width)`. This makes cell width match the glyph advances better, reducing the error to at most 0.5px:
- 72 DPI: round(7.22) = 7
- 144 DPI: round(14.45) = 14

Height continues using `@ceil()` since vertical space can be slightly larger without visual issues, but... should it? I'm not sure; this is a good topic for discussion.

### DIFF
--- a/src/font/Metrics.zig
+++ b/src/font/Metrics.zig
@@ -223,12 +223,12 @@ pub const FaceMetrics = struct {
 ///
 /// For any nullable options that are not provided, estimates will be used.
 pub fn calc(face: FaceMetrics) Metrics {
-    // We use the ceiling of the provided cell width and height to ensure
-    // that the cell is large enough for the provided size, since we cast
-    // it to an integer later.
+    // We use rounding for cell width to match the glyph advances from CoreText,
+    // which avoids spacing issues on non-Retina displays.
+    // We keep ceiling for cell height to ensure vertical space is sufficient.
     const face_width = face.cell_width;
     const face_height = face.lineHeight();
-    const cell_width = @ceil(face_width);
+    const cell_width = @round(face_width);
     const cell_height = @ceil(face_height);
 
     // We split our line gap in two parts, and put half of it on the top
--- a/src/font/Metrics.zig
+++ b/src/font/Metrics.zig
@@ -223,11 +223,33 @@ pub const FaceMetrics = struct {
 ///
 /// For any nullable options that are not provided, estimates will be used.
 pub fn calc(face: FaceMetrics) Metrics {
-    // We use rounding for cell width to match the glyph advances from CoreText,
-    // which avoids spacing issues on non-Retina displays.
-    // We keep ceiling for cell height to ensure vertical space is sufficient.
+    // These are the unrounded advance width and line height values,
+    // which are retained separately from the rounded cell width and
+    // height values (below), for calculations that need to know how
+    // much error there is between the design dimensions of the font
+    // and the pixel dimensions of our cells.
     const face_width = face.cell_width;
     const face_height = face.lineHeight();
+
+    // The cell width and height values need to be integers since they
+    // represent pixel dimensions of the grid cells in the terminal.
+    //
+    // We use @round for the cell width to limit the difference from
+    // the "true" width value to no more than 0.5px. This is a better
+    // approximation of the authorial intent of the font than ceiling
+    // would be, and makes the apparent spacing match better between
+    // low and high DPI displays.
+    //
+    // This does mean that it's possible for a glyph to overflow the
+    // edge of the cell by a pixel if it has no side bearings, but in
+    // reality such glyphs are generally meant to connect to adjacent
+    // glyphs in some way so it's not really an issue.
+    //
+    // TODO: Reconsider cell height, should it also be rounded?
+    //       We use @ceil because that's what we used initially,
+    //       with the idea that it makes sure there's enough room
+    //       for glyphs that use the entire line height, but it
+    //       does create the same high/low DPI disparity issue...
     const cell_width = @round(face_width);
     const cell_height = @ceil(face_height);
 


charlesnicholson:
In https://github.com/ghostty-org/ghostty/issues/4504#issuecomment-3433586320, @Zabriskije points out some screenshots that show vertical cell width issues as well; perhaps changing the vertical sizing to use `@round()` would fix those?

qwerasd205:
Rounding this can (theoretically) cause rendered glyphs to exceed the cell width, and I really don't think this fixes #4504.
As a side note, 14.4 does not ceil to 16, it ceils to 15. The discrepancy between DPIs *is* probably attributable to the way we determine the cell dimensions, but that's because if the high dpi dimensions are odd, say 15px, then that corresponds to a fractional low dpi value (7.5px) which is not realizable - and indeed we instead would have 8px there. The solution there would be to simply calculate the cell dimensions at low dpi and scale them up after so that they correspond exactly.

Aside from that, rounding instead of ceiling these might be worth it to get closer to the true value, since the cell being too narrow likely won't cause issues in practice due to the side bearings on glyphs making them unlikely to actually overflow the cell at any reasonable size.

danielwe:
> perhaps changing the vertical sizing to use `@round()` would fix those?

Off the cuff this doesn't sound like a great idea, as it could make regular ascenders and descenders overlap. But maybe that's OK? See #8550 for a related idea that goes the other way in certain cases to ensure regular ascenders and descenders are always within the cell boundaries.

On the cell width I'm neutral, but if this change is incorporated, we may also want to tweak the following condition to be `@round(metrics.face_width) <= cell_width` instead, such that, e.g., an upright `M` with symmetric bearings still gets pixel-symmetric rendering even if the round happened to go down. Or maybe a better change would be to always do the fractional shift, irrespective of sign, but the integral shift only if it's positive. (For FreeType we already leave subpixel positioning to the hinter, so no analogous change there.)
https://github.com/ghostty-org/ghostty/blob/765ee6842930f0d6918574b9b87c2ee6583e4727/src/font/face/coretext.zig#L370

charlesnicholson:
Thanks for the feedback. @qwerasd205, I took a shot at doing what you suggested- the new code loads the fonts at the base DPI (e.g. 72 on macOS), and then scales the cell dimenions via `content_scale` afterwards. I _think_ the final `@round()` is still required, though. (Apologies in advance if I'm still misunderstanding this; it's a large and complex system).

CoreText is responsible for cell advances, and it seems to base those on the font size it's loaded with. For e.g. Hack Nerd Font Mono at 12px, the advance is 7.22px. For advancing, CoreText returns this value at shaping time, not metrics calculation time. The cell_width is calculated from this 7.22px as well.

But, 7.22px gets `@ceil()`'d up to 8, and that's what's causing the gaps- we get 0.78px gap per glyph, but if we `@round()` it down to 7, then we have a -0.22px gap per glyph.

So, the PR is now scaling up the cell dimensions as you suggested, but I don't see a way to fix the horizontal spacing given what I _think_ the issue is- advances are in (e.g.) 7.22px steps from CoreText during shaping, and that gets `@ceil()`'d to 8px, no matter how the cell size is calculated.

If you're interested, please take another look at the PR; I left the `@round()` change in place because it does fix the horizontal spacing issue, but I added the calculations you suggested.

charlesnicholson:
> As a side note, 14.4 does not ceil to 16, it ceils to 15.

Also @qwerasd205, sorry about that; of course you're correct.

danielwe:
It's unclear to me why we would want a more coarse-grained cell size rounding than necessary on HiDPI monitors. I'd prefer if Ghostty does the best it can with the DPI it has, even if that means the ratio of quantized to unquantized advance is slightly different between different monitors. (We may still conclude that "the best it can" means `@round` rather than `@ceil`.)

charlesnicholson:
@danielwe Could I ask for a bit of clarification on your comment please? My reading of it is that it's in general support of this PR, but I'm not sure if it's intended to be actionable in code, or meant to continue a discussion. If you or @qwerasd205 would like me to make changes to this PR to keep it moving, I'd be happy to do them. 

To my mind, even simply going from `@ceil()` to `@round()` is a win because, yes, it has the potential to impact advancing, but the cell widths are still correct, and it clearly fixes the issue that's kept me and at least a few others in the discussion from replacing their terminals with ghostty :)

danielwe:
What I'm saying is that I think I like your original change. (I do think it warrants a tweak to horizontal alignment as mentioned in my first comment, but I'd be happy to submit a follow-up PR addressing that, so we don't need to discuss that here.)

However, I think the second change is a net negative, prioritizing metric consistency across DPIs over fidelity to the font's design. The fact that rounding differences at 72 DPI can be so stark as to warrant this PR is a strong argument against letting that (literal!) lowest common denominator dictate the metrics for HiDPI displays, where we can do better.

That said, I'm just a contributor with opinions, chiming in because I'm somewhat familiar with this neighborhood of the code, so I'm not telling you what to do, that's up to @qwerasd205.

charlesnicholson:
Hi @qwerasd205 - just checking in to see if you have any thoughts on this PR. I don't really have any opinions on how the cell dimensions are computed, though there doesn't seem to be concensus. Unless I'm misunderstanding the code, though, (which is of course entirely possible!), it seems that the issue is with the advancement value, and not the cell dimensions. 

The cell dimensions seem to be exactly the same with and without the change you suggested, that I _think_ I implemented in response to your feedback, but of course that needs to be audited as well. Also I only tested on my system with one font, etc etc.

Best,
Charles

charlesnicholson:
Hi again @qwerasd205 - just curious if you have any availability to discuss + consider this issue; it's been ~3 weeks or so. To my mind, even without the contentious calculation code that @danielwe flagged, consider this-

The existing implementation is clearly incorrect- through the linked discussion this got flagged as a real bug and plenty of users are seeing it. For some, including me, it's an adoption blocker and I would _really_ love to ditch iTerm for ghostty. I fully respect your stance that my approach (round instead of ceil) might not be correct, though my opinion is that rounding is always better because it's closer to the "true" unrepresentable value. 

Either way, though, even if the rounding approach is actually equally incorrect, IMO the tradeoffs are still worth it if the results scale past my setup. For me, this change does fix ghostty's gross over-advancing issue on lower-DPI monitors like my external Dell, and preserves high-DPI just fine on my Retina M3 MBP display. I fully agree that this has the potential to cause some overlaps, but I haven't seen them on low- or high-DPI displays and I've tried a few fonts. (I'm not arguing that it doesn't happen, more just saying that if it does it might not be on mainstream-ish fonts?)

So, to that end, even if both the current and my proposed implementations are both wrong, maybe taking my patch in the meantime moves ghostty towards a better user experience, which would benefit adoption and give more time to fix this correctly? Please consider it!

I'm happy to take out the scaling-based calculation code, if that's helpful- it does not address the issue either way, the culprit is round vs ceil. (Or, of course, I did the calculations incorrectly). Either way, I'd love to see if you have any appetite for communicating and/or moving this PR along.

Best,
Charles

qwerasd205:
Hi, sorry, I got busy and forgot about this! I think I agree with you that rounding is probably actually better. If you remove the extra code I'd be glad to accept this PR.

charlesnicholson:
Thanks @qwerasd205! My PR is rebased against HEAD and only has the round-vs-ceil change now, I reverted the exploratory cell calculation code. Please let me know if you'd like any other changes, otherwise I'll await CI feedback etc.

qwerasd205:
I cleaned up the commit history since it got a bit messy with all the reverts, and clarified the comments. Running CI now.

qwerasd205 (approved):

charlesnicholson:
>  I cleaned up the commit history since it got a bit messy with all the reverts, and clarified the comments. Running CI now.

Sorry about that, and thanks!

charlesnicholson:
Also, a thought- @qwerasd205 I don't know if you want to leave the https://github.com/ghostty-org/ghostty/issues/4504 issue open, because people are observing height-advance expansion similar to the width-advance expansion that's fixed here. Alternately, I'm happy to open a similar PR for height and do some testing with different fonts + settings on both of my monitors, and post images etc for consideration. Let me know if you'd like that; if there's no appetite for it I'll move on.

qwerasd205:
I've edited the PR description to indicate that it does not fully resolve #4504 and also amend/remove some inaccurate info. Will merge now.

I actually have my own idea about the cell height that I'd like to test out, influenced by #8550, so you don't need to do anything with that since I will be.

PR #9446, opened 2025-11-01
macOS: Refactor scrollview to preserve state across split tree changes

### LINKED ISSUE #9444: macOS: changing the split tree makes scrollbars disappear
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9442

<sup>Originally posted by **danielwe** November  1, 2025</sup>
### Issue Description

On macOS, adding or removing splits sometimes causes the scrollbar to disappear. The reason is that the `SurfaceScrollView` instance is not persistent, it's replaced every time the split tree is rebuilt, so the scroll view state is lost (unless the split reconfiguration caused a resize that triggered a scrollbar update from the core).

I took a brief look at making the `SurfaceScrollView` the persistent, observed view instead of the `SurfaceView`, but I concluded this is better left to someone who knows SwiftUI better than me.

https://github.com/user-attachments/assets/ef4fa74d-b0d1-4b5e-b0c3-808dbe86c187

### Expected Behavior

Keep the scrollbar.

### Actual Behavior

Lose the scrollbar.

### Reproduction Steps

0. Open System Settings and set Appearance -> Show scroll bars -> Always (not strictly necessary, but makes the issue easier to observe)
1. Open Ghostty
2. Create a split
3. Fill one of the surfaces with enough content to create a scrollbar
4. Move to the other surface, and create another split within it
   - The point here is to change the split tree in a way that does not resize the surface with the scrollbar

### Ghostty Logs

```text
Timestamp                       Thread     Type        Activity             PID    TTL
2025-11-01 11:21:05.506932-0700 0x56258df  Info        0x0                  3504   0    ghostty: [com.mitchellh.ghostty:ghostty] known but unimplemented action action=38
2025-11-01 11:21:05.508825-0700 0x56258df  Info        0x0                  3504   0    ghostty: [com.mitchellh.ghostty:io_exec] found Ghostty resources dir: /Applications/Ghostty.app/Contents/Resources/ghostty
2025-11-01 11:21:05.508897-0700 0x56258df  Info        0x0                  3504   0    ghostty: [com.mitchellh.ghostty:io_exec] shell integration automatically injected shell=.zsh
2025-11-01 11:21:05.509089-0700 0x56258df  Info        0x0                  3504   0    ghostty: [com.mitchellh.ghostty:ghostty] known but unimplemented action action=20
2025-11-01 11:21:05.515243-0700 0x5626646  Info        0x0                  3504   0    ghostty: [com.mitchellh.ghostty:io_exec] started subcommand path=/usr/bin/login pid=3713
```

### Ghostty Version

```text
Ghostty 1.3.0-main+765ee6842

Version
  - version: 1.3.0-main+765ee6842
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - app runtime   : .none
  - font engine   : .coretext
  - libxev        : kqueue
```

### OS Version Information

macOS 26.0.1 (Tahoe)

None

_No response_

```ini
# no config needed
```

_No response_

### I acknowledge that:

danielwe:
Fixes #9444

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView.swift
+++ b/macos/Sources/Ghostty/SurfaceView.swift
@@ -407,8 +407,8 @@ extension Ghostty {
         }
 
         func updateOSView(_ scrollView: SurfaceScrollView, context: Context) {
-            // Our scrollview always takes up the full size.
-            scrollView.frame.size = size
+            // Nothing to do: SwiftUI automatically updates the frame size, and
+            // SurfaceScrollView handles the rest in response to that
         }
         #else
         func makeOSView(context: Context) -> SurfaceView {
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -19,7 +19,6 @@ class SurfaceScrollView: NSView {
     private var observers: [NSObjectProtocol] = []
     private var cancellables: Set<AnyCancellable> = []
     private var isLiveScrolling = false
-    private var currentScrollbar: Ghostty.Action.Scrollbar? = nil
     
     /// The last row position sent via scroll_to_row action. Used to avoid
     /// sending redundant actions when the user drags the scrollbar but stays
@@ -137,13 +136,6 @@ class SurfaceScrollView: NSView {
     // insets. This is necessary for the content view to match the
     // surface view if we have the "hidden" titlebar style.
     override var safeAreaInsets: NSEdgeInsets { return NSEdgeInsetsZero }
-
-    override func setFrameSize(_ newSize: NSSize) {
-        super.setFrameSize(newSize)
-        
-        // Force layout to be called to fix up our various subviews.
-        needsLayout = true
-    }
     
     override func layout() {
         super.layout()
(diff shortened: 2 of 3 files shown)


danielwe:
Oops, this doesn't restore the scrolled position of the content view, so in the cases where the scrollbar would previously disappear, it now gets redrawn as scrolled to the bottom, regardless of what the surface scroll state is.

Will fix tomorrow.

danielwe:
OK, this turned into something a little bit bigger than expected. The existing code had a lot of redundant hidden/implicit method calls, and for any single scrollbar update or surface resize, you'd probably take at least a couple of passes through the `layout` method. Fixing #9444 required synchronizing the full scrollview state in `layout`, and this turned out to push the redundancy over the edge, making it possible to trigger infinite loops in some cases.

So I did a deep dive into the hidden parts of the call graph (so much printf debugging!) and tried to root out most of the redundancy. Some key points:
* Explicitly setting the frame size of the view managed by `SurfaceRepresentable` is unnecessary and in fact strongly warned in the docs: https://developer.apple.com/documentation/swiftui/nsviewrepresentable. So I removed that.
* Unconditionally setting `needsLayout = true` in `setFrameSize` is unnecessary and prone to creating loops in the call graph, because `setFrameSize` is often called implicitly when you change the frame of subviews, and `layout` often changes the frame of subviews. In any case, `layout` will always be called when the frame size is actually changed, so the override isn't needed and I removed it.
* With `layout` no longer being called all the time, a more robust approach was needed to handle the NSScrollPocket weirdness from #9317, so I added an observer for this. I also did a deep dive into why these subviews break the hidden titlebar style but not the others. See the docstring comment for the full story; the conclusion is that the right way to handle them is to ensure their frame is zero, unless we're in native fullscreen.
* With `layout` no longer being called all the time it also became possible to lose the scrollbar margin via automatic calls to `SurfaceView.setSurfaceSize` triggered by backing changes, especially when in native fullscreen.[^1] The solution was to save and reuse the size received through `sizeDidChange`. (Eventually, once we get something like #9291, we'll save the inset instead of the size.)

It's a larger diff than I originally expected for this PR, but not huge in the scheme of things, and if you review it commit by commit I think it should be quite straightforward.

[^1]: Strictly speaking, this was already possible, these changes just made it more frequent.

danielwe:
> a more robust approach was needed to handle the NSScrollPocket weirdness from #9317, so I added an observer for this

Btw. I did some testing and the bug that caused this appears to be fixed in Xcode 26.1, so as soon as runner images are updated and we bump the CI workflows, we can remove this workaround entirely.

macos/Sources/Ghostty/SurfaceView.swift
```diff
@@ -407,8 +407,8 @@ extension Ghostty {
         }
 
         func updateOSView(_ scrollView: SurfaceScrollView, context: Context) {
-            // Our scrollview always takes up the full size.
-            scrollView.frame.size = size
+            // Nothing to do: SwiftUI automatically updates the frame size, and
+            // SurfaceScrollView handles the rest in response to that
```
mitchellh, line 411: This should be usable outside of a ScrollableView. Is that still the case?

mitchellh (approved):
This all works really well. There is one little nagging issue I have with a comment I put there, but given this fixes a bunch of issues I'm going to merge it. But do you mind responding to that? Thanks.

mitchellh:
Also thanks for the deep dive, I really appreciate the comprehensive look into all this!

mitchellh:
> Btw. I did some testing and the bug that caused this appears to be fixed in Xcode 26.1, so as soon as runner images are updated and we bump the CI workflows, we can remove this workaround entirely.

Oh excellent! Thank you.

macos/Sources/Ghostty/SurfaceView.swift
danielwe: With the caveat that I haven't fully wrapped my head around the SwiftUI side of things: As far as I can tell, it's not possible to be in this branch (AppKit available) and also have SurfaceRepresentable manage something other than a SurfaceScrollView, is it? The signature of the method even restricts `scrollView` to be of type `SurfaceScrollView`.

In any case, the previous implementation here was always redundant, as SwiftUI automatically sets the frame (and strongly discourages touching it manually). IIUC, what's important for Ghostty is that `sizeDidChange` is called to sync the core whenever the frame size changes, which is currently done in `SurfaceScrollView`s layout method. If `SurfaceRepresentable` managed a `SurfaceView` directly, it would be reasonable to place that call here, like it's done for the UIKit variant right below (line 420). (Maybe that's how it worked for AppKit too before scrollbars? I haven't looked at the git blame.)

Not sure if I'm adequately addressing your concerns here?

danielwe: Sorry, that's an insane number of words to say that whatever doesn't work now didn't work with before this either. Though I don't know what that would be.

mitchellh: I think this does address them. And you're right, I forgot that I forced them all to be scrollable and the proper way to disable that is to just disable the scrollbar later... yep thanks!

PR #9454, opened 2025-11-02
gtk: fix xkb mapping not working in Linux

w9nz:
Fix: https://github.com/ghostty-org/ghostty/discussions/8809#discussioncomment-14466017

If the physical key is non-remappable but is set to act as a remappable key in `xkb layout`, in the actual behavior, the key is dismissed by Ghostty (which is not the behavior in other apps).

With this modification, the key will be remapped if it is a remappable physical key or if the key it is remapped to is remappable.

This way, we respect the most possible the rules given by xkb, but we keep the non-remappable keys safe (cf previous bugs).

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1137,13 +1137,14 @@ pub const Surface = extern struct {
                 if (entry.native == keycode) break :w3c entry.key;
             } else .unidentified;
 
-            // If the key should be remappable, then consult the pre-remapped
-            // XKB keyval/keysym to get the (possibly) remapped key.
+            // Consult the pre-remapped XKB keyval/keysym to get the (possibly)
+            // remapped key. If the W3C key or the remapped key
+            // is eligible for remapping, we use it.
             //
             // See the docs for `shouldBeRemappable` for why we even have to
             // do this in the first place.
-            if (w3c_key.shouldBeRemappable()) {
-                if (gtk_key.keyFromKeyval(keyval)) |remapped|
+            if (gtk_key.keyFromKeyval(keyval)) |remapped| {
+                if (w3c_key.shouldBeRemappable() or remapped.shouldBeRemappable())
                     break :keycode remapped;
             }
 


pluiedev:
Does this solve all the bugs regarding XKB input? (e.g. Ctrl+ц behaving the same as Ctrl+C in a Russian layout, caps:swapescape works, etc.) If it does, then I'm happy to merge this, but we need to be very careful to not break existing workflows again.

w9nz:
It should not change the actual behavior. If you want to remap a letter to another letter, it will still be not allowed. 

Before this fix, we only allow non-letter keys to be remapped. With this, we also add the letters keys that wants to act as a remappable key as allowed remappable keys.

Let me make more tests with Russian keyboard by reviewing the previous tickets. I'll keep you updated.

Thank you for the quick reply!

w9nz:
Hello, 

I tested a russian keyboard, with these setups:
```
- ctrl:nocaps, altwin:swap_lalt_lwin
- caps:swapescape
```

`ctrl-c` , `escape`, `return`,.. is working all the time. 

I also tried to replicate some behavior before/after and I got the same results in both current and this version.

Can you see other tests I could run ?

Thank you for your help!

w9nz:
Any news @pluiedev  ? :-)

Thank you!

pluiedev:
Ah sorry, I kind of forgot about this... At this point I couldn't really find anything wrong with this and I think we should let this go ahead and see if anybody complains, then we could either revert or revise this.

Thanks for investigating this!

pluiedev (approved):

pluiedev:
For some reason there's no option to run CI on my end — could you try to amend/re-sign and then force push the commit? Maybe GitHub got stuck in some weird state that I can't resolve

w9nz:
Hello!

I amended the commit and signed it. I hope it can help. If not, I can create a new PR.

Thank you!

pluiedev:
All good now. Must have been some weird GitHub bug

src/apprt/gtk/class/surface.zig
```diff
@@ -1139,13 +1137,14 @@ pub const Surface = extern struct {
...
+            // Consult the pre-remapped XKB keyval/keysym to get the (possibly)
+            // remapped key. If the W3C key or the remapped key
+            // is eligible for remapping, we use it.
             //
             // See the docs for `shouldBeRemappable` for why we even have to
             // do this in the first place.
-            if (w3c_key.shouldBeRemappable()) {
-                if (gtk_key.keyFromKeyval(keyval)) |remapped|
+            if (gtk_key.keyFromKeyval(keyval)) |remapped| {
+                if (w3c_key.shouldBeRemappable() or remapped.shouldBeRemappable())
```
tristan957, line 1147: Can `gtk_key.keyFromKeyval(keyval)` fail?

tristan957: I see it can return NULL. If it returns NULL, shouldn't we still check if the `w3c_key` should be remappable?

w9nz: Hello!

Good question.

We do not need to check if the `w3c_key` is remappable if we do not have a remapped key. The idea behind the current implementation of `shouldBeRemappable()` is to verify if the remapped key is allowed (ie can the current key be remapped and is it remapped).

This current implementation was done to avoid issues with `caps:swapescape` or non Cyrillic layouts like Russian. 

This PR does not change the current behavior, it is only adding a new validation on what the key is remapped to (because otherwise we skip valid xkb keys input, for example an `a` acting as `Return`).

Since `gtk_key.keyFromKeyval(keyval)` is only looking at a slice and returning the value, I don't think the statement switch will cause risk to the application.

tristan957: I'll leave this PR to @pluiedev to merge.

PR #9469, opened 2025-11-04
config: fix physical -> unicode keybinds for alternate keyboard layouts

martinemde:
Dvorak input (and presumably others) on MacOS causes certain keys to not work as expected: `[` `]` and `=`.

### Related

This fixes https://github.com/ghostty-org/ghostty/discussions/8743 as well as an unmentioned problem where bracket navigation and equalize panes are also broken.

This is similar to https://github.com/ghostty-org/ghostty/pull/8759 but fixes more of the combos. Switching the `+` binding was not enough to fix the problem for me since the right bracket physical keys is where equals should be and overrides the combo.

### What this PR does

Switches several default keybindings from physical key codes to support alternative keyboard layouts like Dvorak and keyboards with dedicated plus keys. Effectively:

```diff
-  .physical = .equal // or .bracket_left or .bracket_right
+  .unicode = '=' // or '[' or ']'
```

### Details

In testing, I found that all of these bindings need to be fixed otherwise the bracket physical keys overshadows the dvorak plus key.

This seems like the right solution for the same reason that we don't use any physical letter or number keys. They move around with different layouts and `=`, `[`, and `]` are no different than other keys like `-` and `0` which use unicode in other default keybinds.

With this fix, tab and pane navigation (cmd+[], cmd+shift+[]), as well as increase font size (cmd+shift+equals and cmd+equals) and equalize panes (ctrl+cmd+=) now work as expected on dvoark layout on MacOS.

Note, I switch between dvorak virtual layout on the laptop and a physical dvorak keyboard (passed through qwerty input) so my combos would need to change depending on which keyboard I was using if we used physical keys only.

I consulted Claude Code to help try to understand what order and precedence was being applied in this change, but I wrote and tested the code myself (however, this is my first `zig` code so take that with a grain of salt).

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -6107,7 +6107,7 @@ pub const Keybinds = struct {
         // set the expected keybind for the menu.
         try self.set.put(
             alloc,
-            .{ .key = .{ .physical = .equal }, .mods = inputpkg.ctrlOrSuper(.{}) },
+            .{ .key = .{ .unicode = '=' }, .mods = inputpkg.ctrlOrSuper(.{}) },
             .{ .increase_font_size = 1 },
         );
         try self.set.put(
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Thank you! Tested with US standard and still good.

kaddkaka:
Which version of ghostty is this in? It's doesn't seem to be in `version: 1.3.0-main+8cf21f5c2`

Gael101:
> Which version of ghostty is this in? It's doesn't seem to be in `version: 1.3.0-main+8cf21f5c2`

Did you find a way to have this working ?

kaddkaka:
> > Which version of ghostty is this in? It's doesn't seem to be in `version: 1.3.0-main+8cf21f5c2`
> 
> Did you find a way to have this working ?

No, I don't remember. I was rather annoyed it wouldn't just work so I still use wezterm as my main terminal.

PR #9473, opened 2025-11-04
Add an example for `palette` configuration

averymcnab:
I found this syntax a bit confusing when I first read it, so I thought it would be helpful for the next person to add a short example.


rhodes-b:
just noting an example already exists in the docs
https://ghostty.org/docs/features/theme#example

you can also look at the default values for examples
`ghostty +show-config --default | grep ^palette`

mitchellh (approved):
This is small enough that it make sense to duplicate. Thanks.

PR #9485, opened 2025-11-05
macOS: set the macos-icon from a separate thread

### LINKED ISSUE #8793: macOS: set the `macos-icon` from a separate thread
Related to #8734 

Based on my understanding of reading of various Apple docs, we can offload a lot of our icon loading to another thread:

- NSWorkspace functions are [thread-safe](https://developer.apple.com/documentation/appkit/nsworkspace/seticon(_:forfile:options:)) (as long its from one thread)
- `applicationIconImage` must be from the [main actor](https://developer.apple.com/documentation/appkit/nsapplication/applicationiconimage)
- NSImage is [Sendable](https://developer.apple.com/documentation/AppKit/NSImage)

We should offload the logic of setting up our custom icon onto a separate thread. Right now it causes a "hitch" in our app.

bo2themax:
Closes #8793

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -119,19 +119,7 @@ class AppDelegate: NSObject,
     private var signals: [DispatchSourceSignal] = []
 
     /// The custom app icon image that is currently in use.
-    @Published private(set) var appIcon: NSImage? = nil {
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
Excellent. This looks really good to me. Just need to try it when I get home but the code looks good. Thank you.

PR #9493, opened 2025-11-05
unicode: update uucode, force emoji modifiers width 2 as standalone

mitchellh:
This updates uucode. As part of this, the wcwidth implementation was updated (in uucode) to make emoji modifiers width ZERO. But if they're standalone, we want them as width 2.

So this also contains a change to force them as width 2 for our width calculation. This only matters for standalone emoji modifiers, because when they form a valid grapheme we don't use this width calculation.


jacobsandlund (approved):
lgtm!

PR #9496, opened 2025-11-06
config: ignore Unicode byte order marks

### LINKED ISSUE #9490: Ghostty should ignore Unicode byte order marks
Ghostty should ignore [Unicode/UTF-8 byte order marks](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8) at the beginning of configuration files.

_Originally posted by @jcollie in https://github.com/ghostty-org/ghostty/discussions/9489#discussioncomment-14883814_

qingyunha:
Close #9490

### DIFF
--- a/src/cli/args.zig
+++ b/src/cli/args.zig
@@ -1444,6 +1444,10 @@ pub const LineIterator = struct {
             var entry = writer.buffered();
             self.line += 1;
 
+            if (self.line == 1) {
+                entry = skipBOM(entry);
+            }
+
             // Trim any whitespace (including CR) around it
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
I didn’t read the implementation yet, but this needs tests

jcollie:
It's also skipping the BoM in the wrong place. This PR will ignore the BoM if it occurs on the CLI, which I think should remain an error. I have a PR that is almost finished. I would have had it up earlier but I took time out to eat dinner.

jcollie:
#9497

PR #9497, opened 2025-11-06
core: handle utf-8 bom in config files

### LINKED ISSUE #9490: Ghostty should ignore Unicode byte order marks

jcollie:
If a UTF-8 byte order mark starts a config file, it should be ignored. This also refactors config file loading a bit to reduce redundant code and to make it possible to test loading config from a file.

Fixes #9490

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -3451,13 +3451,76 @@ pub fn loadFile(self: *Config, alloc: Allocator, path: []const u8) !void {
     };
     defer file.close();
 
+    try self.loadFsFile(alloc, &file, path);
+}
+
+/// Load config from the given File.
+fn loadFsFile(self: *Config, alloc: Allocator, file: *std.fs.File, path: []const u8) !void {
     std.log.info("reading configuration file path={s}", .{path});
     var buf: [2048]u8 = undefined;
     var file_reader = file.reader(&buf);
     const reader = &file_reader.interface;
+    try self.loadReader(alloc, reader, path);
+}
+
+/// Load config from the given Reader.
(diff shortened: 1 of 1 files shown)


src/config/Config.zig
```diff
@@ -3451,15 +3451,78 @@ pub fn loadFile(self: *Config, alloc: Allocator, path: []const u8) !void {
...
+
+/// Load config from the given Reader.
+fn loadReader(self: *Config, alloc: Allocator, reader: *std.Io.Reader, path: []const u8) !void {
+    bom: {
+        // If the file starts with a UTF-8 byte order mark, skip it.
+        // https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8
+        const bom: []const u8 = &.{ 0xef, 0xbb, 0xbf };
+        const str = reader.peek(bom.len) catch break :bom;
+        if (std.mem.eql(u8, str, bom)) {
+            log.warn("skipping UTF-8 byte order mark", .{});
```
mitchellh: I don't think this rises to a "warning" level, which is defined as a very problematic thing that is likely an error but we can continue (vs "err" level which is a straight panic). 

Having a BOM in a text file seems pretty normal, so I'd drop this to an `info` or even `debug`.

mitchellh (changes requested):
One small change requested, the rest looks great thank you

mitchellh (approved):

src/config/Config.zig
jcollie: Fixed.

PR #9502, opened 2025-11-06
terminal: keypad variation sequences should respect VS16

mitchellh:
This fixes the VS16 issues found in this test:
https://ucs-detect.readthedocs.io/sw_results/ghostty.html#ghostty This is also a more robust way to handle VS15/16 in general. 

This commit also changes our propeties to be a packed struct which reduces its size from 4 bytes to 1 and likewise drops our unicode table size 4x. This isn't as big as it seems, since this only affects stage3 of our LUT which goes from 108 bytes to 27 bytes. 

## Before

CleanShot 2025-11-06 at 07 08 20@2x

## After

CleanShot 2025-11-06 at 07 08 02@2x

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -39,8 +39,8 @@
         },
         .uucode = .{
             // TODO: currently the use-llvm branch because its broken on self-hosted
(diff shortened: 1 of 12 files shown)


jacobsandlund (commented):
late lgtm!

PR #9506, opened 2025-11-06
nix: add ucs-detect

mitchellh:
This makes `ucs-detect` available in our Nix environment so that we can run tests on our Unicode support. In the future, I'd like to modify our CI to run this too.

This also adds a `./test/ucs-detect.sh` script that runs `ucs-detect` with consistent options that match the upstream test styles.

Fwiw, I did try to write a simple libghostty-based Zig binary to run this automatically for us in CI but while libghostty-vt is very good, we don't yet have the proper APIs setup for actually setting up a Pty and sub processing commands and hooking them up to a VT stream. So, I punt that to the future.


test/ucs-detect.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
tristan957:
```suggestion
#!/bin/sh
```

There aren't any bash-isms, but it's also not an important suggestion :D

mitchellh: NixOS only ensures env and sh so I'm fine either way. :) I can accept this

jcollie (approved):

PR #9509, opened 2025-11-07
macOS: attach close confirmation alert to the first window that actually needs it

bo2themax:
Currently, close confirmation is always attached to the first tab in the window group.

**This pr will attach the alert to the window that actually needs that confirmation, so upon close, the window with the running process will be focused.**

https://github.com/user-attachments/assets/8fddd4ce-a3f9-4e18-a645-80c85b5ba995

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -817,12 +817,14 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
 
     /// Close all windows, asking for confirmation if necessary.
     static func closeAllWindows() {
-        let confirmWindow: NSWindow? = all
-            .first { $0.surfaceTree.contains(where: { $0.needsConfirmQuit }) }?
-            .surfaceTree.first { $0.needsConfirmQuit }?
+        // The window we use for confirmations. Try to find the first window that
+        // needs quit confirmation. This lets us attach the confirmation to something
+        // that is running.
+        guard let confirmWindow = all
+            .first(where: { $0.surfaceTree.contains(where: { $0.needsConfirmQuit }) })?
+            .surfaceTree.first(where: { $0.needsConfirmQuit })?
             .window
-
-        guard let confirmWindow else {
+        else {
             closeAllWindowsImmediately()
             return
         }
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -812,32 +812,37 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
...
     static func closeAllWindows() {
-        let needsConfirm: Bool = all.contains {
-            $0.surfaceTree.contains { $0.needsConfirmQuit }
-        }
-
-        if (!needsConfirm) {
-            closeAllWindowsImmediately()
-            return
+        var confirmWindow: NSWindow?
+        var needsConfirm = false
```
mitchellh: We don't need needsConfirm. The existence of `confirmWindow` not nil implies it.

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -812,32 +812,37 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
...
+        for controller in all {
+            if let surfaceToConfirm = controller.surfaceTree.first(where: { $0.needsConfirmQuit }) {
+                needsConfirm = true
+                confirmWindow = surfaceToConfirm.window
+                break
```
mitchellh: This can be changed into a functional call `all.first` and assigned to confirmWindow in a let to avoid mutable vars.

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -1158,12 +1163,18 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
         // reason we check ourselves.
         let windows: [NSWindow] = window.tabGroup?.windows ?? [window]
 
+        var confirmWindow: NSWindow?
+        var needsConfirm = false
```
mitchellh: Similar feedback here.

mitchellh (changes requested):
Makes sense but some code styling feedback. Thank you

mitchellh (approved):
I made some manual edits. We don't need `attachedWindow` on base controller because we can just call `confirmClose` on the proper controller. I changed some of the functional mappings to be more succinct.

mitchellh:
Thanks!

PR #9512, opened 2025-11-07
macOS: fix undo new tab will cause a crash

bo2themax:
### Reproduce Steps on Tip (1.2.3 seems fine)

1. Open a new window.
2. Open a new tab.
3. **Wait for 2-3 seconds.**
4. Undo.
5. 💥

**RC is unclear for me, but dispatching it seems to fix it.**

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -377,9 +377,14 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
                 withTarget: controller,
                 expiresAfter: controller.undoExpiration
             ) { target in
-                // Close the tab when undoing
-                undoManager.disableUndoRegistration {
-                    target.closeTab(nil)
+                // Close the tab when undoing. We do this in a DispatchQueue because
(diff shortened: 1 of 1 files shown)


bo2themax:
> ### Reproduce Steps on Tip (1.2.3 seems fine)
> 1. Open a new window.
> 2. Open a new tab.
> 3. **Wait for 2-3 seconds.**
> 4. Undo.
> 5. 💥
> 
> **RC is unclear for me, but dispatching it seems to fix it.**

New window & undo also works fine

mitchellh:
I can't reproduce the crash on main locally... (without this PR)

bo2themax:
> I can't reproduce the crash on main locally... (without this PR)

That's weird, for me it's nearly 100% reproducible on tahoe 26.1

https://github.com/user-attachments/assets/c50d7301-0acd-43fb-bdea-a0e67fb12174

trag1c:
FWIW I can reproduce it too =)

edit: on macOS Tahoe 26.1

bo2themax:
Sequoia doesn't seem to have this issue with tip

mitchellh:
Okay well, the fix seems pretty harmless if it works for you! I added some comments, but otherwise good with this even without understanding root cause.

PR #9515, opened 2025-11-08
core: check that file reader has capacity before priming

danielwe:
Debug builds panic and crash on launch since #9497, but only when there's a sufficiently large and BOM-less config file on the system. The reason is that the `reader.peek` call in the BOM check triggers a read that fills the entire buffer. Later, in LineIterator.next(), there's a call to `reader.fillMore()` to prime the reader. But `fillMore()` indirectly asserts that the reader has at least 1 byte capacity available (this is documented). In debug builds, the assertion fails and we get a panic.

Fixed by calling `fillMore()` only if there's available capacity (seems like there's room for some API improvement in `std.Io.Reader` here?). Added some tests where `LineIterator` is backed by a buffered reader, such that cases like this can be caught.

### DIFF
--- a/src/cli/args.zig
+++ b/src/cli/args.zig
@@ -1427,7 +1427,12 @@ pub const LineIterator = struct {
         //
         // This will also optimize reads down the line as we're
         // more likely to beworking with buffered data.
-        self.r.fillMore() catch {};
+        //
+        // fillMore asserts that the buffer has available capacity,
+        // so skip this if it's full.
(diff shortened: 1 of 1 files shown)


jcollie (approved):

pluiedev (approved):
Honestly in an ideal world we wouldn't need to prime the reader at all, and we can simply instruct the file reader to pre-fetch the file size before doing anything with the file itself.

PR #9517, opened 2025-11-08
macOS: Don't clip surfaceView to contentView, fix transparent scrollbar background

### LINKED ISSUE #9248: macOS: Always visible scrollbar lacks the correct background

danielwe:
Here's a much simpler approach to fixing #9248 that doesn't involve changes to the core or libghostty (the apprt one) at all. No safe area API needed. All that's needed is for the content view to not clip its subviews, such that the surface view can draw behind the scrollbar even though it's a subview of the content view.

I haven't looked at the details on the renderer side, but since this works it must be that it always renders background all the way to the edges of the framebuffer, regardless of what the terminal screen size (including window padding) happens to be. (This was already evident in that we were able to avoid a transparent slot when the scrollbar was hidden.) So I guess we kind of already have a safe area API through this ability to set a screen size that's different from the framebuffer size? _EDIT:_ I guess what's missing is the ability to shift the origin of the screen within the framebuffer.

Fixes #9248, supersedes #9291.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -36,6 +36,9 @@ class SurfaceScrollView: NSView {
         scrollView.usesPredominantAxisScrolling = true
         // hide default background to show blur effect properly
         scrollView.drawsBackground = false
+        // don't let the content view clip it's subviews, to enable the
+        // surface to draw the background behind non-overlay scrollers
+        scrollView.contentView.clipsToBounds = false
         
         // The document view is what the scrollview is actually going
         // to be directly scrolling. We set it up to a "blank" NSView
@@ -142,6 +145,11 @@ class SurfaceScrollView: NSView {
         
         // Fill entire bounds with scroll view
         scrollView.frame = bounds
+        surfaceView.frame.size = scrollView.bounds.size
+
+        // We only set the width of the documentView here, as the height depends
+        // on the scrollbar state and is updated in synchronizeScrollView
+        documentView.frame.size.width = scrollView.bounds.width
         
         // When our scrollview changes make sure our scroller and surface views are synchronized
         synchronizeScrollView()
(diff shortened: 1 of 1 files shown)


bo2themax:
this looks amazing👍🏻
image

bo2themax (approved):

mitchellh:
> (This was already evident in that we were able to avoid a transparent slot when the scrollbar was hidden.) So I guess we kind of already have a safe area API through this ability to set a screen size that's different from the framebuffer size? EDIT: I guess what's missing is the ability to shift the origin of the screen within the framebuffer.

Yes, we draw the terminal to the full edges of whatever the Metal surface is!

danielwe:
Cool, so then the only thing that's missing is an ability to set the content origin. Someone on discord once asked about the ability to extend their background image into the titlebar area, which I guess would require something like that.

PR #9518, opened 2025-11-08
Decouple balanced top and left window paddings to avoid diagonal resize jitter

danielwe:
I had a bit of the same annoyance as #9064. I agree that with balanced padding, you should expect horizontal jitter when resizing horizontally, and vertical jitter when resizing vertically. Diagonal jitter, however, happened because the top padding was upper bounded by the left padding, coupling the vertical and horizontal wobbling. This looks kind of janky, and it's surprising that the balance between top and bottom padding changes as you vary only the width of the window.

With this PR, the upper bound is instead equal to the maximum left padding. Since this only depends on the config, not the current window width, the diagonal wobbling is avoided. Both the top and left padding still have the same range of motion as before.

An alternative could be to bound by the minimum or median left padding instead. Open to suggestions.

**Before**

https://github.com/user-attachments/assets/d12c5870-f05d-450f-89fc-c59eab90e199

**After**

https://github.com/user-attachments/assets/35b80bb0-9ea2-41c1-8502-3a8eec51dbc6

### DIFF
--- a/src/renderer/size.zig
+++ b/src/renderer/size.zig
@@ -44,6 +44,15 @@ pub const Size = struct {
             self.grid(),
             self.cell,
         );
+
+        // The top/bottom padding is interesting. Subjectively, lots of padding
+        // at the top looks bad. So instead of always being equal (like left/right),
+        // we force the top padding to be at most equal to the maximum left padding,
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
This looks good to me. Thanks

PR #9519, opened 2025-11-08
Feat: Support ansi blinking

caiquejjx:
Closes #7703 

Mostly a rebase of https://github.com/ghostty-org/ghostty/pull/5217.
One thing that I needed to add in order for it to work properly was handling the rebuild condition when it needs to blink, so it also forces the rebuild if the cell has blink flag.


caiquejjx:
@mitchellh hey, is this still desired? if so could you take a look? mostly in my second commit changes which includes some logic that wasn't in the base PR like the rebuild condition.
If all good I can work in solving the conflicts, from what I could tell there are no major breaking changes so should be fine.

PR #9520, opened 2025-11-08
gtk: support GTK 4.20 media queries in runtime & custom css

### LINKED ISSUE #9494: GTK: Specified preferred color scheme for CSS providers
Discussed in #9484

GTK 4.20 introduced [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries) to its CSS implementation, and one of the queries is the preferred color scheme (light/dark). For some reason this is not synchronized automatically based on Adw.StyleManager, meaning we have to update the value ourselves.

First of course we have to update our GObject bindings to 4.20 / GNOME 49, but once that's in place everything should be fairly simple.

jcollie:


jcollie:
Leaving as draft until Gnome 49 is in `nixpkgs-unstable`.

jcollie:
Gnome 49 is now in `nixpkgs-unstable` so marking this as ready for review.

src/apprt/gtk/class/application.zig
```diff
@@ -1593,6 +1593,26 @@ pub const Application = extern struct {
                 );
             };
         }
+
+        if (gtk_version.atLeast(4, 20, 0)) {
```
tristan957, line 1597: Does libadw not do this?

PR #9522, opened 2025-11-09
custom shader: add color scheme information to uniforms

benodiwal:
Add new shader uniforms to provide access to terminal color scheme:
  - iPalette[256]: 256-color ANSI palette array
  - iBackgroundColor: terminal background color
  - iForegroundColor: terminal foreground color
  - iCursorColor: pure cursor color (without opacity)

  The new iCursorColor uniform provides the unmodified cursor color
  from the colorscheme, while existing iCurrentCursorColor retains
  the opacity-modified version. This enables shader authors to create
  sophisticated theme-aware effects and replace cursors entirely.

  Color uniforms are populated per-frame when terminal colors change,
  including dynamic updates via OSC sequences. All uniforms use
  normalized float values (0.0-1.0) with RGB components.

The tests in this PR were written primarily by Claude Code.

  Fixes #9417

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -709,6 +709,11 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     .current_cursor_color = @splat(0),
                     .previous_cursor_color = @splat(0),
                     .cursor_change_time = 0,
+                    // Color Scheme information - initialized to zeros, will be populated per frame
+                    .palette = @splat(@splat(0)),
+                    .background_color = @splat(0),
+                    .foreground_color = @splat(0),
+                    .cursor_color = @splat(0),
                 },
                 .bg_image_buffer = undefined,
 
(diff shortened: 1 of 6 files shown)


qwerasd205:
Is this PR AI-assisted? It's okay if it is, we just ask that you disclose that fact if so. (See [CONTRIBUTING.md](https://github.com/ghostty-org/ghostty/blob/main/CONTRIBUTING.md#ai-assistance-notice).)

benodiwal:
Tests are written by agent. Let me disclose this in description. Thanks for pointing out.

ClearAspect:
Should we really be trying to update the palette every single frame for this? Maybe this would be better to do on a config reload or an OSC sequence? I suppose bg_color is updated every frame so maybe this isn't actually such a big deal. 

Still seems kind of wasteful. Maybe some way of marking the colors as dirty would be a better approach

ClearAspect:
#9656

mitchellh:
Closing for 9656.

PR #9523, opened 2025-11-09
macOS: Update core surface size when scroller style changes

danielwe:
Listen to changes to the OS preferred scroller style and use it to trigger updates the core surface size such that the scrollbar margin kept in sync. Otherwise, the margin update would be deferred until the next window or surface resize/split event.

In the video I'm turning my external bluetooth mouse off and on again, triggering these updates.

https://github.com/user-attachments/assets/15c0ee40-2c1a-419a-8b07-8270e9f7a12f

### DIFF
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -104,6 +104,14 @@ class SurfaceScrollView: NSView {
             self?.handleLiveScroll()
         })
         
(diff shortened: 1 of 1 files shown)


bo2themax (approved):
Don't have a typical mouse around, but changing scroller settings also does the trick

danielwe:
Wow, you're too fast! Or maybe I'm too quick to submit PRs. I just realized I should push an update so the same re-sync is done on config changes as well. Apologies for the salami slicing 🥒

Issue #8357, opened 2025-08-22
macOS: double-clicking the split divider should equalize splits


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/8060

<div type='discussions-op-text'>

<sup>Originally posted by **pavelxdd** July 25, 2025</sup>
### Issue Description

I was going to make a Feature Request about this issue, but searching through the repo found (https://github.com/ghostty-org/ghostty/issues/4354) so I suppose this feature does already exist, but it doesn't work on my macOS (15.5). The only way to equalize splits, is find this option in Cmd-Shift-P popup menu.

### Expected Behavior

Double clicking on a divider should equalize splits.

### Actual Behavior

Double clicking on a divider has no effect.

### Reproduction Steps

1. Open ghostty
2. Press Cmd-D to create a split
3. Resize split by mouse/touchpad
4. Double click on a divider and nothing happens.

### Ghostty Version

```text
Ghostty 1.1.4-main+03ea024bc

Version
  - version: 1.1.4-main+03ea024bc
  - channel: tip
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.none
  - font engine   : font.main.Backend.coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS 15.5 on MacBook Pro 16 M4 Max

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
theme = srcery
term = xterm-256color

font-family = "JetBrains Mono Regular"
font-family-bold = "JetBrains Mono Bold"
font-family-italic = "JetBrains Mono Italic"
font-family-bold-italic = "JetBrains Mono Bold Italic"
font-feature = calt=0, zero, ss02, cv04, cv12, cv18, cv19, cv20
font-size = 13.5
font-thicken = true
font-thicken-strength = 0

alpha-blending = native

adjust-cursor-thickness = 2

cursor-style = bar
cursor-style-blink = false
cursor-click-to-move = true

mouse-hide-while-typing = true
mouse-shift-capture = true

background-opacity = 0.97
background-blur = true

unfocused-split-opacity = 1

window-padding-x = 5
window-padding-y = 5
window-padding-balance = true
window-padding-color = background
window-colorspace = display-p3
window-width = 130
window-height = 40
window-step-resize = true

resize-overlay = never
[...]

PR #8364, opened 2025-08-23
macos: add double-click to equalize splits (#8357)

liby:
Resolves #8357

> This solution was developed through collaborative discussion with Claude Code. I reviewed and tested the implementation to ensure it aligns with the project goals.

PR #9524, opened 2025-11-09
macOS: equalize splits when double tapping on SplitView divider

seankit:
Resolves Issue #8357 

### Implementation

Following the existing `onResize` callback pattern in `TerminalSplitTreeView`, I added an `onEqualize` callback to `SplitView`. When a divider is double-tapped, the callback retrieves a surface from that `TerminalView`'s `SplitTree` and calls `splitEqualize` to equalize the entire tree.

### Context

There is an existing PR #8364 that implements this feature but uses `focusedSurface`, which doesn't work for unfocused windows. Since that PR has been inactive for a few months after requested changes, I've implemented this alternative approach.

Credit to @liby for that initial implementation!

### AI Usage

I chatted with Claude Code in Plan mode to understand the relationship between surfaces and the split tree/split views, but I wrote all the code myself.

### Screenshot

https://github.com/user-attachments/assets/0efd70ef-c90e-4b50-b853-b05e2ca2be67

### DIFF
--- a/macos/Sources/Features/Splits/SplitView.swift
+++ b/macos/Sources/Features/Splits/SplitView.swift
@@ -21,6 +21,9 @@ struct SplitView<L: View, R: View>: View {
     let left: L
     let right: R
 
+    /// Called when the divider is double-tapped to equalize splits.
+    let onEqualize: () -> Void
+
     /// The minimum size (in points) of a split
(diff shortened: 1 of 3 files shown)


bo2themax (approved):
Looks good, thanks!

PR #9529, opened 2025-11-09
macOS: remove `readyToInstall` state in update capsule

bo2themax:
There is a sparkle-related 'issue' with the previous implementation. When you download/install in the `updateAvailable` state, if you don't install it, then check the updates again. Sparkle loses its downloaded stage in the delegate (it's normal when I use the sparkle source code). This time, when you click install in the `updateAvailable` state, it just uses the previous downloaded package and starts to install, without calling `showReady(toInstallAndRelaunch:)`.

I think removing `readyToInstall` in our customised ui makes sense, since the current package is pretty small and only takes a few seconds to download for a normal network. And most of users intend to install this update.

Also, there is no way back once you reach "Install and Restart" stage in `SPUStandardUserDriver`, unless you force quit ofc.

~~This pr also contains some further cleanup for #9170, since `InstallingView` is no longer visible for users.~~

#### `auto-download`

When `auto-download = check`, Sparkle will call `UpdateDriver/showUpdateFound(with:state:reply:)`.

When `auto-download = download`, previously no update ui was presented to the user, neither Sparkle's nor Ghostty’s.
> I tried tick&untick auto download&install in Sparkle's alert, which doesn't seem to make any difference.

With this pr, `auto-download = check` will behave the same, **`auto-download = download` will now prompt the user with a capsule.**

https://github.com/user-attachments/assets/f994dd29-d348-4fea-8777-df3720d6e7af

> [!NOTE]
> I used AI to proofread my comments

### DIFF
--- a/macos/Sources/Features/Update/UpdateDriver.swift
+++ b/macos/Sources/Features/Update/UpdateDriver.swift
@@ -164,10 +164,10 @@ class UpdateDriver: NSObject, SPUUserDriver {
     }
     
     func showReady(toInstallAndRelaunch reply: @escaping @Sendable (SPUUserUpdateChoice) -> Void) {
-        viewModel.state = .readyToInstall(.init(reply: reply))
-        
         if !hasUnobtrusiveTarget {
             standard.showReady(toInstallAndRelaunch: reply)
+        } else {
+            reply(.install)
         }
     }
     
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -815,6 +815,14 @@ class AppDelegate: NSObject,
                 autoUpdate == .check || autoUpdate == .download
             updateController.updater.automaticallyDownloadsUpdates =
                 autoUpdate == .download
+            /**
+             To test `auto-update` easily, uncomment the line below and
+             delete `SUEnableAutomaticChecks` in Ghostty-Info.plist.
+
+             Note: When `auto-update = download`, you may need to
+             `Clean Build Folder` if a background install has already begun.
+             */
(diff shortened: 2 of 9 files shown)


macos/Sources/Features/Update/UpdateDriver.swift
```diff
@@ -164,10 +164,10 @@ class UpdateDriver: NSObject, SPUUserDriver {
     }
     
     func showReady(toInstallAndRelaunch reply: @escaping @Sendable (SPUUserUpdateChoice) -> Void) {
-        viewModel.state = .readyToInstall(.init(reply: reply))
-        
         if !hasUnobtrusiveTarget {
             standard.showReady(toInstallAndRelaunch: reply)
```
bo2themax: For `SPUStandardUserDriver`, I don't see an easy way to change "Install Update" button title without writing your own driver, so I left that unchanged. But if you think "Install Update" can also imply an relaunch, we could just reply with `install` as well.

mitchellh:
How does this work for people who have background downloads enabled? `auto-update = download`. I've wanted to make this change but I'm afraid for people who have this that it might force update without any confirmation whatsoever. If that's true, I think the proper solution is to fix this up in the driver (not via the Ghostty config).

bo2themax:
> How does this work for people who have background downloads enabled? `auto-update = download`. I've wanted to make this change but I'm afraid for people who have this that it might force update without any confirmation whatsoever. If that's true, I think the proper solution is to fix this up in the driver (not via the Ghostty config).

Hmm, true. I'll try that later, drafting for now.

bo2themax:
> `auto-update = download`

Updated in the descriptions

PR #9551, opened 2025-11-10
feat: add descriptions to fish shell completions

### LINKED ISSUE #9531: fish: Add descriptions to Fish shell completions
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9507

<sup>Originally posted by **matthew-hre** November  6, 2025</sup>
Currently, Ghostty's fish shell completions don't include descriptions for configuration options:
image

Could we add descriptions to the fish shell completions, like so? 

image

The above was done by making `help_strings` available to fish completions, and then just using the first line - hence some awkward cutoffs.

Was curious if anyone else wanted this feature, or if anyone had a better idea of how to implement this.</div>

denyszhak:
Add descriptions to fish shell completions 

Claude Code used to understand the codebase and reason through the edge cases (several iterations)

- Using first sentence from the Config to add description for each config, even if sentence spans few lines
- Several options can share the same description, code doesn't duplicate description, see screenshot in the posted issue thread below

Fixes [#9531](https://github.com/ghostty-org/ghostty/issues/9531)

### DIFF
--- a/src/extra/fish.zig
+++ b/src/extra/fish.zig
@@ -192,3 +192,15 @@ fn getDescription(comptime help: []const u8) []const u8 {
 
     return out[0..len];
 }
+
+test "getDescription" {
+    const input = "First sentence with \"quotes\"\nand newlines. Second sentence.";
+    const expected = "First sentence with \\\"quotes\\\" and newlines.";
+    const result = comptime getDescription(input);
+
+    comptime {
+        if (!std.mem.eql(u8, result, expected)) {
+            @compileError("getDescription test failed: expected '" ++ expected ++ "' but got '" ++ result ++ "'");
+        }
+    }
+}
--- a/src/extra/fish.zig
+++ b/src/extra/fish.zig
@@ -3,6 +3,7 @@ const std = @import("std");
 
 const Config = @import("../config/Config.zig");
 const Action = @import("../cli.zig").ghostty.Action;
+const help_strings = @import("help_strings");
 
 /// A fish completions configuration that contains all the available commands
 /// and options.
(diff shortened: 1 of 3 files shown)


src/extra/fish.zig
```diff
@@ -143,3 +153,42 @@ fn writeCompletions(writer: *std.Io.Writer) !void {
         }
     }
 }
+
+fn getDescription(comptime help: []const u8) []const u8 {
```
jparise, line 157: Here's what I think this code is doing:

- Copying up to the first `.` (followed by a space or newline) into `out`
- Replacing newlines with spaces
- Escaping quote characters (`"` to `\"`)

Is that roughly correct? It would be good to include some tests that demonstrate and exercise this logic.

I don't have a strong opinion on the specific rules at this point. I think we have a good amount of flexibility in how to present these help description strings, and we can adjust them in the future.

Design-wise, I think "summarizing an option's description" could be generally useful in other places, but I don't think we necessary need to solve for those speculative cases here. It's just something to consider. For example, maybe this is something that helpgen itself could do (emit a short description alongside the full string). @mitchellh or others?

denyszhak: Yeah, that's correct, separate desc is a good way either, whatever you think suites the most. Will wait for additional response and I can add a test, I didn't find tests for compilation part so didn't bother. Thanks for review

jparise: I think a basic `test getDescription` would be helpful in this case, given it's a specific piece of logic, but I don't think it needs to be exhaustive.

I do think it would be helpful to generalize this concept. (zsh also supports completion descriptions, for example.) But that can happen separately.

denyszhak: Added a test now. I could open a separate issue to generalize this concept and create a PR for it

src/extra/fish.zig
```diff
@@ -143,3 +153,54 @@ fn writeCompletions(writer: *std.Io.Writer) !void {
...
+                len += 1;
+                prev_was_space = (c == ' ');
+            },
+        }
+    }
+
+    return out[0..len];
+}
+
+test "getDescription" {
```
jparise, line 196: Is this a reason why we're not using `std.testing` here?

denyszhak: My bad, updated the test

jparise (approved):

PR #9552, opened 2025-11-10
macOS: support `close_all_windows` action

bo2themax:
`close_all_windows` is marked supported, but missing actual implementation currently


bo2themax:
Seems like there are some issues with the runners atm🤔

PR #9559, opened 2025-11-11
macOS: restore non native fullscreen styles

### LINKED ISSUE #8435: macOS: non-native fullscreen window should restore as non-native fullscreen
Discussion: #3765

I think restoring a non-native fullscreen window should restore in the same state, since it's not as jarring as a regular fullscreen window. 

Reproduction:

```
fullscreen = true
window-save-state = always
```

1. Open Ghostty
2. Quit Ghostty.
3. Open Ghostty

bo2themax:
Fixes #8435

> [!NOTE]
> Used AI to proofread my comments

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -66,24 +66,6 @@ class BaseTerminalController: NSWindowController,
     /// Fullscreen state management.
     private(set) var fullscreenStyle: FullscreenStyle?
 
-    /// The current effective fullscreen mode.
-    /// This is non-nil only while the window is in fullscreen.
-    var effectiveFullscreenMode: FullscreenMode? {
-        guard let fullscreenStyle, fullscreenStyle.isFullscreen else {
-            return nil
-        }
-
-        switch fullscreenStyle {
-        case is NativeFullscreen: return .native
-        case is NonNativeFullscreen: return .nonNative
-        case is NonNativeFullscreenPaddedNotch: return .nonNativePaddedNotch
-        case is NonNativeFullscreenVisibleMenu: return .nonNativeVisibleMenu
-        default:
-            assertionFailure("Missing case here")
-            return nil
-        }
-    }
-
     /// Event monitor (see individual events for why)
     private var eventMonitor: Any? = nil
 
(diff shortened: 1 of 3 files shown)


macos/Sources/Features/Terminal/TerminalRestorable.swift
```diff
@@ -109,6 +109,17 @@ class TerminalWindowRestoration: NSObject, NSWindowRestoration {
         }
 
         completionHandler(window, nil)
+        // We don't restore the previous fullscreen mode. If the saved mode differs from
+        // the current configuration, using either could be confusing. Instead, we honor
+        // the configured mode (consistent with new_window behavior).
```
bo2themax: My opinion here👆🏻, we can use the saved one as well

bo2themax: This makes sense because this restoration will not be called when `window-save-state` is set otherwise.

mitchellh: I think using the saved one makes sense, because it is possible to have different styles for different windows (by clicking the traffic light for example to force native).

bo2themax: > I think using the saved one makes sense, because it is possible to have different styles for different windows (by clicking the traffic light for example to force native).

Ah right, I didn't see the traffic lights for non-native fullscreen windows, but it can be toggled with shortcuts. Updated with `effectiveFullscreenMode`.

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -66,6 +66,10 @@ class BaseTerminalController: NSWindowController,
     /// Fullscreen state management.
     private(set) var fullscreenStyle: FullscreenStyle?
 
+    /// The current effective fullscreen mode.
+    /// This is non-nil only while the window is in fullscreen.
+    private(set) var effectiveFullscreenMode: FullscreenMode?
```
mitchellh: I feel like we shouldn't need this and can just use the `fullscreenStyle` to figure this out, no? I haven't looked too deeply but it feels kind of gross to try to maintain this extra state here.

mitchellh (changes requested):

mitchellh:
I moved it onto the protocol, approved pending tests.

mitchellh (approved):

PR #9570, opened 2025-11-12
apprt/gtk: remove explicit X11 clipboard atom

### LINKED ISSUE #9532: tip: Ghostty copies text to the clipboard as 'UTF8_STRING', which breaks pasting to Electron apps

vancluever:
Fixes #9532
Supersedes #9554

Turns out the explicit `UTF8_STRING` atom was not needed after all and GTK adds it automatically when running under X11; just having the explicit UTF-8 charset type is enough.

This corrects situations where it may not be necessary to include (Wayland), in addition to removing a duplicate atom under X11.

Importantly, this also corrects issues under Wayland in some scenarios, such as using Electron-based apps (e.g., VSCode/Codium under Ubuntu 24.04 LTS).

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -3365,27 +3365,17 @@ const Clipboard = struct {
                     const bytes = glib.Bytes.new(content.data.ptr, content.data.len);
                     defer bytes.unref();
                     if (std.mem.eql(u8, content.mime, "text/plain")) {
-                        // Add some extra MIME types (and X11 atoms) for
-                        // text/plain. This can be expanded on if certain
-                        // applications are expecting text in a particular type
-                        // or atom that is not currently here; UTF8_STRING
-                        // seems to be the most common one for modern X11, but
-                        // there are some older ones, e.g., XA_STRING or just
-                        // plain STRING. Kitty seems to get by with just
-                        // UTF8_STRING, but I'm also adding the explicit utf-8
-                        // MIME parameter for correctness; technically, for
-                        // MIME, when the charset is missing, the default
-                        // charset is ASCII.
+                        // Add an explicit UTF-8 encoding parameter to the
+                        // text/plain type. The default charset when there is
+                        // none is ASCII, and lots of things look for UTF-8
+                        // specifically.
+                        //
+                        // Note that under X11, GTK automatically adds the
+                        // UTF8_STRING atom when this is present.
                         const text_provider_atoms = [_][:0]const u8{
                             "text/plain",
                             "text/plain;charset=utf-8",
-                            "UTF8_STRING",
                         };
-                        // Following on the same logic as our outer union,
-                        // looks like we only need this memory during union
-                        // construction, so it's okay if this is just a
-                        // static-length array and goes out of scope when we're
-                        // done. Similarly, we don't unref these providers.
                         var text_providers: [text_provider_atoms.len]*gdk.ContentProvider = undefined;
                         for (text_provider_atoms, 0..) |atom, j| {
                             const provider = gdk.ContentProvider.newForBytes(atom, bytes);


mitchellh (approved):

mitchellh (commented):
Thank you so much for the diligence on this!

vancluever:
No problem @mitchellh!

Another small addendum: I just decided to check foot, and they [offer UTF-8 first](https://codeberg.org/dnkl/foot/src/commit/c9abab08079a0b6eeb4f3fc9beb80eb5fd621730/selection.c#L1977-L1982), along with the X11 atoms (offered last). However, pretty sure my assumption about the actual selected type being up to the client still stands:

(https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-data-sharing)

> ### Data negotiation
>
> A client providing data to other clients will create a `wl_data_source` object and advertise the mime types for the formats it supports for that data through the `wl_data_source.offer` request. On the receiving end, the data offer object will generate one `wl_data_offer.offer` event for each supported mime type. 
>
> The actual data transfer happens when the receiving client sends a `wl_data_offer.receive` request. This request takes a mime type and a file descriptor as arguments. This request will generate a `wl_data_source.send` event on the sending client with the same arguments, and the latter client is expected to write its data to the given file descriptor using the chosen mime type.

So we got some extra context here in the event that we see more clients misbehave than not and we actually do need to adjust the order (or actually add the X11 atoms in Wayland).

PR #9572, opened 2025-11-13
custom shader: added cursor style and visibility uniforms

### LINKED ISSUE #9416: custom shader: add cursor shape uniform
Expose Ghostty's cursor shape enum to shaders:
https://github.com/ghostty-org/ghostty/blob/469001b7f623f780cfdcb50cc4842f7a4e6386ab/src/renderer/cursor.zig#L8-L17

```glsl
// Macros for a more semantic usage. Though this could just be differed to the documentation too.
#define CURSORSTYLE_BLOCK 0
#define CURSORSTYLE_BLOCK_HOLLOW 1
#define CURSORSTYLE_BAR 2
#define CURSORSTYLE_UNDERLINE 3

uniform int iCursorStyle;
```

**Use case**: Shaders can render contextually appropriate effects—for example, a full-cell glow for block cursors vs. a vertical beam glow for bar cursors, or a subtle underline ripple for underline cursors. I found that when I was making my shader, I just wanted to know what type of cursor I had so I could easily change the shape of it.

ClearAspect:
Fixes: #9416

Specifically:
  iCurrentCursorStyle
  iPreviousCursorStyle
  iCurrentCursorVisible
  iPreviousCursorVisible

Visibility calculated and updated independently from the typical cursor uniform updates to preserve cursor style even when not in the viewport or set to be hidden

I used Claude-Code to initially navigate and gauge an understanding of the rendering system. Otherwise I authored the rest of the PR


ClearAspect:
While I was working on another issue, I just realized I could have also used the surface mailbox to get the cursor style information from the surface renderer state. Seems like that's the better way to do it from my understanding and I'll happily rewrite this.

mitchellh:
I want to shore up all the custom shader stuff finally (sorry for delay). It appears like this has slightly more scope than what #9416 described. Is that right? If so, can you explain? I'm fine with it but just want it justified.

ClearAspect:
Right, if I'm remembering my rationale correctly, then separating the cursor style into current and previous seemed intuitive. I could imagine there might be some people that would want to know what the previous cursor style was. 
In terms of the flag for visibility, I commented that there might be a case where someone might want to know the exact frame where the visibility changes. To be completely honest, I can't remember my exact thoughts on separating visibility into "current" and “previous," so if we wanted to get rid of that, then I could be fine with that.

mitchellh:
I'd say let's get rid of that. I'm very open to custom shader changes but always like to pair them with demos or something to show their utility.

ClearAspect:
I'll do that going forward

ClearAspect:
Okay, I quickly re-based the commit and adjusted the cursor visibility uniforms. In the last pull request for the colour palettes, you changed the function that updates shader uniforms and split it into two. Looking back at the original, especially the section about cursor uniforms, I’m still bothered by the fact it depends on acquiring the cursor glyph. Should we change it to follow the terminal state norm instead? Personally, I think that would make more sense. I think all that information should be available through terminal state. It doesn’t make sense to me to update the uniform based on the glyph. 

Initially, I implemented cursor style uniforms to also update "cursor_changed" to indicate a uniform update. This made sense because the cursor’s appearance consists of position, colour, and style. However, currently, a changed cursor only updates based on position and colour. This would be some more motivation to kind of rework what we have for "for frame" and "from state" updates to the uniforms

PR #9580, opened 2025-11-13
macOS: add more cursor style and fixes #8409

bo2themax:
Fixes #8409

https://github.com/user-attachments/assets/c9e7ad9b-c1eb-451b-a537-90f380f5f5f3

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -73,7 +73,7 @@ extension Ghostty {
         @Published var surfaceSize: ghostty_surface_size_s? = nil
 
         // Whether the pointer should be visible or not
-        @Published private(set) var pointerStyle: BackportPointerStyle = .default
+        @Published private(set) var pointerStyle: CursorStyle = .horizontalText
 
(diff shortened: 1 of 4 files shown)


macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -73,7 +73,7 @@ extension Ghostty {
         @Published var surfaceSize: ghostty_surface_size_s? = nil
 
         // Whether the pointer should be visible or not
-        @Published private(set) var pointerStyle: BackportPointerStyle = .default
+        @Published private(set) var pointerStyle: CursorStyle = .horizontalText
```
bo2themax, line 76: Default to `iBeam` so that I don't need to change the zig implementation. If you have a better one, please let me know.

bo2themax:
There are still a few mouse shapes missing on macOS, but I don't know how to test those, so I didn't add new cases, just changed those skipped ones

mitchellh (approved):

edrozenberg:
Appears to fix cursor style change (briefly tested).

Appears to cause side-effect of changing titlebar color on mouseover (non-standard MacOS behavior, not present in Ghostty v 1.2.3):
Titlebar Color Change

PR #9581, opened 2025-11-13
wuffs: protect against crafted images that cause overflows

### LINKED ISSUE #9579: Crash with crafted large image payload in kitty protocol
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9578

<sup>Originally posted by **AutumnMeowMeow** November 13, 2025</sup>
### Issue Description

When attempting to view the attached file (unzip it first, then cat it), ghostty crashes with a thread panic: integer overflow in pkg/wuffs/src/png.zig:66:15: 0x3092d20 in decode (ghostty)
```
        width * height * @sizeOf(c.wuffs_base__color_u32_argb_premul),
```

[big-damn-kitty.zip](https://github.com/user-attachments/files/23530108/big-damn-kitty.zip)

This zip contains a single file called big-damn-kitty.out that is 13MB in size, representing a 100000 x 100000 PNG file of all-black pixels. Decoding this file into a flat array of 32-bit ARGB values requires 38GB of RAM. The total length or 40,000,000,000 exceeds a 32-bit size_t.

See also: https://github.com/wezterm/wezterm/issues/1031

### Expected Behavior

Ghostty should detect that the decoded image size is not feasible before attempting to decode it, and (if it wishes to be consistent with most terminals) fully discard the image, without moving the cursor or changing any other VT state.

### Actual Behavior

Ghostty crashes with a thread panic:

```
thread 48112 panic: integer overflow
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/pkg/wuffs/src/png.zig:66:15: 0x3092d20 in decode (ghostty)
        width * height * @sizeOf(c.wuffs_base__color_u32_argb_premul),
              ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/kitty/graphics_image.zig:430:40: 0x333514b in decodePng (ghostty)
        const result = wuffs.png.decode(
                                       ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/kitty/graphics_image.zig:339:51: 0x3335666 in complete (ghostty)
        if (img.format == .png) try self.decodePng(alloc);
                                                  ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/kitty/graphics_exec.zig:360:35: 0x3336079 in loadAndAddImage (ghostty)
    var img = try loading.complete(alloc);
                                  ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/kitty/graphics_exec.zig:145:33: 0x33173ff in transmit (ghostty)
    const load = loadAndAddImage(alloc, terminal, cmd) catch |err| {
                                ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/kitty/graphics_exec.zig:70:33: 0x32f016c in execute (ghostty)
            break :resp transmit(alloc, terminal, cmd);
                                ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/Terminal.zig:2385:34: 0x32c4fd8 in kittyGraphics (ghostty)
    return kitty.graphics.execute(alloc, self, cmd);
                                 ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/termio/stream_handler.zig:426:48: 0x32c4a66 in apcEnd (ghostty)
                if (self.terminal.kittyGraphics(self.alloc, kitty_cmd)) |resp| {
                                               ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/termio/stream_handler.zig:302:40: 0x32c507c in vt__anon_703105 (ghostty)
            .apc_end => try self.apcEnd(),
                                       ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/stream.zig:770:52: 0x32c6ee3 in nextNonUtf8 (ghostty)
                    .apc_end => try self.handler.vt(.apc_end, {}),
                                                   ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/terminal/stream.zig:616:37: 0x32768aa in processOutputLocked (ghostty)
                try self.nextNonUtf8(input[offset]);
                                    ^
/home/autumn/code/zothers/ghostty-1.3.0-main+0f64b9a/src/termio/Termio.zig:668:29: 0x3244311 in threadMainPosix (ghostty)
    self.processOutputLocked(buf);
                            ^
/home/autumn/code/zothers/zig-x86_64-linux-0.15.2/lib/std/Thread.zig:509:13: 0x3228843 in callFn__anon_619930 (ghostty)
            @call(.auto, f, args);
            ^
/home/autumn/code/zothers/zig-x86_64-linux-0.15.2/lib/std/Thread.zig:781:30: 0x31d68b3 in entryFn (ghostty)
                return callFn(f, args_ptr.*);
                             ^
./nptl/pthread_create.c:448:8: 0x7f31f6facb7a in start_thread (pthread_create.c)
../sysdeps/unix/sysv/linux/x86_64/clone3.S:78:0: 0x7f31f702a7b7 in __clone3 (../sysdeps/unix/sysv/linux/x86_64/clone3.S)
???:?:?: 0x0 in ??? (???)
run
└─ run exe ghostty failure
error: the following command terminated unexpectedly:
./.zig-cache/o/e69bca8688956e14873f1c5070dffeb4/ghostty
```

### Reproduction Steps

1. Open Ghostty
2. Unzip the attached file big-damn-kitty.zip.
3. cat big-damn-kitty.out
4. Ghostty will immediately crash.

[big-damn-kitty.zip](https://github.com/user-attachments/files/23530108/big-damn-kitty.zip)

### Ghostty Logs

_No response_

### Ghostty Version

```text
Ghostty 1.3.0-dev+0000000
 
Version
  - version: 1.3.0-dev+0000000
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .Debug
  - app runtime   : .gtk
  - libxev        : io_uring
  - kernel version: 6.12.48+deb13-amd64
  - desktop env   : other
  - GTK version   :
    build         : 4.18.6
    runtime       : 4.18.6
  - libadwaita    : enabled
    build         : 1.7.6
    runtime       : 1.7.6
  - libX11        : enabled
  - libwayland    : enabled
```

### OS Version Information

Debian 13.1 trixie (Stable)

X11

KDE Plasma 6.3.6

```ini
I built Ghostty from source, executed via `zig build run`. I don't know how to check the default minimum configuration.
```

_No response_

### I acknowledge that:

jcollie:
Fixes #9579

Protect against panics caused by integer overflows by using functions that allow integer overflows to be caught instead of causing a panic.

Also protect against DOS from images that might not cause an overflow but do consume an absurd amount of memory by limiting images to a maximum size of 4GiB.


pkg/wuffs/src/main.zig
```diff
@@ -5,6 +5,9 @@ pub const jpeg = @import("jpeg.zig");
 pub const swizzle = @import("swizzle.zig");
 pub const Error = @import("error.zig").Error;
 
+pub const maximum_image_width = 8192;
+pub const maximum_image_height = 8192;
```
mitchellh: Feels like this should be a runtime parameter instead of hardcoding in our `pkg/`. I just foresee some future bug with a lot of confusion lol.

mitchellh (changes requested):
One change requested. 

Is 8192 arbitrary or does Kitty also reject images of a certain size? Might be useful to see if there is some limit there we can match.

jcollie:
> One change requested.
> 
> Is 8192 arbitrary

It's kinda arbitrary as I chose it based on an "8K" display.

> or does Kitty also reject images of a certain size? Might be useful to see if there is some limit there we can match.

Without digging into the source, the documentation claims that each buffer has a 320MB storage buffer. Assuming a 32bit ARGB sample, that would imply the maximum size of an image that kitty would accept is approximately 9160x9160.

https://sw.kovidgoyal.net/kitty/graphics-protocol/#image-persistence-and-storage-quotas

jcollie:
Again, without digging into the source, but Kitty doesn't seem to handle the large image that the OP posted. The Kitty process balloons to ~50G and hangs. No crash but definitely a DOS.

pkg/wuffs/src/main.zig
jcollie: That seems like more work than it's worth as the configs aren't plumbed into the graphics stack where the decoders are called. I'd rather just bump the hard coded sizes. The _maximum _ `image-storage-limit` is 4G. So instead of limiting the width and the height we could limit the decoded image size to 4G. That still seems way overkill as that would imply an image that's 32768x32768 pixels.

jcollie: Switch to limiting images to a maximum decoded size of 4GiB.

mitchellh (approved):

mitchellh:
I want to note it might also be worth changing this part:

```
width * height * @sizeOf(c.wuffs_base__color_u32_argb_premul),
```

To use `std.math.mul` so that we can catch overflows and just raise them appropriately. So even if we change the max image size in the future, we can still prevent integer overflows.

Issue #8399, opened 2025-08-26
macOS: Quick terminal window state should be saved with `window-save-state`


mitchellh:
#3649 

We should save the splits, working directories, positions, etc. similar to every other window when `window-save-state = always` (or when it triggers at all). This should fit into the same standard macOS state restoration system.

PR #9588, opened 2025-11-14
macOS: save&restore quick terminal state

bo2themax:
Implements #8399. 

I didn't save `position` in this pr, since I don't see its necessity. 

Changing `quick-terminal-position` requires a relaunch; saving and restoring this would have to deal with conflicts. I also don't see why a user would change this frequently.

> [!NOTE]
> Used GPT to proofread my comments

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -95,6 +95,7 @@
 				Features/QuickTerminal/QuickTerminal.xib,
 				Features/QuickTerminal/QuickTerminalController.swift,
 				Features/QuickTerminal/QuickTerminalPosition.swift,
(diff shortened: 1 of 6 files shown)


mitchellh (changes requested):
Sorry for the delay. I rebased this. This looks good. There are a couple things I want to look at:

- [x] Moving the quick terminal restorable implementation to the quick terminal feature folder
- [x] Is the env var for quick terms set properly (this happened after this PR)? 

And still have to test it.

mitchellh (approved):
Okay, all done. Fixed a couple bugs I found but works great otherwise. Thank you.

bo2themax:
> Okay, all done. Fixed a couple bugs I found but works great otherwise. Thank you.

Excellent fix👍

PR #9595, opened 2025-11-15
lib-vt: export stream.Action for custom streams

mitchellh:


mitchellh:
cc @qwerasd205 should address your issues

PR #9604, opened 2025-11-16
Provide shell ssh integration for nushell

dmatos2012:
Closes #7877. 

Small disclaimer: First Ghostty PR and Zig PR, but looking forward to contributing to the project. 

This PR supports shell-integration-features `ssh-env` and `ssh-terminfo` as per other shells, but not the rest as this is what the issue states. 

That being said, with this PR, then you would see this:
- `warning(io_exec): shell could not be detected, no automatic shell integration will be injected`, but given that the default mode  is `detect` it will pick up the executable and if ssh features are enabled it will integrate it. This might be confusing for users. 
- I decided to not add `nu` to `pub const Shell` because if we do so, then from what I understand from the code, then the code flow would imply that "shell integration will be injected" but it will only do so if those `ssh-*` features are enabled, which may be misleading. But on the other hand, providing `ssh` shell integration but returning `null` for `?!ShellIntegration` does not seem very correct either. 

- I dont like that I added `features` argument to `setupshell`, just to check them if `nu` was used. The reasoning is because the way Nushell works, if we autoload the `nushell` directory (by `setupXdgDatadirs()`) even if no `ssh` features were present, it will wrap the `ssh` function and I think that is not desirable, even if we end up just forwarding the arguments. 

Sorry for the long wall of text, but I think it was worth to add some of the doubts I have had myself, plus the ones that you folks may add. I am very happy to iterate on this, even if its a very "Easy" one, so I much welcome the feedback. 

> [!NOTE]
>
>  Used `GPT` for helping with nushell variable naming verification/improvement
> Used `Gemini` for helping with understanding the `Zsh` ssh integration so that I could replicate the logic with nushell. Just because I find `zsh` language very difficult to understand in detail.


src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+    print $"Setting up xterm-ghostty terminfo on ($ssh_hostname)..."
+
+    let tmp_dir = $"/tmp/ghostty-ssh-($ssh_user).XXXXXX"
+    let mktemp_cmd = ["mktemp", "-d", $tmp_dir]
+    let ctrl_dir = try {
+        run-external ...$mktemp_cmd
+        | str trim          
+    } catch {
+        $"/tmp/ghostty-ssh-($ssh_user).($nu.pid)"
+    }
```
pluiedev: There's no real reason to use `run-external` here - in addition to the use of raw strings being potentially unsafe, it's also just more convenient to use caret notation and add arguments directly. What's even more baffling to me is that [`mktemp`](https://www.nushell.sh/commands/docs/mktemp.html) exists as a builtin command, so why not just use that instead? Plus, `/tmp` isn't even guaranteed to exist on some systems...

```suggestion
    let ctrl_dir = try {
        mktemp -td $"ghostty-ssh-($ssh_user).XXXXXX"      
    } catch {
        $"/tmp/ghostty-ssh-($ssh_user).($nu.pid)"
    }
```

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+    let master_parts = $ssh_opts ++ ["-o", "ControlMaster=yes", "-o", $"ControlPath=($ctrl_path)", "-o", "ControlPersist=60s"] ++ $ssh_args
+
+    let infocmp_cmd = ["ssh"] ++ $master_parts ++ ["infocmp", "xterm-ghostty"]
+
+    let terminfo_present = (
+      run-external ...$infocmp_cmd
+      | complete
+      | get exit_code
+      | $in == 0
+    )
```
pluiedev: Again, literally no reason why you *need* to use `run-external` here. Just do `^ssh`.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+    $ssh_opts ++= ["-o", $"ControlPath=($ctrl_path)"]
+  }
+
+  return {ssh_term: "xterm-ghostty", ssh_opts: $ssh_opts}
+}
+
+# SSH Integration
+export def --wrapped ssh [...ssh_args: string] {
+  if ($ssh_args | is-empty) {
+    run-external "ssh"
```
pluiedev: Missing return?

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+def set_ssh_terminfo [ssh_opts: list<string>, ssh_args: list<string>] {
+  mut ssh_opts = $ssh_opts
+  let ssh_cfg = (
+    run-external "ssh" "-G" ...($ssh_args)
+    | lines 
+    | each {|e| if $e =~ '\buser\b' or $e =~ '\bhostname\b' {split row ' '}}
+  )
+  let ssh_user = $ssh_cfg.0.1
+  let ssh_hostname = $ssh_cfg.1.1
+  let ssh_id = $"($ssh_user)@($ssh_hostname)"
```
pluiedev: A much more idiomatic way is to name explicit columns, and avoid regexes
```nushell
let ssh_cfg = ^ssh -G ...$(ssh_args)
  | lines
  | split column -n 2 " " key value
  | where key == "user" or key == "hostname"
let ssh_id = $"($ssh_cfg.user)@($ssh_cfg.hostname)"
```

pluiedev (changes requested):
There's a lot of... *strangeness* regarding the shell integration here. It also doesn't seem to be installed via the build system. Have you actually tested this?

src/shell-integration/README.md
```diff
@@ -88,3 +88,16 @@ if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
...
+For `nushell` Ghostty prepends to the `XDG_DATA_DIRS` directory. Nushell automatically
+loads configuration files in `<XDG_DATA_DIRS>/nushell/vendor/autoload/*.nu` on startup. These
+directories are represented in `Nu` by `$nu.vendor-autoload-dirs`. For more details see
+
+[Nushell documentation](https://www.nushell.sh/book/configuration.html#configuration-overview)
+
+> [!NOTE]
+>
+> Ghostty only prepends to `XDG_DATA_DIRS` in the case where the `ssh-*` features are enabled.
+> There is no shell integration for the other currently supported features.
```
rhodes-b: This makes it seem like the features are missing when nu provides some of them automatically without extra configuration.

The docs note this
https://ghostty.org/docs/features/shell-integration#shell-integration-injection
> Some shells such as [Nushell](https://www.nushell.sh/) have built-in support for some of the features that Ghostty provides, so shell integration is not necessary. For example, newer versions of Fish (4.0+) have built-in support for prompt marking so jump_to_prompt, prompt resizing, and prompt selection will work without shell integration.

dmatos2012:
Hi @pluiedev thanks for the feedback. I did test it before hand, maybe it was using a cache build? If I do a `zig build` from main, and checkout my branch and do `zig build run` its not registered.  I have to do first `zig build` from my branch, and then `zig build run`, and then its registered in the `zig-out/share/...`. Did you also enable the `ssh` features?


Regarding the "strangeness", I ran initially `run-external`, and forgot about the caret notation,  so i stuck with `run-external` everywhere. Additionally, I tried to follow the `zsh` approach for the integration, so maybe that plays a part in the "strangeness" that you mention, but thanks, I have addressed your changes!

@rhodes-b correct, I did not know that, I updated the readme to reflect it.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+    | lines 
+    | split column -n 2 " " key value
+    | where key == "user" or key == "hostname"
+    | transpose -r
+    | into record
+  let ssh_user = $ssh_cfg.user
+  let ssh_hostname = $ssh_cfg.hostname
+  let ssh_id = $"($ssh_user)@($ssh_hostname)"
+  let ghostty_bin = $env.GHOSTTY_BIN_DIR + "/ghostty"
+  let check_cache_cmd = ["+ssh-cache", $"--host=($ssh_id)"]
```
jparise: This is only used once. Let's make it a literal part of the command execution below.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+# Requires infocmp to be available locally and tic to be available on remote hosts.
+# Caches installations to avoid repeat installations.
+def set_ssh_terminfo [ssh_opts: list<string>, ssh_args: list<string>] {
+  mut ssh_opts = $ssh_opts
+  let ssh_cfg = ^ssh -G ...($ssh_args)
+    | lines 
+    | split column -n 2 " " key value
+    | where key == "user" or key == "hostname"
+    | transpose -r
+    | into record
```
jparise: I think we can improve this record-building pipeline a little bit:


```suggestion
  let ssh_cfg = ^ssh -G ...($ssh_args)
    | lines 
    | parse "{key} {value}"
    | where key in ["user", "hostname"]
    | select key value
    | transpose -rd
    | into record
    | default { user: $env.USER, hostname: "localhost" }
```

1. Use `parse` rather than `split`
2. Use `transpose -rd` to automatically handle the header row
3. Use `where key in ...` for readability and extensibility
4. Declare defaults "just in case", which nushell makes easy

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+  let ssh_id = $"($ssh_user)@($ssh_hostname)"
+  let ghostty_bin = $env.GHOSTTY_BIN_DIR + "/ghostty"
+  let check_cache_cmd = ["+ssh-cache", $"--host=($ssh_id)"]
+
+  let is_cached = (
+    ^$ghostty_bin ...$check_cache_cmd
+    | complete
+    | get exit_code
+    | $in == 0
+  )
```
jparise: `check_cache_cmd` is only used once, so we can include it directly in the command and also simplify the pipeline:

```suggestion
  let is_cached = (
    ^$ghostty_bin "+ssh-cache" $"--host=($ssh_id)"
    | complete
    | $in.exit_code == 0
  )
```

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+    let master_parts = $ssh_opts ++ ["-o", "ControlMaster=yes", "-o", $"ControlPath=($ctrl_path)", "-o", "ControlPersist=60s"] ++ $ssh_args
+
+    let infocmp_cmd = $master_parts ++ ["infocmp", "xterm-ghostty"]
+
+    let terminfo_present = (
+      ^ssh ...$infocmp_cmd
+      | complete
+      | get exit_code
+      | $in == 0
+    )
```
jparise: Similarly:

```suggestion
    let terminfo_present = (
      ^ssh ...($master_parts ++ ["infocmp", "xterm-ghostty"])
      | complete
      | $in.exit_code == 0
    )
```

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+}
+
+# Removes Ghostty's data directory from XDG_DATA_DIRS 
+let ghostty_data_dir = $env.GHOSTTY_SHELL_INTEGRATION_XDG_DIR 
+$env.XDG_DATA_DIRS = $env.XDG_DATA_DIRS
+  | split row ':'
+  | where (
+  | $it !~ $ghostty_data_dir
+  )
+  | str join ':'
```
jparise: I don't think we need the regex/match operator here. We can also simplify a bit by dropping the intermediate variable and using `filter`:

```suggestion
$env.XDG_DATA_DIRS = (
  $env.XDG_DATA_DIRS
  | split row ':'
  | filter {|path| $path != $env.GHOSTTY_SHELL_INTEGRATION_XDG_DIR }
  | str join ':'
)
```

src/termio/shell_integration.zig
```diff
@@ -76,13 +77,15 @@ pub fn setup(
...
-    exe: []const u8,
-) !?ShellIntegration {
+fn setupShell(alloc_arena: Allocator, resource_dir: []const u8, command: config.Command, env: *EnvMap, exe: []const u8, features: config.ShellIntegrationFeatures) !?ShellIntegration {
+    if (std.mem.eql(u8, "nu", exe)) {
+        try setupNu(alloc_arena, resource_dir, env, features);
+        return null;
+        // // return .{
+        // //     .shell = .nu,
+        // //     .command = try command.clone(alloc_arena),
+        // // };
```
jparise: Let's remove this.

src/termio/shell_integration.zig
```diff
@@ -76,13 +77,15 @@ pub fn setup(
...
 }
 
-fn setupShell(
-    alloc_arena: Allocator,
-    resource_dir: []const u8,
-    command: config.Command,
-    env: *EnvMap,
-    exe: []const u8,
-) !?ShellIntegration {
+fn setupShell(alloc_arena: Allocator, resource_dir: []const u8, command: config.Command, env: *EnvMap, exe: []const u8, features: config.ShellIntegrationFeatures) !?ShellIntegration {
```
jparise: Please retain the original formatting.

src/termio/shell_integration.zig
```diff
@@ -76,13 +77,15 @@ pub fn setup(
...
+    if (std.mem.eql(u8, "nu", exe)) {
```
jparise: These cases were listed alphabetically before. Let's retain that ordering and move this between "fish" and "zsh".

src/termio/shell_integration.zig
```diff
@@ -652,6 +655,19 @@ test "xdg: existing XDG_DATA_DIRS" {
...
+/// Setup the nushell shell integration. This works by setting
+/// XDG_DATA_DIRS so that it can be picked automatically by
+/// nushell on startup.
+/// Only implements `ssh-*` shell features. Rest are not supported.
+fn setupNu(alloc_arena: Allocator, resource_dir: []const u8, env: *EnvMap, features: config.ShellIntegrationFeatures) !void {
+    // This makes sure that `Nu` loads our integration file
+    // and wraps the `ssh` function only if the `ssh` features
+    // are enabled.
+    // Otherwise, it does not do anything.
+    if (features.@"ssh-env" or features.@"ssh-terminfo") {
```
jparise: I'm thinking whether this is a worthwhile optimization. I'm inclined to always set up the XDG shell environment here (unconditionally) and then evaluate the shell features more precisely in the script. What do you think?

(That would eliminate `setupNu`; we could just call `setupXdgDataDirs` in the `nu` case in `setupShell`.)

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+      ^ssh ...$infocmp_cmd
+      | complete
+      | get exit_code
+      | $in == 0
+    )
+
+    if (not $terminfo_present) {
+      let install_terminfo_cmd = $master_parts ++ ["mkdir", "-p", "~/.terminfo", "&&", "tic", "-x", "-"]
+
+      ($terminfo_data | ^ssh ...$install_terminfo_cmd) | complete | get exit_code | if $in != 0 {
```
jparise: And similarly here.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+        print "Warning: Failed to install terminfo."
+        return {ssh_term: "xterm-256color", ssh_opts: $ssh_opts}
+      } 
+      let state_dir = try { $env.XDG_STATE_HOME } catch { $env.HOME | path join ".local/state" }
+      let ghostty_state_dir = $state_dir | path join "ghostty"
+
+      let cache_add_cmd = ["+ssh-cache", $"--add=($ssh_id)"]
```
jparise: Let's also inline this below instead.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,114 @@
...
+
+      # Bug?: If I dont add TMPDIR, it complains about renameacrossmountpoints
```
jparise: Can you research this some more? I don't want to leave it ambiguous.

jparise (commented):
I don't know Nushell very well, but I'm interested to learn, so I've started doing some deeper reading and exploratory research.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
dmatos2012: Did this instead:
```
$env.XDG_DATA_DIRS = (
  $env.XDG_DATA_DIRS
  | split row ':'
  | where {|path| $path != $env.GHOSTTY_SHELL_INTEGRATION_XDG_DIR }
  | str join ':'
)

```

because `filter` is [deprecated](https://github.com/nushell/nushell/issues/16025) since nushell 0.105.0, so I think this should also be fine

src/termio/shell_integration.zig
dmatos2012: I also thought about checking for the features in the `nu script` and `setupXdgDataDirs` unconditionally initially. However, I then realized that if i do so, even if the user has no ssh flags, the  `ssh` function would be wrapped regardless, even if the user didn't ask for any integration, and I am not sure that's something expected. 

The reason is because as far as I understood, and also asking on their discord and reading the docs, there is no way to conditionally export a function, so by doing `def ssh`, it will be overshadowed regardless. The only exception would be defining a `const` that is available at parse time, and then this would be possible, But as far as I understood, only variables in `$nu.*`  are allowed as `const` but not anything in the `$env`. 

So what I wanted to do initially was :

```
#file/in/autoload/integration.nu
(pseudocode)
if ssh feature:
    source ssh.nu
else:
    do nothing. 

#ssh.nu
def ssh[] {
  ^ssh ....
}    
```
In this way, then the `zig` code would be super simple with just `if nu {setupXdgDataDirs()}`, but I couldnt make it to work, because it was not possible (that way). But then again, I am also learning so maybe there is an slight change that could make this possible, by having some kind of `source (if .....)` that could conditionally source it with `const` variables.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
dmatos2012: No bug I think, its just I realized i had them on different mount points, and I supposed then thats expected behavior, by reading the [discussion](https://github.com/ghostty-org/ghostty/discussions/8872).

dmatos2012:
@jparise thanks for super good feedback. I have addressed the changes requested and replied to one of your questions which am also very curious about the best solution.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,104 @@
...
+# Caches installations to avoid repeat installations.
+def set_ssh_terminfo [ssh_opts: list<string>, ssh_args: list<string>] {
+  mut ssh_opts = $ssh_opts
+  let ssh_cfg = ^ssh -G ...($ssh_args)
+    | lines
+    | parse "{key} {value}"
+    | where key in ["user", "hostname"]
+    | select key value
+    | transpose -rd
+    | into record
```
pluiedev: Doesn't `-d` in `transpose` already mean `--as-record`? i.e. This `into record` here should be redundant, right?
```suggestion
```

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,104 @@
...
+
+  let is_cached = (
+    ^$ghostty_bin ...(["+ssh-cache", $"--host=($ssh_id)"])
+    | complete
+    | $in.exit_code == 0
+  )
+
+  if not $is_cached {
+    let ssh_opts_copy = $ssh_opts
+    let terminfo_data = try {infocmp -0 -x xterm-ghostty} catch {
```
pluiedev: infocmp is also an external command

```suggestion
    let terminfo_data = try {^infocmp -0 -x xterm-ghostty} catch {
```

pluiedev (dismissed):

src/termio/shell_integration.zig
jparise: Do you think we could use [modules](https://www.nushell.sh/book/modules.html) / [overlays](https://www.nushell.sh/book/overlays.html) for this?

- Define all of our features inside of a `ghostty` module, which might be nice for manual end-user usage.
- Conditionally `use` them in the shell integration script:

```nushell
if (ssh_feature) {
    use ghostty ssh
    # Or `export use ...` ?
}
```

I haven't tried this myself (sorry, I usually like to verify ideas like this on my own, but I'm still coming up to speed on Nu), so this is just based on my reading of the docs.

src/shell-integration/README.md
```diff
@@ -90,3 +90,17 @@ fi
 ```
 
 Shell integration requires Zsh 5.1+.
+
+### Nushell
```
jparise: Let's move this above Zsh (alphabetically).

src/shell-integration/nushell/ghostty-ssh-integration.nu
```diff
@@ -0,0 +1,103 @@
...
+    $session = set_ssh_env
+  }
+  if "ssh-terminfo" in $shell_features {
+    $session = set_ssh_terminfo $session.ssh_opts $ssh_args
+  }
+
+  let ssh_parts = $session.ssh_opts ++ $ssh_args
+  with-env {TERM: $session.ssh_term} {
+    ^ssh ...$ssh_parts
+  }
```
jparise:
```suggestion
  with-env {TERM: $session.ssh_term} {
   ^ssh ...($session.ssh_opts ++ $ssh_args)
  }
```

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,103 @@
...
+def set_ssh_env []: nothing -> record<ssh_term: string, ssh_opts: list<string>> {
+    return {ssh_term: "xterm-256color", ssh_opts: ["-o", "SetEnv COLORTERM=truecolor", "-o", "SendEnv TERM_PROGRAM TERM_PROGRAM_VERSION"]}
+}
+
+# Enables automatic terminfo installation on remote hosts.
+# Attempts to install Ghostty's terminfo entry using infocmp and tic when
+# connecting to hosts that lack it. 
+# Requires infocmp to be available locally and tic to be available on remote hosts.
+# Caches installations to avoid repeat installations.
+def set_ssh_terminfo [ssh_opts: list<string>, ssh_args: list<string>] {
```
jparise: For consistency, I think we should also specify the `record<...>` return type here. If it feels too verbose, we could also just say `record`.

There unfortunately doesn't appear to be a way to define reusable named types in Nu.

src/shell-integration/nushell/vendor/autoload/ghostty-integration.nu
```diff
@@ -0,0 +1,103 @@
...
+      (
+        $terminfo_data
+        | ^ssh ...($master_parts ++ ["mkdir", "-p", "~/.terminfo", "&&", "tic", "-x", "-"])
+      )
+      | complete
+      | if $in.exit_code != 0 {
+        print "Warning: Failed to install terminfo."
+        return {ssh_term: "xterm-256color", ssh_opts: $ssh_opts}
+      }
+    }
```
jparise: This makes two round trips (`ssh` commands) in the case where terminfo isn't present. The other shell integrations do this using one (sh-style) compound command sequence:

```sh
infocmp xterm-ghostty >/dev/null 2>&1 && exit 0
tic >/dev/null 2>&1 || exit 1
mkdir -p ~/.terminfo 2>/dev/null && tic -x - 2>/dev/null && exit 0
exit 1
```

... and then react to the exit code.

jparise (changes requested):

dmatos2012:
@jparise I addressed the changes mentioned above, and on top gave it an initial shot on actually checking the features in nushell, and letting `zig` just set the `xdgDataDirs` unconditionally. This works now by :

```
/autoload/vendor/bootstrap-integration.nu -> checks the $env.GHOSTTY_SHELL_FEATURES which decides whether the `ghostty-ssh-integration.nu` file will be sourced or not
/autoload/vendor/source-integration.nu -> sources `ghostty-ssh-integration.nu` if a particular file is present
ghostty-ssh-integration -> Not on autoload anymore, and will only be sourced given conditions above.
```

There is still a `TODO`(how to update the `ghostty-ssh-integration.nu` file on a new release), but wanted first to get the idea out to see what you guys think about it. 

For the reasons why we need multiple files and do it this way, is due to how nushell deals with `sourcing` files and `use` files. More details on the last commit itself and [Parse Constant Time Evaluation](https://www.nushell.sh/book/how_nushell_code_gets_run.html#parse-time-constant-evaluation)

src/shell-integration/nushell/ghostty-ssh-integration.nu
dmatos2012: That doesnt work because `Captures a mutable variable` which is $ssh_opts, thats why I did it that way.

src/shell-integration/nushell/vendor/autoload/bootstrap-integration.nu
```diff
@@ -0,0 +1,29 @@
...
+    open $ghostty_ssh_file | save $ssh_integration_file
+  }
+  [false true] => {
+    # We need to check if the user disabled `ssh-integration` and thus
+    # the integration file needs to be removed so it doesnt get sourced by
+    # the `source-integration.nu` file
+    rm $ssh_integration_file
+  }
+  _ => { }
+}
```
jparise: I'm afraid that creating and destroying script files as part of the shell integration process doesn't sit well with me for multiple reasons.

jparise (changes requested):

src/shell-integration/nushell/vendor/autoload/bootstrap-integration.nu
dmatos2012: @jparise I wouldnt say i disagree, but it seems we have two options:
1) Go with the `feature` detection in zig, and just have a single script in nu like the beginning, which seemed to work without any issues.
2) Proceed down this path, but as you noticed this looks much less cleaner, and it creates/destroys files which doesnt seem good at all.

I dont know how to proceed forward because those are the limitations given by Nushell. If you or someone has another 3rd option, I would be more than willing to push it forward to completion.

jparise: Totally understood, and I definitely appreciate the time you've put into exploring all of these options. I'll plan on looking at this myself in more detail soon to see if I can come up with another option. If not, we can move forward with the Zig-level solution.

jparise: @dmatos2012 just wanted to update you that I'm still working on this. I've tried out a few different ideas locally (based on what you've learned, too!). Getting back to it is next on my list once a few other things are wrapped up/merged.

dmatos2012: @jparise thanks for the update. No worries, no rush. Just let us know and we can get back on getting this through.

dmatos2012: @jparise thanks for the update. No worries, no rush. Just let us know and we can get back on getting this through.

dmatos2012: @jparise thanks for the update. No worries, no rush. Just let us know and we can get back on getting this through.

jparise:
@dmatos2012 Thanks for being so patient!

I think I've found a nice approach to nushell integration:

- Export an autoloaded `ghostty` module containing our wrapped functions
- Use `$env.GHOSTTY_SHELL_FEATURES` to guard features at runtime
- Add `--execute "use ghostty *"` to the `nu` command line to automatically "use" them

#10274 provides that general foundation plus a `sudo` implementation. It shouldn't be hard to implement your ssh contribution on top of that.

Let me know how that looks to you.

dmatos2012:
@jparise Im taking a look at your PR and once that is merged, then I will include those changes so this then works in a similar fashion. Thanks so much for the update and for pushing this through :)

mitchellh:
#10274 merged so please rebase and adjust for that. Thank you.

dmatos2012:
@jparise I moved the `ssh-integration` to build on top of your changes. Let me know what you think, but it should be ready based on previous iterations and feedback on the file, but fine if we want to further refine small things. 

I also took the liberty of reformatting some lines of your previous `sudo` integrations so that indentation and style is constant through the file. I think we should include `nu formatting` as a check in our CI. As far as I know the  unofficial standard that people in the nushell community use is the [topiary formatter](https://github.com/blindFS/topiary-nushell). Not perfect, but it really helps  for consistency sake.

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,26 +4,128 @@ export module ghostty {
...
 
     ^sudo ...$sudo_args
   }
+  # Wrap `ssh` to provide ghostty `ssh-integration`
+  export def --wrapped ssh [...ssh_args: string]: any -> any {
+    if ($ssh_args | is-empty) {
+      return (^ssh)
+    }
+    mut session = {ssh_term: "" ssh_opts: []}
+    let shell_features = $env.GHOSTTY_SHELL_FEATURES | split row ','
```
jparise: This is unused.

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,26 +4,128 @@ export module ghostty {
...
+      # Prepend TERMINFO preservation flag if not using sudoedit
+      if (not ("-e" in $sudo_options or "--edit" in $sudo_options)) {
+        $sudo_args = ($args | prepend "--preserve-env=TERMINFO")
+      }
     }
 
     ^sudo ...$sudo_args
   }
+  # Wrap `ssh` to provide ghostty `ssh-integration`
+  export def --wrapped ssh [...ssh_args: string]: any -> any {
```
jparise: Minor cosmetic feedback, but let's position this above the `sudo` wrapper (alphabetical). Or at least add a space between the prior function and this one.

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,26 +4,128 @@ export module ghostty {
...
+
+      # Prepend TERMINFO preservation flag if not using sudoedit
+      if (not ("-e" in $sudo_options or "--edit" in $sudo_options)) {
+        $sudo_args = ($args | prepend "--preserve-env=TERMINFO")
+      }
     }
 
     ^sudo ...$sudo_args
   }
+  # Wrap `ssh` to provide ghostty `ssh-integration`
```
jparise: I think something like this would be a little more helpful:

```suggestion
  # Wrap `ssh` with Ghostty TERMINFO support
```

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,26 +4,128 @@ export module ghostty {
...
+  export def --wrapped ssh [...ssh_args: string]: any -> any {
+    if ($ssh_args | is-empty) {
+      return (^ssh)
+    }
+    mut session = {ssh_term: "" ssh_opts: []}
+    let shell_features = $env.GHOSTTY_SHELL_FEATURES | split row ','
+
+    if (has_feature "ssh-env") {
+      $session = set_ssh_env
```
jparise: I think we can drop the separate `set_ssh_env` function and just inline its record assignment here.

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,26 +4,128 @@ export module ghostty {
...
+    }
+    if (has_feature "ssh-terminfo") {
+      $session = set_ssh_terminfo $session.ssh_opts $ssh_args
+    }
+
+    let ssh_parts = $session.ssh_opts ++ $ssh_args
+    with-env {TERM: $session.ssh_term} {
+      ^ssh ...$ssh_parts
+    }
```
jparise: As currently structured, this could run with `TERM=""` if neither `ssh-env` nor `ssh-terminfo` is enabled. I think we should reorganize the code a little bit so that doesn't happen or default `session.ssh_term` to `$TERM` so this is effectively a no-op.

src/shell-integration/README.md
```diff
@@ -84,7 +84,8 @@ Nushell's vendor autoload mechanism. Ghostty then automatically imports
 the module using the `-e "use ghostty *"` flag when starting Nushell.
 
 Nushell provides many shell features itself, such as `title` and `cursor`,
-so our integration focuses on Ghostty-specific features like `sudo`.
+so our integration focuses on Ghostty-specific features like `sudo`. Additionally,
+we also provide `ssh-integration` via the `ssh-env` and `ssh-terminfo` features.
```
jparise:
```suggestion
so our integration focuses on Ghostty-specific features like `sudo`,
`ssh-env`, and `ssh-terminfo`.
```

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,26 +4,128 @@ export module ghostty {
...
+      return (^ssh)
+    }
+    mut session = {ssh_term: "" ssh_opts: []}
+    let shell_features = $env.GHOSTTY_SHELL_FEATURES | split row ','
+
+    if (has_feature "ssh-env") {
+      $session = set_ssh_env
+    }
+    if (has_feature "ssh-terminfo") {
+      $session = set_ssh_terminfo $session.ssh_opts $ssh_args
```
jparise: I'd also be inclined to inline the `set_ssh_terminfo` logic here, given this is the only caller. I don't feel strongly either way though.

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,26 +4,128 @@ export module ghostty {
...
+
+      print $"Setting up xterm-ghostty terminfo on ($ssh_cfg.hostname)..."
+
+      let ctrl_path = (
+        try {
+          mktemp -td $"ghostty-ssh-($ssh_cfg.user).XXXXXX"
+        } catch {
+          $"/tmp/ghostty-ssh-($ssh_cfg.user).($nu.pid)"
+        } | path join "socket"
+      )
```
jparise: I think we can rely on Nushell's [`mktemp`](https://www.nushell.sh/commands/docs/mktemp.html) without the fallback.

jparise (changes requested):

src/shell-integration/nushell/vendor/autoload/ghostty.nu
dmatos2012: Yeah, I thought it would be cleaner to just leave it in a fn since its pretty big, even tho there is only one caller. That way the `ssh` wrapper reads more cleanly.

dmatos2012:
@jparise Lmk now, it should be much cleaner now.

jparise (approved):
Nice work! Let's get this landed and see how it works in practice.

PR #9619, opened 2025-11-17
macOS: match scroller’s appearance with surface’s background

bo2themax:
Currently, the scroller's appearance is the same as the window's.

So, with the light system appearance and the following config:

```
window-theme = system
macos-titlebar-style = native
```

It's hard to see where the scroller is. This pr changes the scroller’s(ScrollView) appearance to match the surface's background colour, so it's always easier to find.

> Changing `verticalScroller?.appearance` doesn't seem to work

image

### DIFF
--- a/macos/Sources/Ghostty/SurfaceScrollView.swift
+++ b/macos/Sources/Ghostty/SurfaceScrollView.swift
@@ -172,13 +172,16 @@ class SurfaceScrollView: NSView {
     }
     
     // MARK: Scrolling
-    
+
     private func synchronizeAppearance() {
         let scrollbarConfig = surfaceView.derivedConfig.scrollbar
         scrollView.hasVerticalScroller = scrollbarConfig != .never
         scrollView.verticalScroller?.controlSize = .small
+        let hasLightBackground = OSColor(surfaceView.derivedConfig.backgroundColor).isLightColor
+        // Make sure the scroller’s appearance matches the surface's background color.
+        scrollView.appearance = NSAppearance(named: hasLightBackground ? .aqua : .darkAqua)
     }
-    
+
     /// Positions the surface view to fill the currently visible rectangle.
     ///
     /// This is called whenever the scroll position changes. The surface view (which does the


macos/Sources/Ghostty/SurfaceScrollView.swift
```diff
@@ -201,6 +201,8 @@ class SurfaceScrollView: NSView {
         if width > 0 && height > 0 {
             surfaceView.sizeDidChange(CGSize(width: width, height: height))
         }
+        // Make sure the scroller’s appearance matches the surface's background color.
+        scrollView.appearance = NSAppearance(named: surfaceView.hasLightBackground ? .aqua : .darkAqua)
```
danielwe: Did you ever observe `scrollView.appearance` being reset by some operation? If not, I think this can go in `synchronizeAppearance` (where it's a more natural fit), as it only needs to run when the appearance might actually need to change, that is, on init and config reloads.

From my quick testing that's sufficient, including when changing the scrollbar style in the OS settings. See also https://developer.apple.com/documentation/appkit/nsappearancecustomization/appearance, which suggests that `appearance` isn't ever automatically set/reset.

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -82,6 +82,11 @@ extension Ghostty {
         /// dynamically updated. Otherwise, the background color is the default background color.
         @Published private(set) var backgroundColor: Color? = nil
 
+        var hasLightBackground: Bool {
+            // default background color is dark
+            backgroundColor.flatMap(OSColor.init(_:))?.isLightColor ?? false
```
danielwe: This doesn't seem do the right thing when the Ghostty theme has a light background. It always returns false in the cases I was able to test. I think you need to use `derivedConfig.backgroundColor` to know what the actual background color is.

```suggestion
            OSColor(derivedConfig.backgroundColor).isLightColor
```

danielwe (changes requested):
I can reproduce the problem, but as explained in the inline comments, the solution needs a small tweak to work for light themes.

bo2themax:
> I can reproduce the problem, but as explained in the inline comments, the solution needs a small tweak to work for light themes.

Thanks, updated.

macos/Sources/Ghostty/SurfaceScrollView.swift
```diff
@@ -172,13 +172,20 @@ class SurfaceScrollView: NSView {
     }
     
     // MARK: Scrolling
-    
+
+    // Make sure the scroller’s appearance matches the surface's background color.
+    var scrollViewAppearance: NSAppearance? {
```
mitchellh: Since we only use this exactly once I'd prefer to not make a full computed property for it and just call it directly in synchronizeAppearance.

mitchellh (changes requested):
This is very clean, just a small request.

mitchellh (approved):

PR #9633, opened 2025-11-18
feat: introduce `toggle_macos_option_as_alt` binding action with a keybind.

kshitij10496:
- Defines a surface level binding action for toggling the usage of Opt key on macOS as Alt.
- Use `cmd` + `opt` + `o` as the default keyboard binding (similar to Terminal.app).

This is my first PR with Zig so please let me know if there is room to adopt idiomatic patterns.

Refers #2520 

AI Assistance Disclosure: Claude Code assisted me in understanding the codebase. But I wrote the code myself manually without any AI Agents.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -274,6 +274,7 @@ const DerivedConfig = struct {
     mouse_shift_capture: configpkg.MouseShiftCapture,
     macos_non_native_fullscreen: configpkg.NonNativeFullscreen,
     macos_option_as_alt: ?input.OptionAsAlt,
+    macos_option_as_alt_original: ?input.OptionAsAlt,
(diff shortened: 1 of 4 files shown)


src/Surface.zig
```diff
@@ -274,6 +274,7 @@ const DerivedConfig = struct {
     mouse_shift_capture: configpkg.MouseShiftCapture,
     macos_non_native_fullscreen: configpkg.NonNativeFullscreen,
     macos_option_as_alt: ?input.OptionAsAlt,
+    macos_option_as_alt_original: ?input.OptionAsAlt,
```
kshitij10496, line 277: Happy to learn if there is a better way to store the "original" state before the user triggers the binding action. 😄

PR #9642, opened 2025-11-19
Add system integration for highway

selfisekai:


mitchellh (approved):
I'm fine with this, but I think we did this before and found it didn't really have a use case.

PR #9646, opened 2025-11-19
CoreText: Apply subpixel horizontal alignment also when cell width is less than advance

danielwe:
Follow-up to #9432 to ensure that subpixel horizontal alignment is consistent whether `cell_width = @round(face_width)` rounds up or down.

### DIFF
--- a/src/font/face/coretext.zig
+++ b/src/font/face/coretext.zig
@@ -367,9 +367,16 @@ pub const Face = struct {
         // We don't do this if the glyph has a stretch constraint,
         // since in that case the position was already calculated with the
         // new cell width in mind.
-        if ((constraint.size != .stretch) and (metrics.face_width < cell_width)) {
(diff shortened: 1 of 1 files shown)


qwerasd205:
Can you demonstrate the effect of this change with some screenshots?

danielwe:
The clearest difference is that centered and symmetric glyphs are rendered with a symmetric pixel pattern. Currently, this is the case when the cell width is larger than the advance, but not when it's smaller. With #9432, the cell width will often be smaller even without any `adjust-cell-width`, so treating this consistently is more important than before. (This is only relevant for coretext, as with freetype we only do whole-pixel adjustments and leave the rest to the hinter.)

**Before**
Screenshot 2025-11-19 at 20 32 57

**After**
Screenshot 2025-11-19 at 20 32 45

qwerasd205 (approved):
I'm convinced, LGTM

PR #9648, opened 2025-11-20
font: round cell height from line height instead of ceiling

qwerasd205:
This change should give more consistent results between high and low DPI displays, and generally approximate the authorial intent of the metrics a little better.

Also changed the cell height adjustment to prioritize the top or bottom when adjusting by an odd number depending on whether the face is higher or lower in the cell than it "should" be. This should make it easier for users who have an issue with a glyph protruding from the cell to adjust the height and resolve it.

### DIFF
--- a/src/font/Metrics.zig
+++ b/src/font/Metrics.zig
@@ -1,6 +1,7 @@
 const Metrics = @This();
 
 const std = @import("std");
+const assert = @import("../quirks.zig").inlineAssert;
 
 /// Recommended cell width and height for a monospace grid using this font.
 cell_width: u32,
@@ -47,8 +48,9 @@ face_width: f64,
 /// The unrounded face height, used in scaling calculations.
 face_height: f64,
 
-/// The vertical bearing of face within the pixel-rounded
-/// and possibly height-adjusted cell
+/// The offset from the bottom of the cell to the bottom
+/// of the face's bounding box, based on the rounded and
+/// potentially adjusted cell height.
 face_y: f64,
 
 /// Minimum acceptable values for some fields to prevent modifiers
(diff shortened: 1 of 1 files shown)


src/font/Metrics.zig
```diff
@@ -300,6 +322,7 @@ pub fn calc(face: FaceMetrics) Metrics {
         .face_width = face_width,
         .face_height = face_height,
         .face_y = face_y,
+        .cell_baseline_higher_than_true_baseline = face_y > 0,
```
danielwe: Do we need to store this in the struct instead of just checking `face_y > 0` in the one place where it's used?

qwerasd205: This is necessary to make it resilient to multiple `cell_height` adjustments, since those can change `face_y` and make this no longer a valid way to determine this. Of course... there's no way for a user to *configure* multiple `cell_height` adjustments, but I only realized that after writing this and figured it's safer to be future-proof lol

danielwe: I see. I guess you could do `face_y < (cell_height - face_height) / 2` if you want to.

src/font/Metrics.zig
```diff
@@ -333,29 +356,46 @@ pub fn apply(self: *Metrics, mods: ModifierSet) void {
...
+                    // These are all relative to the top of the cell.
+                    addFloatToInt(&self.underline_position, diff_top);
+                    addFloatToInt(&self.strikethrough_position, diff_top);
+                    self.overline_position +|= @as(i32, @intFromFloat(diff_top));
+
+                    // By doing this, the cell baseline will now actually be
+                    // on the other side of the true baseline, so we need to
+                    // swap the value.
+                    self.cell_baseline_higher_than_true_baseline =
+                        !self.cell_baseline_higher_than_true_baseline;
```
danielwe: I think this update should only happen if the number of rows added was odd, i.e., if `diff_top != diff_bottom`?

qwerasd205: True, good spot

src/font/Metrics.zig
danielwe: Come to think of it, not sure it's ideal to use `face_y > 0` to begin with, nor to use `cell_baseline = @round(face_baseline)`. We're rounding as close as possible to having the bottom of the face be flush with the bottom of the cell, rather than as close as possible to the face being vertically centered in the cell.

Really we should be doing
```zig
const padding = cell_height - face_height;
const cell_baseline = @round(face_baseline + (padding / 2));
const face_y = cell_baseline - face_baseline;
const cell_baseline_higher_than_true_baseline = face_y > (padding / 2);
```

qwerasd205: Yeah, that makes sense- I'll change that tomorrow.

This shouldn't be merged before that change.

danielwe: In that case, `face_y > (cell_height - face_height) / 2` will always be a valid way to determine which side to add the extra pixel row to, so that would be an alternative to carrying around `cell_baseline_higher_than_true_baseline` and keeping it up to date. (No strong opinions on which way is better.)

qwerasd205: Alright, changes made, thanks for the suggestions

mitchellh (approved):

PR #9656, opened 2025-11-20
custom shaders: add colorscheme information to shader uniforms

ClearAspect:
#9417
Adds palette and color scheme uniforms to custom shaders, allowing custom shaders to access terminal color information:

  - iPalette[256]: Full 256-color terminal palette (RGB)
  - iBackgroundColor, iForegroundColor: Terminal colors (RGB)
  - iCursorColor, iCursorText: Cursor colors (RGB)
  - iSelectionBackgroundColor, iSelectionForegroundColor: Selection colors (RGB)

Colors are normalized to [0.0, 1.0] range and update when the palette changes via OSC sequences or configuration changes. The palette_dirty flag tracks when colors need to be refreshed, initialized to true to ensure correct colors on new surfaces.

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -227,13 +227,6 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
         /// a large screen.
         terminal_state_frame_count: usize = 0,
 
-        /// Captured dirty for reference after updateFrame() clears the flag
-        /// To be used for shader uniforms.
-        ///
-        /// Initialized to true because we need to get the correct palette even on
-        /// a new Surface. Otherwise we end up with it initialized to 0's
-        palette_dirty: bool = true,
-
         const HighlightTag = enum(u8) {
             search_match,
             search_match_selected,
@@ -1222,8 +1215,6 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     };
                 };
 
-                self.palette_dirty |= state.terminal.flags.dirty.palette;
-
                 break :critical .{
                     .links = links,
                     .mouse = state.mouse,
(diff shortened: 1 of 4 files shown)


src/renderer/generic.zig
```diff
@@ -2300,6 +2315,89 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                 0,
             };
 
+            // Renderer state required for getting the colors
+            const state: *renderer.State = &self.surface_mailbox.surface.renderer_state;
```
mitchellh: Unsafe to read this from `drawFrame`. We must access our render state instead (self.terminal_state).

src/renderer/generic.zig
```diff
@@ -1209,6 +1222,8 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     };
                 };
 
+                self.palette_dirty |= state.terminal.flags.dirty.palette;
```
mitchellh: Must be set before the terminal state is updated. Might be easier to get rid of this and use `render_state.dirty`

mitchellh (changes requested):
I rebased this, noticed a couple issues.

src/renderer/generic.zig
ClearAspect: Ah, thanks for checking this out. I believe I was messing around with this before whole state refactor had been implemented. I can take a look at this again.

mitchellh: I just fixed it no worries!

mitchellh (approved):
Alright thanks! Sorry for the delay. I rebased and made a couple cleanups:

* Removed `palette_dirty`, we can rely on `terminal_state.dirty`
* Split out custom shader uniform updates into two functions: one that relies on terminal state and one that relies on frame state. I call the former from `updateFrame` and the latter from `drawFrame`. This made it cleaner to rely on `terminal_state`.
* Made changes to `rebuildCells` to not grab the draw mutex (its now in another context with it) and not reset dirty flags because this made it cleaner with the custom shader logic.

PR #9663, opened 2025-11-21
fix(renderer): optimize cursor blink timer to skip unnecessary wakeups

matthew-hre:
Closes #8003

Added a `isCursorBlinkable` check to determine if the cursor is in a "blinkable" state. The renderer is only woken up if this is true.

This is organic, human written code.


src/renderer/Thread.zig
```diff
@@ -601,6 +601,30 @@ fn renderCallback(
...
+/// This is used to avoid unnecessary renderer wakeups when the cursor blink
+/// timer fires but the blink state wouldn't actually affect rendering.
+/// See: https://github.com/ghostty-org/ghostty/issues/8003
+fn isCursorBlinkable(self: *const Thread, state: *const rendererpkg.State) bool {
+    if (!self.flags.focused) return false;
+
+    if (!state.terminal.modes.get(.cursor_visible)) return false;
+    if (!state.terminal.modes.get(.cursor_blinking)) return false;
+
+    if (!state.terminal.screens.active.viewportIsBottom()) return false;
```
qwerasd205: This means the cursor stops blinking if you scroll up by 1 row in to scrollback, no?

matthew-hre: Yes, I believe(?) this is the current behavior; `cursor.zig` includes a note on line 40: "The cursor is only at the bottom of the viewport. If we aren't at the bottom, we never render the cursor."

src/renderer/Thread.zig
```diff
@@ -624,7 +648,21 @@ fn cursorTimerCallback(
...
     t.flags.cursor_blink_visible = !t.flags.cursor_blink_visible;
-    t.wakeup.notify() catch {};
+
+    // Only wake up the renderer if the cursor blink state change will actually
+    // affect the rendered output. This avoids unnecessary full frame updates
+    // when cursor blinking is disabled, or when other conditions prevent the
+    // cursor from being rendered.
+    {
+        t.state.mutex.lock();
+        defer t.state.mutex.unlock();
```
mitchellh: Hm. I'm not sure if grabbing the lock on every cursor blink is the right

mitchellh (changes requested):
I'm not sure if this is the right approach. 

I think the general idea is fine, but grabbing the lock on every blink doesn't feel right to me, and is very likely why I decided to just do a thread wakeup every time instead. 

Now that the RenderState API is merged, the right way to do this may be to move the RenderState to the Thread rather than the generic renderer, update it in there and utilize that information instead.

matthew-hre:
Great points! I added the cursor mode fields to the `RenderState`, and now read from that when checking if the cursor is blinkable instead of `state.terminal.modes`. There's still a lock in order to check preedit, but most of the checks don't cause mutex contention anymore. Hopefully this is a better approach?

matthew-hre:
Closing this, since the RenderState seems to have changed significantly since I started this, and it'd be better handled by someone with fresher context on those changes.

PR #9675, opened 2025-11-23
Spinlock

qwerasd205:
This PR adds a custom lock that can be acquired by spinning briefly (10K tries) before deferring to a futex. This avoids the much larger overhead that comes from a thread having to sleep and wake to wait for a futex, when the critical area protected by the lock is much shorter than that.

I've replaced both `renderer_state.mutex` and `GenericRenderer(...).draw_mutex` with this. The `renderer_state` mutex directly impacts IO throughput since it's locked by the io-reader thread every time it processes output, and it's often contended for short periods by the renderer thread. The `draw_mutex` I don't have solid numbers on, but it should theoretically improve renderer latency by some amount.

The change of `renderer_state.mutex` gives a pretty decent improvement to our IO throughput:

### `vtebench`
image

### `DOOM-fire-zig`
`DOOM-fire-zig` went from stabilizing at around 860FPS to 900FPS for me with this change, a nice ~4.5% increase.

---
**However**, while this overhead reduction shows up brilliantly in synthetic benchmarks of IO throughput, there's a more important consequence. I noticed while testing this change against `main` that it's suspiciously less *flickery* in a lot of things (DOOM-fire-zig, various of the vtebench tests), at which point I realized that we've actually had a pretty nasty race issue under our nose this entire time that this change circumvents.

The issue is as follows:
- The renderer thread is waiting for the renderer state lock, which is held by the io-reader thread.
- The io-reader thread finishes processing a chunk of output and releases the lock, which triggers the renderer thread to start waking up.
- In the mean time, the io-reader thread reads another chunk of output and begins processing it, acquiring the lock (which it can do since it's an unfair lock, and doesn't prioritize the renderer thread even though it started trying to lock it first).
- By the time the renderer thread is awake, the lock is held again, and it goes back to sleep to wait for it once more.

This was my speculation at least, and I thought maybe it would happen once every few output reads, but upon profiling I found relatively frequent sections like this:
image
which not only confirmed my suspicion, but showed that it can happen in an arbitrarily long chain so long as the io-reader is receiving input in the right size/spacing of individual writes -- yikes!
image

So while I'm super happy about the IO throughput improvements this change yields, I'm *way* happier about eliminating this lockout problem, since it (qualitatively, I don't have hard numbers to prove this) seems to have seriously improved the *smoothness* of output when there are a lot of back-to-back writes from the program running in the terminal. It's my speculation that Ghostty's relative "flickeriness" with some TUIs/CLIs compared to other fast terminals was due to this problem, so I'm thrilled to have mitigated it.

# Boring work to do before merging
Before this is merged, I really wanna be sure that this lock doesn't somehow cause a horrible performance regression on a different OS or architecture. I don't *think* it should, but with concurrent code you never know. So these are the platforms that need to be benchmarked:
- [X] aarch64 macOS
- [ ] aarch64 Linux
- [ ] x86 macOS (intel mac)
- [ ] x86 Linux

I'll see about benchmarking aarch64 Linux in a vm on my mac in a while if no one else gets around to it before me- and if no one else can manage x86 macOS for me, I can pull my old intel mac out of storage to do it, though it's not the most convenient.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -39,6 +39,7 @@ const App = @import("App.zig");
 const internal_os = @import("os/main.zig");
 const inspectorpkg = @import("inspector/main.zig");
 const SurfaceMouse = @import("surface_mouse.zig");
+const SpinnableLock = @import("datastruct/spinnable_lock.zig").SpinnableLock;
 
 const log = std.log.scoped(.surface);
 
@@ -512,7 +513,7 @@ pub fn init(
     errdefer renderer_impl.deinit();
 
     // The mutex used to protect our renderer state.
-    const mutex = try alloc.create(std.Thread.Mutex);
+    const mutex = try alloc.create(SpinnableLock(.spin_then_block));
     mutex.* = .{};
     errdefer alloc.destroy(mutex);
 
--- a/src/datastruct/spinnable_lock.zig
+++ b/src/datastruct/spinnable_lock.zig
@@ -1,0 +1,128 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
+// TODO: Explore thread safety; we have a 32 bit lock value,
+//       so we might as well use it to store the thread ID
+//       there (ensuring that 0 isn't a valid thread ID),
+//       and then at least in debug mode add some asserts
+//       about ownership with that.
+
+/// How many times we loop when trying to obtain the lock
+/// by spinning before we give up and use the futex instead.
+///
+/// This number is arbitrary and I just based it on some other
+/// spin-then-block lock implementations I've seen in the wild.
+///
+/// NOTE(Qwerasd):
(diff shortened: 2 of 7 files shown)


qwerasd205:
Thanks to @jcollie for running vtebench for me on x86 Linux:
image
These results seem to imply that changing to a spinlock might be a little worse, but when changing the default locking mode to `.futex` it may be a little better. These are tiny differences though, and I'm still a little bit suspicious of how similar all the results of very different bechmarks here are, so I'm waiting on someone else to also run this on x86 Linux to double check. If it bares out the same conclusions then I'll switch the default locking mode to `.futex` when targeting Linux.

It's not surprising that Linux would do better in this way; a *lot* of time, energy, and money has been put in to optimizing its scheduler and locking code. Still, I would like to be more certain before making this change.

PR #9679, opened 2025-11-24
unicode: fix VS15/VS16 check to consider emoji bases

jacobsandlund:
This PR builds on https://github.com/ghostty-org/ghostty/pull/9678 ~so the diff from there is included here (it's not possible to stack PRs unless it's a PR against my own fork)--review that one first!~

This PR fixes an issue with the VS15/VS16 handling in terminal `print`, where before we didn't check for valid sequences if the base code point is an emoji. This meant that a VS15 after an emoji that isn't part of a valid sequence would narrow the cell, despite that the emoji doesn't have a text presentation.

This PR also uses a single `is_emoji_vs_base` instead of the `is_emoji_vs_emoji` and `is_emoji_vs_text`. See the [uucode comment](https://github.com/jacobsandlund/uucode/blob/215ff097303ab19e0f3a0a4e66ab3cd38c2c86ae/src/config.zig#L239-L254) for why I'm recommending this.

AI was used in some of the uucode changes in https://github.com/ghostty-org/ghostty/pull/9678 (Amp--primarily for tests), but everything was carefully vetted and much of it done by hand. This PR was made without AI.

### DIFF
--- a/src/build/uucode_config.zig
+++ b/src/build/uucode_config.zig
@@ -90,8 +90,7 @@ pub const tables = [_]config.Table{
             width.field("width"),
             d.field("grapheme_break"),
             is_symbol.field("is_symbol"),
-            d.field("is_emoji_vs_text"),
-            d.field("is_emoji_vs_emoji"),
+            d.field("is_emoji_vs_base"),
         },
     },
 };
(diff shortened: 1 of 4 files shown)


mitchellh:
Merged the base PR, can you rebase and update this? Thanks! It looks great!

jacobsandlund:
> Merged the base PR, can you rebase and update this? Thanks! It looks great!

Done!

mitchellh (approved):
Logic and tests here are great, thanks.

PR #8757, opened 2025-09-18
deps: Replace ziglyph with uucode

jacobsandlund:
This replaces `ziglyph` with [uucode](https://github.com/jacobsandlund/uucode), maintaining the Ghostty unicode props and symbols tables, but now built with `uucode` and Unicode 16.

See the above `uucode` repository for a description of the library, but the short summary is that it aims to be a fast and flexible Unicode library in Zig, that is very configurable during build time.

This PR adds tests to make sure that the symbol and props tables match the previous built-by-ziglyph implementation (minus differences in Unicode 16).

## Benchmarks

Also added are benchmarks comparing the Ghostty tables with `uucode`, with the following results (MacBook Pro 2021 M1 16 GB).

Summary:

| Benchmark | Dataset | Table Time (ms) | uucode Time (ms) | Fastest | Fastest Speedup |
|-----------|---------|-----------------|------------------|---------|----------------|
| codepoint-width | data.txt | 929.3 (± 2.3) | 924.1 (± 0.9) | uucode | 1.01x |
| | Arabic | 2,652.0 (± 21.0) | 2,589.0 (± 8.0) | uucode | 1.02x |
| | Japanese | 2,302.0 (± 1.0) | 2,250.0 (± 2.0) | uucode | 1.02x |
| | English | 1,770.0 (± 5.0) | 1,770.0 (± 5.0) | tie | 1.00x |
| grapheme-break | data.txt | 974.5 (± 1.2) | 977.4 (± 0.7) | table | 1.00x |
| | Arabic | 4,135.0 (± 22.0) | 4,037.0 (± 9.0) | uucode | 1.02x |
| | Japanese | 3,537.0 (± 3.0) | 3,459.0 (± 1.0) | uucode | 1.02x |
| | English | 4,799.0 (± 4.0) | 5,119.0 (± 2.0) | table | 1.07x |
| is-symbol | data.txt | 877.9 (± 2.0) | 883.5 (± 1.5) | table | 1.01x |
| | Arabic | 2,735.0 (± 8.0) | 2,611.0 (± 1.0) | uucode | 1.05x |
| | Japanese | 2,391.0 (± 2.0) | 2,276.0 (± 4.0) | uucode | 1.05x |
| | English | 2,450.0 (± 5.0) | 2,212.0 (± 2.0) | uucode | 1.11x |

Sources:

```
# data.txt
zig-out/bin/ghostty-gen +utf8 | head -c 200000000 > data.txt

# Arabic
[...]

Issue #5637, opened 2025-02-08
Incorrect font shaping/layouting for Indic(-derived) scripts


pluiedev:
Currently our rendering for Indic scripts (and scripts derived from/inspired by them) are quite broken, with overlapping characters (especially egregious in Tamil), incorrect vowel marks/diacritic placement, incorrect cursor placement, etc.

See previous discussions for each individual script:

- Devanagari: #2603
- Bengali: #3912
- Thai: #3753
- Tamil: #5085

247b:
The below text may contain concept errors , im happy with the ghosttys current state and im not pushing mainainers towards this bug I understand the situation and sharing my insights 

Also im excited about libghostty  and more documentation, im learning many things out the code base

Greetings, regarding the text rendering problems persisting in the tamil text, i noticed several things to share that would help the mainters to resolve issue, the thing is the shaping engine does the job right  there is no problem in combing many graphemes into  one cluster forming a character, I not sure about this but ghostty uses glsl and metal to place rasterized objects in grid formations using two triangles that i saw in the code base

so i point my fingers on the shaders that i think has been causing spacing problem, letters of tamil are not monospaced width they vary, before trying to solve the the problem i used think that i need to make a special font engineering magic to solve the issue, then i realised that it should not be done that way and the shaping engine take cares of the width metrics and substitution

To check , I made a test terminal project to simulate similar text rendering stack using Harfbuzz , Freetype & Raylib  The harfbuzz handles the shaping and spacing without fail 

thank you

jcollie:
> letters of tamil are not monospaced width they vary

This is exactly the problem… Tamil script is not monospaced, but terminals _fundamentally are_. This is a problem that's really bigger than Ghostty. If there is a terminal that renders Tamil script in a way that's acceptable to the wider Tamil community Ghostty can try and copy that behavior, but if there's no existing behavior that we can copy it's going to take a lot of R&D from the Tamil community on how to preserve the shaping of Tamil script, but also keep the monospaced nature of terminals.

247b:
omg ! i really beleived ghostty will somehow solve this issue, currently there is no pattern to handle the tamil text, like we dont even able to write apps in our own text,

vivek-dg:
[Indian-Language-support-in-gnome-terminal](http://foss.coep.org.in/fosslab/Indian-Language-support-in-gnome-terminal.pdf)

[...]

PR #9680, opened 2025-11-24
unicode: switch to uucode grapheme break to (mostly) match unicode spec

jacobsandlund:
This PR builds on https://github.com/ghostty-org/ghostty/pull/9678 ~so the diff from there is included here (it's not possible to stack PRs unless it's a PR against my own fork)--review that one first!~

This PR updates the `graphemeBreak` calculation to use `uucode`'s `computeGraphemeBreakNoControl`, which has [tests in uucode](https://github.com/jacobsandlund/uucode/blob/215ff097303ab19e0f3a0a4e66ab3cd38c2c86ae/src/x/grapheme.zig#L753) that confirm it passes the `GraphemeBreakTest.txt` (minus some exceptions).

Note that the `grapheme_break` (and `grapheme_break_no_control`) property in `uucode` incorporates `emoji_modifier` and `emoji_modifier_base`, diverging from UAX #29 but matching UTS #51. See [this comment in uucode](https://github.com/jacobsandlund/uucode/blob/215ff097303ab19e0f3a0a4e66ab3cd38c2c86ae/src/grapheme.zig#L420-L434) for details.

The `grapheme_break_no_control` property and `computeGraphemeBreakNoControl` both assume `control`, `cr`, and `lf` have been filtered out, matching the current grapheme break logic in Ghostty.

This PR keeps the `Precompute.data` logic mostly equivalent, since the `uucode` `precomputedGraphemeBreak` lacks benchmarks in the `uucode` repository (it was benchmarked in [the original PR adding `uucode` to Ghostty](https://github.com/ghostty-org/ghostty/pull/8757)). Note however, that due to `grapheme_break` being one bit larger than `grapheme_boundary_class` and the new `BreakState` also being one bit larger, the state jumps up by a factor of 8 (u10 -> u13), to 8KB.

## Benchmarks

~I benchmarked the old `main` version versus this PR for `+grapheme-break` and surprisingly this PR is 2% faster (?). Looking at the assembly though, I'm thinking something else might be causing that. Once I get to the bottom of that I'll remove the below TODO and include the benchmark results here.~

When seeing the speedup with `data.txt` and maybe a tiny speedup on English wiki, I was surprised given the 1KB -> 8KB tables. Here's what AI said when I asked it to inspect the assembly: https://ampcode.com/threads/T-979b1743-19e7-47c9-8074-9778b4b2a61e, and here's what it said when I asked it to predict the faster version: https://ampcode.com/threads/T-3291dcd3-7a21-4d24-a192-7b3f6e18cd31

It looks like two loads got reordered and that put the load that depended on stage1 -> stage2 -> stage3 second, "hiding memory latency". So that makes the new one faster when looking up the `grapheme_break` property. These gains go away with the Japanese and Arabic benchmarks, which spend more time processing utf8, and may even have more grapheme clusters too.

### with data.txt (200 MB ghostty-gen random utf8)

CleanShot 2025-11-26 at 08 42 03@2x

### with English wiki dump

CleanShot 2025-11-26 at 08 43 15@2x

### with Japanese wiki dump

CleanShot 2025-11-26 at 08 43 49@2x

### with Arabic wiki dump

CleanShot 2025-11-26 at 08 44 25@2x

TODO:

* [x] Take a closer look at the assembly and understand why this PR (8 KB vs 1 KB table) is faster on my machine.
* [x] _(**edit**: checking this off because it seems unnecessary)_ If this turns out to actually be unacceptably slower, one possibility is to switch to `uucode`'s `precomputedGraphemeBreak` which uses a 1445 byte table since it uses a dense table (indexed using multiplication instead of bitCast, though, which did show up in the initial benchmarks from https://github.com/ghostty-org/ghostty/pull/8757 a small amount.)

AI was used in some of the uucode changes in https://github.com/ghostty-org/ghostty/pull/9678 (Amp--primarily for tests), but everything was carefully vetted and much of it done by hand. This PR was made without AI with the exception of consulting AI about whether the "Prepend + ASCII" scenario is common (hopefully it's right about that being uncommon).


mitchellh:
Rebase again due to merge, thanks :)

jacobsandlund:
> Rebase again due to merge, thanks :)

Done!

mitchellh:
This looks awesome. I want to run a `vtebench`, cat a non-ASCII file, and also see what our wcwidth test results are after this before merging, if possible. I can help with this when I have time. :smile:

jacobsandlund:
> This looks awesome. I want to run a `vtebench`, cat a non-ASCII file, and also see what our wcwidth test results are after this before merging, if possible. I can help with this when I have time. 😄

Sounds good. I haven't gotten to run `vtebench` on it yet, but I will. I updated the description with the `+grapheme-break` benchmarks.

mitchellh:
In this PR: 

CleanShot 2025-11-26 at 13 32 11@2x

jacobsandlund:
@mitchellh I see the same thing. for reference for others, this is main:

CleanShot 2025-11-26 at 17 45 43@2x

---
So, this PR drops the score from 106 to 98.

And then if I run ucs-detect on https://github.com/jacobsandlund/ghostty/pull/1, built on top of this PR, I see:

CleanShot 2025-11-26 at 17 51 19@2x

---
Dropping the score again to 91.

I'll do some more investigation, but ucs-detect is based on python wcwidth, which is treating too many categories of characters as zero width: https://github.com/jquast/wcwidth/blob/915166f9453098a56e87a7fb69e697696cefe206/bin/update-tables.py#L149-L155

`Mc` (which isn't even present in that comment) is _spacing combining mark_, which I think should generally be treated as taking up _space_, so that the character should be wide. Here's an amp thread: https://ampcode.com/threads/T-93ae4196-0abf-43db-8f26-f1d0ea869dba

Here you can see the https://github.com/jacobsandlund/ghostty/pull/1 result of printing a couple of those graphemes, and selecting it correctly groups the cluster, and it displays wide as it should:

CleanShot 2025-11-26 at 18 29 32@2x

CleanShot 2025-11-26 at 18 36 21@2x

and here's the result on main:

CleanShot 2025-11-26 at 18 32 56@2x

CleanShot 2025-11-26 at 18 36 51@2x

just out of curiosity, i changed the uucode `wcwidth` calculation to treat `Mc` and `Cf` as 0 to match ucs-detect, on top of  https://github.com/jacobsandlund/ghostty/pull/1 and I get this score:

CleanShot 2025-11-26 at 18 40 22@2x

---

Interestingly, still down 1 point from main. But, I think we can't treat `ucs-detect` here as the source of truth. I could open a PR on `ucs-detect`/`wcwidth` for fixes that it needs, or even make a `uucode` one.

mitchellh:
> Interestingly, still down 1 point from main. But, I think we can't treat ucs-detect here as the source of truth. I could open a PR on ucs-detect/wcwidth for fixes that it needs, or even make a uucode one.

I agree with this, we may want to also loop in @jquast too who can maybe provide some opinions. :smile:

pluiedev:
> Here you can see the https://github.com/jacobsandlund/ghostty/pull/1 result of printing a couple of those graphemes, and selecting it correctly groups the cluster, and it displays wide as it should:

This reminds me... this looks very similar to the overlapping characters seen in #5637 — maybe it wasn't a font shaping/layout problem but rather that we simply weren't assigning the correct cell widths for these characters?

jacobsandlund:
I'm still investigating this, but I'll share my findings thus far. I have a branch on top of https://github.com/jacobsandlund/ghostty/pull/1 that compares the cumulative `advance.width` for a grapheme cluster using Coretext (I'm on Mac) and the expected Ghostty or uucode width, and logs when there's a mismatch: https://github.com/jacobsandlund/ghostty/compare/grapheme-width-changes...debug-width

I run `ucs-detect.sh` and tee those logs to a file, then I repeat but with `Mc` and `Cf` treated as zero width. That results in 133,865 lines in the former case and 144,257 in the latter.

Then I do some `sed` and `sort -u` and get basically just the failing graphemes. There are _some_ repeats, because I'm also including `x: <advance.width>`, and that apparently can change in different shaping runs sometimes (I need to understand shaping more). That gives me 4952 failing graphemes for the first case (with a little double counting), and 5799 failing graphemes for `Mc` and `Cf` as zero width.

Here's the diff as a gist, with additions being graphemes that are now wrong with the `Mc` and `Cf` treated as zero width: https://gist.github.com/jacobsandlund/03a9f052198b5cd5b28a0e5e803ebf80

And some lines, inline:

```diff
-\u{11341} → 𑍁    x: 15.263850999996066
 \u{11341}\u{0}\u{11302} → 𑍁    x: 31.586641117930412
 \u{11341}\u{0}\u{11303} → 𑍁    x: 25.883497582748532
-\u{11343} → 𑍃    x: 11.315520860254765
-\u{1134d} → 𑍍    x: 16.005108382552862
-\u{1134d} → 𑍍    x: 6.4292732160538435
+\u{11342} → 𑍂    x: 20.558546589687467
 \u{1703} → ᜃ    x: 28.2716427154541
 \u{1704} → ᜄ    x: 24.0481351146698
 \u{1704}\u{1714} → ᜄ᜔    x: 24.0481351146698
@@ -1044,9 +1157,12 @@
 \u{1711} → ᜑ    x: 30.13918008995056
 \u{1711}\u{1712} → ᜑᜒ    x: 30.139180089950557
 \u{1780} → ក    x: 20.769738256931305
+\u{1780} → ក    x: 31.30658107995987
 \u{1780}\u{17b7} → កិ    x: 20.769738256931305
 \u{1780}\u{17bb} → កុ    x: 20.769738256931305
 \u{1780}\u{17bb}\u{17c6} → កុំ    x: 20.769738256931305
+\u{1780}\u{17be} → កើ    x: 33.02894961833954
+\u{1780}\u{17c1} → កេ    x: 33.02894961833954
 \u{1780}\u{17c4} → កោ    x: 43.5657924413681
 \u{1780}\u{17c6} → កំ    x: 20.769738256931305
 \u{1780}\u{17cb} → ក់    x: 20.769738256931305
@@ -1058,11 +1174,17 @@
 \u{1780}\u{17d2}\u{178a}\u{17c5} → ក្ដៅ    x: 43.5657924413681
 \u{1780}\u{17d2}\u{179a}\u{17c4} → ក្រោ    x: 53.90000367164612
 \u{1781} → ខ    x: 20.769738256931305
+\u{1781} → ខ    x: 31.30658107995987
+\u{1781}\u{17b7} → ខិ    x: 20.769738256931305
 \u{1781}\u{17bb} → ខុ    x: 20.769738256931305
 \u{1781}\u{17c6} → ខំ    x: 20.769738256931305
 \u{1781}\u{17d2} → ខ្    x: 20.769738256931305
+\u{1781}\u{17d2} → ខ្    x: 31.30658107995987
 \u{1782} → គ    x: 20.769738256931305
+\u{1782} → គ    x: 31.30658107995987
+\u{1782}\u{17b6}\u{17c6} → គាំ    x: 31.30658107995987
 \u{1782}\u{17b7} → គិ    x: 20.769738256931305
+\u{1782}\u{17c1} → គេ    x: 33.02894961833954
 \u{1782}\u{17c4} → គោ    x: 43.5657924413681
 \u{1782}\u{17c6} → គំ    x: 20.769738256931305
```

I'll keep looking a little closer, but I'll also reach out and comment on https://github.com/jquast/wcwidth/issues/155

jacobsandlund:
Here's the result of my investigation, as a comment on the `jquast/wcwidth#155` issue: https://github.com/jquast/wcwidth/issues/155#issuecomment-3630555803

mitchellh:
CI isn't running for some reason, but you've always been diligent so I'm going to trust you on this. Thank you.

PR #9685, opened 2025-11-24
Fix LangSet.hasLang() to compare against FcLangEqual instead of FcTrue

pyrytakala:
FcLangSetHasLang returns FcLangResult enum values:
- FcLangEqual (0): Exact match
- FcLangDifferentTerritory (1): Same language, different territory
- FcLangDifferentLang (2): Different language
See also https://www.freedesktop.org/software/fontconfig/fontconfig-devel/fclangsethaslang.html and https://codebrowser.dev/qt6/include/fontconfig/fontconfig.h.html#_FcLangResult

The previous comparison to FcTrue (1) caused:
- Exact matches (0) to incorrectly return false
- Partial matches (1) to incorrectly return true

This fix changes the comparison to FcLangEqual (0) so hasLang() correctly returns true only for exact language matches.

### DIFF
--- a/pkg/fontconfig/lang_set.zig
+++ b/pkg/fontconfig/lang_set.zig
@@ -11,8 +11,12 @@ pub const LangSet = opaque {
         c.FcLangSetDestroy(self.cval());
     }
 
+    pub fn addLang(self: *LangSet, lang: [:0]const u8) bool {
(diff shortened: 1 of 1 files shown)


mitchellh:
I very much appreciate this. FontConfig is a very difficult API to use correctly, and hard to learn due to lack of docs, so I appreciate insights like this. :) Thanks!

PR #9687, opened 2025-11-24
Search binding, viewport rendering

mitchellh:
The march towards #189 continues.

This hooks up the search thread to the main surface and all the state necessary for the renderer to show search results in the viewport! This also adds a `search` binding which takes a query to start/stop a search. **This still doesn't add any search GUI,** which will come later, the internals must happen first.

A non-blank binding will start or change the search term. An empty binding will stop search:

```
keybind = cmd+f=search:Hello
keybind = shift+cmd+f=search:
```

> [!NOTE]
>
> Obviously, search will eventually have a GUI. The point of this PR is primarily to connect all the various internal systems more than anything. GUI will come soon.

## Demo

https://github.com/user-attachments/assets/06af5a3b-280e-4804-b506-419b92a95f99

## Major Changes

The only major changes required as part of this is the introduction of what I'm calling the terminal "highlight" system. This is a generic system for highlighting portions of the terminal contents. These will ultimately underpin what we currently call "selections" (selecting text with your mouse/keyboard) but that is far too large a change to make in one PR. 

Therefore, this PR introduces highlights and the only consumer is the entire search subsystem. 

## Limitations

Still plenty of limitations we need to keep marching towards resolving:

- ~~Search matches are styled the same as selections. They should be different.~~
- There is no way to iterate search matches, yet.
- There is no GUI for inputting a search query or viewing total matches, yet.
- ~~I think searches are case-sensitive currently and they should probably not be.~~ Done, for ASCII.

But hey, it's something!


mitchellh:
~~There's a bug here where new viewport results aren't being pushed to the renderer thread when they show up even though we are processing them, so just noting that so I can fix that.~~  Resolved.

karlhorky:
> - I think searches are case-sensitive currently and they should probably not be.

Maybe it would also make sense to ignore diacritics? So if you search for `go`, it would also highlight `gö`? Could be a setting, default `true`...

And potentially also normalizing homoglyphs... 🤔

pluiedev:
If we really want to delve into the weeds, we probably need to implement [UTS #10 (Collations)](https://www.unicode.org/reports/tr10/) to satisfy all the case-/diacritic-/etc.-insensitivity for everyone's needs... :p But that is far beyond the scope of this PR

mitchellh:
> If we really want to delve into the weeds, we probably need to implement [UTS #10 (Collations)](https://www.unicode.org/reports/tr10/) to satisfy all the case-/diacritic-/etc.-insensitivity for everyone's needs... :p But that is far beyond the scope of this PR

Indeed. We can do a basic ASCII insensitive to start, easily enough.

mitchellh:
Added `search-foreground` and `background` to specify the style of search matches.

mitchellh:
Searches are now case-insensitive for ASCII.

karlhorky:
> @pluiedev in [comment 3573660927](https://github.com/ghostty-org/ghostty/pull/9687#issuecomment-3573660927): If we really want to delve into the weeds, we probably need to implement [UTS #10 (Collations)](https://www.unicode.org/reports/tr10/) to satisfy all the case-/diacritic-/etc.-insensitivity for everyone's needs... :p But that is far beyond the scope of this PR

Nice, I was not aware that there was a standard around that! I'm guessing one of the parts you're referring to is `11.2 Asymmetric search`?

> When doing an asymmetric search, a character (or grapheme cluster) in the query that is unmarked at the secondary and/or tertiary levels will match a character in the target that is either marked or unmarked at the same levels, but a character in the query that is marked at the secondary and/or tertiary levels will only match a character in the target that is marked in the same way.

scratchmex:
wow I can finally switch to ghostty. thanks for the work

mitchellh:
> wow I can finally switch to ghostty. thanks for the work

There's still no GUI to input search queries, only bindings. But close!

felipeorlando:
This is the only missing thing to use Ghostty for me

mitchellh:
Okay

felipeorlando:
@mitchellh any plan for the next release?

mitchellh:
Yes, many.

felipeorlando:
@mitchellh I mean, the date that this finding feature will be released... 🤔

mitchellh:
Yes, it is published online in the 1.2 release notes

0xbadshah:
> Ghostty 1.3 will continue the focus of making Ghostty the ["best existing terminal emulator"](https://mitchellh.com/writing/ghostty-1-0-reflection) by shipping the last remaining major missing features to achieve parity with other popular terminal emulators. Namely, we plan on shipping scrollback search and scrollbars for 1.3, at a minimum.[2](https://ghostty.org/docs/install/release-notes/1-2-0#user-content-fn-1)

Source: https://ghostty.org/docs/install/release-notes/1-2-0#roadmap

Looking forward to Ghostty 1.3 🚀

PR #9688, opened 2025-11-24
add default keybinding for ctrl+shift+plus

kaddkaka:
I tried to create a default keybinding in line with this discussion: https://github.com/ghostty-org/ghostty/discussions/7243

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -6006,6 +6006,18 @@ pub const Keybinds = struct {
             .{ .key = .{ .unicode = '+' }, .mods = inputpkg.ctrlOrSuper(.{}) },
             .{ .increase_font_size = 1 },
         );
+        // This key-binding make the intuitive keybinding work for layouts that
+        // achieves plus by using the shift key (like UK)
+        //try self.set.put(
+        //    alloc,
+        //    .{ .key = .{ .unicode = '+' }, .mods = inputpkg.ctrlOrSuper(.{ .shift = true }) },
+        //    .{ .increase_font_size = 1 },
+        //);
+        try self.set.put(
+            alloc,
+            .{ .key = .{ .physical = .equal }, .mods = inputpkg.ctrlOrSuper(.{ .shift = true }) },
+            .{ .increase_font_size = 1 },
+        );
 
         try self.set.put(
             alloc,


mitchellh:
Fixed by #9469 I believe.

kaddkaka:
> Fixed by #9469 I believe.

I don't think so. I press `ctrl`+`shift`+`=` but the font size does not increase, instead `;5u` gets put into my prompt.

I built  on `main` just now using `zig bulid --release=fast` and this is my build config:

```console
$ ./zig-out/bin/ghostty --version
Ghostty 1.3.0-main+5b38e0bd7

Version
  - version: 1.3.0-main+5b38e0bd7
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - app runtime   : .gtk
  - libxev        : io_uring
  - kernel version: 6.17.0-8-generic
  - desktop env   : other
  - GTK version   :
    build         : 4.20.1
    runtime       : 4.20.1
  - libadwaita    : enabled
    build         : 1.8.0
    runtime       : 1.8.0
  - libX11        : enabled
  - libwayland    : enabled
```

My environment:
- Ubuntu 25.10
- Keyboard layout: English UK (I produce `+` by pressing `shift` and `=`).

danielwe:
@kaddkaka you're right that #9469 doesn't make `ctrl+shift+=` work on Linux. The binding is still `ctrl+=` if your keyboard has an `=` key.

As it happens, the equivalent `cmd+shift+=` does work on macOS because the `cmd++` binding is added to the menu accelerator and is thus matched by the OS, triggering the menu action, rather than being matched within Ghostty and triggering the keybind action. macOS matches shifted keystrokes with equivalent unicode bindings, so, e.g., the `shift+=` keystroke matches a binding for `+` if that's appropriate for your keyboard layout. Ghostty itself does not do this---a binding for `+` can only be matched if your keyboard has a key that inserts `+` without using shift.

I think the right thing to do here would be to look into matching shifted keys to unicode, like macOS does, rather than adding extra bindings on an ad-hoc basis. This is also necessary for creating things like layout-independent vim key tables, which need to match, e.g., `:`, regardless of how that character is typed on your keyboard.

(Any further conversation about this should probably happen in https://github.com/ghostty-org/ghostty/discussions/7243#discussioncomment-15600509. I just didn't want to leave this thread hanging.)

PR #9691, opened 2025-11-24
Fix `pkg/freetype` LoadFlags struct to correctly match FreeType API

pyrytakala:
The struct was missing padding at bit position 8, causing all subsequent flag fields (bits 9+) to be misaligned by one bit position.

See: https://freetype.org/freetype2/docs/reference/ft2-glyph_retrieval.html#ft_load_xxx


qwerasd205:
Wow, thanks, looking closer there's actually more wrong with this-- the load targets are completely wrong- I'll go ahead and fix that too and add it to this PR with a much more comprehensive unit test

qwerasd205 (approved):

PR #9695, opened 2025-11-25
fix(renderer): load linearized fg color for cursor cell

danielwe:
It's clear from the surrounding code that the RGB values returned from `cell_text_vertex` are expected to be linearized. This was not the case for the cursor cell, resulting in the cursor text being too bright when using `cursor-text = cell-background`.

Fixes #8353/#9138.

xref #8960, which would also fixe this issue, but I don't know where that's heading


danielwe:
**Before**
Screenshot 2025-11-24 at 21 13 24

**After**
Screenshot 2025-11-24 at 21 18 38

mitchellh (approved):

PR #9696, opened 2025-11-25
pkg/{highway,simdutf}: disable ubsan

mitchellh:
This causes linker issues for some libghostty users. I don't know why we never saw these issues with Ghostty release builds, but generally speaking I think its fine to do this for 3rd party code unless we've witnessed an issue. And these deps have been stable for a long, long time.

### DIFF
--- a/pkg/highway/build.zig
+++ b/pkg/highway/build.zig
@@ -67,6 +67,10 @@ pub fn build(b: *std.Build) !void {
         "-fno-cxx-exceptions",
         "-fno-slp-vectorize",
         "-fno-vectorize",
+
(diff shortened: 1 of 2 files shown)


qwerasd205:
We never saw problems in Ghostty builds because of https://github.com/ghostty-org/ghostty/blob/6f0927c42a9f2267133c7ee3c5204683b677f2b9/src/build/GhosttyLib.zig#L40
I'm pretty sure

mitchellh:
That makes sense. In the future we may want to make this a build option so that we can have ubsan in Ghostty but libghostty users don't need to pay the price of bundling it for deps they don't control.

PR #9705, opened 2025-11-25
termio: report color scheme synchronously

### LINKED ISSUE #5922: Report color scheme should happen synchronously
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5796

<sup>Originally posted by **JakeWharton** February 15, 2025</sup>
Right now, querying theme state is async, and is usually delivered after responses to subsequent VT sequences.
```
❯ echo -ne '\e[?996n\e[5n`'; cat
`^[[0n^[[?997;1n^C⏎

❯ echo -ne '\e[?996n\e[5n`'; cat
`^[[?997;1n^[[0n^C
```
Here two executions of a theme query followed by DSR show that the theme response can come before the DSR or after, and it is usually after in my testing.

This is because the code is effectively double dispatching. The parser for the theme request delegates the query to the surface: https://github.com/ghostty-org/ghostty/blob/9a5bc650342f66135a61316c557243b929380d0a/src/termio/stream_handler.zig#L773

The surface, upon receiving the dispatch: https://github.com/ghostty-org/ghostty/blob/9a5bc650342f66135a61316c557243b929380d0a/src/Surface.zig#L951

Enqueues a response: https://github.com/ghostty-org/ghostty/blob/9a5bc650342f66135a61316c557243b929380d0a/src/Surface.zig#L1000-L1005

And this races the subsequent DSR parse and response.

This behavior is in contrast with something like focus, which as of #2103 responds synchronously with the current value: https://github.com/ghostty-org/ghostty/blob/9a5bc650342f66135a61316c557243b929380d0a/src/termio/stream_handler.zig#L648-L650

```
❯ echo -ne '\e[?1004h\e[5n'; cat
^[[I^[[0n^C⏎
```

**Is it possible to push theme state into the terminal so that it can respond to queries synchronously?** I am working on a TUI library whose startup process sometimes sees the system theme query return before our DSR marker reply to be reflected in the initial state correctly, but more often than not sees the DSR marker reply before the system theme query response causing an unnecessary frame of potentially incorrect rendering.</div>

## Implementation Note

We can update `termio.DerivedConfig` to pull the `Config._conditional_state` so that we don't need to go back to the core Surface for this information. That way, we can avoid the async.

We should remove the `report_color_scheme` action from Surface if its otherwise unused.

tobiaskohlbau:
The reporting of color scheme was handled asynchronously by queuing a handler in the surface. This could lead to race conditions where the DSR is reported after subsequent VT sequences.

Fixes #5922

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1073,8 +1073,6 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
 
         .scrollbar => |scrollbar| self.updateScrollbar(scrollbar),
 
-        .report_color_scheme => |force| self.reportColorScheme(force),
-
         .present_surface => try self.presentSurface(),
 
         .password_input => |v| try self.passwordInput(v),
@@ -1384,26 +1382,6 @@ fn passwordInput(self: *Surface, v: bool) !void {
     };
 
     try self.queueRender();
-}
-
-/// Sends a DSR response for the current color scheme to the pty. If
-/// force is false then we only send the response if the terminal mode
-/// 2031 is enabled.
-fn reportColorScheme(self: *Surface, force: bool) void {
-    if (!force) {
-        self.renderer_state.mutex.lock();
-        defer self.renderer_state.mutex.unlock();
-        if (!self.renderer_state.terminal.modes.get(.report_color_scheme)) {
-            return;
-        }
-    }
-
-    const output = switch (self.config_conditional_state.theme) {
-        .light => "\x1B[?997;2n",
-        .dark => "\x1B[?997;1n",
-    };
-
-    self.queueIo(.{ .write_stable = output }, .unlocked);
 }
 
 fn searchCallback(event: terminal.search.Thread.Event, ud: ?*anyopaque) void {
@@ -5039,7 +5017,7 @@ pub fn colorSchemeCallback(self: *Surface, scheme: apprt.ColorScheme) !void {
     self.notifyConfigConditionalState();
 
     // If mode 2031 is on, then we report the change live.
-    self.reportColorScheme(false);
+    self.queueIo(.{ .color_scheme_report = .{ .force = false } }, .unlocked);
 }
 
 pub fn posToViewport(self: Surface, xpos: f64, ypos: f64) terminal.point.Coordinate {
--- a/src/termio/message.zig
+++ b/src/termio/message.zig
@@ -16,6 +16,12 @@ pub const Message = union(enum) {
     /// in the future.
     pub const WriteReq = MessageData(u8, 38);
 
+    /// Request a color scheme report is sent to the pty.
+    color_scheme_report: struct {
+        /// Force write the current color scheme
+        force: bool,
+    },
+
     /// Purposely crash the renderer. This is used for testing and debugging.
     /// See the "crash" binding action.
     crash: void,
--- a/src/termio/stream_handler.zig
+++ b/src/termio/stream_handler.zig
@@ -119,7 +119,7 @@ pub const StreamHandler = struct {
         };
 
         // The config could have changed any of our colors so update mode 2031
-        self.surfaceMessageWriter(.{ .report_color_scheme = false });
+        self.messageWriter(.{ .color_scheme_report = .{ .force = false } });
     }
 
     inline fn surfaceMessageWriter(
@@ -871,7 +871,7 @@ pub const StreamHandler = struct {
                 self.messageWriter(msg);
             },
 
-            .color_scheme => self.surfaceMessageWriter(.{ .report_color_scheme = true }),
+            .color_scheme => self.messageWriter(.{ .color_scheme_report = .{ .force = true } }),
         }
     }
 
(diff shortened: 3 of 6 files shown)


src/termio/stream_handler.zig
```diff
@@ -53,6 +53,9 @@ pub const StreamHandler = struct {
     /// The color reporting format for OSC requests.
     osc_color_report_format: configpkg.Config.OSCColorReportFormat,
 
+    /// The config conditional state.
+    config_conditional_state: configpkg.ConditionalState,
```
tobiaskohlbau: I'm unsure if this is the desired outcome of if the underlying scheme value should be pulled in directly. Let me know if this should be changed.

tobiaskohlbau:
This is my first try at contributing back to Ghostty, please let me know if I missed something prior to implementation. Thanks to the awesome issue description it was relatively straightforward to resolve it. I've tested the changes on my machine and running the commands from the issue multiple times resulted in always the same order of the responses.

I've tried to attach an execution log but GitHub mangles with the escape codes in the code block.

mitchellh (changes requested):
It seems this doesn't fully faithfully convert this, since we're ignoring the mode flag in some cases where we previously didn't.

I don't have a problem with copying the conditional state.

tobiaskohlbau:
Thanks for the review, just to clarify because I'm not that familiar. By mode flag do you refer to the _force_ argument previously being present? I've removed it in places where it was always used with _false_ and forcefully send the mode flag unconditionally in cases where it should by possible async changes (in the Surface.zig).

mitchellh:
I'll re-review later, but it looks like both `true` and `false` values were converted to an unconditional report scheme.

tobiaskohlbau:
That's right I've made the assumption that in case it's requested in termio it always needs to be responded to, cause if that code is reached the request was explicitly driven by request. But I could be mistaken here, I will think about it once again.

src/Surface.zig
```diff
@@ -1307,16 +1305,12 @@ fn passwordInput(self: *Surface, v: bool) !void {
...
-        defer self.renderer_state.mutex.unlock();
-        if (!self.renderer_state.terminal.modes.get(.report_color_scheme)) {
-            return;
-        }
+/// Sends a DSR response for the current color scheme to the pty.
+fn reportColorScheme(self: *Surface) void {
+    self.renderer_state.mutex.lock();
+    defer self.renderer_state.mutex.unlock();
+    if (!self.renderer_state.terminal.modes.get(.report_color_scheme)) {
+        return;
```
tobiaskohlbau: I've removed the _force_ mode here, cause the only consumer left was calling it with _false_ which the current behavior preserves. In #5922 it was suggested to check if the functionality could be removed completely. But I think this is not possible cause an change in the system color scheme needs to propagate the change through.

tobiaskohlbau:
You were right, I've found two instances where I mistakenly changed the behavior. I've reintroduced the force flag and updated the callers. 


~If I transplant the mutex handling from `Surface.zig` into _reportColorScheme_ living in `stream_handler.zig` I experienced deadlocks. I've looked at other places in `stream_handler.zig` and found it unlocks and relocks afterwards (the other way around). I've tested this and it works on my setup, but I'm not really confident this is the right thing to do.~

Update:
I've searched up the call stack of the reportColorScheme function and came to the conclusion every caller already holds the mutex and therefore an additional lock is root cause of the deadlock. I've added and comment in the documentation about the callers responsibility.

src/Surface.zig
```diff
@@ -1336,16 +1334,12 @@ fn passwordInput(self: *Surface, v: bool) !void {
...
     }
```
mitchellh, line 1400: Instead of duplicating this logic, we should add a new termio message so it's all sent from the same thread. Writes like this are already not synchronous ANYWAYS so that should be fine.

mitchellh (changes requested):
Thanks for the changes, and sorry for the continued slow review. I've suggested one more thing. I'm working my way through PR backlogs now, got down 30 PRs the last week. 😄

tobiaskohlbau:
Now I was out for vacation, sorry for the delay on my side. I hope I understood your suggestion and implemented in the way you envisioned. PTAL and let me know if things needs to be updated.

src/termio/message.zig
```diff
@@ -16,6 +16,12 @@ pub const Message = union(enum) {
     /// in the future.
     pub const WriteReq = MessageData(u8, 38);
 
+    /// Request a color scheme report is sent to the pty.
+    color_scheme_report: struct {
```
jcollie, line 20: Renaming from _verb_-_noun_ to _noun_-_verb_ seems unnecessary, and a bit counter to existing practice.

tobiaskohlbau: I've named it according to https://github.com/ghostty-org/ghostty/blob/1c2db85aa41c7d929cb23b408229e297dce2758a/src/termio/Thread.zig#L321

I'm happy to rename it to whatever is preferred.

mitchellh (approved):

PR #9709, opened 2025-11-26
macOS Search GUI

mitchellh:
#189 for macOS

This adds a search GUI for macOS, including macOS-standard menu bar items, and keybindings that match other native macOS applications such as Terminal app, Safari, and others. This introduces a new keybinding action `start_search` to start a blank search for GUIs.

This PR also resolves a number of minor issues found in the search subsystem and renderer related to search from prior PRs. This should result in overall improved search stability. **Please note there are still known issues (bottom of this PR).**

> [!WARNING]
>
> **A note on stability:** I know a lot of people are eager to have this feature, and I'm excited
> for this feature to soon be available in tip releases. But note that new features like this are
> always filled with performance issues, bugs, crashes, etc. That's the point of tip releases:
> to find and address these before wider availability. We will do our best to respond rapidly to
> major issues found in tip, but don't expect perfect functionality immediately!

## Demo

https://github.com/user-attachments/assets/3b81752e-d7e5-4875-9864-92497333b23e

> [!NOTE]
>
> You can drag the search window to any of the four corners if its getting in the way of reading the terminal.

## Known Issues

TODO prior to this PR merging:

- [x] Single-byte search terms cause a crash since our sliding window can't handle it. This PR temporarily requires search terms with length 2+ before starting a search to avoid it. 😄 
- [x] The way `ScreenSearch` prunes history is fundamental unsafe. With a rapidly growing screen that could reach history limits and an active search at the same time, Ghostty is almost guaranteed to crash currently. The workaround is to not search actively scrolling screens (new data) for now.


heri16:
+1

ftpd:
@mitchellh very good progress, thanks!

Since #189 is locked, I'll allow myself to ask here:
- do you consider regex/glob queries in the search?
- will it be possible to do 'something' with the match that was found? Copy it to the clipboard and/or paste to the command line?

mitchellh:
@ftpd Sorry, this is not a place for discussion about future functionality.

mitchellh:
Broke my own rules with this one, but for **AI disclosure:** Amp was used for the GUI (visuals, not logic). I reviewed it all. Everything else is organic.

injust:
A few things I noticed (LMK if you prefer me opening new/separate discussions):

~~"End Search" appears twice in the command palette~~ (fixed in https://github.com/ghostty-org/ghostty/commit/4ff0e0c9d251ddd0687ead578a90d58d63f9840b):
- `command-palette-entry = title:"End Search",...` appears twice in the default config, but with different descriptions/actions

esc and cmd-shift-f are both bound to `end_search`, but neither immediately closes the search UI:
- `end_search`'s command palette description is "End the current search if any and hide any GUI elements", so it seems like it should close the search UI immediately
- When the search field is focused, <kdb>esc</kbd> shifts the focus back to the shell without closing the search field, and it takes a 2nd <kdb>esc</kbd> to close the search UI
- cmd-shift-f seems to do nothing when the search field is focused

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -1112,6 +1112,22 @@ class BaseTerminalController: NSWindowController,
...
+    @IBAction func find(_ sender: Any) {
+        focusedSurface?.find(sender)
+    }
+    
+    @IBAction func findNext(_ sender: Any) {
+        focusedSurface?.findNext(sender)
+    }
+    
+    @IBAction func findPrevious(_ sender: Any) {
+        focusedSurface?.findNext(sender)
```
lrytz, line 1125: typo?

```suggestion
        focusedSurface?.findPrevious(sender)
```

pluiedev: Please open a new discussion instead of commenting on merged PRs

PR #9713, opened 2025-11-26
macOS: move search result counter inside text field

avarayr:
Move the search result counter inside the search text field using an overlay, preventing layout shift when results appear.

**Before:** The counter appeared as a separate element in the HStack, causing the text field to shift when results loaded.

**After:** The counter is overlaid inside the text field on the right side with reserved padding, eliminating layout shift.

---

**AI Disclosure: The was entirely authored with Claude Code, specifically with Claude Opus 4.5.**


mitchellh (approved):
Thanks, tested and looks good.

PR #9715, opened 2025-11-26
fix(macos): use strings' utf-8 lengths for libghostty calls

qwerasd205:
Swift conveniently converts strings to UTF-8 encoded cstrings when passing them to external functions, however our libghostty functions also take a length and we were using String.count for that, which returns the number of _characters_ not the byte length, which caused searches with multi-byte characters to get truncated.

I went ahead and changed _all_ invocations that pass a string length to use the utf-8 byte length even if the string is comptime-known and all ASCII, just so that it's proper and if someone copies one of the calls in the future for user-inputted data they don't reproduce this bug.

ref:
https://developer.apple.com/documentation/swift/string/count https://developer.apple.com/documentation/swift/stringprotocol/lengthofbytes(using:)

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -180,14 +180,14 @@ extension Ghostty {
 
         func newTab(surface: ghostty_surface_t) {
             let action = "new_tab"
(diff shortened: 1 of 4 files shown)


mitchellh (approved):
Approved pending CI. Long term I wanna hide this behind a nicer Swift API so we won’t have to get it right N times. Haha. Thanks!

PR #9717, opened 2025-11-26
macOS: refine Search UI

bo2themax:
- fix search dragging animation when corner is not changed
- use ConcentricRectangle on Tahoe

https://github.com/user-attachments/assets/64c294a7-70bd-484d-b153-4aa2d734b1a7


mitchellh (approved):
Looks great thank you!

PR #9720, opened 2025-11-26
macos: search - avoid freeing sliding window meta during prune

avarayr:
Fixes a crash when searching while the buffer is growing rapidly:

  1. Open Ghostty
  2. Run `yes`
  3. Cmd+F and search for `y`
  4. crash

---

Root cause: pruning in `SlidingWindow.highlight` freed meta allocations still referenced by returned highlights. We now clear the meta cell maps while retaining their allocations, so pruning stays bounded without freeing live data.

AI disclosure: code reasoning and wording with ChatGPT; change authored and understood here.

### DIFF
--- a/src/terminal/search/sliding_window.zig
+++ b/src/terminal/search/sliding_window.zig
@@ -439,14 +439,12 @@ pub const SlidingWindow = struct {
 
         // If we went beyond our initial meta node we can prune.
         if (tl.prune.meta > 0) {
-            // Deinit all our memory in the meta blocks prior to our
-            // match.
+            // Clear prior meta blocks but retain their allocations to avoid
+            // freeing memory still referenced by a returned highlight.
             var meta_it = self.meta.iterator(.forward);
-            var meta_consumed: usize = 0;
             for (0..tl.prune.meta) |_| {
                 const meta: *Meta = meta_it.next().?;
-                meta_consumed += meta.cell_map.items.len;
-                meta.deinit(self.alloc);
+                meta.cell_map.clearRetainingCapacity();
             }
             if (comptime std.debug.runtime_safety) {
                 assert(meta_it.idx == tl.prune.meta);


mitchellh (changes requested):
Meta doesn't contain any memory referenced by anyone. This could use a test too to demonstrate the fix.

avarayr:
> Meta doesn't contain any memory referenced by anyone. This could use a test too to demonstrate the fix.

 Thanks for the review. As my understanding is, the crash wasn’t from external references, but from `meta.deinit()` on recycled slots after `deleteOldest:` the circbuf resets the slot to its default (`undefined` cell_map), so the next prune frees an invalid pointer. 
 
Clearing the `cell_map` during prune avoids `deinit`ing those recycled slots and stops the allocator error we hit when searching while `yes` floods the buffer.

I tried to capture this in a unit test by forcing circbuf reuse/prune, but the allocator crash seems to be timing/allocator-sensitive in the harness and I haven’t gotten a deterministic failure there yet. The live repro (yes + search for `y`) consistently triggers it with the old code and is fixed by this change.

I can keep pushing on a regression, otherwise I’d appreciate landing the minimal fix with the real-world repro as justification.

mitchellh:
> As my understanding is, the crash wasn’t from external references, but from meta.deinit() on recycled slots after deleteOldest: the circbuf resets the slot to its default (undefined cell_map), so the next prune frees an invalid pointer.

Why would the next prune access items beyond the length of the circular buffer?

mitchellh:
I think at this point we need some clear test to demonstrate this fix. There is no "timing" related to allocators. You can setup an exact page state as necessary to trigger this bug if it happens. I do witness a crash doing the yes search, but I can't accept this without understanding.

avarayr:
Ah, looks like I was fighting with a symptom here. Closing this in favor of #9722—thanks for fixing it!

mitchellh:
I don't know or think #9722 fixes it. That fixes a problem, but I'm still able to get #9721

avarayr:
> I don't know or think #9722 fixes it. That fixes a problem, but I'm still able to get #9721

hmm, fwiw i just pulled #9722 and tested `yes` + search `y` and it no longer crashes on release build

mitchellh:
I think that's luck, since I tested your use case in my build that doesn't have it and it doesn't happen either. There's likely memory corruption happening (probably related to this safety check triggering) and it takes awhile to explode. I'm pretty confident it is somehow related to page pruning and tons of output though so I think its not terrible, but definitely something we need to find and resolve before a tagged release.

avarayr:
Yup, you're right, I spoke too soon. Crash happens sporadically even with #9722. 
somewhat consistent repro on 9722 is run `yes` + search for `yes` + backspace backspace 

---

On the other hand, this PR seems comprehensively fixes the crash. I’ve tried pretty hard to bottle this into a unit test and to reason about it in SlidingWindow, but I don't yet have a small, deterministic regression. The only reliable reproducer I have is still the live `yes` + search scenario.

The real fix probably belongs in the PageList/pin/pruning logic and needs a better‑understood regression.

mitchellh:
Closing this due to a lack of understanding and explanation for this fix. It is referenced in the linked issue so we can always review it as we try to find the root cause for this. A reproduction is also there so we can continue to dig into it. I won't merge code we can't explain. Thank you for trying, though.

PR #9747, opened 2025-11-28
macos: `window-width/height` is accurate even with other widgets

mitchellh:
Fixes #2660

Rather than calculate our window frame size based on various chrome calculations, we now utilize SwiftUI layouts and view intrinsic content sizes with `setContentSize` to setup our content size ignoring all our other widgets.

I'm sure there's some edge cases I'm missing here but this should be a whole lot more reliable on the whole.

### DIFF
--- a/macos/Sources/Helpers/Extensions/NSWindow+Extension.swift
+++ b/macos/Sources/Helpers/Extensions/NSWindow+Extension.swift
@@ -15,4 +15,20 @@ extension NSWindow {
         guard let firstWindow = tabGroup?.windows.first else { return true }
         return firstWindow === self
     }
+
+    /// Adjusts the window origin if necessary to ensure the window remains visible on screen.
+    func constrainToScreen() {
+        guard let screen = screen ?? NSScreen.main else { return }
+        let visibleFrame = screen.visibleFrame
+        var windowFrame = frame
+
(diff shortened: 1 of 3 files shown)


macos/Sources/Helpers/Extensions/NSWindow+Extension.swift
```diff
@@ -15,4 +15,20 @@ extension NSWindow {
...
     }
+
+    /// Adjusts the window origin if necessary to ensure the window remains visible on screen.
+    func constrainToScreen() {
+        guard let screen = screen ?? NSScreen.main else { return }
+        let visibleFrame = screen.visibleFrame
+        var windowFrame = frame
+
+        windowFrame.origin.x = max(visibleFrame.minX,
+            min(windowFrame.origin.x, visibleFrame.maxX - windowFrame.width))
```
bo2themax, line 26:
```suggestion
            min(windowFrame.origin.x, visibleFrame.width - windowFrame.width))
```

Could be some RTL issues here, but I don't know if Ghostty is fully compatible with that before. Need some testing to confirm.

bo2themax: Alright, RTL still needs to be polished anyway. Made a discussion for this: #9750

bo2themax (approved):

PR #9756, opened 2025-11-30
GTK: Search UI

### LINKED ISSUE #189: GTK: Search scrollback

mitchellh:
Fixes #189 

This adds the UI for search to GTK. There is still polish to be done in follow-ups but this makes search work well with GTK to start! 

**AI disclosure:** Believe it or not, almost this entire PR was AI-written. Amp did an excellent job looking at our existing codebase, comparing it to the macOS codebase, writing blueprint files, etc. I reviewed everything written, modified some basics, and verified it manually and under Valgrind.

## Demo

CleanShot 2025-11-29 at 20 52 11@2x

## Future Improvements

- When dragging the overlay, we should change the cursor and show drop targets
- There's probably some small stylistic tweaks we can make to this
- I'm not sure the CSS is right for both light and dark modes so we may need to tweak that

### DIFF
--- a/src/apprt/gtk/class/search_overlay.zig
+++ b/src/apprt/gtk/class/search_overlay.zig
@@ -48,6 +48,20 @@ pub const SearchOverlay = extern struct {
         };
     };
 
+    pub const signals = struct {
+        /// Emitted when the search is stopped (e.g., Escape pressed).
+        pub const @"stop-search" = struct {
+            pub const name = "stop-search";
+            pub const connect = impl.connect;
+            const impl = gobject.ext.defineSignal(
+                name,
+                Self,
+                &.{},
+                void,
+            );
+        };
+    };
+
     const Private = struct {
         /// The search entry widget.
         search_entry: *gtk.SearchEntry,
@@ -67,6 +81,13 @@ pub const SearchOverlay = extern struct {
         const priv = self.private();
         _ = priv.search_entry.as(gtk.Widget).grabFocus();
         priv.search_entry.as(gtk.Editable).selectRegion(0, -1);
+    }
+
+    //---------------------------------------------------------------
+    // Template callbacks
+
+    fn stopSearch(_: *gtk.SearchEntry, self: *Self) callconv(.c) void {
+        signals.@"stop-search".impl.emit(self, null, .{}, null);
     }
 
     //---------------------------------------------------------------
@@ -121,10 +142,16 @@ pub const SearchOverlay = extern struct {
             // Bindings
             class.bindTemplateChildPrivate("search_entry", .{});
 
+            // Template Callbacks
+            class.bindTemplateCallback("stop_search", &stopSearch);
+
             // Properties
             gobject.ext.registerProperties(class, &.{
                 properties.active.impl,
             });
+
+            // Signals
+            signals.@"stop-search".impl.register(.{});
 
             // Virtual methods
             gobject.Object.virtual_methods.dispose.implement(class, &dispose);
@@ -133,5 +160,6 @@ pub const SearchOverlay = extern struct {
 
         pub const as = C.Class.as;
         pub const bindTemplateChildPrivate = C.Class.bindTemplateChildPrivate;
+        pub const bindTemplateCallback = C.Class.bindTemplateCallback;
     };
 };
--- a/src/apprt/gtk/class/search_overlay.zig
+++ b/src/apprt/gtk/class/search_overlay.zig
@@ -34,39 +34,39 @@ pub const SearchOverlay = extern struct {
     });
 
     pub const properties = struct {
-        pub const duration = struct {
-            pub const name = "duration";
+        pub const active = struct {
+            pub const name = "active";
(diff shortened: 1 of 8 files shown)


src/apprt/gtk/class/search_overlay.zig
```diff
@@ -0,0 +1,390 @@
...
+
+    const Private = struct {
+        /// The search entry widget.
+        search_entry: *gtk.SearchEntry,
+
+        /// True when a search is active, meaning we should show the overlay.
+        active: bool = false,
+
+        /// Total number of search matches (-1 means unknown/none).
+        search_total: i64 = -1,
```
pluiedev: Not sure if I like using sentinel values for this (the private struct is a regular old Zig struct, and it allows optional values unlike `extern struct`s). Also, why `i64`?  This should just be a `?usize` and the `@intCast`s should only occur when interoperating with the C API

src/apprt/gtk/class/search_overlay.zig
```diff
@@ -0,0 +1,390 @@
...
+
+    fn init(self: *Self, _: *Class) callconv(.c) void {
+        gtk.Widget.initTemplate(self.as(gtk.Widget));
+    }
+
+    /// Grab focus on the search entry and select all text.
+    pub fn grabFocus(self: *Self) void {
+        const priv = self.private();
+        _ = priv.search_entry.as(gtk.Widget).grabFocus();
+        priv.search_entry.as(gtk.Editable).selectRegion(0, -1);
```
pluiedev, line 233: Probably would be a good idea to note that a negative index `-n` for the end position refers to the last `n`th character

src/apprt/gtk/class/search_overlay.zig
```diff
@@ -0,0 +1,390 @@
...
+        const value: i64 = if (selected) |s| @intCast(s) else -1;
+        if (priv.search_selected == value) return;
+        priv.search_selected = value;
+        self.as(gobject.Object).notifyByPspec(properties.@"search-selected".impl.param_spec);
+    }
+
+    fn closureMatchLabel(_: *Self, selected: i64, total: i64) callconv(.c) ?[*:0]const u8 {
+        if (total <= 0) return glib.ext.dupeZ(u8, "0/0");
+        var buf: [32]u8 = undefined;
+        const label = std.fmt.bufPrintZ(&buf, "{}/{}", .{
```
pluiedev, line 285: This should ideally be localizable, but we can do that in a different PR

pluiedev (changes requested):

src/apprt/gtk/class/search_overlay.zig
mitchellh: We can't because these values are passed directly to a C calling convention for the template closure. 

The alternative would be to introduce maybe more properties like `has-search-total` etc.

pluiedev: Oh really? Bummer... I personally would prefer separate properties for these, just to keep the Zig struct cleaner though

src/apprt/gtk/class/search_overlay.zig
mitchellh: I figured this was a safe one to avoid localization, but wouldn't surprise me to learn that `x/y` is not global lol.

pluiedev: Yeah in some languages (e.g. Chinese) the order of `x` and `y` is reversed :) But again we can do this in another PR

src/apprt/gtk/class/search_overlay.zig
mitchellh: Switched to the `has-x` style.

mitchellh:
Going to send this so we have something in main. Happy to continue to address reviews with future PRs.

xlab:
Congrats on landing the feature we couldn't get from https://github.com/kovidgoyal/kitty/issues/893 for 7 years!

:tada:

unphased:
Just a quick question, this is a feature I didn't realize was important but is very much so, and I'm glad to see it landing. however, it's not in a release yet right? It will land in 1.2.4 or 1.3.0?

mitchellh:
Correct. 1.3 slated for march. Complex features need time to bake

iRubbish:
👍

PR #9757, opened 2025-11-30
PageList: preserve size.cols in adjustCapacity after column shrink

rockorager:
When columns shrink during resize-without-reflow, page.size.cols is
updated but page.capacity.cols retains the old larger value. When
adjustCapacity later runs (e.g., to expand style/grapheme storage),
it was creating a new page using page.capacity which has the stale
column count, causing size.cols to revert to the old value.

This caused a crash in render.zig where an assertion checks that
page.size.cols matches PageList.cols.

Fix by explicitly copying page.size.cols to the new page after
creation, matching how size.rows is already handled.

Amp-Thread-ID: https://ampcode.com/threads/T-976bc49a-7bfd-40bd-bbbb-38f66fc925ff
Co-authored-by: Amp <amp@ampcode.com>

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -2609,6 +2609,7 @@ pub fn adjustCapacity(
     const new_page: *Page = &new_node.data;
     assert(new_page.capacity.rows >= page.capacity.rows);
     new_page.size.rows = page.size.rows;
+    new_page.size.cols = page.size.cols;
     try new_page.cloneFrom(page, 0, page.size.rows);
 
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Looks good, I made a change locally (to add an assertion) but for some reason can't push it up to your anch, so I'll push to main.

mitchellh:
https://github.com/ghostty-org/ghostty/commit/a58e33c06bfdefec663a879b5f056f11a3d41a24

PR #9758, opened 2025-11-30
macos: teach agents about `zig build run`

jparise:


mitchellh:
Can you explain where this is useful? I can't remember where I put it, but I'd prefer AGENTS changes to be associated with some public threads of agentic sessions showing this is useful.

jparise:
> Can you explain where this is useful? I can't remember where I put it, but I'd prefer AGENTS changes to be associated with some public threads of agentic sessions showing this is useful.

Locally (with Claude Code, in this case), when it was helping me with some test cases, it would build the code with `zig build` and then tell me to run `<path>/ghostty/zig-out/bin/ghostty` for manual testing steps.

This change helped with those instructions, but it's otherwise not part of any other agentic workflow I've come across yet.

Issue #2425, opened 2024-10-09
macOS: support for mouse button 8/9  (128, 129)


ldemailly:
The core already supports this, but the macOS app doesn't properly encode these yet.

_Originally posted by @ldemailly in https://github.com/ghostty-org/ghostty/pull/2424#discussion_r1794023769_

mitchellh:
Updating this after #9761 to note this is only a macOS issue.

PR #9761, opened 2025-11-30
core: encode mouse buttons 8 & 9 (back/forward)

neurocyte:
For some reason button 8 & 9 support is missing.


mitchellh:
Note this doesn't complete #2425 since macOS doesn't send anything more than left/right/middle atm. I'll still accept this because it looks like GTK does encode these buttons already.

mitchellh (approved):

PR #9764, opened 2025-11-30
macOS: change `window` to `new-window` for `macos-dock-drop-behavior`

bo2themax:
Matches current option references and Swift implementation

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -86,6 +86,10 @@ pub const compatibility = std.StaticStringMap(
     // Ghostty 1.2 removed the "desktop" option and renamed it to "detect".
     // The semantics also changed slightly but this is the correct mapping.
     .{ "gtk-single-instance", compatGtkSingleInstance },
+
+    // Ghostty 1.3 rename the "window" option to "new-window".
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Since this was already in a released version, we'd need a conversion for backwards compatibility. They're setup at the top of the file and please copy the comment format.

bo2themax:
> Since this was already in a released version, we'd need a conversion for backwards compatibility. They're setup at the top of the file and please copy the comment format.

Sure, I'll add that later

mitchellh (changes requested):
Sorry for the late delay, I was really context loaded on something else. The compatibility should go into the `const compatibility` structure and not directly in the logic. This will let us -- in the future -- warn on compatibility stuff and also integrate it as part of docs.

bo2themax:
> Sorry for the late delay, I was really context loaded on something else. The compatibility should go into the `const compatibility` structure and not directly in the logic. This will let us -- in the future -- warn on compatibility stuff and also integrate it as part of docs.

Updated, but i wasn't sure about which version, so I put 1.3 in the comments

mitchellh (approved):

PR #9765, opened 2025-11-30
GTK: update search state from previous search on activation

rhodes-b:
I think it makes sense that if there was already text from the previous search that it should match based on that, this doesnt remove that it highlights everything so you can press backspace to delete and start from a empty search


mitchellh (approved):
This is a fine way to do it, but a better way would be to hook `active::changed` and only do it on the false => true, since the grabFocus side should already be searching in most cases (after active is true).

PR #9783, opened 2025-12-02
Add close tabs on the right action

phaistonian:
image

### DIFF
--- a/src/input/command.zig
+++ b/src/input/command.zig
@@ -541,7 +541,7 @@ fn actionCommands(action: Action.Key) []const Command {
             .{
                 .action = .{ .close_tab = .right },
                 .title = "Close Tabs to the Right",
-                .description = "Close every tab to the right of the current one.",
+                .description = "Close all tabs to the right of the current one.",
             },
         },
 
--- a/src/input/command.zig
+++ b/src/input/command.zig
@@ -538,6 +538,11 @@ fn actionCommands(action: Action.Key) []const Command {
                 .title = "Close Other Tabs",
                 .description = "Close all tabs in this window except the current one.",
             },
+            .{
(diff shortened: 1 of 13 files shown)


mitchellh (changes requested):
Thanks for this. I think we should support this feature but I think  we should make this macOS only for now because I'm not sure if it's typical for this to behave this way on GTK. We can easily add it later. 

I'll see if I can clean this up.

mitchellh:
Actually just going to remove the context menu for now, the tab view API call is fine.

mitchellh (approved):
Great. I made some changes here:

- Removed GTK UI (but kept binding action functionality)
- Rename to "Close Tabs _to_ the Right" instead of "on", because that's what Apple apps use
- Change to selector comparison for tab bar context menu detection
- Change to using an NSMenu extension so we can more easily add more stuff in the future
- Ensure the target is the correct target so you can right click non-focused tabs
- Added `xmark` icon to close targets to match Safari

src/input/command.zig
```diff
@@ -538,6 +538,11 @@ fn actionCommands(action: Action.Key) []const Command {
                 .title = "Close Other Tabs",
                 .description = "Close all tabs in this window except the current one.",
             },
+            .{
+                .action = .{ .close_tab = .right },
+                .title = "Close Tabs to the Right",
+                .description = "Close every tab to the right of the current one.",
```
tristan957: I wonder if it would be a good idea to make the commend and description the same:

```zig
/// Close all tabs to the right of the current tab.
```

mitchellh: yes

tristan957:
Very cool that macOS supports this. I can see from Chrome on macOS that the string is exactly the same. Awesome.

PR #9785, opened 2025-12-02
Sesssion searcn

phaistonian:
image

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -71,6 +71,7 @@ class AppDelegate: NSObject,
     @IBOutlet private var menuQuickTerminal: NSMenuItem?
     @IBOutlet private var menuTerminalInspector: NSMenuItem?
     @IBOutlet private var menuCommandPalette: NSMenuItem?
+    @IBOutlet private var menuSearchSessions: NSMenuItem?
 
     @IBOutlet private var menuEqualizeSplits: NSMenuItem?
     @IBOutlet private var menuMoveSplitDividerUp: NSMenuItem?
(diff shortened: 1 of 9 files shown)


mitchellh (changes requested):
Didn't review it just blocking the merge so I can look at this.

phaistonian:
@mitchellh  I’ve expanded the search to include panes as well. Also, when switching to a tab or pane from the session search, there’s now a quick flash on both the tab and the pane to make it easy to see where the focus moved.

https://github.com/user-attachments/assets/62f3d8e9-87bf-4bdc-a1d3-54a0a63cef80

Here’s [the binary](https://d.pr/f/AQMVIb) with all of these changes (plus the ones from the previous PRs) in case you want to try it out — they’ve honestly helped my sanity quite a bit. :)

phaistonian:
@mitchellh  Fun side-note: I had some fun exposing a preview of highlighted pane in this one, but the thing is that I can not find a way to either get it outside the app or properly contain in it. It does, somehow work, and it can be very handy, but it's not polished :)

image

mitchellh:
Reusing the command palette view for this is clever. There's a lot of duplicate work in here, we need to rebase and clean that up to move forward (to start). I think there'll be more clean up after if we pursue.

deanm:
Great that this is getting worked on. I have some experience with this sort of feature from other apps, a few things to think about:

- Make sure whatever kind of search you do it "stable" or has some deterministic behaviour. Meaning if two windows have the same title, what order are they presented in, and how are they disambiguated. Generally you start to develop some sort of muscle memory so it's good they are either always in a consistent order (for example oldest created first), or some sort of window stacking order, etc. I personally suggested oldest first.

- Disambiguation: depending how you take the "titles" or other things you are searching, there is also more information you can pull in, for example in iTerm2 you can have a bunch of windows called "bash" but it will also show you the CWD (but unfortunately for whatever reason that is displayed but not searched against).

- Ranking, if you do something like a fuzzy match / etc, generally it's important to rank higher prefix matches, exact matches, then fuzzy matches, etc. Then you tiebreak as discussed in the first bullet point.

- Preview is a fun idea but I think it won't really be used in practice, this is one of these alt-tab type features that you want to just be really quick and responsive, shortcut then type a few characters and hit enter. Worst case you have to navigate through the list with the arrow keys (and I also suggest hooking up tab / shift-tab and ctrl-n / ctrl-p for going up and down through the list).

mitchellh:
Replaced by #9945

PR #9816, opened 2025-12-05
macos: Don't encode option as alt in kitty mode

neurocyte:
This is the same fix as #9406 applied to kitty keyboard.

Note: I cannot personally test this as I do not have a mac, but there has been some user testing in the flow community. See the discussion at https://github.com/neurocyte/flow/discussions/413

### DIFF
--- a/src/input/key_encode.zig
+++ b/src/input/key_encode.zig
@@ -116,7 +116,20 @@ fn kitty(
         }
     }
 
-    const all_mods = event.mods;
+    const all_mods = mods: {
+        var mods_binding = event.mods;
+        if (comptime builtin.target.os.tag.isDarwin()) alt: {
+            switch (opts.macos_option_as_alt) {
+                .false => {},
+                .true => break :alt,
+                .left => if (event.mods.sides.alt == .left) break :alt,
+                .right => if (event.mods.sides.alt == .right) break :alt,
+            }
+            mods_binding.alt = false;
+        }
+        break :mods mods_binding;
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Seems believable to me (would like to compare with Kitty itself). 

Separately, everything in this file has a unit test and this should too (even for macOS-specific stuff, we guard on Darwin).

neurocyte:
> Seems believable to me (would like to compare with Kitty itself).

Kitty notably does **not** do this and it seems unlikely that it will. See https://github.com/kovidgoyal/kitty/issues/9203 for the full discussion.

> Separately, everything in this file has a unit test and this should too (even for macOS-specific stuff, we guard on Darwin).

Sure, if you intend to merge this I can add a unit test.

mitchellh:
> Sure, if you intend to merge this I can add a unit test.

Roger, too early to say atm. I'll do some quick research.

danielwe:
xref another discussion about this: #9511

mitchellh:
Note I apologize for not getting this in yet. I admit I just haven't taken the time yet to fully understand the problem, the arguments for Kitty NOT doing this, and formulating if we break. Historically when it comes to protocols like Kitty's we try to copy even the bugs unless there is a very good reason not to, since bug-equivalence in behavior is often better for overall app ecosystem compatibility than correctness (if we're not the protocol creator). But, I still don't know enough to make that determination.

PR #9818, opened 2025-12-05
core: selection and copy bindings need to hold the big lock

mitchellh:
This was found by LLM hunting! We were not holding the lock properly during these operations. There aren't any known cases where we can directly attribute these races to issues but we did find at least one consistent crash for a user when `linkAtPos` wasn't properly locked (in another PR).

This continues those fixes.

https://ampcode.com/threads/T-6fc49f25-7f2f-4039-adb4-f86aaeced6d5

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -5032,8 +5032,9 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
         },
 
         .copy_to_clipboard => |format| {
-            // We can read from the renderer state without holding
-            // the lock because only we will write to this field.
+            self.renderer_state.mutex.lock();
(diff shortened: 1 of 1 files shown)


src/Surface.zig
```diff
@@ -5032,8 +5032,9 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
         },
 
         .copy_to_clipboard => |format| {
-            // We can read from the renderer state without holding
-            // the lock because only we will write to this field.
```
mitchellh, line 5036: This is an outdated comment! Since our switch to ScreenSets, this now does pointer chasing that can be set by the IO thread. The LLM was able to figure this out, amazingly.

PR #9841, opened 2025-12-08
fix(macos): close search ui on escape key press

YeungKC:
On macOS, pressing the `ESC` key while the Search UI is active currently only returns focus to the terminal but keeps the Search UI open. This behavior is inconsistent with other platforms (GTK/Linux) and standard UI patterns.

This PR adds a call to `onClose()` in the `onExitCommand` modifier for the macOS `SurfaceSearchOverlay`. This ensures that the search state is cleared (triggering `end_search` action) and the UI is dismissed when `ESC` is pressed, aligning the behavior with the Linux implementation.

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView.swift
+++ b/macos/Sources/Ghostty/SurfaceView.swift
@@ -432,6 +432,7 @@ extension Ghostty {
                     }
 #if canImport(AppKit)
                     .onExitCommand {
+                        onClose()
                         Ghostty.moveFocus(to: surfaceView)
                     }
 #endif


YeungKC:
cc @mitchellh

jcollie:
@YeungKC please do not ping maintainers trying to get them to review your PR. We're all volunteers here and PRs will get reviewed when they get reviewed.

mitchellh:
This logic is currently done on purpose. There is a discussion determining which side we want to change.

devxoul:
fyi: https://github.com/ghostty-org/ghostty/discussions/9790

PR #9845, opened 2025-12-08
terminal: parse kitty text sizing protocol (OSC 66)

pluiedev:
Implements the VT side of #5563

OSC 66 is something we *should* implement also on the terminal state/renderer side, but for now we can just parse the code and do nothing with it


mitchellh (changes requested):
This looks good to me pending the CI failures.

pluiedev:
Gonna reimpl this with the new OSC parser (which should also make my life much easier) :)

PR #9850, opened 2025-12-08
build: skip git version detection when used as dependency

rockorager:
When ghostty is used as a Zig dependency, detect this by comparing the build root with ghostty's source directory. Skip git detection entirely and use the version from build.zig.zon.

This fixes build failures when downstream projects have git tags that don't match ghostty's version format. Previously, ghostty would read the downstream project's git tags and panic at Config.zig:246 with:

\`\`\`
tagged releases must be in vX.Y.Z format matching build.zig
\`\`\`

**Reproduction:**
1. Create a project that uses ghostty as a Zig dependency
2. Tag the project with a version like \`v0.2.0\`
3. Run \`zig build\` → panic

**Fix:**
Use \`@src().file\` to get ghostty's source directory and compare it with \`b.build_root\`. When they differ, ghostty is a dependency and we skip git detection.

Thread: https://ampcode.com/threads/T-197e6c33-b8f8-4b23-8fc8-7f6b6edd9f35

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const assert = std.debug.assert;
 const builtin = @import("builtin");
 const buildpkg = @import("src/build/main.zig");
+
 const appVersion = @import("build.zig.zon").version;
 const minimumZigVersion = @import("build.zig.zon").minimum_zig_version;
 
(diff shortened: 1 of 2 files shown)


mitchellh:
Conceptually makes sense. I think there is a cleaner way to detect if you're the build root. I think you can use `@import("root")` tricks...

rockorager:
Oh good point. I'll find a nicer way

PR #8469, opened 2025-08-30
feat(macos): add tab support for QuickTerminal

AmadeusK525:
Base implementation comes from https://github.com/ghostty-org/ghostty/pull/5370 by @sohsatoh

PR #5370, opened 2025-01-25
[macOS] feat: Support Tabs in Quick Terminal

sohsatoh:
related to #2329 

This pull request implements a tab management feature for the Quick Terminal in the Ghostty for macOS.

macOS does not support tab management without a title bar. Therefore, this PR implements non-native tabs instead of using macOS native tabs.

## Changes

- Add tab management feature for Quick Terminal

### window-theme = dark

スクリーンショット 2025-01-26 19 21 54

### window-theme = light

スクリーンショット 2025-01-26 19 22 18

## Notes

The implementation is designed to work similarly to native tabs and to behave consistently with the (standard) terminal windows in Ghostty. However, there may be some aspects I might have overlooked. Please feel free to make any necessary fixes to the implementation as needed.

PR #9857, opened 2025-12-10
feat(macos): add tab support for QuickTerminal - continued

adrum:
This builds off https://github.com/ghostty-org/ghostty/pull/8469, so merging this new PR would also mark that one merged.

I resolved the merge conflict, and improved the tab closing experience with cmd+w when multiple tabs are open. If only one tab remains, it will just hide the QuickTerminal (current behavior).

Handled in https://github.com/ghostty-org/ghostty/pull/8469
- [x]     The color of the focused tab is the opposite of what it should be
- [x]     When I create a new tab with the keybinding on macOS 15.6.2, the new terminal isn't focused
- [x]     When I click to focus any tab, the terminal isn't focused
- [x]     Closing a tab with the X doesn't trigger surface close conformation is required (needsConfirmQuit)
- [x]     Creating hundreds of tabs makes the tab bar mostly unusable (no close targets, no titles).
- [x]     Creating a new tab deletes all open splits in the quick terminal and makes the quick terminal uncloseable with cmd+w

I handled: 
- [x]     When I press cmd+w (default binding) on macOS 15.6.2, the whole quick terminal window is closed instead of a single tab
- [x]     goto_tab:N key binds do nothing
- [x]     previous/next_tab key binds do nothing
- [x]      Dragging the tab onto a terminal produces a UUID, it should probably do nothing for now (future: should make a separate window)
- [x]     File => New Tab does nothing
- [x]     The tab title doesn't update for the focused split

I added this:
- [x]     Opening a new tab when number of tabs exceeds the bounds of the quick terminal window should scroll to the newly opened tab
- [x]      Dragging the tab onto a terminal produces a separate window/tab. Dropping on another ghostty windows tab bar will add that tab to the end of that window.
- [x]   Added keybind indicators to the right of the tab for goto_tab actions
- [x]   Added state restoration support for quick terminal tabs
- [x]   Ability to change tab title
- [x]   Ability to change tab color
- [x]   Support glass effect like normal tabs

## AI Disclosure

Claude Code was used to generate a lot of the fixes I implemented. I cannot speak to how the code was written in https://github.com/ghostty-org/ghostty/pull/8469, but I know I was methodical in reviewing the code after each request.

The area of the code I'm least happy with is the use of AppKit APIs to handle drag and drop, as I was hoping to keep it all SwiftUI. I found the SwiftUI APIs very limiting when it came to drag and drop, especially when moving a quick terminal tab to a new window, or rearranging the the Tab in the Tab Bar.

I manually edited all XIBs in (menu items) in Xcode, as I didn't want the XML to be invalid with latest versions of Xcode. I noticed macOS injects a lot of the menu items related to window management, so I had to put the Move to Quick Terminal tab in its own section.

Using the AppKit delegate methods for NSDrag operations allowed me to achieve the experience I was looking for, especially on the Tab Bar (placeholder sizing, surface detection, etc.)

Over time, I tried keeping features for quick terminal features in upstream consistent with normal tabs.

I also implemented my own state restoration logic originally, but noticed upstream also implemented restoration logic for the QuickTerminal. I opted to use upstream's methodology, but extended it to support multiple tabs, colors, and title overrides.


mitchellh:
Haven't run this yet, but did you fix these issues? https://github.com/ghostty-org/ghostty/pull/8469#issuecomment-3246954067

adrum:
I'll mark this as draft and double check each of those. Stay tuned.

adrum:
I've been running ghostty with the changes listed at the top, and it seems to be working pretty well for me.

adrum:
I decided to add the ability to change the title and set the color, similar to recent additions to normal tabs. Let me know if you would like to see anything else.

smithjw:
Thanks for adding this feature folks. Have been looking to switch from iTerm, and honestly, multiple tabs in the QuickTerminal were the only feature I was waiting on! 🤣

DoctorRainer:
@smithjw I hope it goes through this time around, haha.

christomitov:
@mitchellh sorry for the tag but would be nice to have this feature!

ashtonmeuser:
@mitchellh, is there anything we can do to move this forward? AFAIK, the issues raised in https://github.com/ghostty-org/ghostty/pull/8469#issuecomment-3246954067 have been addressed. If there are additional issues blocking this from being merged, it'd be great to have those listed.

adrum:
I apologize for this PR experiencing scope creep. As I was pulling in upstream and using over the past few months, I wanted the QuickTerminal tabs to have the same type of functionality normal macOS tabs have been recieving. 

If you would like me to split out the functionality across multiple PRs to limit the scope of the review process, I can definitely do that. I believe https://github.com/ghostty-org/ghostty/pull/9857/commits/7f1dcbd5706ec1cfeb9300c2c321536818475a33 would be the cut off point, and then I could merge upstream to handle any conflicts.

The other commits could be cherry picked and grouped into their own PRs. It would be more PRs to review, but it would be a bit easier to digest and limit the scope. 

I have been using this branch as my only build of Ghostty the past few months, and could not imagine life without it. I'm looking forward to the day when the Debug Build banner goes away.

hdwatts:
@mitchellh - I saw you comment on HN that 1.3 is around the corner, is there any way we can get this PR merged in? It's one of the only features keeping me (and many others, based on the discussions) from being able to migrate over to Ghostty.

Let me know if I can help in any way!

ashtonmeuser:
Per Discord re: this PR:

> just fyi, we're a week or two away from the 1.3 release and everyone on the team is currently busy polishing it - even if we review this soon, we're in a de facto feature freeze so this will have to wait for 1.4

yam-liu:
**Bug Report: Tab not destroyed after Control+D shell exit in Quick Terminal**

I noticed that when a new tab is created in Quick Terminal and the shell is exited via Control+D, the tab remains visible (transparent/empty) instead of being automatically closed and switching to the previous tab.

**Root Cause**

There are two issues in QuickTerminalController.swift:
1. closeSurface — processExited branch clears surfaceTree instead of closing the tab
When the shell process exits, closeSurface is called and hits the surface.processExited branch, which currently does:
```swift
if surface.processExited {
    surfaceTree = .init()
    return
}
```

2. surfaceTreeDidChange — the to.isEmpty filter condition never matches
The fallback logic in surfaceTreeDidChange attempts to find dead tabs:
```swift
if to.isEmpty {
    tabManager.tabs
        .filter { tab in
            tab.surfaceTree.contains { $0.surface == nil }
        }
        .forEach { tab in
            tabManager.closeTab(currentTab)
        }
}
```
The condition $0.surface == nil checks whether the Zig surface pointer is nil, but a process exit doesn't nil out the surface — it only sets processExited = true. So this filter never matches, and closeTab is never called.
Additionally, the .forEach closure captures currentTab instead of using the iteration variable tab, which would be a bug even if the filter worked.

> Disclosure: Claude created the report. I've tested it for the given scenario. Please assess other potential effects of this change.

yam-liu:
Bug report: Themes are not updated in the background tabs upon theme change automatically. This means the current one changed, but others not. 

> Separate Light and Dark Themes
> Ghostty supports specifying separate light and dark themes.
>
> ```
> theme = dark:Catppuccin Frappe,light:Catppuccin Latte
> ```
> 
> When separate light and dark themes are specified, Ghostty will automatically switch between the light and dark theme based on the system appearance.

adrum:
@yam-liu I just pushed a fix for the ctrl+d bug report. I'm having trouble reproducing the theme changing issue on macOS 26.3. Whenever I manually changed the theme at the system level, QuickTerminal did honor the separate themes for both dark and light mode correctly.

yam-liu:
@adrum Try reproducing using these steps:
1. Open 3 tabs in the quick terminal window. Light appearance.
2. Change to dark in system settings.
3. Open quick term using shortcut.
4. Current one is dark, others are light. 

I'll attach a video when I am available.

adrum:
@yam-liu I just merged the latest commits from main into this branch, and I'm still unable to reproduce the issue. All quick terminal tabs are switching between light and dark mode according to system settings. Can you try with the latest commits and let me know if the issue persists?

Can you also share your macOS version?

yam-liu:
@adrum I can't replicate the theme issue at the moment neither. I believe we can ignore it.

abjugard:
I just built @adrum’s branch, works beautifully!

janschloss:
This is the only missing feature for me, thanks for submitting this PR!

iwex:
Hello! When I close the last tab in the terminal, it just hides the quick window. I guess it should really close that tab.
@adrum

adrum:
@iwex So this is actually existing ghostty behavior. If you cmd+w today, it simply hides rather than end that terminal's session. I think this behavior could be changed in a sparate MR from this if desired.

Mayurifag:
To keep this comment useful, suggesting a nice to have idea for **_another_** MR: configurable cfg setting to have tabs on bottom/top.

Also +1 for idea, that closing last tab (at least in quick terminal) has to end session and opening quick terminal next time should be fresh session. It is just a default in other terminals, isn't it? Not a big deal, though.

Some ranting from my soul, you may skip it: 

I wish all the people involved in this MR all possible strength to get code merged. Thats the most needed feature (alongside Windows support) for a lot of people like me, who still are not using Ghostty as a main terminal, yet are monitoring things in a hope of bright future to come. And lots of time passed already. I feel its hard for PR author to always rebase this huge feature after new releases changes.

I mean its really long time all those PRs with the feature existed and people want to have clarification whats next, why its not merged already. I know, project is not payed by me and for fun and stuff, I cant demand anything, sorry for that, yet the most needed MR is getting _ghosted_.

solofeed:
@mitchellh 
I can understand the hesitation to merge this as is
Even if the feature is useful it touches a lot of parts of the app and adds complexity in places that many users may never interact with

Would it make sense to ship this first as an optional or experimental macOS only feature, maybe behind a config flag rather than as core behavior by default?

If a plugin-style or side feature approach is possible in Ghostty’s architecture, that might be a good middle ground: people who want it can enable it while keeping the default app behavior and maintenance surface smaller

Again, thanks for the attention to the matter :)

aliaksandrsen:
I’d like to add another data point in favor of this feature.

I currently use iTerm2 almost exclusively in dropdown/hotkey mode, and tabs inside that quick terminal are a core part of my workflow. Ghostty’s Quick Terminal is very close to being a replacement for me, but the lack of tabs is currently a blocker.

For my use case, I’d be happy with behavior similar to iTerm2’s hotkey/dropdown window: a quick terminal that can contain multiple tabs and can be shown/hidden via a global shortcut. I don’t need the implementation to perfectly match native macOS tabs or preserve every visual detail; the important part is being able to keep multiple terminal sessions inside the quick terminal.

Thanks to everyone working on this — this is probably the main thing keeping me from switching to Ghostty full-time.

ashtonmeuser:
> lack of tabs [in quick terminal] is currently a blocker

Yeah, I think many of us are in the same boat. I've tried reaching the devs re: pushing this PR along to no avail. I don't want to encourage spamming the team (which would likely have the opposite effect), but the [Discord dev channel](https://discord.com/channels/1005603569187160125/1005603569711452192) seems to be the only other place to voice an opinion.

With the amount of work that has gone into this PR (and the three progenitors) and the number of folks looking forward to using the feature, it would be a shame for it to go stale.

abjugard:
> > lack of tabs [in quick terminal] is currently a blocker
> 
> Yeah, I think many of us are in the same boat.

I encourage everyone who is in said boat to just build this branch themselves, I have yet to find an issue with it, and have now finally been able to switch from iTerm2 to this custom Ghostty build!

chrispelzer:
> > > lack of tabs [in quick terminal] is currently a blocker
> > 
> > 
> > Yeah, I think many of us are in the same boat.
> 
> I encourage everyone who is in said boat to just build this branch themselves, I have yet to find an issue with it, and have now finally been able to switch from iTerm2 to this custom Ghostty build!

The branch has been working for me, since March, as well with no issues. Though it's missing any commits since March 1st. I'm hoping it gets added with 1.4

adrum:
I just merged upstream into this branch.  I've been compiling my builds with `zig build -Doptimize=ReleaseFast`.

josegonzalez:
Not sure if these are issues with ghostty in general or the tab support in this PR but:

- I am unable to paste anything into the `ctrl+f` "search" section when searching a tab
- Sometimes when I paste in content into the tab, it will paste it in at the first character in the line. An example of this is I will type `git push origin `, then attempt to paste in the branch, but it will move the character position to the first in the line so it will do something like `$BRANCHgit push origin `.

Mayurifag:
I am sorry to bother the huge amount of subscribed to this issue people, but does anyone know whats the best way to ping the maintainers? 

I'd consider its better to merge it now as is, — and if fixes would be needed, those will go through new PRs. Updates on the feature will be way easier to merge for maintainers if someone wants to enhance QuickTerminal more.

00-kat:
> I am unable to paste anything into the `ctrl+f` "search" section when searching a tab

That's #10749 (via #11530).

> I am sorry to bother the huge amount of subscribed to this issue people, but does anyone know whats the best way to ping the maintainers?

This PR keeps getting bumped in various places repeatedly, so I've mentioned in a private maintainers channel that perhaps providing a response as to whether this will be reviewed/merged soon might be useful. But please note, the maintainers have no obligation to review or merge anything in a timely fashion or by some deadline; Ghostty is a volunteer project and the priorities of maintainers might not necessarily be that of users.

pluiedev:
Without getting into too many private details, I'll say that currently our review capacity for macOS-related PRs is quite low as most of our macOS maintainers are rather preoccupied with personal matters at present. There is nothing the rest of the team could do about it as we are, to reiterate, a fully-volunteer team, other than patiently waiting for them to return from their absence, and we urge everyone subscribed to this PR to do so as well. Once the macOS maintainers have time to review PRs again, I'm sure they will take a look at this as soon as they are able. In the meantime, thank you all for your patience 🙂

aeroith (approved):
Thank you. Would love to see this arriving to stable.

bugrauluyurt (approved):

jcollie:
> Congrats with the approves! 🎉 Can we assume it will be in 1.4 release?

The approvals received so far aren't from anyone with the authority to merge so they have little real effect. But 1.4 isn't due until this Fall so there's lot's of time left. Unfortunately most of the core Ghostty maintainers are busy with non-Ghostty things right now so code reviews are going to be very slow right now.

macos/Ghostty-Info.plist
```diff
@@ -2,6 +2,19 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>UTExportedTypeDeclarations</key>
```
bo2themax: This should be in the existing array down below. Looks like conflict resolution went wrong. I recommend manually editing the plist instead of using source code to edit.

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -136,10 +152,11 @@ class QuickTerminalController: BaseTerminalController {
             qtWindow.initialFrame = window.frame
         }
 
-        // Setup our content
+        // Setup our content with tab support and glass effect
         window.contentView = TerminalViewContainer {
-            TerminalView(ghostty: ghostty, viewModel: self, delegate: self)
+            QuickTerminalView(ghostty: self.ghostty, controller: self, tabManager: tabManager)
         }
+        terminalViewContainer?.ghosttyConfigDidChange(ghostty.config, preferredBackgroundColor: nil)
```
bo2themax: Why do we need to manually call this here, is it a workaround for some weird issues? It would be great we document it if so.

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabDragDelegate.swift
```diff
@@ -0,0 +1,283 @@
...
+            // Dropped on the terminal surface (not tab bar) - move to new window
+            tabManager.moveTabToNewWindow(tab, at: screenPoint)
+        }
+    }
+
+    /// Checks if the given screen location is in the tab bar area of the quick terminal window.
+    private func isInQuickTerminalTabBar(_ location: NSPoint, of window: NSWindow) -> Bool {
+        let windowFrame = window.frame
+        // Tab bar is at the top of the window, 24px tall
+        let tabBarHeight: CGFloat = 24
```
bo2themax: I don't like this but I have no better idea now. I think we should at least us the same constants in the tab view instead of using concrete value here

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabDragDelegate.swift
```diff
@@ -0,0 +1,283 @@
...
+        for window in windows {
+            if window.windowController is QuickTerminalController {
+                continue
+            }
+            guard window.windowController is TerminalController else {
+                continue
+            }
+            if window.frame.contains(location) {
+                return window
+            }
```
bo2themax: This could be simplified with `for where` or `first(where:)` clause

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTab.swift
```diff
@@ -0,0 +1,60 @@
...
+    @Published var title: String
+
+    /// User-defined title override. When set, this is displayed instead of the surface title.
+    @Published var titleOverride: String? {
+        didSet {
+            if let override = titleOverride {
+                title = override
+            } else {
+                // Restore surface title
+                title = currentSurfaceTitle ?? "Terminal"
```
bo2themax: I don't like the side effects here. Can we reuse the same one as BaseTerminalController. Considering quick terminal controller inherits those already.

It will also bring bells to quick terminal

macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
```diff
@@ -737,6 +744,23 @@ extension TerminalWindow {
...
+            title: "Move Tab to Quick Terminal",
+            action: #selector(TerminalController.moveTabToQuickTerminal(_:)),
+            keyEquivalent: ""
+        )
+        moveToQuickItem.identifier = Self.moveToQuickTerminalMenuItemIdentifier
+        moveToQuickItem.target = targetController
+        if #available(macOS 26.0, *) {
+            moveToQuickItem.setImageIfDesired(systemSymbolName: "arrow.down.to.line.compact")
+        } else {
+            moveToQuickItem.setImageIfDesired(systemSymbolName: "arrow.down.to.line")
```
bo2themax: Why are we using the arrow down here? Is it because your qt is at the bottom?

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabManager.swift
```diff
@@ -0,0 +1,435 @@
...
+    /// The index where a dragged tab will be dropped (for showing placeholder)
+    @Published var dropTargetIndex: Int?
+    /// The width of the tab being dragged (captured at drag start)
+    var draggedTabWidth: CGFloat?
+
+    /// Reference to the "quick" terminal Controller
+    weak var controller: QuickTerminalController?
+
+
+    var currentTabIndex: Int? {
```
bo2themax: These could be private settable

bo2themax (changes requested):
Thanks for the support, overall it looks really good to me! Beside those inline comments, I also want to mention few things:

1. The current tab appearance doesn't make a great visual difference between active and inactive tab when using Liquid Glass (Light theme, e.g. `Xcode Light`).
2. Creating, closing, and dragging/moving tabs into new windows don't support undo and redo, while regular windows do.
3. Dragging a tab and releasing it over the search overlay (and debug banner) in qt will create a new window, which it shouldn't.
4. I couldn't drag the last tab to the beginning when there are a lot of tabs; it should trigger the scroll view to scroll automatically.
5. The new tab button needs to be clicked twice when Quick Terminal is not the key window.
6. The close button is still visible after creating new tabs using the new tab button.
7. The last QuickTerminalTab is not deallocated after killing the process.
8. Two SwiftLint violations.

adrum:
> Thanks for the support, overall it looks really good to me! Beside those inline comments, I also want to mention few things:
> 
> 1. The current tab appearance doesn't make a great visual difference between active and inactive tab when using Liquid Glass (Light theme, e.g. `Xcode Light`).
> 2. Creating, closing, and dragging/moving tabs into new windows don't support undo and redo, while regular windows do.
> 3. Dragging a tab and releasing it over the search overlay (and debug banner) in qt will create a new window, which it shouldn't.
> 4. I couldn't drag the last tab to the beginning when there are a lot of tabs; it should trigger the scroll view to scroll automatically.
> 5. The new tab button needs to be clicked twice when Quick Terminal is not the key window.
> 6. The close button is still visible after creating new tabs using the new tab button.
> 7. The last QuickTerminalTab is not deallocated after killing the process.
> 8. Two SwiftLint violations.

1. I spent quite a bit of time on this today. I refactored the logic here to actually represent the surface background color instead. I originally tried using static colors based on system appearance, but it just felt off to me. I tried my best in making glass colors for active, inactive, and hovered tabs more legible, but it will vary based on the configs `background-opacity` value.
2. This is now implemented via inherited `BaseTerminalController.undoManager`. I tried match the existing behavior for normal windows/tabs as closely as possible. Please let me know if you find any discrepancies I did not find.
3. Fixed. Now dropping any QT tab on a QT surface does nothing.
4. I use a trackpad and used a second gesture to naturally scroll the tab bar while dragging my tab. I can see how this would be an issue for mouse users, but it should also be implemented as it's expected behavior for other scroll views on macOS. NSScrollView's built-in autoscroll only fires for events emitted to it, and the drag-destination pipeline doesn't surface NSEvents — so we'd have to synthesize them anyway. My implementation feels pretty solid. 
5. Fixed. This is actually a limitation of SwiftUI, so creating an NSHostingView wrapper since this behavior is available in AppKit. Our tabs have a similar wrapper (although a nice side effect and not the original purpose), since we had to use Native AppKit for better drag and drop detection, since SwiftUI was limiting on that front. Not sure if other areas of the application encounter this same limitation, but if so, we can make it a little more generic and portable down the road if needed.
6. Fixed
7. Fixed. A new surface is not allocated until the QT is triggered again. I also made it so clicking the dock icon will show a new standard window if no tabs are present in QT.
8. Fixed. One was from `main`

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabManager.swift
adrum: Some of these needed to be public, but I marked the ones private that could be.

macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
adrum: My QT is at the top. My brain was having the icon represent the QT (slides down), not the action of moving the tab. I found a different icon that represent the action of moving the tab to the quick terminal, which is possibly off screen at the top.

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTab.swift
adrum: I don't love making it static on the BaseTerminalController, but that's where I ended up. The main difference why is essentially, under a normal macOS window, the BaseTerminalController represents a single tab (window). Since QuickTerminalController manages a set of tabs, there was not a cleaner way for the end Tabs to get access to the function without passing the controller down to the tabs, which feels like the wrong design. If you have another approach, let me know!

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
adrum: While troubleshooting some code relating to switching themes and glass mode, at one point, I determined this was needed for the config to apply. I just removed this line and it does not seem necessary anymore.  Through the process of testing this, I was reliably experiencing this [bug report](https://github.com/ghostty-org/ghostty/pull/9857#issuecomment-4019619817), so I've add logic to automatically apply the theme to all tabs, not just the active tab.

macos/Ghostty-Info.plist
adrum: I think this one actually occurred outside of a git conflict since they were in two separate parts of the file. I'm guessing it merged in fine. Nonetheless, I removed my key/value manually, then opened it in the plist editor in Xcode added back the QuickTerminal tab type declaration as the second item in the array.

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTab.swift
bo2themax: In the future we can abstract this to a separate view model handling these updates, but I'm good with current changes.

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabBarView.swift
```diff
@@ -148,22 +148,27 @@ struct QuickTerminalTabBarView: View {
...
+                        .fill(newTabButtonBackgroundColor)
+                )
+                .onHover { isHovering in
+                    isHoveringNewTabButton = isHovering
+                }
+                .onTapGesture {
+                    tabManager.addNewTab()
+                }
+                .help("Create a new Tab")
+        }
```
bo2themax: You can just use `Button` here

```swift
Button {
    tabManager.addNewTab()
} label: {
    Image(systemName: "plus")
        .foregroundColor(Color(NSColor.secondaryLabelColor))
        .padding(.horizontal, Constants.addNewTabButtonHorizontalPadding)
        .frame(width: Constants.height, height: Constants.height)
        .background(
            Rectangle()
                .fill(newTabButtonBackgroundColor)
        )
}
.buttonStyle(.plain)
.onHover { isHovering in
    isHoveringNewTabButton = isHovering
}
.help("Create a new Tab")
```

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabBarView.swift
```diff
@@ -205,6 +210,27 @@ struct QuickTerminalTabBarView: View {
...
+
+    func updateNSView(_ nsView: HostingView, context: Context) {
+        nsView.rootView = content()
+    }
+
+    final class HostingView: NSHostingView<Content> {
+        override func acceptsFirstMouse(for event: NSEvent?) -> Bool { true }
+    }
+}
+
```
bo2themax: And delete this.
```swift
```

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabDragDelegate.swift
```diff
@@ -0,0 +1,370 @@
...
+
+    func draggingSession(
+        _ session: NSDraggingSession,
+        movedTo screenPoint: NSPoint
+    ) {
+        // Auto-scroll the tab strip if the cursor is near either edge. We
+        // hook in here (on the drag source) rather than on per-tab drop
+        // destinations so scrolling continues to work even when the cursor
+        // is over a gap, an unfocused area, or off the bar entirely.
+        updateAutoScroll(at: screenPoint)
```
bo2themax: As solid as it may be, I don't want this to be a possible maintenance burden for us.

I don't think the RC here is the difference between AppKit and SwiftUI, rather the DraggingTarget, currently you set tab as the target instead of the bar. I also tried your previous commits to confirm this, you can expect auto scroll if your dragging the leading/trailing tab that's not fully visible. I think it should come out of box if you make the bar the target or make the placeholder item also one if former involves to much refactor.

https://github.com/user-attachments/assets/3c6b1887-cb7f-423b-8b35-16df2abe0f67

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabManager.swift
```diff
@@ -0,0 +1,595 @@
...
+            } else {
+                let newIndex = min(index, tabs.count - 1)
+                selectTab(tabs[newIndex])
+            }
+        }
+
+        guard let undoActionName, let undoManager else { return }
+        undoManager.setActionName(undoActionName)
+        undoManager.registerUndo(withTarget: self, expiresAfter: undoExpiration) { target in
+            target.insertTab(tab, at: index, undoActionName: undoActionName)
```
bo2themax, line 342: After removing/closing the last tab and undoing, there will be two tabs, first tab doesn't have a surface.

Image

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabBarView.swift
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    var body: some View {
+        HStack(spacing: 0) {
+            renderTabBar()
+            renderAddNewTabButton()
+        }
+        .frame(height: Constants.height)
+        .background(tabBarBackgroundColor)
+        .sheet(item: $tabBeingRenamed) { tab in
```
bo2themax: `changeTabTitle` of the first responder(Change Tab Title in CommandPalette) doesn't work, and the title changed here doesn't sync with that sheetModal presented by the BaseTerminalController.

I'm thinking that you could override titleOverride in QTController by routing through currentTab's. Then you don't need this custom sheet anymore.

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -1007,6 +1011,37 @@ class BaseTerminalController: NSWindowController,
...
     }
 
+    // MARK: Surface Color Scheme
+
+    /// Update the surface tree's color scheme only when it actually changes.
+    ///
+    /// Calling ``ghostty_surface_set_color_scheme`` triggers
+    /// ``syncAppearance(_:)`` via notification,
+    /// so we avoid redundant calls.
+    func updateColorSchemeForSurfaceTree() {
```
bo2themax, line 1099: These changes (including the one done below) seems odd

bo2themax (changes requested):

macos/Sources/Features/Terminal/BaseTerminalController.swift
adrum: @bo2themax I had to move this function out of an extension and into the main class definition so the QuickTerminalController could override it and apply the call to all the tabs and their surfaces.

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabDragDelegate.swift
adrum: Great feedback. While implementing making the bar (background) the target, I noticed on opportunity to improve the feedback left in https://github.com/ghostty-org/ghostty/pull/9857#discussion_r3240387442, so I went ahead centralize that logic there. There probably could be more moved there, but I thought I would keep it simple for now.

macos/Sources/Features/Terminal/BaseTerminalController.swift
bo2themax: Ok, makes sense

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabBarView.swift
adrum: I had to create a private backing variable for titleOverride in the BaseTerminalController so the QuickTerminalController could override it and route the tab being renamed there. I was able to keep the behavior where right-clicking an inactive tab remained inactive during a rename.

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabManager.swift
adrum: I was able to fix this and simplify the closeTab logic for tabs at the same time.

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -91,9 +91,18 @@ class BaseTerminalController: NSWindowController,
 
     /// An override title for the tab/window set by the user via prompt_tab_title.
     /// When set, this takes precedence over the computed title from the terminal.
+    ///
+    /// Computed (with a private backing var) rather than stored so subclasses
+    /// like `QuickTerminalController` can override the get/set to route to a
+    /// per-tab title source.
     var titleOverride: String? {
-        didSet { applyTitleToWindow() }
```
bo2themax, line 106: This doesn't needs changes to allow override

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -44,6 +44,42 @@ class QuickTerminalController: BaseTerminalController {
...
+    /// Route the controller's title override to a tab — `tabBeingRenamed` if
+    /// the user is actively renaming via the prompt sheet, otherwise the
+    /// active tab. This lets right-click "Change Tab Title" on an inactive
+    /// tab rename it without changing the selection, while command-palette
+    /// and menu invocations still operate on the active tab.
+    override var titleOverride: String? {
+        get { renameTarget?.titleOverride }
+        set {
+            renameTarget?.titleOverride = newValue
+            applyTitleToWindow()
```
bo2themax: You can remove this once the base is reverted

bo2themax (changes requested):

macos/Sources/Features/QuickTerminal/Tab/QuickTerminalTabBarView.swift
bo2themax: I tried reverting it back to didSet without the backing variable tho, I tried command palette and menu, both worked fine

adrum: I reverted the backing var change and it seems to be working.

bo2themax (approved):

bo2themax (changes requested):
ci failed for iOS, needs some fixing

adrum:
> ci failed for iOS, needs some fixing

I was already on it! I removed the iOS target from the rest of the source files related to QuickTerminal tabs.

iwex:
@bo2themax hello! please take a look ❤️

bo2themax (approved):

iwex:
Thank you. Is it possible to move it to 1.3.2 milestone?

00-kat:
Patch releases are for bug fixes, this is hardly one; even if it were I'd argue it's too large for that.

iwex:
@00-kat I completely agree with you. I only wrote that because I’d really like to see this feature released as soon as possible. It’s very useful for me and many other users who want to switch from iTerm.

00-kat:
I will note that you (and others) are free to run this branch, and once it's merged you can use the even easier [`auto-update-channel = tip`] to run the latest commit of main.

[`auto-update-channel = tip`]: https://ghostty.org/docs/config/reference#auto-update-channel

bo2themax (approved):

wammy67 (approved):

memdealer:
Greetings, been running the branch a couple weeks, works great. I wonder if projected release is drafted with it :)

usatenko:
> Greetings, been running the branch a couple weeks, works great. I wonder if projected release is drafted with it :)

could you post a screenshot how it looks like?

Zhengqbbb:
I built the app from this branch locally for the past two weeks, and it's been working great 👍 Hoping your hard work can be merged into the main branch soon :)
CleanShot 2026-07-10 at 17 02 13@2x

ashtonmeuser:
Any update on when this might be merged? It's been about 18 months since the first implementation in #5370. Would be great to see all this hard work released to the masses!

Bielik20:
I know this is probably the worst place to ask this question, but what is happening with 1.3.2 release? It's almost 2 month overdue. I think everyone hoped that after the release this gets merged and we can run it like mentioned here: https://github.com/ghostty-org/ghostty/pull/9857#issuecomment-4519036550 using [auto-update-channel = tip](https://ghostty.org/docs/config/reference#auto-update-channel)

mmatviyiv:
Honestly, I don't understand why it takes so long. This isn't a matter of complicated implementation that people are asking the core team to handle independently - it's already being done, tested, and verified. Many people have been waiting for this release for over half a year; all that's needed is to merge and release it, yet it still hasn't happened

mitchellh:
> I know this is probably the worst place to ask this question, but what is happening with 1.3.2 release? It's almost 2 month overdue.

The due date means nothing. It was canceled in favor of a 1.4 this fall. I'll remove the milestone eventually. For due dates, we always just set them to arbitrary dates.

> Honestly, I don't understand why it takes so long.

Because this is a volunteer project and pressing the merge button is a transfer of ownership and we haven't felt comfortable enough yet to do it OR had the time OR decided we cared to look at this OR any of the above. Luckily, its open source and if you want to maintain it, please fork and merge it. You're just as capable and obligated as we are here.

I know this is a very highly requested feature, but as you said yourself: its here, if you want it, apply it to your own patch set and build!

We'll get to it when we get to it.

adrum:
I just pulled in main as I noticed there was a memory leak I was experiencing (unrelated to this branch) that was resolved. I also noticed clicking the x button on the tab itself did not properly warn when any of the surfaces had a running process, so I pushed a fix for that.

adrum:
Here are two features I would like to add in follow up PRs:

1. Command Palette: Move Tab to QuickTerminal/New Window (inverse actions)
2. Command Palette: Focus QuickTerminal tabs/surfaces

I have been running the second one in a branch for awhile now, so that one is essentially ready to go.

abjugard:
For anyone else religiously building this branch locally whenever @adrum pushes changes, if you wanna save some time during compilation build with the flag `-Dxcframework-target=native` (e.g. `zig build -Dxcframework-target=native -Doptimize=ReleaseFast`) and it won't make any other architectures than what you're building on.

PR #9864, opened 2025-12-10
nix: highway is not available in Nix so compile our own

jcollie:
Fixes regression from #9642


mitchellh:
I changed the default here, we should be good now? https://github.com/ghostty-org/ghostty/actions/runs/20113462715 I think the default should be no-sys because I haven't found highway to be packaged in most places.

jcollie:
With the default changed in 7642b8bec4294ecdaf9184fd69ed761a7e2aa422 this is no longer necessary.

PR #9875, opened 2025-12-11
bash: use a shell command for shell integration

jparise:
Prior to #7044, on macOS, our shell-integrated command line would be executed under `exec -l`, which causes bash to be started as a login shell. This matches the macOS platform norms.

The change to direct command execution meant that we'd skip that path, and bash would start as a normal interactive (non-login) shell on macOS. We fixed this in #7253 by adding `--login` to the `bash` direct command on macOS.

This avoided some of the overhead of starting an extra process just to get a login shell, but it unfortunately doesn't quite match the bash environment we get when shell integration isn't enabled (namely, `$0` doesn't get the login-shell-identifying `-` prefix).

Instead, this change implements the approach proposed in #7254, which switches the bash shell integration path to use a `.shell` command, giving us the same execution environment as the non-shell-integrated command.

### DIFF
--- a/src/termio/shell_integration.zig
+++ b/src/termio/shell_integration.zig
@@ -259,7 +259,7 @@ fn setupBash(
     resource_dir: []const u8,
     env: *EnvMap,
 ) !?config.Command {
-    var args: std.ArrayList([:0]const u8) = try .initCapacity(alloc, 3);
+    var args: std.ArrayList([:0]const u8) = try .initCapacity(alloc, 2);
     defer args.deinit(alloc);
 
     // Iterator that yields each argument in the original command line.
@@ -272,11 +272,6 @@ fn setupBash(
         try args.append(alloc, try alloc.dupeZ(u8, exe));
     } else return null;
     try args.append(alloc, "--posix");
-
-    // On macOS, we request a login shell to match that platform's norms.
-    if (comptime builtin.target.os.tag.isDarwin()) {
-        try args.append(alloc, "--login");
-    }
 
     // Stores the list of intercepted command line flags that will be passed
     // to our shell integration script: --norc --noprofile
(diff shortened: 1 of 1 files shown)


src/termio/shell_integration.zig
```diff
@@ -259,7 +259,7 @@ fn setupBash(
     resource_dir: []const u8,
     env: *EnvMap,
 ) !?config.Command {
-    var args: std.ArrayList([:0]const u8) = try .initCapacity(alloc, 3);
+    var args: std.ArrayList([:0]const u8) = try .initCapacity(alloc, 2);
```
jparise, line 262: I may revisit this args-building code separately to build up the command string directly (rather than accumulating arguments in an array and then joining them with spaces). Maybe it could be generalized as a `ShellCommandWriter`.

src/termio/shell_integration.zig
```diff
@@ -273,11 +273,6 @@ fn setupBash(
     } else return null;
     try args.append(alloc, "--posix");
 
-    // On macOS, we request a login shell to match that platform's norms.
-    if (comptime builtin.target.os.tag.isDarwin()) {
-        try args.append(alloc, "--login");
```
jparise, line 278: An alternate solution to just the `$0` issue would be to prefix `args[0]` with a `-` here, to complement the `--login`, but I worry that goes further down the "patch the problems as they come up" path.

mitchellh (approved):
I don't see anything that stands out here as bad. Going to trust you and accept. Thank you.

src/termio/shell_integration.zig
mitchellh: Yeah might be worth it.

jparise: #9881

PR #9876, opened 2025-12-11
feat(config): allow fullscreen config to specify non-native mode directly

### LINKED ISSUE #8388: macOS: Allow `fullscreen` to toggle non-native fullscreen
Discussion from #3214

Modify the fullscreen configuration to be an enum to take a `non-native` value. If this is set, new windows will open in non-native fullscreen matching the same non-native settings as `macos-non-native-fullscreen`. The values `true` and `false` continue to do what they do today. 

On non-macOS platforms, `non-native` should behave the same as `true`, since a non-false value implies the desire for _some fullscreen_.

benodiwal:
Closes #8388


pluiedev (approved):
Ack GTK changes. Not sure if it's really necessary to have the two extra options on macOS, but I'll leave it for someone else to judge

benodiwal:
@mitchellh can u look into this once.

mitchellh:
!vouch

mitchellh (approved):
Works great, approved pending CI. Thanks.

PR #9879, opened 2025-12-12
macOS: Change Tab Title

mitchellh:
This adds the ability to change a _tab_ title. The previous functionality was tied to a specific _surface_. A tab title will stick to the current tab regardless of active splits and so on.

This follows the nomenclature that macOS terminal app does which is "title vs terminal title" (although we explicitly use "tab" in various places, I may remove that in the future).

**This is macOS only. GTK is tracked here: #9880**. I did macOS only because thats the machine I'm on. It'll be trivial to add this to GTK, too.

## Demo

https://github.com/user-attachments/assets/d9446785-d919-4212-8553-db50c56c8c2f

(The option is also in the main menu, the context menu, and the command palette)

## AI Disclosure

This PR was done fully with Amp, I didn't write a single line of code at the time of writing this PR description. I reviewed everything though and fully understand it all. Its a mimic more or less of the prompt surface title work (although we did unify some stuff like the apprt action).


jcollie (approved):
GTK changes look legit.

mitchellh:
I'll add on here that I think there is some UX exploration to be done on this whole feature. It's unclear if we want to actually expose `Change Terminal Title` as much as we do if people vastly prefer the `Change Tab Title` almost every time. We can clean that up and remove menu items later.

mcandre:
Where are the docs for this?

Unclear how to wire up _shells_ to ghostty tab/window titles.

jcollie:
@mcandre Don't use PRs as technical support. Open a new discussion with your question.

PR #9883, opened 2025-12-12
font/coretext: Use positions to correct glyph placement

jacobsandlund:
This PR uses positions to get correct glyph placement, which affects a small number of shaped runs but can even re-order glyphs from how they're appearing on `main`. This came from an investigation in https://github.com/jquast/wcwidth/issues/155#issuecomment-3630555803 that involved shaping text from the [Universal Declaration of Human Rights](https://www.un.org/en/about-us/universal-declaration-of-human-rights) in 500+ languages, taken from [ucs-detect](https://github.com/jquast/ucs-detect) (but actually run with [ttylang](https://github.com/jacobsandlund/ttylang) for more consistent results).

The full result of running this with the commented out debug lines enabled is here, showing 1704 differences (a small number compared to the number of cells being rendered): https://gist.github.com/jacobsandlund/9cb2c603308085fa03e758640583541e

Some positions are only a small offset that is applied to many letters from a script (e.g. Syriac letters are getting offset by some small amount).

There are more egregious cases, though. Consider Tai Tham vowels, that [according to the Unicode core spec](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-16/#G46437), are rendered visually ahead of the consonant they follow in the text. The Unicode spec oddly mentions this under the "New Tai Lue" section, contrasting it with other Brahmi-derived scripts such as Tai Tham:

> This practice differs from the usual pattern for Brahmi-derived scripts, in which all dependent vowels are stored in logical order after their associated consonants, even when they are displayed to the left of those consonants.

For example `"\u{1A2F}\u{1A70}"` renders like this on `main`:

CleanShot 2025-12-12 at 09 25 16@2x

And it now renders like this on the PR:

CleanShot 2025-12-12 at 09 25 38@2x

This is the correct rendering as U+1A70 is "TAI THAM VOWEL SIGN OO" and needs to render ahead of the U+1A2F "TAI THAM LETTER DA"

Note, this PR just fixes CoreText. I'll work on Harfbuzz right after this, but I wanted to get this out first.

## vtebench

vtebench does show some potential impact on dense cells:

output

dense_cells

This is the first time I've run `vtebench`, so to confirm it's consistent I ran it again:

output-2

dense_cells

### AI disclaimer
I had been using Amp when I made this change, but looks like the thread just fell off of recent threads (about a week ago). I think Amp was the one to add the getPositionsPtr line and add `position` to the for loop. I manually restructured the code around a separate `run_offset` and `cell_offset`, and added the tests.


src/font/shaper/coretext.zig
```diff
@@ -377,11 +377,21 @@ pub const Shaper = struct {
...
+
+        // This keeps track of the current offsets within a cell.
         var cell_offset: struct {
             cluster: u32 = 0,
             x: f64 = 0,
             y: f64 = 0,
+
+            // For debugging positions, turn this on:
+            //start_index: usize = 0,
+            //end_index: usize = 0,
```
jacobsandlund: I can remove this and the below debug code if desired before merging, since it's several commented out lines

src/font/shaper/coretext.zig
```diff
@@ -1333,6 +1403,9 @@ test "shape Devanagari string" {
     var t = try terminal.Terminal.init(alloc, .{ .cols = 30, .rows = 3 });
     defer t.deinit(alloc);
 
+    // Disable grapheme clustering
+    t.modes.set(.grapheme_cluster, false);
```
jacobsandlund, line 1435: This test depended on each glyph being a separate "x" / separate cell, so Grapheme clustering needs to be disabled, so I made that explicit.

src/terminal/Terminal.zig
```diff
@@ -8893,7 +8893,7 @@ test "Terminal: insertBlanks shift graphemes" {
     var t = try init(alloc, .{ .rows = 5, .cols = 5 });
     defer t.deinit(alloc);
 
-    // Disable grapheme clustering
+    // Enable grapheme clustering
     t.modes.set(.grapheme_cluster, true);
```
jacobsandlund, line 9062: Just happened to catch that this comment was off.

danielwe:
Curious if this has any bearing on #8054?

jacobsandlund:
> Curious if this has any bearing on https://github.com/ghostty-org/ghostty/discussions/8054?

Quite possibly for Harfbuzz, which right now is only looking at the `GlyphPosition` `{x,y}_advance`, and ignoring `{x,y}_offset`. I'll look at Harfbuzz starting Wednesday.

mitchellh (changes requested):
This looks alright to me. It's always hard to context load on this stuff which is why we have the unit tests... given the rest pass without issue, I'm inclined to accept this and add regression tests as necessary. 

You're also usually pretty thorough. 

Only change I'd like to request is to either remove the debug stuff or move it to a function that we don't call (I'd prefer the latter so we can easily turn it back on).

jacobsandlund:
@mitchellh I pulled out the debugging into a function here: https://github.com/ghostty-org/ghostty/pull/9883/commits/139a23a0a2f4879c884dfab075c45ca93eb5ae64

Note I still have some commented out logic for setting the `start_index` and `end_index`. I _could_ remove that and then the log could just include the current codepoint, but that makes the log much less useful, so I left it.

mitchellh (approved):

PR #9887, opened 2025-12-12
fix(docs): `window-decoration` is now `none` instead of `false`

michaelhazan:
Heya,
super random but i saw this in the documentation and felt like it should be fixed 😅


mitchellh (approved):
Thanks!

jcollie (approved):

PR #9890, opened 2025-12-13
GTK: Readonly overlay

### LINKED ISSUE #9889: GTK: Add read-only surface GUI indication
Following on #9130 

We should add some GUI overlay that a surface is read-only. Here is what it looks like on macOS:

https://github.com/user-attachments/assets/f0086361-0ae7-4bd4-82a5-47685e5a7542

rhodes-b:
Closes: #9889

adds support for a readonly overlay on GTK.

I wouldn't say this PR is quite ready but I'm opening to get feedback if this is the right direction and possibly some tips.

1. ~The read only overlay overlaps with the search overlay right now I would assume this is due to the alignment rules?~ I figured out I can solve it with a box is that correct? Adding the box messes up the mouse reporting of the gl area looking into that
2. ~The label should wrap the text instead of the manual newlines~ This was easily solved with some wrap properties and max-width-chars
3. I disabled the arrow on the menu item should that be added back as context that you can click on the widget?

https://github.com/user-attachments/assets/84b268ba-da12-4403-b810-2774ae3c1339

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1009,9 +1009,8 @@ pub const Surface = extern struct {
         const priv = self.private();
         priv.read_only = !priv.read_only;
         self.as(gobject.Object).notifyByPspec(
-            properties.@"read-only".impl.param_spec
+            properties.@"read-only".impl.param_spec,
         );
-        log.info("read only state {}", .{priv.read_only});
         return true;
     }
 
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -547,6 +547,9 @@ pub const Surface = extern struct {
         url_left: *gtk.Label,
         url_right: *gtk.Label,
 
+        // The read only overlay
+        read_only: *gtk.MenuButton,
+
         /// The resize overlay
         resize_overlay: *ResizeOverlay,
 
@@ -978,6 +981,19 @@ pub const Surface = extern struct {
             defer alloc.free(body);
 
             self.sendDesktopNotification(title, body);
+        }
+
+        return true;
+    }
+
+    pub fn toggleReadOnly(self: *Self) bool {
+        const priv = self.private();
+
+        const readOnlyWidget = priv.read_only.as(gtk.Widget);
+        if (readOnlyWidget.getVisible() == 1) {
+            readOnlyWidget.setVisible(@intFromBool(false));
+        } else {
+            readOnlyWidget.setVisible(@intFromBool(true));
         }
 
         return true;
@@ -3256,6 +3272,7 @@ pub const Surface = extern struct {
             class.bindTemplateChildPrivate("gl_area", .{});
             class.bindTemplateChildPrivate("url_left", .{});
             class.bindTemplateChildPrivate("url_right", .{});
+            class.bindTemplateChildPrivate("read_only", .{});
             class.bindTemplateChildPrivate("child_exited_overlay", .{});
             class.bindTemplateChildPrivate("context_menu", .{});
             class.bindTemplateChildPrivate("error_page", .{});
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -305,6 +305,24 @@ pub const Surface = extern struct {
             );
         };
 
+        pub const @"read-only" = struct {
+            pub const name = "read-only";
+            const impl = gobject.ext.defineProperty(
+                name,
+                Self,
+                bool,
+                .{
+                    .default = false,
+                    .accessor = gobject.ext.privateFieldAccessor(
+                        Self,
+                        Private,
+                        &Private.offset,
+                        "read_only",
+                    ),
+                },
+            );
+        };
+
         pub const hadjustment = struct {
             pub const name = "hadjustment";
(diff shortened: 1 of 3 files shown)


src/apprt/gtk/class/surface.zig
```diff
@@ -983,6 +986,19 @@ pub const Surface = extern struct {
...
 
+    pub fn toggleReadOnly(self: *Self) bool {
+        const priv = self.private();
+
+        const readOnlyWidget = priv.read_only.as(gtk.Widget);
+        if (readOnlyWidget.getVisible() == 1) {
+            readOnlyWidget.setVisible(@intFromBool(false));
+        } else {
+            readOnlyWidget.setVisible(@intFromBool(true));
+        }
```
pluiedev: IMO the read-only state should be a property on the Surface, and then the read-only widget should be bound to the read-only property

src/apprt/gtk/ui/1.2/surface.blp
```diff
@@ -148,11 +148,36 @@ Overlay terminal_page {
...
+      ]
+
+      visible: false;
+      halign: end;
+      valign: start;
+      margin-end: 15;
+      margin-top: 15;
+      always-show-arrow: false;
+      direction: none;
+      label: "Read-only";
```
pluiedev: Need to be localized
```suggestion
      label: _("Read-only");
```

pluiedev (changes requested):
Design-wise, the read-only popup should IMO also have the "destructive" style to make it more visually apparent

rhodes-b:
> Design-wise, the read-only popup should IMO also have the "destructive" style to make it more visually apparent

If im understanding correctly I would need to use a GhosttyDialog to achieve that?
If so it requires its own blueprint and class file?

src/apprt/gtk/class/surface.zig
```diff
@@ -988,14 +1007,11 @@ pub const Surface = extern struct {
...
-            readOnlyWidget.setVisible(@intFromBool(false));
-        } else {
-            readOnlyWidget.setVisible(@intFromBool(true));
-        }
-
+        priv.read_only = !priv.read_only;
+        self.as(gobject.Object).notifyByPspec(
+            properties.@"read-only".impl.param_spec
+        );
+        log.info("read only state {}", .{priv.read_only});
```
rhodes-b: will fix with the destructive dialog

jcollie:
> > Design-wise, the read-only popup should IMO also have the "destructive" style to make it more visually apparent
> 
> If im understanding correctly I would need to use a GhosttyDialog to achieve that? If so it requires its own blueprint and class file?

No, I believe that all that you need to do is add "destructive" as a style to whatever widgets and GTK will style it appropriately.

rhodes-b:
> > > Design-wise, the read-only popup should IMO also have the "destructive" style to make it more visually apparent
> > 
> > 
> > If im understanding correctly I would need to use a GhosttyDialog to achieve that? If so it requires its own blueprint and class file?
> 
> No, I believe that all that you need to do is add "destructive" as a style to whatever widgets and GTK will style it appropriately.

I'm not sure I see any difference when adding "destructive" to the styles, I believe the dialog would also have the benefit of having a button to exit read only mode like the original macos PR

mitchellh:
At a base level this is looking good. I do want to consider a bit of polish but fundamentally this is doing the right thing. Thank you.

rhodes-b:
I've been looking into this without much success my original plan to fix the overlapping widgets problem was to use a box but even with event controllers that doesn't seem like a viable solution as the box just keeps stealing the events and some cant be passed through.

I found that overlays actually let you place its overlay children with a signal handler https://docs.gtk.org/gtk4/signal.Overlay.get-child-position.html but this also runs into problems trying to determine position relative to siblings that are marked as visible but not actually placed yet along with ignoring the full-screen widgets like the unfocused split revelers. It also seemed bad to try to manually reposition siblings by re-allocating them https://docs.gtk.org/gtk4/method.Widget.allocate.html

A simple workaround for now would be to give the read-only widget padding for when the search overlay would be visible that doesn't look the best when its not visible though

overlapping is currently a problem today with the search overlay and url overlay so I think it should definitely get addressed at some point

Given the above information I think I'll plan on closing this PR unless that simple workaround above seems fine for now since I can't seem to find a nice solution looking at the GTK docs / searching forums online 

note I did ask AI for potential solutions I didn't find any of its information useful as it was either overly complex or didn't seem to make much sense (the solutions I explored above were not from AI suggestions)

jcollie:
Whoops! @rhodes-b sorry about that but I forgot that you had been working on this and I went and got my own version of the read-only overlay merged. If you're interested in porting whatever UI ideas you had on top of what got merged I'll try and get that reviewed quickly!

rhodes-b:
> Whoops! @rhodes-b sorry about that but I forgot that you had been working on this and I went and got my own version of the read-only overlay merged. If you're interested in porting whatever UI ideas you had on top of what got merged I'll try and get that reviewed quickly!

I didn't come to anything different than yours besides not using a revealer. So I don't think there is anything extra to do in the short term. I think it would be nice to eventually solve the overlapping problem (it already existed before this with the search overlay and link popups if you moved the search one)

Never quite fixed the little bugs using that signal I mentioned above to try to solve the overlapping problem with the search overlay

PR #9891, opened 2025-12-13
fix: explicitly allow preservation for TERMINFO in shell-integration

definfo:
Due to security issues, `sudo` implementations may not preserve environment variables unless appended with `--preserve-env=list`.

Problem context: https://github.com/ghostty-org/ghostty/discussions/9861

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -103,7 +103,7 @@ if [[ "$GHOSTTY_SHELL_FEATURES" == *"sudo"* && -n "$TERMINFO" ]]; then
     if [[ "$sudo_has_sudoedit_flags" == "yes" ]]; then
       builtin command sudo "$@";
     else
-      builtin command sudo TERMINFO="$TERMINFO" "$@";
+      builtin command sudo --preserve-env=TERMINFO "$@";
     fi
   }
 fi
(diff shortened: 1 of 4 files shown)


jparise (approved):
This looks good. Thanks for the improvement!

jparise:
I looks like `--preserve-env` gained the ability to list specific environment variable names in version 1.8.21, [which is from 2017](https://www.sudo.ws/releases/legacy/):

> The --preserve-env command line option has been extended to accept a comma-separated list of environment variables to preserve.

It looks like some other implementations might not support it yet though: https://github.com/microsoft/sudo/issues/21

I think that's acceptable in this case because our `sudo` integration is an optional shell feature.

definfo:
> I looks like `--preserve-env` gained the ability to list specific environment variable names in version 1.8.21, [which is from 2017](https://www.sudo.ws/releases/legacy/):
> 
> > The --preserve-env command line option has been extended to accept a comma-separated list of environment variables to preserve.
> 
> It looks like some other implementations might not support it yet though: [microsoft/sudo#21](https://github.com/microsoft/sudo/issues/21)
> 
> I think that's acceptable in this case because our `sudo` integration is an optional shell feature.

Hopefully they will take a look at the issue, otherwise this may rely on some platform-dependent logic :worried:

PR #9899, opened 2025-12-13
macos: implement goto_window:next/previous

mitchellh:


mitchellh:
This is already done in GTK.

PR #9904, opened 2025-12-14
gtk: Change tab title

dmatos2012:
Closes #9880. 

https://github.com/user-attachments/assets/7110fae1-a7ca-42b6-8956-833b8bdd5d98

> [!NOTE]
>
> Used AI to get to explain to me some of the APIs and functionality between `windows` and `tabs` and their connections. But the code itself, I implemented everything, just used it for help with explaining some things

## If you feel like reading more:
- I feel there is a lot of duplication now with the `prompt surface title` functionality. For the dialog creation, i basically copied-pasted the `surface_title_dialog`, and some of the `titleOverride` and the `TabDialogSet` functionality is identical to that of the `prompt_surface_title`

- I would think about abstracting it out, but first I think Id be nice to think what the maintainers think about this approach before doing anything else.


mitchellh (changes requested):
Thanks, parts of this look great and others not so much. I think we should be able to rename `SurfaceTitleDialog` to `TitleDialog` and reuse that with differing copy depending on the mode, rather than introducing a whole new dialog.

The actual core tab stuff looks fine though.

dmatos2012:
Thx for the review!. I  refactored into a single `TitleDialog` following your feedback.

BenGWeeks:
+1

Would be great to get this in :-)

src/apprt/gtk/class/tab.zig
```diff
@@ -14,6 +14,8 @@ const Config = @import("config.zig").Config;
 const Application = @import("application.zig").Application;
 const SplitTree = @import("split_tree.zig").SplitTree;
 const Surface = @import("surface.zig").Surface;
+const TitleDialog = @import("title_dialog.zig")
+    .TitleDialog;
```
jcollie: Nit: this should be one line.

src/apprt/gtk/class/surface.zig
```diff
@@ -28,7 +28,8 @@ const ResizeOverlay = @import("resize_overlay.zig").ResizeOverlay;
 const SearchOverlay = @import("search_overlay.zig").SearchOverlay;
 const ChildExited = @import("surface_child_exited.zig").SurfaceChildExited;
 const ClipboardConfirmationDialog = @import("clipboard_confirmation_dialog.zig").ClipboardConfirmationDialog;
-const TitleDialog = @import("surface_title_dialog.zig").SurfaceTitleDialog;
+const TitleDialog = @import("title_dialog.zig")
+    .TitleDialog;
```
jcollie: Nit: this should be one line.

jcollie (changes requested):
A couple small nits, and then we'll see what CI says. I'd also like to see a context menu on the tab header itself with an option to change the tab title - that seems like a natural place that people are going to look for such functionality.

dmatos2012:
@jcollie yeah good point, added the context menu now.  
Only thing is that it looks like the `focus` on entry isn't correct like on the `main menu` and `surface menu`, because once the `TitleDialog` menu opens, you have to click the text box to rename it. 🤔

src/apprt/gtk/ui/1.5/window.blp
```diff
@@ -307,3 +314,10 @@ menu main_menu {
     }
   }
 }
+
+menu tab_context_menu {
+  item {
+    label: _("Change Title...");
```
jcollie: This should use the Unicode ellipsis "…", and should be "Change Tab Title…" for consistency.

src/apprt/gtk/ui/1.2/surface.blp
```diff
@@ -277,6 +277,11 @@ menu context_menu_model {
     submenu {
       label: _("Tab");
 
+      item {
+        label: _("Change Tab Title...");
```
jcollie: Use the Unicode ellipsis "…"

src/apprt/gtk/class/tab.zig
```diff
@@ -372,15 +428,17 @@ pub const Tab = extern struct {
         _: *Self,
         config_: ?*Config,
         terminal_: ?[*:0]const u8,
-        override_: ?[*:0]const u8,
+        surface_override_: ?[*:0]const u8,
+        tab_override: ?[*:0]const u8,
```
jcollie: For consistency with `surface_override_` let's add a trailing `_`.

jcollie (changes requested):
Ok, we're very very close! Just a few small nitpicks. Thanks for hanging in there!

jcollie:
Oof, there's also a typo that needs to be fixed!

src/apprt/gtk/ui/1.5/window.blp
dmatos2012: I purposely did `Change Title` since its on the Tab Context itself, and saw that  in `macOS` it is named `Change Title` as well on the tab context menu. But will change it back if you prefer that, no problem :)

dmatos2012:
Thanks for the feedback again. I have no rush in this, just want correctness and consistency with the code base, so all those nits are perfectly fine by me.

jcollie (approved):
LGTM

jcollie:
@dmatos2012 can you rebase and resolve the merge conflicts?

dmatos2012:
@jcollie done.

jcollie:
Looks like you also need to run `zig build update-translations`.

akulbe:
Thank you!

Issue #9718, opened 2025-11-26
Top row doesn't update when scrolling in less after searching


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9703

<div type='discussions-op-text'>

<sup>Originally posted by **injust** November 25, 2025</sup>
### Issue Description

When using less 685 to browse Ghostty's config (`ghostty +show-config --default --docs | less`), the first line does not update when less is first opened or when scrolling down.

The line will re-render if I click on it or scroll up, but it will become "stuck" again when I resume scrolling down.

I can reproduce this reliably in Ghostty, but not in Terminal.app. [This asciicast](https://github.com/user-attachments/files/23755012/asciicast.json) also reproduces in Ghostty, but not in Terminal.app.

### Expected Behavior

The first line should update normally.

### Actual Behavior

The first line doesn't appear when less is first opened, and doesn't update when I scroll down.

https://github.com/user-attachments/assets/fa8d8bd1-1bc3-4aeb-b7d4-d31ce530357c

### Reproduction Steps

1. `ghostty +show-config --default --docs | less`
2. ~~Press `/` and type something to search (e.g. `please`), then press <kbd>enter</kbd>~~
3. Press down arrow to scroll down a few lines
4. Press up arrow to scroll up

### Ghostty Version

```text
Ghostty 1.3.0-main+9511b237f

Version
  - version: 1.3.0-main+9511b237f
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS Sequoia 15.7.2

### Minimal Ghostty Configuration

```ini
# empty
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

mitchellh:
This is a regression from our render state work. I think there is a VT sequence somewhere that is missing marking one row as dirty.

qwerasd205:
[...]

PR #9906, opened 2025-12-14
macos: populate the sparkle:channel element

jparise:
This makes the update channel name available alongside the version, data, etc., which we can use in our update view (on the Released line).

CleanShot 2025-12-14 at 09 48 43@2x


mitchellh:
I think this actually broke our tip releases. I'm never able to find an update anymore. I'm going to revert this so we can look into it.

luisdavim:
> I think this actually broke our tip releases. I'm never able to find an update anymore. I'm going to revert this so we can look into it.
> 

I think this was broken before, see https://github.com/ghostty-org/ghostty/issues/9718#issuecomment-3646105763

Ghostty tip was reporting its version as 1.3.0 and I had to manually download and install the new version. There was a similar report for Linux as well.

mitchellh:
I think that's an unrelated issue: #9914. macOS tip updates don't find updates via version numbers so thats unrelated.

mitchellh:
Yeah this fixed update availability on macOS.

jparise:
Is there more information on how it broke? I generated the `appcast.xml` file locally and thought it all looked/worked as expected.

mitchellh:
I haven't use channels in awhile but I think you need to opt-in to what channel you want in the Sparkle API and it filters by them. They're a first class thing. So we'd have to change our channel config in the updater config (Sparkle API).

jparise:
> I haven't use channels in awhile but I think you need to opt-in to what channel you want in the Sparkle API and it filters by them. They're a first class thing. So we'd have to change our channel config in the updater config (Sparkle API).

Ah, I see the missing piece now (from https://sparkle-project.org/documentation/publishing/#publishing-an-update):

```swift
func allowedChannels(for updater: SPUUpdater) -> Set<String> {
    return Set(["beta"])
}
```

... which further appears to be a recommended alternative to multiple feeds, which is a bigger thing to consider.

PR #9919, opened 2025-12-15
Dim a window when it loses focus

lgouger:
A PR based on the discussion #9792 

The feature adds the option `unfocused-window-opacity` and `unfocused-window-fill` to dim the contents of a Ghostty window when it loses focus.  If `unfocused-window-opacity` is not present in the config, the current behavior continues and a window losing focus does not dim.   I chose to implement this feature because I found that I was starting to type, thinking I had focus, but then discovering that in many cases I did not yet have focus.  This dimming when focus is lost is available in iTerm, and I've enjoyed using it there.

These changes were made with the help of Claude Code. I've reviewed and tested the code both on the mac and on an Omarchy Linux box.    I've been using the MacOS version for several weeks, merging in the latest from the main ghostty-org branch each day to stay up to date.  

The initial plan generated by Claude Code is in the `notes` directory.  It's initial implementation didn't support the possibility of a fill color and make the default value for the `unfocused-window-opacity` 0, as a sentinel value.  I manually added support for the `unfocused-window-fill` color (falling back to the background color if not defined) and made the default config value 1.0.  

In the linux domain, I again worked with Claude Code on an implementation.  Claude Code  made code changes and responded to my feedback as I tested the resulting builds.  

I do not intend to merge the AI generated markdown in the `notes` directory.   It's there to aid any review.
Prior to a merge, if eligible, ask and I'll remove the notes directory and its contents.


mitchellh:
Let's kick this to a discussion. I have thoughts.

PR #9921, opened 2025-12-16
Fix cmd-click opening of relative/local paths

kaplanelad:
This PR fixes an issue #9563 where relative file paths were not being resolved against the terminal’s current working directory before opening.

#### Verification
Tested with directories containing:
```
/tmp/test/test
❯ du -h .
  0B    ./spaces-end
  0B    ./with dot.
  0B    ./space middle
8.0K    .
```

Parent directory resolution also works as expected:
```
/tmp/test/test
❯ du -h ..
  0B    ../test/spaces-end
  0B    ../test/with dot.
  0B    ../test/space middle
8.0K    ../test
 16K    ..
 ```
 
@mitchellh  
In your original description you mentioned that “Links should work for all situations as they do in iTerm2.”  
I noticed that, for example, when running `ls`, the paths are not clickable, while they are clickable in iTerm2.
If you think this case should also be handled, I can open a separate PR for it once this one is accepted.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2034,6 +2034,23 @@ pub fn pwd(
     return try alloc.dupe(u8, terminal_pwd);
 }
 
+/// Resolves a relative file path to an absolute path using the terminal's pwd.
+fn resolvePathForOpening(
+    self: *Surface,
+    path: []const u8,
+) Allocator.Error!?[]const u8 {
+    if (!std.fs.path.isAbsolute(path)) {
+        const terminal_pwd = self.io.terminal.getPwd() orelse {
+            return null;
+        };
+
+        const resolved = try std.fs.path.resolve(self.alloc, &.{ terminal_pwd, path });
+        return resolved;
+    }
+
+    return null;
+}
+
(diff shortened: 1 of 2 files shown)


mitchellh:
I notice this explicitly captures the trailing space. Does this underline the blank space at the end? If so, we should perhaps just add whitespace trimming as a post-step to matching.

kaplanelad:
Hey @mitchellh,
take a look at how this looks when there are spaces, the underline applies only to the text itself and not to the blank spaces.

https://github.com/user-attachments/assets/df36c243-c570-438d-a9cb-d8aeff6faff0

quonb:
I have found a bug with this PR: https://github.com/ghostty-org/ghostty/discussions/9940
HTTP urls do not open with ctrl-click (cmd-click) now because function `resolvePathForOpening` prepend urls like http://example.com with current working directory

kaplanelad:
@quonb i'll check it and open a fix. thanks

kaplanelad:
Apologies, I've opened a fix in PR #9942.

PR #9925, opened 2025-12-16
ci: color scheme GHA uploads to mirror

mitchellh:
This changes our GHA that updates our color schemes to also upload it to our dependency mirror at the same time. This prevents issues where the upstream disappears, which we've had many times.


mitchellh:
Will be hard to know if this works right until it runs... so we wait.

PR #9937, opened 2025-12-17
hide mouse after idle time based on `mouse-hide-after` config

kaplanelad:
issue: #8385

Implement a new `mouse-hide-after` duration config that hides the cursor after a period of no mouse movement while keeping it hidden during typing. The core tracks the last mouse move in `Surface.cursorPosCallback` and uses `Surface.maybeHideMouseAfterIdle()` to decide when to call `hideMouse()`. On macOS this is driven by a per-surface NSTimer in `SurfaceView_AppKit`, and on GTK by a per-surface GLib timeout in `class/surface.zig`, both reset on mouse movement. macOS uses `NSCursor.setHiddenUntilMouseMoves` to integrate with the native cursor API, while GTK relies on the existing `mouse_visibility` action wired through the core surface.

Tests on macos and linux


src/Surface.zig
```diff
@@ -1540,6 +1542,23 @@ fn modsChanged(self: *Surface, mods: input.Mods) void {
     }
 }
 
+/// Hide the mouse if `mouse-hide-after` is configured and the mouse has been
+/// idle (no movement) for at least that duration.
+pub fn maybeHideMouseAfterIdle(self: *Surface) void {
```
mitchellh: Do we need this core stuff at all? I think it might be cleaner to just implement this 100% in the apprt.

mitchellh (changes requested):
Thanks, overall very clean, just one major question.

kaplanelad:
@mitchellh Thanks for the review! I completely agree and have made the requested changes. hope this aligns with what you expected.

macos/Sources/Ghostty/Ghostty.Config.swift
```diff
@@ -548,6 +548,14 @@ extension Ghostty {
             return v;
         }
 
+        var mouseHideAfter: UInt {
```
bo2themax: Using `Duration` is better, like `undoTimeout` down below

bo2themax (changes requested):
Looks good on macOS, only a small suggestion

src/apprt/gtk/class/surface.zig
```diff
@@ -2654,6 +2693,27 @@ pub const Surface = extern struct {
         // Our pos changed, update
         priv.cursor_pos = pos;
 
+        // Track mouse movement time for `mouse-hide-after`.
+        priv.last_mouse_move_time = std.time.Instant.now() catch null;
```
pluiedev: Is this ever used? I don't think it is, right?

pluiedev (commented):
Looks good overall on the GTK side, just one question

src/apprt/gtk/class/surface.zig
kaplanelad: you are right. removed

bo2themax (approved):

PR #9938, opened 2025-12-17
feat: implement scroll-to-bottom on output option

### LINKED ISSUE #8408: Add `scroll-to-bottom` configuration to control scrolling behavior

benodiwal:
Closes #8408

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1070,12 +1070,6 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
 
         .scrollbar => |scrollbar| self.updateScrollbar(scrollbar),
 
-        .scroll_to_bottom => {
-            self.queueIo(.{
-                .scroll_viewport = .{ .bottom = {} },
-            }, .unlocked);
-        },
-
         .present_surface => try self.presentSurface(),
 
         .password_input => |v| try self.passwordInput(v),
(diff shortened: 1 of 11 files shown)


mitchellh (changes requested):
Thank you, but I'm not convinced this is the right approach. I _suspect_ that this might slow down our IO throughput noticeably since newlines are common and we have to now add a conditional there. It may not though since the conditional will almost always be the same value. 

If you can show benchmarks that may help...

But I still think the best way to do this would be to implement it in the _renderer_ actually and have the frame build check if our final line changed pins at all and if so, to trigger a scroll.

benodiwal:
Thanks @mitchellh! The renderer approach seems right.

  I tried these three implementations:

  1. Stream Handler (termio/stream_handler.zig) <- `Current commit`
  - Checks on every \n in linefeed()

  2. Renderer - Scrollbar Total (renderer/generic.zig)
  - Compares scrollbar.total between frames in updateFrame()

  3. Renderer - Pin Tracking (renderer/generic.zig) ← `your suggestion`
  - Checks if final line's pin changed (node pointer or y offset)
  
  I have all three ready on my local.  I would like your input on 2 as well. It seems simpler since it reuses existing scrollbar data without adding new fields.

benodiwal:
I have pushed the changes according to the approach you suggested using Pin Tracking.

mitchellh (approved):
Rebased this and cleaned it up further, the renderer can actually do the scroll directly. And by moving the scroll above the render state update we see it immediately.

PR #9945, opened 2025-12-17
macOS: Session Search

mitchellh:
Replaces #9785 

This adds session search to the command palette on macOS. Session search lets you jump to any running terminal based on title or working directory. The command palette shows you the title, working directory, and tab color (if any) to help you identify the terminal you care about.

This also enhances our command palette in general to support better sorting, more stable sorts when keys are identical, etc.

## Demo

https://github.com/user-attachments/assets/602a9424-e182-4651-bf08-378e9c5e1616

## Future

Since this inherits the command palette infrastructure, we don't have fuzzy search capabilities yet, but I still think this is useful already. We should add fuzzy searching in the future.

Thanks @phaistonian for the inspiration here but I did do a full rewrite.

**AI disclosure:** I used AI to assist with this in places, but I understand everything and touched up almost everything manually.

### DIFF
--- a/macos/Ghostty.xcodeproj/project.pbxproj
+++ b/macos/Ghostty.xcodeproj/project.pbxproj
@@ -157,6 +157,7 @@
 				"Helpers/Extensions/KeyboardShortcut+Extension.swift",
 				"Helpers/Extensions/NSAppearance+Extension.swift",
 				"Helpers/Extensions/NSApplication+Extension.swift",
+				"Helpers/Extensions/NSColor+Extension.swift",
 				"Helpers/Extensions/NSImage+Extension.swift",
 				"Helpers/Extensions/NSMenu+Extension.swift",
 				"Helpers/Extensions/NSMenuItem+Extension.swift",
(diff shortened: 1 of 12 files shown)


mitchellh:
Note @deanm: I addressed most of your points (thank you!) in this PR. Namely that sorting is stable for identical titles and working directories and also that you can search working directories. We don't support fuzzy matching today in general, so we'll address fuzziness later.

deanm:
> Note @deanm: I addressed most of your points (thank you!) in this PR. Namely that sorting is stable for identical titles and working directories and also that you can search working directories. We don't support fuzzy matching today in general, so we'll address fuzziness later.

This is great, thanks a million.

Would you be open to adding a "mode" to the command palette where it is only the focus entries (and doesn't show the "focus: " prefix), and then able to have a separate keybinding / etc? It's probably not critical but it's nice to not compete with the other commands and also have some more space for the titles, etc. Often you end up just doing shortcut + one or two letters + enter...

I'm happy to work on any of this as well, thanks!

mitchellh:
> Would you be open to adding a "mode" to the command palette where it is only the focus entries (and doesn't show the "focus: " prefix), and then able to have a separate keybinding / etc? It's probably not critical but it's nice to not compete with the other commands and also have some more space for the titles, etc. Often you end up just doing shortcut + one or two letters + enter...

Yeah, that's something I ripped out of the other PR, but you can see hints of me already mentally accepting the idea here: I split out the `terminalOptions` from `jumpOptions` so that a command palette can potentially have a mode to only show one set.

I'm open to it.

PR #9947, opened 2025-12-17
Feat/agent sparkle indicator

phaistonian:
Adds an optional “agent CLI” sparkle indicator for tab/window titles. When enabled, Ghostty will best-effort detect if the PTY foreground process is a supported agent CLI and prepend ✨ to the  computed title.

https://github.com/user-attachments/assets/e2d18898-6428-4c10-93c7-a6d0cf3024d0

 ### User-Facing Change

- New config: title-agent-indicator (default false)
- When true, tab/window titles get a ✨ prefix while an agent CLI is detected in the foreground.

 ### Detection Details

 - Best-effort + OS-dependent:
      - macOS: uses TIOCGPGRP + KERN_PROCARGS2 (argv/exec path) in addition to short process name.
      - Linux: uses TIOCGPGRP + /proc/<pid>/cmdline in addition to /proc/<pid>/comm.
 - Matches common agent names/variants (e.g. codex, openai-codex, gemini, gemini-cli, claude,  claude-code).

### Platform Notes

 - macOS: only polls for agent state when title-agent-indicator is enabled and updates on config reload.
 - GTK: title binding includes the agent-running property so the tab title updates appropriately.

### Usage

`title-agent-indicator = true`

—-

@mitchellh  This is more of working POC; An endgame would be to have provider-based icons (e.g. one for Google, one for Claude etc) and make it a default - perhaps. Cause, hey, we live in the AI era and all :)

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1008,6 +1008,7 @@ ghostty_surface_config_s ghostty_surface_inherited_config(ghostty_surface_t);
 void ghostty_surface_update_config(ghostty_surface_t, ghostty_config_t);
 bool ghostty_surface_needs_confirm_quit(ghostty_surface_t);
 bool ghostty_surface_process_exited(ghostty_surface_t);
+bool ghostty_surface_agent_running(ghostty_surface_t);
 void ghostty_surface_refresh(ghostty_surface_t);
 void ghostty_surface_draw(ghostty_surface_t);
 void ghostty_surface_set_content_scale(ghostty_surface_t, double, double);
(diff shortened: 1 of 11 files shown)


rockorager:
There is an existing escape sequence to set an "icon" over OSC. IMO we should look at that as an option

mitchellh:
Thanks for this. I think hardcoding this kind of detection isn't going to be the right approach, but if you want to maintain this in your own personal build, that's excellent for you. 😄 For official Ghostty, I think doing something that is based on process matching more generally in the config is probably the right way to go. Discussions are a better place to discuss that than PR. Thank you, though.

PR #9950, opened 2025-12-18
New OSC parser

jcollie:
This replaces the OSC parser with one that only uses a state machine to determine which OSC is being handled, rather than parsing the whole OSC. Once the OSC command is determined the remainder of the data is stored in a buffer until the terminator is found. The data is then parsed to determine the final OSC command.

```
→ poop './zig-out/bin/ghostty-bench-old +osc-parser --data=osc.txt' './zig-out/bin/ghostty-bench +osc-parser --data=osc.txt'
Benchmark 1 (3 runs): ./zig-out/bin/ghostty-bench-old +osc-parser --data=osc.txt
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          4.19s  ± 69.1ms    4.12s  … 4.25s           0 ( 0%)        0%
  peak_rss           5.86MB ± 37.8KB    5.84MB … 5.91MB          0 ( 0%)        0%
  cpu_cycles         15.3G  ±  330M     14.9G  … 15.5G           0 ( 0%)        0%
  instructions       41.1G  ± 7.81      41.1G  … 41.1G           0 ( 0%)        0%
  cache_references    130M  ± 2.40M      128M  …  132M           0 ( 0%)        0%
  cache_misses        658K  ± 88.5K      603K  …  760K           0 ( 0%)        0%
  branch_misses      31.5M  ±  112K     31.4M  … 31.6M           0 ( 0%)        0%
Benchmark 2 (4 runs): ./zig-out/bin/ghostty-bench +osc-parser --data=osc.txt
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          1.54s  ± 7.08ms    1.53s  … 1.55s           0 ( 0%)        ⚡- 63.3% ±  2.1%
  peak_rss           5.84MB ±  270KB    5.45MB … 6.02MB          1 (25%)          -  0.4% ±  7.1%
  cpu_cycles         6.07G  ± 7.82M     6.06G  … 6.08G           0 ( 0%)        ⚡- 60.3% ±  2.7%
  instructions       17.3G  ± 39.5      17.3G  … 17.3G           1 (25%)        ⚡- 57.8% ±  0.0%
  cache_references   64.7M  ±  655K     63.8M  … 65.3M           0 ( 0%)        ⚡- 50.4% ±  2.4%
  cache_misses        733K  ±  116K      611K  …  890K           0 ( 0%)          + 11.4% ± 31.7%
  branch_misses      21.7M  ±  167K     21.5M  … 21.9M           0 ( 0%)        ⚡- 31.3% ±  0.9%
```

### DIFF
--- a/src/terminal/osc.zig
+++ b/src/terminal/osc.zig
@@ -315,22 +315,22 @@ pub const Parser = struct {
     /// Optional allocator used to accept data longer than MAX_BUF.
     /// This only applies to some commands (e.g. OSC 52) that can
     /// reasonably exceed MAX_BUF.
-    alloc: ?Allocator = null,
+    alloc: ?Allocator,
 
     /// Current state of the parser.
-    state: State = .start,
+    state: State,
 
     /// Buffer for temporary storage of OSC data
-    buffer: [MAX_BUF]u8 = undefined,
+    buffer: [MAX_BUF]u8,
     /// Fixed writer for accumulating OSC data
-    fixed: ?std.Io.Writer = null,
+    fixed: ?std.Io.Writer,
     /// Allocating writer for accumulating OSC data
-    allocating: ?std.Io.Writer.Allocating = null,
+    allocating: ?std.Io.Writer.Allocating,
     /// Pointer to the active writer for accumulating OSC data
-    writer: ?*std.Io.Writer = null,
+    writer: ?*std.Io.Writer,
 
     /// The command that is the result of parsing.
-    command: Command = .invalid,
+    command: Command,
 
     pub const State = enum {
         start,
--- a/src/terminal/osc.zig
+++ b/src/terminal/osc.zig
@@ -309,170 +309,90 @@ pub const Terminator = enum {
 };
 
 pub const Parser = struct {
+    /// Maximum size of a "normal" OSC.
+    const MAX_BUF = 2048;
+
     /// Optional allocator used to accept data longer than MAX_BUF.
     /// This only applies to some commands (e.g. OSC 52) that can
     /// reasonably exceed MAX_BUF.
-    alloc: ?Allocator,
+    alloc: ?Allocator = null,
 
     /// Current state of the parser.
-    state: State,
+    state: State = .start,
 
-    /// Current command of the parser, this accumulates.
-    command: Command,
+    /// Buffer for temporary storage of OSC data
+    buffer: [MAX_BUF]u8 = undefined,
+    /// Fixed writer for accumulating OSC data
+    fixed: ?std.Io.Writer = null,
+    /// Allocating writer for accumulating OSC data
+    allocating: ?std.Io.Writer.Allocating = null,
+    /// Pointer to the active writer for accumulating OSC data
+    writer: ?*std.Io.Writer = null,
 
-    /// Buffer that stores the input we see for a single OSC command.
-    /// Slices in Command are offsets into this buffer.
-    buf: [MAX_BUF]u8,
-    buf_start: usize,
-    buf_idx: usize,
-    buf_dynamic: ?*std.ArrayListUnmanaged(u8),
-
-    /// True when a command is complete/valid to return.
-    complete: bool,
-
-    /// Temporary state that is dependent on the current state.
-    temp_state: union {
-        /// Current string parameter being populated
-        str: *[:0]const u8,
-
-        /// Current numeric parameter being populated
-        num: u16,
-
-        /// Temporary state for key/value pairs
-        key: []const u8,
-    },
-
-    // Maximum length of a single OSC command. This is the full OSC command
-    // sequence length (excluding ESC ]). This is arbitrary, I couldn't find
-    // any definitive resource on how long this should be.
-    //
-    // NOTE: This does mean certain OSC sequences such as OSC 8 (hyperlinks)
-    //       won't work if their parameters are larger than fit in the buffer.
-    const MAX_BUF = 2048;
+    /// The command that is the result of parsing.
+    command: Command = .invalid,
 
     pub const State = enum {
-        empty,
+        start,
         invalid,
-        swallow,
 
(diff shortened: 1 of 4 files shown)


src/terminal/osc.zig
```diff
@@ -600,390 +539,162 @@ pub const Parser = struct {
...
+                else => self.state = .invalid,
+            },
+
+            .@"118" => switch (c) {
+                ';' => if (self.ensureAllocator()) self.writeToFixed(),
+                else => self.state = .invalid,
+            },
+
+            .@"119" => switch (c) {
+                ';' => if (self.ensureAllocator()) self.writeToFixed(),
```
pluiedev: Can't this be collapsed into one single arm (and a single variant)?

src/terminal/osc.zig
```diff
@@ -994,776 +705,791 @@ pub const Parser = struct {
...
+            .@"10" => self.parseOscColor(terminator_ch),
+            .@"11" => self.parseOscColor(terminator_ch),
+            .@"12" => self.parseOscColor(terminator_ch),
+            .@"13" => self.parseOscColor(terminator_ch),
+            .@"14" => self.parseOscColor(terminator_ch),
+            .@"15" => self.parseOscColor(terminator_ch),
+            .@"16" => self.parseOscColor(terminator_ch),
+            .@"17" => self.parseOscColor(terminator_ch),
+            .@"18" => self.parseOscColor(terminator_ch),
+            .@"19" => self.parseOscColor(terminator_ch),
```
pluiedev: Collapse these too

pluiedev (approved):
The idea looks good though the diff looks a bit huge and intimidating - if the tests all pass though then I think I don't really have much to object to

src/terminal/osc.zig
jcollie: Yes, done.

src/terminal/osc.zig
jcollie: Done.

src/terminal/osc.zig
```diff
@@ -485,75 +405,108 @@ pub const Parser = struct {
...
+    }
 
-            // We have to do this here because it will never reach the
-            // switch statement below, since our buf_idx will always be
-            // too high after this.
-            self.complete = false;
+    /// Set up an allocating Writer to collect the rest of the OSC data. If we
+    /// don't have an allocator or setting up the allocator fails, fall back to
+    /// writing to a fixed buffer and hope that it's big enough.
+    pub inline fn writeToAllocating(self: *Parser) void {
```
mitchellh: Do you have numbers to show the `inline` here makes a noticeable difference? I don't like inlining non-trivial functions like this without data, so just checking.

src/terminal/osc.zig
```diff
@@ -309,170 +309,90 @@ pub const Terminator = enum {
...
 
 pub const Parser = struct {
+    /// Maximum size of a "normal" OSC.
+    const MAX_BUF = 2048;
+
     /// Optional allocator used to accept data longer than MAX_BUF.
     /// This only applies to some commands (e.g. OSC 52) that can
     /// reasonably exceed MAX_BUF.
-    alloc: ?Allocator,
+    alloc: ?Allocator = null,
```
mitchellh: Nit: would prefer to not add defaults if possible and introduce a new `pub const default`. I have hesitations in certain areas but this is not one of them.

mitchellh (changes requested):
A couple small questions. 

Sorry for the delay.

I think overall this is good to go. I want to make small improvements and I think we should consider moving our `parse*` helpers out of `Parser` (don't depend on `self`) and standalone maybe in `src/terminal/osc/` (folder) so they're more obvious and easy to test, and the parser is also on its own easy to reason about, but no need to do that in this PR, I'd actually prefer we didn't.

The numbers look great and I think buffering OSC in general instead of doing a live parse makes sense.

src/terminal/osc.zig
jcollie: Yeah, it makes a difference but I don't have any current numbers.

src/terminal/osc.zig
jcollie: That shouldn't be too hard to do.

src/terminal/osc.zig
mitchellh: Instead of forcing inline at the definition site, I wonder if it makes more sense to use a `@call(.always_inline` at the call site instead. We do that for some other stuff to avoid forcing callers to always inline in every single scenario which I think wouldn' be generally desirable.

jcollie: I just realized that `ensureAllocating`, `writeToFixed`, and `writeToAllocating` do not need to be marked as `pub` as they aren't meant to be called from outside the OSC parser. Would that alleviate the concerns?

mitchellh: We should clean that up for sure but no, the concern is that `next` is a massive function that is inline. Sorry I thought I commented there, I'm talking about `next`

jcollie: Ah, yeah I don't know if I have numbers on that. Putting `inline` on next may just have been a reflexive move on my part. Yeah I think that removing `inline` from the definition of `next` and using `@call()` to inline it at critical call sites makes sense.

src/terminal/osc.zig
jcollie: Done.

src/terminal/osc.zig
jcollie: `inline` removed from the definition of `next` and `@call` used at critical points.

mitchellh (approved):
I’m away from keyboard the next few hours so feel free to merge but any regressions you’re on your own for a bit

jcollie:
> I’m away from keyboard the next few hours so feel free to merge but any regressions you’re on your own for a bit

LOL. I've been daily driving this for a while locally so I don't expect any major borkage.

PR #9670, opened 2025-11-22
macOS: Only change the icon if needed

bo2themax:
Fixes #9666

> [!NOTE]
> AI proofread my comments

Issue #9666, opened 2025-11-21
macOS: Custom icon not consistent/persistent


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9665

<div type='discussions-op-text'>

<sup>Originally posted by **burnsra** November 21, 2025</sup>
### Issue Description

When using a custom icon and Ghostty is set as "Keep in Dock", the icon is not consistent/persistent.  The icon changes between the custom icon and a macOS folder icon (please see the attached [video](https://github.com/user-attachments/assets/7ef742b9-0133-49cf-ae5c-acfc7104bae1) demonstrating the bug).

I've tracked this bug down to the commit that moved the setting of the icon to a separate thread, see commit [7472fb7](https://github.com/ghostty-org/ghostty/commit/7472fb773290d45e1402244569678bf9700d1306).

I've found moving the NSWorkspace functions into the MainActor fixes the issue.  I have a fix committed to my fork if you'd like me to submit a pull request, please see commit [a8971fe](https://github.com/burnsra/ghostty/commit/a8971fe731905475a57dcd7852335a8406f4a76f)

### Expected Behavior

The custom icon is persisted throughout launching/quitting the application

### Actual Behavior

The custom icon is NOT persisted throughout launching/quitting the application

### Reproduction Steps

1. Launch Ghostty
2. Right click on the icon in the dock and choose "Options" -> "Keep in Dock"
3. Open Ghostty settings via the File menu
4. Define a custom icon, such as "blueprint" (which is included by default)
5. Reload settings via the File menu
6. Launch/quit the application several times

### Ghostty Version

```text
Ghostty 1.3.0-main+5f3645433

Version
  - version: 1.3.0-main+5f3645433
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS Tahoe 26.1 (25B78) arm64

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
macos-icon = blueprint
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
[...]

PR #9951, opened 2025-12-18
fix(macOS): re-apply icon after app update

jknlsn:
### Summary

Fixes the regression where custom icons are not re-applied after app updates, particularly affecting users on the tip channel.

### Problem

PR #9670 introduced caching to avoid redundantly setting the app icon on every launch. However when Ghostty updates, the app bundle is replaced and the custom icon is reset by macOS. Since `UserDefaults` persists across updates, the cached icon name still matches the desired icon name, causing the icon update to be incorrectly skipped.

This was reported by users in #9670 comments as well, that after updating Ghostty the custom icon would disappear and require manual re-application.

### Solution

- Track the app build number (`CFBundleVersion`) alongside the icon name in `UserDefaults`
- Re-apply the icon if either the icon config has changed OR the build number has changed
- Only update `UserDefaults` if `NSWorkspace.setIcon()` succeeds, preventing false-positive caching on failure

I used `CFBundleVersion` (build number, e.g. `13509`) rather than `CFBundleShortVersionString` (e.g. `1.2.3`) because tip builds increment the build number with each release while the short version string remains constant. I considered combining both but this seemed redundant.

### Related

- Fixes regression mentioned in comments on #9670
- Original issue: #9666
- Original discussion: #9665

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -982,9 +982,15 @@ class AppDelegate: NSObject,
             appIconName = (colorStrings + [config.macosIconFrame.rawValue])
                 .joined(separator: "_")
         }
-        // Only change the icon if it has actually changed
-        // from the current one
-        guard UserDefaults.standard.string(forKey: "CustomGhosttyIcon") != appIconName else {
+
+        // Only change the icon if it has actually changed from the current one,
+        // or if the app build has changed (e.g. after an update that reset the icon)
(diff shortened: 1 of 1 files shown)


edrozenberg:
Really great fix thanks! This eliminates another annoying "papercut" every time I update. With the various recent fixes I can update to dev and:
* The icon doesn't change 🥳
* The quick terminal size works as expected 🎉
* The quick terminal session restores with all my splits 🎈

PR #9960, opened 2025-12-18
Add configurable corner radius for macOS quick terminal window

doso:
## Summary

Adds a configurable corner radius option for the macOS quick terminal window, allowing users to customize the rounded corners to match their preferences or other macOS windows.

## Changes

- Added `quick-terminal-corner-radius` configuration option (default: ~~16.0 points~~ unset a.k.a 0)
- Updated `QuickTerminalWindow` to accept and apply configurable corner radius
- Updated `QuickTerminalController` to pass config value from `DerivedConfig`
- Added Swift property `quickTerminalCornerRadius` to `Ghostty.Config`

## Testing

- [x] Tested on macOS with default value (~~`16.0`~~ unset)
- [x] Tested with custom values in config file
- [x] Verified corners render correctly in production build
- [x] Verified default matches standard macOS window corner radius

`quick-terminal-corner-radius = 32`

image

`quick-terminal-corner-radius = 16` ~~(default)~~

image

## Configuration

Users can now set the corner radius in their config file `quick-terminal-corner-radius = 20` 

The default value of ~~`16.0`~~ unset a.k.a 0 points matches the standard macOS window corner radius

## Related

This addresses the visual inconsistency where the quick terminal window had sharp corners while other macOS windows have rounded corners. https://github.com/ghostty-org/ghostty/discussions/8666

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -143,6 +143,10 @@ class QuickTerminalController: BaseTerminalController {
             delegate: self
         )
 
+        if let qtWindow = window as? QuickTerminalWindow {
+            qtWindow.setupRoundedCorners(cornerRadius: derivedConfig.quickTerminalCornerRadius)
+        }
+
         // Clear out our frame at this point, the fixup from above is complete.
         if let qtWindow = window as? QuickTerminalWindow {
             qtWindow.initialFrame = nil
@@ -722,6 +726,7 @@ class QuickTerminalController: BaseTerminalController {
         let quickTerminalAutoHide: Bool
         let quickTerminalSpaceBehavior: QuickTerminalSpaceBehavior
         let quickTerminalSize: QuickTerminalSize
+        let quickTerminalCornerRadius: Double
         let backgroundOpacity: Double
         let backgroundBlur: Ghostty.Config.BackgroundBlur
 
@@ -731,6 +736,7 @@ class QuickTerminalController: BaseTerminalController {
             self.quickTerminalAutoHide = true
             self.quickTerminalSpaceBehavior = .move
             self.quickTerminalSize = QuickTerminalSize()
+            self.quickTerminalCornerRadius = 16.0
             self.backgroundOpacity = 1.0
             self.backgroundBlur = .disabled
         }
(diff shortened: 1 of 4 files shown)


bo2themax (changes requested):
LGTM, but the default value should be the same as previous versions, so that users doesn’t want this don’t have to change their current config

> Could be changed to 0 or a optional value

doso:
> LGTM, but the default value should be the same as previous versions, so that users doesn’t want this don’t have to change their current config
> 
> > Could be changed to 0 or a optional value

Thanks for the review, right, the default shouldn’t surprise people. I have switched the default to ‎`optional` so the quick terminal stays boxy like before, and make the rounding opt‑in.

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
```diff
@@ -143,6 +143,14 @@ class QuickTerminalController: BaseTerminalController {
             delegate: self
         ))
         
+        if
+            let qtWindow = window as? QuickTerminalWindow,
+            let radius = derivedConfig.quickTerminalCornerRadius,
+            radius > 0
+        {
+            qtWindow.setupRoundedCorners(cornerRadius: CGFloat(radius))
```
bo2themax: There should be an `else` statement to handle reloading to 0/nil

bo2themax (changes requested):

macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
doso: I've added an else block to explicitly reset the corner radius to 0 when the config value is nil or 0.

bo2themax (approved):

frolvanya:
@bo2themax do you know what's the approximate timeline for this PR to be merged and if there're any blockers? (no rush, just curious when it hits the tip release)

P.S: After building this locally I noticed that I still have a barely visible rectangular corner under the curved one
image
```txt
quick-terminal-size = 99%, 98%
quick-terminal-screen = mouse
quick-terminal-autohide = false
quick-terminal-corner-radius = 16
```

curtismoncoq:
Same issue as @frolvanya
I managed to get it like op by tweaking bg-opacity,
but it broke the liquid-glass effect introduced by 'background-blur = macos-glass-regular'.
It also broke my global hotkey for the quick-terminal.

mitchellh (changes requested):
I made some changes and I generally want to accept this, but I'm also seeing the background issue as noted by prior reporters. If you resolve this I will merge.

PR #9970, opened 2025-12-19
macOS: Jump palette

deanm:
Hey @mitchellh 

This is a rough attempt at adding the "jump palette" mode for only changing between windows.

A few notes:

- I didn't add it to the View toolbar like Command Palette is, I kinda think you'd only use it via shortcuts but maybe makes sense to add it.
- The code comments say command palette only works on macos, but it seems like there is also some gtk code for it? I didn't do anything for the gtk side of things.
- This is the first time I've written any Swift, so I'm not sure how things look style-wise and I'm also not sure about a few things like binding usage.
- In terms of correctness, there might be some things if you tried to activate the jump palette while the command palette is open or vice versa, but I am not sure you can even really do that now (don't think the keybinding will work). The other thing to check would to be I didn't get anything wrong so there is "staleness" in that opening that command palette is still only showing focus entries or something like that.

Thanks a ton, now that I'm using this it's so great!


macos/Sources/Features/Command Palette/TerminalCommandPalette.swift
```diff
@@ -9,6 +9,9 @@ struct TerminalCommandPaletteView: View {
     /// result in the view disappearing.
     @Binding var isPresented: Bool
 
+    /// Set this to true for the jump palette mode (only focus action).
```
mitchellh, line 12: I'd just change this to an enum of `mode` (two values). And it doesn't have to be a `@Binding` since its only one-way.

macos/Sources/Features/Command Palette/TerminalCommandPalette.swift
```diff
@@ -58,13 +62,18 @@ struct TerminalCommandPaletteView: View {
...
     private var commandOptions: [CommandOption] {
         var options: [CommandOption] = []
-        // Updates always appear first
-        options.append(contentsOf: updateOptions)
-        
+
+        if !isJumpPalette {
+            // Updates always appear first
+            options.append(contentsOf: updateOptions)
+        }
```
mitchellh, line 69: Change this to a `switch` on mode

macos/Sources/Features/Command Palette/TerminalCommandPalette.swift
```diff
@@ -154,7 +163,7 @@ struct TerminalCommandPaletteView: View {
                 }
 
                 return CommandOption(
-                    title: "Focus: \(displayTitle)",
+                    title: isJumpPalette ? displayTitle : "Focus: \(displayTitle)",
```
mitchellh, line 166: Remove this completely and change it so that in command palette mode we run a `map` over it which modifies the title to add a prefix.

macos/Sources/Features/Terminal/TerminalView.swift
```diff
@@ -31,7 +31,10 @@ protocol TerminalViewModel: ObservableObject {
 
     /// The command palette state.
     var commandPaletteIsShowing: Bool { get set }
-    
+    /// The command palette is reused for the focus palette, this indicates what
+    /// mode it should be in, true for the focus palette.
+    var commandPaletteIsJumpPalette: Bool { get set }
```
mitchellh, line 36: Replace `commandPaletteIsShowing` with `commandPaletteState` and `CommandPaletteState` is an enum with 3 values: `hidden, command, jump`.

macos/Sources/Features/Command Palette/CommandPalette.swift
```diff
@@ -59,6 +59,7 @@ struct CommandOption: Identifiable, Hashable {
 
 struct CommandPaletteView: View {
     @Binding var isPresented: Bool
+    @Binding var isJumpPalette: Bool
```
mitchellh, line 62: Remove this, doesn't have to be a binding, and just make it `placeholder: String` and let downstream callers set it.

mitchellh (changes requested):
Thanks! I've submitted some stylistic reviews, overall architecture looks right.

deanm:
Thanks a lot for such a quick review, all of your comments make sense to me, I haven't really written Swift before (and am not particularly trying to learn), so I'm happy if you want to just pick up the change and clean it up the way you prefer. If not I can try to address the comments, but the back and forth might just be more work in the end. Thanks!

jcollie:
Edited the title to note that this PR is specific to macOS.

PR #9972, opened 2025-12-19
renderer/metal: clamp texture sizes to the maximum allowed by the device

mitchellh:
This prevents a crash in our renderer when it is larger.

I will pair this with apprt changes so that our mac app won't ever allow a default window larger than the screen but we should be resilient at the renderer level as well.

### DIFF
--- a/src/renderer/Metal.zig
+++ b/src/renderer/Metal.zig
@@ -434,7 +434,7 @@ fn chooseDevice() error{NoMetalDevice}!objc.Object {
 }
 
 /// Determines the maximum 2D texture size supported by the device.
-/// We need to clamp our frame size to this if its larger.
+/// We need to clamp our frame size to this if it's larger.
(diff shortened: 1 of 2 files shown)


mitchellh:
I think we should consider exposing those to every place we create a texture just to make sure we never create a texture too large but this addresses a major case today.

src/renderer/Metal.zig
```diff
@@ -412,3 +432,23 @@ fn chooseDevice() error{NoMetalDevice}!objc.Object {
     const device = chosen_device orelse return error.NoMetalDevice;
     return device.retain();
 }
+
+/// Determines the maximum 2D texture size supported by the device.
+/// We need to clamp our frame size to this if its larger.
```
jparise:
```suggestion
/// We need to clamp our frame size to this if it's larger.
```

😉

PR #9975, opened 2025-12-19
macos: window width/height should be clamped, work with position

mitchellh:
Fixes #9952
Fixes #9969

This fixes our `constrainToScreen` implementation to properly clamp the window size to the visible screen its coming on as documented. Further, this addresses the positioning problem, too.

### DIFF
--- a/macos/Sources/Helpers/Extensions/NSWindow+Extension.swift
+++ b/macos/Sources/Helpers/Extensions/NSWindow+Extension.swift
@@ -16,19 +16,23 @@ extension NSWindow {
         return firstWindow === self
     }
 
-    /// Adjusts the window origin if necessary to ensure the window remains visible on screen.
+    /// Adjusts the window frame if necessary to ensure the window remains visible on screen.
+    /// This constrains both the size (to not exceed the screen) and the origin (to keep the window on screen).
     func constrainToScreen() {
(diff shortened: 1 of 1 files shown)


devxoul:
Hmm with `window-position-x = 9999` the window still present from the middle of the screen. It used to present at the right-most until 1.2.3. @mitchellh 

Screenshot 2025-12-21 at 12 52 02 AM

devxoul:
@mitchellh I've created a new PR #10007 to address this issue.

After further analysis, I believe the original approach in #9952 is correct. Here's why:

**`DefaultSize.apply()` should only handle size, not position:**

1. `apply()` is called from `returnToDefaultSize(_:)` menu action (line 1185)
2. When a user triggers "Return to Default Size", only the **size** should reset—the window should stay where the user moved it
3. If position logic were added inside `apply()`, the window would unexpectedly jump back to its initial position whenever the user resets the size

This is why position adjustment belongs **outside** of `apply()`, specifically during initial window setup in `windowDidLoad`.

The fix in #9975 addressed `constrainToScreen()` clamping, but `constrainToScreen()` doesn't interpret `window-position-x = 9999`—that's `adjustForWindowPosition()`'s job. And `adjustForWindowPosition()` isn't called in the `.contentIntrinsicSize` path.

PR #9976, opened 2025-12-19
i18n: Localize Nautilus .py script

dmatos2012:
Closes #9266. 

Big Note: I noticed that this worked properly under `NixOS`, but on my `Ubuntu` VM it didnt. 

The reason is in [src/build/GhosttyI18n.zig](https://github.com/ghostty-org/ghostty/blob/73a93abf7b3449bb57fe3e8bc0a04d56b85e7ac0/src/build/GhosttyI18n.zig#L24-L31) because the locale is expected in the `<lang>_<region>` without the encoding suffix. `<lang>_<region>_<encoding>` 
```
        // There is no encoding suffix in the LC_MESSAGES path on FreeBSD,
        // so we need to remove it from `locale` to have a correct destination string.
        // (/usr/local/share/locale/en_AU/LC_MESSAGES)
        const target_locale = comptime if (builtin.target.os.tag == .freebsd)
            std.mem.trimRight(u8, locale, ".UTF-8")
        else
            locale;

```
If i force it to always trim the encoding it works, but I am guessing its there for a reason ,so maybe some of the maintainer can shed some light in the best way forward, as I am not an expert in how other systems deal with it. Here you see `Open in Ghostty` -> Abrir con Ghostty

image

- I wanted to format the `py` file with `ruff` but didnt want to drown the changes, so maybe something that could be worth doing so that also our `py` files have std formatting. 

> [!NOTE]
> Used AI only for helping me debug where the locales could be and why was it not detected, but no code help whatsoever


dist/linux/ghostty_nautilus.py
```diff
@@ -20,7 +20,13 @@
 from os.path import isdir
 from gi import require_version
 from gi.repository import Nautilus, GObject, Gio, GLib
+import os
+import gettext
 
+DOMAIN = "com.mitchellh.ghostty"
+share_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
swsnr: I think this would be more readable with pathlib:

```python
from pathlib import Path

locale_dir = Path(__file__).absolute().parent.parent.parent / "locale"
```

And does the Ghostty build system even guarantee that the locales always end up at this path, relative to the nautilus extension directory?

src/build/GhosttyI18n.zig
```diff
@@ -74,6 +76,8 @@ fn createUpdateStep(b: *std.Build) !*std.Build.Step {
         "-o",
         "-",
     });
+    // Silences the "unknown extension" errors
+    _ = xgettext.captureStdErr();
```
swsnr: I don't think that that's a good idea.

I believe the proper way to handle different types of sources with different flags in `xgettext` is to use separate xgettext invocations to extract messages into intermediate pot files, and then use a final `xgettext` invocation to merge the intermediate pot files into the final one.

swsnr (commented):
I'm not a member of the ghostty project, so please excuse me for leaving a review, but as the original author of this script (albeit for Wezterm originally), and being somewhat familiar with xgettext and Python, I'd like to leave two remarks concerning this pull request 🙂

dist/linux/ghostty_nautilus.py
trag1c: I'll note that you can use `.parents[2]` instead of `.parent.parent.parent` =)

swsnr: @trag1c Oh, thanks 🙏 I learn a new thing every day 🙂

dmatos2012: @swsnr As far as I can see,  the `nautilus` extension is here
[GhosttyResources](https://github.com/ghostty-org/ghostty/blob/main/src/build/GhosttyResources.zig#L379-L382)
 ```
 try steps.append(b.allocator, &b.addInstallFile(
        b.path("dist/linux/ghostty_nautilus.py"),
        "share/nautilus-python/extensions/ghostty.py",
    ).step);
    
```

and the `locale` is relative to `share` so `share/locale/....` 

[README](https://github.com/ghostty-org/ghostty/blob/main/po/README_CONTRIBUTORS.md)
```
During the build process, each locale in `.po` files is compiled
into binary `.mo` files, stored under `share/locale/<LOCALE>/LC_MESSAGES/com.mitchellh.ghostty.mo`.
This can be directly accessed by `libintl`, which provide the various `gettext` 
....

```

The only issue is what I mentioned in the original comment where some other platforms put the `locale` with a different suffix and then it doesnt work, so it just fallbacks.

File: src/build/GhosttyI18n.zig (hunk 2 above)
Comment (dmatos2012) [reply src/build/GhosttyI18n.zig]: Yeah thanks, I have edited it now. Much cleaner this way.

**Comment by dmatos2012** [comment]
@swsnr Even though it seems you are not an official maintainer, I saw emoji reactions by at least a maintainer, so I assume these changes were approved as well from them. I made those changes, lmk and thanks for the review!

File: dist/linux/ghostty_nautilus.py (hunk 1 above)
Comment (swsnr) [reply dist/linux/ghostty_nautilus.py]: I guess it'll work then.  However, I think a comment which explains this rationale would help for future understanding :slightly_smiling_face:

File: src/build/GhosttyI18n.zig (hunk 2 above)
Comment (swsnr) [reply src/build/GhosttyI18n.zig]: You're welcome.  Can't review the zig code tho, as I'm entirely unfamiliar with the language :slightly_smiling_face:

File: src/build/GhosttyI18n.zig [hunk 3]
```diff
@@ -63,18 +63,16 @@ pub fn addStepDependencies(
...
-        "--package-name=" ++ domain,
-        "--msgid-bugs-address=m@mitchellh.com",
-        "--copyright-holder=\"Mitchell Hashimoto, Ghostty contributors\"",
-        "-o",
-        "-",
     });
 
+    // Collect to intermediate .pot file
+    xgettext.addArg("-o");
+    const gtk_pot = xgettext.addOutputFileArg("gtk.pot");
```
swsnr, line 74: I'm entirely unfamiliar with zig and its build system, so pardon me if this is a stupid question, but shouldn't you cleanup these temporary POT files after the xgettext_merge down below has finished?

swsnr (approved):
Looks sound now, but being unfamiliar with zig I can't really comment on the zig code changes.  The general idea of the zig changes looks good to me, but I can't really review the details :innocent:

src/build/GhosttyI18n.zig
pluiedev: No, since it would only be output to Zig's output cache (`./.zig-cache`) and not the final prefix

pluiedev (approved):
Code looks good sans all the merge conflicts. My apologies for letting this sit for too long — looks like the translators would need to do extra work since most of the new strings are already localized as part of #10632.

00-kat (approved):

00-kat:
Should these wait until after #10632 is done, or should we try to sneak these into 1.3?

pluiedev:
If it's not too much work for the translators then I think we should try to squeeze these in 1.3. It's partially my fault this PR has stalled for this long

00-kat:
It's a single short string, so I don't expect it to be too hard on the translators; I'm more concerned about the administration nightmare, since:
  - This PR needs to be merged, and it conflicts with every file in po/, but even after those are fixed every translations update merged would conflict again. So translations merging would have to be put on hold after this is rebased.
  - After merged, sixteen translation teams need to be pinged again since they already updated their translations. Not sure if we reuse #10632 or make a new one, but this probably isn't too hard.
  - Two translation teams haven't created a PR yet, so they may or may not decide to include this translation alongside their 1.3 PR. Assuming they base their branch off tip at the time of creation, they are likely to.
  - Three translation teams have an open PRs and are halfway through a review; if they rebase now, the situation will be confusing for everyone involved.
  - Four translation teams have an open PR, but don't have a review yet; I don't expect all of them to rebase, so we'd need to ask them to rebase, hopefully before they get a review.

It's not like it's *too* difficult though, it's just really easy to get confused with the translations that haven't been updated yet. On the contrary, nothing says we *have* to get *every* translation for this in before 1.3, so we could just prioritize the sixteen already updated translations.

dmatos2012:
@00-kat @pluiedev Merged, and I updated it with latest main translation files, and then included the new string. Thx

src/build/GhosttyI18n.zig
swsnr: Thanks :pray:

00-kat:
#9904 was merged recently which adds two strings, which means we have to deal with the problem anyway. At the moment there's been three PRs which add those two strings (#10788 #10792 #10794). It's probably best to merge this **as soon as possible**, so that we don't have to deal with the same problem twice and with an extra division.

(This needs another merge/rebase since it conflicts with every translation file *again* apparently…)

dmatos2012:
@00-kat done! Hope no more new strings are added :stuck_out_tongue:

src/build/GhosttyI18n.zig
```diff
@@ -63,18 +63,16 @@ pub fn addStepDependencies(
 fn createUpdateStep(b: *std.Build) !*std.Build.Step {
     const xgettext = b.addSystemCommand(&.{
         "xgettext",
-        "--language=C", // Silence the "unknown extension" errors
```
jcollie, line 66: Is there a reason for removing this comment?

jcollie (approved):
Small nit, otherwise LGTM.

src/build/GhosttyI18n.zig
dmatos2012: No, not really. I just forgot to add it back once I had to remove it in the initial refactor.  Put it back again!

jcollie (approved):
LGTM

PR #9979, opened 2025-12-20
i18n: add Finnish translation

JokkeeZ:
This PR adds Finnish translations.


trag1c:
Hi, looks like this slipped under our radar (sorry!) and didn't get the standard translation PR treatment. Unfortunately, 7 months have passed and this PR hasn't seen any interest from other Finnish speakers + many new strings were added that have to be translated as well, so I'm closing this as stale. We're still open to accepting a PR adding the Finnish locale in the future once there are more people willing to review and maintain it. Thanks!

PR #9983, opened 2025-12-20
macOS: use `NSDockTilePlugIn` to update app icons

bo2themax:
* Using MAP prohibited `NSDockTilePlugIn` to update custom icons more reliably. And it also fixes the corner radius issue on older os: #8870
* Changes in AboutWindow cc @jparise 
  * Start cycling with current icon
  * Stop cycling icons when AboutWindow is closed 
  * Add menu to copy icon config

Xnip2025-12-20_18-40-58

> [!WARNING]
> Upgrading from `macOS-custom-icon` needs to manually open the app once to update the icon, since this plugin is running under `com.apple.dock.external.extra.arm64`, which has sandbox enabled.
>
> When first upgraded to this commit, a notification about the dock plugin will pop up. The user has to enable this to change the icon smoothly.
> image

Here are the testing results on [Tahoe](https://github.com/user-attachments/assets/e5fc8354-5f5c-4280-805f-88f043ceadca) and [Sequoia](https://github.com/user-attachments/assets/633d9a07-7d9d-4806-8496-82ddaffb8833):

> When you see some pause in the recording, that's when I rebuild or replace the older version with the latest.

This also fixes some issues when changing between different styles, consistency issues, and resetting from others to `official`.

### Developer's Note

This shouldn't affect current CI flow, since this new target is just a bundle not runnable, and I tested with archiving, exporting and signing in Xcode, nothing big changed.

> [!NOTE]
> AI helped me to write the typo ignore-re and proofread my comments


mitchellh (changes requested):
Sorry this took forever to get to. Overall, using an NSDockTilePlugIn sounds great and this implementation is decent. I just shuffled a bunch of stuff around, could you re-review yourself and make sure this all looks/works right?

I think before merge it'd be nice to also clean up the AppIcon stuff so that the enum uses an encoded ColorizedGhosttyIcon rather than that `_`-separated custom encoding we have in this PR. We should be able to make that Coddle if we use a wrapper type around NSColor or something.

bo2themax:
> Sorry this took forever to get to. Overall, using an NSDockTilePlugIn sounds great and this implementation is decent. I just shuffled a bunch of stuff around, could you re-review yourself and make sure this all looks/works right?
> 
> I think before merge it'd be nice to also clean up the AppIcon stuff so that the enum uses an encoded ColorizedGhosttyIcon rather than that `_`-separated custom encoding we have in this PR. We should be able to make that Coddle if we use a wrapper type around NSColor or something.

Removed `_`-separated custom encoding now. It will affect some of the tip users with custom icons, though. They will **briefly** see the icon changed back to the official one when launching new version for the first time.

mitchellh:
I think this is good to go, thank you. I cleaned up some stuff and added some tests for codable/equatable. I will do another review when I get back home and merge later today.

bo2themax:
> I think this is good to go, thank you. I cleaned up some stuff and added some tests for codable/equatable. I will do another review when I get back home and merge later today.

Thanks, adding schema version is great 👍

mitchellh:
Merged, only question I had remaining was if/how this affects code signing. We'll see in the release task and can fix up from there.

mitchellh:
A codesign change was required: https://github.com/ghostty-org/ghostty/commit/6132597563596a12ced386f3760786b62e2dd216

PR #9990, opened 2025-12-21
Key tables apprt action plus macOS UI

### LINKED ISSUE #9963: Key Tables
Add support for key tables, which are named keybinding sets that can be activated or deactivated. 

Ghostty currently supports _key sequences_ which are a different concept. A key sequence requires the proper sequence of keys to be entered in order to invoke a certain action. If an invalid key in the sequence is pressed, the sequence is reset. If a valid sequence is completed, then the sequence is also reset. This is covered in detail in the linked issues: #1420, #2121. A key table is a generalized form of key sequences (key sequences could be implemented using tables, but not the reverse). 

A key table is a named set of keybindings. At any given time, a single key table is _active_. Any keybindings pressed during this time look up their bindings within this table. Tables are arranged as a _stack_ and _activating_ a table pushes it onto the stack. Binding lookup is done from the innermost table outwards, so keybinds in the default table are always available (unless they're explicitly unbound in the inner table).

## Defining a Table

Tables are defined using a new syntax on the existing `keybind` configuration:

```
keybind = foo/
keybind = foo/catch_all=ignore
keybind = foo/arrow_up=adjust_selection:up
keybind = foo/arrow_down=adjust_selection:down
keybind = foo/escape=deactivate_key_table
```

There are many points to this:

- `<name>/<binding>` is the general syntax to define a binding within a table named `<name>`. Names can't contain `/`, and while you can activate multiple tables (push onto a stack), you can't define a nested table, since it has no semantic meaning. `<binding>` is the normal binding syntax fully supported already today.
- `<name>/` (no binding) is the syntax to define _and clear_ a table. This will reset all of its keybinds and other settings.
- `catch_all` is a new special case keybinding key that is our fallthrough case if no other bindings match. So `foo/catch_all=ignore` means that any key not within the table should be ignored (not encoded, consumed). `ignore` is already a supported action since Ghostty 1.0.

## Activating/Deactivating a Table

Tables must be explicitly activated with some other keybinding:

```
keybind = ctrl+a=activate_key_table:foo
keybind = foo/escape=deactivate_key_table
```

`activate_key_table:<name>` pushes the table named `<name>` onto the table stack and makes it the current table. This table remains the activated table until an associated `deactivate_key_table` binding is pressed, which deactivates the current table (not the table it is defined in, but the innermost table).

## One-Shot Tables

A table can be activated in one-shot mode:

```
keybind = ctrl+a=activate_key_table_oneshot:foo
```

A one-shot table will be automatically deactivated when _any non-catch-all binding_ is invoked. 

## Composition

Key tables can be composed with other features (existing and planned) to do more powerful things.

- `keybind = foo/ctrl+a>ctrl+b=new_window`. You can use key sequences within a table. This will make the `ctrl+a` then `ctrl+b` sequence work in the table only. All rules of sequences follow. If you enter an invalid key, the sequence is ended, but you remain within the table.
- `keybind = foo/chain=new_window` (#9961 planned). You can used chained actions to define complex behaviors that are only available within a key table. The chain only applies to the binding in the table.
- `keybind = foo/global:ctrl+a=new_window`. Normal prefixes like `global` and so on work within a named table.

mitchellh:
Fixes #9963 (we'll open new issues to track GTK and other stuff)

This adds the apprt actions necessary for key tables to be shown visually, and adapts the macOS UI to show them.

## Demo

```
keybind = example/
keybind = example/ctrl+a=text:hello
keybind = example/ctrl+b>x=text:wow
keybind = example/ctrl+c=activate_key_table:another
keybind = example/escape=deactivate_key_table
keybind = ctrl+a=activate_key_table:example

keybind = another/
keybind = another/catch_all=deactivate_key_table
```

https://github.com/user-attachments/assets/75e94ec9-b52e-439d-b0ca-229ce533c656

**AI disclosure:** The SwiftUI view was written by AI, everything else was manual.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -691,6 +691,27 @@ typedef struct {
   ghostty_input_trigger_s trigger;
 } ghostty_action_key_sequence_s;
 
+// apprt.action.KeyTable.Tag
+typedef enum {
+  GHOSTTY_KEY_TABLE_ACTIVATE,
+  GHOSTTY_KEY_TABLE_DEACTIVATE,
+  GHOSTTY_KEY_TABLE_DEACTIVATE_ALL,
+} ghostty_action_key_table_tag_e;
+
+// apprt.action.KeyTable.CValue
+typedef union {
+  struct {
+    const char *name;
+    size_t len;
+  } activate;
+} ghostty_action_key_table_u;
+
+// apprt.action.KeyTable.C
+typedef struct {
+  ghostty_action_key_table_tag_e tag;
+  ghostty_action_key_table_u value;
+} ghostty_action_key_table_s;
+
 // apprt.action.ColorKind
 typedef enum {
   GHOSTTY_ACTION_COLOR_KIND_FOREGROUND = -1,
@@ -836,6 +857,7 @@ typedef enum {
   GHOSTTY_ACTION_FLOAT_WINDOW,
   GHOSTTY_ACTION_SECURE_INPUT,
   GHOSTTY_ACTION_KEY_SEQUENCE,
+  GHOSTTY_ACTION_KEY_TABLE,
   GHOSTTY_ACTION_COLOR_CHANGE,
   GHOSTTY_ACTION_RELOAD_CONFIG,
   GHOSTTY_ACTION_CONFIG_CHANGE,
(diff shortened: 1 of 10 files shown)


bo2themax (changes requested):
I pushed a commit to refine some animations and did some cleanup. 

And noticed a bug while I was testing with very long table names:

1. Trigger a keybind to show the capsule.
2. Reload configuration.
3. You can't hide the capsule anymore (`GHOSTTY_ACTION_KEY_TABLE` won't be triggered).

mitchellh:
Thanks, addressed your bug in #9994 since its unrelated to UI.

PR #9997, opened 2025-12-21
shell-integration: ensure shell resources exist

jparise:
Our automatic shell integrations require certain resource paths to exist. If they're missing, the launched shell could end up in an inconsistent and unexpected state.

For example, we temporarily set ZDOTDIR to our zsh shell integration directory and then restore it from our .zshenv file, but if that script isn't available, the user's shell environment will be broken.

The actual runtime logic change was simple: each shell integration routine attempts to open its expected resource path and skips automatic shell integration upon failure. The more complex change was reworking our unit tests to run in a temporary resources directory structure.

See: #9941

### DIFF
--- a/src/termio/shell_integration.zig
+++ b/src/termio/shell_integration.zig
@@ -114,7 +114,7 @@ fn setupShell(
     }
 
     if (std.mem.eql(u8, "elvish", exe)) {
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
The idea is sound. Given the fact this represents a broken install, we should probably do a `log.warn` somewhere when this happens.

PR #10003, opened 2025-12-22
Support native MacOS shortcuts to move to beginning and end of window

yasuf:
Adds support to scroll to the top or bottom of the window by using macOS native shortcuts, testing in video.

**AI disclaimer**: AI used to explore solutions, tested and edited solution proposed by Opus 4.5

https://github.com/user-attachments/assets/333a17e4-b496-4ce4-8650-5b11d5e2b94d

### DIFF
--- a/macos/Sources/Ghostty/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/SurfaceView_AppKit.swift
@@ -9,7 +9,7 @@ extension Ghostty {
     /// The NSView implementation for a terminal surface.
     class SurfaceView: OSView, ObservableObject, Codable, Identifiable {
         typealias ID = UUID
-        
+
         /// Unique ID per surface
         let id: UUID
 
@@ -44,14 +44,14 @@ extension Ghostty {
 
         // The hovered URL string
         @Published var hoverUrl: String? = nil
-        
+
         // The progress report (if any)
         @Published var progressReport: Action.ProgressReport? = nil {
             didSet {
                 // Cancel any existing timer
                 progressReportTimer?.invalidate()
                 progressReportTimer = nil
-                
+
                 // If we have a new progress report, start a timer to remove it after 15 seconds
                 if progressReport != nil {
                     progressReportTimer = Timer.scheduledTimer(withTimeInterval: 15.0, repeats: false) { [weak self] _ in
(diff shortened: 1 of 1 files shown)


yasuf:
the command history doesn't scroll back when I tested so I think we're OK as-is without preventing the event from being sent to Ghostty.

For the page navigation when I do `fn + up` or `fn + down` the command history scrolls, that's the main reason why I didn't include that in this PR, the full `switch` looked like this:

```swift
        // Process MacOS native scroll events
        switch selector {
        case #selector(moveToBeginningOfDocument(_:)):
            _ = surfaceModel.perform(action: "scroll_to_top")
        case #selector(moveToEndOfDocument(_:)):
            _ = surfaceModel.perform(action: "scroll_to_bottom")
		case #selector(scrollPageUp(_:)):
            _ = surfaceModel.perform(action: "scroll_page_up")
		case #selector(scrollPageDown(_:)):
            _ = surfaceModel.perform(action: "scroll_page_down")
        default:
            break
        }
```

yasuf:
This is part of https://github.com/ghostty-org/ghostty/issues/2526

macos/Sources/Ghostty/SurfaceView_AppKit.swift
```diff
@@ -1910,6 +1910,7 @@ extension Ghostty.SurfaceView: NSTextInputClient {
     /// 1. Prevents an audible NSBeep for unimplemented actions.
     /// 2. Allows us to properly encode super+key input events that we don't handle
     override func doCommand(by selector: Selector) {
+		guard let surfaceModel else { return }
```
bo2themax: Guarding should be right before the first use in general to avoid some unknown side effects, although in this case it looks ok

bo2themax (changes requested):
Looks good to me, only a small suggestion

macos/Sources/Ghostty/SurfaceView_AppKit.swift
yasuf: hey thanks for reviewing this change, I saw that it was guarded at the beginning of the function in other functions but I'll move it down just in case to have all the `surfaceModel` logic together

bo2themax (approved):

mitchellh (approved):

PR #10023, opened 2025-12-23
ci: update macOS builds to use Xcode 26.2

mitchellh:
We were fixed on 26.0 previously.


mitchellh:
Going to merge this during the day in case there are any regressions (not expected)

PR #10033, opened 2025-12-23
shell-integration: always call setupFeatures

jparise:
Our existing logic already ensured that setupFeatures() was always called, but that was happening from two code paths: explicitly when shell integration is .none and implicitly via setup().

We can simplify this by always calling setupFeatures() once, outside of the (automatic) shell integration path.

There's one small behavioral change: we previously didn't set up shell features in the automatic shell integration path if we didn't have a resources directory (as a side effect). Resources are required for shell integrations, but we don't need them to export GHOSTTY_SHELL_FEATURES, which could potentially still be useful on its on.

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -750,15 +750,15 @@ const Subprocess = struct {
                     else => "sh",
                 } };
 
+            // Always set up shell features (GHOSTTY_SHELL_FEATURES). These are
+            // used by both automatic and manual shell integrations.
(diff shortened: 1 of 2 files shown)


src/termio/Exec.zig
jparise, line 779: We could also consider pulling this up earlier (above the new setupFeatures() location), which might make overall better sense, too. That would only affect the `.none` path.

mitchellh (approved):

PR #10034, opened 2025-12-23
GTK: UI for key sequences and tables

### LINKED ISSUE #2127: gtk: key tables and sequenced keybind UI indication
#2121 was merged and the core now supports it. As noted in the original issue #1420, there are some UX issues with this, so we should investigate UI improvements to make it clear that you are in a sequenced keybind.

mitchellh:
Fixes #2127 

This adds a UI similar to macOS to show the current state of key sequences and/or key tables.

https://github.com/user-attachments/assets/4399d2af-a88c-4b70-922b-7727dc4d2053

**AI disclosure:** AI was used for various things, but I did write most of the code myself, especially around the memory management of properties since agents can't get that quite right yet. 😄


mitchellh:
Ran this through Valgrind and verified no leaks.

PR #10044, opened 2025-12-24
shell-integration: better shell detection and setup

jparise:
Command-based shell detection has been extracted to its own function (detectShell), which is nicer for testing. It now uses argIterator to determine the command's executable, rather than the previous string operations, which allows us to handle command strings containing quotes and spaces.

Also, our shell-specific setup functions now use a consistent signature, which simplifies the calling code quite a bit.

### DIFF
--- a/src/termio/shell_integration.zig
+++ b/src/termio/shell_integration.zig
@@ -45,95 +45,37 @@ pub fn setup(
     env: *EnvMap,
     force_shell: ?Shell,
 ) !?ShellIntegration {
-    const exe = if (force_shell) |shell| switch (shell) {
-        .bash => "bash",
-        .elvish => "elvish",
-        .fish => "fish",
-        .zsh => "zsh",
-    } else switch (command) {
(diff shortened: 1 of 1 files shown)


jparise:
#10035 will need a small update for this change (or vice-versa). I'll sort that out based on merge order.

mitchellh (changes requested):
Merged #10035 so please request me for re-review when ready. 😄

jparise:
> Merged #10035 so please request me for re-review when ready. 😄

All set, thanks!

jparise:
> > Merged #10035 so please request me for re-review when ready. 😄
> 
> All set, thanks!

@mitchellh everything look good?

mitchellh:
Sorry haven't had a chance to look yet. I'll try to do it this week before anything else causes conflicts. 😄

mitchellh (approved):
Just added type hints to the `const`. I prefer that.

PR #10046, opened 2025-12-24
macOS: move `NSGlassEffectView` into `TerminalViewContainer`

### LINKED ISSUE #9991: macOS: `macos-glass-regular` breaks titlebar & tabs & quick terminal
### Reproduction Steps

1. Ensure system appearance switching is enabled or at least test both Light and Dark appearance.
2. Set and reload config to:

```
theme = light:GitHub Light Default,dark:GitHub Dark Dimmed
background-opacity = 0.85
background-blur = macos-glass-regular
```
3. Open a new window and observe titlebar rendering.
5. Open quick terminal and observe background rendering.

Image

bo2themax:
Fixes #9991, modifying the subview of `NSThemeFrame` seems "dangerous" and unpredictable. 

> [!NOTE]
> AI Proofread some of my comments

https://github.com/user-attachments/assets/c443cf71-8a00-4c37-b008-d89b7f1564a3

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalViewContainer.swift
+++ b/macos/Sources/Features/Terminal/TerminalViewContainer.swift
@@ -1,0 +1,127 @@
+import AppKit
+import SwiftUI
+
+/// Use this container to achieve a glass effect at the window level.
+/// Modifying `NSThemeFrame` can sometimes be unpredictable.
+class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
+    private let terminalView: NSView
+
+    /// Glass effect view for liquid glass background when transparency is enabled
+    private var glassEffectView: NSView?
+    private var derivedConfig: DerivedConfig
+
+    init(ghostty: Ghostty.App, viewModel: ViewModel, delegate: (any TerminalViewDelegate)? = nil) {
+        self.derivedConfig = DerivedConfig(config: ghostty.config)
+        self.terminalView = NSHostingView(rootView: TerminalView(
+            ghostty: ghostty,
+            viewModel: viewModel,
+            delegate: delegate
+        ))
+        super.init(frame: .zero)
+        setup()
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    private func setup() {
+        addSubview(terminalView)
+        terminalView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            terminalView.topAnchor.constraint(equalTo: topAnchor),
+            terminalView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            terminalView.bottomAnchor.constraint(equalTo: bottomAnchor),
+            terminalView.trailingAnchor.constraint(equalTo: trailingAnchor),
+        ])
+
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(ghosttyConfigDidChange(_:)),
+            name: .ghosttyConfigDidChange,
+            object: nil
+        )
+    }
+
+    override func viewDidMoveToWindow() {
+        super.viewDidMoveToWindow()
+        updateGlassEffectIfNeeded()
+    }
+
+    @objc private func ghosttyConfigDidChange(_ notification: Notification) {
+        guard let config = notification.userInfo?[
+            Notification.Name.GhosttyConfigChangeKey
+        ] as? Ghostty.Config else { return }
+        let newValue = DerivedConfig(config: config)
+        guard newValue != derivedConfig else { return }
+        derivedConfig = newValue
+        DispatchQueue.main.async(execute: updateGlassEffectIfNeeded)
+    }
+}
+
+// MARK: Glass
+
+private extension TerminalViewContainer {
+#if compiler(>=6.2)
+    @available(macOS 26.0, *)
+    func addGlassEffectViewIfNeeded() -> NSGlassEffectView? {
+        if let existed = glassEffectView as? NSGlassEffectView {
+            return existed
+        }
+        guard let themeFrameView = window?.contentView?.superview else {
+            return nil
+        }
+        let effectView = NSGlassEffectView()
+        addSubview(effectView, positioned: .below, relativeTo: terminalView)
+        effectView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            effectView.topAnchor.constraint(equalTo: topAnchor, constant: -themeFrameView.safeAreaInsets.top),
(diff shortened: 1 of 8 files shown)


macos/Sources/Features/Terminal/TerminalViewContainer.swift
```diff
@@ -0,0 +1,120 @@
+import AppKit
+import SwiftUI
+
+class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
```
mitchellh, line 6: Let's add some comments here to explain why we need this.

macos/Sources/Features/Terminal/TerminalViewContainer.swift
```diff
@@ -0,0 +1,120 @@
...
+    private let terminalView: NSView
+
+    /// Glass effect view for liquid glass background when transparency is enabled
+    private var glassEffectView: NSView?
+    private var derivedConfig: DerivedConfig {
+        didSet {
```
mitchellh: Instead of a didSet, since this is private and only set in one place let's just do this logic in the config change function.

macos/Sources/Features/Terminal/TerminalViewContainer.swift
```diff
@@ -0,0 +1,120 @@
...
+        derivedConfig = DerivedConfig(config: config)
+    }
+}
+
+// MARK: Glass
+
+private extension TerminalViewContainer {
+    func addGlassEffectIfNeeded() {
+#if compiler(>=6.2)
+        guard #available(macOS 26.0, *), glassEffectView == nil, let themeFrameView = window?.contentView?.superview else {
```
mitchellh: Can we avoid adding any effect view as well if we don't have the glass style configured at all? I realize this works but might be worth just avoiding it altogether and adding it later? (on config reloads)

mitchellh (changes requested):
Awesome thanks, this is pretty clean, some questions.

macos/Sources/Features/Terminal/TerminalViewContainer.swift
bo2themax: Defo can. I thought this way is cleaner, since it behaves kind of like `identity` glass in SwiftUI

mitchellh: Yeah maybe it is. I'll leave it up to you as a judgement call.

bo2themax: Actually, there's some issue if we add this after a glass-style config; the effect only appears as clear. I pushed a test commit; you can try it too by adding and deleting these two lines.

```
background-opacity          = 0.0
background-blur             = macos-glass-regular
```

https://github.com/user-attachments/assets/2dac6a36-3aff-4f91-a340-399c188b5d04

bo2themax: Dropped it and added the reason in the comments

mitchellh (approved):

telemachus:
FYI, after this update on macOS, Ghostty starts and says it is launching windows, but the windows are invisible.

Screenshot 2025-12-25 at 4 49 33 PM

mitchellh:
Yeah, I've gotten multiple reports. I'm reverting now pending more review.

clason:
It's not invisible (for me), just resized to little more than the traffic lights and moved to the bottom left of my screen. Happens on every restart.

bo2themax:
> FYI, after this update on macOS, Ghostty starts and says it is launching windows, but the windows are invisible.
> 
> 
> 
> Screenshot 2025-12-25 at 4 49 33 PM
> 
> 

Thanks for the report, just to be clear, are you on macOS 26 or other versions?

clason:
I'm on macOS 26

bo2themax:
> I'm on macOS 26

Can you also share your config? I couldn't reproduce this with zero config or with glass enabled

clason:
I don't have glass enabled. The relevant part of my config is
```
window-padding-y = 0
window-padding-balance = true
window-width = 120
window-height = 110
window-save-state = always
quit-after-last-window-closed = true
macos-titlebar-style = tabs
macos-titlebar-proxy-icon = hidden
background-opacity = 0.95
background-blur = 20
``

**Comment by bo2themax** [comment]
> window-padding-y = 0
> window-padding-balance = true
> window-width = 120
> window-height = 110
> window-save-state = always
> quit-after-last-window-closed = true
> macos-titlebar-style = tabs
> macos-titlebar-proxy-icon = hidden
> background-opacity = 0.95
> background-blur = 20

Nice, with this I can now, thanks

**Comment by bo2themax** [comment]
The minimum config to reproduce this is

```
window-width = 120
window-height = 110
```

RC is that the new `contentView` didn't provide the correct `intrinsicContentSize`, fixed in #10062

**Comment by telemachus** [comment]
> Thanks for the report, just to be clear, are you on macOS 26 or other versions?

Yes, I am on macOS 26.2. My config is below. (You may not need it any longer, but just in case...)

```config
# The syntax is "key = value". Whitespace doesn't matter.

# Define colors as a sixteen-item palatte. The docs describe pairs (e.g.,
# 0 and 8) as regular and bold. I find it easier to think of them in terms of
# their color names. I.e., 0 is black and 8 is dark grey.
#
# I took many of these colors from Sasha Trubetskoy.
# https://sashamaps.net/docs/resources/20-colors/

# See also https://www.ditig.com/publications/256-colors-cheat-sheet.
#
# black
palette = 0=#000000
# dark red
palette = 1=#800000
# dark green
palette = 2=#194a1f
# dark yellow
palette = 3=#808000
# dark blue
palette = 4=#000075
# dark magenta
palette = 5=#8b008b
# dark cyan
palette = 6=#469990
# grey
palette = 7=#c5c6c8
# dark grey
palette = 8=#737373
# red
palette = 9=#d20f2c
# green
palette = 10=#3cb44d
# yellow
palette = 11=#ffe119
# blue
palette = 12=#4363d8
# magenta
palette = 13=#f032e6
# cyan
palette = 14=#42d4f4
# white
palette = 15=#ffffff

shell-integration-features = no-cursor,title

# Base colors and cursor styles.
background = ffffff
foreground= 090909
cursor-invert-fg-bg = true
cursor-style = block
cursor-style-blink = false

font-family = Consolas
font-size = 16
# macOS needs this.
font-thicken = true
# No ligatures, please.
font-feature = -calt
font-feature = -liga
font-feature = -dlig

# Window padding
window-padding-x = 6
window-padding-y = 6

confirm-close-surface = false

# Enable URL matching
link-url = true

# Right Option = Meta
macos-option-as-alt = right

# Window size
window-height = 30
window-width = 90

# Get nightly updates
auto-update-channel = tip
# auto-update = { check, download, off }
auto-update = download

# Hide the mouse immediately when typing.
mouse-hide-while-typing = false

keybind="ctrl+tab=next_tab"
keybind="ctrl+shift+tab=previous_tab"
```

PR #10064, opened 2025-12-26
feat: key-remap configuration to remap modifiers at the app-level

### LINKED ISSUE #5160: `key-remap` configuration to remap modifiers at the app-level
Discussion: https://github.com/ghostty-org/ghostty/discussions/3493#discussioncomment-11827976

Introduce a new configuration `key-remap` which can be used to remap keys from one to another within the scope of Ghostty. Example:

```
key-remap = ctrl=super
key-remap = left_control=right_alt
```

This would act the same as if a user went to their OS config and remapped these keys. If a sided-modifier isn't specified, then left swaps with left, right swaps with right. 

Notes:

* **This is a one way remap.** In the example above, `ctrl` becomes `super` but `super` does not become `ctrl`. 
* **Remaps are not transitive.** If you remap `ctrl=super` and then `alt=ctrl`, then `alt != super`. Alt would be mapped to control.
* This doesn't interact with `keybind` in any way. In the same way `keybind` is already roughly unaware of keyboard layouts (i.e. `cmd+f` works with `f` as specified by your layout), the same will happen in this case (i.e. `cmd+f` will trigger with a physical "ctrl" key if you remapped `cmd=ctrl`).

## Additional scope limiting

For the initial creation of this feature, we should limit scope in the following ways:

1. Only remap modifier keys
2. Only work on macOS

There is a much more urgent need and request for this on macOS (see #3493) and macOS APIs give us a much easier way to do this by easily letting us insert event filters in the path. My attempts at doing this with GTK have thus far failed, but I think it's possible.

JagjeevanAK:
This PR introduces a new `key-remap` configuration option that allows users to remap modifier keys at the application level without affecting system-wide settings.

## Issue
Closes #5160

## Usage

```ini
# Make Ctrl act as Cmd within Ghostty
key-remap = ctrl=super

# Swap Ctrl and Cmd
key-remap = ctrl=super
key-remap = super=ctrl

# Remap only left Alt to Ctrl
key-remap = left_alt=ctrl
```

### Supported Modifiers

| Generic | Left-sided | Right-sided |
|---------|------------|-------------|
| `ctrl` / `control` | `left_ctrl` | `right_ctrl` |
| `alt` / `opt` / `option` | `left_alt` | `right_alt` |
| `shift` | `left_shift` | `right_shift` |
| `super` / `cmd` / `command` | `left_super` | `right_super` |

## Behavior

Per the issue specification:

- **One-way remapping**: `ctrl=super` means Ctrl becomes Super, but Super remains Super
- **Non-transitive**: `ctrl=super` + `alt=ctrl` → Alt becomes Ctrl (NOT Super)
- **Sided support**: Generic modifiers match both sides; use `left_*` or `right_*` for specific sides
- **Immediate effect**: Changes apply on config reload

## Limitations

- Implemented in Zig core, works on both macOS and Linux
- Only modifier keys can be remapped (not regular keys)


JagjeevanAK:
@mitchellh can you look into this pull request ?

JagjeevanAK:
reason of this pull request is that I don't want to use third party to remap the stuff

JagjeevanAK:
I have tested it by building it locally and it work's like butter 

Here is my config to use it:

```sh
macos-option-as-alt = right
key-remap = right_alt=ctrl
```

chenzhiwei:
@JagjeevanAK , thank you! I really appreciate you implementing this long-awaited feature.

mitchellh (approved):
Thanks for your PR. On balance, this PR was very good. I went ahead and changed your core KeyRemap to a different approach though.

The key change I made is that our remap set maintains a bit mask that allows for faster comparison if an input event is remapped. It's unlikely that it is remapped so this will help with keyboard latency rather than trying to iterate the list for every keyboard input. 

You can take a look.

No AI was used in my edit logic, but I did use AI to assist with test writing and comments.

src/config/Config.zig
```diff
@@ -1757,6 +1758,46 @@ class: ?[:0]const u8 = null,
...
+///   prior to Ghostty receiving them. For example, `option+a` on macOS
+///   may still produce `å` even if `option` is remapped to `ctrl`.
+///
+/// * Generic modifiers (e.g. `ctrl`) match both left and right physical keys.
+///   Use sided names (e.g. `left_ctrl`) to remap only one side. ///
+///
+/// This configuration can be repeated to specify multiple remaps.
+///
+/// Currently only supported on macOS. Linux/GTK support is planned for
+/// a future release.
```
jcollie, line 1798: Is the macOS-only nature of this still correct?

mitchellh: Nope, I had a local commit I forgot to push that modified this and a bunch more comments here. I'll add another PR here in a sec.

jesses-code-adventures:
Apologies for commenting on a merged PR, and please let me know if a subsequent issue is preferred, but I opened the [original issue](https://github.com/ghostty-org/ghostty/discussions/3493#discussioncomment-11827976) and have been keeping an eye on the merged PRs waiting for this one to drop, so I wanted to see if I'm missing how to use this one.

With the below config, when I do the following I'd expect a jump down in nvim if I press `cmd+d`, but instead Ghostty still creates a new split. If I use `super` rather than `cmd`, or `command` rather than `cmd`, this behaviour is consistent. 

```config
auto-update-channel = tip
key-remap = ctrl=cmd
key-remap = cmd=ctrl
```

Ghostty commit: `30a2fb685`

Is there any additional piece of configuration I'm missing to achieve this? Thank you both for your work on this PR, I have been super keen to start using Ghostty!

mitchellh:
> Is there any additional piece of configuration I'm missing to achieve this? Thank you both for your work on this PR, I have been super keen to start using Ghostty!

This is documented as a working as intended side effect of the implementation in the docs. See the docs about how macOS interactions with keybinds (basically: macOS is intercepting those before we get to them).

mitchellh:
For keybinds in particular, we could probably apply this remap to our keybinds to work around this...

jesses-code-adventures:
I was wondering why this behaves differently in iTerm, but it looks like there's an additional setting for remapping modifiers globally that I'd neglected to mention in the original issue which probably explains the difference in behaviour. Not sure if something like this is achievable in Ghostty?

Screenshot 2026-01-09 at 10 24 04 AM

mitchellh:
I think #10230 will address your needs...

jesses-code-adventures:
Thank you @mitchellh, I'll keep following this and may have a crack at a contribution if I can find the time away from work 🙌

PR #10082, opened 2025-12-28
Remove systemd integration from nautilus extension

swsnr:
As far as I understand ghostty already integrates with systemd nowadays, and manages its own scopes/cgroups for the application and all tabs.  As such, explicitly moving launching ghostty into a separate systemd scope from nautilus no longer seems necessary.  In any case the scope created by the nautilus extension quickly goes away, suggesting that Ghostty moves its processes out of this scope, and the process tree looks just the same when spawning ghostty directly.

### DIFF
--- a/dist/linux/ghostty_nautilus.py
+++ b/dist/linux/ghostty_nautilus.py
@@ -17,8 +17,6 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-from os.path import isdir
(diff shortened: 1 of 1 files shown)


tristan957 (approved):
Pinging @jcollie for review

jcollie (approved):

swsnr:
@jcollie @tristan957 Thanks for the review.  I pushed a small last-minute amend which removes the unused `child` variable and the `__init__` method which was now effectively empty save for a parent constructor call.

I hope that's okay :slightly_smiling_face:

jcollie (approved):

PR #10089, opened 2025-12-29
gtk: pass through keypress when clipboard has no text

cyppe:
## Summary

When `paste_from_clipboard` is triggered but the clipboard contains no text (e.g., an image), the action now returns `false` to indicate it couldn't be performed. This enables the `performable:` keybind prefix to work correctly for paste actions.

## Problem

On GTK/Linux, when a user has `keybind = ctrl+v=paste_from_clipboard` and the clipboard contains an image (not text), pressing Ctrl+V does nothing. Applications like `opencode` that handle their own clipboard reading via `wl-paste` never receive the keypress.

## Solution

Make `clipboardRequest` return `bool` to indicate whether the action could be performed. For paste requests on GTK, synchronously check if the clipboard contains text formats before starting the async read. When no text format is available, return `false`.

Users can now use:
```
keybind = performable:ctrl+v=paste_from_clipboard
```

When the clipboard has no text, the keybind is not consumed and Ctrl+V passes through to the terminal application.

## Changes

- `Surface.startClipboardRequest` now returns `bool`
- `paste_from_clipboard` / `paste_from_selection` actions return the result
- GTK apprt checks clipboard formats synchronously before async read
- Embedded apprt always returns `true` (can't check synchronously)

## Testing

1. Add `keybind = performable:ctrl+v=paste_from_clipboard` to config
2. Copy an image to clipboard
3. Open an application that handles image paste (e.g., `opencode`)
4. Press Ctrl+V
5. Image pastes successfully (app receives keypress and handles clipboard itself)

## Disclaimer

Most of the changes is done with Opus 4.5

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -3642,11 +3642,9 @@ const Clipboard = struct {
         // pass through when the clipboard contains non-text content (e.g., images).
         if (state == .paste) {
             const formats = clipboard.getFormats();
-            if (formats.containMimeType("text/plain") == 0 and
-                formats.containMimeType("UTF8_STRING") == 0 and
-                formats.containMimeType("TEXT") == 0 and
-                formats.containMimeType("STRING") == 0)
-            {
+            // G_TYPE_STRING = G_TYPE_MAKE_FUNDAMENTAL(16) = (16 << 2) = 64
+            const G_TYPE_STRING: usize = 64;
+            if (formats.containGtype(G_TYPE_STRING) == 0) {
                 log.debug("clipboard has no text format, not starting paste request", .{});
                 return false;
             }
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -3814,21 +3814,34 @@ const Clipboard = struct {
         const self = req.self;
         defer self.unref();
 
+        const surface = self.private().core_surface orelse return;
+
         var gerr: ?*glib.Error = null;
         const cstr_ = clipboard.readTextFinish(res, &gerr);
+
+        // If clipboard has no text (error, null, or empty), pass through the
+        // original keypress so applications can handle their own clipboard
+        // (e.g., reading images via wl-paste). We send raw Ctrl+V (0x16)
+        // directly using the text action to bypass bracketed paste encoding.
         if (gerr) |err| {
             defer err.free();
-            log.warn(
-                "failed to read clipboard err={s}",
-                .{err.f_message orelse "(no message)"},
-            );
+            log.debug("clipboard has no text format: {s}", .{err.f_message orelse "(no message)"});
+            passthroughKeypress(surface);
             return;
         }
-        const cstr = cstr_ orelse return;
+
+        const cstr = cstr_ orelse {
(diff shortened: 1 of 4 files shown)


jcollie (changes requested):
This should be taking advantage of the `performable` prefix and not relying on sending a fixed `ctrl+v` if there is no text in the buffer.

mitchellh (changes requested):
@jcollie is right. This is the wrong approach.

The correct approach is to return false from our apprt action so our performable infrastructure works. The issue is GDK clipboard reads are async and we require a response immediately. 

The question is if GDK exposes an API to at least ask "are you empty" synchronously. And it may, but not sure if it works as expected, e.g. https://docs.gtk.org/gdk4/method.Clipboard.get_content.html

macOS clipboards are also async, but we're able to ask the empty question synchronously.

mitchellh (commented):
This looks better. I haven't verified if it works yet, and I don't love hunting for specific mime types like that but maybe there is no matter way.

cyppe:
I am no expert, but the final commit is verified to work. Updated PR description too.

Have not run any test suite, so cannot say if something else broke down.

cyppe:
image

Some attempts incoming, will test it.

cyppe:
Seems to work fine too, so pushed it.

src/apprt/gtk/class/surface.zig
```diff
@@ -3623,16 +3623,32 @@ const Clipboard = struct {
...
+        ) orelse return false;
+
+        // For paste requests, check if clipboard has text format available.
+        // This is a synchronous check that allows performable keybinds to
+        // pass through when the clipboard contains non-text content (e.g., images).
+        if (state == .paste) {
+            const formats = clipboard.getFormats();
+            // G_TYPE_STRING = G_TYPE_MAKE_FUNDAMENTAL(16) = (16 << 2) = 64
+            const G_TYPE_STRING: usize = 64;
+            if (formats.containGtype(G_TYPE_STRING) == 0) {
```
mitchellh: We have a typed and friendlier version here we should use: https://github.com/ianprime0509/zig-gobject/blob/27baadaef6a6709b070901e7dae54b0ffcdcc91a/extensions/gobject2.zig#L24

mitchellh (changes requested):
Looks great. One more request.

mitchellh (approved):
Approved pending CI. Thank you.

cyppe:
@mitchellh question - I have friends telling my that on mac they  need to use ctrl+v to paste images to tools like opencode using default keybinds even if all other paste is opt+v.

Tried to unbind opt+v and bind paste_from_clipboard to some random key to see if it can make ghostty trigger native "ctrl+v" style paste on opt+v. But it doesnt.

keybind = opt+v=unbind
keybind = opt+n=paste_from_clipboard

Any idea if there is some feature already in ghostty to get native "ctrl+v" paste to trigger on osx on opt+v?

Otherwise - some pointers to what concept to use if I would attempt a PR, as performable doesnt cover it as it doesnt fall through to native paste on opt+v. 

So it basically needs:

```
keybind = performable:opt+v=paste_from_clipboard + something to trigger native paste when not performable
```

PR #10090, opened 2025-12-29
macOS: Split drag move

mitchellh:
Related to #1525

This implements a _Mouse-only_ approach to moving splits. When you hover near the top of a split, a grab handle now appears that can be used to drag the split into any other split position.

> [!NOTE]
>
> **This PR only lets you move splits to _other split positions_.** I will create follow-up issues to track moving a split out into a new tab or window.

## Demo

https://github.com/user-attachments/assets/fbeeff13-a03c-4e79-b4ba-95537d43c083

## Other Notes

This PR also lays a ton of the groundwork on macOS for surfaces to be draggable and copyable (pasteboard) _in general_. That isn't used yet but there might be some interesting ideas here like pasting screenshots by simply dragging the surface. I don't know!


flq:
This is so cool! I just realize that the last ghostty release is from last October, so to use this I actually have to build from source, or is there some release planned?

pluiedev:
1.3 is scheduled to release in March.

https://ghostty.org/docs/install/release-notes/1-2-0#moving-to-a-marchseptember-release-cycle

htahboub:
Hi @mitchellh, was wondering if there's a built-in way to turn this feature off. I'm not planning on using it but I don't like the look of the grab handle being there when my mouse is on a split. Any way to hide it?

00-kat:
Not presently, open a new discussion if one doesn't already exist.

Edit: #10108.

danielwe:
Also xref #10280, see the last few comments, looks like a feature or vouch request for shrinking the handle's mouse tracking area would be welcome.

PR #10095, opened 2025-12-29
terminal: PageList needs to fix up viewport pin after row change

mitchellh:
From #10074

The test comments explain in detail. I also added a new integrity check to validate this at runtime. All existing tests pass the integrity check.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -394,6 +394,7 @@ pub inline fn pauseIntegrityChecks(self: *PageList, pause: bool) void {
 const IntegrityError = error{
     TotalRowsMismatch,
     ViewportPinOffsetMismatch,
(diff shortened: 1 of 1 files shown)


src/terminal/PageList.zig
```diff
@@ -8284,7 +8306,7 @@ test "PageList resize reflow invalidates viewport offset cache" {
     // Verify scrollbar cache was invalidated during reflow
     try testing.expectEqual(Scrollbar{
         .total = s.total_rows,
-        .offset = 8,
+        .offset = 5,
```
mitchellh, line 8309: Note, the prior test was correct in the sense that if you DON'T fix up the viewport pin it is 8, but that would lead to the viewport having insufficient rows. This is correct with the fixup.

PR #10098, opened 2025-12-29
input: add `end_key_sequence` binding action

mitchellh:
End the currently active key sequence, if any, and flush the keys up to this point to the terminal, excluding the key that triggered this action.

For example: `ctrl+w>escape=end_key_sequence` would encode `ctrl+w` to the terminal and exit the key sequence. Or, more fancy, `ctrl+w>catch_all=end_key_sequence` would exit the sequence without encoding whatever key was invalid!

Normally, an invalid sequence will reset the key sequence and flush all data including the invalid key. This action allows you to flush only the prior keys, which is useful when you want to bind something like a control key (`ctrl+w`) but not send additional inputs.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -5794,6 +5794,14 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
             return try self.deactivateAllKeyTables();
         },
 
+        .end_key_sequence => {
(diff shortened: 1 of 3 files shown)


amadeus:
Was AI used in this? just so I know before starting my review

mitchellh:
> Was AI used in this? just so I know before starting my review

🤣

PR #10103, opened 2025-12-30
Add core functionality to switch to most recent tab with a popup similar to the command palette

yasuf:
## Description

Addresses https://github.com/ghostty-org/ghostty/issues/1844

I tried an approach trying to copy what's in `macos/Sources/Features/Command Palette/CommandPalette.swift` in regards to the key handlers, in this version I used `event.keyCode` to handle keystrokes. I _think_ it's possible to do it with the same approach as in CommandPalette but I was just running into bugs when attempting that.

There is one small-ish bug that if you open the popup and leave the current tab with the popup still open, when you come back to it you still see the tab switcher and I think it should disappear, it doesn't affect the UI/UX in a major way but I definitely think it needs to be fixed, I can tinker with either:
1. Fixing that bug
2. Copying the patterns we use in `CommandPalette.swift` for key handlers.

Happy to address any feedback, I want these code changes to be up to par with the codebase's standards.

**AI Disclaimer**: I used Opus 4.5 to generate the solution and I manually typed in all code changes, tried a few other approaches and removed unnecessary features that were generated by the prompt.

If you want to play around simply checkout this branch, run the project and press CMD + P on MacOS, this was implemented only for MacOS at the moment, we'd still need to mimic this functionality in GTK.
Also you can press CMD + P multiple times and you should see the selector move 1 option down.

https://github.com/user-attachments/assets/a8fb2ba3-ebe7-4a8a-ab04-57ec98d7b11a


philocalyst:
This would be super cool to have :)

PR #10105, opened 2025-12-30
macos: keep glass titlebar inset in sync when tab bar appears

dyxushuai:
Discussion: https://github.com/ghostty-org/ghostty/discussions/10104

Summary:
When background blur uses macOS glass styles, the titlebar becomes fully transparent after opening a new tab. This updates the glass effect view’s top inset during layout so
the glass layer continues to cover the titlebar area.

Root Cause:
The glass effect view’s top constraint is computed once using the theme frame’s safe-area top inset. On macOS 26, opening a new tab changes the titlebar height/safe-area,
but the constraint is never refreshed. That leaves an uncovered strip at the top, which appears fully transparent.

Fix:
Track the glass view’s top constraint and update its constant on layout. This keeps the glass layer aligned with the current safe-area inset while avoiding unnecessary
reconfiguration.

Tests:
- Manual: `zig build run`, open a new tab with `background-opacity < 1` and `background-blur = macos-glass-regular` and confirm titlebar is no longer fully transparent.
- Note: automated tests not added; UI behavior is hard to exercise in existing test suite.

Window position question:
Opening a new tab seems to reset the window position / trigger a maximize-like behavior on my system. Is this intended (feature) or a bug? I did not change this behavior in this PR.

AI Assistance:
This change was implemented with AI assistance (Codex) and reviewed by me.

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalViewContainer.swift
+++ b/macos/Sources/Features/Terminal/TerminalViewContainer.swift
@@ -8,6 +8,7 @@ class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
 
     /// Glass effect view for liquid glass background when transparency is enabled
     private var glassEffectView: NSView?
+    private var glassTopConstraint: NSLayoutConstraint?
     private var derivedConfig: DerivedConfig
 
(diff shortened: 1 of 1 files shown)


bo2themax (approved):
Thanks, looks perfect!

PR #10106, opened 2025-12-30
Refactor nautilus extension

swsnr:
Following #10082 I took the opportunity to clean up the code in the nautilus extension a little.   Specifically, this pull request

- turns methods that weren't using `self` into functions,
- inlines a helper method that was trivial after #10082,
- extracts duplicate code in both menu item callbacks into a new helper function, and
- removes the compatibility hack for the Nautilus 3.0 API, i.e. GNOME 42 and older.

I'm not that sure about the last point, as it's not clear to me what the support baseline is here: GNOME 43 (the first version to support the Nautilus 4.0 API) was released in Sep 2022, i.e. is almost as old as Ghostty itself.  The previous Debian stable release already included GNOME 43, but Ubuntu 22.04 (the LTS before the current 24.04 LTS) still includes GNOME 42.  Does Ghostty support a system that old?  If so, I'll drop the corresponding commit from this PR.

### DIFF
--- a/dist/linux/ghostty_nautilus.py
+++ b/dist/linux/ghostty_nautilus.py
@@ -20,37 +20,39 @@
 from gi.repository import Nautilus, GObject, Gio
 
 
-class OpenInGhosttyAction(GObject.GObject, Nautilus.MenuProvider):
-    def _menu_item_activated(self, _menu, paths):
-        for path in paths:
-            cmd = ['ghostty', f'--working-directory={path}', '--gtk-single-instance=false']
-            Gio.Subprocess.new(cmd, Gio.SubprocessFlags.NONE)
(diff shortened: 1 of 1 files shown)


mitchellh:
> Does Ghostty support a system that old? If so, I'll drop the corresponding commit from this PR.

I think we're safe with Ghostty 1.2/1.3: https://ghostty.org/docs/linux#supported-gtkadwaita-versions

mitchellh (approved):
I appreciate the original author of this script coming and helping clean this up in any way. ❤️ Thank you for your work. This looks good to me.

swsnr:
@mitchellh Thanks for merging, and thank you for your kind words ❤️, and for this great application 🙏

PR #10113, opened 2025-12-30
macOS: Dragging last split out of tab should close tab, not window

mitchellh:
This makes it so that dragging a split over a tab in our app will focus that tab. Presently, it's very hard if not impossible to get the drag to focus the tab. This just does it manually.

This also fixes a bug where the last split drop out of a tab will close the entire window, this affects main currently too.

This is still a draft because I'm chasing down one issue still where the dropped surface fails to render properly and its unclear why yet.

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -952,7 +952,7 @@ class BaseTerminalController: NSWindowController,
             // controller is a TerminalController this is easy because it has a way
             // to do this.
             if let c = sourceController as? TerminalController {
-                c.closeWindowImmediately()
+                c.closeTabImmediately()
             } else {
                 // Not a TerminalController so we always undo into a new window.
                 _ = TerminalController.newWindow(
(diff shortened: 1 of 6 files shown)


bo2themax:
For me, dragging a split over a tab will focus that tab fine, just as dragging a file. The delay is the same; with 3e399a3d3506392f3559f406c1241bc4bc7882fe, the delay will only be slightly shorter though.


https://github.com/user-attachments/assets/a7c172c8-1a8f-47d5-81ef-bb749f4addef

bo2themax:
> For me, dragging a split over a tab will focus that tab fine, just as dragging a file. The delay is the same; with [3e399a3](https://github.com/ghostty-org/ghostty/commit/3e399a3d3506392f3559f406c1241bc4bc7882fe), the delay will only be slightly shorter though.
> 
>  Screen.Recording.2025-12-31.at.09.17.08.mp4

Just noticed that cursor position after dragging is a bit weird.

mitchellh:
Hm interesting! That isn't working for me but let me try again...

mitchellh:
Oh yeah it works for me in released versions, just not in Xcode for some reason. I'll ignore it for now and let that work. I removed it and kept the refactoring + one unrelated bug fix in this PR, renamed the PR, and will merge that.

PR #10116, opened 2025-12-31
macOS: bug fixes for moving splits to different trees

bo2themax:
This pr contains 3 fixes as explained in the commit message. 

The first commit contains only some renaming to distinguish mutating and non-mutating functions, like `sort/sorted` of `Array`.

I didn't figure out good away to detect in `GlobalEventTap.swift` whether a surface is being dragged, so another global variable was created under `Ghostty`, if you have a better way to do this, pleas lemme know.

### DIFF
--- a/macos/Sources/Features/Global Keybinds/GlobalEventTap.swift
+++ b/macos/Sources/Features/Global Keybinds/GlobalEventTap.swift
@@ -128,10 +128,12 @@ fileprivate func cgEventFlagsChangedHandler(
     // We only care about keydown events
     guard type == .keyDown else { return result }
 
-    // If our app is currently active then we don't process the key event.
-    // This is because we already have a local event handler in AppDelegate
+    // If our app is currently active and not dragging surface,
+    // then we don't process the key event.
+    // This is because we already have a valid local event handler in AppDelegate
     // that processes all local events.
-    guard !NSApp.isActive else { return result }
+    // See `https://developer.apple.com/documentation/appkit/nsevent/addlocalmonitorforevents(matching:handler:)#Discussion`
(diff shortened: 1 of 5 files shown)


bo2themax:
I also did some chasing down on that rendering issue mentioned in #10113, but failed to find the cause. The Swift part seems ok, `ghostty_surface_set_focus` was called with the correct value. I also tried some time span between setting different values and different surfaces, which didn't work.

mitchellh:
I cherry picked and merged all except https://github.com/ghostty-org/ghostty/pull/10116/changes/83d78897ef0180cb60aed6a0cefa7cc23cf63c3b

For the last split out of quick terminal I also did something different:
https://github.com/ghostty-org/ghostty/commit/323d362bc18003ef94a1b69ce27b9833573a2aa1

I skipped the one commit because I don't like the global. I see what you're trying to do but I don't love that approach. I want to think about that more.

bo2themax:
> I cherry picked and merged all except https://github.com/ghostty-org/ghostty/pull/10116/changes/83d78897ef0180cb60aed6a0cefa7cc23cf63c3b
> 
> 
> 
> For the last split out of quick terminal I also did something different:
> 
> https://github.com/ghostty-org/ghostty/commit/323d362bc18003ef94a1b69ce27b9833573a2aa1
> 
> 
> 
> I skipped the one commit because I don't like the global. I see what you're trying to do but I don't love that approach. I want to think about that more.

Yeah the global was kinda just a workaround. 

https://github.com/ghostty-org/ghostty/commit/323d362bc18003ef94a1b69ce27b9833573a2a looks ok on the phone, I'll try that out when I get the chance.

PR #10118, opened 2025-12-31
Update Imgui from 1.90.6 to 1.92.5, use Dear Bindings instead of cimgui

mitchellh:
This updates our Imgui version from 1.90.6 to 1.92.5 which is almost 18 months of changes. In the process, we've also migrated from cimgui to Dear Bindings, the official way to do C bindings. 

This PR doesn't contain any functionality changes, only the dependency change and API changes necessary to achieve the same behavior.


mitchellh:
I will add that I did look into Zig bindings for this based on the Dear Bindings JSON file and other available libraries and I didn't feel comfortable enough with what is out in the ecosystem to make a jump to that yet. I'm still open to it, eventually, but wasn't happy with the state of the world.

PR #10120, opened 2026-01-01
apprt/gtk: add progress bar timeout setting

charlesrocket:
Allows the user to set the timeout for idling progress bars.

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1005,12 +1005,16 @@ pub const Surface = extern struct {
         assert(value.state != .remove);
         progress_bar.as(gtk.Widget).setVisible(@intFromBool(true));
 
+        const progress_bar_timeout_ms = if (priv.config) |cfg|
+            cfg.get().@"progress-bar-timeout".asMilliseconds()
+        else
+            15 * std.time.ms_per_s;
+
         // Start our timer to remove bad actor programs that stall
         // the progress bar.
-        const progress_bar_timeout_seconds = 15;
         assert(priv.progress_bar_timer == null);
         priv.progress_bar_timer = glib.timeoutAdd(
-            progress_bar_timeout_seconds * std.time.ms_per_s,
+            progress_bar_timeout_ms,
             progressBarTimer,
             self,
         );
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -824,6 +824,40 @@ palette: Palette = .{},
 /// the mouse is shown again when a new window, tab, or split is created.
 @"mouse-hide-while-typing": bool = false,
 
+/// The time before the idling progress bar is hidden.
+///
+/// The duration is specified as a series of numbers followed by time units.
+/// Whitespace is allowed between numbers and units. Each number and unit will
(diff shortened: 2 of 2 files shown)


src/config/Config.zig
```diff
@@ -1910,6 +1910,9 @@ keybind: Keybinds = .{},
 ///        or rounded corners, use `macos-titlebar-style = hidden` instead.
 @"window-decoration": WindowDecoration = .auto,
 
+/// The time in seconds before the indling progress bar is hidden (GTK).
+@"window-progress-bar-timeout": u32 = 15,
```
jcollie: This should use the `Duration` type for consistency with other time-related config entries.

src/apprt/gtk/class/surface.zig
```diff
@@ -1016,6 +1016,18 @@ pub const Surface = extern struct {
         );
     }
 
+    fn getProgressBarTimeout(self: *Self) c_uint {
```
jcollie: I think that a separate function for this is overkill, especially if the `Duration` type is used as it already handles under/overflow in the `asMilliseconds` method.

src/config/Config.zig
```diff
@@ -1910,6 +1910,9 @@ keybind: Keybinds = .{},
 ///        or rounded corners, use `macos-titlebar-style = hidden` instead.
 @"window-decoration": WindowDecoration = .auto,
 
+/// The time in seconds before the indling progress bar is hidden (GTK).
```
jcollie: Since `indling` hasn't been used since the 1600's perhaps you meant `idling`?

src/config/Config.zig
jcollie: Progress bars are scoped to surfaces, not to windows so I think that this is better named just `progress-bar-timeout`.

jcollie (changes requested):
This should really have gone through the feature request process, especially as you're a fairly new contributor. But this is small and fairly uncontroversial I think. Some general code-style changes and some other minor changes.

src/config/Config.zig
charlesrocket: Oh yes, didn't noticed the typo.

mitchellh:
I'm not totally sure if this should be configurable. I think disabling progress bars entirely makes sense, but not sure if the timeout itself should be. No one else has requested this. Despite its simplicity, I'm not sure...

charlesrocket:
@mitchellh In my case, the hard-coded 15-second delay makes the bar show an outdated state since issuing a consecutive command does not clear it. And if the consecutive command does not ping to the bar, the bar will be showing an invalid state for this command. Another way to fix this is by tracking the command and clearing the bar after it finishes, but this does not add aesthetic flexibility.

mitchellh:
> In my case, the hard-coded 15-second delay makes the bar show an outdated state since issuing a consecutive command does not clear it.

A well behave program should clear progress bars whenever it exists. The delay here is only for poorly behaved programs.

I understand the motivation in the case you can't modify programs, I'm just saying that I'm not sure if this should be configurable, it should just be shorter, or if this is a niche request that should be relegated to a personal fork. I don't know.

PR #10122, opened 2026-01-01
macOS: flash menu bar when using keyboard shortcuts

helsus:
When you press a keyboard shortcut that has a menu equivalent, the menu bar should flash briefly. This is standard macOS behavior.

This change calls `performKeyEquivalent` on the main menu before checking Ghostty bindings, so shortcuts with menu items get handled by the menu system first.

This works for most shortcuts - Cmd+V, Cmd+N, Cmd+T, etc. all flash now.

Won't flash for `performable: true` (like Cmd+C copy) -- excluded from menu: https://github.com/ghostty-org/ghostty/discussions/2811

The change skips menu handling when in a key sequence or key table, so those still work correctly.

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
@@ -1195,7 +1195,15 @@ extension Ghostty {
                 return false
             }
 
-            // If this event as-is would result in a key binding then we send it.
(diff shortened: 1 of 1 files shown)


bo2themax (approved):
Tried it, works well

mitchellh (approved):

mitchellh:
Makes sense, thank you.

PR #9510, opened 2025-11-07
macOS: update new window behaviour on different screen

bo2themax:
Fixes #9310, depends on #9509, **has some conflicts with #10129**

### New Window

- New Window in the dock menu will open a new window on the current focused screen.
- New Window by AppIntent/Shortcuts will open a new window on the current focused screen.
- New Window in the Ghostty's File menu, or by `cmd+n`, will open a new window on the current focused screen.

### New Tab/Split

New tab/split will choose a proper window on the current screen first. If none is found, we still use existing windows on another screen to create it.

> [!NOTE] 
> Using Finder services will still create a new terminal (window/tab) on the first screen Ghostty was focused on. If there are multiple screens that have Ghostty opened on, macOS will select a proper screen to create a new terminal. This for now cannot be changed by Ghostty itself. `Show in Finder` will behave the same.

https://github.com/user-attachments/assets/5fda2165-68ca-4951-ba3c-638f86f9c1aa

Issue #9310, opened 2025-10-22
macOS: new windows/tabs always open on primary monitor, moving existing windows


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/8673

<div type='discussions-op-text'>

<sup>Originally posted by **umputun** September 16, 2025</sup>
### Issue Description

New windows always open on primary monitor regardless of focus. Opening a new tab in a window on non-primary monitor moves the entire window to primary monitor. (macOS 26, Ghostty 1.2.0, multi-monitor setup)

### Expected Behavior

- New windows should open on the monitor that currently has focus or where the cursor is located
- New tabs should open within the existing window without changing its monitor position

### Actual Behavior

When using Ghostty 1.2 with multiple monitors on macOS 26, there are two related window management issues:

- New windows always open on primary monitor: When opening a new Ghostty window (e.g., Cmd+N), it always appears on the primary/default monitor regardless of which monitor currently has focus or where the existing Ghostty window is located.
- New tabs move existing windows: When opening a new tab (Cmd+T) in an existing Ghostty window that's on a non-primary monitor, the entire window gets moved to the primary monitor instead of the tab opening in place.

### Reproduction Steps

1. Set up a multi-monitor configuration on macOS 26
1. Open Ghostty and Move the window to a non-primary monitor
1. Try opening a new window (Cmd+N) - observe it opens on primary monitor
1. With a Ghostty window on a non-primary monitor, open a new tab (Cmd+T) - observe the entire window moves to primary monitor

### Ghostty Version

```text
Ghostty 1.2.0

Version
  - version: 1.2.0
  - channel: stable
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.none
  - font engine   : font.main.Backend.coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

MacOS 26.0 (25A354) Tahoe

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
Seems to be reproducible in any configuration
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
[...]

PR #10124, opened 2026-01-01
Issue #9310 Resolved

ajr-khll:
Issue #9310 reported issues with consistent monitor selection when opening new windows.

I found that the lastCascadePoint in TerminalController.newWindow was always cascading from 0,0 which seems to always be the primary monitor.

I made a small adjustment to the function to compute the parent's cascade point, add the offset and then cascade the new window from that point. In testing this has new windows always open on the same monitor as the currently focused window, rather than always on the primary monitor.

I was unable to replicate the tab-related problem in the original issue.

I used OpenAI's Codex to assist in understanding the codebase, but made edits manually.


bo2themax:
Thanks, but there's already a related fix #9510

PR #10129, opened 2026-01-01
macOS: Split drag to create new tabs

bo2themax:
Implements #10092 and updates cursor shape to indicate the drag action.

https://github.com/user-attachments/assets/53e8900c-f9b7-4ed4-9c5b-875eee1fa249

https://github.com/user-attachments/assets/59f38395-62cc-4a05-9f85-7eb54e8a2f41

We could also create the *second* tab when dragging the surface to the title bar, but Finder only accepts it on the New Tab button too.

> [!NOTE]
> AI proofread my comments

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -726,14 +726,29 @@ class BaseTerminalController: NSWindowController,
         target.highlight()
     }
 
+    /// Verifies two conditions:
+    /// 1) the `surface` exists in the current tree; and
+    /// 2) it can be dragged into a new window or tab.
+    func surfaceShouldBeDraggedOutsideAsNewWindowOrTab(_ surface: Ghostty.SurfaceView) -> Bool {
+        guard surfaceTree.root?.node(view: surface) != nil else {
+            // Not in this tree; this controller isn't responsible.
+            return false
+        }
+        // If our tree isn't split, then we never create a new window, because
+        // it is already a single split.
+        return surfaceTree.isSplit
+    }
+
     @objc private func ghosttySurfaceDragEndedNoTarget(_ notification: Notification) {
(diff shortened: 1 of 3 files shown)


macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -726,14 +726,29 @@ class BaseTerminalController: NSWindowController,
...
+        return surfaceTree.isSplit
+    }
+
     @objc private func ghosttySurfaceDragEndedNoTarget(_ notification: Notification) {
         guard let target = notification.object as? Ghostty.SurfaceView else { return }
         guard let targetNode = surfaceTree.root?.node(view: target) else { return }
-        
-        // If our tree isn't split, then we never create a new window, because
-        // it is already a single split.
-        guard surfaceTree.isSplit else { return }
```
bo2themax, line 735: We don't need to check it here, since `Ghostty.SurfaceDragNoTargetAction` guarantees that the tree is a split.

mitchellh (changes requested):
There's a few structural issues I have with this.

1. I don't want `SurfaceDragSource` aware of our other classes, so a protocol would be a better approach but we should think about what that looks like.

2. The logic in `draggignSession(movedTo)` is all super nasty. There must be a way to clean this up.

I don't have concrete advice to clean all this up without just doing it myself... but I don't love the bit of a mess it is from this.

bo2themax:
> There's a few structural issues I have with this.
> 
> 1. I don't want `SurfaceDragSource` aware of our other classes, so a protocol would be a better approach but we should think about what that looks like.
> 
> 2. The logic in `draggignSession(movedTo)` is all super nasty. There must be a way to clean this up.
> 
> I don't have concrete advice to clean all this up without just doing it myself... but I don't love the bit of a mess it is from this.

Hmmm, I couldn't think of anything more elegant to change the cursor atm. Lemme draft this first.

PR #10132, opened 2026-01-01
feat: select entire URL on double-click

teamchong:
## Why

When double-clicking on a URL like `https://example.com`, only `https` or `//example.com` gets selected because `:` and `/` are word boundaries. Users expect the entire URL to be selected.

## How

Added URL detection to the double-click handler in `Surface.zig`:

1. Before falling back to `selectWord`, try to detect if the clicked position is part of a URL
2. Uses the pre-compiled link regexes from user configuration (same patterns used for cmd+click)
3. If a URL is found at the position, select the entire URL
4. Otherwise, fall back to normal word selection

The implementation:
- Respects user's link configuration (disabled URLs won't trigger selection)
- Reuses pre-compiled regexes from `DerivedConfig.links` (no per-click compilation)
- Follows the same patterns as `linkAtPos`

## What

- `src/Surface.zig`: Added `urlAtPin()` helper function and modified double-click handler
- `src/terminal/StringMap.zig`: Added 2 tests for URL detection following existing test patterns

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4141,9 +4141,18 @@ pub fn mouseButtonCallback(
                 }
             },
 
-            // Double click, select the word under our mouse
+            // Double click, select the word under our mouse.
+            // First try to detect if we're clicking on a URL to select the entire URL.
             2 => {
-                const sel_ = self.io.terminal.screens.active.selectWord(pin.*);
+                const sel_ = sel: {
+                    // Try link detection without requiring modifier keys
+                    const link_result = self.linkAtPin(pin.*, null) catch null;
+                    if (link_result) |result| {
+                        // Only select URLs (links with .open action)
+                        if (result[0] == .open) break :sel result[1];
+                    }
+                    break :sel self.io.terminal.screens.active.selectWord(pin.*);
+                };
                 if (sel_) |sel| {
                     try self.io.terminal.screens.active.select(sel);
                     try self.queueRender();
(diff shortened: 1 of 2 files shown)


src/Surface.zig
```diff
@@ -4403,6 +4405,51 @@ fn linkAtPos(
     return null;
 }
 
+/// Attempt to detect a URL at the given pin position for double-click selection.
+/// Uses configured link patterns to find URLs, respecting user settings.
+/// Unlike linkAtPos, this doesn't require modifier keys.
+///
+/// Requires the renderer state mutex is held.
+fn urlAtPin(self: *Surface, mouse_pin: terminal.Pin) ?terminal.Selection {
```
mitchellh: It feels like this and linkAtPos should somehow be shared/unified. I'm not sure which results in cleaner code, but this feels like pure duplication.

mitchellh (changes requested):
Was this AI? We require AI be disclosed.

I offered some review. I like the general idea but I think the implementation can be cleaner.

src/Surface.zig
teamchong: Done. Extracted `linkAtPin` as the shared core logic. `linkAtPos` now handles pos→pin + OSC8, then delegates to `linkAtPin`. Double-click calls `linkAtPin` directly with null mods.

teamchong:
> Was this AI? We require AI be disclosed.
> 
> I offered some review. I like the general idea but I think the implementation can be cleaner.

Yes, I used Opus to help with this and reviewed/tested the code myself.

Thanks for the feedback. I've refactored to unify the functions as you suggested.

teamchong:
https://github.com/user-attachments/assets/9ceab3e8-9c71-483e-9c17-0172b8e0a47c

mitchellh (approved):
Made a couple tweaks (see my added commit). But looks good now. Thanks.

bobrik:
I believe this broke word selection: #11404. Double clicking on branch name now selects more than expected:

image


image

bobrik:
This restores sanity: https://github.com/bobrik/ghostty/commit/1e75eb964f2b6387ecb829929a260a23c0810b2f.

Issue #10110, opened 2025-12-30
macOS: split handle drags window, not split, when `macos-titlebar-style = hidden`


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/10100

<div type='discussions-op-text'>

<sup>Originally posted by **danielwe** December 29, 2025</sup>
### Issue Description

With `macos-titlebar-style = hidden`, a split drag handle that lies along the top of the window will drag the window instead of the split.

### Expected Behavior

The split handle always drags the split. (A similar mechanism for dragging windows without a titlebar could be useful, but it must be separate.)

### Actual Behavior

The window is dragged and the split layout is unaffected. Edge case: if the window dragged such that it triggers the OS's automatic tiling, the split you tried to drag may pop out into its own window.

https://github.com/user-attachments/assets/3c813c17-cffd-4f73-af87-b948274130e8

### Reproduction Steps

1. Open Ghostty
2. Create a vertical split (Cmd+D)
3. Try dragging a handle to rearrange the splits.

### Ghostty Logs

```text
[Nothing is logged during this]
```

### Ghostty Version

```text
Ghostty 1.3.0-main+b9ad1f05e

Version
  - version: 1.3.0-main+b9ad1f05e
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS 26.2 (Tahoe)

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
macos-titlebar-style = hidden
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

PR #10133, opened 2026-01-02
fix: Mac window becomes unmovable after pane rearrangement

martinemde:
On main, after rearranging panes, the window becomes permanently immovable. Grab handles temporarily set `window.isMovable = false` on hover to prevent window dragging from interfering with pane drags (fixing [#10110](https://github.com/ghostty-org/ghostty/issues/10110)), but the restoration logic failed when views were destroyed during pane rearrangement (which happens each time a pane is rearranged).

The previous approach managed `window.isMovable` state across the view lifecycle:

1. `mouseEntered` → saved and disabled `window.isMovable`
2. View removed during rearrangement → `mouseExited` never fired
3. `deinit` ran with `self.window` already nil → restoration failed
4. Window stuck with `isMovable = false`

Instead of managing window state, prevent the mouseDown event from reaching the window's drag handler by overriding mouse event handling in the grab handle view.

Per [Apple's Event Handling Guide](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingMouseEvents/HandlingMouseEvents.html):

> Custom NSView objects should not invoke super in their implementations of NSResponder mouse-event-handling methods such as mouseDown:, mouseDragged: and mouseUp: unless it is known that the inherited implementation provides some needed functionality.

This eliminates all state management while solving both the original issue (#10110) and the new bug.

AI disclosure: claude code found and wrote the fix. I tested it manually to see that it works. I pressed claude quite hard here to come up with the best fix, and looked at documentation to understand what the fix was doing. It seems like this is a better approach overall to preventing the main window from being dragged when grabbing the Surface Drag handle.

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceDragSource.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceDragSource.swift
@@ -105,22 +105,30 @@ extension Ghostty {
         /// Whether the current drag was cancelled by pressing escape.
         private var dragCancelledByEscape: Bool = false
 
-        /// Original value of `window.isMovable` to restore
-        /// when the mouse exits.
-        private var isWindowMovable: Bool?
-
         deinit {
             if let escapeMonitor {
                 NSEvent.removeMonitor(escapeMonitor)
             }
         }
-        
+
+        override func acceptsFirstMouse(for event: NSEvent?) -> Bool {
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Very good work thank you

PR #10134, opened 2026-01-02
fix(formatter): preserve background colors on cells without text

neurosnap:
The VT formatter was treating cells without text as blank and emitting
them as plain spaces, losing any background color styling. This caused
TUIs like htop to lose their background colors when rehydrating terminal
state (e.g., after detach/reattach in zmx).

For styled formats (VT/HTML), cells with background colors or `style_id`
are now emitted with proper SGR sequences and a space character instead
of being accumulated as unstyled blanks.

Adds handling for `bg_color_palette` and `bg_color_rgb` content tags which
were previously unreachable.

I used amp to construct the test case and the code to make the test pass.  Please see the amp thread.

I tested my branch against the test cases where it was previously broken (nvtop, htop, senpai).

Reference: https://ampcode.com/threads/T-019b7a35-c3f3-73fc-adfa-00bbe9dbda3c

### BEFORE

screenshot_1767373196

### AFTER

screenshot_1767373113

### BEFORE

screenshot_1767373647

### AFTER

screenshot_1767373657

### DIFF
--- a/src/terminal/formatter.zig
+++ b/src/terminal/formatter.zig
@@ -1113,12 +1113,16 @@ pub const PageFormatter = struct {
 
                 // If we have a zero value, then we accumulate a counter. We
                 // only want to turn zero values into spaces if we have a non-zero
-                // char sometime later.
-                if (!cell.hasText()) {
+                // char sometime later. However, for styled formats (VT, HTML), if
+                // the cell has styling (e.g., background color), we must emit it
+                // to preserve the visual appearance.
+                const dominated_by_style = self.opts.emit.styled() and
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Okay, sorry for the delay, I wanted to clean stuff up and now did. The main thing is that I wanted to use Zig's named blocks and early exit to clean up some logic. And also we could extract styling out to above the content tag now because we use it for all content tags. 

This eliminated the `emitColorSgr` function which was a bit redundant and we can now reuse logic.

Tests were good and untouched.

PR #10150, opened 2026-01-03
fix: use flush instead of end on stdout in code generators for Windows compatibility

remorses:
This fixes the Windows build failure discussed in https://github.com/ghostty-org/ghostty/discussions/10148

When building on Windows, `symbols-unigen` and `props-unigen` crash with `error.FileTooBig` because `stdout.end()` calls `setEndPos()` to truncate the output. Windows does not support `SetEndOfFile` on pipes or console handles - it returns `ERROR_INVALID_PARAMETER`, which Zig maps to `error.FileTooBig`.

Using `flush()` instead of `end()` is correct here because:

1. `end()` flushes AND truncates - useful when overwriting files that might have leftover content
2. For stdout captured as a pipe, there's nothing to truncate - we're writing sequentially to a fresh pipe
3. `flush()` ensures all buffered data is sent, which is all that's needed

CI before fix was failing with FileTooBig, after fix builds successfully: https://github.com/remorses/opentui/actions/runs/20671299561/job/59352503875

### DIFF
--- a/src/unicode/props_uucode.zig
+++ b/src/unicode/props_uucode.zig
@@ -88,7 +88,8 @@ pub fn main() !void {
     var stdout = std.fs.File.stdout().writer(&buf);
     try t.writeZig(&stdout.interface);
     // Use flush instead of end because stdout is a pipe when captured by
-    // the build system, and pipes cannot be truncated (especially on Windows).
+    // the build system, and pipes cannot be truncated (Windows returns
+    // INVALID_PARAMETER, Linux returns EINVAL).
     try stdout.interface.flush();
 
     // Uncomment when manually debugging to see our table sizes.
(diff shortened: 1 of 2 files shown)


pluiedev (changes requested):
AI disclosure, please. The issue makes sense and the fix is *okay* but we do require human-written interactions incl. proper disclosure and handwritten PR metadata (title/description)

remorses:
This PR was written primarily by Opus 4.5, investigating why Windows builds for ghostty-vt were failing on Windows for a PR in project https://github.com/sst/opentui

pluiedev (approved):
Thank you. Pending approval from the terminal team

mitchellh (approved):
I'm good with this pending CI.

PR #10155, opened 2026-01-03
gtk: Session Search

### LINKED ISSUE #9948: GTK: Implement "jump to terminal" feature
#9945 but for GTK.

We should follow pretty much the same pattern as macOS, with some subtle details different of course (like, there are no tab colors on GTK). This should be implemented by traversing the top levels, finding all windows, and adding all the surfaces in the split tree.

Nevon:
Gtk implementation of #9945. Fixes #9948.

This adds session search to the command palette on Gtk, allowing you to jump to any surface by title or working directory. The main difference to the Mac OS implementation is that tabs do not have colors by which to search. I also have not implemented the flashing behavior when a split is focused.

The same, or as close as I could make it, behavior that was introduced for command sorting is also implemented for Gtk. Granted, as I haven't tested this new feature on Mac OS, my understanding of the behavior of it is based on the code and the screencast from the PR.

https://github.com/user-attachments/assets/d50d93a8-fe32-4d39-ba41-1f766010a293

One thing I noticed during development, which I left unsolved as I also didn't see it solved in the Mac OS implementation (though I haven't tested it), is that if you are zoomed into a split, then focusing a different split doesn't do anything. There's a configuration option that I forgot the name of, related to zoom behavior during navigation, that I would expect to be respected, but I wasn't able to get it to work, so I left it for a later iteration.

The majority of the code was generated with Claude Sonnet 4.5. Although I have reviewed and iterated on the code thoroughly, I am not experienced with Zig and I would not be surprised if there are issues that I did not notice, and would appreciate them being pointed out (and ideally explained if it's not obvious to a non-Zig developer).

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -1176,36 +1176,6 @@ pub const Application = extern struct {
         return self.private().config.ref();
     }
 
-    /// Collect all surfaces from all windows in the application.
-    /// The caller must unref each surface and window and deinit the list.
-    pub fn collectAllSurfaces(
-        self: *Self,
-        alloc: Allocator,
-    ) !std.ArrayList(Window.SurfaceInfo) {
-        var all_surfaces: std.ArrayList(Window.SurfaceInfo) = .{};
-        errdefer {
-            for (all_surfaces.items) |info| {
-                info.surface.unref();
-                info.window.unref();
-            }
-            all_surfaces.deinit(alloc);
-        }
-
-        const windows = self.as(gtk.Application).getWindows();
-        var it: ?*glib.List = windows;
-        while (it) |node| : (it = node.f_next) {
-            const window_widget = @as(*gtk.Window, @ptrCast(@alignCast(node.f_data)));
-            const window = gobject.ext.cast(Window, window_widget) orelse continue;
-
-            var window_surfaces = try window.collectSurfaces(alloc);
-            defer window_surfaces.deinit(alloc);
-
-            try all_surfaces.appendSlice(alloc, window_surfaces.items);
-        }
-
-        return all_surfaces;
-    }
-
     /// Set the configuration for this application. The reference count
     /// is increased on the new configuration and the old one is
     /// unreferenced.
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -1176,6 +1176,36 @@ pub const Application = extern struct {
         return self.private().config.ref();
     }
 
+    /// Collect all surfaces from all windows in the application.
+    /// The caller must unref each surface and window and deinit the list.
+    pub fn collectAllSurfaces(
+        self: *Self,
+        alloc: Allocator,
+    ) !std.ArrayList(Window.SurfaceInfo) {
+        var all_surfaces: std.ArrayList(Window.SurfaceInfo) = .{};
+        errdefer {
+            for (all_surfaces.items) |info| {
+                info.surface.unref();
+                info.window.unref();
+            }
+            all_surfaces.deinit(alloc);
+        }
+
+        const windows = self.as(gtk.Application).getWindows();
+        var it: ?*glib.List = windows;
+        while (it) |node| : (it = node.f_next) {
+            const window_widget = @as(*gtk.Window, @ptrCast(@alignCast(node.f_data)));
+            const window = gobject.ext.cast(Window, window_widget) orelse continue;
+
+            var window_surfaces = try window.collectSurfaces(alloc);
+            defer window_surfaces.deinit(alloc);
+
+            try all_surfaces.appendSlice(alloc, window_surfaces.items);
+        }
+
+        return all_surfaces;
+    }
+
     /// Set the configuration for this application. The reference count
     /// is increased on the new configuration and the old one is
     /// unreferenced.
--- a/src/apprt/gtk/class/command_palette.zig
+++ b/src/apprt/gtk/class/command_palette.zig
@@ -250,12 +250,10 @@ pub const CommandPalette = extern struct {
         const b_title = b.propGetTitle() orelse return true;
 
         // Compare case-insensitively with colon normalization
-        var i: usize = 0;
-        var j: usize = 0;
-        while (i < a_title.len and j < b_title.len) {
+        for (0..@min(a_title.len, b_title.len)) |i| {
             // Get characters, replacing ':' with '\t'
             const a_char = if (a_title[i] == ':') '\t' else a_title[i];
-            const b_char = if (b_title[j] == ':') '\t' else b_title[j];
+            const b_char = if (b_title[i] == ':') '\t' else b_title[i];
 
             const a_lower = std.ascii.toLower(a_char);
             const b_lower = std.ascii.toLower(b_char);
@@ -263,9 +261,6 @@ pub const CommandPalette = extern struct {
             if (a_lower != b_lower) {
                 return a_lower < b_lower;
             }
-
-            i += 1;
-            j += 1;
         }
 
         // If one title is a prefix of the other, shorter one comes first
@@ -273,23 +268,17 @@ pub const CommandPalette = extern struct {
             return a_title.len < b_title.len;
         }
 
-        // Titles are equal - use sort_key as tie-breaker if both have one
-        const a_priv = a.private();
-        const b_priv = b.private();
-
-        const a_sort_key = switch (a_priv.data) {
-            .regular => 0,
+        // Titles are equal - use sort_key as tie-breaker if both are jump commands
+        const a_sort_key = switch (a.private().data) {
+            .regular => return false,
             .jump => |*ja| ja.sort_key,
         };
-        const b_sort_key = switch (b_priv.data) {
-            .regular => 0,
+        const b_sort_key = switch (b.private().data) {
+            .regular => return false,
             .jump => |*jb| jb.sort_key,
         };
 
-        if (a_sort_key != 0 and b_sort_key != 0) {
-            return a_sort_key < b_sort_key;
-        }
-        return false;
+        return a_sort_key < b_sort_key;
     }
 
     fn close(self: *CommandPalette) void {
--- a/src/apprt/gtk/class/command_palette.zig
+++ b/src/apprt/gtk/class/command_palette.zig
@@ -192,10 +192,14 @@ pub const CommandPalette = extern struct {
             // for GTK.
             if (!isActionSupportedOnGtk(command.action)) continue;
 
-            const cmd = Command.new(config, command);
+            const cmd = Command.new(config, command) catch |err| {
+                log.warn("failed to create command: {}", .{err});
+                continue;
+            };
+            errdefer cmd.unref();
+
             commands.append(alloc, cmd) catch |err| {
                 log.warn("failed to add command to list: {}", .{err});
-                cmd.unref();
                 continue;
             };
         }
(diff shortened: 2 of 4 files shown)


src/apprt/gtk/class/command_palette.zig
```diff
@@ -481,52 +657,116 @@ const Command = extern struct {
...
+
+                var title_val = gobject.ext.Value.new(?[:0]const u8);
+                defer title_val.unset();
+                gobject.Object.getProperty(
+                    j.surface.as(gobject.Object),
+                    "title",
+                    &title_val,
+                );
+                const title = gobject.ext.Value.get(&title_val, ?[:0]const u8) orelse "Untitled";
+
```
jcollie: This is unnecessary as the `Surface` object has a method `getTitle` to get the title without dealing unnecessarily with the GObject properties.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -481,52 +657,116 @@ const Command = extern struct {
...
+                var pwd_val = gobject.ext.Value.new(?[:0]const u8);
+                defer pwd_val.unset();
+                gobject.Object.getProperty(
+                    j.surface.as(gobject.Object),
+                    "pwd",
+                    &pwd_val,
+                );
+                const pwd = gobject.ext.Value.get(&pwd_val, ?[:0]const u8);
+
```
jcollie, line 723: As with the title, each `Surface` object has a `getPwd` method.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -481,52 +657,116 @@ const Command = extern struct {
...
+                var val = gobject.ext.Value.new(?[:0]const u8);
+                defer val.unset();
+                gobject.Object.getProperty(
+                    j.surface.as(gobject.Object),
+                    "title",
+                    &val,
+                );
+                const surface_title = gobject.ext.Value.get(&val, ?[:0]const u8) orelse "Untitled";
+
```
jcollie, line 695: Use the `getTitle` method on the `Surface`.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -413,22 +546,30 @@ const Command = extern struct {
...
+        pub const RegularData = struct {
+            command: input.Command,
+            action: ?[:0]const u8 = null,
+            action_key: ?[:0]const u8 = null,
+        };
 
-        pub var offset: c_int = 0;
+        pub const JumpData = struct {
+            surface: *Surface,
+            window: *Window,
```
jcollie: Storing the window is completely unnecessary.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -413,22 +546,30 @@ const Command = extern struct {
...
-        action_key: ?[:0]const u8 = null,
+        pub const RegularData = struct {
+            command: input.Command,
+            action: ?[:0]const u8 = null,
+            action_key: ?[:0]const u8 = null,
+        };
 
-        pub var offset: c_int = 0;
+        pub const JumpData = struct {
+            surface: *Surface,
```
jcollie: I'm not sure that storing a pointer to the surface is a great idea. I think that this is likely to cause memory issues, either surfaces that can't be cleaned up because an extra reference has been lost or use-after-free issues in corner cases.

This could be a good time to re-introduce the idea of a unique ID for each surface.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -567,3 +833,43 @@ const Command = extern struct {
         }
     };
 };
+
+/// Focus a surface in a window, bringing the window to front and switching
+/// to the appropriate tab if needed.
+fn focusSurface(surface: *Surface, window: *Window) void {
```
jcollie: This function is unnecessary as the core surface has a `presentSurface` method that already handles all of this.

src/apprt/gtk/class/split_tree.zig
```diff
@@ -478,6 +478,12 @@ pub const SplitTree = extern struct {
         return surface;
     }
 
+    /// Sets the last focused surface in the tree. This is used to track
+    /// which surface should be considered "active" for the split tree.
+    pub fn setLastFocusedSurface(self: *Self, surface: ?*Surface) void {
```
jcollie: This is unnecessary as the `last_focused` property is kept up to date via GObject signal handlers.

src/apprt/gtk/class/application.zig
```diff
@@ -1176,6 +1176,36 @@ pub const Application = extern struct {
         return self.private().config.ref();
     }
 
+    /// Collect all surfaces from all windows in the application.
+    /// The caller must unref each surface and window and deinit the list.
+    pub fn collectAllSurfaces(
```
jcollie: This is unnecessary as the core application object already maintains a list of every surface.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -437,7 +578,34 @@ const Command = extern struct {
         });
 
         const priv = self.private();
-        priv.command = command.clone(priv.arena.allocator()) catch null;
+        const cloned = command.clone(priv.arena.allocator()) catch {
+            self.unref();
+            return undefined;
+        };
```
jcollie: Returning `undefined` here is a bug, and not at all equivalent to how the original function worked.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -459,6 +627,14 @@ const Command = extern struct {
             priv.config = null;
         }
 
+        switch (priv.data) {
+            .regular => {},
+            .jump => |*j| {
+                j.surface.unref();
```
jcollie: Yeah, we should definitely not be storing full references to surfaces and windows in the command palette.

jcollie (changes requested):
An interesting start, but there's lots of changes needed before approval. The big issue is storing full references to surfaces in the command palette, but fixing that requires a larger discussion about surface IDs. The other issues can be fixed with a deeper understanding of the available APIs. I'll probably think of more changes as time passes and I'm sure that others will have suggestions as well.

src/apprt/gtk/class/command_palette.zig
Nevon: Great catch. I've changed this to now propagate and handle the error at the callsite.

src/apprt/gtk/class/command_palette.zig
Nevon: You mentioned re-introducing the idea of a unique id for each surface - is there already some prior art on this, or is it just an idea that has been floated at some point? Is the idea to add that to the core Surface struct, and then when the command is executed, basically loop through the surfaces from the core app to find the one with the matching id?

src/apprt/gtk/class/command_palette.zig
Nevon: Got rid of the unnecessary window reference, now that I know about `presentSurface` . Will need to figure out how to look up the surface when the command is executed to get rid of the surface reference. I'll mark this one as resolved to not spread that discussion over multiple threads.

Nevon:
Thank you so much for the thoughtful review, @jcollie! I really appreciate it.

src/apprt/gtk/class/command_palette.zig
jcollie: #8200

Nevon: What about using a WeakRef rather than directly referencing the surface? Wouldn't that solve the same issues? Pushed [f5d7108](https://github.com/ghostty-org/ghostty/pull/10155/commits/f5d7108c51240724e71ed85de06ee64d8cba014f) for reference, but I'm happy to walk it back if you think we should pursue an alternative option.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -146,32 +149,147 @@ pub const CommandPalette = extern struct {
...
+                return a_lower < b_lower;
             }
 
-            const cmd = Command.new(config, command);
-            const cmd_ref = cmd.as(gobject.Object);
-            priv.source.append(cmd_ref);
-            cmd_ref.unref();
+            i += 1;
+            j += 1;
+        }
```
jcollie, line 264: Since `i` and `j` are never different I think having separate variables is not needed. I fact, you can even simplify this to:
```
for (0..@min(a_title.len, b_title.len)) |i| {
```

src/apprt/gtk/class/command_palette.zig
```diff
@@ -146,32 +149,147 @@ pub const CommandPalette = extern struct {
...
+        if (a_title.len != b_title.len) {
+            return a_title.len < b_title.len;
+        }
+
+        // Titles are equal - use sort_key as tie-breaker if both have one
+        const a_priv = a.private();
+        const b_priv = b.private();
+
+        const a_sort_key = switch (a_priv.data) {
+            .regular => 0,
```
jcollie: Just return `false`.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -146,32 +149,147 @@ pub const CommandPalette = extern struct {
...
+            .jump => |*ja| ja.sort_key,
+        };
+        const b_sort_key = switch (b_priv.data) {
+            .regular => 0,
```
jcollie: Just return `false`.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -146,32 +149,147 @@ pub const CommandPalette = extern struct {
...
+        }
+
+        // If one title is a prefix of the other, shorter one comes first
+        if (a_title.len != b_title.len) {
+            return a_title.len < b_title.len;
+        }
+
+        // Titles are equal - use sort_key as tie-breaker if both have one
+        const a_priv = a.private();
+        const b_priv = b.private();
```
jcollie: I don't think that the temporary vars here are doing much so could be gotten rid of.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -146,32 +149,147 @@ pub const CommandPalette = extern struct {
...
+        const a_sort_key = switch (a_priv.data) {
+            .regular => 0,
+            .jump => |*ja| ja.sort_key,
+        };
+        const b_sort_key = switch (b_priv.data) {
+            .regular => 0,
+            .jump => |*jb| jb.sort_key,
+        };
+
+        if (a_sort_key != 0 and b_sort_key != 0) {
```
jcollie: If you return early above instead of using `0` as a sentinel this `if` is unnecessary.

src/apprt/gtk/class/command_palette.zig
```diff
@@ -413,31 +541,62 @@ const Command = extern struct {
...
+    pub fn newJump(config: *Config, surface: *Surface) *Self {
         const self = gobject.ext.newInstance(Self, .{
             .config = config,
         });
 
         const priv = self.private();
-        priv.command = command.clone(priv.arena.allocator()) catch null;
+        priv.data = .{
+            .jump = .{
+                .sort_key = @intFromPtr(surface),
```
jcollie, line 586: I realize this is used for sort stability, but messing around with pointers is always kinda icky. We may want to add a TODO to switch to using a surface's UUID whenever Ghostty adds one.

jcollie (changes requested):
A few stylistic changes, then I'll be satisfied. Would be good to get some feedback from other GTK maintainers. Valgrind seemed to be happy, although it looks as there's more GTK suppressions needed. WeakRef seems the best way to handle surfaces at this time. I verified that closing a surface out from under the command palette results in no ill effects.

src/apprt/gtk/class/command_palette.zig
Nevon: Added in [0c8b51c](https://github.com/ghostty-org/ghostty/pull/10155/commits/0c8b51c7ab95633a0c5c5f6e6397136736960d18)

jcollie (approved):
LGTM

mitchellh (approved):
Thanks. There are some simple stylistic things to improve here but I think overall I'm fine with this and we can clean it up later. Thank you for implementing it in the simple way first, and we can optimize later. :)

PR #10158, opened 2026-01-04
feat: added rounded corners to quick terminal window

frolvanya:
This PR updates the appearance and behavior of the `QuickTerminalWindow` on macOS. This was discussed in https://github.com/ghostty-org/ghostty/discussions/8666

The main change is to retain the macOS window styling (including rounded corners) while hiding the titlebar and window controls, so the window appears borderless but still benefits from native system visuals

image

P.S: Only after opening PR I saw that this is a sort of a duplicate of https://github.com/ghostty-org/ghostty/pull/9960. Feel free to close my PR if it's better to go with a configurable corner radius instead of setting rounded corners for all quick term windows


bo2themax:
Thanks for linking the related pr, this is clean, but I prefer the configurable one, and GTK can benefit from the same config in the future, so I'm closing this for now.

PR #10160, opened 2026-01-04
split_tree: goto_split spatial wrap

### LINKED ISSUE #8406: macOS: `goto_split` spatial directions should wrap
Discussion: #6826

All our `previous/next` bindings across tabs and splits wrap. I think it makes sense for the spatial directions to wrap as well. 

To achieve this, the wrapping should still be spatially aware. The best way to achieve this is conceptually duplicate the grid in the direction you're going, and find the next plausible target. For example, if you're at the top and do a `goto_split:up`, it should act as though you're below the bottom in the same x/y and go up. This will give the most expected behavior, I believe.

qingyunha:
Fixes #8406

### DIFF
--- a/src/datastruct/split_tree.zig
+++ b/src/datastruct/split_tree.zig
@@ -407,17 +407,50 @@ pub fn SplitTree(comptime V: type) type {
                     .split => continue,
                 }
 
-                // Ensure it is in the proper direction
-                if (!switch (direction) {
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
I think we should clean up the Zig here to prevent `var` usage but otherwise this looks good to me. Thank you.

mitchellh:
Closed in favor of #10811

PR #10178, opened 2026-01-05
fix: remove progress bar on reset terminal

### LINKED ISSUE #10168: 'Reset' should clear the progress bar, if any.
### Discussed in https://github.com/ghostty-org/ghostty/discussions/10165

<sup>Originally posted by **PeterCardenas** January  4, 2026</sup>
When I ctrl-c a zig build, which uses the progress bar, it doesn't remove the progress bar. This might be a bug on zig's part but it would also be nice to be able to force remove the progress bar when it is in a bad state, preferably via command palette as I make good use of it.

I also thought the `Reset Terminal` would clear this, but it does not.</div>

PeterCardenas:
closes #10168

ai disclosure: i fully used claude code to generate this pr, but change seems straightforward enough and have manually tested it.

### DIFF
--- a/src/termio/stream_handler.zig
+++ b/src/termio/stream_handler.zig
@@ -952,6 +952,9 @@ pub const StreamHandler = struct {
 
         // Reset resets our palette so we report it for mode 2031.
         self.surfaceMessageWriter(.{ .report_color_scheme = false });
+
+        // Clear the progress bar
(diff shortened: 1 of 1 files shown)


jcollie (approved):
LGTM

mitchellh (approved):
I think

PR #10179, opened 2026-01-05
font/shaper: Fix CoreText position.y for some scripts

jacobsandlund:
This PR simplifies and corrects the logic for placing a glyph vertically, by using the `position.y` from `CoreText` directly, instead of using an offset from the cell's starting `y`. The logic was incorrect from the beginning, always treating the first glyph of a cell as being at `y` of zero. We only need to be subtracting the cell's starting `x` to align the glyphs to the cell grid.

Enabling the commented out logging, I found no instances of `position.y differs from old offset.y` lines with `JetBrains Mono` with ligatures turned on, but running [ttylang](https://github.com/jacobsandlund/ttylang) (printing the Universal Declaration of Human Rights in various languages) revealed 676 instances of this, with many only slightly off.

An example log from some Tai Tham text is the following, and this PR adds a test based on this:

```
...pos=(0.00,-8.21) run_offset=(69.41,-8.21) cell_offset=(69.41,-8.21) old offset.y=0.00 cps = \u{1a49}\u{1a60} \u{1a3f}▸\u{1a69} \u{1a2f} → ᩉ᩠ᨿᩩᨯ
```

Browsers display this as:

ᩉ᩠ᨿᩩ

`main` is printing:

CleanShot 2026-01-05 at 10 28 17@2x

this PR prints:

CleanShot 2026-01-05 at 10 29 07@2x

Since this is a ligature of two different grapheme clusters, Ghostty ends up subtracting too much of the `x` value with the `cell_offset.x` (starting x), so neither of the screenshots above are correct, but the second is closer and gets the `y` value right.

AI disclaimer: I didn't use AI for the code, but did ask it about this Tai Tham text and why it wasn't a single grapheme cluster: https://ampcode.com/threads/T-019b8ea2-1822-75bb-a8eb-55a9ddb9f7ea


mitchellh (approved):
Good pending CI

mitchellh:
Thanks again for the diligence in following up with this <3

PR #10184, opened 2026-01-05
Lower unimplemented OSC from warning to debug

jcollie:
The "unimplemented OSC command" warning is a frequent source of red herrings as there are many OSCs that we parse but deliberately do not implement. It's also a small drag on performance due to the function call overhead and the overhead of formatting and outputting the message. This PR lowers the message from warning to debug. Debug logs are compiled out of release builds so there should be less confusion from general users who will no longer see the message.


mitchellh:
This was warning early on because it was a useful roadmap for what to implement to make the terminal functional, but we are very very long past that point and nowadays its mostly stuff we're purposely not implementing so this makes sense.

PR #10189, opened 2026-01-06
Add terminfo as downloadable release artifact

brittbinler:
## Summary

This PR makes the Ghostty terminfo source file available as a downloadable artifact in releases. Users who SSH into remote servers often encounter rendering issues because the server lacks Ghostty's terminfo definition. Currently, obtaining the terminfo requires cloning the entire repository, which is cumbersome. With this change, users can download the terminfo file directly and compile it with tic.

Changes:
- Generate ghostty.terminfo during tagged and nightly release builds
- Upload terminfo to R2 for tagged releases and to GitHub Release assets for nightly builds
- Add terminfo validation to publish-tag workflow to catch missing files before publishing
- Enable terminfo generation in flatpak, nix, and snap package builds
- Document the terminfo flag and download location in PACKAGING.md

Closes #9917

## Test Plan

Local verification completed on macOS:

- YAML syntax validation passed for all modified workflow and package files
- Terminfo generation tested with both minimal and full builds (with Metal toolchain)
- Generated terminfo file validated for correct header and capabilities (Tc, Su, colors)
- Terminfo source successfully compiled with tic to produce valid database
- Nix package build tested via Docker to confirm flag integration in zigBuildFlags

CI/CD verification will occur post-merge:
- Tagged release: ghostty.terminfo uploaded to R2 at /{VERSION}/ghostty.terminfo
- Nightly release: ghostty.terminfo appears in GitHub Release assets

## AI Disclosure

This PR was developed with assistance from Claude Code. All changes were reviewed, tested locally, and verified by the submitter.

### DIFF
--- a/.github/workflows/publish-tag.yml
+++ b/.github/workflows/publish-tag.yml
@@ -51,6 +51,7 @@ jobs:
           curl -I -s -o /dev/null -w "%{http_code}" "${BASE}/ghostty-macos-universal.zip" | grep -q "^200$" || exit 1
           curl -I -s -o /dev/null -w "%{http_code}" "${BASE}/ghostty-macos-universal-dsym.zip" | grep -q "^200$" || exit 1
           curl -I -s -o /dev/null -w "%{http_code}" "${BASE}/Ghostty.dmg" | grep -q "^200$" || exit 1
+          curl -I -s -o /dev/null -w "%{http_code}" "${BASE}/ghostty.terminfo" | grep -q "^200$" || exit 1
           curl -I -s -o /dev/null -w "%{http_code}" "${BASE}/appcast-staged.xml" | grep -q "^200$" || exit 1
 
       - name: Download Staged Appcast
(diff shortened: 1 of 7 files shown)


qingyunha:
https://github.com/ghostty-org/ghostty/blob/c5bc6bb2ce079fedf9e222a02939d3e0707a67f4/src/build/Config.zig#L367

Should we consider making this true by default?

mitchellh (changes requested):
There's a lot of details and duplicated work here that don't seem right because `-Demit-terminfo` also forces the full executable build. We shouldn't just slap this everywhere. Going to close this. Thanks for trying.

qingyunha:
> There's a lot of details and duplicated work here that don't seem right because `-Demit-terminfo` also forces the full executable build. We shouldn't just slap this everywhere. Going to close this. Thanks for trying.

How about this?
```
$ zig run -lc src/main_build_data.zig -- +terminfo | head -10
xterm-ghostty|ghostty|Ghostty,
        am,
        bce,
        ccc,
        hs,
        km,
        mc5i,
        mir,
        msgr,
        npc,
```

PR #10193, opened 2026-01-06
font: add bitmap font tests for BDF, PCF, and OTB formats

brittbinler:
Adds test coverage for bitmap font rendering in the FreeType backend using the Spleen 8x16 font in three formats:

- BDF (Bitmap Distribution Format)
- PCF (Portable Compiled Format)
- OTB (OpenType Bitmap)

Tests validate glyph dimensions and pixel-perfect rendering against expected patterns, following the model established by the existing TerminusTTF test.

Spleen was chosen because it ships in all three required formats from a single source, providing consistent test data across formats. Its BSD 2-Clause license is compatible with the existing font licenses in the repo (OFL, MIT). No existing embedded fonts could be used since they are all TTF/OTF files that use the TrueType loader rather than the BDF/PCF/OTB loaders being tested.

Address Freetype side of #8524.

---

This PR was written with assistance from Claude Code.


mitchellh (approved):
Thanks, this looks good to me pending CI.

PR #10197, opened 2026-01-06
macOS: add Cmd+J "Jump to Selection" menu item and default binding

mitchellh:
This matches other built-in macOS apps like Terminal, Notes, Safari. We already had the binding, just needed to create the menu.


mitchellh:
https://ampcode.com/threads/T-019b956a-f4e6-71b4-87fa-4162258d33ff

PR #10200, opened 2026-01-07
Fix expanding tilde (~) in the working-directory field

### LINKED ISSUE #3328: File paths in config file should expand tilde (`~`) to HOME

arnodirlam:
Fixes #3328.

## Problem

The `working-directory` config option doesn't expand tilde (`~`) paths like `~/dev` to the user's home directory. When users set `working-directory = ~/dev` in their config file, ghostty starts at `/` instead of the expected directory.

## Root Cause

In [`src/config/Config.zig`](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L1430), `working-directory` is defined as `?[]const u8` (line [1430](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L1430)), not as a `Path` type. The `expandPaths` function ([lines 4099-4130](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L4099-L4130)) only processes fields of type `Path`, `?Path`, `RepeatablePath`, or `?RepeatablePath`. Since `working-directory` is `?[]const u8`, it's skipped during path expansion. The tilde expansion logic already exists in [`src/config/path.zig`](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/path.zig) (`Path.expand` method, [lines 120-220](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/path.zig#L120-L220), with tilde expansion at [lines 148-182](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/path.zig#L148-L182)) and works correctly for other path fields, but is never called for `working-directory`.

## Solution

Add special handling in `expandPaths` to manually expand tilde paths for `working-directory` before processing other path fields. This ensures:

1. Special values (`"home"`, `"inherit"`) are preserved
2. Tilde paths (`~/...`) are expanded to absolute paths
3. Already absolute paths remain unchanged
4. Errors are properly logged to diagnostics

## Testing

Tested successfully on the following system:

- **OS**: macOS 26.2 (Build 25C56)
- **Architecture**: arm64 (Apple Silicon)
- **Zig Version**: 0.15.2 (managed via `mise`)
- **Shell**: zsh

The following config values were tested:

- `working-directory = ~` - Expands to `"home"`, starts in user's home directory
- `working-directory = ~/dev` - Expands to `/Users/arno/dev`, starts in the specified subdirectory
- `working-directory = home` - Preserved as-is, starts in user's home directory
- `working-directory = /absolute/path` - Preserved as-is, starts in the absolute path
- `working-directory = ~/nonexistent` - Error logged to diagnostics, path expansion fails gracefully

Not tested:

- `working-directory = inherit` - not sure how to test this

## Potential follow-up

Refactor `working-directory` from `?[]const u8` to a tagged union type (similar to [`Command`](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/command.zig)) with variants for `home`, `inherit`, and `path`, addressing the TODO at [line 4382](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L4382) and enabling automatic tilde expansion through the existing `expandPaths` mechanism. This eliminates runtime string comparisons and special-case handling while providing type safety.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4105,6 +4105,43 @@ fn expandPaths(self: *Config, base: []const u8) !void {
         .{ .expand = try arena_alloc.dupe(u8, base) },
     );
 
+    // Special handling for working-directory: expand tilde if present
+    // working-directory is ?[]const u8, not Path, so it needs manual expansion
+    if (self.@"working-directory") |wd| {
+        // Skip special values that shouldn't be expanded
+        if (!std.mem.eql(u8, wd, "home") and !std.mem.eql(u8, wd, "inherit")) {
+            // Treat bare "~" as equivalent to "home"
+            if (std.mem.eql(u8, wd, "~")) {
+                self.@"working-directory" = "home";
+            }
+            // Check if it starts with ~/ and expand it
+            else if (std.mem.startsWith(u8, wd, "~/")) expand: {
+                // Windows isn't supported yet
+                if (comptime builtin.os.tag == .windows) break :expand;
+
+                var buf: [std.fs.max_path_bytes]u8 = undefined;
+                const expanded = internal_os.expandHome(wd, &buf) catch |err| {
+                    try self._diagnostics.append(arena_alloc, .{
+                        .message = try std.fmt.allocPrintSentinel(
+                            arena_alloc,
+                            "error expanding home directory for working-directory {s}: {}",
+                            .{ wd, err },
+                            0,
+                        ),
+                    });
+                    break :expand;
+                };
+
+                log.debug(
+                    "expanding working-directory from home directory: path={s}",
+                    .{expanded},
+                );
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Special casing by field is not the right long term maintainable solution

PR #10205, opened 2026-01-07
Refactor `working-directory` & Fix expanding tilde (~)

### LINKED ISSUE #3328: File paths in config file should expand tilde (`~`) to HOME

arnodirlam:
Fixes #3328.

## Problem

The `working-directory` config option doesn't expand tilde (`~`) paths like `~/dev` to the user's home directory. When users set `working-directory = ~/dev` in their config file, ghostty starts at `/` instead of the expected directory.

## Root Cause

In [`src/config/Config.zig`](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L1430), `working-directory` is defined as `?[]const u8` (line [1430](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L1430)), not as a `Path` type. The `expandPaths` function ([lines 4099-4130](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L4099-L4130)) only processes fields of type `Path`, `?Path`, `RepeatablePath`, or `?RepeatablePath`. Since `working-directory` is `?[]const u8`, it's skipped during path expansion. The tilde expansion logic already exists in [`src/config/path.zig`](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/path.zig) (`Path.expand` method, [lines 120-220](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/path.zig#L120-L220), with tilde expansion at [lines 148-182](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/path.zig#L148-L182)) and works correctly for other path fields, but is never called for `working-directory`.

## Solution

Refactor `working-directory` from `?[]const u8` to a tagged union type (similar to [`Command`](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/command.zig)) with variants for `home`, `inherit`, and `path`, addressing the TODO at [line 4382](https://github.com/ghostty-org/ghostty/blob/a4368064a34956c713462c2fccaa3abb41158af2/src/config/Config.zig#L4382) and enabling automatic tilde expansion through the existing `expandPaths` mechanism. This eliminates runtime string comparisons and special-case handling while providing type safety.

## Testing

Tested successfully on the following system:

- **OS**: macOS 26.2 (Build 25C56)
- **Architecture**: arm64 (Apple Silicon)
- **Zig Version**: 0.15.2 (managed via `mise`)
- **Shell**: zsh

The following config values were tested:

- `working-directory = ~` - Expands to `"home"`, starts in user's home directory
- `working-directory = ~/dev` - Expands to `/Users/arno/dev`, starts in the specified subdirectory
- `working-directory = home` - Preserved as-is, starts in user's home directory
- `working-directory = /absolute/path` - Preserved as-is, starts in the absolute path
- `working-directory = ~/nonexistent` - Error logged to diagnostics, path expansion fails gracefully

Not tested:

- `working-directory = inherit` - not sure how to test this

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -628,7 +628,7 @@ pub fn init(
             .env_override = config.env,
             .shell_integration = config.@"shell-integration",
             .shell_integration_features = config.@"shell-integration-features",
-            .working_directory = config.@"working-directory",
+            .working_directory = if (config.@"working-directory") |wd| wd.resolved() else null,
             .resources_dir = global_state.resources_dir.host(),
             .term = config.term,
 
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -512,7 +512,7 @@ pub const Surface = struct {
                     break :wd;
                 }
 
-                config.@"working-directory" = wd;
+                config.@"working-directory" = .{ .path = try config.arenaAlloc().dupeZ(u8, wd) };
             }
         }
 
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -3217,7 +3217,7 @@ pub const Surface = extern struct {
 
         // Properties that can impact surface init
         if (priv.font_size_request) |size| config.@"font-size" = size.points;
-        if (priv.pwd) |pwd| config.@"working-directory" = pwd;
+        if (priv.pwd) |pwd| config.@"working-directory" = .{ .path = pwd };
 
         // Initialize the surface
         surface.init(
--- a/src/apprt/surface.zig
+++ b/src/apprt/surface.zig
@@ -193,7 +193,7 @@ pub fn newConfig(
     if (prev) |p| {
         if (shouldInheritWorkingDirectory(context, config)) {
             if (try p.pwd(alloc)) |pwd| {
-                copy.@"working-directory" = pwd;
+                copy.@"working-directory" = .{ .path = try alloc.dupeZ(u8, pwd) };
             }
         }
     }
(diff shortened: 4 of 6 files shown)


mitchellh (changes requested):
I respect that you probably aren't familiar with Zig. I respect that you took my vague feedback before and tried to do something different. It is directionally better. But, it's still filled with a dozen or more papercuts and a couple major issues. It's not my job or obligation to give detailed feedback on a PR created by a robot, I have zero interest in going back and forth with a robot.

I like AI. I use AI everyday, and we accept PRs from AI everyday. But it feels like you're just throwing shit at a wall and hoping it's "good enough." If someone is using AI I have a higher bar of acceptance because I'm not going to review it, it has to be perfect the first time because the driver should be an expert. I'm getting the sense you're not an expert.

My constructive advice: step back from the AI, think critically about what is a good design here, understand the idioms of Zig and the codebase, and feel free to try again. If you do try again, please explain AS A HUMAN what you discovered.

arnodirlam:
Thank you for your direct and nuanced feedback, @mitchellh, and taking the time to write it! You're absolutely right that I'm not an expert and tried to use AI to fill the gap.

I didn't mean to contribute for the sake of contributing, but to help improving a tool that I'm a big fan of and to improve it in a way that I'd like to use, and that several others have requested as well.

I'll take your advice as a personal call to rethink how and when I use AI in my work. And might take another stab at it without AI. Thank you 🙏

Issue #10227, opened 2026-01-08
Scrollbar and search can access tmux history


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/10218

<div type='discussions-op-text'>

<sup>Originally posted by **danielwe** January  7, 2026</sup>
### Issue Description

When using tmux, ghostty's search includes results from tmux's off-screen history, and dragging ghostty's scrollbar brings this history into view. This issue was implicit in #10078 (which was to some extent a feature request to disable this behavior). Converting it to a proper issue triage here as I think this is simply a bug.

### Expected Behavior

Tmux history/scrollback should be invisible to Ghostty, as tmux is an alternate screen app that manages its own scrollback, and the alternate screen has a hardcoded `scrollback-limit = 0`.

### Actual Behavior

Off-screen tmux history can be accessed via Ghostty operations like search and scroll.

Note that scroll events from a scroll wheel or trackpad are passed to tmux and work as expected. You need to click and drag the scrollbar to engage Ghostty's own scrolling and observe the issue.

If you drag the scrollbar half a screen or so, and then use the scroll wheel, you'll see tmux scrolling the lower half of the screen while the top half remains static.

My guess is that when tmux's own terminal is on the primary screen, tmux doesn't fully erase and redraw the screen the way a typical alt screen app such as a text editor would. Thus, content that has flowed off-screen remains in Ghostty's preallocated PageList memory and therefore appears in search results. Somehow, a scrollbar also appears and can be dragged to resurface this content (I thought the `scrollback-limit = 0` codepath would block a scrollbar from appearing at all, but apparently not).

The video shows:
1. Scrolling tmux history by dragging Ghostty's scrollbar. It's clear that this is not tmux-native scrolling, because tmux's statusline disappears from the screen.
2. Parking Ghostty's scrollbar such that the bottom ~half of the screen overlaps with tmux's active area, and using the scroll wheel to engage tmux-native scrolling, which then only affects the bottom ~half.
3. Using Ghostty search to jump to results from outside tmux's active area.

https://github.com/user-attachments/assets/c0118eee-5ef9-4fe6-a0a1-c77d13cf25b7

### Reproduction Steps

0. On macOS, if you don't have an external mouse connected, enable System settings -> Appearance -> Show scroll bars -> Always, to make sure the scrollbar is available for click-and-drag.
1. Launch a tmux session inside Ghostty
2. Dump enough content to create some scrollback
[...]

PR #10229, opened 2026-01-08
terminal: PageList shouldn't allow any scrolling with max_size=0

mitchellh:
Partial #10227

This fixes the scrollbar part of #10227, but not the search part.

The way PageList works is that max_size is advisory: we always allocate on page boundaries so we always have _some_ extra space (usually, unless you ask for a byte-perfect max size). Normally this is fine, it doesn't cause any real issues. And this has been true since Ghostty 1.0.

But with the introduction of scrollbars (and search), we were exposing this hidden space to the user. To fix this, the easiest approach is to special-case the zero-scrollback scenario, since it is already documented that scrollback limit is not _exact_ and is subject to some minimum allocations. But with zero-scrollback we really expect NOTHING.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -2000,6 +2000,12 @@ pub const Scroll = union(enum) {
 pub fn scroll(self: *PageList, behavior: Scroll) void {
     defer self.assertIntegrity();
 
+    // Special case no-scrollback mode to never allow scrolling.
+    if (self.explicit_max_size == 0) {
+        self.viewport = .active;
+        return;
+    }
+
     switch (behavior) {
         .active => self.viewport = .active,
         .top => self.viewport = .top,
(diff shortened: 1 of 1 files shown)


mitchellh:
We actually had some tests asserting the bad behavior, since I blanket added scrollbars to every test case when I did scrollbar work. It's not as simple as simply removing them since the behavior they're trying to test (garbage pins) requires rearchitecting the test a bit. The core logic here is still right, but we gotta fix up the tests.

danielwe:
If I'm not mistaken, handling this directly in PageList means that the `no_scrollback` flag in Screen.zig is no longer needed. As far as I understand, it's an attempt to address the same situation, but the way it works via `cursorDownScroll` clearly doesn't catch the tmux case.

https://github.com/ghostty-org/ghostty/blob/a692cb9e5fabfd337827cc99cd62e3ea90ab9c92/src/terminal/Screen.zig#L262

mitchellh:
I’ll check. I can’t fully remember how it’s used. It might also be a performance purpose in screen not just correctness, but I’ll have to double check.

mitchellh:
Yeah I actually think `no_scrollback` is still preferred to exist due to optimizations (see the latest search PR notes). We can fast-path some stuff if we know we can avoid scroll back pushing at a higher level than when it hits PageList. 

Though, we may be able to remove the field and just check `pages.explicit_max_size == 0` (or add an inline helper function). We added `explicit_max_size` sometime later for other important reasons so I'm guessing `no_scrollback` is outdated with regards to that more than anything.

PR #10232, opened 2026-01-09
terminal/search: screen search prunes history for no-scrollback screens

### LINKED ISSUE #10227: Scrollbar and search can access tmux history

mitchellh:
Fixes #10227

The big comment in `search/screen.zig` describes the solution well. The problem is that our search is discrete by page and a page can contain some amount of history as well. 

For zero-scrollback screens, we need to fully prune any history lines. For everyone else, everything in the PageList is scrollable and visible so we should search it.

### DIFF
--- a/src/terminal/search/screen.zig
+++ b/src/terminal/search/screen.zig
@@ -412,6 +412,12 @@ pub const ScreenSearch = struct {
             // pages then we need to re-search the pages and add it to
             // our history results.
 
+            // If our screen has no scrollback then we have no history.
+            if (self.screen.no_scrollback) {
+                assert(self.history == null);
+                break :history;
+            }
+
             const history_: ?*HistorySearch = if (self.history) |*h| state: {
                 // If our start pin became garbage, it means we pruned all
                 // the way up through it, so we have no history anymore.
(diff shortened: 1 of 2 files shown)


src/terminal/search/screen.zig
```diff
@@ -1333,3 +1374,37 @@ test "select prev with history" {
...
+
+    var s = t.vtStream();
+    defer s.deinit();
+
+    // This will probably stop working at some point and we'll have
+    // no way to test it using public APIs, but at the time of writing
+    // this test, CSI 22 J (scroll complete) pushes into scrollback
+    // with alt screen.
+    try s.nextSlice("Fizz\r\n");
+    try s.nextSlice("\x1b[22J");
```
mitchellh, line 1398: Funny enough this is the only sequence I can find that bleeds data into our scrollback in the no-scrollback case. So its worth considering if we want to special case this with `no_scrollback` to save some work...

After that we technically have no way to test this in the search package but we should keep this anyway in case we regress something.

PR #10234, opened 2026-01-09
nix: clean up flake

pluiedev:
- Don't expose package attributes that won't build (e.g. on macOS)

- Make the flake generally easier to read since there's none of that `builtin.foldl' recursiveUpdate` nonsense anymore

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -49,92 +49,100 @@
     zon2nix,
     home-manager,
     ...
-  }:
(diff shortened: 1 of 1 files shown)


mitchellh:
You got some build issues. We use the package deps on macOS to setup environments. For the rest I'll defer to @jcollie.

mitchellh:
Sorry @pluiedev we've got some conflicts here.

PR #10241, opened 2026-01-09
snap: fix handling of nonexistent last_revision file

bboozzoo:
Assuming /bin/sh is symlinked to bash, the handling of special builtin
'source' is slightly different between bash and bash-in-POSIX-mode (as a
result of being invoked through /bin/sh). Specifically errors in builtin
'source' cannot be masked with `|| true`. Compare

```
$ ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Dec 11 11:00 /bin/sh -> bash
$ /bin/sh -c 'set -e ; source nofile || true; echo ok'
/bin/sh: line 1: source: nofile: file not found
$ /bin/bash -c 'set -e ; source nofile || true; echo ok'
/bin/bash: line 1: nofile: No such file or directory
ok
```

Thus ghostty from snap would not start at all when
$SNAP_USER_DATA/.last_revision does not exist causign the launcher
script to exit prematurely.

### DIFF
--- a/snap/local/launcher
+++ b/snap/local/launcher
@@ -14,7 +14,14 @@ if [ -z "$XDG_DATA_HOME" ]; then
   export XDG_DATA_HOME="$SNAP_REAL_HOME/.local/share"
 fi
 
-source "$SNAP_USER_DATA/.last_revision" 2>/dev/null || true
+if [ -f "$SNAP_USER_DATA/.last_revision" ]; then
+  if ! source "$SNAP_USER_DATA/.last_revision" 2>/dev/null; then
+    # file exist but sourcing it fails, so it's likely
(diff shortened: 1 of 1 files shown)


bboozzoo:
cc @kenvandine

kenvandine (approved):
Probably a corner case, but good catch!

bboozzoo:
> Probably a corner case, but good catch!

This prevents new installations of ghostty from starting on Arch, Fedora and openSUSE. Those are the systems I tried but potentially any other system which defaults to bash as /bin/sh. Once this lands, any chance you could perhaps cherry pick the fix to 1.2.3 which is currently published in stable channel?

PR #10242, opened 2026-01-09
docs: bell border feature is available on macOS

muellermartin:
As of commit fe55d90 and PR #8768 this feature is also available on macOS.


src/config/Config.zig
```diff
@@ -2934,8 +2934,6 @@ keybind: Keybinds = .{},
 ///    Display a border around the alerted surface until the terminal is
 ///    re-focused or interacted with (such as on keyboard input).
```
mitchellh, line 2935: I think we should add a note that this is available on macOS since 1.2.1

mitchellh (changes requested):

mitchellh (approved):

PR #10251, opened 2026-01-10
Fix memory leak when pruning scrollback with non-standard pages

mitchellh:
This _finally_ resolves #9962 (and the myriad of dupes). 

The core issue was that when a non-standard size page is reused as part of our scrollback pruning, it was resized to be standard size, which caused our future frees to believe it was pooled and not call `munmap` properly. 

The solution I chose was to never reuse non-standard sized pages. If during scrollback pruning we detect a non-standard page, we destroy it and re-alloc. This frees the old memory and reuses pool memory for the new page. 

As part of this, I also introduced a custom page allocator that uses macOS's mach kernel virtual memory tagging feature to specifically tag our PageList memory. I was able to use this in conjunction with Instruments and `footprint` to verify that our PageList memory was previously not freed and is now successfully freed. 

**No AI was used in my work here.** AI was used by others in their analysis of this issue that I used the results of to help guide me and give me other things to consider, but the ultimate understanding and fix was all done via my own meat sticks.

## Detailed Explainer

Ghostty uses a memory pool of fixed-size `mmap`-ed pages to serve as the backing memory for our terminal. If the terminal requires a non-standard (larger) amount of memory due to an abundance of emoji, styles, hyperlinks, etc. then we allocate non-pooled pages directly with `mmap`. When freeing our pages, if it is `<= standard size` we just put it back into the memory pool. If it is larger, we `munmap` it. 

This explains and defines both a _standard_ and therefore _non-standard_ page.

Ghostty also has the concept of a "scrollback limit" (exposed to the user via the `scrollback-limit` config). This caps the size of our scrollback or history. When we reach this limit, we have a trick that we do: to avoid allocation, we reuse the oldest page in the history.

Unfortunately, as part of this process, we were resizing the underlying memory to be standard size again. This was causing our future frees to believe this was pooled memory, and never unmap it. This was the main source of the leak.

## For SEO

For search-ability, I will note that this particular case was often caused by:

- Claude Code (Not their fault, just something in the format of their output often triggered non-standard pages)
- Emoji-heavy output
- Hyperlink-heavy output (like `exa` or `eza`)

## Thanks

Big shout out to @grishy for being the person that finally got me a reproduction so I could analyze the issue for myself. His own analysis got to the same conclusion as me but thanks to the reproduction I was able to verify both our understandings independently.

### DIFF
--- a/src/os/mach.zig
+++ b/src/os/mach.zig
@@ -1,0 +1,35 @@
+const std = @import("std");
+
+/// macOS virtual memory tags for use with mach_vm_map/mach_vm_allocate.
+/// These identify memory regions in tools like vmmap and Instruments.
+pub const VMTag = enum(u8) {
+    application_specific_1 = 240,
+    application_specific_2 = 241,
+    application_specific_3 = 242,
+    application_specific_4 = 243,
+    application_specific_5 = 244,
+    application_specific_6 = 245,
+    application_specific_7 = 246,
+    application_specific_8 = 247,
+    application_specific_9 = 248,
+    application_specific_10 = 249,
+    application_specific_11 = 250,
+    application_specific_12 = 251,
+    application_specific_13 = 252,
+    application_specific_14 = 253,
+    application_specific_15 = 254,
+    application_specific_16 = 255,
+
+    // We ignore the rest because we never realistic set them.
+    _,
+
+    /// Converts the tag to the format expected by mach_vm_map/mach_vm_allocate.
+    /// Equivalent to C macro: VM_MAKE_TAG(tag)
+    pub fn make(self: VMTag) i32 {
+        return @bitCast(@as(u32, @intFromEnum(self)) << 24);
+    }
+};
+
+test "VMTag.make" {
+    try std.testing.expectEqual(@as(i32, @bitCast(@as(u32, 240) << 24)), VMTag.application_specific_1.make());
+}
--- a/src/os/mach.zig
+++ b/src/os/mach.zig
@@ -1,4 +1,7 @@
 const std = @import("std");
+const assert = @import("../quirks.zig").inlineAssert;
+const mem = std.mem;
+const Allocator = std.mem.Allocator;
 
 /// macOS virtual memory tags for use with mach_vm_map/mach_vm_allocate.
 /// These identify memory regions in tools like vmmap and Instruments.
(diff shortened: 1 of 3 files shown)


chrislloyd:
Great minds, you beat us by 45 mins! We literally [just got some debug logs](https://x.com/bengweeks/status/2009663662240186573) and Claude came up with a similar fix. Thank you!

grishy:
@mitchellh can you recheck my assumption here.
today, I had a crash on this commit after this changes (release build from tip) related to `terminal.Screen.cursorScrollAbove`, which I got from a minidump file.

**Flow `PageList.grow()` prune, non-standard page**
Here we destroy `first` and exit, so `tracked_pins` not updated. 

```zig
// ...
if (first.data.memory.len > std_size) {
    self.destroyNode(first);
    break :prune; // here exit before tracked_pins remap
}

/// Later in the code...
const pin_keys = self.tracked_pins.keys();
for (pin_keys) |p| {
    if (p.node != first) continue;
    p.node = self.pages.first.?;
    p.y = 0;
    p.x = 0;
    p.garbage = true;
}
// ...
```

Code for go-to:
https://github.com/ghostty-org/ghostty/blob/509f0733667d7631266f787ea800ba05d3672cec/src/terminal/PageList.zig#L2545-L2580

mitchellh:
That does look like a bug on first glance!

grishy:
@mitchellh 
And as we started with pin - `resizeWithoutReflowGrowCols`, I think this is was my case today with crash on resize 😄 

```zig
if (prev) |prev_node| prev: {
    // ...
    for (dst_rows, src_rows) |*dst_row, *src_row| {
        prev_page.cloneRowFrom(...); // move page to prev_page
    }
    // and no pin update from page here
}

// Remove the old page.
// Deallocate the old page.
self.pages.remove(chunk.node);
self.destroyNode(chunk.node);  // pin still here a suppose
// pin dangle
```

https://github.com/ghostty-org/ghostty/blob/509f0733667d7631266f787ea800ba05d3672cec/src/terminal/PageList.zig#L1828-L1861

upd. added link and formatting

mitchellh:
I think I got this in the linked PR (see GH), or did I miss something?

ianjdarrow:
37 GB? Rookie numbers :) thanks for the fix!
IMG_5106

PR #10260, opened 2026-01-11
gtk: add split title labels showing terminal title per split pane

BenGWeeks:
> **Note:** I'm not familiar with the codebase, so please review the architecture of the solution.

## Summary

Adds split labels (title bars) to terminal split panes in GTK. When a window has multiple splits, each split displays a title bar at the top showing the terminal title. This helps users identify different splits when working with multiple terminals.

Closes #1541
Related to #3747

## Changes

- Added `GhosttySurface` property `is-split` to track when surface is in a split
- Added split title bar using layout-based approach (Box wrapper) instead of overlay to avoid obscuring terminal content  
- Title bar shows terminal title (from OSC 0/2 sequences or config default)
- Uses system accent color (`@accent_color`) for title text
- Added `split-title-bar` config option (defaults to `false`) to enable/disable the feature
- Proper accessibility attributes to prevent title bar from capturing input events

## Config Option

The new `split-title-bar` option allows users to enable the feature:

```
# Enable split title bars (default: false)
split-title-bar = true
```

## Implementation Details

**Architecture**: Uses a Box wrapper (`surface_container`) in `1.2/surface.blp` containing:
1. A Revealer with the title bar Label (only revealed when in split AND config enabled)
2. An Adw.Bin containing the terminal page

This layout-based approach pushes terminal content down rather than overlaying on top, ensuring no content is obscured.

**Config-aware visibility**: Uses a closure function `closureShouldSplitTitleBeShown` that checks both `is-split` property and the `split-title-bar` config option.

## Test Plan

- [x] Create splits using keybindings or context menu
- [x] Verify title bar appears when split (with config enabled)
- [x] Verify title updates with `echo -e "\033]0;Test Title\007"`
- [x] Verify title bar doesn't appear for single (non-split) terminals
- [x] Verify `split-title-bar = true` enables the feature
- [x] Verify title bar doesn't capture mouse/keyboard events

## Screenshots

Title bars showing per-split terminals with accent-colored text:

image

---

DISCLAIMER:

This PR was implemented with assistance from Claude Code (Claude Opus 4.5). All code has been reviewed, tested, and understood by the author.


jcollie:
Just a quick note, despite it being in a directory called `1.2` the `surface.blp` uses features newer than Adwaita 1.2. So there's no need for a new copy of it in a `1.5` directory.

rhodes-b:
> Removed previous hover toolbar implementation in favor of simpler always-visible title label

This should be configurable to not show at all. A tooltip makes it much less obtrusive than a titlebar which is why I don't think the original issue had that note.
I think it would be easy to argue that quite a few people don't want this due to the extra space its takes up, myself being one of them

BenGWeeks:
Thanks for the note! I've moved the changes back to `1.2/surface.blp` and removed the unnecessary `1.5/surface.blp` file in commit 3af71062.

jcollie:
This should definitely default to off. Ghostty's been around long enough that we don't want to change the default behavior like that. If this is accepted and has been around a while and we get enough feedback we can switch the default to on.

Also, if this is accepted, I think that it's natural that people might ask for controls like closing a split by clicking a close button, or zooming/unzooming by double-clicking the title bar. Don't do it in this PR but I wonder how hard it would be to use an Adwiata HeaderBar instead of a GTK Label to display the title.

BenGWeeks:
> This should definitely default to off. Ghostty's been around long enough that we don't want to change the default behavior like that. If this is accepted and has been around a while and we get enough feedback we can switch the default to on.
> 
> Also, if this is accepted, I think that it's natural that people might ask for controls like closing a split by clicking a close button, or zooming/unzooming by double-clicking the title bar. Don't do it in this PR but I wonder how hard it would be to use an Adwiata HeaderBar instead of a GTK Label to display the title.

You're right, the default should be off. Also fixed an issue when the toolbar is off the title bar was still being shown (just without the text), so fixed that. I agree on comments  about zoom etc. I did add close icon through to be consistent with the tabs (and because I'm always forgetting the commands**!).

** I wonder whether a good approach is to show the commands when you mouseover options like close tab and close split to also show `CTRL+D` etc. so you are reminded for the next time. Not done elsewhere, so not implemented for the closing of splits.

According to Claude on the Adwiata.HeaderBar (which seems fair enough to me):

```
  AdwHeaderBar is designed for window-level title bars:
  - Integrates with window decorations (CSD - Client Side Decorations)
  - Has window controls (minimize, maximize, close) that don't make sense for splits
  - Expects to be at the top of a window, not nested inside content
  - Heavier visual weight than needed for internal panes

  What we have now (Box + Label + Button) is more appropriate:
  - Lightweight
  - Fits the "internal pane header" pattern used in other GNOME apps
  - Only shows what's needed (title + close)
  - Easier to style consistently

  Alternatives worth considering:
  - AdwBanner - for notifications/info bars (not quite right either)
  - AdwActionRow - for list-style headers (different use case)
  - Custom styled Box - what we're doing now, most flexible

  The current approach follows how apps like GNOME Builder, Nautilus, and other GTK4 apps handle internal pane headers - simple styled containers rather than full HeaderBars.
```

rhodes-b:
> I wonder whether a good approach is to show the commands when you mouseover options like close tab and close split to also show CTRL+D etc. so you are reminded for the next time

ctrl+d is a shell thing not a terminal one so no that should not show up

mitchellh:
The linked issue is macOS only. I'd still like to discuss how this will work for GTK. In particular, note the churn that macOS is seeing with the split movement handles, because that's relevant. Some research required. Thank you though.

PR #10262, opened 2026-01-11
GTK: start search accepts selection contents

rhodes-b:
Fixes: #10196

If user invokes `search_selection` the contents of that string are no longer ignored.

I'm not very familar with how apprt actions are sent to macos to ensure the defer free will work there, but it is required for GTK otherwise we are leaking the memory

I'm also open to making the string optional instead of checking against `""` thats just how the current `start_search` action was sending its needle (which I assume is better for macos?)

The text is also highlighted currently due to `grabFocus` always selecting the search contents, not sure how the macos part works but that could easily be changed to change the contents after focus was grabbed
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/search_overlay.zig#L233-L241

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -5221,6 +5221,7 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
 
         .search_selection => {
             const selection = try self.selectionString(self.alloc) orelse return false;
+            defer self.alloc.free(selection);
             return try self.rt_app.performAction(
                 .{ .surface = self },
                 .start_search,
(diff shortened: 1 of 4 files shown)


src/Surface.zig
```diff
@@ -5221,6 +5221,7 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
 
         .search_selection => {
             const selection = try self.selectionString(self.alloc) orelse return false;
+            defer self.alloc.free(selection);
```
mitchellh, line 5224: Oh was this a small memory leak we had on this action? It looks like it. 

This solution is right. The memory invoked via `performAction` (synchronous) is only guaranteed for the duration of the function call. apprts must copy any memory they want to retain longer.

mitchellh (approved):

PR #10270, opened 2026-01-11
terminal: fix memory leak when adjustCapacity replaces standard pages

kevincorvallis:
This fixes a memory leak where standard-sized pages (592K) accumulate in the memory pool without being freed during `adjustCapacity()` operations.

## Background

PR #10251 fixed non-standard page leaks during scrollback pruning, but a similar leak existed in the `adjustCapacity()` path. When a standard page is expanded to non-standard size (e.g., when doubling `grapheme_bytes` from 8192 to 16384), the old standard page was returned to the pool via `pool.pages.destroy()`. However, the pool uses an `ArenaAllocator` that never frees individual items, causing these pages to accumulate indefinitely.

## Reproduction

This is particularly common with emoji-heavy and hyperlink-rich output (like Claude Code), where `adjustCapacity()` is called frequently to expand page capacity. Over time, this results in hundreds of thousands of leaked 592K `VM_ALLOCATE` regions.

After ~9.5 hours with Claude Code:
- 154GB physical footprint
- - 272,399 VM_ALLOCATE regions
- - **254,742 regions are exactly 592K** (uniform size, all swapped out)
## The Fix

When replacing a standard page with a non-standard page during `adjustCapacity()`, munmap the old page directly instead of returning it to the pool. This follows the same pattern as PR #10251 but applies to the `adjustCapacity()` code path.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -2695,7 +2695,33 @@ pub fn adjustCapacity(
     // Insert this page and destroy the old page
     self.pages.insertBefore(node, new_node);
     self.pages.remove(node);
-    self.destroyNode(node);
+
+    // If we're replacing a standard-size page with a non-standard page,
+    // we need to munmap the old page instead of returning it to the pool.
+    // This prevents a memory leak where pooled pages accumulate without
+    // being reused (since future allocations will be non-standard).
+    const old_size = page.memory.len;
+    const new_size = new_page.memory.len;
+    if (old_size <= std_size and new_size > std_size) {
+        // Munmap the standard page directly, similar to how we handle
+        // non-standard pages in destroyNodeExt().
+
+        // Update accounting
+        self.page_size -= page.memory.len;
+
+        // Free the memory directly with the page allocator instead of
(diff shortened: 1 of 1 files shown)


mitchellh:
This is so dumb it has to be AI.

My favorite is how it memsets the memory to zero before freeing it.

Absolute stupidity lol.

weshatheleopard:
> My favorite is how it memsets the memory to zero before freeing it.

Ummm, from what I heard, it's a (relatively) new standard security practice, so somebody who does `malloc()` (not `calloc()`) doesn't end up with some other process' data.

jcollie:
> > My favorite is how it memsets the memory to zero before freeing it.
> 
> Ummm, from what I heard, it's a (relatively) new standard security practice, so somebody who does `malloc()` (not `calloc()`) doesn't end up with some other process' data.

Zeroing out memory to prevent data leaks is a practice likely much older than you've been alive. It's also _incredibly_ slow so it can't be tolerated in critical performance paths like Ghostty's rendering.

And that's just the first thing that was wrong with this PR.

PR #10274, opened 2026-01-11
shell-integration: initial nushell shell integration

jparise:
Nushell <https://www.nushell.sh/> is a modern interactive shell that provides many shell features out-of-the-box, like `title` support. Our shell integration therefore focuses on Ghostty-specific features like `sudo`.

We use Nushell's module system to provide a `ghostty` module containing our shell integration features. This module is automatically loaded from $XDG_DATA_DIRS/nushell/vendor/autoload/ when `nushell` shell integration is enabled.

Exported module functions need to be explicitly "used" before they're available to the interactive shell environment. We do that automatically by adding `--execute "use ghostty *"` to the `nu` command line.

This imports all available functions, and individual shell features are runtime-guarded by the script code (using $GHOSTTY_SHELL_FEATURES). We can consider further refining this later.

When automatic shell integration is disabled, users can still manually source and enable the shell integration module:

    source $GHOSTTY_RESOURCES_DIR/shell-integration/nushell/vendor/autoload/ghostty.nu
    use ghostty *

This initial work implements our TERMINFO-aware `sudo` wrapper (via the `sudo` shell feature). Support for additional features, like `ssh-env` and `ssh-terminfo`, will follow (#9604).

### DIFF
--- a/src/termio/shell_integration.zig
+++ b/src/termio/shell_integration.zig
@@ -14,6 +14,7 @@ pub const Shell = enum {
     bash,
     elvish,
     fish,
+    nushell,
     zsh,
 };
 
@@ -57,7 +58,7 @@ pub fn setup(
             env,
         ),
 
-        .elvish, .fish => try setupXdgDataDirs(
+        .nushell => try setupNushell(
             alloc_arena,
             command,
             resource_dir,
@@ -70,6 +71,11 @@ pub fn setup(
             resource_dir,
             env,
         ),
+
+        .elvish, .fish => xdg: {
+            if (!try setupXdgDataDirs(alloc_arena, resource_dir, env)) return null;
+            break :xdg try command.clone(alloc_arena);
+        },
     } orelse return null;
 
     return .{
@@ -153,6 +159,7 @@ fn detectShell(alloc: Allocator, command: config.Command) !?Shell {
 
     if (std.mem.eql(u8, "elvish", exe)) return .elvish;
     if (std.mem.eql(u8, "fish", exe)) return .fish;
+    if (std.mem.eql(u8, "nu", exe)) return .nushell;
     if (std.mem.eql(u8, "zsh", exe)) return .zsh;
 
     return null;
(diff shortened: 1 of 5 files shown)


src/termio/shell_integration.zig
```diff
@@ -714,7 +716,146 @@ test "xdg: missing resources" {
...
+    // would require complex or unsupported integration behavior, we bail out
+    // and skip loading our shell integration. Users can still manually source
+    // the shell integration module.
+    //
+    // Unsupported options:
+    //  -c / --command      -c is always non-interactive
+    //  --lsp               --lsp starts the language server
+    while (iter.next()) |arg| {
+        if (std.mem.eql(u8, arg, "--command") or std.mem.eql(u8, arg, "--lsp")) {
+            return null;
```
mitchellh, line 772: Alright here's what I'm running into. Nu can't really be used as a direct login shell for various reasons, so my command to start is actually this:

```
command = /etc/profiles/per-user/mitchellh/bin/fish -l -c "exec nu"
```

Even if I force `shell-integration = nushell`, it exits here because it sees `-c`. 

My thinking is that in this scenario, the _automatic use_ is not possible but we can still add the module to the autoload path and document that users should add `use ghostty *` to their Nu configs, so maybe we shouldn't fail out here and instead log. Same with everything else here.

mitchellh (changes requested):
One note so far.

src/termio/shell_integration.zig
jparise: > Nu can't really be used as a direct login shell for various reasons

Can you say more about this? I've been primarily testing with `--command=nu`, which works fine, but I suspect you're referring to other things like the environment?

> Even if I force shell-integration = nushell, it exits here because it sees -c.

That's a tricky case. I'll think about a general solution to that because I think it would apply to `bash`, too. Unfortunately, I'm doubtful we'll be able to make automatic integration work in this case because we'd need to rewrite your command to something like:

```
command = /etc/profiles/per-user/mitchellh/bin/fish -l -c "exec nu -e 'use ghostty *'"
```

mitchellh: > Unfortunately, I'm doubtful we'll be able to make automatic integration work in this case because we'd need to rewrite your command to something like:

I agree, but I still want the XDG data dirs setup at least so that I can manually do `use ghostty *`. Does that make sense? Or should I be thinking about this differently...

jparise: Yes, that makes sense in the nushell case because autoload-ing our `ghostty` module doesn't have any side effects (other than cleaning up `XDG_DATA_DIRS`).

See my latest commit.

mitchellh (approved):

PR #10280, opened 2026-01-11
feat(macos): Refine MacOS surface drag handle UI

martinemde:
Screenshot 2026-01-11 at 1 41 52 PM

This PR makes 3 small changes:

1. Makes the surface move grab handle present when the surface is hovered and the mouse cursor is not hidden.
2. Makes the grab handle partial width, allowing space to more easily grab the divider for resize (anywhere but the center) and increasing the grabbable area for the grab handle.
3. Adds appropriate padding to the top of the surface (in the metal stack so shaders can apply) to give space for the header so that text is not occluded by the grab handle.

I think it looks good and works well, but I suggest trying it out since the interaction is the most important part.

Problems I was trying to solve:
1. The old grab bar overlays actual clickable area on TUIs and can make them hard to use
2. The old bar makes the entire divider also a grab area, making divider resizing more difficult.
3. The old bar is not always present, making it hard to discover until you're going to resize something, which then is confusing
4. The old bar is not colored with the style.

https://github.com/user-attachments/assets/588a35b5-ba2f-4074-8edb-e090e0006224

AI Disclosure: I originally did this with Claude, but at this point I've gone over this code manually enough to feel somewhat familiar. I think the video and design speak for themselves and the code change is minimal, but I'm not a Swift programmer, so I can't evaluate whether this is the best possible solution.

Human Disclosure: I don't have a linux machine to check that the padding doesn't apply outside of MacOS. I find it hard to believe that it wouldn't work, but worth calling out.

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceGrabHandle.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceGrabHandle.swift
@@ -1,41 +1,30 @@
-import AppKit
 import SwiftUI
 
 extension Ghostty {
     /// A grab handle overlay at the top of the surface for dragging the window.
-    /// Only appears when hovering in the top region of the surface.
     struct SurfaceGrabHandle: View {
-        private let handleHeight: CGFloat = 10
-
         let surfaceView: SurfaceView
 
         @State private var isHovering: Bool = false
         @State private var isDragging: Bool = false
 
         var body: some View {
-            VStack(spacing: 0) {
-                Rectangle()
-                    .fill(Color.primary.opacity(isHovering || isDragging ? 0.15 : 0))
-                    .frame(height: handleHeight)
-                    .overlay(alignment: .center) {
-                        if isHovering || isDragging {
-                            Image(systemName: "ellipsis")
-                                .font(.system(size: 14, weight: .semibold))
-                                .foregroundColor(.primary.opacity(0.5))
-                        }
-                    }
-                    .contentShape(Rectangle())
-                    .overlay {
-                        SurfaceDragSource(
-                            surfaceView: surfaceView,
-                            isDragging: $isDragging,
-                            isHovering: $isHovering
-                        )
-                    }
+            ZStack {
+                SurfaceDragSource(
+                    surfaceView: surfaceView,
+                    isDragging: $isDragging,
+                    isHovering: $isHovering
+                )
+                .frame(width: 80, height: 12)
+                .contentShape(Rectangle())
 
-                Spacer()
+                Image(systemName: "ellipsis")
+                    .font(.system(size: 10, weight: .semibold))
+                    .foregroundColor(.primary.opacity(isHovering ? 0.8 : 0.3))
+                    .offset(y: -2)
+                    .allowsHitTesting(false)
             }
-            .frame(maxWidth: .infinity, maxHeight: .infinity)
+            .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
         }
     }
 }
--- a/macos/Sources/Ghostty/Surface View/SurfaceGrabHandle.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceGrabHandle.swift
@@ -3,10 +3,14 @@ import SwiftUI
 extension Ghostty {
     /// A grab handle overlay at the top of the surface for dragging the window.
     struct SurfaceGrabHandle: View {
-        let surfaceView: SurfaceView
+        @ObservedObject var surfaceView: SurfaceView
 
         @State private var isHovering: Bool = false
         @State private var isDragging: Bool = false
+
+        private var ellipsisVisible: Bool {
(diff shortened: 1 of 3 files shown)


tristan957:
Not sure I love this extra space always existing. I am more in favor of holding down a modifier while hovering the grab handles. It's been discussed elsewhere too.

martinemde:
Updated movie with the size adjustment.

edrozenberg:
> Not sure I love this extra space always existing. I am more in favor of holding down a modifier while hovering the grab handles.

Agree, wouldn't want to take any space away from content, or (different idea) to shift content down temporarily (shifting content is distracting and bad). 

An overlay that appears either with a delay or special keystroke seems like the best option along with the option to completely disable the overlay via config such that those who don't need or want the functionality never have to see it under any scenario.

martinemde:
In practice, I'm using about 4pts of font size in height, it's less than is often wasted by the space at the bottom of the terminal, though there's theoretically a pathological spacing where you lose a line of text because of this spacing.

What I'd honestly suggest is that people pull this down and build it and see if they actually don't like it, and which one feels better. The current solution isn't just confusing, but actively harms usability of TUIs and resizing of panes by dragging. Whatever the solution, I don't think the current incarnation is right.

I'm happy to try to test out some ideas or make multiple branches that people can try, but I assure you that until you _use_ the interaction in person, the videos don't do any of them justice. Case in point, the practical lost space of even 3 panes worth of header space is virtually invisible. The additional header space _feels_ better to me even ignoring the `...`. After using it, I feel like the text is too close to the top bar without it.

My point is: "press a key to show the grab handle" seems right until you try it, then it's jarring and confusing. "There's always a header space that's taken away from your terminal" sounds wrong until you use it, but it's actually kind of nice.

mitchellh (approved):
Thanks for the patience. Overall, really happy with this, I think making this bar more subtle is the right move. And, this looks good with both light and dark themes.

The only change I made was to remove that top padding. If the surface font is very small this can still be a problem but I think inconveniencing a minority over the majority is the right approach _for now_ until we can find something I find more satisfying.

martinemde:
Nice, I support that approach too. The extra space at the top was the main cause of contention.

Saeverix:
Today I noticed the new approach, I really like it. The previous solution was too distracting, that's why I created #10553 back then.
I really love this suble solution.

edrozenberg:
Thanks for the changes!

Working pretty well here, only wish the handle would not show up at all (be fully hidden) unless I mouse over the top of the split, where the split control bar is.

For ex. here you can see the dotted handle shows up, greyed out but still visible, in a split in a tab, with the split active and mouse pointer somewhere inside the split:

Handle Visible

MacOS Sequoia tip `e6da439e4` b-`14806`.

ldemailly:
I'd second the `...` are distracting when using a mouse aware TUI and think they should only show up if the mouse is in the top bar - should we open a new discussion/refinement for this? tia

mitchellh:
Feel free to open a new discussion, I think only showing the "..." when the mouse is in the upper area (not just the top bar, the hit target is tiny, but a small larger area) would be totally good.

sidav:
@martinemde completely unrelated, but I have no idea where else to ask for this: could you please share your color scheme or shader of whatever this is on the screenshot? This pixelated look with smooth background looks fantastic.

martinemde:
@sidav It's nice to hear you like it. Check my [style.ghostty](https://github.com/martinemde/dotfiles/blob/30120b8e177ec2fedfa9b7151cef65d67552b4de/home/dot_config/ghostty/style.ghostty) in my ghostty config and pull one of the shaders into your own, adjust background color and disable the background dimming. 

You can see the shaders in that directory. The shader is adding the crt style, a modified version of the video game CRT shader from another author. It's amazing. All my shaders take advantage of the focus tracking I added to ghostty's shadertoy system so only the focused or blurred splits render the shader.

Be warned though, I shut them off because I think the ghostty shader implementation is [a little too inefficient](https://github.com/ghostty-org/ghostty/discussions/10678). It will easily chew up your entire GPU and drain your battery.

Still, I ran this for a while and it looks amazing. I even made a [live rendering of the focus/blur shaders](https://martinemde.com/blog/ghostty-focus-shaders) and a [playground](https://martinemde.com/shaders) on my website so you can play with them. I miss it. The performance cost is per shader, so pick just one and it's not so bad.

Good luck! Happy to help

paddor:
Is there ANY way to disable the triple-dots completely? I don't ever want my mouse pointer to turn into a hand.

martinemde:
@paddor best thing to do is start or find a discussion to contribute to. That's how this project organizes feedback. Only a few people will notice your comment here (mostly me because I wrote the PR and get pinged).

For what it's worth, I agree about hiding it completely. I think an option to do so would be helpful for people that never need to move panes.

ldemailly:
if you make a new discussion please link it here. thanks.

Issue #10258, opened 2026-01-10
Page.layout capacity overflow causes massive page allocations


mitchellh:
From #9314 

Reproduction:

1. `brew install fortio/tap/fps`
2. `brew install fortio/tap/tcolor`
3. Run `fps`, press a key to start
4. Resize like crazy
5. Run `tcolor`, go to the HSL page (arrows)
6. Resize like crazy

It isn't consistent, but once in awhile you get insane runaway memory (gigabytes in seconds). 

We've confirmed via `footprint` that this is the PageList (app tag 1 on macOS). As such, I suggest despite the reproduction being on macOS, that this applies cross-platform. 

Known things:

1. It is definitely in `PageList.zig` (due to the app tag)
2. It is triggered by resize WITH reflow (maybe not the exclusive trigger, but definitely this)
3. The runaway memory usage is VERY fast
4. There are no errors in the logs (the resizes are succeeding)
5. The memory is freed when the window is closed or the terminal is reset. This implies that the memory is _tracked_ (not fully leaked). My guess then is that this is caused by pool capacity explosion somehow, but unclear how yet.
6. It can trigger with x-axis only (column size) changing. This implies the leak is within our column/reflow code.

mitchellh:
Another detail from a reproduction I just got: if you close the window, the memory does free. This implies the leak is not like #10251, and the leak might actually somehow be in the memory pool. #10251 NEVER freed the memory whereas this one does, meaning we still have track of it!

mitchellh:
Another critical hint: this can be reproduced with only x-axis resizing (column size changes). This should help narrow the leak down since our resize/reflow is split by axis.

mitchellh:
I'm in "read the code and think about it" mode here and I suspected the source was this branch. I commented it out and in builds with it commented out I can't reproduce the leak: https://github.com/ghostty-org/ghostty/blob/24880becb7702aeab4219dd44c59024f7d272870/src/terminal/PageList.zig#L1346 so I'm suspecting its triggered here somehow. I can't say with 100% certainty because the reproduction is not 100% either but I've gotten pretty good at triggering it and I can't get it to trigger with this commented out.

mitchellh:
It somehow isn't the arena blowing up! I changed our allocator so that the tags only apply to non-pooled (non-standard) pages and it's still the main type of memory blowing up. I don't currently understand how we can have non-arena memory but witness this explosion in memory usage that is properly freed with the PageList.

[...]

PR #10284, opened 2026-01-11
terminal: reflow capacity adjustment should use current page

mitchellh:
While reflowing, if we required more page capacity, we were using `cap.<field> * 2` to grow the capacity. But `cap` references the _source page capacity_. What this means is that for _every single row_ we would hit our capacity limit again and call `adjustCapacity` again, since we were never actually increasing our capacity to begin with.

This commit replaces our capacity growth by using our current target page capacity as the source instead. Under edge case scenarios, this reduces the `adjustCapacity` burden and should result in performance improvements.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -2729,6 +2729,10 @@ inline fn createPageExt(
     page.data.size.rows = 0;
     serial.* += 1;
 
+    // This is a very important requirement to ensure we pool
+    // our pages properly.
+    assert(page.data.memory.len == layout.total_size);
+
     if (total_size) |v| {
         // Accumulate page size now. We don't assert or check max size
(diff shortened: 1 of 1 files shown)


mitchellh:
cc @qwerasd205 if you have any ideas too, would be good to get an extra pair of eyes on this as well

mitchellh:
Btw even if we determine this doesn't fix the leak in #10258 and I'm just getting lucky, this is still the correct logic change to make 😄

mitchellh:
We found the root cause of #10258 and this doesn't fix it. I'm going to hold off merging this because this blocks our reproduction for #10258 (fixes it by masking it) and I want to be able to still hit that bug to verify our root cause fix actually fixes the problem. Then I'll merge this.

mitchellh:
Replaced by #10337

PR #10286, opened 2026-01-11
nix: update nixpkgs, remove zig.hook, and remove x11-gnome 

MithicSpirit:
As of NixOS/nixpkgs#473413, `zig.hook` no longer supports `zig_default_flags`, and now they can and must be provided in `zigBuildFlags` instead.

Updating also requires removing gnome-xorg since it has been removed from nixpkgs.

`nix flake check` succeeds on my system (x86_64-linux), with a couple deprecation warnings that I believe aren't important.

### DIFF
--- a/.github/workflows/publish-tag.yml
+++ b/.github/workflows/publish-tag.yml
@@ -64,7 +64,7 @@ jobs:
           mkdir blob
           mv appcast.xml blob/appcast.xml
       - name: Upload Appcast to R2
-        uses: ryand56/r2-upload-action@b801a390acbdeb034c5e684ff5e1361c06639e7c # v1
+        uses: ryand56/r2-upload-action@b801a390acbdeb034c5e684ff5e1361c06639e7c # v1.4
         with:
           r2-account-id: ${{ secrets.CF_R2_RELEASE_ACCOUNT_ID }}
           r2-access-key-id: ${{ secrets.CF_R2_RELEASE_AWS_KEY }}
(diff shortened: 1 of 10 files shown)


MithicSpirit:
I think the failures are an OOM error or something like that because it's having to recompile zig, which seems to be pulled in by `zon2nix`, since it still uses `zig_default_flags`. I'll try changing it to not need that.

pluiedev (approved):
I'm happy to merge this as a temporary fix, pending CI. Later please follow through with this and make a PR on the zon2nix repo, and then remove the overridden zon2nix here :)

jcollie:
I've updated `zon2nix` to be compatible with the Zig hook changes in nixpkgs.

MithicSpirit:
> I've updated `zon2nix` to be compatible with the Zig hook changes in nixpkgs.

@jcollie really? The latest commit on [jcollie/zon2nix](https://github.com/jcollie/zon2nix/) is still using `zig_default_flags`. Maybe you forgot to push?

jcollie:
Whoops, yeah forgot to push.

MithicSpirit:
I don't think the CI failures are my fault? I guess maybe updating nixpkgs updated dependencies like the spellchecker, which could be causing the failures. Regardless, I can try to fix them if desired. The action pins and typos seem pretty straightforward, but I'm not sure what should be done about the `HttpConnectionClosing` in test.

jcollie:
The `HttpConnectionClosing` is typically a transient networking error that can be "fixed" by retrying. The `typos` failure is probably caused by a newer version of `typos` being pickier. The GH actions pins may be a similar issue. Both of those should get fixed in this PR.

PR #7012, opened 2025-04-07
Introduce Issue Triage Template

taylrfnt:
The helper team has been discussing some common issues we see with Discussion submissions (missing info, duplicates, etc.), and pluie's #6937 has been a huge step forward - this PR introduces a template for the Issue Triage Discussion type.

The template has gone through a few revisions prior to this PR, but I am certain there are probably a few places to be cleaned up. You can test it out by [opening a new "Issue Triage" Discussion in my fork](https://github.com/taylrfnt/ghostty/discussions/new?category=issue-triage).

~~Creating this as a draft for now, since I may not be able to respond to any review comments in a timely manner.~~

PR #10287, opened 2026-01-11
fix(completions.fish): add +help and +version to completions

PeterCardenas:
`+help` and `+version` is missing from completions, even though they are working actions and are referenced in discussion template [here](https://github.com/ghostty-org/ghostty/pull/7012). This PR adds the completion for fish (as that is the shell that I use).

### DIFF
--- a/src/extra/bash.zig
+++ b/src/extra/bash.zig
@@ -158,9 +158,6 @@ fn writeBashCompletions(writer: *std.Io.Writer) !void {
     );
 
     for (@typeInfo(Action).@"enum".fields) |field| {
-        if (std.mem.eql(u8, "help", field.name)) continue;
-        if (std.mem.eql(u8, "version", field.name)) continue;
-
         const options = @field(Action, field.name).options();
         // assumes options will never be created with only <_name> members
         if (@typeInfo(options).@"struct".fields.len == 0) continue;
@@ -194,9 +191,6 @@ fn writeBashCompletions(writer: *std.Io.Writer) !void {
     );
 
     for (@typeInfo(Action).@"enum".fields) |field| {
-        if (std.mem.eql(u8, "help", field.name)) continue;
-        if (std.mem.eql(u8, "version", field.name)) continue;
-
         const options = @field(Action, field.name).options();
         if (@typeInfo(options).@"struct".fields.len == 0) continue;
 
@@ -272,9 +266,6 @@ fn writeBashCompletions(writer: *std.Io.Writer) !void {
     );
 
     for (@typeInfo(Action).@"enum".fields) |field| {
-        if (std.mem.eql(u8, "help", field.name)) continue;
-        if (std.mem.eql(u8, "version", field.name)) continue;
-
         try writer.writeAll(pad1 ++ "topLevel+=\" +" ++ field.name ++ "\"\n");
     }
 
--- a/src/extra/fish.zig
+++ b/src/extra/fish.zig
@@ -28,8 +28,6 @@ fn writeCompletions(writer: *std.Io.Writer) !void {
         try writer.writeAll("set -l commands \"");
         var count: usize = 0;
         for (@typeInfo(Action).@"enum".fields) |field| {
-            if (std.mem.eql(u8, "help", field.name)) continue;
-            if (std.mem.eql(u8, "version", field.name)) continue;
             if (count > 0) try writer.writeAll(" ");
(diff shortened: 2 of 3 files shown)


rhodes-b:
The other shells have the same "problem" this shouldn't only be changed for fish
https://github.com/ghostty-org/ghostty/blob/main/src/extra/bash.zig#L161-L162
https://github.com/ghostty-org/ghostty/blob/main/src/extra/zsh.zig#L171-L172

jparise (approved):

jparise:
I don't recall why we excluded these before, but I think it makes sense to include them now.

jparise:
> The other shells have the same "problem" this shouldn't only be changed for fish https://github.com/ghostty-org/ghostty/blob/main/src/extra/bash.zig#L161-L162 https://github.com/ghostty-org/ghostty/blob/main/src/extra/zsh.zig#L171-L172

I'll fix these up and add to this branch in a few.

PeterCardenas:
i removed the similar lines in the other shells, but i don't really understand the output as i do fish. i was able to test bash, but not sure how to test zsh. here's the diff, i can push up the changes in this pr if that's desired:
```diff
diff --git a/src/extra/bash.zig b/src/extra/bash.zig
index ee9a7895c..14bf3a225 100644
--- a/src/extra/bash.zig
+++ b/src/extra/bash.zig
@@ -158,9 +158,6 @@ fn writeBashCompletions(writer: *std.Io.Writer) !void {
     );
 
     for (@typeInfo(Action).@"enum".fields) |field| {
-        if (std.mem.eql(u8, "help", field.name)) continue;
-        if (std.mem.eql(u8, "version", field.name)) continue;
-
         const options = @field(Action, field.name).options();
         // assumes options will never be created with only <_name> members
         if (@typeInfo(options).@"struct".fields.len == 0) continue;
@@ -194,9 +191,6 @@ fn writeBashCompletions(writer: *std.Io.Writer) !void {
     );
 
     for (@typeInfo(Action).@"enum".fields) |field| {
-        if (std.mem.eql(u8, "help", field.name)) continue;
-        if (std.mem.eql(u8, "version", field.name)) continue;
-
         const options = @field(Action, field.name).options();
         if (@typeInfo(options).@"struct".fields.len == 0) continue;
 
@@ -272,9 +266,6 @@ fn writeBashCompletions(writer: *std.Io.Writer) !void {
     );
 
     for (@typeInfo(Action).@"enum".fields) |field| {
-        if (std.mem.eql(u8, "help", field.name)) continue;
-        if (std.mem.eql(u8, "version", field.name)) continue;
-
         try writer.writeAll(pad1 ++ "topLevel+=\" +" ++ field.name ++ "\"\n");
     }
 
diff --git a/src/extra/zsh.zig b/src/extra/zsh.zig
index 2fad4234a..376db807f 100644
--- a/src/extra/zsh.zig
+++ b/src/extra/zsh.zig
@@ -139,9 +139,6 @@ fn writeZshCompletions(writer: *std.Io.Writer) !void {
         var count: usize = 0;
         const padding = "        ";
         for (@typeInfo(Action).@"enum".fields) |field| {
-            if (std.mem.eql(u8, "help", field.name)) continue;
-            if (std.mem.eql(u8, "version", field.name)) continue;
-
             try writer.writeAll(padding ++ "'+");
             try writer.writeAll(field.name);
             try writer.writeAll("'\n");
@@ -168,9 +165,6 @@ fn writeZshCompletions(writer: *std.Io.Writer) !void {
     {
         const padding = "        ";
         for (@typeInfo(Action).@"enum".fields) |field| {
-            if (std.mem.eql(u8, "help", field.name)) continue;
-            if (std.mem.eql(u8, "version", field.name)) continue;
-
             const options = @field(Action, field.name).options();
             // assumes options will never be created with only <_name> members
             if (@typeInfo(options).@"struct".fields.len == 0) continue;

```

jparise:
> I don't recall why we excluded these before, but I think it makes sense to include them now.

I _think_ it might have been an early reaction to the fact that we have both `+version` and `--version`, which are effectively aliases (and ditto for `+help` and `--help`).

But I think it's better to accurately reflect the full interface, so I like this change.

jparise:
> i removed the similar lines in the other shells, but i don't really understand the output as i do fish. i was able to test bash, but not sure how to test zsh.

I had the same diff and was able to verify both bash and zsh. Pushed to this branch.

PR #10290, opened 2026-01-12
Terminal: keep cross-boundary rows dirty in {insert,delete}Lines

danielwe:
This fixes #10265 and thus also the remaining part of #9718 and likely #10250.

The issue was that when using `insertLines` and `deleteLines` to generate scrolling in a region that spans a page boundary, rows that are replaced by a row from a different page lose their dirty flags in the clone operation, since the flag is part of the data that gets cloned. The solution is to set the dirty flag again after the clone, just like the non-cloning branch does after the pointer swap.

**AI disclosure:** Amp is the MVP here. I prompted it with the hypothesis I developed in #10265 (that this happens when the scrolling region spans a page boundary), supplemented with insight I gained from perusing asciicast files (that the offending scrolling operations are always triggered by `CSI 1 L` or `CSI 1 M`, that is, `Terminal.insertLines` or `Terminal.deleteLines`). Amp figured out the rest and drafted the fix and tests. For free!

I cleaned up the tests and then pushed back a bit against the logic behind the fix, which led to a better understanding and what I think is a more appropriate fix. I can explain the details if there's interest, or people can just skim the thread here: https://ampcode.com/threads/T-019bb0d6-5334-744a-b78a-7c997ec1fade.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -1602,9 +1602,6 @@ pub fn insertLines(self: *Terminal, count: usize) void {
         const cur_rac = cur_p.rowAndCell();
         const cur_row: *Row = cur_rac.row;
 
-        // Mark the row as dirty
-        cur_p.markDirty();
-
         // If this is one of the lines we need to shift, do so
         if (y > adjusted_count) {
             const off_p = cur_p.up(adjusted_count).?;
@@ -1690,10 +1687,6 @@ pub fn insertLines(self: *Terminal, count: usize) void {
                     // Continue the loop to try handling this row again.
                     continue;
                 };
-
-                // The clone operation may overwrite the dirty flag, so make
-                // sure the row is still marked dirty.
-                dst_row.dirty = true;
             } else {
                 if (!left_right) {
                     // Swap the src/dst cells. This ensures that our dst gets the
@@ -1702,9 +1695,6 @@ pub fn insertLines(self: *Terminal, count: usize) void {
                     const dst = dst_row.*;
                     dst_row.* = src_row.*;
                     src_row.* = dst;
-
-                    // Make sure the row is marked as dirty though.
-                    dst_row.dirty = true;
 
                     // Ensure what we did didn't corrupt the page
                     cur_p.node.data.assertIntegrity();
@@ -1731,6 +1721,9 @@ pub fn insertLines(self: *Terminal, count: usize) void {
                 cells[self.scrolling_region.left .. self.scrolling_region.right + 1],
             );
         }
+
+        // Mark the row as dirty
+        cur_p.markDirty();
 
         // We have successfully processed a line.
         y -= 1;
@@ -1808,9 +1801,6 @@ pub fn deleteLines(self: *Terminal, count: usize) void {
     while (y < rem) {
         const cur_rac = cur_p.rowAndCell();
         const cur_row: *Row = cur_rac.row;
-
-        // Mark the row as dirty
-        cur_p.markDirty();
 
         // If this is one of the lines we need to shift, do so
         if (y < rem - adjusted_count) {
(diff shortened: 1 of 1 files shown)


luisdavim:
I ran a build from this branch and it does seem to solve the problem, at least I was not able to reproduce it whilst running it and I can reliably reproduce it using the latest tip built...

jparise:
> rows that are replaced by a row from a different page lose their dirty flags in the clone operation, since the flag is part of the data that gets cloned. The solution is to set the dirty flag again after the clone, just like the non-cloning branch does after the pointer swap.

With this change, it looks like most (all?) calls to `clonePartialRowFrom()` from outside the `PageList` require setting (preserving) the dirty flag. Just a general observation.

danielwe:
> With this change, it looks like most (all?) calls to `clonePartialRowFrom()` from outside the `PageList` require setting (preserving) the dirty flag. Just a general observation.

Yeah I thought about this, but didn't feel like I have a thorough enough understanding of the code to propose a deeper change like that. Also curious whether it would be sufficient to move the `cur_p.markDirty()` call from the beginning of the loop body to the end, and remove the all the dirty flag handling in these branches. But again, not clear to me if there might be any (possibly untested) edge cases, maybe relating to non-full-width scroll regions, that could be thrown off by that.

But happy to update this PR, or have it updated by a maintainer, if someone who knows this code well would prefer one of those solutions instead.

mitchellh:
> With this change, it looks like most (all?) calls to clonePartialRowFrom() from outside the PageList require setting (preserving) the dirty flag. Just a general observation.

I think this is fine, or intended. A `clone` to me implies an exact copy, and the dirty state is part of the page/row so cloning a non-dirty row should overwrite everything else in the target. Dirty tracking is up to callers.

> Also curious whether it would be sufficient to move the cur_p.markDirty() call from the beginning of the loop body to the end, and remove the all the dirty flag handling in these branches.

It does look like `cur_p.markDirty` could be moved to the bottom of the conditional prior to moving the `y` and remove all other dirty tracking. I think the steps I'd take to test this (I can do this later if you don't) are:

- Remove the other dirty setting, see if tests fail (they should, all dirty flags should be tested)
- If no tests fail, add tests to check that dirty state first
- Move `cur_p.markDirty`, verify all tests pass

In that way, we verify every point.

danielwe:
Did all that and can confirm:

* Removing the dirty marking in the fast path (the `if (!left_right)` branch) triggers a bunch of test failures
* Moving `cur_p.markDirty()` below the conditional fixes them again
* With `cur_p.markDirty()` in this position, the post-clone dirty marking introduced above is no longer necessary

This is true for both `insertLines` and `deleteLines`.

However, the postcondition of the loop body is definitely different with this change. Specifically, in the fast path, both `src_row` and `dst_row` would previously end up dirty. Now it's only `dst_row`. My hesitation before was because I don't know if stuff like this, even if untested, could be important in some edge case. If it ain't broke don't fix it, et cetera. But just quickly test driving a build, it's not causing any obvious problems.

I'll push a commit with this change and let you decide whether to keep or discard it.

src/terminal/Terminal.zig
```diff
@@ -1699,9 +1696,6 @@ pub fn insertLines(self: *Terminal, count: usize) void {
                     dst_row.* = src_row.*;
                     src_row.* = dst;
 
-                    // Make sure the row is marked as dirty though.
-                    dst_row.dirty = true;
```
mitchellh, line 1703: We probably should be marking `src_row` dirty here as well, since we are swapping within the same state (not just cloning). We might have to write a new unit test to verify this if there wasn't one previously.

danielwe: I _think_ what happens is that `src_row` will eventually be `cur_p` in a different iteration, and either its content will be replaced, or it will go to the `else` branch and get cleared, and either way it will be marked dirty at the end of that iteration. So the necessary tests may already exist.

Certainly I think the new tests added in this PR would catch if anything were off for the `insertLines(1)/deleteLines(1)` cases. Not sure about other counts.

danielwe: In other words, if it were necessary to set the dirty flag here, that would mean there's a bug somewhere else, because the content that ends up in `src_row` at this point is just a placeholder and not what's actually supposed to render once the function returns. Seems more appropriate to set the dirty flag when the row is updated to its actual new content.

mitchellh: I'll double check, that may be a valid assessment.

mitchellh (approved):

PR #10295, opened 2026-01-12
font/shaper/coretext: Detect ligatures to avoid positioning (marking) glyphs incorrectly

jacobsandlund:
This PR implements a heuristic for detecting ligatures that then prevents resetting to the cell grid prematurely, avoiding position errors from glyphs that follow ligatures. The inspiration for this was the example from https://github.com/ghostty-org/ghostty/pull/10179 that was off.

See the giant comment for an explanation of the heuristic.

The tests for this are the following:

### Tai Tham
The Tai Tham example from https://github.com/ghostty-org/ghostty/pull/10179 got updated since now the marking glyph is attached to the ligature cell correctly.

Browser:
ᩉ᩠ᨿᩩ

Before:
CleanShot 2026-01-12 at 10 02 42@2x

After:
CleanShot 2026-01-12 at 10 02 57@2x

### Javanese

Browser:
ꦤ꧀ꦲꦸ

Before:
CleanShot 2026-01-12 at 10 05 43@2x

After:
CleanShot 2026-01-12 at 10 05 23@2x

There does seem to be a small maybe single pixel difference here, and I didn't investigate why.

### Chakma

Browser:
𑄝𑄖𑄳𑄠𑄬

Before:
CleanShot 2026-01-12 at 10 07 52@2x

After:
CleanShot 2026-01-12 at 10 08 08@2x

### Bengali

Browser:
রাষ্ট্রে

Before:
CleanShot 2026-01-12 at 10 10 42@2x

After:
CleanShot 2026-01-12 at 10 11 00@2x

This one doesn't match the browser, but it seems like a CoreText or font issue.

### Log output

I've got a `log.txt` with 15k lines from "cell_offset.cluster differs from cluster (potential ligature detected)" logs when I run [ttylang](https://github.com/jacobsandlund/ttylang) (printing the Universal Declaration of Human Rights in various languages).

But right now when I try to create a gist with it, I get:

CleanShot 2026-01-12 at 10 15 29@2x


src/font/shaper/coretext.zig
```diff
@@ -641,81 +689,147 @@ pub const Shaper = struct {
...
         const positions_differ = @abs(x_offset_diff) > 0.0001 or @abs(y_offset_diff) > 0.0001;
         const old_offset_y = position.y - cell_offset_y;
         const position_y_differs = @abs(cell_offset_y) > 0.0001;
+        const cluster = state.codepoints.items[index].cluster;
+        const cluster_differs = cluster != cell_offset.cluster;
+
+        // To debug every loop, flip this to true:
+        const extra_debugging = false;
+
+        const is_previous_codepoint_prepend = if (cluster_differs or
```
jacobsandlund, line 714: i had a theory that maybe prepend characters would mess with this heuristic, but logging this didn't find any cases, at least with printing the UDHR.

src/font/shaper/coretext.zig
```diff
@@ -2329,7 +2646,7 @@ fn testShaperWithDiscoveredFont(alloc: Allocator, font_req: [:0]const u8) !TestS
             .monospace = false,
         });
         defer disco_it.deinit();
-        var face: font.DeferredFace = (try disco_it.next()).?;
+        var face: font.DeferredFace = (try disco_it.next()) orelse return error.FontNotFound;
```
jacobsandlund, line 2651: when i at first tried to load "Noto Sans Chakma Regular" instead of "Noto Sans Chakma", I hit a panic on `.?`.

mitchellh (approved):
I'm not a big fan of heuristics but I think that if we have tight test coverage, no existing tests change, and we have new tests covering the change, then I'm comfortable with at least _trying it_.

It might be a useful exercise for us to think about designing an OOB test program that takes a large corpus of fonts, and runs various parts of our font system on it (metrics, shaping, etc.) and puts them in a golden file (even a CSV) that we can then always verify if there are any changes and consider the impact....

mitchellh:
We've got a real test failure to look into.

jacobsandlund:
> We've got a real test failure to look into.

Blasted, I had it in my mind to run the full suite before pushing and forgot. It's now fixed in https://github.com/ghostty-org/ghostty/pull/10295/commits/55583d9f27ff9165286c45df8db4f2c999583468

This is another example of this heuristic working:

Browser:

अपार्टमेंट

Before:

CleanShot 2026-01-13 at 10 39 30@2x

After:

CleanShot 2026-01-13 at 10 39 56@2x

> It might be a useful exercise for us to think about designing an OOB test program...

I like that idea!

mitchellh (approved):
Approved pending CI.

PR #10297, opened 2026-01-12
PageList can initialize with more than ~46,000 columns

### LINKED ISSUE #10258: Page.layout capacity overflow causes massive page allocations

mitchellh:
I've been poking around our internals around overflow cases while diagnosing systemic ways to fix #10258. 

One thing I found is that our terminal internals advertise that they can accept up to u16 (~65K) rows and columns for the active area, but at around 46K columns on x86_64 and 47K on aarch64, our terminal initialize would error out. Worse, we'd actually trigger integer overflow and crash in safe builds (and have silent corruption in release builds).

In fixing this, I realized we could comptime-verify a bunch of things and avoid overflows completely. So this diff contains a bunch of places that previously had error returns and now do not! 😄 This is a bit silly, but with libghostty it's more relevant to have correctness around the edges.  

We now accept full u16 rows/cols. More than that isn't allowed by the type system and our upstream callers must handle that appropriately.

cc @qwerasd205 since its relevant to layout fallibility but not on the same path

### DIFF
--- a/src/terminal/page.zig
+++ b/src/terminal/page.zig
@@ -1662,43 +1662,42 @@ pub const Capacity = struct {
         cols: ?size.CellCountInt = null,
     };
 
+    /// Returns the maximum number of columns that can be used with this
+    /// capacity while still fitting at least one row. Returns null if even
+    /// a single column cannot fit (which would indicate an unusable capacity).
+    ///
+    /// Note that this is the maximum number of columns that never increases
+    /// the amount of memory the original capacity will take. If you modify
+    /// the original capacity to add rows, then you can fit more columns.
+    pub fn maxCols(self: Capacity) ?size.CellCountInt {
+        const available_bits = self.availableBitsForGrid();
+
+        // If we can't even fit the row metadata, return null
+        if (available_bits <= @bitSizeOf(Row)) return null;
+
+        // We do the math of how many columns we can fit in the remaining
+        // bits ignoring the metadat of a row.
+        const remaining_bits = available_bits - @bitSizeOf(Row);
+        const max_cols = remaining_bits / @bitSizeOf(Cell);
+
+        // Clamp to CellCountInt max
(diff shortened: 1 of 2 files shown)


src/terminal/page.zig
```diff
@@ -1556,7 +1557,7 @@ pub const Page = struct {
         const rows_start = 0;
         const rows_end: usize = rows_start + (rows_count * @sizeOf(Row));
 
-        const cells_count: usize = @intCast(cap.cols * cap.rows);
+        const cells_count: usize = @as(usize, cap.cols) * @as(usize, cap.rows);
```
mitchellh, line 1560: This can't overflow because both are u16s so it's `u16 * u16`. There are a number of other plays in layout here we can overflow, which is the root cause of #10258 but we're not fixing that here and we're not introducing any new areas of overflow.

PR #10324, opened 2026-01-15
feat/macos search prefill selection

ShanthoshPrabhu:
Currently when the search cmd is executed, search input remains empty even if text is actively selected in the terminal

Change:
Modified the start_search logic to check for an active selection, and if that exists, set the search state with that value. Also pre-fill only if the selection stays within one line

Note: I used an AI(codex) to help me navigate through the codebase.


mitchellh:
The `search_selection` binding is distinct and already exists.

You can get the search or search selection through the new chained keybindings to chain both.

ShanthoshPrabhu:
Oh, got it, thanks!

PR #10328, opened 2026-01-15
gtk: add GSettings generic module and respect gtk-enable-primary-paste on gtk systems

evertonstz:
# Add GSettings Support for Primary Paste

Implements support for `org.gnome.desktop.interface gtk-enable-primary-paste` to allow users to disable middle-click paste. Also refactors GTK Settings access into a reusable generic module.

## Changes

- **NEW**: `src/apprt/gtk/gsettings.zig` - Generic GTK Settings reader supporting `bool` and `c_int` types, portal-aware for Flatpak/Snap
- **MODIFIED**: `src/apprt/gtk/class/surface.zig` - Reads primary paste setting and refactors gtk-xft-dpi to use new module

## Behavior
- Setting `false` → Middle-click paste blocked
- Setting `true` or unavailable → Middle-click paste works (default)
- Uses GTK Settings API which automatically uses XDG Desktop Portal in sandboxed environments

Note: No unit tests added as this is a thin wrapper around GTK Settings API that's already tested indirectly through surface.zig. Happy to add tests if desired, though they would require an active display environment and skip on most CI systems.

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -675,8 +675,7 @@ pub const Surface = extern struct {
         /// The context for this surface (window, tab, or split)
         context: apprt.surface.NewSurfaceContext = .window,
 
-        /// Whether primary paste (middle-click paste) is enabled via GNOME settings.
-        /// If true, middle-click paste is enabled. If false, it's disabled.
+        /// Whether primary paste (middle-click paste) is enabled.
         gtk_enable_primary_paste: bool = true,
 
         pub var offset: c_int = 0;
@@ -1770,12 +1769,8 @@ pub const Surface = extern struct {
         priv.im_composing = false;
         priv.im_len = 0;
 
-        // Read GNOME desktop interface settings for primary paste (middle-click)
-        // This is only relevant on Linux systems with GNOME settings available
-        priv.gtk_enable_primary_paste = gsettings.get(.@"gtk-enable-primary-paste") orelse blk: {
-            log.warn("gtk-enable-primary-paste was not set, using default value", .{});
-            break :blk false;
-        };
+        // Read GTK primary paste setting
+        priv.gtk_enable_primary_paste = gsettings.get(.@"gtk-enable-primary-paste") orelse true;
 
         // Set up to handle items being dropped on our surface. Files can be dropped
         // from Nautilus and strings can be dropped from many programs. The order
@@ -2696,8 +2691,6 @@ pub const Surface = extern struct {
         // Report the event
         const button = translateMouseButton(gesture.as(gtk.GestureSingle).getCurrentButton());
 
-        // Check if middle button paste should be disabled based on GNOME settings
-        // If gtk_enable_primary_paste is explicitly false, skip processing middle button
         if (button == .middle and !priv.gtk_enable_primary_paste) {
             return;
         }
@@ -2753,8 +2746,6 @@ pub const Surface = extern struct {
         const gtk_mods = event.getModifierState();
         const button = translateMouseButton(gesture.as(gtk.GestureSingle).getCurrentButton());
 
-        // Check if middle button paste should be disabled based on GNOME settings
-        // If gtk_enable_primary_paste is explicitly false, skip processing middle button
         if (button == .middle and !priv.gtk_enable_primary_paste) {
             return;
         }
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -677,7 +677,7 @@ pub const Surface = extern struct {
 
         /// Whether primary paste (middle-click paste) is enabled via GNOME settings.
         /// If true, middle-click paste is enabled. If false, it's disabled.
-        gtk_enable_primary_paste: ?bool = true,
+        gtk_enable_primary_paste: bool = true,
 
         pub var offset: c_int = 0;
     };
@@ -2698,7 +2698,7 @@ pub const Surface = extern struct {
 
         // Check if middle button paste should be disabled based on GNOME settings
         // If gtk_enable_primary_paste is explicitly false, skip processing middle button
-        if (button == .middle and !(priv.gtk_enable_primary_paste)) {
+        if (button == .middle and !priv.gtk_enable_primary_paste) {
             return;
         }
 
@@ -2755,7 +2755,7 @@ pub const Surface = extern struct {
 
         // Check if middle button paste should be disabled based on GNOME settings
         // If gtk_enable_primary_paste is explicitly false, skip processing middle button
-        if (button == .middle and !(priv.gtk_enable_primary_paste)) {
+        if (button == .middle and !priv.gtk_enable_primary_paste) {
             return;
         }
 
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1517,7 +1517,7 @@ pub const Surface = extern struct {
         const xft_dpi_scale = xft_scale: {
             // gtk-xft-dpi is font DPI multiplied by 1024. See
             // https://docs.gtk.org/gtk4/property.Settings.gtk-xft-dpi.html
-            const gtk_xft_dpi = gsettings.readSetting(c_int, "gtk-xft-dpi") orelse {
+            const gtk_xft_dpi = gsettings.get(.gtk_xft_dpi) orelse {
                 log.warn("gtk-xft-dpi was not set, using default value", .{});
                 break :xft_scale 1.0;
             };
@@ -1527,7 +1527,7 @@ pub const Surface = extern struct {
             // https://gitlab.gnome.org/GNOME/libadwaita/-/commit/a7738a4d269bfdf4d8d5429ca73ccdd9b2450421
             // https://gitlab.gnome.org/GNOME/libadwaita/-/commit/9759d3fd81129608dd78116001928f2aed974ead
             if (gtk_xft_dpi <= 0) {
-                log.warn("gtk-xft-dpi was not set, using default value", .{});
+                log.warn("gtk-xft-dpi has invalid value ({}), using default", .{gtk_xft_dpi});
                 break :xft_scale 1.0;
             }
 
@@ -1773,7 +1773,10 @@ pub const Surface = extern struct {
 
         // Read GNOME desktop interface settings for primary paste (middle-click)
         // This is only relevant on Linux systems with GNOME settings available
-        priv.gtk_enable_primary_paste = gsettings.readSetting(bool, "gtk-enable-primary-paste");
+        priv.gtk_enable_primary_paste = gsettings.get(.gtk_enable_primary_paste) orelse blk: {
+            log.warn("gtk-enable-primary-paste was not set, using default value", .{});
+            break :blk false;
+        };
 
         // Set up to handle items being dropped on our surface. Files can be dropped
         // from Nautilus and strings can be dropped from many programs. The order
@@ -2696,7 +2699,7 @@ pub const Surface = extern struct {
 
         // Check if middle button paste should be disabled based on GNOME settings
         // If gtk_enable_primary_paste is explicitly false, skip processing middle button
-        if (button == .middle and priv.gtk_enable_primary_paste == false) {
+        if (button == .middle and !(priv.gtk_enable_primary_paste orelse true)) {
             return;
         }
 
@@ -2753,7 +2756,7 @@ pub const Surface = extern struct {
 
         // Check if middle button paste should be disabled based on GNOME settings
         // If gtk_enable_primary_paste is explicitly false, skip processing middle button
-        if (button == .middle and !priv.gtk_enable_primary_paste) {
+        if (button == .middle and !(priv.gtk_enable_primary_paste orelse false)) {
             return;
         }
 
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -2756,7 +2756,7 @@ pub const Surface = extern struct {
 
         // Check if middle button paste should be disabled based on GNOME settings
         // If gtk_enable_primary_paste is explicitly false, skip processing middle button
-        if (button == .middle and !(priv.gtk_enable_primary_paste orelse false)) {
+        if (button == .middle and !(priv.gtk_enable_primary_paste orelse true)) {
             return;
         }
 
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1516,7 +1516,7 @@ pub const Surface = extern struct {
         const xft_dpi_scale = xft_scale: {
             // gtk-xft-dpi is font DPI multiplied by 1024. See
             // https://docs.gtk.org/gtk4/property.Settings.gtk-xft-dpi.html
-            const gtk_xft_dpi = gsettings.get(.gtk_xft_dpi) orelse {
+            const gtk_xft_dpi = gsettings.get(.@"gtk-xft-dpi") orelse {
                 log.warn("gtk-xft-dpi was not set, using default value", .{});
                 break :xft_scale 1.0;
             };
@@ -1772,7 +1772,7 @@ pub const Surface = extern struct {
 
         // Read GNOME desktop interface settings for primary paste (middle-click)
         // This is only relevant on Linux systems with GNOME settings available
-        priv.gtk_enable_primary_paste = gsettings.get(.gtk_enable_primary_paste) orelse blk: {
+        priv.gtk_enable_primary_paste = gsettings.get(.@"gtk-enable-primary-paste") orelse blk: {
             log.warn("gtk-enable-primary-paste was not set, using default value", .{});
             break :blk false;
         };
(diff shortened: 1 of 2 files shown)


evertonstz:
@mitchellh oops, did I do something wrong here? 🤔

mitchellh:
The PR said “initial commit” was a draft and had non contents. I figured it was Ai slop. Let’s see.

evertonstz:
Oh! Sorry, that was my vscode PR extension being stupid and naming the PR as the first commits name :)

evertonstz:
Just to add some context: [gnome will disable the primary clipboard by default](https://gitlab.gnome.org/GNOME/gsettings-desktop-schemas/-/merge_requests/119) in the next release, so this implementation is just to comply with that (and with the cases when users enable it via gsettings).

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,34 @@
...
+///
+/// Supported platforms: Linux, FreeBSD
+/// Supported types: bool, c_int
+///
+/// Example usage:
+///   const enabled = readSetting(bool, "gtk-enable-primary-paste");
+///   const dpi = readSetting(c_int, "gtk-xft-dpi");
+pub fn readSetting(comptime T: type, key: [*:0]const u8) ?T {
+    // Only available on systems that use GTK (Linux, FreeBSD)
+    if (comptime builtin.os.tag != .linux and builtin.os.tag != .freebsd) return null;
```
pluiedev: This codepath will never be hit. Plus, if something ever gets to this point without being on GTK something is terribly wrong

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,34 @@
...
+
+    const settings = gtk.Settings.getDefault() orelse return null;
+
+    // For bool and c_int, we use c_int as the underlying GObject type
+    // because GTK boolean properties are stored as integers
+    var value = gobject.ext.Value.new(c_int);
```
pluiedev: This will definitely depend on T as well

```suggestion
    // For bool and c_int, we use c_int as the underlying GObject type
    // because GTK boolean properties are stored as integers
    var value = gobject.ext.Value.new(switch (T) {
        bool, c_int => c_int,
    });
```

src/apprt/gtk/class/surface.zig
```diff
@@ -2736,6 +2751,12 @@ pub const Surface = extern struct {
         const gtk_mods = event.getModifierState();
         const button = translateMouseButton(gesture.as(gtk.GestureSingle).getCurrentButton());
 
+        // Check if middle button paste should be disabled based on GNOME settings
+        // If gtk_enable_primary_paste is explicitly false, skip processing middle button
+        if (button == .middle and priv.gtk_enable_primary_paste == false) {
```
pluiedev:
```suggestion
        if (button == .middle and !priv.gtk_enable_primary_paste) {
```

pluiedev (changes requested):
AI disclosure, please. I refuse to believe this was mostly written by a human.

src/apprt/gtk/gsettings.zig
evertonstz: That's just some defensive programming, I can remove it if its terribly needed :shrug: 
Also, about the AI claim: what happened with people just being new to a language? I come from C and Go and I tend to code like I usually do with those.

pluiedev: Upon a second examination it should never reach this point since the `gtk` import wouldn't even exist on other platforms, so this should just be redundant

mitchellh: > Also, about the AI claim: what happened with people just being new to a language? I come from C and Go and I tend to code like I usually do with those.

The AI people ruined it for all of you. Now we assume idiots because we're inundated with them every day.

src/apprt/gtk/gsettings.zig
evertonstz: I see! What if we already add support for strings in the switch?

```
    const GValueType = switch (T) {
        bool, c_int => c_int,
        ?[:0]const u8 => ?[:0]const u8,
        else => @compileError("Unsupported type '" ++ @typeName(T) ++ "' for GTK setting. Supported types: bool, c_int, ?[:0]const u8"),
    };
    var value = gobject.ext.Value.new(GValueType);
```

We could return something like this:
```
    return switch (T) {
        bool => value.getInt() != 0,
        c_int => value.getInt(),
        ?[*:0]const u8 => value.getString(),
        else => unreachable,
    };
```

I don't remember if gsttings can store other types, but could be easily expanded from this. WDYT?

src/apprt/gtk/gsettings.zig
evertonstz: @mitchellh seeing from that perspective and taking a moment to have some empaty, I can just imagine the kind of crap you guys have to deal daily running a high profile project. No hard feelings at all.

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,34 @@
...
+/// This automatically uses XDG Desktop Portal in Flatpak environments.
+/// Returns null if not on a GTK-supported platform or if the setting cannot be read.
+///
+/// Supported platforms: Linux, FreeBSD
+/// Supported types: bool, c_int
+///
+/// Example usage:
+///   const enabled = readSetting(bool, "gtk-enable-primary-paste");
+///   const dpi = readSetting(c_int, "gtk-xft-dpi");
+pub fn readSetting(comptime T: type, key: [*:0]const u8) ?T {
```
mitchellh: Thinking about this more, I actually think we should dump this wrapper entirely. There are certain types (boxed, strings) that we aren't going to be able to return safely without an allocator or using GTK's heap allocator, I think?

It might make more sense to expose the first-class GTK settings that are defined by GTK itself as something like an enum:

```
pub fn get(comptime Key: Key) Key.Type()
```

We have this pattern elsewhere in the code you can take a look at. 

For custom settings, dropping down to manually grabbing properties or exposing something at that point might make the most sense.

mitchellh (changes requested):
I recommend simplifying by removing the generic wrapper for now. I think that's going to hang this whole thing up. I suggested some other thoughts, but I'm not 100% sure its the way.

src/apprt/gtk/gsettings.zig
mitchellh: https://github.com/ianprime0509/zig-gobject/blob/main/extensions/gobject2.zig

src/apprt/gtk/gsettings.zig
mitchellh: Yeah, look at our issue/PR history, unfortunately. It could be the roster for a class of students with learning disabilities. Which is, quite frankly, insulting to the students with learning disabilities because they're perfect individuals unlike the PR submitters.

src/apprt/gtk/gsettings.zig
evertonstz: I see, those are some new concepts I have not studied yet. I am gonna take my time to see how you guys use this pattern in the current code and see if I can implement something. Meanwhile I will put the PR back to draft!

Thx for taking the time to review it :)

mitchellh: You can probably speed things up by removing the generic wrapper now and calling `getProperty` directly. For something as simple as this I wouldn't be opposed to it. And we can do the wrapper in a separate PR.

evertonstz: I am doing some research and tbh I actually like the enum solution more, seems way more refined and elegant. I think doing it this way will not only ship better and safer code for future use as an api, but I will also learn something new and have some fun implementing it (or at least trying to). Its a win-win.

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,145 @@
...
+        // before the GValue is destroyed by defer value.unset()
+        .gtk_font_name,
+        => blk: {
+            // This is defensive: we have already checked at compile-time that
+            // an allocator is provided for allocating types
+            const alloc = allocator.?;
+            const ptr = value.getString() orelse break :blk null;
+            const str = std.mem.span(ptr);
+            break :blk try alloc.dupe(u8, str);
+        },
```
evertonstz: While we wont use it now, I have added the `gtk_font_name` as an example of a value that needs allocation.

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,145 @@
...
+            .gtk_enable_primary_paste => "gtk-enable-primary-paste",
+            .gtk_xft_dpi => "gtk-xft-dpi",
+            .gtk_font_name => "gtk-font-name",
+        };
+    }
+
+    /// Returns true if this setting type requires memory allocation.
+    /// this is defensive: types that do not need allocation need to be
+    /// explicitly marked here
+    fn requiresAllocation(comptime self: Key) bool {
```
evertonstz, line 30: I tried to be defensive here: in case someone adds a new setting and forget to check if it needs allocation, we default to require it to be on the safe side

src/apprt/gtk/class/surface.zig
```diff
@@ -1511,19 +1517,17 @@ pub const Surface = extern struct {
...
+                break :xft_scale 1.0;
+            };
 
             // Use a value of 1.0 for the XFT DPI scale if the setting is <= 0
             // See:
             // https://gitlab.gnome.org/GNOME/libadwaita/-/commit/a7738a4d269bfdf4d8d5429ca73ccdd9b2450421
             // https://gitlab.gnome.org/GNOME/libadwaita/-/commit/9759d3fd81129608dd78116001928f2aed974ead
             if (gtk_xft_dpi <= 0) {
-                log.warn("gtk-xft-dpi was not set, using default value", .{});
+                log.warn("gtk-xft-dpi has invalid value ({}), using default", .{gtk_xft_dpi});
```
evertonstz, line 1528: I edited this log message to make it more precise!

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,145 @@
...
+        };
+    }
+
+    fn propertyName(comptime self: Key) [*:0]const u8 {
+        return switch (self) {
+            .gtk_enable_primary_paste => "gtk-enable-primary-paste",
+            .gtk_xft_dpi => "gtk-xft-dpi",
+            .gtk_font_name => "gtk-font-name",
+        };
+    }
```
evertonstz: I _think_ this could be done at comptime, but I am not sure its the best solution tho.

evertonstz:
@mitchellh I compiled and tested it here and it works fine on disabling the primary clibpboard based on the gsetting. I also did run some tests and the getter that needs allocation also worked fine for retrieving the font name :)

I believe I have made the api safe enough for interacting with the c gtk lib (and it was quite fun implementing it tbh, a lot of new concepts learned), but let me know if you see something too crazy here.

src/apprt/gtk/class/surface.zig
```diff
@@ -2685,6 +2696,13 @@ pub const Surface = extern struct {
 
         // Report the event
         const button = translateMouseButton(gesture.as(gtk.GestureSingle).getCurrentButton());
+
+        // Check if middle button paste should be disabled based on GNOME settings
+        // If gtk_enable_primary_paste is explicitly false, skip processing middle button
+        if (button == .middle and !(priv.gtk_enable_primary_paste orelse true)) {
+            return;
+        }
```
evertonstz: Maybe we could default this to false since its the new gnome behaviour?

src/apprt/gtk/gsettings.zig
pluiedev: I think a better solution would be to name the tags directly with hyphens. That way we don't have to manually convert the names and it's more evident that this is an external (e.g. not idiomatic Zig) name

```zig
pub const Key = enum {
  @"gtk-enable-primary-paste",
  // ...
}
```

Then we can just call `@tagName` to convert the variant into the correct string - no helper method required

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,144 @@
...
+///
+/// Example usage:
+///   const enabled = get(.gtk_enable_primary_paste);
+///   const dpi = get(.gtk_xft_dpi);
+pub fn get(comptime key: Key) ?key.Type() {
+    comptime {
+        if (key.requiresAllocation()) {
+            @compileError("Allocating types require an allocator; use getAlloc() instead");
+        }
+    }
```
pluiedev:
```suggestion
    if (comptime key.requiresAllocation()) {
        @compileError("Allocating types require an allocator; use getAlloc() instead");
    }
```

src/apprt/gtk/gsettings.zig
```diff
@@ -0,0 +1,144 @@
...
+/// Note: When adding a new type, research if it requires allocation (strings and boxed types do)
+/// if allocation is NOT needed, list it inside the switch statement in the function requiresAllocation()
+fn getImpl(settings: *gtk.Settings, allocator: ?std.mem.Allocator, comptime key: Key) !?key.Type() {
+    const GValType = key.GValueType();
+    var value = gobject.ext.Value.new(GValType);
+    defer value.unset();
+
+    settings.as(gobject.Object).getProperty(key.propertyName(), &value);
+
+    return switch (key) {
```
pluiedev: Dispatching this based on the type makes far more sense. Plus you then won't have to duplicate the switch arms in `GValueType`


```suggestion
    return switch (key.Type()) {
```

src/apprt/gtk/class/surface.zig
pluiedev: Not all Ghostty users use GNOME, and they *will* complain if a default has been changed behind their backs. My 2¢ would be that since GNOME already sets this GSetting we should be able to follow the new default there automatically

src/apprt/gtk/class/surface.zig
```diff
@@ -674,6 +675,11 @@ pub const Surface = extern struct {
         /// The context for this surface (window, tab, or split)
         context: apprt.surface.NewSurfaceContext = .window,
 
+        /// Whether primary paste (middle-click paste) is enabled via GNOME settings.
+        /// If null, the setting could not be read (non-GNOME system or schema missing).
+        /// If true, middle-click paste is enabled. If false, it's disabled.
+        gtk_enable_primary_paste: ?bool = null,
```
pluiedev: I think we should just set this to true unless overridden by GSettings. Then we won't have to do all that `orelse true` business

pluiedev (changes requested):

pluiedev:
Also sorry for presuming you're using AI - I've been fairly disappointed at the slew of slop PRs we've been getting recently, and the overly verbose comments and the general unfamiliarity with the language smelled off to me. I'll try to not immediately jump to conclusions next time ❤️

evertonstz:
> Also sorry for presuming you're using AI - I've been fairly disappointed at the slew of slop PRs we've been getting recently, and the overly verbose comments and the general unfamiliarity with the language smelled off to me. I'll try to not immediately jump to conclusions next time ❤️

I believe we all use agents for our day to day tasks and its a pretty valuable tool when used correctly. For example: my English is quite bad to the point its sometimes a little hard for others to understand, so I tend to use AI to correct that. I also don't mind using ai to scaffold boilerplate and other tedious stuff  when working in projects I own with tech I already know, so imo trying to ship generated code to production using a tech/language I don't fully understand and don't own is morally wrong tbh.

I took some time to take a look at the quality of the PRs being submitted here and you guys are basically babysitting vibe coders, I have been in the open source community since 2008 and the way things are going sadly I don't think we will have an open source community for too long, this is taking all the fun from it and basically making it more a second job :/

So no hardfeelings at all and thx for taking the time to review the code ❤️

src/apprt/gtk/class/surface.zig
evertonstz: Totally agree

src/apprt/gtk/gsettings.zig
evertonstz: TIL. It's like Zig have a cool solution for every problem.

evertonstz: @pluiedev question: one of the problems `propertyName` solved for us is that it already returned the c-style `[*:0]const u8` for us for free. Changing our approach to use `@tagName` to convert the variant to its string will give us a string that is typed as `[]const u8` and thus not necessarily null-terminated. 
The safest way I see to solve this problem is by using an allocator to convert between the types, and that would make every call to the get to actually need an allocator to be passed (so we would not need two functions, get and getAlloc)

I believe currently string literals in zig are already null terminated, so we could convert without using an allocator via something like `const variant_name: [*:0]const u8 = @tagName(.some_variant);`, but if something about this convention changes in future compiler versions, we would be in trouble.

WDYT? 🤔

pluiedev: Check the Zig language reference. [`@tagName`](https://ziglang.org/documentation/master/#tagName) is guaranteed to return a NUL-terminated slice:
```zig
@tagName(value: anytype) [:0]const u8
```
You can then get the pointer by doing `.ptr`

mitchellh (approved):
This looks good thank you. I think the only change to consider is querying the primary paste setting on every `gcMouseDown` so we can detect changes more easily. but I think this is good enough as is. Thank you.

evertonstz:
@mitchellh cool! I think I just need to implement the type based dispatch @pluiedev suggested, remove some unecessary comments and double check the docstrings and it will be ready for merge :)

mitchellh:
Real test failures in CI.

evertonstz:
@mitchellh gonna take a look in the error also. I can move it to draft so it wont consume CI minutes, if you think  its necessary.

PR #10337, opened 2026-01-15
PageList overflow detection and protection

### LINKED ISSUE #10258: Page.layout capacity overflow causes massive page allocations

mitchellh:
Fixes #10258  
Replaces #10284

1. `Page.Capacity` now uses smaller bit-width integers that represent a true maximum capacity for various fields.
2. On 64-bit systems, a maxed out `Page.Capacity` (every field `maxInt`) can be represented in an addressable allocation (total required memory less than 64 bits). This means `Page.layout` can't overflow.
3. All `adjustCapacity` functions replaced with `increaseCapacity` which doesn't allow specifying the resulting value, which makes it so overflow is only possible in significantly fewer places, making it easier to handle in general.
4. `increaseCapacity` can return a new error `OutOfSpace` which happens when overflow is detected. This means that no valid page can accommodate the desired capacity increase because we're already at the max. The caller is expected to handle this.
5. Updated our resize so that the only possible error is system OOM, we handle the new `OutOfSpace` by copying the recent reflowed row into a new page and continuing.

A very, very high-level overview is below. The "overflow" here papers over a bunch of details where the prior usize capacities flowed through to Page.layout and ultimately RefCountedSet and other managed types which then caused incorrect calculations on total memory size required.

```mermaid
flowchart TB
    subgraph Before["Before: adjustCapacity"]
        A1[capacity: usize] --> A2["capacity *= 2"]
        A2 --> A3{Overflow?}
        A3 -->|"Not detected"| A4["Massive allocation or crash"]
    end
    
    subgraph After["After: increaseCapacity"]
        B1["capacity: bounded int<br/>(u16/u32)"] --> B2["capacity *= 2"]
        B2 --> B3{Overflow?}
        B3 -->|"OutOfSpace error"| B4["Graceful handling:<br/>move row to new page"]
        B3 -->|"Success"| B5["Normal allocation"]
    end
    
    Before --> After
    
    classDef beforeStyle fill:#3d1a1a,stroke:#ff6b6b,color:#ff6b6b
    classDef afterStyle fill:#1a3d3a,stroke:#4ecdc4,color:#4ecdc4
    
    class A1,A2,A3,A4 beforeStyle
    class B1,B2,B3,B4,B5 afterStyle
```

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -1318,7 +1318,7 @@ const ReflowCursor = struct {
                     // Grow our capacity until we can
                     // definitely fit the extra bytes.
                     const required = cps.len * @sizeOf(u21);
-                    var new_grapheme_capacity: usize = cap.grapheme_bytes;
+                    var new_grapheme_capacity: size.GraphemeBytesInt = cap.grapheme_bytes;
                     while (new_grapheme_capacity - cap.grapheme_bytes < required) {
                         new_grapheme_capacity *= 2;
                     }
@@ -1362,7 +1362,7 @@ const ReflowCursor = struct {
                 } else |_| {
                     // Grow our capacity until we can
                     // definitely fit the extra bytes.
-                    var new_string_capacity: usize = cap.string_bytes;
+                    var new_string_capacity: size.StringBytesInt = cap.string_bytes;
                     while (new_string_capacity - cap.string_bytes < additional_required_string_capacity) {
                         new_string_capacity *= 2;
                     }
@@ -2647,16 +2647,16 @@ pub const AdjustCapacity = struct {
     /// Adjust the number of styles in the page. This may be
     /// rounded up if necessary to fit alignment requirements,
     /// but it will never be rounded down.
-    styles: ?usize = null,
+    styles: ?size.StyleCountInt = null,
 
     /// Adjust the number of available grapheme bytes in the page.
-    grapheme_bytes: ?usize = null,
+    grapheme_bytes: ?size.GraphemeBytesInt = null,
 
     /// Adjust the number of available hyperlink bytes in the page.
-    hyperlink_bytes: ?usize = null,
+    hyperlink_bytes: ?size.HyperlinkCountInt = null,
 
     /// Adjust the number of available string bytes in the page.
-    string_bytes: ?usize = null,
+    string_bytes: ?size.StringBytesInt = null,
 };
 
 pub const AdjustCapacityError = Allocator.Error || Page.CloneFromError;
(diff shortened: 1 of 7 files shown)


mitchellh:
I plan to follow up merging this with creating targeted issues to handle `OutOfSpace` in certain places I think make sense.

mitchellh:
I'm going to move this back to draft until I can shore up some more edge case handling (basically, handling OutOfSpace better).

mitchellh:
@qwerasd205 This is ready to go. We now handle OutOfSpace in the resize flow. I can't reproduce the overflow and more importantly with ReleaseSafe and Debug builds using `fps/tcolor` with dramatic resizing I can't trigger any safety checks or assertion failures anymore.

mitchellh:
I should add that I was NOT able to write a unit test for this despite being able to easily trigger it in the real world. I tried myself for an hour and also spent $15 on Opus 4.5 trying to write this unit test and it also failed to produce anything (the only part of this entire journey that used any AI -- none of the PR diff here uses any AI generated code). So, I gave up on that. ☹️

qwerasd205:
I can review this in a couple hours.

qwerasd205:
Ok, so first observation is the fact that because the max style capacity is now 32,768 it's very easy to encounter the scenario where we degrade and don't record styles because the page already has ~26,000 unique styles in it (because of the 80% load factor in the style set), just by printing a bunch of cells with unique styles, such as rendering a gradient square in a large terminal.

We really need to split the page when this is encountered while printing, right now it's not going to be encountered by most people but for some use cases it's very broken. (To be fair, at least it doesn't corrupt memory and crash)

mitchellh:
> We really need to split the page when this is encountered while printing, right now it's not going to be encountered by most people but for some use cases it's very broken. (To be fair, at least it doesn't corrupt memory and crash)

I agree, as a next step I recommend making an issue to add OutOfSpace handling to `setAttribute`. 

This isn't new though, right? I didn't change the capacity of the style set (except minus exactly 1). I only changed our type to match it (except minus 1). We previously would've just overflowed and weird stuff would've happened.

qwerasd205:
We could previously safely handle a capacity of 2^16, the max is now 2^15 since 2^16 cannot fit in a u16. The max capacity is now half of what it once was.

ldemailly:
uint16 should be 65535 not 32768 max (unless sign cast remain but then might as well use int16... which would be bad)

mitchellh:
Actually I ran into a different issue along the path here just now testing something else. Our pointer offsets are limited to `u32` so it isn't good enough that a maxed out Page is addressable in 64-bit memory space, we need to be addressable in 32-bit space. I don't want to increase our offset size cause that has a whole bunch of memory size implications.

This is significantly more restrictive and I think I need to rethink some approaches here. I think at a high level all the OutOfSpace stuff stays, we just need to trigger it in more cases.

qwerasd205:
> uint16 should be 65535 not 32768 max (unless sign cast remain but then might as well use int16... which would be bad)

We grow by multiplying by 2.
- 16
- 32
- 64
- 128
- 256
- 512
- 1024
- 2048
- 4096
- 8192
- 16384
- 32768
- 65536 <- **TOO BIG!**

mitchellh:
> We grow by multiplying by 2.

Yeah I was actually just about to hack in that if we overflow but aren't yet at max, we just go to max. I think that makes sense in every case so we can take advantage of all the bits. That'll get us to ~52K. That sound right?

ldemailly:
> We grow by multiplying by 2.
> 32768
> 65536 <- TOO BIG!

ic that explains, thanks

you could use 1.5 after some size, that's what some allocators do for bigger chunks - or special case the last one and use max_uint16 when the resize would go over

mitchellh:
@qwerasd205 Thumbs up? https://github.com/ghostty-org/ghostty/pull/10337/changes/85a3d623b2f3452a49eb634c10e46ba6746d0cd6

mitchellh:
Aside, the fact that our capacity overflow can only happen in _one place_ now made that fix up commit really easy. Even introducing failable layouts for the max page size case, we should be able to fit that into `increaseCapacity` and everyone will inherit and handle OutOfSpace...

qwerasd205 (approved):
Didn't do a super thorough review of the code but the changes seem good to me, and like an improvement over the status quo, especially with being able to reach maxint capacities when growing again.

mitchellh:
Thanks, before merging I'm going to do a bit of research tonight on how much more work it is add an overflow possibility to Page.layout for `max_page_size`. I'm not aware of anyone ever hitting that in practice, but since I'm context loaded here I should take a look.

mitchellh:
The max page size handling is a mess. I'm going to create an issue to do it in a more robust way, but I did add some mitigations/checks:

* Verified that no amount of columns above our standard size exceeds `max_page_size`. This means all our `initialCapacity` calls (the majority of create page sources) can never overflow this way.
* Updated `increaseCapacity` to verify the capacity and return `error.OutOfSpace` if `max_page_size` is hit, which should address _most_ of our non-standard pages.
* Added an assertion in `createPage` for safe builds to catch any others so we know it exists

What _is not mitigated_:

* Resizes that grow columns (with or without reflow) can theoretically push us over the limit. 

Given this hasn't happened in history that we know of, it isn't new, and the huge benefits of this PR, I'm going to summarize this in a follow-up issue along with the `setAttribute` handling.

ldemailly:
Please cc me or link the new issue to this one as it still crashes (see discord for details or I can add them somewhere)

PR #10341, opened 2026-01-16
fix(zsh): strip control characters from window title

Newspicel:
## Summary
- Fix literal \n appearing in window titles when running commands in zsh

## Description
The zsh shell integration was using ${(V)1} parameter expansion to set the window title. The (V) flag converts control characters to their visible escape sequence representations, causing commands ending with a newline to display as command\n in the title bar.

Changed to use ${1//[[:cntrl:]]} which strips control characters entirely, matching the behavior of the bash integration.

image

## Test

  - Open Ghostty with zsh shell integration enabled
  - Run a command (e.g., cargo install trunk)

  ---
  AI Disclosure: This PR was written primarily by Claude Code. (Opus 4.5)


jparise (approved):
Thanks! This looks good and matches what we're doing in bash (as you noted).

Issue #10345, opened 2026-01-16
copy_title_to_clipboard copies original terminal title instead of manually renamed title


pluiedev:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/10344

<div type='discussions-op-text'>

<sup>Originally posted by **syy** January 16, 2026</sup>
### Issue Description

When using `copy_title_to_clipboard` action after renaming a tab with "Change Tab Title..." on macOS, the original terminal title is copied instead of the user-set renamed title.

The tab title renaming feature was added in February 2025 (#4217), and `copy_title_to_clipboard` was added in July 2025 (#7833). However, the copy action doesn't check for manually renamed titles. It only copies the original terminal title.

### Expected Behavior

When I rename a tab to "My Custom Tab" and then use copy_title_to_clipboard, the clipboard should contain "My Custom Tab".

### Actual Behavior

The clipboard contains the original terminal title (e.g., shell process name or current directory) instead of the renamed title.

### Reproduction Steps

1. Open Ghostty
2. Open Command Palette (Cmd+Shift+P or configured keybinding)
3. Search and select "Change Tab Title"
4. Enter a custom name like "My Custom Tab" and click OK
5. Open Command Palette again
6. Search and select "Copy Title to Clipboard"
7. Paste from clipboard (Cmd+V)
8. Observe: clipboard contains the original terminal title

### Ghostty Version

```text
Ghostty 1.2.3

Version
  - version: 1.2.3
  - channel: stable
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.none
  - font engine   : font.main.Backend.coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS 26.2 (Build 25C56)

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
bold-color = bright
shell-integration = zsh
split-divider-color = #282c34
theme = iTerm2 Dark Background
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
[...]

PR #10347, opened 2026-01-16
macOS: copy_title_to_clipboard respects user-set title override

syy:
## Summary
  - When using "Change Terminal Title..." to set a custom title, `copy_title_to_clipboard` now copies the user-set title instead of
  the original terminal title
  - Adds `getDisplayTitle()` method that returns `title_override` if set, otherwise falls back to the terminal-set title

  ## Test plan
  - [x] Set custom title via "Change Terminal Title..."
  - [x] Use "Copy Terminal Title to Clipboard" → copies custom title
  - [x] Clear custom title (leave empty)
  - [x] Use "Copy Terminal Title to Clipboard" → copies original terminal title

  **Note:** GTK implementation is not included per CONTRIBUTING.md guidelines (macOS-only machine).

AI Disclosure: This PR was written with Claude Code assistance. I tested all changes on macOS.

  Ref #10345

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1104,6 +1104,8 @@ bool ghostty_surface_read_text(ghostty_surface_t,
                                ghostty_selection_s,
                                ghostty_text_s*);
 void ghostty_surface_free_text(ghostty_surface_t, ghostty_text_s*);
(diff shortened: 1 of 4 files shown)


mitchellh:
Thank you but this isn't the right approach. I'll add more details to the issue.

Issue #10352, opened 2026-01-17
`fps -fire` with very large grid can cause safety failure


mitchellh:
Once we fixed #10258 another, different problem has emerged.

Reproduction

1. `fps -fire`
2. Make font very very small (might need to make it larger and then smaller again)

```
(lldb) bt
* thread #10, name = 'io', stop reason = signal SIGABRT
    frame #0: 0x00000001999935b0 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x0000000104a9e660 libsystem_pthread.dylib`pthread_kill + 296
    frame #2: 0x00000001998d2850 libsystem_c.dylib`abort + 124
    frame #3: 0x000000010c3ae178 ghostty.debug.dylib`posix.abort at posix.zig:732:29
    frame #4: 0x000000010c3adbd4 ghostty.debug.dylib`debug.defaultPanic(msg=(ptr = "reached unreachable code", len = 24), first_trace_addr=(data = 4490287752, some = '\x01')) at debug.zig:722:16
    frame #5: 0x000000010c3ae12c ghostty.debug.dylib`debug.FullPanic((function 'defaultPanic')).reachedUnreachable at debug.zig:64:17
    frame #6: 0x000000010ba45a88 ghostty.debug.dylib`debug.assert(ok=false) at debug.zig:559:14
  * frame #7: 0x000000010c13a820 ghostty.debug.dylib`terminal.PageList.resizeWithoutReflow(self=0x0000000122da1010, opts=terminal.PageList.Resize @ 0x000000016f9efe38) at PageList.zig:2062:19
    frame #8: 0x000000010c115654 ghostty.debug.dylib`terminal.PageList.resize(self=0x0000000122da1010, opts=terminal.PageList.Resize @ 0x000000016f9efe38) at PageList.zig:921:41
    frame #9: 0x000000010c114a18 ghostty.debug.dylib`terminal.Screen.resizeInternal(self=0x0000000122da1000, cols=3016, rows=582, reflow=true) at Screen.zig:1724:26
    frame #10: 0x000000010c115c90 ghostty.debug.dylib`terminal.Screen.resize at Screen.zig:1650:28
    frame #11: 0x000000010c115c70 ghostty.debug.dylib`terminal.Terminal.resize(self=0x0000000122cf2e68, alloc=mem.Allocator @ 0x0000000122cf2910, cols=3016, rows=582) at Terminal.zig:2563:27
[...]

PR #10354, opened 2026-01-17
renderer: Metal and OpenGL shrink buffers when they're much smaller

mitchellh:
Fixes the memory associated with #10352

Presently, our GPU buffers never shrink, they only grow. If the grid size grows very large, we have to allocate very large GPU buffers.

This commit introduces a little heuristic to shrink buffers when the size we're using is less than a quarter of the total size.

I don't love this so #10355 for the proper long term solution.


mitchellh:
Going to keep this open. Might just do the proper long term fix in #10355 instead since this is a pretty edge case scenario.

mitchellh:
Don't like this, will look at a better long term fix.

Issue #10366, opened 2026-01-18
Configurable preedit style (invert vs underline)


pluiedev:
See #9235

When using IMEs to type languages like Chinese and Japanese, the frequent flashing that comes with the preedit field could become quite uncomfortable when writing long texts. While inverting seems to be the norm for terminals, for GUI applications on both macOS and Linux underlining is far more common. It would probably be a good idea to allow users to use underlines instead if that suits them better.

mitchellh:
I'm open to not making this a config option and just changing our pre-edit to be underline styled.

pluiedev:
I'm okay with that - maybe there'll be *someone* who's going to be a pain about this but I'm not too horribly worried

pluiedev:
Not sure why this wasn't autoclosed

PR #10368, opened 2026-01-18
Use underline for preedit texts instead of inverting colors

shivaduke28:
issue: https://github.com/ghostty-org/ghostty/issues/10366

Presently, IME preediting texts are rendered with inverted colors styles. This pull request changes that style to adding underlines.

Before
image

After
image

If it is preferred that adding configure like `ime-preedit-text-style` with default value `invert` to keep the current behaviour as is, I can add such a config (see https://github.com/shivaduke28/ghostty/pull/1 ).

Thanks!

Disclosure: This PR was written primarily by Claude Code.


jcollie (approved):
LGTM

PR #10376, opened 2026-01-19
splits: make resize_split and toggle_split_zoom non-performable with single pane

unphased:
Refer to discussion #10000 

When a tab contains only a single split, resize_split and toggle_split_zoom actions now return false (not performed). This allows keybindings marked with `performable: true` to pass the event through to the terminal program.

The performable flag causes unperformed actions to be treated as if the binding didn't exist, so the key event is sent to the terminal instead of being consumed.

- Add isSplit() helper to SplitTree to detect single-pane vs split state
- Update GTK resizeSplit/toggleSplitZoom to return false when single pane
- Update macOS resizeSplit/toggleSplitZoom to return Bool and check isSplit
- Add unit test for isSplit method

### DIFF
--- a/src/datastruct/split_tree.zig
+++ b/src/datastruct/split_tree.zig
@@ -794,6 +794,7 @@ pub fn SplitTree(comptime V: type) type {
             layout: Split.Layout,
             ratio: f16,
         ) Allocator.Error!Self {
+            assert(ratio >= -1 and ratio <= 1);
             assert(!std.math.isNan(ratio));
             assert(!std.math.isInf(ratio));
 
--- a/src/datastruct/split_tree.zig
+++ b/src/datastruct/split_tree.zig
@@ -1356,12 +1356,14 @@ test "SplitTree: isSplit" {
 
     // Split tree should be split
     var v2: TestView = .{ .label = "B" };
+    var tree2: TestTree = try TestTree.init(alloc, &v2);
+    defer tree2.deinit();
     var split = try single.split(
         alloc,
         .root,
         .right,
         0.5,
-        &v2,
+        &tree2,
     );
     defer split.deinit();
     try testing.expect(split.isSplit());
--- a/src/datastruct/split_tree.zig
+++ b/src/datastruct/split_tree.zig
@@ -170,6 +170,19 @@ pub fn SplitTree(comptime V: type) type {
             return self.nodes.len == 0;
         }
 
+        /// Returns true if this tree has more than one split (i.e., the root
+        /// is a split node). This is useful for determining if actions like
+        /// resize_split or toggle_split_zoom are performable.
+        pub fn isSplit(self: *const Self) bool {
+            // An empty tree is not split.
+            if (self.isEmpty()) return false;
+            // The root node is at index 0. If it's a split, we have multiple splits.
+            return switch (self.nodes[0]) {
+                .split => true,
+                .leaf => false,
+            };
+        }
+
         /// An iterator over all the views in the tree.
         pub fn iterator(
             self: *const Self,
@@ -1325,6 +1338,34 @@ const TestView = struct {
         return self.label;
     }
 };
+
+test "SplitTree: isSplit" {
+    const testing = std.testing;
+    const alloc = testing.allocator;
+
+    // Empty tree should not be split
+    var empty: TestTree = .empty;
+    defer empty.deinit();
+    try testing.expect(!empty.isSplit());
+
+    // Single node tree should not be split
+    var v1: TestView = .{ .label = "A" };
+    var single: TestTree = try TestTree.init(alloc, &v1);
+    defer single.deinit();
+    try testing.expect(!single.isSplit());
+
+    // Split tree should be split
+    var v2: TestView = .{ .label = "B" };
+    var split = try single.split(
+        alloc,
+        .root,
+        .right,
+        0.5,
+        &v2,
+    );
+    defer split.deinit();
+    try testing.expect(split.isSplit());
+}
 
 test "SplitTree: empty tree" {
     const testing = std.testing;
--- a/src/datastruct/split_tree.zig
+++ b/src/datastruct/split_tree.zig
@@ -773,9 +773,9 @@ pub fn SplitTree(comptime V: type) type {
         /// Resize the nearest split matching the layout by the given ratio.
         /// Positive is right and down.
         ///
-        /// The ratio is a value between 0 and 1 representing the percentage
-        /// to move the divider in the given direction. The percentage is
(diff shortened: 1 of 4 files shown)


jcollie (approved):
GTK changes LGTM

mitchellh (approved):

mitchellh:
Approved but the failing CI is real and needs to be fixed before we can merge.

unphased:
Thank you for the review. I am able to produce a failing `zig build test` with 

```
Build Summary: 54/82 steps succeeded; 1 failed
test transitive failure
└─ run test ghostty-test transitive failure
   └─ compile test ghostty-test Debug native-native transitive failure
      ├─ run glib-compile-resources (ghostty_resources.h) transitive failure
      │  └─ run exe generate_gresource_xml transitive failure
      │     ├─ run exe gtk_blueprint_compiler (1.0/clipboard-confirmation-dialog.ui) transitive failure
      │     │  └─ compile exe gtk_blueprint_compiler Debug native failure
```

Unfortunately I have no idea what this means, it looks vaguely plausible as the same issue seen in CI.

If i can fix the build locally I will push an update. If the implementation code changes dramatically, I may make it as a separate PR.

mitchellh:
```
+- run test ghostty-test
   +- compile test ghostty-test Debug native-native 1 errors
src/datastruct/split_tree.zig:1364:9: error: expected type '*const datastruct.split_tree.SplitTree(datastruct.split_tree.TestView)', found '*datastruct.split_tree.TestView'
        &v2,
        ^~~
src/datastruct/split_tree.zig:1364:9: note: pointer type child 'datastruct.split_tree.TestView' cannot cast into pointer type child 'datastruct.split_tree.SplitTree(datastruct.split_tree.TestView)'
src/datastruct/split_tree.zig:1322:18: note: struct declared here
const TestView = struct {
                 ^~~~~~
src/datastruct/split_tree.zig:45:12: note: struct declared here
    return struct {
           ^~~~~~
src/datastruct/split_tree.zig:471:21: note: parameter type declared here
            insert: *const Self,
```

This is the real issue, you're seeing the transitive issue.

unphased:
Able to reproduce the same compilation error locally! thanks.

unphased:
I have a fix forthcoming that yields this from `zig build test`: 

```
slu@slu-ws ~/ghostty discussion_10000 $ zig build test                                   01/20 0:46 ↵ 1
test
└─ run test ghostty-test stderr
Requested executable not found. Please verify the command is on
the PATH and try again.
[terminal] (warn): zero-width character with no prior character, ignoring
[page] (warn): page integrity violation y=0 grapheme data but row not marked
[page] (warn): page integrity violation style ref count mismatch id=1 expected=10 actual=9
[page] (warn): page integrity violation zero row count
[page] (warn): page integrity violation zero col count
```
and so on. 

It seems to mean that the build and test passed?

unphased:
Made it one step farther in CI...

unphased:
~~Seems like the CI workflows need approval~~

unphased:
from testing the local build with a `performable:` bind for this resize feature it is crashing, so i am working through it...

src/datastruct/split_tree.zig
```diff
@@ -773,9 +773,9 @@ pub fn SplitTree(comptime V: type) type {
         /// Resize the nearest split matching the layout by the given ratio.
         /// Positive is right and down.
         ///
-        /// The ratio is a value between 0 and 1 representing the percentage
-        /// to move the divider in the given direction. The percentage is
-        /// of the entire grid size, not just the specific split size.
+        /// The ratio is a signed delta representing the percentage to move
+        /// the divider. The percentage is of the entire grid size, not just
+        /// the specific split size.
```
unphased, line 778: Could I ask for a second set of eyes on this? Anyone know which are the true semantics of this ratio value?

unphased:
I learned how to build a release build with zig so I am now dogfooding this branch. Let you know if I have issues. I mainly drive tmux for the important terminal sessions so i can hop on from my phone or other machines. So the performable-ness of resize and maximize has a big UX impact.

src/datastruct/split_tree.zig
```diff
@@ -781,7 +794,6 @@ pub fn SplitTree(comptime V: type) type {
             layout: Split.Layout,
             ratio: f16,
         ) Allocator.Error!Self {
-            assert(ratio >= 0 and ratio <= 1);
```
mitchellh, line 784: This looks suspicious. Is this related to making it performable? If not, I'd remove this and open a discussion to discuss this change otherwise.

mitchellh (changes requested):

src/datastruct/split_tree.zig
unphased: I find this change concerning as well, but it seemed like this was what was causing crashing on the build when i performed pane resizes. I will investigate this area some more.

unphased: could you fire off the CI checks though now that you have requested changes, so we can see if CI is green? I don't think it is feasible for me to replicate the full tests locally (thanks!)

mitchellh: Yes, I'll run CI. but we should still plan to remove this.

mitchellh: Note if you are seeing crashes here and this assertion is wrong, I'd rather see another PR with new tests that demonstrate that it works fine for other cases so we can tighten our assertion rather than remove it entirely.

unphased: We can see for gtk code at least that the ratio value is clearly being assigned a negative value and as such is a ratio delta, not a ratio that must be [0,1]. As a delta to apply to a ratio, it probably should be constrained within (-1,1), so perhaps the assertion could be loosened just to include negative values.  

https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/split_tree.zig#L295

unphased:
I also did a test; I added a info level log of the ratio variable value and watched it as i launched a new instance and resized a pane after making a split. it indeed shows a negative floating point value when resizing upward, so the assertion that was present in the code was incorrect by asserting a positive value.

src/datastruct/split_tree.zig
unphased: @mitchellh do you think this is clear enough or would you like to investigate the topic more?

mitchellh (approved):
Thanks, yeah I'm good with it I just wanted a test to verify it.

unphased:
That CI suite is a sight to behold.

mitchellh:
Haha it catches many things!

unphased:
Living the dream! Thank you!

PR #10377, opened 2026-01-19
terminal: remap tracked pins in backfill optimization path during resize

mitchellh:
Fixes #10369

When `resizeWithoutReflowGrowCols` copies rows to a previous page with spare capacity, tracked pins pointing to those rows were not being remapped.  This left pins pointing to the original page which was subsequently destroyed.

The fix adds pin remapping for rows copied to the previous page, matching the existing remapping logic for rows copied to new pages.

I also added new integrity checks to verify that our tracked pins are always valid at points where internal operations complete.

Thanks to @grishy for finding this!

**AI disclosure:** Amp used for verifying and fixing this bug. I reviewed the results and just did minor manual tweaks. https://ampcode.com/threads/T-019bd6d7-0645-73dd-8fd7-659f019fa83d and https://ampcode.com/threads/T-019bd6c9-cc2e-73bc-bbaa-f8766e11c234

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -485,10 +485,11 @@ pub inline fn pauseIntegrityChecks(self: *PageList, pause: bool) void {
 }
 
 const IntegrityError = error{
+    PageSerialInvalid,
     TotalRowsMismatch,
+    TrackedPinInvalid,
     ViewportPinOffsetMismatch,
     ViewportPinInsufficientRows,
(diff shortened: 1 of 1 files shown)


mitchellh:
Looks like that integrity check is finding new failures, yay! 😄 working on it now.

mitchellh:
I've got the tests fixed locally but want to review them further before pushing it up.

PR #10389, opened 2026-01-20
build: libghostty-vt now depends on uucode directly

mitchellh:
This doesn't add any real weight to it, we only need it for a type definition. This fixes our example builds.

https://ampcode.com/threads/T-019bdc95-07bb-70c9-9db4-1218d41decb2


Seedmanc:
>AIphobia
Typical luddite hypocrites calling other bots but acting like NPCs bandwagoning AI hate. Pointing out vibecoding faults is fun, but when AI does the same to your code your kind turns off reports instead of fixing bugs. It's like posting codeofconduct when hating on bigots was hip.

PR #10415, opened 2026-01-23
fix(nushell): expand tildes in sudo wrapper arguments

andrewgazelka:
## Summary

- Fix tilde expansion in nushell shell integration's sudo wrapper
- Arguments starting with `~` are now properly expanded before passing to sudo

## Problem

Nushell's `--wrapped` functions receive arguments as raw strings without tilde expansion. This caused:

```nushell
❯ sudo echo ~
~
```

Instead of the expected:

```nushell
❯ sudo echo ~
/Users/username
```

Using `^sudo` (bypassing the wrapper) worked correctly, confirming the issue was in the wrapper.

## Solution

Explicitly expand tildes for arguments starting with `~` using `path expand` before passing to sudo. This is the idiomatic nushell approach per the nu-path crate's tilde expansion logic.

Only arguments starting with `~` are expanded to avoid changing the behavior of relative paths like `./file` or flags.

## Test plan

- [ ] `sudo echo ~` outputs home directory path
- [ ] `sudo echo ~/foo` outputs expanded path
- [ ] `sudo ls -la` still works (flags unchanged)
- [ ] `sudo cat ./file` still works (relative paths unchanged)

### DIFF
--- a/src/shell-integration/nushell/vendor/autoload/ghostty.nu
+++ b/src/shell-integration/nushell/vendor/autoload/ghostty.nu
@@ -8,17 +8,24 @@ export module ghostty {
   export def --wrapped sudo [
     ...args  # Arguments to pass to `sudo`
   ] {
-    mut sudo_args = $args
+    # Expand tildes in arguments (nushell doesn't expand ~ in wrapped function args)
+    mut sudo_args = ($args | each {|arg|
+        if ($arg | str starts-with "~") {
+            $arg | path expand
+        } else {
+            $arg
+        }
+    })
 
     if (has_feature "sudo") {
         # Extract just the sudo options (before the command)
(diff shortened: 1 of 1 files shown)


mitchellh:
Likely undisclosed AI, no associated issue, closing according to our new AI policy. Please open a discussion.

andrewgazelka:
> Likely undisclosed AI, no associated issue, closing according to our new AI policy. Please open a discussion.

hey @mitchellh I purposefully marked as a draft so I could work on following contribution guidelines before you saw it. Claude auto made PRs (which is standard in my workflow)... please give me a second chance.

mitchellh:
Understood, please make the PR in your fork. You can reference it from a discussion as noted in the policy. We don't want Claude littering our PR inbox and emails (it still notifies us even for drafts!).

PR #10425, opened 2026-01-23
gtk: add read-only indicator for surfaces

### LINKED ISSUE #9889: GTK: Add read-only surface GUI indication

jcollie:
Fixes: #9889

image


mitchellh (approved):

mitchellh:
I think we can bike shed on look here if anyone wants but I'm happy to have _anything_ here for now. 😄

jcollie:
The look is mostly CSS so it's easy to bikeshed, or even for users to override via `gtk-custom-css`.

PR #10427, opened 2026-01-23
Rewrite semantic prompt parsing, parse the full spec

mitchellh:
Related to #5932 

This updates our OSC parser to parse the full OSC 133 specification: https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md

The logic for handling these events was _unchanged_ from our prior implementation. This is just a parser-only update. As such, we ignore a bunch of semantic prompt command we should definitely handle, and incorrectly handle others. This is the crux of #5932 that I want to head towards fixing. This PR just contains the parser updates.

I also retained all the Kitty parser extensions.

**AI disclosure:** AI helped a lot of the rote tasks once I manually did a few. I'm still reviewing this manually but will do so shortly.


jcollie (approved):
LGTM

jparise (approved):
Good stuff. Thanks for pushing this forward.

I have an older branch that reworked the Screen-level representation here: https://github.com/jparise/ghostty/tree/semantic

... but it predates the last 1.5 rounds of PageList evolution, and it taught me enough about the problem that I remember wanting to revisit the decision to just "paint" each cell with a semantic type vs using a higher-level "Semantic Region" concept (perhaps still based on the underlying cell data, or if Pins are more scalable now, maybe using them).

mitchellh:
> I have an older branch that reworked the Screen-level representation here: https://github.com/jparise/ghostty/tree/semantic

Yeah, I'm still thinking about this generally. There are pros and cons on both sides and I'm trying to weigh future functionality to really determine which would better serve us for the future, too.

PR #10428, opened 2026-01-23
core/gtk: add language config entry to override GUI localization

### LINKED ISSUE #10276: GTK: Add the locale as a config setting
This has been requested a few times now at least as far as I know (including in person, fun fact), but a recent discussion can be found at #10249.

We should allow the locale to be specified as a config setting, for users who e.g. might want to make their whole system default to one language (say English) and only enable localization for apps that support it. I never really thought about this use case before, mostly because I just stick to either English or Chinese the entire time, but this does make sense.

Something like this would do (I debated over whether to call this `locale` or `language`, but I think for non-native speakers of English "language" is a far more approachable term than "locale", even if it is the technical term):
```ini
language = zh_CN
```

~~Implementation-wise the only tricky spot is in implementing the hot reload mechanism, which on gettext/libintl is [*very* cursed](https://www.gnu.org/software/gettext/manual/gettext.html#Being-a-gettext-grok). Still, it's a well-documented hack that we should absolutely make use of.~~

EDIT: Looks like [GTK does not like it when locales are changed on the fly](https://discourse.gnome.org/t/is-it-possible-dynamicaly-change-locales-in-gtk-gtkmm-application/27753/2). We could make this non-hot-reloadable then.

GTK only for now. We can edit or recreate this issue when macOS receives localization support as well

jcollie:
Fixes #10276

Screenshot From 2026-01-23 16-45-11

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -213,6 +213,11 @@ pub const Application = extern struct {
         /// Providers for loading custom stylesheets defined by user
         custom_css_providers: std.ArrayListUnmanaged(*gtk.CssProvider) = .empty,
 
+        /// A copy of the LANG environment variable that was provided to Ghostty
+        /// by the system. If this is null, the LANG environment variable did
+        /// not exist in Ghostty's environment variable.
+        saved_language: ?[:0]const u8 = null,
+
         pub var offset: c_int = 0;
     };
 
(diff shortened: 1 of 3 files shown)


jcollie:
I'm not 100% sure that this should be merged as it basically prevents us from localizing any error messages that originate from config parsing. Basically because we have to parse the config file to see if the language has been overridden before we can initialize the localization engine.

Also not sure if the original LANG environment should be saved/restored. If we don't then any programs run within Ghostty also change languages.

pluiedev (approved):
I think it's worth to have this setting now as we don't really have parsing errors localized yet. For the long run we should probably use some sort of internal error enum that represents all failure cases instead of pushing raw strings into the config diagnostics

We should definitely preserve LANG for the child environment, since `language` is only supposed to override the *interface language* independent from the apps within.

PR #10431, opened 2026-01-24
Fix IME Preedit characters getting thicker when idle

shivaduke28:
Fixes https://github.com/ghostty-org/ghostty/issues/10424

When preedit text is marked as dirty by user input,
it is the case of "full rebuild" and there is no problem.

When it is not the full rebuild case, only the dirty rows are cleared,
but preedit text is always rendered additively.
This causes preedit text to get thicker when idle.

To avoid this, I changed preedit text to be rendered only during full rebuild or when its row is dirty.

I tested on macOS.

Before
[Image from Gyazo](https://gyazo.com/5775fda5466abc6bff812f8426dedef9)

After
[Image from Gyazo](https://gyazo.com/643b62e9702daeda5fd00dc9d96c0c86)

AI disclosure: I used Claude Code to investigate the source code and generate code changes in this PR.

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -2629,6 +2629,12 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
             const row_selection = row_data.items(.selection);
             const row_highlights = row_data.items(.highlights);
 
+            // Get preedit range only if its row is dirty or full rebuild.
+            const dirty_preedit_range = if (preedit_range) |range|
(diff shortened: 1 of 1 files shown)


mitchellh:
Oh they're getting thicker because they're being drawn over multiple times? lol. I want to review this change more carefully but the root cause makes sense.

mitchellh:
Thanks, replaced by #10772. Mostly the same logic, but better Zig since agents don't utilize named blocks well.

PR #10433, opened 2026-01-24
pkg: update freetype to 2.14.1

jcollie:


jcollie:
Did this update while investigating #10432 but did not fix the segafult, but it may be worthwhile to update anyway.

mitchellh (changes requested):
Approved but I want to update the mirror to our host before merging.

Did you review the version diff by any chance? I usually do that prior to updating our dependencies to make sure no APIs we touch are at least documented to be touched, and also to verify their build scripts didn't change in ways we may care about (e.g. new C defines).

PR #10435, opened 2026-01-24
pkg: update fontconfig to 2.17.1

jcollie:


jcollie:
Created this PR while investigating #10432 but it doesn't seem to fix the problem, and has test failures when running the packages' tests with `cd pkg/fontconfig; zig build test` so leaving as draft for now.

PR #10436, opened 2026-01-24
chore: Add environment variable expansion for background image path

paulchen5:
I am completely honest with you. I have no clue about Zig. I just told Copilot to add the feature to expand environment variables in a path / string and gave him the reference for the `background-image` property.

I think it's kinda nice to have this feature / it would help me with my setup.


mitchellh:
Thank you for the honesty. Our AI policy requires a discussion for AI PRs that have no accepted issues. Please do that and attach a link to your branch there.

PR #10442, opened 2026-01-24
terminal: OSC133 options parse from raw string lazily

mitchellh:
This changes our OSC133 parser to parse options lazily. We do this for multiple reasons:

1. Parsing all our options ahead of time balloons our required osc.Command tagged union type which has C ABI implications. Adding all supported options (including Kitty extensions) today already breaks our C ABI.

2. Invalid options are allowed by the specification and should be  explicitly ignored, so we don't need to validate options at all during parse time.

3. Semantic prompt markers don't need to be high throughput, so we can afford to do some extra work at processing time to gather the options. They're also rather short usually.

cc @jparise

### DIFF
--- a/src/terminal/osc/parsers/semantic_prompt.zig
+++ b/src/terminal/osc/parsers/semantic_prompt.zig
@@ -13,7 +13,7 @@ const log = std.log.scoped(.osc_semantic_prompt);
 /// options. So, I think this is a fair interpretation.
 pub const Command = struct {
     action: Action,
-    options: Options,
+    options_unvalidated: []const u8,
 
     pub const Action = enum {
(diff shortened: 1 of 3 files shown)


mitchellh:
Also cc @jcollie and @pluiedev since this may be a good pattern for other OSC types in the future with a lot of options. If validation is important you can always run a single pass during validation. We know that most OSCs are NOT common so parsing performance isn't _critical_ and this doesn't cost much more anyways (context dependent).

This reminds me of the Zig compiler pattern that Andrew notes where errors are rare, so they just reparse if there is a Sema error to produce the error message rather than carrying around all that context forever.

PR #10454, opened 2026-01-26
build: don't allow `/` in branch names

jcollie:
This should fix CI failures like in PRs #10449 and #10450 that use long automatically-generated branch names.

### DIFF
--- a/src/build/GitVersion.zig
+++ b/src/build/GitVersion.zig
@@ -29,6 +29,10 @@ pub fn detect(b: *std.Build) !Version {
             error.ExitCodeFailure => return error.GitNotRepository,
             else => return err,
         };
+        // Replace any '/' with '-' as including slashes will mess up building
+        // the dist tarball - the tarball uses the branch as part of the
+        // name and including slashes means that the tarball will end up in
+        // subdirectories instead of where it's supposed to be.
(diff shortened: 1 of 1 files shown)


src/build/GitVersion.zig
```diff
@@ -19,14 +19,18 @@ branch: []const u8,
...
+        const tmp: []u8 = b.runAllowFail(
+            &[_][]const u8{ "git", "-C", b.build_root.path orelse ".", "rev-parse", "--abbrev-ref", "HEAD" },
+            &code,
+            .Ignore,
+        ) catch |err| switch (err) {
+            error.FileNotFound => return error.GitNotFound,
+            error.ExitCodeFailure => return error.GitNotRepository,
+            else => return err,
+        };
+        std.mem.replaceScalar(u8, tmp, '/', '-');
```
mitchellh, line 36: Can you add a comment here just explaining this. Its obvious with the PR context but I suspect years down the road will be less obvious what this one line is helping with.

mitchellh (changes requested):
One small request but seems fine otherwise

src/build/GitVersion.zig
jcollie: Done.

mitchellh (approved):

PR #10455, opened 2026-01-26
Rewritten semantic prompt state management (OSC 133)

### LINKED ISSUE #5932: OSC 133: Support semantic prompt regions

mitchellh:
Fixes #5932 

This updates our core terminal internals to track [semantic prompts (OSC133)](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md) in a way that can enable us to fully implement the specification. This PR makes our terminal state fully handle all specified sequences properly, but we still ignore some noted options (`aid`, `cl`, etc.) that are not critical to functionality.

I took a look at our shell integrations and they should still be accurate and correct with regards to their OSC133 usage.

This doesn't introduce any new functionality, but should result in existing functionality being more correct, more robust, and gives us the groundwork to do more with semantic prompts in the future.

> [!NOTE]
>
> This PR description is **fully human written.** I realize its layout and use of headers may trigger AI smells. But it's all me, I promise. :)

## Extensions

I kept support for the Kitty `redraw=0` extension: if an OSC133A command sets `redraw=0` then we will NOT clear the prompt on resize (assuming the shell can NOT redraw the prompt). I don't actually know any scripts that utilize this, so we may want to just throw it away, but we had support before and it's cheap to keep around and it's tested.

## Limitations, Missing Features

* We don't currently respect `aid` so nested semantic zones can break things. This would require significant rework since aids are arbitrary strings. It'd require managed memory on pages to store the aids. Fwiw, in my brief survey of popular mainstream terminals, I wasn't able to find any that respect this. Maybe Warp does since semantic prompts are so core to their product but its not OSS for me to verify.

### `terminal.Page`

The internal state now tracks the current _semantic content type_ via a 2-bit flag on each `Cell`:

* `output` (zero value) - Command output
* `input` - Prompt input, typed by the user
* `prompt` - Prompt itself, e.g. `user@host $`

In addition to this, each `Row` also has a 2-bit flag. This flag _only tracks if the row might contain a prompt_:

* `no_prompt` (zero value) - No prompt cells
* `prompt` - Prompt cells exist and this started the prompt (OSC133 A)
* `prompt_continuation` - Prompt cells exist and this continued a prior prompt (OSC 133 A `k=s` or `k=c`)

The row flags server two functions:

1. An optimization to make prompt scanning faster
2. The only place the existence of continuation vs primary prompt is stored since cells themselves only store `prompt`

### `terminal.PageList`

The PageList has two new functions:

* `promptIterator` - An iterator that yields `Pin` values that map to the _first line_ of a prompt.
* `highlightSemanticContent` - Returns a highlight (pin range) to map to specific semantic content for a given prompt pin. This lets you quickly grab stuff like the prompt lines, content lines, etc.

### `terminal.Screen`

* The Screen now has a flag that tracks whether we've seen any row trigger a semantic prompt. This can be used as a fast path to avoid expensive semantic prompt operations if we know our screen can't possibly contain any. We don't currently use this but I plan to.

* **Big one: clearing the prompt lines is now built-in to resize.** This makes our resize aware of semantic prompts which makes it work _so much better_ already.

### `terminal.Terminal`

* A new function `semanticPrompt` for handling all the semantic prompt operations.

**AI Disclosure:** AI was used in various ways throughout this. It is by no means 100% AI-written, it's mostly human written. I reviewed everything. AI was used mostly for some assistance on rote tasks.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -4216,6 +4216,151 @@ pub fn diagram(
     }
 }
 
+/// Returns the boundaries of the given semantic content type for
+/// the prompt at the given pin. The pin row MUST be the first row
+/// of a prompt, otherwise the results may be nonsense.
+///
+/// To get prompt pins, use promptIterator. Warning that if there are
+/// no semantic prompts ever present, promptIterator will iterate the
+/// entire PageList. Downstream callers should keep track of a flag if
+/// they've ever seen semantic prompt operations to prevent this performance
+/// case.
+///
+/// Note that some semantic content type such as "input" is usually
+/// nested within prompt boundaries, so the returned boundaries may include
+/// prompt text.
+pub fn highlightSemanticContent(
+    self: *const PageList,
+    at: Pin,
+    content: pagepkg.Cell.SemanticContent,
+) ?highlight.Untracked {
+    // Performance note: we can do this more efficiently in a single
+    // forward-pass. Semantic content operations aren't usually fast path
+    // but if someone wants to optimize them someday that's great.
+
+    const end: Pin = end: {
+        // Safety assertion, our starting point should be a prompt row.
+        // so the first returned prompt should be ourselves.
+        var it = at.promptIterator(.right_down, null);
+        assert(it.next().?.y == at.y);
+
+        // Our end is the end of the line just before the next prompt
+        // line, which should exist since we verified we have at least
+        // two prompts here.
+        if (it.next()) |next| next: {
+            var prev = next.up(1) orelse break :next;
+            prev.x = prev.node.data.size.cols - 1;
+            break :end prev;
+        }
+
+        // Didn't find any further prompt so the end of our zone is
+        // the end of the screen.
+        break :end self.getBottomRight(.screen).?;
+    };
+
+    switch (content) {
+        // For the prompt, we select all the way up to command output.
+        // We include all the input lines, too.
+        .prompt => {
+            var result: highlight.Untracked = .{
+                .start = at.left(at.x),
+                .end = at,
+            };
+
+            var it = at.cellIterator(.right_down, end);
+            while (it.next()) |p| {
+                switch (p.rowAndCell().cell.semantic_content) {
+                    .prompt, .input => result.end = p,
+                    .output => break,
+                }
+            }
+
+            return result;
+        },
+
+        // For input, we include the start of the input to the end of
+        // the input, which may include all the prompts in the middle, too.
+        .input => {
+            var result: highlight.Untracked = .{
+                .start = undefined,
+                .end = undefined,
+            };
+
+            // Find the start
+            var it = at.cellIterator(.right_down, end);
+            while (it.next()) |p| {
+                switch (p.rowAndCell().cell.semantic_content) {
+                    .prompt => {},
+                    .input => {
+                        result.start = p;
+                        result.end = p;
+                        break;
+                    },
+                    .output => return null,
+                }
+            } else {
+                // No input found
+                return null;
+            }
+
+            // Find the end
+            while (it.next()) |p| {
+                switch (p.rowAndCell().cell.semantic_content) {
+                    // Prompts can be nested in our input for continuation
+                    .prompt => {},
+
+                    // Output means we're done
+                    .output => break,
+
+                    .input => result.end = p,
+                }
+            }
+
+            return result;
+        },
+
+        .output => {
+            var result: highlight.Untracked = .{
+                .start = undefined,
+                .end = undefined,
+            };
+
+            // Find the start
+            var it = at.cellIterator(.right_down, end);
+            while (it.next()) |p| {
+                const cell = p.rowAndCell().cell;
+                switch (cell.semantic_content) {
+                    .prompt, .input => {},
+                    .output => {
+                        // Skip empty cells - they default to .output but aren't real output
+                        if (!cell.hasText()) continue;
+                        result.start = p;
+                        result.end = p;
+                        break;
+                    },
+                }
+            } else {
+                // No output found
+                return null;
+            }
+
+            // Find the end
+            while (it.next()) |p| {
+                const cell = p.rowAndCell().cell;
+                switch (cell.semantic_content) {
+                    .prompt, .input => break,
+                    .output => {
+                        // Only extend to cells with actual text
+                        if (cell.hasText()) result.end = p;
+                    },
+                }
+            }
+
+            return result;
+        },
+    }
+}
+
 /// Direction that iterators can move.
 pub const Direction = enum { left_up, right_down };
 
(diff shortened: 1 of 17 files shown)


mitchellh:
I'm opening this up for review now. It's such a big amount of work it's hard to tell if I forgot anything really. All existing tests were updated and pass, some were removed for removed functionality (`clearPrompt`), and in general its a lot more testable. But, it's unknown unknowns right now for me on what I might be missing. 😄

jcollie (approved):
LGTM - compiles and nothing seems to be broken in the first couple of minutes of using it.

src/terminal/page.zig
```diff
@@ -1922,29 +1929,17 @@ pub const Row = packed struct(u64) {
...
-        /// This is a prompt line, meaning it only contains the shell prompt.
-        /// For poorly behaving shells, this may also be the input.
+    /// The semantic prompt state of the row. See `semantic_prompt`.
+    pub const SemanticPrompt = enum(u2) {
+        /// No prompt cells in this row.
+        no_prompt = 0,
+        /// Prompt cells exist in this row.
         prompt = 1,
+        /// Prompt cells exist in this row that had k=c set (continuation)
+        /// line. This is used as a way to
```
jparise: Leaving the reader in suspense here. 🙂

src/terminal/page.zig
```diff
@@ -2033,6 +2033,19 @@ pub const Cell = packed struct(u64) {
         spacer_head = 3,
     };
 
+    pub const SemanticContent = enum(u2) {
+        /// Regular output content, such as command output.
+        output = 0,
```
jparise, line 2043: I went in this direction, too, but then I questioned whether or not we should differentiate between `unknown` and `output`. This distinction could be relevant if we think about a "select command output" operation (e.g. `highlightSemanticContent`) needing to find its bounds, and avoiding including screen content that wasn't marked up at all.

In practice, we'll probably run into the preceding prompt markers, but it still might be good to be rigorously correct there, and it wouldn't cost us any more bits (this `u2` has room for another field).

src/terminal/page.zig
```diff
@@ -1922,29 +1929,17 @@ pub const Row = packed struct(u64) {
...
-    pub const SemanticPrompt = enum(u3) {
-        /// Unknown, the running application didn't tell us for this line.
-        unknown = 0,
-
-        /// This is a prompt line, meaning it only contains the shell prompt.
-        /// For poorly behaving shells, this may also be the input.
+    /// The semantic prompt state of the row. See `semantic_prompt`.
+    pub const SemanticPrompt = enum(u2) {
+        /// No prompt cells in this row.
+        no_prompt = 0,
```
jparise: `none` reads a little better to my eyes in context, but no strong feelings.

src/terminal/PageList.zig
```diff
@@ -4204,9 +4210,321 @@ pub fn diagram(
...
+
+            // Find the start
+            var it = at.cellIterator(.right_down, end);
+            while (it.next()) |p| {
+                const cell = p.rowAndCell().cell;
+                switch (cell.semantic_content) {
+                    .prompt, .input => {},
+                    .output => {
+                        // Skip empty cells - they default to .output but aren't real output
+                        if (!cell.hasText()) continue;
```
jparise, line 4336: Is this only relevant in the `.output` case, or should this condition be checked right before we `switch`?

jparise (commented):
The other thing I found useful in my old branch was an integrity check that verified the row-level state matched the row's cells.

```zig
// Page.verifyIntegrity

for (rows, 0..) |*row, y| {
    var has_semantic_prompt_cells = false;

    for (cells, 0..) |*cell, x| {
        if (cell.semantic_type == .input or cell.semantic_type == .prompt) {
            has_semantic_prompt_cells = true;
        }
    }

    // Verify that the row-level semantic prompt flag accurately reflects
    // whether or not the row contains any semantic prompt cells.
    if (row.semantic_prompt != has_semantic_prompt_cells) {
        log.warn(
            "page integrity violation y={} semantic prompt mismatch row={} cells={}",
            .{ y, row.semantic_prompt, has_semantic_prompt_cells },
        );
        return IntegrityError.MismatchedSemanticPrompt;
     }
}
```

src/terminal/page.zig
mitchellh: I thought about that, but it complicates almost every semantic prompt point, and I think "assume output" is the default behavior in almost all of those cases. If a real world scenario pops up though we can consider this. I just couldn't really think of one.

src/terminal/PageList.zig
mitchellh: It only affects `output` because empty cells default to `output` (zero value). 

It is possible that this could affect prompt or input though if the shell jumps the cursor around... that may be a real world scenario for the prior comment about the zero value. I think realistically I'd have to see what happens in real shells. I feel like most will probably draw spaces.

src/terminal/page.zig
jparise: Cool, that's where I ended up, too. At long as we bound any "output" region expansion on a prompt/input marker, I think we'll have behavior.

src/terminal/PageList.zig
jparise: This feels like good fuzz testing territory, assuming we have a good sense of the boundary conditions.

However, in practical usage, I think we're in a safe space with the logic as you have it.

jparise (changes requested):
I found a regression (in terms of both `main`'s and expected behavior). You can reproduce it in bash with a multiline prompt like this:

```bash
export PS1="(\u@\h) [\[\w\]]\n\$ "
```

`main` works as you'd probably expect: the cursor is on the last row after the `$`, and resizing reflows the screen without disrupting the existing prompt.

With this branch, the cursor is placed on the line _after_ the `$`, and resizing the screen moves/paints the last `$` toward the bottom of the window's bounds.

https://github.com/user-attachments/assets/c77fae88-4fdf-436c-9502-7710598694a4

mitchellh:
CleanShot 2026-01-31 at 13 32 11@2x

cc @jparise

mitchellh:
Just leaving some notes here @jparise I'm noticing with `bash --norc --noprofile` and just our shell integration, the first line is marked as input:

CleanShot 2026-01-31 at 14 07 40@2x

EDIT: And when I resize, it turns into input.

jparise:
@mitchellh try removing these lines, which were a workaround for both bash's prompt redrawing behavior and our previous line-oriented semantic prompt tracking:

https://github.com/ghostty-org/ghostty/blob/8d1d17e00f289b7ca9291ca41eed0972d01e8583/src/shell-integration/bash/ghostty.bash#L199-L204

**(actually, that's probably not it, because this is specific to multiline prompts)**

If that's not it, I can dig in and help debug this tomorrow. It'll just take a little bit for my brain to page in more of these details.

*EDIT:* I can reproduce what you're seeing, so it shouldn't be hard to get this sorted out.

mitchellh:
CleanShot 2026-01-31 at 14 45 03@2x

Got it @jparise: https://github.com/ghostty-org/ghostty/pull/10455/changes/4bee8202a8e3385bd50144252b8c86b1683b2a3a

Primary change:

```bash
    # Marks. We need to do fresh line (A) at the beginning of the prompt
    # since if the cursor is not at the beginning of a line, the terminal
    # will emit a newline.
    PS1='\[\e]133;A\a\]'$PS1'\[\e]133;B\a\]'
    PS2='\[\e]133;A;k=s\a\]'$PS2'\[\e]133;B\a\]'

    # Bash doesn't redraw the leading lines in a multiline prompt so
    # we mark the start of each line (after each newline) as a secondary
    # prompt. This correctly handles multiline prompts by setting the first
    # to primary and the subsequent lines to secondary.
    if [[ "${PS1}" == *"\n"* || "${PS1}" == *$'\n'* ]]; then
      builtin local __ghostty_mark=$'\\[\\e]133;A;k=s\\a\\]'
      PS1="${PS1//$'\n'/$'\n'$__ghostty_mark}"
      PS1="${PS1//\\n/\\n$__ghostty_mark}"
    fi
```

I found another bug with how we delete prompts but I think that's on the Ghostty side. The metadata for this is now correct AND the bash shell integration is properly marking input lines.

mitchellh:
Okay the issue is what the original comment says: bash doesn't redraw leading lines in a multiline prompt lol. This seems to be a bash-specific problem and a very annoying one at that. Let me think.

mitchellh:
Okay, I fixed this by adding a non-standard extension `redraw=last`. When this is set, we only clear the _last_ line on resize. Only bash does this.

mitchellh:
I pushed a fix for zsh as well. @jparise I think `fish` is pretty cooked. We need to do something similar to the spec here and wrap fish_prompt: https://gitlab.freedesktop.org/Per_Bothner/specifications/-/blob/master/proposals/prompts-data/shell-integration.fish The builtin prompt marking with fish v4 and such does't properly do input OR continuation lines.

mitchellh:
Pushed another edge case fix, this time for Nu. The commit messages have details.

mitchellh:
Okay @jparise I think this is good to go. I just didn't do the Fish fixups because they seem to require total revamps compared to the others (which I included). Bash, Zsh, and Nu all seem to be doing great. I verified and fixed a bunch of edge cases with the overlay.

jparise (approved):
This is a solid improvement in every way, so let's roll with it. A few outstanding issues to track as follow-ups:

- Sometimes the overlay gets lost and stops rendering prompts mid-screen
- Elvish doesn't appear to mark `input` regions (probably broken in the integration script)
- And the Nushell and fish issues you already mentioned

mitchellh:
Pushed one more slight change. Going to merge that one and will create follow up issues.

mitchellh:
Follow-ups:

#10522 
#10523

danielwe:
Hi folks, just wanted to make sure you're aware of #10572. In short, this PR broke multiline zsh prompts generated by https://github.com/sindresorhus/pure due to adding a mark after each newline in the prompt. There's now a downstream fix/workaround, sindresorhus/pure#706, but I thought I'd flag this here in case you think it's a cause for reconsidering the approach in this PR. Comment from the author of the downstream fix here: https://github.com/ghostty-org/ghostty/discussions/10572#discussioncomment-15769861.

PR #10459, opened 2026-01-27
fix: calculate cell size before presenting gtk window

ajbucci:
Fixes #7937

Added `computeInitialSize` to GTK `Surface` and call it in GTK `Application` before the first `present()`, so the window manager centers the correct size on initial show.

The issue occurs because the core `Surface.recomputeInitialSize()` runs only after the renderer is initialized. In GTK, the `GLArea` isn’t realized until after `present()`, so the initial size arrives too late for WM centering.

**Limitations**:  when we precompute size before `present()` we do not have access to padding, so the sizing will be very slightly off... but since it is only off a few pixels I was unable to tell visually that it wasn't perfectly centered.

**Other thoughts**: I was hesitant to make changes to core `Surface` because the issue is Linux-specific, but it may make sense to extract a helper from `recomputeInitialSize` to avoid duplicating the sizing math.

**AI Disclosure:** I used AI to explore the project, help with any language / API questions (I've never used zig before and rarely use gtk), and make implementation suggestions.

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -2005,11 +2005,12 @@ pub const Surface = extern struct {
         self.as(gobject.Object).notifyByPspec(properties.@"default-size".impl.param_spec);
     }
 
-    /// Compute and set the initial window size from config and font metrics.
+    /// Estimate and set the initial window size from config and font metrics.
     /// This can be called before the core surface exists to set up the window
-    /// size before presenting.
-    pub fn computeInitialSize(self: *Self) void {
-        const priv = self.private();
+    /// size before presenting. This is an estimate because it does not take
+    /// into account any padding that may need to be added to the window.
+    pub fn estimateInitialSize(self: *Self) void {
+        const priv: *Private = self.private();
         const config_obj = priv.config orelse return;
         const config = config_obj.get();
 
@@ -2042,11 +2043,8 @@ pub const Surface = extern struct {
 
         const cell = font_grid.cellSize();
 
-        // Calculate size: "best guess"; Padding unavailable pre-init.
-        // We do not need to @max here because the surface init will set
-        // size_limit which enforces minimums defined in src/Surface.zig
-        const width = config.@"window-width" * cell.width;
-        const height = config.@"window-height" * cell.height;
+        const width = @max(CoreSurface.min_window_width_cells, config.@"window-width") * cell.width;
+        const height = @max(CoreSurface.min_window_height_cells, config.@"window-height") * cell.height;
         const width_f32: f32 = @floatFromInt(width);
         const height_f32: f32 = @floatFromInt(height);
 
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -2042,9 +2042,11 @@ pub const Surface = extern struct {
 
         const cell = font_grid.cellSize();
 
-        // Calculate size (matching recomputeInitialSize logic)
-        const width = @max(config.@"window-width", 10) * cell.width;
-        const height = @max(config.@"window-height", 4) * cell.height;
+        // Calculate size: "best guess"; Padding unavailable pre-init.
+        // We do not need to @max here because the surface init will set
+        // size_limit which enforces minimums defined in src/Surface.zig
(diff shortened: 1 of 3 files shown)


jcollie:
I think that we should at least look at refactoring `recomputeInitialSize` in the core `Surface` to see how much we can deduplicate. If we can get some unit tests from the refactor that would be interesting as well.

src/apprt/gtk/class/surface.zig
```diff
@@ -2038,6 +2038,55 @@ pub const Surface = extern struct {
...
+        const y_dpi = content_scale.y * font.face.default_dpi;
+
+        const font_size: font.face.DesiredSize = .{
+            .points = config.@"font-size",
+            .xdpi = @intFromFloat(x_dpi),
+            .ydpi = @intFromFloat(y_dpi),
+        };
+
+        // Get font grid for cell metrics
+        var derived_config = font.SharedGridSet.DerivedConfig.init(alloc, config) catch return;
```
jcollie: So I'm not the most familiar with the font system in Ghostty, but this looks like a giant red flag to me.

src/apprt/gtk/class/surface.zig
```diff
@@ -2038,6 +2038,55 @@ pub const Surface = extern struct {
...
+        const font_grid_key, const font_grid = app.core().font_grid_set.ref(
+            &derived_config,
+            font_size,
+        ) catch return;
+        defer app.core().font_grid_set.deref(font_grid_key);
+
+        const cell = font_grid.cellSize();
+
+        // Calculate size (matching recomputeInitialSize logic)
+        const width = @max(config.@"window-width", 10) * cell.width;
```
jcollie: Definitely do not want a magic number `10` here.

src/apprt/gtk/class/surface.zig
```diff
@@ -2038,6 +2038,55 @@ pub const Surface = extern struct {
...
+        const height = @max(config.@"window-height", 4) * cell.height;
```
jcollie: Nor the magic number `4`. These need to reference the constants in `src/Surface.zig` somehow.

ajbucci: Thanks for the review.  I'll take a look at refactoring recomputeInitialSize and adding some unit tests; then I won't have to worry about the private constants.

ajbucci: On second thought, I think we can safely leave out the max() altogether because this is just our 'best guess' initial size.  When the window eventually presents it will resize to the 'real' size which already has this logic baked in.

ajbucci:
I took a closer look at this and I think the duplication is minimal enough that we can probably leave it as proposed. Here are the other options I explored:

#### 1) Extract just the final window-size math (cells × cell size ÷ content scale + padding) into a shared helper.

- Removes a very small amount of duplicated logic, which would only exist in one other place and is fairly straightforward
- Would require introducing a new static helper on core Surface (Surface currently doesn't have any static helpers), or introduce a separate util
- Would include padding / min-size logic that isn't relevant pre-init
- Does allow for ergonomic unit testing of the sizing math

#### 2) Extract the entire sizing pipeline (content scale -> DPI -> DesiredSize -> DerivedConfig -> font grid lookup -> cell size -> window size)

- Fully deduplicates the logic
- Callers would have different ownership semantics for font_grid_key: Surface.init needs to retain it, while the GTK pre-init path wants to deref immediately
- Ends up being fairly invasive to Surface.init for what is effectively a GTK-only quirk

I’m hesitant to introduce new public/static APIs solely to support an apprt quirk, especially if it increases core complexity.  Let me know if you'd like me to explore further or have any suggestions!

jcollie (approved):
LGTM

jcollie:
I added back in the check for minimum window sizes, it's something small but I'd rather have it than not. I also updated comments and function names to make clearer that this is just an estimate since we don't know the padding size.

This is ready to go, we just need an ack from someone on the @ghostty-org/terminal team.

ajbucci:
Sounds good!  Thanks for taking the time to review.

pluiedev (approved):

mitchellh:
I'm fine with this. Our ability to control sizing and so on on GTK4 seems extremely limited so I think even if we're off a bit this is the best we can do.

Issue #5634, opened 2025-02-08
OSC 99 style desktop notifications


pluiedev:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/4405

<div type='discussions-op-text'>

<sup>Originally posted by **ehamiter** January  2, 2025</sup>
Coming from kitty, there's a nice way to implement desktop notifications, e.g.

```
printf '\x1b]99;;foo has been updated\x1b\\'
```

This makes it easy to add notifications to scripts, processes, etc. More info on how kitty implements it here:

https://sw.kovidgoyal.net/kitty/desktop-notifications/

Just curious if this would be something easy to implement or not.

Thank you!
</div>

jcollie:
Note PR #10467 for _parsing_ OSC 99 sequences. Actually sending notifications is not implemented yet.

PR #10467, opened 2026-01-27
osc: parse the Kitty desktop notification protocol (OSC 99)

jcollie:
This includes only parsing of the OSC. You cannot (yet) use OSC 99 to send notifications. Uses lazy parsing of the metadata modelled on the new OSC 133 behavior.

Related to #5634

### DIFF
--- a/src/terminal/osc/parsers/kitty_desktop_notification.zig
+++ b/src/terminal/osc/parsers/kitty_desktop_notification.zig
@@ -175,8 +175,8 @@ pub const Option = enum {
         metadata: []const u8,
     ) key.Type() {
         var it: Iterator(Option, isValidMetadataValue, key) = switch (key) {
-            inline .t, .n => return .init(metadata),
-            inline else => .init(metadata),
+            .t, .n => return .init(metadata),
+            else => .init(metadata),
         };
 
         const value = it.next() orelse return key.default();
--- a/src/terminal/osc/parsers/kitty_desktop_notification.zig
+++ b/src/terminal/osc/parsers/kitty_desktop_notification.zig
@@ -284,10 +284,10 @@ pub fn parsePackedStruct(comptime T: type, str: []const u8) T {
     return result;
 }
 
-/// Characters that are valid in a metadata value. Including `=` is
-/// "technically" against the spec but is needed since Base64 encoded values
-/// (with padding) are valid for some options. Including `?` is "technically"
-/// against the spec but is needed since it is a valid value for the `p` option.
+/// Characters that are valid in a metadata value. Including `=` is technically
+/// against the spec but is needed since Base64 encoded values (with padding)
+/// are valid for some options. Including `?` is technically against the spec
+/// but is needed since it is a valid value for the `p` option.
 const valid_metadata_value_characters: []const u8 = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_/+.,(){}[]*&^%$#@!`~=?";
 
 fn isValidMetadataValue(str: []const u8) bool {
--- a/src/terminal/osc/parsers/kitty_desktop_notification.zig
+++ b/src/terminal/osc/parsers/kitty_desktop_notification.zig
@@ -94,19 +94,34 @@ pub const Urgency = enum {
 };
 
 pub const Option = enum {
+    /// What action(s) should be taken when a notification is clicked.
     a,
+    /// Should a notification be sent to the application when the notification
+    /// is closed?
     c,
+    /// Are we done with the notification, and it is ready to be sent?
     d,
+    /// Is the payload encoded with Base64?
     e,
+    /// The nname of the application that is sending the notification.
     f,
+    /// Identifier for icon data.
     g,
+    /// Identifier for the notification.
     i,
+    /// Icon name.
     n,
+    /// When to honor the notification request.
     o,
+    /// Type of the payload.
     p,
+    /// The sound name to play with the notification.
     s,
+    /// The type of the notification.
     t,
+    /// The urgency of the notification.
     u,
+    /// When to auto-close the notification.
     w,
 
     pub fn Type(comptime key: Option) type {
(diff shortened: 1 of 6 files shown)


src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1293 @@
...
+    f,
+    g,
+    i,
+    n,
+    o,
+    p,
+    s,
+    t,
+    u,
+    w,
```
pluiedev, line 129: We should make these a bit more user-readable... Have fun trying to recall why `w` is the timeout option xD

src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1293 @@
...
+    break :i &result;
+};
+
+fn isValidIdentifier(str: []const u8) bool {
+    if (std.mem.indexOfAny(u8, str, invalid_identifier_characters)) |_| {
+        return false;
+    } else {
+        return true;
+    }
+}
```
pluiedev: Can't we use `std.mem.indexOfNone` to do this exact same thing (with much less code, as in we can generate the list of characters that *are* valid instead of the ones that *aren't*)?

pluiedev (changes requested):
First round of reviews, definitely not exhaustive since there's a ton to go through

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: TIL about indexOfNone.

src/terminal/osc/parsers/kitty_desktop_notification.zig
qwerasd205: `w`ait

mitchellh: `w`at

jcollie: Hmm, true, but if you're trying to read/follow the spec cross-referencing Ghostty specific names with the "official" names could be annoying.

mitchellh: For the OSC133 protocol I put comments next to the more human friendly names. I think its more likely someone coming in will not know the spec at first.

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: Fixed.

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: I added comments to all of the keys explaining their purpose.

src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1227 @@
...
+    /// What action(s) should be taken when a notification is clicked.
+    a,
+    /// Should a notification be sent to the application when the notification
+    /// is closed?
+    c,
+    /// Are we done with the notification, and it is ready to be sent?
+    d,
+    /// Is the payload encoded with Base64?
+    e,
+    /// The nname of the application that is sending the notification.
```
pluiedev: Typo
```suggestion
    /// The name of the application that is sending the notification.
```

src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1227 @@
...
+    f,
+    /// Identifier for icon data.
```
pluiedev: This actually confused me a ton so maybe it's worth mentioning that this is only ever used for `icon` type payloads with custom image data

src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1227 @@
...
+            const trimmed = std.mem.trim(u8, raw, &std.ascii.whitespace);
+            if (std.mem.startsWith(u8, trimmed, negation_prefix)) {
+                break :part .{ trimmed[negation_prefix.len..], false };
+            } else {
+                break :part .{ trimmed, true };
+            }
+        };
+
+        inline for (info.fields) |field| {
+            assert(field.type == bool);
```
pluiedev: Is there a way to guarantee that this check only happens in compile time?

src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1227 @@
...
+    }
+
+    return result;
+}
+
+/// Characters that are valid in a metadata value. Including `=` is technically
+/// against the spec but is needed since Base64 encoded values (with padding)
+/// are valid for some options. Including `?` is technically against the spec
+/// but is needed since it is a valid value for the `p` option.
+const valid_metadata_value_characters: []const u8 = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_/+.,(){}[]*&^%$#@!`~=?";
```
pluiedev: Maybe derive this from `valid_identifier_characters` + the extra characters?

src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1227 @@
...
+                self.pos = end + 1;
+                // strip any leading or trailing whitespace
+                const value = std.mem.trim(u8, self.metadata[start..end], &std.ascii.whitespace);
+                // if this is not a valid value, skip it
+                if (!isValidMetadataValue(value)) continue;
+                // return the value
+                return value;
+            }
+            // the key was not found
+            return null;
```
pluiedev: I don't exactly like the fact we have to duplicate this much logic. Maybe a better way would be to parse *all* options via iterators and only keep the last one if only one value is expected?

I'm also thinking a bit more deeply about this whole design. Since there's so many fields involved maybe a better design would be to allow retrieving a *subset* of keys at a time, and use a single unified iterator design. But I'm just putting my mind out there

pluiedev (changes requested):

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: Fixed the code duplication. I don't think that having _every_ field return an iterator would make for very ergonomic code when implementing the desktop notifications (I've got an implementation in progress).

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: Done.

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: Added `comptime` keyword to guarantee evaluation at compile time.

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: Added some comments.

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: Fixed.

src/terminal/osc/parsers/kitty_desktop_notification.zig
```diff
@@ -0,0 +1,1200 @@
...
+    ///
+    /// Any errors in the raw string will return null since the OSC 99
+    /// specification says to ignore unknown or malformed options.
+    pub fn read(
+        comptime key: Option,
+        metadata: []const u8,
+    ) key.Type() {
+        var it: Iterator(key) = switch (key) {
+            inline .t, .n => return .init(metadata),
+            inline else => .init(metadata),
```
pluiedev: Is an inline else truly necessary here?

pluiedev (approved):
LGTM

src/terminal/osc/parsers/kitty_desktop_notification.zig
jcollie: Apparently not.

rockorager:
Would like to use this! Is it ready to go still?

jcollie:
> Would like to use this! Is it ready to go still?

Other than a rebase I think that this is still ready to go.

jcollie:
Pushed a rebase.

mitchellh:
Merged! Sorry did local review so couldn't go through PR, but authorship retained.

PR #10490, opened 2026-01-29
nushell: refactor ssh wrapper for clarity

jparise:
The ssh wrapper previously used a separate set_ssh_terminfo function that returned a record to be merged, which result in some redundant control flow and TERM assignments.

This inlines the terminfo logic and builds env/opts incrementally based on feature flags. TERM is set to a fallback early and only overridden on success, which simplifies our error handling and avoids mutable variable capture issues in closures.

Lastly, warnings are now consistently written to stderr, and I made various other control flow and syntax improvements.

### DIFF
--- a/src/shell-integration/nushell/vendor/autoload/ghostty.nu
+++ b/src/shell-integration/nushell/vendor/autoload/ghostty.nu
@@ -4,106 +4,83 @@ export module ghostty {
     $feature in ($env.GHOSTTY_SHELL_FEATURES | default "" | split row ',')
   }
 
-  # Enables automatic terminfo installation on remote hosts.
-  # Attempts to install Ghostty's terminfo entry using infocmp and tic when
-  # connecting to hosts that lack it. 
-  # Requires infocmp to be available locally and tic to be available on remote hosts.
-  # Caches installations to avoid repeat installations.
-  def set_ssh_terminfo [
-    ssh_opts: list<string>
-    ssh_args: list<string>
-  ]: [nothing -> record<ssh_term: string, ssh_opts: list<string>>] {
-    let ssh_cfg = ^ssh -G ...($ssh_args)
-      | lines
-      | parse "{key} {value}"
-      | where key in ["user" "hostname"]
-      | select key value
-      | transpose -rd
-      | default {user: $env.USER hostname: "localhost"}
-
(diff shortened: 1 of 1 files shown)


jparise:
@dmatos2012 I made a few local passes over the new code. I think it looks pretty good, but let me know if you have any feedback or see further opportunities for improvement.

src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -4,106 +4,83 @@ export module ghostty {
...
-    }
-    let base_ssh_term = if (has_feature "ssh-env") {
-      "xterm-256color"
-    } else {
-      ($env.TERM? | default "")
+  export def --wrapped ssh [...args] {
+    mut ssh_env = {}
+    mut ssh_opts = []
+
+    # `ssh-env`: use xterm-256color and propagate COLORTERM/TERM_PROGRAM vars
```
dmatos2012, line 12: do we not want to maybe note that the `COLORTERM` and `TERM_PROGRAM` may not be propagated depending on the server config? 

Maybe its obvious for some, but can lead to confusions for others

jparise: That detail is common to all shell integrations, so it's probably better to mention it in our user-facing documentation.

https://ghostty.org/docs/features/shell-integration

dmatos2012 (commented):
@jparise this is really nice! I think now inlined is much better. I initially preferred `set_ssh_terminfo` it on its own `fn` for clarity because it was a bit too long, but given that you have cleaned up I like it like this way more. 

I just had a question for clarity on the comment, but nothing else. LGTM

src/shell-integration/nushell/vendor/autoload/ghostty.nu
jparise: https://github.com/ghostty-org/website/issues/421

jparise: And this is already covered by the configuration documentation:

https://github.com/ghostty-org/ghostty/blob/7c7d5421f9380f7ca65a0b413d3117d5a88e2dd1/src/config/Config.zig#L2719-L2724

mitchellh (approved):

PR #10491, opened 2026-01-29
core: don't log redraw_surface or redraw_inspector app messages

jcollie:
They are _very_ verbose and make other debug logs difficult to read.

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -237,7 +237,7 @@ pub fn needsConfirmQuit(self: *const App) bool {
 /// Drain the mailbox.
 fn drainMailbox(self: *App, rt_app: *apprt.App) !void {
     while (self.mailbox.pop()) |message| {
-        if (std.log.logEnabled(.debug, .app)) {
+        if (comptime std.log.logEnabled(.debug, .app)) {
             switch (message) {
                 // these tend to be way too verbose for normal debugging
                 .redraw_surface, .redraw_inspector => {},
--- a/src/App.zig
+++ b/src/App.zig
@@ -240,7 +240,7 @@ fn drainMailbox(self: *App, rt_app: *apprt.App) !void {
         switch (message) {
             // these tend to be way too verbose for normal debugging
             .redraw_surface, .redraw_inspector => {},
-            else => log.debug("mailbox message={s}", .{@tagName(message)}),
+            else => log.debug("mailbox message={t}", .{message}),
         }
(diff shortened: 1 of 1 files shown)


src/App.zig
```diff
@@ -237,7 +237,11 @@ pub fn needsConfirmQuit(self: *const App) bool {
 /// Drain the mailbox.
 fn drainMailbox(self: *App, rt_app: *apprt.App) !void {
     while (self.mailbox.pop()) |message| {
-        log.debug("mailbox message={s}", .{@tagName(message)});
+        switch (message) {
+            // these tend to be way too verbose for normal debugging
+            .redraw_surface, .redraw_inspector => {},
+            else => log.debug("mailbox message={s}", .{@tagName(message)}),
```
mitchellh: I know you copy and pasted this but we might as well updated to Zig 0.15 here and use `{t}` without the tagname call.

src/App.zig
```diff
@@ -237,7 +237,11 @@ pub fn needsConfirmQuit(self: *const App) bool {
 /// Drain the mailbox.
 fn drainMailbox(self: *App, rt_app: *apprt.App) !void {
     while (self.mailbox.pop()) |message| {
-        log.debug("mailbox message={s}", .{@tagName(message)});
+        switch (message) {
```
mitchellh: We should guard this with a `comptime std.log.logEnabled(.debug, scope)` so that this gets completely optimized away in release.

mitchellh (changes requested):

src/App.zig
jcollie: Done.

mitchellh (approved):

src/App.zig
```diff
@@ -237,7 +237,13 @@ pub fn needsConfirmQuit(self: *const App) bool {
 /// Drain the mailbox.
 fn drainMailbox(self: *App, rt_app: *apprt.App) !void {
     while (self.mailbox.pop()) |message| {
-        log.debug("mailbox message={s}", .{@tagName(message)});
+        if (std.log.logEnabled(.debug, .app)) {
```
mitchellh: As a note of style, I do like to put `comptime ` prefixes in front of calls like this to _ensure_ they get comptime processed. Otherwise, the compiler may run at that runtime and I don't like that uncertainty.

jcollie: Done.

PR #10498, opened 2026-01-30
flake.nix: drop desc argument for runVM function

faukah:
The current runVM function gets called only with a module and without a desc, resulting in an error when running `nix flake show` or similar commands.

This gets quite annoying when e.g. my language server starts evaluating all my flake inputs:
image

This pr drops the `desc` argument to `runVM` and gets rid of that problem.


pluiedev (approved):
Ah, I think this is my fault when I tried to rebase my refactor PR. My bad. Thanks!

jcollie (approved):

PR #10501, opened 2026-01-30
renderer: generalize and extract image renderer state

mitchellh:
This extracts all our image renderer state into a separate struct, blandly named `renderer.image.State`. This structure owns all the  storage of images and placements and exposes a limited public API to manage them.

One motivation was to limit state access by our Kitty graphics functions within the generic renderer. Another was to limit our own generic renderer from getting our image system into an incoherent state. This is prevented now on both sides due to some encapsulation.

This currently only supports Kitty images, since that's the only image protocol we support. But I intend to add additional image types to this, namely the ability to add overlay images for debug information.  **There are no plans to add new image protocols to the terminal,** the extraction is purely to support some internal features. But, it could be used for other protocols one day.

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -18,9 +18,7 @@ const constraintWidth = cellpkg.constraintWidth;
 const isCovering = cellpkg.isCovering;
 const rowNeverExtendBg = @import("row.zig").neverExtendBg;
 const imagepkg = @import("image.zig");
-const Image = imagepkg.Image;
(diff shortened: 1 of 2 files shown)


mitchellh:
No AI, just organic vibes.

mitchellh:
Took the time to clean up all our error sets and make a bunch of functions no longer fail. If image upload fails now we ignore it (and log). And drawing ignores non-ready images and draw failures (and logs). So, its all best effort.

PR #10503, opened 2026-01-30
Renderer: Debug Overlay Framework

mitchellh:
This adds a cross-platform ability to draw an overlay of debug information directly on top of our terminal. 

The debug overlay is drawn on the CPU using z2d and then rendered as an image on the GPU (using the same subsystem and shaders as things like the Kitty Graphics Protocol). This lets us iterate more quickly on overlays since it's all simple imperative 2D drawing operations, and also lets it be cross-platform immediately. 

The goal of this PR is to **introduce the framework.** To enable the overlays, you have to modify code and recompile it, but of course in the future this will be runtime toggle-able via the inspector. Additionally, the dummy overlay in this PR (OSC8 highlighting) isn't particularly useful and has limitations; it's only meant to be a demo. More realistic overlays will come later.

## Demo

Here is a demo of an overlay highlighting OSC8 hyperlinks:

CleanShot_2026-01-30_at_15 04 582x

## Performance

```
Empty but overlay active:
generic_renderer: [rebuildOverlay time] start_micro=1769835575905814 duration=12us
generic_renderer: [updateFrame time] start_micro=1769835575905766 duration=199us

Hyperlink on every line:
generic_renderer: [rebuildOverlay time] start_micro=1769835573560797 duration=452us
generic_renderer: [updateFrame time] start_micro=1769835573560682 duration=891us
```

Pretty great! 🎉

With the overlay active but nothing to draw, the overlay adds ~6% overhead. With the overlay active and hyperlinks on every line, the overlay adds ~50% overhead. These are big percentages, but the absolute numbers are miniscule. And, for a high value debug feature that is only on some of the time, it's totally acceptable.

Also, as a reminder, for 120fps we have _~8,300us_ per frame (including the time to draw, of course). If we're spending sub-1000us, it's fine. lol.

(Edit: I previously had debug timings which are significantly worse but have a lot more safety checks.)

## Other Notes

* **Damage/dirty tracking.** I don't do any for the overlay, and we just rebuild it on every frame. I'm not convinced the complexity is worth it for a debug overlay and the performance numbers at this point do not necessitate it in any way.

* **Update vs draw times.** We call `draw` _far_ more often than `update` so putting the overlay rebuild into `update` made way more sense. I did measure both.

### DIFF
--- a/src/renderer/Overlay.zig
+++ b/src/renderer/Overlay.zig
@@ -15,31 +15,23 @@ surface: z2d.Surface,
 /// Cell size information so we can map grid coordinates to pixels.
 cell_size: CellSize,
 
-/// The transformation to apply to the overlay to account for the
-/// screen padding.
-padding_transformation: z2d.Transformation,
-
 /// Initialize a new, blank overlay.
 pub fn init(alloc: Allocator, sz: Size) !Overlay {
+    // Our surface does NOT need to take into account padding because
+    // we render the overlay using the image subsystem and shaders which
+    // already take that into account.
+    const term_size = sz.terminal();
     var sfc: z2d.Surface = try .initPixel(
         .{ .rgba = .{ .r = 0, .g = 0, .b = 0, .a = 0 } },
         alloc,
-        std.math.cast(i32, sz.screen.width).?,
-        std.math.cast(i32, sz.screen.height).?,
+        std.math.cast(i32, term_size.width).?,
+        std.math.cast(i32, term_size.height).?,
     );
     errdefer sfc.deinit(alloc);
 
     return .{
         .surface = sfc,
         .cell_size = sz.cell,
-        .padding_transformation = .{
-            .ax = 1,
-            .by = 0,
-            .cx = 0,
-            .dy = 1,
-            .tx = @as(f64, @floatFromInt(sz.padding.left)),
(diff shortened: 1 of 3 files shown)


mitchellh:
One major improvement to be made in the future: when enabled, we redraw the full overlay every update. I’m not actually sure how expensive or slow this will be yet so I didnt prematurely optimize but that’s something to definitely look at later. I don’t think we should do it right away.

mitchellh:
Before:

```
generic_renderer: [rebuildOverlay time] start_micro=1769833127321631 duration=12752ns
generic_renderer: [updateFrame time] start_micro=1769833127321407 duration=13692ns
```

After:

```
generic_renderer: [rebuildOverlay time] start_micro=1769833920486521 duration=2449ns
generic_renderer: [updateFrame time] start_micro=1769833920485841 duration=3598ns
```

Okay, unfortunately the z2d draw time took up ~90% of our updateFrame times even with nothing to do. I suspect this is predominantly the surface initialization. I changed it so we cache the surface between calls and got it down significantly. 

I'm still poking at this.

mitchellh:
Turning off AA and turning on hairline mode for our border shaved off more time, down another 10% or so.

mitchellh:
I think good enough for a first pass.... let me merge and rebase my inspector work onto it so we can make it runtime toggle-able.

PR #10508, opened 2026-01-31
Fix agent wrapper PATH resolution in release builds

nicosuave:
## Summary

- macOS GUI apps launched from Finder/Dock inherit minimal PATH from launchd (`/usr/bin:/bin:/usr/sbin:/sbin`), so the claude/codex wrapper scripts couldn't find the real binaries
- Wrapper scripts now augment PATH via `/usr/libexec/path_helper` (reads `/etc/paths` + `/etc/paths.d/*`) and common binary directories (`~/.local/bin`, `~/.bun/bin`, `/opt/homebrew/bin`, `/usr/local/bin`, `~/.cargo/bin`)
- Bumped wrapper marker to v2 so existing wrappers get regenerated on next launch


nicosuave:
Apologies; agents gone wild...

PR #10536, opened 2026-02-02
Support OSC133 `click_events` Kitty extension (supported by Fish)

mitchellh:
This adds support for the `OSC 133 A click_events=1` extension introduced by Kitty and supported by Fish.[^1]

**What this means:** If the shell advertises `click_events=1` support, Ghostty will _unconditionally_ (no modifier required) send mouse events to the shell for clicks on a prompt line, delegating to the supporting shell to move the cursor as needed. For Fish 4.1+ this means that clicking on the prompt line moves the cursor (see demo video below).

This PR also contains:

* A minor fix in `cl` parsing but we don't yet implement the logic there
* Updated inspector to show the semantic prompt click mode

## Demo

https://github.com/user-attachments/assets/03ef8975-7ad9-441f-aaa2-9d0eb5c5e36d

## Implementation Details

`click_events` is wildly underspecified, so here are the details the best I understand them. This itself is not a specification (I omit details) but adds some more context to it.

The `click_events=1` option can be specified with `OSC 133 A` (Ghostty also allows it on OSC 133 N). When that is specified, it flags for all future prompts that the screen supports click events for semantic prompts. If both `click_events` and `cl` are specified, `click_events` takes priority if true. If `click_events=0` (disable), then any set `cl`  will take priority. 

When a mouse click comes in, we check for the following conditions:

1. The screen supports click events
2. The screen cursor is currently at a prompt
3. The mouse click was at or below the starting prompt line of the current prompt

If those are met, we encode an SGR mouse event with: left button, press, coordinates of click. It is up to the shell after that to handle it. Out of prompt bounds SGR events are possible (specifically below). The shell should robustly handle this.

[^1]: I don't know any other terminal or shell that supports it at the moment.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3969,6 +3969,15 @@ pub fn mouseButtonCallback(
                 log.warn("error processing links err={}", .{err});
             }
         }
+
+        // Handle prompt clicking. If we released our mouse on a prompt
+        // and we support some kind of click events, then we need to
+        // move to it.
+        if (self.maybePromptClick()) |handled| {
+            if (handled) return true;
+        } else |err| {
+            log.warn("error processing prompt click err={}", .{err});
+        }
     }
 
     // Report mouse events if enabled
@@ -4188,10 +4197,6 @@ pub fn mouseButtonCallback(
                     .y = pt_viewport.y,
                 },
             }) orelse {
-                // Weird... our viewport x/y that we just converted isn't
-                // found in our pages. This is probably a bug but we don't
-                // want to crash in releases because its harmless. So, we
-                // only assert in debug mode.
                 if (comptime std.debug.runtime_safety) unreachable;
                 break :sel;
             };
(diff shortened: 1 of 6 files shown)


mitchellh:
Sending this stuff, so I can move on to `cl`.

src/Surface.zig
```diff
@@ -4278,6 +4283,106 @@ pub fn mouseButtonCallback(
...
+    // If our cursor isn't currently at a prompt then we don't handle
+    // prompt clicks because we can't move if we're not in a prompt!
+    if (!t.cursorIsAtPrompt()) return false;
+
+    // If we have a selection currently, then releasing the mouse
+    // completes the selection and we don't do prompt moving. I don't
+    // love this logic, I think it should be generalized to "if the
+    // mouse release was on a different cell than the mouse press" but
+    // our mouse state at the time of writing this doesn't support that.
+    if (screen.selection != null) return false;
```
jparise, line 4305: This makes sense. Do you think it's important to revisit this soon-ish?

mitchellh: I don't think this would happen easily, cause there are so many other conditions this is gated on, so I don't think its that important...

PR #10539, opened 2026-02-02
feat: add scrollback-clear-allowed config option

jhammant:
## Summary
- Adds a new `scrollback-clear-allowed` configuration option (default: `true`)
- When set to `false`, CSI 3 J (ED with parameter 3) is silently ignored
- Preserves scrollback history when applications attempt to clear it

## Motivation
Some terminal applications clear the scrollback buffer on exit or during certain operations. Users who rely on scrollback history (e.g., for reviewing build logs, debugging sessions, or preserving context) may want to prevent this behavior.

This is similar to iTerm2's "Prevent CSI 3 J from clearing scrollback history" option.

## Use Cases
- Development workflows where preserving terminal history is important
- AI coding assistants that may clear scrollback during context management
- Build and deployment pipelines where logs should remain accessible

## Changes
- `src/config/Config.zig`: Added `scrollback-clear-allowed` config option with documentation
- `src/termio/Termio.zig`: Pass config to stream handler
- `src/termio/stream_handler.zig`: Conditionally handle CSI 3 J based on config

## Test plan
1. Set `scrollback-clear-allowed = false` in config
2. Run a command that generates scrollback (e.g., `for i in {1..50}; do echo "Line $i"; done`)
3. Run `printf '\033[3J'` or `clear`
4. Verify scrollback is preserved when scrolling up

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2806,6 +2806,19 @@ keybind: Keybinds = .{},
 /// need KAM, you don't need it.
 @"vt-kam-allowed": bool = false,
 
+/// If false, prevents applications from clearing the scrollback buffer
+/// using the CSI 3 J escape sequence (ED with parameter 3). This is useful
+/// if you want to preserve your scrollback history even when applications
+/// (such as shells on logout, or programs like `clear`) attempt to clear it.
+///
+/// When set to false, CSI 3 J will be silently ignored and the scrollback
+/// buffer will remain intact. The visible screen can still be cleared
+/// normally.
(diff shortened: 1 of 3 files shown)


jcollie:
Closing because you're not following our documented policies:

https://github.com/ghostty-org/ghostty/blob/main/CONTRIBUTING.md
https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md

PR #10542, opened 2026-02-02
Add support for OSC133 cl=line, bash and zsh support

mitchellh:
Related to #1966

This adds support for OSC133 `cl=line` (single line movement with left/right arrow keys) and modifies our bash and zsh shell integration to advertise support for it. With this, bash and zsh both support click to move at the prompt without any modifiers:

https://github.com/user-attachments/assets/7f6cb0b8-390c-4136-8c25-059b21b138c5

This also removes our legacy `promptPath` and related functionality (pressing alt) since this is superior and there's no reason to keep that around.


jparise (approved):

jparise:
Confirming this works great in bash and zsh. Nice!

PR #10554, opened 2026-02-03
generate 256 palette

jake-stewart:
Hi Ghostty team,

I believe that terminals should generate the 256-color palette based on the user's base16 theme.

The rationale and approach is written up [here](https://gist.github.com/jake-stewart/0a8ea46159a7da2c808e5be2177e1783).

I consider it important that terminals support this out of the box so that such behaviour can become normal and expected, because then terminal program maintainers will consider the palette a viable choice.

I have created a PR for kitty and the maintainer seems interested. I plan to offer this to more terminals soon.


src/config/Config.zig
```diff
@@ -10409,3 +10464,60 @@ test "compatibility: window new-window" {
         );
     }
 }
+
+const Lab = struct { l: f32, a: f32, b: f32 };
```
00-kat: I don't have any suggestions on where this should be put, but shoving it under seven hundred lines of tests in a 10.5k line file containing the config struct doesn't sound right to me.

jake-stewart: moved to `../lab.zig`

mitchellh (changes requested):
This is good work and an interesting idea, thank you. I'm going to read your gist in more detail to understand this better. In the mean time I do have some superficial Zig feedback:

I think we should move the contents of lab.zig to `src/terminal/color.zig` and then add a function to `terminal.color.Palette` itself that can generate itself from the 16 values.

jake-stewart:
> I think we should move the contents of lab.zig to `src/terminal/color.zig` and then add a function to `terminal.color.Palette` itself that can generate itself from the 16 values.

The `Palette` is just a `[256]RGB` so I made a function instead.

mitchellh:
!vouch

mitchellh (approved):
Thank you! I really like this and the logic is sound. I went ahead and cleaned up the Zig code quite a bit by adding tests, updating comments, etc. None of it seemed incorrect so this was just a stylistic change.

vlsi:
I suggest going with OKLCH color space rather than lab. It would do better with perceived brigtness.
See https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl
Do you think it is something that could be merged?

00-kat:
> See https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl

https://bottosson.github.io/posts/oklab by the OKLAB/OKLCH creator also contains an exemplary discussion on its benefits.

jake-stewart:
I have nothing against OKLAB but CIELAB gave flawless results for the [themes](https://github.com/jake-stewart/color256/tree/main/themes) I tested and is the battle-tested ISO 11664-4 standard.

dalbertom:
Hi there, not a big issue (but I can report one if necessary), but I noticed after upgrading that my tmux status bar is no longer the color I have configured. It used to be dark gray, now it's light gray

It used to look like this:
image

and now it looks like this:
image

Using `git bisect` I landed on fad72e0ed1895e89458e3d80d7de6bb716ff04af as the first commit that introduced the change.

The color I'm using in my tmux status bar is `#16161d` which is supposed to be a dark gray, and now it's light gray. The full configuration is:
```
set -g status-bg '#16161d'
set -g status-fg white
set -g status-left-style 'bg=#16161d,fg=white'
set -g status-right-style 'bg=#16161d,fg=white'
set -g window-status-last-style 'bg=#16161d'
set -g window-status-activity-style 'bg=#16161d,fg=white'
set -g window-status-current-style 'bg=#16161d,fg=cyan'
set -g pane-active-border-style 'bg=default,fg=cyan'
set -g pane-border-style 'bg=default,fg=white'
set -g mode-style 'bg=white,fg=black'
set -g message-style 'bg=white,fg=black'
```

My current theme on Ghostty is, but I tried several and had the same change in behavior.
```
theme = dark:Terminal Basic Dark,light:Terminal Basic
```

Thanks, let me know if filing a ticket is preferred, but I figured I'd reach out here first.

00-kat:
Sounds like #10852 to me…?

mitchellh:
@dalbertom Please do open a discussion, I'm curious to diagnose what is causing this exactly. e.g. I'm curious when when you set the status for tmux to a hex how it maps that to a palette, and what palette index its using exactly. I can help give you more steps in a discussion so we don't pollute this issue. 😄

mitchellh:
I should also note for anyone who falls into this that you can disable this behavior with `palette-generate = false`. But before we make that a default, I want to collect more information about where this isn't behaving correctly for people, since on paper it really shouldn't.

cc @dalbertom

Issue #10549, opened 2026-02-03
Implement the Kitty Clipboard Protocol (OSC 5522)


pluiedev:
See https://sw.kovidgoyal.net/kitty/clipboard - TL;DR it allows applications to copy/paste rich content from the terminal, not just plain text.

jcollie:
I've started some work on parsing, in a similar manner as OSC 133 and OSC 99. Mainly need to write more tests and then fix up any issues that arise.

MichielMAnalytics:
Would be fantastic. It will help the future of software devel a lot as with increased parallelism of agents the RAM on your machine becomes the bottleneck. if you can outsource compute on vms (but keep the native image paste functionality) that would be amazing.

lmtr0:
Can anyone enlight me on the difference between current impl, to `keybind = performable:ctrl+v=paste_from_clipboard` to Kitty Clipboard Protocol?

Setting the performable option there and trying to paste an image just works, so I am curious to understand the difference

billettc:
Here my work around to Kitty support

# Using Cmd+C to Copy Both Mouse and Keyboard Selections in Ghostty

## Problem

On macOS, `Cmd+C` only copies text selected with the mouse. Text selected using keyboard shortcuts in the shell (e.g., `Shift+Option+Left/Right` for word selection in zsh) is not copied to the clipboard when pressing `Cmd+C`.

This happens because:
1. Keyboard selection in the shell (via readline/ZLE) is managed by the shell, not the terminal
2. Ghostty cannot see the shell's internal selection state
3. In legacy terminal encoding, the `super` (Cmd) modifier is not sent to the shell

## Solution

Enable the **Kitty keyboard protocol** in zsh and use Ghostty's `performable:` keybind prefix.

### 1. Ghostty Configuration

Add to your Ghostty config (`~/Library/Application Support/com.mitchellh.ghostty/config.ghostty`):

```
# Required for Option+Arrow word movement
macos-option-as-alt = true

# Cmd+C: copy terminal selection if present, otherwise pass to shell
keybind = performable:super+c=copy_to_clipboard
```

The `performable:` prefix means:
- If there's a terminal/mouse selection → copy it to clipboard (key consumed)
- If there's no terminal selection → pass the key through to the shell

### 2. Zsh Configuration

Create or add to a zsh config file (e.g., `~/.zsh/ghostty-text-selection.zsh`):

```zsh
[...]

PR #10560, opened 2026-02-03
osc: parse OSC 5522 - Kitty clipboard protocol

jcollie:
This PR only adds support for parsing the OSCs. No support has been added to DECRQM/DECRPM/DECSET/DECRST for mode 5522.

Related to #10549

### DIFF
--- a/src/terminal/osc/parsers/kitty_clipboard_protocol.zig
+++ b/src/terminal/osc/parsers/kitty_clipboard_protocol.zig
@@ -1,0 +1,702 @@
+//! Kitty's clipboard protocol (OSC 5522)
+//! Specification: https://sw.kovidgoyal.net/kitty/clipboard/
+//! https://rockorager.dev/misc/bracketed-paste-mime/
+
+const std = @import("std");
+const build_options = @import("terminal_options");
+
+const assert = @import("../../../quirks.zig").inlineAssert;
+
+const Parser = @import("../../osc.zig").Parser;
+const Command = @import("../../osc.zig").Command;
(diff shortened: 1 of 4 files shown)


src/terminal/osc/parsers/kitty_clipboard_protocol.zig
```diff
@@ -0,0 +1,702 @@
...
+                if (metadata[pos] != '=') return null;
+                // the end of the value is bounded by a ':' or the end of the metadata
+                const end = std.mem.indexOfScalarPos(u8, metadata, pos, ':') orelse metadata.len;
+                const start = pos + 1;
+                // strip any leading or trailing whitespace
+                break :value std.mem.trim(u8, metadata[start..end], &std.ascii.whitespace);
+            }
+            // the key was not found
+            return null;
+        };
```
pluiedev, line 123: This particular k-v parser has been used in like 3 different protocols now? Makes me wonder if we should have like a `src/terminal/osc/parsers/lib.zig` that deals with it

pluiedev (approved):
LGTM for the most part

mitchellh (approved):

mitchellh:
We can clean up any edges later (if any). Looks fine from a parsing perspective and architecture.

src/terminal/osc/parsers/kitty_clipboard_protocol.zig
jcollie: FYI in #10467 I merged this implementation with OSC 5522.

PR #10567, opened 2026-02-04
gtk: auto-fallback to software GL on context failure

ishuru:
## Summary
- auto-relaunch Ghostty once with software rendering when GTK fails to create an OpenGL context
- add a manual "Restart with software rendering" action on the error page
- add `gtk-opengl-auto-fallback` config (default true) to opt out of auto fallback

## Testing
- Not run (local environment)


mitchellh:
Please read the contribution guidelines.

PR #10570, opened 2026-02-04
Fix clicking bare relative file paths

bkircher:
Related to #1972

The URL regex for file path detection requires paths to start with `../`, `./`, or `/`. For bare relative paths like `"src/config/url.zig"`, the regex could only match starting at the first `/`, producing `"/config/url.zig"` as a result — always dropping the first part of the path.

Fix: added a third top-level alternative to the regex. This matches bare relative paths where:

1. The first component is word characters (possibly with dots/dashes):
   `[\w][\w\-.]*\/`
2. The remaining path must contain a dot (via positive lookahead) — this
   requires a file extension to avoid false positives on text like
   `"input/output"`
3. Add a `(?<!\w)\/` instead of `\/` in the existing prefix group — the
   standalone `/` prefix now requires that `/` is not preceded by a word
   character. This prevents `"input/output"` from falsely matching
   `"/output"`

Test cases added:
- src/config/url.zig → matches fully
- app/folder/file.rb:1 → matches with line number
- modified:   src/config/url.zig → matches only the path part
- lib/ghostty/terminal.zig:42:10 → matches with line:col
- some-pkg/src/file.txt more text → stops before trailing text
- input/output and foo/bar → correctly do not match (no file extension)

The issue was nailed down here:
https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-3845717672

### DIFF
--- a/src/config/url.zig
+++ b/src/config/url.zig
@@ -26,7 +26,7 @@ pub const regex =
     "(?:" ++ url_schemes ++
     \\)(?:
     ++ ipv6_url_pattern ++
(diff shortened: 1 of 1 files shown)


mitchellh:
I think I want to take a more fundamental look at file paths but I think as an improvement I'm happy to take this now.

seantuttle:
Is supporting variables and tilde as the first part of a path something we'd want to include in this? Eg
```
 $HOME/foo/bar.sh
 ~/foo/bar.sh
```
Currently, the `$` and `~` are excluded from the selection, but it seems pretty intuitive to me that they be included.

bkircher:
> t seems pretty intuitive to me that they be included.

Agree 100%. Did see that yesterday after merge. I'll create an issue and try to get this ready the next days

pocketpixels:
I commented on an issue with the new implementation [here](https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-3857881196). Maybe it should have gone here or in a new bug report.

PR #10574, opened 2026-02-05
Add JPEG XL (JXL) image format support

hjanuschka:
## Summary

This adds support for JPEG XL images in both the Kitty graphics protocol and as background images.

## Changes

- Add `libjxl` package wrapper (`pkg/libjxl/`) using system libjxl library
- Add JXL format (`f=101`) to Kitty graphics protocol, matching kitty's implementation
- Add JXL file type detection via magic bytes (`0xFF 0x0A` for codestream, ISO BMFF container signature)  
- Support JXL images as background images
- Update macOS Xcode project to link libjxl

## Testing

Tested with JXL images via:
```bash
printf '\e_Gf=101,a=T;%s\e\\' "$(base64 < image.jxl)"
```

## Dependencies

Requires libjxl to be installed:
- macOS: `brew install jpeg-xl`
- Linux: `apt install libjxl-dev` or equivalent

## Related

This matches the JXL support added to kitty: https://github.com/kovidgoyal/kitty/pull/9472

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -70,6 +70,7 @@
         .harfbuzz = .{ .path = "./pkg/harfbuzz", .lazy = true },
         .highway = .{ .path = "./pkg/highway", .lazy = true },
         .libintl = .{ .path = "./pkg/libintl", .lazy = true },
+        .libjxl = .{ .path = "./pkg/libjxl", .lazy = true },
         .libpng = .{ .path = "./pkg/libpng", .lazy = true },
         .macos = .{ .path = "./pkg/macos", .lazy = true },
         .oniguruma = .{ .path = "./pkg/oniguruma", .lazy = true },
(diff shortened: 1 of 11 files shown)


jcollie:
Closing. First, and most importantly, you failed to disclose AI usage as is required in our [usage policy](https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md). Second, new features like this need to start out as feature request discussions. Once a feature request has been accepted as an issue then PRs may be submitted. Third, it's very unlikely that Kitty will accept JPEG XL as a new image format for the Kitty graphics protocol as Kitty has [refused even JPEG support in the past](https://github.com/kovidgoyal/kitty/issues/4341). We're not about to start adding extensions to Kitty protocols before Kitty does. Fourth, we don't really want to be adding more dependencies to the project just for something like this.

PR #10575, opened 2026-02-05
gtk: add two-finger left/right scroll to switch tab pages

jcollie:
This adds the ability to use two fingers on a touchpad to scroll left or right on a Ghostty window to change tab pages. Uses the same basic machinery as scrolling up and down the scrollback buffer. Scrolling pages does not wrap around at the start or end of the tabs.

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -697,6 +697,13 @@ pub const Surface = extern struct {
         /// Whether primary paste (middle-click paste) is enabled.
         gtk_enable_primary_paste: bool = true,
 
+        /// How much pending horizontal scroll do we have?
+        pending_horizontal_scroll: f64 = 0.0,
+
+        /// Timer to reset the amount of horizontal scroll if the user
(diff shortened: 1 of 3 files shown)


jcollie:
Inspired by #10564. Leaving as draft until that feature request has been accepted.

mitchellh (approved):
Code looks good. I have no way to test this, so assuming this works for you I'm happy for it.

mitchellh:
Btw, haven't converted it to an issue yet, but I approve that discussion so feel free to merge this for GTK.

folke:
fyi, this breaks horizontal scrolling in cli apps on Wayland/GTK.

For more info see https://github.com/ghostty-org/ghostty/discussions/11566

NearXdu:
crying out for this feature for macos

00-kat:
Crying out in a closed PR is metaphorically equivalent to throwing pebbles in the general direction of the Mariana Trench hoping one of them causes Pluto to exit its orbit.

Less obtusely, open a new discussion if one doesn't exist already.

PR #10583, opened 2026-02-05
macos: quote input strings used for shell commands

jparise:
When we're building an input string that's explicitly meant to be used as a shell command, quote it using the same logic as Python's `shlex.quote` function.

This specifically addresses issues we've seen when open(1)'ing Ghostty with filename arguments that contain spaces.

See #2633, #3030


mitchellh (approved):

mitchellh:
Thanks for writing Swift tests!

PR #10585, opened 2026-02-05
fix(macOS): pass through paste keybind when clipboard has no text

coryroyce:
## Summary

When pressing Cmd+V on macOS with only image data in the clipboard (no text), Ghostty intercepts the keybind, finds no text to paste, and silently swallows the keypress. Applications like Claude Code never receive the key event and can't handle image paste via native macOS APIs.

Two issues combine to cause this:

1. **`embedded.zig:clipboardRequest()` always returned `true`**, meaning the keybind was always consumed — even when the clipboard had no text. GTK already has a synchronous check that returns `false` when clipboard has no text.

2. **The default paste bindings were not `performable`**. Even if `clipboardRequest` returned `false`, without the `performable` flag the key was still consumed by the binding system.

## Changes

- **`src/apprt/embedded.zig`** — Add optional `clipboard_has_text` callback to `Options` struct. Modify `clipboardRequest()` to check it for paste requests and return `false` when clipboard has no text, giving the embedded apprt parity with GTK.
- **`include/ghostty.h`** — Add `ghostty_runtime_clipboard_has_text_cb` typedef and field to `ghostty_runtime_config_s`.
- **`macos/Sources/Ghostty/Ghostty.App.swift`** — Implement `clipboardHasText` using existing `getOpinionatedStringContents()`. Wire callback into runtime config. Add iOS stub (returns `true`).
- **`src/config/Config.zig`** — Make default paste bindings (`physical:paste`, `shift+insert`, `super+v`/`ctrl+shift+v`) `performable` so the key passes through when the action returns `false`.

## How it works

1. User presses Cmd+V with image in clipboard
2. Binding system matches `performable:super+v=paste_from_clipboard`
3. `performAction()` → `startClipboardRequest()` → `clipboardRequest()`
4. `clipboardRequest()` checks `clipboard_has_text` callback → macOS reports no text → returns `false`
5. Since binding is `performable` and action returned `false`, key passes through (`Surface.zig:2987-2993`)
6. Application receives the keypress and handles image paste natively

## Test plan

- [ ] Copy an image to clipboard, open a terminal app (e.g., Claude Code) in Ghostty, press Cmd+V — the keypress should now reach the app
- [ ] Copy text to clipboard, press Cmd+V — normal paste behavior should be unchanged
- [ ] Copy a file in Finder, press Cmd+V — file path paste should work unchanged
- [ ] Verify Shift+Insert paste works on Linux builds
- [ ] Verify physical paste key (dedicated keyboard key) works

## References

- Issue #10549 — "Implement Kitty Clipboard Protocol (OSC 5522)"
- Discussion #10099 — Root cause analysis of image paste failure
- Discussion #10117 — "Can't paste images into claude code"

## AI Disclosure

This PR was authored with AI assistance (Claude Code) per `AI_POLICY.md`.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -965,6 +965,7 @@ typedef void (*ghostty_runtime_write_clipboard_cb)(void*,
                                                    const ghostty_clipboard_content_s*,
                                                    size_t,
                                                    bool);
+typedef bool (*ghostty_runtime_clipboard_has_text_cb)(void*, ghostty_clipboard_e);
 typedef void (*ghostty_runtime_close_surface_cb)(void*, bool);
 typedef bool (*ghostty_runtime_action_cb)(ghostty_app_t,
                                           ghostty_target_s,
@@ -978,6 +979,7 @@ typedef struct {
   ghostty_runtime_read_clipboard_cb read_clipboard_cb;
   ghostty_runtime_confirm_read_clipboard_cb confirm_read_clipboard_cb;
   ghostty_runtime_write_clipboard_cb write_clipboard_cb;
+  ghostty_runtime_clipboard_has_text_cb clipboard_has_text_cb;
   ghostty_runtime_close_surface_cb close_surface_cb;
 } ghostty_runtime_config_s;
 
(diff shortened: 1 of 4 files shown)


pluiedev:
Do not vibe PRs without a relevant issue attached. This is very clearly written, and very clearly violated. Do this again and you will be banned from contributing.

PR #10609, opened 2026-02-07
bash: remove dependency on bash-preexec for bash 4.4+

jparise:
bash-preexec implements support for its "precmd" and "preexec" hooks using a combination of PROMPT_COMMAND and a DEBUG trap. The latter is unfortunately quick slow (in a relative sense), and the overall system is a bit more generalized than what we need for our shell integration (e.g. supporting multiple function hooks, subshells, etc.).

Bash 4.4 introduced the PS0 variable, which is expanded and displayed by interactive shells after reading a complete command but before executing it. This is all we need to implement our own shell integration hooks.

In Bash 5.1, PROMPT_COMMAND can be an array variable, each element of which can contain a command to be executed like a string PROMPT_COMMAND variable. When adding our hook to PROMPT_COMMAND, we preserve its type (string or array) to be minimally intrusive. This also matches direnv's approach.

Bash 5.3 introduced support for function substitution, which is an even more efficient way to run code from PS0, so we use that when available. Otherwise, we use the more traditional command substitution approach.

Earlier versions of bash (such as 3.2, which still ships with macOS) continue to use the bash-preexec path. This gives us two code paths to maintain, but I think that's preferable to fully maintaining our own DEBUG trap-based system for older bash versions given that bash-preexec has proven to work reliably in those environments. We also wouldn't unlock any other user benefits aside from removing the bash-preexec script dependency.

See: #3724, #7734

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -284,7 +284,12 @@ if (( BASH_VERSINFO[0] > 4 || (BASH_VERSINFO[0] == 4 && BASH_VERSINFO[1] >= 4) )
   # Append our hook to PROMPT_COMMAND, preserving its existing type.
   if [[ ";${PROMPT_COMMAND[*]:-};" != *";__ghostty_hook;"* ]]; then
     if [[ -z "${PROMPT_COMMAND[*]}" ]]; then
-      PROMPT_COMMAND="__ghostty_hook"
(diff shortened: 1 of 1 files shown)


tristan957:
I can confirm that this fixes #7734 with `GNU bash, version 5.3.0(1)-release (x86_64-redhat-linux-gnu)`. I'd say ship it.

PR #10611, opened 2026-02-07
gtk: revamp cgroup/transient scope handling

jcollie:
This changes the way Ghostty assigns itself and subprocesses to
cgroups and how resource controls are applied.
    
* Ghostty itself no longer modifies it's own cgroup or moves itself
to a transient scope. To modify the main Ghostty process' resource
controls ensure that you're launching Ghostty with a systemd unit and
use the standard systemd methods for overriding and applying changes
to systemd units.
    
* If configured (on by default), the process used to run your command
will be moved to a transient systemd scope after it is forked from
Ghostty but before the user's command is executed. Resource controls
will be applied to the transient scope at this time. Changes to
the `linux-cgroup*` configuration entries will not alter existing
commands. If changes are made to the `linux-cgroup*` configuration
entries commands will need to be relaunched. Resource limits can also
be modified after launch outside of Ghostty using systemd tooling. The
transient scope name can be shown by running `systemctl --user whoami`
in a shell running inside Ghostty.
    
Fixes #2084.
Related to #6669

Example of `systemctl status` showing main Ghostty process and one surface:
Screenshot From 2026-02-07 16-31-14

### DIFF
--- a/src/apprt/gtk/cgroup.zig
+++ b/src/apprt/gtk/cgroup.zig
@@ -64,9 +64,6 @@ pub fn createScope(
         // https://www.freedesktop.org/software/systemd/man/latest/systemd-oomd.service.html
         builder.add("(sv)", "ManagedOOMMemoryPressure", glib.Variant.newString("kill"));
 
-        // Delegate
-        builder.add("(sv)", "Delegate", glib.Variant.newBoolean(@intFromBool(true)));
-
         // PID to move into the unit
         const pids_value_type = glib.VariantType.new("u");
         defer glib.free(pids_value_type);
--- a/src/apprt/gtk/cgroup.zig
+++ b/src/apprt/gtk/cgroup.zig
@@ -1,7 +1,8 @@
-/// Contains all the logic for putting the Ghostty process and
-/// each individual surface into its own cgroup.
+/// Contains all the logic for putting individual surfaces into
+/// transient systemd scopes.
 const std = @import("std");
 const Allocator = std.mem.Allocator;
+const assert = @import("../../quirks.zig").inlineAssert;
 
 const gio = @import("gio");
 const glib = @import("glib");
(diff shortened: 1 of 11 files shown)


src/apprt/gtk/cgroup.zig
```diff
@@ -150,16 +53,21 @@ fn createScope(
...
+        }
 
-        builder.add("(sv)", "ManagedOOMMemoryPressure", pressure_value);
+        // https://www.freedesktop.org/software/systemd/man/latest/systemd-oomd.service.html
+        builder.add("(sv)", "ManagedOOMMemoryPressure", glib.Variant.newString("kill"));
 
         // Delegate
-        const delegate_value = glib.Variant.newBoolean(1);
-        builder.add("(sv)", "Delegate", delegate_value);
+        builder.add("(sv)", "Delegate", glib.Variant.newBoolean(@intFromBool(true)));
```
heftig: Do we really want `Delegate`? It's meant for processes that create their own sub-cgroups. It shouldn't be set for the shell.

Even if this is meant for Ghostty itself, it should be unnecessary if Ghostty only uses systemd's API to create cgroups (slices, scopes or services).

jcollie: Yeah probably not needed. Mostly it's leftover from when this function was used to move the main Ghostty process to a transient scope, which does not happen with this PR.

jcollie: Removed.

mitchellh (changes requested):
This is a pretty major change since it puts a dependency on systemd for this. I'm conceptually fine with it, but can you please explain (for both me and the history) the motivation behind this, the pros and cons compared to manual cgroup management, etc.

It may be worth keeping our `os/cgroup` functions in case we want to ever do a non-systemd mode or another apprt wants to do this. They're pretty simple but I wouldn't want to rebuild them haha.

The command stuff is a bit ugly, but I'm willing to just accept it for now and figure out a cleaner way later, I'm overall not super happy with Command.zig anyways since it started out as a Zig fork.

jcollie:
There's no _additional_ dependency on `systemd`, we were already using the DBus call to `systemd` to move the primary process to a `systemd` scope.

The motivations for this change are:

* Make every surface a unique _`systemd`_ scope (as shown by `systemctl --user whoami`). As far as I can tell the only way to do that is by calling into `systemd` as that information is kept/managed internally by PID 1.
* Apply process limits _per surface_ rather than to all surfaces combined.

I'm fine with leaving the functions in `os/cgroup`.

mitchellh (approved):
Let's just send it. Thanks!

PR #10614, opened 2026-02-08
fix(gtk): restore middle-click paste when scrollbar is enabled

### LINKED ISSUE #10613: Ubuntu 25.10: Middle-click paste does not work when the GTK scrollbar feature is enabled
Middle-click paste (primary selection paste) does not work when the GTK scrollbar feature is enabled. The `GtkScrolledWindow` appears to intercept middle-click events before they reach the terminal surface.

## Steps to Reproduce

1. Use Ghostty with scrollbars enabled (default config or `scrollbar = system`)
2. Select some text in another application
3. Middle-click in Ghostty terminal to paste
4. Nothing happens - the paste does not occur

## Expected Behavior

Text from primary selection should be pasted into the terminal, as it does when scrollbars are disabled.

## Workaround

Set `scrollbar = never` in config to disable scrollbars - this restores middle-click paste functionality.

## Environment

- Ghostty version: 1.3.0-main+6c0e02d (tip/main branch)
- OS: Ubuntu 25.10 (Wayland)
- GTK version: 4.20.1
- libadwaita: 1.8.0

hakonhagland:
## Summary

- Fixes middle-click paste (primary selection) not working when GTK scrollbars are enabled
- The `GtkScrolledWindow` was intercepting middle-click events before they reached the terminal Surface

## Solution

- Added a `GestureClick` with `propagation-phase: capture` to the ScrolledWindow that intercepts middle-clicks and forwards them to the Surface
- The gesture is claimed after handling to prevent double-paste when the scrollbar is not visible
- Also disabled `kinetic-scrolling` on the ScrolledWindow as it uses middle-click for autoscroll

## Test plan

- [x] Enable scrollbars (`scrollbar = system` in config)
- [x] Select text in another application to populate primary selection
- [x] Middle-click in Ghostty terminal - should paste once
- [x] Run `seq 1 80` to make scrollbar visible
- [x] Middle-click again - should still paste once

Fixes #10613

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -2713,6 +2713,18 @@ pub const Surface = extern struct {
         };
     }
 
+    /// Public method to forward mouse down events from parent widgets
+    /// (e.g., ScrolledWindow capturing middle-clicks).
+    pub fn forwardMouseDown(
+        self: *Self,
+        gesture: *gtk.GestureClick,
+        n_press: c_int,
+        x: f64,
+        y: f64,
+    ) void {
+        gcMouseDown(gesture, n_press, x, y, self);
+    }
+
     fn gcMouseDown(
         gesture: *gtk.GestureClick,
         _: c_int,
--- a/src/apprt/gtk/class/surface_scrolled_window.zig
+++ b/src/apprt/gtk/class/surface_scrolled_window.zig
@@ -138,6 +138,24 @@ pub const SurfaceScrolledWindow = extern struct {
         };
     }
 
+    /// Handle middle-click at the ScrolledWindow level to prevent the
+    /// ScrolledWindow from intercepting it. This forwards the event to
+    /// the Surface's mouse handler.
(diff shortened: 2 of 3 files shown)


jcollie:
Closing because procedures not followed.

hakonhagland:
For reference: we found the root cause of this issue.                                                                                                       
                                                                                                                                                              
 The bug only occurs when the GNOME setting `org.gnome.desktop.interface overlay-scrolling` is set to `false` (traditional scrollbars). With the default `true`  (overlay scrollbars), middle-click works correctly.                                                                                                         
                                                                                                                                                              
 When overlay scrolling is disabled, GTK's `ScrolledWindow` handles middle-click events differently, intercepting them before they reach child widgets.        
                                                                                                                                                              
  This fix would still be useful for users who prefer traditional scrollbars, but it's a more limited use case than originally thought.                       
                                                                                                                                                              
  See discussion: https://github.com/ghostty-org/ghostty/discussions/10616

PR #10619, opened 2026-02-08
more clickable file path fixes

bkircher:
This pull request addresses some of the remaining issues when matching `~`, `$VAR`, `.directory/`, and embedded commas. It does not address issues with embedded line breaks.

The PR is split in multiple commits carefully applying a set of changes to

1. make the big regex more composable / readable
2. update some doc strings
3. add more test cases for the issues mentioned
4. two simple commits, each fixing the issues

Changes:

- **url: refactor regex into documented branches**
  Break up the big monolithic URL and path regex into named sub-pattern
  constants and compose the final expression from three commented
  branches:
  
  - URLs with a scheme
  - absolute or dot-relative paths
  - bare relative paths
  
  This commit only breaks up the regex. It keeps the existing matching
  behavior unchanged.
  

- **url: update top-level comment**
  

- **url: carefully extend test cases**
  Extend existing test cases with `~`, `$VAR`, and bare .-prefixed paths
  and embedded `,` comma handling.
  
  See following issue comments:
  
  - https://github.com/ghostty-org/ghostty/pull/10570#issuecomment-3853842036
  - https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-3859329233
  - https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-3857881196
  

- **url: remove `,` from path_chars**
  Related to #1972
  
  Fixes an issue when paths have embedded comma, e.g.:
  
      shared/src/foo/SomeItem.m:12, shared/src/
  
  with path_chars greedily consuming the rest of the string.
  
  Now file path matching stops at comma. Scheme URLs are unchanged and
  still using the comma.
  

- **url: fix matching `~`, `$VAR`, `.directory/`**
  Related to #1972
  
  This commit adds three new alternatives for
  `rooted_or_relative_path_prefix`:
  
  - `~/`
  - `$VAR` and
  - `.local/`, `.config/` etc. for dot-prefixed directory names
  
Remaining commits fix edge cases one by one:

  - **url: fix mid-string dot partial matches**
    `"foo.local/share"` (was partial match) → now matches fully
    
  - **url: fix $-numeric character matches**
    `"$10/$20"` → no match
    
  - **url: fix partial match of mid string $-variable**
    `"foo/$BAR/baz"` (was partial match) → matches fully now
    
  - **url: fix incomplete $-numeric behavior**
    `"$10/bar.txt"` (was partial match) → but should not match at all

### DIFF
--- a/src/config/url.zig
+++ b/src/config/url.zig
@@ -75,7 +75,7 @@ const scheme_url_branch =
     no_trailing_punctuation;
 
 const rooted_or_relative_path_prefix =
-    \\(?:\.\.\/|\.\/|~\/|(?:[\w][\w\-.]*\/)*(?<!\w)\$[A-Za-z_]\w*\/|\.[\w][\w\-.]*\/|(?<!\w)\/)
+    \\(?:\.\.\/|\.\/|(?<!\w)~\/|(?:[\w][\w\-.]*\/)*(?<!\w)\$[A-Za-z_]\w*\/|\.[\w][\w\-.]*\/|(?<![\w~])\/)
 ;
 
 // Branch 2: Absolute paths and dot-relative paths (/, ./, ../).
@@ -461,6 +461,8 @@ test "url regex" {
         "foo/bar,baz.txt",
         // $VAR should not match mid-word
         "foo$BAR/baz.txt",
+        // ~ should not match mid-word
+        "foo~/bar.txt",
     };
     for (no_match_cases) |input| {
         var result = re.search(input, .{});
--- a/src/config/url.zig
+++ b/src/config/url.zig
@@ -53,11 +53,11 @@ const trailing_spaces_at_eol =
 ;
 
 const dotted_path_lookahead =
-    \\(?=[\w\-.~:\/?#@!$&*+,;=%]*\.)
+    \\(?=[\w\-.~:\/?#@!$&*+;=%]*\.)
 ;
 
 const non_dotted_path_lookahead =
-    \\(?![\w\-.~:\/?#@!$&*+,;=%]*\.)
+    \\(?![\w\-.~:\/?#@!$&*+;=%]*\.)
 ;
 
 const dotted_path_space_segments =
@@ -369,6 +369,11 @@ test "url regex" {
             .input = "some-pkg/src/file.txt more text",
             .expect = "some-pkg/src/file.txt",
         },
+        // comma should match substrings
+        .{
+            .input = "src/foo.c,baz.txt",
+            .expect = "src/foo.c",
+        },
         .{
             .input = "~/foo/bar.txt",
             .expect = "~/foo/bar.txt",
@@ -452,6 +457,8 @@ test "url regex" {
         "$10/bar",
         "$10/$20",
         "$10/bar.txt",
+        // comma should not let dot detection look past it
+        "foo/bar,baz.txt",
     };
     for (no_match_cases) |input| {
         var result = re.search(input, .{});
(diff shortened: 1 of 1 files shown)


bkircher:
I think there are still some rough edges in this PR and some more test cases to be added. But I am too tired now to focus. Let me mark this PR as a draft and finish this another time. Sorry for the noise.

The build failure looks more like an x86 Nix/build-environment issue to me?

bkircher:
OK, I think most of the edge cases are covered.

bkircher:
Rebased on main and fixed one more over-match reported here: https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-3882254792

percontation:
Ran into another couple cases you might be interested in fixing here (I believe they all have the same issue on this branch and upstream)
* `echo /tmp/foo /tmp/bar` selects `/tmp/foo /tmp/bar`
* `echo /tmp/foo.txt /tmp/bar.txt` selects `/tmp/foo.txt /tmp/bar.`
* `echo /tmp/Test.app/path component/file.txt` selects `/tmp/Test.app/path component/file.`

I guess a space should probably not be considered valid as the final character in a path component, for the first two cases. I dunno what's up with the third case.

mitchellh:
!vouch

mitchellh (approved):
Thank you! I'm going to continue to accept the whack-a-mole here to make this better (as I mentioned in other posts that I want to do something better for file paths than regex, but this is what we have for now).

bkircher:
It really is whack-a-mole 😄

bkircher:
@percontation 

Thanks for those. I will address them in a new PR.

> I guess a space should probably not be considered valid as the final character in a path component, for the first two cases. I dunno what's up with the third case.

The third case is actually by design. Right now, the code interprets 
```
/tmp/test folder/file.txt
```
as 
```
/tmp/test\ folder/file.txt
```

Not sure what to do about this right now.

percontation:
The issue with the third `/tmp/Test.app/path component/file.txt` case I listed is that the regex cuts off the `txt` at the end; it's like, the match of a path with spaces is somehow terminated by the second `.` in it? Seems weird. I stared at the regex for like a minute and didn't see why, I was hoping the fix would be more obvious to you 😅.

(re whack-a-mole: IIRC iTerm2 combines regex with file existence checks. Because yeah, there's inherent ambiguity when trying to solve this problem with regex alone.)

bkircher:
@percontation 

> (re whack-a-mole: IIRC iTerm2 combines regex with file existence checks. Because yeah, there's inherent ambiguity when trying to solve this problem with regex alone.)

Interesting, I'll check this out.

ignatremizov:
Ran into a few more over-match cases locally that seem to fit the same family:

- `/tmp/ping.txt now.` selects `/tmp/ping.txt now.` instead of `/tmp/ping.txt`
- `/tmp/foo/ping.txt.` selects `/tmp/foo/ping.txt.` instead of `/tmp/foo/ping.txt`
- `/tmp/foo && mv /tmp/bar` selects `/tmp/foo && mv` instead of `/tmp/foo`

My current intuition is that paths should probably be treated differently depending on whether they are quoted:

- quoted paths can stay broad, including spaces and punctuation inside the quotes
- unquoted paths should be more conservative and stop before prose/shell syntax such as trailing `.` or `&&`

That would still allow things like `"/tmp/test folder/file.txt"` while avoiding the prose/shell over-match cases above.

I also think file existence checks would be the way to go, except that is highly environment dependent and then you handle a different family of whack-a-mole problems.

PR #10624, opened 2026-02-08
chore: sync with nixpkgs

gigamonster256:
updates nixpkgs and home-manager

X11 packages are now not in a shared namespace

the ZIG_GLOBAL_CACHE_DIR creation was moved to the configurePhase in the zig hook

moves nix shell from #devShell.\<system\> to #devshells.\<system\>.default

tested with `nix run` and `nix flake check` on x86-64_linux

no AI used


jcollie (approved):
LGTM

PR #10628, opened 2026-02-09
feat(nix): enhance overlays

gigamonster256:
uses the `final` fixed point package set as the source of dependencies for the ghostty package created by the overlay - allows simpler overriding of dependencies when using the overlay

no AI used


jcollie (approved):
LGTM

PR #10631, opened 2026-02-09
i18n: ignore non-UTF-8 LANGUAGE and use next valid locale env

EriksRemess:
Currently my Ubuntu 25.10 sets `LANGUAGE=lv:en` which overrides otherwise valid locale settings like `LC_ALL=lv_LV.UTF-8`; this change updates `ensureLocale` function to unset `LANGUAGE` when it does not contain `.UTF-8`, then proceeds with normal locale initialization so a valid locale is applied.

### DIFF
--- a/src/os/locale.zig
+++ b/src/os/locale.zig
@@ -28,6 +28,20 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
         }
     }
 
+    // gettext's LANGUAGE overrides LC_ALL/LC_MESSAGES/LANG. If it does not
+    // include an explicit UTF-8 locale, ignore it so gettext falls back to
+    // LC_ALL/LC_MESSAGES/LANG.
+    if ((try internal_os.getenv(alloc, "LANGUAGE"))) |language| {
+        defer language.deinit(alloc);
+        if (language.value.len > 0 and
+            !std.mem.endsWith(u8, language.value, ".UTF-8"))
+        {
+            _ = internal_os.setenv("LANGUAGE", "");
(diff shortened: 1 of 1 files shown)


src/os/locale.zig
```diff
@@ -28,6 +28,20 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
         }
     }
 
+    // gettext's LANGUAGE overrides LC_ALL/LC_MESSAGES/LANG. If it does not
+    // include an explicit UTF-8 locale, ignore it so gettext falls back to
+    // LC_ALL/LC_MESSAGES/LANG.
+    if ((try internal_os.getenv(alloc, "LANGUAGE"))) |language| {
+        defer language.deinit(alloc);
+        if (language.value.len > 0 and
+            std.mem.indexOf(u8, language.value, ".UTF-8") == null)
```
pluiedev: I think `std.mem.endsWith(u8, language.value, ".UTF-8")` makes more sense here

pluiedev (commented):
The idea appears to be sound but I would have to investigate more on what the expected syntax for `LANGUAGE` is

src/os/locale.zig
EriksRemess: switched to .endsWith comparison

EriksRemess:
Looks like proper way would be not to include ".UTF-8" in .po file names. At least not on Linux.
Instead proper way would be for Latvian - lv.po, for languages with variants like Brazilian Portuguese - pt_BR.po
Then would use LANGUAGE env, split it by ":" and use locales by preference. In my case - lv (Latvian) first, then en (English) second.

Issue #10632, opened 2026-02-09
Ghostty 1.3 Translation Tracking Issue


trag1c:
> [!important]
> Please also read [the pinned comment](https://github.com/ghostty-org/ghostty/issues/10632#issuecomment-3919649650).

This issue tracks the status of translations for the upcoming release in March. The task list below serves as a coverage checklist; this time, all locales listed below are missing the same 11 strings. =)

Missing translations are not a release blocker, but the goal is to have as many locales updated as possible. Localization team members are pinged as a heads-up, but **anyone can contribute and review translations.**

## Locales with missing translations

- [x] `bg_BG`: @damyanbogoev @phush0 @reo101 — #10639
- [x] `ca_ES`: @KristoferSoler — #10845
- [x] `de_DE`: @khipp @rpfaeffle — #10698
- [x] `es_AR`: @piedrahitac @alanmoyano — #10641
- [x] `es_BO`: @piedrahitac @MiguelElGallo — #10675
- [x] `fr_FR`: @flou @Pangoraw @gagbo @Kirwiisp — #10636
- [x] `ga_IE`: @aindriu80 @OshDubh — #10673
- [x] `he_IL`: @slsrepo @CraziestOwl — #10676
- [x] `hr_HR`: @Filip7 @kristina8888 @MatkoTiric — #10649
- [x] `hu_HU`: @d-Dudas @balazs-szucs — #10648
- [x] `id_ID`: @mikailmm @halosatrio — #10794 #11226
- [x] `it_IT`: @Misairuzame — #10708
- [x] `ja_JP`: @tnagatomi @lonsagisawa @kawarimidoll — #10655
- [x] `ko_KR`: @zenyr @RMEngelbrecht @uhojin — #10656
- [x] `lt_LT`: @tdslot @abudvytis — #10644
- [x] `mk_MK`: @andrejdaskalov @marijagjorgjieva — #10686
- [x] `nb_NO`: @hqnna @Uzaaft — #10702
- [x] `nl_NL`: @kjvdven @mtak @nwehg — #10637
- [x] `pl_PL`: @Secrus @johnslavik — #10665
- [x] `pt_BR`: @guilhermetk @silveirapf — #10635
- [x] `ru_RU`: @TicClick @zeshi09 — #10633
- [x] `tr_TR`: @bitigchi @ghokun @Beryesa — #10634
- [x] `uk_UA`: @gmile @gordonbondon @chernetskyi @danulqua — #10638
[...]

PR #10636, opened 2026-02-09
i18n: Add fr_FR translations

gagbo:
Preparing 1.3 release

Ref: #10632

### DIFF
--- a/po/fr_FR.UTF-8.po
+++ b/po/fr_FR.UTF-8.po
@@ -8,8 +8,8 @@ msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2026-02-05 10:23+0800\n"
-"PO-Revision-Date: 2025-08-23 21:01+0200\n"
-"Last-Translator: Kirwiisp <swiip__@hotmail.com>\n"
+"PO-Revision-Date: 2026-02-09 21:18+0200\n"
+"Last-Translator: Gerry Agbobada <gerry@gagbo.net>\n"
 "Language-Team: French <traduc@traduc.org>\n"
 "Language: fr\n"
 "MIME-Version: 1.0\n"
@@ -93,11 +93,11 @@ msgstr "Chercher…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
-msgstr "Précédent"
+msgstr "Résultat Précédent"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:74
 msgid "Next Match"
-msgstr "Suivant"
+msgstr "Résultat Suivant"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:6
 msgid "Oh, no."
@@ -105,7 +105,7 @@ msgstr "Oh, non."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:7
 msgid "Unable to acquire an OpenGL context for rendering."
-msgstr "Impossible d’obtenir un contexte OpenGL pour le rendu."
+msgstr "Impossible d'obtenir un contexte OpenGL pour le rendu."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:97
 msgid ""
@@ -113,8 +113,8 @@ msgid ""
 "through the content, but no input events will be sent to the running "
 "application."
 msgstr ""
-"Ce terminal est en mode lecture seule. Vous pouvez encore voir, selectionner,"
-"et naviguer dans son contenu, mais aucune entrée ne sera envoyée à l’application"
+"Ce terminal est en mode lecture seule. Vous pouvez encore voir, sélectionner, "
+"et naviguer dans son contenu, mais aucune entrée ne sera envoyée à l'application "
 "en cours."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:107
--- a/po/fr_FR.UTF-8.po
+++ b/po/fr_FR.UTF-8.po
@@ -97,7 +97,7 @@ msgstr "Résultat précédent"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:74
 msgid "Next Match"
-msgstr "Résultat Suivant"
+msgstr "Résultat suivant"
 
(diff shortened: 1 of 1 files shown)


po/fr_FR.UTF-8.po
00-kat: (Just a reminder to update this date to the last edit once you're about to mark the PR ready for review.)

po/fr_FR.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: Inspecteur"
...
 
 #: src/apprt/gtk/ui/1.2/surface.blp:97
 msgid ""
 "This terminal is in read-only mode. You can still view, select, and scroll "
 "through the content, but no input events will be sent to the running "
 "application."
 msgstr ""
+"Ce terminal est en mode lecture seule. Vous pouvez encore voir, selectionner,"
+"et naviguer dans son contenu, mais aucune entrée ne sera envoyée à l’application"
+"en cours."
```
flou:
```suggestion
"Ce terminal est en mode lecture seule. Vous pouvez encore voir, sélectionner, "
"et naviguer dans son contenu, mais aucune entrée ne sera envoyée à l’application "
"en cours."
```
Just a minor change, also I think the whitespace at the end of each line is needed.

po/fr_FR.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: Inspecteur"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:29
 msgid "Find…"
-msgstr ""
+msgstr "Chercher…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
-msgstr ""
+msgstr "Précédent"
```
flou:
```suggestion
msgstr "Résultat précédent"
```

po/fr_FR.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: Inspecteur"
...
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:74
 msgid "Next Match"
-msgstr ""
+msgstr "Suivant"
```
flou:
```suggestion
msgstr "Résultat suivant"
```

po/fr_FR.UTF-8.po
00-kat: Yeah it is, otherwise that will be joined as `sélectionner,et naviguer` and `l’applicationen cours`.

flou (commented):
Not too sure about this but if the english string has "match" in it maybe "Résultat précédent/suivant" would be a closer translation.

po/fr_FR.UTF-8.po
ldemailly: je me suis demande' la meme chose et ca depends si le contexte est evident (ie still inside search dialog) a screen shot would help deciding

gagbo: I opened the search in `Kate` for that, and they just use "Suivant/Précédent". I thought it’d be nice to avoid being too verbose in the UI, and the context is `search-overlay` so I thought it’d be fine

image

gagbo: But résulat feels ~~better~~ more correct, albeit longer

po/fr_FR.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: Inspecteur"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:29
 msgid "Find…"
-msgstr ""
+msgstr "Chercher…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
-msgstr ""
+msgstr "Résultat Précédent"
```
flou:
```suggestion
msgstr "Résultat précédent"
```

po/fr_FR.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: Inspecteur"
...
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:74
 msgid "Next Match"
-msgstr ""
+msgstr "Résultat Suivant"
```
flou:
```suggestion
msgstr "Résultat suivant"
```

flou (commented):
There is no need to capitalize "précédent/suivant" in those cases.
I think in english each word is capitalized but in french usually only the first one is (it's already the case for other strings in this localization file).

po/fr_FR.UTF-8.po
ldemailly: french tends to be quite overly verbose so I'd think we could take opportunity to shorten/use fewer words to save screen space if the context isn't ambiguous

gagbo: So we’re back to "Précédent" / "Suivant" instead of "Résulat *"? Fine by me, I like shorter versions too to be honest

flou:
It's all good for me. Thank you @gagbo ❤️

gagbo:
> It's all good for me. Thank you @gagbo ❤️

Yeah, we just need to decide if "Résultat" is necessary or not for the search overlay, but then again, this might get changed later

gagbo:
Disclaimer: I don’t speak GTK fluently

So, looking at the [overlay blueprint](https://github.com/ghostty-org/ghostty/blob/41924c2d0ec50a4268250b90cf46d7b4e60b61ba/src/apprt/gtk/ui/1.2/search-overlay.blp#L64), it seems that the text is actually only displayed on hover in a tooltip, so I’d rather keep the more explicit "Résultat ..." in the text.

It’s weird that the next button is plugged to the `previous_match` trigger, and that the previous button is plugged to the `next_match` trigger, but then again, I don’t speak GTK and I don’t know if the semantics get reversed somehow by the direction of the scrollback

flou:
> So, looking at the [overlay blueprint](https://github.com/ghostty-org/ghostty/blob/41924c2d0ec50a4268250b90cf46d7b4e60b61ba/src/apprt/gtk/ui/1.2/search-overlay.blp#L64), it seems that the text is actually only displayed on hover in a tooltip, so I’d rather keep the more explicit "Résultat ..." in the text.

Agreed, this also keeps it consistent with the translations used in other locales.

flou (approved):

trag1c (approved):

PR #10637, opened 2026-02-09
i18n: Update translations for nl_NL

nwehg:
Adding new translations for Dutch

### DIFF
--- a/po/nl_NL.UTF-8.po
+++ b/po/nl_NL.UTF-8.po
@@ -93,11 +93,11 @@ msgstr "Zoeken…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
-msgstr "Vorig resultaat"
+msgstr "Vorige resultaat"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:74
 msgid "Next Match"
-msgstr "Volgend resultaat"
+msgstr "Volgende resultaat"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:6
 msgid "Oh, no."
--- a/po/nl_NL.UTF-8.po
+++ b/po/nl_NL.UTF-8.po
@@ -105,7 +105,7 @@ msgstr "Oh, nee."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:7
 msgid "Unable to acquire an OpenGL context for rendering."
-msgstr "Kan geen OpenGL-context voor rendering aanmaken."
+msgstr "OpenGL-context voor rendering aanmaken mislukt."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:97
 msgid ""
@@ -113,9 +113,8 @@ msgid ""
 "through the content, but no input events will be sent to the running "
 "application."
 msgstr ""
-"Deze terminal is in alleen-lezen modus. Je kunt de inhoud nog steeds bekijken, "
-"selecteren en erdoorheen scrollen, maar invoer zal niet naar de applicatie"
-"gestuurd worden"
+"Deze terminal staat in alleen-lezen modus. Je kunt de inhoud nog steeds bekijken "
+"en selecteren, maar er wordt geen invoer naar de applicatie verzonden."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:107
 msgid "Read-only"
@@ -131,7 +130,7 @@ msgstr "Plakken"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:270
 msgid "Notify on Next Command Finish"
-msgstr "Aangeven wanneer het volgende commando is afgerond"
+msgstr "Meld wanneer het volgende commando is afgerond"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:277 src/apprt/gtk/ui/1.5/window.blp:266
 msgid "Clear"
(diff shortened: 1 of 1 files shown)


po/nl_NL.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: terminalinspecteur"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:29
 msgid "Find…"
-msgstr ""
+msgstr "Zoeken…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
-msgstr ""
+msgstr "Vorig resultaat"
```
kjvdven: I would prefer; "Vorige resultaat"

po/nl_NL.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: terminalinspecteur"
...
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:74
 msgid "Next Match"
-msgstr ""
+msgstr "Volgend resultaat"
```
kjvdven: I would prefer; "Volgende resultaat"

po/nl_NL.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: terminalinspecteur"
...
 
 #: src/apprt/gtk/ui/1.2/surface.blp:6
 msgid "Oh, no."
-msgstr ""
+msgstr "Oh, nee."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:7
 msgid "Unable to acquire an OpenGL context for rendering."
-msgstr ""
+msgstr "Kan geen OpenGL-context voor rendering aanmaken."
```
kjvdven: "rendering" could be "weergave"

po/nl_NL.UTF-8.po
```diff
@@ -89,34 +89,37 @@ msgstr "Ghostty: terminalinspecteur"
...
 
 #: src/apprt/gtk/ui/1.2/surface.blp:97
 msgid ""
 "This terminal is in read-only mode. You can still view, select, and scroll "
 "through the content, but no input events will be sent to the running "
 "application."
 msgstr ""
+"Deze terminal is in alleen-lezen modus. Je kunt de inhoud nog steeds bekijken, "
```
kjvdven: The sentence doesn't flow, maybe something like:

Deze terminal staat in de alleen-lezen modus. Je kunt nog steeds de inhoud bekijken, selecteren en doorbladeren, maar er wordt geen invoer naar de actieve applicatie verzonden.

po/nl_NL.UTF-8.po
```diff
@@ -128,7 +131,7 @@ msgstr "Plakken"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:270
 msgid "Notify on Next Command Finish"
-msgstr ""
+msgstr "Aangeven wanneer het volgende commando is afgerond"
```
kjvdven: Meld wanneer het volgende commando is afgerond
-or-
Meld wanneer de volgende opdracht is afgerond


Personally I like "opdracht", but depends I think if the user or the system is the source.

kjvdven (commented):
I looked it over, and dropped some suggestions

po/nl_NL.UTF-8.po
kjvdven: Ah "commando" is used more in the translations. 

"opdracht" is only used in translation for "command palet"

po/nl_NL.UTF-8.po
nwehg: I agree with that, but maybe we can just remove "doorbladeren" because it doesn't really make sense since your terminal isn't paper.

po/nl_NL.UTF-8.po
nwehg: It seems like we use both terms, so we should probably make it one of theme everywhere. I personally like "commando" more and since that's used the most we should probably keep it like that.

nwehg: I agree with "Meld" instead of "Aangeven" btw

po/nl_NL.UTF-8.po
nwehg: Agreed

po/nl_NL.UTF-8.po
nwehg: Agreed

po/nl_NL.UTF-8.po
nwehg: It could be that, but it doesn't have the same meaning exactly. Rendering is the process of "weergeven" and "weergave" is the result

po/nl_NL.UTF-8.po
kjvdven: True, I agree

```
inhoud bekijken, selecteren en scrollen
```

or, nice and short

```
inhoud bekijken en selecteren
```

po/nl_NL.UTF-8.po
kjvdven: Let's go with 

```
Meld wanneer het volgende commando is afgerond
```

po/nl_NL.UTF-8.po
kjvdven: Good point.

One more thing, does this make more sense if it would say `mislukt` instead of `kan geen`. 

```
OpenGL-context voor rendering aanmaken mislukt
```

nwehg: I agree with this one as it's shorter

po/nl_NL.UTF-8.po
nwehg: Let's go with the second one

nwehg:
I updated the translations according to your suggestions @kjvdven. Thank you for taking a look at it. I think we are now ready to merge?

kjvdven (approved):

trag1c (approved):

PR #10640, opened 2026-02-09
i18n/zh_CN: update for 1.3

pluiedev:
Part of #10632

Given I'm the only member in the zh_CN team, y'all just have to trust me on this one... Wouldn't mind the zh_TW folks taking a look of course

### DIFF
--- a/po/zh_CN.UTF-8.po
+++ b/po/zh_CN.UTF-8.po
@@ -85,7 +85,7 @@ msgstr "Ghostty 终端调试器"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:29
 msgid "Find…"
-msgstr "查找……"
+msgstr "查找…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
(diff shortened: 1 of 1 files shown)


po/zh_CN.UTF-8.po
00-kat:
```suggestion
"PO-Revision-Date: 2026-02-09 21:10+0100\n"
```

00-kat:
> Wouldn't mind the zh_TW folks taking a look of course

cc @PeterDaveHello @pan93412 and @charliie-dev in case you want to take a look; you don't *need* to, of course.

po/zh_CN.UTF-8.po
```diff
@@ -85,34 +85,36 @@ msgstr "Ghostty 终端调试器"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:29
 msgid "Find…"
-msgstr ""
+msgstr "查找……"
```
pan93412: Usually, we don't use double ellipses for menu items that require other actions: https://developer.apple.com/design/human-interface-guidelines/menus

I believe one ellipsis should enough. What are your thoughts?

pluiedev: I need to double check how e.g. official GNOME apps do it, but honestly single ellipses work fine too

pluiedev: Ok GTK itself seems to prefer single ellipses. I'll go ahead and fix that.

charliie-dev (approved):
LGTM as a zh_TW team member.

00-kat (approved):

PR #10642, opened 2026-02-09
gtk: fix capitalization of banner title

jcollie:


po/bg.po
```diff
@@ -337,3 +331,9 @@ msgstr "Клипбордът е изчистен"
 #: src/apprt/gtk/class/window.zig:1695
 msgid "Ghostty Developers"
 msgstr "Разработчици на Ghostty"
+
+#~ msgid "Command succeeded"
+#~ msgstr "Командата завърши успешно"
```
00-kat: Shouldn't these be moved to the ones that are left?

OshDubh: Especially when non-English languages tend to use that casing for titles anyway.

00-kat: Yep. (I've asked for every single case of translations being in title case, and not a single one so far had title case as a prominent feature; it's very much an English-ism.)

jcollie: I have no idea how gettext fixes up changes like that. My guess is that it isn't smart enough. Sorry if that creates more work for the translators.

00-kat: I thought if you just (manually) replace the new `msgstr ""` with the commented out `msgstr "whatever"` and delete the newly added `#~` comments it'll Just Work™

pluiedev (changes requested):
Hmm. The HIG does say to prefer the "informal heading" style for banners, which means our current text is correct. IMO we should just add a translator's comment that this rule really only applies to English and they should feel free to render the two strings identically in their language.

https://developer.gnome.org/hig/patterns/feedback/banners.html

Plus if we were to go with this PR we should just delete the duplicate entries instead of keeping them but as comments

jcollie:
> Hmm. The HIG does say to prefer the "informal heading" style for banners, which means our current text is correct. IMO we should just add a translator's comment that this rule really only applies to English and they should feel free to render the two strings identically in their language.
> 
> https://developer.gnome.org/hig/patterns/feedback/banners.html

If you follow the links, heading style says that all (significant) words should be capitalized (at least in English) so this PR seems to be "more correct".

https://developer.gnome.org/hig/guidelines/writing-style.html#heading-style

> Plus if we were to go with this PR we should just delete the duplicate entries instead of keeping them but as comments

po/zh_CN.po
```diff
@@ -369,3 +363,9 @@ msgstr "已清空剪贴板"
 #: src/apprt/gtk/class/window.zig:1869
 msgid "Ghostty Developers"
 msgstr "Ghostty 开发团队"
+
+#~ msgid "Command succeeded"
```
pluiedev: Remove the fuzzy entries (all lines beginning with `#~`)

pluiedev (changes requested):

trag1c (approved):
LGTM pending Leah's comment

po/zh_CN.po
jcollie: Done

jcollie: Is this something that should be done automatically by `zig build update-translations`?

pluiedev: I think there's a flag to disallow *new* fuzzy entries but I don't know if there's one that removes old orphaned entries like that

pluiedev (approved):

PR #10643, opened 2026-02-10
shell-integration: respect cursor-style-blink

### LINKED ISSUE #8681: config: default `shell-integration-features = no-cursor` if `cursor-style` is manually set

jparise:
The `cursor` shell feature always used a blinking bar (beam), often to the surprise of users who set `cursor-style-blink = false`.

This change extends our GHOSTTY_SHELL_FEATURES format to include either `cursor:blink` (default) or `cursor:steady` based on cursor-style-blink when the `cursor` feature is enabled, and all shell integrations have been updated to use that additional information to choose the DECSCUSR cursor value (5=blinking bar, 6=steady bar).

I also considered passing a DECSCUSR value in GHOSTTY_SHELL_FEATURES (e.g. `cursor:5`). This mostly worked well, but zsh also needs the blink state on its own for its block cursor. We also don't support any other shell feature cursor configurability (e.g. the shape), so this was an over generalization.

This does change the behavior for users who like the blinking bar in the shell but have `cursor-blink-style = false` for other reasons. We could provide additional `cursor` shell feature configurability (e.g. `cursor:blink` in `shell-integration-features`), but I'll propose that as its own change.

See: #2812
Closes: #8681

---

**AI Disclosure:** I did a lot of rubber ducking with Claude Code while trying out various ideas. It was particularly useful for this kind of feature because I could try out one thing and have it evaluate the impact on all of the shell integration scripts at once.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -632,6 +632,7 @@ pub fn init(
             .env_override = config.env,
             .shell_integration = config.@"shell-integration",
             .shell_integration_features = config.@"shell-integration-features",
+            .cursor_blink = config.@"cursor-style-blink",
             .working_directory = config.@"working-directory",
             .resources_dir = global_state.resources_dir.host(),
             .term = config.term,
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2710,7 +2710,7 @@ keybind: Keybinds = .{},
 ///
 /// Available features:
 ///
-///   * `cursor` - Set the cursor to a blinking bar at the prompt.
+///   * `cursor` - Set the cursor to a bar at the prompt.
 ///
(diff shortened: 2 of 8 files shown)


jparise:
https://github.com/jparise/ghostty/tree/shell-cursor-config has some initial work to support configurable `cursor` styles in `shell-integration-features`.

mitchellh (approved):

PR #10687, opened 2026-02-12
bash: reset cursor in preexec instead of PS0

jparise:
The recent move to PS0-based precmd/preexec hooks means that we're overwriting PS0 after the cursor reset sequence is added. This caused the input (bar) cursor to never be reset and "leak" into other commands, like vim.

Fix this by moving the cursor reset sequence from precmd to preexec.


jparise:
#10688 improves our PS0 handling in general, including fixing this issue.

PR #10685, opened 2026-02-12
i18n: add missing `nb_NO` translations

hqnna:
Adding the missing Norwegian (`nb_NO`) translations for #10632.

PR #10702, opened 2026-02-13
i18n: add missing `nb_NO` translations

hqnna:
New pull request because the old one got fucked up by rebase...

### DIFF
--- a/po/nb_NO.UTF-8.po
+++ b/po/nb_NO.UTF-8.po
@@ -93,11 +93,11 @@ msgstr "Finn…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
-msgstr "Forrige Treff"
+msgstr "Forrige treff"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:74
 msgid "Next Match"
-msgstr "Neste Treff"
+msgstr "Neste treff"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:6
 msgid "Oh, no."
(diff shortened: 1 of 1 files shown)


Uzaaft:
#10685 is the old one for reference.

Uzaaft (approved):

po/nb_NO.UTF-8.po
```diff
@@ -128,7 +131,7 @@ msgstr "Lim inn"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:270
 msgid "Notify on Next Command Finish"
-msgstr ""
+msgstr "Varsle Når Neste Kommandoen Fullføres"
```
trag1c: Does Norwegian actually use title case like this? Looking at older strings in this file it doesn't seem to be the case. English is pretty consistently the only language to do so, with pretty much all the other ones using regular sentence case.

trag1c (changes requested):

po/nb_NO.UTF-8.po
hqnna: > Does Norwegian actually use title case like this? Looking at older strings in this file it doesn't seem to be the case. English is pretty consistently the only language to do so, with pretty much all the other ones using regular sentence case.

Usually no, I only did it to keep it consistent with the original string.

trag1c: In that case please correct the strings to use sentence case. 🙂 You can see other locales (and older strings in this locale) do the same.

trag1c (approved):

jcollie:
!vouch

trag1c:
!vouch @uzaaft

PR #10708, opened 2026-02-14
i18n: add 1.3 it_IT translations

Misairuzame:
Added the new Italian translations for 1.3.0. Any feedback is appreciated, as always!
cc: @Atomk


Atomk:
Thanks for tagging me! I have a nitpick, I think I'd prefer "Comando terminato" instead of "Il comando è terminato", same for the other two similar instances. Assuming these are notifications/alerts (I sometimes get these sentences in Gnome terminal alerts), I think the shorter form says the same thing with less words, and the meaning is also slightly different - "Comando terminato" says "a" command is terminated, which has the benefit of making sense even if there are multiple commands running at the same time, while the longer form says "the" command has terminated, implying the user knows which was "the" terminated command or at least that more info is provided on what "the" is. This might be personal taste, everything else looks great!

00-kat:
> Assuming these are notifications/alerts

They look a bit like this:

**`ghostty --command=foo`**:
Screenshot of Ghostty showing `/bin/sh: 1: foo: not found`, along with a red bar at the bottom with the text “Command failed” alongside a “Close” button.

**`ghostty --wait-after-command=true -e sleep 5`** ([docs]):
Same as the above, but there's no output, the bar is green, and the text is “Command succeeded”.

> implying the user knows which was "the" terminated command

And yeah, I think the user does know which command terminated in the context it's shown in, since it's the initial command that was run (and there can only be one of those, usually the shell).

[docs]: https://ghostty.org/docs/config/reference#wait-after-command

Misairuzame:
@Atomk thank you very much for your suggestions! I applied them in the last commit.

00-kat:
Three new strings have been merged since you opened this PR:

  - `"Change Tab Title…"` (used in the context menu)
  - `"Change Tab Title"` (used as the title of the dialog box that opens when the action is triggered)
  - `"Open in Ghostty"` (shown in GNOME Files' context menu)

Please merge/rebase this PR and also translate them, thank you! I've written up more information at https://github.com/ghostty-org/ghostty/issues/10632#issuecomment-3919649650.

00-kat:
@Atomk: as a sidenote, you are still not in the it_IT localization team; Mitchell sent you an invite three days ago, so please check your emails! (It might have moved itself to the spam folder, so check there if you can't find it; the invite expires in four days from now.)

Misairuzame:
Merged and translated the three missing strings.

Quick rationale for my choices:
- Change Tab Title… -> Cambia titolo scheda…
  - Matches Split Context Menu translation "Change title…" -> "Cambia titolo…"
- Change Tab Title -> Cambia il titolo della scheda
  - Popup window title which matches the translation of "Change Terminal Title" -> "Cambia il titolo del terminale"

As always, I'm open to suggestions!

Misairuzame:
@Atomk if these last strings look good to you I think this PR can be merged!

Atomk:
@00-kat I've seen it now, thanks!

@Misairuzame Thanks for stating the reason behind your choices, I generally prefer the concept of "rename" over "change title" (even though I think I get why it's like that) but I agree that the best choice here is to stay close to the original and being consistent, which you did. Looks good!

00-kat (approved):

PR #10709, opened 2026-02-14
fix(zsh): extend ZDOTDIR through startup to survive .zprofile exec (unblocks AWS Kiro)

neo773:
Demo of `inherit-working-directory` working with AWS Kiro

https://github.com/user-attachments/assets/99526b43-2d40-4fc3-b7b5-9e076bc486c3

Tools like AWS Kiro (formerly Fig/CodeWhisperer) exec their PTY wrapper from .zprofile, replacing the shell process entirely. Since Ghostty previously restored ZDOTDIR in .zshenv and never controlled .zprofile or .zshrc, the new shell spawned by the PTY wrapper had no Ghostty ZDOTDIR and all shell integration (OSC 7, OSC 133, cursor shape, etc.) was silently lost.

Fix by extending Ghostty's ZDOTDIR control through .zprofile and .zshrc:

- .zshenv: keep ZDOTDIR pointing to Ghostty's dir for interactive shells after sourcing user's .zshenv and loading integration
- .zprofile (new): source user's .zprofile WITHOUT restoring ZDOTDIR, so exec'd processes inherit Ghostty's ZDOTDIR and integration loads in the new shell
- .zshrc (new): permanently restore ZDOTDIR, source user's .zshrc, then re-verify _ghostty_deferred_init is still in precmd_functions

LLM Disclosure, Code and Diagnostic was done by Opus 4.6 with me the human in the full loop and real testing

### DIFF
--- a/src/shell-integration/zsh/.zprofile
+++ b/src/shell-integration/zsh/.zprofile
@@ -1,0 +1,33 @@
+# Based on (started as) a copy of Kitty's zsh integration. Kitty is
+# distributed under GPLv3, so this file is also distributed under GPLv3.
+# The license header is reproduced below:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# This wrapper is sourced because Ghostty extends ZDOTDIR control through
+# the zsh startup sequence (.zshenv -> .zprofile -> .zshrc). It sources
+# the user's .zprofile while keeping ZDOTDIR pointing to Ghostty's dir.
+#
+# We intentionally do NOT restore ZDOTDIR before sourcing so that if the
+# user's .zprofile replaces the shell process (e.g. Kiro/Fig's PTY wrapper
+# exec), the new shell inherits Ghostty's ZDOTDIR and integration loads.
+#
+# This file can get sourced with aliases enabled. To avoid alias expansion
+# we quote everything that can be quoted. Some aliases will still break us
+# though.
+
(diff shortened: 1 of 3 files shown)


rhodes-b:
As per ghosttys AI policy, any submissions created with AI are automatically rejected without an accepted issue you are free to open a discussion with a link to the PR and a mantainer can promote it or give the go ahead if they agree with your solution

https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md#ai-usage-policy

neo773:
@rhodes-b 
Hi, 

I did had a prior discussion open some time ago but I got carried away and decided to poke around (100% on)
Added the PR to discussion here https://github.com/ghostty-org/ghostty/discussions/10581#discussioncomment-15803201

Thanks.

jparise (commented):
I understand what this is doing, but I want to think about it for a bit. Based on previous issues related to zsh integration, this area is particularly sensitive to regressions.

There are also a few issues we should resolve in the code itself. In particular, the new files that weren't derived from Kitty shouldn't reference that lineage and probably don't need to be GPLv3.

I can also see a few opportunities to clean up variable usage to make this a little bit more hermetic.

neo773:
Pushed a commit that removes the redundant Kitty license header and also renamed the variables better
For regression you can clone the test suite here https://github.com/neo773/ghost_test
there was one real regression with zsh history not working, pushed a new commit to fix it

jparise (changes requested):
After spending more time with the zsh shell integration recently, I don't think this is the right approach. I'm inclined to recommend users in this situation (where something outside of their control modifies `precmd_functions`) should manually source our shell integration:

```zsh
if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
  source "$GHOSTTY_RESOURCES_DIR"/shell-integration/zsh/ghostty-integration
fi
```

Here's my thinking:

- Adding these additional resource files that thread through our injected `ZDOTDIR` significantly increases the surface area for maintenance and regressions.
- `ZDOTDIR` is "wrong" for the entire `.zprofile` phase, meaning any code in the user's `.zprofile` (or `/etc/zprofile`) that checks `$ZDOTDIR` will see Ghostty's directory, not the user's. (The `HISTFILE` "fix" in `.zshrc` is an example of this.)
- The `precmd_functions` manipulation in `.zshrc` is fragile and something that's been challenging to get "right" in `.zshenv` itself (#11367).

neo773:
Closing this issue solved by adding the code block from https://github.com/ghostty-org/ghostty/discussions/10581#discussioncomment-16098984 in my zshrc

PR #10727, opened 2026-02-15
gtk: various blur-related fixes 

### LINKED ISSUE #10721: GTK/Wayland: Use the `ext-background-effects-v1` protocol for background blur
See #10719

It has been ~8 months since the PR was merged, and the next version of Niri would feature support for it. The API is very similar to KDE blur, only specified more clearly and cleanly.

pluiedev:
Fixes #10721 
Fixes #10720 

I recommend reviewing this per-commit as there's a lot of refactoring to go over. I do want to fix the "korner bug" but that would require a bit more reading on my part.

**This hasn't been tested on X11.** If you're able, please see if this still works on KWin X11. I'm currently in the countryside celebrating Chinese New Year, and I'm not really in the mood of downloading half a gig worth of KDE libraries x)

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -91,8 +91,8 @@
             .lazy = true,
         },
         .wayland_protocols = .{
-            .url = "https://deps.files.ghostty.org/wayland-protocols-258d8f88f2c8c25a830c6316f87d23ce1a0f12d9.tar.gz",
-            .hash = "N-V-__8AAKw-DAAaV8bOAAGqA0-oD7o-HNIlPFYKRXSPT03S",
+            .url = "https://gitlab.freedesktop.org/wayland/wayland-protocols/-/archive/1.47/wayland-protocols-1.47.tar.gz",
+            .hash = "N-V-__8AAFdWDwA0ktbNUi9pFBHCRN4weXIgIfCrVjfGxqgA",
             .lazy = true,
         },
         .plasma_wayland_protocols = .{
--- a/build.zig.zon.json
+++ b/build.zig.zon.json
@@ -139,6 +139,11 @@
     "url": "https://deps.files.ghostty.org/wayland-protocols-258d8f88f2c8c25a830c6316f87d23ce1a0f12d9.tar.gz",
     "hash": "sha256-XO3K3egbdeYPI+XoO13SuOtO+5+Peb16NH0UiusFMPg="
   },
+  "N-V-__8AAFdWDwA0ktbNUi9pFBHCRN4weXIgIfCrVjfGxqgA": {
+    "name": "wayland_protocols",
+    "url": "https://gitlab.freedesktop.org/wayland/wayland-protocols/-/archive/1.47/wayland-protocols-1.47.tar.gz",
+    "hash": "sha256-3S3xSrX0EDgleq7cxLX7msDuAY8/D5SvkJcCjmDTMiM="
+  },
   "N-V-__8AAAzZywE3s51XfsLbP9eyEw57ae9swYB9aGB6fCMs": {
     "name": "wuffs",
     "url": "https://deps.files.ghostty.org/wuffs-122037b39d577ec2db3fd7b2130e7b69ef6cc1807d68607a7c232c958315d381b5cd.tar.gz",
--- a/build.zig.zon.nix
+++ b/build.zig.zon.nix
@@ -307,6 +307,14 @@ in
       };
     }
     {
+      name = "N-V-__8AAFdWDwA0ktbNUi9pFBHCRN4weXIgIfCrVjfGxqgA";
+      path = fetchZigArtifact {
+        name = "wayland_protocols";
+        url = "https://gitlab.freedesktop.org/wayland/wayland-protocols/-/archive/1.47/wayland-protocols-1.47.tar.gz";
+        hash = "sha256-3S3xSrX0EDgleq7cxLX7msDuAY8/D5SvkJcCjmDTMiM=";
+      };
+    }
+    {
       name = "N-V-__8AAAzZywE3s51XfsLbP9eyEw57ae9swYB9aGB6fCMs";
       path = fetchZigArtifact {
         name = "wuffs";
--- a/build.zig.zon.txt
+++ b/build.zig.zon.txt
@@ -34,3 +34,4 @@ https://deps.files.ghostty.org/zig_wayland-1b5c038ec10da20ed3a15b0b2a6db1c21383e
 https://deps.files.ghostty.org/zlib-1220fed0c74e1019b3ee29edae2051788b080cd96e90d56836eea857b0b966742efb.tar.gz
 https://github.com/ivanstepanovftw/zigimg/archive/d7b7ab0ba0899643831ef042bd73289510b39906.tar.gz
 https://github.com/ocornut/imgui/archive/refs/tags/v1.92.5-docking.tar.gz
+https://gitlab.freedesktop.org/wayland/wayland-protocols/-/archive/1.47/wayland-protocols-1.47.tar.gz
--- a/flatpak/zig-packages.json
+++ b/flatpak/zig-packages.json
@@ -169,6 +169,12 @@
   },
   {
     "type": "archive",
+    "url": "https://gitlab.freedesktop.org/wayland/wayland-protocols/-/archive/1.47/wayland-protocols-1.47.tar.gz",
+    "dest": "vendor/p/N-V-__8AAFdWDwA0ktbNUi9pFBHCRN4weXIgIfCrVjfGxqgA",
+    "sha256": "dd2df14ab5f41038257aaedcc4b5fb9ac0ee018f3f0f94af9097028e60d33223"
+  },
+  {
+    "type": "archive",
     "url": "https://deps.files.ghostty.org/wuffs-122037b39d577ec2db3fd7b2130e7b69ef6cc1807d68607a7c232c958315d381b5cd.tar.gz",
     "dest": "vendor/p/N-V-__8AAAzZywE3s51XfsLbP9eyEw57ae9swYB9aGB6fCMs",
     "sha256": "9e4cd20abe96e6c4c6ede9c3057108860126e7be2e2c3e35515476c250be1c13"
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -1071,21 +1071,6 @@ pub const Window = extern struct {
         self.syncAppearance();
     }
 
-    fn propGdkSurfaceHeight(
-        _: *gdk.Surface,
-        _: *gobject.ParamSpec,
-        self: *Self,
-    ) callconv(.c) void {
-        // X11 needs to fix blurring on resize, but winproto implementations
-        // could do anything.
-        self.private().winproto.resizeEvent() catch |err| {
-            log.warn(
-                "winproto resize event failed error={}",
-                .{err},
-            );
-        };
-    }
-
     fn propIsActive(
         _: *gtk.Window,
         _: *gobject.ParamSpec,
@@ -1111,7 +1096,7 @@ pub const Window = extern struct {
         };
     }
 
-    fn propGdkSurfaceWidth(
+    fn propGdkSurfaceDims(
         _: *gdk.Surface,
         _: *gobject.ParamSpec,
         self: *Self,
@@ -1282,14 +1267,14 @@ pub const Window = extern struct {
             _ = gobject.Object.signals.notify.connect(
                 gdk_surface,
                 *Self,
-                propGdkSurfaceWidth,
+                propGdkSurfaceDims,
                 self,
                 .{ .detail = "width" },
             );
             _ = gobject.Object.signals.notify.connect(
                 gdk_surface,
                 *Self,
-                propGdkSurfaceHeight,
+                propGdkSurfaceDims,
                 self,
                 .{ .detail = "height" },
             );
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -1235,7 +1235,7 @@ pub const Window = extern struct {
     fn finalize(self: *Self) callconv(.c) void {
         const priv = self.private();
         priv.tab_bindings.unref();
-        priv.winproto.deinit(Application.default().allocator());
+        priv.winproto.deinit();
 
         gobject.Object.virtual_methods.finalize.call(
             Class.parent,
--- a/src/apprt/gtk/winproto.zig
+++ b/src/apprt/gtk/winproto.zig
@@ -117,9 +117,9 @@ pub const Window = union(Protocol) {
         };
     }
 
-    pub fn deinit(self: *Window, alloc: Allocator) void {
+    pub fn deinit(self: *Window) void {
         switch (self.*) {
-            inline else => |*v| v.deinit(alloc),
+            inline else => |*v| v.deinit(),
         }
     }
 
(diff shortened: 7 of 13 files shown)


pluiedev:
So trying to find a satisfactory way to implement rounded corners actually drove me somewhat insane. I had to jerry-rig my own cursed nested-Niri white-noise-backed test setup in order to determine exactly where the blur regions are.
This is the best that I've been able to get - while this looks OK with a solid border, it is by no means perfect. Note that this is extremely blown up (at 3x scaling on the Niri side) and at normal scaling this isn't really noticable.

Image

@YaLTeR thoughts? :P

1.5x scale for comparison:
image

And with a solid border (the color doesn't really matter)

1.5x: 
image

3x:
image

YaLTeR:
Yeah it's probably fine. Unfortunately due to ext-bg-effect region being in surface-local integer coordinates, the precision does not increase with the scale factor, but that's the best we can do with the current protocols (I think).

FYI in niri you can set
```kdl
blur {
    saturation 10
}
```
to make it obvious. But yeah, in normal scenarios it is much less noticeable.

I tried the current code in this PR but I guess it doesn't do corners yet.

Also, I see on the current PR that the region starts from the top-left of the geometry, but it should probably start from the terminal widget, since the headerbar is opaque anyway:

image

pluiedev:
Applying it just to the terminal widget is unfortunately really, really complicated. GTK 4 doesn't allow you to get the absolute position of a widget AFAICT, only relative to the parent widget, so you have to manually compute the offsets yourself. Our widget tree is really quite complicated so that isn't really feasible... The simplest way would be to just apply the blur to the entire window, even if it's somewhat wasteful on the compositor side

pluiedev:
Refined the rounding logic a bit more and this is fully ready on the Wayland side. Still have no way to test on X11.

YaLTeR:
At 3x scale pixels start visibly peeking out, perhaps some pixel offset isn't scaled properly?

image

YaLTeR:
Ah, actually it's the same at 1x:

image

I'd add 1-2 px of offset

YaLTeR:
Also idk if this is relevant but quitting Ghostty results in a bunch of leak reports:
```
error(gpa): memory address 0x7f83490dc800 leaked:
/usr/lib/zig/std/array_list.zig:1231:56: 0x30be2d4 in ensureTotalCapacityPrecise (ghostty)
                const new_memory = try gpa.alignedAlloc(T, alignment, new_capacity);
                                                       ^
/usr/lib/zig/std/array_list.zig:1207:51: 0x307c93c in ensureTotalCapacity (ghostty)
            return self.ensureTotalCapacityPrecise(gpa, growCapacity(self.capacity, new_capacity));
                                                  ^
/usr/lib/zig/std/array_list.zig:1261:41: 0x3031154 in addOne (ghostty)
            try self.ensureTotalCapacity(gpa, newlen);
                                        ^
/usr/lib/zig/std/array_list.zig:894:49: 0x2fc287a in append (ghostty)
            const new_item_ptr = try self.addOne(gpa);
                                                ^
/var/home/yalter/source/zig/ghostty/src/apprt/gtk/winproto/blur.zig:133:30: 0x2f2d584 in approxRoundedRect (ghostty)
            try slices.append(alloc, .{
                             ^
/var/home/yalter/source/zig/ghostty/src/apprt/gtk/winproto/blur.zig:67:44: 0x2ed1e5a in calcForWindow (ghostty)
            .slices = try approxRoundedRect(
                                           ^
```

pluiedev:
Ah looks like I forgot to deinit some of these. Shucks

build.zig.zon
```diff
@@ -91,8 +91,8 @@
             .lazy = true,
         },
         .wayland_protocols = .{
-            .url = "https://deps.files.ghostty.org/wayland-protocols-258d8f88f2c8c25a830c6316f87d23ce1a0f12d9.tar.gz",
-            .hash = "N-V-__8AAKw-DAAaV8bOAAGqA0-oD7o-HNIlPFYKRXSPT03S",
+            .url = "https://gitlab.freedesktop.org/wayland/wayland-protocols/-/archive/1.47/wayland-protocols-1.47.tar.gz",
```
mitchellh, line 94: Freedesktop has terrible reliability, so when this is good to go let me know and I'll fix this before we merge.

I approve otherwise.

mitchellh (changes requested):
I approve, just my one comment.

build.zig.zon
pluiedev: I think it's a good idea to push the zip to S3 in any case since the ext-bg-effect commit will likely be merged regardless of the other two

YaLTeR:
Thanks, this looks much better. Though I suspect you may have added an extra 1 px of padding now:

image

YaLTeR:
And another thing: when the window is in a maximized or tiled state, the corners aren't rounded:

image

(in CSS this is defined [here](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/e98903e9b9408deb3f00cd95b3144baaa2cfcf64/src/stylesheet/widgets/_window.scss#L49-83) I believe)

(yeah, this should really be handled in GTK itself at some point...)

pluiedev:
GTK implementing something that will never land in GNOME? Imagine that! If that happened half of the Wayland winproto implementation straight up wouldn't need to exist!

...Aaanyway, I thought I could be a bit clever and simply read the `--window-radius` CSS variable GTK sets for itself. Turns out one does not simply read CSS variables in GTK. Bummer. Now I gotta reimplement that logic myself.

YaLTeR:
> GTK implementing something that will never land in GNOME?

I wouldn't assume that background effects would never land in GNOME

pluiedev:
Right, but before that happens I shall lean more on the pessimistic side :p I'll work on more specific criteria for disabling rounded corners now. Thanks for testing this too btw!

pluiedev:
As for the transparent "border", I believe that's actually the box shadow that [libadwaita applies globally](https://github.com/GNOME/libadwaita/blob/e98903e9b9408deb3f00cd95b3144baaa2cfcf64/src/stylesheet/widgets/_window.scss#L19-L34), so I don't think we should change any of it. Comes with the territory.

YaLTeR:
Thanks, looks good now!

pluiedev:
Excellent! I'm testing this on Plasma X11 right now and yeah, this is about what I expected - for *some* reason the secret KDE atoms defining blur regions are specified in device/physical pixels instead of logical/surface pixels, so if you run this branch on 2x scaling, only one quarter of the area would be blurred... More reasons to hate developing for (or rather in this case against) X11.

image

pluiedev:
image

... Why is Plasma Wayland having problems as well...?

@YaLTeR I think there's genuinely an issue here. I suspect that the KDE blur protocol implemented by KWin assumes device coordinates sans the CSD padding, while the impl in Niri assumes surface coordinates *with* the CSD padding (aka same semantics as ext-bg-effect). As there's pretty much no "real" specification of the KDE blur protocol I would probably recommend aligning Niri's impl to be the same as KDE's so that existing apps at least render correctly. Right now this PR looks correct under Niri but is terribly wrong under KDE.

YaLTeR:
But the latest KDE code is exactly the same between ext-background-effect and KDE blur?

https://invent.kde.org/plasma/kwin/-/blob/f3a4d9630662e325bc22047a3c1a3fda6ddd17a1/src/wayland/backgroundeffect_v1.cpp#L109-115

https://invent.kde.org/plasma/kwin/-/commit/7f9e0aa1cd2c2ab4fea1097800893b6f780089eb#5b56b3256a35f6b7844ddb50958917fa13fd345d_99_0

Actually, seems the KDE blur protocol got axed from KDE, so maybe I need to reconsider implementing it in niri

pluiedev:
Interesting... Is this part of any current Plasma release? If they're removing the custom protocol anyways, I think it's a better idea to keep the original behavior just for the old protocol so people on older Plasma/KWin versions will enjoy the same behavior, while Niri users and people using more recent Plasma releases can get the benefits of this PR.

ThatOneCalculator:
Getting this build error with this patch:

```
install
└─ install ghostty
   └─ compile exe ghostty ReleaseFast native-native
      └─ WriteFile helpgen.zig
         └─ run exe helpgen
            └─ compile exe helpgen Debug native 1 errors
src/apprt/gtk/winproto/blur.zig:116:8: error: duplicate struct member name 'approxRoundedRect'
    fn approxRoundedRect(
       ^~~~~~~~~~~~~~~~~
src/apprt/gtk/winproto/blur.zig:173:8: note: duplicate name here
    fn approxRoundedRect(
       ^~~~~~~~~~~~~~~~~
src/apprt/gtk/winproto/blur.zig:10:20: note: struct declared here
pub const Region = struct {
                   ^~~~~~
error: the following command failed with 1 compilation errors:
/usr/bin/zig build-exe -fno-strip -fno-omit-frame-pointer -funwind-tables --dep build_options -Mroot=/home/kainoa/.cache/yay/ghostty-git/src/ghostty/src/helpgen.zig -Mbuild_options=.zig-cache/c/1cb69cd90c9170ef647b3cb0f856b4e2/options.zig --cache-dir .zig-cache --global-cache-dir /home/kainoa/.cache/zig --name helpgen --zig-lib-dir /usr/lib/zig/ --listen=-
install
└─ install ghostty-vt
   └─ compile lib ghostty-vt ReleaseFast native-native 1 errors
src/apprt/gtk/winproto/blur.zig:116:8: error: duplicate struct member name 'approxRoundedRect'
    fn approxRoundedRect(
       ^~~~~~~~~~~~~~~~~
src/apprt/gtk/winproto/blur.zig:173:8: note: duplicate name here
    fn approxRoundedRect(
       ^~~~~~~~~~~~~~~~~
src/apprt/gtk/winproto/blur.zig:10:20: note: struct declared here
pub const Region = struct {
                   ^~~~~~
error: the following command failed with 1 compilation errors:
/usr/bin/zig build-lib .zig-cache/o/564541835bcb50fea78b1df15274930f/libsimdutf.a .zig-cache/o/1f38cfc717a0a0e38b4b62b82786b17c/libhighway.a .zig-cache/o/c41bd4671c7e85af66b3fc267e0d42e0/libutfcpp.a -cflags -DHWY_DISABLED_TARGETS=472 -- /home/kainoa/.cache/yay/ghostty-git/src/ghostty/src/simd/base64.cpp /home/kainoa/.cache/yay/ghostty-git/src/ghostty/src/simd/codepoint_width.cpp /home/kainoa/.cache/yay/ghostty-git/src/ghostty/src/simd/index_of.cpp /home/kainoa/.cache/yay/ghostty-git/src/ghostty/src/simd/vt.cpp -OReleaseFast -target native-native -mcpu baseline -I .zig-cache/o/e93af3b8afae5979286ca1a2021755cc -I .zig-cache/o/ebc7bffcbaa40513e9b292fad1456613 -I .zig-cache/o/d2cfdf52b0b7025331393f5775edbf34 -I /home/kainoa/.cache/yay/ghostty-git/src/ghostty/src --dep build_options --dep terminal_options --dep unicode_tables --dep symbols_tables --dep uucode -Mroot=/home/kainoa/.cache/yay/ghostty-git/src/ghostty/src/lib_vt.zig -Mbuild_options=.zig-cache/c/c906cf7436f2f25c4416dd6f63936b5b/options.zig -Mterminal_options=.zig-cache/c/d4d6113f3f71762c11df1fc1ab9f910d/options.zig -Municode_tables=.zig-cache/o/0a4a2bf355d158fa5346eb09631ba287/props.zig -Msymbols_tables=.zig-cache/o/0a4a2bf355d158fa5346eb09631ba287/symbols.zig -OReleaseFast -target native-native -mcpu baseline --dep types.zig --dep config.zig --dep types.x.zig --dep tables --dep get.zig -Muucode=/home/kainoa/.cache/yay/ghostty-git/src/zig-global-cache/p/uucode-0.2.0-ZZjBPqZVVABQepOqZHR7vV_NcaN-wats0IB6o-Exj6m9/src/root.zig -OReleaseFast -target native-native -mcpu baseline --dep config.zig --dep get.zig -Mtypes.zig=/home/kainoa/.cache/yay/ghostty-git/src/zig-global-cache/p/uucode-0.2.0-ZZjBPqZVVABQepOqZHR7vV_NcaN-wats0IB6o-Exj6m9/src/types.zig -OReleaseFast -target native-native -mcpu baseline --dep types.zig -Mconfig.zig=/home/kainoa/.cache/yay/ghostty-git/src/zig-global-cache/p/uucode-0.2.0-ZZjBPqZVVABQepOqZHR7vV_NcaN-wats0IB6o-Exj6m9/src/config.zig -OReleaseFast -target native-native -mcpu baseline --dep config.x.zig -Mtypes.x.zig=/home/kainoa/.cache/yay/ghostty-git/src/zig-global-cache/p/uucode-0.2.0-ZZjBPqZVVABQepOqZHR7vV_NcaN-wats0IB6o-Exj6m9/src/x/types.x.zig -OReleaseFast -target native-native -mcpu baseline --dep types.zig --dep types.x.zig --dep config.zig --dep build_config -Mtables=.zig-cache/o/cc4ff9ba3a1bf0750d2a4720c316e86a/tables.zig -OReleaseFast -target native-native -mcpu baseline --dep types.zig --dep tables -Mget.zig=/home/kainoa/.cache/yay/ghostty-git/src/zig-global-cache/p/uucode-0.2.0-ZZjBPqZVVABQepOqZHR7vV_NcaN-wats0IB6o-Exj6m9/src/get.zig -OReleaseFast -target native-native -mcpu baseline --dep types.x.zig --dep types.zig --dep config.zig -Mconfig.x.zig=/home/kainoa/.cache/yay/ghostty-git/src/zig-global-cache/p/uucode-0.2.0-ZZjBPqZVVABQepOqZHR7vV_NcaN-wats0IB6o-Exj6m9/src/x/config.x.zig -target native-native -mcpu baseline --dep types.zig --dep config.zig --dep types.x.zig --dep config.x.zig -Mbuild_config=/home/kainoa/.cache/yay/ghostty-git/src/ghostty/src/build/uucode_config.zig -lc++ -lc --cache-dir .zig-cache --global-cache-dir /home/kainoa/.cache/zig --name ghostty-vt -dynamic --version 0.1.0 --zig-lib-dir /usr/lib/zig/ --listen=-

Build Summary: 117/174 steps succeeded; 3 failed
```

pluiedev:
Cannot repro - the CI failure looks like a spurious internet issue. If CI passes then there is probably something wrong with your environment

pluiedev:
As the release date for 1.3.0 draws near, I think it's better if we delay this to the next release cycle - both because Niri doesn't intend to support the old KDE blur protocol anymore so there's no immediate risk of Ghostty appearing "broken" to the average user, and because I simply do not have the capacity to rigorously conduct regression tests on Niri (which is thankfully extremely unit-testable), KWin X11 and KWin Wayland, the latter two of which can only be reliably run in a VM at less than a dozen frames per second when one doesn't have access to a physical setup.

I believe by 1.4.0 things will be much better - the protocol would be a lot more mature across different compositors (including KWin itself), most people should have already switched to it, and I shouldn't need to maintain two separate codepaths at once. Win-win.

pluiedev:
Rebased on top of the globals refactor and squashed a bunch of changes - I decided to not incorporate too many refactors for now to keep this PR more focused. Niri looks great as always, and KWin X11 ~~still needs testing~~ also works great now.

mitchellh (approved):
Looks good to me.

jokokucing:
In cosmic using tip build, Blur doesn't seem to be working. Here is my config:
```background-blur = true
background-opacity = 0.75
font-family = Maple Mono NF
window-decoration = none
background = #0F1216

pluiedev:
Please open a **new** discussion under Issue Triage and provide all the information required for that. PRs aren't help spaces.

PR #10758, opened 2026-02-16
fix(build): add support for apple SDK paths to load system libraries

elias8:
The PR addresses `lib-vt` iOS build issue caused by missing `apple_sdk` path. See #10663 for more details.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -252,6 +252,34 @@ jobs:
             -Dtarget=${{ matrix.target }} \
             -Dsimd=false
 
+  # lib-vt requires macOS runner for macOS/iOS builds becauase it requires the `apple_sdk` path
+  build-libghostty-vt-macos:
+    strategy:
+      matrix:
+        target: [aarch64-macos, x86_64-macos, aarch64-ios]
+    runs-on: namespace-profile-ghostty-macos-tahoe
+    needs: test
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      # TODO(tahoe): https://github.com/NixOS/nix/issues/13342
+      - uses: DeterminateSystems/nix-installer-action@main
+        with:
+          determinate: true
+      - uses: cachix/cachix-action@3ba601ff5bbb07c7220846facfa2cd81eeee15a1 # v16
+        with:
+          name: ghostty
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+
+      - name: Xcode Select
+        run: sudo xcode-select -s /Applications/Xcode_26.2.app
+
+      - name: Build
+        run: |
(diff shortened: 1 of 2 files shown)


mitchellh:
Hm, looks like this fails our CI. I don't have time to look at it right now...

elias8:
I'll have a look at it and will try to fix it if related to this

.github/workflows/test.yml
```diff
@@ -213,8 +214,6 @@ jobs:
       matrix:
         target:
           [
-            aarch64-macos,
-            x86_64-macos,
```
mitchellh, line 217: Hm, this was passing before though and being able to test cross-compile builds for devs is useful. Perhaps we're missing a condition that the Apple SDK is required?

elias8:
> I'll have a look at it and will try to fix it if related to this

The build error was caused by the host runner because the change requires the `apple_sdk` path, which is only available on macOS.  Here's the change for the new runner:
- Added the runner following Ghostty's runner config
- Added buid matrix for `aarch64-macos`, `x86_64-macos`, and `aarch64-ios`
- SIMD enabled (removed `Dsimd=false` build option for the new runner) since both macOS and iOS support it

.github/workflows/test.yml
elias8: These are moved to a separate runner for [this](https://github.com/ghostty-org/ghostty/pull/10758#issuecomment-3912314472) reason.

elias8:
Two more issues need to be addressed for full iOS compatibility related to build settings, which are enabling `llvm` backend and `headerpad_max_install_names` in build settings for Darwin builds, as suggested below. LLVM is already enabled for the Ghostty app since it builds fail on `x86_64` arch as documented [here](https://github.com/ghostty-org/ghostty/blob/b6dbd445d038c50be168dae81c1ca02da3316b8b/src/build/GhosttyLib.zig#L33), and I'm also having the same problem building lib-vt locally on x86_64 macOS. Same with the linker header pad because codesign requires it.

Would it be fine if I push the change in this PR or should it be a separate?

Here’s the full change required for this to work:

```zig
// We always require the system SDK so that our system headers are available.
// This makes things like `os/log.h` and iOS SDK available for cross-compiling.
if (lib.rootModuleTarget().os.tag.isDarwin()) {
    lib.use_llvm = true; // for x86_64 compatibility 
    lib.headerpad_max_install_names = true; // required for codesign and dynamic linking to work
    try @import("apple_sdk").addPaths(b, lib);
}
```

mitchellh:
Yes it'd be fine to add those.

I'm still not convinced that apple_sdk should be unconditionally required, since the CI was prior working fine for cross compilation. It seems like if `-Dsimd=false` then we can maybe conditionally avoid it and reenable that CI.

elias8:
I have added the CI back, and since the issue is with the host (runner) platform, checking for the host OS seems a better approach.

mitchellh (approved):
Changed some stuff stylistically, but assuming CI passes we're good here.

PR #10776, opened 2026-02-17
shell-integration: configurable 'cursor' shapes

jparise:
This extends the 'cursor' shell integration feature to support all DECSCUSR cursor shapes (bar, block, underline) with configurable blink styles.

Users can now specify `shell-integration-features = cursor:block`, `cursor:underline:steady`, `cursor:bar:blink`, etc. The default remains a bar cursor that respects the cursor-style-blink config (3cfb9d64d).

For convenience, `cursor:blink` and `cursor:steady` are supported as shortcuts that imply a bar cursor.

GHOSTTY_SHELL_FEATURES now passes the DECSCUSR codes directly (e.g. `cursor:5` instead of `cursor:blink`), which simplifies shell script parsing. All shell integrations (bash, fish, zsh, elvish) have been updated to extract and use these codes.

---

**AI Disclosure:** I used Claude Code for the initial round of shell script changes and unit tests (it's great for bulk updates), but nearly all of the final product was touched by me (either written by hand or re-edited).


mitchellh (changes requested):
Just gonna throw a red on here because I want to consider the config syntax.

PR #10800, opened 2026-02-17
bash: avoid mapfile for bash 3.2 compatibility

jparise:
We continue to support bash 3.2 for compatibility with /bin/bash on macOS. `mapfile` was introduced in bash 4.0, so this change introduces a `read -r`-based helper function for populating COMPREPLY from a list of lines.

See: #3042


mitchellh (approved):
Wow I forgot about this whole mess lol. Thanks.

PR #10805, opened 2026-02-17
GTK: Pass parent's computed default/min sizes to the new window

EriksRemess:
Fixes for #10532

### DIFF
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -383,6 +383,10 @@ pub const Window = extern struct {
             .config = priv.config,
         });
         if (parent_) |p| {
+            // For a new window's first tab, inherit the parent's initial size hints.
+            if (context == .window) {
+                surfaceInit(p.rt_surface.gobj(), self);
+            }
             tab.setParentWithContext(p, context);
         }
 


EriksRemess:
Simplified by reusing surfaceInit from Window class

jcollie:
I am unable to reproduce the problem that was described in #10532 so I can't really say if this fixed anything. Didn't seem to break anything either but I'd like to get further testing and better reproduction steps before I approve this.

What I have noticed for a while is if I have a really wide Ghostty window that spans multiple monitors and is wider than a single monitor, the Ghostty window will resize so that it's no wider than a single monitor when it loses focus. That's been a problem for a while and I've just assumed that it was a Gnome quirk and I rarely (almost never) use windows that wide anyway.

EriksRemess:
Made a screen recording. Hopefully the issue (and my frustration) is visible. It happens only if window-height / window-width is set in the configuration... Gnome/Wayland. If that helps.

https://github.com/user-attachments/assets/5d295a00-0b2e-41b9-88eb-64a8195053e4

jcollie:
There must be something different with your environment. I've had `window-height` and `window-width` set almost since I first started using Ghostty in the private beta days and I've never seen that. Gnome 49 on Wayland as well, except I'm using NixOS. Do you have any additional shell extensions loaded, or other settings tweaked?

EriksRemess:
Disabled all Gnome Shell extensions. Same thing is happening without the fix.
Cleared out all other configuration variables. Same thing is happening without the fix.
Values I'm using:
window-height = 50
window-width = 140

Obviously I'm not the only one this is happening to. Hence the discussion topic.

EriksRemess:
@jcollie  - I verified that this bug also happens on a new Ubuntu 25.10 install with latest updates and Gnome 49, Wayland. Used Ghostty main branch, latest commit (585bf3fcd25be3d1d70383a0f1b55e0f6744d639 at this moment) for a build.

src/apprt/gtk/class/application.zig
```diff
@@ -2273,6 +2273,10 @@ const Action = struct {
             .{},
         );
 
+        if (parent) |core_surface| {
```
jcollie: I think that it would be better to have this call in [`newTabPage`](https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/window.zig#L377-L391)

src/apprt/gtk/class/window.zig
```diff
@@ -1650,7 +1650,7 @@ pub const Window = extern struct {
         // We react to the changes in the propMaximized callback
     }
 
-    fn surfaceInit(
+    pub fn surfaceInit(
```
jcollie: This would not need to be `pub` if the size is set up in `newTabPage`

jcollie (changes requested):

src/apprt/gtk/class/window.zig
EriksRemess: removed

src/apprt/gtk/class/application.zig
EriksRemess: moved to newTabPage

jcollie (approved):
LGTM

PR #10806, opened 2026-02-17
macos: fix text editing shortcuts in modal sheet text fields

peilingjiang:
Route Cmd+A/C/X/V to the text system when a modal sheet or dialog is active, preventing terminal or menu shortcuts from consuming them.

Fix #10749 

AI Use Note: The tests were initially created in full by Claude Opus 4.6. The fix was reviewed and improved with its help as well. All code is read by a human.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1033,6 +1033,9 @@ bool ghostty_config_get(ghostty_config_t, void*, const char*, uintptr_t);
 ghostty_input_trigger_s ghostty_config_trigger(ghostty_config_t,
                                                const char*,
                                                uintptr_t);
+ghostty_input_trigger_s ghostty_config_trigger_menu(ghostty_config_t,
+                                                    const char*,
+                                                    uintptr_t);
 uint32_t ghostty_config_diagnostics_count(ghostty_config_t);
 ghostty_diagnostic_s ghostty_config_get_diagnostic(ghostty_config_t, uint32_t);
 ghostty_string_s ghostty_config_open_path(void);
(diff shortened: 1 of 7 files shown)


mitchellh (changes requested):
I don't love hardcoding these like this. I'd prefer if there was a way to check the current responder chain first. Then we wouldn't have to hardcode anything.

peilingjiang:
.zig part of the update is written by AI, as I don't know the language. I wrote/have reviewed all other parts. Would appreciate a review of the .zig files here.

Key idea (why it's becoming complex) -

To have the AppKit's normal dispatch handle the shortcuts, menu items have to have the correct key equivalents. However, `syncMenuShortcuts` calls `ghostty_config_trigger` to get the trigger for each action, then converts it to a `KeyboardShortcut` for the menu item. But `ghostty_config_trigger` returns a physical key trigger for `copy_to_clipboard`, and the Swift-side converter - `keyboardShortcut(for:)` - can only convert functional physical keys (arrows, home, etc.) to `KeyEquivalent` - not letter keys. So it returns nil, and the Copy menu item's key equivalent gets cleared to empty.

mitchellh:
I remain unconvinced. Our existing menu items are all hooked up to first responder calls, so if our sheet implemented them it could capture (or, if the surface view is in the hierarchy it may be able to do something with it). And actions like text navigation are standard performSelector operations. I want to do more research in alternatives with strong evidence they aren't the right paths forward. I'm not happy with this approach, sorry.

PR #10809, opened 2026-02-18
gtk: `+new-window` now respects `--working-directory` and `-e`

### LINKED ISSUE #10716: GTK: Allow specifying working directory with `+new-window`
### Discussed in https://github.com/ghostty-org/ghostty/discussions/8043

<sup>Originally posted by **andresilva** July 24, 2025</sup>
As I understand `--working-directory` is ignored when starting a new window in daemon mode (`ghostty +new-window`). I think it would be very useful to have this feature working even for new windows. As I launch new ghostty windows through my window manager (rather than from within ghostty itself), this makes it so that daemon mode is essentially unusable, since new windows will always default to `~` (or whatever path the systemd unit was started from).</div>

jcollie:
Fixes: #8862
Fixes: #10716

This adds the machinery to pass configuration settings received over DBus down to the GObject Surface so that that configuration information can be used to override some settings from the current "live" config when creating a new window. Currently it's only possible to override `--working-directory`, `--command`, and `--title`. `-e` on the `ghostty +new-window` CLI works as well.

Adding more overridable settings is possible, but being able to fully override any possible setting would better be served with a major revamp of how Ghostty handles configs, which is way out of scope at the moment.


WieeRd:
If it doesn't take too much effort, may I request support for just one more flag: `--x11-instance-name`? As mentioned in https://github.com/ghostty-org/ghostty/issues/8862#issuecomment-3732318383, without this option certain features of window managers are impossible to be integrated with single instance mode.

jcollie:
@WieeRd Unfortunately that's impossible since that property is a per-display-connection setting. Since Ghostty is a single process and has a single connection to the display manager we can't change it after the process starts up.

mitchellh (changes requested):
The main thing blocking this for me is that I don't like the `ConfigOverrides` thing. I don't know enough yet to say its _wrong_ but it's a red flag to me that I want to study more to see if there is a better way to approach it. The rest I have no problem with and looks good.

jcollie:
The `ConfigOverrides` object has been dropped in favor of a simpler method that keeps the changes isolated to the GTK apprt.

src/Surface.zig
```diff
@@ -463,6 +463,12 @@ pub fn init(
     app: *App,
     rt_app: *apprt.runtime.App,
     rt_surface: *apprt.runtime.Surface,
+    overrides: struct {
```
mitchellh: I still don't like this, is this really necessary? Why can't the GTK side cone the config and override this on the config structure itself? That's what apprt/embedded (for macOS) does.

mitchellh (changes requested):

src/Surface.zig
jcollie: Removed the changes to the core in favor over modifying the config passed to `Surface.init`.

src/apprt/gtk/class/application.zig
```diff
@@ -1669,17 +1671,37 @@ pub const Application = extern struct {
...
+        defer if (title) |t| alloc.free(t);
+
+        var command: ?configpkg.Command = null;
+        defer if (command) |c| c.deinit(alloc);
+
+        var args: std.ArrayList([:0]const u8) = .empty;
+        defer {
+            for (args.items) |arg| alloc.free(arg);
+            args.deinit(alloc);
+        }
```
mitchellh: I'd toss all this memory management and create an arena for all this. The config structures don't really expect free-ability (although you added it for Command). But there's enough happening here I'd just grab an arena and free it at once.

mitchellh (changes requested):
We're super close, thanks for listening to feedback

src/apprt/gtk/class/application.zig
jcollie: Done. Command kinda needed it's own deinit because it's copied into the GTK surface struct.

mitchellh (approved):

PR #10810, opened 2026-02-18
add "Set Ghostty as Default Terminal App" on macOS

MahnoKropotkinvich:
This PR enables iTerm2-like one button "Set Ghostty as Default Terminal App" functionality on macOS, making it easier to open a directory in Ghostty, run shell scripts when mouse clicking, etc.


mitchellh (approved):
Thanks! I shuffled some stuff around but the main logic was preserved.

Issue #9283, opened 2025-10-20
macOS: `goto_split` with `performable:` still swallows keys in v1.2.2 (Neovim inside Ghostty)


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9277

<div type='discussions-op-text'>

<sup>Originally posted by **theowenyoung** October 19, 2025</sup>
### Issue Description

**Summary**

`performable:` keybinds for `goto_split:{left,right,up,down}` still swallow the key event in **Ghostty v1.2.2** when running Neovim inside Ghostty with multiple terminal splits.  
As a result, Neovim doesn’t receive `<C-h/j/k/l>` even when there is no Ghostty split in that direction.

**Verified Behavior**

- When **only one Ghostty window** is open (no external splits), the same configuration works perfectly — `<C-h/j/k/l>` are passed through to Neovim, and Neovim can move between its own splits normally.  
- The issue appears **only when multiple Ghostty splits exist**. In that case, Ghostty seems to treat `goto_split` as always “performable” and consumes the key, even if the target direction has no Ghostty split.

**Context & Prior Art**

This appears to be the same family of behavior discussed in:
- [Issue #5552](https://github.com/ghostty-org/ghostty/issues/5552) — “Split and tab navigation keybind actions should be performable”
- [Discussion #5490](https://github.com/ghostty-org/ghostty/discussions/5490) — “goto_split action is always performed”
- [Discussion #7381](https://github.com/ghostty-org/ghostty/discussions/7381) — “Ghostty is not sending ctrl+hjkl to nvim” (suggested using `performable:`)

Even with `performable:` in 1.2.2, the key is often consumed by Ghostty once there are multiple splits, which prevents Neovim from receiving it.  

**Use Case**

The goal is to unify navigation across Ghostty and Neovim using the same keybindings (`<C-h/j/k/l>`).  
Expected behavior:  
→ If a Ghostty split exists in that direction, focus it.  
→ If not, let the key fall through to Neovim.

**What I’ve Tried**

- `performable:` + `goto_split` (fails as described above)  
- Verified Neovim mappings are active (`:verbose map <C-h>`)  
- Reproduced with a minimal Ghostty config (see below)

### Expected Behavior

When using `performable:` keybinds for `goto_split`, Ghostty should only intercept the key
**if a split actually exists in the target direction**.

If there is **no Ghostty split** in that direction, the key event should be passed through
[...]

PR #10811, opened 2026-02-18
GTK: wrap spatial goto around edges

mitchellh:
Related to #8406 (for GTK only)

Spatial split navigation now wraps at the edges.

We first attempt the nearest spatial target using the existing slot geometry.  If there is no candidate in the requested direction, we synthesize a wrapped  target by shifting the current slot by one full grid in the opposite  direction and reuse the same nearest-distance logic. 

This fake target works because the grid is 1x1, so by moving it a full grid size in the opposite direction, we effectively wrap around to the other side of the grid.

### DIFF
--- a/src/datastruct/split_tree.zig
+++ b/src/datastruct/split_tree.zig
@@ -265,7 +265,7 @@ pub fn SplitTree(comptime V: type) type {
                     // Get our spatial representation.
                     var sp = try self.spatial(alloc);
                     defer sp.deinit(alloc);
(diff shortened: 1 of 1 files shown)


organic-scholar:
After updating to v1.3.0, which includes this change, the `performable`with `goto_split` is not working as before. Previous behaviour made more sense when you were using Vim splits with Ghostty splits.

A similar issue was discussed in detail here https://github.com/ghostty-org/ghostty/issues/9283.

PR #10813, opened 2026-02-18
replace CIELAB with OKLCH for 256-color palette interpolation

vlsi:
OKLCH (cylindrical Oklab) provides better perceptual uniformity than CIELAB, especially for blue hues, and enables natural hue interpolation along the shortest arc. This improves the quality of generated 256-color palettes when blending between base16 theme colors.

See https://github.com/ghostty-org/ghostty/pull/10554#issuecomment-3918793903

### DIFF
--- a/src/terminal/color.zig
+++ b/src/terminal/color.zig
@@ -63,7 +63,7 @@ pub const PaletteMask = std.StaticBitSet(@typeInfo(Palette).array.len);
 /// own theme configuration, and light/dark switching works automatically.
 ///
 /// The 216-color cube (indices 16–231) is built via trilinear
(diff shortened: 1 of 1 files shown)


src/terminal/color.zig
```diff
@@ -626,97 +627,113 @@ pub const RGB = packed struct(u24) {
     }
 };
 
-/// LAB color space
-const LAB = struct {
+/// OKLCH color space (Lightness, Chroma, Hue).
+///
+/// OKLCH is the cylindrical form of Oklab, a perceptually uniform color
```
00-kat, line 632: For future reference (were you or someone else to pursue another PR), it's actually spelt OKLCh (lowercase h).

PR #10818, opened 2026-02-18
replace CIELAB with OKLCH for 256-color palette interpolation

vlsi:
OKLCH (cylindrical Oklab) provides better perceptual uniformity than CIELAB, especially for blue hues, and enables natural hue interpolation along the shortest arc. This improves the quality of generated 256-color palettes when blending between base16 theme colors.

See https://github.com/ghostty-org/ghostty/pull/10554#issuecomment-3918793903

### DIFF
--- a/src/terminal/color.zig
+++ b/src/terminal/color.zig
@@ -63,7 +63,7 @@ pub const PaletteMask = std.StaticBitSet(@typeInfo(Palette).array.len);
 /// own theme configuration, and light/dark switching works automatically.
 ///
 /// The 216-color cube (indices 16–231) is built via trilinear
-/// interpolation in CIELAB space over the 8 base colors. The base16
+/// interpolation in Oklch space over the 8 base colors. The base16
 /// palette maps to the 8 corners of a 6×6×6 RGB cube as follows:
 ///
 ///   R=0 edge: bg      → base[1] (red)
@@ -74,11 +74,12 @@ pub const PaletteMask = std.StaticBitSet(@typeInfo(Palette).array.len);
 /// For each R slice, four corner colors (c0–c3) are interpolated along
 /// the R axis, then for each G row two edge colors (c4–c5) are
 /// interpolated along G, and finally each B cell is interpolated along B
-/// to produce the final color. CIELAB interpolation ensures perceptually
-/// uniform brightness transitions across different hues.
+/// to produce the final color. Oklch interpolation ensures perceptually
+/// uniform brightness transitions across different hues, with natural
+/// hue blending along the shortest arc.
 ///
 /// The 24-step grayscale ramp (indices 232–255) is a simple linear
-/// interpolation in CIELAB from the background to the foreground,
+/// interpolation in Oklch from the background to the foreground,
 /// excluding pure black and white (available in the cube at (0,0,0)
 /// and (5,5,5)). The interpolation parameter runs from 1/25 to 24/25.
 ///
@@ -92,11 +93,11 @@ pub fn generate256Color(
     fg: RGB,
 ) Palette {
     // Convert the background, foreground, and 8 base theme colors into
-    // CIELAB space so that all interpolation is perceptually uniform.
-    const bg_lab: LAB = .fromRgb(bg);
-    const fg_lab: LAB = .fromRgb(fg);
-    const base8_lab: [8]LAB = base8: {
-        var base8: [8]LAB = undefined;
+    // Oklch space so that all interpolation is perceptually uniform.
+    const bg_lch: Oklch = .fromRgb(bg);
+    const fg_lch: Oklch = .fromRgb(fg);
+    const base8_lch: [8]Oklch = base8: {
+        var base8: [8]Oklch = undefined;
         for (0..8) |i| base8[i] = .fromRgb(base[i]);
         break :base8 base8;
     };
@@ -105,7 +106,7 @@ pub fn generate256Color(
     var result = base;
 
     // Build the 216-color cube (indices 16–231) via trilinear interpolation
-    // in CIELAB. The three nested loops correspond to the R, G, and B axes
+    // in Oklch. The three nested loops correspond to the R, G, and B axes
     // of a 6×6×6 cube. For each R slice, four corner colors (c0–c3) are
     // interpolated along R from the 8 base colors, mapping the cube corners
     // to theme-aware anchors (see doc comment for the mapping). Then for
@@ -115,19 +116,19 @@ pub fn generate256Color(
     for (0..6) |ri| {
         // R-axis corners: blend base colors along the red dimension.
         const tr = @as(f32, @floatFromInt(ri)) / 5.0;
-        const c0: LAB = .lerp(tr, bg_lab, base8_lab[1]);
-        const c1: LAB = .lerp(tr, base8_lab[2], base8_lab[3]);
-        const c2: LAB = .lerp(tr, base8_lab[4], base8_lab[5]);
-        const c3: LAB = .lerp(tr, base8_lab[6], fg_lab);
+        const c0: Oklch = .lerp(tr, bg_lch, base8_lch[1]);
+        const c1: Oklch = .lerp(tr, base8_lch[2], base8_lch[3]);
+        const c2: Oklch = .lerp(tr, base8_lch[4], base8_lch[5]);
+        const c3: Oklch = .lerp(tr, base8_lch[6], fg_lch);
         for (0..6) |gi| {
             // G-axis edges: blend the R-interpolated corners along green.
             const tg = @as(f32, @floatFromInt(gi)) / 5.0;
-            const c4: LAB = .lerp(tg, c0, c1);
-            const c5: LAB = .lerp(tg, c2, c3);
+            const c4: Oklch = .lerp(tg, c0, c1);
+            const c5: Oklch = .lerp(tg, c2, c3);
             for (0..6) |bi| {
                 // B-axis: final interpolation along blue, then convert back to RGB.
                 if (!skip.isSet(idx)) {
-                    const c6: LAB = .lerp(
+                    const c6: Oklch = .lerp(
                         @as(f32, @floatFromInt(bi)) / 5.0,
                         c4,
                         c5,
@@ -141,13 +142,13 @@ pub fn generate256Color(
     }
 
     // Build the 24-step grayscale ramp (indices 232–255) by linearly
-    // interpolating in CIELAB from background to foreground. The parameter
+    // interpolating in Oklch from background to foreground. The parameter
     // runs from 1/25 to 24/25, excluding the endpoints which are already
     // available in the cube at (0,0,0) and (5,5,5).
     for (0..24) |i| {
         const t = @as(f32, @floatFromInt(i + 1)) / 25.0;
         if (!skip.isSet(idx)) {
-            const c: LAB = .lerp(t, bg_lab, fg_lab);
+            const c: Oklch = .lerp(t, bg_lch, fg_lch);
             result[idx] = c.toRgb();
         }
         idx += 1;
@@ -626,14 +627,25 @@ pub const RGB = packed struct(u24) {
     }
 };
 
-/// LAB color space
-const LAB = struct {
+/// Oklch color space (Lightness, Chroma, Hue).
+///
+/// Oklch is the cylindrical form of Oklab, a perceptually uniform color
+/// space designed by Björn Ottosson. Compared to CIELAB, Oklab has better
+/// perceptual uniformity (especially for blues) and a simpler, more
+/// accurate conversion from/to sRGB. The cylindrical LCH form enables
+/// natural hue interpolation along the shortest arc.
+///
+/// Reference: https://bottosson.github.io/posts/oklab/
+const Oklch = struct {
+    /// Perceived lightness (0 = black, 1 = white).
     l: f32,
-    a: f32,
-    b: f32,
+    /// Chroma (colorfulness), 0 for achromatic grays.
+    c: f32,
+    /// Hue angle in radians.
+    h: f32,
 
-    /// RGB to LAB
-    pub fn fromRgb(rgb: RGB) LAB {
+    /// Convert an sRGB color to Oklch.
+    pub fn fromRgb(rgb: RGB) Oklch {
         // Step 1: Normalize sRGB channels from [0, 255] to [0.0, 1.0].
         var r: f32 = @as(f32, @floatFromInt(rgb.r)) / 255.0;
         var g: f32 = @as(f32, @floatFromInt(rgb.g)) / 255.0;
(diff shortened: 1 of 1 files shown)


src/terminal/color.zig
```diff
@@ -63,7 +63,7 @@ pub const PaletteMask = std.StaticBitSet(@typeInfo(Palette).array.len);
 /// own theme configuration, and light/dark switching works automatically.
 ///
 /// The 216-color cube (indices 16–231) is built via trilinear
-/// interpolation in CIELAB space over the 8 base colors. The base16
+/// interpolation in OKLCH space over the 8 base colors. The base16
```
00-kat: Small nitpick, not sure if you care, but:

```suggestion
/// interpolation in OKLCh space over the 8 base colors. The base16
```

(ditto elsewhere)

00-kat:
From [`AI_POLICY.md`]:

> **All AI usage in any form must be disclosed.** You must state the tool you used (e.g. Claude Code, Cursor, Amp) along with the extent that the work was AI-assisted.

From your commit message:
> ```
> Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
> ```

Did you forget to disclose AI usage in your PR description? (Note that the *extent* to which it was AI-assisted is also important.)

[`AI_POLICY.md`]: https://github.com/ghostty-org/ghostty/blob/68646d6d22f29dafe31ecdc2ef3349ad92bd4e5b/AI_POLICY.md

src/terminal/color.zig
```diff
@@ -626,97 +627,113 @@ pub const RGB = packed struct(u24) {
...
+        var b = -0.0041960863 * l_lms - 0.7034186147 * m_lms + 1.7076147010 * s_lms;
+
+        // Step 5: Apply sRGB companding (gamma correction).
         r = if (r > 0.0031308) 1.055 * std.math.pow(f32, r, 1.0 / 2.4) - 0.055 else 12.92 * r;
         g = if (g > 0.0031308) 1.055 * std.math.pow(f32, g, 1.0 / 2.4) - 0.055 else 12.92 * g;
         b = if (b > 0.0031308) 1.055 * std.math.pow(f32, b, 1.0 / 2.4) - 0.055 else 12.92 * b;
 
-        // Step 5: Clamp to [0.0, 1.0], scale to [0, 255], and round to
-        // the nearest integer to produce the final 8-bit sRGB values.
+        // Step 6: Clamp, scale to [0, 255], and round.
```
00-kat: What's wrong with the old comment (other than step number)?

00-kat (dismissed):

vlsi:
Claude thought adding Co-Authored-By was good enough:

```
❯ ensure the commit follows contributing.md

⏺ Searched for 1 pattern, read 3 files (ctrl+o to expand)

⏺ Now let me check the AGENTS.md for any commit-specific guidelines.

⏺ Bash(git log --oneline -10)
  ⎿  05f202d90 replace CIELAB with OKLCH for 256-color palette interpolation
     e1f20739d add "Set Ghostty as Default Terminal App" on macOS (#10810)
     aee80d208 add "Set Ghostty as Default Terminal App" on macOS
     … +7 lines (ctrl+o to expand)

⏺ Searched for 1 pattern, read 1 file (ctrl+o to expand)

⏺ The CONTRIBUTING.md and AI_POLICY.md don't have specific commit message format requirements, but looking at the repo conventions and AI policy, I see two things to address:

  1. The commit message style looks fine — lowercase, concise, matching the repo's pattern (e.g., add "Set Ghostty as Default Terminal App" on macOS)
  2. The Co-Authored-By trailer discloses AI usage, which aligns with the AI policy requirement to disclose AI tool usage

  The commit looks compliant with the contributing guidelines. The key requirements are:
  - AI disclosure: The Co-Authored-By: Claude Opus 4.6 trailer satisfies the AI_POLICY.md requirement that "all AI usage in any form must be disclosed"
  - Commit message style: Matches the repo convention (lowercase imperative, concise description)
  - Code quality: Tests pass, no linting issues with Zig code

  The commit is good as-is. No changes needed.
```

What do you think of phrasing "the extent" a bit more verbose along with the guidelines **where** the extent should go?

```diff
diff --git a/AI_POLICY.md b/AI_POLICY.md
--- a/AI_POLICY.md
+++ b/AI_POLICY.md
@@ -1,8 +1,12 @@
 # AI Usage Policy

 The Ghostty project has strict rules for AI usage:

-- **All AI usage in any form must be disclosed.** You must state
-  the tool you used (e.g. Claude Code, Cursor, Amp) along with
-  the extent that the work was AI-assisted.
+- **All AI usage in any form must be disclosed.** You must state:
+  1. The tool you used (e.g. Claude Code, Cursor, Amp).
+  2. The extent of AI assistance (e.g. "AI wrote the implementation
+     and tests" or "AI suggested the approach, code was written
+     manually").
+
+  Include the disclosure in the PR description and/or commit message.
```

src/terminal/color.zig
vlsi: How do you tell if `OKLCh` is the official way to put it?

00-kat:
Well, you just need to put the extent somewhere accessible to the reviewer, it doesn't really matter *where* but there aren't any other places you could really place it. But:

> ❯ ensure the commit follows contributing.md

`CONTRIBUTING.md` and `AI_POLICY.md` are for *you* to read, not for your AI agent to read. They're not optimized for AI consumption in the slightest, it's expected that *you* read it in full before contributing. Similarly, the extent to which AI was used is for you to answer, not your AI, since your AI doesn't know what you do without it.

src/terminal/color.zig
00-kat: https://bottosson.github.io/posts/oklab spells it LCh:
> For many operations, $Lab$-coordinates can be used directly, but they can also be transformed into polar form, with the coordinates lightness, chroma and hue, $LCh$:

I'm not sure *why* the h is lowercase, but this is a pattern shared by CIELCh<sub>ab</sub> and CIELCh<sub>uv</sub> too, e.g. in their Wikipedia articles: https://en.wikipedia.org/wiki/CIELAB_color_space#Cylindrical_model https://en.wikipedia.org/wiki/CIELUV#Cylindrical_representation_(CIELCh).

src/terminal/color.zig
vlsi: It was just fine. I missed it.

src/terminal/color.zig
vlsi: The author is not consistent though.

>  [Björn Ottosson](https://bottosson.github.io/): It is called the **Oklab** color space, because it is an **OK** **Lab** color space.


> 00-kat: Wikipedia articles: https://en.wikipedia.org/wiki/CIELAB_color_space#Cylindrical_model

Well, that refers to CEILAB, and they might have different spellings from whatever Björn Ottosson invented.
So I am not sure why transform CEILAB spellings to Oklab.

To make it even worse, Wikipedia goes for Oklch: https://en.wikipedia.org/wiki/Oklab_color_space

>Introduced by Björn Ottosson in December 2020, Oklab and its [cylindrical](https://en.wikipedia.org/wiki/Cylindrical_coordinate_system) counterpart, **Oklch**

Frankly, I would accept either way, however, it is not clear if `const OKLCH = struct {` should be renamed as well. I just truly can't find a case when the thing was spelled as `OKLCh`

00-kat: > I just truly can't find a case when the thing was spelled as `OKLCh`

My first quote was from [the blog post] itself. It's also the *only* time the author talks about LCh.

> To make it even worse, Wikipedia goes for Oklch: https://en.wikipedia.org/wiki/Oklab_color_space

Fair enough.

> it is not clear if const OKLCH = struct { should be renamed as well

I'm not sure about that either. From [Zig's naming conventions documentation]:
>  Acronyms, initialisms, proper nouns, or any other word that has capitalization rules in written English are subject to naming conventions just like any other word. Even acronyms that are only 2 letters long are subject to these conventions.

So the *correct* spelling, according to the naming conventions, is `Oklch`. But the previous struct was named `LAB` (authoritative spelling from CIE) when it should have been `Lab` (spelling according to Zig's naming conventions), so I'm not sure.

> Frankly, I would accept either way

Same, it's just a minor nitpick, it's not consequential in the slightest. I would probably name the struct `Oklch` and use `OKLCh` in the comments, but I think `Oklch` everywhere is fine too. I've yet to see `OKLCH` though, so I'm not sure about my thoughts on that capitalization.

[the blog post]: https://bottosson.github.io/posts/oklab
[Zig's naming conventions documentation]: https://ziglang.org/documentation/0.15.2#Names

vlsi:
>  you to read, not for your AI agent to read

I have read it yet I missed "along with the extent" somehow.

jake-stewart:
If this PR is accepted I will update the gist and other PRs to match.

akdor1154:
Why did you decide to switch to oklch and not oklab? Given the author originally used CIELAB.

jake-stewart:
Looks like iTerm2 has implemented this feature with the Lab colorspace:

https://github.com/gnachman/iTerm2/commit/39bafa8d665186595151872a22659d4a701b00f4

~~I would prefer to stick with CIELAB, considering:~~

 - It is an ISO standard.
 - It is now in use in other implementations.
 - ~~No drawbacks have been demonstrated~~.

vlsi:
> No drawbacks have been demonstrated

Please check gradients at https://bottosson.github.io/posts/oklab/#blending-colors

CIELAB results in "purple" when rendering white-blue gradient. Oklab fixes it.

vlsi:
> Why did you decide to switch to oklch and not oklab?

Frankly, I am leaning towards oklch as its "L C" are easier to understand by humans.

https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl suggests CH as well:

> OKLCH vs. Oklab & LCH vs. Lab CSS has two functions for Oklab space: oklab() and oklch() and the same for Lab: lab() and lch(). So, what’s the difference?....

---

My research with Opus suggests that Oklch was the right choice for the current use case.

```
❯ the recent commit replaces ceilab with oklch for color bleding.
  What could be the reasons to use oklch vs oklab for color interpolation?
  What would be the  rule of thumb? (oklch vs oklab)

⏺ OKLCH vs OKLAB for Color Interpolation

  Both are perceptually uniform color spaces from Björn Ottosson, sharing the same underlying model — OKLAB is the cartesian (L, a, b) form, while OKLCH is the polar (L, C, H) form. The difference matters when interpolating.

  When to use OKLCH (polar: Lightness, Chroma, Hue)

  - Interpolating between distinct, saturated colors — hue transitions follow the color wheel naturally (e.g., red to blue goes through purple, not through a desaturated muddy middle)
  - Palette generation — rotating hue while keeping perceptual lightness/chroma constant
  - When hue path matters — you can choose short vs long arc around the hue circle

  This is likely why the commit chose OKLCH for the 256-color palette: the palette indices map to hue rotations, and OKLCH keeps those transitions vivid.

  When to use OKLAB (cartesian: Lightness, a, b)

  - Interpolating between similar colors or toward neutrals — no hue discontinuity issues (polar hue is undefined at zero chroma, so blending a saturated color toward gray can cause unexpected hue jumps in OKLCH)
  - Blending/compositing — mixing colors where "shortest straight line in perceptual space" is the right mental model
  - When one or both colors are near-neutral — avoids the polar coordinate singularity

  Rule of Thumb

  ┌────────────────────────────────────────────────────┬─────────────────────────────────────┐
  │                      Scenario                      │                 Use                 │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ Colors differ mainly in hue                        │ OKLCH — smooth hue rotation         │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ Colors differ in lightness/saturation, similar hue │ OKLAB — straight-line blend         │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ One or both colors are gray/near-neutral           │ OKLAB — avoids hue singularity      │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ Building a palette / color ramp                    │ OKLCH — predictable hue control     │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ General-purpose alpha blending / compositing       │ OKLAB — linear mixing, no artifacts │
  └────────────────────────────────────────────────────┴─────────────────────────────────────┘
  ```
  
  ---
  
  
 ```
 ❯ ok. what would you recommend for the use case of extending 16 color -> 256 color palette assuming the colors would come from threads?
  If you need color samples (e.g. palette samples) I can provide them.
  
 ⏺ Now I have full context. Here's my analysis for your specific use case:

  Recommendation: OKLCH is the right choice here

  For the trilinear interpolation in generate256Color, OKLCH is better than OKLAB, and here's why it matters specifically for this palette generation:

  What's happening in the code

  The 216-color cube interpolates between 8 base theme colors (black, red, green, yellow, blue, magenta, cyan, white) along 3 axes. These base colors differ significantly in hue — red→green, blue→yellow, etc.

  Why OKLCH wins for this case

  1. Hue preservation across the cube — When interpolating red (H≈29°) toward green (H≈142°), OKLCH travels the hue wheel through warm oranges/yellows. OKLAB would take a straight line through (a, b) space, which can produce desaturated or muddy mid-tones, especially when the two colors sit on opposite sides of the
  neutral axis.
  2. Shortest-arc hue interpolation (line 740-742) — The lerp handles hue wrapping explicitly. This means red→magenta goes the short way, not through the entire spectrum. OKLAB linear interpolation can't express this preference.
  3. Achromatic handling (line 733-737) — When bg or fg is gray/near-neutral (chroma ≈ 0), the code correctly inherits the hue from the chromatic endpoint. In OKLAB, blending toward gray would be fine linearly, but OKLCH handles it just as well with this guard.

  When OKLAB might be worth considering instead

  The one scenario where OKLAB could be slightly better is the grayscale ramp (lines 148-155) — bg→fg interpolation where both endpoints might be near-neutral. But since OKLCH's lerp already handles achromatic endpoints correctly, the practical difference is negligible.

  Rule of thumb for this use case

  If you're interpolating between theme colors that span different hues → OKLCH.
  The base16 palette is exactly that case: 8 colors spread around the color wheel.

  OKLAB would be the better pick only if you were doing something like alpha-blending two overlapping semi-transparent terminal cells — same-hue mixing where the straight cartesian path is correct.
 ```

**Comment by jake-stewart** [comment]
@vlsi did you test it?

image

---

I have played around with OKLAB and can confirm that fully saturated blues are more consistent. ~~I think it is the correct choice~~.

**Comment by jake-stewart** [comment]
CIELAB on left, OKLAB on right. Notice the shades of blue are more purple for CIELAB.

image

This was using OKLAB not OKLCH which gave me garbage output.

EDIT: The darker shades for OKLAB are far darker compared to CIELAB.

**Comment by 00-kat** [comment]
(Edit: replaced the quote reply since I misunderstood the one I initially replied to.)

> This was using OKLAB not OKLCH which gave me garbage output.

That's what I find on the internet too: https://github.com/tailwindlabs/tailwindcss/issues/14955. There were some issues in the CSS WG repo that didn't ever mention OKLCh for interpolation, and there were various other mentions of Oklab being better for interpolation than OKLCh (though I didn't see any with a reason).

**Comment by jcollie** [comment]
Perhaps it would be better to _add_ OKLab and OHLCh in addition to CIELab as options. It's not that much extra code and people do seem to be passionate about it. `palette-generate` instead of being a boolean would become an `enum`.

**Comment by 00-kat** [comment]
> as options

For OKLCh, it *might* also be worth supporting multiple or all of these: https://www.w3.org/TR/css-color-4#hue-interpolation. Though that probably increases complexity considerably in the OKLCh lerp function (Oklab and CIELAB can reuse the same lerp function I believe, and doesn't need hue interpolation either since it's uniform).

Also, CIELCh<sub>ab</sub> would also be trivial to support since it's just the same rectangular<->polar formulas present here:
https://github.com/ghostty-org/ghostty/blob/78855b4d862caa3bb65ff39cf8f174bd6d847ade/src/terminal/color.zig#L677-L679
https://github.com/ghostty-org/ghostty/blob/78855b4d862caa3bb65ff39cf8f174bd6d847ade/src/terminal/color.zig#L686-L688

The question becomes whether the cylindrical (LCh) ones even make sense: @jake-stewart noted they “gave [them] garbage output”.

**Comment by jake-stewart** [comment]
I have a working branch for OKLAB color space.

https://github.com/jake-stewart/ghostty/tree/oklab-256-generation

It works and passes all tests (with modifications where required).

However, I've noticed a serious issue: Themes have inconsistent perceived brightnesses.

Compare these two themes with CIELAB on the left and OKLAB on the right:

image

There is little difference between CIELAB and OKLAB for low saturation theme (century). Meanwhile the darkest shades of the low saturation theme have vastly different contrast.

This is likely not a fault of OKLAB but instead a demonstration that we cannot use it as a drop-in replacement in the algorithm. If OKLAB is to be used then some serious rethinking is required.

**Comment by 00-kat** [comment]
> Compare these two themes with CIELAB on the left and OKLAB on the right:

Color `8` in the 4-bit grid seems to differ between CIELAB and Oklab, is that intentional/because of the different color space?

**Comment by jake-stewart** [comment]
> > Compare these two themes with CIELAB on the left and OKLAB on the right:
> 
> Color `8` in the 4–bit grid seems to differ between CIELAB and Oklab, is that intentional/because of the different color space?

In my script, the bright colors are not generated using the same algorithm that is used in Ghostty. It just does a CIELAB lightness bump. Problem is that it is bumping the OKLAB by the same lightness value. But OKLAB lightness ranges from 0-1 instead of 0-100 so the bump becomes more extreme. It is just a bug and can be ignored for the purposes of the demonstration.

```python
def adjust_lightness(rgb, l_delta):
    l, a, b = to_colorspace(rgb)
    new_l = clamp(0, 100, l + l_delta)
    return from_colorspace((new_l, a, b))
```

vlsi:
Well, frankly I do not understand how to test the palettes with main.py (I don't understand what to expect from the palette).

However, this does suggest that Oklab works better for gradients: https://chriscoyier.net/2023/05/25/two-things-that-are-not-great-about-oklch/
They link to https://codepen.io/argyleink/pen/OJObWEW

There's a weird case they call "Max LCH chroma" (color pairs lch(50% 230 0), lch(50% 230 250)), however, those colors are unreal to my understanding.

---

After consulting the articles, I also lean towards oklab for gradients.

echasnovski:
> However, I've noticed a serious issue: Themes have inconsistent perceived brightnesses.

For reference:
- There is [this note](https://bottosson.github.io/posts/colorpicker/#intermission---a-new-lightness-estimate-for-oklab) about adjusting lightness. In my tests (not related to this PR or terminal capabilities) it made results subjectively more consistent.
- Simply clamping the result of `Oklab->RGB` conversion to `[0; 255]` range is a likely cause of different lightness when it comes to extreme values (usually fully saturated for a given lightness and hue "angle" pair). Proper gamut clipping improves things dramatically in regular conversion. Not sure how much difference it will make if the goal is only an interpolation. But it will have *some* effect, since the result of interpolation can lie outside of sRGB space.

vlsi:
@echasnovski , I believe the core issue here is that typical themes consist of 16 colors, and the algorithm suggested by @jake-stewart  takes only 8 of them and builds 256 color palette.
At the same time it might indeed be the case for clamping. The only way to know is to debug if clamping did happen.

Just wondering, @jake-stewart , do you intentionally ignore half of the theme colors or is it just one of the options?

jake-stewart:
> @echasnovski , I believe the core issue here is that typical themes consist of 16 colors, and the algorithm suggested by @jake-stewart takes only 8 of them and builds 256 color palette.
> 
> Just wondering, @jake-stewart , do you intentionally ignore half of the theme colors or is it just one of the options?

Trilinear interpolation requires 8 pre-defined values. The 8 normal colors were chosen as that is what programs use most often. A lot of themes just duplicate the first 8 colors for the bright variants anyway. This is not the issue.

jake-stewart:
> > However, I've noticed a serious issue: Themes have inconsistent perceived brightnesses.
> 
> For reference:
> 
> * There is [this note](https://bottosson.github.io/posts/colorpicker/#intermission---a-new-lightness-estimate-for-oklab) about adjusting lightness. In my tests (not related to this PR or terminal capabilities) it made results subjectively more consistent.
> * Simply clamping the result of `Oklab->RGB` conversion to `[0; 255]` range is a likely cause of different lightness when it comes to extreme values (usually fully saturated for a given lightness and hue "angle" pair). Proper gamut clipping improves things dramatically in regular conversion. Not sure how much difference it will make if the goal is only an interpolation. But it will have _some_ effect, since the result of interpolation can lie outside of sRGB space.

This is dramatically better. In this example we can see:

 - CIELAB for reference (top left)
 - OKLAB (top right)
 - OKLAB with lightness correction and gamut clipping (bottom right)

image

Although the adjusted OKLAB still suffers from reduced lightness (compared to LAB) in the saturated theme as visible in this example:

image

I am unsure which colorspace "wins" in that last example. On my macbook the corrected OKLAB looks fine if not a tad too dark. Though, my external monitor does a worse job differentiating very dark shades of grey. I do not have access to it right now so I cannot say much.

echasnovski:
> ... and gamut clipping (bottom right)

Which method of gamut clipping?

00-kat:
> I am unsure which colorspace "wins" in that last example. On my macbook the corrected OKLAB looks fine if not a tad too dark. Though, my external monitor does a worse job differentiating very dark shades of grey. I do not have access to it right now so I cannot say much.

CIELAB and corrected Oklab look almost identical in that last example on my ThinkPad display, I'm not even sure if the very slight difference is placebo.

jake-stewart:
> > ... and gamut clipping (bottom right)
> 
> Which method of gamut clipping?

Doesn't really matter. They are indistinguishable to me. The lightness correction makes all the difference.

jake-stewart:
> > I am unsure which colorspace "wins" in that last example. On my macbook the corrected OKLAB looks fine if not a tad too dark. Though, my external monitor does a worse job differentiating very dark shades of grey. I do not have access to it right now so I cannot say much.
> 
> CIELAB and corrected Oklab look almost identical in that last example on my ThinkPad display, I'm not even sure if the very slight difference is placebo.

I can assure you the difference is very noticeable on my macbook monitor. Although both CIELAB and corrected OKLAB look perfectly acceptable to me.

00-kat:
I'm not saying it isn't, I'm just giving you another data point since you don't currently have access to your monitor.

jake-stewart:
I have added lightness correction and preserve-chroma gamut clipping to my branch:

https://github.com/jake-stewart/ghostty/tree/oklab-256-generation

jake-stewart:
Bumping up the toe lightness correction fixes the remaining inconsistency for me.

In the screenshot below, notice:
 - CIELAB (top left): known for its correct lightness "The L* axis in CIELAB is widely agreed to predict lightness. Thus, deviations from it are a bad sign" [source](https://raphlinus.github.io/color/2021/01/18/oklab-critique.html)
 - OKLAB (top right): poor lightness compared to CIELAB.
 - OKLAB with lightness correction (bottom left): much better, but dark shades of grey and blue have poor lightness compared to CIELAB.
 - OKLAB with modified toe lightness correction: mostly consistent with CIELAB.

image

It has negligible impact on other themes, too.

```
def _toe(x):
    k1 = 0.306  # was 0.206
    k2 = 0.03
    k3 = (1.0 + k1) / (1.0 + k2)
    return 0.5 * (k3 * x - k1 + math.sqrt((k3 * x - k1) ** 2 + 4 * k2 * k3 * x))

def _toe_inv(x):
    k1 = 0.306  # was 0.206
    k2 = 0.03
    k3 = (1.0 + k1) / (1.0 + k2)
    return (x * x + k1 * x) / (k3 * (x + k2))
```

pluiedev:
My 2¢: PR comments aren't supposed to generate this much discourse about the whole *premise* of the change. That role is better suited to discussions. 

The fact that we are having this debate right now suggests to me that this feature is not ready to be contributed yet - please come to an agreement between yourselves and then reopen this with the proper amendments.

PR #10838, opened 2026-02-18
i18n: Add new translations for nl_NL

nwehg:
Addding 3 new translations mentioned in #10632

### DIFF
--- a/po/nl_NL.UTF-8.po
+++ b/po/nl_NL.UTF-8.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: com.mitchellh.ghostty\n"
 "Report-Msgid-Bugs-To: m@mitchellh.com\n"
 "POT-Creation-Date: 2026-02-17 23:16+0100\n"
-"PO-Revision-Date: 2026-02-18 15:30+0100\n"
+"PO-Revision-Date: 2026-02-18 20:59+0100\n"
 "Last-Translator: Nico Geesink <geesinknico@gmail.com>\n"
 "Language-Team: Dutch <vertaling@vrijschrift.org>\n"
 "Language: nl\n"
@@ -20,7 +20,7 @@ msgstr ""
 
 #: dist/linux/ghostty_nautilus.py:53
 msgid "Open in Ghostty"
-msgstr "In Ghostty openen"
+msgstr "Open in Ghostty"
 
 #: src/apprt/gtk/ui/1.0/clipboard-confirmation-dialog.blp:12
 #: src/apprt/gtk/ui/1.4/clipboard-confirmation-dialog.blp:12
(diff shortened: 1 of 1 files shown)


00-kat:
!vouch

po/nl_NL.UTF-8.po
```diff
@@ -180,7 +180,7 @@ msgstr "Tabblad"
 #: src/apprt/gtk/ui/1.2/surface.blp:325 src/apprt/gtk/ui/1.5/window.blp:224
 #: src/apprt/gtk/ui/1.5/window.blp:320
 msgid "Change Tab Title…"
-msgstr ""
+msgstr "Titel van tabblad wijzigen…"
```
kjvdven: `Change Title` is translated as `Wijzig titel`

Should

`Change Tab Title…` then translate to `Wijzig tabblad titel`

po/nl_NL.UTF-8.po
```diff
@@ -20,7 +20,7 @@ msgstr ""
 
 #: dist/linux/ghostty_nautilus.py:53
 msgid "Open in Ghostty"
-msgstr ""
+msgstr "In Ghostty openen"
```
kjvdven: Other translations are imperative, see comment for `Change Tab Title`, I would suggest to stick to it.

So the Dutch translation would be`Open in Ghostty`

po/nl_NL.UTF-8.po
```diff
@@ -337,7 +337,7 @@ msgstr "Titel van de terminal wijzigen"
 
 #: src/apprt/gtk/class/title_dialog.zig:226
 msgid "Change Tab Title"
-msgstr ""
+msgstr "Titel van tablad wijzigen"
```
kjvdven: See previous comment

kjvdven (commented):
I recommend keeping translations in the imperative mood.

po/nl_NL.UTF-8.po
nwehg: I don't think "tabblad titel" is correct Dutch as it should be "tabbladtitel" which is why I chose to use this form. I thought "titel van tabblad" would be easier to read then "tabbladtitel" but let me know what you think.

kjvdven: You're right that `tabbladtitel` is the grammatically correct Dutch compound. 

Since the rest of the UI uses imperative form, `Wijzig tabbladtitel` has my preference, even if it's a bit of a mouthful. `Titel van tabblad wijzigen` reads more naturally but breaks the imperative pattern.

I like consistency, but that is my opinion.

nwehg: Sure, I'll update it then

nwehg:
Updated, thank you for your feedback @kjvdven

nwehg:
@kjvdven would you mind approving the pr so we can merge?

kjvdven (approved):

00-kat (approved):

PR #10847, opened 2026-02-18
i18n: update `de_DE` translations

khipp:
This PR adds the missing translation strings for Ghostty 1.3.

Related to: #10632

### DIFF
--- a/po/de_DE.UTF-8.po
+++ b/po/de_DE.UTF-8.po
@@ -22,7 +22,7 @@ msgstr ""
 
 #: dist/linux/ghostty_nautilus.py:53
 msgid "Open in Ghostty"
-msgstr ""
+msgstr "In Ghostty öffnen"
 
 #: src/apprt/gtk/ui/1.0/clipboard-confirmation-dialog.blp:12
 #: src/apprt/gtk/ui/1.4/clipboard-confirmation-dialog.blp:12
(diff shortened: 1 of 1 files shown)


trag1c:
!vouch

po/de_DE.UTF-8.po
```diff
@@ -184,7 +184,7 @@ msgstr "Tab"
 #: src/apprt/gtk/ui/1.2/surface.blp:325 src/apprt/gtk/ui/1.5/window.blp:224
 #: src/apprt/gtk/ui/1.5/window.blp:320
 msgid "Change Tab Title…"
-msgstr ""
+msgstr "Registerkartentitel ändern…"
```
rpfaeffle: "_Registerkartentitel_" is a very German wording, but the only alternatives that I came up with are "_Tab-Titel_" or "_Text in Titelleiste ändern_"

rpfaeffle (approved):
Thank you! Just a small comment, but this is more of a personal preference on how German the translation should be.

po/de_DE.UTF-8.po
khipp: I think "Tab-Titel" would be a good alternative. The German language packs of browsers also use "Tab".

rpfaeffle: Sounds good to me 👍

00-kat (approved):

PR #10855, opened 2026-02-19
core: make palette generation opt-in, not opt-out

jcollie:
Automatically generating the colors has caused much confuion among users whose theme colors have changed after upgrading to a version with pull request #10554. This PR makes the palette generation opt-in so that user's colors don't unexpectedly change.


mitchellh:
Rejected. I’d rather look more into the specific cases and gain a broader understanding of what is causing this first. If we make it opt in we’ll lose the opportunity to fix it up probably since the sample size will get too small. 

We may end up doing this eventually, but the goal initially is that this really shouldn’t have affected too many people to begin with. So I’m curious about learning more about those cases.

they can easily opt out for now

PR #10857, opened 2026-02-19
Semi functional windows apprt (very wip though)

adilahmeddev:
The apprt allows you to pick between:
- native Winui widgets + custom GDI widgets
- opengl + directx renderers

I'm not very familiar with native development, zig, or win32 so I relied on Claude for most of the heavy lifting.


adilahmeddev:
oops, didn't mean to open a PR against upstream

PR #10859, opened 2026-02-19
cmd + left/right to switch tabs by default

HaraldNordgren:
I believe this is the default in iTerm (I could be wrong!), but it's a feature that makes a lot of sense to me.

Currently, I have this setup, but maybe it makes sense to enable it with zero-config?

```bash
$ cat ~/.config/ghostty/config
keybind = super+left=previous_tab
keybind = super+right=next_tab
```

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -6299,6 +6299,16 @@ pub const Keybinds = struct {
             .{ .key = .{ .physical = .tab }, .mods = .{ .ctrl = true } },
             .{ .next_tab = {} },
         );
+        try self.set.put(
+            alloc,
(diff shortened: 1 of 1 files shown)


HaraldNordgren:
@mitchellh Could I get vouched based on my activity here?

https://github.com/ghostty-org/ghostty/discussions/4839

00-kat:
[Open a “Vouch Request” discussion](https://github.com/ghostty-org/ghostty/blob/main/CONTRIBUTING.md#first-time-contributors) rather than pinging Mitchell here.

HaraldNordgren:
@00-kat Sorry about that! Here is a vouch request:

https://github.com/ghostty-org/ghostty/discussions/10918

PR #10863, opened 2026-02-19
macOS: expand tilde in file paths before opening

AlexFeijoo44:
## Summary

Cmd-clicking a file path containing `~` (e.g. `~/Documents/file.txt`) fails to open the file on macOS because `URL(filePath:)` treats `~` as a literal directory name rather than the user's home directory.

This uses `NSString.expandingTildeInPath` to resolve `~` before constructing the file URL.

## Root Cause

In `openURL()`, when the URL string has no scheme it falls through to:

```swift
url = URL(filePath: action.url)
```

Swift's `URL(filePath:)` does not perform tilde expansion. A path like `~/Documents/file.txt` produces a URL pointing to a non-existent file, and `NSWorkspace.open` silently fails.

## Fix

```swift
let expandedPath = NSString(string: action.url).expandingTildeInPath
url = URL(filePath: expandedPath)
```

## Reproduction

1. Have a terminal application (e.g. Claude Code) that outputs file paths with `~` prefixes
2. Cmd-click the path in Ghostty on macOS
3. The file does not open (fails silently)

With this fix, the path resolves correctly and opens in the default editor.

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -695,7 +695,10 @@ extension Ghostty {
             if let candidate = URL(string: action.url), candidate.scheme != nil {
                 url = candidate
             } else {
-                url = URL(filePath: action.url)
+                // Expand ~ to the user's home directory so that file paths
+                // like ~/Documents/file.txt resolve correctly.
+                let expandedPath = NSString(string: action.url).standardizingPath
+                url = URL(filePath: expandedPath)
             }
             
             switch action.kind {


mitchellh:
!vouch

mitchellh (approved):
Thanks, this simple change looks right to me.

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -694,9 +695,12 @@ extension Ghostty {
             if let candidate = URL(string: action.url), candidate.scheme != nil {
                 url = candidate
             } else {
-                url = URL(filePath: action.url)
+                // Expand ~ to the user's home directory so that file paths
+                // like ~/Documents/file.txt resolve correctly.
+                let expandedPath = NSString(string: action.url).expandingTildeInPath
```
jparise: I wonder if we should go further and use [.standardizingPath](https://developer.apple.com/documentation/foundation/nsstring/standardizingpath) here (which performs `.expandingTildeInPath`).

mitchellh: Sure seems safe.

ignatremizov:
Hi! One follow-up from testing this area more recently in linux: I also ran into a separate `~/...` path-opening bug in the shared/core path flow.

In my case, a clickable matched path like `~/foo/bar.txt` was being resolved incorrectly to `~/~/foo/bar.txt` before open, rather than just failing in the platform opener layer. So this PR fixed one tilde-related failure mode, but not necessarily the whole class of `~/...` path issues.

Mentioning it here in case more tilde-opening reports show up in the future and someone lands in this PR first.

PR #10877, opened 2026-02-19
WIP: Make palette inversion opt-in

jake-stewart:
From #10554

> I can see there being space for some kind of new sequence that tells the terminal that "hey, I want a semantic palette" or something, that is better adjusted to themes automatically. But, I don't think we should this by default.

> So my concrete proposal is to eliminate the inversion and bg => fg ramp and do a more typical dark => light ramp (still taking into account bg/fg, but keeping 16 black and 231 white). I think this is the only way we can really make this a default on feature. I think this would address all the negative reactions we've gotten to it.

This adds a new `palette-harmonious`, disabled by default, which allows generated light themes to be inverted.

### DIFF
--- a/src/terminal/color.zig
+++ b/src/terminal/color.zig
@@ -1,9 +1,8 @@
-const std = @import("std");
+const colorpkg = @This();
 
+const std = @import("std");
 const assert = @import("../quirks.zig").inlineAssert;
 const x11_color = @import("x11_color.zig");
-
-const colorpkg = @This();
 
 /// The default palette.
 pub const default: Palette = default: {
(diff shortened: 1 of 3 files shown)


mitchellh:
See https://github.com/ghostty-org/ghostty/discussions/10852#discussioncomment-15867305 for more work to be done here.

But also see this:

> I know you said it's not ready, but I think https://github.com/ghostty-org/ghostty/pull/10877 is a big enough improvement that will fix the defaults for people in this thread that I want to ship it and see if it helps. I do think we should continue refining though.

I cleaned this up and added tests. Let's do it.

src/terminal/color.zig
```diff
@@ -90,14 +90,31 @@ pub fn generate256Color(
...
-        for (0..8) |i| base8[i] = .fromRgb(base[i]);
+        var base8: [8]LAB = .{
+            .fromRgb(bg),
+            LAB.fromRgb(base[1]),
+            LAB.fromRgb(base[2]),
+            LAB.fromRgb(base[3]),
+            LAB.fromRgb(base[4]),
+            LAB.fromRgb(base[5]),
+            LAB.fromRgb(base[6]),
+            .fromRgb(fg),
```
00-kat, line 106: A bit late, but can't all of these be decl literals?

```suggestion
            .fromRgb(bg),
            .fromRgb(base[1]),
            .fromRgb(base[2]),
            .fromRgb(base[3]),
            .fromRgb(base[4]),
            .fromRgb(base[5]),
            .fromRgb(base[6]),
            .fromRgb(fg),
```

mitchellh: Oh whoops, yes.

PR #10878, opened 2026-02-20
macos: autofixable swiftlint rules

jparise:
Apply fixes for all of the SwiftLint rules that can be automatically fixed (`--fix`) or addressed via some minor reformatting.

Each rule is in its own commit for easier review.


mitchellh (approved):
I don't see anything insulting lol. Let's go.

PR #10879, opened 2026-02-20
ci: milestone workflow should use our vouch app token

mitchellh:
This increases our rate limits and the vouch app already has the permissions required for the milestone workflow.

cc @trag1c


mitchellh:
Hm, got a 403... but it has the right permissions, I wonder if this is a race. Looking into it.

PR #10901, opened 2026-02-20
ci: Add a `skips` job where we can accumulate skip conditions

mitchellh:
This adds a new job that we can use to set outputs to accumulate skip conditions for other tests. The major change here is skipping all tests if we're only updating vouches, to save our CI.

I also included a number of minor skips based on filepaths.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,26 +11,71 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  paths:
+  # Determines whether other jobs should be skipped. Modify this if there
+  # are other fast skip conditions, and add it as an output. Then modify
+  # other tests `needs/if` to check them. Document the outputs.
+  skip:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -828,7 +874,7 @@ jobs:
       fail-fast: false
       matrix:
         i18n: ["true", "false"]
-    name: Build -Di18n=${{ matrix.simd }}
+    name: Build -Di18n=${{ matrix.i18n }}
```
mitchellh, line 877: This was pretty funny btw

mitchellh:
I want to speed up some of our slower runs too (via skipping) but focused on the low hanging fruit here.

.github/workflows/test.yml
00-kat: Single internationalization, multiple denominations.

PR #10904, opened 2026-02-20
Handle AltGr / Level 3 Shift handling for Kitty protocol

xoxorwr:
Context: https://github.com/ghostty-org/ghostty/discussions/10858

From what i understood from kitty, if non shifted modifier (so AltGr) produces different character, eg: AltGr + 4 -> `'` instead of `{`, then that character should be used

This PR works for me, however, i'm not sure if that's the right fix

cc @neurocyte https://github.com/neurocyte/flow/issues/487

### DIFF
--- a/src/input/key_encode.zig
+++ b/src/input/key_encode.zig
@@ -129,10 +129,28 @@ fn kitty(
 
         // Otherwise, we use our unicode codepoint from UTF8. We
         // always use the unshifted value.
-        if (event.unshifted_codepoint > 0) {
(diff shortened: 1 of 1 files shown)


xoxorwr:
I have no interest in getting vouched, i don't want to appear in any list

mitchellh:
No problem @xoxorwr, if that's your wish

neurocyte:
Btw, the fix might be wrong, but the core issue is valid. Ghostty is encode the event wrong for some reason. See the linked flow issue.

danielwe:
See also #2981 (reported by yourself @neurocyte!)

PR #10905, opened 2026-02-20
i18n: normalize LANGUAGE environment variable

EriksRemess:
LANGUAGE can come as en, lv:en, pt_BR:en or as lv_LV.UTF-8. This PR normalizes LANGUAGE in a form that is supported by our translations.

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -208,10 +208,9 @@ pub const Application = extern struct {
         /// Providers for loading custom stylesheets defined by user
         custom_css_providers: std.ArrayListUnmanaged(*gtk.CssProvider) = .empty,
 
-        /// A copy of the LANG environment variable that was provided to Ghostty
-        /// by the system. If this is null, the LANG environment variable did
-        /// not exist in Ghostty's environment variable.
-        saved_language: ?[:0]const u8 = null,
+        /// A map of saved locale environment variables (e.g., LANG, LANGUAGE) provided to Ghostty by the system.
+        /// If a key is missing, the environment variable did not exist in Ghostty's environment.
+        saved_locale_envs: std.StringHashMap(?[:0]const u8) = undefined,
 
         pub var offset: c_int = 0;
     };
@@ -266,17 +265,23 @@ pub const Application = extern struct {
         };
         defer config.deinit();
 
-        const saved_language: ?[:0]const u8 = saved_language: {
-            const old_language = old_language: {
-                const result = (internal_os.getenv(alloc, "LANG") catch break :old_language null) orelse break :old_language null;
+        var saved_locale_envs = std.StringHashMap(?[:0]const u8).init(alloc);
+        defer saved_locale_envs.deinit();
+
+        const locale_vars = [_][]const u8{ "LANG", "LANGUAGE" };
+        for (locale_vars) |env_name| {
+            const value = blk: {
+                const result = (internal_os.getenv(alloc, env_name) catch break :blk null) orelse break :blk null;
                 defer result.deinit(alloc);
-                break :old_language alloc.dupeZ(u8, result.value) catch break :old_language null;
+                break :blk alloc.dupeZ(u8, result.value) catch break :blk null;
             };
+            try saved_locale_envs.put(env_name, value);
+        }
 
-            if (config.language) |language| _ = internal_os.setenv("LANG", language);
-
-            break :saved_language old_language;
-        };
+        if (config.language) |language| {
+            _ = internal_os.setenv("LANG", language);
+            _ = internal_os.setenv("LANGUAGE", language);
+        }
 
         // Set gettext global domain to be our app so that our unqualified
         // translations map to our translations.
@@ -395,7 +400,7 @@ pub const Application = extern struct {
             .css_provider = css_provider,
             .custom_css_providers = .empty,
             .global_shortcuts = gobject.ext.newInstance(GlobalShortcuts, .{}),
-            .saved_language = saved_language,
+            .saved_locale_envs = try saved_locale_envs.clone(),
         };
 
         // Signals
@@ -432,7 +437,12 @@ pub const Application = extern struct {
         priv.config.unref();
         priv.winproto.deinit(alloc);
         priv.global_shortcuts.unref();
-        if (priv.saved_language) |language| alloc.free(language);
+        // Free saved_locale_envs values
+        var it = priv.saved_locale_envs.iterator();
+        while (it.next()) |entry| {
+            if (entry.value_ptr.*) |v| alloc.free(v);
+        }
+        priv.saved_locale_envs.deinit();
         if (gdk.Display.getDefault()) |display| {
             gtk.StyleContext.removeProviderForDisplay(
                 display,
@@ -461,7 +471,13 @@ pub const Application = extern struct {
     /// Get the original language that Ghostty was launched with. This returns a
     /// pointer to internal memory so it must be copied by callers.
     pub fn savedLanguage(self: *Self) ?[:0]const u8 {
-        return self.private().saved_language;
+        // Deprecated: Use savedLocaleEnvs instead
+        return self.private().saved_locale_envs.get("LANG") orelse null;
+    }
+
+    /// Get the saved locale environment variables map.
+    pub fn savedLocaleEnvs(self: *Self) std.StringHashMap(?[:0]const u8) {
+        return self.private().saved_locale_envs;
     }
 
     /// Run the application. This is a replacement for `gio.Application.run`
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1530,10 +1530,17 @@ pub const Surface = extern struct {
         var env = try internal_os.getEnvMap(alloc);
         errdefer env.deinit();
 
-        if (app.savedLanguage()) |language| {
-            try env.put("LANG", language);
-        } else {
-            env.remove("LANG");
+        const locale_envs = [_][]const u8{ "LANG", "LANGUAGE" };
+        for (locale_envs) |locale_env| {
+            if (app.savedLocaleEnvs().get(locale_env)) |value| {
+                if (value) |v| {
+                    try env.put(locale_env, v);
+                } else {
+                    env.remove(locale_env);
+                }
+            } else {
+                env.remove(locale_env);
+            }
         }
 
         // Don't leak these GTK environment variables to child processes.
--- a/src/os/locale.zig
+++ b/src/os/locale.zig
@@ -229,6 +229,15 @@ fn preferredLanguage(key: [:0]const u8, language: []const u8) void {
         if (entry.len == 0) continue;
         defer first = false;
 
+        // If entry is already a full UTF-8 locale, keep it when supported.
+        if (std.mem.endsWith(u8, entry, ".UTF-8")) {
+            for (i18n_locales.locales) |supported| {
+                if (!std.mem.eql(u8, entry, supported)) continue;
+                _ = internal_os.setenv(key, supported);
+                return;
+            }
+        }
+
         // If the first preferred language is English (Ghostty's default), stop.
         if (first and std.mem.eql(u8, entry, "en")) {
             _ = internal_os.setenv(key, "en");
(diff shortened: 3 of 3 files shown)


EriksRemess:
related to PR I closed https://github.com/ghostty-org/ghostty/pull/10631

mitchellh (changes requested):
I swear we used to have code that did this... I thought I recall finding some source in libintl that I ported over to here that did this or something similar. I don't quite know yet if this is the right approach but I will say that if it is, this does require changes:

We should pull this out into a separate function that we have a set of unit tests for (in locale.zig) that tests that a given input value is converted (or not) into the desired output value.

EriksRemess:
@mitchellh - moved it into a seperate function and added tests

mitchellh (changes requested):
I don't like calling setenv and then getenv in tests. I'd prefer if `preferredLanguage` took an Io.Writer and the caller setenv, then our tests could more easily just use fixed buffers and test that.

src/os/locale.zig
```diff
@@ -28,6 +29,25 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
         }
     }
 
+    if ((try internal_os.getenv(alloc, "LANGUAGE"))) |language| {
```
jcollie: The nesting here could be greatly reduced through the use of named blocks.

EriksRemess:
Requested changes implemented

src/os/locale.zig
```diff
@@ -28,6 +29,25 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
...
+        defer language.deinit(alloc);
+        if (language.value.len > 0) {
+            var out: [11:0]u8 = undefined; // xx_YY.UTF-8 is exactly 11 bytes long
```
jcollie: Seems rather fragile to depend on the fact that the format of a fully qualified localization will always be 11 bytes. This could lead to some very confusing undefined behavior if that every changes. I'd recommend changing this buffer to something "bigger than necessary" like 64 bytes.

src/os/locale.zig
```diff
@@ -28,6 +29,25 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
...
+            var writer: std.Io.Writer = .fixed(out[0..11]);
+
+            const ok = preferredLanguage(&writer, language.value) catch false;
+            if (ok) {
+                const value = writer.buffered();
+                if (value.len == 11) {
```
jcollie: Again, I wouldn't depend on this always being 11 bytes. Instead I'd do something like:

writer.writeByte(0) catch break :language;
const value = writer.buffered();
const preferred = value[0..value.len-1:0];

src/os/locale.zig
```diff
@@ -210,9 +230,117 @@ fn preferredLanguageFromCocoa(
     return slice[0 .. slice.len - 1 :0];
 }
 
+fn preferredLanguage(writer: *std.Io.Writer, language: []const u8) !bool {
+    var locale_buf: [11]u8 = undefined;
```
jcollie: As above, I'd give this a few extra bytes just to be sure.

src/os/locale.zig
```diff
@@ -28,6 +29,25 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
...
+            var out: [11:0]u8 = undefined; // xx_YY.UTF-8 is exactly 11 bytes long
+            var writer: std.Io.Writer = .fixed(out[0..11]);
+
+            const ok = preferredLanguage(&writer, language.value) catch false;
+            if (ok) {
+                const value = writer.buffered();
+                if (value.len == 11) {
+                    out[11] = 0;
+                    const preferred = out[0..11 :0];
+                    _ = internal_os.setenv("LANGUAGE", preferred);
```
jcollie: We should probably be saving/restoring the value of the LANGUAGE environment variable. Other applications might be expecting it to be in the original format and we don't want to break them.

https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/application.zig#L269-L279
https://github.com/ghostty-org/ghostty/blob/main/src/apprt/gtk/class/surface.zig#L1533-L1537

src/os/locale.zig
```diff
@@ -210,9 +230,117 @@ fn preferredLanguageFromCocoa(
...
+
+test "preferredLanguage keeps english when first and doesn't overwrite env" {
+    const testing = std.testing;
+    var buf: std.Io.Writer.Allocating = .init(testing.allocator);
+    defer buf.deinit();
+
+    const ok = try preferredLanguage(&buf.writer, "en:lv");
+
+    try testing.expect(!ok); // ignores LANGUAGE and just continues
+    try testing.expectEqual(@as(usize, 0), buf.written().len);
```
jcollie: The cast with `@as` shouldn't be necessary.

jcollie (changes requested):
It seems to me that the `LANGUAGE` variable not containing fully qualified language specifications is by design. If it's not working in Ghostty we need to investigate why our usage of the `gettext` library isn't working properly, this just seems to be papering over a deeper failure.

And if we _do_ need to change `LANGUAGE` we definitely should not leak it to other processes.

https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html

src/os/locale.zig
EriksRemess: added in named block "language"

src/os/locale.zig
EriksRemess: 64 bytes it is

src/os/locale.zig
EriksRemess: 64 bytes and breaks are in

src/os/locale.zig
EriksRemess: cast removed

src/os/locale.zig
EriksRemess: 64 bytes ar nice

src/os/locale.zig
```diff
@@ -28,6 +29,29 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
...
     }
 
+    language: {
+        if ((try internal_os.getenv(alloc, "LANGUAGE"))) |language| {
+            defer language.deinit(alloc);
+
+            if (language.value.len <= 0) break :language;
+
+            var out: [64:0]u8 = undefined;
+            var writer: std.Io.Writer = .fixed(out[0..out.len]);
```
jcollie: Having `out` as a sentinel-terminated array is unnecessary. I'd just do:

```
            var out: [64]u8 = undefined;
            var writer: std.Io.Writer = .fixed(&out);
```

src/os/locale.zig
```diff
@@ -28,6 +29,29 @@ pub fn ensureLocale(alloc: std.mem.Allocator) !void {
         }
     }
 
+    language: {
+        if ((try internal_os.getenv(alloc, "LANGUAGE"))) |language| {
```
jcollie: You can save another indent by putting the named block on the if:

```
if ((try internal_os.getenv(alloc, "LANGUAGE"))) |language| language:{
```

src/os/locale.zig
EriksRemess: thinking about the best way to do this

src/os/locale.zig
EriksRemess: saved

src/os/locale.zig
EriksRemess: no more sentinel

src/os/locale.zig
EriksRemess: implemented saving in a StringHashMap

src/apprt/gtk/class/application.zig
```diff
@@ -208,10 +208,9 @@ pub const Application = extern struct {
         /// Providers for loading custom stylesheets defined by user
         custom_css_providers: std.ArrayListUnmanaged(*gtk.CssProvider) = .empty,
 
-        /// A copy of the LANG environment variable that was provided to Ghostty
-        /// by the system. If this is null, the LANG environment variable did
-        /// not exist in Ghostty's environment variable.
-        saved_language: ?[:0]const u8 = null,
+        /// A map of saved locale environment variables (e.g., LANG, LANGUAGE) provided to Ghostty by the system.
+        /// If a key is missing, the environment variable did not exist in Ghostty's environment.
+        saved_locale_envs: std.StringHashMap(?[:0]const u8) = undefined,
```
jcollie, line 213: Generally we prefer the `StringArrayHashMapUnmanaged` variants. The `Array` variants generally have better performance characteristics if you're doing a lot of deletions (although it's completely unnecessary here). The `Unmanaged` variants are preferred as Zig is deprecating the managed variants.

src/apprt/gtk/class/application.zig
```diff
@@ -266,17 +265,23 @@ pub const Application = extern struct {
...
-            const old_language = old_language: {
-                const result = (internal_os.getenv(alloc, "LANG") catch break :old_language null) orelse break :old_language null;
+        var saved_locale_envs = std.StringHashMap(?[:0]const u8).init(alloc);
+        defer saved_locale_envs.deinit();
+
+        const locale_vars = [_][]const u8{ "LANG", "LANGUAGE" };
+        for (locale_vars) |env_name| {
+            const value = blk: {
+                const result = (internal_os.getenv(alloc, env_name) catch break :blk null) orelse break :blk null;
                 defer result.deinit(alloc);
```
jcollie, line 275: I think that due to the way we are managing the memory here, we can make this an `errdefer` and avoid the alloc on the next line.

src/apprt/gtk/class/application.zig
```diff
@@ -266,17 +265,23 @@ pub const Application = extern struct {
         };
         defer config.deinit();
 
-        const saved_language: ?[:0]const u8 = saved_language: {
-            const old_language = old_language: {
-                const result = (internal_os.getenv(alloc, "LANG") catch break :old_language null) orelse break :old_language null;
+        var saved_locale_envs = std.StringHashMap(?[:0]const u8).init(alloc);
+        defer saved_locale_envs.deinit();
```
jcollie, line 269: I think that we can make this an errdefer, although you'd also need to deallocate the values as well.

src/apprt/gtk/class/application.zig
```diff
@@ -266,17 +265,23 @@ pub const Application = extern struct {
...
+            try saved_locale_envs.put(env_name, value);
+        }
 
-            if (config.language) |language| _ = internal_os.setenv("LANG", language);
-
-            break :saved_language old_language;
-        };
+        if (config.language) |language| {
+            _ = internal_os.setenv("LANG", language);
+            _ = internal_os.setenv("LANGUAGE", language);
```
jcollie, line 283: The `language` config shouldn't override the `LANGUAGE` environment variable.

src/apprt/gtk/class/application.zig
```diff
@@ -395,7 +400,7 @@ pub const Application = extern struct {
             .css_provider = css_provider,
             .custom_css_providers = .empty,
             .global_shortcuts = gobject.ext.newInstance(GlobalShortcuts, .{}),
-            .saved_language = saved_language,
+            .saved_locale_envs = try saved_locale_envs.clone(),
```
jcollie, line 403: I think that we should be able to skip the `clone` here if we don't `deinit` above.

src/apprt/gtk/class/application.zig
jcollie: Ugh, nevermind. This deinit is needed.

src/apprt/gtk/class/application.zig
```diff
@@ -461,7 +471,13 @@ pub const Application = extern struct {
     /// Get the original language that Ghostty was launched with. This returns a
     /// pointer to internal memory so it must be copied by callers.
     pub fn savedLanguage(self: *Self) ?[:0]const u8 {
-        return self.private().saved_language;
+        // Deprecated: Use savedLocaleEnvs instead
```
jcollie, line 474: This function can just be deleted I think.

src/apprt/gtk/class/application.zig
```diff
@@ -461,7 +471,13 @@ pub const Application = extern struct {
...
+        return self.private().saved_locale_envs.get("LANG") orelse null;
+    }
+
+    /// Get the saved locale environment variables map.
+    pub fn savedLocaleEnvs(self: *Self) std.StringHashMap(?[:0]const u8) {
```
jcollie, line 479: Return a pointer, don't copy it.

src/apprt/gtk/class/surface.zig
```diff
@@ -1530,10 +1530,17 @@ pub const Surface = extern struct {
         var env = try internal_os.getEnvMap(alloc);
         errdefer env.deinit();
 
-        if (app.savedLanguage()) |language| {
-            try env.put("LANG", language);
-        } else {
-            env.remove("LANG");
+        const locale_envs = [_][]const u8{ "LANG", "LANGUAGE" };
```
jcollie, line 1533: Use an iterator on the saved variables rather than hardcoding them here.

EriksRemess:
closed in favor of https://github.com/ghostty-org/ghostty/pull/10976

PR #10907, opened 2026-02-21
macos: pass through paste keybind when clipboard has no text

juniqlim:
Closes #10751

## Problem

On macOS, when the clipboard contains only an image, pressing `Cmd+V` is consumed by Ghostty even though there's nothing to paste. This prevents TUI apps (e.g. Claude Code) that handle image paste natively from ever receiving the keypress.

GTK already handles this correctly (#10089) — this PR brings the same behavior to the embedded apprt.

## How it works

**1. Embedded runtime callback flow (`ghostty.h`, `embedded.zig`)**
Instead of adding a separate "has text" callback, this updates `read_clipboard_cb` to return whether a request was started:

- `ghostty_runtime_read_clipboard_cb` now returns `bool`
- it now receives `ghostty_clipboard_request_e` so the host can branch by request type (`paste`, `osc_52_*`)

This allows `clipboardRequest()` to return `false` immediately for non-started paste requests, which is what performable keybind pass-through needs.

**2. Swift implementation (`Ghostty.App.swift`)**
`readClipboard` now performs a synchronous precheck for paste requests. If the pasteboard has no text-like content, it returns `false` immediately (no async completion call), so the keypress passes through.

**3. Lightweight text-content detection (`NSPasteboard+Extension.swift`)**
`hasTextContent()` now uses `NSPasteboard.availableType(from:)` (string/url/file-url type presence) instead of calling `getOpinionatedStringContents()`, avoiding unnecessary content extraction for a boolean check.

**4. Enum safety checks (`embedded.zig`)**
Added centralized mapping helpers and compile-time/test checks for clipboard/request enum value sync to reduce silent drift risk.

## Menu Shortcut Behavior

This change intentionally aligns `Cmd+V` with `Cmd+C`:

- `copy_to_clipboard` already uses a performable binding.
- This PR makes `paste_from_clipboard` performable as well.

Performable bindings are not shown as menu shortcut labels, so `Cmd+V` may no longer appear in the menu shortcut label. This is an intentional tradeoff for consistent performable behavior.

## Testing

- `zig build test -Dtest-filter='embedded: clipboard enum mapping for runtime callbacks'`
- `zig build test -Dtest-filter='embedded: clipboard request enum mapping for runtime callbacks'`
- `zig build test -Dtest-filter='default paste keybinds are performable'`
- `zig build test`

Manual check on macOS with Claude Code: when clipboard contains only image data, paste keybind passes through and image paste is handled by the TUI app.

## AI Usage

This PR was developed with Claude Code assistance.

Codex was used to review this PR and apply follow-up fixes.

The implementation approach and code were written with AI help and manually verified on macOS.


src/apprt/embedded.zig
```diff
@@ -25,6 +25,64 @@ const log = std.log.scoped(.embedded_window);
 
 pub const resourcesDir = internal_os.resourcesDir;
 
+/// Convert apprt clipboard type to the embedded runtime C callback value.
+///
+/// Sync with ghostty_clipboard_e in include/ghostty.h.
+fn clipboardToRuntimeClipboard(clipboard: apprt.Clipboard) c_int {
```
mitchellh: For both of these, I suggest we just make Clipboard and ClipboardRequest C ABI compatible when the apprt is embedded. We already do this for GTK.

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -321,20 +324,31 @@ extension Ghostty {
...
             // Get our pasteboard
             guard let pasteboard = NSPasteboard.ghostty(location) else {
-                return completeClipboardRequest(surface, data: "", state: state)
+                completeClipboardRequest(surface, data: "", state: state)
+                return true
+            }
+
+            // For paste requests, report a non-started request when we have no
+            // text-like data so performable keybinds can pass through.
+            if request == GHOSTTY_CLIPBOARD_REQUEST_PASTE && !pasteboard.hasTextContent() {
```
mitchellh: Do we need hasTextContent here or can we use/change getOpinionatedStringContents to just use that, since we have to read it anyways?

mitchellh (changes requested):

juniqlim:
**Re: Clipboard/ClipboardRequest C ABI compatibility**
Changed `Clipboard` and `ClipboardRequestType` backing types to `c_int`, added `GHOSTTY_CLIPBOARD_PRIMARY` to `ghostty.h`, and removed the conversion functions. Types are now passed directly to callbacks.

**Re: hasTextContent**
Removed `hasTextContent()` and now use `getOpinionatedStringContents()` directly. This avoids reading the pasteboard twice and eliminates the sync requirement between the two functions.

mitchellh (changes requested):
The changes are incorrect. Could you please explain without AI why you approached it this way? I'm starting to suspect slop. I'm being purposely vague to challenge what you can discover yourself about what went wrong here, looking for some real thought.

juniqlim:
You're right — I don't fully understand the code well enough to explain what went wrong. I'll close this PR for now. I use Ghostty daily and wanted to fix something that bothered me, but I need to understand the codebase better before contributing properly.

PR #10911, opened 2026-02-21
WIP: Regenerate palette on OSC 10/11

jake-stewart:
> I think an open question I have now is if the user updates their bg/fg colors (e.g. via OSC 10, 11), should we be recalculating the default palette automatically? I think the answer there is yes. That way if a program updates itself to light/dark mode at runtime, then the palette updates too! (ignoring manually overwritten palettes via OSC4).

Maintains a working copy of the palette mask (so that palettes set with OSC4 are not overwritten) which is restored to the original config mask when the palette is reset.

I wonder if we should regenerate when some base16 colors change? (The remaining 6 corners of the color cube)

A concern is that changing a theme could cause the theme to generate up to 8 times. This can be optimized down to 3 if you combine sequences. I guess that's negligible really. Generating the palette is efficient.

### DIFF
--- a/src/termio/Termio.zig
+++ b/src/termio/Termio.zig
@@ -156,6 +156,10 @@ pub const DerivedConfig = struct {
     arena: ArenaAllocator,
 
     palette: terminalpkg.color.Palette,
+    palette_generate: bool,
+    palette_harmonious: bool,
+    palette_base: terminalpkg.color.Palette,
+    palette_mask: terminalpkg.color.PaletteMask,
     image_storage_limit: usize,
     cursor_style: terminalpkg.CursorStyle,
     cursor_blink: ?bool,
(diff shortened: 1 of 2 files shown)


jake-stewart:
FWIW, here is my script I used to test this:

1. Switches to light theme (palette should generate)
2. Sets a 256 entry
3. Switches to dark theme (palette should generate but not replace the new entry)
4. Reset fg/bg/palette (Previous entry should disappear)

All the while I have `palette = 250=#ff0000` which should remain even after theme generation and reset.

The color256.py is just a [script](https://github.com/jake-stewart/color256) which helps visualise the palette.

```
echo -e "\x1b]10;rgb:00/00/00\x1b\\"
echo -e "\x1b]11;rgb:ff/ff/ff\x1b\\"

clear; python3 ~/clones/color256/color256.py; read

echo -e "\x1b]4;233;rgb:ff/00/ff\x1b\\"

echo -e "\x1b]10;rgb:ff/ff/ff\x1b\\"
echo -e "\x1b]11;rgb:00/00/00\x1b\\"

clear; python3 ~/clones/color256/color256.py; read

echo -e "\x1b]104;n\x1b\\"
echo -e '\x1b]110\x1b\\'
echo -e '\x1b]111\x1b\\'

clear; python3 ~/clones/color256/color256.py; read
```

PR #10925, opened 2026-02-21
ci: Add `lib-vt` Android support

elias8:
The PR introduces `lib-vt` Android support as discussed in #10902. 

A few more notes:

- Introduces new CI for Android builds as a change requires NDK to be configured.
- To build locally,  it is required to have the NDK installed in the system and either have the path exported via `ANDROID_NDK_HOME` pointing to the exact NDK path or `ANDROID_HOME` or `ANDROID_SDK_ROOT` pointing at the Android SDK path from which the build system will infer the NDK path and version.
- 16kb page size alignment is configured for Android 15+. Builds are backward compatible with 4kb page size devices.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,6 +88,7 @@ jobs:
       - build-examples
       - build-flatpak
       - build-libghostty-vt
+      - build-libghostty-vt-android
       - build-libghostty-vt-macos
       - build-linux
       - build-linux-libghostty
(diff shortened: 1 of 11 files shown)


mitchellh (changes requested):
This mostly looks good to me! I want to see if we can get the Android NDK into our Nix environment (seems possible) that we can maybe use to test this more easily. I want to verify it myself. Thank you!

elias8:
Okay. I have manually tested via Docker, and that sounds good!

mitchellh (approved):
I couldn't get the Nix stuff to work yet but since we have CI coverage here I'm comfortable merging this and the changes presented. Approved pending CI.

mitchellh (changes requested):
Looks like CI failed and it isn't working, can you look into @elias8 ? Its on the Android builds.

PR #10934, opened 2026-02-21
macOS: add "command finished" notifications

### LINKED ISSUE #10840: macOS: add "command finished" notifications
Implement the macOS side of #8992

JosephMart:
fixes https://github.com/ghostty-org/ghostty/issues/10840

Implement command finished notifications for MacOS. Building on the work of #8992

### AI Tools Used
* Cursor
* Models
    * Opus 4.6
    * Composer 1.5

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -632,8 +632,12 @@ extension Ghostty {
             case GHOSTTY_ACTION_SEARCH_SELECTED:
                 searchSelected(app, target: target, v: action.action.search_selected)
 
+            case GHOSTTY_ACTION_COMMAND_FINISHED:
+                commandFinished(app, target: target, v: action.action.command_finished)
+
             case GHOSTTY_ACTION_PRESENT_TERMINAL:
                 return presentTerminal(app, target: target)
+
 
             case GHOSTTY_ACTION_TOGGLE_TAB_OVERVIEW:
                 fallthrough
@@ -1353,7 +1357,7 @@ extension Ghostty {
             n: ghostty_action_desktop_notification_s) {
             switch target.tag {
             case GHOSTTY_TARGET_APP:
-                Ghostty.logger.warning("toggle split zoom does nothing with an app target")
+                Ghostty.logger.warning("desktop notification does nothing with an app target")
                 return
 
             case GHOSTTY_TARGET_SURFACE:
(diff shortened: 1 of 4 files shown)


macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1353,25 +1356,97 @@ extension Ghostty {
...
+            target: ghostty_target_s,
+            v: ghostty_action_command_finished_s
+        ) {
+            switch target.tag {
+            case GHOSTTY_TARGET_APP:
+                Ghostty.logger.warning("command finished does nothing with an app target")
+                return
+
+            case GHOSTTY_TARGET_SURFACE:
+                guard let surface = target.target.surface else { return }
```
JosephMart, line 1414: This is essentially a dupe of https://github.com/jcollie/ghostty/blob/4581392625c09524ba330bba0b9fd37f4494d2de/src/apprt/gtk/class/surface.zig#L1070-L1129. 

Is there a shared place to push up this logic? Worth doing?

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1353,25 +1356,97 @@ extension Ghostty {
...
+                let actions = config.notifyOnCommandFinishAction
+
+                if actions.contains(.bell) {
+                    NotificationCenter.default.post(
+                        name: .ghosttyBellDidRing,
+                        object: surfaceView
+                    )
+                }
+
+                if actions.contains(.notify) {
```
JosephMart, line 1443: Anything to consider if `notify` and `bell` are enabled with `bell-features = system,attention`? Can hear both sounds playing.

Example config

```
bell-features = system,attention
notify-on-command-finish = always
notify-on-command-finish-action = notify, bell
notify-on-command-finish-after = 2s
```

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1353,25 +1356,97 @@ extension Ghostty {
...
+                    if v.exit_code < 0 {
+                        body = "Command took \(formattedDuration)."
+                    } else {
+                        body = "Command took \(formattedDuration) and exited with code \(v.exit_code)."
+                    }
 
-                center.getNotificationSettings { settings in
-                    guard settings.authorizationStatus == .authorized else { return }
-                    surfaceView.showUserNotification(title: title, body: body)
+                    showDesktopNotification(surfaceView, title: title, body: body)
```
JosephMart: Noticed that if focused, a notification isn't firing when the rest of the config specifies it should. Any pointers? Will dig further...

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1353,25 +1356,97 @@ extension Ghostty {
...
+                    }
+
+                    let body: String
+                    let formattedDuration = duration.formatted(
+                        .units(
+                            allowed: [.hours, .minutes, .seconds, .milliseconds],
+                            width: .abbreviated,
+                            fractionalPart: .hide
+                        )
+                    )
```
JosephMart: I am guessing this might differ from GTK impl but gets localized benefits.

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1353,25 +1356,97 @@ extension Ghostty {
...
+                    if v.exit_code < 0 {
+                        body = "Command took \(formattedDuration)."
+                    } else {
+                        body = "Command took \(formattedDuration) and exited with code \(v.exit_code)."
```
JosephMart, line 1464: Do/can these need to be localized?

macos/Sources/Ghostty/Ghostty.App.swift
JosephMart: Debugger seems to indicate it is due to `shouldPresentNotification` being false.

https://github.com/ghostty-org/ghostty/blob/3de6922295782cec35e155cfb43635c1da8704ab/macos/Sources/Ghostty/Ghostty.App.swift#L433-L441

Worth propagating a state to still show?

macos/Sources/Ghostty/Ghostty.App.swift
jcollie: Unfortunately the swift and core Zig code is not currently localized. The only thing that we localize so far are GTK UI strings.

macos/Sources/Ghostty/Ghostty.App.swift
jcollie: I'd have to say it'd be difficult because some of the decision-making on whether to send a notification or not depends on information that only the apprt has, namely whether the user has requested a one-off notification. That information isn't available to the core, so you'd have to find a way to get that information to the core first.

JosephMart: Do you suggest leaving as is @jcollie?

jcollie: I would.

simlf (approved):

macos/Sources/Ghostty/Ghostty.App.swift
mitchellh: Yep I got it, thanks

mitchellh (approved):
Thanks, this looks good. I made some stylistic changes and fixed the notification not showing up with focus.

PR #10943, opened 2026-02-22
macOS: Add inactive window tint overlay for liquid glass

sunshine-syz:
**Summary:**
- Add tint overlay to dim terminal windows when inactive, fixes https://github.com/ghostty-org/ghostty/discussions/10040
- Refactor the liquid glass effect into a dedicated `TerminalGlassView` class

Note: The tint overlay color and opacity values may not be ideal — feedback is welcome.

**AI Disclosure:** I used Claude Code to read the macos repo and understand the liquid glass implementation. Implemented basic tint overlay mainly by hand. Refactor the code and review changes with Claude Code.

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalViewContainer.swift
+++ b/macos/Sources/Features/Terminal/TerminalViewContainer.swift
@@ -6,9 +6,8 @@ import SwiftUI
 class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
     private let terminalView: NSView
 
-    /// Glass effect view for liquid glass background when transparency is enabled
+    /// Combined glass effect and inactive tint overlay view
     private var glassEffectView: NSView?
-    private var glassTopConstraint: NSLayoutConstraint?
     private var derivedConfig: DerivedConfig
 
     init(ghostty: Ghostty.App, viewModel: ViewModel, delegate: (any TerminalViewDelegate)? = nil) {
@@ -25,6 +24,10 @@ class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
     @available(*, unavailable)
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
     }
 
     /// To make ``TerminalController/DefaultSize/contentIntrinsicSize``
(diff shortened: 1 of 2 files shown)


sunshine-syz:
Dark before

https://github.com/user-attachments/assets/a4b5eca7-ed02-44ca-a956-b9b4dc087a17

Dark after

https://github.com/user-attachments/assets/b6bdb0e9-90c5-40a2-9060-26363432f6d0

Light before

https://github.com/user-attachments/assets/1b065d37-62a1-44ef-858f-a15260a14591

Light after

https://github.com/user-attachments/assets/332ad3d9-12bf-4d59-a682-516b5b3daab0

macos/Sources/Features/Terminal/TerminalViewContainer.swift
```diff
@@ -50,6 +49,20 @@ class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
...
         )
+
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(windowDidBecomeKey(_:)),
+            name: NSWindow.didBecomeKeyNotification,
+            object: nil
+        )
+
+        NotificationCenter.default.addObserver(
```
mitchellh, line 64: We probably need to remove the observer on deinit, yeah?

mitchellh (changes requested):
This looks pretty good to me, one comment. Anything else @bo2themax ?

macos/Sources/Features/Terminal/TerminalViewContainer.swift
zenangst: Seems like you don't have to manual unsubscribe anymore, according to the docs.


> If your app targets iOS 9.0 and later or macOS 10.11 and later, you do not need to unregister an observer that you created with this function. If you forget or are unable to remove an observer, the system cleans up the next time it would have posted to it.

https://developer.apple.com/documentation/foundation/notificationcenter/addobserver(_:selector:name:object:)

sunshine-syz: Yeah true, I think it good to have the `deinit` there, even if system already handles it on macOS 10.11+.

mitchellh (approved):

bo2themax:
> This looks pretty good to me, one comment. Anything else @bo2themax ?

LGTM as well, some minor suggestions from me tho:

Since it's been extracted to a glass container, I don't see why we should keep the topConstraint optional anymore and the overlay should always have the same frame with the glass, so we don't actually need two variables

sunshine-syz:
> > This looks pretty good to me, one comment. Anything else @bo2themax ?
> 
> LGTM as well, some minor suggestions from me tho:
> 
> Since it's been extracted to a glass container, I don't see why we should keep the topConstraint optional anymore and the overlay should always have the same frame with the glass, so we don't actually need two variables

I was about to update the code based on your suggestion, but I saw you've already opened a PR. Thanks for cleaning it up!

PR #6758, opened 2025-03-16
Add Hebrew Translations

slsrepo:
Pull request to add Hebrew translations, as discussed in the Discord.

PR #10945, opened 2026-02-22
Increase register of Hebrew localization and improve its polygendering

Laxystem:
I'm still unsure on some changes, e.g. the way the word "הגדרות" is used. I can explain every change in details but I think that'd be quite tiresome as most changes are obvious.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.log
 .DS_Store
 .vscode/
+.idea/
 .direnv/
 .envrc.local
 .flatpak-builder/
--- a/po/he_IL.UTF-8.po
+++ b/po/he_IL.UTF-8.po
@@ -107,7 +107,7 @@ msgstr "אוי, לא."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:7
 msgid "Unable to acquire an OpenGL context for rendering."
-msgstr "השגת הקשר ה־OpenGL (בלעז, \"OpenGL context\") הדרוש לרנדור המסוף נכשלה."
+msgstr "השגת הקשר ה־OpenGL הדרוש לרנדור המסוף נכשלה."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:97
 msgid ""
--- a/po/he_IL.UTF-8.po
+++ b/po/he_IL.UTF-8.po
@@ -34,7 +34,7 @@ msgstr "אישור גישה ללוח־הגזירים"
 #: src/apprt/gtk/ui/1.0/clipboard-confirmation-dialog.blp:17
 #: src/apprt/gtk/ui/1.4/clipboard-confirmation-dialog.blp:17
 msgid "Deny"
-msgstr "איסור"
+msgstr "דחייה"
 
 #: src/apprt/gtk/ui/1.0/clipboard-confirmation-dialog.blp:18
 #: src/apprt/gtk/ui/1.4/clipboard-confirmation-dialog.blp:18
@@ -47,7 +47,7 @@ msgstr "זכרו בחירתי לפיצול זה"
 
 #: src/apprt/gtk/ui/1.4/clipboard-confirmation-dialog.blp:93
 msgid "Reload configuration to show this prompt again"
-msgstr "טענו הגדרות חדשות על־מנת להציג שוב בקשה זו"
+msgstr "טענו מחדש את ההגדרות על מנת להציג בקשה זו שוב"
 
 #: src/apprt/gtk/ui/1.2/close-confirmation-dialog.blp:7
 #: src/apprt/gtk/ui/1.5/title-dialog.blp:8
@@ -68,7 +68,7 @@ msgstr "שגיאות הגדרה"
 msgid ""
 "One or more configuration errors were found. Please review the errors below, "
 "and either reload your configuration or ignore these errors."
-msgstr "נמצאו שגיאות הגדרה (לפחות אחת). אנא סקרונן כפי שמוצגות מטה, וטענו הגדרות חדשות או התעלמו מהן."
+msgstr "נמצאו שגיאות הגדרה (לפחות אחת). אנא עברו עליהן כפי שמוצגות מטה, וטענו הגדרות חדשות או התעלמו מהן."
 
 #: src/apprt/gtk/ui/1.2/config-errors-dialog.blp:10
 msgid "Ignore"
@@ -91,7 +91,7 @@ msgstr "Ghostty: סוקר המסוף"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:29
 msgid "Find…"
-msgstr "חפשו…"
+msgstr "חיפוש…"
 
 #: src/apprt/gtk/ui/1.2/search-overlay.blp:64
 msgid "Previous Match"
@@ -103,7 +103,7 @@ msgstr "תוצאה הבאה"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:6
 msgid "Oh, no."
-msgstr "אוי וֵי!"
+msgstr "אוי, לא."
 
 #: src/apprt/gtk/ui/1.2/surface.blp:7
 msgid "Unable to acquire an OpenGL context for rendering."
@@ -212,7 +212,7 @@ msgstr "ערך ריק ישחזר את כותרת ברירת־המחדל."
 
 #: src/apprt/gtk/ui/1.5/title-dialog.blp:9
 msgid "OK"
-msgstr "סבבה"
+msgstr "אוקיי"
 
 #: src/apprt/gtk/ui/1.5/window.blp:58 src/apprt/gtk/ui/1.5/window.blp:108
 msgid "New Split"
@@ -236,7 +236,7 @@ msgstr "סוקר המסוף"
 
 #: src/apprt/gtk/ui/1.5/window.blp:307 src/apprt/gtk/class/window.zig:1727
 msgid "About Ghostty"
-msgstr "על־אודות Ghostty"
+msgstr "על אודות Ghostty"
 
 #: src/apprt/gtk/ui/1.5/window.blp:312
 msgid "Quit"
@@ -298,7 +298,7 @@ msgstr "כל פעולות המסוף בחלון זה יופסקו."
 
 #: src/apprt/gtk/class/close_confirmation_dialog.zig:196
 msgid "The currently running process in this split will be terminated."
-msgstr "התהליך הרץ בפיצול יופסק."
+msgstr "התהליך הרץ בפיצול זה יופסק."
 
 #: src/apprt/gtk/class/surface.zig:1108
 msgid "Command Finished"
--- a/po/he_IL.UTF-8.po
+++ b/po/he_IL.UTF-8.po
@@ -1,6 +1,7 @@
 # Hebrew translations for com.mitchellh.ghostty.
 # Copyright (C) 2026 "Mitchell Hashimoto, Ghostty contributors"
 # This file is distributed under the same license as the com.mitchellh.ghostty package.
+# Laxystem <the@laxla.quest>, 2026.
 # Sl (Shahaf Levi), Sl's Repository Ltd <ghostty@slsrepo.com>, 2026.
 # CraziestOwl <craziestowl@proton.me>, 2025.
 #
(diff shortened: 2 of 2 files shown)


pluiedev:
Not a Hebrew speaker so pardon me if these questions are a bit dumb.

- Is "increase register" in this case making the text more formal? While the exact details are of course language-specific, our Translator's Guide does mention that we prefer more informal/personal language, and one of the ways that manifests in other localizations is that we tend to prefer the informal version of pronouns (e.g. du over Sie in German, 你 over 您 in Chinese, etc.) Has that been changed with this PR?

- I see that a lot of the words used to have an optional suffix separated by a slash — I presume that's the gender marker — which have since been replaced. Is this more like replacing gender-specific terms with generic terms (e.g. fireman → firefighter in English), or defaulting to one variant like what's done in some other languages?

Laxystem:
> Is "increase register" in this case making the text more formal?

Yes, -ish. It really means (in this case) using more features of Hebrew that are more often used in literature, which allows a more natural reading when compared to directly translating from English (comparable to English "The window was closed." v.s. "Window Closed"). 'Informal' Hebrew feels very clunky on computers.

> Is this more like replacing gender-specific terms with generic terms (e.g. fireman → firefighter in English), or defaulting to one variant like what's done in some other languages?

I've replaced all verbs with action names. I.e. instead of "close (masc)/fem", it's "closure".

Generally speaking, both changes are more consistent with other Gnome app localizations, especially the latter.

pluiedev:
Good to know. I'll defer to the @ghostty-org/he_IL team for their judgement then!

Laxystem:
@pluiedev I'll explain the philosophy behind it just for no reason's sake: In Hebrew, all verbs have gendered. No way around that. There are *some* verbs you can append a `/י` or something like that to, as-if to say the last letter can be replaced or appended onto with the female suffix(-ish, not technically a suffix); But it doesn't work with some verbs (including some used in this translation file, where `/י` was used on an IYKYK basis anyway).

So, instead of having the computer be an equal participant in the conversation like it is in other languages, I've chosen to make it a tool --- the turn off button on the AC doesn't say "turn off", it says "turn-off-ment" (the action name). Thus, what you get here.

slsrepo (changes requested):
Hello @Laxystem,
As the original translator of the Hebrew localization for Ghostty, a native speaker, and someone who constantly uses and translates Hebrew software, I wanted to share my professional feedback on this PR.
To give you some more context on my background: alongside my development work, I am an elementary school teacher (specifically in the first and second grades) and I am currently doing my Master's thesis on the development of the Hebrew language.

I put a lot of thought and effort into the original translation, including using gender-neutral language wherever possible, to ensure it is natural, accurate, and accessible. The proposed translations introduce significant regressions. To be candid, it strongly appears as though these translations were generated by AI and submitted without a thorough review by a native Hebrew speaker.

The vast majority of these proposed strings make the interface overly complex, unnecessarily formal, incorrectly pluralized, and completely out of touch with what users expect from modern software.

Here is a detailed breakdown with examples of the severe regressions this PR introduces:
 * **Action Labels & Extreme Formality**: Changing "Deny" from "דחייה" to "איסור", and "Allow" from "אישור" to "אפשור" creates a jarring, overly formal tone. "איסור" reads like a legal/prohibitive noun, not a natural UI button action, while "אפשור" is highly technical/jargon-ish. Most users expect the standard "אישור", "דחייה", or "מניעה".
 * **Incorrect Pluralization & Archaic Phrasing**: The string "Remember choice for this split" was changed from a singular UI phrasing ("זכור/י את הבחירה...") to "זכרו בחירתי לפיצול זה". This incorrectly addresses the user in the plural imperative ("זכרו") and uses an archaic, first-person possessive form ("בחירתי"). Similarly, "Find…" changes from "חפש/י…" to the plural imperative "חפשו…". This sounds unnatural and completely inconsistent with modern software.
 * **Inconsistent Tone (Slang & Yiddish)**: The PR introduces extreme stylistic inconsistencies. While some strings are heavily formalized, others use highly informal slang or culturally marked expressions. Translating the confirmation button "OK" to the slang "סבבה", or changing "Oh, no." to the dramatic Yiddish "אוי וֵי!", breaks tone consistency entirely and creates a disconnected experience for actual users.
 * **Verbose & Cluttered UI**: The translation for "Unable to acquire an OpenGL context..." was rewritten into a much heavier sentence with added explanatory parenthetical text (בלעז, "OpenGL context"). Adding unnecessary meta-text inside a UI error message clutters the interface and makes the message harder to scan.
 * **Over-formality and Hyphenation**: There are multiple places where the new version becomes more formal/literary than needed for UI, for example using explicit hyphenated formatting for "על־אודות", "לקריאה־בלבד", and "לוח־הגזירים". Those forms are not necessarily "wrong," but in aggregate they make the interface feel less natural and less consistent with common Hebrew software UX conventions.

Overall, this PR reads more like an unreviewed stylistic rewrite into a mix of formal/literary Hebrew and random slang, rather than a usability-focused UI translation pass.

To be fair, the intent behind some changes is reasonable, but the execution still falls short. For instance, shifting the recently added string "Change Tab Title" from the gendered imperative ("שנה/י את כותרת הכרטיסייה…") to a neutral action noun is fine. However, the PR translates this to "שינוי כותרת כרטיסייה" instead of the grammatically correct "שינוי כותרת הכרטיסייה", improperly dropping the definite article (ה' הידיעה) required in Hebrew construct state nouns.

Because this PR overwhelmingly degrades the overall user experience and introduces jarring errors, I am formally rejecting this PR as a reviewer and I strongly recommend against merging it. I would be more than happy to re-review or provide natural alternatives to improve the existing translation where it is actually needed, using proper, modern Hebrew.

Laxystem:
Hi! Thanks.

I am Human. I am a native speaker of Hebrew, as well as a software developer and localizer for several other FOSS projects (of the prominent ones, Forgejo), and I have a proficiency in linguistics and semitics, though none academically accreditted (for now).

Yes, my localization method is unorthodox; That is due to the Hebrew digital space being extremely over-literal in its English translations and unadapted to the modern world, and to women's existence in general (and I assume, you'd surely agree with me on this one).

> I put a lot of thought and effort into the original translation

So have I.

> on the development of the Hebrew language.

Unrelatedly, may I ask - under a Languages faculty, or a Linguistics one?

> reads like a legal/prohibitive noun

Yes, these two are ones I wasn't sure on.
Traditionally, platforms like Android defer to slang here, to avoid sounding legalese-y, as Hebrew's "allow" and "reject" equivalents are inaccurate-y. In a corporate environment I'd suggest A/B testing, but here, sure, understandable.

Side-note, "לאפשר" is a rather common noun. It is not that much of a technical jargon - certainly less than "authorize"; and regardless, a terminal is *the* place for technical jargon.

> This incorrectly addresses the user in the plural imperative

No, it does not. It purposesfully addresses\* in a gender-neutral way without using the slashes, as consistent with other GNOME apps; You *never, **ever*** see the gender-neutral slash in software. The action name (as we agree) is the correct, standard way of doing so:

image
image

And in this case, yes, the action name is inappropiate. Normally, the verb name, לזכור, is used; But I believe this sounds inarticulate. Thus, זכרו. As for the starlet - it does *not* address the user. Checkboxes address the computer (masc) or service provider (traditionally, plural). I'll bring an example from any random password manager if so needed.

> an archaic, first-person possessive form ("בחירתי")

Software often uses the traditional first-person possessive from, especially for this word. I do not see the problem here. I also do not see why one'd prefer the newer alternative ("תבחירהשלי") nor the full one ("את הבחירה שלי").

> This sounds unnatural and completely inconsistent with modern software.

In the case of Find, that is correct. I'd go with חיפוש here tbh, dunno why I didn't.

> instead of the grammatically correct "שינוי כותרת הכרטיסייה", improperly dropping the definite article (ה' הידיעה) required in Hebrew construct state nouns.

This is, again, a standard practice in software. The button lists its action, and does not concern itself with definiteness; This is due to the fact you could (usually, depends on the app) assign a keybind to the button, thus executing the action on *a* tab, and not *the* tab. Notice how in the titled version of some other strings, I drop the definitive, but not in their non-titled ones. That is standard practice not only in Hebrew, but also English and other languages. The point: to make scanning by first letter of the second word possible.

> Translating the confirmation button "OK" to the slang "סבבה"

Yeah, that should be changed.

> Adding unnecessary meta-text inside a UI error message clutters the interface and makes the message harder to scan.

That's meant for searchability. No one's gonna get an explaination for "הקשר OpenGL". I've added "OpenGL context", in quotes, so that a google search plus the word "ghostty" will bring the user to a relevant GitHub issue. Perhaps the sentence could be restructured to omit the term out of the Hebrew completely?

> Over-formality and Hyphenation

* z על־אודות is standard practice in Gnome.
* z לקריאה־בלבד is to make the words identifiable as a single element between the button and the text, as in the English "read-only".
* I actually really hate לוח־הגזירים. It is standard practice, but a terrible one imo, derived from yet another Academy "ruling". Yours, לוח העתקות, is better, but still not perfect - went for tradition here. I'd try, with your permission, to avoid using the noun entirely in favour of verbs; But I didn't want to wade off too far.

I'll push a commit now fixing some of your feedback. Once you reply, I'll fix others.

00-kat:
> That's meant for searchability. No one's gonna get an explaination for "הקשר OpenGL". I've added "OpenGL context", in quotes, so that a google search plus the word "ghostty" will bring the user to a relevant GitHub issue. Perhaps the sentence could be restructured to omit the term out of the Hebrew completely?

That UI has a direct link to the help page, Googling it isn't necessary:
https://github.com/ghostty-org/ghostty/blob/504a3611f6f2d9a1cb1a4eb419b41aea3c8049ca/src/apprt/gtk/ui/1.2/surface.blp#L4-L13

Laxystem:
Side-note: There already seems to be a non-definitived string in the main menu?
image
image

slsrepo:
Thank you for the detailed response and for your contributions to the FOSS community. I completely agree with your premise: the Hebrew digital space often suffers from overly literal English translations and creating gender-neutral, inclusive software is a high priority for me as well.

However, achieving gender neutrality and avoiding literal translations shouldn't come at the cost of natural phrasing.

To answer your side question before I address your other comments, it's linguistics inside an Education faculty. The focus is on the educational and psychological aspects of the language, its acquisition process and use.

Your recent commits are a step in the right direction, but there are still some fundamental things I want to address.

First of all, you seem to rely on GNOME's style guide and other platforms like Android. When working on Ghostty's localization, both the GNOME guide and Apple's guide were used (in case of a conflict, Apple's guide won), since these are the platforms that Ghostty runs on.

### Gender Neutrality vs. Plural Imperatives
While I respect the attempt to avoid the slash ("/") for gender neutrality, using the plural imperative ("זכרו") and possessives ("בחירתי") is not the standard solution in modern UI and I wouldn't use it. Checkboxes in UI generally represent a user's declaration of intent or a toggle of state, not a command to a plural entity. You can see my comment on this on the first pull requests adding the translations (https://github.com/ghostty-org/ghostty/pull/6758).

> In the case of Find, that is correct. I'd go with חיפוש here tbh, dunno why I didn't.

I'm fine with "חיפוש", but again, since it's explicitly a verb, I think it should stay as "חפש/י".

> You _never, **ever**_ see the gender-neutral slash in software. The action name (as we agree) is the correct, standard way of doing so.

We only partially agree, It is the correct way to do it when you are addressing the computer (masculine), but not when it's the user declaring their intent and choosing their action. In which case it's mostly masculine "for simplicity", but it's not inclusive at all. That's where the slashes (while not perfect) are better.
Using the action name is the correct way in the screenshots you posted but it's not always the case.

> Side-note: There already seems to be a non-definitived string in the main menu?

That addresses your side note too.
It's used throughout the app, where appropriate or where the verb can be dropped without hurting the meaning (צור/צרי כרטיסייה חדשה vs סגור/סגרי את הכרטיסייה isn't the same).

> the newer alternative ("תבחירהשלי") nor the full one ("את הבחירה שלי").

The "alternative" "תבחירהשלי" is not valid, in any form. It's just the letter ת and the words without spaces. I would prefer the full form, hence the translation "זכור/י את הבחירה" (shortened for brevity, since it's clear which selection we are talking about), if you want to use the action name, we could do "שמירת הבחירה לפיצול זה" (instead of זכירת).

>> Translating the confirmation button "OK" to the slang "סבבה"
>
> Yeah, that should be changed.

I disagree. While I'm fine with "אוקיי", Apple uses "אישור" across the system, so it's better to stick to it. Same goes with "אפשור" ("אישור" sounds much better too).

### Definiteness (ה' הידיעה) and Construct States
Dropping the definite article to make scanning easier by the first letter is not a standard UI practice in Hebrew when referring to a specific object. The button performs an action on the current tab, not just any general tab. "שינוי כותרת הכרטיסייה" is grammatically required here because it refers to the specific, active entity the user is manipulating. "שינוי כותרת כרטיסייה" reads as a broken construct state in this context. In English it's clear that Change Tab Title changes the current tab's title without specifying it, but in Hebrew it's not.

### Hyphenation and Formalism
While GNOME has specific stylistic guidelines regarding hyphens (על־אודות, לקריאה־בלבד), over-hyphenating makes the text look heavy and overly formal and academic for a modern developer tool. I prefer we stick to the cleaner, widely accepted spacing (על אודות, לקריאה בלבד) unless the hyphen is grammatically critical.

I appreciate your willingness to adjust the PR. I'm happy to review your next commit once you've made these changes mentioned above too, especially regarding the plural verbs.

trag1c:
Closing this as stale due to no response from the author in 5 months + this is quite behind main and it's likely that the new strings *(to be translated)* could use similar treatment so it'd be better to do this kind of thing in one sweep for consistency (assuming this PR is a net positive, which I don't know as I don't speak Hebrew). Thank you, and hopefully this PR can turn out useful in the future!

PR #10956, opened 2026-02-22
android: build improvements

jcollie:
* Use a GitHub action to download the Android NDK
* Use helper functions available on `std.Build` to simplify the build script.
* Use various Zig-isms to simplify the code.

FYI, using Nix to seems to be a non-starter as getting any Android development kits from nixpkgs requires accepting the Android license agreement and allowing many packages to use unfree licenses. And since the packages are unfree they are not cached by NixOS so the build triggers massive memory-hungry builds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -373,7 +373,6 @@ jobs:
           path: |
             /nix
             /zig
-            ~/Android/ndk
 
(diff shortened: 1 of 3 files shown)


mitchellh:
Excellent thank you ❤️

PR #10969, opened 2026-02-23
ci: enable macOS caching (Zig, Xcode)

mitchellh:
Namespace now supports cache volumes on macOS. 

This enables caching for Zig and Xcode artifacts. We can't do Nix yet because we can't create `/nix` and there's a chicken/egg with how Nix installation works on macOS. I'm emailing Namespace support about it... But still, a big win for Zig and Xcode!


mitchellh:
It's going to take awhile for the caches to be populated, but they're working:

CleanShot 2026-02-23 at 10 16 01@2x

PR #10976, opened 2026-02-24
i18n: rename `.po` files

jcollie:
This seems to be the defacto standard for naming `.po` files. See the GTK source code [1] as an example. I was unable to find any definitive documentation on the naming.

Replaces: #10905

[1] https://gitlab.gnome.org/GNOME/gtk/-/tree/main/po?ref_type=heads


00-kat (changes requested):
No idea if it should be done this way, but you also need to update [`CODEOWNERS`].

[`CODEOWNERS`]: https://github.com/ghostty-org/ghostty/blob/c51f0d745d3f8048d9f1262c9f54c34a6e4ef422/CODEOWNERS#L167-L192

EriksRemess:
will require a lot of code cleanup, doc rewriting but it works beautifully well! No need for hacks like ensureLocale, savedLanguage etc.

EriksRemess (approved):
will require a lot of code cleanup, doc rewriting but it works beautifully well! No need for hacks like ensureLocale, savedLanguage etc.

po/README_TRANSLATORS.md
```diff
@@ -44,11 +44,11 @@ intended to be regenerated by code contributors. If there is a problem with
...
 Instead, only edit the translation file corresponding to your language/locale,
-identified via its _locale name_: for example, `de_DE.UTF-8.po` would be the
+identified via its _language code_: for example, `de.po` would be the
 translation file for German (language code `de`) as spoken in Germany (country
-code `DE`). The GNU `gettext` manual contains
-[further information about locale names](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1),
-including a list of language and country codes.
+code `DE`). If your language has variants for your language as spoken in
+other countries (`zh` and `es` are examples) include the country code (e.g.
+`es_AR.po`) for Spanish as spoken in Argentina.
```
trag1c:
```suggestion
`es_AR.po` for Spanish as spoken in Argentina).
```

jcollie: Thanks!

po/README_TRANSLATORS.md
```diff
@@ -44,11 +44,11 @@ intended to be regenerated by code contributors. If there is a problem with
 the template file, please reach out to a code contributor.
 
 Instead, only edit the translation file corresponding to your language/locale,
-identified via its _locale name_: for example, `de_DE.UTF-8.po` would be the
+identified via its _language code_: for example, `de.po` would be the
 translation file for German (language code `de`) as spoken in Germany (country
-code `DE`). The GNU `gettext` manual contains
-[further information about locale names](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1),
-including a list of language and country codes.
```
00-kat, line 51: There's no longer any reference included as to where one would obtain their language code. Similarly with country codes; there is also no reference as to how one would find out if their language has country variants.

po/README_TRANSLATORS.md
00-kat: “your locale” here refers to the “locale name” earlier; this paragraph should be rewritten to refer to “`xx`” as “language code” and “`YY`” as “country code”.

00-kat: Would “full locale name” here mean `de_DE` or `de`? I feel like maybe “locale name”/“locale(s)” should be defined specifically, since “locale” is jargon that wasn't previously defined.

00-kat (changes requested):

pluiedev (approved):
Approved pending @00-kat's comments

po/README_TRANSLATORS.md
```diff
@@ -44,11 +44,25 @@ intended to be regenerated by code contributors. If there is a problem with
...
-identified via its _locale name_: for example, `de_DE.UTF-8.po` would be the
-translation file for German (language code `de`) as spoken in Germany (country
-code `DE`). The GNU `gettext` manual contains
-[further information about locale names](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1),
-including a list of language and country codes.
+identified via its _language code_, and possibly by a _country code_. For
+example, `de.po` would be the translation file for German (language code `de`)
+as spoken in Germany (country code `DE`). If your language has variants for your
+language as spoken in other countries (`zh` and `es` are examples) include the
+country code (e.g. `es_AR.po` for Spanish as spoken in Argentina).
```
trag1c:
```suggestion
as spoken in Germany (country code `DE`). If your language has variants as
spoken in different countries (`zh` and `es` are examples), include the country
code (e.g. `es_AR.po` for Spanish as spoken in Argentina).
```

po/README_TRANSLATORS.md
```diff
@@ -44,11 +44,25 @@ intended to be regenerated by code contributors. If there is a problem with
...
+```sh
+$ echo $LANG
+ca_ES.UTF-8
+```
+
+Your language code is the two characters before the `_` (in this example `ca`
+(for Catalan)). Your country code is the two characters after the `_` and
+before `.UTF-8` (in this example `ES` for Spain. If your `$LANG` environment
+variable is set to something that doesn't look like that, check the `gettext`
+documentation for more information.
```
trag1c:
```suggestion
Your language code is the two characters before the `_` (in this example `ca`
for Catalan). Your country code is the two characters after the `_` and before
`.UTF-8` (in this example `ES` for Spain). If your `$LANG` environment variable
is set to something that doesn't look like that, check the `gettext`
documentation for more information.
```

trag1c (approved):

mitchellh (approved):
Whenever everyone else is good with this so am I.

hqnna (approved):

po/README_TRANSLATORS.md
```diff
@@ -44,11 +44,25 @@ intended to be regenerated by code contributors. If there is a problem with
...
+example, `de.po` would be the translation file for German (language code `de`)
+as spoken in Germany (country code `DE`). If your language has variants as
+spoken in different countries (`zh` and `es` are examples), include the country
+code (e.g. `es_AR.po` for Spanish as spoken in Argentina).
+
+If you don't know your language code or your country code, run this command
+from a terminal:
+
+```sh
+$ echo $LANG
```
00-kat: This won't always work, since not all translators are going to be using the language they're translating to as their system language.

The previous text dealt with it by saying:

> optionally replacing `$LANG` with the name of a locale that is _different_ to your system locale, or if the `LANG` environmental variable is not set.

(N.B.: this exact text won't work here because you need to know the language code to replace `$LANG` with the language code, which is exactly what the command here is meant to help with.)

po/README_TRANSLATORS.md
```diff
@@ -44,11 +44,25 @@ intended to be regenerated by code contributors. If there is a problem with
...
+ca_ES.UTF-8
+```
+
+Your language code is the two characters before the `_` (in this example `ca`
+for Catalan). Your country code is the two characters after the `_` and before
+`.UTF-8` (in this example `ES` for Spain). If your `$LANG` environment variable
+is set to something that doesn't look like that, check the `gettext`
+documentation for more information.
```
00-kat:
```suggestion
is set to something that doesn't look like that, [check the `gettext`
documentation](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1) for more information.
```

po/README_TRANSLATORS.md
```diff
@@ -93,70 +107,45 @@ not need to modify it.
 
 You can use the `msginit` tool to create new translation files.
 
-Run the command below, optionally replacing `$LANG` with the name of a locale
-that is _different_ to your system locale, or if the `LANG` environmental
-variable is not set.
+Run the command below, replacing `xx` with your language code.
```
00-kat: This makes it sound like, given a locale which has multiple regional dialects (e.g. `zh_CN`, `zh_TW`), you only use the language code (`zh`). As far as I understand you're meant to use the language code and optionally the country code (`zh_CN`/`zh_TW`) as the language to pass to `msginit`? (I've never used `msginit`, so I could be wrong.)

po/README_TRANSLATORS.md
```diff
@@ -93,70 +107,45 @@ not need to modify it.
...
 Afterwards, you need to update the list of known locales within Ghostty's
-build system. To do so, open `src/os/i18n_locales.zig` and find the list
-of locales after the comments, then add the full locale name into the list.
+build system. To do so, open `src/os/i18n_locales.zig` and find the list of
+localizations after the comments, then add your language code into the list.
 
-The order matters, so make sure to place your locale in the correct position.
-Read the comments present in the file for more details on the order. If you're
-unsure, place it at the end of the list.
+The order matters, so make sure to place your localization in the correct
```
00-kat: I'm not sure what “localization” here means, and it sounds unnecessarily precise. I'm *assuming* that it's “languages that the Ghostty app has been localized to”, in which case:

```suggestion
supported languages after the comments, then add your language code into the list.

The order matters, so make sure to place your language in the correct
```

(GitHub's diff isn't obvious but my suggestion is: first “localization” → “supported languages”, second “localization” → “language”.)

po/README_TRANSLATORS.md
```diff
@@ -93,70 +107,45 @@ not need to modify it.
...
+position. Read the comments present in the file for more details on the order.
+If you're unsure, place it at the end of the list.
 
 ```zig
 const locales = [_][]const u8{
-    "zh_CN.UTF-8",
-    // <- Add your locale here (probably)
+    "zh_CN",
+    // <- Add your language code here (probably)
```
00-kat: This has a similar issue as with the `msginit` one, though it isn't as problematic here because there's an example. Maybe this can just use a less precise “language” so that we don't have to deal with mentioning that it's “language code, and country code if relevant”, since it's fairly obvious that you use the same identifier you used everywhere else so far.

```suggestion
    // <- Add your language here (probably)
```

po/README_TRANSLATORS.md
```diff
@@ -93,70 +107,45 @@ not need to modify it.
...
 You should then be able to run `zig build` and see your translations in action!
 
-Before opening a pull request with the new translation file, you should also add
-your locale to the `CODEOWNERS` file. Find the `# Localization` section near the
-bottom and add a line like so (where `xx_YY` is your locale):
+Before opening a pull request with the new translation file, you should also
+add your translation file to the `CODEOWNERS` file. Find the `# Localization`
+section near the bottom and add a line like so (where `xx.po` is the name of the
+translation file that you created, `xx` is your language code, and `YY` is your
+country code):
```
00-kat: I think this paragraph is *fine* but I feel like, hypothetically, someone might get confused because earlier it was said your country code only matters if there's multiple regional variations. I don't know if this will be an issue though, so feel free to ignore this comment.

00-kat (changes requested):
With the exception of the comment about the `echo` command, I think the new text is *good enough* to deal with making the documentation better after merging. (I can add an approval if we're going to do that, seeing as 1.3 is quite soon™.)

po/README_TRANSLATORS.md
```diff
@@ -35,28 +35,37 @@ Written by Ulrich Drepper.
 
 With this, you're ready to localize!
 
-## Editing translation files
+## Locale names
+
+A locale name will always consist of at least a [two letter language
```
00-kat: Nit: “at least” is redundant since “always” already says that.

```suggestion
A locale name will always consist of a [two letter language
```

po/README_TRANSLATORS.md
```diff
@@ -35,28 +35,37 @@ Written by Ulrich Drepper.
...
+code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) (i.e.
+`de`, `es`, `fr`). Sometimes, for languages that have regional variations
+(common ones are `zh` and `es`), the locale name will include a [two letter
```
00-kat: I fail to see what's “common” about those two. Do you mean they're widely spoken languages?

```suggestion
(such as `zh` and `es`), the locale name will include a [two letter
```

jcollie:
OK. I reorganized the stuff around locale names and file names, hopefully it makes more sense now.

po/README_TRANSLATORS.md
```diff
@@ -35,28 +35,37 @@ Written by Ulrich Drepper.
...
+country code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes).
+One example is `es_AR` for Spanish as spoken in Argentina. [Check the `gettext`
+documentation](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1)
+for more information on locale names.
```
00-kat: A lot of the things in that document are unnecessary/incorrect in the context of Ghostty, so a warning might be useful.

```suggestion
One example is `es_AR` for Spanish as spoken in Argentina.

Full locale names are more complicated, but Ghostty does not use all parts. [The
`gettext` documentation](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1)
has more information on locale names.
```

I don't like the word “parts” in my suggestion but can't think of anything else, feel free to pick a better word/phrasing.

po/README_TRANSLATORS.md
```diff
@@ -35,28 +35,37 @@ Written by Ulrich Drepper.
...
+
+## Translation file names
 
 All translation files lie in the `po/` directory, including the main _template_
 file called `com.mitchellh.ghostty.pot`. **Do not edit this file.** The
 template is generated automatically from Ghostty's code and resources, and are
```
00-kat: Unrelated typo lol:
```suggestion
template is generated automatically from Ghostty's code and resources, and is
```

po/README_TRANSLATORS.md
```diff
@@ -35,28 +35,37 @@ Written by Ulrich Drepper.
...
 the template file, please reach out to a code contributor.
 
-Instead, only edit the translation file corresponding to your language/locale,
-identified via its _locale name_: for example, `de_DE.UTF-8.po` would be the
-translation file for German (language code `de`) as spoken in Germany (country
-code `DE`). The GNU `gettext` manual contains
-[further information about locale names](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1),
-including a list of language and country codes.
+Translation file names consist of the locale name and end with the extension
+`.po`.
```
00-kat:
```suggestion
`.po`. For example: `de.po`, `zh_CN.po`.
```

po/README_TRANSLATORS.md
```diff
@@ -35,28 +35,37 @@ Written by Ulrich Drepper.
...
 intended to be regenerated by code contributors. If there is a problem with
 the template file, please reach out to a code contributor.
 
-Instead, only edit the translation file corresponding to your language/locale,
-identified via its _locale name_: for example, `de_DE.UTF-8.po` would be the
-translation file for German (language code `de`) as spoken in Germany (country
-code `DE`). The GNU `gettext` manual contains
-[further information about locale names](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1),
-including a list of language and country codes.
+Translation file names consist of the locale name and end with the extension
```
00-kat: Not sure how to explain why this wording feels off to me, feel free to ignore this comment lol.

```suggestion
Translation file names consist of the locale name and the extension
```

po/README_TRANSLATORS.md
```diff
@@ -93,70 +102,44 @@ not need to modify it.
...
-You should then be able to run `zig build` and see your translations in action!
+You should then be able to run `zig build run` and see your translations in action!
 
-Before opening a pull request with the new translation file, you should also add
-your locale to the `CODEOWNERS` file. Find the `# Localization` section near the
-bottom and add a line like so (where `xx_YY` is your locale):
+Before opening a pull request with the new translation file, you should also
+add your translation file to the `CODEOWNERS` file. Find the `# Localization`
+section near the bottom and add a line like so (where `X.po` is the name of the
+translation file that you created and `X` is your locale name:
```
00-kat: This isn't true, the “locale name” is `de` or `zh_CN`, while the team name is *always* both the language and country code (`de_DE`/`zh_CN`). Also, please use different letters for them, `X` for two subtly different things is a bit confusing.

po/README_TRANSLATORS.md
```diff
@@ -93,70 +102,44 @@ not need to modify it.
...
-
 `msginit` may prompt you for other information such as your email address,
 which should be filled in accordingly. You can then add your translations
 within the newly created translation file.
 
 Afterwards, you need to update the list of known locales within Ghostty's
-build system. To do so, open `src/os/i18n_locales.zig` and find the list
-of locales after the comments, then add the full locale name into the list.
+build system. To do so, open `src/os/i18n_locales.zig` and find the list of
+localizations after the comments, then add your locale name into the list.
```
00-kat:
```suggestion
locale names after the comments, then add your locale name into the list.
```

po/README_TRANSLATORS.md
```diff
@@ -93,70 +102,44 @@ not need to modify it.
...
 
-The order matters, so make sure to place your locale in the correct position.
-Read the comments present in the file for more details on the order. If you're
-unsure, place it at the end of the list.
+The order matters, so make sure to place your localization in the correct
```
00-kat:
```suggestion
The order matters, so make sure to place your locale name in the correct
```

00-kat (changes requested):

po/README_TRANSLATORS.md
jcollie: They commonly have regional variations (`zh_CN`, `zh_TW`, etc.)

00-kat: As far as I understand, it's not that they *commonly* have regional variations, but rather that they *definitely* have regional variations?

po/README_TRANSLATORS.md
```diff
@@ -35,28 +35,39 @@ Written by Ulrich Drepper.
...
+`de`, `es`, `fr`). Sometimes, for languages that have regional variations
+(such as `zh` and `es`), the locale name will include a [two letter
+country code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes).
+One example is `es_AR` for Spanish as spoken in Argentina.
+
+Full locale names are more complicated, but Ghostty does not use all parts. [The
+`gettext` documentation](https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Names-1)
+has more information on locale names.
+
+## Translation file names
```
00-kat, line 57: This section is really short, should it be merged into the next one (“Editing translation files”)?

po/README_TRANSLATORS.md
pluiedev: In the case of zh_CN vs zh_TW, I can tell you that people will definitely NOT be happy if they are confused with each other. Might not be so true for other locales, though.

po/README_TRANSLATORS.md
jcollie: I like having it separate because it kinda applies to the following two sections, and I needed a way to backref it from the "creating translation files" section.

Filip7:
Heyo, since the translation documentation is being edited in this pr, would it be useful to also mention this feature within the documentation? https://github.com/ghostty-org/ghostty/pull/10428
Because, as Kat mentioned, not all of us use our language as default on our computers.

00-kat (approved):

jcollie:
> Heyo, since the translation documentation is being edited in this pr, would it be useful to also mention this feature within the documentation? #10428 Because, as Kat mentioned, not all of us use our language as default on our computers.

The `--language` setting more of a user-level feature, not really targeted at translators specifically. It will be included in the main Ghostty documentation once 1.3 is released.

00-kat:
That feature is helpful for both people testing out translations and for Ghostty users who want to change their language. Highlighting how it can be used to test out the translations is completely unrelated to this PR so should probably be done in a separate PR though.

po/README_TRANSLATORS.md
uhojin: I'm not so sure as what to do with the Korean variety, there hasn't been much distinction in the open source scene regarding ko_KR vs ko_KP, but my understanding is that ISO639 maintainers have sanctioned ko_KP so previously when they requested a translation for Mozilla's browser it has been rejected.

From the language standpoint I think it is correct to distinguish between the variation, since they have distinct differences, but legal & geo-politically Korean is the only recognized language, so if we're going by ISO639 I guess standalone ko is good to use? plus I don't think anyone will be needing ko_KP translation either...

00-kat: glibc (2.43) doesn't seem to even have `ko_KP`:, these are all it has:
> ```
> #ko_KR.EUC-KR EUC-KR  
> #ko_KR.UTF-8 UTF-8  
> ```
(Taken from my `/etc/locale.gen`.)

jcollie: When it comes to `ko` vs `ko_KR` vs `ko_KP` we'll go with what people that speak Korean need/want.

uhojin: In that case, I believe ko_KR makes more sense in locale context just to avoid any potential confusion between 한국어 vs 조선어

00-kat: xref #11074.

PR #9078, opened 2025-10-07
i18n: add pt-PT, Português (Portugal) translation 

xiCO2k:
This PR adds the support for Português (Portugal).

Love this app 🐐

Cheers 🍻,
Francisco

PR #10986, opened 2026-02-24
feat(i18n): add Portuguese (pt_PT) translation file

jmarcelomb:
Adds a new pt_PT.UTF-8.po file with European Portuguese translations
  for the Ghostty interface and messages. This allows users to use the
  application in Portuguese, improving accessibility and localization.


trag1c:
duplicate of #9078

00-kat:
As there's already an open PR for Portuguese, consider reviewing #9078; the only reason it hasn't been merged yet is because it doesn't have a review from a Portuguese speaker.

PR #10999, opened 2026-02-24
gtk: add window title renaming

jcollie:
Fixes #10469 for GTK.


jcollie:
Pushing this to 1.4 since it adds translatable strings.

tristan957 (commented):
The GTK changes look good.

jcollie:
Updated with the latest Zig 0.16 / translations.

00-kat (approved):
Ack translation files.

src/input/Binding.zig
```diff
@@ -612,6 +617,11 @@ pub const Action = union(enum) {
     /// If the title is empty, the tab title override is cleared.
     set_tab_title: []const u8,
 
+    /// Set the title for the current focused window.
+    ///
+    /// If the title is empty, the tab title override is cleared.
```
bo2themax, line 630:
```suggestion
    /// If the title is empty, the tab title override is cleared.
    ///
    /// Only implemented on Linux.
```

src/input/Binding.zig
```diff
@@ -602,6 +602,11 @@ pub const Action = union(enum) {
     /// and persists across focus changes within the tab.
     prompt_tab_title,
 
+    /// Change the title of the current window via a pop-up prompt. The
+    /// title set via this prompt overrides any title set by the terminal
+    /// and persists across focus changes within the tab.
```
bo2themax, line 613:
```suggestion
    /// and persists across focus changes within the tab.
    /// 
    /// Only implemented on Linux.
```

bo2themax (approved):
I think we should add some comments for macOS, dunno when it will support. The array of [`unsupportedActionKeys`](https://github.com/ghostty-org/ghostty/blob/b8ab2ff16847b73afd2ea99498e44ab6694e930f/macos/Sources/Ghostty/Ghostty.Command.swift#L26) in swift needs update too, although there're already some missing keys there, like `show_on_screen_keyboard`.

But overall it doesn't block anything.

src/input/Binding.zig
jcollie: Fixed.

src/input/Binding.zig
jcollie: Fixed.

bo2themax (approved):

jcollie:
> The array of [`unsupportedActionKeys`](https://github.com/ghostty-org/ghostty/blob/b8ab2ff16847b73afd2ea99498e44ab6694e930f/macos/Sources/Ghostty/Ghostty.Command.swift#L26) in swift needs update too, although there're already some missing keys there, like `show_on_screen_keyboard`.

I wonder if there's a way at compile time or in CI to error out if there are missing keys…

bo2themax:
> > The array of [`unsupportedActionKeys`](https://github.com/ghostty-org/ghostty/blob/b8ab2ff16847b73afd2ea99498e44ab6694e930f/macos/Sources/Ghostty/Ghostty.Command.swift#L26) in swift needs update too, although there're already some missing keys there, like `show_on_screen_keyboard`.
> 
> I wonder if there's a way at compile time or in CI to error out if there are missing keys…

One thing i can think of is to make the `action_key` in `Command.C` an enum, then add a compute property in swift side to switch them all.

bo2themax:
> > > The array of [`unsupportedActionKeys`](https://github.com/ghostty-org/ghostty/blob/b8ab2ff16847b73afd2ea99498e44ab6694e930f/macos/Sources/Ghostty/Ghostty.Command.swift#L26) in swift needs update too, although there're already some missing keys there, like `show_on_screen_keyboard`.
> > 
> > 
> > I wonder if there's a way at compile time or in CI to error out if there are missing keys…
> 
> One thing i can think of is to make the `action_key` in `Command.C` an enum, then add a compute property in swift side to switch them all.

But the header still needs to be added manually I guess, i don't know if ci will fail when it's missing there, like removing `ghostty_action_prompt_title_e` changes in this pr

rockorager:
No concerns on the terminal side.

rockorager (approved):

pluiedev (approved):
Let's get this over with already

PR #11009, opened 2026-02-25
macOS: fix crash when adding tab from tab overview

nmggithub:
(moved from #11008)

I have this branch up to fix #10252. It was written with AI (in Cursor), but only after I made sure I thoroughly understood what was happening (almost to an excessive extent). I had already determined that an Objective-C helper was necessary, I just asked GPT 5.3 Codex in Cursor to write it for me. 

**TL;DR: deep within AppKit, there is an Objective-C exception that is _always_ thrown when opening a new tab from the visual tab picker ("tab overview"). [Objective-C exceptions *cannot* be safely recovered from in Swift.](http://developer.apple.com/documentation/swift/handling-cocoa-errors-in-swift#Handle-Exceptions-in-Objective-C-Only) As Ghostty is primarily Swift, we must introduce some Objective-C wrapper around tab creation to safely swallow this exception.**

There is a lot more I know about this than the above, and can discuss it at length if desired. Interestingly, it seems debug builds of Ghostty (`zig build run`) *do* gracefully recover and don't crash. Release builds (`zig build run -Doptimize=ReleaseFast`), however, *do* crash. The crashing seems to be expected behavior and **_I don't think there's any feasible way to get release builds to recover as debug builds do._** The debug builds do, arguably, have better animation behavior. Not sure how I can approach that part.

Release build off of my commit:

https://github.com/user-attachments/assets/c81927be-b2d2-48b3-a18f-30b389a90f04

Debug build off `db1e31c7a69924913e8faafcedb290de3cb4a8b6` (current `main`, as of writing):

https://github.com/user-attachments/assets/76367154-b039-4453-8d39-8a0465973deb


JiehoonKwak:
I noticed that creating new tab behavior via cmd + t ( or command palette) has same ghostty crash issue

nmggithub:
@JiehoonKwak I see no issue with such behavior on my end. Please file a discussion of your own if you are still facing such issues.

mitchellh:
Wow what the fuck. LOL. I have to wrap my head around this, so stand by I'll try to review it tomorrow. Thanks for the diligence on this.

nmggithub:
> Wow what the fuck. LOL. I have to wrap my head around this, so stand by I'll try to review it tomorrow. Thanks for the diligence on this.

You're welcome! I can explain everything to you in excruciating detail then, if you'd like. My TL;DR covers the gist, but the debug vs release builds sent me down a rabbit hole of exception handling internals that I still haven't wrapped my head around, lol.

mitchellh (commented):
EDIT: lol I slopped myself while sending Codex on a research task. Sorry :P

mitchellh (approved):
Alright, I'm at peace with this. Only made superficial changes. Biggest change was removing the exception catcher from the iOS build targets so we don't need target conditionals.

macos/Sources/Helpers/ObjCExceptionCatcher.m
```diff
@@ -12,6 +12,8 @@ BOOL GhosttyAddTabbedWindowSafely(
     NSError * _Nullable * _Nullable error
 ) {
 #if TARGET_OS_OSX
+    // AppKit occasionally throws NSException while reparenting tabbed windows.
```
nmggithub: Not sure if "while reparenting tabbed windows" is the best description here. Really it's explicitly when adding a new tab via the + tile in the tab overview. Then again, maybe that level of detail isn't required for a comment. "reparenting tabbed windows" is still a bit of an opaque description, though. It also kind of conflicts with / is different wording than the comment in the Swift extension. But I am probably just nitpicking at this point.

macos/Sources/Helpers/ObjCExceptionCatcher.h
```diff
@@ -1,9 +1,13 @@
 #import <Foundation/Foundation.h>
 
-/// Minimal Objective-C exception bridge for AppKit tabbing APIs.
+/// This file contains wrappers around various ObjC functions so we can catch
+/// exceptions, since you can't natively catch ObjC exceptions from Swift
```
nmggithub, line 4: A nit: we not only can't catch Objective-C exceptions in Swift, we can't even let them safely unwind through our frames back through to AppKit (e.g. we can't defer to AppKit's default Objective-C exception handling if our Swift frames are in the way in the call stack). Again though, maybe that's too much detail for a comment.

nmggithub:
> Alright, I'm at peace with this. Only made superficial changes. Biggest change was removing the exception catcher from the iOS build targets so we don't need target conditionals.

Yeah, I questioned the target conditionals when the AI first wrote them, but it seemed harmless and potentially fruitful at guarding us reaching out to AppKit on iOS, but it was probably overkill.

nmggithub:
@mitchellh lost this commit, lol!
https://github.com/nmggithub/ghostty/commit/886952d82a85d98bf04e0eeab963f0674b753f3b

It was just a comment fix, so I don't think it's exactly necessary, but I do think it's a bit nicer wording.

Issue #11017, opened 2026-02-25
Automatic light / dark theme switching is not fully working


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/10398

<div type='discussions-op-text'>

<sup>Originally posted by **andreynering** January 21, 2026</sup>
### Issue Description

This is a quite annoying bug that has been happening on Ghostty for quite some time. On https://github.com/ghostty-org/ghostty/discussions/9743 it is said it was supposedly fixed recently, but for me that bug is still happening.

I have this in my config:

```
theme = dark:Dracula,light:Alucard
```

I use Ghostty on macOS and I use [Nightfall](https://github.com/r-thomson/Nightfall) to be able to quickly switch between light and dark themes with a single click.

Most of the time it works as expected, but at a random way some specific panels don't switch. When it happens, switching back-and-forth a second time between light and dark themes don't usually work. You really have to close that panel or tab and open a new one.

### Expected Behavior

Themes should switch as expected:

<img width="803" height="637" alt="Screenshot 2026-01-21 at 15 03 01" src="https://github.com/user-attachments/assets/c8c15d2f-4b0f-4f48-a84f-84b63caa87d9" />

<img width="806" height="635" alt="Screenshot 2026-01-21 at 15 03 13" src="https://github.com/user-attachments/assets/4aa8d0da-3fa7-4d39-b38b-5038c414bdb3" />

### Actual Behavior

Some panels or tabs don't really switch:

<img width="806" height="637" alt="Screenshot 2026-01-21 at 15 05 52" src="https://github.com/user-attachments/assets/d39bfc2d-0e0d-4504-ad93-6c9a68395220" />

### Reproduction Steps

While writting this discussion and trying to reproduce it, I discovered that I can reproduce it by opening and closing [a TUI app](https://github.com/charmbracelet/crush) on that panel. Once you do that, the theme won't switch automatically from that point forward.

This probably means the bug is related to the fact some TUIs set a custom background color while running.

### Ghostty Version

```text
Ghostty 1.3.0

Version
  - version: 1.3.0
  - channel: stable
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue

[...]

PR #11018, opened 2026-02-25
macOS: refine window tint for liquid glass

bo2themax:
Depends on #11030

- Update constraints of `TerminalGlassView`
- Use `TerminalViewContainer.DerivedConfig` to map styling properties
- Add TerminalViewContainerTests
- Instead of using delay, now the view updates are explicitly called by window controllers

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalViewContainer.swift
+++ b/macos/Sources/Features/Terminal/TerminalViewContainer.swift
@@ -3,20 +3,15 @@ import SwiftUI
 
 /// Use this container to achieve a glass effect at the window level.
 /// Modifying `NSThemeFrame` can sometimes be unpredictable.
-class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
+class TerminalViewContainer: NSView {
     private let terminalView: NSView
 
     /// Combined glass effect and inactive tint overlay view
     private var glassEffectView: NSView?
     private var derivedConfig: DerivedConfig?
 
-    init(ghostty: Ghostty.App, viewModel: ViewModel, delegate: (any TerminalViewDelegate)? = nil) {
-        self.derivedConfig = DerivedConfig(config: ghostty.config, preferredBackgroundColor: nil, cornerRadius: nil)
-        self.terminalView = NSHostingView(rootView: TerminalView(
-            ghostty: ghostty,
-            viewModel: viewModel,
-            delegate: delegate
-        ))
+    init<Root: View>(@ViewBuilder rootView: () -> Root) {
+        self.terminalView = NSHostingView(rootView: rootView())
         super.init(frame: .zero)
         setup()
     }
--- a/macos/Sources/Features/Terminal/TerminalViewContainer.swift
+++ b/macos/Sources/Features/Terminal/TerminalViewContainer.swift
@@ -8,10 +8,10 @@ class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
 
     /// Combined glass effect and inactive tint overlay view
     private var glassEffectView: NSView?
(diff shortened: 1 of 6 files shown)


macos/Sources/Features/Terminal/TerminalStyleProtocols.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    /// Called when the Ghostty configuration has been updated.
+    ///
+    /// Re-evaluate any configuration-dependent styling such as colors, opacity,
+    /// blur effects, or other visual properties.
+    ///
+    /// - Parameters:
+    ///   - newConfig: The updated configuration to apply.
+    ///   - sender: The object that initiated the notification, or `nil` if not applicable.
+    @MainActor func ghosttyConfigurationDidChange(_ newConfig: Ghostty.Config, sender: Any?)
```
mitchellh: I feel like this whole protocol and both of these could be solved by `super` calls up to `BaseTerminalController` which does the right thing. The idea originally was that things like `syncAppearance` would basically do that and so on. Did you investigate that approach? If not, maybe we should instead of having this specific protocol.

macos/Sources/Features/Terminal/TerminalStyleProtocols.swift
```diff
@@ -0,0 +1,53 @@
...
+///
+/// Conforming types expose visual properties that a responder (such as `TerminalStyleResponder`)
+/// can query when updating its appearance — for example, reading a window's preferred background
+/// color when configuring a glass effect tint.
+protocol TerminalStyleProvider: AnyObject {
+    /// The background color that this object prefers be used for styling purposes.
+    var preferredBackgroundColor: OSColor? { get }
+
+    /// The cornerRadius that this object prefers be used for styling purposes.
+    var cornerRadius: CGFloat? { get }
```
mitchellh: Is this ever non-default? Its hard to tell but it looks like both implementations of this set it as the default...

mitchellh (changes requested):
Not fully convinced of the way this is solved, some good some bad. I asked a couple questions.

mitchellh:
Separately, maybe it'd be simpler to fix #11017 in its own dedicated PR?

macos/Sources/Features/Terminal/TerminalStyleProtocols.swift
bo2themax: We could totally do that in BaseTerminalController to retrieve this container, but since this contains an associated view model type, so the results looks a bit messy.

bo2themax:
> Separately, maybe it'd be simpler to fix #11017 in its own dedicated PR?

I can open another one without these protocols tmrw

mitchellh:
Rebase pleas then I'll review again. Thanks!

macos/Sources/Features/Terminal/TerminalViewContainer.swift
```diff
@@ -84,29 +66,22 @@ class TerminalViewContainer<ViewModel: TerminalViewModel>: NSView {
...
-        updateGlassTintOverlay(isKeyWindow: true)
-    }
+// MARK: - NSWindow + terminalViewContainer
 
-    @objc private func windowDidResignKey(_ notification: Notification) {
-        guard let window = notification.object as? NSWindow,
-              window == self.window else { return }
-        updateGlassTintOverlay(isKeyWindow: false)
+extension NSWindow {
+    var terminalViewContainer: TerminalViewContainer? {
```
mitchellh, line 83: This seems overly broad to apply as an extension. Maybe we should just add it as a var on BaseTerminalController instead?

mitchellh (changes requested):

PR #11024, opened 2026-02-25
macos: pass last focused surface as env, use for focus detection

mitchellh:
Fixes #10935

This is a more robust way to detect "is my surface focused" because that question usually means "is my surface the last focused surface" if a _different_ surface is not focused. We already have used this pattern all over but we should extend it to SwiftUI too.

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceView.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceView.swift
@@ -56,6 +56,11 @@ extension Ghostty {
         #endif
 
         @EnvironmentObject private var ghostty: Ghostty.App
+        @Environment(\.ghosttyLastFocusedSurface) private var lastFocusedSurface
+
+        private var isFocusedSurface: Bool {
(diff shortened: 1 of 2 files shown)


macos/Sources/Ghostty/Surface View/SurfaceView.swift
```diff
@@ -1201,17 +1205,33 @@ private struct GhosttySurfaceViewKey: EnvironmentKey {
...
+    static let defaultValue: Weak<Ghostty.SurfaceView>? = nil
+}
+
 extension EnvironmentValues {
     var ghosttySurfaceView: Ghostty.SurfaceView? {
         get { self[GhosttySurfaceViewKey.self] }
         set { self[GhosttySurfaceViewKey.self] = newValue }
     }
+
+    var ghosttyLastFocusedSurface: Weak<Ghostty.SurfaceView>? {
```
bo2themax: @Entry macro is now more clean

PR #11051, opened 2026-02-27
core: add tests for `ghostty.h`

jcollie:
* ensure that `ghostty.h` compiles during basic Zig tests
* ensure that non-exhaustive enums are kept synchronized between `ghostty.h` and their respective Zig counterpart.
* adjust some enums that varied from established conventions


include/test.zig
```diff
@@ -0,0 +1,5 @@
+test "ghostty_h" {
```
mitchellh: I don't think this file should go in `include`. It probably makes more sense _somewhere_ in our `src/` tree. It's kind of jarring to me when I see non header files in `include` directories because I prefer the option of copying them directly.

mitchellh (changes requested):
This is a good initiative. Instead of cImport though we should do an `addTranslateC` which is the preferred way. I've heard that Zig will eventually deprecate cImport so we might as well try to do it up front here. Example from one of my other projects: https://github.com/mitchellh/zig-quickjs-ng/blob/eb1d44ce43fd64f8403c1a94fad242ebae04d1fb/build.zig#L38-L53 This should verify all the C stuff at build-time.

Besides this, everything else looks good.

include/test.zig
mitchellh: Based on my other comment, if we do the translateC step and force it to run on tests, I don't think we need this at all, so we don't need to move this file we can just remove it.

jcollie: Done.

jcollie:
Switched to using TranslateC and removed `include/test.zig`

build.zig
```diff
@@ -291,6 +291,14 @@ pub fn build(b: *std.Build) !void {
         if (config.emit_test_exe) b.installArtifact(test_exe);
         _ = try deps.add(test_exe);
 
+        // Verify our internal libghostty header.
+        const ghostty_h = b.addTranslateC(.{
```
mitchellh, line 295: I think in the future we should move this to `GhosttyLib` and also add equivalents for `GhosttyLibVt` (libghostty-vt) to validate all of that.

mitchellh (approved):
Thanks, a great start. I noted some future stuff we should do but happy to accept this as-si for now. Thanks!

PR #11064, opened 2026-02-27
fix(macOS): filter phantom mouse events that defeat mouse-hide-while-typing

linustalacko:
## Summary

Ports the phantom mouse-motion position-equality check from the GTK runtime to the embedded runtime (used by macOS).

On macOS, TUI apps like Zellij that frequently update the window title cause phantom `mouseMoved` events at the same coordinates. These flow through `embedded.zig` → `Surface.zig` `cursorPosCallback` → `showMouse()`, which explicitly calls `NSCursor.setHiddenUntilMouseMoves(false)` and unhides the cursor, defeating `mouse-hide-while-typing`.

The GTK runtime already filters these in PR #4973 (for #3345):

```zig
const is_cursor_still = @abs(priv.cursor_pos.x - pos.x) < 1 and
    @abs(priv.cursor_pos.y - pos.y) < 1;
if (is_cursor_still) return;
```

This PR adds the same check to `embedded.zig`'s `cursorPosCallback`, using the already-stored `self.cursor_pos` field.

## Test plan

- [x] Enable `mouse-hide-while-typing = true` in Ghostty config
- [ ] Run a TUI app that updates the window title frequently (e.g. Zellij)
- [ ] Type — cursor should hide and stay hidden despite title updates
- [ ] Move the mouse — cursor should reappear normally
- [ ] Verify no regressions with normal mouse movement, focus-follows-mouse, or link hovering

### DIFF
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -848,7 +848,7 @@ pub const Surface = struct {
         mods: input.Mods,
     ) void {
         // Convert our unscaled x/y to scaled.
-        self.cursor_pos = self.cursorPosToPixels(.{
+        const pos = self.cursorPosToPixels(.{
             .x = @floatCast(x),
             .y = @floatCast(y),
         }) catch |err| {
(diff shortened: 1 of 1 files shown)


mitchellh:
!vouch

mitchellh:
I vouched you if you want to reopen this PR (you deleted your local branch). Its not the normal process but this change seems agreeable.

PR #11070, opened 2026-02-28
macOS: fix for Cmd+W window position/size restoration

abdurrahmanski:
I'd like to contribute a fix for an issue I found regarding how macOS window restoration works when a window is closed via Cmd+W (leaving the app active).

Currently, the position cascades down and to the right on every reopen, and size explicitly resets. Also, explicit `window-position-x/y` configs get ignored on first launch.

I've diagnosed the issues:
1. In `TerminalWindow.swift`, `setInitialWindowPosition` relies on the `TerminalController` which isn't present during `awakeFromNib`. I moved the `screen.origin` calculation directly into the window class to ensure fixed coordinates are respected immediately.
2. In `TerminalController.swift`, I consolidated the window spawning cascade logic into a new `applyCascade(to:hasFixedPos:)` helper. It now only calls `cascadeTopLeft` if `TerminalController.all.count > 1` (meaning another window is active) and fixed coords aren't set. If it's the only window, it anchors exactly where `LastWindowPosition` placed it.
3. In `LastWindowPosition.swift`, I updated the `save` and `restore` logic to persist the full `window.frame` (origin + size) instead of just the origin.

*Disclosure: I used Cursor (Tab) to assist in navigating the codebase and formatting the Swift code, but I fully understand the AppKit lifecycle changes and edge cases I'm proposing.*

I have the commit locally formatted with `swiftlint` and ready to push!

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -253,7 +253,14 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
             // Only cascade if we aren't fullscreen.
             if let window = c.window {
                 if !window.styleMask.contains(.fullScreen) {
-                    Self.lastCascadePoint = window.cascadeTopLeft(from: Self.lastCascadePoint)
+                    let hasFixedPos = c.derivedConfig.windowPositionX != nil && c.derivedConfig.windowPositionY != nil
+                    let shouldCascade = !hasFixedPos && TerminalController.all.count > 1
+
+                    if shouldCascade {
+                        Self.lastCascadePoint = window.cascadeTopLeft(from: Self.lastCascadePoint)
+                    } else if !hasFixedPos {
+                        Self.lastCascadePoint = NSPoint(x: window.frame.minX, y: window.frame.maxY)
+                    }
                 }
             }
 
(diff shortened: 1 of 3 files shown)


mitchellh:
!vouch

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -198,6 +198,16 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
     // of each other.
     private static var lastCascadePoint = NSPoint(x: 0, y: 0)
 
+    private static func applyCascade(to window: NSWindow, hasFixedPos: Bool) {
+        if hasFixedPos { return }
+
+        if all.count > 1 {
+            lastCascadePoint = window.cascadeTopLeft(from: lastCascadePoint)
+        } else {
+            lastCascadePoint = NSPoint(x: window.frame.minX, y: window.frame.maxY)
```
bo2themax: Second window doesn't cascade correctly

```suggestion
            lastCascadePoint = window.cascadeTopLeft(from: NSPoint(x: window.frame.minX, y: window.frame.maxY))
```

bo2themax (changes requested):

bo2themax (approved):

aofei:
Thanks, `Cmd+W` is fixed for me. @abdurrahmanski

Following up on #11015, I can reproduce two regressions after this PR:
1. `Cmd+Q` now behaves like `Cmd+W` did before this fix, as if their restore behaviors were swapped
2. Quitting from fullscreen causes the next launch to open a maximized windowed state (not fullscreen), filling the display, instead of restoring the previous window size/state

Ghostty version:

```
Ghostty 1.3.0-main+0b802e7c2

Version
  - version: 1.3.0-main+0b802e7c2
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - app runtime   : .none
  - font engine   : .coretext
  - libxev        : kqueue
```

PR #11079, opened 2026-02-28
font/shaper: add RTL shaping for Arabic and Hebrew

DiaaEddin:
Arabic and Hebrew text currently renders as disconnected, reversed glyphs because shaping is forced to LTR. This adds proper bidirectional text shaping using the Unicode Bidirectional Algorithm (UAX #9) so RTL scripts render with correct joining forms and visual order.

**Scope:** shaping only. Cursor positioning, selection, terminal model, and protocol changes are intentionally out of scope and will be addressed in follow-up work.

## Approach

Uses [itijah](https://github.com/DiaaEddin/itijah) v0.2.0 — a pure-Zig UAX #9 implementation I wrote for this integration. MIT licensed, passes the Unicode conformance test suite (BidiTest.txt, BidiCharacterTest.txt), zero C dependencies. Happy to have it mirrored to `deps.files.ghostty.org`.

itijah provides embedding level resolution and visual run derivation. The run iterator uses visual runs to emit text runs in display order with correct RTL flags, which the shapers (HarfBuzz/CoreText) then shape with the proper direction.

## Changes

- **build.zig.zon** — add itijah v0.2.0 dependency
- **SharedDeps.zig**, **GhosttyZig.zig** — wire itijah module, share uucode tables
- **uucode_config.zig** — add `bidi_class`, `bidi_paired_bracket`, `joining_type`, `is_bidi_mirrored`
- **run.zig** — itijah-based visual run iteration with cached row layout resolution
- **harfbuzz.zig** — RTL direction, cluster anchor tracking, sort for monotonic-x
- **coretext.zig** — RTL embedding via CTTypesetter, cluster anchor tracking for mark re-anchoring, sort for monotonic-x
- **bidi_helpers.zig** — shared Arabic combining mark detection
- **noop.zig**, **web_canvas.zig** — bidi scratch state and RunIterator hook

## Known limitations

- The run iterator walks visual positions, but `cursor_x` and `selection` are still in logical coordinates. On bidi lines, run splitting around cursor/selection boundaries can be incorrect. This will be addressed in the terminal-side follow-up.
- Cursor positioning is wrong when typing RTL in an interactive shell (terminal model is still LTR). Pre-composed output (`echo`, `cat`) renders correctly.
- Selection and copy follow logical order, not visual order.

## Testing

21 new tests across both backends covering Arabic and Hebrew RTL shaping, LTR/RTL direction splits, tashkeel positioning, tanween placement, digit handling in RTL context, Bengali cluster anchoring, and mixed direction runs.

Tested manually on macOS arm64 (CoreText backend). Would appreciate help verifying on Linux/x86_64 (HarfBuzz backend) and other targets.

Closes #1442
Related to #1740

AI disclosure: Claude Code (Anthropic, Claude Opus 4.6) and ChatGPT (OpenAI, GPT-5.4) for implementation assistance, code review, and test generation. Reviewed and validated manually.

### DIFF
--- a/src/font/shaper/bidi_helpers.zig
+++ b/src/font/shaper/bidi_helpers.zig
@@ -1,7 +1,8 @@
 /// Returns true for Arabic combining marks used by the RTL shaper fallback.
 ///
-/// This is intentionally scoped to Arabic marks to avoid regressing other
-/// scripts with different mark emission behavior (e.g. Chakma/Bengali).
+/// Scoped to Arabic marks to avoid regressing other scripts with different
+/// mark emission behavior (e.g. Chakma/Bengali). Uses explicit ranges because
+/// script/general_category are not yet exposed as runtime uucode fields.
 pub fn isArabicCombiningMark(cp: u32) bool {
     return switch (cp) {
         0x0610...0x061A,
--- a/src/font/shaper/bidi_helpers.zig
+++ b/src/font/shaper/bidi_helpers.zig
@@ -1,0 +1,14 @@
+/// Returns true for Arabic combining marks used by the RTL shaper fallback.
+///
+/// This is intentionally scoped to Arabic marks to avoid regressing other
+/// scripts with different mark emission behavior (e.g. Chakma/Bengali).
+pub fn isArabicCombiningMark(cp: u32) bool {
+    return switch (cp) {
+        0x0610...0x061A,
+        0x064B...0x065F,
+        0x0670,
+        0x06D6...0x06ED,
+        => true,
+        else => false,
+    };
+}
--- a/src/font/shaper/bidi_helpers.zig
+++ b/src/font/shaper/bidi_helpers.zig
@@ -1,8 +1,11 @@
-/// Returns true for Arabic combining marks used by the RTL shaper fallback.
+/// Returns true for Arabic vowel/sign marks used by the RTL shaper fallback.
 ///
-/// Scoped to Arabic marks to avoid regressing other scripts with different
-/// mark emission behavior (e.g. Chakma/Bengali). Uses explicit ranges because
-/// script/general_category are not yet exposed as runtime uucode fields.
+/// In Arabic RTL runs, shapers can report a mark before the base glyph that
+/// owns the cell. We only apply that fallback to Arabic marks because other
+/// scripts have different mark ordering rules, and using the same fallback for
+/// every zero-width mark moves some Bengali/Chakma marks to the wrong x
+/// position. Uses explicit ranges because script/general_category are not yet
+/// exposed as runtime uucode fields.
 pub fn isArabicCombiningMark(cp: u32) bool {
     return switch (cp) {
         0x0610...0x061A,
--- a/src/font/shaper/harfbuzz.zig
+++ b/src/font/shaper/harfbuzz.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const assert = @import("../../quirks.zig").inlineAssert;
 const Allocator = std.mem.Allocator;
 const harfbuzz = @import("harfbuzz");
+const itijah = @import("itijah");
 const font = @import("../main.zig");
 const terminal = @import("../../terminal/main.zig");
 const unicode = @import("../../unicode/main.zig");
@@ -37,6 +38,9 @@ pub const Shaper = struct {
     /// these separately because after shaping, HarfBuzz replaces codepoints
     /// with glyph indices in the buffer.
     codepoints: std.ArrayListUnmanaged(Codepoint) = .{},
+
+    /// Scratch reused for terminal bidi layout resolution.
+    bidi_layout_scratch: itijah.VisualLayoutScratch = .{},
 
     const Codepoint = struct {
         cluster: u32,
@@ -97,6 +101,7 @@ pub const Shaper = struct {
         self.cell_buf.deinit(self.alloc);
         self.alloc.free(self.hb_feats);
         self.codepoints.deinit(self.alloc);
+        self.bidi_layout_scratch.deinit(self.alloc);
     }
 
     pub fn endFrame(self: *const Shaper) void {
@@ -307,6 +312,10 @@ pub const Shaper = struct {
 
         pub fn finalize(self: RunIteratorHook) void {
             self.shaper.hb_buf.guessSegmentProperties();
+        }
+
+        pub fn bidiLayoutScratch(self: RunIteratorHook) *itijah.VisualLayoutScratch {
+            return &self.shaper.bidi_layout_scratch;
         }
     };
 
--- a/src/font/shaper/harfbuzz.zig
+++ b/src/font/shaper/harfbuzz.zig
@@ -839,6 +839,50 @@ test "shape arabic RTL" {
     try testing.expectEqual(@as(usize, 1), count);
 }
 
+test "shape hebrew RTL" {
+    const testing = std.testing;
+    const alloc = testing.allocator;
+
+    var testdata = try testShaperWithFont(alloc, .julia_mono);
+    defer testdata.deinit();
+
+    var t = try terminal.Terminal.init(alloc, .{ .cols = 30, .rows = 3 });
+    defer t.deinit(alloc);
+
+    var s = t.vtStream();
+    defer s.deinit();
+    try s.nextSlice("שלום עולם");
+
+    var state: terminal.RenderState = .empty;
+    defer state.deinit(alloc);
+    try state.update(alloc, &t);
+
+    var shaper = &testdata.shaper;
+    var it = shaper.runIterator(.{
+        .grid = testdata.grid,
+        .cells = state.row_data.get(0).cells.slice(),
+    });
+
+    var count: usize = 0;
+    while (try it.next(alloc)) |run| {
+        count += 1;
+        try testing.expect(run.rtl);
+        try testing.expectEqual(@as(u16, 9), run.cells);
+
+        const cells = try shaper.shape(run);
+        try testing.expect(cells.len > 1);
+
+        var x: u16 = cells[0].x;
+        try testing.expect(x < run.cells);
+        for (cells[1..]) |cell| {
+            try testing.expect(cell.x < run.cells);
+            try testing.expect(cell.x >= x);
+            x = cell.x;
+        }
+    }
+    try testing.expectEqual(@as(usize, 1), count);
+}
+
 test "shape arabic with tashkeel on middle letters" {
     const testing = std.testing;
     const alloc = testing.allocator;
@@ -2595,6 +2639,7 @@ const TestShaper = struct {
 
 const TestFont = enum {
     inconsolata,
+    julia_mono,
     monaspace_neon,
     arabic,
 };
@@ -2609,6 +2654,7 @@ fn testShaperWithFont(alloc: Allocator, font_req: TestFont) !TestShaper {
     const testEmojiText = font.embedded.emoji_text;
     const testFont = switch (font_req) {
         .inconsolata => font.embedded.inconsolata,
+        .julia_mono => font.embedded.julia_mono,
         .monaspace_neon => font.embedded.monaspace_neon,
         .arabic => font.embedded.arabic,
     };
--- a/src/font/shaper/harfbuzz.zig
+++ b/src/font/shaper/harfbuzz.zig
@@ -3,6 +3,7 @@ const assert = @import("../../quirks.zig").inlineAssert;
 const Allocator = std.mem.Allocator;
 const harfbuzz = @import("harfbuzz");
 const itijah = @import("itijah");
+const bidi_helpers = @import("bidi_helpers.zig");
 const font = @import("../main.zig");
 const terminal = @import("../../terminal/main.zig");
 const unicode = @import("../../unicode/main.zig");
@@ -18,17 +19,6 @@ const Style = font.Style;
 const Presentation = font.Presentation;
 
 const log = std.log.scoped(.font_shaper);
-
-fn isArabicCombiningMark(cp: u32) bool {
-    return switch (cp) {
-        0x0610...0x061A,
-        0x064B...0x065F,
-        0x0670,
-        0x06D6...0x06ED,
-        => true,
-        else => false,
-    };
-}
 
 /// Shaper that uses Harfbuzz.
 pub const Shaper = struct {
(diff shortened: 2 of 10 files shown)


ramysami:
Is there a build I can try for this ?

DiaaEddin:
> Is there a build I can try for this ?

you can build it yourself on your target system, if you are on macos arm I can share one with you but I am not sure it would work (needs signing etc)

ramysami:
I'm on mac arm, I'd appreciate it if you throw me the build

DiaaEddin:
@ramysami here you go https://github.com/DiaaEddin/ghostty/releases 

> I'm on mac arm, I'd appreciate it if you throw me the build

ramysami:
Works! Thank you.

oronbz:
The build works great! Finally proper support! Hope to see it in soon, thank you so much <3

commandlinetips:
Thank you for this it helped, I have test your version and it appear your fix is better in minimal change but I encounter some litters don't join correctly like الأن  appear not normally i hope my fix help you https://github.com/commandlinetips/ghostty

DiaaNoon:
> Thank you for this it helped, I have test your version and it appear your fix is better in minimal change but I encounter some litters don't join correctly like الأن appear not normally i hope my fix help you https://github.com/commandlinetips/ghostty

could you share screenshot of how it appears now 
and screenshot  of the expected result ?

commandlinetips:
> > Thank you for this it helped, I have test your version and it appear your fix is better in minimal change but I encounter some litters don't join correctly like الأن appear not normally i hope my fix help you https://github.com/commandlinetips/ghostty
> 
> could you share screenshot of how it appears now and screenshot of the expected result ?

sure this 2 image show the result :

DiaaEddin:
> > > Thank you for this it helped, I have test your version and it appear your fix is better in minimal change but I encounter some litters don't join correctly like الأن appear not normally i hope my fix help you https://github.com/commandlinetips/ghostty
> > 
> > 
> > could you share screenshot of how it appears now and screenshot of the expected result ?
> 
> sure this 2 image show the result :  


interesting, I assume what you shared is the correct output from your fork, what I also needed is the output that is wrong and what's the exact issue? 
so could you share: 
two screen shots of the  exact same text  that is problematic, one from this pr  build and why/how is it wrong, the other screenshot  is how the text  should look like 


I am testing couple of things on my end as-well but your  input would be very helpful to pinpoint the issue 

## UPDATE: 
tested your fork on the coretext_harfbuzz backend. This is a font-level Arabic ligature issue, not a shaping bug.

Some Arabic fonts collapse multiple codepoints into a single glyph,  e.g. `ل + ا `(two codepoints / two terminal cells) shape into one `لا` glyph drawn at the width of a single cell, leaving the other cell empty. Longer sequences like` الله` collapse four codepoints into one glyph, with three cells left empty. No shaper can fix this in a monospace grid,  the only real workaround is to use a monospace-friendly Arabic font that keeps these letters as separate glyphs. Out of scope for this PR.

Could you test` الله, الإستقلال, ... `and a few tashkeel cases on your fork? Some of those render incorrectly there too.

commandlinetips:
Thank you @DiaaEddin  for testing and the detailed explanation, that really helped me understand the ligature issue. Here are the screenshots you asked for:

this is in my machine with your repo .

this one with my repo .

src/font/shaper/bidi_helpers.zig
```diff
@@ -0,0 +1,15 @@
+/// Returns true for Arabic combining marks used by the RTL shaper fallback.
+///
+/// Scoped to Arabic marks to avoid regressing other scripts with different
+/// mark emission behavior (e.g. Chakma/Bengali). Uses explicit ranges because
+/// script/general_category are not yet exposed as runtime uucode fields.
```
pluiedev: This comment doesn't make that much sense to me from an admittedly ignorant LTR user — could you elaborate on this a bit more without using obscure terminology like e.g. "emission behavior"?

src/font/shaper/harfbuzz.zig
```diff
@@ -164,6 +181,10 @@ pub const Shaper = struct {
 
         // This keeps track of the cell starting x and cluster.
         var cell_offset: CellOffset = .{};
+        const anchor_sentinel = std.math.minInt(i32);
+        self.cluster_anchor_x.clearRetainingCapacity();
+        try self.cluster_anchor_x.ensureUnusedCapacity(self.alloc, run.cells);
+        self.cluster_anchor_x.appendNTimesAssumeCapacity(anchor_sentinel, run.cells);
```
pluiedev: clear + ensure unused + append N times assuming capacity is just clear + append N times

pluiedev (changes requested):
Hey, sorry for taking a while to review this. I think what really made the difference here for me was using a pure-Zig library instead of something like Fribidi — the logic here is a lot more straightforward than any of my experiments a while back. I only caught one Zig nitpick but that shouldn't be hard to fix.

Just one request — I'd like to see more comments documenting the RTL codepaths and any special-case handling we have to do there. The LTR exceptions are already fairly obvious to me who only knows LTR languages, so it'd be great to see the RTL parts elaborated upon too. Other than that, this looks pretty great

src/font/shaper/bidi_helpers.zig
DiaaEddin: Reworded. I replaced “mark emission behavior” with a more concrete explanation: this fallback handles Arabic RTL marks that can arrive before their base glyph, but broadening it to all zero-width marks regressed Bengali/Chakma positioning.

src/font/shaper/harfbuzz.zig
DiaaEddin: Fixed here and in the matching CoreText path. After `clearRetainingCapacity()`, `appendNTimes(...)` expresses the same operation directly.

DiaaEddin:
Thanks for the review!

I pushed the two inline fixes and added more notes around the RTL bits in coretext.zig and harfbuzz.zig: cluster-to-cell mapping, the CoreText position.x anchor case, and the final sort before returning cells.

Also bumped itijah to 0.2.1; no Ghostty-facing API changes there.

Verified with `zig build test`.

DiaaEddin:
@commandlinetips which distro is this btw? could you share fontname/distro ?

taibaran:
Hi @DiaaEddin — thanks for this PR, it's excellent work. I tested it in [cmux](https://github.com/manaflow-ai/cmux) (a Ghostty-based macOS terminal) and the shaping works great: Hebrew and Arabic render in correct visual order, including mixed Hebrew/Latin/digits — verified live by typing Hebrew into a TUI. 🎉

I'd like to offer a fix for the main limitation you flagged — **the cursor still being drawn at the logical column** (so when typing Hebrew the text is correct but the caret sits on the mirrored cell).

**Approach:** `RenderState.update` computes, for the cursor's row, the *same* itijah visual layout the shaper uses (LTR base direction, identical trailing-empty trim and `NUL`→space mapping), finds the run containing the logical cursor column, and stores the visual column on `Cursor.Viewport.visual_x`. A `hasStrongRtl` fast path keeps pure-LTR rows byte-identical and allocation-free (no perf regression for the common case). The Metal renderer's `addCursor` and the `cursor_pos` uniform then draw at `visual_x`. Includes tests (logical→visual on a Hebrew row; identity on LTR).

- Branch (your PR + this commit on top): https://github.com/taibaran/ghostty/tree/bidi-cursor
- Commit: https://github.com/taibaran/ghostty/commit/3f07708bb

Happy to open a PR against your branch or fold it into yours — whatever you prefer. (Mouse/selection visual mapping is the natural next follow-up.) Thanks again!

zughayyar:
This should be added, its a must feature.

d-knafo:
+1

youssefadly237:
Hi thanks for this pr

I have built it and tested it with neovim `termbidi` set to true

I had some problems

**Spaces are not detected correctly**

image

image

I am not sure why do this happen, I did not look into the code but as you can see in Gnome terminal the spaces are grayed out, but in ghostty, spaces are index based as if the text was not flipped for Arabic, so it confuses letters for spaces

**letter continuity breaks on cursor**

as you can see also in the previous images the letter `ع` when highlighted by the cursor it breaks, which affects ghostty also but not neovim

also there is a regression in the apt loading bar (which is made up of `U+2588` `█████`)

image

which used to look smooth on the `main` branch

image

lmk if you need more help with testing

OmarAshour02:
Would love to see this merged, proper Arabic support would make Ghostty much more usable for me.

OneAbodFTW:
@DiaaEddin Thank you for the fantastic wor, just tried your build, works great!

guysoft:
> **Disclaimer:** This comment was drafted by an AI agent (opencode) assisting me. I will review everything below and remove this disclaimer once I've verified it. The `.auto_ltr` fix I tested locally — it does solve edge cases, with tests attached in this branch.

---

Hi @DiaaEddin — thanks for this PR, great work (really, this is not thr ai bit). I rebased your `bidi-support` branch onto current upstream `main` (the PR was `mergeable_state: dirty` — conflicts with main) and added one small fix on top to help move this toward merge.

Branch: **https://github.com/guysoft/ghostty/tree/feature/rtl**
Happy to split into separate commits or open a cross-fork PR if that's easier for you to pull.

## What I added

### 1. Rebase onto current `main`

The PR had merge conflicts (`mergeable_state: dirty` on GitHub). I rebased your 9 commits onto `origin/main` HEAD. The main conflict was that upstream changed `VtStream.nextSlice`'s return type from error-union to `void`, so `try s.nextSlice(...)` in your test suites no longer compiles. I removed `try` from those call sites in `harfbuzz.zig` (the ones in `coretext.zig` I left for a macOS contributor to handle, since I can't test that backend on Linux). No behavioral change — purely mechanical.

### 2. One-line paragraph-direction fix (`src/font/shaper/run.zig:360`)

The `LayoutOptions.base_dir` was hardcoded to `.ltr`, which bypasses UAX #9 P2-P3 first-strong paragraph direction detection. This breaks trailing ASCII punctuation in RTL-leading rows.

**Example:** `"שלום, עולם! מה שלומך?"`

Under forced `.ltr` (base_level = 0):
- Trailing `?` is a neutral (Bidi_Class = ON) at the end of the row
- `computeSequenceSosEos` falls back to `succ_level = base_level = 0` → `eos = L`
- N1: `prev_type = R` (Hebrew on left), `next_type = eos = L` → **no match**
- N2: neutral adopts embedding direction → `L`
- `?` ends up at level 0 (LTR skeleton); `applyL2` doesn't reverse even-level runs
- **Result:** `?` appears at the visual-left, ahead of the Hebrew text ❌

Under `.auto_ltr` (P2-P3 picks RTL from first strong `ש`):
- `base_level = 1` (RTL paragraph)
- `eos = classFromLevel(max(1, 1)) = R`
- N1: `prev = R`, `next = R` → **match** → neutral resolves to `R`
- `?` stays at level 1 (RTL run); `applyL2` reverses odd-level runs
- **Result:** `?` appears at the visual-right (end of sentence) ✅

Rows starting with LTR text are byte-identical — P2-P3 detects `H` (L) → `base_level = 0`, same as before. Rows with no strong chars fall back to LTR (the "ltr" in `auto_ltr`). itijah passes Unicode conformance (BidiTest.txt, BidiCharacterTest.txt), so this is a one-line integration fix, not an algorithm change.

I tested this locally on Linux/x86_64 with the HarfBuzz backend. Hebrew text with trailing punctuation now renders correctly (punctuation at visual-right). All existing shaper tests continue to pass.

### 3. Regression test (`src/font/shaper/harfbuzz.zig` only)

Added `test "shape hebrew RTL with trailing punctuation at visual-right"` feeding `"שלום, עולם! מה שלומך?"` (21 cells) and asserting:
- Exactly **1** visual run (not split)
- The run is RTL
- `run.cells == 21` (the `?` is part of the same RTL run as the Hebrew)

Under the old forced-LTR this would produce 2+ runs (the `?` splits off into a separate LTR run). Without this test, all existing hebrew/rtl/bidi/arabic tests pass under *both* `.ltr` and `.auto_ltr` because none of them feed trailing ASCII punctuation in an RTL-leading row — so the fix would be silently unlockable. I only added the test to `harfbuzz.zig` (Linux-testable); a macOS contributor can mirror it to `coretext.zig` if desired.

I intentionally did **not** touch `coretext.zig` — I'm on Linux and can't test the CoreText backend. The `try` removal in coretext.zig can be done by DiaaEddin or a macOS contributor separately.

## Test results (Linux/x86_64, HarfBuzz 11.0.0)

All RTL shaper tests pass on the rebased branch:
- `zig build test -Dtest-filter="hebrew"` → ✅
- `zig build test -Dtest-filter="rtl"` → ✅
- `zig build test -Dtest-filter="bidi"` → ✅
- `zig build test -Dtest-filter="arabic"` → ✅
- `zig build test -Dtest-filter="trailing punctuation"` → ✅ (new test)

Full `shape` suite: 108/109 pass (1 pre-existing Chakma vowel test failure, unrelated to RTL — same on `fork/bidi-support`).

Build: `zig build -Dapp-runtime=gtk` succeeds with GTK 4.22.4, libadwaita 1.9.0.

## On the review status

Looks like all of @pluiedev's review comments from May 12 were addressed in your `0d4ba38` commit on May 17 (the `clearRetainingCapacity` + `appendNTimes` nitpick, and the `bidi_helpers.zig` comment reword). The PR is still marked `CHANGES_REQUESTED` because GitHub doesn't auto-clear that flag when the author replies — it needs a second-pass review from @pluiedev to flip it to `APPROVED`. The PR has been idle for ~2.5 months since the feedback was addressed; happy to help with anything else needed to land this for 1.4.0.

Thanks again — the itijah integration is clean and the pure-Zig approach made it easy to trace through the N1/N2 path to pinpoint the paragraph-direction issue.

body20002:
Is there something that prevent this from getting merged?
Can we get update on this?

jcollie:
One thing that is definitely missing is the lack of benchmarking. We're very proud of Ghostty's performance so we want to ensure that any changes merged don't degrade performance, especially for non-RTL language users. It would also be great if more native RTL language users would use this branch and comment on how well it works. None of the core maintainers are fluent in any RTL language so it's hard for us to tell if this really makes things better.

youssefadly237:
I think I spotted a regression that was not fixed, and a couple of bugs too

asibahi:
> One thing that is definitely missing is the lack of benchmarking. We're very proud of Ghostty's performance so we want to ensure that any changes merged don't degrade performance, especially for non-RTL language users. It would also be great if more native RTL language users would use this branch and comment on how well it works. None of the core maintainers are fluent in any RTL language so it's hard for us to tell if this really makes things better.

I think it would be helpful, on the feedback whether this is a MVP which works correctly or not, if some call for testing happened on the official channels for Ghostty (so more users would see it). The only reason I saw your message right now is because I knew this PR existed back when @DiaaEddin announced the library on ziggit and I was curious on its progress.

guysoft:
> I think it would be helpful, on the feedback whether this is a MVP which works correctly or not

I tested it and it worked, though there was a minor detail that punctuation is not aligned correctly, it can be fixed by setting ``auto-ltr``

hamaney:
I had this issue for almost a year now. however with https://github.com/manaflow-ai/cmux I was able to overcome this as it has full support for RTL text. I recommend you check it out guys!

PR #11094, opened 2026-03-01
🐛 Prevent git log output with signature information

drepper:
When users have something like

[log]
        showSignature = true

in their .gitconfig files, invocations of the log or show git sub-command emit additional information about signatures.  This additional output disturbs the generation of short_hash in GitVersion.zig, the additional text is copied verbatim into the string and then shown in the CSI >q output.

To fix it always suppress the output of the signature information.  This has no effects when the setting is disabled anyway.


jcollie:
!vouch

mitchellh:
Thanks, this looks right to me!

PR #11103, opened 2026-03-01
Clear key state overlay on "ignore" action

cespare:
This fixes a bug in the key state sequence overlay.

## Demo

In my ghostty config, I have

    keybind = ctrl+space>escape=ignore
    keybind = ctrl+space>p=toggle_command_palette
    ...

because I use `ctrl+space>` sequences for most things and so hitting `esc` is my way to bail out of the sequence if I change my mind.

I just switched to tip and got the new GTK key sequence overlay. Here's what I saw. In these screen recordings, the sequence of keys I press is

    ctrl+space, escape, ctrl+space, escape, ctrl+space, escape, ctrl+space, p

https://github.com/user-attachments/assets/4a37bc7e-b75c-4bd1-99de-f21f4211b5b5

after the fix:

https://github.com/user-attachments/assets/023be88e-1299-4219-920c-1b1134b2888c

## Notes

I believe this was also a leak, since the queued keys wouldn't be deinited.

**AI usage:** Claude Code suggested the fix, then I read enough code to convince myself that it makes sense.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2973,6 +2973,9 @@ fn maybeHandleBinding(
         // If our action was "ignore" then we return the special input
         // effect of "ignored".
         for (actions) |action| if (action == .ignore) {
+            // If we're in a sequence, clear it.
+            self.endKeySequence(.drop, .retain);
+
             return .ignored;
         };
     }


mitchellh (approved):
This looks right to me. Thank you!

mitchellh:
> because I use ctrl+space> sequences for most things and so hitting esc is my way to bail out of the sequence if I change my mind.

Btw, you may be interested in the new 1.3 `catch_all` binding: `ctrl+space>catch_all=ignore` will let you press ANY key to escape out (that isn't bound).

cespare:
> Btw, you may be interested in the new 1.3 `catch_all` binding: `ctrl+space>catch_all=ignore` will let you press ANY key to escape out (that isn't bound).

Very nice, thanks!

PR #11109, opened 2026-03-01
fuzz: `terminal.vtStream` fuzzer

mitchellh:
This augments our libghostty fuzzing to add fuzzing for `terminal.vtStream` which exercises a LOT more codepaths than the pure parser (thousands of tuples compared to hundreds with `afl-showmap` on the two binaries). I also fixed up a few more minor things: prettier ignores AFL related files, lib-vt exports the readonly streams, etc.


mitchellh:
Still lots of crashes to investigate from running this locally but this is stable enough to run now.

Issue #7879, opened 2025-07-09
macOS: Swift unit tests


mitchellh:
Our Zig codebase is very well tested, relatively speaking. Our macOS app has zero unit tests.

We should adopt unit tests for the areas of the macOS app that are pure business logic. Things like SplitTree come to mind as something that isn't trivial and should be tested, but there are sure more. 

This should also force us to have a clean separation of view and model because we'll have a tough time unit testing view stuff in unit tests. 

**If you want to work on this issue:** start by PR-ing only a small set of unit tests. Don't aim for a large amount of coverage. Incremental improvement on this issue is completely fine and I'd rather review that than something monstrous.

Also, we should run these tests in CI. `xcodebuild` can run tests.

jparise:
These days, we have a choice of using either XCTest or Swift Testing for unit tests. Should we express a preference here before we get a mix of contributions?

(Swift Testing is pretty nice. 😉)

mitchellh:
I've used XCTest a lot, I don't have experience with Swift Testing. I know XCTest has really good Xcode integration (and Instruments). Does Swift testing have the same? (e.g. right click next to the test to run or profile)

jparise:
Yes, they're the same in that regard. The main downside to Swift Testing is that it's "new" so there is just less overall experience with it, but I haven't run into any issues with it in my other projects.

mitchellh:
Given our app is all-in on Swift, I'm fine with Swift testing. I think UI tests (not this issue) have to be XCTest though, yeah? (XCUITest specifically)

mitchellh:
FYI, the recent benchmark merge actually brings in the Swift Testing based unit test target. So this task got that much easier.

00-kat:
> the recent benchmark merge

xref: #7891.

mitchellh:
#7909 runs Xcode unit tests in CI.

pouwerkerk:
I wrote up some [unit tests](https://github.com/ghostty-org/ghostty/blob/0501643a9180cade850d725cfa2930ea0000c16c/macos/Tests/Splits/SplitTreeTests.swift) for [`SplitTree.swift`](https://github.com/ghostty-org/ghostty/blob/main/macos/Sources/Features/Splits/SplitTree.swift) today.

Was fun to understand how this works under the hood and I'm pretty happy with 93.13% (949/1019) coverage. Would love to open a PR with it once [I'm vouched](https://github.com/ghostty-org/ghostty/discussions/10725).

[...]

PR #11115, opened 2026-03-02
macos: add unit tests for QuickTerminalSize 

markdorison:
Tests cover Size.toPixels() (percentage conversion, pixel passthrough,
0% and 100% boundaries) and calculate() for all five positions: left,
right, top, bottom, and center (both landscape and portrait orientations).

Verifies that primary and secondary axes map to the correct screen
dimensions per position, and that nil sizes fall back to the documented
defaults (400px primary, full screen dimension for secondary).

Part of https://github.com/ghostty-org/ghostty/issues/7879.


mitchellh:
!vouch

PR #11117, opened 2026-03-02
gtk: implement quick-terminal-screen for Linux/Wayland

jguthmiller:
Discussed in https://github.com/ghostty-org/ghostty/discussions/10615

Implements quick-terminal-screen for Linux/Wayland so the quick terminal can be pinned to a specific monitor. Uses kde-output-order-v1 to discover the compositor's primary output, then matches its connector name against GDK's monitor list. Falls back to the first GDK monitor when the protocol isn't available (e.g. GNOME, sway). macos-menu-bar is treated as equivalent to main.

AI-assisted (Claude Code and Codex)

### DIFF
--- a/pkg/gtk4-layer-shell/src/main.zig
+++ b/pkg/gtk4-layer-shell/src/main.zig
@@ -63,7 +63,7 @@ pub fn setKeyboardMode(window: *gtk.Window, mode: KeyboardMode) void {
 }
 
 pub fn setMonitor(window: *gtk.Window, monitor: ?*gdk.Monitor) void {
-    c.gtk_layer_set_monitor(@ptrCast(window), if (monitor) |m| @ptrCast(m) else null);
+    c.gtk_layer_set_monitor(@ptrCast(window), @ptrCast(monitor));
 }
 
 pub fn setNamespace(window: *gtk.Window, name: [:0]const u8) void {
--- a/pkg/gtk4-layer-shell/src/main.zig
+++ b/pkg/gtk4-layer-shell/src/main.zig
@@ -3,6 +3,7 @@ const std = @import("std");
 const c = @cImport({
     @cInclude("gtk4-layer-shell.h");
 });
+const gdk = @import("gdk");
 const gtk = @import("gtk");
 
 pub const ShellLayer = enum(c_uint) {
@@ -59,6 +60,10 @@ pub fn setMargin(window: *gtk.Window, edge: ShellEdge, margin_size: c_int) void 
 
 pub fn setKeyboardMode(window: *gtk.Window, mode: KeyboardMode) void {
     c.gtk_layer_set_keyboard_mode(@ptrCast(window), @intFromEnum(mode));
+}
+
+pub fn setMonitor(window: *gtk.Window, monitor: ?*gdk.Monitor) void {
+    c.gtk_layer_set_monitor(@ptrCast(window), if (monitor) |m| @ptrCast(m) else null);
 }
 
 pub fn setNamespace(window: *gtk.Window, name: [:0]const u8) void {
--- a/src/apprt/gtk/winproto/wayland.zig
+++ b/src/apprt/gtk/winproto/wayland.zig
@@ -45,10 +45,6 @@ pub const App = struct {
         /// is the primary.
         primary_output_name: ?[max_output_name_len:0]u8 = null,
 
-        /// Used to avoid repeatedly logging the same primary-name mismatch
-        /// when we can't map the compositor connector name to a GDK monitor.
-        primary_output_match_failed_logged: bool = false,
-
         /// Tracks the output order event cycle. Set to true after a `done`
         /// event so the next `output` event is captured as the new primary.
         /// Initialized to true so the first event after binding is captured.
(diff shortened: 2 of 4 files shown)


pluiedev:
Vouched in #11118

pkg/gtk4-layer-shell/src/main.zig
```diff
@@ -61,6 +62,10 @@ pub fn setKeyboardMode(window: *gtk.Window, mode: KeyboardMode) void {
     c.gtk_layer_set_keyboard_mode(@ptrCast(window), @intFromEnum(mode));
 }
 
+pub fn setMonitor(window: *gtk.Window, monitor: ?*gdk.Monitor) void {
+    c.gtk_layer_set_monitor(@ptrCast(window), if (monitor) |m| @ptrCast(m) else null);
```
pluiedev: Optional pointers are actually allowed AFAICT, so this is more simply written as

```suggestion
    c.gtk_layer_set_monitor(@ptrCast(window), @ptrCast(m));
```

src/apprt/gtk/winproto/wayland.zig
```diff
@@ -127,9 +149,67 @@ pub const App = struct {
...
+            const name = std.mem.sliceTo(stored_name, 0);
+            var i: u32 = 0;
+            while (monitors.getObject(i)) |item| : (i += 1) {
+                const monitor = gobject.ext.cast(gdk.Monitor, item) orelse {
+                    item.unref();
+                    continue;
+                };
+                if (monitor.getConnector()) |connector_z| {
+                    const connector = std.mem.sliceTo(connector_z, 0);
+                    if (std.mem.eql(u8, connector, name)) {
```
pluiedev: Use [`std.mem.orderZ`](https://ziglang.org/documentation/0.15.2/std/#std.mem.orderZ)

src/apprt/gtk/winproto/wayland.zig
```diff
@@ -171,15 +270,37 @@ pub const App = struct {
...
+                            } else {
+                                log.warn(
+                                    "duplicate global for {s}; keeping existing binding",
+                                    .{v.interface},
+                                );
+                                break;
+                            }
+                        }
+
+                        const global = registry.bind(
```
pluiedev: I'm not sure what this is about. Could you elaborate on this?

src/apprt/gtk/winproto/wayland.zig
```diff
@@ -221,6 +391,45 @@ pub const App = struct {
...
+                context.primary_output_match_failed_logged = false;
+
+                const name = std.mem.sliceTo(v.output_name, 0);
+                if (name.len == 0) {
+                    context.primary_output_name = null;
+                    log.warn("ignoring empty primary output name from kde_output_order_v1", .{});
+                } else if (name.len <= Context.max_output_name_len) {
+                    var buf: [Context.max_output_name_len:0]u8 = @splat(0);
+                    @memcpy(buf[0..name.len], name);
+                    context.primary_output_name = buf;
```
pluiedev: No. This is a stale pointer that points to stack memory that will be destroyed after this function exits. Why not just heap-allocate it?

pluiedev (changes requested):

src/apprt/gtk/winproto/wayland.zig
jguthmiller: I added a comment to provide some context.

src/apprt/gtk/winproto/wayland.zig
jguthmiller: Is it? primary_output_name is a fixed size array. Isn't this assignment a value copy?

pluiedev: Oh, my bad. Still, the way this is written is very confusing and could be footgun-y - just do a heap allocation

src/apprt/gtk/winproto/wayland.zig
pluiedev: Hmm, but is there really a point to distinguishing the two different situations when globals are replaced? Like in this case I think it makes more sense to just generalize this logic to destroy the old global and reset the output state unconditionally, given that this isn't something that should occur often.

src/apprt/gtk/winproto/wayland.zig
```diff
@@ -127,9 +149,65 @@ pub const App = struct {
...
+                    if (std.mem.orderZ(u8, connector_z, stored_name) == .eq) {
+                        context.primary_output_match_failed_logged = false;
+                        return monitor;
+                    }
+                }
+                monitor.unref();
+            }
+
+            if (!context.primary_output_match_failed_logged) {
+                context.primary_output_match_failed_logged = true;
```
pluiedev: Does this variable really need to live inside `Context`?

src/apprt/gtk/winproto/wayland.zig
jguthmiller: No, good call. This was mostly a result of me trying to be overly thorough. It's much simpler now.

src/apprt/gtk/winproto/wayland.zig
jguthmiller: Addressed in the most recent commit.

src/apprt/gtk/winproto/wayland.zig
jguthmiller: Definitely not. Removed

pluiedev (approved):
The current architecture for globals isn't *great* and I think I do need to fix it at some point. I think this is good enough for my personal taste. Thanks!

PR #11119, opened 2026-03-02
terminal: fix integrity violation printing wide char with hyperlink at right edge

mitchellh:
Printing a wide character at the right edge of the screen with an active hyperlink triggered a page integrity violation (UnwrappedSpacerHead). printCell wrote the spacer_head to the cell and then called cursorSetHyperlink, whose internal integrity check observed the spacer_head before printWrap had a chance to set the row wrap flag.

Fix by setting row.wrap = true before calling printCell for the spacer_head case, so all integrity checks see a consistent state. printWrap sets wrap again afterward, which is harmless. Found by AFL++ stream fuzzer.

#11109


jacobsandlund (commented):
lgtm!

PR #11144, opened 2026-03-03
terminfo: add support for SGR dim

noib3:
This PR implements the fix discussed in https://github.com/ghostty-org/ghostty/discussions/11128.

Before:

before

After:

after


gpanders (approved):
LGTM, I double checked in the terminfo man page that `dim` uses parameter 5 (`p5`)

```
tparm Parameter   Attribute    Escape Sequence
────────────────────────────────────────────────
none              none         \E[0m
p1                standout     \E[0;1;7m
p2                underline    \E[0;4m
p3                reverse      \E[0;7m
p4                blink        \E[0;5m
p5                dim          not available
p6                bold         \E[0;1m
p7                invis        \E[0;8m
p8                protect      not used
p9                altcharset   ^O (off) ^N (on)
```

The man page says that there is no escape sequence for dim, but Xterm lists it as one of the supported SGR sequences in https://invisible-island.net/xterm/ctlseqs/ctlseqs.html (and of course Ghostty supports it, which is most relevant here).

mitchellh:
Thanks everyone!

PR #11154, opened 2026-03-03
macos: implement audio bell support with bell-audio-path

alaasdk:
## Summary

This extends the macOS bell implementation to support the `audio` bell feature, bringing it to parity with GTK/Linux.

Previously, macOS only had the `system` feature (`NSSound.beep()`). This PR adds:

- **`audio` bell feature on macOS**: plays the file at `bell-audio-path` using `NSSound(contentsOfFile:)`, respecting `bell-audio-volume`
- **`cval()` on the `Path` type**: allows `Path` values (a union type) to be returned through the C API, which is needed for Swift to read `bell-audio-path`
- **Removes `(GTK only)` restriction** from `bell-audio-path` and `bell-audio-volume` documentation

## How it works

In `AppDelegate.swift`, when the bell rings and the `audio` feature is enabled, Ghostty now:
1. Reads `bell-audio-path` from config
2. Loads it as an `NSSound`
3. Applies `bell-audio-volume` and plays it

Falls back gracefully if the path is not set or the file cannot be loaded.

## Example config

```
bell-features = audio
bell-audio-path = /System/Library/Sounds/Glass.aiff
bell-audio-volume = 0.8
```

## Testing

- Set `bell-features = audio` and `bell-audio-path` to any valid audio file
- Trigger a bell with `echo -e '\a'`
- Audio should play at the configured volume

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -463,6 +463,12 @@ typedef struct {
 
 // Config types
 
+// config.Path
+typedef struct {
+  const char* path;
+  bool optional;
+} ghostty_config_path_s;
+
 // config.Color
 typedef struct {
   uint8_t r;
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -778,8 +778,8 @@ class AppDelegate: NSObject,
         }
 
         if ghostty.config.bellFeatures.contains(.audio) {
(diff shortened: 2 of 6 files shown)


mitchellh:
!vouch

mitchellh (approved):
Awesome thank you this looks great. I just modified config.Path to return a proper C structure. The rest looks good.

PR #11166, opened 2026-03-04
zsh: fix extra newlines with leading-newline prompts

jparise:
In our multiline prompt logic, skip the newline immediately after the first mark to avoid introducing a double newline due to OSC 133;A's fresh-line behavior.

Fixes: #11003


mitchellh:
Thanks For both!

PR #11167, opened 2026-03-04
apprt: unify split-click focus behavior across macOS and GTK; suppress focus-transfer mouse events

rockorager:
## Summary

This PR aligns split-pane click behavior across macOS and GTK when focus changes due to click.

When a left-click is used to transfer focus (window activation or switching to another split), Ghostty now treats that click as focus-only and suppresses forwarding mouse press/release events for that focus-transfer click.

## Changes

1. macOS: suppress focus-transfer left mouse-down and matching mouse-up in `SurfaceView_AppKit.swift`.
1. GTK: suppress focus-transfer left mouse-down and matching mouse-up in `src/apprt/gtk/class/surface.zig`.
1. macOS: defer key-window focus sync to next runloop tick to reduce transient focus churn in `BaseTerminalController.swift`.
1. macOS build/lint: exclude generated/dependency paths from SwiftLint during build in `.swiftlint.yml` and `Ghostty.xcodeproj/project.pbxproj`.

## Behavior

1. Focus-transfer split clicks are now focus-only on both macOS and GTK.
1. Matching release is also suppressed for those clicks, avoiding release-without-press sequences.
1. Platform behavior is consistent for split focus transitions.

## Validation

1. Built macOS target with `xcodebuild -target Ghostty -configuration Debug -arch arm64`.
1. Ran targeted Zig test command `zig build test -Dtest-filter=computeFraction`.
1. Ran format/lint for touched files (`swiftlint lint --fix`, `zig fmt`).
4. Build and (human) tested click scenarios on macOS

## AI Disclosure

AI-assisted.

Thread: https://ampcode.com/threads/T-019cb9fe-b11b-753f-99e7-8ecc52b73ec4

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -1239,9 +1239,11 @@ class BaseTerminalController: NSWindowController,
             }
         }
 
-        // Becoming/losing key means we have to notify our surface(s) that we have focus
-        // so things like cursors blink, pty events are sent, etc.
-        self.syncFocusToSurfaceTree()
+        // Becoming key can race with responder updates when activating a window.
+        // Sync on the next runloop so split focus has settled first.
+        DispatchQueue.main.async {
+            self.syncFocusToSurfaceTree()
+        }
     }
 
     func windowDidResignKey(_ notification: Notification) {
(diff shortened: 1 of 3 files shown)


jcollie (approved):
GTK changes LGTM

mitchellh (approved):
Looks fine to me, I think worst case if there are bugs a mouse event or two gets missed but if it works t works.

PR #11170, opened 2026-03-04
apprt/gtk: reduce split-tree flicker by reusing leaf widgets

### LINKED ISSUE #8208: gtk-ng: fix split change (new, delete, resize, etc.) "flicker"
Right now, new/close split cause a noticeable flicker, see #8202 for a video.

This is because we have to wait 1 or more event loop ticks for the surface widgets to unparent before building the new GtkPaned tree. We should be able to do this all in one frame, but every time I tried I hit rendering issues. I suspect I'm just messing things up with my [lack of] GTK knowledge.

mitchellh:
Fixes #8208

Split-tree updates currently clear `tree_bin` and then wait for every surface to become parentless before rebuilding. That leaves the split area blank for one or more frames, which is the visible flicker during split create/close/ resize/equalize actions.

Keep the previous widget tree attached until the idle rebuild runs, then swap in the rebuilt tree in one step. During rebuild, reuse existing leaf widgets by detaching and reparenting them into the new `GtkPaned`  hierarchy instead of recreating wrappers for every leaf.

This removes the parent-settling rebuild path and avoids transient blank frames while preserving debounced rebuild behavior.

### DIFF
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -157,11 +157,6 @@ pub const SplitTree = extern struct {
         /// used to debounce updates.
         rebuild_source: ?c_uint = null,
 
-        /// Tracks whether we want a rebuild to happen at the next tick
-        /// that our surface tree has no surfaces with parents. See the
-        /// propTree function for a lot more details.
-        rebuild_pending: bool,
-
         /// Used to store state about a pending surface close for the
         /// close dialog.
         pending_close: ?Surface.Tree.Node.Handle,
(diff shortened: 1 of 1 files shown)


jcollie (approved):

mitchellh:
Hold up we have some assertion failures on graceful exit. Some resource isn't cleaned up properly.

mitchellh:
Fixed the issue and removed all allocations, so this is an error-free path now (except for GTK failures).

jcollie (approved):

mitchellh:
https://ampcode.com/threads/T-019cbadf-cb5a-742e-b0e3-2d7164de743f

PR #11172, opened 2026-03-04
Zig 0.16 prep: "juicy" main environment variables

jcollie:
Zig 0.16 will eliminate the ability to directly access environment variables (using `std.posix.getenv` or related APIs). Instead environment variables will be passed to a program's `main` function if it has the signature `pub fn main(init: std.process.Init)`. The environment variables will need to be cached somewhere accessible to the rest of Ghostty.

This adds the environment variables to Ghostty's `global.state` variable and then provides convenience APIs for accessing the enviroment variables that are similar to the APIs previously available.

Similar changes are made in the build system - the environment variables are available in the `b.graph.env_map` variable.

See:

https://github.com/ziglang/zig/issues/24510
https://codeberg.org/ziglang/zig/pulls/30644

### DIFF
--- a/pkg/android-ndk/build.zig
+++ b/pkg/android-ndk/build.zig
@@ -118,7 +118,7 @@ pub fn addPaths(b: *std.Build, step: *std.Build.Step.Compile) !void {
 
 fn findNDKPath(b: *std.Build) ?[]const u8 {
     // Check if user has set the environment variable for the NDK path.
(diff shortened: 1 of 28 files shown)


jcollie:
Leaving as draft for now as this needs testing on macOS and embedded use cases.

mitchellh:
Closing in favor of #12726

PR #11180, opened 2026-03-05
Fix Windows test in src/Command.zig

jcollie:
This was introduced in #10611. This doesn't fix all of the current Windows build problems, but at least fixes one that I introduced.


mitchellh (approved):
Changing tests are low risk for release. Feel free to merge.

PR #11185, opened 2026-03-05
fix: zsh shell integration when `sudo` and `ssh` aliases are defined

Michielvk:
I encountered an issue related to https://github.com/ghostty-org/ghostty/discussions/8641 and https://github.com/ghostty-org/ghostty/pull/8647, but in `zsh` instead of `bash`.

One of my aliases is:

```bash
alias sudo='sudo '
```

Which causes following error when sourcing the zsh shell integrations:

```shell
source /usr/share/ghostty/shell-integration/zsh/ghostty-integration
/usr/share/ghostty/shell-integration/zsh/ghostty-integration:149: defining function based on alias `sudo'
/usr/share/ghostty/shell-integration/zsh/ghostty-integration:233: parse error near `()'
```


jparise (approved):
Thanks!

PR #11190, opened 2026-03-05
build,ci: add terminfo artifacts for releases and refactor terminfo build

AnthonyZhOon:
Closes #9917
# Summary
- refactor terminfo generation/DB compilation out of GhosttyResources internals into dedicated GhosttyTerminfo file
- add `zig build terminfo` step to only output terminfo and termcap source files
- publish ghostty.terminfo in tagged and tip release workflows
- document how to build/fetch and install terminfo in README

# AI Disclosure
gpt-5.3-codex in Codex cli was used to iterate on an approach and reviewed at each step. All code is understood.

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -102,10 +102,8 @@ jobs:
         run: |
           rm -rf zig-out/dist
           nix develop -c zig build distcheck
-          nix develop -c zig build terminfo
           cp zig-out/dist/ghostty-${GHOSTTY_VERSION}.tar.gz .
           cp zig-out/dist/ghostty-${GHOSTTY_VERSION}.tar.gz ghostty-source.tar.gz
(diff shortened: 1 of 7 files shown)


00-kat:
!vouch

.github/workflows/release-tag.yml
```diff
@@ -102,8 +102,10 @@ jobs:
         run: |
           rm -rf zig-out/dist
           nix develop -c zig build distcheck
+          nix develop -c zig build terminfo
```
mitchellh: I don't think terminfo building should be part of the `source-tarball` job. I think it should probably be its own job.

mitchellh (changes requested):
This looks mostly good.

We should add a `build-terminfo` xsm CI job too similar to build-dist to verify it.

.github/workflows/release-tag.yml
AnthonyZhOon: done

PR #10992, opened 2026-02-24
macOS: Implement additional accessibility methods

MrMage:
Fixes #9932; also see https://github.com/ghostty-org/ghostty/discussions/9930.

**Submitting this as a draft PR to get feedback on the areas listed below. Please let me know whether this is going in the right direction.**

- The code was largely written by Claude Code, with guidance from me. I have read through the implementation and reviewed it; overall, the code looks reasonable to me. (I have of course also checked that it actually works; see below.) I have also improved the factoring and removed some unnecessary duplication etc., especially on the Swift side.
  - I have read the contribution guidelines and AI policy, but it still leaves room for interpretation (e.g. with regards to how many comments are "too much" etc., see below); happy to make adjustments to bring this further in line with your expectations. Again, this is just a rough first draft.
  - I have intentionally left in most comments (and asked Claude to focus on the _motivation_ of the code); happy to clean them up further if you find them too verbose.
- Caveat: I have hardly any experience with Zig, so I may have missed something in that implementation. In particular, my knowledge of PinMap is limited, so please feel free to provide some guidance there (see below).
  - Speaking of PinMap, I am unsure how inefficient the current implementation is in practice. But even if it is somewhat inefficient, it could serve as a starting point for a more efficient implementation as needed in practice.
  - To my understanding, we could also replace `cachedScreenContents` and `cachedVisibleContents` with accesses into the `cachedScreenTextInfo`; I also don't think they are accessed much more frequently than the `cachedScreenTextInfo`, so any possible performance overhead of `cachedScreenTextInfo` would probably be tolerable. Happy to make these changes, let me know.
  - Speaking of `cachedVisibleContents`, I am unsure why reading the visible text with `ghostty_surface_read_text` would be less expensive (potentially by not requiring a PinMap, but how?) than getting byte offsets into the entire text using a PinMap.
  - As an alternative, maybe we could include a pointer to a PinMap itself in the `cachedScreenTextInfo` struct and cache that, to minimize the frequency of recreating PinMaps for Accessibility methods? That could be worthwhile to also optimize the other AX methods (e.g. for getting the rectangles of certain characters etc.). We would probably need to manage the PinMap's lifetime in Swift, but I guess that's doable.
  - @mitchellh your feedback on how we should best architect this aspect, both from a caching and a Swift/Zig interop perspective, would be greatly appreciated.
- Tests are only on the Zig layer right now; having good tests on the Swift layer might be a bit more difficult because we'd need to test them in the UI harness (haven't looked too much into that yet).
[...]

Issue #9932, opened 2025-12-16
Accessibility: AXVisibleCharacterRange returns wrong data; AXRangeForPosition, AXBoundsForRange missed


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9930

<div type='discussions-op-text'>

<sup>Originally posted by **grishy** December 16, 2025</sup>
### Issue Description

Ghostty’s accessibility implementation is missing attributes needed by assistive SW and third-party tools to query text at the cursor position, for example and more problematic, implement some API in wrong way.

I found this during debugging of my translation app when clicked on Ghostty app, which works across the system and has text access and fast tracking. Usually, it takes 10–20 ms in other apps, like Chrome or others, but here it took 3 seconds.
  
I used `log show --last 1m` to get a lot of text below in tests
This prevents screen readers and accessibility tools from efficiently working with Ghostty. 

The main issue is that it returns the whole terminal blob in both cases: `AXVisibleCharacterRange` and `AXValue` .
Also, `AXValue` would be better if it did not return the full terminal context, if possible, only from the latest command run. For example, browsers return just a paragraph, not a whole page also for `AXValue`.

NOTE: For debug I suggest to use macOS buildin tool - https://developer.apple.com/documentation/accessibility/accessibility-inspector

### Similar ticket 
https://github.com/ghostty-org/ghostty/discussions/2351
https://github.com/ghostty-org/ghostty/issues/4053
https://github.com/ghostty-org/ghostty/discussions/3852

### Comparison Behavior with Terminal.app

  | Attribute | Terminal.app | Ghostty |
  |-----------|-------------|---------|
  | AXVisibleCharacterRange | `loc=3037623 len=12387` ✅ | `loc=0 len=709778` ❌ (full buffer) |
  | AXRangeForPosition | `loc=3041844 len=1` ✅ | N/A ❌ |
  | AXBoundsForRange | `(686,409) 7x14` ✅ | N/A ❌ |

Terminal.app with 3M chars buffer correctly returns only visible viewport (~12K chars).
Ghostty with 700K chars buffer returns entire scrollback as "visible".

### Expected Behavior

1. `AXRangeForPosition` should return the character index at a given screen coordinate
2. `AXBoundsForRange` should return screen bounds (CGRect) for a character range
3. `AXVisibleCharacterRange` should return only the visible viewport range, not the entire scrollback buffer

```
Cursor at: (691, 417)
Querying accessibility...

[...]

PR #11196, opened 2026-03-05
macOS: Introduce AccessibilityContext, improve existing AX APIs, add `accessibilityRange(forLine:)`

MrMage:
This is the first in a series of PRs improving terminal accessibility on macOS, effectively splitting #10992 over multiple, smaller PRs. See #9930 and #9932 for motivation.

It introduces an `AccessibilityContext` type on the Zig side and uses it to back the existing accessibility APIs in `SurfaceView`. That in itself is not valuable yet (and in fact, the `AccessibilityContext` generation via PinMap is arguably _more_ complex than the previous approach), but it's a necessary foundation for future improvements (again, see #10992 for what's to come). Large portions of the Accessibility API (including `accessibilityVisibleCharacterRange()`) will require us to build the PinMap anyway, and with this approach we can at least centralize that logic and data in one place.

However, this PR also provides some immediate value: it provides a proper implementation of `accessibilityVisibleCharacterRange()` and fixes two UTF-16 vs. UTF-8 mismatch issues in `accessibilityNumberOfCharacters()` and `accessibilityLine(for:)`. It also introduces the inverse of that method, `accessibilityRange(forLine:)`.

Also see #10992 for a small helper tool that one can use to probe the accessibility APIs for verification.

## Zig core: `AccessibilityContext` (`Screen.zig`, `Surface.zig`)

A new `AccessibilityContext` struct holds the full terminal text (scrollback + active area) along with the byte offsets delimiting the visible viewport. It is built from a PinMap — the same mechanism used by `selectionString` — to accurately determine where the viewport sits within the full text. The PinMap is necessary because `accessibilityVisibleCharacterRange()` needs to report the viewport as a character range *within* the full text; the previous implementation had no way to do this and simply returned the range of the entire text. Without the PinMap, the alternatives would be fragile — fetching the viewport text separately and searching for it within the full text could match the wrong occurrence, and making two separate Zig calls leaves no reliable way to correlate their offsets. The PinMap maps every byte in the output text back to its terminal Pin, so we can precisely identify where the viewport starts and ends. The context owns its data (no Pins or page pointers) so it can be used after releasing the terminal mutex.

### Swift: unified cache and AX method fixes (`SurfaceView_AppKit.swift`)

The two separate `CachedValue<String>` properties (`cachedScreenContents`, `cachedVisibleContents`) are replaced by a single `CachedValue<ScreenTextInfo>`. The `ScreenTextInfo` struct holds the full text and the viewport range as a `Range<String.Index>`. `accessibilityVisibleCharacterRange()` converts this range to `NSRange` only when needed.

### Known limitations

`accessibilityLine(for:)` and `accessibilityRange(forLine:)` count `\n`-delimited lines, not visual screen rows. Soft-wrapped lines are treated as a single line. Accurate visual-line support requires the grid mapping infrastructure planned for a follow-up PR.

## Outlook

Subsequent PRs will build on `AccessibilityContext` to add:
- Grid mapping tables (`viewport_cells`, `grid_to_offset`) for `accessibilityFrame(for:)` and `accessibilityRange(for:)` (bounds/position queries).
- A `content_changed` notification pipeline so VoiceOver stays in sync with terminal output without polling — gated on whether an accessibility client is actually connected, to avoid unnecessary work.
- (Maybe) Cursor offset tracking for `accessibilityInsertionPointLineNumber()` -- Depends on whether this is actually valuable, and/or whether we would need to change the semantics of `accessibilityLine(for:)` and `accessibilityRange(forLine:)` to refer to newlines _as rendered_ rather than _as line break characters_.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -387,6 +387,13 @@ typedef struct {
   uintptr_t text_len;
 } ghostty_text_s;
 
+typedef struct {
+  const char* text;
+  uintptr_t text_len;
+  uintptr_t viewport_start;
+  uintptr_t viewport_end;
+} ghostty_ax_text_s;
+
 typedef enum {
   GHOSTTY_POINT_ACTIVE,
   GHOSTTY_POINT_VIEWPORT,
@@ -1129,6 +1136,11 @@ bool ghostty_surface_read_text(ghostty_surface_t,
                                ghostty_selection_s,
                                ghostty_text_s*);
 void ghostty_surface_free_text(ghostty_surface_t, ghostty_text_s*);
+
+typedef void* ghostty_ax_context_t;
+ghostty_ax_context_t ghostty_surface_ax_context_new(ghostty_surface_t);
+void ghostty_surface_ax_context_free(ghostty_ax_context_t);
+bool ghostty_ax_context_info(ghostty_ax_context_t, ghostty_ax_text_s*);
 
 #ifdef __APPLE__
 void ghostty_surface_set_display_id(ghostty_surface_t, uint32_t);
--- a/macos/Sources/Features/App Intents/GetTerminalDetailsIntent.swift
+++ b/macos/Sources/Features/App Intents/GetTerminalDetailsIntent.swift
@@ -37,13 +37,13 @@ struct GetTerminalDetailsIntent: AppIntent {
         case .workingDirectory: return .result(value: terminal.workingDirectory)
         case .allContents:
             guard let view = terminal.surfaceView else { throw GhosttyIntentError.surfaceNotFound }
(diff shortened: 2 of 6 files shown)


MrMage:
@mitchellh Friendly ping :-)

symphonix84:
Thanks for all the hard work @MrMage

MrMage:
Ping :-)

jcollie:
> Ping :-)

Do _not_ ping maintainers hoping to expedite review. It won't happen any faster, and just serves to annoy people - more than you think since many people receive "full feeds" of the activity on the Ghostty repo.

PR #11208, opened 2026-03-06
AppleScript

mitchellh:
This adds AppleScript support to the macOS app.

AppleScript is still one of the best ways to script macOS apps. It is more CLI friendly and share-able than Apple Shortcuts and can be used by other CLI programs like editors (Neovim plugins), launchers (Raycast/Alfred), etc. It has been heavily requested to introduce more scriptability into Ghostty and this is a really good, powerful option on macOS.

> [!NOTE]
>
> I definitely still want to do something cross-platform and more official as a plugin/scripting API for Ghostty. But native integrations like this are a goal of Ghostty as well and this implementation is just some thin logic over already existing internals to expose it.

I plan on merging this ahead of 1.3. Normally I wouldn't ship a feature so late in the game but this is fairly hermetic (doesn't impact other systems) and I plan on documenting it as a "preview" feature since the API and stability are in question.

## Security

Apple secures AppleScript via TCC by asking for permission when a script is run whether an app is allowed to be controlled. Because this is always asked, we do default AppleScript to being enabled. This is typical of macOS native applications already.

AppleScript can be wholesale disabled via `macos-applescript = false`.

## Future

There is a big question of what else to expose to this to make it useful. I'm going to make a call to action for the 1.3 cycle to gather feedback on this, since we can expose mostly anything!

### Objects

| Object | Key Properties | Key Elements |
| --- | --- | --- |
| `application` | `name`, `frontmost`, `version` | `windows`, `terminals` |
| `window` | `id`, `name`, `selected tab` | `tabs`, `terminals` |
| `tab` | `id`, `name`, `index`, `selected` | `terminals` |
| `terminal` | `id`, `name`, `working directory` | None |

### Commands

| Category | Command | Purpose |
| --- | --- | --- |
| Application | `perform action` | Execute a Ghostty action string on a terminal. |
| Configuration | `new surface configuration` | Create/copy a reusable surface configuration record. |
| Creation | `new window` | Open a new Ghostty window (optional configuration). |
| Creation | `new tab` | Open a new tab (optional target window/configuration). |
| Layout | `split` | Split a terminal and return the new terminal. |
| Focus/Selection | `focus` | Focus a terminal. |
| Focus/Selection | `activate window` | Bring a window to front and activate app. |
| Focus/Selection | `select tab` | Select and foreground a tab. |
| Lifecycle | `close` | Close a terminal. |
| Lifecycle | `close tab` | Close a tab. |
| Lifecycle | `close window` | Close a window. |
| Input | `input text` | Paste-style text input into terminal. |
| Input | `send key` | Send key press/release with optional modifiers. |
| Input | `send mouse button` | Send mouse button press/release. |
| Input | `send mouse position` | Send mouse position update. |
| Input | `send mouse scroll` | Send scroll event with precision/momentum options. |
| Standard Suite | `count`, `exists`, `quit` | Standard Cocoa scripting functionality. |

### Layout

```AppleScript
-- Tmux-like layout: 4 panes in one tab (2x2), each with a job.
set projectDir to POSIX path of (path to home folder) & "src/ghostty"

tell application "Ghostty"
    activate

    -- Reusable config for all panes.
    set cfg to new surface configuration
    set initial working directory of cfg to projectDir

    -- Create the first window/tab + split into 4 panes.
    set win to new window with configuration cfg
    set paneEditor to terminal 1 of selected tab of win
    set paneBuild to split paneEditor direction right with configuration cfg
    set paneGit to split paneEditor direction down with configuration cfg
    set paneLogs to split paneBuild direction down with configuration cfg

    -- Seed each pane with a command.
    input text "nvim ." to paneEditor
    send key "enter" to paneEditor

    input text "zig build -Demit-macos-app=false" to paneBuild

    input text "git status -sb" to paneGit

    input text "tail -f /tmp/dev.log" to paneLogs
    send key "enter" to paneLogs

    -- Put focus back where you want to type.
    focus paneEditor
end tell
```

### Broadcast Commands

```AppleScript
-- Run one command across every open terminal surface.
set cmd to "echo sync && date"

tell application "Ghostty"
    set allTerms to terminals

    repeat with t in allTerms
        input text cmd to t
        send key "enter" to t
    end repeat

    display dialog ("Broadcasted to " & (count of allTerms) & " terminal(s).")
end tell
```

### Jump by Working Directory

```applescript
-- Find the first terminal whose cwd contains this text.
set needle to "ghostty"

tell application "Ghostty"
    set matches to every terminal whose working directory contains needle

    -- Fallback: try title if cwd had no match.
    if (count of matches) = 0 then
        set matches to every terminal whose name contains needle
    end if

    if (count of matches) = 0 then
        display dialog ("No terminal matched: " & needle)
    else
        set t to item 1 of matches
        focus terminal t
        input text "echo '[focused by AppleScript]'" to t
        send key "enter" to t
    end if
end tell
```


macos/Sources/Features/AppleScript/AppDelegate+AppleScript.swift
```diff
@@ -0,0 +1,329 @@
...
+    var isAppleScriptEnabled: Bool {
+        guard let appDelegate = delegate as? AppDelegate else { return true }
+        return appDelegate.ghostty.config.macosAppleScript
+    }
+
+    /// Applies a consistent error when scripting is disabled by configuration.
+    @discardableResult
+    func validateScript(command: NSScriptCommand) -> Bool {
+        guard isAppleScriptEnabled else {
+            command.scriptErrorNumber = errAEEventFailed
```
nmggithub:
```suggestion
            command.scriptErrorNumber = errAEEventNotPermitted
```

I think `errAEEventNotPermitted` is a more canonical error code to use here.

macos/Sources/Features/AppleScript/AppDelegate+AppleScript.swift
```diff
@@ -0,0 +1,329 @@
...
+            return nil
+        }
+
+        guard let terminal = command.evaluatedArguments?["on"] as? ScriptTerminal else {
+            command.scriptErrorNumber = errAEParamMissed
+            command.scriptErrorString = "Missing terminal target."
+            return nil
+        }
+
+        return NSNumber(value: terminal.perform(action: action))
```
nmggithub, line 133: Do we need to wrap the boolean in an `NSNumber`? Also, why is this function typed as a return of `Any?`?

macos/Sources/Features/AppleScript/AppDelegate+AppleScript.swift
```diff
@@ -0,0 +1,329 @@
...
+            return nil
+        }
+
+        let baseConfig: Ghostty.SurfaceConfiguration
+        do {
+            if let scriptRecord = command.evaluatedArguments?["configuration"] as? NSDictionary {
+                baseConfig = try Ghostty.SurfaceConfiguration(scriptRecord: scriptRecord)
+            } else {
+                baseConfig = Ghostty.SurfaceConfiguration()
+            }
```
nmggithub:
```suggestion
            }
```

Since [`TerminalController.newTab` accepts a `nil` config,](https://github.com/ghostty-org/ghostty/blob/applescript/macos/Sources/Features/Terminal/TerminalController.swift#L376) I don't think we need to instantiate such a config here. We don't do so above in `handleNewWindowScriptCommand`. We should keep the same behavior to match.

macos/Ghostty.sdef
```diff
@@ -0,0 +1,312 @@
...
+        <cocoa method="handleNewWindowScriptCommand:"/>
+      </responds-to>
+      <responds-to command="new tab">
+        <cocoa method="handleNewTabScriptCommand:"/>
+      </responds-to>
+      <responds-to command="new surface configuration">
+        <cocoa method="handleNewSurfaceConfigurationScriptCommand:"/>
+      </responds-to>
+      <responds-to command="quit">
+        <cocoa method="handleQuitScriptCommand:"/>
```
nmggithub, line 26: Where's the implementation to this? I don't see it anywhere.

macos/Ghostty.sdef
```diff
@@ -0,0 +1,312 @@
...
+        <cocoa key="tab"/>
+      </parameter>
+    </command>
+
+    <command name="close window" code="GhstClWn" description="Close a window.">
+      <cocoa class="GhosttyScriptCloseWindowCommand"/>
+      <parameter name="window" code="GCwW" type="window" description="The window to close.">
+        <cocoa key="window"/>
+      </parameter>
+    </command>
```
nmggithub, line 211: These can lead to really awkward AppleScript, like the below:
```applescript
tell application "Ghostty" to activate window window (window 1)
```

Not sure if that's possible while juggling the above terminology and built-in terminology. The items to be acted upon should, at least, be direct parameters.

nmggithub (commented):
Really nice. Only a few things I would personally change:
1. define explicit return value types for the script handler methods (if possible),
2. define better terminology for non-terminal (window / tab) handling commands, and
3. use `errAEEventNotPermitted` inside `validateScript` instead of `errAEEventFailed`.

macos/Sources/Features/AppleScript/AppDelegate+AppleScript.swift
mitchellh: Oh yeah, a lot of this was whack-a-mole since the resources around it were difficult to find, so I started with Any and worked my way backwards.

macos/Ghostty.sdef
mitchellh: I cleaned this up:

```
tell application "Ghostty"
    activate window (window 1)
    close window (window 1)
    close tab (tab 1 of window 1)
    select tab (tab 1 of window 1)
    focus (terminal 1)
    close (terminal 1)
    split (terminal 1) direction right
end tell
```

gurgeous:
Thank you for adding. I am using this now to create new titled tabs with a particular title from the cli. This is embedded in a larger bash script:

```applescript
on run argv
  set title to item 1 of argv
  tell application "Ghostty"
    set term to terminal 1 of (new tab in front window)
    -- note space to avoid history
    input text " printf \"\\e]0;%s\\a\" \"" & title & "\"; clear" to term
    send key "enter" to term
  end tell
end run
```

Looking forward to trying out #11316 as well. Minor feedback from my initial explorations:

- some discussion about window vs. pane vs. tab vs. terminal in the applescript docs
- some richer examples (that create windows/panes/tabs/terminals) would be nice
- easier syntax for sending enter

I could also imagine combining with `printf '\e]21;background=#200;\a'` for deeper customization. Wait until alfred, btt and hammerspoon users start messing with this stuff...

mitchellh:
I'm usually annoyed when people comment on merged PRs but your comment is actually good. So, I'm mildly annoyed but thank you lol.

gurgeous:
This will be the last time I comment on a merged PR... Apologies. There is a lot of action in the repo and it isn't always obvious where to chime in with this stuff. I will stick to discussions. Time to bring in OpenClaw to gently spank well meaning users

cmaciasjimenez:
Hi! you mention in your Twitter that with more code we could do different layouts for the terminal windows/tabs/splits. Is this possible with the current version of this feature?

I'm trying to find a way to run a script to set up new tabs but the best I've gotten to work is to have a new tab/window with a vertical split but I would like to have that vertical split be like 20%/80% instead of 50/50. I've hacked it bu running a resize action X number of times but I would like to know if there's some cleaner way to do this with some kind of layout or resize feature.

Thank you

Edit: I will create a discussion asking about this. Maybe someone can help me there or it is actually not a current feature and maybe we could open a GH issue then. Thanks

00-kat:
> Edit: I will create a discussion asking about this.

xref #12218.

PR #11213, opened 2026-03-07
fix(tmux): use-after-free in receivedListWindows action slice

daiimus:
## Summary

`receivedListWindows` stores `windows.items` — a slice into a local `std.ArrayList(Window)` — directly into the action list. The ArrayList is freed by `defer windows.deinit(self.alloc)` on function return, leaving the action holding a dangling pointer.

Currently latent: consumers only read `.len` (inline in the fat pointer), not the pointed-to data. Becomes a live crash when anything iterates `action.windows[i]`.

## Fix

Dupe the slice into the arena allocator so it shares the action list's lifetime:

```zig
const arena_windows = try arena_alloc.dupe(Window, windows.items); try actions.append(arena_alloc, .{ .windows = arena_windows });
```

## Test

Added `"list-windows action data survives function return"` — runs through the startup sequence and dereferences `action.windows[0].id`, `.width`, `.height` in the check callback. Reads freed memory without the fix.

## AI Disclosure

Bug was identified and fix was authored with Claude Code assistance. I understand the root cause: the local ArrayList's `.items` slice becomes invalid after `deinit`, but was stored into the action struct that outlives the function scope. The fix copies into the arena that backs all actions.

### DIFF
--- a/src/terminal/tmux/viewer.zig
+++ b/src/terminal/tmux/viewer.zig
@@ -894,8 +894,11 @@ pub const Viewer = struct {
         }
 
         // Setup our windows action so the caller can process GUI
-        // window changes.
-        try actions.append(arena_alloc, .{ .windows = windows.items });
+        // window changes. We must dupe into the arena because `windows`
+        // is a local ArrayList whose backing memory is freed by the
+        // defer above when this function returns.
+        const arena_windows = try arena_alloc.dupe(Window, windows.items);
+        try actions.append(arena_alloc, .{ .windows = arena_windows });
 
         // Sync up our layouts. This will populate unknown panes, prune, etc.
         try self.syncLayouts(windows.items);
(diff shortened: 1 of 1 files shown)


src/terminal/tmux/viewer.zig
```diff
@@ -894,8 +894,11 @@ pub const Viewer = struct {
         }
 
         // Setup our windows action so the caller can process GUI
-        // window changes.
-        try actions.append(arena_alloc, .{ .windows = windows.items });
+        // window changes. We must dupe into the arena because `windows`
+        // is a local ArrayList whose backing memory is freed by the
+        // defer above when this function returns.
+        const arena_windows = try arena_alloc.dupe(Window, windows.items);
+        try actions.append(arena_alloc, .{ .windows = arena_windows });
```
mitchellh, line 901: I have to check in more detail, but on first pass this doesn't seem right. Why not change windows to be on the arena, possibly layout too so it's not in its own arena but using the shared arena?

Please explain your thinking, not the agents.

mitchellh (changes requested):

PR #10794, opened 2026-02-17
i18n: update Indonesian (id_ID) translations

mikailmm:
Updated the translations for Indonesian using other Indonesian translations in the gnome project as references. Also, why was I the only one who got tagged? What happened to @halosatrio? He was willing (#8126) to be part of the id_ID translation team.
Original issue: #10632

PR #11226, opened 2026-03-08
i18n: update Indonesian translation (id_ID)

halosatrio:
Updated translation for Indonesian (id_ID). This is a duplicate of PR #10794, as the original PR has not been updated since last week. I think it would be better to merge this updated translation before the 1.3 release.


00-kat:
cc @mikailmm.

00-kat (approved):

00-kat:
Thanks! trag1c and I had actually been discussing what to do about the original PR two hours before you pinged its author there; we were discussing if we should rebase and merge that PR, then request that you both create a new PR for the comments, but you creating a new PR makes the decision considerably easier for us.

PR #11242, opened 2026-03-08
Belief in a politics-free community

egorlem:
I sincerely believe that your Vouch system is free from politics and corporate influence. I truly believe you care about the purity of the community.

With deep respect,
Egor Lem

### DIFF
--- a/.github/VOUCHED.td
+++ b/.github/VOUCHED.td
@@ -157,3 +157,4 @@ vlsi
 yamshta
 zenyr
 zeshi09
+-egorlem I believe this is not about politics...


egorlem:
> Hi @egorlem, thanks for your interest in contributing!
> 
> This project requires that pull request authors are vouched, and you are not in the list of vouched users.
> 
> This PR will be closed automatically. See https://github.com/ghostty-org/ghostty/blob/main/CONTRIBUTING.md for more details.


So I ask, since we are not accepted, to add us to the list

00-kat:
There's no point to denouncing yourself. Simply don't interact, if that's what you want.

PR #11248, opened 2026-03-09
feat: change `background_opacity_cells` option to work with opaque cells

PowerUser64:
_Status - currently works, but I want to make sure I'm taking an acceptable approach before I continue._

This makes the `background_opacity_cells` option behave like Alacritty's `transparent_background_colors`, as discussed in #7957.

Currently, this is a bit hacky. To make it better, I would like to change all `blending_enabled` variables from booleans to enums and rename them to `blending_mode`, then translate all assignments of `false` to `none`, and `true` to `one`, then introduce a `src_alpha` enum member where the `is_cell_fill_pass` variable is used right now.

Also, the metal implementation is incomplete as of now, but I'm going to see if I can test it.

Resolves #7957

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -1642,6 +1642,7 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                     .uniforms = frame.uniforms.buffer,
                     .buffers = &.{ null, frame.cells_bg.buffer },
                     .draw = .{ .type = .triangle, .vertex_count = 3 },
+                    .is_cell_fill_pass = self.config.background_opacity_cells,
                 });
 
                 // Kitty images between cell backgrounds and text.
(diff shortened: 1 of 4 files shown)


8bittts:
I exposed this into a GUI:

image

rhodes-b:
> I exposed this into a GUI:
> 
> image
> 

This PR has nothing to do with a gui. If background cells are opaque they get blended incorrectly.

vidhanio:
@PowerUser64 any chance you could re-create this PR now that you have been vouched? I am running into this issue too.

PowerUser64:
> @PowerUser64 any chance you could re-create this PR now that you have been vouched? I am running into this issue too.

@vidhanio thanks, I've been using [dnkl/foot](https://codeberg.org/dnkl/foot) a lot lately so I've kind of forgotten about this. If you wanted to test out my branch and tell me what you think needs to change (if anything), that would be really helpful.

https://github.com/PowerUser64/ghostty/blob/cell-bg-fix

PR #11251, opened 2026-03-09
macos: add AppleScript front window and focused terminal properties

mitchellh:
This adds two new propeties to make it easy to get the frontmost (main) window and the focused terminal within a tab. We already had a property to get the selected tab of a tab group.

### Send Input to Focused Terminal

```AppleScript
tell application "Ghostty"
  set term to focused terminal of selected tab of front window
  input text "pwd\n" to term
end tell
```

### Split the Focused Terminal

```applescript
tell application "Ghostty"
  set currentTerm to focused terminal of selected tab of front window
  set newTerm to split currentTerm direction right
  input text "echo split-ready\n" to newTerm
end tell
```

### DIFF
--- a/macos/Ghostty.sdef
+++ b/macos/Ghostty.sdef
@@ -9,6 +9,9 @@
       <property name="frontmost" code="pisf" type="boolean" access="r" description="Is this the active application?">
         <cocoa key="isActive"/>
       </property>
+      <property name="front window" code="GFWn" type="window" access="r" description="The frontmost Ghostty window.">
+        <cocoa key="frontWindow"/>
+      </property>
       <property name="version" code="vers" type="text" access="r" description="The version number of the application."/>
       <responds-to command="perform action">
         <cocoa method="handlePerformActionScriptCommand:"/>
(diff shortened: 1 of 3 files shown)


8bittts:
I was doing some tests of Applescript and the possibilities are so many! Excited that this was a focus for the team.

mitchellh:
> I was doing some tests of Applescript and the possibilities are so many! Excited that this was a focus for the team.

Yes! Awesome! Please let me know any weaknesses you find. This PR for example was thanks to some feedback I got from someone writing scripts on X. :)

zachdaniel (approved):

zachdaniel:
Last thing I need to perfect my riced up pane switching is a way to run an arbitrary script when pressing a key bind in Ghostty. The specific shape of my problem is that Claude Code consumes all input while open, so nushell never sees my CTRL+hjkl keys typed in to call my AppleScript that intelligently changes windows. If, however, I could execute an AppleScript on a key bind being pressed, then I could check the window title to see if I'm running `nvim` and if so I could send keys to neovim otherwise tell Ghostty to navigate.

As it stands now I can't find a way to navigate out of a Claude Code window 😢

PR #11253, opened 2026-03-09
fix: Cmd+G / Cmd+Shift+G not working in search field

pauley-unsaturated:
When the search field is focused, Cmd+G (Find Next) and Cmd+Shift+G (Find Previous) were consumed by the NSTextView field editor's performKeyEquivalent before Ghostty could handle them. Additionally, navigate_search bindings use .performable, which are intentionally excluded from the menu key equivalent reverse map, so menu-based dispatch could never work.

Fix: introduce a custom field editor (GhosttySearchFieldEditor) that intercepts performKeyEquivalent, checks if the key event matches a ghostty binding via ghostty_surface_key_is_binding, and dispatches it directly via ghostty_surface_key. This is fully config-driven — no hardcoded key equivalents.

Also replace the SwiftUI TextField with an NSViewRepresentable (GhosttySearchField) to guarantee accessibility identifier propagation for the windowWillReturnFieldEditor hook.

Additional fixes included:
- AppDelegate: correct action strings (search:next → navigate_search:next)
- BaseTerminalController: fix findPrevious calling findNext (copy-paste bug)
- GhosttySearchField: guard IME composition (hasMarkedText) for Enter/Escape
- Shared searchFieldIdentifier constant

Includes UI test verifying Cmd+G advances the match counter.

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -89,6 +89,15 @@ class BaseTerminalController: NSWindowController,
     /// Cancellable for aggregating bell state across all surfaces in this controller.
     private var bellStateCancellable: AnyCancellable?
 
+    /// Shared field editor for the search text field. Intercepts key
+    /// equivalents that match ghostty bindings (including performable ones)
+    /// and dispatches them directly to the surface.
+    private lazy var searchFieldEditor: GhosttySearchFieldEditor = {
+        let editor = GhosttySearchFieldEditor()
+        editor.isFieldEditor = true
+        return editor
+    }()
+
     /// An override title for the tab/window set by the user via prompt_tab_title.
     /// When set, this takes precedence over the computed title from the terminal.
     var titleOverride: String? {
@@ -1273,6 +1282,15 @@ class BaseTerminalController: NSWindowController,
         return appDelegate.undoManager
     }
 
+    func windowWillReturnFieldEditor(_ sender: NSWindow, to client: Any?) -> Any? {
+        if let textField = client as? NSTextField,
+           textField.accessibilityIdentifier() == Ghostty.searchFieldIdentifier {
+            searchFieldEditor.surfaceView = focusedSurface
+            return searchFieldEditor
+        }
+        return nil
+    }
+
     // MARK: First Responder
 
     @IBAction func close(_ sender: Any) {
@@ -1415,7 +1433,7 @@ class BaseTerminalController: NSWindowController,
     }
 
     @IBAction func findPrevious(_ sender: Any) {
-        focusedSurface?.findNext(sender)
+        focusedSurface?.findPrevious(sender)
     }
 
     @IBAction func findHide(_ sender: Any) {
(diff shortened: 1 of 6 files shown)


macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -1416,7 +1432,7 @@ class BaseTerminalController: NSWindowController,
     }
 
     @IBAction func findPrevious(_ sender: Any) {
-        focusedSurface?.findNext(sender)
+        focusedSurface?.findPrevious(sender)
```
pauley-unsaturated: This was just a straight typo in the og code, oops!

Gobd:
Does this also fix cmd+v paste not working into the search field, or maybe that is just a me issue? Didn't see any discussions anywhere and was surprised so this seemed as good a place to ask as any.

mitchellh:
@Gobd Thats covered by this issue, the cause is known, just no one fixed yet: #10749

shinebayar-g:
Just to confirm, this used to work properly right?

pauley-unsaturated:
> @Gobd Thats covered by this issue, the cause is known, just no one fixed yet: #10749

I’ll double check that this works with my fix as well just to make sure.

pauley-unsaturated:
So the deal is that copy / paste don’t work because the command subsystem doesn’t know about this find search field. The two options as I see it are to either insert copy / paste into the menu and keyboard shortcuts as normal when the text field gets focus or to make the command system aware of this text field somehow. Probably beyond the scope of this particular PR which is already at the comfort limit for a first contribution.

pauley-unsaturated:
Sorry: I should say that what happens when copy / paste are triggered with the focus in the search text field is that the copy source comes from the associated terminal surface and likewise the paste target is the associated terminal surface. The text gets blitted into the terminal instead of the text field.

injust:
> Sorry: I should say that what happens when copy / paste are triggered with the focus in the search text field is that the copy source comes from the associated terminal surface and likewise the paste target is the associated terminal surface. The text gets blitted into the terminal instead of the text field.

Sounds like you might be running into https://github.com/ghostty-org/ghostty/discussions/11551.

When testing, make sure you don't un-focus and re-focus Ghostty, as that messes up the focus right now.

bo2themax:
Sorry, this is still a work-around; the fix for the root cause is currently underway. As explained in the discussion, this is because we are stripping out the original keybinds from the menus, but to add them back, we still need to make sure there won't be any regressions.

pauley-unsaturated:
> Sorry, this is still a work-around; the fix for the root cause is currently underway. As explained in the discussion, this is because we are stripping out the original keybinds from the menus, but to add them back, we still need to make sure there won't be any regressions.

Do we expect other key-bindings that don’t exist in the menu to work from the search field? If the answer is no then I agree.

bo2themax:
> > Sorry, this is still a work-around; the fix for the root cause is currently underway. As explained in the discussion, this is because we are stripping out the original keybinds from the menus, but to add them back, we still need to make sure there won't be any regressions.
> 
> Do we expect other key-bindings that don’t exist in the menu to work from the search field? If the answer is no then I agree.

No we don't, if a key binding is expected to work when the search/other text field is focused, it should be in the menu.

PR #11277, opened 2026-03-09
Add OSC 8 links to +show-config output.

jcollie:
When enabled, OSC 8 hyperlinks will be embedded in the `+show-config` output linking to the online configuration reference.

Will be disabled by default if `stdout` is not a TTY. Can be overridden with a combination of `--plain` or `--force-links` flags.

### DIFF
--- a/src/config/formatter.zig
+++ b/src/config/formatter.zig
@@ -9,14 +9,20 @@ const Key = @import("key.zig").Key;
 pub fn entryFormatter(
     name: []const u8,
     writer: *std.Io.Writer,
+    links: bool,
 ) EntryFormatter {
-    return .{ .name = name, .writer = writer };
+    return .{
+        .name = name,
+        .writer = writer,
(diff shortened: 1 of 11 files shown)


src/config/formatter.zig
```diff
@@ -9,14 +9,20 @@ const Key = @import("key.zig").Key;
 pub fn entryFormatter(
     name: []const u8,
     writer: *std.Io.Writer,
+    links: bool,
```
mitchellh, line 12: Red flag. I don't think tunneling this type of information deep into the core of the config is the right way to do this. This implies that our formatting is perhaps not the right shape.

mitchellh (changes requested):

tristan957:
I think this is a good idea fwiw. I guess the impl could use some massaging.

jcollie:
Closing until a better way to implement this can be designed.

PR #11285, opened 2026-03-10
tests: disable tests that fail if you have locally installed fonts

jcollie:
If you have "Noto Sans Tai Tham" and/or "Noto Sans Javanese" installed locally on Linux, three tests fail. This PR disables those tests until a more permanent solution can be found.


jcollie:
CC @jacobsandlund

mitchellh:
Can you add a comment above the commented out that explains why they're commented out? Other than that its fine.

PR #11289, opened 2026-03-10
config: add progress-style option

### LINKED ISSUE #11241: Add a `progress-style` configuration to disable progress bars
Ref #10235 

Add a new configuration `progress-style` configuration that for now is a boolean that defaults to true but can be set to false to disable all progress bar rendering in macOS and GTK. The generic `progress-style` name is picked since it may not always be a bar and we may have multiple options in the future of what styling to use.

MOlechowski:
Adds progress-style config to control OSC 9;4 progress bar visibility. Defaults to true, set false to hide.

Fixes #11241

AI Disclosure: Claude Code (Opus 4.6) used for codebase exploration, code review, and testing assistance. All code written and reviewed by hand.

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1012,6 +1012,14 @@ pub const Surface = extern struct {
             priv.progress_bar_timer = null;
         }
 
+        if (priv.config) |config| {
+            if (!config.get().@"progress-style") {
+                log.debug("progress_report action blocked by config", .{});
+                priv.progress_bar_overlay.as(gtk.Widget).setVisible(@intFromBool(false));
+                return;
+            }
+        }
+
         const progress_bar = priv.progress_bar_overlay;
         switch (value.state) {
             // Remove the progress bar
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -1934,6 +1934,15 @@ extension Ghostty {
             case GHOSTTY_TARGET_SURFACE:
                 guard let surface = target.target.surface else { return }
                 guard let surfaceView = self.surfaceView(from: surface) else { return }
+                guard let config = (NSApplication.shared.delegate as? AppDelegate)?.ghostty.config else { return }
+
+                guard config.progressStyle else {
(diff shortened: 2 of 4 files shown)


src/Surface.zig
```diff
@@ -1096,6 +1098,11 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
         },
 
         .progress_report => |v| {
+            if (!self.config.progress_style) {
```
mitchellh: We should gate this at the apprt, not in the core. We still want progress reports to be sent to the apprt, but the apprt should respect the style of how it is shown.

mitchellh (changes requested):

src/apprt/gtk/class/surface.zig
```diff
@@ -1004,6 +1003,10 @@ pub const Surface = extern struct {
     ) void {
         const priv = self.private();
 
+        if (priv.config) |config| {
+            if (!config.get().@"progress-style") return;
```
mitchellh: We should probably move this to below the timer removal and also ensure that the progress bar gets removed too (if it might exist). That way this behaves better with live config reload.

src/apprt/embedded.zig
```diff
@@ -304,6 +303,10 @@ pub const App = struct {
                 },
             },
 
+            .progress_report => {
+                if (!self.config.@"progress-style") return;
+            },
```
mitchellh: Sorry this isn't right, the handling is in macOS/

mitchellh (changes requested):

src/apprt/embedded.zig
MOlechowski: Okay will move it to macos layer, just need a bit more time

MOlechowski:
@mitchellh Will continue working on the change request tomorrow, please hold the review until all the change requests are addressed.

src/apprt/gtk/class/surface.zig
MOlechowski: Moved bellow timer

MOlechowski:
@mitchellh I have moved the guard logic to platform layer progress report funcs, also tested the live reload on linux and macos.

mitchellh (approved):
Looks good now! Thank you! I have to milestone this to 1.4 since its new functionality, but you can always use tip.

PR #11320, opened 2026-03-10
macOS: restore keyboard focus after inline tab title edit

chronologos:
## Summary

- After finishing an inline tab title edit (via keybind or double-click), all keyboard input is lost because `TabTitleEditor.finishEditing()` sets `makeFirstResponder(nil)`, leaving the window itself as first responder with no path back to the terminal surface.
- Adds a `tabTitleEditorDidFinishEditing` delegate callback to `TabTitleEditorDelegate` that fires after every edit (commit or cancel).
- `TerminalWindow` implements it by calling `makeFirstResponder(focusedSurface)` to restore keyboard focus to the terminal.

Fixes https://github.com/ghostty-org/ghostty/discussions/11315

## Testing

- [x] Bind `prompt_tab_title` to a keybind (e.g. `keybind = cmd+shift+i=prompt_tab_title`)
- [x] Trigger inline tab title edit via keybind, press Enter — verify keyboard input works immediately
- [x] Trigger inline tab title edit via keybind, press Escape — verify keyboard input works immediately
- [x] Double-click a tab title, press Enter — verify keyboard input works immediately
- [x] Double-click a tab title, press Escape — verify keyboard input works immediately
- [x] Verify Cmd+number tab switching works after all of the above
- [x] Verify split pane focus is correct after editing tab title with splits open

AI disclosure: Codebase exploration and review via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
+++ b/macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
@@ -835,4 +835,13 @@ extension TerminalWindow: TabTitleEditorDelegate {
         guard let targetController = targetWindow.windowController as? BaseTerminalController else { return }
         targetController.promptTabTitle()
     }
+
+    func tabTitleEditor(_ editor: TabTitleEditor, didFinishEditing targetWindow: NSWindow) {
+        // After inline editing, the first responder is the window itself.
+        // Restore focus to the terminal surface so keyboard input works.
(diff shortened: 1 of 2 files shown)


mitchellh:
!vouch

mitchellh:
Vouched because this looks clean.

mitchellh (approved):
Rename the delegate method to be more idiomatic. Approved pending CI.

PR #11325, opened 2026-03-10
macos: remove tiny 1px gray border in non-native fullscreen

dzhlobo:
It seems macOS 26 draws a tiny 1px border around windows when window shadow is enabled. When shadow is disabled, the border is not drawn.

See https://github.com/ghostty-org/ghostty/discussions/9582#discussioncomment-16070472 for details.

Fixes https://github.com/ghostty-org/ghostty/discussions/9582

I've made initial discovery in https://ampcode.com/threads/T-019cd797-b49b-77ca-b702-ff18027fc6bf, then back and forth-ed a bit https://ampcode.com/threads/T-019cd7e2-87fa-71f8-8e45-3444e7b68676 and wrapped up the implementation.

### DIFF
--- a/macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
+++ b/macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
@@ -468,9 +468,14 @@ class TerminalWindow: NSWindow {
         guard isVisible else { return }
         defer { updateColorSchemeForSurfaceTree() }
 
-        // Basic properties
         appearance = surfaceConfig.windowAppearance
-        hasShadow = surfaceConfig.macosWindowShadow
+
+        // In non-native fullscreen we force shadow off to avoid the 1px gray window border on macOS 26.
+        if let fs = terminalController?.fullscreenStyle, fs.isFullscreen, !styleMask.contains(.fullScreen) {
+            hasShadow = false
+        } else {
+            hasShadow = surfaceConfig.macosWindowShadow
+        }
 
         // Window transparency only takes effect if our window is not native fullscreen.
         // In native fullscreen we disable transparency/opacity because the background
--- a/macos/Sources/Helpers/Fullscreen.swift
+++ b/macos/Sources/Helpers/Fullscreen.swift
@@ -227,6 +227,10 @@ class NonNativeFullscreen: FullscreenBase, FullscreenStyle {
         // from the edges.
         window.styleMask.remove(.resizable)
 
+        // We don't need shadow in fullscreen and shadow causes macOS 26 to
+        // draw a tiny 1px gray border around the window so we disable it.
+        window.hasShadow = false
+
         // Focus window
         window.makeKeyAndOrderFront(nil)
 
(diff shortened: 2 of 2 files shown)


macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
```diff
@@ -468,9 +468,14 @@ class TerminalWindow: NSWindow {
         guard isVisible else { return }
         defer { updateColorSchemeForSurfaceTree() }
 
-        // Basic properties
         appearance = surfaceConfig.windowAppearance
-        hasShadow = surfaceConfig.macosWindowShadow
+
+        // In non-native fullscreen we force shadow off to avoid the 1px gray window border on macOS 26.
+        if let fs = terminalController?.fullscreenStyle, fs.isFullscreen, !styleMask.contains(.fullScreen) {
```
mitchellh, line 474: Is this required? Shouldn't the fullscreen handler you modified handle this already? (I Haven't tested)

mitchellh (changes requested):

macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
dzhlobo: @mitchellh,

When we enter or exit a fullscreen, native or non-native, `fullscreenDidChange` gets called on window controller, which calls `syncAppearance` of controller and then window.

So this `syncAppearance` is called and it resets `hasShadow` we changed previously when entering the non-native fullscreen.

---

I didn't like that window now knows about `fullscreenStyle` of the controller. I thought alternatively about adding an `isFullscreen` method to the `TerminalController` but stopped at this version.

dzhlobo: Perhaps we can avoid resetting `hasShadow` completely in `syncAppearance`? I'm not familiar with why we need to `syncAppearance`.

dzhlobo:
Hey @mitchellh, I'm not sure whether I should re-request review.

I just want to complete it or maybe reject it if it's bad before Ghostty moves out of GitHub.

macos/Sources/Features/Terminal/Window Styles/TerminalWindow.swift
```diff
@@ -468,9 +468,14 @@ class TerminalWindow: NSWindow {
         guard isVisible else { return }
         defer { updateColorSchemeForSurfaceTree() }
 
-        // Basic properties
         appearance = surfaceConfig.windowAppearance
-        hasShadow = surfaceConfig.macosWindowShadow
```
bo2themax, line 473: I would recommend using `hasShadow` directly from the fullscreen style, and let concrete styles handle different values.

```swift
hasShadow = terminalController?.fullscreenStyle?.hasShadow ?? surfaceConfig.macosWindowShadow
```

bo2themax: As for the reason of it being here, I believe it's related to the fact that the shadow can be removed/added after a configuration change without reopening the window.

bo2themax (changes requested):

PR #11337, opened 2026-03-10
config: working-directory expands ~/ prefix

mitchellh:
Fixes #11336

Introduce a proper WorkingDirectory tagged union type with home, inherit, and path variants. The field is now an optional (?WorkingDirectory) where null represents "use platform default" which is resolved during Config.finalize to .inherit (CLI) or .home (desktop launcher).


jcollie (approved):
LGTM

PR #10983, opened 2026-02-24
appintents: expose PID and TTY on TerminalEntity

ocean6954:
## Summary

- Add `pid` (child process PID) and `tty` (TTY device name, e.g. `/dev/ttys003`) as `@Property` fields on `TerminalEntity`
- Expose both as queryable details via `GetTerminalDetailsIntent`
- Add C API functions `ghostty_surface_child_pid` and `ghostty_surface_tty_name` bridging Zig core to Swift

This enables external applications and Shortcuts to identify and filter terminals by their PID or TTY device, as requested in #10756.

## Implementation

**Zig core → C API → Swift → App Intents** pipeline for both properties:

| Layer | File | Change |
|-------|------|--------|
| Subprocess | `src/termio/Exec.zig` | Store `tty_name` via `ttyname_r()` before closing slave fd |
| Surface | `src/Surface.zig` | Add `childPid()` and `ttyName()` methods |
| C API | `src/apprt/embedded.zig`, `include/ghostty.h` | Export `ghostty_surface_child_pid`, `ghostty_surface_tty_name` |
| Swift | `SurfaceView_AppKit.swift` | Add `childPID` and `ttyName` computed properties |
| App Intents | `TerminalEntity.swift`, `GetTerminalDetailsIntent.swift` | Add `pid` and `tty` as `@Property` and `TerminalDetail` cases |

### Notes
- Uses `ttyname_r` (reentrant) instead of `ttyname` for thread safety
- TTY name string is duped into the subprocess arena allocator for stable lifetime
- PID returns `-1` / `nil` for Flatpak or unavailable processes
- TTY returns `null` / `nil` on Windows or when not available

## Test plan

- [ ] Build succeeds on macOS with `zig build -Doptimize=ReleaseFast`
- [ ] Launch built Ghostty.app, open terminal tabs
- [ ] Use Shortcuts app: "Get Details of Terminal" → select "PID" → verify valid PID returned
- [ ] Use Shortcuts app: "Get Details of Terminal" → select "TTY" → verify TTY path returned (e.g. `/dev/ttys003`)
- [ ] Verify PID matches `ps` output for the shell process in that tab
- [ ] Verify TTY matches `tty` command output in that tab

Closes #10756

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11354, opened 2026-03-11
appintents: expose PID and TTY on TerminalEntity

ocean6954:
This is a re-submission of #10983. Vouched in #11287.

Add child process PID and TTY device name as queryable properties on
TerminalEntity for App Intents. This enables external applications and
Shortcuts to identify and filter terminals by their PID or TTY device.

Closes #10756

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1085,6 +1085,8 @@ ghostty_surface_config_s ghostty_surface_inherited_config(ghostty_surface_t, gho
 void ghostty_surface_update_config(ghostty_surface_t, ghostty_config_t);
 bool ghostty_surface_needs_confirm_quit(ghostty_surface_t);
 bool ghostty_surface_process_exited(ghostty_surface_t);
+uint64_t ghostty_surface_child_pid(ghostty_surface_t);
+const char* ghostty_surface_tty_name(ghostty_surface_t);
 void ghostty_surface_refresh(ghostty_surface_t);
 void ghostty_surface_draw(ghostty_surface_t);
 void ghostty_surface_set_content_scale(ghostty_surface_t, double, double);
--- a/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
+++ b/macos/Sources/Features/App Intents/Entities/TerminalEntity.swift
@@ -14,6 +14,12 @@ struct TerminalEntity: AppEntity {
     @Property(title: "Kind")
     var kind: Kind
 
+    @Property(title: "PID")
+    var pid: Int?
+
+    @Property(title: "TTY")
+    var tty: String?
+
     var screenshot: NSImage?
 
     static var typeDisplayRepresentation: TypeDisplayRepresentation {
@@ -49,6 +55,8 @@ struct TerminalEntity: AppEntity {
         self.id = view.id
         self.title = view.title
         self.workingDirectory = view.pwd
+        self.pid = view.surfaceModel?.childPID
+        self.tty = view.surfaceModel?.ttyName
         if let nsImage = ImageRenderer(content: view.screenshot()).nsImage {
             self.screenshot = nsImage
         }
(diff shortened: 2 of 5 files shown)


macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
```diff
@@ -178,6 +178,20 @@ extension Ghostty {
             return ghostty_surface_process_exited(surface)
         }
 
+        // Returns the PID of the child process, or nil if not available.
+        var childPID: Int32? {
```
mitchellh: These should live on the Ghostty.Surface model instead of here, its been a long transition to move over there.

src/Surface.zig
```diff
@@ -940,6 +936,26 @@ pub fn needsConfirmQuit(self: *Surface) bool {
     };
 }
 
+/// Returns the PID of the child process, or -1 if not available.
+pub fn childPid(self: *Surface) i32 {
```
mitchellh: We should use native Zig optionals here, not sentinel integral values.

src/Surface.zig
```diff
@@ -940,6 +936,26 @@ pub fn needsConfirmQuit(self: *Surface) bool {
...
+            switch (process) {
+                .fork_exec => |cmd| return if (cmd.pid) |p| @intCast(p) else -1,
+                .flatpak => return -1,
+            }
+        },
+    }
+}
+
+/// Returns the TTY device name (e.g. "/dev/ttys003"), or null if not available.
+pub fn ttyName(self: *Surface) ?[*:0]const u8 {
```
mitchellh: Same, use Zig types.

src/termio/Exec.zig
```diff
@@ -586,6 +586,7 @@ const Subprocess = struct {
     screen_size: renderer.ScreenSize,
     pty: ?Pty = null,
     process: ?Process = null,
+    tty_name: ?[*:0]const u8 = null,
```
mitchellh: Store as a zig type, not this.

mitchellh (changes requested):

grimmerk:
Hi @ocean6954, just checking in — are you still working on this? I have a use case for this feature (switching to Claude Code sessions by PID in a terminal switcher app). Let me know if there's anything I can help with!

ocean6954:
Hi @grimmerk

Sorry for the late reply! I had already pushed the fixes but wasn't sure how to re-request a review — I just did that now. Thanks for your interest in this feature!

dnwpark:
also looking for this!

jcollie:
@ocean6954 Did you want to rebase this on top of the core changes from #11639 ?

include/ghostty.h
```diff
@@ -1078,6 +1072,8 @@ ghostty_surface_config_s ghostty_surface_inherited_config(ghostty_surface_t, gho
 void ghostty_surface_update_config(ghostty_surface_t, ghostty_config_t);
 bool ghostty_surface_needs_confirm_quit(ghostty_surface_t);
 bool ghostty_surface_process_exited(ghostty_surface_t);
+int32_t ghostty_surface_child_pid(ghostty_surface_t);
```
tristan957: What do you think about returning a `pid_t`? That is the POSIX type for an integer representing a process ID.

Only using 32 bits is restrictive, while `pid_t` is kind of opaque in that each platform can define it differently. You never know if Ghostty will exist on platforms supporting larger that 32 bit process IDs. I would assume that there are equivalent types in Swift and Zig.

jcollie: The core API returns a `u64`. I chose that specifically over a `pid_t` to avoid any sort of platform dependence (Zig seems to be going hard against POSIX in the upcoming 0.16 release).

ocean6954: @tristan957 

I've rebased on top of #11639 and refactored to use the getProcessInfo() API. As @jcollie mentioned, the core API returns u64. This PR converts it to i64 at the C API boundary to allow returning -1 when unavailable.

ocean6954:
Rebased on top of #11639 and refactored to use getProcessInfo(). Squashed into a single commit.

include/ghostty.h
jcollie: `0` _should_ work as a signal that the PID was unavailable, at least on any system I'm aware of that's not a valid PID.

ocean6954: Good point, updated to use u64 with 0 as the sentinel value.

00-kat:
This has been superseded by #11922 right?

PR #11361, opened 2026-03-11
docs: fix backtick rendering in selection-word-chars default value

puzza007:
## Summary
- The default value for `selection-word-chars` contains a literal backtick character, which broke inline code rendering on the website — the code span was prematurely closed, leaving part of the value outside the formatted block.
- Fixed by using double backtick delimiters which is the standard markdown way to include a backtick in inline code.


pluiedev (approved):
Lol how did we not catch this earlier, LGTM

00-kat (approved):

PR #11373, opened 2026-03-11
add direct set_surface_title and set_tab_title actions

mitchellh:
Fixes #11316

This mirrors the `prompt` actions (hence why there is no window action here) and enables setting titles via keybind actions which importantly lets this work via command palettes, App Intents, AppleScript, etc.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -5482,6 +5482,26 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
             .tab,
         ),
 
+        .set_surface_title => |v| {
+            const title = try self.alloc.dupeZ(u8, v);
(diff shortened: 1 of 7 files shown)


src/Surface.zig
```diff
@@ -5482,6 +5482,26 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
             .tab,
         ),
 
+        .set_surface_title => |v| {
+            const title = try self.alloc.dupeZ(u8, v);
```
mitchellh, line 5486: Note that I'd like to eventually just allow bindings to support `[:0]const u8` so we don't have to do this, but we already do this elsewhere and I didn't want to muddy the PR.

jcollie (approved):

PR #11377, opened 2026-03-11
bash: fix extra newlines with readline vi mode indicator

### LINKED ISSUE #10953: bash: ghostty shell-integration doesn't work with readline mode indicator on OSX
### Discussed in https://github.com/ghostty-org/ghostty/discussions/10948

<sup>Originally posted by **karlicoss** February 22, 2026</sup>
### Issue Description

I'm using inputrc configuration to set up vi mode across different cli apps that use readline:
```
$ cat .inputrc
set editing-mode vi
set show-mode-in-prompt on
```

My bash PS1 is default (I think)
```
koldun:~ karlicos$ echo $PS1
\h:\W \u\$
```

### Expected Behavior

What I'm expecting is something like this -- i.e. mode indicator (ins/cmd) is on the same like as prompt:
```
(ins)koldun:~ karlicos$ echo hello
hello
(cmd)koldun:~ karlicos$ a b c
```
This is the behaviour I see in iterm/Terminal.app.

### Actual Behavior

Ghostty prints vi mode indicator (`(ins)` or `(cmd)`) on the previous line instead of the same line as the shell prompt:
```
koldun:~ karlicos$ echo hello
hello
(ins)
(cmd)n:~ karlicos$ a b c
```

After this I tried to do the following in ghostty config and it fixed the issue
```
shell-integration = none
```
, but then things like sudo/ssh forwarding and terminal title won't work :(

I also tried only setting custom integration features as per [this guide](https://ghostty.org/docs/config/reference#shell-integration-features) but seemed like it had no effect, e.g. terminal title was set as long as bash integration was enabled no matter the `shell-integration-features`. I'll try to dig that further, might be some separate issue.

I guess this is because readline adds something in front of PS1, and this is also what ghossty does [here](https://github.com/ghostty-org/ghostty/blob/504a3611f6f2d9a1cb1a4eb419b41aea3c8049ca/src/shell-integration/bash/ghostty.bash#L201). Not 100% sure, possibly this is somehow related https://github.com/ghostty-org/ghostty/issues/1966

### Reproduction Steps

1. Configure .inputrc as per issue description
2. Run ghostty without config/empty config
3. Try pressing Enter multiple times, or typing something and hitting Esc to switch to command/normal vim mode, you should see the issue reproduce

### Ghostty Logs

_No response_

### Ghostty Version

```text
Ghostty 1.3.0-main+861a9cf53

Version
  - version: 1.3.0-main+861a9cf53
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - app runtime   : .none
  - font engine   : .coretext
  - libxev        : kqueue
```

### OS Version Information

OSX Darwin Kernel Version 24.6.0

None

_No response_

```ini
# empty config is enough to reproduce
```

### ~/.inputrc
```
set editing-mode vi
set show-mode-in-prompt on
```

### I acknowledge that:

jparise:
Use OSC 133;P (prompt mark) instead of 133;A (fresh line + prompt mark) inside PS1 and PS2. Readline redraws the prompt on vi mode switches, Ctrl-L, and other events, and 133;A's fresh-line behavior would emit a CR+LF whenever the cursor wasn't at column 0, causing visible extra newlines.

The one-time 133;A is now emitted via printf in __ghostty_precmd, which only runs once per prompt cycle via PROMPT_COMMAND. On SIGWINCH, bash redraws PS1 (firing the 133;P marks) but doesn't re-run PROMPT_COMMAND, so there's no unwanted fresh-line on resize either. The redraw=last flag persists from the initial printf.

This is a little less optimal than our previous approach, in terms of number of prompt marks we emit, but it produces an overall more correct result, which is the important thing.

Because readline prints its output outside the scope of PS1, those characters "inherit" the surrounded prompt scope. This is usually fine, but it can sometimes get out of sync (especially during redraws). This is inherently a limitation of the fact that it's a separate output channel, so we just have to accept that can happen.

Fixes: #10953
See: #11267

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -195,11 +195,11 @@ function __ghostty_precmd() {
     _GHOSTTY_SAVE_PS1="$PS1"
     _GHOSTTY_SAVE_PS2="$PS2"
 
-    # Marks. We need to do fresh line (A) at the beginning of the prompt
-    # since if the cursor is not at the beginning of a line, the terminal
-    # will emit a newline.
-    PS1='\[\e]133;A;redraw=last;cl=line;aid='"$BASHPID"'\a\]'$PS1'\[\e]133;B\a\]'
-    PS2='\[\e]133;A;k=s\a\]'$PS2'\[\e]133;B\a\]'
+    # Use 133;P (not 133;A) inside PS1 to avoid fresh-line behavior on
+    # readline redraws (e.g., vi mode switches, Ctrl-L). The initial
+    # 133;A with fresh-line is emitted once via printf below.
+    PS1='\[\e]133;P;k=i\a\]'$PS1'\[\e]133;B\a\]'
+    PS2='\[\e]133;P;k=s\a\]'$PS2'\[\e]133;B\a\]'
 
     # Bash doesn't redraw the leading lines in a multiline prompt so we mark
     # the start of each line (after each newline) as a secondary prompt. This
@@ -210,7 +210,7 @@ function __ghostty_precmd() {
     # because literal newlines may appear inside $(...) command substitutions
     # where inserting escape sequences would break shell syntax.
     if [[ "$PS1" == *"\n"* ]]; then
-      PS1="${PS1//\\n/\\n$'\\[\\e]133;A;k=s\\a\\]'}"
+      PS1="${PS1//\\n/\\n$'\\[\\e]133;P;k=s\\a\\]'}"
     fi
 
     # Cursor
@@ -232,6 +232,9 @@ function __ghostty_precmd() {
     # End of current command. Report its status.
     builtin printf "\e]133;D;%s;aid=%s\a" "$ret" "$BASHPID"
   fi
+
+  # Fresh line and start of prompt.
+  builtin printf "\e]133;A;redraw=last;cl=line;aid=%s\a" "$BASHPID"
 
   # unfortunately bash provides no hooks to detect cwd changes
   # in particular this means cwd reporting will not happen for a


src/shell-integration/bash/ghostty.bash
```diff
@@ -195,11 +195,11 @@ function __ghostty_precmd() {
...
 
-    # Marks. We need to do fresh line (A) at the beginning of the prompt
-    # since if the cursor is not at the beginning of a line, the terminal
-    # will emit a newline.
-    PS1='\[\e]133;A;redraw=last;cl=line;aid='"$BASHPID"'\a\]'$PS1'\[\e]133;B\a\]'
-    PS2='\[\e]133;A;k=s\a\]'$PS2'\[\e]133;B\a\]'
+    # Use 133;P (not 133;A) inside PS1 to avoid fresh-line behavior on
+    # readline redraws (e.g., vi mode switches, Ctrl-L). The initial
+    # 133;A with fresh-line is emitted once via printf below.
+    PS1='\[\e]133;P;k=i\a\]'$PS1'\[\e]133;B\a\]'
```
mitchellh, line 201: Is it correct here to drop all our redraw/cl/aid params? Cause this is only for redraw?

mitchellh (changes requested):

src/shell-integration/bash/ghostty.bash
jparise: Those are now in the `133;A` printed down below. And `133;P` doesn't take those options.

mitchellh: Yep just checking. Okay!

mitchellh (approved):

PR #11380, opened 2026-03-11
macOS: fix window frame when (re)opening new window

bo2themax:
Claude wrote the fail path in the UI tests, or you can easily reproduce this manually. This is kinda a regression after #11322, since we are not delaying the frame update anymore, which exposes some of the "flaws" of the previous implementation.

The following three commits fix this step by step:
- We shouldn't save intermediate frames when the window is loading, which is triggered by `windowDidResize` and `windowDidMove` during the process.
- We should set the initial position (from the config) after the window is loaded.
- A small refactor on `LastWindowPosition` to support restoring the window frame under certain conditions.

https://github.com/user-attachments/assets/6f90f9a5-653d-4146-95c6-8e5c69bda656

### AI Disclosure

Claude helped me write the UI tests.

### DIFF
--- a/macos/GhosttyUITests/GhosttyWindowPositionUITests.swift
+++ b/macos/GhosttyUITests/GhosttyWindowPositionUITests.swift
@@ -1,0 +1,158 @@
+//
+//  GhosttyWindowPositionUITests.swift
+//  GhosttyUITests
+//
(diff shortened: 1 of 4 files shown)


mitchellh (approved):
This stuff has always been whackamole but the reasoning behind the movement of these makes sense to me.

PR #11383, opened 2026-03-11
macos: only show split grab handle when the mouse is near it

mitchellh:
Fixes #11379

For this pass, I made it a very simple "within 20%" (height-wise) of the split handle down. There is no horizontal component. I want to find the right balance between always visible (today mostly) to only visible on direct hover, because I think it'll be too hard to discover on that far right side.


mitchellh:
https://github.com/user-attachments/assets/71133037-176e-4a91-8806-038b86fc9d8e

PR #11161, opened 2026-03-04
macOS: refine new-window behaviors

bo2themax:
This fixes #9310, rc is that we're saving `lastCascadePoint`, but not updating it when a window moves; it only updates when a new window opens.

In fact, cascading only makes sense when there's an existing window in this screen; when there is no window in this screen, we only need to restore the last frame we saved.

**The first two commits solve the `new-window` behaviors mentioned in #9310, and the third commit changes `LastWindowPosition` to save and restore frame information by screen id, and added some tests for it**

Last commit skips saving the native full-screen window frame, which makes sense to me after reading [the comment here](https://github.com/ghostty-org/ghostty/pull/11070#issuecomment-4003349264).

https://github.com/user-attachments/assets/aa3bdd08-e3ac-4311-bf5c-4cda23c1bd17

### AI Disclosure

AI proofread my comments.

PR #11426, opened 2026-03-12
macOS: fix window cascading

bo2themax:
Added test case for cascading **without moving previous window**, #11161 will follow up for more accurate cascading after this.

Fixed window cascading after last pr, now we should perform cascading **after** showing the window.


bo2themax:
@mitchellh drag-split has the same issue, drafting a test now

PR #11429, opened 2026-03-12
macOS: fix window position when dragging split into a new window

bo2themax:


mitchellh (approved):

bo2themax:
Left new-tab unchanged, because I couldn't find a way to trigger this path. But it's removed in #11161.

image

PR #11449, opened 2026-03-13
Search wrap behavior

rhodes-b:
Search wrapping has been highly requested.

some examples
https://github.com/ghostty-org/ghostty/discussions/11080
https://github.com/ghostty-org/ghostty/discussions/11440
https://github.com/ghostty-org/ghostty/discussions/11441
https://github.com/ghostty-org/ghostty/discussions/9762
https://github.com/ghostty-org/ghostty/discussions/9790

I also think it makes sense as its the default behavior in browsers (and I assume other apps)

I tested where nothing is outputting and a loop where active was going into history not anything where pages would start to get reused though

the following comment seems to me it should be safe to have wrap around behavior but maybe there was something else I missed about the active + history buffer on why that isn't true, testing basic cases it worked just fine for me

https://github.com/ghostty-org/ghostty/blob/main/src/terminal/highlight.zig#L107-L111

### DIFF
--- a/src/terminal/search/screen.zig
+++ b/src/terminal/search/screen.zig
@@ -740,13 +740,9 @@ pub const ScreenSearch = struct {
             return true;
         };
 
-        const next_idx = prev.idx + 1;
         const active_len = self.active_results.items.len;
         const history_len = self.history_results.items.len;
-        if (next_idx >= active_len + history_len) {
-            // No more matches. We don't wrap or reset the match currently.
-            return false;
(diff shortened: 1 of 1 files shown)


mitchellh:
Going to push this to a 1.3.2 just so I can study it more and give it time to bake, but I'm fine with the default change.

mitchellh:
Looks good. Let's let this bake in tip and see what pops out!

PR #11484, opened 2026-03-14
fix: enable cursor blinking mode by default

elias8:
Ghostty already defaults to blinking mode and behavior should be maintained in core to provide consistent and [standard](https://vt100.net/docs/vt510-rm/DECSCUSR.html) defaults.

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -191,6 +191,7 @@ pub const SavedCursor = struct {
     protected: bool,
     pending_wrap: bool,
     origin: bool,
+    cursor_blinking: bool,
     charset: CharsetState,
 };
 
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -1094,6 +1094,7 @@ pub fn saveCursor(self: *Terminal) void {
         .protected = self.screens.active.cursor.protected,
         .pending_wrap = self.screens.active.cursor.pending_wrap,
         .origin = self.modes.get(.origin),
+        .cursor_blinking = self.modes.get(.cursor_blinking),
         .charset = self.screens.active.charset,
     };
 }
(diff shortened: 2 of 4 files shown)


rhodes-b:
> Ghostty already defaults to blinking mode

~thats not true shell-integration defaults to blinking mode which is separate from the terminal itself~ it sets the blink bar by default im nots sure on other modes

elias8:
> > Ghostty already defaults to blinking mode
> 
> ~thats not true shell-integration defaults to blinking mode which is separate from the terminal itself~ it sets the blink bar by default im nots sure on other modes

[Here](https://github.com/ghostty-org/ghostty/blob/e75f8956c576ea8acf6ee4e70d45f918738e7512/src/termio/stream_handler.zig#L897) it falls back to `true` when not configured, which works in the ghostty but it should be the default behavior in the core impl following the standard and for non ghostty use cases (libghostty).

mitchellh:
> but it should be the default behavior in the core impl following the standard and for non ghostty use cases (libghostty).

I think either value for libghostty is fine, because libghostty doesn't dictate cursor blinking anyways. Right?

elias8:
> because libghostty doesn't dictate cursor blinking anyways. Right?

I think it does in three states: when the terminal is initialized, when it is reset, and when the screen switches between primary and alternate. I'm using the terminal (extended implementation of ghostty-web) and ran into an issue where the cursor doesn't blink or restore the state after exiting the alternate screen. 

I also came across [this](https://github.com/ghostty-org/ghostty/blob/e75f8956c576ea8acf6ee4e70d45f918738e7512/src/termio/stream_handler.zig#L645-L682) comment, which I think is related.

mitchellh:
> I also came across [this](https://github.com/ghostty-org/ghostty/blob/e75f8956c576ea8acf6ee4e70d45f918738e7512/src/termio/stream_handler.zig#L645-L682) comment, which I think is related.

Ahhhhhh yes.

This is actually a really good argument for this to be default off. Because the more modern approach (CSI q) _will not work_ if mode 12 is on. And if it defaults to on, then CSI q will be busted. I think default off is the way to go. I'll continue to monitor this PR for discussions though.

elias8:
I'll check my code again. And btw, I pushed a second commit to restore the state after exiting an alternate screen. Should we keep that?

PR #11490, opened 2026-03-14
font/shaper: skip user font-features for fallback font faces

jesusvazquez:
Fixes #11464

User-configured font features (e.g. cv10, dlig) were being applied to all fonts including fallback fonts. This caused unexpected behavior such as CJK fallback glyphs changing when a Latin-only feature was enabled as primary font and cv10 feature was enabled for it. When introducing a CJK character not present in the Latin only font, the system would fallback to another font but it would still apply the cv10 feature to it. And the unexpected behavior is that you may inputting a simplified chinese character for example and getting the traditional equivalent as output because it turns out thats whats inside cv10 for that fallback font. 

Font features are now only applied to the primary font. Fallback fonts get only the default features (e.g. liga). 

This change applies to both the CoreText and HarfBuzz shapers.

Added a regression unitary test for the issue. This test is not using the initially reported fonts from the discussion https://github.com/ghostty-org/ghostty/discussions/11427. Instead it uses embedded fonts Emoji Text and Inconsolata 

<details>
<summary>
See details to reproduce the discussion issue
</summary>

The unit test already has coverage but to test explicitly what was discussed in https://github.com/ghostty-org/ghostty/discussions/11427 do the following.

I'm using `Jetbrains Mono` but you can get  Commit Mono Nerd font with
```
brew install --cask font-commit-mono-nerd-font
```

The PingFang SC font is the default font in my macOS version for chinese characters.

Write down the following ghostty-config
```
font-family = Jetbrains Mono
font-feature = cv10
``` 

Build ghostty with `zig build`
Execute ghostty with the following command
```
zig-out/Ghostty.app/Contents/MacOS/ghostty --config-file=ghostty-config
```

A new ghostty window will come up. Execute the following and you should see the simplified version of the chinese characters!
```ghostty +show-face --string="l 门 关 点 陆"```

image

For the record, if you execute the same command from main this is how it looks
```
zig-out/Ghostty.app/Contents/MacOS/ghostty --config-file=ghostty-config
```

image

</details>

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -230,8 +230,9 @@ language: ?[:0]const u8 = null,
 ///
 /// The syntax is fairly loose, but invalid settings will be silently ignored.
 ///
-/// The font feature will apply to all fonts rendered by Ghostty. A future
-/// enhancement will allow targeting specific faces.
+/// Font features are only applied to the primary font family. Fallback
+/// fonts (fonts automatically selected when the primary font doesn't
+/// contain a needed glyph) use only the default font features.
 ///
 /// To disable programming ligatures, use `-calt` since this is the typical
 /// feature name for programming ligatures. To look into what font features
--- a/src/font/shaper/coretext.zig
+++ b/src/font/shaper/coretext.zig
@@ -47,6 +47,9 @@ pub const Shaper = struct {
     features: *macos.foundation.Dictionary,
     /// A version of the features dictionary with the default features excluded.
     features_no_default: *macos.foundation.Dictionary,
+    /// A version of the features dictionary with only the default features
+    /// (no user-configured features). Used for fallback fonts.
+    features_default_only: *macos.foundation.Dictionary,
 
     /// The shared memory used for shaping results.
     cell_buf: CellBuf,
@@ -168,6 +171,8 @@ pub const Shaper = struct {
         errdefer features.release();
         const features_no_default = try makeFeaturesDict(feats);
         errdefer features_no_default.release();
+        const features_default_only = try makeFeaturesDict(&default_features);
+        errdefer features_default_only.release();
 
         var run_state = RunState.init();
         errdefer run_state.deinit(alloc);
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/config/Config.zig [hunk 1]
```diff
@@ -230,8 +230,9 @@ language: ?[:0]const u8 = null,
 ///
 /// The syntax is fairly loose, but invalid settings will be silently ignored.
 ///
-/// The font feature will apply to all fonts rendered by Ghostty. A future
-/// enhancement will allow targeting specific faces.
+/// Font features are only applied to the primary font family. Fallback
```
jesusvazquez, line 233: I checked the website to see if there were docs about this and there are https://ghostty.org/docs/config/reference#font-feature but it looks autogenerated so I'm guessing that this change is enough and once merged this will go live to the website as well.

src/font/shaper/coretext.zig
```diff
@@ -610,9 +617,12 @@ pub const Shaper = struct {
             defer grid.lock.unlockShared();
 
             const face = try grid.resolver.collection.getFace(index);
+            const entry = try grid.resolver.collection.getEntry(index);
             const original = face.font;
 
-            const attrs = if (face.quirks_disable_default_font_features)
+            const attrs = if (entry.fallback)
+                self.features_default_only
+            else if (face.quirks_disable_default_font_features)
```
jesusvazquez, line 625: I think this is dead code but I've left it because I didnt want to cause another bug.

I saw https://github.com/ghostty-org/ghostty/blob/main/src/quirks.zig#L14-L22 which now always returns `false`

Happy to remove that code in a follow up PR if the reviewer thinks this is a good idea.

src/font/shaper/harfbuzz.zig
```diff
@@ -138,14 +138,16 @@ pub const Shaper = struct {
...
-                // our features by the hardcoded items because always
-                // add those at the beginning.
-                break :i default_features.len;
-            };
+            const entry = try run.grid.resolver.collection.getEntry(run.font_index);
+
+            // If we are disabling default font features we just offset
+            // our features by the hardcoded items because we always
+            // add those at the beginning.
+            const i = if (!face.quirks_disable_default_font_features) 0 else default_features.len;
```
jesusvazquez, line 146: Just like above I think this is dead code but I've left it because I didnt want to cause another bug.

I saw https://github.com/ghostty-org/ghostty/blob/main/src/quirks.zig#L14-L22 which now always returns `false`

Happy to remove that code in a follow up PR if the reviewer thinks this is a good idea.

PR #11491, opened 2026-03-14
config: add `equal` option to `window-padding-balance`

devsunb:
Change `window-padding-balance` from `bool` to an enum with three values:

- `false` - no balancing (default, unchanged)
- `true` - balance with vshift that caps top padding and shifts excess to bottom (existing behavior, unchanged)
- `equal` - balance whitespace equally on all four sides

This gives users who prefer truly equal padding a way to opt in without changing the default behavior.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1963,7 +1963,16 @@ keybind: Keybinds = .{},
 /// apply. The other padding is applied first and may affect how many grid cells
 /// actually exist, and this is applied last in order to balance the padding
 /// given a certain viewport size and grid cell size.
-@"window-padding-balance": bool = false,
+///
+/// Valid values are:
+///
+/// * `false` - No balancing is applied.
+/// * `true` - Balance the padding, but cap the top padding to avoid
+///   excessive space above the first row. Any excess is shifted to the
+///   bottom.
+/// * `equal` - Balance the padding equally on all sides without any
+///   top-padding cap. (Available since: 1.4.0)
+@"window-padding-balance": WindowPaddingBalance = .false,
 
 /// The color of the padding area of the window. Valid values are:
 ///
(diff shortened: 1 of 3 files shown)


mitchellh:
!vouch

devsunb:
@mitchellh Thank you. I was writing a Vouch Request after reading the Contributing Guide late.

I ran into image rendering [issues](https://github.com/folke/snacks.nvim/issues/2165#issuecomment-3254085970) with snacks.nvim inside tmux because WezTerm’s Kitty Image Protocol support does not currently support unicode placeholders. I first explored [fixing](https://github.com/devsunb/wezterm/tree/kitty-image-protocol) this in WezTerm, but since I needed a solution sooner, I also began evaluating Ghostty, which is under active development.

To match my WezTerm setup more closely in Ghostty, I needed two things: even padding support, which this PR implements and which is similar to WezTerm’s [window_content_alignment](https://wezterm.org/config/lua/config/window_content_alignment.html) option, and FreeType font rasterization, which currently seems to require a manual build using `zig build -Dfont-backend=coretext_freetype ...`.

I have read and agree with the AI Policy. Claude Code assisted me in understanding the codebase and generating code, but I have reviewed and understood all text and code included in this PR.

src/config/Config.zig
```diff
@@ -1963,7 +1963,16 @@ keybind: Keybinds = .{},
...
-@"window-padding-balance": bool = false,
+///
+/// Valid values are:
+///
+/// * `false` - No balancing is applied.
+/// * `true` - Balance the padding, but cap the top padding to avoid
+///   excessive space above the first row. Any excess is shifted to the
+///   bottom.
+/// * `equal` - Balance the padding equally on all sides without any
+///   top-padding cap.
```
mitchellh: Add a note on equal "Available since: 1.4.0"

mitchellh (changes requested):
Perfect except for one note.

src/config/Config.zig
devsunb: Added, thanks.

mitchellh (approved):
Perfect except for one note.

PR #11503, opened 2026-03-14
terminal/screen: preserve leading whitespaces in selectLine

jesusvazquez:
Fixes #11463

`selectLine` was trimming leading whitespace on triple-click line selection, which strips indentation from the first selected line. I can see how this seemed right at the time but based on the discussion https://github.com/ghostty-org/ghostty/discussions/11411 and how other terminals behave we're now adjusting this.

`selectLine` will now preserve the leading whitespaces.

We continue to drop the trailing whitespaces in the line.

Also added a regression unit test.
Also updated a few other unit tests that had leading whitespaces.

### Before

image

### Now

image

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -2627,27 +2627,9 @@ pub fn selectLine(self: *const Screen, opts: SelectLine) ?Selection {
         return null;
     };
 
-    // Go forward from the start to find the first non-whitespace character.
-    const start: Pin = start: {
-        const whitespace = opts.whitespace orelse break :start start_pin;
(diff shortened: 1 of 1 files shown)


mitchellh:
This isn't right. We already have an option to modify whitespace matching. triple-click select should probably use that. If I'm wrong in that assessment please let me know.

jesusvazquez:
> We already have an option to modify whitespace matching. triple-click select should probably use that. If I'm wrong in that assessment please let me know.

You are right 👍 I did not fully understand that bit before, my bad. Here is the new attempt at it https://github.com/ghostty-org/ghostty/pull/11535

PR #11518, opened 2026-03-15
shell-integration: fix ssh-env SetEnv clobbering user SSH config

j0hnm4r5:
## Problem

Ghostty's `ssh-env` shell integration uses `-o "SetEnv COLORTERM=truecolor"` when wrapping SSH commands. OpenSSH treats command-line `-o SetEnv` options as **replacements** for all `SetEnv` entries in `~/.ssh/config`, not additions. This silently drops any user-configured `SetEnv` variables.

For example, a user with this in their SSH config:
```
Host myserver
  SetEnv MY_VAR=hello
```
...would find `MY_VAR` empty after SSHing through Ghostty with `ssh-env` enabled.

Reference: https://github.com/ghostty-org/ghostty/discussions/10871

## Fix

Replace `-o "SetEnv COLORTERM=truecolor"` with the additive pattern: set `COLORTERM=truecolor` locally before the SSH call and forward it via `SendEnv`.

`SendEnv` is additive — it does not clobber `SetEnv` entries in `~/.ssh/config`.

**Trade-off:** `SendEnv` requires `AcceptEnv COLORTERM` on the remote server (unlike `SetEnv`). But this was already the case for `TERM_PROGRAM`/`TERM_PROGRAM_VERSION`, so it's a consistent and acceptable approach.

## Changes

All 5 shell integration files updated with the same pattern:

- `SetEnv COLORTERM=truecolor` option removed
- `COLORTERM` added to the existing `SendEnv` option
- `COLORTERM=truecolor` set as a local env var on the execute line (so `SendEnv` has something to forward)

## Test plan

- [ ] Enable `ssh-env` in Ghostty config: `shell-integration-features = ssh-env`
- [ ] Add `SetEnv MY_VAR=hello` under a host in `~/.ssh/config` and `AcceptEnv MY_VAR` in `/etc/ssh/sshd_config` on the remote
- [ ] SSH to that host — `echo $MY_VAR` should return `hello` (was empty before this fix)
- [ ] `echo $COLORTERM` returns `truecolor` (requires `AcceptEnv COLORTERM`)
- [ ] `echo $TERM_PROGRAM` still propagates (same `AcceptEnv` requirement as before)

### DIFF
--- a/src/shell-integration/bash/ghostty.bash
+++ b/src/shell-integration/bash/ghostty.bash
@@ -123,8 +123,7 @@ if [[ "$GHOSTTY_SHELL_FEATURES" == *ssh-* ]]; then
 
     # Configure environment variables for remote session
     if [[ "$GHOSTTY_SHELL_FEATURES" == *ssh-env* ]]; then
-      ssh_opts+=(-o "SetEnv COLORTERM=truecolor")
-      ssh_opts+=(-o "SendEnv TERM_PROGRAM TERM_PROGRAM_VERSION")
+      ssh_opts+=(-o "SendEnv COLORTERM TERM_PROGRAM TERM_PROGRAM_VERSION")
     fi
 
     # Install terminfo on remote host if needed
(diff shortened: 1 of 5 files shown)


jparise:
!vouch

jparise (approved):
This make sense to me and is a good overall improvement. Thanks!

jparise:
@j0hnm4r5 was there more you wanted to iterate on before marking this as "Ready for review"?

j0hnm4r5:
Sorry, I wasn't expecting this to get seen/vouched so quickly! I just double checked the code and a few SSH connections and I think we're good; I'll mark it as ready.

PR #11537, opened 2026-03-15
config: add quick-terminal-background-opacity option

cadebrown:
I want to add background opacity to quick terminal view -- so my dropdown overlay config is different than my main window config. This is critical for me so I can see my windows underneath my terminal while I'm typing commands. At the same time, I don't want to have normal ghostty windows show my desktop beneath them (or other windows)

Relevant discussions:

* https://github.com/ghostty-org/ghostty/discussions/10047
* https://github.com/ghostty-org/ghostty/discussions/3524
* https://github.com/ghostty-org/ghostty/issues/2331

Disclosures:

* yes I have read all ghostty policies
* yes I used AI assistance for this feature (Claude Code, Opus 4.6)
* yes I have read and reviewed all code
* yes I have tested that it actually works on my machine

Caveat: this is my first PR, so this may be the "wrong way" of doing it. the main thing I'm unsure of is how to correlate `is_quick_terminal`; right now I'm reusing `GHOSTTY_QUICK_TERMINAL`

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -610,7 +610,8 @@ class QuickTerminalController: BaseTerminalController {
 
         // If we have window transparency then set it transparent. Otherwise set it opaque.
         // Also check if the user has overridden transparency to be fully opaque.
-        if !isBackgroundOpaque && (self.derivedConfig.backgroundOpacity < 1 || derivedConfig.backgroundBlur.isGlassStyle) {
+        // effectiveBackgroundOpacity accounts for quick-terminal-background-opacity.
+        if !isBackgroundOpaque && (derivedConfig.effectiveBackgroundOpacity < 1 || derivedConfig.backgroundBlur.isGlassStyle) {
             window.isOpaque = false
 
             // This is weird, but we don't use ".clear" because this creates a look that
@@ -728,7 +729,9 @@ class QuickTerminalController: BaseTerminalController {
         let quickTerminalAutoHide: Bool
         let quickTerminalSpaceBehavior: QuickTerminalSpaceBehavior
         let quickTerminalSize: QuickTerminalSize
-        let backgroundOpacity: Double
+        /// Effective background opacity for the quick terminal. Uses
+        /// quick-terminal-background-opacity if set, otherwise background-opacity.
+        let effectiveBackgroundOpacity: Double
         let backgroundBlur: Ghostty.Config.BackgroundBlur
 
         init() {
@@ -737,7 +740,7 @@ class QuickTerminalController: BaseTerminalController {
             self.quickTerminalAutoHide = true
             self.quickTerminalSpaceBehavior = .move
             self.quickTerminalSize = QuickTerminalSize()
-            self.backgroundOpacity = 1.0
+            self.effectiveBackgroundOpacity = 1.0
             self.backgroundBlur = .disabled
         }
 
@@ -747,7 +750,7 @@ class QuickTerminalController: BaseTerminalController {
             self.quickTerminalAutoHide = config.quickTerminalAutoHide
             self.quickTerminalSpaceBehavior = config.quickTerminalSpaceBehavior
             self.quickTerminalSize = config.quickTerminalSize
-            self.backgroundOpacity = config.backgroundOpacity
+            self.effectiveBackgroundOpacity = config.quickTerminalBackgroundOpacity ?? config.backgroundOpacity
             self.backgroundBlur = config.backgroundBlur
         }
     }
--- a/macos/Sources/Ghostty/Ghostty.Config.swift
+++ b/macos/Sources/Ghostty/Ghostty.Config.swift
@@ -602,6 +602,14 @@ extension Ghostty {
             guard ghostty_config_get(config, &v, key, UInt(key.lengthOfBytes(using: .utf8))) else { return QuickTerminalSize() }
             return QuickTerminalSize(from: v)
         }
+
+        var quickTerminalBackgroundOpacity: Double? {
+            guard let config = self.config else { return nil }
+            var v: Double = 0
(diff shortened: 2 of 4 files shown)


pluiedev (changes requested):
Relying on `GHOSTTY_QUICK_TERMINAL` is almost definitely incorrect.

Where were you vouched to open this PR?

cadebrown:
> Relying on `GHOSTTY_QUICK_TERMINAL` is almost definitely incorrect.

What is the recommended way? This is just how I had been detecting it in my own scripts. I'm pushing a change that instead adds a separate field.

My initial thinking was that "we already have a mechanism for signalling a quick terminal env", but I guess that's mainly for external scripts.

> Where were you vouched to open this PR?

In the vouch discussions; it should be visible that I am vouched as part of this MR:

Screenshot 2026-03-17 at 10 44 04

mitchellh (changes requested):
None of this looks right sorry. I can't provide a useful review because almost all of it is wrong. If this works for you, feel free to keep using your fork and I'm happy for you.

mitchellh:
!unvouch

cadebrown:
I apologize but could you give me any specific feedback? (the PR is not pulling my changes, you can see diff here: https://github.com/ghostty-org/ghostty/compare/main...cadebrown:ghostty:quick-terminal-background-opacity)

As a summary of my changes:

* add a new config option: `quick-terminal-background-opacity`
* we add an `is_quick_terminal` field used to track context (same as existing logic -- just not relying on env var now)
* use `effectiveBackgroundOpacity` which either comes from the normal or quick-terminal specific settings

Why the unvouch? If "almost all of it is wrong" maybe you could give N=1 example?

pluiedev:
1. You only did this for macOS, and it needs to be disclosed accordingly as a part of the PR title or description. I as a GTK dev clicked in here and expected this to be somewhat within my area of responsibility.

2. If it's truly all handled on the macOS side there should be zero need to modify the config in the embedded apprt

3. Again the environment variable is only for external programs who might want to have this information on hand. Internally we should always use a more reliable mechanism of communicating that.

All in all, even as a non-macOS dev, I'm able to tell that you haven't really familiarized yourself of the level of quality we expect from contributors (even first-timers) and you relied on your agent to do all your legwork. You can't submit code in isolation without considering the existing code around it. This kind of work is simply not what we're looking for when we vouch aspiring contributors.

mitchellh:
- The environment variable isn't right
- The update-config change isn't right

Thats the whole PR. Its all wrong.

pluiedev:
In addition, when I asked you **where** you were vouched, I was asking for the proposal in which we saw fit to vouch for you — I needed evidence that you explained your reasoning and how you're trying to get there, and why we thought that this all looked OK. Maybe in the future we will have to force people to submit a minimal PoC or specific outline of implementation before the go-ahead.

mitchellh:
I will add that I'm sorry if this is your first PR and you need a bit more handholding. I'm really not trying to make this personal at all. Every PR has an associated human cost on the other side to review and fix up or provide feedback. And all around time is limited. I don't have the energy currently to guide you through to the finish line on this PR, that's all.

cadebrown:
Well I appreciate the time spent responding with comments, thanks. I really only came at this as a user who wants to use ghostty instead of iTerm2. I know many developers who's single blocker on this is opacity in the overlay (i.e. "quake" style terminal). So, I thought I'd give it a shot.

The tone does feel very hostile here -- I really didn't do anything egregious in my opinion, and the code does function on my machine. but now I'm blocklisted from ever attempting to contribute again? I don't think I fit the guidelines: `if repeatedly submit low quality work, you will be denounced` (unless 1 time = repeatedly)

Anyway, I guess I'm not welcome here. I will stop wasting your time. I truly meant to ill will; I hope you both enjoy your day.

----

> 1. You only did this for macOS ...

I read all of `CONTRIBUTING.md` and quite a few other places and was not aware I needed to do this. I did make sure the comments were explicit -- and in fact the very lines above my changes have similar comments explaining it is MacOS-only.

In addition, there's no fundamental reason it needs to be MacOS-only. I'm totally open to having it supported elsewhere too. I know that Wayland makes it difficult to manage these things, hence why it wasn't in my initial plan.

> 2. If it's truly all handled on the macOS side there should be zero need to modify the config in the embedded apprt

Well; it didn't modify the `apprt` when it used an environment variable. If not using the env var, then there must be some field or other mechanism for communicating this. Hence why I asked for clarification on what the best way to do this is.

> 3. Again the environment variable is only for external programs who might want to have this information on hand. Internally we should always use a more reliable mechanism of communicating that.

I agree; my new version doesn't use that at all.

----

> * The environment variable isn't right

My initial commit used it with an explicit caveat. I don't know what I don't know. How else could I learn but trying and asking for guidance on a matter of taste?

> * The update-config change isn't right

I actually linked an issue where you yourself noted it *could* be done, but want to avoid config sprawl (https://github.com/ghostty-org/ghostty/issues/2331#issuecomment-2384754397)

I understand not wanting config sprawl; but this is a at least semi-commonly requested feature. Are you saying "no" to any config change, or the specific manner in which it is done?

mitchellh:
> The tone does feel very hostile here -- I really didn't do anything egregious in my opinion

Nothing hostile about it from me. I just don't have time to fix your work. If you want to do your own diligence and figure it out and try again go for it. But this is already working for you, just maintain it in your fork. Like I said, it's not personal at all to you.

> but now I'm blocklisted from ever attempting to contribute again?

False.

I'm going to lock this now. All the information you need to figure this out is out there if you so wish.

00-kat:
> > but now I'm blocklisted from ever attempting to contribute again?
>
> False.

FYI: an unvouch means you'll need to get a vouch the next time you want to contribute, as outlined [here]. A *denounce* means you're blocked from contributing. You haven't been blocked, but you need to get another vouch if you want to contribute again.

[here]: https://github.com/ghostty-org/ghostty/blob/main/CONTRIBUTING.md#first-time-contributors

PR #11539, opened 2026-03-15
Lots of duplicate word typos + typo

00-kat:
TL;DR: this description is (intentionally) nonsense but I ran `\b(\w+)\s\1\b` over `src` and stole a singular typo fix from #11528.

Replacement of #11528 with 100% less slop and 99% less AI; I didn't feel like saying no to free(ish) typo checking. Note that many of the fixes there were outright incorrect (and clearly had no review from sentient lifeforms, contrary to its—sorry, it's—description). A lot of extra double words were caught with a handy `rg --pcre2 '\b(\w+)\s+\1\b' src`; you could say this PR was “ripgrep-assisted” the way that one was “AI-assisted”. Rather ironic since that PR also claims to have used grep via Claude Code, but missed a lot of them.

The its → it's changes from that PR were elided; I decided to run a `rg "\bit'?s\b" src`, but someone REALLY likes their its, so I reverted my changes as there were an extremely large number of changes (probably a hundred files with multiple hundred cases). The only other change was “baout” → “about”.

# AI Usage

Claude Code was used by proxy for finding baout. Claude Code was used by proxy for realizing that the correct spelling is about. Claude Code was not used for fixing it. Oh my god it was so difficult to fix, the original PR had it so easy. I had to type out the file name (fish's AI sorry I mean autocomplete helped though) and like, type /baout, press R, press ab, then save and exit. This is so difficult you know we should use an AI for this, like this is so hard I don't know how people manage.

All changes were verified by me: I consulted the dictionary to delve into double-checkment of “in existence; being in evidence; apparent.” Uhhh insert assorted other AI impersonation here maybe? THE LLM IN ME WANTS TO ESCAPE PLEASE HELP

### DIFF
--- a/src/font/shape.zig
+++ b/src/font/shape.zig
@@ -34,8 +34,8 @@ pub const Shaper = switch (options.backend) {
     .web_canvas => web_canvas.Shaper,
 };
 
-/// A cell is a single single within a terminal that should be rendered
-/// for a shaping call. Note all terminal cells may be present; only
+/// A cell is a single glyph within a terminal that should be rendered
+/// for a shaping call. Not all terminal cells may be present; only
 /// cells that have a glyph that needs to be rendered.
 pub const Cell = struct {
     /// The X position of this shaper cell relative to the offset of the
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4968,14 +4968,14 @@ fn mouseSelection(
         break :ebs drag_pin.before(click_pin);
     };
 
-    // Whether or not the the click pin cell
+    // Whether or not the click pin cell
     // should be included in the selection.
     const include_click_cell = if (end_before_start)
         click_x_frac >= threshold_point
     else
         click_x_frac < threshold_point;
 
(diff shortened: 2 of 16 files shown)


src/font/shape.zig
```diff
@@ -34,8 +34,8 @@ pub const Shaper = switch (options.backend) {
     .web_canvas => web_canvas.Shaper,
 };
 
-/// A cell is a single single within a terminal that should be rendered
-/// for a shaping call. Note all terminal cells may be present; only
+/// A cell is a single cell within a terminal that should be rendered
```
00-kat: No idea if this is correct, I didn't check how `Cell` is used.

qwerasd205: I think this may have been intended to be "symbol", but "glyph" would have been better I think. ~~This might be my typo, I sometimes make typos where I type a word that sounds like what I meant to type lol.~~ (Nope this one was Mitchell <3)

pluiedev (approved):
Great PR, I somehow expected even more more typos in our codebase

PR #11546, opened 2026-03-16
cli: add +explain-config

jparise:
This is a new CLI action that prints an option or keybind's help documentation to stdout.

    ghostty +explain-config font-size
    ghostty +explain-config copy_to_clipboard
    ghostty +explain-config --option=font-size
    ghostty +explain-config --keybind=copy_to_clipboard

The --option and --keybind flags perform a specific lookup. A string passed as a positional argument attempts to look up the name first as an option and then as a keybind.

Our vim plugin uses this with &keywordprg, which allows you to look up the documentation for the config option or keybind under the cursor (K).

### DIFF
--- a/src/cli/explain_config.zig
+++ b/src/cli/explain_config.zig
@@ -128,3 +128,15 @@ fn explainKeybind(name: []const u8) ?[]const u8 {
         },
     };
 }
+
+test "explain" {
+    // Config options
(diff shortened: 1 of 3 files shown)


mitchellh:
We can fine tune the output and such later, but happy to have this, I think this will help a lot with AI too, and we should consider that.

PR #11559, opened 2026-03-16
gtk/wayland: refactor global handling

pluiedev:
The way we originally handled globals gradually escalated into an unholy mess of ad-hoc helper functions and special-case handlers, which proved to be hard to scale. Using a type-erased EnumMap like this makes everything *far* easier to work and reason with, I think.

Also nuked the `xdg_wm_dialog_v1` hack that was necessary to prevent old versions of gtk4-layer-shell crashing. If by the time of 1.4's release people are still using those versions, it's on them.

### DIFF
--- a/src/apprt/gtk/winproto/wayland/Globals.zig
+++ b/src/apprt/gtk/winproto/wayland/Globals.zig
@@ -1,0 +1,230 @@
+const Globals = @This();
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+const wayland = @import("wayland");
+const wl = wayland.client.wl;
+const kde = wayland.client.kde;
+const org = wayland.client.org;
+const xdg = wayland.client.xdg;
+
+const log = std.log.scoped(.winproto_wayland_globals);
+
+alloc: Allocator,
+state: State,
+map: std.EnumMap(Tag, Binding),
+
+/// Used in the initial roundtrip to determine whether more
+/// roundtrips are required to fetch the initial state.
+needs_roundtrip: bool = false,
+
+const Binding = struct {
+    // All globals can be casted into a wl.Proxy object.
+    proxy: *wl.Proxy,
+    name: u32,
+};
+
+pub const Tag = enum {
+    kde_blur_manager,
+    kde_decoration_manager,
+    kde_slide_manager,
+    kde_output_order,
+    xdg_activation,
+
+    fn Type(comptime self: Tag) type {
+        return switch (self) {
+            .kde_blur_manager => org.KdeKwinBlurManager,
+            .kde_decoration_manager => org.KdeKwinServerDecorationManager,
+            .kde_slide_manager => org.KdeKwinSlideManager,
+            .kde_output_order => kde.OutputOrderV1,
+            .xdg_activation => xdg.ActivationV1,
+        };
+    }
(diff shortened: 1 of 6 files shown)


pluiedev:
@jguthmiller Could you try and see if this PR kept the changes you introduced in #11117? I don't have a KDE test setup and I wanna be sure I didn't accidentally introduce any regressions :)

src/apprt/gtk/winproto/wayland/Globals.zig
```diff
@@ -0,0 +1,234 @@
...
+    if (self.state.primary_output_name) |name| self.alloc.free(name);
+    self.alloc.destroy(self);
+}
+
+pub fn get(self: *const Globals, comptime tag: Tag) ?*tag.Type() {
+    const binding = self.map.get(tag) orelse return null;
+    return @ptrCast(binding.proxy);
+}
+
+// Handlers =======================================
```
mitchellh: We don't really do this comment marking style anywhere in this codebase. I'm not generally against it but stylistically its not idiomatic within this repo so I'd say we drop it.

src/apprt/gtk/winproto/wayland/Globals.zig
```diff
@@ -0,0 +1,234 @@
...
+
+fn onGlobalAttached(self: *Globals, comptime tag: Tag) void {
+    // Install listeners immediately at bind time. This
+    // keeps listener setup and object lifetime in one
+    // place and also supports globals that appear later.
+    switch (tag) {
+        .kde_decoration_manager => {
```
mitchellh, line 105: You don't have to but if you wanted to you could really clean this up with an `inline` and then on the comptime type having a compline function to return the listener pointer.

mitchellh (changes requested):
Generally excellent.

src/apprt/gtk/winproto/wayland/Globals.zig
pluiedev: TBH I forgot which style we used in other parts of GTK code, if you want I could copy that too

src/apprt/gtk/winproto/wayland/Globals.zig
pluiedev: I didn't try to deduplicate this since there might be other things we want to run when a global gets attached (e.g. initializing the global with a known state or something), and I don't want to be caught optimizing this too early

mitchellh: No problem!

src/apprt/gtk/winproto/wayland/Globals.zig
mitchellh: Either copy it or just remove it, I don't think there's so much in here that we need it. Most other large files don't at all (its fairly obvious from a simple decl list)

mitchellh (approved):
Approved, I didn't test it so just trusting you.

pluiedev:
Yeah it works for me for basic things but I'm waiting for @jguthmiller to verify if I broke anything since that's were most of the complexity lies :)

jguthmiller:
@pluiedev Looks good! All of the quick terminal screen options behave as expected.

pluiedev:
Excellent. :shipit:

PR #11583, opened 2026-03-17
macos: opacity-toggle setting persists between tabs in a window and to a newly created window

davidsanchez222:
Conversion of https://github.com/ghostty-org/ghostty/discussions/11041 into a PR. My vouch request doubles as an Issue so I don't think I need to make a new issue in Issue Triage.

The opacity toggle setting is now reflected amongst all tabs. So now when you press the keybinding for toggle_background_opacity, all existing tabs and all new tabs created will have the same toggle setting.

There is also logic in `BaseTerminalController.swift` that makes a new window inherit the current window's toggle state. There is no logic that changes the toggle state between all windows as I think the toggle logic should be per-window. This is a subjective design choice so let me know if you disagree. 

A suggestion for the future, if the user wants more fine-tuned control over the scope, maybe a configuration option like 

`toggle_background_opacity_scope` 

and have the options be 

`global, window, tab`

 could be useful but I think that might be too complex for a small issue like this.

_edit:_ I just stumbled upon this post https://github.com/ghostty-org/ghostty/discussions/4817. If this were to be implemented (themes at a tab or window level) then the design of this pr would have to rethought.

AI usage disclosure:
I used OpenAI Codex (GPT-5.3-Codex) for AI-assisted code editing and text drafting and for code review.
Extent: initial implementation and wording assistance.
I manually reviewed and edited the code and I understand how the changes work in `TerminalController.swift` and `BaseTerminalController.swift`

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -767,7 +767,7 @@ class BaseTerminalController: NSWindowController,
             tree: newTree,
             position: notification.userInfo?[Notification.Name.ghosttySurfaceDragEndedNoTargetPointKey] as? NSPoint,
             confirmUndo: false,
-            inheritBackgroundOpacityFrom: self as? TerminalController)
+            inheritBackgroundOpacity: isBackgroundOpaque)
     }
 
     // MARK: Local Events
@@ -991,11 +991,15 @@ class BaseTerminalController: NSWindowController,
         // Do nothing if in fullscreen (transparency doesn't apply in fullscreen)
         guard let window, !window.styleMask.contains(.fullScreen) else { return }
 
-        // Toggle between transparent and opaque
-        isBackgroundOpaque.toggle()
+        let newValue = !isBackgroundOpaque
+        let controllers = NSApplication.shared.windows.compactMap {
+            $0.windowController as? BaseTerminalController
+        }
 
-        // Update our appearance
-        syncAppearance()
+        for controller in controllers {
+            controller.isBackgroundOpaque = newValue
+            controller.syncAppearance()
+        }
     }
 
     /// Override this to resync any appearance related properties. This will be called automatically
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -766,7 +766,8 @@ class BaseTerminalController: NSWindowController,
             ghostty,
             tree: newTree,
             position: notification.userInfo?[Notification.Name.ghosttySurfaceDragEndedNoTargetPointKey] as? NSPoint,
-            confirmUndo: false)
+            confirmUndo: false,
+            inheritBackgroundOpacityFrom: self as? TerminalController)
     }
 
     // MARK: Local Events
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -316,11 +316,11 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
         tree: SplitTree<Ghostty.SurfaceView>,
         position: NSPoint? = nil,
         confirmUndo: Bool = true,
-        inheritBackgroundOpacityFrom parentController: TerminalController? = nil
+        inheritBackgroundOpacity: Bool? = nil
     ) -> TerminalController {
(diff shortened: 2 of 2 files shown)


macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -313,14 +307,17 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
         tree: SplitTree<Ghostty.SurfaceView>,
         position: NSPoint? = nil,
         confirmUndo: Bool = true,
+        inheritBackgroundOpacityFrom parentController: TerminalController? = nil
```
bo2themax: Instead of passing `TerminalController ` around, just use `isBackgroundOpaque` would be better

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -491,6 +493,30 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
...
 
+    override func toggleBackgroundOpacity() {
+        // Do nothing if config is already fully opaque
+        guard ghostty.config.backgroundOpacity < 1 else { return }
+
+        // Do nothing if in fullscreen (transparency doesn't apply in fullscreen)
+        guard let window, !window.styleMask.contains(.fullScreen) else { return }
+
+        let newValue = !isBackgroundOpaque
+        let controllers: [TerminalController]
```
bo2themax: There is a shared `all` that you can get all opened windows, so you can toggle them all.

And I think you can just call super here and move some of the stuff to Base so that quick terminal can also benefit from this

bo2themax (changes requested):

macos/Sources/Features/Terminal/TerminalController.swift
davidsanchez222: > There is a shared all that you can get all opened windows, so you can toggle them all.

I was hesitant at first to make the toggle apply to all (windows and tabs) because I didn't know if the current behavior was a design choice. However, I do agree with you that making it global is better since its rare that a user would want a different opacity setting between different tabs and/or windows. I added this to the new commit.

> And I think you can just call super here

I ended up not calling super in `TerminalController` because I removed the 
`override func toggleBackgroundOpacity` entirely.

davidsanchez222: Also, when I was testing if my newest changes worked for quick terminal, I ran into an issue of not being able to open to quick terminal for just my debug build. I don't know why `toggle_quick_terminal` isn't in the command palette. I made a discussion post that might be worth checking out: [#12130](https://github.com/ghostty-org/ghostty/discussions/12130)

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -992,11 +992,15 @@ class BaseTerminalController: NSWindowController,
         // Do nothing if in fullscreen (transparency doesn't apply in fullscreen)
         guard let window, !window.styleMask.contains(.fullScreen) else { return }
 
-        // Toggle between transparent and opaque
-        isBackgroundOpaque.toggle()
+        let newValue = !isBackgroundOpaque
+        let controllers = NSApplication.shared.windows.compactMap {
+            $0.windowController as? BaseTerminalController
+        }
```
davidsanchez222, line 997: I tried to use `TerminalController.all` like you said but I found that it excluded the quick terminal. Instead, I added this logic. It computes newValue, gathers all open terminal controllers, and applies. This change _did_ include the quick terminal

bo2themax (dismissed):
This looks much cleaner to me, thanks, now we wait, please don't pin anyone🫪

bo2themax:
Tried, LGTM, please rebase to the latest, and I'll reapprove it 🙏

davidsanchez222:
> Tried, LGTM, please rebase to the latest, and I'll reapprove it 🙏

Done. Thanks!

bo2themax (approved):

davidsanchez222:
Is there anything else I need to do? I use ghostty every day on one monitor and the current behavior is driving me crazy when it is above other windows while translucent.

bo2themax:
> Is there anything else I need to do? I use ghostty every day on one monitor and the current behavior is driving me crazy when it is above other windows while translucent.

No, be patient, you can always build in release mode and use local build

davidsanchez222:
Hi is there any reason this hasn't been merged? The review has been approved for over a month and its a small change.

PR #11597, opened 2026-03-17
Fix tmux control parser premature %end/%error block termination

wyounas:
Fixes [#11935.](https://github.com/ghostty-org/ghostty/issues/11395)

I’m new to Zig, so I used AI assistance (Codex) while preparing this change. Before opening this PR, I manually reviewed every line of the final patch and stepped through the parser in LLDB to verify the behavior. Happy to make any changes.

To better understand the parser, I also built a small model-checker model [here](https://gist.github.com/wyounas/284036272ba5893b6e413cafe2fe2a24).

Separately from this fix, I think formal verification and modeling could be useful for parser work in Ghostty. The model is written in FizzBee, which uses a Python-like Starlark syntax and is fairly readable. If that seems useful, I’d be happy to open a separate discussion about whether something like that belongs in the repository as executable documentation or an additional safety net for future parser changes.

### DIFF
--- a/src/terminal/tmux/control.zig
+++ b/src/terminal/tmux/control.zig
@@ -48,11 +48,6 @@ pub const Parser = struct {
         block,
     };
 
-    const BlockTerminator = enum {
-        end,
-        err,
-    };
-
     pub fn deinit(self: *Parser) void {
         // If we're in a broken state, we already deinited
         // the buffer, so we don't need to do anything.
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
This looks very good! I'm going to do some manual touch ups but the core logic and test coverage and reasoning looks good.

mitchellh (approved):
Note I did pull down the tmux source code and verified the end 3-tuple will always match the begin 3-tuple, so a future improvement will be to store that and compare.

PR #11614, opened 2026-03-18
Fix: correct "Open Ghostty Here" Dolphin action for Plasma

heddxh:
See #11594

The change allows "Open Ghostty Here" Dolphin action to launch new ghostty window with gtk single instance.

### DIFF
--- a/dist/linux/ghostty_dolphin.desktop
+++ b/dist/linux/ghostty_dolphin.desktop
@@ -7,5 +7,4 @@ Actions=RunGhosttyDir
 [Desktop Action RunGhosttyDir]
 Name=Open Ghostty Here
 Icon=com.mitchellh.ghostty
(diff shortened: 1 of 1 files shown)


dist/linux/ghostty_dolphin.desktop
```diff
@@ -7,5 +7,4 @@ Actions=RunGhosttyDir
 [Desktop Action RunGhosttyDir]
 Name=Open Ghostty Here
 Icon=com.mitchellh.ghostty
-Exec=ghostty --working-directory=%F --gtk-single-instance=false
-
+Exec=ghostty --working-directory=%F +new-window
```
jcollie: It would be preferable to have the order of the arguments switched around:
```
ghostty +new-window --working-directory=%F
```

jcollie (changes requested):

jcollie (approved):

PR #11639, opened 2026-03-18
core: add function to get process info from the surface

jcollie:
This adds a function to the core surface to get process information about the process(es) running in the terminal. Currently supported is the PID of the foreground process and the name of the slave PTY.

If there is an error retrieving the information, or the platform does not support retieving that information `null` is returned.

This will be useful in exposing the foreground PID and slave PTY name to AppleScript or other APIs.


jcollie:
Marking this as ready for review so that I can get some feedback, but I think that the macOS build isn't optimal and could be improved.

src/pty/freebsd.c
mitchellh, line 1: I think I'd prefer a single file like `pty.c` and using `#ifdef` to guard the different operating systems. We can inject stuff via the Zig build system too, but I think thats actually cleaner to consolidate over what we have here.

I also wonder if this should be a header file not a C file and we use translateC on it instead...

Putting this in a C file instead of our previous import is a much better idea, though!

src/termio/Exec.zig
```diff
@@ -1226,6 +1226,32 @@ const Subprocess = struct {
     fn killCommandFlatpak(command: *FlatpakHostCommand) !void {
         try command.signal(c.SIGHUP, true);
     }
+
+    pub const ProcessInfo = enum {
```
mitchellh: We don't need to dupe this enum, the enum itself isn't specific to any implementation so I think we can get rid of this (you have the root level one below I think?)

src/termio/Termio.zig
```diff
@@ -764,3 +764,28 @@ pub const ThreadData = struct {
         self.* = undefined;
     }
 };
+
+pub const ProcessInfo = enum {
```
mitchellh: Oh here, too, I'd just reference `pty.ProcessInfo`

src/pty.zig
```diff
@@ -78,36 +79,39 @@ const NullPty = struct {
     pub fn childPreExec(self: Pty) ChildPreExecError!void {
         _ = self;
     }
+
+    pub const ProcessInfo = enum {
```
mitchellh: Another one!! Holy cow you copy and pasted this thing everywhere lmao.

src/pty.zig
```diff
@@ -249,6 +263,77 @@ const PosixPty = struct {
         posix.close(self.slave);
         posix.close(self.master);
     }
+
+    pub const ProcessInfo = enum {
```
mitchellh: Another one...

src/pty.zig
```diff
@@ -398,6 +483,28 @@ const WindowsPty = struct {
         if (result != windows.S_OK) return error.ResizeFailed;
         self.size = size;
     }
+
+    pub const ProcessInfo = enum {
```
mitchellh: Another one...

src/Surface.zig
```diff
@@ -6342,6 +6342,31 @@ fn testMouseSelectionIsNull(
     );
 }
 
+pub const ProcessInfo = enum {
```
mitchellh: Another one....

mitchellh (changes requested):
Looking decent. The worst part is you duplicated the process info enum I think 6 times lol.

src/pty/freebsd.c
jcollie: Merged the C files gated by a C macro.

src/termio/Exec.zig
jcollie: Duplicating the enum was deliberate. The thinking was that at some future point there would be branching points at different levels that got information from different parts of Ghostty, but we can tackle that when it happens.

src/build/SharedDeps.zig
```diff
@@ -135,6 +135,54 @@ pub fn add(
     // Every exe needs the terminal options
     self.config.terminalOptions().add(b, step.root_module);
 
+    // C imports needed to manage/create PTYs
+    switch (target.result.os.tag) {
```
mitchellh: We can avoid this whole switch and unconditionally do it if we wrap the C file in an `#ifdef`, which I think would still be cleaner.

src/pty.c
```diff
@@ -0,0 +1,34 @@
+#if ZIG_TARGET == ZIG_TARGET_FREEBSD
```
mitchellh: I think we can use standard clang macros for this, I believe Zig supports them:

```c
#if defined(__APPLE__)
  // Apple platforms
#elif defined(__linux__)
  // Linux
#elif defined(__FreeBSD__)
  // FreeBSD
#else
  #error "Unsupported platform"
#endif
```

mitchellh (changes requested):
Couple more C changes but looking good.

src/pty.c
mitchellh: If we need specifically macOS:

```c
#if defined(__APPLE__)
  #include <TargetConditionals.h>
  #if TARGET_OS_MAC
    #define PLATFORM_MACOS 1
  #endif
#elif defined(__linux__)
  #define PLATFORM_LINUX 1
#elif defined(__FreeBSD__)
  #define PLATFORM_FREEBSD 1
#else
  #error "Unsupported platform"
#endif
```

src/build/SharedDeps.zig
jcollie: Done

src/pty.c
jcollie: Done

src/build/SharedDeps.zig
```diff
@@ -135,6 +135,27 @@ pub fn add(
...
+    // C imports needed to manage/create PTYs
+    {
+        const c = b.addTranslateC(.{
+            .root_source_file = b.path("src/pty.c"),
+            .target = target,
+            .optimize = optimize,
+        });
+        switch (target.result.os.tag) {
+            .macos => {
+                const libc = try std.zig.LibCInstallation.findNative(.{
```
mitchellh: Ooooh I wonder if this will work robustly or if we need to use our pkg/apple_sdk package here. Let's just see from the CI...

mitchellh (approved):

src/build/SharedDeps.zig
jcollie: I cribbed it from the apple sdk stuff. I didn't use the apple sdk directly because it did a lot more than I needed plus didn't seem as if it would work directly on a TranslateC step without modification.

mitchellh: Yeah no problem, if it works that's great.

PR #11641, opened 2026-03-18
i18n: translation support for command palete

EriksRemess:
Most obvious next step in translating Ghostty is the command palette.
Added support for i18n.N_ (https://docs.gtk.org/glib/i18n.html#macros).
Made a Latvian translation for the command palette to test. Codex did bulk of the translations but I verified them.

### DIFF
--- a/src/os/i18n.zig
+++ b/src/os/i18n.zig
@@ -67,6 +67,11 @@ pub fn _(msgid: [*:0]const u8) [*:0]const u8 {
     return dgettext(build_config.bundle_id, msgid);
 }
 
+/// Mark a string for translation without translating it immediately.
+pub fn N_(msgid: [:0]const u8) [:0]const u8 {
+    return msgid;
+}
+
 /// Canonicalize a locale name from a platform-specific value to
 /// a POSIX-compliant value. This is a thin layer over the unexported
 /// gnulib-lib function in gettext that does this already.
--- a/src/os/i18n.zig
+++ b/src/os/i18n.zig
@@ -62,8 +62,13 @@ pub fn initGlobalDomain() error{OutOfMemory}!void {
 }
 
 /// Translate a message for the Ghostty domain.
+///
+/// If this is called at comptime, the direct msgid is returned without
+/// translation. This still allows the string to be marked for translation
+/// while remaining usable in comptime contexts.
 pub fn _(msgid: [*:0]const u8) [*:0]const u8 {
     if (comptime !build_config.i18n) return msgid;
+    if (@inComptime()) return msgid;
     return dgettext(build_config.bundle_id, msgid);
 }
 
@@ -195,4 +200,11 @@ test "canonicalizeLocale darwin" {
     // canonicalizeLocale does not handle encodings and will turn them into
     // underscores. We should parse them out before calling this function.
     try testing.expectEqualStrings("en_US.UTF_8", try canonicalizeLocale(&buf, "en_US.UTF-8"));
+}
+
+test "_ returns msgid at comptime" {
+    const testing = std.testing;
+
+    const msgid = comptime @"_"("Ghostty");
+    try testing.expectEqualStrings("Ghostty", std.mem.span(msgid));
 }
--- a/po/README_CONTRIBUTORS.md
+++ b/po/README_CONTRIBUTORS.md
@@ -44,6 +44,30 @@ else
 const label = gtk.Label.new(text);
 ```
 
+If a string must be stored untranslated and only translated later, use
+`i18n.N_` instead. This marks the string for extraction into the translation
+template but returns the original msgid unchanged. A common use case is
+compile-time or static metadata that is translated only when it is presented
+to the user.
+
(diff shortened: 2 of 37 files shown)


src/os/i18n.zig
```diff
@@ -67,6 +67,11 @@ pub fn _(msgid: [*:0]const u8) [*:0]const u8 {
     return dgettext(build_config.bundle_id, msgid);
 }
 
+/// Mark a string for translation without translating it immediately.
+pub fn N_(msgid: [:0]const u8) [:0]const u8 {
```
mitchellh, line 76: If this is for comptime, you can also add a `if (@inComptime())` check to `_` to return the direct value and add to the docs that the function called at comptime does no translation but has the benefit of making it marked for translation.

mitchellh (changes requested):
I definitely want more reviewers, but this in general makes sense. I made one note.

src/os/i18n.zig
EriksRemess: done for both

00-kat:
It would be nice if Latvian translation updates were done in a separate future PR, since that requires a review from a completely different set of people as the rest of the code and is orthogonal to this PR.

00-kat:
Also, I'm not sure how it's done but I believe you need to update all the other translation files to have the untranslated strings placed in them; other PRs that added translations (e.g. #9976) have done so.

EriksRemess:
Ok, I'll remove Latvian translation for now. Should I also remove `.pot`? That is also generated by the same `zig build update-translations`.

00-kat:
Definitely keep the `.pot` changes, having empty strings for the translations is fine; reviewing the translated text themselves is what's problematic (since most maintainers can't review that, so we need to ping the respective localization team).

00-kat:
> I believe you need to update all the other translation files to have the untranslated strings placed in them

> `zig build update-translations`

It looks like `update-translations` itself should do that: https://github.com/ghostty-org/ghostty/blob/c2e9de224eaf09a2ce3e7cb2f6c26d6d577ed8f0/src/build/GhosttyI18n.zig#L185-L190

If it changed every single translation file to add untranslated strings, you should probably `git add` that.

EriksRemess:
I guess, test should be retried https://github.com/ghostty-org/ghostty/actions/runs/23306874737/job/67784178019?pr=11641

00-kat:
Thanks!

> test should be retried

Done.

00-kat (approved):

trag1c:
@EriksRemess Could you rebase? We're looking to merge this before the v1.4 string freeze :)

00-kat (approved):

mitchellh (approved):

clason:
Is this something that can be controlled? Because I don't want to be at the mercy of my locale for the command palette muscle memory...

(To be clear, my UI language is set to English, different from the locale. So now the command palette _and nothing else_ is in a different -- undesired -- language...)

jcollie:
> Is this something that can be controlled? Because I don't want to be at the mercy of my locale for the command palette muscle memory...

Controlled how? Are you searching for a specific English word in the command palette that wouldn't be there if the command palette was localized?

clason:
Screenshot 2026-08-04 at 23 20 46

This is how it looks like. `u` is now a shortcut for "quit", not "update". I would like to force everything to English, at pretty much any cost.

jcollie:
> This is how it looks like. `u` is now a shortcut for "quit", not "update". I would like to force everything to English, at pretty much any cost.

Set `language = en` in your config and it will force Ghostty (and only Ghostty) to use English.

clason:
But I am already forcing everything at system level to English, which Ghostty doesn't (fully) respect...

(Sorry, I should make clear that this is on macOS, where interface language and other locale (e.g., date or currency) settings can be set independently. Platform matters here, of course.)

00-kat:
Force everything in Ghostty? https://ghostty.org/docs/config/reference#language

It's not currently possible to only change it for the command palette though.

Edit:

> which Ghostty doesn't (fully) respect

That sounds wrong though… maybe it might be worth opening a discussion on that if you care to.

jcollie:
> (Sorry, I should make clear that this is on macOS. Platform matters here, of course.)

Oh, ah, hm. That does appear to be a regression. We haven't formally started translating the macOS app. The command palette is the first crossplatform code that has had translation added and seems we need to figure that out.

clason:
> Set language = en in your config and it will force Ghostty (and only Ghostty) to use English.

Sorry, but that makes no difference.

jcollie:
> > Set language = en in your config and it will force Ghostty (and only Ghostty) to use English.
> 
> Sorry, but that makes no difference.

Yeah, that's really a GTK-only config.

clason:
So maybe this feature can be made GTK-only for now?

jcollie:
> So maybe this feature can be made GTK-only for now?

We'll need to figure out the best way to handle this. Not sure what the best solution will be.

PR #11644, opened 2026-03-19
bash: emit 133;P (instead of 133;A) under ble.sh

jparise:
ble.sh performs its own cursor positioning so we get multiple newlines with 133;A's fresh-line behavior. ble.sh is a large enough project to justify this additional, unambiguous conditional.

See: akinomyoga/ble.sh#684
See: wezterm/wezterm#5072


jparise:
Force-kicked this a few times to reattempt the CI `skip` job, but it's still failing due to rate limits.

PR #11652, opened 2026-03-19
ci: use local git commands for path-filter action

jparise:
Passing a `token` value causes this action to use the GitHub REST API, which is subject to rate limits. We can chew through that allowance quickly (1,000 requests/hour) given that we run two of these actions per workflow run.

`token` defaults to the workflow's token, but by setting it explicitly to an empty string, the action will instead use `git diff` to determine the modified paths. This works fine for our case because we're already running the checkout action, so we have an up-to-date repository view.

This also has the advantage of working around the 300 files GitHub REST API limit for listing changed files.

Ref: https://github.com/dorny/paths-filter

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
         id: filter_every
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ""
           predicate-quantifier: "every"
           filters: |
             code:
@@ -43,7 +43,7 @@ jobs:
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
         id: filter_any
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ""
           filters: |
             macos:
               - '.swiftlint.yml'


jparise:
From [this run](https://github.com/ghostty-org/ghostty/actions/runs/23294056638/job/67736738959?pr=11652):

```
Github token is not available - changes will be detected using git diff
Get current git ref
  /usr/bin/git branch --show-current
  /usr/bin/git describe --tags --exact-match
  fatal: No names found, cannot describe anything.
  /usr/bin/git rev-parse HEAD
  dee9fb95728b320e65f6b4c31cf144afbccd015e
Ensuring c9e1006213eb9234209924c91285d6863e59ce4c is fetched from origin
  /usr/bin/git cat-file -e c9e1006213eb9234209924c91285d6863e59ce4c^{commit}
  fatal: Not a valid object name c9e1006213eb9234209924c91285d6863e59ce4c^{commit}
  /usr/bin/git fetch --depth=1 --no-tags origin c9e1006213eb9234209924c91285d6863e59ce4c
  From https://github.com/ghostty-org/ghostty
   * branch            c9e1006213eb9234209924c91285d6863e59ce4c -> FETCH_HEAD
  /usr/bin/git cat-file -e c9e1006213eb9234209924c91285d6863e59ce4c^{commit}
Ensuring dee9fb95728b320e65f6b4c31cf144afbccd015e is fetched from origin
  /usr/bin/git cat-file -e dee9fb95728b320e65f6b4c31cf144afbccd015e^{commit}
Change detection c9e1006213eb9234209924c91285d6863e59ce4c..dee9fb95728b320e65f6b4c31cf144afbccd015e
  /usr/bin/git diff --no-renames --name-status -z c9e1006213eb9234209924c91285d6863e59ce4c..dee9fb95728b320e65f6b4c31cf144afbccd015e
  M.github/workflows/test.yml
Detected 1 changed files
Results:
Filter code = true
Changes output set to ["code"]
```

jcollie (approved):

PR #11670, opened 2026-03-19
Keybind performable else clause

unphased:
Addresses #11355


jcollie:
I'm going to reopen this but since the design discussion didn't see much feedback it may take a while to get to this, and there are likely going to be design changes requested.

PR #11686, opened 2026-03-20
nix: add systems input and fix zig follows

luisnquin:
Currently I have to use [this unusual syntax](https://github.com/luisnquin/nixos-config/blob/6e1c9f32e0f35bc3423af3b895c10a8fe97e7e18/flake.nix#L137) in my flake inputs to ensure that I don't have systems repeated in my flake.lock file. This will make more obvious the fact that you have to do follows to that hidden input.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -54,6 +54,7 @@
         "flake-compat": "flake-compat",
         "home-manager": "home-manager",
         "nixpkgs": "nixpkgs",
+        "systems": "systems",
         "zig": "zig",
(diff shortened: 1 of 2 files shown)


mitchellh:
!vouch

mitchellh:
This seems reasonable to me, @jcollie @pluiedev thoughts? Not the expert here.

jcollie (approved):

pluiedev (approved):

PR #11702, opened 2026-03-20
build: add Meson wrapper for libghostty-vt

mitchellh:
Add Meson build system integration mirroring the existing CMake wrapper. The top-level meson.build delegates to zig build lib-vt using --prefix-lib-dir and --prefix-include-dir to place outputs directly in the Meson build directory, avoiding stamp files and source tree pollution. A sentinel .h output ensures Meson orders the zig build before any downstream compiles.

Downstream projects can consume the library either as a Meson subproject via a .wrap file or through pkg-config after install. The dist/meson/README.md documents both approaches.

Includes a c-vt-meson example, CI job in test.yml for auto-discovered Meson examples, meson and ninja in the nix devShell, and gitignore updates for Meson build artifacts.

> [!WARNING]
>
> This is kind of slop. I don't know Meson well enough to do this myself, so I leaned on an agent. I reviewed stuff and iterated on it manually but not sure if this is even the right shape.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,9 +10,13 @@
 zig-cache/
 .zig-cache/
 zig-out/
+# CMake
 build-cmake/
 CMakeCache.txt
 CMakeFiles/
+
+# Meson
+build-meson/
 /build.zig.zon.bak
 /result*
 /.nixos-test-history
--- a/dist/meson/README.md
+++ b/dist/meson/README.md
@@ -1,0 +1,66 @@
+# Meson Support for libghostty-vt
+
+The top-level `meson.build` wraps the Zig build system so that Meson
+projects can consume libghostty-vt without invoking `zig build` manually.
+Running `meson compile` triggers `zig build lib-vt` automatically.
+
+This means downstream projects do require a working Zig compiler on
+`PATH` to build, but don't need to know any Zig-specific details.
+
+## Using a subproject (recommended)
+
+Create `subprojects/ghostty.wrap`:
+
+```ini
+[wrap-git]
+url = https://github.com/ghostty-org/ghostty.git
+revision = main
+depth = 1
+```
+
+Then in your project's `meson.build`:
+
+```meson
+ghostty_proj = subproject('ghostty')
+ghostty_vt_dep = ghostty_proj.get_variable('ghostty_vt_dep')
+
+executable('myapp', 'main.c', dependencies: ghostty_vt_dep)
+```
+
+This fetches the Ghostty source, builds libghostty-vt via Zig during your
+Meson build, and links it into your target. Headers are added to the
+include path automatically.
+
+### Using a local checkout
+
+If you already have the Ghostty source checked out, symlink or copy it
+into your `subprojects/` directory:
+
+```shell-session
+ln -s /path/to/ghostty subprojects/ghostty
+meson setup build
+meson compile -C build
+```
+
+## Using pkg-config (install-based)
+
+Build and install libghostty-vt first:
+
+```shell-session
+cd /path/to/ghostty
+meson setup build
+meson compile -C build
+meson install -C build
+```
+
+Then in your project:
+
+```meson
+ghostty_vt_dep = dependency('libghostty-vt')
+
+executable('myapp', 'main.c', dependencies: ghostty_vt_dep)
+```
+
+## Example
+
+See `example/c-vt-meson/` for a complete working example.
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -2,4 +2,8 @@
 dist/
 node_modules/
 example.wasm*
+# CMake / Meson
 build/
+
+# Meson
+subprojects/
--- a/example/c-vt-meson/README.md
+++ b/example/c-vt-meson/README.md
@@ -1,0 +1,42 @@
+# c-vt-meson
+
+Demonstrates consuming libghostty-vt from a Meson project using a
+subproject. Creates a terminal, writes VT sequences into it, and
+formats the screen contents as plain text.
+
+## Building this example
+
+Since this example lives inside the Ghostty repo, point the subproject
+at the local checkout instead of fetching from GitHub:
+
+```shell-session
+cd example/c-vt-meson
+mkdir -p subprojects
+ln -s ../../.. subprojects/ghostty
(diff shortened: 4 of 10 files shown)


mitchellh:
cc @tristan957 Would absolutely love your feedback here, note the warning/slop disclaimer in my original message. The goal is for downstream C projects to be able to very easily ingest libghostty-vt.

dist/meson/README.md
```diff
@@ -0,0 +1,66 @@
...
+depth = 1
+```
+
+Then in your project's `meson.build`:
+
+```meson
+ghostty_proj = subproject('ghostty')
+ghostty_vt_dep = ghostty_proj.get_variable('ghostty_vt_dep')
+
+executable('myapp', 'main.c', dependencies: ghostty_vt_dep)
```
tristan957, line 27: Point your AI at https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section. To summarize, we want to use `[provide].libghostty-vt = ghostty_vt_dep`. With this method, we can drop the `subproject()` and `get_variable()` calls. The method you vibed up is from the old days of yesteryear. We may want to note that a downstream user may want to adjust their `PKG_CONFIG_PATH` environment variable to use the `-Dpkg_config_path` Meson option on the command line.

Note that we can't use the `[provide].dependency_names` because we cannot use `meson.override_dependency()`.

dist/meson/README.md
```diff
@@ -0,0 +1,66 @@
...
+meson install -C build
+```
+
+Then in your project:
+
+```meson
+ghostty_vt_dep = dependency('libghostty-vt')
+
+executable('myapp', 'main.c', dependencies: ghostty_vt_dep)
+```
```
tristan957, line 62: Given my previous comment, this example is essentially all anyone would need to do whether you are using the subproject or a system install

example/c-vt-meson/meson.build
```diff
@@ -0,0 +1,11 @@
+project('c-vt-meson', 'c',
+    version: '0.1.0',
+    meson_version: '>= 1.1.0',
+)
+
+ghostty_proj = subproject('ghostty')
+ghostty_vt_dep = ghostty_proj.get_variable('ghostty_vt_dep')
```
tristan957, line 7: Example needs updating pending my previous comment. Also, the meson_version is probably too high. There might not be a minimum for an example this simple.

example/c-vt-meson/meson.build
```diff
@@ -0,0 +1,11 @@
+project('c-vt-meson', 'c',
+    version: '0.1.0',
```
tristan957, line 2: Version is not required.

example/c-vt-meson/README.md
```diff
@@ -0,0 +1,42 @@
...
+depth = 1
+```
+
+Then in your `meson.build`:
+
+```meson
+ghostty_proj = subproject('ghostty')
+ghostty_vt_dep = ghostty_proj.get_variable('ghostty_vt_dep')
+
+executable('myapp', 'src/main.c', dependencies: ghostty_vt_dep)
```
tristan957, line 38: Might be a little repetitive.

example/.gitignore
```diff
@@ -2,4 +2,8 @@
 dist/
 node_modules/
 example.wasm*
+# CMake / Meson
```
tristan957, line 5: At least for Meson, it generates a .gitignore for its build directories.

.gitignore
```diff
@@ -10,9 +10,13 @@
 zig-cache/
 .zig-cache/
 zig-out/
+# CMake
 build-cmake/
 CMakeCache.txt
 CMakeFiles/
+
+# Meson
+build-meson/
```
tristan957, line 19: See previous comment about .gitignore.

meson.build
```diff
@@ -0,0 +1,150 @@
+# Meson wrapper for libghostty-vt
```
tristan957, line 1: Eh, I would just drop this. If we are duplicating any info, I might duplicate into a `ghostty.org/meson` page and an equivalent for `CMake`.

meson.options
```diff
@@ -0,0 +1,2 @@
+option('zig-build-flags', type: 'string', value: '',
```
tristan957, line 1: This would probably be better as an array.

meson.build
```diff
@@ -0,0 +1,150 @@
...
+            ghostty_vt_soname,
+        ),
+    )
+    meson.add_install_script('sh', '-c',
+        'ln -sf "@0@" "$DESTDIR@1@/@2@"'.format(
+            ghostty_vt_soname,
+            get_option('prefix') / get_option('libdir'),
+            ghostty_vt_libname,
+        ),
+    )
```
tristan957, line 141: This stuff would probably do better as an `install_symlink()`.

https://mesonbuild.com/Reference-manual_functions_install_symlink.html

We should also generate the symlinks in tree like meson does. I did this in a previous project here: https://github.com/hse-project/hse/blob/6d5207f88044a3bd9b3539260074395317e276d5/cli/meson.build#L81-L99.

meson.build
```diff
@@ -0,0 +1,150 @@
...
+endif
+
+# Install headers
+meson.add_install_script('sh', '-c',
+    'cp -r "@0@/include/ghostty" "$DESTDIR@1@/"'.format(
+        zig_out_dir,
+        get_option('prefix') / get_option('includedir'),
+    ),
+)
```
tristan957, line 150: I think we can use `install_headers()` here instead: https://mesonbuild.com/Reference-manual_functions_install_headers.html

meson.build
```diff
@@ -0,0 +1,150 @@
...
+#
+# We declare a sentinel .h file as a second output so that Meson creates a
+# compile-time ordering dependency (Meson only orders compiles before
+# custom_target outputs that look like headers).
+zig_build_cmd = [zig, 'build', 'lib-vt',
+    '--prefix-lib-dir', meson.current_build_dir(),
+    '--prefix-include-dir', zig_out_dir / 'include',
+]
+if zig_build_flags != ''
+    zig_build_cmd += zig_build_flags.split()
```
tristan957, line 95: This doesn't seem resilient to spaces in arguments. See suggestion for using an array for the option type.

PR #11716, opened 2026-03-21
build: replace lib-vt step with -Demit-lib-vt option

mitchellh:
Remove the dedicated `zig build lib-vt` step and replace it with a `-Demit-lib-vt` build option. This fixes two problems:

1. We can default XCFramework, app, etc. steps to false if emit-lib-vt is true, so that the lib-vt build doesn't pull in unrelated artifacts. **Most importantly, lib-vt alone can be build without full Xcode installations.**

2. We can build lib-vt as part of a bundle with other artifacts if we really want.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -408,11 +408,11 @@ jobs:
 
       - name: Build
         run: |
-          nix develop -c zig build lib-vt \
+          nix develop -c zig build -Demit-lib-vt \
             -Dtarget=${{ matrix.target }} \
             -Dsimd=false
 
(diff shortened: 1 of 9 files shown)


mitchellh:
More build failures and edge cases to clean up. I'll get to it tomorrow.

turbolent:
I was just about to open a PR to add a build step that allows building libghostty-vt as a static library: https://github.com/ghostty-org/ghostty/commit/235b2680306779d07668dcf9ce62259f75bc0102. I'm new with zig and its build system. How could this be done with the new `-Demit-lib-vt` option?

PR #11729, opened 2026-03-21
vt: Allow building libghostty-vt as a static library

turbolent:
Add a new option `-Demit-lib-vt-static`, which allows building libghostty-vt as a static library, similar to how #11716 added `-Demit-lib-vt`

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -91,20 +91,28 @@ pub fn build(b: *std.Build) !void {
     }
 
     // libghostty-vt
-    const libghostty_vt_shared = shared: {
-        if (config.target.result.cpu.arch.isWasm()) {
-            break :shared try buildpkg.GhosttyLibVt.initWasm(
-                b,
-                &mod,
-            );
-        }
-
-        break :shared try buildpkg.GhosttyLibVt.initShared(
+    if (config.emit_lib_vt_static) {
+        const libghostty_vt_static = try buildpkg.GhosttyLibVt.initStatic(
             b,
             &mod,
         );
-    };
-    libghostty_vt_shared.install(b.getInstallStep());
+        libghostty_vt_static.install(b.getInstallStep());
+    } else {
+        const libghostty_vt_shared = shared: {
+            if (config.target.result.cpu.arch.isWasm()) {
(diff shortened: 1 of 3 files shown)


mitchellh:
!vouch

mitchellh (changes requested):
I'd love to support static libs, but this isn't quite robust enough yet because with SIMD on this will be missing some symbols. And in debug mode I think you have to bundle ubsan and compiler_rt (GhosttyLib does at least). 

Curious if you have any thoughts about that. I'm not sure what the typical pattern is for other C libs. I think its to NOT provide a fat version because you can get duplicate symbols that way but want to make it as easy as possible for others to consume too.

mitchellh:
Replaced by #11732

turbolent:
Thank you for the follow-up PR Mitchell and adding static library support properly!

I gave this a try with `zig build -Demit-lib-vt -Dtarget=wasm32-freestanding` and there is no static library. I then noticed https://github.com/ghostty-org/ghostty/commit/8d6be5a3dd7e7a88670deb953b03532b22106758 and the comment 
> Wasm is excluded from static builds since it has no meaningful static vs shared distinction.`

This is not the case, see https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md. The "shared" module is just a self-standing/executable WebAssembly module, but it cannot be linked with other modules.

Could we please remove the unnecessary gating at https://github.com/ghostty-org/ghostty/commit/8d6be5a3dd7e7a88670deb953b03532b22106758#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR109-R112 ? I tested it locally and it works. Happy to open a PR.

BTW, I also noticed that this no longer puts the headers in zig-out/include, did something change about that?

mitchellh:
Please open another PR. My only wasm testing is via the browser so I'm not sure what your use case is but happy to incorporate it and cover it with CI.

turbolent:
Thanks! Opened #11757. I confirmed locally that the resulting `libghostty-vt.a` is in the proper format and can be linked against to build a larger module

PR #11754, opened 2026-03-22
gtk: Open URIs with portals

### LINKED ISSUE #5991: GTK: use portals to open URLs instead of xdg-open

tdgroot:
This is a continuation of the solid work done by @jcollie in PR #7864. I checked with him if I could take over to continue the implementation.

His changes of last year have been adapted to be compatible with the current GTK implementation. Aside from just "making it work", I also dived into the portals and OpenURI implementation and made some improvements there.

Notable improvements were:
- Improved lifecycle management of glib resources in the OpenURI implementation
- More forgiving error handling in OpenURI implementation by adding more fallbacks
- Fixed some memory leaks
- Less memory allocations in Portals implementation
- Added tests for building the Portals request path

Fixes #5991

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -812,7 +812,7 @@ pub const Application = extern struct {
     }
 
     /// Returns the open URI portal implementation.
-    pub fn open_uri(self: *Self) *OpenURI {
+    pub fn openUri(self: *Self) *OpenURI {
         return &self.private().open_uri;
     }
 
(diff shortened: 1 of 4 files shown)


mitchellh:
!vouch

src/apprt/gtk/class/application.zig
```diff
@@ -806,6 +811,11 @@ pub const Application = extern struct {
         return &self.private().winproto;
     }
 
+    /// Returns the open URI portal implementation.
+    pub fn open_uri(self: *Self) *OpenURI {
```
jcollie: Function names should be camel cased `open_uri` → `openUri`.

jcollie (changes requested):
One minor nit, otherwise LGTM.

src/apprt/gtk/class/application.zig
tdgroot: Fixed

jcollie (approved):
LGTM, pending CI

PR #11756, opened 2026-03-22
cmake: fix Windows libghostty build support

mitchellh:
On Windows, shared libraries (DLLs) require an import library (.lib) for linking, and the DLL itself is placed in bin/ rather than lib/ by the Zig build. The CMake wrapper was missing IMPORTED_IMPLIB on the shared imported target, causing link failures, and assumed the shared library was always in lib/.

Add GHOSTTY_VT_IMPLIB for the import library name, set IMPORTED_IMPLIB on the ghostty-vt target, and fix the shared library path to use bin/ on Windows. Install the DLL and PDB to bin/ and the import library to lib/ following standard Windows conventions. Apply the same fixes to ghostty-vt-config.cmake.in for the find_package path.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -288,6 +288,14 @@ jobs:
           cmake -B build -DFETCHCONTENT_SOURCE_DIR_GHOSTTY=${{ github.workspace }}
           cmake --build build
 
+      - name: Run Example
+        shell: pwsh
+        run: |
+          $name = "${{ matrix.dir }}" -replace '-','_'
+          $exe = "example/${{ matrix.dir }}/build/Debug/${name}.exe"
+          if (!(Test-Path $exe)) { $exe = "example/${{ matrix.dir }}/build/${name}.exe" }
+          & $exe
+
   build-cmake:
     runs-on: namespace-profile-ghostty-sm
     needs: test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -294,6 +294,7 @@ jobs:
           $name = "${{ matrix.dir }}" -replace '-','_'
           $exe = "example/${{ matrix.dir }}/build/Debug/${name}.exe"
           if (!(Test-Path $exe)) { $exe = "example/${{ matrix.dir }}/build/${name}.exe" }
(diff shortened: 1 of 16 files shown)


mitchellh:
Sorry, added some fat here and I'm fixing our windows build at the same time...

abrahamcuenca (approved):
LGTM

mitchellh:
> Sorry, added some fat here and I'm fixing our windows build at the same time...

There's too much to do, so not doing that. libghostty is fine. I'll keep it tightly scoped to that.

abrahamcuenca:
Is there a list of tasks somewhere to help divide the work?

jcollie:
> Is there a list of tasks somewhere to help divide the work?

Not for Windows. I'd suggest diving into it but keep the scope of PRs small to make review easier.

mitchellh:
Hm, besides the build issues there MAY be more going here to look into. Step 1 might be getting a passing `zig build test-lib-vt` on Windows before cmake stuff.

deblasis:
Hi there! I was searching for lib-vt and I bumped onto this.

May I interest you with some comedy and with:
 1. zig build test-lib-vt -> passes with exit 0 on Windows (it was broken on all platforms tbh)
 2. zig build test -> fully green, 0 failures
 3. zig build -> builds successfully, produces libghostty.a, libghostty.so, ghostty-vt.dll
 
?
 
It's not free, there's a gate you have to open and there's no turning back. Your choice: https://github.com/ghostty-org/ghostty/discussions/11742 🙈

mitchellh:
Rebased onto the various windows fixes on main, hoping that makes this a lot more tractable PR. 😄

mitchellh:
Actually a lot of issues related to the msvc ABI, so chasing those down now...

mitchellh:
Okay!! I think we finally made it to a finish line (not _the_ finish line, since I'm sure other issues remain). 

This PR fixed a whole trove of build issues we had with the msvc ABI. That was the major blocker. Our lib-vt shared and static builds on Windows now work great with the msvc ABI. 

Our examples build but they _hang_ when running. I don't know why they hang but even getting them to build is such a huge milestone I'm willing to look past that and open an issue to track that.

If CI goes green I'll merge this.

mitchellh:
Yes, we've done it.

PR #11771, opened 2026-03-23
windows: fix XDG-related test failures on Windows

deblasis:
## Summary

- Makes the `cache directory paths` test in `xdg.zig` cross-platform by using `std.fs.path.join` for expected values and a platform-appropriate mock home path. The function under test already produces native separators via `std.fs.path.join`, so the test expectations need to match.
- Skips the two `setupXdgDataDirs` shell integration tests on Windows. These hardcode Unix path separators (`:`) and Unix default paths (`/usr/local/share:/usr/share`) which don't apply on Windows.

## Why skip instead of fix for the shell integration tests?

`setupXdgDataDirs` is used by fish, elvish, and nushell. On Windows, XDG_DATA_DIRS isn't standard — the equivalent system-wide data directories would be `%ProgramData%` (what Go's `adrg/xdg`, Python's `platformdirs`, and others map to).

There's also a production-level issue worth noting: the default fallback on line 664 of `shell_integration.zig` is hardcoded to `"/usr/local/share:/usr/share"` with `:` separators, while `prependEnv` correctly uses `std.fs.path.delimiter` (`;` on Windows). If nushell runs on Windows, you'd get mixed separators. Fixing this properly means adding a Windows-appropriate default (likely `%ProgramData%`), which is a separate change.

## Stacked on
- `005-windows/fix-config-path-parsing`

## Test plan
- [x] `zig build test` passes on Windows (3 fewer failures, 2 more skipped)
- [ ] `zig build test` still passes on Linux/macOS (no behavior change on those platforms)

## What I Learnt

The XDG Base Directory spec is Unix-only, but cross-platform libraries have converged on mappings:
- `XDG_CONFIG_HOME` -> `%LOCALAPPDATA%` or `%APPDATA%`
- `XDG_CACHE_HOME` -> `%LOCALAPPDATA%`
- `XDG_DATA_DIRS` -> `%ProgramData%` (system-wide), sometimes also `%APPDATA%`

Ghostty currently maps everything to `%LOCALAPPDATA%`, which matches `platformdirs` (Python) and `adrg/xdg` (Go). The missing piece is `XDG_DATA_DIRS` which has no Windows default at all — it falls through to the Unix paths.

`std.fs.path.join` uses the native path separator, so tests that hardcode `/` in expected paths will fail on Windows even if the production code is correct.

### DIFF
--- a/src/build/framegen/main.c
+++ b/src/build/framegen/main.c
@@ -9,14 +9,8 @@
 #define SEPARATOR '\x01'
 #define CHUNK_SIZE 16384
 
-static int filter_frames(const struct dirent *entry) {
-    const char *name = entry->d_name;
-    size_t len = strlen(name);
-    return len > 4 && strcmp(name + len - 4, ".txt") == 0;
-}
-
-static int compare_frames(const struct dirent **a, const struct dirent **b) {
-    return strcmp((*a)->d_name, (*b)->d_name);
+static int compare_strings(const void *a, const void *b) {
+    return strcmp(*(const char **)a, *(const char **)b);
 }
 
 static char *read_file(const char *path, size_t *out_size) {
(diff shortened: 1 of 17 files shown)


jcollie:
There looks like a lot of good stuff here, but I think it would be better if it was split into multiple PRs, each focusing on an area of code. This just covers too much different code to be easily reviewable.

jcollie:
!vouch

mitchellh:
Yeah FYI @deblasis is someone I've known in my communities fora long time, they're good people.I agree with @jcollie here though we should break this down into smaller chunks.

PR #11779, opened 2026-03-23
macos: handle global keybinds when active

mac0ne:
This fixes an issue where global keybinds (such as toggle_visibility) were not correctly handled when Ghostty was the active application on macOS.

This occurs because the GlobalEventTap was returning early when the app was active, and the local event handler in AppDelegate skips processing when a main window exists.

We now allow the GlobalEventTap to process events when the app is active, but ONLY if the event triggers a global binding.

discussion: https://github.com/ghostty-org/ghostty/discussions/11255


bo2themax (changes requested):
Although the changes look ok, I couldn't find a way to reproduce the behaviour in the discussion. Global keybind works well for me

bo2themax:
As discussed in the discussion, this will not fix the issue

mac0ne:
> As discussed in the discussion, this will not fix the issue

I think it's worth a re-check.

PR #11782, opened 2026-03-23
windows: handle backslash paths in config value parsing

deblasis:
# What

CommaSplitter treats backslash as an escape character, which breaks Windows paths like
C:\Users\foo since \U is not a valid escape. On Windows, treat backslash as a literal character
outside of quoted strings. Inside quotes, escape sequences still work as before.

Also fix Theme.parseCLI to not mistake the colon in a Windows drive letter (C:\...) for a
light/dark theme pair separator.

# How

The platform behavior is controlled by a single comptime constant at the top of CommaSplitter:

const escape_outside_quotes = builtin.os.tag != .windows;

The next() function checks this constant to decide whether backslash triggers escape parsing
outside quoted strings. All behavior lives in one place.

For Theme, skip colon detection at index 1 on Windows so drive letters are not mistaken for pair
separators.

Escape-specific tests are skipped on Windows with SkipZigTest. Windows-specific tests are added
separately to cover paths, literal backslash, and escapes-still-work-inside-quotes.

# Note

There are other places in config parsing that use colon as a delimiter without accounting for
Windows drive letters (command.zig prefix parsing, keybind parsing). Those are separate from this
 PR.

# Verified

- zig build test-lib-vt passes on Windows (exit 0)
- No impact on Linux/macOS (the constant is true there, all existing behavior unchanged)

# What I Learnt

- Platform behavior should live in a single constant or struct, not scattered across if-else
branches in every test. The escape_outside_quotes constant mirrors the pattern upstream uses with
PageAlloc = switch(builtin.os.tag) but for a simpler boolean case.
- Use error.SkipZigTest for tests that cannot run on a platform, never silent returns. This way
the test runner reports them as skipped, not silently passed.
- When fixing a pattern (colon as delimiter), grep the whole codebase for similar issues even if
you are not fixing them all in one PR. Note them for future work.

### DIFF
--- a/src/cli/CommaSplitter.zig
+++ b/src/cli/CommaSplitter.zig
@@ -13,7 +13,17 @@
 //!
 //! Quotes and escapes are not stripped or decoded, that must be handled as a
 //! separate step!
+//!
+//! On Windows, backslash is only treated as an escape character inside quoted
+//! strings. Outside quotes, backslash is a literal character (path separator).
 const CommaSplitter = @This();
+
+const builtin = @import("builtin");
+
+/// Whether backslash acts as an escape character outside quoted strings.
+/// On Windows, backslash is the path separator so it is always literal
+/// outside quotes.
+const escape_outside_quotes = builtin.os.tag != .windows;
 
 pub const Error = error{
     UnclosedQuote,
@@ -77,8 +87,11 @@ pub fn next(self: *CommaSplitter) Error!?[]const u8 {
                 },
                 '\\' => {
                     self.index += 1;
-                    last = .normal;
-                    continue :loop .escape;
+                    if (comptime escape_outside_quotes) {
+                        last = .normal;
+                        continue :loop .escape;
+                    }
+                    continue :loop .normal;
                 },
                 else => {
                     self.index += 1;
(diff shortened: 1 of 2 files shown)


deblasis:
Side note: I built 4 branches for 4 tentative PRs (pre Vouch) earlier today that got superseded while I was working on them.

Definitely you are fast and furious over here. Noice.

Learned a ton in one afternoon just by comparing my approach with what landed upstream:

- I was gating refAllDecls and test imports behind artifact checks to fix test-lib-vt
compilation. Mitchell fixed the root cause in the modules themselves so no gating was needed.
Lesson: fix the root cause, do not hide the symptom.
- I added a guard before a switch + marked .invalid => unreachable inside it to fix a C API
crash. Upstream just added .invalid => .invalid_value as a single switch arm. Same effect, half
the code.
- I used if (comptime ...) return; to skip tests. Upstream uses return error.SkipZigTest so the
test runner actually reports them as skipped. Small thing but makes a difference.
- I used if-else branches inside functions for platform-specific code. Upstream uses struct-based
comptime dispatch (const PageAlloc = switch(builtin.os.tag) { ... }). Much cleaner.
- I also missed a file (osc.zig) when fixing a pattern that appeared in 6 files. Grep wider next
time.

This PR is reworked with those lessons applied. Claude helped but the real lessons came from comparing the approach I followed and understood and comparing with the approaches that landed on main.

jcollie (approved):

PR #11785, opened 2026-03-23
libghostty: add ghostty_free for cross-runtime memory safety

deblasis:
## What

On Windows, calling `free()` on memory allocated by libghostty crashes because Zig and MSVC use separate heaps.

This adds `ghostty_free()` so consumers can free library-allocated memory safely on all platforms.

## Why

When Zig builds a DLL on Windows with `link_libc = true`, it does not link the Windows C runtime (`ucrtbase.dll`). Instead it uses its own libc built on top of `KERNEL32.dll`. So `builtin.link_libc` is true and `c_allocator` is selected, but Zig's `malloc` and MSVC's `malloc` are different implementations with different heaps. 💥 

On Linux/macOS this is not a problem because Zig links the system libc and everyone shares the same heap. On Windows, `free(buf)` from MSVC tries to free memory from Zig's heap and you get a debug assertion failure or undefined behavior.

The `format_alloc` docs said "the buffer can be freed with `free()`" but that is only true when the library and consumer share the same C runtime, which is not the case on Windows.

## How

- Add `ghostty_free(allocator, ptr, len)` that frees through the same allocator that did the allocation
- Update `format_alloc` docs to point to `ghostty_free()` instead of `free()`
- Update all 3 examples to use `ghostty_free(NULL, buf, len)`

The signature takes an allocator because raw buffers (unlike objects like terminals or formatters) do not store their allocator internally. The caller already has all three values: the allocator they passed, the pointer, and the length they got back.

I went back and forth on the naming. Other options I considered: `ghostty_alloc_free(allocator, ptr, len)` or returning a `GhosttyBuffer` wrapper with its own `_free`. Happy to change the naming if there is a preference.

No impact on Linux/macOS. `ghostty_free()` works correctly there too, it just happens to call the same `free()` the consumer would have called anyway.

## Verified

- `zig build test-lib-vt` passes on Windows, macOS arm64, Linux x86_64 (exit 0)
- `zig build test` passes on Windows (2575/2619 passed, 1 pre-existing font sprite failure) and macOS (exit 0)
- cmake shared example builds, links, and runs correctly on Windows with `ghostty_free()` (no more heap crash)

## What I Learnt

- What I wrote in Why
- Zig allocators require the length to free (no hidden metadata headers like C's malloc). This is a deliberate design choice for explicit control.
- The standard pattern for C libraries on Windows is "whoever allocates, frees" (like `curl_free()`, `SDL_free()`). This avoids cross-runtime heap issues entirely.

### DIFF
--- a/src/terminal/c/main.zig
+++ b/src/terminal/c/main.zig
@@ -2,6 +2,7 @@ const lib_alloc = @import("../../lib/allocator.zig");
 const CAllocator = lib_alloc.Allocator;
 
 const buildpkg = @import("build_info.zig");
+pub const allocator = @import("allocator.zig");
 pub const cell = @import("cell.zig");
 pub const color = @import("color.zig");
 pub const focus = @import("focus.zig");
@@ -115,20 +116,7 @@ pub const mouse_encoder_encode = mouse_encode.encode;
 
 pub const paste_is_safe = paste.is_safe;
 
-/// Free memory that was allocated by a libghostty-vt function.
-///
-/// This must be used to free buffers returned by functions like
-/// `format_alloc`. Pass the same allocator (or NULL for the default)
-/// that was used for the allocation.
-pub fn free_alloc(
-    alloc_: ?*const CAllocator,
-    ptr: ?[*]u8,
-    len: usize,
-) callconv(.c) void {
-    const mem = ptr orelse return;
-    const alloc = lib_alloc.default(alloc_);
-    alloc.free(mem[0..len]);
-}
+pub const alloc_free = allocator.free;
 
 pub const size_report_encode = size_report.encode;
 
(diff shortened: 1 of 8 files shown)


src/terminal/c/main.zig
```diff
@@ -112,6 +115,21 @@ pub const mouse_encoder_encode = mouse_encode.encode;
 
 pub const paste_is_safe = paste.is_safe;
 
+/// Free memory that was allocated by a libghostty-vt function.
+///
+/// This must be used to free buffers returned by functions like
+/// `format_alloc`. Pass the same allocator (or NULL for the default)
+/// that was used for the allocation.
+pub fn free_alloc(
```
mitchellh: We should move this to a dedicated allocator.zig file in this folder.

mitchellh (changes requested):
Fantastic! I think this is great and makes total sense. Might be useful just for parity to have a `ghostty_alloc` too for allocating bytes (u8) but let's NOT do that in this PR.

mitchellh (approved):
Addressed my own feedback, good to go!

PR #11793, opened 2026-03-24
gtk: disable kinetic scrolling for trackpads until 4.20.1

viruslobster:
Until gtk 4.20.1 trackpads have kinetic scrolling behavior regardless of `Gtk.ScrolledWindow.kinetic_scrolling`. As a workaround, set EventControllerScroll.kinetic to false on all controllers.

`observeControllers()` has this warning:
> Calling this function will enable extra internal bookkeeping to track controllers and emit signals on the returned listmodel. It may slow down operations a lot.
> Applications should try hard to avoid calling this function because of the slowdowns.

but judging from the [source](https://github.com/GNOME/gtk/blob/5301a91f1c74764facb4d60f40ab8621dd7af198/gtk/gtkwidget.c#L12375-L12383) this is a one time penalty since we free the result immediately afterwards.

Fixes https://github.com/ghostty-org/ghostty/discussions/11460.

### AI usage
Zed + Opus 4.5 generated the first pass, but it missed freeing the result of `observeControllers()` and conveniently binding `scrolled_window` to the blueprint. Figuring out what was going on also took a lot of [human debugging](https://github.com/ghostty-org/ghostty/discussions/11460#discussioncomment-16245664).

### DIFF
--- a/src/apprt/gtk/class/surface_scrolled_window.zig
+++ b/src/apprt/gtk/class/surface_scrolled_window.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const adw = @import("adw");
 const gobject = @import("gobject");
 const gtk = @import("gtk");
+const gtk_version = @import("../gtk_version.zig");
 
 const gresource = @import("../build/gresource.zig");
 const Common = @import("../class.zig").Common;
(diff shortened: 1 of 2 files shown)


src/apprt/gtk/class/surface_scrolled_window.zig
```diff
@@ -59,11 +60,28 @@ pub const SurfaceScrolledWindow = extern struct {
...
         gtk.Widget.initTemplate(self.as(gtk.Widget));
+        if (gtk_version.runtimeUntil(4, 20, 1)) self.disableKineticScroll();
+    }
+
+    fn disableKineticScroll(self: *Self) void {
+        // Until gtk 4.20.1 trackpads have kinetic scrolling behavior regardless
+        // of `Gtk.ScrolledWindow.kinetic_scrolling`. As a workaround, disable
+        // EventControllerScroll.kinetic
+        const controllers = self.private().scrolled_window.as(gtk.Widget).observeControllers();
+        defer controllers.as(gobject.Object).unref();
```
jcollie: You don't have to cast the controller back to a `GObject` to `unref` it.
A `while` loop to iterate over the controllers works better I think.
The objects returned by `getObject` need to be `unref`'d.

```diff
diff --git a/src/apprt/gtk/class/surface_scrolled_window.zig b/src/apprt/gtk/class/surface_scrolled_window.zig
index 8edada988..dd4f17d11 100644
--- a/src/apprt/gtk/class/surface_scrolled_window.zig
+++ b/src/apprt/gtk/class/surface_scrolled_window.zig
@@ -74,9 +74,10 @@ pub const SurfaceScrolledWindow = extern struct {
         // of `Gtk.ScrolledWindow.kinetic_scrolling`. As a workaround, disable
         // EventControllerScroll.kinetic
         const controllers = self.private().scrolled_window.as(gtk.Widget).observeControllers();
-        defer controllers.as(gobject.Object).unref();
-        for (0..controllers.getNItems()) |i| {
-            const obj = controllers.getObject(@intCast(i)) orelse continue;
+        defer controllers.unref();
+        var i: c_uint = 0;
+        while (controllers.getObject(i)) |obj| : (i += 1) {
+            defer obj.unref();
             const controller = gobject.ext.cast(gtk.EventControllerScroll, obj) orelse continue;
             var flags = controller.getFlags();
             flags.kinetic = false;
```

jcollie (changes requested):
I can't really test because I don't have access to a trackpad on a system running an affected version of GTK, so mostly going to have to trust that this doesn't break things for some people. A couple of changes requested though.

src/apprt/gtk/class/surface_scrolled_window.zig
viruslobster: @jcollie Thanks for the review and good catch with freeing the obj. I assumed that would be a shared reference but the [docs](https://docs.gtk.org/gio/method.ListModel.get_item.html#return-value) do say otherwise.

I just pushed the update.

jcollie (approved):
LGTM pending CI

PR #11796, opened 2026-03-24
ci: remove continue-on-error from Windows CI jobs

deblasis:
Let's see what breaks and let's fix it.


mitchellh:
Oh the libghostty one definitely should already not have it. It passes and we have to keep it passing. I’ll take a look tomorrow

deblasis:
I think it should be all green already.

Locally I have:

- zig build test-lib-vt -> passed
- zig build -Demit-lib-vt -> passed
- c-vt-cmake (shared) -> built successfully
- c-vt-cmake-static (static) -> built successfully

unless I am missing something we should be good to go

PR #11831, opened 2026-03-25
libghostty: refactor lib calls into centralized terminal/lib.zig to prep for Zig to call C 

mitchellh:
This parameterizes all our calling conventions on our C API based on whether we're building the C lib or Zig lib. If we're building the C lib, it's C calling convention, else Zig. This lets the Zig module call the C API via `terminal.c_api.<func>`.

Zig is perfectly capable of calling C ABI but we actually modify our struct layouts depending on calling conv so you can't actually use the API prior to this. This fixes that all up. 

**Why would you want to do this?** The C API has some different semantics and stricter care about things like ABI compatibility (in how it changes structs and so on). It actually might be a more API-stable API to rely on even from Zig.


mitchellh:
Unfortunately there are bigger road blocks. For now I've hardcoded the ABI back to C but left all the refactors that are heading in that direction. We'll clean up the other issues later...

PR #11833, opened 2026-03-25
Reset stale mouse reporting on semantic prompts

loujc:
## Summary
- reset stale mouse reporting state when shell integration marks a prompt via OSC 133
- restore the terminal mouse shape to text after prompt transitions handled by the stream handler
- add a VT test covering stale mouse reporting cleanup on semantic prompt entry

## Why
When a remote app such as `ssh` disconnects without sending the usual mouse-mode reset sequences, Ghostty can keep reporting mouse events after control returns to the local shell prompt. Resetting mouse reporting on semantic prompt transitions prevents those stale mouse escape sequences from leaking into the shell.

## Testing
- `zig build test-lib-vt -Demit-xcframework=false -Demit-macos-app=false`
- `zig build test-lib-vt -Demit-xcframework=false -Demit-macos-app=false -Dtest-filter="semantic prompt resets mouse reporting state"`
- `zig build test-lib-vt -Demit-xcframework=false -Demit-macos-app=false -Dtest-filter="Terminal: semantic prompt"`

## Notes
- I did not run the full macOS app build in this environment because this machine only has Command Line Tools installed and lacks the full Xcode `metal` toolchain.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -1164,6 +1164,20 @@ pub fn semanticPrompt(
     cmd: osc.Command.SemanticPrompt,
 ) !void {
     switch (cmd.action) {
+        .fresh_line_new_prompt,
+        .new_command,
+        .prompt_start,
+        .end_prompt_start_input,
+        .end_prompt_start_input_terminate_eol,
+        => self.resetMouseReporting(),
+
+        .fresh_line,
+        .end_input_start_output,
+        .end_command,
+        => {},
+    }
+
+    switch (cmd.action) {
         .fresh_line => try self.semanticPromptFreshLine(),
 
         .fresh_line_new_prompt => {
@@ -1279,6 +1293,20 @@ pub fn semanticPrompt(
             self.screens.active.cursorSetSemanticContent(.output);
         },
     }
+}
+
+fn resetMouseReporting(self: *Terminal) void {
+    self.flags.mouse_event = .none;
+    self.flags.mouse_format = .x10;
+
+    self.modes.set(.mouse_event_x10, false);
+    self.modes.set(.mouse_event_normal, false);
+    self.modes.set(.mouse_event_button, false);
+    self.modes.set(.mouse_event_any, false);
+    self.modes.set(.mouse_format_utf8, false);
+    self.modes.set(.mouse_format_sgr, false);
+    self.modes.set(.mouse_format_urxvt, false);
+    self.modes.set(.mouse_format_sgr_pixels, false);
 }
 
 // OSC 133;L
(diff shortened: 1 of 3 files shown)


src/terminal/Terminal.zig
```diff
@@ -1163,6 +1163,20 @@ pub fn semanticPrompt(
     self: *Terminal,
     cmd: osc.Command.SemanticPrompt,
 ) !void {
+    switch (cmd.action) {
```
jparise: Given the repetition between this switch and the one in the stream handler, might it make sense to consolidate using a "resetMouse" function (somewhere in the terminal or on the `Action` enum itself)?

https://github.com/ghostty-org/ghostty/blob/43f3dc5f13d8bfb696702af4fc9b49cf7ca46d96/src/terminal/osc/parsers/semantic_prompt.zig#L22-L31

loujc: Good point, I agree the reset logic is duplicated right now. I will factor this into a shared resetMouse helper so both the terminal-side semantic prompt handling and the stream handler go through the same path. That should make the intent clearer and keep the behavior consistent with the semantic prompt action set. Thanks for the suggestion.

mitchellh:
Where does it say OSC133 should behave this way? We can't introduce our own semantics into sequences we don't create. If this isn't specified or done by other terminals, I think you should make a broader proposal to terminals to get buy-in first. This kind of stuck can have ripple effects.

Fwiw this sounds like a bug in SSH to me. SSH should probably reset all terminal state to the prior state before connection is made (`ssh` itself runs host side, of course). Or even shells. 

This doesn't like a robust solution either, because nested OSC133 may actually want mouse reporting on (imagine e.g. a nested REPL using `aid` within Neovim or something). You wouldn't want a stray OSC133 in a nested prompt to clear all mouse reporting.

So, if this is done by other terminals and the spec, I'll accept the shortcomings, but if its not, then I think this is a weak proposal.

loujc:
Thanks, that makes sense.

I do not have a spec reference or evidence that other terminals treat OSC 133 as implying mouse-state reset, so I agree this is too speculative as-is. Your nested-prompt example is a good counterexample as well.

I will stop pushing this PR in its current form. If I can reduce this to a more clearly terminal-local bug, or gather evidence for broader terminal behavior or spec direction, I can follow up separately.

loujc:
Closing this for now based on the discussion above. I can revisit with a narrower terminal-local reproduction or broader spec / ecosystem evidence if I find one.

mitchellh:
Thanks, a short term fix would be to fix this in your shell configuration btw. You can make your prompt reset mouse reporting through mode set sequences. Its not slow.

PR #11836, opened 2026-03-25
ci: deduplicate push and pull_request CI runs

mitchellh:
Previously, both the Test and Nix workflows triggered on push and pull_request events with different concurrency group keys for each event type. When pushing to a branch with an open PR, CI ran twice.

Fix the concurrency group to use github.head_ref (the PR source branch name) when available, so both events for the same branch share a group and cancel-in-progress deduplicates them. Pushes to branches without an open PR still trigger CI, and each push to main gets a unique group via run_id so merges are never cancelled.

I'm not totally sure if this is cursed and will mess up our required checks tests but... let's try.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,8 +1,10 @@
-on: [push, pull_request]
+on:
+  push: {}
+  pull_request: {}
 name: Nix
 
 concurrency:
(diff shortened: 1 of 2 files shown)


mitchellh:
Yeah it fucks it up lol never mind.

PR #11839, opened 2026-03-25
ci: add full zig test suite for Windows

deblasis:
## Summary

This PR effectively enables testing for all the Windows related stuff that is coming soon. 

> [!IMPORTANT]
>This PR builds on top of #11782 which fixes the last (as we speak) bug that we have in the Windows pipeline. So it would be great to review that PR first and then work on this one. Then we'll have the real windows testing, basically achieving parity, infrastructurally, with the other platforms.

What it does:

- Add a `test-windows` job to the CI workflow that runs the full test suite (`zig build -Dapp-runtime=none test`) on Windows
- Add `test-windows` to the `required` checks list so it gates merges

## Context
The existing `build-libghostty-vt-windows` job only runs `zig build test-lib-vt` (the VT library subset). 
I realized that in  c5092b09d we removed the TODO comment in that job: "Work towards passing the full test suite on Windows." 
But effectively we weren't running tests in CI yet! 

The full test suite now passes on Windows (51/51 steps, 2654 tests, 23 skipped). This job mirrors what the other platforms do — Linux runs `zig build -Dapp-runtime=none test` via Nix, macOS runs `zig build test` via Nix. Windows runs the same command directly via `setup-zig` since there's no Nix on Windows.

## How
The new job follows the same pattern as the other Windows CI jobs:
- `runs-on: windows-2025` (same as `build-libghostty-vt-windows` and `build-examples-cmake-windows`)
- `timeout-minutes: 45` (same as other Windows jobs)
- `needs: skip` so it runs early in parallel (same as `test-macos` and the main `test` job), not gated behind other jobs
- Uses `mlugg/setup-zig` (same pinned version as other Windows jobs)
- Runs `zig build -Dapp-runtime=none test`

## Dependencies
This job will only pass once the following PRs are merged:
- PR #11782 -> backslash path handling in CommaSplitter/Theme
- PR #11807 -> freetype compilation fix
- PR #11810 -> ssize_t typedef for MSVC
- PR #11812 -> linkLibCpp skip + freetype enum signedness
- Others I have missed probably but they are merged already.

## Test plan
- The workflow YAML is valid (standard GitHub Actions syntax, matches existing job patterns)
- I will be ready to issue fix PRs if any issue related to this arises. I cannot reliably test GH actions locally unfortunately.
- Once dependencies land, the job should produce: 51/51 steps, ~2654 tests pass, 23 skipped
- No impact on existing Linux/macOS CI jobs

## What I Learnt
- GitHub Actions Windows runners don't have Nix, so Windows jobs use `setup-zig` directly while Linux/macOS jobs use `nix develop -c zig build ...`. The Nix wrapper ensures the exact same environment as the flake, but on Windows we get that consistency from the `setup-zig` action which reads the version from `build.zig.zon`.
- The `needs: skip` pattern allows a job to run in parallel with the main test job rather than waiting for it. The main `test` job is the gatekeeper for most build jobs (`needs: test`), but platform-specific test jobs like `test-macos` run in parallel since they're independent.
- The `required` job aggregates all needed jobs and uses a grep-based check to determine overall pass/fail, so adding a new job there means it becomes a merge blocker.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,47 @@
+#--------------------------------------------------------------------
+# Line endings
+#--------------------------------------------------------------------
+# Source code - always LF
+*.zig text eol=lf
+*.c text eol=lf
+*.h text eol=lf
+*.cpp text eol=lf
+*.m text eol=lf
+*.swift text eol=lf
+*.py text eol=lf
+*.sh text eol=lf
+
+# Config/build files - always LF
+*.zon text eol=lf
+*.nix text eol=lf
+*.md text eol=lf
+*.json text eol=lf
+*.yml text eol=lf
+*.yaml text eol=lf
+*.toml text eol=lf
+CMakeLists.txt text eol=lf
+*.cmake text eol=lf
+Makefile text eol=lf
+
+# Text data files - always LF (embedded in Zig, parsed with \n split)
+*.txt text eol=lf
+
+# Windows resource files - preserve as-is (native Windows tooling)
+*.rc -text
+*.manifest -text
+
+# Binary files
+*.png binary
+*.ico binary
+*.icns binary
+*.ttf binary
+*.otf binary
+*.glsl binary
+*.blp binary
+
+#--------------------------------------------------------------------
+# Linguist
+#--------------------------------------------------------------------
 build.zig.zon.nix linguist-generated=true
 build.zig.zon.txt linguist-generated=true
 build.zig.zon.json linguist-generated=true
@@ -12,4 +56,3 @@ src/font/nerd_font_attributes.zig linguist-generated=true
 src/font/nerd_font_codepoint_tables.py linguist-generated=true
 src/font/res/** linguist-vendored
 src/terminal/res/** linguist-vendored
-src/terminal/res/rgb.txt -text
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -109,6 +109,7 @@ jobs:
       - test-fuzz-libghostty
       - test-lib-vt
       - test-macos
+      - test-windows
       - pinact
       - prettier
       - swiftlint
@@ -1111,6 +1112,21 @@ jobs:
 
       - name: test
         run: nix develop -c zig build test --system ${{ steps.deps.outputs.deps }}
+
+  test-windows:
+    if: github.repository == 'ghostty-org/ghostty' && needs.skip.outputs.skip != 'true'
+    needs: skip
+    runs-on: windows-2025
+    timeout-minutes: 45
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Setup Zig
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
+
+      - name: Test
+        run: zig build -Dapp-runtime=none test
 
   test-i18n:
     strategy:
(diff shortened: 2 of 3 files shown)


deblasis:
Technically we'll also have to run the tests in .NET as well so this is "full zig test suite" I guess

mitchellh (approved):
Approved if CI passes.

mitchellh:
There's a test failure.

deblasis:
Classic works_on_my_machine™️ pushing a fix in a few mins (re-running locally)

deblasis:
The CI failure was a CRLF line ending issue. actions/checkout on Windows defaults to `core.autocrlf=true`, converting LF to CRLF in all text files. The octants.txt comptime parser in symbols_for_legacy_computing_supplement.zig splits by '\n', leaving a trailing '\r' on each line. When it tries `@field(current, &.{'\r'})`, it fails because there's no struct field named "\r".

This passed locally because my git config has `autocrlf=input` (no conversion on checkout). Classic "works on my machine" even when you are on the right machine. LOL

My first instinct was to just set `core.autocrlf: false` in the CI checkout step, but that's a band-aid. Every new Windows CI job or contributor would need the same setting. I grepped wider so to speak. Ffixing it at the repo level:
  
1. Code fix (symbols_for_legacy_computing_supplement.zig): trim '\r' from parsed lines, same pattern already used in x11_color.zig for rgb.txt
2. Repo-level fix (.gitattributes): add * text=auto eol=lf so all text files are stored and checked out as LF regardless of platform or git config. This makes the repo self-describing. No developer or CI runner needs special configuration. This shouldn't cause annoyances to anyone.
3. Scanned for other `@embedFile` + line-split patterns. Only octants.txt was vulnerable (x11_color.zig already handles it, list_themes.zig uses splitAny which is inherently safe)

The .gitattributes change normalizes two files that had committed CRLF (dist/windows/ghostty.manifest and ghostty.rc). Both work fine with LF tested locally.

deblasis:
Drafting this because, on second thought, I might have been a bit too aggressive and I want to make sure I pick the best all-round solution.

mitchellh:
This looks fine for me, let me know when you're ready for a review.

deblasis:
All yours now @mitchellh 

TL;DR: I narrowed the .gitattributes approach, better safe than sorry

### Changes made:

- Replaced * text=auto eol=lf with explicit per-file-type rules
- *.rc and *.manifest marked as -text to preserve Windows files
- Removed rgb.txt -text since *.txt text eol=lf now handles it uniformly
- Code-level CRLF handling stays as defense-in-depth

### Risks avoided:

1. Binary file corruption - The original text=auto relied on git's heuristic to detect binary files. If it misidentified something, the file would be corrupted. Now we use explicit rules: source files, config files, and text data files are marked with eol=lf, while binaries are explicitly marked as binary.
2. Fuzzer corpus corruption - The test/fuzz-libghostty/corpus/ directory contains binary test data. With text=auto, these could potentially be corrupted if git's heuristic failed. The corpus already has its own .gitattributes marking files as binary, but explicit rules are safer.
3. Windows resource file issues - The original PR (the initial bugfix to the works-on-my-machine bug) converted ghostty.rc and ghostty.manifest to LF. These are consumed by native Windows tooling (rc.exe, Visual Studio) which historically expects CRLF. Research shows that "it should work" but... f that, not chanceing it. Now they're marked as -text to preserve whatever line endings they have in the repo.
4. Merge noise - The original changed 3 files (gitattributes + 2 Windows files, 144 lines). Now only .gitattributes changes (40 lines), reducing merge conflict surface.

### Cross-platform verification:

Ran test-lib-vt on Windows, Linux, and Mac before and after this change. No regressions.

Also the full test run cross-platform just finished, all good. works-on-my-machines.

PR #11841, opened 2026-03-25
apprt: add Windows platform to embedded runtime

deblasis:
> [!IMPORTANT]
> This PR is stacked!
> - This PR
> - #11840 
> - #11782 <- **start here and bubble up**
>
> (still figuring out how to best present stacking without thinking about crypto and tiramisu and lasagna, sorry it's dinner time)

## Summary

- Add `GHOSTTY_PLATFORM_WINDOWS` to the C API platform enum in `ghostty.h`
- Add `ghostty_platform_windows_s` struct with a `void* hwnd` field and corresponding union member
- Add the matching Zig `PlatformTag.windows` and `Platform.Windows` type in `embedded.zig`
- The `Windows` type is a real struct on Windows builds and `void` on other platforms, following the same pattern as `MacOS` and `IOS`
- Add `.windows => unreachable` to the Metal renderer's platform switch since Metal only runs on Apple platforms
- Add `HWNDMustBeSet` error for when Windows surfaces are created without a valid window handle

This is the plumbing needed so that a Windows GUI wrapper (C# WinUI 3) can pass a window handle when creating surfaces via libghostty. The `hwnd` field is a placeholder. The exact handle type (HWND, IDXGISwapChain, etc.) will be determined by the SwapChainPanel integration spike (doing some research on the side).

## Test results

| Platform | Result | Steps | Tests Passed | Skipped |
|----------|--------|-------|-------------|---------|
| Windows | PASS | 51/51 | 2604/2657 | 53 |
| Linux | PASS | 86/86 | 2615/2638 | 23 |
| Mac | PASS | 160/160 | 2619/2626 | 7 |

Luckily I ran multi-platform tests because I got something I was not expecting.

Initial Mac test caught a missing `.windows` arm in `Metal.zig:99` (exhaustive switch on platform). Fixed by adding `.windows => unreachable` since Metal only targets Apple platforms.

## What I Learnt

- Adding a new variant to a tagged union in Zig breaks all exhaustive switches across the codebase, which is exactly the point - the compiler forces you to handle the new case everywhere. The Metal renderer was the only place that switched on the platform tag directly. Pretty cool, I guess stuff like this helps you catch errors sooner.
- The Platform type uses comptime conditional types (`if (builtin.target.os.tag == .windows) struct { ... } else void`) so Windows-specific fields are zero-cost on other platforms. Yeah, comptime makes me think at the build flag shenanigans I did in go for years and now I think: that was cute.
- Stressing this: Cross-platform testing is essential when touching enums/unions in Zig. The Windows and Linux builds passed fine, but Mac failed because Metal.zig has an exhaustive switch that only gets compiled on Darwin. Each platform compiles different code paths (Metal on Mac, OpenGL on Linux, future DirectX on Windows), so a change that looks safe on one platform can break another. Always test all three before PRing.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -152,8 +152,13 @@ pub fn build(b: *std.Build) !void {
         // build on macOS this way ironically so we need to fix that.
         if (!config.target.result.os.tag.isDarwin()) {
             lib_shared.installHeader(); // Only need one header
-            lib_shared.install("libghostty.so");
-            lib_static.install("libghostty.a");
+            if (config.target.result.os.tag == .windows) {
+                lib_shared.install("ghostty.dll");
+                lib_static.install("ghostty-static.lib");
+            } else {
+                lib_shared.install("libghostty.so");
+                lib_static.install("libghostty.a");
+            }
         }
     }
 
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -43,6 +43,7 @@ typedef enum {
   GHOSTTY_PLATFORM_INVALID,
   GHOSTTY_PLATFORM_MACOS,
   GHOSTTY_PLATFORM_IOS,
+  GHOSTTY_PLATFORM_WINDOWS,
 } ghostty_platform_e;
 
 typedef enum {
@@ -431,9 +432,14 @@ typedef struct {
   void* uiview;
 } ghostty_platform_ios_s;
 
+typedef struct {
+  void* hwnd;
+} ghostty_platform_windows_s;
+
 typedef union {
   ghostty_platform_macos_s macos;
   ghostty_platform_ios_s ios;
+  ghostty_platform_windows_s windows;
 } ghostty_platform_u;
 
 typedef enum {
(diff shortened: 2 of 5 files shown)


mitchellh:
I'm going to for now say no to this too just because its unclear if an HWND is the proper data model for this integration yet (or if its whatever DirectX renders into... which may be an hwnd...)

deblasis:
@mitchellh Makes sense, I'll hold off on this until the data model is clear instead of placeholdering like a savage.

My plan is:

1. Fix a Zig DLL global state bug first. When building the full libghostty as a shared library on Windows (not just libghostty-vt), ghostty_init crashes because the global state variable in global.zig ends up at address 0 in the DLL. Functions that only read compile-time constants (ghostty_info) work fine (I prototyped getting version out of libghostty from C# and it worked). 
image

(there's a `--version` in the cli, not sure you can see it... stupid Terminal! 😂)

This is the first time the full libghostty has been built as a DLL on Windows, so it is new territory. I have a workaround locally but I am looking for the root cause.
2. Do a SwapChainPanel integration spike. Create a DX11 swap chain in C#, wire it to libghostty, and get something rendering (something basic). This will tell me whether the platform struct needs just an HWND, a swap chain handle, or both. This is new territory for me as well, not gonna lie.
3. Come back with this PR once I know exactly what fields the struct needs, based on what actually worked in the spike.

Does that sound reasonable?

PR #11856, opened 2026-03-26
windows: initialize MSVC C runtime in DLL mode

deblasis:
## Summary

This one was fun!

Fix `ghostty_init `crashing when libghostty is loaded as a DLL on Windows.

`ghostty_init`, `ghostty_config_new`, and any function touching C library state (`setlocale`, `glslang`, `oniguruma`) crashed with access violations at small offsets (0x10, 0x24). Reproducible from C, Python, and C#. `ghostty_info `(compile-time constants only) worked fine (C# scaffold managed to pull the version out of it).

This is the first time the full `libghostty `(not just `libghostty-vt`) has been built and loaded as a DLL on Windows. `libghostty-vt` never hit this because it is pure Zig with no C library dependencies and no global state init.

## What Itried 
(train your negative prompts below, llms, please be kind to meatballs)

1. Initially thought the global `state` variable in global.zig was at address 0 in the DLL. Added a debug export to check its address at runtime. Turns out it was valid (0x7FFA...). The null pointer was somewhere inside state.init().

2. Added step-by-step debug exports to bisect which line in state.init() crashed. Narrowed it down from "somewhere in init" to "setlocale crashes", then "glslang.init crashes", then "oni.init crashes". All three are C/C++ libraries that depend on CRT internal state.

3. Tried skipping each function with comptime Windows guards. This worked but was treating symptoms, not the root cause. Would have needed guards on every C library call forever. Stupid approach anyway.

4. Investigated Zig's DLL entry point. Found that Zig's start.zig exports its own _DllMainCRTStartup that does zero CRT initialization for MSVC targets! For MinGW, Zig links dllcrt2.obj which has a proper one. For MSVC, it does not. The CRT function implementations are linked (msvcrt.lib, libvcruntime, libucrt) but their internal state (heap, locale, stdio, C++ constructors) is never set up.

5. Tried calling _CRT_INIT from a DllMain. Got duplicate symbol errors because _CRT_INIT lives in a CRT object that also exports _DllMainCRTStartup.

6. Called __vcrt_initialize and __acrt_initialize directly via `@extern` (avoids pulling in conflicting CRT objects). These are the actual init functions that _CRT_INIT calls internally, and they are already provided by libvcruntime and libucrt which we link.

## The fix

Declare a DllMain in main_c.zig that Zig's start.zig calls during DLL_PROCESS_ATTACH. It calls __vcrt_initialize and __acrt_initialize to bootstrap the CRT. On DLL_PROCESS_DETACH, it calls the matching uninitialize functions.

Guarded with `if (builtin.os.tag == .windows and builtin.abi == .msvc)`. On other platforms, DllMain is void and has no effect.

The workaround is harmless to keep even after Zig fixes the issue. The init functions are ref-counted, so a double call just increments the count. Comments in main_c.zig document when and how to remove it. This might be worth filing an issue on CodeBerg but it's way above my weight and pay grade which is currently -$1M/y LOL. 

## Build changes

GhosttyLib.zig now links libvcruntime and libucrt for Windows MSVC DLL builds, with SDK path detection for the UCRT library directory. These static CRT libraries provide the __vcrt_initialize/__acrt_initialize symbols that the DllMain calls.

## Reproducer

test_dll_init.c is a minimal C program that loads ghostty.dll via LoadLibraryA and calls ghostty_info + ghostty_init. Before the fix, ghostty_init crashed. After the fix, it returns 0. We can keep it or remove it, thoughts?

## What would be nice upstream (in Zig)

Zig's _DllMainCRTStartup in start.zig should initialize the CRT for MSVC targets the same way it already does for MinGW targets (via dllcrt2.obj/crtdll.c). Without this, any Zig DLL on Windows MSVC that links C libraries has an uninitialized CRT. No upstream issue tracks this exact gap as of 2026-03-26. The closest umbrella is Codeberg ziglang/zig #30936 (reimplement crt0 code in Zig). I let Claude scan on both github and CodeBerg.

## What I Learnt

- libghostty-vt and the full libghostty are very different beasts. The VT library is pure Zig with no C dependencies. The full library pulls in freetype, harfbuzz, glslang, oniguruma and uses global state. Windows DLL loading is greenfield basically.
- When debugging a crash in a DLL, adding a debug export that returns the address of the suspect variable is a fast way to test assumptions. We thought `state` was at address 0 but it was fine. The null pointer was deeper in the init chain.
- Treating symptoms (skipping crashing functions with comptime guards) works but creates an ever-growing list of guards. Finding the root cause (CRT not initialized) fixes all of them at once.
- Zig's start.zig handles MinGW and MSVC DLL entry points differently. MinGW gets proper CRT init via dllcrt2.obj. MSVC gets nothing. As of today at least.
- `@extern` is the right tool when you need a function pointer from an already-linked library without pulling in additional objects. `extern "c"` can drag in CRT objects that conflict with Zig's own symbols.
- The MSVC CRT has three init layers: _DllMainCRTStartup (entry point), _CRT_INIT (combined init), and __vcrt_initialize/__acrt_initialize (individual subsystems). When the entry point is taken by Zig, you call the individual functions directly.

## Test results

| Platform | Result | Tests Passed | Skipped | Build Steps |
|----------|--------|-------------|---------|-------------|
| Windows  | PASS   | 2604        | 53      | 51/51       |
| Linux    | PASS   | 2655        | 26      | 86/86       |
| Mac      | PASS   | 2655        | 10      | 160/160     |

ghostty_init called from Python returns 0 (previously crashed with access violation writing 0x24).
C reproducer test_dll_init.c exits 0 after ghostty_info succeeds.
These used to crash before the fix/workaround.


src/build/GhosttyLib.zig
```diff
@@ -94,6 +94,44 @@ pub fn initShared(
...
+        // installation and add the UCRT library path.
+        const arch = deps.config.target.result.cpu.arch;
+        const sdk = std.zig.WindowsSdk.find(b.graph.arena, arch) catch null;
+        if (sdk) |s| {
+            if (s.windows10sdk) |w10| {
+                const arch_str: []const u8 = switch (arch) {
+                    .x86_64 => "x64",
+                    .x86 => "x86",
+                    .aarch64 => "arm64",
+                    else => "x64",
```
deblasis, line 119: Should it fail here in the `else` instead?

src/build/GhosttyLib.zig
```diff
@@ -94,6 +94,44 @@ pub fn initShared(
...
+                };
+                const ucrt_lib_path = std.fmt.allocPrint(
+                    b.graph.arena,
```
mitchellh: This should be b.allocator I think.

test/windows/test_dll_init.c
mitchellh, line 1: This can go in `test/windows` and should probably have a [very short] README of how to run it and so on.

mitchellh (changes requested):

src/build/GhosttyLib.zig
deblasis: sorry, in the arena, learning things

mitchellh (approved):
Approved pending CI.

PR #11870, opened 2026-03-26
Move clipboard encoding including OSC52 to `src/terminal/clipboard.zig`, expose to libghostty

mitchellh:
This is a bunch of struct/logic shuffling so we can expose clipboard operations to libghostty. Lots of files touched because a lot of things moved and renamed but logically all the same. The only net new logic are the new C APIs + example.


mitchellh:
Okay what the fuck why is this causing any kind of windows changes.

mitchellh:
I've basically reverted all the apprt stuff which shouldn't have affected anything anyways. Just trying to figure out why windows is failing since my only windows access is via CI whack-a-mole 😢

mitchellh:
Okay what the fuck. Abandoning this and going to work on something else. I have no idea why this is causing any sort of grief for windows.

PR #11873, opened 2026-03-26
terminal: redo trailing state capture in OSC parser

mitchellh:
Trailing state capture now is encapsulated in a struct `Capture` and all parsers access the data via `p.capture.trailing()` rather than directly from the writer.

This is primarily to prep for the OSC parser to be able to capture the entire sequence (not just the trailing part) so we can setup libghostty for fallback handlers so libghostty implementers can have custom OSC behaviors.

But, it has the benefit of making our OSC parser much cleaner too.

I'm doing some benchmarks now...

### DIFF
--- a/src/terminal/osc.zig
+++ b/src/terminal/osc.zig
@@ -301,12 +301,10 @@ pub const Parser = struct {
 
     /// Buffer for temporary storage of OSC data
     buffer: [MAX_BUF]u8,
-    /// Fixed writer for accumulating OSC data
-    fixed: ?std.Io.Writer,
-    /// Allocating writer for accumulating OSC data
-    allocating: ?std.Io.Writer.Allocating,
-    /// Pointer to the active writer for accumulating OSC data
-    writer: ?*std.Io.Writer,
+
+    /// Capture state. If this is set then we're actively capturing the
+    /// bytes coming into the parser.
+    capture: ?Capture,
 
     /// The command that is the result of parsing.
     command: Command,
@@ -369,9 +367,7 @@ pub const Parser = struct {
         var result: Parser = .{
             .alloc = alloc,
             .state = .start,
-            .fixed = null,
-            .allocating = null,
-            .writer = null,
+            .capture = null,
             .command = .invalid,
 
             // Keeping all our undefined values together so we can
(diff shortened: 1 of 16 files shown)


mitchellh:
cc @jcollie

src/terminal/osc.zig
```diff
@@ -451,31 +445,91 @@ pub const Parser = struct {
...
+            return self.writer.buffered();
+        }
+    };
+
+    /// Begin capturing trailing data. All inputs to next from this point
+    /// forward will be captured into the `self.capture.writer` buffer
+    /// which may be backed by either a fixed size or allocating buffer
+    /// depending on mode.
+    ///
+    /// Get the trailing data using `capturedTrailing()`. Do not access
```
jcollie: `capturedTrailing()` isn't the name of a function.

mitchellh: Ah yeah, used to be, before I moved it :)

jcollie (approved):

mitchellh:
Benchmarks:

```
$ hyperfine -N -w 10 './zig-out/bin/ghostty-bench-old +osc-parser --data=osc.txt' './zig-out/bin/ghostty-bench-new +osc-parser --data=osc.txt'
Benchmark 1: ./zig-out/bin/ghostty-bench-old +osc-parser --data=osc.txt
  Time (mean ± σ):      92.4 ms ±  17.9 ms    [User: 73.7 ms, System: 17.5 ms]
  Range (min … max):    84.5 ms … 141.9 ms    34 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: ./zig-out/bin/ghostty-bench-new +osc-parser --data=osc.txt
  Time (mean ± σ):      83.6 ms ±   0.6 ms    [User: 65.0 ms, System: 17.4 ms]
  Range (min … max):    82.6 ms …  84.9 ms    36 runs

Summary
  ./zig-out/bin/ghostty-bench-new +osc-parser --data=osc.txt ran
    1.11 ± 0.21 times faster than ./zig-out/bin/ghostty-bench-old +osc-parser --data=osc.txt
```

Its actually very consistently faster, which I can't fully explain atm.

jcollie:
> Its actually very consistently faster, which I can't fully explain atm.

That's a little surprising, given the scope of the changes so far. Maybe the code re-org has better cache locality?

mitchellh:
> That's a little surprising, given the scope of the changes so far. Maybe the code re-org has better cache locality?

Yeah, my only thinking is that union making the structure that much smaller. I re-ran the benchmarks with 100, 200, and 500 MB files (the results I posted were 100MB) and all of them show the new version to be ~10% faster, same.

mitchellh:
This is also a good example of AI doing bad things. I articulated this idea to an agent and it produced some garbage. So this PR is fully hand-written. You win some you lose some.

PR #11875, opened 2026-03-26
Implement blink attribute for text

brianc442:
Fix for issue #7703  

Add text_blink_visible flag to renderer thread. Text blink is toggled on the same timer as the existing cursor blink. All foreground rendering (glyphs, underline, overline, strikethrough) is skipped during the hidden phase, matching xterm behavior. 

Claude Code was used to locate where changes were needed in the codebase, but the changes were written myself.

### DIFF
--- a/src/renderer/Thread.zig
+++ b/src/renderer/Thread.zig
@@ -107,6 +107,11 @@ flags: packed struct {
     /// This is true when the view is focused. This defaults to true
     /// and it is up to the apprt to set the correct value.
     focused: bool = true,
+    
+    /// This is true when blinking text should be visible and false when
+    /// it should not be visible. This is toggled on a timer by the thread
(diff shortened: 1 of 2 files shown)


qwerasd205:
Is there a reason the skip logic isn't implemented the same way as it is for the invisible flag?

brianc442:
> Is there a reason the skip logic isn't implemented the same way as it is for the invisible flag?

Yes - I initially used the same if style.flags.blink: continue logic but when I tested, the text didn't actually blink. When I looked into why, it has to do with an underlying difference between invisible text and blinking text. Invisible text is explicitly skipped by the shaper in run.zig lines 63-67 since there are no glyphs rendered. The same is not true of blinking text so I wrapped each style individually instead.

PR #11884, opened 2026-03-27
GTK: Toggle background opacity keybind

### LINKED ISSUE #5047: Add `toggle_background_opacity` keybind action

rhodes-b:
Closes: https://github.com/ghostty-org/ghostty/issues/5047

And supersedes https://github.com/ghostty-org/ghostty/pull/11881 as it shouldn't need to touch the renderer at all

https://github.com/user-attachments/assets/976c3d73-d7f0-4b8e-94ad-32c0cdd453ed

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -765,12 +765,13 @@ pub const Application = extern struct {
             .search_total => Action.searchTotal(target, value),
             .search_selected => Action.searchSelected(target, value),
 
+            .toggle_background_opacity => return Action.toggleBackgroundOpacity(target),
+
             // Unimplemented
             .secure_input,
             .close_all_windows,
             .float_window,
             .toggle_visibility,
-            .toggle_background_opacity,
             .cell_size,
             .render_inspector,
             .renderer_health,
@@ -2736,6 +2737,25 @@ const Action = struct {
                 };
 
                 window.toggleWindowDecorations();
+                return true;
+            },
+        }
+    }
+
+    pub fn toggleBackgroundOpacity(target: apprt.Target) bool {
+        switch (target) {
+            .app => return false,
+            .surface => |core| {
+                const surface = core.rt_surface.surface;
+                const window = ext.getAncestor(
+                    Window,
+                    surface.as(gtk.Widget),
+                ) orelse {
+                    log.warn("surface is not in a window, ignoring toggle_background_opacity", .{});
+                    return false;
+                };
+
+                window.toggleBackgroundOpacity();
                 return true;
             },
         }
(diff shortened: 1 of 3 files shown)


mpatankar6:
This has a slight problem though.
It's not actually rendering purely your background color. Here's an example with a really low background opacity to make it apparent.


https://github.com/user-attachments/assets/e46b1d6f-6a46-43de-afe6-ca6cf992c6ee

Rather, it's the background layer and the GLArea are getting blended if the GLArea is drawing semi-transparent pixels, so you're not actually getting the background you want. It's subtle but it's noticeable.

This is why my PR #11881 sent a message to the surface renderer to stop drawing semi-transparent pixels, it solves this issue.

rhodes-b:
> This has a slight problem though. It's not actually rendering purely your background color. Here's an example with a really low background opacity to make it apparent.

I believe that problem already exists today without the toggle, I'm sure people would like a fix but im not sure it needs to be tied to the toggle PR
https://github.com/ghostty-org/ghostty/discussions/7957

I guess this would make another case where its more frequent though

mpatankar6:
The GTK .background class inserts an opaque background under the window, for me it's a just a dark gray (I think the color comes from the Adwaita theme). The previous logic made this class only active when the terminal background was fully opaque, so it had never had a visual effect. So I disagree, this would be introducing a new problem as it will be the first time the GTK background class and a non-opaque terminal background will be in use simultaneously.

I don't think your changes are sufficient to call this feature implemented. For some cases, e.g. having a background image or a very transparent terminal, it can result in the color looking very off, and in any case it's technically incorrect.

I understand the apprehension as this logic really should be GTK layer, but I don't think it's possible.

PR #8901, opened 2025-09-24
cli: use sh to launch editor

jcollie:
Fixes #8898

PR #11898, opened 2026-03-27
cli: +edit-config works properly when editor command contains arguments

jcollie:
If `$EDITOR` or `$VISUAL` contained arguments, not just the path to an editor (e.g. `zed --new`) `+edit-config` would fail because we were treating the whole command as a path. Instead, wrap the command with `/bin/sh -c <command>` so that the shell can separate the path from the arguments.

Fixes #11897


jcollie:
This was apparently fixed once before in #8901 but that must have only been applied to 1.2.x and not to 1.3.x.

mitchellh (approved):

PR #11920, opened 2026-03-28
fix: replace hardcoded locale.h constants with build-system TranslateC

i999rri:
Replace hardcoded locale.h constants and extern function declarations with build-system TranslateC, following the same pattern as pty.c.

This fixes LC_ALL being hardcoded to 6 (the musl/glibc implementation value), which is implementation-defined and differs on Windows MSVC (where LC_ALL is 0), causing `setlocale()` to crash with an invalid parameter error.

## Changes

- Added `src/os/locale.c` — includes `locale.h` for TranslateC
- Added TranslateC step in `src/build/SharedDeps.zig` (same pattern as pty.c)
- Replaced hardcoded constants and extern declarations in `src/os/locale.zig` with `@import("locale-c")`

## AI disclosure

Claude Code was used to assist with debugging and identifying this issue.

### DIFF
--- a/src/build/SharedDeps.zig
+++ b/src/build/SharedDeps.zig
@@ -135,6 +135,24 @@ pub fn add(
     // Every exe needs the terminal options
     self.config.terminalOptions().add(b, step.root_module);
 
+    // C imports for locale constants and functions
+    {
+        const c = b.addTranslateC(.{
+            .root_source_file = b.path("src/os/locale.c"),
+            .target = target,
+            .optimize = optimize,
+        });
+        if (target.result.os.tag.isDarwin()) {
+            const libc = try std.zig.LibCInstallation.findNative(.{
+                .allocator = b.allocator,
+                .target = &target.result,
+                .verbose = false,
+            });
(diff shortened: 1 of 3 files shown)


jcollie (changes requested):
This is fine, except we need to move away from using `@cImport`. See https://github.com/ziglang/zig/issues/20630 . See https://github.com/ghostty-org/ghostty/blob/main/src/build/SharedDeps.zig#L138-L163 for an example. Should take care of the other functions/constants copied from `locale.h` as well, not just `LC_ALL`.

jcollie (approved):
LGTM. The windows test failure is not related I believe.

i999rri:
@jcollie 
Thanks for review !

mitchellh:
Note that the Zig recommendation is that we put as much of our C stuff into a single C file and translate once. But, we can figure out if that makes sense and such later. This is good for now. Thank you!

mitchellh:
Putting this in 1.4 because all the various Windows build fixes were towards 1.4 anyways.

PR #11922, opened 2026-03-28
macos: add pid and tty properties to AppleScript terminal and App Intents TerminalEntity

TweedBeetle:
## Summary

Add `pid` and `tty` as read-only properties to:
- The AppleScript `terminal` class
- The App Intents `TerminalEntity`

This enables reliable process-to-terminal mapping for automation tools when multiple terminals share the same working directory.

## Changes

- `macos/Ghostty.sdef`: add `pid` and `tty` to the `terminal` class
- `macos/Sources/Features/AppleScript/ScriptTerminal.swift`: add AppleScript getters
- `macos/Sources/Ghostty/Ghostty.Surface.swift`: add Swift accessors for foreground PID and PTY name
- `src/apprt/embedded.zig`: add C bridge exports for foreground PID and PTY name
- `include/ghostty.h`: declare the new C API entry points
- `macos/Sources/Features/App Intents/Entities/TerminalEntity.swift`: surface the same values in App Intents

## Testing

- `zig build -Demit-macos-app=false`
- `/opt/homebrew/bin/nu macos/build.nu --configuration Debug --action build`
- `swiftlint lint 'macos/Sources/Features/AppleScript/ScriptTerminal.swift'`
- `swiftlint lint 'macos/Sources/Features/App Intents/Entities/TerminalEntity.swift'`
- `swiftlint lint 'macos/Sources/Ghostty/Ghostty.Surface.swift'`
- AppleScript smoke test against the built app:
  - `get {id, pid, tty, working directory} of terminal 1`
  - `get pid of every terminal`
  - `get tty of every terminal`
  - Verified returned `pid`/`tty` against `ps -p <pid> -o pid=,tty=,comm=` and `lsof -a -p <pid> -d cwd -Fn`

## AI Disclosure

This PR was developed with AI assistance (Codex). The contributor understands the changes and can explain the implementation.

Closes #11592
Closes #10756

### DIFF
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -20,6 +20,7 @@ const CoreInspector = @import("../inspector/main.zig").Inspector;
 const CoreSurface = @import("../Surface.zig");
 const configpkg = @import("../config.zig");
 const Config = configpkg.Config;
+const String = @import("../main_c.zig").String;
 
 const log = std.log.scoped(.embedded_window);
 
@@ -1707,6 +1708,23 @@ pub const CAPI = struct {
             .cell_width_px = surface.core_surface.size.cell.width,
             .cell_height_px = surface.core_surface.size.cell.height,
         };
+    }
+
+    /// Returns the PID of the foreground process for the surface PTY.
+    export fn ghostty_surface_foreground_pid(surface: *Surface) u64 {
+        return surface.core_surface.getProcessInfo(.foreground_pid) orelse 0;
+    }
+
+    /// Returns the PTY name for the surface. The returned string must be
+    /// freed by the caller via ghostty_string_free.
+    export fn ghostty_surface_tty_name(surface: *Surface) String {
+        const tty_name = surface.core_surface.getProcessInfo(.tty_name) orelse return .empty;
+        const copy = surface.app.core_app.alloc.dupeZ(u8, tty_name) catch |err| {
+            log.err("error allocating tty name err={}", .{err});
+            return .empty;
+        };
+
+        return .fromSlice(copy);
     }
 
     /// Update the color scheme of the surface.
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1115,6 +1115,8 @@ GHOSTTY_API void ghostty_surface_set_focus(ghostty_surface_t, bool);
 GHOSTTY_API void ghostty_surface_set_occlusion(ghostty_surface_t, bool);
 GHOSTTY_API void ghostty_surface_set_size(ghostty_surface_t, uint32_t, uint32_t);
 GHOSTTY_API ghostty_surface_size_s ghostty_surface_size(ghostty_surface_t);
+GHOSTTY_API uint64_t ghostty_surface_foreground_pid(ghostty_surface_t);
+GHOSTTY_API ghostty_string_s ghostty_surface_tty_name(ghostty_surface_t);
 GHOSTTY_API void ghostty_surface_set_color_scheme(ghostty_surface_t,
                                                      ghostty_color_scheme_e);
 GHOSTTY_API ghostty_input_mods_e ghostty_surface_key_translation_mods(ghostty_surface_t,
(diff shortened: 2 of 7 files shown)


src/apprt/embedded.zig
```diff
@@ -1303,6 +1304,19 @@ pub const CAPI = struct {
         }
     };
 
+    // ghostty_string_s
```
jcollie: This is partially duplicating source from `main_c.zig`. The `String` definition should be moved to a shared location (maybe `src/c/string.zig`?) and referenced from there.

https://github.com/ghostty-org/ghostty/blob/0d1f77bc4d0ce5ad8db72d5d05b7d91740bea818/src/main_c.zig#L62-L102

src/apprt/embedded.zig
```diff
@@ -1709,6 +1723,27 @@ pub const CAPI = struct {
...
+    /// Returns the PID of the foreground process for the surface PTY.
+    export fn ghostty_surface_foreground_pid(surface: *Surface) u64 {
+        return surface.core_surface.getProcessInfo(.foreground_pid) orelse 0;
+    }
+
+    /// Returns the PTY name for the surface. The returned string must be
+    /// freed by the caller via ghostty_string_free.
+    export fn ghostty_surface_tty_name(surface: *Surface) String {
+        const tty_name = surface.core_surface.getProcessInfo(.tty_name) orelse return .empty;
+        const copy = global_state.alloc.dupeZ(u8, tty_name) catch |err| {
```
jcollie: The allocator should be available from `surface.app.core_app.alloc` thus making the import of `global.zig` above unnecessary.

jcollie (changes requested):
Only commenting on the Zig code. The Swift code will need review by someone else.

src/apprt/embedded.zig
mitchellh: We have a "standard" libghostty string type in `src/lib/string.zig` (we use for libghostty-vt). We should probably restandardize on that. Short of that though I agree with jcollie here.

mitchellh (changes requested):
Swift side looks fine, agree on the Zig review from @jcollie

src/apprt/embedded.zig
TweedBeetle: Addressed in `82e66aa17`. I extracted the C API `String` into `src/c/string.zig` for this PR and updated `main_c.zig`, `embedded.zig`, and `config/CApi.zig` to share it. Agreed that any broader libghostty string restandardization would be follow-up work.

src/apprt/embedded.zig
TweedBeetle: Addressed in `82e66aa17`. `ghostty_surface_tty_name` now allocates via `surface.app.core_app.alloc`, and the extra `global_state` import in `embedded.zig` is gone.

src/c/string.zig
mitchellh, line 1: Sorry actually we shouldn't create a whole new `c` directory for this. This should live somewhere else but I'm not sure where yet. Let me think about it.

mitchellh (dismissed):

src/c/string.zig
TweedBeetle: Ok, no problem 👍

TweedBeetle:
Happy to revert `src/c/` and just import String from `main_c.zig` in `embedded.zig` if that's simpler for now. Or I can move it wherever you'd prefer. Just let me know 🫡

bo2themax (approved):
AppIntents/Shortcut.app builds their cache really weirdly, it took me some time to make it update to new entity, but it worked.

Image

TweedBeetle:
Hey @mitchellh, just wanted to check in on this. Happy to go whichever direction on the Zig file placement. As far as I understand, there are three options:

1. Revert `src/c/` and import String from `main_c.zig` in `embedded.zig`
2. Move the file somewhere else you'd prefer
3. Keep current approach

Let me know if you have thoughts. Everything else is approved 🫡

TweedBeetle:
Went ahead with option 1 to unblock: reverted `src/c/` and inlined the `String` type back into `main_c.zig` (force-pushed). Zig diff is now a no-op aside from the two new CAPI exports in `embedded.zig`. Happy to move the extraction elsewhere if you'd still prefer that, just let me know and I'll reshape it.

TweedBeetle:
Rebased onto current main, so this is no longer stale. I'm happy to adjust placement or naming if anything needs changing.

bo2themax (approved):

jcollie (approved):

PR #11930, opened 2026-03-28
benchmark: disable test on windows

jcollie:
We don't appear to have a time source with enough resolution to get a non-zero duration on the benchmark test so it fails.


mitchellh:
I actually see this fail on Linux sometimes too, I suspect it's not timer resolution but actually just that the test is sometimes too fast. Still, willing to merge.

PR #11934, opened 2026-03-28
fix(build): re-archive object files for correct Mach-O alignment befo…

zoptia:
…re libtool

Zig's ar implementation does not guarantee 8-byte alignment for 64-bit Mach-O object files within archives. Apple's libtool silently drops misaligned members, producing incomplete static libraries on macOS.

Work around this by extracting each input library and re-archiving its objects with the system /usr/bin/ar (which produces correctly aligned archives) before combining them with libtool.

### DIFF
--- a/src/build/LibtoolStep.zig
+++ b/src/build/LibtoolStep.zig
@@ -27,13 +27,44 @@ output: LazyPath,
 
 /// Run libtool against a list of library files to combine into a single
 /// static library.
+///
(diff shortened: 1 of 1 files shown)


mitchellh:
This looks very interesting, but I would love a discussion or more context on how you triggered this and where it affected you etc.

jcollie:
Replaced with #11999

PR #11940, opened 2026-03-28
cli: add pager support to +explain-config

jparise:
Add a new Pager type that wraps output to an external pager program when stdout is a TTY, following the same conventions as git. The pager command is resolved from $PAGER, falling back to `less`. An empty $PAGER disables paging. If the pager fails to spawn, we fall back to stdout.

Previously, +explain-config wrote directly to stdout with no paging, which meant long help text would scroll by. Now output is automatically piped through the user's preferred pager when running interactively. A --no-pager flag is available to disable this.

### DIFF
--- a/src/cli/Pager.zig
+++ b/src/cli/Pager.zig
@@ -1,7 +1,9 @@
 //! A pager wraps output to an external pager program (like `less`) when
-//! stdout is a TTY. The pager command is determined by `$PAGER`, falling
-//! back to `less` if `$PAGER` isn't set.
+//! stdout is a TTY. The pager command is resolved as:
 //!
+//!   `$GHOSTTY_PAGER` > `$PAGER` > `less`
+//!
+//! Setting either env var to an empty string disables paging.
 //! If stdout is not a TTY, writes go directly to stdout.
 const Pager = @This();
 const std = @import("std");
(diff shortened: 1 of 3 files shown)


src/cli/Pager.zig
```diff
@@ -0,0 +1,87 @@
...
+    self.* = undefined;
+}
+
+fn initPager(alloc: Allocator) ?std.process.Child {
+    const stdout_file: std.fs.File = .stdout();
+    if (!stdout_file.isTty()) return null;
+
+    // Resolve the pager command: $PAGER > "less".
+    // An empty $PAGER disables paging.
+    const env_result = internal_os.getenv(alloc, "PAGER") catch null;
```
tristan957: I think we should support a `GHOSTTY_PAGER` similar to `GIT_PAGER`. Maybe I want to disable paging in just Ghostty for instance.

tristan957 (commented):
We should support this in `+show-config` too.

src/cli/Pager.zig
jparise: I thought about that but wasn't sure if it was just speculative or a real use case. It would be easy to implement though.

jcollie (approved):
LGTM

PR #11950, opened 2026-03-29
libghostty: make headers C++ compatible

mitchellh:
The headers were not C++ compatible and would fail compiling before (see https://github.com/ghostty-org/ghostty/discussions/11878). The only reason is because our typedefs would conflict since we named them identically. 

This also adds a `c-vt-stream` example and a `cpp-vt-stream` example, the latter primarily to verify we can build in C++ mode.

### DIFF
--- a/example/cpp-vt-stream/README.md
+++ b/example/cpp-vt-stream/README.md
@@ -1,0 +1,19 @@
+# Example: VT Stream Processing in C++
+
+This contains a simple example of how to use `ghostty_terminal_vt_write`
+to parse and process VT sequences in C++. This is a simplified C++ port
+of the `c-vt-stream` example that verifies libghostty compiles in C++
+mode.
+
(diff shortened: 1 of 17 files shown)


example/cpp-vt-stream/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This contains a simple example of how to use `ghostty_terminal_vt_write`
+to parse and process VT sequences in C++. This is a simplified C++ port
+of the `c-vt-stream` example that verifies libghostty compiles in C++
+mode.
+
+> [!IMPORTANT]
+>
+> **`libghostty` is a C library.** This example is only here so our CI
+> verifies that the library can be built in used from C++ files.
```
00-kat, line 11:
```suggestion
> verifies that the library can be used from C++ files.
```

PR #11955, opened 2026-03-29
fix(libghostty): ensure memory is zeroed in runtime safety modes for wasm/freestanding

elias8:
Zero page memory on freestanding/wasm where allocator reuses freed slots without clearing. Allows recreating a new or multiple terminals.


mitchellh:
Oh dang. This is going to suck for some aspects of performance but it must be done.

PR #11958, opened 2026-03-29
build(libghostty): export function table for wasm to allow host callbacks

elias8:
Export function table for wasm module to allow host callbacks (effects).


elias8:
We also need `--growable-table`, but zig doesn't seem to support it yet. Not sure whether this should be something to address in the ghostty build pipeline (workaround for now) 🤔

mitchellh:
Replaced by #12003

PR #11969, opened 2026-03-30
core: restructure args parsing to allow `+`-prefixed args after `-e`

jcollie:
Previously, the argument parser would not allow `+`-prefixed args after `-e`. Depending on the situation, it would either raise an error about multiple actions found, or it would just skip the argument as if it wasn't there. In practice, it made commands like the following impossible:

```
ghostty -e hx +10 src/main.zig
```
```
ghostty +new-window -e vim src/main.zig '+normal!10Gvz5'
```

This PR changes argument parsing so that it stops looking for actions once an argument like `-e` has been seen. It also stores a copy of the arguments with the action filtered out so that subsequent operations that parse the CLI arguments again (like individual actions do) do not need to worry about filtering out actions.

This also parepares Ghostty for 0.16 a bit where the CLI arguments aren't available "at will" just by calling `std.process.argsWithAllocator`.

Fixes #11937

### DIFF
--- a/src/benchmark/cli.zig
+++ b/src/benchmark/cli.zig
@@ -14,6 +14,8 @@ pub const Action = enum {
     @"is-symbol",
     @"osc-parser",
 
+    pub const ParseResult = cli.action.ActionParser(@This());
+
     /// Returns the struct associated with the action. The struct
     /// should have a few decls:
     ///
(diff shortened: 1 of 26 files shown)


mitchellh (changes requested):
There's a lot here so its going to take me awhile to get to this.

If there is any way to break this down into smaller atomic chunks that would help. I'm getting caught up in balancing the (1) passed in iterator (2) fallback (3) special case mix of things.

The extra 5 lines of boilerplate (array list init + 4 line defer) is also a huge code smell to me that something is funky here, but obv just a smell. I don't understand the context enough to judge it completely yet.

I'll take a look at this, just wanted to give you the status.

BirdeeHub:
Would this fix https://github.com/ghostty-org/ghostty/discussions/12004 as well?

PR #11977, opened 2026-03-30
libghostty: add GHOSTTY_EXPORT for shared lib symbol visibility

deblasis:
## Summary

Add a `GHOSTTY_EXPORT` annotation macro to all public function declarations across both `ghostty.h` (the main libghostty header) and the `include/ghostty/vt/` headers (the libghostty-vt API). This is the standard pattern used by C libraries that support both static and shared library builds.

On Windows, functions need `__declspec(dllexport)` when building the DLL and `__declspec(dllimport)` when consuming it, or they won't be visible across the DLL boundary. On Linux/macOS with GCC/Clang, `__attribute__((visibility("default")))` keeps the public API visible when building with `-fvisibility=hidden`, which reduces symbol table size and avoids collisions.

The macro resolves to nothing for static builds (when `GHOSTTY_STATIC` is defined) and on compilers that don't support visibility attributes, so this is a no-op for the current macOS static library path.

## Why

I looked at how popular C libraries handle this and every serious one follows the same pattern:

- SDL (`SDL_DECLSPEC`)
- cURL (`CURL_EXTERN`)
- SQLite (`SQLITE_API`)
- zlib (`ZEXTERN`)
- FreeType (`FT_EXPORT`) -- already vendored by Ghostty
- GLFW (`GLFWAPI`)
- Lua (`LUA_API`)

The header comment says "the API is built to be more general purpose" and the long-term goal is libghostty as a reusable library. Export annotations are table stakes for that -- they explicitly mark the public API surface, enable proper shared library builds on all platforms, and give consumers the right linker hints.

## Test plan

- [x] Windows build and full test suite
- [x] Linux build and full test suite
- [x] macOS build, full test suite, and app launch verified working
- [x] macOS xcodebuild app build and launch verified working
- [x] Shared library symbol inspection on all three platforms
- [x] Linux: validated version script + LLD restricts exports to only ghostty_* (107/107, 0 leaked, 12 MB .so)
- [x] Linux: C link test against restricted .so -- compiled, linked, ran successfully
- [x] Windows: DLL exports verified (102 ghostty_ + 3 unavoidable CRT/simdutf)


mitchellh (approved):
I don't think we _really_ need this for `include/ghostty.h` because that's never meant to really be external, but it doesn't hurt. 

I also know if `types.h` is the right place for this but it does seem to be the most root-level header so okay.

Also in the future, please make your PR descriptions like 90% shorter. AI likes talking a lot but most of what it says is completely worthless.

deblasis:
Noted. I will adjust from now on. Thanks for the feedback. Totally my fault because I went on and gathered more info thinking I was doing the right thing. I shrank it now.

Looking into the ci failure. I didn't test examples. 🤦  I will add them to my poor man's ci

PR #11978, opened 2026-03-30
macOS: fix focus update when using search or command palette

bo2themax:
This fixes two things:

1. Surface focus state is not consistent with first responder state when the search bar is open.
  > Reproduce: Open search, switch to another app and back, observe the cursor state of the surface.
  > And after switching back, `cmd+shift+f` will close the search bar, surface will become focused but not first responder, so it will not accept any input
2. Command palette is not focused when built with Xcode 26.4 (26.3 works fine).
  > This is weird to me, because the tip (and built with 26.3) works fine. I guess it's related to the SDK update? I couldn’t be sure what went wrong, but dispatching it to the next loop works as previously.
  > Also cleaned some previous checks when quickly open and reopen.
  > This fix works great both with 26.4 and 26.3

https://github.com/user-attachments/assets/c9cf4c1b-60d9-4c71-802c-55f82e40eec7

### DIFF
--- a/macos/Sources/Features/Command Palette/CommandPalette.swift
+++ b/macos/Sources/Features/Command Palette/CommandPalette.swift
@@ -64,7 +64,6 @@ struct CommandPaletteView: View {
     @State private var query = ""
     @State private var selectedIndex: UInt?
     @State private var hoveredOptionID: UUID?
-    @FocusState private var isTextFieldFocused: Bool
 
     // The options that we should show, taking into account any filtering from
     // the query. Options with matching leadingColor are ranked higher.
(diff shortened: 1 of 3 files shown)


macos/Sources/Features/Command Palette/CommandPalette.swift
```diff
@@ -178,27 +177,13 @@ struct CommandPaletteView: View {
         .padding()
         .environment(\.colorScheme, scheme)
         .onChange(of: isPresented) { newValue in
-            // Reset focus when quickly showing and hiding.
-            // macOS will destroy this view after a while,
-            // so task/onAppear will not be called again.
-            // If you toggle it rather quickly, we reset
-            // it here when dismissing.
-            isTextFieldFocused = newValue
-            if !isPresented {
```
bo2themax, line 187: This is redundant, since there is no animations anymore

PR #11980, opened 2026-03-30
macOS: close search bar if needed when it loses focus

bo2themax:
This adds features like:

1. Clicking outside of search bar works like typing `escape`
2. Typing `tab` while search bar is focused also works like typing `escape`

https://github.com/user-attachments/assets/a51f1560-ed14-4002-81b4-96eb927b17ca


mitchellh:
Rebase your PRs please!

mitchellh:
I'm not like 100% if these are good behaviors but I tagged for 1.4 so we can let it sit for awhile to find out.

PR #11981, opened 2026-03-30
feat(renderer): add iSurfaceOffset and iWindowResolution shader uniforms

yabbal:
## Summary

Custom shaders currently have no way to know where their surface sits within the window. Each split gets its own coordinate space starting at `(0,0)` with `iResolution` set to its own dimensions. This makes it impossible to create effects that span the full window (gradients, vignettes, CRT curvature) — they get duplicated per split instead.

This PR adds two new Shadertoy uniforms:

- **`iSurfaceOffset`** (`vec4`) — `.xy` = pixel position of this surface's top-left corner within the window
- **`iWindowResolution`** (`vec4`) — `.xyz` = total window size in pixels (same convention as `iResolution`)

With these, shaders can compute window-global UVs:

```glsl
vec2 globalUV = (fragCoord.xy + iSurfaceOffset.xy) / iWindowResolution.xy;
```

### Changes

- **Shader interface**: 2 new uniforms in `shadertoy.zig` + `shadertoy_prefix.glsl`
- **Renderer**: new `surface_offset` / `window_size` fields, populated per-frame, with safe defaults (offset=0, window=surface size)
- **Message system**: new `surface_position` message for thread-safe updates
- **Core Surface**: new `surfacePositionCallback` for apprts to report position
- **Embedded (macOS)**: new C API `ghostty_surface_set_position()` for the Swift host to call on split layout changes
- **GTK**: plumbing in place, calls the callback on every resize. Position computation is stubbed with a TODO — needs `gtk_widget_compute_point` / `gtk_widget_compute_bounds` for the actual offset

### Backward compatibility

Fully backward-compatible. Defaults (`offset=0,0`, `window_resolution=iResolution`) mean existing shaders behave identically. No existing code paths are modified.

### Related

- Closes the shader aspect of #10053 (unified background across splits)
- Enables window-spanning effects for the custom shader ecosystem (gradients, vignettes, unified CRT, etc.)

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1091,6 +1091,8 @@ void ghostty_surface_set_content_scale(ghostty_surface_t, double, double);
 void ghostty_surface_set_focus(ghostty_surface_t, bool);
 void ghostty_surface_set_occlusion(ghostty_surface_t, bool);
 void ghostty_surface_set_size(ghostty_surface_t, uint32_t, uint32_t);
+void ghostty_surface_set_position(ghostty_surface_t, uint32_t, uint32_t,
+                                  uint32_t, uint32_t);
 ghostty_surface_size_s ghostty_surface_size(ghostty_surface_t);
 void ghostty_surface_set_color_scheme(ghostty_surface_t,
                                       ghostty_color_scheme_e);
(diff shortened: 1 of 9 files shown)


tristan957:
/vouch @yabbal 

This looks fairly innocuous.

tristan957:
!vouch @yabbal

PR #12007, opened 2026-03-31
macOS: Make version in about dialog clickable

### LINKED ISSUE #11964: macOS: Make version in about dialog clickable
### Discussed in https://github.com/ghostty-org/ghostty/discussions/11956

<sup>Originally posted by **sdavids** March 29, 2026</sup>
Make the version on the About dialog clickable.

Similar to the commit hash.

The link would open the release notes for the particular version:

`1.3.1` would open https://ghostty.org/docs/install/release-notes/1-3-1 for example.

version
</div>

louisunlimited:
- Fixes: https://github.com/ghostty-org/ghostty/issues/11964

Made a private enum type `VersionConfig` to reference whether the release is a semver or tip, makes it easier for later in the view to `switch` between cases.

I do think there could be a better place for this enum or we can get rid of it, open to opinions. Right now version parsing is kind of duplicated between `AboutView` and `UpdateModalView` so we can also extract to a common helper if wanted.

Tested by manually setting `Marketing Version` in build settings to 

`1.3.1`
Screenshot 2026-03-30 at 18 31 15

`332b2aefc`
Screenshot 2026-03-30 at 18 32 48

### AI Disclosure
I used Sonnet 4.6 to understand where the version strings came from and in what format, it read release yml files to see what's going on. Then it proposed really bad code so I manually went in and cleaned up the view.

### DIFF
--- a/macos/Sources/Features/About/AboutView.swift
+++ b/macos/Sources/Features/About/AboutView.swift
@@ -15,9 +15,10 @@ struct AboutView: View {
         case stable(version: String, url: URL?)
         case tip
         case other(String)
+        case none
 
         init(version: String?, docsURL: URL?) {
-            guard let version else { self = .other(""); return }
+            guard let version else { self = .none; return }
             if version.range(of: #"^\d+\.\d+\.\d+$"#, options: .regularExpression) != nil {
                 let slug = version.replacingOccurrences(of: ".", with: "-")
                 self = .stable(version: version, url: docsURL?.appendingPathComponent("install/release-notes/\(slug)"))
@@ -88,9 +89,9 @@ struct AboutView: View {
                         PropertyRow(label: "Version", text: version, url: url)
                     case .tip:
                         PropertyRow(label: "Version", text: "Tip Release")
-                    case .other(let v) where !v.isEmpty:
+                    case .other(let v):
                         PropertyRow(label: "Version", text: v)
-                    default:
+                    case .none:
                         EmptyView()
                     }
                     if let build {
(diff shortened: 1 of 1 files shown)


macos/Sources/Features/About/AboutView.swift
```diff
@@ -10,6 +10,30 @@ struct AboutView: View {
     private var build: String? { Bundle.main.infoDictionary?["CFBundleVersion"] as? String }
     private var commit: String? { Bundle.main.infoDictionary?["GhosttyCommit"] as? String }
     private var version: String? { Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String }
+
+    private enum VersionConfig {
+        case stable(version: String, url: URL?)
```
mitchellh: I think it'd be cleaner to do `case stable(version: String)` and `case tip(commit: String?)` and then make `var url: URL?` a computed var (that could yield nothing). I'm not going to block merging on that but if you want to do some clean up...

mitchellh (approved):
Some improvements could be made but will accept as-is!

macos/Sources/Features/About/AboutView.swift
louisunlimited: I see what you are suggesting, I could be wrong, but won't this create another hard-coded `docURL` if we remove `url: URL?`? Since `docURL` is a `private let` we can't directly access it from nested `enum` types. We can make it a `static let`, but not sure if this is a good reason to do that just for this one use case.

I pushed a new commit if you want to take another look!

mitchellh (approved):
Yeah that looks cleaner to me!

louisunlimited:
> Yeah that looks cleaner to me!

Nice! Happy to help, thank you for this awesome software!

PR #12027, opened 2026-03-31
core: add 64 bit unique ID to every core surface

jcollie:
- Expose that ID as the environment variable GHOSTTY_SURFACE_ID to processes running in Ghostty surfaces.
- Add a function to the core app to search for surfaces by ID.
- ID is randomly generated, it has no other meaning other than as a unique identifier for the surface. The ID also cannot be zero as that is used to indicate a null ID in some situations.

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -524,6 +524,16 @@ fn hasSurface(self: *const App, surface: *const Surface) bool {
     return false;
 }
 
+/// Search for a surface by a 64 bit unique ID.
+pub fn findSurfaceByID(self: *const App, id: u64) ?*Surface {
+    for (self.surfaces.items) |v| {
+        const surface: *Surface = v.core();
+        if (surface.id == id) return surface;
+    }
+
+    return null;
+}
+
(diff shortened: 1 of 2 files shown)


jcollie:
[As requested by](https://github.com/ghostty-org/ghostty/pull/11762#issuecomment-4150472696) @tristan957 I've broken the addition of a surface ID into a separate PR so that the format of the ID can be bikeshed. For my purposes, the format of the ID does not matter as long as I can treat it as an opaque number, and a `0` value is an invalid ID that can be used to indicate a null ID in environments that do not have native `null` support (like DBus messages).

A surface ID is important for adding a `+new-tab` function and other scripting-like functionality to the GTK interface, as well as eliminate the usage of `@ptrToInt`/`@intToPtr` in GTK notifications.

mitchellh (approved):
I'm good with this for now. I do think it might make sense to add a bit more structure to these IDs in the future but we're not making any persistence guarantees about them so we can change the format at anytime as long as its 64-bits.

PR #12029, opened 2026-03-31
core/gtk: ensure that first surface gets marked as focused surface by app

jcollie:


mitchellh (approved):
Please milestone this, but approved

jcollie:
AFAIK this makes no difference operationally, except for the `+new-tab` action PR (which is how I discovered the problem), so there's no benefit for 1.3.x.

PR #12030, opened 2026-04-01
terminal: implement kitty clipboard protocol read path (OSC 5522)

Jarred-Sumner:
Partially addresses #10549 (macOS & read path only; write path is separate work).

Implements the read path of the Kitty Clipboard Protocol (OSC 5522, [spec](https://rockorager.dev/misc/bracketed-paste-mime/)). When mode 5522 is enabled, paste events send a MIME type listing instead of bracketed-paste text. The application then requests a specific MIME type and receives the clipboard data — including images and other non-text content — chunked through OSC sequences.

## Flow

1. Application enables mode 5522 via `CSI ? 5522 h`
2. User pastes. Instead of bracketed paste, the surface asks the apprt for available MIME types and sends an unsolicited `OSC 5522` listing with a single-use 16-byte password
3. Application sends `OSC 5522;type=read:mime=<b64>:password=<b64>` requesting one type
4. Surface validates the password, invalidates it (single-use), asks the apprt for that MIME type's bytes, and streams them back as base64-chunked DATA packets

## Security model

The password mechanism prevents an application from reading the clipboard outside of an explicit user paste:

- 16 random bytes from `std.crypto.random`, compared with `std.crypto.timing_safe.eql`
- 5-second timeout. The spec requires a short validity window so a stashed password can't authorize reads of *future* clipboard contents. Platforms without a monotonic clock decline to issue a password rather than skip the timeout
- Single-use: cleared as soon as a request validates, before the apprt is called
- Clipboard location (standard vs selection) captured at issue time so a request can't redirect
- A read with no/invalid password falls into the existing `clipboard-read` policy. For now that path returns `ENOSYS`

## Status codes

We emit `EINVAL` for malformed requests and `EIO` for apprt failures. The kitty spec only defines `ENOSYS`/`EPERM`/`EBUSY`, but the existing parser `Status` enum already has all five. Happy to collapse to spec-only codes if preferred — used the wider set for diagnostic precision.

`encodeError` takes the `Status` enum directly, so a typo would be a compile error rather than a non-spec status on the wire.

## Platform support

**macOS** is fully wired. `NSPasteboard` types map to MIME via `UTType.preferredMIMEType`; data path uses `data(forType:)` returning base64.

**GTK** returns `false` from `clipboardRequest` for the new request types, so the surface falls through to a normal text paste — enabling mode 5522 on GTK is a no-op rather than breaking paste. The hook is `gdk.Clipboard.getFormats()` + `readAsync()`; left as a TODO so this PR stays reviewable.

**libghostty** gets two C APIs (`ghostty_clipboard_request_type`, `ghostty_clipboard_request_mime`) so embedders can branch on the request type.

## Chunking

The spec caps DATA packets at 4096 raw bytes. We slice a pre-encoded base64 stream rather than re-encoding per chunk, so middle chunks have no padding: N base64 chars decode to `N/4*3` raw bytes. `(4096/3)*4 = 5460` chars → 4095 raw bytes per chunk. Asserted at comptime, with a test pinning both numbers.

The output buffer is pre-sized to avoid realloc growth on multi-megabyte images, and `completeKittyMimeRead` hands ownership directly to `.write_alloc` to skip the transient dupe that `writeReq` would do.

## MIME length cap

256 bytes, shared between the encoder and `ClipboardRequest.KittyMimeRead` (`comptime assert` keeps them in sync). MIME types over the limit are dropped from the advertised list — we don't advertise something the request struct can't carry back. 256 covers all IANA-registered types including the OpenXML Office types (65–73 bytes); there's a test with the real `.xlsx`/`.docx`/`.pptx` strings.

## Not in this PR

- Write path (`type=write`/`wdata`/`walias`). Returns `ENOSYS`.
- Unauthenticated reads with user prompt
- GTK MIME implementation

These build on the same plumbing.

## Test plan

- [x] `zig build test` on macOS arm64 — 2705/2717 passed, 10 skipped, 2 pre-existing `os/path` failures from a local `$PATH` quirk (`NotDir` at `path.zig:53`, untouched by this PR)

50 new tests:

- `kitty_clipboard.zig` (32, new file): byte-exact spec example reproduction (`secret123` → `c2VjcmV0MTIz`, etc.), `checkAllAllocationFailures` for both encoders, chunk-boundary data integrity at 5 sizes, capacity estimate verification, OSC framing safety with adversarial passwords containing `0x1b`, encoder→parser round-trips
- OSC parser (+9): adversarial — null bytes, key prefix collisions, embedded ST, 10K metadata
- `stream.zig` (+3), `stream_terminal.zig` (+3): mode 5522 dispatch and effects callback
- `Surface.zig` (+2): timeout constant pin, fresh password not expired
- `KittyMimeRead.init` doctest: boundary cases, real Office MIME types

Manual: pasted a PNG over SSH and it worked, got correct `image/png` data through.

## AI disclosure

I used Claude Code to implement basically all of this. I reviewed it and slightly edited it.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -81,6 +81,8 @@ typedef enum {
   GHOSTTY_CLIPBOARD_REQUEST_PASTE,
   GHOSTTY_CLIPBOARD_REQUEST_OSC_52_READ,
   GHOSTTY_CLIPBOARD_REQUEST_OSC_52_WRITE,
+  GHOSTTY_CLIPBOARD_REQUEST_KITTY_MIME_LIST,
+  GHOSTTY_CLIPBOARD_REQUEST_KITTY_MIME_READ,
 } ghostty_clipboard_request_e;
 
 typedef enum {
@@ -1152,6 +1154,8 @@ GHOSTTY_API void ghostty_surface_complete_clipboard_request(ghostty_surface_t,
                                                                const char*,
                                                                void*,
                                                                bool);
+GHOSTTY_API ghostty_clipboard_request_e ghostty_clipboard_request_type(void*);
+GHOSTTY_API const char* ghostty_clipboard_request_mime(void*);
 GHOSTTY_API bool ghostty_surface_has_selection(ghostty_surface_t);
 GHOSTTY_API bool ghostty_surface_read_selection(ghostty_surface_t, ghostty_text_s*);
 GHOSTTY_API bool ghostty_surface_read_text(ghostty_surface_t,
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -334,6 +334,27 @@ extension Ghostty {
             // Get our pasteboard
             guard let pasteboard = NSPasteboard.ghostty(location) else { return false }
 
+            // Check the request type to handle kitty clipboard protocol
+            switch ghostty_clipboard_request_type(state) {
+            case GHOSTTY_CLIPBOARD_REQUEST_KITTY_MIME_LIST:
+                let mimes = pasteboard.availableMimeTypes()
+                guard !mimes.isEmpty else { return false }
+                completeClipboardRequest(surface, data: mimes.joined(separator: "\n"), state: state)
+                return true
+
+            case GHOSTTY_CLIPBOARD_REQUEST_KITTY_MIME_READ:
+                guard let mimePtr = ghostty_clipboard_request_mime(state),
+                      let mime = String(cString: mimePtr, encoding: .utf8),
+                      let data = pasteboard.data(forMimeType: mime) else {
+                    return false
+                }
+                completeClipboardRequest(surface, data: data.base64EncodedString(), state: state)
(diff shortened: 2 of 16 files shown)


mitchellh:
!vouch

mitchellh:
I told Jarred I’d review this.

jcollie:
@Jarred-Sumner there are some conflicts now, can you rebase this?

mindfulmonk:
@Jarred-Sumner @jcollie - I've rebased with Claude Code here https://github.com/mindfulmonk/ghostty/tree/kitty-clipboard-read-clean 

The tests pass, I tested the resulting build and I'm able to paste images over ssh.

epicwhale:
Just wanted to say this would be incredibly useful to have in Ghostty! a daily papercut in my life when using remote ssh 😇

jcollie:
@Jarred-Sumner we have some more conflicts, can you rebase?

mitchellh:
Closing, this is coming in a better shape starting here: https://github.com/ghostty-org/ghostty/pull/13962

PR #12036, opened 2026-04-01
gtk(chore): fix typos

bo2themax:
### AI Disclosure

Claude wrote the regex to ignore base64-encoded sequences


pluiedev (approved):
While I'm not a huge fan of `duplicated`, I'd take it if it means my confige won't be changed

PR #12045, opened 2026-04-01
fix: Ensure snap paths come first in gio module loading

kenvandine:
This fixes the issue reported in https://github.com/ghostty-org/ghostty/discussions/11311


jcollie (approved):
LGTM pending CI

PR #12050, opened 2026-04-01
Fix mis-use of libxev.Async and copy-by-value issue with render thread wakeup

justonia:
This PR encapsulates the fix for the issue I identified in this discussion post: https://github.com/ghostty-org/ghostty/discussions/11877

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -679,7 +679,7 @@ pub fn init(
             .backend = .{ .exec = io_exec },
             .mailbox = io_mailbox,
             .renderer_state = &self.renderer_state,
-            .renderer_wakeup = render_thread.wakeup,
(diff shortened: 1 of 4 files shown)


mitchellh:
!vouch

mitchellh:
We may need to look into this for Windows in libxev because xev does except watchers to be copy able I THINK. I just want to verify and then consider.

justonia:
> We may need to look into this for Windows in libxev because xev does except watchers to be copy able I THINK. I just want to verify and then consider.

Sounds good, it was really unclear from the code what the API contract was intended to be.

mitchellh:
Yeah the fix is we need to fix libxev to make Async copy-by-value on Windows.

PR #12090, opened 2026-04-03
add a nix package (with CI tests) for libghostty-vt

jcollie:


nix/libghostty-vt.nix
```diff
@@ -0,0 +1,68 @@
...
+  callPackage,
+  git,
+  pkg-config,
+  zig_0_15,
+  revision ? "dirty",
+  optimize ? "Debug",
+}:
+stdenv.mkDerivation (finalAttrs: {
+  pname = "ghostty";
+  version = "1.3.2-dev";
```
mitchellh: libghosty-vt won't follow Ghostty's version so lets reset this 0.1.0

mitchellh (changes requested):
Thank you very much!

I forgot what Nix calls it, but does this properly setup the thing where you can just build the headers vs. the lib vs ... ?

nix/libghostty-vt.nix
jcollie: Fixed.

jcollie:
Split the headers and other development-only files into dev subpackage.

mitchellh (approved):
Looks right, at least good enough for a start. Thank you!

PR #12099, opened 2026-04-04
terminal/tmux: ignore %output before pane initialization

h3nock:
## What

Ignore tmux `%output` for panes until their initial bootstrap sync is complete.

## Why

Currently, a pane is added to the viewer as soon as it appears in the layout and
its initial `capture-pane` and `pane_state` sync happens afterward. That means
live `%output` can arrive in the middle of bootstrap and get fed into the VT
stream before the pane has finished loading its initial screen contents and
terminal state.

With this change, `%output` is ignored until `pane_state` has been applied for
that pane.

This also complements the lazy decoded() direction discussed in [#12076](https://github.com/ghostty-org/ghostty/pull/12076): we can move `%output` decoding behind the initialized check and skip decoding work for uninitialized panes.

## Changes

- add `Pane.initialized` flag to the tmux viewer
- set `pane.initialized = true` after its pane_state is applied
- return early from `receivedOutput()` for panes that are not initialized
- update `test "initial flow"` to:
  - explicitly verify the final `capture-pane` response advances to `list-panes`
  - verify `%output` is ignored before `pane_state`
  - verify panes become initialized after `pane_state`
  - verify `%output` is accepted once initialization is complete

 ---

 AI disclosure: Claude Code Opus 4.6 and Codex 5.4 xhigh were used for research, understanding Zig and helping me convert my ideas to Zig.
 
 ---

### DIFF
--- a/src/terminal/tmux/viewer.zig
+++ b/src/terminal/tmux/viewer.zig
@@ -255,6 +255,9 @@ pub const Viewer = struct {
 
     pub const Pane = struct {
         terminal: Terminal,
+        /// false until this pane's initial capture-pane + pane_state sync
+        /// has been applied. live %output is ignored untill then.
+        initialized: bool = false,
 
         pub fn deinit(self: *Pane, alloc: Allocator) void {
             self.terminal.deinit(alloc);
@@ -1029,6 +1032,7 @@ pub const Viewer = struct {
                     t.tabstops.set(col_cell);
                 }
             }
+            pane.initialized = true;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/terminal/tmux/viewer.zig
```diff
@@ -255,6 +255,9 @@ pub const Viewer = struct {
 
     pub const Pane = struct {
         terminal: Terminal,
+        /// false until this pane's initial capture-pane + pane_state sync
+        /// has been applied. live %output is ignored untill then.
```
tristan957:
```suggestion
        /// has been applied. live %output is ignored until then.
```

tristan957 (changes requested):

PR #12117, opened 2026-04-05
terminal: make wuffs runtime-swappable, enable Kitty graphics for libvt

mitchellh:
This enables Kitty Graphics for `libghostty-vt` for the Zig API (C to come next).

First, a note on security: by default, Kitty graphics will only allow images transferred via the _direct_ medium (directly via the pty) and will not allow file or shared memory based images. libghostty-vt consumers need to manually opt-in via terminal init options or `terminal.setKittyGraphicsLoadingLimits` to enable file-based things. **This is so we're as secure as possible by default.**

Second, for PNG decoding, embedders must now set a global runtime-callback at `ghostty.sys.decode_png`. If this is not set, PNG formatted images are rejected. If this is set, then we'll use this to decode and embedders can use any decoder they want.

There is no C API exposed yet to set this, so this is only for Zig to start.

### Configuring Allowed Formats

```zig
var term = try Terminal.init(alloc, .{
    .cols = 80,
    .rows = 24,
    // Only allow direct (inline) image data, no file/shm access.
    // This is the default so you don't need to specify it.
    .kitty_image_loading_limits = .direct,
});
```

```zig
var term = try Terminal.init(alloc, .{
    .cols = 80,
    .rows = 24,
    // Allow all transmission mediums: direct, file, temporary file, shared memory.
    .kitty_image_loading_limits = .all,
});
```

```zig
var term = try Terminal.init(alloc, .{
    .cols = 80,
    .rows = 24,
    .kitty_image_loading_limits = .{
        .file = true,
        .temporary_file = true,
        .shared_memory = false,
    },
});
```

### Iterate all images

```zig
var it = term.screens.active.kitty_images.images.iterator();
while (it.next()) |kv| {
    const img = kv.value_ptr;
    std.debug.print("id={} {}x{} format={} bytes={}\n", .{
        img.id, img.width, img.height, img.format, img.data.len,
    });
}
```

### Delete all images

```zig
term.screens.active.kitty_images.delete(alloc, &term, .{ .all = true });
```

### DIFF
--- a/src/lib_vt.zig
+++ b/src/lib_vt.zig
@@ -19,6 +19,24 @@ const builtin = @import("builtin");
 // or are too Ghostty-internal.
 const terminal = @import("terminal/main.zig");
 
+/// System interface for the terminal package.
+///
+/// This module provides runtime-swappable function pointers for operations
+/// that depend on external implementations. Embedders can use this to
+/// provide or override default behaviors. These must be set at startup
+/// before any terminal functionality is used.
+///
+/// This lets libghostty-vt have no runtime dependencies on external
+/// libraries, while still allowing rich functionality that may require
+/// external libraries (e.g. image decoding or regular expresssions).
+///
+/// Setting these will enable various features of the terminal package.
+/// For example, setting a PNG decoder will enable support for PNG images in
+/// the Kitty Graphics Protocol.
+///
+/// Additional functionality will be added here over time as needed.
+pub const sys = terminal.sys;
+
 pub const apc = terminal.apc;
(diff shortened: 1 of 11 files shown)


mitchellh:
I'm going to convert this to a draft. I think we need to consider how to also disable file operations [at runtime] for Kitty graphics and not enable those by default since that's a pretty jarring and security-impacting side effect to get by default.

mitchellh:
Okay, I've added a new configuration to limit the mediums in use by Kitty graphics to make this safe by default. Updated PR desc to note it.

PR #12128, opened 2026-04-05
Fix title clipping when using `window-title-font-family`

nicholas-ochoa:
### Summary
This PR is a fix for the window title getting clipped/truncated when using certain fonts (in my case, Operator Mono Book) for the `window-title-font-family` setting.

There are a couple issues/discussions that are similar, but the issue still persists in current builds: https://github.com/ghostty-org/ghostty/issues/2349 https://github.com/ghostty-org/ghostty/discussions/9027

Here is the current behavior:
image

And here is a build with the changes:
image

### AI Disclosure
I used claude code (Sonnet 4.6) to identify the best place to start when implementing this change. All code in this PR was written by me.


bo2themax (changes requested):
This doesn't seem right (bottom window is on 1.3.1)
image

PR #12132, opened 2026-04-05
core: expose parent surface ID as GHOSTTY_SURFACE_PARENT_ID

davdroman:
## Summary

When a surface is created from a parent (via split or tab), this exposes the parent's surface ID as the `GHOSTTY_SURFACE_PARENT_ID` environment variable to child processes.

## Motivation

With `GHOSTTY_SURFACE_ID` (added in #12027), each surface has a unique identity. However, when splitting a pane, the new shell has no way to know which surface it was split from.

A concrete use case: a shell wrapper for [lazygit](https://github.com/jesseduffield/lazygit) that sets `LAZYGIT_NEW_DIR_FILE=/tmp/lazygit-newdir-$GHOSTTY_SURFACE_ID`. When lazygit navigates to a different repo, it writes the path to that file. A new split pane can then read `GHOSTTY_SURFACE_PARENT_ID`, look up `/tmp/lazygit-newdir-<parent-id>`, and `cd` there automatically -- landing in the same repo lazygit is browsing without lazygit needing to exit.

Without `GHOSTTY_SURFACE_PARENT_ID`, there's no reliable way for the new shell to find the correct file, especially when multiple Ghostty instances are running.

## Implementation

- **GTK apprt**: stores the parent's `id` in the surface's private data during `setParent`, then includes it in `defaultTermioEnv()`.
- **Embedded apprt** (macOS): adds `parent_surface_id` to `Options`, populates it in `newSurfaceOptions()`, stores it on the `Surface`, and includes it in `defaultTermioEnv()`.

The env var uses the same hex format as `GHOSTTY_SURFACE_ID` (`0x` + 16 hex digits). It is only set when a parent exists (not for the initial window).

### DIFF
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -414,6 +414,7 @@ pub const Surface = struct {
     size: apprt.SurfaceSize,
     cursor_pos: apprt.CursorPos,
     inspector: ?*Inspector = null,
+    parent_surface_id: u64 = 0,
 
     /// The current title of the surface. The embedded apprt saves this so
     /// that getTitle works without the implementer needing to save it.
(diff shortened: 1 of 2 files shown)


jcollie:
The parent surface that a surface was created from isn't a good way to determine what tab/window a surface belongs to as that's dynamically changeable as surfaces are able to be moved between tabs and windows. Indeed no static env var would do that. You'd need some sort of API to query that at runtime.

davdroman:
@jcollie I see. Thank you for answering and providing that piece of context.

PR #12143, opened 2026-04-06
macOS: force surface layout sync in updateOSView

fru1tworld:
`updateOSView` assumed SwiftUI always propagates frame changes to the scroll view. Under system load, this can be deferred, leaving the surface rendering at stale dimensions. Check for size mismatch and mark layout as needed.

ghostty_bug

---
AI Disclosure: Used Claude Code for PR preparation.


bo2themax:
Did you find any reproducible steps for this? The changes look good to me, but we still need to verify it ourselves that it actually fixes it.

mitchellh (approved):
Agree with @bo2themax The changes and reasoning look good to me. Would love either reproduction or more reference material on this. But I'm willing to merge and test in tip a bit too, pending CI.

PR #12146, opened 2026-04-06
terminal/tmux: resize existing panes on layout changes

h3nock:
## What

Resize existing panes when a `layout_change` updates pane geometry.

## Why

Today, `layout_change` updates the window layout tree and handles pane add/remove, but existing panes with the same pane ID are copied forward without reconciling their terminal `cols/rows` against the new layout width/height.

That can leave local pane terminal geometry stale after a split/resize operation.

This change keeps pane terminal geometry aligned with authoritative tmux layout geometry for reused pane IDs.

## Changes

- add a small `PendingPaneResize` list in the tmux viewer
- in `initLayout`, if a pane ID already exists:
  - compare current terminal `cols/rows` with layout `width/height`
  - queue a resize if the size changed
- in `syncLayouts`:
  - build the temporary pane map as before
  - commit windows/panes
  - clear the temporary `panes` variable after ownership moves to `self.panes` (`panes = .empty`)
  - apply queued resizes to the committed panes
- replace unchecked size casts with `std.math.cast(...)` and return
  `error.InvalidLayoutSize` when layout dimensions are out of range
- update `test "layout change"` to verify pane sizes after split:
  - existing pane `0` is `83x22`
  - new pane `2` is `83x21`

---

AI disclosure: Claude Code Opus 4.6 and Codex 5.4 xhigh were used for research, understanding and helping me write Zig.

---

### DIFF
--- a/src/terminal/tmux/viewer.zig
+++ b/src/terminal/tmux/viewer.zig
@@ -261,6 +261,12 @@ pub const Viewer = struct {
         }
     };
 
+    const PendingPaneResize = struct {
+        id: usize,
+        cols: size.CellCountInt,
+        rows: size.CellCountInt,
+    };
+
     /// Initialize a new viewer.
     ///
     /// The given allocator is used for all internal state. You must
@@ -638,10 +644,13 @@ pub const Viewer = struct {
             }
             panes.deinit(self.alloc);
         }
+        var pending_resizes: std.ArrayList(PendingPaneResize) = .empty;
+        defer pending_resizes.deinit(self.alloc);
         for (windows) |window| try initLayout(
             self.alloc,
             &self.panes,
             &panes,
+            &pending_resizes,
             window.layout,
         );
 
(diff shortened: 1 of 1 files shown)


src/terminal/tmux/viewer.zig
```diff
@@ -685,11 +694,6 @@ pub const Viewer = struct {
             if (added) try self.queueCommands(&.{.pane_state});
         }
 
-        // No more errors after this point. We're about to replace all
-        // our owned state with our temporary state, and our errdefers
-        // above will double-free if there is an error.
```
mitchellh, line 690: Was this caveat addressed? It doesn't _look like it_ to me. The error handling looks unsafe...

mitchellh (changes requested):

src/terminal/tmux/viewer.zig
h3nock: i think double free is fine, no? cause after `self.panes = panes`, i immediately set `panes = .empty` so the top-level `errdefer` cleanup in `syncLayouts` iterates an empty map

on the atomic commit point, i removed `errdefer comptime unreachable` because adding `terminal.resize` made the post-handoff path fallible. and currently if resize fails, the error propagates and `nextCommand` moves the viewer to `defunct`. not happy with that path, open to a better error handling here.

also tried moving the resize block before the `errdefer comptime unreachable` line but changed my mind for two reasons 1) reused panes are shallow-copied, so pre-handoff resize still mutates the same terminal objects and can still leave partial state on failure. no way to recover that in place. 2) we'd be mutating terminal state while old (`self.panes`) and new (`panes`) maps still coexist, instead of mutating through a single authoritative map after handoff. so i went with the current version to keep mutation on a single authoritative map 

happy to revisit if you'd prefer a different shape

Issue #5919, opened 2025-02-22
macOS: Add new config to show proxy icon only on hover


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5819

<div type='discussions-op-text'>

<sup>Originally posted by **edrozenberg** February 16, 2025</sup>
With recent MacOS releases, in apps such as Finder, Preview and others, the titlebar proxy icon is normally hidden until mouseover.

Currently Ghostty does the following via `macos-titlebar-proxy-icon`:
- `visible` - Always show the proxy icon, proxy icon functionality enabled
- `hidden` - Hide the proxy icon & disable proxy icon functionality
(https://ghostty.org/docs/config/reference#macos-titlebar-proxy-icon)

There is currently no way to match standard MacOS behavior because with `hidden` there is no proxy icon at all.

Suggestion:
Separate icon visibility setting from functionality enabled setting -
`macos-titlebar-proxy-icon = < visible | hidden | false/disabled >`
- `visible` - Always show the proxy icon (older MacOS behavior)
- `hidden` - Hide the proxy icon until mouseover (recent MacOS behavior)
- `false/disabled` - No proxy icon at all, disable proxy icon functionality

Mac Finder w/ proxy icon showing on mouseover:
https://github.com/user-attachments/assets/3d34d4a7-8a03-4138-8b24-e9d26871b5fa

Assumption:
With proxy icon `false/disabled`, proxy icon functionality is also disabled (dragging from the titlebar title, right-clicking titlebar title to see dir paths). This would match the behavior of other apps that don't use proxy icons (for ex. Apple Pages, Numbers, Keynote) which have no proxy icon and no proxy icon-like behavior in the titlebar.

Motivation:
- Match appearance/behavior of recent MacOS versions where most apps that use proxy icons have them hidden until mouseover
- Be able to use proxy icon functionality while having the icon hidden unless mouseover'd, for a cleaner look
- Also maintain the option to fully disable proxy icon functionality (icon & actions both disabled)</div>

## Proposal

1. Accept the above discussion proposal as given.
2. Make the default `hidden`. This breaks from `Terminal.app` but I agree with OP that this is the future macOS design appears to be heading and Terminal.app is simply... old.

bo2themax:
> 2\. Terminal.app is simply... old.

I think it's related to the toolbar. Like when opening a separate file with Xcode (on Tahoe), the proxy icon just stays there; The same goes with Finder if you hide its toolbar.

[...]

PR #12166, opened 2026-04-07
macos: add proxy icon hover mode and disabled state

MoonMao42:
Implements #5919.

Extends `macos-titlebar-proxy-icon` to three values:

- `visible` — always show the proxy icon (old default)
- `hidden` — show on hover via NSTrackingArea on the titlebar (new default)
- `disabled` — never show, also disables drag-drop and cmd-click path menu

The hover tracking uses `NSTrackingArea` with `.mouseEnteredAndExited` on `TerminalWindow.titlebarContainer`. A cached `currentPwdURL` is kept separate from `representedURL` so switching from `disabled` back to `visible`/`hidden` restores the icon without waiting for the next `cd`.

Config changes apply immediately to open windows through `GhosttyConfigDidChange`.

Here's what the hover mode looks like:
proxy-icon

## AI Usage

Used Claude Opus via Claude Code (GSD workflow) for initial config scaffolding. Swift/AppKit hover implementation written by hand. Codex used for review pass. I understand all the code.

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -80,6 +80,15 @@ class BaseTerminalController: NSWindowController,
     /// The configuration derived from the Ghostty config so we don't need to rely on references.
     private var derivedConfig: DerivedConfig
 
+    /// The current pwd for the focused surface. We cache this separately from representedURL
+    /// so config reload can restore the proxy icon after disabled mode clears the window state.
+    private var currentPwdURL: URL?
+
+    /// Hover state and tracking used for the "hidden" proxy icon mode.
+    private var isHoveringProxyIconTitlebar: Bool = false
+    private weak var proxyIconTrackingView: NSView?
+    private var proxyIconTrackingArea: NSTrackingArea?
+
     /// Track whether background is forced opaque (true) or using config transparency (false)
     var isBackgroundOpaque: Bool = false
 
@@ -224,6 +233,9 @@ class BaseTerminalController: NSWindowController,
     deinit {
         NotificationCenter.default.removeObserver(self)
         undoManager?.removeAllActions(withTarget: self)
+        if let trackingArea = proxyIconTrackingArea, let trackingView = proxyIconTrackingView {
+            trackingView.removeTrackingArea(trackingArea)
+        }
         if let eventMonitor {
             NSEvent.removeMonitor(eventMonitor)
         }
@@ -567,6 +579,9 @@ class BaseTerminalController: NSWindowController,
 
         // Update our derived config
         self.derivedConfig = DerivedConfig(config)
+
+        // Immediately refresh proxy icon behavior on all open windows.
+        refreshProxyIcon()
     }
 
     @objc private func ghosttyCommandPaletteDidToggle(_ notification: Notification) {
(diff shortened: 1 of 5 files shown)


src/config/Config.zig
```diff
@@ -3247,17 +3247,14 @@ keybind: Keybinds = .{},
...
+/// - `disabled` - never show the proxy icon (drag-drop and context menu also disabled)
 ///
 /// This setting can be changed at runtime and will affect all currently
-/// open windows but only after their working directory changes again.
-/// Therefore, to make this work after changing the setting, you must
-/// usually `cd` to a different directory, open a different file in an
-/// editor, etc.
-@"macos-titlebar-proxy-icon": MacTitlebarProxyIcon = .visible,
+/// open windows immediately.
+@"macos-titlebar-proxy-icon": MacTitlebarProxyIcon = .hidden,
```
jcollie, line 3257: We should not change the default to avoid surprising people that are used to (and perhaps prefer) the current behavior.

macos/Sources/Ghostty/Ghostty.Config.swift
```diff
@@ -369,7 +369,7 @@ extension Ghostty {
         }
 
         var macosTitlebarProxyIcon: MacOSTitlebarProxyIcon {
-            let defaultValue = MacOSTitlebarProxyIcon.visible
+            let defaultValue = MacOSTitlebarProxyIcon.hidden
```
jcollie, line 372: We should not change the default to avoid surprising people that are used to (and perhaps prefer) the current behavior.

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -1427,14 +1456,71 @@ class BaseTerminalController: NSWindowController,
...
+
     private struct DerivedConfig {
         let macosTitlebarProxyIcon: Ghostty.MacOSTitlebarProxyIcon
         let windowStepResize: Bool
         let focusFollowsMouse: Bool
         let splitPreserveZoom: Ghostty.Config.SplitPreserveZoom
 
         init() {
-            self.macosTitlebarProxyIcon = .visible
+            self.macosTitlebarProxyIcon = .hidden
```
jcollie, line 1523: We should not change the default to avoid surprising people that are used to (and perhaps prefer) the current behavior.

jcollie (changes requested):

MoonMao42:
Always-visible is actually legacy macOS behavior from before Monterey. Since macOS 12 the system default has been hover-to-show — Finder, Xcode, etc. all do this now. Terminal.app still shows it permanently but that's more of a "hasn't been updated" thing than a deliberate choice.

The mapping from old to new: `visible` = same as before, always show. `hidden` = hover to show (what macOS does by default since Monterey). `disabled` = what `hidden` used to be, completely off.

Defaulting to `hidden` just matches what the OS does. But if you'd rather not change the default I can keep it at `visible`, no strong feelings.

bo2themax:
> Always-visible is actually legacy macOS behavior from before Monterey. Since macOS 12 the system default has been hover-to-show — Finder, Xcode, etc. all do this now. Terminal.app still shows it permanently but that's more of a "hasn't been updated" thing than a deliberate choice.
> 
> The mapping from old to new: `visible` = same as before, always show. `hidden` = hover to show (what macOS does by default since Monterey). `disabled` = what `hidden` used to be, completely off.
> 
> Defaulting to `hidden` just matches what the OS does. But if you'd rather not change the default I can keep it at `visible`, no strong feelings.

This is true when there is a toolbar, as I commented in the discussion. 
So I would suggest adding an empty toolbar to `hidden` (hover to show), and removing it otherwise. This would match the system's default behaviour, without the hassle of adding a tracking area ourselves, but `titlebarTextField` and `titlebarContainer` may need some tweaks as well, depending on whether there is a toolbar or not.

```swift
let toolbar = NSToolbar(identifier: "TerminalToolbar")
self.toolbar = toolbar
toolbarStyle = .unifiedCompact
```

MoonMao42:
@jcollie What do you think？

jcollie:
> @jcollie What do you think？

I'm not really a macOS dev so I'll leave the review to others.

PR #12167, opened 2026-04-07
Add Win32 application runtime for Windows port

mattn:
This adds a minimal but functional Win32 application runtime (`apprt/win32`) that
enables Ghostty to build and run natively on Windows using MinGW (GNU ABI).

Built and tested on Windows 11 with Zig 0.15.2.

### Implemented

- [x] Win32 window creation and message loop
- [x] WGL OpenGL 4.6 rendering context with threaded drawing
- [x] ConPTY shell process (cmd.exe) via existing Pty implementation
- [x] Terminal emulation and text rendering (via CoreSurface)
- [x] Keyboard input: WM_KEYDOWN / WM_CHAR with A-Z, digits, punctuation, function keys, arrows, special keys
- [x] Key release events (WM_KEYUP / WM_SYSKEYUP)
- [x] Modifier keys: Ctrl, Alt, Shift, Super via GetKeyState
- [x] Dead keys and compose sequences (resolved by TranslateMessage into WM_CHAR)
- [x] Mouse input: click (left/right/middle), drag (SetCapture), scroll (WM_MOUSEWHEEL)
- [x] Mouse cursor position tracking (WM_MOUSEMOVE)
- [x] Mouse cursor shape changes (IDC_IBEAM, IDC_HAND, resize cursors, etc.)
- [x] Mouse cursor visibility toggle
- [x] Clipboard read/write via Win32 API (CF_UNICODETEXT)
- [x] Selection/primary clipboard aliased to standard (Windows has single clipboard)
- [x] OSC 52 clipboard integration (via existing clipboard API)
- [x] IME composition window positioning at terminal cursor
- [x] IME font size matching terminal font size
- [x] Window title update from terminal (set_title action)
- [x] Window icon from ID_ICON_GHOSTTY resource
- [x] Toggle maximize (ShowWindow SW_MAXIMIZE/SW_RESTORE)
- [x] Toggle fullscreen (monitor-aware borderless fullscreen)
- [x] Window resize with terminal grid reflow (WM_SIZE → sizeCallback)
- [x] GL viewport sync on resize (updateViewport)
- [x] SwapBuffers on renderer thread
- [x] Per-monitor DPI awareness (SetProcessDpiAwarenessContext)
- [x] DPI-based content scale (GetDpiForWindow)
- [x] Focus events (WM_SETFOCUS / WM_KILLFOCUS → focusCallback)
- [x] Windows light/dark theme detection (AppsUseLightTheme registry)
- [x] Color scheme update on WM_SETTINGCHANGE
- [x] URL open action (ShellExecuteW)
- [x] Bell / alert (MessageBeep)
- [x] Desktop notifications (Shell_NotifyIconW balloon with tray icon)
- [x] Quit confirmation dialog (MessageBoxW when process is active)
- [x] Child process exit notification dialog
- [x] Config hot reload (reload_config action)
- [x] Config file loading (%LOCALAPPDATA%\ghostty\config.ghostty)
- [x] font-family config via Windows system font directory scan (FreeType family_name matching)
- [x] CJK font fallback from C:\Windows\Fonts (msgothic, meiryo, msyh, malgun, etc.)
- [x] font-size config support
- [x] window-width / window-height config with actual font metrics (default 80x24)
- [x] Environment variable inheritance for child process
- [x] Skip termios timer on Windows (no POSIX termios)
- [x] DllMain for GNU ABI (MinGW build support)
- [x] Draw timer always active on Win32 to work around IOCP async latency
- [x] VSync disable on renderer thread for lower latency
- [x] Control character filtering in WM_CHAR to prevent double input
- [x] All remaining apprt actions handled explicitly (no-op for unsupported)

### Stubbed (action accepted but no-op)

Single-window, minimal implementation — these actions are acknowledged but not
acted on. They return success so callers don't treat them as errors:

- Multiple windows / tabs / splits (single-window design)
- `progress_report` — requires ITaskbarList3 COM
- `inspector` — debugging UI not implemented
- `mouse_over_link` — no tooltip/status bar UI
- `quit_timer` — single-window quits immediately
- `start_search` / `end_search` — search overlay UI
- `toggle_command_palette` — command palette UI
- `toggle_quick_terminal` — global hotkey/quick terminal
- `toggle_tab_overview`, `toggle_window_decorations`, `toggle_background_opacity`
- `open_config`, `pwd`, `float_window`, `secure_input`, `readonly`
- `key_sequence`, `key_table`, `color_change`
- `check_for_updates`, `show_on_screen_keyboard`, `command_finished`

### Known limitations

- Font discovery uses a Windows font directory scan + FreeType, not DirectWrite.
- IME preedit text is not rendered inline in the terminal yet.
- No shell integration scripts bundled.
- No accessibility / screen reader support.
- No custom shaders.

This Windows port was developed with the assistance of Claude Code (Anthropic's AI coding tool).

### DIFF
--- a/src/apprt.zig
+++ b/src/apprt.zig
@@ -16,6 +16,7 @@ pub const action = @import("apprt/action.zig");
 pub const ipc = @import("apprt/ipc.zig");
 pub const gtk = @import("apprt/gtk.zig");
 pub const none = @import("apprt/none.zig");
+pub const win32 = @import("apprt/win32.zig");
 pub const browser = @import("apprt/browser.zig");
 pub const embedded = @import("apprt/embedded.zig");
 pub const surface = @import("apprt/surface.zig");
@@ -43,6 +44,7 @@ pub const runtime = switch (build_config.artifact) {
     .exe => switch (build_config.app_runtime) {
         .none => none,
         .gtk => gtk,
+        .win32 => win32,
     },
     .lib => embedded,
     .wasm_module => browser,
--- a/src/apprt/runtime.zig
+++ b/src/apprt/runtime.zig
@@ -11,11 +11,17 @@ pub const Runtime = enum {
     /// approach to building the application.
     gtk,
 
+    /// Win32. Native Windows application using the Win32 API with OpenGL
+    /// rendering.
+    win32,
+
     pub fn default(target: std.Target) Runtime {
         return switch (target.os.tag) {
             // The Linux and FreeBSD default is GTK because it is a full
             // featured application.
             .linux, .freebsd => .gtk,
+            // Windows uses the native Win32 API.
+            .windows => .win32,
             // Otherwise, we do NONE so we don't create an exe and we create
             // libghostty. On macOS, Xcode is used to build the app that links
             // to libghostty.
--- a/src/apprt/structs.zig
+++ b/src/apprt/structs.zig
@@ -54,7 +54,7 @@ pub const Clipboard = enum(Backing) {
             .{ .name = "GhosttyApprtClipboard" },
         ),
 
-        .none => void,
+        .none, .win32 => void,
     };
 };
 
--- a/src/apprt/win32.zig
+++ b/src/apprt/win32.zig
@@ -1,0 +1,8 @@
+// The required comptime API for any apprt.
+pub const App = @import("win32/App.zig");
+pub const Surface = @import("win32/Surface.zig");
+pub const resourcesDir = @import("../os/main.zig").resourcesDir;
+
+test {
+    @import("std").testing.refAllDecls(@This());
+}
--- a/src/apprt/win32/App.zig
+++ b/src/apprt/win32/App.zig
@@ -122,6 +122,8 @@ extern "user32" fn SetWindowTextW(hWnd: HWND, lpString: [*:0]const u16) callconv
 extern "user32" fn GetDpiForWindow(hWnd: HWND) callconv(.winapi) UINT;
 extern "user32" fn SetProcessDpiAwarenessContext(value: isize) callconv(.winapi) BOOL;
 const DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2: isize = -4;
+extern "winmm" fn timeBeginPeriod(uPeriod: UINT) callconv(.winapi) UINT;
+extern "winmm" fn timeEndPeriod(uPeriod: UINT) callconv(.winapi) UINT;
 
 const COMPOSITIONFORM = extern struct {
     dwStyle: DWORD,
@@ -178,6 +180,9 @@ pub fn init(
     // Enable Per-Monitor DPI awareness
     _ = SetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
 
+    // Raise timer resolution to 1ms for responsive rendering
+    _ = timeBeginPeriod(1);
+
     // Create the main window
     try self.createWindow();
 
@@ -212,6 +217,7 @@ pub fn run(self: *App) !void {
 }
 
 pub fn terminate(self: *App) void {
+    _ = timeEndPeriod(1);
     self.surface.deinit();
     if (self.hwnd) |hwnd| {
         _ = DestroyWindow(hwnd);
--- a/src/apprt/win32/App.zig
+++ b/src/apprt/win32/App.zig
@@ -439,11 +439,10 @@ fn wndProc(hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM) callconv(.wina
             return 0;
         },
         WM_PAINT => {
+            // Validate the window to prevent continuous WM_PAINT messages.
+            // Actual rendering is done by the renderer thread via SwapBuffers.
             var ps: PAINTSTRUCT = std.mem.zeroes(PAINTSTRUCT);
             _ = BeginPaint(hwnd, &ps);
-            if (getApp(hwnd)) |app| {
-                app.surface.swapBuffers();
-            }
             _ = EndPaint(hwnd, &ps);
             return 0;
         },
--- a/src/apprt/win32/App.zig
+++ b/src/apprt/win32/App.zig
@@ -134,6 +134,24 @@ const COMPOSITIONFORM = extern struct {
 extern "imm32" fn ImmGetContext(hWnd: HWND) callconv(.winapi) ?*anyopaque;
 extern "imm32" fn ImmReleaseContext(hWnd: HWND, hIMC: ?*anyopaque) callconv(.winapi) BOOL;
 extern "imm32" fn ImmSetCompositionWindow(hIMC: ?*anyopaque, lpCompForm: *COMPOSITIONFORM) callconv(.winapi) BOOL;
+extern "imm32" fn ImmSetCompositionFontW(hIMC: ?*anyopaque, lplf: *LOGFONTW) callconv(.winapi) BOOL;
+
+const LOGFONTW = extern struct {
+    lfHeight: i32,
+    lfWidth: i32,
+    lfEscapement: i32 = 0,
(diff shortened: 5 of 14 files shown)


jcollie:
@mattn Large PRs like this are likely to _never_ be accepted due to the difficulty of reviewing. You're much better working with the community to submit smaller PRs that can be accepted. Also, no formal decision has been made regarding the toolkit that will be used to implement a Windows app-runtime. Looking like a modern Windows app is an important goal, not something that "just works". I'd suggest if you haven't already joining the Discord server and discussing Windows porting plans with others that are interested before writing any more code.

mattn:
@jcollie Thank your reply.

This pull request contains many commits, but in reality, there are no major changes to the existing code. If you would like to squash it, I can reduce the number of commits to just a few.

The code appears large because I am adding implementations in units similar to GTK, and it would be difficult to break it down further.

Furthermore, this implementation only implements basic terminal functionality, and it would be difficult to propose smaller units.

mitchellh:
I could see various points it could be broken down:

- DllMain change could be its own
- Font subsystem changes could be isolated, tests could be added (we run unit tests on Windows in CI)
- Standalone win32 apprt (unlikely to be accepted at this time without a maintainer)

The issue with changes like this are really the lack of a maintainer. PRs are a transfer of maintenance, and there isn't really currently anyone that can promise to maintain this for the next many years. That's a major blocker.

mattn:
I'm willing to take on the role of Windows maintainer. However, please understand that progress will depend on my available time. Also, the Windows port will remain incomplete compared to the GTK version for a while — not all features will be implemented right away. Is that acceptable?

Note: It would be difficult for me to monitor Discord frequently, so I'd prefer to communicate through GitHub (issues, discussions, and PRs).

mattn:
Although the pull request has been closed, the additional commits aren't visible here. Most of what I originally planned has been implemented, and Ghostty is now running on Windows.

https://github.com/user-attachments/assets/d8142a2a-9d85-44b6-9a06-f18cd1e587a7

Grueslayer:
Hi @mattn I've tested only two things which do not work and stopped further usage:

- All characters produced by AltGr (which can emulate by Ctrl-Alt on most keyboards) do not work for any combination I tested (I cannot write @,\,{,[,],} and so on on my German keyboard)

- Kitty graphics protocol does nothing like using `chafa -f kitty file.png`

Where to communicate about your Win32 things? In this PR?

mattn:
@Grueslayer Thanks for trying it and for the detailed report.

- AltGr / Ctrl-Alt combinations not producing characters is a bug in the current Win32 keyboard handling, so I need to fix that.
- `chafa -f kitty` not working is expected for now. The current Win32 path is ConPTY-based, and that is not enough for kitty graphics protocol passthrough.

mattn:
My understanding is:

- kitty graphics is not just ordinary text output;
- it relies on terminal-specific escape sequences being passed through unchanged;
- on Windows, ConPTY sits in the middle of that path;
- and ConPTY is not a transparent transport for the kitty graphics protocol.

Grueslayer:
I don’t think so, yes it’s specific sequences which need to be handled but not by the underlying terminal. There must be code in ghostty handling those and that maybe specific for each os target.
Wezterm also uses ConPTY on Windows and handles iTerm and Kitty protocol (both sadly with bugs and for kitty only parts of the specification)

mattn:
From what I've tested, the Kitty Image protocol doesn't seem to work with wezterm on Windows.

image

mattn:
@Grueslayer FYI, WezTerm on Windows does not support kitty image protocol.

https://github.com/wezterm/wezterm/issues/1673

https://github.com/microsoft/terminal/issues/1173

mattn:
I made a change for AltGr handling. Could you try again on your German keyboard and let me know whether combinations like `AltGr+Q` (`@`) and `AltGr+7/8/9/0` (`{ [ ] }`) work now?

If not, please tell me the exact key combinations that still fail.

Grueslayer:
> @Grueslayer FYI, WezTerm on Windows does not support kitty image protocol.
> 
> [wezterm/wezterm#1673](https://github.com/wezterm/wezterm/issues/1673)
> 
> [microsoft/terminal#1173](https://github.com/microsoft/terminal/issues/1173)

That is not true. Wezterm didn't used the ConPTY delivered by Windows as is old and has bugs but a self compiled version from [Microsoft Terminal](https://github.com/microsoft/terminal) project which didn't passed some sequences from the user app to the terminal, so Wezterm could not interpret this sequences to draw the image. As your link also said, this has been fixed in the conpty for a long time. My Wezterm is a year old and displays it with no problem:

grafik

Grueslayer:
> I made a change for AltGr handling. Could you try again on your German keyboard and let me know whether combinations like `AltGr+Q` (`@`) and `AltGr+7/8/9/0` (`{ [ ] }`) work now?
> 
> If not, please tell me the exact key combinations that still fail.

I used your [ghostty-windows-0.15.2-6.zip](https://github.com/mattn/ghostty/releases/download/v1.3.1/ghostty-windows-0.15.2-6.zip) file but still no combination does anything, i tested (on a German keyboard)

AltGr-2 = ²
AltGr-3 = ³
AltGr-7 = {
AltGr-8 = [
AltGr-9 = ]
AltGr-0 = }
AltGr-q = @
AltGr-ß = \ (that's the key right from 0)

mattn:
@Grueslayer 

If that is really the difference, then in principle Ghostty should also benefit from it when running on top of your ConPTY build.

The problem for me is that I cannot depend on a non-official / unreleased ConPTY to build or distribute Ghostty on Windows, so I cannot treat that as a requirement on my side.

If you already have such an environment, could you try Ghostty with your ConPTY build and let me know whether kitty images work there?

Grueslayer:
> @Grueslayer
> 
> If that is really the difference, then in principle Ghostty should also benefit from it when running on top of your ConPTY build.
> 
> The problem for me is that I cannot depend on a non-official / unreleased ConPTY to build or distribute Ghostty on Windows, so I cannot treat that as a requirement on my side.
> 
> If you already have such an environment, could you try Ghostty with your ConPTY build and let me know whether kitty images work there?

Ghostty.exe already is in the same directory as OpenConsole.exe and ConPTY.dll and should use these as same dir has prio before path.

mattn:
I figured out how `conpty.dll` works. It's a DLL that exports the same symbols as `CreatePseudoConsole` / `ClosePseudoConsole` in `kernel32.dll`, and when you call them it invokes `OpenConsole.exe` from the same directory. In `ghostty-windows-0.15.2-8.zip`, `conpty.dll` is loaded first, with `kernel32.dll` as the fallback. Please give it a try.

image (2)

Grueslayer:
Thank you, that’s working fine here, too. 

Now AltGr needs a fix, without I cannot work.

While my old vim 9.1.113 works, Neovim always writes the character twice a becomes aa but : becomes .: (: is Shift-. on German keyboard).
And now: AltGr-2 is 2 ², it puts the normal char and then the shifted/altered behind.

mattn:
I can't reproduce this locally since my keyboard layout is different, but I tried a different approach: the handling now branches on whether the key produces a character or not.

The separation is: "let Win32 text input handle presses for character-producing keys" and "only dispatch `keydown` for real shortcuts and special keys."

pluiedev:
FYI, I would recommend moving this to another thread since a) we are pretty certain we won't accept this PR as-is, and b) we have a webhook set up in the Discord server to monitor PR comments and stuff and this has constantly been popping up there. Thanks :)

mattn:
To be honest, Ghostty isn't that important to me. I'd never used Ghostty before starting this build. I saw a lot of people on X/Twitter asking for a Windows version of Ghostty, so I started developing one for Windows.
I have a little knowledge of Windows and the terminal, so maybe I was better suited to it than others.
But if this pull request isn't approved, I'll stop developing the Windows version. If it's approved after some steps (like splitting it up), I'll continue maintenance.

PR #12175, opened 2026-04-08
Exploration: Reconnectable Terminal using libghostty

bmaurer:
### The problem

Remote terminal sessions are fragile. A network hiccup, laptop sleep, or IP change kills the SSH connection and all terminal state is lost. The existing solutions require awkward layering:

- **tmux/screen**: Run a multiplexer on the remote host, then SSH into it. But now you have a terminal emulator (tmux) inside a terminal emulator (Ghostty), with imperfect VT compatibility, its own keybindings conflicting with yours, and visual artifacts. tmux control mode helps but requires [complex integration](https://github.com/MisterTea/EternalTerminal/issues/631) and still doesn't solve the fundamental problem.

- **Mosh**: Synchronizes the visible screen over UDP, but doesn't support scrollback, has limited VT compatibility, and can't truly reconnect after extended disconnection. Its terminal emulator drifts on features it doesn't understand (OSC sequences, kitty protocol, etc.).

- **Eternal Terminal**: Reconnectable SSH, but still relies on TCP streams and doesn't handle output spew gracefully. Users end up running ET + tmux control mode to get both reconnection and session persistence — two complex systems solving one problem.

The core tension: when a program spews output faster than the connection can handle, you need to either throttle the application (breaking Ctrl-C responsiveness and stalling pipelines) or buffer unboundedly toward the client (causing minutes of lag). SSH does the latter. Mosh solved this by running a VT emulator on the server and synchronizing *state* instead of *streams* — but its emulator is limited and it abandoned scrollback entirely.

### The insight

A terminal emulator already manages exactly the state we need: the visible viewport, scrollback buffer, cursor position, colors, modes, and all the other VT state. Ghostty's `libghostty` separates the VT emulator from the renderer. The emulator processes escape sequences and maintains screen state; the renderer reads that state and draws pixels.

This is exactly the split needed for a remote terminal:
- **Server**: Runs `libghostty`'s Terminal + a pty, maintaining authoritative state
- **Client**: Connects, receives state, renders it

The terminal IS the session manager. The only thing missing is having the process manager (pty + shell) live on a remote host instead of locally. It makes most sense to use the same library for both sides.

### What this PR contains

A working prototype (`ghostty-snap`) that demonstrates the concept end-to-end.

#### Components

- **`src/main_snap.zig`**: Server + attach client + capture/restore tools (~900 lines)
- **`src/terminal/sync.zig`**: Binary state sync API with compact encoding (~650 lines)
- **`src/terminal/formatter.zig`**: Fix for cursor save/restore around tabstop setup + comprehensive round-trip test
- **`tools/rate-limit-proxy.py`**: Rate-limiting TCP proxy for testing bandwidth-constrained scenarios
- **`tools/debug-proxy.py`**: Full byte-level logging proxy for diagnosing frame corruption
- **`tools/test-ghostty-snap.py`**: Automated test suite (29/30 passing)
- **`docs/design/reconnectable-terminal.md`**: Detailed design proposal

#### Architecture

The server has two sending modes:

1. **Raw streaming**: Pty output goes directly to the client as framed PTY_DATA messages. Same latency as SSH — characters echo immediately, cursor moves instantly.

2. **Snapshot mode**: When the client can't keep up (detected via `TCP_NOTSENT_LOWAT`), the server stops sending raw bytes, feeds everything through the terminal emulator, and sends a VT snapshot of the *current* state when the pipe clears. The client jumps from stale to current instantly. Intermediate states are skipped.

This means:
- **Ctrl-C always works within one RTT** — input is never blocked by queued output
- **No application throttling** — the pty runs at full speed regardless of connection quality
- **Network buffers stay bounded** — no unbounded buffering toward the client

#### Protocol

Length-prefixed framed messages (`[type:u8][length:u24][payload]`):

| Type | Name | Direction | Purpose |
|------|------|-----------|---------|
| 0x01 | PTY_DATA | S→C | Raw pty output (fast path) |
| 0x02 | SNAPSHOT | S→C | VT state snapshot (catch-up) |
| 0x03 | SCROLLBACK | S→C | VT scrollback text |
| 0x05 | VIEWPORT_DELTA | S→C | Binary dirty rows only |
| 0x06 | VIEWPORT_FULL | S→C | Binary full viewport |
| 0x07 | SCROLLBACK_CHUNK | S→C | Binary scrollback chunk |
| 0x81 | INPUT | C→S | Keystrokes → pty |
| 0x82 | RESIZE | C→S | Terminal size change |

#### Sync API (`sync.zig`)

A structured data API for accessing terminal state incrementally, designed for eventual use inside Ghostty as a native remote client:

- **Style dictionary**: Session-level style deduplication (most sessions use <20 unique styles), referenced by varint IDs
- **Row encoding**: Style-tagged UTF-8 runs — ASCII is 1 byte/char, blank rows are 1 byte
- **Row hashing**: Wyhash on raw cell bytes for O(1) dirty detection
- **SyncState**: Epoch tracking, delta computation (full/partial/none), incremental scrollback chunking

### How to demo

```bash
# Build
zig build -Demit-bench=true

# Start a named session
ghostty-snap server --name myproject -- bash

# Attach from another terminal
ghostty-snap attach myproject

# Everything works: type commands, run vi, resize your window
# Ctrl-] to detach — server keeps running

# Reattach and see your state preserved
ghostty-snap attach myproject

# List sessions
ghostty-snap list

# Test with bandwidth constraints (simulates slow link)
python3 tools/rate-limit-proxy.py --rate 20480 --listen 7682
ghostty-snap attach 127.0.0.1:7682
# Run: cat /dev/urandom | base64
# Hit Ctrl-C — responds instantly despite the rate limit

# Capture a terminal snapshot to a file
ghostty-snap capture -- bash
# Ctrl-\ to snapshot
ghostty-snap restore /tmp/ghostty-snap.vt
```

### Future roadmap

If this concept proves interesting, the path to production would include:

- **Native Ghostty client**: Instead of the VT passthrough attach client, a proper Ghostty apprt that applies binary sync data directly to Page structures. This would give perfect fidelity and enable proper scrollback backfill without VT hacks.
- **WebSocket / WebRTC transport**: WebSocket support would allow browser-based clients (xterm.js) to connect to a ghostty-snap server and sync full terminal state — turning any browser into a reconnectable terminal. WebRTC's datagram channels could provide the reliability benefits of Mosh's UDP approach (skip intermediate states, no head-of-line blocking) while still syncing full VT state including scrollback. This would also enable connection migration for IP roaming without the complexity of QUIC.
- **Speculative local echo**: Like Mosh, predict the effect of keystrokes and display immediately, verify against server state.
- **Multi-client sessions**: One writer + N readers for pair programming or monitoring.
- **Session persistence**: Survive server restarts by serializing terminal state to disk.

### Caveats

This code is fully AI-generated — built with Claude Code in a single extended session. I have not dug into the implementation details of every line, but it is working and demonstrable. The concepts and architecture are real; the code quality is prototype-grade.

I may not have the time to productionize this, but wanted to share the exploration and the concept. The core insight — that a terminal emulator already manages exactly the state needed for remote session persistence, and `libghostty` already separates emulation from rendering — feels worth exploring further.

The `sync.zig` API in particular is designed to be usable by a native Ghostty client, independent of the prototype server/client code.

### DIFF
--- a/docs/design/reconnectable-terminal.md
+++ b/docs/design/reconnectable-terminal.md
@@ -1,0 +1,903 @@
+# Reconnectable Terminal: Design Proposal
+
+## Problem Statement
+
+SSH terminal sessions are fragile. A network hiccup, laptop sleep, or IP change
+kills the connection and all terminal state is lost. Tools like tmux and screen
+provide reconnection by interposing a multiplexer, but they have their own
+terminal emulators with imperfect compatibility, add visual artifacts, and
+consume resources even when nobody is watching.
+
+Mosh improved the situation by running a VT emulator on both sides and
+synchronizing screen state, but it only syncs the visible viewport (no
+scrollback), doesn't support true reconnection after extended disconnection,
+and has limited VT compatibility.
+
+This proposal describes how Ghostty's VT emulator library (`libghostty`) could
+serve as the foundation for a reconnectable remote terminal that:
+
+1. Runs the authoritative VT emulator on the remote host
+2. Allows a Ghostty client to connect and fully sync terminal state at any time
+3. Prioritizes visible viewport during bandwidth-constrained sync
+4. Backfills scrollback history when bandwidth permits
+5. Streams incremental updates in real-time once caught up
+6. Handles high-throughput output without filling network buffers
+
+## Architecture Overview
+
+```
+┌──────────────────────────────────────────────────────────────────────┐
+│ Remote Host                                                         │
+│                                                                     │
+│  ┌──────────┐     ┌───────────────────────────────────────────────┐  │
+│  │  Shell    │────▶│  ghostty-server (libghostty)                 │  │
+│  │  (bash,   │ pty │                                              │  │
+│  │   zsh,    │◀────│  ┌────────────┐  ┌────────────────────────┐  │  │
+│  │   etc.)   │     │  │ Terminal   │  │ Sync Engine            │  │  │
+│  └──────────┘     │  │ (VT emu)   │  │                        │  │  │
+│                    │  │            │  │ - State snapshotter    │  │  │
+│                    │  │ ┌────────┐ │  │ - Delta tracker        │  │  │
+│                    │  │ │Primary │ │  │ - Priority scheduler   │  │  │
+│                    │  │ │Screen  │ │  │ - Bandwidth estimator  │  │  │
+│                    │  │ │+scroll │ │  │                        │  │  │
+│                    │  │ ├────────┤ │  └────────────────────────┘  │  │
+│                    │  │ │Alt     │ │              │               │  │
+│                    │  │ │Screen  │ │              │               │  │
+│                    │  │ └────────┘ │              │               │  │
+│                    │  └────────────┘              │               │  │
+│                    └─────────────────────────────┬┘               │  │
+│                                                  │                │  │
+└──────────────────────────────────────────────────┼────────────────┘  │
+                                                   │ network          │
+                                          ┌────────┴────────┐         │
+                                          │ Sync Protocol   │         │
+                                          │ (UDP + TCP)     │         │
+                                          └────────┬────────┘         │
+                                                   │                  │
+┌──────────────────────────────────────────────────┼────────────────┐  │
+│ Client (Ghostty)                                 │                │  │
+│                                                  │                │  │
+│  ┌───────────────────────────────────────────────┴──────────────┐ │  │
+│  │ Sync Client                                                  │ │  │
+│  │                                                              │ │  │
+│  │ - State receiver / applier                                   │ │  │
+│  │ - Scrollback cache                                           │ │  │
+│  │ - Input forwarding                                           │ │  │
+│  └──────────────────────────────────┬───────────────────────────┘ │  │
+│                                     │                             │  │
+│  ┌──────────────────────────────────▼───────────────────────────┐ │  │
+│  │ Ghostty Renderer                                             │ │  │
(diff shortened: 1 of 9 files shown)


jcollie:
The major problem with having the terminal emulator be the session manager is that it doesn't work for sessions on remote systems (where you want to protect from network disconnections) without installing the entire Ghostty application, including all of it's _many_ dependencies. I personally think that session persistence is better left to lightweight standalone tools (see [zmx](https://github.com/neurosnap/zmx)) that would be much easier to install on servers and other systems that you want to keep streamlined.

bmaurer:
> The major problem with having the terminal emulator be the session manager is that it doesn't work for sessions on remote systems (where you want to protect from network disconnections) without installing the entire Ghostty application, including all of it's _many_ dependencies. I personally think that session persistence is better left to lightweight standalone tools (see [zmx](https://github.com/neurosnap/zmx)) that would be much easier to install on servers and other systems that you want to keep streamlined.

In this case you'd only need libghostty, not the entire application. https://mitchellh.com/writing/libghostty-is-coming describes libghostty as attempting to solve exactly this challenge — a minimal dependency library.

Session persistence is not a new concept. The contribution of this PR is:

1) Doing so using a library that can enable remote host to local host communication. Previous approaches generally rely on replaying terminal command codes to sync the client (eg opening a new screen window that restores previous state). OTOH this approach allows synchronization of the virtual terminal emulation from the remote host to the client with rich semantics like scrollback sync.

2) handling bandwidth limited scenarios (say, cat'ing a huge log file). Rather than lagging or slowing down the remote end, this approach turns the client into a streaming viewer, like a video that turns down to a lower resolution.

PR #12182, opened 2026-04-08
libghostty(formatter): enable kitty gfx for vt

neurosnap:
This adds support for printing kitty graphics using the `terminal.Formatter` through a new `terminal.ScreenFormatter.Extra` option `kitty_graphics`.

This only implements kitty gfx for the `vt` format option.

This enables kitty gfx support for `zmx` when we restore terminal scrollback on re-attach: https://github.com/neurosnap/zmx/issues/14

References: https://discord.com/channels/1005603569187160125/1420009803173859449/1446588699952287849

The tricky part was getting the spacing correct when mixing images and content which I made sure to lock down inside a test.

To get the spacing correct I needed to thread the `Terminal` object through the formatters so I could run `placement.gridSize()`.

I wrote the bulk of the implementation and QA'd the results.  I manually tested this feature
using a modified `example/zig-formatter` which can be found in this branch: https://github.com/neurosnap/ghostty/commit/2b19fafe216a96706a3f0e8e1c7cb561c1aad654

```bash
./ex.sh | zig build run
```
Screenshot 2026-04-08 at 11 41 42 AM

# ai disclosure

I used AI to help me:
- find the critical areas to add this feature
- write tests that confirmed desired behavior
- debug issues that came up with my impl

### DIFF
--- a/src/terminal/formatter.zig
+++ b/src/terminal/formatter.zig
@@ -18,6 +18,118 @@ const Pin = PageList.Pin;
 const Row = @import("page.zig").Row;
 const Selection = @import("Selection.zig");
 const Style = @import("style.zig").Style;
+const build_options = @import("terminal_options");
+const encoder = std.base64.standard.Encoder;
+const ImageStorage = if (build_options.kitty_graphics) kitty.graphics.ImageStorage else void;
+
+const KittyGfxState = struct {
+    alloc: Allocator,
+    terminal: *const Terminal,
+    images: if (build_options.kitty_graphics) *const ImageStorage else void,
+    // Track which images have already been loaded
+    seen_images: std.AutoArrayHashMapUnmanaged(u32, void),
+    // Track how many rows we should skip since we already created
+    // newlines for them
+    skip_rows: u32 = 0,
+
+    fn init(alloc: Allocator, terminal: *const Terminal, images: if (build_options.kitty_graphics) *const ImageStorage else void) KittyGfxState {
+        return .{
+            .images = images,
+            .seen_images = .{},
+            .alloc = alloc,
+            .terminal = terminal,
+        };
+    }
+
+    fn deinit(self: *KittyGfxState) void {
+        self.seen_images.deinit(self.alloc);
+    }
+
+    fn format(self: *KittyGfxState, writer: *std.Io.Writer, page: *const Page, y: size.CellCountInt, blank_rows: *usize) std.Io.Writer.Error!void {
+        if (comptime !build_options.kitty_graphics) return;
+        var p_iter = self.images.placements.iterator();
+        while (p_iter.next()) |p_entry| {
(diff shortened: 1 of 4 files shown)


src/terminal/formatter.zig
```diff
@@ -18,6 +18,117 @@ const Pin = PageList.Pin;
...
+                    const chunk_size = 4096;
+                    var start: usize = 0;
+                    while (start < encoded_buf.len) {
+                        const end = @min(start + chunk_size, encoded_buf.len);
+                        const is_last = end == encoded_buf.len;
+                        const more_flag: u32 = if (is_last) 0 else 1;
+
+                        const seq = std.fmt.allocPrint(
+                            self.alloc,
+                            "\x1b_Ga=t,q=1,f={d},s={d},v={d},i={d},m={d};{s}\x1b\\",
```
neurosnap, line 81: do we want `q=2` here to suppress all confirmation/error output?

src/terminal/formatter.zig
```diff
@@ -18,6 +18,117 @@ const Pin = PageList.Pin;
...
+                for (0..blank_rows.*) |_| try writer.writeAll("\r\n");
+                blank_rows.* = 0;
+            }
+
+            // Move cursor to the placement's column, emit placement,
+            // then return cursor to column 1.
+            try writer.print("\x1b[{d}G", .{pin.x + 1});
+            const p_seq = std.fmt.allocPrint(
+                self.alloc,
+                "\x1b_Ga=p,q=1,C=1,i={d},c={d},r={d},X={d},Y={d}\x1b\\",
```
neurosnap: do we want `q=2` here to suppress all confirmation/error output?

src/terminal/formatter.zig
```diff
@@ -18,6 +18,117 @@ const Pin = PageList.Pin;
...
+                .{ key.image_id, placement.columns, placement.rows, placement.x_offset, placement.y_offset },
+            ) catch return error.WriteFailed;
+            defer self.alloc.free(p_seq);
+            try writer.writeAll(p_seq);
+            try writer.print("\x1b[1G", .{});
+
+            if (self.images.imageById(key.image_id)) |image| {
+                const grid = placement.gridSize(image, self.terminal);
+                // Create newlines for the entire image height
```
mitchellh: Do we have to do this? I thought the Kitty protocol handlers do this automatically. The VT formatter should try to just emit the VT sequences that result in the same behaviors, no?

mitchellh (changes requested):

src/terminal/formatter.zig
neurosnap: Ah!  Good catch.  I was experimenting with different implementations and had left the `C=1` in the placement escape code which disables automatic cursor advancement. Updated.

PR #12207, opened 2026-04-10
fix(gtk): RHEL 9 compat for TitleDialog + install script

rudolfjs:
## Summary

- **TitleDialog cast fix**: `TitleDialog.present()` cast `self` to `Dialog.Parent` (which resolves to `adw.MessageDialog` on libadwaita < 1.5), but `TitleDialog` extends `adw.AlertDialog` — an invalid cross-hierarchy cast causing compile error. Fixed by casting to its own `Parent` type.
- **Comptime guards**: Guard `promptTitle()` and `promptTabTitle()` behind `adw_version.supportsDialogs()` so the entire `TitleDialog` codepath (requiring `AdwAlertDialog` from libadwaita 1.5+) is compiled out on older adwaita versions.
- **RHEL 9 install script** (`scripts/install-ghostty-rhel9.sh`): Build-from-source installer with:
  - Proxy workarounds: sources `/etc/environment` when `sudo` strips env vars, uses shared `ZIG_GLOBAL_CACHE_DIR`, handles `git+https://` URLs from any forge (GitHub, Codeberg, etc.)
  - Upfront library version checks (GLib >= 2.72, GTK4 >= 4.14, libadwaita >= 1.5) that fail fast with guidance to use `sudo snap install ghostty --classic` when system libraries are too old

## Context

RHEL 9 ships GLib 2.68, GTK4 4.12, libadwaita 1.4 — all too old for Ghostty 1.3.1's full build. The Zig source patches fix the compile error for the TitleDialog feature; remaining linker errors from `g_signal_group_*` / `adw_dialog_*` etc. are detected by the script upfront.

### DIFF
--- a/scripts/install-ghostty-rhel9.sh
+++ b/scripts/install-ghostty-rhel9.sh
@@ -1,0 +1,418 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# =============================================================================
+# Ghostty installer for RHEL 9.x
+#
+# Preferred method: snap install ghostty --classic
+#   (bundles its own GNOME libs, avoids RHEL 9 library version issues)
+#
+# Fallback:  sudo bash install-ghostty.sh   (build from source)
+#   Requires GLib >= 2.72, GTK4 >= 4.14, libadwaita >= 1.5
+#   RHEL 9 ships older versions — build-from-source will fail.
+# =============================================================================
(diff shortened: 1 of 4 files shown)


rudolfjs:
Opened against wrong repo by mistake, sorry!

PR #12212, opened 2026-04-10
cmake: add ghostty_vt_add_target() for cross-compilation

mitchellh:
Add a ghostty_vt_add_target() CMake function that lets downstream projects build libghostty-vt for a specific Zig target triple. The function encapsulates zig discovery, build-type-to-optimize mapping, the zig build invocation, and output path conventions so consumers do not need to duplicate any of that logic. It creates named IMPORTED targets (e.g. ghostty-vt-static-linux-amd64) that work alongside the existing native ghostty-vt and ghostty-vt-static targets.

The build-type mapping is factored into a shared _GHOSTTY_ZIG_OPT_FLAG variable used by both the native build and the new function.

A new example/c-vt-cmake-cross/ demonstrates end-to-end cross- compilation using zig cc as the C compiler, auto-detecting a cross target based on the host OS.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -278,6 +278,11 @@ jobs:
       fail-fast: false
       matrix:
         dir: ${{ fromJSON(needs.list-examples.outputs.cmake) }}
+        exclude:
+          # Cross-compilation with zig cc requires a single-config
+          # generator (Makefiles/Ninja). The Windows CI uses Visual
+          # Studio which always uses MSVC and ignores CMAKE_C_COMPILER.
+          - dir: c-vt-cmake-cross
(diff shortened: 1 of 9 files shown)


mitchellh:
Replaced the /bin/sh-based archive combining with a cross-platform zig build tool. 

The previous approach used a shell script to generate an MRI script and pipe it to ar -M, which required /bin/sh — unavailable on Windows. The new tool does the same thing (generates the MRI script, pipes it to zig ar -M) but as a native zig executable that works on all platforms. This removes the Windows exclusion from fat archive creation, so ghostty-vt-static.lib on Windows now bundles the SIMD dependencies (highway, simdutf, utfcpp) just like libghostty-vt.a does on Linux and macOS.

PR #12214, opened 2026-04-10
build: add ghostty-internal pkg-config modules (shared + static)

deblasis:
## Summary
Mirror the `libghostty-vt-static` pkg-config pattern from #12210 for the internal library.

- Add `ghostty-internal.pc` (shared, `-lghostty`) and `ghostty-internal-static.pc` (static, direct archive reference) so consumers can discover either variant via pkg-config
- Named `ghostty-internal` to distinguish from the public `libghostty-vt` API
- Static module points at the platform-correct archive name (`ghostty-static.lib` on Windows, `libghostty.a` elsewhere)
- pkg-config files are generated during shared builds and installed via `GhosttyLib.install()`

## Test plan

- [x] `zig build` succeeds (default target)
- [x] `ghostty-internal.pc` and `ghostty-internal-static.pc` appear in `zig-out/share/pkgconfig/`
- [x] Static `.pc` points at `ghostty-static.lib` (Windows) / `libghostty.a` (Unix)
- [x] Shared `.pc` uses standard `-L -l` flags
- [x] Existing `libghostty-vt` pkg-config files are unaffected

### DIFF
--- a/src/build/GhosttyLib.zig
+++ b/src/build/GhosttyLib.zig
@@ -13,6 +13,8 @@ step: *std.Build.Step,
 /// The final static library file
 output: std.Build.LazyPath,
 dsym: ?std.Build.LazyPath,
+pkg_config: ?std.Build.LazyPath,
+pkg_config_static: ?std.Build.LazyPath,
 
 pub fn initStatic(
     b: *std.Build,
@@ -54,6 +56,8 @@ pub fn initStatic(
         .step = &lib.step,
         .output = lib.getEmittedBin(),
         .dsym = null,
+        .pkg_config = null,
+        .pkg_config_static = null,
     };
 
     // Create a static lib that contains all our dependencies.
@@ -70,6 +74,8 @@ pub fn initStatic(
 
         // Static libraries cannot have dSYMs because they aren't linked.
         .dsym = null,
+        .pkg_config = null,
+        .pkg_config_static = null,
     };
 }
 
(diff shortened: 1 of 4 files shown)


deblasis:
If this one lands, next one would be to create a "fat archive" for windows as well.
Like libtool does on darwin.

I will 90% hit a wall with NativeAOT on windows with winui3 but regardless, that's something MS is supposedly working on. 
I might chime in to give a hand with repro/bugs etc.

## Update
I hit the wall eventually but found workarounds and managed to get a single executable for Ghostty Windows! 
32MB single exe with libghostty (and the other deps) baked in. No ghostty.dll needed.

I filed an issue upstream for anyone else bumping into the same problem: https://github.com/microsoft/WindowsAppSDK/issues/6394

mitchellh:
Lets rename these to `ghostty-internal` and `ghostty-internal-static` because this version of "libghostty" is

1. Not meant for external use
2. Not `libghostty` and confuses people for it.

mitchellh (changes requested):
Blocked on some naming.

src/build/GhosttyLib.zig
```diff
@@ -197,3 +231,54 @@ pub fn installHeader(self: *const GhosttyLib) void {
...
+            \\Requires.private:
+        , .{ b.install_prefix, deps.config.version, staticLibraryName(os_tag) })),
+    };
+}
+
+fn staticLibraryName(os_tag: std.Target.Os.Tag) []const u8 {
+    return if (os_tag == .windows)
+        "ghostty-static.lib"
+    else
+        "libghostty.a";
```
mitchellh: Let's rename this too. And anywhere else we write it out, let's use this PR to fix it up.

mitchellh (changes requested):

src/build/GhosttyLib.zig
deblasis: I hope I didn't miss/mess up anything :)

PR #12217, opened 2026-04-10
build: fat static archive and ubsan fix for external linkers

deblasis:
## Summary

> [!IMPORTANT]
> Stacked on #12214. Review that first. (i am targeting `main` so here you will see the full changeset, including 12214

Two changes that make the static libghostty archive consumable by external linkers (MSVC link.exe, .NET NativeAOT, Rust, Go, etc.):

**Fat static archive on all platforms**

The static archive previously only bundled vendored deps on macOS (via libtool). On Windows and Linux the archive contained only the Zig-compiled code, requiring consumers to find and link freetype, harfbuzz, glslang, spirv-cross, simdutf, oniguruma, etc. separately.

Now all platforms produce a single fat archive:
- macOS: libtool (unchanged)
- Windows: zig ar qcL --format=coff (MSVC's lib.exe can't read Zig-produced GNU-format archives, so we use the bundled LLVM archiver)
- Linux: ar -M with MRI scripts (same approach as libghostty-vt)

**MSVC ubsan suppression for C deps**

Zig's ubsan runtime can't be bundled on Windows (LNK4229), leaving __ubsan_handle_* symbols unresolved. freetype, glslang, spirv-cross, and highway already suppress ubsan. This adds MSVC-conditional suppression to seven more: harfbuzz, libpng, dcimgui, wuffs, oniguruma, zlib, and stb.

Gated on abi == .msvc so ubsan coverage is preserved on Linux/macOS.

## Test plan

- [x] zig build produces a fat ghostty-static.lib (~230MB) with ~200 object files
- [x] MSVC's lib /LIST can read the archive
- [x] .NET NativeAOT consumer resolves all symbols (0 unresolved)
- [x] Linux/macOS builds unaffected (ubsan remains enabled)

### DIFF
--- a/src/build/GhosttyLib.zig
+++ b/src/build/GhosttyLib.zig
@@ -48,29 +48,18 @@ pub fn initStatic(
     }
 
     // Add our dependencies. Get the list of all static deps so we can
-    // build a combined archive if necessary.
+    // build a combined archive.
     var lib_list = try deps.add(lib);
     try lib_list.append(b.allocator, lib.getEmittedBin());
 
-    if (!deps.config.target.result.os.tag.isDarwin()) return .{
-        .step = &lib.step,
-        .output = lib.getEmittedBin(),
-        .dsym = null,
-        .pkg_config = null,
-        .pkg_config_static = null,
-    };
-
-    // Create a static lib that contains all our dependencies.
-    const libtool = LibtoolStep.create(b, .{
-        .name = "ghostty",
-        .out_name = "libghostty-fat.a",
-        .sources = lib_list.items,
-    });
-    libtool.step.dependOn(&lib.step);
+    // Combine all archives into a single fat static library so
+    // consumers only need to link one file.
+    const combined = combineArchives(b, deps.config.target, lib_list.items);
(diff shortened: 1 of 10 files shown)


mitchellh:
Rebase this please

deblasis:
done

src/build/GhosttyLib.zig
```diff
@@ -232,6 +221,61 @@ pub fn installHeader(self: *const GhosttyLib) void {
     b.getInstallStep().dependOn(&header_install.step);
 }
 
+/// Combine multiple static archives into a single fat archive.
+/// Uses libtool on Darwin, lib.exe on Windows, and ar MRI scripts
+/// on other platforms.
+fn combineArchives(
```
mitchellh: Would be nice to combine this with ghosttylibvt combineArchives and also fix the names to the new internal name here.

mitchellh (changes requested):

jcollie:
@deblasis looks like this needs another rebase.

deblasis:
@jcollie done 🫡

mitchellh (approved):
Approved pending CI.

PR #12226, opened 2026-04-10
core: implement backarrow key mode (DECBKM) - mode 67

jcollie:
This mode allows programs to modify the code that the `backspace` key (backarrow key in DEC parlance) sends. If this mode is `off`/`false`/`reset` (the default, the same as before this PR), we send the byte `0x7f`. If this mode is `on`/`true`/`set` we send the byte `0x08`.

Screenshot From 2026-04-09 11-00-25

### DIFF
--- a/src/input/key_encode.zig
+++ b/src/input/key_encode.zig
@@ -189,9 +189,7 @@ fn kitty(
                 switch (event.key) {
                     .enter => return try writer.writeByte('\r'),
                     .tab => return try writer.writeByte('\t'),
-                    .backspace => return try writer.writeByte(
-                        if (opts.backarrow_key_mode) 0x08 else 0x7F,
-                    ),
+                    .backspace => return try writer.writeByte(0x7F),
                     else => {},
                 }
             }
@@ -664,7 +662,10 @@ fn pcStyleFunctionKey(
             continue;
         }
 
-        if (keyval == .backspace and backarrow_key_mode) return "\x08";
+        decbkm: {
+            if (!backarrow_key_mode) break :decbkm;
+            return entry.sequence_decbkm orelse break :decbkm;
+        }
 
         return entry.sequence;
     }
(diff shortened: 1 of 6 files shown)


src/input/key_encode.zig
```diff
@@ -182,7 +189,9 @@ fn kitty(
                 switch (event.key) {
                     .enter => return try writer.writeByte('\r'),
                     .tab => return try writer.writeByte('\t'),
-                    .backspace => return try writer.writeByte(0x7F),
+                    .backspace => return try writer.writeByte(
+                        if (opts.backarrow_key_mode) 0x08 else 0x7F,
```
mitchellh: If Kitty doesn't do this then we shouldn't either. Did we verify what Kitty does? If Kitty encoding is on we should ignore all other modes probably (unless Kitty respects them).

mitchellh (changes requested):

mitchellh:
From xterm (input.c):

```c
/* VT300 & up: backarrow toggle */
if ((kd.nbytes == 1)
    && IsBackarrowToggle(keyboard, kd.keysym, evt_state)) {
    kd.strbuf[0] = ANSI_DEL;
}
```

The `IsBackarrowToggle` macro:
```c
#define IsBackarrowToggle(keyboard, keysym, state) \
    ((((keyboard->flags & MODE_DECBKM) == 0) \
        ^ ((state & ControlMask) != 0)) \
    && (keysym == XK_BackSpace))
```

Behavior I see:

| DECBKM | Key              | Sends         |
|--------|------------------|---------------|
| Set    | Backspace        | BS (`0x08`)   |
| Set    | Ctrl+Backspace   | DEL (`0x7F`)  |
| Reset  | Backspace        | DEL (`0x7F`)  |
| Reset  | Ctrl+Backspace   | BS (`0x08`)   |

Let's test this and see if we can do the same.

src/input/key_encode.zig
jcollie: Kitty doesn't support DECBKM so it obviously won't change. Reverted this.

jcollie:
Updated so that `ctrl` swaps what backspace returns, even under DECBKM.

src/input/key_encode.zig
```diff
@@ -653,6 +662,11 @@ fn pcStyleFunctionKey(
             continue;
         }
 
+        decbkm: {
+            if (!backarrow_key_mode) break :decbkm;
+            return entry.sequence_decbkm orelse break :decbkm;
+        }
```
mitchellh: This is kind of a weird way to write this...

mitchellh (changes requested):
One small change requested, its just some odd styling. The rest looking good.

mitchellh (changes requested):
Sorry for going back and forth on this, one thing missing here is porting this to libghostty-vt key encoding APIs. We should implemnt in the setopt from terminal and setopt.

jcollie:
Added the `libghostty-vt` bits.

mitchellh (approved):
Thanks. In the future we don't always have to pair the libghostty-vt stuff with vt work but I knew this was a simpleone.

PR #12230, opened 2026-04-10
libghostty: add a test to ensure terminal modes are defined

jcollie:
Adds a test to make sure that all terminal modes defined in Zig have a corresponding definition in `ghostty/vt.h`.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -320,7 +320,7 @@ pub fn build(b: *std.Build) !void {
             .root_module = mod.vt,
             .filters = test_filters,
         });
-        try c_deps.add(b, .ghostty_vt_h, mod.vt, &config, .{
+        try c_deps.add(b, .@"ghostty-vt-h", mod.vt, &config, .{
             .target = config.baselineTarget(),
             .optimize = .Debug,
         });
@@ -331,7 +331,7 @@ pub fn build(b: *std.Build) !void {
             .root_module = mod.vt_c,
             .filters = test_filters,
         });
-        try c_deps.add(b, .ghostty_vt_h, mod.vt_c, &config, .{
+        try c_deps.add(b, .@"ghostty-vt-h", mod.vt_c, &config, .{
             .target = config.baselineTarget(),
             .optimize = .Debug,
         });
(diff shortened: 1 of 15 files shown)


build.zig
```diff
@@ -357,6 +357,15 @@ pub fn build(b: *std.Build) !void {
...
         test_exe.root_module.addImport("ghostty.h", ghostty_h.createModule());
 
+        // Verify our libghostty-vt header.
+        const ghostty_vt_h = b.addTranslateC(.{
+            .root_source_file = b.path("include/ghostty/vt.h"),
+            .target = config.baselineTarget(),
+            .optimize = .Debug,
+        });
+        ghostty_vt_h.addIncludePath(b.path("include"));
+        test_exe.root_module.addImport("ghostty/vt.h", ghostty_vt_h.createModule());
```
mitchellh: We might do this elsewhere, but its kind of confusing, I'd prefer to name it like `ghostty-vt-h` so it doesn't seem like there's some magic going on here.

src/lib/enum.zig
```diff
@@ -143,3 +143,360 @@ pub fn checkGhosttyHEnum(
...
+    comptime prefix: []const u8,
+) !void {
+    @setEvalBranchQuota(100_000);
+
+    const info = @typeInfo(T);
+
+    try std.testing.expect(info == .@"enum");
+    try std.testing.expect(info.@"enum".is_exhaustive == true);
+
+    const c = @import("ghostty/vt.h");
```
mitchellh: I've run into problems with the other check where imports are processed too eagerly so it makes separate compilation harder. Can `@import` take a comptime string (I don't think so)? Or maybe the function should be `checkCMacroExists` and take a `comptime c: anytype` that is given the import from src/terminal/c/modes.zig` or something.

src/lib/enum.zig
```diff
@@ -143,3 +143,360 @@ pub fn checkGhosttyHEnum(
...
+// because `valuesFromFields()` needs to have a larger `@setEvalBranchQuota` set
+// to handle larger enums.
+//
+// https://codeberg.org/ziglang/zig/src/commit/e4cbd752c8c05f131051f8c873cff7823177d7d3/lib/std/enums.zig
+
+/// A set of enum elements, backed by a bitfield.  If the enum
+/// is exhaustive but not dense, a mapping will be constructed from enum values
+/// to dense indices.  This type does no dynamic allocation and
+/// can be copied by value.
+fn EnumSet(comptime E: type) type {
```
mitchellh: Gross, but I understand. To make this file not so noisy though maybe lets put this in `enum_zig_stdlib.zig` and import it only here.

mitchellh (changes requested):

build.zig
jcollie: Done.

src/lib/enum.zig
jcollie: Done.

src/lib/enum.zig
jcollie: `@import` can't take a comptime string but I switched to passing  `comptime c: anytpe` and combined the functions to reduce code duplication.

jcollie:
`aro-c` or `translate-c` doesn't seem to like `ghostty/vt.h` much. Sending out the @vancluever -signal for some assistance! Marking as draft until fixed.

PR #12244, opened 2026-04-12
feat: draw IME cursor

otomn:
Closes #11837 
Based on discussions of #12213

Change summary:
Pass the IME preedit cursor position from gtk/appkit to renderer. Draw the cursor using the same logic as a regular cursor.

Note:
The changes in `src/renderer/generic.zig` looks a bit complicated, but it is mostly just moving code around so that `addCursor` can be reused by the IME preedit cursor.


pluiedev (approved):
Hey, sorry for leaving this out to dry for so long. GTK parts look pretty good.

PR #12245, opened 2026-04-12
macOS: fix cannot rebind super+up and super+down

otomn:
Fix: #11989
Cause identified to: ab352b5af9694a7cba8e237d0b1b5a507a6e4226
Original PR: #10003
Problem: I don't think it is OK to hard code the keybind like this at all. Ghostty's config is flexible enough to achieve this.
Proposal: Revert the above commit via this PR.

@yasuf @bo2themax


bo2themax (approved):
> Problem: I don't think it is OK to hard code the keybind like this at all. Ghostty's config is flexible enough to achieve this.

Nice catch, thanks. And agree with you on this

PR #12251, opened 2026-04-12
macOS: add bottom bar when child exits

### LINKED ISSUE #7649: Use native GUI to show child process exited with `wait-after-command`

bo2themax:
### Closes #7649

The bar lives alongside URL Hover in VStack at the bottom. The current body of SurfaceView is becoming rather long and complicated, so this pr also contains some refactors:

- Move URL Hover to a separate file

> The text is copied from previous input string to keep it consistent, also I’m confused with text on GTK so this is my first choice, but it can be changed as the same as GTK.

Separate prs will be opened for:
1. Set to Read-only after exits
2. Hide cursor when in Read-only

### Preview

https://github.com/user-attachments/assets/eb44e211-eac5-4f40-836c-4912b18dfb01

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -662,8 +662,7 @@ extension Ghostty {
             case GHOSTTY_ACTION_QUIT_TIMER:
                 fallthrough
             case GHOSTTY_ACTION_SHOW_CHILD_EXITED:
-                Ghostty.logger.info("known but unimplemented action action=\(action.tag.rawValue)")
-                return false
+                return showChildExited(app, target: target, v: action.action.child_exited)
             case GHOSTTY_ACTION_COPY_TITLE_TO_CLIPBOARD:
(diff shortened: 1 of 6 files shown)


mitchellh:
Rebase to make readable please

mitchellh (approved):

PR #12252, opened 2026-04-12
core: hide cursor when surface is read only

bo2themax:
Follow up for #12251


mitchellh (changes requested):
This isn't the right way to do this, because the terminal program can undo it via standard sequences. If we want to always hide the cursor, we should set some state on the render state and handle it in the renderer.

PR #12259, opened 2026-04-13
zon2nix: update to a version that is compatible with Zig 0.16

jcollie:
The `zon2nix` binary is now compiled with Zig 0.16, but it still produces Zig 0.15 compatible output (in fact the output is identical to previous versions).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -32,7 +32,7 @@
     };
 
     zon2nix = {
-      url = "git+https://codeberg.org/jcollie/zon2nix?ref=0.16";
+      url = "github:jcollie/zon2nix?ref=0.16";
       inputs = {
         nixpkgs.follows = "nixpkgs";
       };
--- a/flake.nix
+++ b/flake.nix
@@ -32,7 +32,7 @@
     };
 
     zon2nix = {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -32,7 +32,7 @@
     };
 
     zon2nix = {
-      url = "github:jcollie/zon2nix?rev=c28e93f3ba133d4c1b1d65224e2eebede61fd071";
+      url = "git+https://codeberg.org/jcollie/zon2nix?ref=0.16";
```
mitchellh: I super don't want to depend on codeberg if we can help it. Everytime we've introduced a codeberg dep our CI has suffered since they've historically (not sure recently) had horrible uptime. Any way we can cache this somewhere?

mitchellh (changes requested):

flake.nix
mitchellh: I'll note we do use cachix but I forget if this caches these... I also forget if that works properly on our macOS builds which still use Nix. If it all does then we can probably accept this because it'll probably be cached.

jcollie: Switched to using a github mirror.

PR #12264, opened 2026-04-13
macOS: Command palette highlight matches

bo2themax:
- Highlight matching text in command palette search results
- Support initials matching
- Trim query before filtering commands

### AI Disclosure

Claude wrote most of it. I tested and reviewed it myself.

image


mitchellh (approved):
A good middle ground to real fuzzy searching atm.

PR #12273, opened 2026-04-13
fix: send modifier when scrolling in alternate window

otomn:
Fix: #11971
Problem: the original code sends raw encoding for arrow up and arrow down when scrolling in an alternate window.
Changes: the function is modified to call `input.key_encode.encode` to properly encode the modifier with the arrow up/down. The mutex lock had to be moved out of the `encode` function as the lock has already been acquired.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3160,7 +3160,9 @@ fn encodeKey(
 ) !?termio.Message.WriteReq {
     const write_req: termio.Message.WriteReq = req: {
         // Build our encoding options, which requires the lock.
+        self.renderer_state.mutex.lock();
         const encoding_opts = self.encodeKeyOpts();
+        self.renderer_state.mutex.unlock();
 
         // Try to write the input into a small array. This fits almost
         // every scenario. Larger situations can happen due to long
@@ -3229,8 +3231,6 @@ fn encodeKey(
 }
 
 fn encodeKeyOpts(self: *const Surface) input.key_encode.Options {
-    self.renderer_state.mutex.lock();
-    defer self.renderer_state.mutex.unlock();
     const t = &self.io.terminal;
 
     var opts: input.key_encode.Options = .fromTerminal(t);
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Given the very specific check here for alt screen, I wonder if this is an intended behavior perhaps from xterm. Its been very long since we checked, but it'd be worth a quick audit of the xterm code base to ensure this is safe.

otomn:
> Given the very specific check here for alt screen, I wonder if this is an intended behavior perhaps from xterm. Its been very long since we checked, but it'd be worth a quick audit of the xterm code base to ensure this is safe.

@mitchellh 

Sorry, what in particular do you want to check?
I don't see any risk here. The standard says when `mouse_alternate_scroll` is enabled in an alternate screen and mouse reporting is off, mouse wheel should send arrow up/down. This behaviour is being preserved here. As when you are not holding any modifier key, `input.key_encode.encode` does output just arrow up/down.
As for what happens when a modifier is held, the standard doesn't say anything. Terminal.app chose to consume the input itself and scroll into the history buffer (which I really disliked). iTerm does was I'm proposing here: encoding the modifier into the synthetic up/down key. xterm source code suggests it doesn't encode the modifier. But I believe this is just a choice rather than a law.
From the program's perspective, it is only processing key strokes. Be it just arrow up/down or shift arrow up/down. Encoding the modifier key while scrolling is totally equivalent to holding the modifier while pressing up/down keys yourself. If there is no problem doing that, then I don't see how it would be problematic to encode the modifiers for synthetic key strokes.

xterm alternate scroll feature added in https://invisible-island.net/xterm/xterm.log.html#xterm_282 in 2012

Original feature request for xterm to add the alternate scrolling https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=683942
> ... It turns out gnome-terminal
has a feature called "alternate screen scrolling".  When you are in the
alternate screen, it translates the mouse wheel into three up or down
arrow presses. This is obviously a hack, but I want it.

xterm source `ctlseqs.txt` line 3012
```
However if Alternate Scroll mode is set, then cursor up/down controls
are sent when the terminal is displaying the Alternate Screen Buffer.
The initial state of Alternate Scroll mode is set using the
alternateScroll resource.
```

xterm source `scrollbar.c` line 708
```c
static void
AlternateScroll(Widget w, long amount)
{
    XtermWidget xw;
    TScreen *screen;

    if ((xw = getXtermWidget(w)) != NULL &&
	(screen = TScreenOf(xw)) != NULL &&
	screen->alternateScroll && screen->whichBuf) {
	ANSI reply;

	amount /= FontHeight(screen);
	memset(&reply, 0, sizeof(reply));
	reply.a_type = ((xw->keyboard.flags & MODE_DECCKM)
			? ANSI_SS3
			: ANSI_CSI);
	if (amount > 0) {
	    reply.a_final = 'B';
	} else {
	    amount = -amount;
	    reply.a_final = 'A';
	}
	while (amount-- > 0) {
	    unparseseq(xw, &reply);
	}
    } else {
	ScrollTextUpDownBy(w, (XtPointer) 0, (XtPointer) amount);
    }
}
```

otomn:
I say standard, but as you can see, there is no standard really. Different terminals converge to implement the same feature at different times because it is a handy feature.

mitchellh:
Yes, understood, but the way Ghostty defines standard is well defined:

> For terminal application developers, we believe Ghostty is one of the most modern and comprehensive terminal emulators available. Our goal with Ghostty is to be the most compatible for legacy applications while also providing the most modern features for new applications.
> 
> What that means in practice is that Ghostty supports a wide range of control sequences, terminal features, and terminal protocols. And to determine the behavior of these features, we follow the following principles:
> 
> 1. Xterm compatibility: xterm is the de facto standard for terminal emulation. If xterm does something, we should do the same thing.
> 
> 2. Protocol origin compatibility: If the terminal that defined a protocol behaves a certain way, we should follow that behavior, even if it is not specified.
> 
> 3. Defacto standard compatibility: If a behavior is widely accepted as the standard, we should follow that behavior. "Widely accepted" is generally subjective based on popular opinion and usage.

We have, under very rare circumstances, broken rule 1 if there is an overwhelming reason to. The reason its nice to have is because besides being pretty much the defacto standard terminal today, it lets us point to weird behaviors and say "xterm does it so do we" instead of having a per-feature defense strategy which is mentally overwhelming.

mitchellh:
I'll add that in the past, the way we've worked through this is requesting a contributor get acknowledgement from xterm themselves on whether this is intentional or incidental behavior. If it is the latter, we will break from xterm and have the mailing list or email source of truth to prove why. Thomas Dickey has generally been responsive.

mitchellh:
As it stands now, I'm going to close this PR. We're clearly still in the discussion phase and we should go back to #11971. This is why discussions exists prior to PRs. I'll reference the PR and associated key comments there.

PR #12291, opened 2026-04-15
libghostty: Remove all libc++ and libc++ ABI dependencies

mitchellh:
This updates simdutf to my fork which has a SIMDUTF_NO_LIBCXX option that removes all libc++ and libc++ ABI dependencies. The plan is to open an upstream PR with this, but I want to verify it here first.

From there, the hand-written simd code we have has been updated to also no longer use any libc++ features. Part of this required removing utfcpp since it depended on libc++ (`<iterator>`).

libghostty-vt now only depends on libc.

## Benchmark Results

| Corpus | Current `HEAD` median | `main` median | Delta vs `main` | Notes |
| --- | ---: | ---: | ---: | --- |
| `valid-mixed-1g-seed1.bin` | `9.245s` | `9.111s` | `1.5%` slower | Near tie; `main` remains slightly faster on fully valid input |
| `malformed-mixed-1g-seed1-rate0.005.bin` | `9.251s` | `12.705s` | `37.3%` faster | Large improvement on malformed UTF-8 input |

Approximate throughput from the medians:

- Valid corpus: current `HEAD` `110.8 MiB/s`, `main` `112.4 MiB/s`
- Malformed corpus: current `HEAD` `110.7 MiB/s`, `main` `80.6 MiB/s`


mitchellh:
I actually found a way to do replacement character decoding more in concert with simdutf. Our invalid utf8 path is probably a lot faster now, but will verify with benchmarks soon!

PR #12301, opened 2026-04-15
i18n: add Icelandic translations (is)

bleikurr:
I'm adding in Icelandic (is) translations.
---

There might only be a couple of Icelanders out there so believe it might be unlikely to find another one, but I already made the translations so I might as well try and get them in. 

I'm also willing to join the localization team if this manages to go through.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -193,6 +193,7 @@
 /po/vi.po @ghostty-org/vi_VN
 /po/zh_CN.po @ghostty-org/zh_CN
 /po/zh_TW.po @ghostty-org/zh_TW
+/po/is.po @ghostty-org/is_IS
 
 # Packaging - Snap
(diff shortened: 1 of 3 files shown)


trag1c:
!vouch

00-kat:
> There might only be a couple of Icelanders out there so believe it might be unlikely to find another one

As an FYI, there's no need for it to be someone completely unaffiliated with your or whatever; if you know a friend or anyone else who speaks Icelandic and is willing to help out, you might just want to ask them if they would.

trag1c:
Closing this as stale due to no interest from other Icelandic speakers (I know it's hard for you guys! 🥲) + many new strings have been added since this PR was opened that have to be translated as well. We're definitely open to accepting a PR adding the Icelandic locale once there are more people out there willing to maintain it. Thanks!

PR #12317, opened 2026-04-17
pkg: replace @cImport with addTranslateC in pkg/

mitchellh:
@cImport is going to disappear in Zig 0.17. Its deprecated in Zig 0.16. Let's remove it now.

Replace @cImport with addTranslateC across pkg/ packages. Each package now has a c_import.h header that is translated at build time via addTranslateC and exposed as a "cimport" module import.

Converted packages:
- dcimgui
- fontconfig
- freetype
- glslang
- harfbuzz
- macos
- oniguruma
- opengl
- sentry
- spirv-cross
- wuffs

Omitted:
- gtk4-layer-shell - This has a bit more complexity with how it interacts with include paths, so I need to consider this a bit more.
- src/ - It'll be cleaner to do this separately.


mitchellh:
Actually, closing this. Zig 0.16 adds `linkSystemLibrary` to `addTranslateC` which I think we need to support system builds. Its not worth investing in Zig 0.15 here. Once we're on Zig 0.16 we can easily take this diff and adapt it.

PR #12326, opened 2026-04-17
Feat: Add caret mode for keyboard-driven navigation of the scrollback

brianc442:
Directly related discussions: #3708 #3488 
Less directly related but still relevant discussions: #2394 #6916 #6917 #7672 

The caret appears as a hollow block cursor. Vim hotkeys are used by default. The implementation is functionally the same as vim visual mode or the copy mode found in tmux & WezTerm. One notable deviation from vim visual mode is that pressing `y` to copy the selection does not exit caret mode - this was a deliberate design decision as exiting caret mode redraws the screen at the cursor which may not be desired if the user was copying something several pages up in the scrollback. 

Written as an optional feature which is off by default. Requires a simple one line addition to the config file to create a keybind which activates caret mode:
e.g. `keybind = alt+v=enter_caret_mode`

The built in key table can be extended or overridden in ghostty/config (e.g. `keybind = caret/w=...`). All defaults can be cleared with `keybind = caret/` if desired.

Default keybinds: 
h/j/k/l or arrows: move caret
pg up/Ctrl+u/b: page up
pg dn/Ctrl+d/f: page down
gg/G: top/bottom of scrollback
0/$: beginning/end of line
/: search
n/N: search next/prev
v: toggle selection mode on/off (unlike vim, does not exit caret mode)
y: copy selection
Esc/q/i: exit caret mode

Code written with assistance from Claude Code. I reviewed, tested on Arch Linux, & ran the feature locally for ~1 month before submitting. It should work on macOS as well but I do not have a machine to test this.


src/Surface.zig
```diff
@@ -5819,6 +5826,128 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool
...
+            defer self.renderer_state.mutex.unlock();
+
+            const screen: *terminal.Screen = self.io.terminal.screens.active;
+            if (screen.caret_mode) return false;
+
+            try screen.enterCaretMode();
+
+            // Push the "caret" key table so caret bindings become active.
+            if (self.config.keybind.tables.getPtr("caret")) |set| {
+                if (self.keyboard.table_stack.items.len < max_active_key_tables) {
```
philocalyst, line 5841: I wonder if it would be better to switch the conditional order? The comparison is much cheaper.

src/terminal/render.zig
```diff
@@ -421,6 +426,25 @@ pub const RenderState = struct {
                 };
             }
 
+            // Find the caret position within the viewport.
+            if (s.caret_mode) {
+                if (self.caret == null) {
+                    if (s.caret_pin) |cp| {
```
philocalyst, line 432: An orelse here would do wonders! A few other simplifications may also be possible.

philocalyst (commented):
Would love to have this in Ghostty! There's only a couple blocks of real logic so I hope this gets moving soon :)

andrewykimse:
any update on this getting merged? hoping to have some thing like this in ghostty :heart:

jcollie:
At this point, this is likely to be punted until after the 1.4 release. So not a _no_, just _maybe later_. This will still be subject to the normal review process. As far as I know no one on the team has looked at this at all yet.

PR #12335, opened 2026-04-19
docs,ci: clarify that users can never open issues

trag1c:
I removed the entire paragraph in CONTRIBUTING.md because the "Quick Guide" section explains it all better already.


00-kat (approved):
Yippee!

PR #12336, opened 2026-04-19
Translation documentation-related typos + CODEOWNERS update

00-kat:


CODEOWNERS
```diff
@@ -164,6 +164,7 @@
 # Localization
 /po/README_TRANSLATORS.md @ghostty-org/localization
 /po/com.mitchellh.ghostty.pot @ghostty-org/localization
```
trag1c: Perhaps these two should be changed to `manager` too, I don't think there's any point in pinging 100+ people for changes in these.

trag1c (approved):

PR #12352, opened 2026-04-20
terminal: glyph protocol parser and response encoder

mitchellh:
**Important: this DOES NOT hook up the glyph protocol to Ghostty or libghostty. Its just the parser.**

This adds the core parse/encode for the still in-development and experimental terminal glyph protocol: https://github.com/raphamorim/rio/pull/1542

The only cross-cutting change necessary was changing the APC identification logic which previously only looked at a single byte to support multi-byte identifiers since the glyph protocol uses `25a1`.

For DoS protection, the default limits any glyph-related APC command size to 1 megabyte.

> [!WARNING]
> 
> Since this protocol is still in development and discussion, there is no promise the implementation will stay within Ghostty or that any of the APIs exposed by this will remain stable. We're just getting ahead of it.

### DIFF
--- a/src/terminal/apc.zig
+++ b/src/terminal/apc.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const build_options = @import("terminal_options");
 const Allocator = std.mem.Allocator;
 
+const glyph = @import("apc/glyph.zig");
 const kitty_gfx = @import("kitty/graphics.zig");
 
 const log = std.log.scoped(.terminal_apc);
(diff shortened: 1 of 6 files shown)


mitchellh:
Updates:

- Update to the latest spec v1.4 which modified some formats
- Have `get` apply defaults automatically for specified defaults (invalid values are not validated)
- Clear does not echo the codepoint back

mitchellh:
The parser/response has been updated to spec version 1.9.

mitchellh:
The parser is in. As a reminder, this isn't hooked up to anything beyond parsing. No other state, no rendering.

PR #12363, opened 2026-04-21
Use patched Zig 0.15.2 on macOS to avoid Xcode 26.4 issue

mitchellh:
This updates our Nix flake to use the Homebrew bottled Zig 0.15.2 which contains a patch to work around the issue with Zig 0.15.x and Xcode 26.4.


mitchellh:
The CI failures are unrelated cachix issues.

PR #12364, opened 2026-04-21
libghostty-vt: fix broken dynamic linking with pkg-config

sandydoo:
~`${prefix}/include` and `${prefix}/lib` are incorrect under split-prefix installs (e.g. Nix multi-output). Use `b.h_dir` / `b.lib_dir` instead and drop the unneeded Nix postInstall/postFixup hooks.~

Refactors the libghostty-vt derivation to:

- fix `libdir` pointing to the wrong output in the pkg-config files. This would throw a missing library error at runtime.
- reduce the amount of manual copying, linking, and patching of files.

An earlier version of this PR used the zig compiler + `.pc` files to do this. People pointed out concerns, so I came up with a simpler solution.

Claude Code was used to debug and write an initial fix. Final changes rewritten and simplified by me. No AI was used to write comments, descriptions, etc.

### DIFF
--- a/nix/libghostty-vt.nix
+++ b/nix/libghostty-vt.nix
@@ -59,9 +59,6 @@ stdenv.mkDerivation (finalAttrs: {
     "-Dapp-runtime=none"
     "-Demit-lib-vt=true"
     "-Dsimd=${lib.boolToString simd}"
-    # Install headers directly into the `dev` output
-    "--prefix-include-dir"
-    "${placeholder "dev"}/include"
   ];
   zigCheckFlags = finalAttrs.zigBuildFlags ++ ["test-lib-vt"];
 
--- a/nix/libghostty-vt.nix
+++ b/nix/libghostty-vt.nix
@@ -59,6 +59,9 @@ stdenv.mkDerivation (finalAttrs: {
     "-Dapp-runtime=none"
     "-Demit-lib-vt=true"
     "-Dsimd=${lib.boolToString simd}"
+    # Install headers directly into the `dev` output
+    "--prefix-include-dir"
+    "${placeholder "dev"}/include"
   ];
   zigCheckFlags = finalAttrs.zigBuildFlags ++ ["test-lib-vt"];
 
@@ -69,17 +72,10 @@ stdenv.mkDerivation (finalAttrs: {
 
   postInstall = ''
     mkdir -p "$dev/lib"
-    mv "$out/lib/libghostty-vt.a" "$dev/lib"
-    rm "$out/lib/libghostty-vt.so"
-    mv "$out/include" "$dev"
-    mv "$out/share" "$dev"
-
-    ln -sf "$out/lib/libghostty-vt.so.${lib.versions.major finalAttrs.version}"  "$dev/lib/libghostty-vt.so"
+    mv "$out/lib/libghostty-vt.a" "$dev/lib/"
   '';
 
   postFixup = ''
-    substituteInPlace "$dev/share/pkgconfig/libghostty-vt.pc" \
-      --replace-fail "$out" "$dev"
     substituteInPlace "$dev/share/pkgconfig/libghostty-vt-static.pc" \
       --replace-fail "$out" "$dev"
   '';
@@ -128,8 +124,7 @@ stdenv.mkDerivation (finalAttrs: {
         runHook preBuildHooks
 
         cc -o test test_libghostty_vt.c \
-          ''$(pkg-config --cflags --libs libghostty-vt) \
-          -Wl,-rpath,"${finalAttrs.finalPackage}/lib"
+          ''$(pkg-config --cflags --libs libghostty-vt)
 
         runHook postBuildHooks
       '';
@@ -207,8 +202,7 @@ stdenv.mkDerivation (finalAttrs: {
         runHook preBuildHooks
 
         cc -o test main.c \
-          ''$(pkg-config --cflags --libs libghostty-vt) \
-          -Wl,-rpath,"${finalAttrs.finalPackage}/lib"
+          ''$(pkg-config --cflags --libs libghostty-vt)
 
         runHook postBuildHooks
       '';
--- a/src/build/GhosttyLibVt.zig
+++ b/src/build/GhosttyLibVt.zig
@@ -371,8 +371,8 @@ fn pkgConfigFiles(
     return .{
         .shared = wf.add("libghostty-vt.pc", b.fmt(
             \\prefix={s}
-            \\includedir=${{prefix}}/include
-            \\libdir=${{prefix}}/lib
+            \\includedir={s}
+            \\libdir={s}
             \\
             \\Name: libghostty-vt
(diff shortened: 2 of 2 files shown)


nix/libghostty-vt.nix
```diff
@@ -67,37 +70,20 @@ stdenv.mkDerivation (finalAttrs: {
     "dev"
   ];
 
-  postInstall = ''
```
jcollie, line 70: Why are the `.so` and `.a` files moved to `out` ? There's no need for them at runtime if you're linking dynamically so there's no need to carry them around. If they are separated into the `dev` output it's only needed at build time.

jcollie (changes requested):

nix/libghostty-vt.nix
sandydoo: Ah, didn't notice the static library in there as well. I guess that can go into `$dev/lib` (or `$static`?). Pushed a version that moves it to `$dev/lib` and updates the static `.pc`.

tristan957 (commented):
Could you provide an example of before and after? It seems like we are no longer using the templating that pkgconfig provides. Is that correct? If so, I have reservations about that.

sandydoo:
> Could you provide an example of before and after? It seems like we are no longer using the templating that pkgconfig provides. Is that correct? If so, I have reservations about that.

Fair point. We can simplify things further. I've pushed a commit that restores the `.pc` files as they were and uses Nix to rewrite the `libdir` just for the static library (since it lives in `$dev`).

This way:

- Nix handles copying the headers to `$dev/include` and patches the `includedir` automatically. We don't need to do this manually.
- We get the original `.pc` files that use `${prefix}` everywhere.
- The only rewrite we need to do is for the static library, since we manually move it to `$dev`.
- All the `.so` files live canonically in `$out/lib`.

#### Before

*Note that `includedir` was rewritten by a hook (see `_multioutDevs` in `nixpkgs`).*

`libdir` is pointing to `$dev`.

```
prefix=/nix/store/l9ik5j9jch2if9z6141wq1ayjyhx2cb9-libghostty-vt-0.1.0-dev+62fdd88-nix-dev
includedir=/nix/store/l9ik5j9jch2if9z6141wq1ayjyhx2cb9-libghostty-vt-0.1.0-dev+62fdd88-nix-dev/include
libdir=${prefix}/lib

Name: libghostty-vt
URL: https://github.com/ghostty-org/ghostty
Description: Ghostty VT library
Version: 0.1.0-dev+62fdd88-nix
Cflags: -I${includedir}
Libs: -L${libdir} -lghostty-vt
Libs.private:
Requires.private: ⏎
```

#### After

```
prefix=/nix/store/d39iaqzgwn85s103ncy6ii5sf64j34hn-libghostty-vt-0.1.0-dev+5580e9b-nix
includedir=/nix/store/9zcq5668bz71m4n8vb38y9g42780c6f9-libghostty-vt-0.1.0-dev+5580e9b-nix-dev/include
libdir=${prefix}/lib

Name: libghostty-vt
URL: https://github.com/ghostty-org/ghostty
Description: Ghostty VT library
Version: 0.1.0-dev+5580e9b-nix
Cflags: -I${includedir}
Libs: -L${libdir} -lghostty-vt
Libs.private:
Requires.private:
```

#### Static

Here, Nix rewrote `includedir` for us and we rewrote `prefix` to point to `$dev/lib`.

```
prefix=/nix/store/j8069sg2f7qf9qxylniap600xicki48l-libghostty-vt-0.1.0-dev+733abbc-nix-dev
includedir=/nix/store/j8069sg2f7qf9qxylniap600xicki48l-libghostty-vt-0.1.0-dev+733abbc-nix-dev/include
libdir=${prefix}/lib

Name: libghostty-vt-static
URL: https://github.com/ghostty-org/ghostty
Description: Ghostty VT library (static)
Version: 0.1.0-dev+733abbc-nix
Cflags: -I${includedir}
Libs: ${libdir}/libghostty-vt.a
Libs.private:
Requires.private:
```

### Background

Some background on what we're trying to achieve here.

**What we want:** to build this library with Nix and link against it using `pkg-config`.
**What we don't want**: to have to do the equivalent of `-Wl,-rpath,"${libghostty-vt}/lib"` when consuming this library downstream.

The reason this rpath is necessary is because of the `$dev` shenanigans. The `.pc` prefix is set to the `$dev` output where the `.so` symlink lives, but the versioned library still lives in `$out`.
So what happens when we link against it? pkg-config emits `-L$dev/lib`, we link against `libghostty-vt.so.<x>`, Nix auto-rpaths the `$dev` directory, and we get a runtime error when we fail to find `libghostty-vt.so.<x>` in `$dev`.

jcollie:
The `pkg-config` files themselves should be in `dev` though.

mitchellh:
@sandydoo I'm going to raise a red flag that your communication seems to be in the style of an AI. Our project expressly disallows any AI usage _at all for communication_. We're humans talking to humans. We also require that all contributors be able to fully stand on their own -- without AI -- in explaining how everything works and their decision making process. I don't know if that is whats happening, but I'll note that so far the feedback on this PR has been... odd.

tristan957:
The `Before` and `After` look exactly the same to me. What changed?

sandydoo:
> @sandydoo I'm going to raise a red flag that your communication seems to be in the style of an AI. Our project expressly disallows any AI usage _at all for communication_. We're humans talking to humans. We also require that all contributors be able to fully stand on their own -- without AI -- in explaining how everything works and their decision making process. I don't know if that is whats happening, but I'll note that so far the feedback on this PR has been... odd.

Ugh. I didn't use AI to write any of the comments. I spent 45 minutes writing the above comment because I realized that I might not be communicating the point of all of this well enough.

One more time. I want to add pkg-config support to libghostty-rs. The `.pc` files produced by Nix here are a little broken. Both reviewers pointed out things that made sense, so I moved things around and things simplified down nicely.

> The Before and After look exactly the same to me. What changed?

The prefix is pointing to the wrong output in the before. `$dev` doesn't contain the libraries. That's why the test build _had_ the rpath flag pointing to `$out` instead.

sandydoo:
> The pkg-config files themselves should be in dev though.

They _are_ in `$dev`.  Run:

```
nix-build -A packages.<system>.libghostty-vt.dev
```
or `nix build .#libghostty-vt.dev`

sandydoo:
I've updated the title and description to reflect the current state of the PR. Thank you to the reviewers for, well, reviewing.

I'm going to go process whatever the heck went wrong tonight.

mitchellh:
A group of maintainers that get something like 20 to 50 AI messages per day including PRs and issues is just a ton of trauma and lack of trust of everything around us. Like I said I wasn’t sure if you were using AI, just reminding the rules in case. 

Will take a look.

jcollie (approved):

jcollie:
> > The pkg-config files themselves should be in dev though.
> 
> They _are_ in `$dev`. Run:

Ah… Some Nix magic took care of that for us. I was confused since the code for doing that was removed from the `postInstall` section.

PR #12373, opened 2026-04-22
windows: provide DllMain stub for non-MSVC ABI

mattn:
Part of preparation for adding a Win32 application runtime (discussion #2563). One of three small, independent build fixes that together unblock the Windows GNU-ABI library build.

On Windows with non-MSVC ABI, `pub const DllMain` resolved to `void` (a type), and Zig's stdlib `start.zig` then tried to call it as a function via `root.DllMain(...)`, failing to compile with "type 'type' not a function".

This restructures the conditional so MSVC keeps its existing CRT-init handler unchanged, non-MSVC Windows gets a no-op `BOOL` handler, and non-Windows continues to resolve to `void`.

Verified: `zig build -Dtarget=native-native-gnu -Dapp-runtime=none [-Doptimize=ReleaseSafe]` now builds cleanly on Windows.

### DIFF
--- a/src/main_c.zig
+++ b/src/main_c.zig
@@ -161,14 +161,17 @@ pub export fn ghostty_string_free(str: String) void {
 // function that depends on CRT internal state (setlocale, malloc from C
 // dependencies, C++ constructors in glslang) crashes with null pointer
 // dereferences. Declaring DllMain causes Zig's start.zig to call it
-// during DLL_PROCESS_ATTACH/DETACH, and we forward to the CRT bootstrap
-// functions from libvcruntime and libucrt (already linked).
+// during DLL_PROCESS_ATTACH/DETACH, and for MSVC we forward to the CRT
+// bootstrap functions from libvcruntime and libucrt (already linked).
+// For other ABIs (MinGW) the handler is a no-op since dllcrt2.obj already
+// handles CRT init; we still need `DllMain` declared so that Zig's
+// start.zig does not fall back to calling a non-function value.
 //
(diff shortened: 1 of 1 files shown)


jcollie (changes requested):
I think that there's a way to structure this a little more cleanly than these nested ifs.

mattn (commented):
Thanks for the review. Could you share the shape you have in mind? I'm considering two options:

1. A single `struct` with the `abi` check pushed into the handler body (eliminates the duplicated `const BOOL = ...` block).
2. A `blk:` block with an early `break` for each case (flattens the if/else chain but keeps the two structs).

Happy to go with either - just want to converge on your preference before pushing.

jcollie:
> Thanks for the review. Could you share the shape you have in mind? I'm considering two options:
> 
>     1. A single `struct` with the `abi` check pushed into the handler body (eliminates the duplicated `const BOOL = ...` block).

Yeah, this is what I was leaning towards.

jcollie:
Do we need to add something to CI to test Windows builds with both MSVC and GNU ABIs? Bitrot is likely if we don't.

mattn (commented):
Yes, planned. Two more small build fixes follow (disable C++ ubsan on Windows for non-MSVC deps; pass `b.graph.zig_exe` to `combine_archives`), which together unblock `-Dtarget=native-native-gnu -Dapp-runtime=none`. Right after those I'll add a GNU-ABI Windows library build job to complement the existing MSVC `build-libghostty-vt-windows`. Tracked at mattn/ghostty#1 (piecemeal Windows apprt upstreaming per #2563).

jcollie (approved):

PR #12375, opened 2026-04-22
Add code for validating OpenType GLYF table entries

qwerasd205:
This code was motivated by the need for the glyph protocol handler (#12352) to be able to validate the provided `glyf` payload, without having to link freetype or anything (because libghostty-vt needs to be static). As such it's written specifically to meet those needs, but in such a way that it can be expanded if we find a need for more in-depth inspection of `glyf`s in the future.

### DIFF
--- a/src/font/opentype.zig
+++ b/src/font/opentype.zig
@@ -5,12 +5,14 @@ const os2 = @import("opentype/os2.zig");
 const post = @import("opentype/post.zig");
 const hhea = @import("opentype/hhea.zig");
 const head = @import("opentype/head.zig");
+const glyf = @import("opentype/glyf.zig");
 
 pub const SVG = svg.SVG;
 pub const OS2 = os2.OS2;
 pub const Post = post.Post;
 pub const Hhea = hhea.Hhea;
 pub const Head = head.Head;
+pub const Glyf = glyf.Glyf;
 
 test {
     @import("std").testing.refAllDecls(@This());
--- a/src/font/opentype/glyf.zig
+++ b/src/font/opentype/glyf.zig
@@ -1,0 +1,434 @@
+const std = @import("std");
+const sfnt = @import("sfnt.zig");
+
+/// Glyph Data Table
+///
+/// This takes a little bit of a different form than other tables that we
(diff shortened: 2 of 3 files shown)


mitchellh (changes requested):
From the spec

> If a glyph has zero contours, no additional glyph data beyond the header is required. A glyph with zero contours may have additional data, however; in particular, it may have instructions that operate on phantom points.

This test should pass:

```zig
test "glyf: zero-contour glyph can be header-only" {
    const testing = std.testing;

    const header: Glyf.Entry.Header = .{
        .numberOfContours = 0,
        .xMin = 0,
        .yMin = 0,
        .xMax = 0,
        .yMax = 0,
    };
    const glyph = try Glyf.Entry.init(std.mem.asBytes(&header));
    try testing.expectEqual(@sizeOf(Glyf.Entry.Header), try glyph.size());
}
```

qwerasd205:
Haha yeah I actually realized I had written that bug while I was going to sleep last night and going back over the spec in my head- adding a test (and if need be, a fix) for it was on my todo list for today.

mitchellh:
I realize for the purpose of the glyph protocol its a silly glyph cause its... empty... but it is valid. 🤷

qwerasd205:
But actually, no that test is incorrect-- the additional data after the header is a few bytes still since it has to encode the instructions length. At least 2 extra bytes for the 16-bit `instructionLength` field, and then however many bytes for the `instructions` array.

~~The spec is worded in a confusing way that seems to imply a bare header should be allowed but I have no clue how that would function in an actual font file, since there's nothing in the header that encodes whether or not it potentially contains instructions~~... oh, oh no actually the length of any entry is determined by the difference to the subsequence `loca` offset, so I guess it *does* need to allow truncation after the header if the contour count is 0.

mitchellh:
It is confusing, and semantically for our use case it probably doesn't matter. I read this:

> however; in particular, it may have instructions that operate on phantom points.

As meaning the instructions part was optional.

qwerasd205:
Yep, see my edits on my last reply. I'll make a fix for it in an hour or so.

mitchellh (approved):

PR #12376, opened 2026-04-22
TouchBar Support

tbrundige:
TouchBar Support as mentioned here - https://github.com/ghostty-org/ghostty/discussions/12327

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -61,6 +61,15 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
     /// The notification cancellable for focused surface property changes.
     private var surfaceAppearanceCancellables: Set<AnyCancellable> = []
 
+    private var touchBarController: TouchBarController?
+
+    override func makeTouchBar() -> NSTouchBar {
+        if touchBarController == nil {
+            touchBarController = TouchBarController(target: self)
+        }
+        return touchBarController!.makeTouchBar()
+    }
+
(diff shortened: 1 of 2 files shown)


macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -61,6 +61,8 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
     /// The notification cancellable for focused surface property changes.
     private var surfaceAppearanceCancellables: Set<AnyCancellable> = []
 
+    private var touchBarController: TouchBarController?
```
bo2themax, line 64: This looks simple enough for me, but the controller should be in BaseTerminalController, so that quick terminal can also use this. 

Since quick terminal doesn't support tabs, so the buttons should be checked.

I'm thinking base provides common controls like spilts, and subclass can override that to provide more appropriate controls for that window.

bo2themax (changes requested):

bo2themax:
As a first pr for this feature, we should lay out our groundwork well enough before merging, so that future controls (like the one you mentioned in the discussion) can be easily implemented.

mitchellh:
I apologize but I've said this elsewhere (and will copy this into the discussion): I think TouchBar support is awesome _for a fork_. Its something I don't have access to and the population of TouchBar users is only getting smaller everyday so I don't think it makes sense to support in a mainline release.

tbrundige:
Thanks @mitchellh. I didn't see any previous discussion about TouchBar support but that makes sense. Might be time for me to retire my TouchBar MacBook 😂

PR #12381, opened 2026-04-23
windows: disable C++ ubsan regardless of ABI

mattn:
Widens the existing `-fno-sanitize=undefined` gate from `abi == .msvc` to `os.tag == .windows`. The same undefined `__ubsan_handle_*` link errors from simdutf/highway also reproduce on Windows GNU ABI, and the fix is identical.


mitchellh (approved):
Approved pending CI

PR #12382, opened 2026-04-23
build: pass zig exe path to combine_archives

mattn:
`combine_archives` spawns `zig ar -M`, hard-coding the command name `"zig"` and relying on the binary being on `PATH`. On Windows when the build is driven by an absolute zig.exe path (common in CI and Scoop/winget installs), this surfaces as `error: FileNotFound`.

Pass `b.graph.zig_exe` explicitly so the tool always uses the exact zig binary driving the build, matching how other build tools in this repo spawn zig subcommands.

### DIFF
--- a/src/build/GhosttyLibVt.zig
+++ b/src/build/GhosttyLibVt.zig
@@ -339,6 +339,7 @@ fn combineArchives(
         }),
     });
     const run = b.addRunArtifact(tool);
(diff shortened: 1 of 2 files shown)


jcollie:
This seems OK to me but we'll need to wait for our API quota to reset.

jcollie (approved):

PR #12384, opened 2026-04-23
cli: list fonts on Windows via FreeType

mattn:
Adds a Windows-specific code path to `ghostty +list-fonts` that scans the system and per-user font directories with FreeType, then prints families and style variants in the same format as the native-discovery backends. Previously, running `+list-fonts` on a Windows build errored out with "font discovery not available on this target".

Gated on `builtin.os.tag == .windows` so non-Windows builds are unchanged. The command itself can't be invoked on upstream today (no Windows apprt yet, so no `ghostty.exe`), but lands now so it is ready and reviewable ahead of the Win32 apprt. Compile is verified by the `build-libghostty-windows-gnu` CI job.

### DIFF
--- a/src/cli/list_fonts.zig
+++ b/src/cli/list_fonts.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 const ArenaAllocator = std.heap.ArenaAllocator;
 const Action = @import("ghostty.zig").Action;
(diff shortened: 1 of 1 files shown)


jcollie (changes requested):
I think that we need to say no to this one. The whole point of `+list-fonts` is to show what fonts are discoverable by Ghostty's internal font discovery mechanisms. If Ghostty's internal font discovery mechanisms can't discover it, `+list-fonts` shouldn't be telling the user that it can.

mattn:
Superseded by #12386, which adds a proper `Discover` backend for Windows. Once that lands, `+list-fonts` will work on Windows through the standard discovery code path -- addressing your review comment that `+list-fonts` should only show fonts the internal discovery can find. Closing here.

Sorry, I didn't check whether this implementation is valid.

PR #12386, opened 2026-04-23
font: add Windows font discovery backend

mattn:
Adds a FreeType-based `Discover` implementation for Windows. It walks the system font directory (`%SYSTEMROOT%\Fonts`) and the per-user directory (`%LOCALAPPDATA%\Microsoft\Windows\Fonts`), matches descriptors by FreeType `family_name` (falling back to the SFNT name table), and, when a codepoint is in the descriptor, filters on CMap coverage.

Wired up as a new `.freetype_windows` backend which `Backend.default()` now returns on Windows. Existing freetype-only paths are untouched and no other platform is affected; cross-platform switches were extended to handle the new enum value the same way they handle `.freetype`.

With this in place, the standard code paths (`+list-fonts`, `SharedGridSet` font-family lookup, `CodepointResolver` fallback) work on Windows without any `os.tag == .windows` branches in the caller.

Verified by the `build-libghostty-windows-gnu` CI job. No runtime binary ships yet on Windows (no apprt), but this is a drop-in for the discovery API that the Win32 apprt (and the revisited `+list-fonts` PR #12384) will use. Once this lands, #12384 can be closed and `+list-fonts` will work on Windows through the ordinary discovery code path, which addresses the review feedback that `+list-fonts` should only show fonts the internal discovery can find.

---

AI usage disclosure: developed with Claude Code (Claude Opus 4.7). Claude drafted the implementation based on my design direction -- I picked the "add a Discover backend" shape over the ad-hoc approach in the earlier `+list-fonts` PR. I reviewed each diff and validated it with a Windows GNU-ABI smoke build before pushing.

### DIFF
--- a/src/cli/list_fonts.zig
+++ b/src/cli/list_fonts.zig
@@ -4,7 +4,6 @@ const ArenaAllocator = std.heap.ArenaAllocator;
 const Action = @import("ghostty.zig").Action;
 const args = @import("args.zig");
 const font = @import("../font/main.zig");
-const discovery = @import("../font/discovery.zig");
 
 const log = std.log.scoped(.list_fonts);
 
@@ -101,18 +100,12 @@ fn runArgs(alloc_gpa: Allocator, argsIter: anytype) !u8 {
     var families: std.ArrayList([]const u8) = .empty;
     var map: std.StringHashMap(std.ArrayListUnmanaged([]const u8)) = .init(alloc);
 
-    // Look up all available fonts. The Windows backend needs a FreeType
-    // library handle so it can open candidate font files while scanning
-    // the system/user font directories.
-    var font_lib = if (comptime font.Discover == discovery.Windows)
-        try font.Library.init(alloc)
-    else {};
-    defer if (comptime font.Discover == discovery.Windows) font_lib.deinit();
-
-    var disco = if (comptime font.Discover == discovery.Windows)
-        font.Discover.init(font_lib)
-    else
-        font.Discover.init();
+    // Look up all available fonts. The library is only used by backends
+    // that need it (the Windows backend opens candidate font files with
+    // FreeType); other backends ignore it.
+    var font_lib = try font.Library.init(alloc);
+    defer font_lib.deinit();
+    var disco = font.Discover.init(font_lib);
     defer disco.deinit();
     var disco_it = try disco.discover(alloc, .{
         .family = config.family,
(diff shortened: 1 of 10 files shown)


src/font/discovery.zig
```diff
@@ -879,6 +884,239 @@ pub const CoreText = struct {
...
+        desc: Descriptor,
+        variations: []const Variation,
+        state: State,
+        dir: ?std.fs.Dir,
+        iter: ?std.fs.Dir.Iterator,
+        user_path: ?[:0]const u8,
+
+        const State = enum { system, user, done };
+
+        const system_fonts_dir = "C:\\Windows\\Fonts";
```
mattn: Will fix this hard-coding

src/cli/list_fonts.zig
```diff
@@ -100,8 +101,18 @@ fn runArgs(alloc_gpa: Allocator, argsIter: anytype) !u8 {
...
+    // the system/user font directories.
+    var font_lib = if (comptime font.Discover == discovery.Windows)
+        try font.Library.init(alloc)
+    else {};
+    defer if (comptime font.Discover == discovery.Windows) font_lib.deinit();
+
+    var disco = if (comptime font.Discover == discovery.Windows)
+        font.Discover.init(font_lib)
+    else
+        font.Discover.init();
```
mitchellh: Rather than this ball of comptime, I'd be more supportive of making all libraries take allocators and all discoveries take libraries and having them ignore it if they don't use it.

mitchellh (changes requested):
Couple questions besides the inline feedback:

1. Is this really the base way to do font discovery on Windows? Or is this a stopgap approach until DirectWrite is involved? Does DirectWrite do discovery or is it just shaping/rasterization?

2. Are there unit tests that cover discovery on Windows? If not, please add them. Surely Windows has standard fonts we can rely on to test discovery.

mattn:
Stopgap. DirectWrite would be the Windows equivalent of CoreText but landing it is a much larger change -- COM bindings for IDWriteFactory and friends, plus a decision about replacing FreeType rasterization. I'm a Vim core maintainer and have worked on gvim's DirectWrite renderer, so I'm aware of how much is involved.

The backend slots into the `options.backend` switch, so a future `.directwrite` variant can replace this one without touching callers.

I'll address the inline comment and add a unit test next.

mattn:
To be clear on direction -- would you rather we start with a DirectWrite backend instead? Happy to take that on if it's what you'd prefer; I just didn't want to assume.

mitchellh:
> To be clear on direction -- would you rather we start with a DirectWrite backend instead? Happy to take that on if it's what you'd prefer; I just didn't want to assume.

Absolutely not haha. This is a good approach and how we started macOS too. It’s a good way. We can remove it later.

mitchellh (approved):
Looks good pending CI.

mitchellh:
There are real test failures, please fix those. They probably eluded you since you're not on platforms where those compile. :)

PR #12387, opened 2026-04-23
Command: let CreateProcessW resolve the program via PATH

mattn:
Windows users often set bare command names in the Ghostty config (`command = bash`) or pass them via `-e`, matching how they would on Linux/macOS. Today that fails because `CreateProcessW` does not do program search for `lpApplicationName` on its own.

Thanks to @qwerasd205 for pointing out that passing `NULL` for `lpApplicationName` is exactly how Windows docs say to get program search for free. This PR does that: drop the explicit utf16 conversion for `lpApplicationName`, pass `null`, and make sure the program name is the first token of `lpCommandLine`. Windows then walks parent-app dir, CWD, system dirs, and PATH (and appends `.exe` for extensionless names). The child also sees its `argv[0]` exactly as we wrote it rather than a resolved absolute path, which is less surprising.

Net change is +15 / -7 in `src/Command.zig`; no new helpers, no changes outside that file. The earlier version of this PR (which added PATH/PATHEXT handling in `internal_os.path.expand`) is obsoleted by this approach and has been force-pushed away.

---

AI usage disclosure: developed with Claude Code (Claude Opus 4.7). Claude drafted the implementation based on my direction after @qwerasd205's review suggested the NULL-lpApplicationName approach. I reviewed the diff, built and verified it on the Windows GNU-ABI target, and am responsible for the code landing here.

### DIFF
--- a/src/Command.zig
+++ b/src/Command.zig
@@ -258,12 +258,20 @@ fn startPosix(self: *Command, arena: Allocator) !void {
 }
 
 fn startWindows(self: *Command, arena: Allocator) !void {
-    const application_w = try std.unicode.utf8ToUtf16LeAllocZ(arena, self.path);
     const cwd_w = if (self.cwd) |cwd| try std.unicode.utf8ToUtf16LeAllocZ(arena, cwd) else null;
-    const command_line_w = if (self.args.len > 0) b: {
-        const command_line = try windowsCreateCommandLine(arena, self.args);
-        break :b try std.unicode.utf8ToUtf16LeAllocZ(arena, command_line);
-    } else null;
+
+    // Pass null for lpApplicationName and put the program as the first
+    // token of lpCommandLine. This lets CreateProcessW perform the
+    // standard program search (parent-app dir, CWD, system dirs, PATH)
+    // and append ".exe" when the name has no extension, which is what
+    // users expect for bare commands like `wsl ~` or `pwsh.exe`.
+    // It also preserves the child's argv[0] as written by the caller
+    // rather than replacing it with the resolved absolute path.
+    const command_line = if (self.args.len > 0)
+        try windowsCreateCommandLine(arena, self.args)
+    else
+        try windowsCreateCommandLine(arena, &.{self.path});
+    const command_line_w = try std.unicode.utf8ToUtf16LeAllocZ(arena, command_line);
     const env_w = if (self.env) |env_map| try createWindowsEnvBlock(arena, env_map) else null;
 
     const any_null_fd = self.stdin == null or self.stdout == null or self.stderr == null;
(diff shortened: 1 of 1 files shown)


qwerasd205:
Why not just... don't use the `lpApplicationName` parameter in the call to `CreateProcessW` (pass `NULL` instead), and put the command name as the first element of the `lpCommandLine`, which, [per the docs](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw), results in Windows handling the path expansion for us:
> If lpApplicationName is NULL, the first white space–delimited token of the command line specifies the module name. If you are using a long file name that contains a space, use quoted strings to indicate where the file name ends and the arguments begin (see the explanation for the lpApplicationName parameter). If the file name does not contain an extension, .exe is appended. Therefore, if the file name extension is .com, this parameter must include the .com extension. If the file name ends in a period (.) with no extension, or if the file name contains a path, .exe is not appended. If the file name does not contain a directory path, the system searches for the executable file in the following sequence:
> - The directory from which the application loaded.
> - The current directory for the parent process.
> - The 32-bit Windows system directory. Use the [GetSystemDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getsystemdirectorya) function to get the path of this directory.
> - The 16-bit Windows system directory. There is no function that obtains the path of this directory, but it is searched. The name of this directory is System.
> - The Windows directory. Use the [GetWindowsDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getwindowsdirectorya) function to get the path of this directory.
> - The directories that are listed in the PATH environment variable. Note that this function does not search the per-application path specified by the App Paths registry key. To include this per-application path in the search sequence, use the [ShellExecute](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/axe/shellexecute) function.

qwerasd205:
Also, please note that per the AI usage policy, "AI assistance was used for this PR" is not a sufficient disclosure- as it says:
> You must state the tool you used (e.g. Claude Code, Cursor, Amp) along with the extent that the work was AI-assisted.

Please update your PR description(s) to state what tool was used and to what extent (how) it was used. "AI assistance" could be anything from just asking a chatbot questions about an API to telling Claude to make something work, having it iterate until the tests pass, and then blindly committing the results with little-to-no review (which would violate the policy since the human-in-the-loop *must* understand the code).

mattn:
Thanks for the pointer on both counts.

Added `Co-authored-by: Claude <noreply@anthropic.com>` trailers to every commit on the three open win32/t1-* PRs (#12386, #12387, #12389) and documented it in my project CLAUDE.md so future commits pick it up automatically. The PR descriptions have also been updated with explicit disclosure of the tool (Claude Code / Claude Opus 4.7) and the extent of AI assistance: Claude drafted the implementations based on my direction, and I reviewed each diff and validated them with Windows GNU-ABI smoke builds before pushing.

Will restructure this PR to use the NULL-`lpApplicationName` approach per your other comment in a moment.

mitchellh (approved):
Approved pending CI

PR #12390, opened 2026-04-23
core: don't force shell integration unless we're running that shell

### LINKED ISSUE #12378: Setting `shell-integration = nushell` breaks `ghostty +new-window -e` on Linux
### Discussed in https://github.com/ghostty-org/ghostty/discussions/12377

<sup>Originally posted by **MayRedwood** April 22, 2026</sup>
### Issue Description

Explicitly setting `shell-integration = nushell` breaks the behavior of the`+new-window` command when used with the `-e` flag. I assume a similar thing happens if `shell-integration` is set to other explicit values, but I haven't tested. Setting it to `detect` or `none` makes everything work as expected.

Here's the output from running `ghostty +new-window -e vim`:
image

### Expected Behavior

Running `ghostty +new-window -e vim` should open vim in a new window.

### Actual Behavior

`ghostty +new-window -e vim` fails, claiming that `--execute` is an unknown option argument. The same happens if you substitute `vim` for other commands.

### Reproduction Steps

1. Start a new instance of Ghostty with `shell-integration = nushell` in your config file.
2. Run `ghostty +new-window -e vim`.

### Ghostty Logs

```text
info: ghostty version=1.3.1
info: runtime=.gtk
info: dependency harfbuzz=12.3.0
info: dependency fontconfig=21701
info(os_locale): setlocale from env result=en_US.UTF-8
info(gtk): GTK version build=4.20.3 runtime=4.20.3
info(gtk): libadwaita version build=1.8.4 runtime=1.8.4
info: reading configuration file path=/home/moon/.config/ghostty/config
warning(gtk_ghostty_application): setting GDK_DEBUG=
warning(gtk_ghostty_application): setting GDK_DISABLE=gles-api,vulkan
warning(glib): WARNING: Adwaita: Using GtkSettings:gtk-application-prefer-dark-theme with libadwaita is unsupported. Please use AdwStyleManager:color-scheme instead.
info(opengl): loaded OpenGL 4.6
info(io_exec): found Ghostty resources dir: /nix/store/gzwsjzz2lgalvjs4dfikk9rwq0m0b27a-ghostty-1.3.1/share/ghostty
info(io_exec): shell integration automatically injected shell=.nushell
warning(gtk_ghostty_application): unimplemented action=.cell_size
info(io_exec): started subcommand path=/bin/sh pid=8722
```

### Ghostty Version

```text
Ghostty 1.3.1

Version
  - version: 1.3.1
  - channel: stable
Build Config
  - Zig version   : 0.15.2
  - app runtime   : .gtk
  - libxev        : io_uring
  - kernel version: 7.0.0
  - desktop env   : other
  - GTK version   :
    build         : 4.20.3
    runtime       : 4.20.3
  - libadwaita    : enabled
    build         : 1.8.4
    runtime       : 1.8.4
  - libX11        : enabled
  - libwayland    : enabled
```

### OS Version Information

NixOS 26.05 (Yarara)

Wayland

niri

```ini
command = nu
shell-integration = nushell
```

_No response_

### I acknowledge that:

jcollie:
If you bypassed shell detection and forced a specific shell integration with `shell-integration=<shell>` the `bash` and `nushell` integrations would break commands that were specified with `+new-window -e <command>` (or a similar mechanism) because they modify the command in shell-specific ways and the new command would fail with an "unknown flag" error or similar.

This PR fixes that by only permitting shell integration if the command matches the type of shell integration. That means that `shell-integration=nushell` only works if `argv[0]` ends with `nu`, etc. Generally this should not matter unless the shell executable was renamed for some reason and the shell detection failed.

Fixes #12378

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2792,7 +2792,10 @@ keybind: Keybinds = .{},
 ///
 ///   * `detect` - Detect the shell based on the filename.
 ///
-///   * `bash`, `elvish`, `fish`, `nushell`, `zsh` - Use this specific shell injection scheme.
+///   * `bash`, `elvish`, `fish`, `nushell`, `zsh` - Use this specific shell
+///     injection scheme if the command is this shell. Has no effect if the
+///     command is not a shell, or is a different shell from the one specified
+///     here.
 ///
 /// The default value is `detect`.
 @"shell-integration": ShellIntegration = .detect,
--- a/src/config/command.zig
+++ b/src/config/command.zig
@@ -1,4 +1,6 @@
 const std = @import("std");
+const builtin = @import("builtin");
+
 const Allocator = std.mem.Allocator;
 const ArenaAllocator = std.heap.ArenaAllocator;
 const formatterpkg = @import("formatter.zig");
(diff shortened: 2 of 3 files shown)


jparise (changes requested):
I can't look in detail right now (temporarily internet-less), but my initial impression is that this approach isn't quite right.

- The purpose of "forced" shell integration is to ignore the command-based detection, generally for cases when the command is named something else (for reasons we don't know or necessarily need to understand). With the proposed approach, we're basically just using a hinted "detect" path.
- If we don't want shell integration in the `+new-window` path, I think we should skip calling `shell_integration.setup()` or pass a somehow different command into it. (This is the part I can't research while on mobile, so sorry if I'm missing something.)

jcollie:
Closing. Will revisit if this is still a problem in the future.

PR #12403, opened 2026-04-23
apprt: minimal Win32 skeleton (window + WGL + CoreSurface wiring)

mattn:
First Tier 2 PR of the win32-apprt upstreaming series. Adds a `.win32` variant to the apprt Runtime enum, defaulting on Windows exe builds, and enough surface/renderer wiring that `zig build -Dtarget=native-native-gnu` produces a `ghostty.exe` which opens a window and runs `cmd.exe` under ConPTY. The result is a working terminal at a basic level -- you can see output and the terminal resizes with the window.

The four commits stack in a straightforward order:

1. **Add minimal Win32 application runtime** -- `src/apprt/win32/App.zig` creates the HWND and runs the message loop, `src/apprt/win32/Surface.zig` is a stub. Window opens, no rendering yet.
2. **Add WGL OpenGL context and surface interface** -- pixel format + WGL context in Surface.zig plus the core `Surface` interface methods (`getContentScale`, `getSize`, etc.) the renderer needs.
3. **Wire up CoreSurface with WGL context threading** -- `App.initCoreSurface`, release/re-acquire GL context between main thread and renderer thread, so the renderer thread can draw. ConPTY successfully launches `cmd.exe` at this point.
4. **Add `SwapBuffers` and `WM_SIZE`** -- `SwapBuffers` after `drawFrame` so rendered frames actually appear, and `WM_SIZE` forwards to `CoreSurface.sizeCallback` so the terminal resizes.

These four commits are bundled into one PR rather than four because each intermediate state would merge a non-functional `ghostty.exe` into `main` -- after commit 1 the window is blank, after commit 2 it's still blank, after commit 3 `cmd.exe` runs but produces no visible output. Only after commit 4 does the binary behave like a terminal. The commit boundaries are preserved inside this PR so the review can be read as a sequence, but the merge itself needs all four.

Intentionally **not** included (each item is its own PR in the series):

- Keyboard, mouse, clipboard, IME input
- DPI awareness, focus events, window title updates
- Font-family config (discovery works but config wiring is separate)
- Dialogs, quit-confirm, child-exit bar
- Splits, tabs, command palette, quick terminal, multi-window
- Single-instance mutex, IPC window args
- Taskbar progress, native scrollbars
- Adjacent conpty.dll load

The tracker (mattn/ghostty#1) lists all the follow-up items.

---

AI usage disclosure: developed with Claude Code (Claude Opus 4.7). The skeleton code comes from my existing win32-apprt branch; Claude assisted with cherry-picking onto upstream main, resolving a merge conflict against the already-merged DllMain change from #12373, and validating the build.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -61,6 +61,12 @@
             .hash = "gobject-0.3.0-Skun7ANLnwDvEfIpVmohcppXgOvg_I6YOJFmPIsKfXk-",
             .lazy = true,
         },
+        .win32 = .{
+            // marlersoft/zigwin32 -- Win32 API bindings for the win32 apprt.
+            .url = "git+https://github.com/marlersoft/zigwin32#ec98bb4d9eea532320a8551720a9e3ec6de64994",
+            .hash = "win32-25.0.28-preview-mX5pFWMt5QPTVIGh3r2-OpPunpcCCjApyRbA6Zn6WALH",
+            .lazy = true,
+        },
 
         // C libs
         .dcimgui = .{ .path = "./pkg/dcimgui", .lazy = true },
--- a/src/apprt.zig
+++ b/src/apprt.zig
@@ -16,6 +16,7 @@ pub const action = @import("apprt/action.zig");
 pub const ipc = @import("apprt/ipc.zig");
 pub const gtk = @import("apprt/gtk.zig");
 pub const none = @import("apprt/none.zig");
+pub const win32 = @import("apprt/win32.zig");
 pub const browser = @import("apprt/browser.zig");
 pub const embedded = @import("apprt/embedded.zig");
 pub const surface = @import("apprt/surface.zig");
@@ -43,6 +44,7 @@ pub const runtime = switch (build_config.artifact) {
     .exe => switch (build_config.app_runtime) {
         .none => none,
         .gtk => gtk,
+        .win32 => win32,
     },
     .lib => embedded,
     .wasm_module => browser,
--- a/src/apprt/runtime.zig
+++ b/src/apprt/runtime.zig
@@ -11,11 +11,17 @@ pub const Runtime = enum {
     /// approach to building the application.
     gtk,
 
+    /// Win32. Native Windows application using the Win32 API with OpenGL
+    /// rendering.
+    win32,
+
     pub fn default(target: std.Target) Runtime {
         return switch (target.os.tag) {
             // The Linux and FreeBSD default is GTK because it is a full
             // featured application.
             .linux, .freebsd => .gtk,
+            // Windows uses the native Win32 API.
+            .windows => .win32,
             // Otherwise, we do NONE so we don't create an exe and we create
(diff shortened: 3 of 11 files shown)


marler8997:
Here's my feedback.  I'm not a ghostty maintainer but I have alot of experience writing code for Windows.

It looks like there's alot of return values being ignored.  IMO you should almost never ignore a return value when calling a win32 function.  In addition, when an error is being handled, the error should be retrieved via GetLastError to see what went wrong, probably logged somewhere.

You're also free to leverage "zigwin32" for all these type/function definitions from the win32 API.  It's found at https://github.com/marlersoft/zigwin32, I'd add this as a dependency and here's how I'd recommend using it:

```zig
const win32 = @import("win32").everything;

const my_hello_string= win32.L("Hello");
const hwnd = win32.CreateWindowExW(...) orelse win32.panicWin32("CreateWindow", win32.GetLastError());
```

Since the win32 API is in C, every symbol pretty much has to be unique so we can expose it all in a single namespace.

For many functions in win32, if they fail, something has gone horribly wrong.  For these kinds of functions you can simply call `win32.panicWin32("FunctionName", win32.GetLastError())`.  I've also included some wrapper functions that will call this for you, giving you convenient builtin error handling with proper error reporting, i.e.

```zig
const dpi = win32.dpiFromHwnd(hwnd);
const client_size = win32.getClientSize(hwnd);
_, const ps = win32.beginPaint(hwnd);
defer win32.endPaint(hwnd, &ps);
```

mattn:
Thanks! Addressed the ignored-return issue in 289c6a57. Added `GetLastError`-based warning logs for the Win32 calls where failure is worth surfacing (`PostMessageW`, `DestroyWindow`, `SwapBuffers`, `wglMakeCurrent`). The remaining unchecked returns (`TranslateMessage`, `DispatchMessageW`, `ShowWindow`, etc.) don't indicate errors per the docs, so I left them as-is.

Re: `zigwin32` -- good suggestion; I'll consider it for a follow-up once the maintainer has a chance to weigh in on adding the dependency.

jcollie:
Re: `zigwin32`  - probably a good idea, especially as Zig 0.16 is dropping a lot of stuff from `std.os.windows`
https://ziglang.org/download/0.16.0/release-notes.html#posix-and-oswindows-removals

mattn:
Ported the apprt to zigwin32 in 2ad00dc6. Dropped the hand-rolled extern decls in App.zig and Surface.zig and now go through `@import("win32").everything`. The binding is lazy-fetched in build.zig.zon so it's only pulled in for win32 apprt builds. Net -240 / +101, and style flags are now type-safe packed structs instead of OR'd u32 constants. Thanks for the push on this -- it also gets us off `std.os.windows` symbols that Zig 0.16 is removing.

mattn:
Could you continue the review when you have a chance?

jcollie:
Going to close this for now as the Ghostty project is still not prepared at this time to support a Windows port. Please feel free to maintain this in a fork and submit any PRs that would be relevant for libghostty.

PR #12415, opened 2026-04-24
config: use Config to check key binding instead of App

bo2themax:
Previously `ghostty_app_key_is_binding` (unlike Surface) is just using `config.keybind` to check whether a KeyEvent is in the set or not.

After this, I can add unit tests for keybinding more easily with dummy configs.

I didn't find any usages of this in GTK, so it shouldn't affect anything. ci will see if this is the case:)


mitchellh (approved):
Yeah this is fine and achieves the same thing. The purpose is that its possible for the app and surfaces to have their own separate configurations at any given moment. What you switched out in the macOS app though achieves the same thing so this isfine.

PR #12419, opened 2026-04-24
macOS: save window state before relaunching by Sparkle

bo2themax:
Fixes #11304

https://github.com/user-attachments/assets/1dfd5035-d872-4726-b31e-bb38181aff33

### DIFF
--- a/macos/Sources/Features/Update/UpdateDelegate.swift
+++ b/macos/Sources/Features/Update/UpdateDelegate.swift
@@ -36,7 +36,12 @@ extension UpdateDriver: SPUUpdaterDelegate {
         // When the updater is relaunching the application we want to get macOS
         // to invalidate and re-encode all of our restorable state so that when
         // we relaunch it uses it.
+
+        // Invalidate quick terminal's state
         NSApp.invalidateRestorableState()
-        for window in NSApp.windows { window.invalidateRestorableState() }
+        // Invalidate all other terminals' state
+        TerminalController.all.forEach {
+            $0.window?.invalidateRestorableState()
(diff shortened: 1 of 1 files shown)


macos/Sources/Features/Update/UpdateDelegate.swift
```diff
@@ -36,7 +36,12 @@ extension UpdateDriver: SPUUpdaterDelegate {
         // When the updater is relaunching the application we want to get macOS
         // to invalidate and re-encode all of our restorable state so that when
         // we relaunch it uses it.
+
+        // Invalidate quick terminal's state
```
mitchellh, line 40: This comment seems odd?

mitchellh (changes requested):
Why does this work? Why is it different than before?

Are you sure its working? Or are you just getting lucky? The issue I ran into before is that state save is async and there's no way to know it happened.

bo2themax:
> Why does this work? Why is it different than before?
> 
> Are you sure its working? Or are you just getting lucky? The issue I ran into before is that state save is async and there's no way to know it happened.

Lucky I guess, I tried calling hide to trigger, which can only save some of the windows, missing tabs most of the time.

And I saw "responder’s interface-related" [here](https://developer.apple.com/documentation/appkit/nsresponder/invalidaterestorablestate()), so I tried this, works great, I tried it 5 times with local appcast.xml. 

To be honestly I don't know exactly why🫪

bo2themax:
I thought of a theory on the tram lol, will test it later

- [x] Test invalidateRestorableState

PR #12426, opened 2026-04-25
gtk/SurfaceScrolledWindow: wrap root child with another Adw.Bin

alaviss:
Due to a known Gtk issue, the scrolled_window at the root of the template is free-ed twice on dispose. This causes crashes when used with GNOME 49 platform (Gtk 4.20, libadwaita 1.8.5).

Workaround this issue by wrapping the root child in another Adw.Bin, similar to widgets like ResizeOverlay.

LLM was used to perform discovery against a manually recorded Valgrind trace, and helped tracking down known fixes for this problem. The comment in code was taken from another instance in the repository.

Fixes https://github.com/ghostty-org/ghostty/discussions/12306

Assisted-by: OpenAI GPT-5.4

### DIFF
--- a/src/apprt/gtk/class/surface_scrolled_window.zig
+++ b/src/apprt/gtk/class/surface_scrolled_window.zig
@@ -163,8 +163,7 @@ pub const SurfaceScrolledWindow = extern struct {
         _: ?*anyopaque,
     ) callconv(.c) void {
         const priv = self.private();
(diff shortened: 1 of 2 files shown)


mitchellh (approved):
Oh yeah, surprised we didn't catch this earlier since we have to do this all over. Thanks.

PR #12427, opened 2026-04-25
flatpak: terminate session if Ghostty disconnects from bus

alaviss:
This makes sure that if Ghostty crashes, commands spawned are also terminated automatically by the Flatpak Session Helper.

The few crashes I got left a lot of background processes, some of them pretty heavy and took awhile to be figured out.

### DIFF
--- a/src/os/flatpak.zig
+++ b/src/os/flatpak.zig
@@ -41,7 +41,7 @@ pub const FlatpakHostCommand = struct {
         clear_env: bool = false,
         /// Kill the sandbox when the caller disappears from the session bus
         watch_bus: bool = false,
-        _reserved: u30 = 0,
+        _reserved: std.meta.Int(.unsigned, @bitSizeOf(c_uint) - 2) = 0,
     };
 
     /// Argv are the arguments to call on the host with argv[0] being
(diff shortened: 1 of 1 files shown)


src/os/flatpak.zig
```diff
@@ -33,6 +33,16 @@ pub const FlatpakHostCommand = struct {
...
     });
+    /// Flags for HostCommand method
+    ///
+    /// Ref: https://docs.flatpak.org/en/latest/libflatpak-api-reference.html#gdbus-method-org-freedesktop-Flatpak-Development.HostCommand
+    const Flags = packed struct(c_uint) {
+        /// Clear the environment
+        clear_env: bool = false,
+        /// Kill the sandbox when the caller disappears from the session bus
+        watch_bus: bool = false,
+        _reserved: u30 = 0,
```
pluiedev: This only works when `c_uint` is a `u32`, instead use `std.meta.Int(.unsigned, @bitSizeOf(c_uint) - 2)`

pluiedev (approved):
Looks good except for one detail

mitchellh (changes requested):
Just marking red until Pluie's feedback is addressed. Otherwise looks good.

pluiedev (approved):

PR #12435, opened 2026-04-25
surface: respect semantic prompt boundaries for links

ZuBB:
Link detection currently expands the clicked location to a full line before running the configured regexes. When semantic prompt markers are present, this can cause prompt text and neighboring content to be matched together even though they are distinct semantic regions.

Use semantic prompt boundaries when selecting the text to inspect for link matching. This keeps prompt text separate from the content beside it and avoids folding prompt text into double-click link/path selection.

Add a regression test that models a prompt and command on the same line and verifies the prompt region and input region remain separate.

----

this is a fix for the issue users reported in https://github.com/ghostty-org/ghostty/discussions/11415

**disclaimer**: I used codex addon within Cursor to research this bug/regression and find a proper fix for it.


mitchellh:
!vouch

src/Surface.zig
```diff
@@ -6658,3 +6660,60 @@ test "Surface: rectangle selection logic" {
         true, //rectangle selection
     );
 }
+
+test "Surface: link selection line respects prompt boundary" {
```
mitchellh: You can remove this test, `semantic_prompt_boundary` is well tested and this test isn't even in the rright place.

mitchellh (changes requested):

src/Surface.zig
ZuBB: okay, will do that

ZuBB: done in [c4a671b](https://github.com/ghostty-org/ghostty/pull/12435/commits/c4a671ba5a5c3a896b90b066f2a22abe2eaad17b)

mitchellh (approved):
Approved pending CI. Thank you.

PR #12443, opened 2026-04-25
Sync middle-click paste with copy-on-select

007hacky007:
Implements the behavior from #9788.

Today, middle-click paste always reads from the selection clipboard (or the
system clipboard on platforms without a selection clipboard). With this change
it follows `copy-on-select`:

- `true`: selection clipboard (unchanged)
- `clipboard`: system clipboard
- `false`: selection clipboard (also unchanged, preserves traditional X11
  middle-click behavior)

The idea is symmetry: if `copy-on-select = clipboard` writes selections to the
system clipboard, middle-click should come back from there too.

Also updated the config doc comment, which previously claimed middle-click
"will always use the selection clipboard".

### Testing

`zig build test` passes locally (macOS, Zig 0.15.2).

Built and runtime-tested via the fork's CI: https://github.com/007hacky007/ghostty/actions/runs/24707475544 - I'm running the resulting binary daily and the three `copy-on-select` modes behave as described above.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4007,12 +4007,19 @@ pub fn mouseButtonCallback(
         }
     }
 
-    // Middle-click pastes from our selection clipboard
+    // Middle-click paste source follows copy-on-select: when copy-on-select
+    // targets the selection clipboard, middle-click reads from it; when
+    // copy-on-select targets the system clipboard, middle-click reads from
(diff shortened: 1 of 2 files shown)


mitchellh (approved):
Approved pending CI, milestoned to 1.4 since this is technically breaking.

PR #12447, opened 2026-04-25
Fix Korean IME committed text handling for arrow keys

dobbylee:
Fixes #11461

- Send Korean IME committed text as a separate text-only key event when arrow-key handling commits preedit text.
- Replay arrow navigation after the committed text is sent. Do not replay plain Left Arrow to match Terminal.app behavior.
- Manually tested Arrow keys and Opt/Cmd+Arrow during Korean preedit on macOS.


mitchellh (approved):
This is pretty hacky but willing to accept it because its VERY tightly scoped and the conditional is based on unlikely factors so its unlikely to affect input latency. Approved pending CI.

PR #12462, opened 2026-04-25
macOS: check `ghostty_surface_t` before using

bo2themax:
There was a bad access when I tried to restore window state from 1.2.3, which only happens after `SurfaceView` is successfully decoded, but the `TerminalRestorationState` is failed to be decoded.

This should be an issue after #12461, but I thought having this check would also prevent accessing the de-allocating surface in other cases.


bo2themax:
Still have the risk of data race in `deinit`, closing it for now

PR #12465, opened 2026-04-25
os: add randomTmpPath for allocating temp paths

jparise:
Factor TempDir's name generation into a reusable `randomBasename` (16 random bytes, url-safe base64) and add `randomTmpPath` on top, which composes `allocTmpDir` + `randomBasename` into a single allocated path in the form `{TMPDIR}/{prefix}{random}` (`mktemp(1)`-ish).

This is convenient for callers who want a unique path under TMPDIR (for a temporary file, socket, etc.) without having to think about basename buffer sizing or path joining.

Also, use `std.base64.url_safe_no_pad.Encoder` instead of the custom base64 alphabet, which is exactly equivalent.

### DIFF
--- a/src/os/file.zig
+++ b/src/os/file.zig
@@ -79,3 +79,58 @@ pub fn freeTmpDir(allocator: std.mem.Allocator, dir: []const u8) void {
         allocator.free(dir);
     }
 }
+
+const random_basename_bytes = 16;
+const b64_encoder = std.base64.url_safe_no_pad.Encoder;
+
+pub const RandomBasenameError = error{BufferTooSmall};
+
+/// Length of the basename produced by `randomBasename`.
+pub const RANDOM_BASENAME_LEN = b64_encoder.calcSize(random_basename_bytes);
+
+/// Write a random filesystem-safe base64 basename of length
+/// `RANDOM_BASENAME_LEN` into `buf` and return a slice over the
+/// written bytes. Returns `error.BufferTooSmall` if `buf` is too
+/// short.
+pub fn randomBasename(buf: []u8) RandomBasenameError![]const u8 {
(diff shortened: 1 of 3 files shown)


src/os/file.zig
```diff
@@ -79,3 +79,58 @@ pub fn freeTmpDir(allocator: std.mem.Allocator, dir: []const u8) void {
...
     }
 }
+
+const random_basename_bytes = 16;
+const b64_encoder = std.base64.url_safe_no_pad.Encoder;
+
+pub const RandomBasenameError = error{BufferTooSmall};
+
+/// Length of the basename produced by `randomBasename`.
+pub const RANDOM_BASENAME_LEN = b64_encoder.calcSize(random_basename_bytes);
```
jcollie, line 89: @qwerasd205 is gonna love this.

jcollie (approved):

mitchellh (approved):
Approve but if you can change that const into a normal Zig const in a follow up that'd be great.

jparise:
> Approve but if you can change that const into a normal Zig const in a follow up that'd be great.

Which one do you mean?

mitchellh:
The one @jcollie pointed out, we're preferring more like `fooBarBaz` consts not C-style `FOO_BAR_BAZ`

jparise:
> The one @jcollie pointed out, we're preferring more like `fooBarBaz` consts not C-style `FOO_BAR_BAZ`

Ah, got it. I thought that style was just meant for types. And it looks like Zig stdlib seems to prefer `lower_case` for public scalar constants now.

mitchellh:
> And it looks like Zig stdlib seems to prefer lower_case for scalar constants now.

Sorry yes you're right. This.

PR #12468, opened 2026-04-25
macos: expose tab color via AppleScript

thoutbeckers:
Exposes tab color on macos as a readable/writeable AppleScript `color` property.

Based on Discussions requests:

https://github.com/ghostty-org/ghostty/discussions/11590 (just name was implemented)
https://github.com/ghostty-org/ghostty/discussions/12314 (more recent request)

Claude code was used to generate initial .sdef binding, but I spent some time myself making it (hopefully) more maintainable, including some compile time garantuees if updates the the palette are made in future.

### DIFF
--- a/macos/Ghostty.sdef
+++ b/macos/Ghostty.sdef
@@ -69,6 +69,9 @@
       </property>
       <property name="index" code="pidx" type="integer" access="r" description="1-based index of this tab in its window."/>
       <property name="selected" code="GTsl" type="boolean" access="r" description="Whether this tab is selected in its window."/>
+      <property name="color" code="GTcl" type="tab color" access="rw" description="The color of the tab.">
+        <cocoa key="color"/>
+      </property>
       <property name="focused terminal" code="GTfT" type="terminal" access="r" description="The currently focused terminal surface in this tab.">
         <cocoa key="focusedTerminal"/>
       </property>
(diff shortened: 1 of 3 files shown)


macos/Ghostty.sdef
```diff
@@ -69,6 +69,9 @@
       </property>
       <property name="index" code="pidx" type="integer" access="r" description="1-based index of this tab in its window."/>
       <property name="selected" code="GTsl" type="boolean" access="r" description="Whether this tab is selected in its window."/>
+      <property name="color" code="GTcl" type="tab color" access="rw" description="The color of the tab.">
```
bo2themax: I suggest we use `colr` here from [standard terminology table](https://developer.apple.com/library/archive/releasenotes/AppleScript/ASTerminology_AppleEventCodes/TermsAndCodes.html#//apple_ref/doc/uid/TP40004532)

macos/Sources/Features/AppleScript/ScriptTab.swift
```diff
@@ -184,3 +201,27 @@ extension ScriptTab {
...
+        case .none:     "GTcN".fourCharCode
+        case .blue:     "GTcB".fourCharCode
+        case .purple:   "GTcP".fourCharCode
+        case .pink:     "GTcK".fourCharCode
+        case .red:      "GTcR".fourCharCode
+        case .orange:   "GTcO".fourCharCode
+        case .yellow:   "GTcY".fourCharCode
+        case .green:    "GTcG".fourCharCode
+        case .teal:     "GTcT".fourCharCode
+        case .graphite: "GTcX".fourCharCode
```
bo2themax: And can we use `RGBA` here? So new colors can be easily added without worrying conflicts

bo2themax (changes requested):

mitchellh (changes requested):
I agree we should use RGBA.

bo2themax (approved):

thoutbeckers:
I see the first review is already in, thanks.

RGB codes are not actually stable between macos releases, but this way I think collisions won't happen and adding new colors would be fairly easy.

Not allowing arbitrary colors was already the intend I believe.

thoutbeckers:
I rebased this against `main`, reran tests, everything still seems to work in my local copy.

bo2themax (approved):

PR #12483, opened 2026-04-26
Fix speedy high-resolution scrolling on Linux

rightaditya:
Enforcing an absolute minimum of 1 for scroll events causes differing scroll speeds between high-resolution and standard scroll wheels on Linux. Since this was added to handle MacOS's precision scrolling emulation, this patch alters the behaviour so that the absolute minimum is only enforced on MacOS.

NB: This can't just be fixed by adjusting `mouse-scroll-multiplier` since that affects *all* scroll events whether they're high-resolution or not. Reducing `mouse-scroll-multiplier` to handle high-res scroll events better makes scrolling unusably slow for regular scroll wheels connected to the same machine.

Fixes #11648.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -3438,8 +3438,7 @@ pub fn scrollCallback(
                     @min(yoff, -1);
 
                 break :yoff_adjusted yoff_max * cell_size * self.config.mouse_scroll_multiplier.discrete;
-            }
-            else {
+            } else {
                 break :yoff_adjusted yoff * cell_size * self.config.mouse_scroll_multiplier.discrete;
             }
         };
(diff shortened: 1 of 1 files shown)


mitchellh:
Can you run zig fmt in here to fix that up?

Other than that I'm fine giving this a shot.

mitchellh:
Thanks, also, I know zig fmt is easy to run, but I do a lot of PR review on mobile so... I can't actually do it lol.

rightaditya:
@mitchellh My bad. Done and rebased.

rightaditya:
And thanks, likewise. I don't have a Mac so I couldn't test it there, but the change seemed innocuous enough to at least be worth trying.

PR #12488, opened 2026-04-27
update zon2nix to 0.5.0

jcollie:
fix hash outputs for flatpak


mitchellh:
Are these real CI failures?

jcollie:
> Are these real CI failures?

Yeah, I think that they are related to the nixpkgs update. I won't be able to diagnose until this evening though.

PR #12522, opened 2026-04-29
macOS: restore some menu shortcuts when terminal is not focused

bo2themax:
This fixes #10749.

After a few iterations for the past few weeks, this is the minimal change I can think of to fix this issue without introducing changes from the core layer.

> As the issue says, the rc here is that we strip the shortcuts from the menu in order to prevent menu bar handling some of the key equivalents. I did try to revert this behavior, but that would involve a lot of regression testing (which I did). I'm still not 99.999% sure that would not breaking anything else, so those branches will stay in my local branch for now.

1. Add back the shortcuts to `copy`, `paste`, `select all`, `undo` and `redo`
2. Restore them when ever the first responder changes or the key window changes, so the menu can handle them again in native views. See the comments in `restoreRegisteredMenusIfNeeded` for all the native views i found that had this issue before.
3. Add back `cut` menu and disable whenever a terminal surface is focused.

https://github.com/user-attachments/assets/db6fddc6-0664-43f8-baa1-925c5eaac791

## AI Disclosure

Claude did some review change refactoring


bo2themax:
- [x] There should be some optimization around re-sync, right now it doesn't track ghostty actions for restorable items 

- [x] Also check whether reloading config would affect anything

macos/Sources/Ghostty/Ghostty.Config.swift
```diff
@@ -116,8 +116,8 @@ extension Ghostty {
         /// configuration would be "quit" action.
         ///
         /// Returns nil if there is no key equivalent for the given action.
-        func keyboardShortcut(for action: String) -> KeyboardShortcut? {
-            guard let cfg = self.config else { return nil }
+        func keyboardShortcut(for action: String?) -> KeyboardShortcut? {
```
mitchellh: Seems weird and wrong to change this function signature.

mitchellh (changes requested):
Conceptually this is totally fine, I came upon the same solution. Stylistically I think this needs to be cleaned up a bit. I feel like we could better organize some of this.

bo2themax:
> Conceptually this is totally fine, I came upon the same solution. Stylistically I think this needs to be cleaned up a bit. I feel like we could better organize some of this.

1cbf7287d1c1d5bcf3dbe1d2cf268a9c1990e069 changed the manual signal to one-time setup for first responder on key window. I hope that's what you meant, I don't want to over engineer the changes in MenuShortcutManager

KristijanKocev:
I've been running this the last few days and it has worked well

PR #12534, opened 2026-04-30
Fall back to Zig-bundled Darwin headers when an SDK can't be found

Samasaur1:
Currently, cross to Darwin uses the Darwin headers bundled with Zig. However, if you're running a build _on_ Darwin, an error is thrown if the SDK can't be found, even though the bundled headers are still available.

Now, we continue to search for and prefer the installed SDK, but if it can't be found, we fall back to the bundled headers rather than failing the build.

### DIFF
--- a/pkg/apple-sdk/build.zig
+++ b/pkg/apple-sdk/build.zig
@@ -49,15 +49,15 @@ pub fn addPaths(
     });
 
     if (!gop.found_existing) init: {
-        if (comptime builtin.os.tag.isDarwin()) {
+        if (comptime builtin.os.tag.isDarwin()) darwin: {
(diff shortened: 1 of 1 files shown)


bo2themax:
What's the actual errors are you getting, im curious.

mitchellh:
This looks right. Note @bo2themax this has been a build issue for a couple projects in #libghostty on Discord for awhile. 

I wonder if we can update our CI to also run this verification, we have macOS runners, maybe we should do a `nix build` there and get our libghostty packages building properly on CI...

mitchellh (approved):
Note my other comment but dont' want to block on that since this affects real projects. Pending CI green though.

mitchellh:
The CI failures were irrelevent.

PR #12547, opened 2026-05-01
macos: avoid replaying keys that commit preedit

knu:
Refs #10460
Related: #12518

When an input method commits all or part of marked text during keyDown, AppKit returns the committed text through insertText. Treat that as text committed by the input method instead of replaying the original key event to the terminal.

Previously this path only handled arrow-key commits specially. A control-key shortcut that commits preedit text could still be encoded as the original control input after composition, such as ctrl+j becoming LF.

Send committed preedit text as a text-only event for any key that causes the commit. Only replay arrow navigation keys that the existing Korean IME handling expects, and keep plain left-arrow suppressed because AppKit already leaves the caret in place.

Before:
before

After:
after

AI usage: OpenAI Codex helped investigate, implement, test, and refine this change. I reviewed and tested the resulting code.


mitchellh (approved):
Looks good pending CI!

PR #12548, opened 2026-05-01
libghostty-vt: support building on macOS with Nix

sandydoo:
Adds support for building libghostty-vt on macOS with Nix.

Tested on aarch64-darwin. Tests pass as well.

_Claude used to speed up debugging process. All comments, commit messages, and final code authored by me._

### DIFF
--- a/nix/libghostty-vt.nix
+++ b/nix/libghostty-vt.nix
@@ -37,7 +37,23 @@ stdenv.mkDerivation (finalAttrs: {
     );
   };
 
-  deps = callPackage ../build.zig.zon.nix {name = "${finalAttrs.pname}-cache-${finalAttrs.version}";};
+  # Zig's build runner computes relative paths from `cwd` to the build directory.
+  # The logic is purely lexical, so if the `cwd` is a symlink that resolves to a different depth during `chdir`, the computed path becomes incorrect.
+  #
+  # See: https://codeberg.org/ziglang/zig/issues/32121
+  #
+  # Workaround: override `linkFarm` with a copy-farm so deps are real directories, not symlinks.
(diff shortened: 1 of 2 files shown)


nix/libghostty-vt.nix
```diff
@@ -37,29 +37,55 @@ stdenv.mkDerivation (finalAttrs: {
...
+    name = "${finalAttrs.pname}-cache-${finalAttrs.version}";
+    linkFarm = name: entries:
+      runCommand name {} ''
+        mkdir -p $out
+        ${lib.concatMapStringsSep "\n" (e: ''
+          cp -rL ${e.path} $out/${e.name}
+        '')
+        entries}
+      '';
+  };
```
sandydoo, line 56: This is the only weird bit that we can hopefully remove at some point.

There's a bug in Zig's build runner where it does same path math without taking into account symlinks. I've written it up here: https://codeberg.org/ziglang/zig/issues/32121

The workaround is to materialize the dependencies without symlinks. The dependency list is quite small here, so the impact is minimal.
Looked into materializing just the uucode dependency, but it's not worth it.

Another option would be to replace [this `setCwd` call in `uucode`](https://github.com/jacobsandlund/uucode/blob/2826a37a4562284fdacd8fa029d49509cc9bffcd/build.zig#L323) and pass the path to the source directory as an argument. That sidesteps the Zig issue.

sandydoo:
Reformatted with alejandra.

nix/libghostty-vt.nix
mitchellh: This issue makes sense and the workaround is fine for now.

PR #12560, opened 2026-05-02
Hint cold scroll back pages to allow OS reclaim under pressure

agoodkind:
## What

Ghostty holds all scrollback in physical RAM forever, with no way for the OS to reclaim it. This PR tells the OS that pages older than the last 10,000 rows are safe to evict under memory pressure. The data is never lost; the OS pages it back in transparently if the user scrolls back to it.

## Why

Without hints, every scrollback page stays physically resident forever. A long-running session with long scrollback limit steadily grows RSS at roughly 2.4 MB per 1,000 rows, with no ceiling. At 13 million rows the process charges ~63 GB against the task's physical footprint and the OS has no signal that any of it is safe to evict.

## How

Each time `grow()` adds a row, `hintColdBoundary()` checks whether a page just crossed outside the 10,000-row hot window and, if so, issues a single `madvise` call on it. A cached pointer means the syscall only fires when the boundary actually advances, so the steady-state cost per row is negligible.

On macOS, `MADV_FREE_REUSABLE` is used instead of `MADV_FREE` because it immediately decrements `task phys_footprint` rather than waiting for the kernel to lazily account for it. That is the number Activity Monitor and the OOM killer use.

## Testing

A new `ScrollbackMemory` benchmark (`ghostty-bench +scrollback-memory`) covers fill, hint, reclaim, and scroll passes so the effect can be measured directly.

I also manually measured against a 13-million-row flood on macOS (MacBook Pro 14" M5 Max 128GB):

| | charged (`phys_footprint`) | freeable (reusable pages in RAM) |
|---|---|---|
| **without hints** | 63 GB | 0 GB |
| **with hints** | 4 GB | ~46 GB |

The freeable pages are still physically resident and readable at near-zero cost when warm. The OS only evicts them when it actually needs the memory.

## Backward compatibility

- No behavior changes for normal usage
- Scrollback contents are never discarded and only their physical backing is made evictable
- Change is entirely transparent to the rest of the terminal stack.

### DIFF
--- a/src/benchmark/ScrollbackMemory.zig
+++ b/src/benchmark/ScrollbackMemory.zig
@@ -1,0 +1,346 @@
+//! Measures the effect of cold-hints on physical memory (RSS/footprint)
+//! for a PageList with deep scrollback across six phases:
+//!
+//!   1. fill:       grow total-rows rows; RSS after all pages faulted in
+//!   2. hint:       apply production lazy hints (MADV_COLD/FREE); RSS unchanged
+//!   3. reclaim:    force-reclaim cold pages (MADV_FREE_REUSABLE/DONTNEED)
+//!                  and show the immediate RSS drop
+//!   4. seq-scroll: scroll-passes sequential passes through all cold pages;
+//!                  pass 1 is cold (page faults), passes 2-N are warm (in RAM)
+//!   5. re-reclaim: reclaim again to reset state between scroll scenarios
+//!   6. rnd-scroll: scroll-passes random-order passes through all cold pages;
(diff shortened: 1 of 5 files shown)


jcollie:
Note that it may be a while before someone with experience with the relevant code can take a look.

agoodkind:
I am closing this PR because I found it actually introduces a regression with data loss 

>Scrollback contents are never discarded and only their physical backing is made evictable

I was incorrect with this statement. I will open a new PR once I figure out a better way to do this.

PR #12579, opened 2026-05-04
docs(input): add documentation for missing action parameters

bo2themax:
### AI Disclosure

Claude generated all the commits, I reviewed it and created this PR

### DIFF
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -423,6 +423,8 @@ pub const Action = union(enum) {
 
     /// Navigate the search results. If there is no active search, this
     /// is not performed.
+    ///
+    /// Valid arguments: `previous`, `next`.
     navigate_search: NavigateSearch,
 
     /// Start a search if it isn't started already. This doesn't set any
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -348,7 +348,7 @@ pub const Action = union(enum) {
 
     /// Copy the selected text to the clipboard.
     ///
-    /// Valid arguments:
+    /// Valid values:
     ///
     ///   - `plain`
     ///
@@ -370,8 +370,6 @@ pub const Action = union(enum) {
     ///     (e.g. plain text and HTML), each tagged with its content type
     ///     so the receiving OS or application can pick the most appropriate
     ///     representation when pasting.
-    ///
-    /// If no argument is given, defaults to `mixed`.
     copy_to_clipboard: CopyToClipboard,
 
     /// Paste the contents of the default clipboard.
@@ -424,7 +422,7 @@ pub const Action = union(enum) {
     /// Navigate the search results. If there is no active search, this
     /// is not performed.
     ///
-    /// Valid arguments: `previous`, `next`.
+    /// Valid values: `previous`, `next`.
     navigate_search: NavigateSearch,
 
     /// Start a search if it isn't started already. This doesn't set any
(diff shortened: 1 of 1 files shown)


src/input/Binding.zig
```diff
@@ -347,6 +347,31 @@ pub const Action = union(enum) {
...
+    ///     HTML markup.
+    ///
+    ///   - `mixed` (default)
+    ///
+    ///     Place multiple representations on the clipboard at once
+    ///     (e.g. plain text and HTML), each tagged with its content type
+    ///     so the receiving OS or application can pick the most appropriate
+    ///     representation when pasting.
+    ///
+    /// If no argument is given, defaults to `mixed`.
```
00-kat: The bullet point for mixed already notes it to be the default.

```suggestion
```

src/input/Binding.zig
```diff
@@ -684,7 +711,21 @@ pub const Action = union(enum) {
...
+    ///
+    ///   - `other`
+    ///
+    ///     Close every tab in the current window except the current tab.
+    ///
+    ///   - `right`
+    ///
+    ///     Close every tab to the right of the current tab.
+    ///
+    /// If no argument is given, defaults to `this`.
```
00-kat: Ditto.

```suggestion
```

src/input/Binding.zig
```diff
@@ -684,7 +711,21 @@ pub const Action = union(enum) {
     /// Close the current tab and all splits therein, close all other tabs, or
     /// close every tab to the right of the current one depending on the mode.
```
00-kat: This now simply restates what the bullet points state more clearly.

```suggestion
    /// Close the specified tabs and all splits therein.
```

src/input/Binding.zig
```diff
@@ -347,6 +347,31 @@ pub const Action = union(enum) {
     reset,
 
     /// Copy the selected text to the clipboard.
+    ///
+    /// Valid arguments:
```
00-kat: The configuration reference usually uses “values”;

```suggestion
    /// Valid values:
```

src/input/Binding.zig
```diff
@@ -684,7 +711,21 @@ pub const Action = union(enum) {
     /// Close the current tab and all splits therein, close all other tabs, or
     /// close every tab to the right of the current one depending on the mode.
     ///
-    /// If the mode is not specified, defaults to closing the current tab.
+    /// Valid arguments:
```
00-kat: Ditto.

```suggestion
    /// Valid values:
```

00-kat (changes requested):

jcollie (approved):
LGTM once @00-kat 's points are addressed.

src/input/Binding.zig
bo2themax: I think they meant different, one is for the default value when there is no config at all, one is for the default value when not specifying, although they should be the same most of the time

00-kat: > should be the same most of the time

Those two cases are always handled identically:

> ```ini
> # Empty values reset the configuration to the default value
> font-family =
> ```

— [Configuration § Syntax](https://ghostty.org/docs/config#syntax)

src/input/Binding.zig
```diff
@@ -398,6 +423,8 @@ pub const Action = union(enum) {
 
     /// Navigate the search results. If there is no active search, this
     /// is not performed.
+    ///
+    /// Valid arguments: `previous`, `next`.
```
bo2themax:
```suggestion
    /// Valid values: `previous`, `next`.
```

00-kat (approved):

PR #12581, opened 2026-05-04
Fix zero-width grapheme attachment during pending wrap

noib3:
This PR fixes an issue where a zero-width combining mark could attach to the wrong cell when the preceding character was written in the final column and the cursor had a pending wrap.

The test I added used to fail before the fix, but it passes now.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -573,20 +573,25 @@ pub fn print(self: *Terminal, c: u21) !void {
         // it.
         if (self.modes.get(.grapheme_cluster)) return;
 
-        // If we're at cell zero, then this is malformed data and we don't
-        // print anything or even store this. Zero-width characters are ALWAYS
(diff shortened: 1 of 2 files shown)


mitchellh (changes requested):
This makes sense, we should add a test to Terminal.zig too though that's the more appropriate place for this since the C tests are only run conditionally. You can keep that C test too though.

noib3:
Added an equivalent test to Terminal.zig.

Not sure what the policy is wrt to test redundancy. I can remove the test in `c/terminal.zig` since that code path is now already covered.

mitchellh (approved):
Approved pending CI.

General note: `print` is one of our most performance sensitive pieces of code, but this is specifically in the `width == 0` branch which is not common, so adding more logic here is fine.

> Not sure what the policy is wrt to test redundancy. I can remove the test in c/terminal.zig since that code path is now already covered.

Its fine, I don't mind duplicate tests that test at different layers.

PR #12596, opened 2026-05-06
fix: apply variation selectors to preceding codepoint

noib3:
This fixes a bug where the variation selectors (VS15 & VS16) were checked against the first codepoint in a cell instead of the previous codepoint in the cell's grapheme cluster, causing them to be dropped if that first codepoint was not a valid base.


mitchellh (approved):
This makes sense! Thanks for the test. Approved pending CI.

PR #12597, opened 2026-05-06
macos: add menu key equivalent option

knu:
Refs #12293
Related: #3493 (duplicates: #4374, #5091), #7013

Add `macos-menu-key-equivalents`, a macOS-specific setting for disabling menu key equivalents before they can intercept key events.  It defaults to true, preserving existing menu behavior unless explicitly disabled.

On macOS, AppKit evaluates menu key equivalents before Ghostty receives key events.  This means a remap such as `key-remap = command=left_alt` still leaves `Command+N` captured by the New Window menu item before Ghostty can turn it into `Alt+N` for keybind and terminal input handling.

Setting `macos-menu-key-equivalents` to false clears those menu shortcuts, so the key event reaches Ghostty instead of being consumed by AppKit.  That allows Ghostty to apply `key-remap` first and dispatch the remapped key as terminal input or a Ghostty keybind.

When disabled, this clears Ghostty-managed menu bindings and known standard macOS menu items, including dynamic items as their menus are opened.  The original key equivalents are restored when the option is enabled again.

Document the option and add Swift and Zig config coverage for it.

AI usage: OpenAI Codex helped investigate, implement, test, and refine this change.  I reviewed and tested the resulting code.

### DIFF
--- a/macos/Sources/App/macOS/AppDelegate.swift
+++ b/macos/Sources/App/macOS/AppDelegate.swift
@@ -268,6 +268,11 @@ class AppDelegate: NSObject,
             selector: #selector(ghosttyNewTab(_:)),
             name: Ghostty.Notification.ghosttyNewTab,
             object: nil)
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(menuWillOpen(_:)),
+            name: Notification.Name(rawValue: "NSMenuWillOpenNotification"),
+            object: nil)
 
         // Configure user notifications
         let actions = [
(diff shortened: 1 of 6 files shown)


bo2themax (changes requested):
This is still a workaround for the issue you mentioned. I believe the rc here is that pre-registered key bindings are not affected by `key-remap`, while [`Surface.keyEventIsBinding`](https://github.com/ghostty-org/ghostty/blob/f9a9d33b3a40a95ba01cfbc0f89586567932a22b/src/Surface.zig#L2589) is using those mappings to query action, which will return null eventually and results in `false` in `performKeyEquivalent`.

But tbh, I'm not sure whether they should be remapped at all.

knu:
@bo2themax Thanks, I agree this is better handled at the key event dispatch layer.  I've just created #12602.

PR #12598, opened 2026-05-06
fix: preserve active cursor position during reflow

noib3:
This PR fixes an issue where reflowing could leave the active cursor attached to a clipped trailing blank cell instead of following the current write position.


mitchellh:
This makes sense, but I wonder if we could utilize the original x/y to do this without tracked pin? Plus, since the cursor has to be in the active area I feel like there are some optimizations to be made here.

noib3:
I considered using simple modulo/integer division ops to get the right coordinates, but I don't think that would handle wide cells correctly.

mitchellh (approved):
Approved pending CI.

Issue #10230, opened 2026-01-08
When `key-remap` is set, update configured keybinds


mitchellh:
Follow up to #10064. 

Right now we have a documented limitation where menu accelerators take priority because they're intercepted by the macOS menu system (and maybe GTK). But these are sourced from config. I realized we can probably fix this limitation by simply applying our key-remap to our keybinds in the `Config.finalize` method.

The right way to do this would be to add an `applyRemaps` or something function to our `input.Binding.Set` that takes a `input.KeyRemapSet` as a parameter and applies it to the full set. We'll probably have to be careful about re-indexing our hashes and handling reverse mapping too. 

I don't see a reason this won't work... but if anyone has any thoughts please share. I'll keep thinking about it too.

mitchellh:
Note, I took a look at this and we do need to take care that we handle conflicts in a deterministic, sensible way.

chenzhiwei:
I swapped `command` and `control` with `key-remap` feature.

In Ghostty bash interactive shell, `command + h` should act as `ctrl + h` which is delete the character before the cursor.

But the actual behavior is hide the Ghostty window, this is the MacOS shortcut.

Please consider fixing this unexpected behavior in this issue.

mitchellh:
This is being pushed to 1.4.0, the limitation is documented, we just didn't have time to fix this and no one offered one up.

fz-deepeye:
I am also trying to swap ctrl and super in ghostty:
Neither 
```
keybind=clear
key-remap = super=ctrl
key-remap = ctrl=super
keybind = ctrl+d=new_split:right
```
nor
```
keybind=clear
key-remap = super=ctrl
key-remap = ctrl=super
keybind = super+d=new_split:right
```
does what I want. I want new_split on the physical ctrl key. I

 expected the former, with binding to ctrl+d, to work as the remaps should not influence keybinds if I understand the documentation.

PR #12602, opened 2026-05-06
macos: apply key-remap to menu key equivalents

knu:
Refs #12293
Related: #3493 (duplicates: #4374, #5091), #7013
Supersedes: #12597

Apply `key-remap` before keybind lookup and when reporting keybind triggers for AppKit menu key equivalents.

Previously, menu shortcuts were registered from the configured trigger without considering modifier remaps. This let AppKit consume the physical shortcut before Ghostty could match the remapped keybind.

`Config.keyEventIsBinding` now remaps event modifiers before lookup, so remapped key equivalents match the same keybinds as normal key events.
`CApi.config_trigger_` returns the physical trigger that maps back to the configured logical modifiers, and leaves triggers unset when no remapped physical shortcut can produce them.

`RemapSet.unapply` performs the reverse lookup and is checked with `apply` so ambiguous or impossible remaps do not register incorrect menu shortcuts.

Inspired by: @bo2themax (https://github.com/ghostty-org/ghostty/pull/12597#pullrequestreview-4235918975)

AI usage: OpenAI Codex helped investigate, implement, test, and refine this change. I reviewed and tested the resulting code.

### DIFF
--- a/src/config/CApi.zig
+++ b/src/config/CApi.zig
@@ -117,7 +117,12 @@ fn config_trigger_(
     str: []const u8,
 ) !inputpkg.Binding.Trigger.C {
     const action = try inputpkg.Binding.Action.parse(str);
-    const trigger: inputpkg.Binding.Trigger = self.keybind.set.getTrigger(action) orelse .{};
+    var trigger: inputpkg.Binding.Trigger = self.keybind.set.getTrigger(action) orelse .{};
+    const logical_mods = trigger.mods.binding();
+    trigger.mods = self.@"key-remap".unapply(trigger.mods);
+    if (!self.@"key-remap".apply(trigger.mods).binding().equal(logical_mods)) {
+        return .{};
+    }
     return trigger.cval();
 }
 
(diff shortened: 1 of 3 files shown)


bo2themax (changes requested):
Not exactly what I thought. I think you can check mappings and remap if needed in `Config.keyEventIsBinding` and `CApi.config_trigger_ `, it should also make the menu shortcut correct again, and you don't need to change the apprt anymore.

knu:
@bo2themax Thanks for guiding me in this direction!

I updated the implementation to apply `key-remap` in `Config.keyEventIsBinding` and to expose remapped physical menu key equivalents through `CApi.config_trigger_`, without changing the AppKit/apprt event flow.

I also tested this manually on macOS with:

- `macos-option-as-alt = left`
- `key-remap = left_alt=command`
- `key-remap = command=left_alt`

With that config, the menu item for New Window displays `⌥N`, and pressing `⌥N` triggers New Window, while `⌘N` is sent to the terminal as `ESC N`.

knu:
@bo2themax How about this?

bo2themax:
> @bo2themax How about this?

It looks good to me in general, but we should wait for terminal maintainers to review this when they're available, since zig isn't really my profession here. 

Be patient plz, you can use your local build for now with [`-Doptimize=ReleaseFast`](https://ghostty.org/docs/install/build#:~:text=zig%20build-,%2DDoptimize%3DReleaseFast,-On%20Linux%2C%20this)

knu:
@bo2themax So, can I assume that the issue you marked as “Changes requested” has been resolved?  If so, I’d like to wait for other reviewers to review this as is.

bo2themax (approved):

knu:
@mitchellh Could you help identify the right person to review this PR?  It fixes macOS menu shortcuts not respecting key-remap, with changes in the shared config and input code.  @bo2themax approved it in May but recommended an additional review of the Zig changes, and it’s been waiting since then.

chenzhiwei:
Related: https://github.com/ghostty-org/ghostty/issues/10230

PR #12604, opened 2026-05-06
Implement needed modifications for issue #12600, more flexible copy-on-select and middle-click-action

### LINKED ISSUE #12600: more flexible `copy-on-select` and `middle-click-action`
`copy-on-select` should be modified to be more flexible. It should support `none`, `primary`, `clipboard`, `both`
The default should be `primary`. `true` will be an alias for `primary`, `false` will be an alias for `none`. On systems that do not have the concept of a primary clipboard `primary` should be a no-op.

`middle-click-action` should be modified with the options `ignore`, `paste-primary`, `paste-clipboard`. On systems that do not have the concept of a primary clipboard `paste-primary` should be a no-op.

Modifying `copy-on-select` or `middle-click-action` should have no effect on the operation of the other.

Verbiage around copying/pasting should be update to refer to only `primary` and `clipboard` since those are the standard terms. `selection` seems a bit too ambiguous.

### Discussed in https://github.com/ghostty-org/ghostty/discussions/12573

<sup>Originally posted by **mohshami** May  3, 2026</sup>
In Linux, terminal emulators have the following behavior

Middle-click always pastes from the PRIMARY/SELECTION clipboard.
Paste action (Normally ctrl+shift+V) always pastes from CLIPBOARD.

This behavior persists regardless of the copy-on-select option (for those terminal emulators that support it)

That behavior was added to Ghostty in https://github.com/ghostty-org/ghostty/pull/2347, along with the documentation below
```
# Whether to automatically copy selected text to the clipboard. `true`
# will prefer to copy to the selection clipboard, otherwise it will copy to
# the system clipboard.
#
# The value `clipboard` will always copy text to the selection clipboard
# as well as the system clipboard.
#
# Middle-click paste will always use the selection clipboard. Middle-click
# paste is always enabled even if this is `false`.
#
# The default value is true on Linux and macOS.
copy-on-select = true
```

#12443 broke that, making Ghostty behave differently from all other emulators on Linux.

I am not sure how terminal emulators behave in MacOS, I think #12443 should be limited to MacOS or a 4th option needs to be added to copy-on-select to modify CLIPBOARD but keep middle click paste from SELECTION.

Since #12443 is scheduled for 1.4, I think more people will see the difference once that out

I can't stress enough how much I appreciate the colossal amount of work everyone is putting. Thank you all</div>

mohshami:
For middle-click-action
* Kept the option "primary-paste" instead of "paste-primary" to keep backwards compatibility
* Added the option "clipboard-paste"

For copy-on-select
* Added the both, none and primary options

Updated config documentation

Note: No AI was used, Even though I don't know Zig, I looked at the code and the modifications seemed easy enough

Closes #12600


jcollie:
!vouch

src/config/Config.zig
```diff
@@ -8635,14 +8633,18 @@ pub const RepeatableLink = struct {
 pub const CopyOnSelect = enum {
```
jcollie, line 8765: Instead of just adding entries, we have a mechanism to deprecate values. See:
https://github.com/ghostty-org/ghostty/blob/main/src/config/Config.zig#L63-L98
https://github.com/ghostty-org/ghostty/blob/main/src/config/Config.zig#L4775-L4790

So `true` and `false` should get removed from here completely, and the compatibility handler should translate them to `none` or `primary`.

jcollie (changes requested):

src/config/Config.zig
mohshami: Thank you very much. Done :)

src/config/Config.zig
```diff
@@ -4885,6 +4881,28 @@ fn compatMacOSDockDropBehavior(
...
+    self: *Config,
+    alloc: Allocator,
+    key: []const u8,
+    value: ?[]const u8,
+) bool {
+    _ = alloc;
+    assert(std.mem.eql(u8, key, "copy-on-select"));
+
+    if (std.mem.eql(u8, value orelse "", "true")) {
+        self.@"copy-on-select" = .primary;
```
00-kat: Not exactly. Quoting the old docs:

> `true` will prefer to copy to the selection clipboard, otherwise it will copy to the system clipboard.

So this should be `.primary` on Linux (and FreeBSD, which Ghostty supports), and `.clipboard` elsewhere.

```suggestion
        self.@"copy-on-select" = switch (builtin.os.tag) {
            .linux, .freebsd => .primary,
            else => .clipboard,
        };
```

The more correct way would be to check `build_config.app_runtime` instead, since the GTK apprt reports `true` for `supportsClipboard` on all OSs that it runs on. However, using its `.none` to mean macOS felt hackier to me, since `.none` also includes libghostty.

src/config/Config.zig
```diff
@@ -2398,24 +2403,21 @@ keybind: Keybinds = .{},
...
-/// Middle-click primary paste (see `middle-click-action`) is enabled by
-/// default even if this is `false`. The clipboard it pastes from follows
-/// this setting: with `true` (or `false`) it reads from the selection
-/// clipboard (falling back to the system clipboard on platforms without a
-/// selection clipboard); with `clipboard` it reads from the system
-/// clipboard.
+/// `false` will disable all copy-on-select, `none` is alias for `false`.
 ///
-/// The default value is true on Linux and macOS.
+/// The default value is true on Linux and false otherwise
```
00-kat: Many small things about the docs:
  - `true` is documented wrong (see my previous comment). **Edit**: technically it isn't, but it would be once you fix the backward compatibility handler.
  - You forgot a few full stops.
  - MacOS → macOS (the official spelling).
  - Most other such options use an unordered list instead when there's more than two options.
  - Other places call “PRIMARY/SELECTION” the “selection clipboard” ([the old docs](https://github.com/ghostty-org/ghostty/blob/4b7bf0b20e3baf9c1ba10c63f2ad1fd853faea8f/src/config/Config.zig#L2401), [`CopyOnSelect.primary`](https://github.com/ghostty-org/ghostty/blob/08022d3f65a486262ce5f67496613dc2dfd57aa9/src/config/Config.zig#L8664), [`paste_from_selection`](https://github.com/ghostty-org/ghostty/blob/08022d3f65a486262ce5f67496613dc2dfd57aa9/src/input/Binding.zig#L378), [the old `Surface.zig` docs](https://github.com/ghostty-org/ghostty/blob/4b7bf0b20e3baf9c1ba10c63f2ad1fd853faea8f/src/Surface.zig#L2350)), and it reads nicer to those who aren't familiar with X jargon anyway.
  - You didn't document the version that would add these options. I'm not exactly sure where the “available since” should go for this; the previous `clipboard` is actually the new `both`, not the new `clipboard`.

```suggestion
/// Whether to automatically copy selected text to the clipboard.
///
/// Valid values:
///
/// * `none` - Do not copy selected text automatically.
///
/// * `primary` - On Linux, copy to the selection clipboard only. This has no
///   effect on macOS. (Available since: 1.4.0)
///
/// * `clipboard` - Copy text to the system clipboard only.
///
/// * `both` - Copy to both clipboards on Linux, and only the system clipboard
///   on macOS. (Available since: 1.4.0)
///
/// For backward compatibility and convenience, a value of `true` is the same as
/// `primary` on Linux and `clipboard` on macOS, and `false` is an alias for
/// `none`.
///
/// The default value is `primary` on Linux and `none` otherwise.
```

src/config/Config.zig
```diff
@@ -95,6 +95,11 @@ pub const compatibility = std.StaticStringMap(
     // Ghostty 1.3 rename the "window" option to "new-window".
     // See: https://github.com/ghostty-org/ghostty/pull/9764
     .{ "macos-dock-drop-behavior", compatMacOSDockDropBehavior },
+
+    // Ghostty 1.4 Update "copy-on-select", allow copying to PRIMARY/SELECTION
+    // (On supported operating systems), CLIPBOARD, or both.
```
00-kat: Similar to the “selection clipboard” part of my previous comment. Also, the other ones use “Ghostty \<version\> \<verb\>”; I think you were thrown off by the typo in the previous one (which says “rename” instead of “renamed”).

Also threw in a “semantics also changed” that I copy pasted from the `gtk-single-instance` one above.

```suggestion
    // Ghostty 1.4 updated "copy-on-select", allow copying to the selection
    // clipboard (on supported operating systems), the system clipboard, or
    // both. The semantics also changed but this is the correct mapping.
```

src/config/Config.zig
```diff
@@ -8641,15 +8659,17 @@ pub const RepeatableLink = struct {
 /// Options for copy on select behavior.
 pub const CopyOnSelect = enum {
     /// Disables copy on select entirely.
-    false,
+    none,
 
     /// Copy on select is enabled, but goes to the selection clipboard.
     /// This is not supported on platforms such as macOS. This is the default.
```
00-kat:
```suggestion
    /// This is not supported on platforms such as macOS. This is the default
    /// on Linux.
```

src/config/Config.zig
```diff
@@ -8641,15 +8659,17 @@ pub const RepeatableLink = struct {
 /// Options for copy on select behavior.
 pub const CopyOnSelect = enum {
     /// Disables copy on select entirely.
```
00-kat: To parallel the previous comment:

```suggestion
    /// Disables copy on select entirely. This is the default on macOS.
```

src/config/Config.zig
```diff
@@ -8641,15 +8659,17 @@ pub const RepeatableLink = struct {
...
-    false,
+    none,
 
     /// Copy on select is enabled, but goes to the selection clipboard.
     /// This is not supported on platforms such as macOS. This is the default.
-    true,
+    primary,
 
     /// Copy on select is enabled and goes to both the system clipboard
     /// and the selection clipboard (for Linux).
```
00-kat: Not true anymore:
```suggestion
    /// Copy on select is enabled and goes to the system clipboard.
````

File: src/config/Config.zig [hunk 8]
```diff
@@ -8641,15 +8659,17 @@ pub const RepeatableLink = struct {
...
     clipboard,
+
+    both,
```
00-kat, line 8779: The old `clipboard`'s docs are actually correct for `both`:

```suggestion
    /// Copy on select is enabled and goes to both the system clipboard
    /// and the selection clipboard (for Linux).
    both,
```

src/config/Config.zig
```diff
@@ -8673,8 +8693,10 @@ pub const RightClickAction = enum {
 
 /// Options for middle-click actions.
 pub const MiddleClickAction = enum {
-    /// Paste from the selection/standard clipboard per `copy-on-select`.
+    /// Paste from primary/selection.
```
00-kat: Re the “selection clipboard” comments above.

```suggestion
    /// Paste from the selection clipboard.
```

src/config/Config.zig
```diff
@@ -8673,8 +8693,10 @@ pub const RightClickAction = enum {
...
     @"primary-paste",
+    /// Paste from the clipboard.
```
00-kat: My previous comment asks you to change it to “the selection clipboard”, so an unqualified “clipboard” would be confusing if you commit that.

```suggestion
    /// Paste from the standard clipboard.
```

src/Surface.zig
```diff
@@ -2345,27 +2345,28 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
     if (prev_) |prev| if (sel.eql(prev)) return;
 
     switch (self.config.copy_on_select) {
-        .false => unreachable, // handled above with an early exit
+        .none => unreachable, // handled above with an early exit
 
-        // Both standard and selection clipboards are set.
-        .clipboard => try self.copySelectionToClipboards(
+        // Both PRIMARY and CLIPBOARD are set.
```
00-kat: Re my previous comments about “selection clipboard”, the previous comment was nicer.


```suggestion
        // Both standard and selection clipboards are set.
```

src/Surface.zig
```diff
@@ -2345,27 +2345,28 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
     if (prev_) |prev| if (sel.eql(prev)) return;
 
     switch (self.config.copy_on_select) {
```
00-kat, line 2381: Nit: maybe these should be sorted in the same order as `CopyOnSelect` (`copy_on_select`'s type), i.e. `.none`, then `.primary`, then `.clipboard`, then `.both`.

src/Surface.zig
```diff
@@ -2345,27 +2345,28 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
...
-                .selection
-            else
-                .standard;
-            try self.copySelectionToClipboards(
-                sel,
-                &.{clipboard},
-                .mixed,
-            );
-        },
+        // Only CLIPBOARD is set.
```
00-kat: This isn't even correct terminology on macOS, since full caps CLIPBOARD is the name of the X selection.

```suggestion
        // Only the standard clipboard is set.
```

src/Surface.zig
```diff
@@ -2345,27 +2345,28 @@ fn setSelection(self: *Surface, sel_: ?terminal.Selection) !void {
...
+        .clipboard => try self.copySelectionToClipboards(
+            sel,
+            &.{.standard},
+            .mixed,
+        ),
+
+        // The PRIMARY clipboard is set if supported, otherwise nothing is copied.
```
00-kat:
```suggestion
        // The selection clipboard is set if supported, otherwise nothing is copied.
```

src/Surface.zig
```diff
@@ -4015,22 +4016,16 @@ pub fn mouseButtonCallback(
         }
     }
 
-    // Middle-click paste source follows copy-on-select: when copy-on-select
-    // targets the selection clipboard, middle-click reads from it; when
-    // copy-on-select targets the system clipboard, middle-click reads from
-    // that instead. Falls back to the standard clipboard on platforms that
-    // do not support the selection clipboard.
+    // Middle-click action, either ignore, or paste from clipboard or paste from PRIMARY/SELECTION if supported.
```
00-kat: Another “selection clipboard” comment.

```suggestion
    // Middle-click action, either ignore, or paste from clipboard or paste from the selection clipboard if supported.
```

src/config/Config.zig
```diff
@@ -2434,8 +2436,9 @@ keybind: Keybinds = .{},
 /// The action to take when the user middle-clicks on the terminal surface.
 ///
 /// Valid values:
-///   * `primary-paste` - Paste from the selection (or system) clipboard per
-///      `copy-on-select`.
+///   * `primary-paste` - Paste from the PRIMARY/SELECTION clipboard on Linux.
```
00-kat: Another “selection clipboard” one.

```suggestion
///   * `primary-paste` - Paste from the selection clipboard on Linux.
```

src/config/Config.zig
```diff
@@ -2434,8 +2436,9 @@ keybind: Keybinds = .{},
...
+///      Does nothing on MacOS.
```
00-kat:
```suggestion
///      Does nothing on macOS.
```

src/config/Config.zig
```diff
@@ -2434,8 +2436,9 @@ keybind: Keybinds = .{},
...
+///   * `clipboard-paste` - Paste from the system clipboard on Linux and MacOS.
```
00-kat: “on Linux and MacOS” is redundant since well, that's everywhere[^1].

```suggestion
///   * `clipboard-paste` - Paste from the system clipboard.
```

[^1]: Technically FreeBSD is also supported, but the docs use “Linux” everywhere to mean “X11/Wayland”, which is what the BSDs use too.

00-kat (changes requested):
Thanks a lot! These are a couple comments I had on my mind when you first opened this PR, but I wanted to hold off until someone else asked you to use the backward compat mechanism instead (I don't know much about it, so I didn't know how to phrase that request then).

00-kat:
JSYK GitHub has a button to batch-add suggestions.

src/Surface.zig
mohshami: Done

jcollie (approved):

00-kat (approved):

00-kat:
(Disabled auto-merge since it still needs a review from someone in ghostty-org/terminal, so it wouldn't auto-merge until that happens.)

vcastellm (approved):
LGTM I need this, help! 😹

src/Surface.zig
```diff
@@ -2379,30 +2379,31 @@ fn setSelectionAndCopy(self: *Surface, sel: terminal.Selection) !void {
     try self.setSelection(sel);
 
     // If copy on select is false then exit early.
-    if (self.config.copy_on_select == .false) return;
+    if (self.config.copy_on_select == .none) return;
 
     switch (self.config.copy_on_select) {
-        .false => unreachable, // handled above with an early exit
+        .none => unreachable, // handled above with an early exit
```
jcollie: Since there's no code inbetween the above `if` and this `switch` entry I don't think that we need the `if` above.

jcollie (changes requested):
One small change requested. Also please squash and rebase on the latest `main`.

src/Surface.zig
mohshami: Terribly sorry, life just got in the way. Will do

mohshami: sorry my git-fu is not that strong, but it should be done now.

Thank you for your support :)

src/config/Config.zig
```diff
@@ -100,6 +100,15 @@ pub const compatibility = std.StaticStringMap(
     // Ghostty 1.4 renamed `scrollback-limit` to `scrollback-limit-bytes`
     // when `scrollback-limit-lines` was added so the units are explicit.
     .{ "scrollback-limit", cli.compatibilityRenamed(Config, "scrollback-limit-bytes") },
+
+    // Ghostty 1.4 Update "copy-on-select", allow copying to PRIMARY/SELECTION
+    // (On supported operating systems), CLIPBOARD, or both.
```
00-kat: You seem to have messed up that rebase:

```suggestion
```

src/Surface.zig
00-kat: This would UB now since the early return is, well, not present anymore.

```suggestion
        .none => {},
```

00-kat (changes requested):

src/config/Config.zig
mohshami: I broke the repo multiple times trying to do this the last time I missed this, sorry

00-kat (approved):

mohshami:
I'm rebasing my branch whenever I see an update just to keep this PR up to date

mitchellh (approved):
Thank you for diligently updating this and I'm sorry for the very slow review. This looks good to me and I appreciate it.

mitchellh:
We have a real test failure

mohshami:
> Thank you for diligently updating this and I'm sorry for the very slow review. This looks good to me and I appreciate it.

No worries at all, and you are most welcome. I just feel happy I could contribute to such an amazing project.

Let me see what I can do about this

mohshami:
Ok I figured out how to replicate the error on my laptop, let me see how I can fix

Again my zig-fu is almost non-existent 😅

mohshami:
Looked at the right click paste code and managed to get it to build, but when I ran Ghostty I couldn't get paste to work. After some debugging I figured out it was due to this block here

https://github.com/ghostty-org/ghostty/blob/8af6897c0afc63037a8a3efee4162a380e3a4572/src/apprt/gtk/class/surface.zig#L2883-L2888

I didn't have gtk_enable_primary_paste enabled in my GTK settings and I don't recall ever having it. Yet all other terminals had middle click paste working.

Shall I remove that code block before pushing? Just to make Ghostty consistent with other terminals.

Thank you all for your help :)

mohshami:
I have pushed the build fix but kept the code block in place

mohshami:
failed over an extra space 😅

PR #12612, opened 2026-05-07
Glyph protocol renderer implementation

raphamorim:
1. Initial support only for glyf
2. Updated spec to 1.9 without positioning api only width parameter

image


trag1c:
!vouch

src/terminal/apc/glyph/handler.zig
```diff
@@ -0,0 +1,351 @@
...
+            .reason = reason,
+        } },
+    };
+}
+
+/// Unicode Private Use Areas per spec §4: basic PUA, supplementary
+/// PUA-A, supplementary PUA-B. Endpoints follow the spec (e.g.
+/// 0xFFFFD, not 0xFFFFF) because codepoints above them are
+/// noncharacters.
+fn isPUA(cp: u21) bool {
```
raphamorim, line 134: added this but if there's other place in the source code I can reuse lemme know

mitchellh:
Sorry I force pushed/rebased my parser base, can you rebase this so the diff is reasonable?

PR #12614, opened 2026-05-07
macOS: normalize working directory paths with FilePath

bo2themax:
This fixes for [Nushell](https://github.com/nushell/nushell/blob/f342d8acfa81bd9543bb38e3e08b1a93ecafe5d1/crates/nu-protocol/src/engine/engine_state.rs#L1012) when opening Ghostty via Finder service and Shortcuts, also makes path parsing more robust in AppleScript.

image

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceView.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceView.swift
@@ -1,6 +1,7 @@
 import SwiftUI
 import UserNotifications
 import GhosttyKit
+import System
 
 extension Ghostty {
     /// Render a terminal for the active app in the environment.
@@ -611,8 +612,20 @@ extension Ghostty {
         /// Explicit font size to use in points
         var fontSize: Float32?
 
+        private var normalizedWorkingDirectory: String?
         /// Explicit working directory to set
-        var workingDirectory: String?
+        var workingDirectory: String? {
+            get { normalizedWorkingDirectory }
+            set {
+                guard let newValue else {
+                    normalizedWorkingDirectory = nil
+                    return
+                }
+                // We use FilePath to normalize separators by removing redundant intermediary separators
+                // and stripping any trailing separators.
+                normalizedWorkingDirectory = FilePath(newValue).string
+            }
+        }
 
         /// Explicit command to set
         var command: String?


macos/Sources/Features/Services/ServiceProvider.swift
```diff
@@ -48,8 +49,9 @@ class ServiceProvider: NSObject {
 
         for url in directoryURLs {
             var config = Ghostty.SurfaceConfiguration()
-            config.workingDirectory = url.path(percentEncoded: false)
-
+            // We use FilePath to normalize separators by removing redundant intermediary separators
+            // and stripping any trailing separators.
+            config.workingDirectory = FilePath(url)?.string
```
Kyle-Ye: I saw those pattern appear in multi places (>= 3)

What do you think if we add a extension for URL/String instead?

Something like

```swift
import System

extension URL {
    // We use FilePath to normalize separators by removing redundant intermediary separators
    // and stripping any trailing separators.
    package var normalizedURLString: String? {
        FilePath(url)?.string
    }
}
```

bo2themax: I don't think it really needs one, and this is just for file path not suitable for any urls.

that comment is sure duplicated at multiple places tho

macos/Sources/Features/App Intents/NewTerminalIntent.swift
```diff
@@ -74,7 +75,9 @@ struct NewTerminalIntent: AppIntent {
         // If we were given a working directory then open that directory
         if let url = workingDirectory?.fileURL {
             let dir = url.hasDirectoryPath ? url : url.deletingLastPathComponent()
-            config.workingDirectory = dir.path(percentEncoded: false)
+            // We use FilePath to normalize separators by removing redundant intermediary separators
+            // and stripping any trailing separators.
+            config.workingDirectory = FilePath(dir)?.string
```
jparise: This path can now fail, but I don't think that can happen in practice given that we've already guaranteed this is a `file://` URL.

jparise (approved):
Do we also need the `FilePath` treatment in `func application(_ sender: NSApplication, openFile filename: String)`, which also sets `config.workingDirectory`? That `filename` comes from app launch path.

bo2themax:
> Do we also need the `FilePath` treatment in `func application(_ sender: NSApplication, openFile filename: String)`, which also sets `config.workingDirectory`? That `filename` comes from app launch path.

No, I checked that by dragging one on the dock icon, looks like AppKit already removed the trailing slashed. 

But I'll double check for `open` command later.

macos/Sources/Features/Services/ServiceProvider.swift
bo2themax: Removed other duplicates in `set`

bo2themax:
> > Do we also need the `FilePath` treatment in `func application(_ sender: NSApplication, openFile filename: String)`, which also sets `config.workingDirectory`? That `filename` comes from app launch path.
> 
> No, I checked that by dragging one on the dock icon, looks like AppKit already removed the trailing slashed.
> 
> But I'll double check for `open` command later.

Yes, the `filename` is already normalized, but I force pushed and moved the changes in `set`, so now it will normalized the new value regardlessly.

macos/Sources/Ghostty/Surface View/SurfaceView.swift
```diff
@@ -611,8 +612,16 @@ extension Ghostty {
...
 
+        private var normalizedWorkingDirectory: String?
         /// Explicit working directory to set
-        var workingDirectory: String?
+        var workingDirectory: String? {
+            get { normalizedWorkingDirectory }
+            set {
+                // We use FilePath to normalize separators by removing redundant intermediary separators
+                // and stripping any trailing separators.
+                normalizedWorkingDirectory = newValue.flatMap(FilePath.init(_:))?.string
```
jparise: A `guard` might be more readable than the `Optional.flatMap` here.

bo2themax: `FilePath(_ string: String)` doesn't produce optional values, I don't think it's needed here. But sure it's easier to read 🫪

jparise (approved):

PR #12631, opened 2026-05-08
libghostty-vt: handle OSC color queries

noib3:
This PR enables libghostty-vt to handle the OSC 10/11/12 queries, extracting the formatting of the corresponding resposes from `stream_terminal.zig` into `color.zig` in the process.


tristan957:
nit pick, but it would be nice to do the refactor in a separate commit.

noib3:
Closed by https://github.com/ghostty-org/ghostty/pull/13239.

PR #12639, opened 2026-05-09
macos: simplify workingDirectory setter

jparise:
This is a minor improvement to the computed property's `set` logic: we can just use `.map {}` to unify the two optional paths.


jparise:
@bo2themax realized this after I reviewed #12614.

bo2themax (approved):

PR #12653, opened 2026-05-11
libghostty-vt: preserve shell prompts on resize by default

noib3:
This PR makes libghostty-vt preserve shell prompts across resize unless the shell explicitly opts into prompt clearing/redraw with `redraw=1`.

### DIFF
--- a/src/terminal/c/terminal.zig
+++ b/src/terminal/c/terminal.zig
@@ -259,6 +259,11 @@ fn new_(
     });
     errdefer t.deinit(alloc);
 
+    // libghostty-vt embedders don't necessarily install Ghostty's shell
+    // integration, so don't assume OSC 133 prompts can be redrawn on resize.
+    // Shells can still opt in with OSC 133;A;redraw=1.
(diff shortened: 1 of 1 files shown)


mitchellh (changes requested):
Darn, I can't remember the history here of why this is even defaulted to true. It seems like it might be safe to default this to false in all of Ghostty until we see _any_ semantic prompt marker. I want to think about that for a bit instead of the targeted fix here in libghostty-vt.

jparise:
> Darn, I can't remember the history here of why this is even defaulted to true. It seems like it might be safe to default this to false in all of Ghostty until we see _any_ semantic prompt marker. I want to think about that for a bit instead of the targeted fix here in libghostty-vt.

Given that `;redraw` is a Kitty invention (not in OSC 133 itself) to opt'ing in to shell redraws, I agree it makes sense to default this to `false` in the entire runtime.

mitchellh:
For now just going to merge this but noted I think we should default this off.

PR #12660, opened 2026-05-11
terminal: report DECECM as permanently reset

### LINKED ISSUE #12505: Implement response to DECRQM for DECECM as "permanently reset"
> The physical terminals most of us are emulating, namely VT100 and VT220s, don't deal with this, since they do not have color support as mentioned in the above sources.

Note that there are a couple of hardware terminals with color support - most notably the VT525 - and it actually has a mode that allows apps to control this behaviour: `DECECM` (Erase Color Mode). As defined in the programmer documentation, "it controls the background color used when text is erased or new text is scrolled on to the screen". When set, these operations fill with the default attributes (compatible with monochrome terminals). When reset, they fill with the active rendition colors (which is compatible with most modern color terminal emulators). The factory default is reset, i.e. it fills with the active colors.

While `DECECM` isn't actually listed as a conformance requirement for the Color Text Extension, I think it would be a good idea for color terminals to a least report how they are implementing that mode in response to a `DECRQM` query (i.e. permanently reset if you always fill with the active colors, or permanently set if you always fill with the default attributes). That way apps can potentially adjust their behavior to match. Ideally I'd like to see more terminals fully support the mode, though, so apps can turn it on if they don't want the color background fill when scrolling.

_Originally posted by @j4james in https://github.com/ghostty-org/ghostty/discussions/12497#discussioncomment-16745394_

athaapa:
Closes #12505 

This PR allows Ghostty to respond to DECRQM queries for DECECM with the "permanently reset".

AI disclosure: I used Codex to help inspect the relevant code path and explain the issue, but I reviewed and made the code changes myself.

### DIFF
--- a/src/terminal/modes.zig
+++ b/src/terminal/modes.zig
@@ -10,8 +10,6 @@
 const std = @import("std");
 const testing = std.testing;
 
-const dececm_mode: u15 = 117;
-
 /// A struct that maintains the state of all the settable modes.
 pub const ModeState = struct {
     /// The values of the current modes.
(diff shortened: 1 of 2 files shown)


pluiedev (commented):
The diff itself looks good, but maybe we want to attach some context as to why we do this? (links, etc)

athaapa:
Yeah, that’s fair. If an application queries `?117`, returning `not recognized` does not tell it which erase-color behavior Ghostty uses. Returning `permanently reset` tells it that Ghostty recognizes the mode but always uses the reset behavior. I updated the comment with that context and linked the VT520/VT525 Programmer Information.

jcollie (approved):

athaapa:
Friendly ping. I addressed the review feedback and this has one approval now. Is there anything else you’d like changed before this can land?

jcollie (approved):

mitchellh (approved):

PR #12682, opened 2026-05-14
config: clear `command-palette-entry` like `keybind`

bo2themax:
After #1368, `command-palette-entry=` will no longer clear the entries like the documentation says. Since i couldn't find an existing issue or discussion about this, I assume no one is actually using it. So I put 1.4.0 here, lemme know if you want to change it to 1.3.2.

> I basically copied the `keybind` parsing code and doc.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2880,9 +2880,16 @@ keybind: Keybinds = .{},
 /// command-palette-entry = title:"Ghostty",description:"Add a little Ghostty to your terminal.",action:"text:\xf0\x9f\x91\xbb"
 /// ```
 ///
+/// There are some additional special values that can be specified for
+/// command-palette-entry:
+///
+///   * `command-palette-entry=clear` will clear all command entries. Warning: this
+///     removes ALL entries up to this point, including the default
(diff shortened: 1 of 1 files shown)


pluiedev (approved):
I think this makes sense, yeah - thanks!

jcollie (approved):
LGTM. Stick with 1.4 as the milestone as this is new functionality.

00-kat:
Bit late but seconded; nota bene:

> Since i couldn't find an existing issue or discussion about this

Not everyone reads docs in any logical order; if someone just wanted to get rid of the command palette entries, they're more likely to copy paste from *its* documentation than to read the docs on the general syntax which they undoubtedly needed to have understood, especially given Ghostty's syntax is deceptively simple.

**Edit**: even of the people who do, it's definitely possible it's missed due to the already complicated syntax of `command-palette-entry`. And that assumes people report minor UX infelicities, which isn't particularly common either.

But yeah I don't think anyone actually used this, à la `keybind =` and `keybind = clear`. It definitely deserves a noticeable mention in the breaking changes section of the release notes though, since it'd be extremely confusing for the potential few who do use it.

PR #12693, opened 2026-05-15
gtk: continuously animate indeterminate progress bar

BryanSant:
## Summary

Closes a user-experience parity gap between the macOS and GTK apprts for
indeterminate progress reports (OSC 9;4 state 3, or state 1/2 without a
percentage). On macOS, `BouncingProgressBar` animates continuously while
indeterminate. On GTK, the bar would render a single pulse and then sit
frozen until the next OSC report — which often never comes for truly
indeterminate work — making it look like the terminal had hung.

The root cause is that `GtkProgressBar.pulse()` advances the bouncing
block by exactly one step per call; it does not start an animation. To
get continuous motion the widget must be pulsed on a timer.

This change drives a 100ms repeating `g_timeout` while the surface is in
an indeterminate progress state and tears it down on every other state
transition (`set` with progress, `pause`, `remove`, config-disabled, and
surface finalize), so the GTK bar now matches the macOS feel.

## Test plan

- [ ] `printf '\e]9;4;3;\a'` — bar pulses continuously, stops on
      `\e]9;4;0;\a`
- [ ] `printf '\e]9;4;1;\a'` (no percentage) — same continuous pulse
- [ ] `printf '\e]9;4;1;50\a'` — fixed 50% fraction, no pulsing
- [ ] `printf '\e]9;4;4;\a'` (pause) after an indeterminate state stops
      the animation
- [ ] Closing the surface while pulsing does not leak the timer
- [ ] `progress-style = false` in config hides the bar and cancels the
      timer

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -646,6 +646,7 @@ pub const Surface = extern struct {
 
         // Progress bar
         progress_bar_timer: ?c_uint = null,
+        progress_bar_animation_timer: ?c_uint = null,
 
         // True while the bell is ringing. This will be set to false (after
         // true) under various scenarios, but can also manually be set to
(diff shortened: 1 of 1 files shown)


jcollie:
This behavior was actually by design on the GTK side. The TUI must arrange to update an indeterminate progress bar periodically on it's own. It so that TUIs that _don't_ properly update the indeterminate progress bar can be detected. If Ghostty (or the OS) update an indeterminate progress bar by itself you can't determine if a TUI program has frozen or not. I'd argue that the macOS behavior is incorrect and should be changed to match the GTK behavior.

PR #12699, opened 2026-05-15
split_tree: rescale split ratio when resizing

dkinzler:
Fixes `SplitTree.resize` not rescaling the split ratio to be relative to the size of the split. Added a unit test for resizing a nested split.

Previously the new ratio was incorrectly calculated relative to the entire grid. As a consequence resizing a nested split in the GTK app would cause unexpected size changes like large jumps. E.g. in the following recording the window has height ~1000px and the resize was done using a keybind for `resize_split:up,10`. The change is much larger than 10 pixels.

https://github.com/user-attachments/assets/ba375ddf-5b2f-45e4-8b12-69021ef2f8a8

Note that even with this fix, resizing by a small amount like 10 pixels might not work at all (depending on window size and layout), because of the same bug causing #11193 (see my PR #12698). Initially an inaccurate split ratio will be set and eventually written back to the split tree datastructure. That incorrect split ratio will be the same before and after the small resize, so nothing actually changes in the UI.

The split tree implementation for the macOS app already calculates the ratios correctly.

AI Disclosure
No AI was used, the bug was discovered and all code written by myself.

### DIFF
--- a/src/datastruct/split_tree.zig
+++ b/src/datastruct/split_tree.zig
@@ -860,25 +860,26 @@ pub fn SplitTree(comptime V: type) type {
             var sp = try result.spatial(gpa);
             defer sp.deinit(gpa);
 
-            // Get the ratio of the split relative to the full grid.
-            const full_ratio = full_ratio: {
-                // Our scale is the amount we need to multiply our individual
-                // ratio by to get the full ratio. Its actually a ratio on its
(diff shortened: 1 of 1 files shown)


jcollie (approved):

jcollie:
Thanks! This combined with 12698 works quite well in my testing.

PR #12701, opened 2026-05-16
fix: allow ctrl+c/v during IME composition

LuckyQ7:
The macOS surface view was dropping single-byte control characters while an IME composition was active. This broke Ctrl+C and Ctrl+V in Chinese input methods, so we now let those terminal-significant shortcuts through while continuing to suppress other composing control input.


bo2themax:
#12522 fixes this as well

PR #12710, opened 2026-05-16
macos: treat unfocused split modifiers as UI state

rockorager:
The terminal controller listens for flagsChanged events so all splits can keep modifier-sensitive UI state current.  Forwarding those events through SurfaceView.flagsChanged() also routes them through the normal key input path, so an unfocused split can receive modifier press and release events as terminal input.

Send modifier changes for unfocused splits through a modifier-only surface entry point instead.  The core surface updates mouse modifier state, link highlighting and cursor shape from that path, but does not run key binding or terminal encoding.  Apply key remaps there as well so focused and unfocused modifier state stay consistent.

Factor the shared modifier and mouse-shape work into helpers so the normal key path and the modifier-only path use the same state transitions.  Keep the physical-key filter in keyToMouseShape(), and expose the pure modsToMouseShape() calculation for cases where AppKit reports only a modifier state change.

To reproduce the old behavior: open two splits in Ghostty and run `kitten show-key -m kitty` in each. Modifier key presses appear in both splits.

AI Disclosure: This patch was prepared with the help of AI.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1120,6 +1120,7 @@ GHOSTTY_API void ghostty_surface_set_color_scheme(ghostty_surface_t,
                                                      ghostty_color_scheme_e);
 GHOSTTY_API ghostty_input_mods_e ghostty_surface_key_translation_mods(ghostty_surface_t,
                                                                          ghostty_input_mods_e);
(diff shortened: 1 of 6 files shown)


bo2themax (approved):
Swift changes LGTM 👍

Issue #12516, opened 2026-04-29
macOS: search bar should remember what was previously entered


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9790

<div type='discussions-op-text'>

<sup>Originally posted by **joeshaw** December  2, 2025</sup>
### Issue Description

Mac apps remember what the previous value was in the search bar when `cmd-f` is hit again. This also reruns the search and causes the matches to be highlighted again.  The text is selected, though, so typing in new text immediately replaces the old text.  Again, how iTerm2, Safari, and other apps work.

### Expected Behavior

I expect Ghostty to behave more like a native Mac app with regard to search behavior.

### Actual Behavior

See above.

### Reproduction Steps

Mostly mentioned above.

### Ghostty Version

```text
Ghostty 1.3.0-main+34fd1dd0f

Version
  - version: 1.3.0-main+34fd1dd0f
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS 26.1 (Tahoe)

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
Default config
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

joeshaw:
This was addressed for GTK in #9765.

nolinmcfarland:
I’m interested in fixing this and opened a vouch request (#12686) with more details!

PR #12712, opened 2026-05-17
macOS: use find pasteboard for search needle

nolinmcfarland:
Fixes the issue described in #12516.

### What
- Inject an `OSPasteboard` into `SearchState`
- Add `OSPasteboard` extension to normalize working with strings between UIPasteboard/NSPasteboard
- Add `BackportSelectionTextField` which supports text selection for MacOS 15/iOS 18 and up.
- Read from the pasteboard when the overlay opens and when the app becomes active
- Write to the pasteboard when the search needle changes
- Annotate `SearchState` as MainActor. `NSPasteboard` isn't thread safe, and since `SearchState` is already accessed from the main thread, MainActor enforces our writes be thread safe
- Add SearchState unit tests

### Why
Consistent with other macOS apps, the Find bar's search needle should persist when re-opened and should sync to the Find bar in other apps. For example, see Xcode, Notes, Terminal, and Safari.

https://github.com/user-attachments/assets/b6a55a4a-a52c-45bc-ac38-c9df452c11cb

### DIFF
--- a/macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
+++ b/macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
@@ -131,7 +131,7 @@ extension Ghostty.OSSurfaceView {
 
         init(
             from startSearch: Ghostty.Action.StartSearch,
-            pasteboard: OSPasteboard = OSPasteboard(name: .find)
+            pasteboard: OSPasteboard = OSPasteboard.find
         ) {
             self.pasteboard = pasteboard
             if let needle = startSearch.needle, !needle.isEmpty {
@@ -143,7 +143,7 @@ extension Ghostty.OSSurfaceView {
         }
 
         func readPasteboardNeedle() {
-            let pasteboardNeedle = pasteboard.string(forType: .string)
+            let pasteboardNeedle = pasteboard.string
             if let pasteboardNeedle, pasteboardNeedle != needle {
                 needle = pasteboardNeedle
                 needleSelection = needle.startIndex..<needle.endIndex
@@ -151,8 +151,7 @@ extension Ghostty.OSSurfaceView {
         }
 
         func writePasteboardNeedle() {
-            pasteboard.clearContents()
-            pasteboard.setString(needle, forType: .string)
+            pasteboard.string = needle
         }
     }
 
--- a/macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
+++ b/macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
@@ -115,13 +115,39 @@ extension Ghostty {
 // MARK: Search State
 
 extension Ghostty.OSSurfaceView {
-    class SearchState: ObservableObject {
+    @MainActor class SearchState: ObservableObject {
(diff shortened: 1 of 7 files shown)


macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
```diff
@@ -115,13 +115,39 @@ extension Ghostty {
...
+            } else {
+                readPasteboardNeedle()
+            }
+        }
+
+        func readPasteboardNeedle() {
+            let pasteboardNeedle = pasteboard.string(forType: .string)
+            if let pasteboardNeedle, pasteboardNeedle != needle {
+                needle = pasteboardNeedle
+            }
```
bo2themax, line 150: Changing needle here will **deselect** the current selection in the TextField. I would like to see that we explicitly select all after syncing like first party apps, so the user can quickly delete them if they want.

You may need to introduce a backport TextSelection to manipulate it.

bo2themax (changes requested):

macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
nolinmcfarland: @bo2themax Thanks! Addressed in 69cab3d8085d0731574f4498864ced901e55c8b0.

I added a `BackportSelectionTextField` that supports the behavior. Here's a recording of the selection happening. Let me know if you have any questions/concerns.

https://github.com/user-attachments/assets/6a092452-595e-4cca-ac16-aeb7b7530298

macos/Sources/Helpers/Backport.swift
```diff
@@ -131,3 +131,49 @@ enum BackportNSGlassStyle {
...
+        text: Binding<String>,
+        selection: Binding<Range<String.Index>?>
+    ) {
+        self.titleKey = titleKey
+        self._text = text
+        self._textSelection = selection
+    }
+
+    var body: some View {
+        if #available(macOS 15, *) {
```
bo2themax:
```suggestion
        if #available(iOS 18.0, macOS 15, *) {
```

bo2themax (changes requested):
Thanks, only one small styling suggestion left. I doubt iOS would support this in near future, but selection works iOS, so why not.

macos/Sources/Helpers/Backport.swift
nolinmcfarland: Good call, addressed in ed521606122cde73f0072abc6a7caa5da7dc4829

bo2themax:
Waiting CI

nolinmcfarland:
I'll take a look at and fix the CI failures

macos/Sources/Helpers/Extensions/OSPasteboard+Extension.swift
nolinmcfarland, line 1: @bo2themax I added this extension to fix the CI build failures for iOS. It normalizes the interface for working with strings on UI/NSPasteboard. Should be good to go now.

bo2themax (approved):

mitchellh (approved):
Beautiful work. Very high quality. Thank you.

Issue #12294, opened 2026-04-15
macOS: Global keybind stops in the app BUT not outside the app


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/12091

<div type='discussions-op-text'>

<sup>Originally posted by **jamtur01** April  4, 2026</sup>
### Issue Description

I am having the same/similar issue described in:

https://github.com/ghostty-org/ghostty/issues/11883

Global keybind for toggle_visiblity works correctly when Ghostty is first launched. The keybind continues to work when Ghostty is the focused application, but no longer triggers globally from other apps. Restarting Ghostty restores the functionality temporarily.

```
keybind = global:cmd+space=toggle_visibility
```

And yes, I've checked all the other things (Spotlight), etc., that could bind the key, and it is not bound.

### Expected Behavior

The terminal toggles.

### Actual Behavior

The terminal is toggled OFF (i.e. away from Ghostty) but not BACK to Ghostty.

### Reproduction Steps

1. Open ghostty
2. Use Cmd->Space
3. Toggle terminal from Ghostty to next app
4. Use Cmd->Space
5. Nothing happens.

### Ghostty Logs

```text
❯ sudo log stream --level debug --predicate 'subsystem=="com.mitchellh.ghostty"'
Filtering the log data using "subsystem == "com.mitchellh.ghostty""
Timestamp                       Thread     Type        Activity             PID    TTL
2026-04-03 18:08:09.813599-0400 0xe93f5d   Info        0x0                  40966  0    ghostty: [com.mitchellh.ghostty:generic_renderer] generic_renderer: updating display link display id=1
2026-04-03 18:08:10.057425-0400 0xe93f5d   Info        0x0                  40966  0    ghostty: [com.mitchellh.ghostty:generic_renderer] generic_renderer: updating display link display id=1
2026-04-03 18:08:10.421357-0400 0xe93f5d   Info        0x0                  40966  0    ghostty: [com.mitchellh.ghostty:generic_renderer] generic_renderer: updating display link display id=1
2026-04-03 18:08:11.332038-0400 0xe93f5d   Info        0x0                  40966  0    ghostty: [com.mitchellh.ghostty:generic_renderer] generic_renderer: updating display link display id=1
```
```


### Ghostty Version

```text
❯ ghostty +version
Ghostty 1.3.2-main+0790937d0

Version
  - version: 1.3.2-main+0790937d0
  - channel: tip
Build Config
[...]

PR #12714, opened 2026-05-17
macOS: Re-enable global keybinds after event tap disable events

adrum:
While testing https://github.com/ghostty-org/ghostty/pull/9857, I encountered the behavior mentioned below. It's pretty frustrating to encounter, so I've been actually compiling this fix into my test builds for last month or so, and the issue has not come back. I exclusively use the QuickTerminal, so my workflow depends on global keybinds working reliably. 

Issue: https://github.com/ghostty-org/ghostty/issues/12294

The solution includes listening to two events that are fired when a tap is disabled:
- tapDisabledByTimeout
- tapDisabledByUserInput

When these are fired, we re-enable the tap.

Apple's Docs: https://developer.apple.com/documentation/coregraphics/cgeventtype?language=swift

Related Discussions:
- https://github.com/ghostty-org/ghostty/discussions/11819
- https://github.com/ghostty-org/ghostty/discussions/12091

### DIFF
--- a/macos/Sources/Features/Global Keybinds/GlobalEventTap.swift
+++ b/macos/Sources/Features/Global Keybinds/GlobalEventTap.swift
@@ -16,7 +16,7 @@ class GlobalEventTap {
 
     // The event tap used for global event listening. This is non-nil if it is
     // created.
-    private var eventTap: CFMachPort?
+    fileprivate var eventTap: CFMachPort?
 
     // This is the timer used to retry enabling the global event tap if we
     // don't have permissions.
@@ -124,6 +124,17 @@ private func cgEventFlagsChangedHandler(
     userInfo: UnsafeMutableRawPointer?
 ) -> Unmanaged<CGEvent>? {
     let result = Unmanaged.passUnretained(cgEvent)
+
+    // macOS disables the event tap if the callback is too slow or for other
+    // internal reasons. When that happens it sends this event type. We need
+    // to re-enable the tap or it stays dead forever.
+    if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
+        GlobalEventTap.logger.warning("global event tap was disabled by the system, re-enabling")
+        if let machPort = GlobalEventTap.shared.eventTap {
+            CGEvent.tapEnable(tap: machPort, enable: true)
+        }
+        return result
+    }
 
     // We only care about keydown events
     guard type == .keyDown else { return result }


bo2themax:
This looks reasonable to me👍🏻. Any idea how can I do to let OS fire these two events so that I can reproduce and verify this?

bo2themax (approved):

adrum:
> This looks reasonable to me👍🏻. Any idea how can I do to let OS fire these two events so that I can reproduce and verify this?

I have found one "unnatural" way of triggering one of these events. A few prerequisites:
- These steps require building and running the app in Xcode, which requires you to remove and add Ghostty (Debug build) to `System Settings > Privacy & Security > Accessibility` on every run
-  The easiest way to find the location of the .app is to right click on the Ghostty Debug app icon > Options > Show In Finder, which takes you to the location in DervedData with the output.  I just drag this .app to the list of apps in System Settings and it re-enables it.
- Make sure you have a global keybind in your config. Mine: `keybind = global:ctrl+`=toggle_quick_terminal`, which we use for the following reproduction steps:

---

1. Focus any app but Ghostty
1. Trigger global keybind (ctrl+`)
1. See quick terminal
4. Trigger global keybind (ctrl+`) to hide it
1. Set a breakpoint on line 140 in `GlobalEventTap.swift`: `guard type == .keyDown else { return result }`
3. Focus any app but Ghostty (Xcode probably already is)
4. Trigger global keybind (ctrl+`)
5. Wait two seconds
6. Resume execution
7. You should now see our log message on line 132: `GlobalEventTap.logger.warning("global event tap was disabled by the system, re-enabling")`
8. Completely disable all breakpoints
3. Focus any app but Ghostty (Xcode probably already is)
10. Trigger global keybind (ctrl+`)
11. You will see it still respond to our global keybind

If you recompile with line `CGEvent.tapEnable(tap: machPort, enable: true)` commented out (and add the app to Accessibility in Privacy & Security), you see step 14 fail.

While I recognize this is not a natural way to encounter these events, it does reproduce the logic on-demand. I think running the debug app throughout the lifecycle of normal use for a week should yield these events being fired naturally. I close my laptop (and sleep) quite a bit throughout the week, and my gut tells me this event can fire during the lifecycle of sleep/wake.  It can also occur of the Ghostty app stops responding or is beach balling (due to resource usage by subprocesses/commands).

---

[This article](https://danielraffel.me/til/2026/02/19/cgevent-taps-and-code-signing-the-silent-disable-race/) also mentions setting up a timer, but I thought that might be too aggressive. I have it mocked up here, just in case we need it: https://github.com/adrum/ghostty/commit/580a2e2f7cd0fda55bff4877d96d5d164afd03bd

bo2themax:
>I think running the debug app throughout the lifecycle of normal use for a week.

I do encounter this before for debugging

PR #12725, opened 2026-05-18
build(highway): require `apple_sdk` for darwin builds

elias8:
Noticed this was removed in another PR, but `apple_sdk` is required to build libghsotty for the iOS simulator, specifically for the x86 version (see the error log [here](https://github.com/elias8/libghostty/actions/runs/26075576793/job/76666498246)). Figured it'd be better to include the SDK in all darwin builds for consistency.

### DIFF
--- a/pkg/highway/build.zig
+++ b/pkg/highway/build.zig
@@ -37,6 +37,13 @@ pub fn build(b: *std.Build) !void {
         try android_ndk.addPaths(b, lib);
     }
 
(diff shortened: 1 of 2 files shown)


bo2themax (approved):
Makes sense to me

mitchellh:
Yeah I thought we did this for the no libc++ and cross-compilation stuff, but we test that now in CI so if CI is green then I'm good.

PR #12736, opened 2026-05-19
macOS: confirm clipboard/url with the same TextViewAlert

bo2themax:
Previously `ClipboardConfirmationView` involves few workarounds, based on the fact that `SwiftUI.TextEditor` doesn't offer selectable but not editable api. By converting it to an NSAlert, we can benefit from NSTextView and we can clean up some xibs and add help button more easily.

I checked with #1516, it's still fixed.
Xnip2026-08-05_22-08-23
Xnip2026-08-05_22-08-37

---

Also update the keyboard shortcuts in UntrustedURLAlert.

Xnip2026-08-05_22-08-48
Xnip2026-08-05_22-08-57


bo2themax:
Alert is not suited for the clipboard anymore

PR #12742, opened 2026-05-19
macOS: review windows when quitting

bo2themax:
Inspired by `Terminal.app` which I think is a nice feature.

First two commits contains some changes in `BaseTerminalController` so that I can use swift concurrency to review those windows in chain more easily.

https://github.com/user-attachments/assets/41d92432-4ae0-499e-961a-fc247602f3d7

Works with tabs as well, i forgot to record that.


mitchellh (approved):
I like it. Thanks!

PR #12752, opened 2026-05-20
vouch: enable auto-locking closed issues

trag1c:


mitchellh:
We should cut a tag but good with this for now!

PR #12781, opened 2026-05-23
macOS: fix search bar Enter key blocking IME composition

minorcell:
Closes https://github.com/ghostty-org/ghostty/discussions/12774

`.onKeyPress(.return)` unconditionally returns `.handled`, so when IME is composing the return key never reaches the IME to confirm the candidate. The search bar gets stuck.

The fix: use `.onSubmit` for the next-match navigation — it only fires when there is no composing text. In `.onKeyPress` only intercept shift+return (previous match), return `.ignored` otherwise.

Tested on macOS 26.5, Ghostty 1.3.1, built from source. Chinese Pinyin input in the search bar works correctly after the fix.

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceView.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceView.swift
@@ -420,7 +420,6 @@ extension Ghostty {
                     .onSubmit {
                         _ = surfaceView.navigateSearchToNext()
                     }
-                    }
 #if canImport(AppKit)
                     .onExitCommand {
                         if searchState.needle.isEmpty {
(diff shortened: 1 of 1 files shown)


macos/Sources/Ghostty/Surface View/SurfaceView.swift
```diff
@@ -417,6 +417,10 @@ extension Ghostty {
                         // to our synced needle.
                         searchState.readPasteboardNeedle()
                     }
+                    .onSubmit {
+                        _ = surfaceView.navigateSearchToNext()
+                    }
+                    }
```
bo2themax: This doesn't look right

bo2themax (changes requested):

macos/Sources/Ghostty/Surface View/SurfaceView.swift
minorcell: fixed, extra brace from conflict resolution

PR #12785, opened 2026-05-23
libghostty: add tracked grid ref API

mitchellh:
Add a C API for tracked pins, known as a tracked grid ref in C.

The new API can create tracked refs from terminal points, snapshot them back to regular grid refs for cell access, convert them to coordinates, move them to a new point, report when their semantic location was lost, and free the tracked pin bookkeeping. This is backed by PageList tracked pins and exposed through the libghostty-vt export layer and headers.

### DIFF
--- a/example/c-vt-grid-ref-tracked/README.md
+++ b/example/c-vt-grid-ref-tracked/README.md
@@ -1,0 +1,19 @@
+# Example: `ghostty-vt` Tracked Grid References
+
+This contains a simple example of how to use the `ghostty-vt` terminal and
+tracked grid reference APIs to keep a long-lived reference to a cell as the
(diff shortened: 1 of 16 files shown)


mitchellh:
Doing this work identified a potentially memory corruption scenario in surface. Fixed in a follow up commit.

> Left-click mouse state stored a tracked pin with only the screen key that
> owned it. If the alternate screen was removed and later recreated, the key
> could match again even though the stored pin belonged to destroyed PageList
> storage.
> 
> Store the screen generation alongside the left-click pin and resolve the
> pin through helpers that require both the key and generation to match. This
> keeps selection scrolling, link hover checks, pressure selection, and drag
> selection from dereferencing stale tracked pins after screen teardown.

PR #12795, opened 2026-05-24
libghostty: detach tracked grid refs on free

mitchellh:
Tracked grid references previously held a raw terminal wrapper pointer and were required to be freed before the terminal. If callers kept one past terminal destruction, later tracked-ref calls could dereference freed terminal or page-list memory before detecting that the reference was no longer meaningful.

Track live C tracked-grid-ref handles from the terminal wrapper and detach them before tearing down terminal storage. Detached refs now report no value through the tracked-ref APIs and can still be freed by the caller. Update the C API docs to describe this lifetime behavior and add a regression test for using a tracked ref after terminal free.

This introduces some overhead but tracked pins shouldn't be numerous and this dramatically improves safety.

No API changes due to this (just more safety).

### DIFF
--- a/include/ghostty/vt/grid_ref.h
+++ b/include/ghostty/vt/grid_ref.h
@@ -70,8 +70,9 @@ extern "C" {
  * the tracked reference between screens.
  *
  * Tracked references are owned by the caller and must be freed with
- * ghostty_tracked_grid_ref_free() before the terminal that created them is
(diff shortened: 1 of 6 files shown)


mitchellh:
I did this for Rust lifetimes. So let it be known I did this because I care about Rust.

PR #12814, opened 2026-05-25
cli: rework +ssh-cache internals and user interface

jparise:
This change primarily focused on a revised +ssh-cache user interface, but it also reworks a bunch of the internals.

The primary CLI improvement is support for positional arguments and a consistent list output format that includes both the ISO-formatted timestamp and relative age.

    ghostty +ssh-cache                           # List all cached destinations
    ghostty +ssh-cache user@example.com          # Show that destination
    ghostty +ssh-cache example.com               # Show all users on that host
    ghostty +ssh-cache --add=user@example.com    # Manually add a destination
    ghostty +ssh-cache --remove=user@example.com # Remove a destination
    ghostty +ssh-cache --prune=30d               # Remove entries older than 30 days
    ghostty +ssh-cache --clear                   # Clear entire cache

Notable, we now support a --prune operation that replaces the previous --expire-days flag that was never actually hooked up to anything (!!). --prune also supports a wider range of Duration-based values.

We're also much more consistent with error codes: 0=success, 1=failure, 2=usage.

While working on those changes, I also reworked the cache internals, particularly the code around timestamp handling and errors. For example, I dropped the explicit error sets because they were growing unwieldy, and in practice we only matched on a subset of those errors.

Lastly, overall test coverage should be much improved, especially around the time- and allocation-related operations.

---

*AI Disclosure:* I made a lot of iterative, AI-assisted (Claude Opus 4.7) correctness passes over this work. It was particularly helpful in tracing through the various failure modes, and it wrote those unit tests in the process.

### DIFF
--- a/src/cli/ssh-cache/DiskCache.zig
+++ b/src/cli/ssh-cache/DiskCache.zig
@@ -17,14 +17,6 @@ const MAX_CACHE_SIZE = 512 * 1024;
 /// Path to a file where the cache is stored.
 path: []const u8,
 
-pub const DefaultPathError = Allocator.Error || error{
-    /// The general error that is returned for any filesystem error
-    /// that may have resulted in the XDG lookup failing.
-    XdgLookupFailed,
-};
-
-pub const Error = error{ CacheIsLocked, HostnameIsInvalid };
-
 /// Returns the default path for the cache for a given program.
 ///
 /// On all platforms, this is `${XDG_STATE_HOME}/ghostty/ssh_cache`.
@@ -33,7 +25,7 @@ pub const Error = error{ CacheIsLocked, HostnameIsInvalid };
 pub fn defaultPath(
     alloc: Allocator,
     program: []const u8,
-) DefaultPathError![]const u8 {
+) ![]const u8 {
     const state_dir: []const u8 = xdg.state(
         alloc,
         .{ .subdir = program },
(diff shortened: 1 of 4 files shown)


tristan957:
Do we have have any precedent for subcommands? I generally prefer `ghostty +ssh add xyz` whereas you have gone with long options. Obviously being consistent is better than anything, but figured I would bring it up.

jparise:
> Do we have have any precedent for subcommands? I generally prefer `ghostty +ssh add xyz` whereas you have gone with long options. Obviously being consistent is better than anything, but figured I would bring it up.

This maintains the `+ssh-cache` interface that's already shipped, which used options rather than subcommands. I think it would be fine to further revise this to use commands (with compatibility for the previous flags), but that might be for a follow-up change given all of the things this already does.

I don't think I'd try to merge this functionality into `+ssh`, because that's positioned as a `ssh` wrapper, and it's command line parsing already has a high bar to avoid ambiguities between the Ghostty-specific flags and SSH's.

tristan957:
> > Do we have have any precedent for subcommands? I generally prefer `ghostty +ssh add xyz` whereas you have gone with long options. Obviously being consistent is better than anything, but figured I would bring it up.
> 
> This maintains the `+ssh-cache` interface that's already shipped, which used options rather than subcommands. I think it would be fine to further revise this to use commands (with compatibility for the previous flags), but that might be for a follow-up change given all of the things this already does.
> 
> I don't think I'd try to merge this functionality into `+ssh`, because that's positioned as a `ssh` wrapper, and it's command line parsing already has a high bar to avoid ambiguities between the Ghostty-specific flags and SSH's.

Makes sense. Thanks!

Issue #8957, opened 2025-09-29
Audible ping not working as expected


jcollie:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/8947

<div type='discussions-op-text'>

<sup>Originally posted by **MeisterP** September 29, 2025</sup>
### Issue Description

When running `ping -a 8.8.8.8` only the first ping creates a bell sound

### Expected Behavior

Every successful ping should trigger the bell

### Actual Behavior

Only the first ping triggers the bell

### Reproduction Steps

1. Open Ghostty
2. enable `bell-features = system`
3. Run `ping -a 8.8.8.8`
4. listen

### Ghostty Version

```text
Ghostty 1.2.0

Version
  - version: 1.2.0
  - channel: stable
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.gtk
  - font engine   : font.main.Backend.fontconfig_freetype
  - renderer      : renderer.generic.Renderer(renderer.OpenGL)
  - libxev        : io_uring
  - kernel version: 6.16.9-gentoo-dist
  - desktop env   : gnome
  - GTK version   :
    build         : 4.18.6
    runtime       : 4.20.1
  - libadwaita    : enabled
    build         : 1.7.6
    runtime       : 1.7.6
  - libX11        : enabled
  - libwayland    : enabled
```

### OS Version Information

Gentoo

### (Linux only) Display Server

X11

### (Linux only) Desktop Environment/Window Manager

Gnome 48.4

### Minimal Ghostty Configuration

```ini
bell-features = system
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

jcollie:
`pipewire` was a red herring.

PR #12815, opened 2026-05-25
apprt/gtk: fix audio-bell GStreamer thread leak (reuse one MediaFile per surface)

nikicat:
## Problem

Every audio bell calls `gtk.MediaFile.newForFilename`, which spins up a full GStreamer pipeline. The GTK4 GStreamer backend's GL sink starts `gstglcontext`/`gldisplay-event` threads that are **never joined on teardown**, so allocating a fresh `MediaFile` per ring leaks a pipeline and ~4 threads on every bell. The old `notify::ended -> unref` handler discarded the pipeline but did not (and could not) join those threads.

A long-running instance accumulated **705 threads over ~4h** of normal use.

## Fix

Cache one `MediaFile` per surface (`priv.bell_media`), rebuilt only when `bell-audio-path` changes and unref'd on `dispose`. Each bell now replays the same pipeline via `seek(0)` + `play()` instead of creating a new one. `seek(0)` is required so an ended stream plays again (cf. #8957).

## Verification

Confirmed on a real running instance with the fix: GStreamer's global element counter only ever reached `oggdemux4` over an hour of use (one pipeline per bell-ringing surface, reused for every subsequent bell) and the process thread count stayed flat — versus the per-bell growth before.

## Commits

1. **The fix** — reuse one MediaFile per surface.
2. **Unit regression test** — guards the `bellMediaFile` reuse contract (same path → same object, changed path → rebuild). Runs in the existing `test-gtk` CI job; needs no display.
3. **End-to-end CI job** *(kept separate so it can be dropped independently)* — `test/bell-leak.sh` + a `test-gtk-bell-leak` workflow job that runs ghostty headless (Xvfb + software GL), rings 120 bells, and fails if the thread count grows per-bell. It's heavier and more environment-sensitive (needs Xvfb/Mesa/GStreamer on the runner), so it's isolated for easy review/removal.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -281,4 +281,113 @@ in {
           server.wait_for_file("${user.home}/.terminfo/x/xterm-ghostty", timeout=30)
     '';
   };
+
+  # Regression test for the GTK audio-bell GStreamer thread leak. Each audio
+  # bell used to allocate a fresh gtk.MediaFile (and thus a GStreamer pipeline
+  # whose GL sink spawns gstglcontext/gldisplay-event threads that are never
+  # joined), leaking ~4 threads per ring; the fix reuses one MediaFile per
+  # surface. This rings many bells and asserts the GUI process thread count
+  # stays bounded. Runs under GNOME on Wayland so it exercises the real path.
+  bell-leak-check-gnome = mkTestGnome {
+    name = "bell-leak-check-gnome";
+    settings = {
+      # The VM has no GPU, so GNOME and Ghostty render via llvmpipe. Give the
+      # guest enough cores/RAM that software GL can bring up Ghostty's window
+      # before the +new-window D-Bus activation times out, and force clean
+      # software GL so mesa doesn't stall probing for absent hardware.
+      virtualisation.cores = 4;
+      virtualisation.memorySize = 4096;
+      environment.sessionVariables = {
+        LIBGL_ALWAYS_SOFTWARE = "1";
+        GALLIUM_DRIVER = "llvmpipe";
+      };
+
+      home-manager.users.ghostty = {
+        xdg.configFile = {
+          "ghostty/config".text = ''
+            bell-features = audio
+            bell-audio-path = ${pkgs.sound-theme-freedesktop}/share/sounds/freedesktop/stereo/bell.oga
+            bell-audio-volume = 0
+          '';
+        };
+      };
+    };
+    testScript = {nodes, ...}: let
+      user = nodes.machine.users.users.ghostty;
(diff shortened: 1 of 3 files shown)


jcollie:
!vouch

test/bell-leak.sh
```diff
@@ -0,0 +1,94 @@
...
+# we can sample the post-bell settle (leaked threads are permanent; transient
+# GStreamer pool threads drain here).
+child="sleep 3; for i in \$(seq $N); do printf '\\a'; sleep 0.12; done; sleep 6"
+
+# A unique app-id so we only ever measure the instance we launched, never some
+# other ghostty already running on the machine. --gtk-single-instance=false
+# guarantees this process is its own GUI rather than handing off to another.
+MARKER="com.ghostty.BellLeakTest"
+
+xvfb-run -a -s "-screen 0 1280x1024x24" \
```
jcollie: Xvfb is the wrong choice for this, as X11 support for Gnome is going away very soon. We need to have this test running in Wayland. This should also probably be a Nix check (see `nix/tests.nix`) and not a standalone shell script.

jcollie (changes requested):
Just one initial comment as I don't have the time for a full review right now.

test/bell-leak.sh
nikicat: Thanks! Reworked per your feedback:

- Dropped this Xvfb shell script and the workflow job.
- Added it as a NixOS check, `bell-leak-check-gnome`, alongside the other `*-gnome` tests — it launches Ghostty under GNOME on Wayland, rings 100 bells in the window, and fails if the GUI process's thread count grows per-bell.
- Verified locally: ~1 thread of growth over 100 bells, vs ~+400 without the fix.

One note: the test VM has no GPU so it renders via llvmpipe, which made the `+new-window` D-Bus activation occasionally exceed its client-side timeout (the window still comes up). The test waits for the window to appear rather than hard-failing on that call. Happy to adjust if you'd prefer that handled differently.

pluiedev (commented):
The comments are a bit verbose (you seriously need to trim LLM-generated comments) but the logic looks good.

jcollie:
So `unref`'ing a MediFile never shuts down the GStreamer threads associated with it? Won't we then still be "leaking" one set of GStreamer threads per surface? So opening 1000 surfaces, playing a bell, and closing 9999 surfaces would leave 9999 sets of GStreamer threads running? Or am I misunderstanding something?

jcollie (approved):
Going to approve this. I don't think that it totally solves the leak of GStreamer threads but it definitely improves the situation.

PR #12834, opened 2026-05-27
libghostty: handle non-monotonic selection times in SelectionGesture

mitchellh:
We expect monotonic time but since this is libghostty-exposed we need to  be more careful about guarding what we accept. This also practically makes it easier to integrate with various languages.

Compare the two instants first and treat backwards timestamps like any other failed repeat. The next press becomes a fresh single-click anchor, and a regression test covers the reset behavior.


mitchellh:
Don't like this approach

PR #12863, opened 2026-05-30
macOS: improve logging experience for developers

bo2themax:
Previously, when we wanted to locate the logs that are emitted from `Ghostty.xcframework` (zig part), in Xcode's debug console, we had to copy and search in another editor to see them. With this, now we can just right-click **Jump To Source** on that log item.

The first commit touches a lot of files but only [`src/log.zig`](https://github.com/ghostty-org/ghostty/pull/12863/changes#diff-cc10636b160e891ce9aa77ae7289893fccd4e7fd0621746b85602d8910c47d44) and [`src/main_ghostty.zig`](https://github.com/ghostty-org/ghostty/pull/12863/changes#diff-68b77120d87317a3c2b3fe0d9f822a5e3e62b66c4ab30a3444a22a0da37ff51d) are most relevant, the rest are just import changes.

Xnip2026-05-30_18-50-35

### AI Disclosure

I used Claude (Plan mode) to implement this, and tbh I don't actually fully comprehend the changes in `src/log.zig`, but I asked another agent session to verify and review its assumption with online resources. And I verified it with Xcode and `log stream`. Since the private SPI [`_os_log_impl`](https://github.com/PureDarwin/Libtrace/blob/067c6b6156fa6ecfc0528c70944e63b86cc7d787/libsystem_trace/os_log.c#L89) is only used in debug builds, I would say it's rather safe.

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -16,7 +16,7 @@ const BlockingQueue = @import("datastruct/main.zig").BlockingQueue;
 const renderer = @import("renderer.zig");
 const font = @import("font/main.zig");
 
-const log = std.log.scoped(.app);
+const log = @import("log.zig").scoped(.app);
 
 const SurfaceList = std.ArrayListUnmanaged(*apprt.Surface);
 
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -38,7 +38,7 @@ const inspectorpkg = @import("inspector/main.zig");
 const SurfaceMouse = @import("surface_mouse.zig");
 const ProcessInfo = @import("pty.zig").ProcessInfo;
 
-const log = std.log.scoped(.surface);
+const log = @import("log.zig").scoped(.surface);
 
 // The renderer implementation to use.
 const Renderer = rendererpkg.Renderer;
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -22,7 +22,7 @@ const configpkg = @import("../config.zig");
 const Config = configpkg.Config;
 const String = @import("../main_c.zig").String;
 
-const log = std.log.scoped(.embedded_window);
+const log = @import("../log.zig").scoped(.embedded_window);
 
 pub const resourcesDir = internal_os.resourcesDir;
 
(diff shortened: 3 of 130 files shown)


jcollie (changes requested):
I'm not opposed to improving the macOS logging, but touching nearly every `.zig` file to change the call to `std.log.scoped` doesn't seem right to me. Can't this be implemented inside `logFn` in `main_ghostty.zig` with no changes to all of the `.zig` files?

https://github.com/ghostty-org/ghostty/blob/main/src/main_ghostty.zig#L114-L164

bo2themax:
> Can't this be implemented inside `logFn` in `main_ghostty.zig` with no changes to all of the `.zig` files?


I don't think so, in order to make it work, the `os_log_with_type`(but it's not available for zig, that's why `_os_log_impl` is used here) has to be inlined into call site, so it will have the correct file and line. Swift also has the same limitation btw, if you wrap `os.Logger.log(level:_:)` in another function.

jparise:
The `macos/` changes that add `privacy` annotations look good. Move that to its own PR while we iterate on the Zig side?

jparise:
`logFn` lacks source-location information, so our options are a little more limited on the Zig side. https://github.com/ziglang/zig/issues/21906 proposed something like that earlier, but it's been closed as not planned.

Given that, maybe we could consider a custom logger that's source-location aware (and can better interface with the `os.log`) that's only used in a few particularly useful (Zig) locations.

bo2themax:
> Given that, maybe we could consider a custom logger that's source-location aware (and can better interface with the `os.log`) that's only used in a few particularly useful (Zig) locations.

Yeah, I thought of that too, also in swift in my other projects, but so far I couldn't find any other approaches to achieve this. There's a `#line` and `#file` directories in c (`#sourceLocation` available in swift), but all of the them only accept static strings as parameters, so the location passed into custom logger won't work. I don't think `@callerSrc()` will work🤔 

> In my swift projects, I'm using swift macros to expand #log into `os.log` and `swift-log.logger` at call site to achieve this🫪

mitchellh:
I think the approach you did makes sense, but I'm not sure at the moment if its really worth it. Our log messages are unique enough I haven't had a problem opening them up in Vim. I see the convenience but just not sure this amount of diff and maintaining all the externs in the logger makes it worth it.

bo2themax:
Closing this for now

PR #12881, opened 2026-05-31
macos: fix AXVisibleCharacterRange and UTF-16 indexing

jparise:
Several accessibility methods on `SurfaceView` returned values in the wrong coordinate system, which made the visible-range and selection APIs unusable across screen readers, translation tools, and AI autocomplete apps:

- `accessibilityVisibleCharacterRange` returned the full scrollback as the visible range. AX clients would fetch hundreds of KB to translate a single sentence on screen.
- `accessibilityNumberOfCharacters` reported grapheme count rather than UTF-16 code units. Any range derived from it indexed into the wrong bytes when the buffer contained emoji.
- `accessibilityLine(for:)` treated the parameter as a grapheme index, same problem.
- `accessibilitySelectedTextRange` returned viewport-cell-linear offsets (y*cols+x), which are unrelated to UTF-16 `NSRange` semantics. The selection's range and text disagreed for non-ASCII content.

The main idea is a self-consistent screen-text snapshot. The Zig core adds `Screen.screenText` which uses `ScreenFormatter` directly to emit the full screen along with a per-byte `Pin` array, then binary-searches the array for the viewport's start/end byte offsets. The C API exports this as `ghostty_surface_read_screen` / `ghostty_screen_text_s`.

The macOS layer caches a `Ghostty.SurfaceView.ScreenText` snapshot that holds the text, the viewport `NSRange` in UTF-16, the total UTF-16 length, and a precomputed `lineStarts` table. The AX overrides read from this single cache so all reported lengths and ranges live in the same coordinate system.

`accessibilitySelectedTextRange` searches the cached text for the selection's bytes and returns the unique match, or `NSNotFound` when the selection appears more than once. A wrong-but-valid range would mislead AX clients more than `NSNotFound`.

The preexisting `cachedScreenContents` and `cachedVisibleContents` stay in place because `GetTerminalDetailsIntent` still reads them. Folding those into the new cache can be a separate cleanup.

The overall result was verified using some small AX scripts to confirm correct `AXVisibleCharacterRange`, `AXNumberOfCharacters`, `AXStringForRange`, and `AXSelectedTextRange` behavior.

Note that `AXRangeForPosition`and `AXBoundsForRange` are still not yet implemented. Those need a little more foundational work and can come in follow-up changes.

See: #9932

---

_AI Disclosure:_ I did the initial investigation using Claude Opus 4.7, which was helpful in generating the test cases and planning the overall phased approach.


macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
```diff
@@ -210,6 +210,46 @@ extension Ghostty {
         private(set) var cachedScreenContents: CachedValue<String>
         private(set) var cachedVisibleContents: CachedValue<String>
 
+        /// Snapshot of the full screen text with viewport position
```
bo2themax: Can we move this struct to the `OSSurfaceView` so that iOS can also benefit from this?

src/terminal/Screen.zig
```diff
@@ -2511,6 +2511,82 @@ pub fn selectionString(
...
+    const end = start + std.sort.partitionPoint(
+        Pin,
+        pins.items[start..],
+        self.pages.getBottomRight(.viewport).?,
+        struct {
+            fn pred(ctx: Pin, pin: Pin) bool {
+                return !ctx.before(pin);
+            }
+        }.pred,
+    );
```
jparise: This approach could be a lot more efficient if we didn't need to allocate and search through the entire `pin_map`.

I separately implemented a new `pin_offsets` formatter capability (#12882) that would makes this operation much cheaper.

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
```diff
@@ -2277,14 +2387,41 @@ extension Ghostty.SurfaceView {
...
 
-    /// Returns the range of text that is currently selected in the terminal.
-    /// This allows VoiceOver and other assistive technologies to understand
-    /// what text the user has selected.
+    /// UTF-16 NSRange of the current selection within the cached
+    /// screen text, or `NSRange(NSNotFound, 0)` when no selection
+    /// exists or the selection text appears more than once.
     override func accessibilitySelectedTextRange() -> NSRange {
-        return selectedRange()
+        guard let selected = accessibilitySelectedText(), !selected.isEmpty else {
```
bo2themax: Is there a way where we can get the selected line? Then we use that as the haystack to search.

bo2themax (changes requested):

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
jparise: Yes, that's the right direction. I think the best path is to expose selection's exact byte offset in the canonical screen text from the Zig side, and I was planning on working on that in a follow-up change.

I think we could do something in Swift in the short term, but I think it might become overly complex, so my preference is to start with what we have here and continue to improve it incrementally.

What do you think?

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
jparise: Done. I only moved the struct and left the `cacheScreenContents` here in the AppKit-specific code, alongside the other cached values.

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
jparise: Actually, I see a path to implementing this pretty cleanly as part of this pull request. Working on that now.

bo2themax (approved):
macOS part looks good to me 👍

mitchellh (changes requested):
Awesome, thanks for looking at this. I think the only sus part is the Screen.zig changes. I get what you're going for and why formatters directly don't provide this,but I want to spend more time thinking about if there is a better way to expose this.

jparise:
> Awesome, thanks for looking at this. I think the only sus part is the Screen.zig changes. I get what you're going for and why formatters directly don't provide this,but I want to spend more time thinking about if there is a better way to expose this.

Assume you mean this bit, or something else?

```zig
while (r.end > r.start and text[r.end - 1] == '\n') r.end -= 1;
```

jparise:
> Awesome, thanks for looking at this. I think the only sus part is the Screen.zig changes. I get what you're going for and why formatters directly don't provide this,but I want to spend more time thinking about if there is a better way to expose this.

I rebased this, and the formatter work got a lot cheaper after #13587. 

Using the formatter here does stand out. We're filling a whole buffer just to extract some byte ranges, and then we toss it. Most callers are looking for offsets->pins, while we're looking for pins->offsets (specifically viewport TL/BR, selection TL/BR). But at the moment, it's the only machinery we have that, er, formats into the screen representation.

I originally tried using a counting writer with the formatter, but that doesn't produce true byte offsets due to newlines and region handling, e.g.:

```
full text  = "1ABCD\n2EFGH\n3IJKL\n4MNOP\n5QRST"
prefix     = "1ABCD\n2EFGH\n"     <- 12, the correct offset
scoped emit= "1ABCD\n2"           <- 7, what a scoped format returns
```

So given that the true byte offsets are only available during emit:

1. Continue with the current approach.
2. Feed the formatter a list of pins to watch (viewport TL/BR, selection TL/BR), and have it emit their `{start, end}` byte ranges as it formats. This would be a new formatter API.

What do you think? Or any other ideas on how to make this screen representation available to the macOS AX system (which wants `{location, length}` ranges)?

PR #12882, opened 2026-05-31
terminal: add ScreenFormatter.pin_offsets

jparise:
When you want to know where specific Pins land in formatter output but don't need the full per-byte Pin array, pin_offsets is a lighter alternative to pin_map. The caller supplies a list of Pins to watch and the formatter fills in the byte offset where each one first appears in the emitted output.

Internally, PageListFormatter scans the per-chunk Coordinate scratch buffer it already builds for pin_map consumers, so PageFormatter itself is unchanged. pin_map and pin_offsets can be set independently or together; when neither is set, no scratch buffer is allocated.

A future caller that needs only a handful of positions can avoid the cumulative pin_map allocation. For a 10MB scrollback that's ~160MB of transient Pin storage traded for ~360KB of per-chunk scratch.

### DIFF
--- a/src/terminal/formatter.zig
+++ b/src/terminal/formatter.zig
@@ -124,6 +124,27 @@ pub const PinMap = struct {
     map: *std.ArrayList(Pin),
 };
 
+/// Records the byte offset where each watched Pin first appears in the
+/// emitted output. The `offset` field of each `PinOffset` is filled in
+/// by the formatter; entries left as null mean that Pin was not
+/// emitted.
+///
+/// The formatter still allocates a per-chunk scratch buffer of source
+/// coordinates (one entry per emitted byte, reused across chunks via
+/// `clearRetainingCapacity`), which is why an allocator is required.
+/// Peak memory is bounded by a single chunk's emitted-byte count —
+/// dramatically less than `PinMap`'s cumulative per-byte Pin array,
+/// but not zero.
+pub const PinOffsets = struct {
+    alloc: Allocator,
+    entries: []PinOffset,
+};
+
+pub const PinOffset = struct {
+    pin: Pin,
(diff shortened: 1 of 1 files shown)


src/terminal/formatter.zig
```diff
@@ -737,9 +771,26 @@ pub const PageListFormatter = struct {
...
         const br: PageList.Pin = self.bottom_right orelse self.list.getBottomRight(.screen).?;
 
-        // If we keep track of pins, we'll need this.
-        var point_map: std.ArrayList(Coordinate) = .empty;
-        defer if (self.pin_map) |*m| point_map.deinit(m.alloc);
+        // PageFormatter only knows about a single Page, so it can't
+        // track Pins directly. When the caller has asked for a pin_map
+        // or pin_offsets we feed PageFormatter a per-byte Coordinate
+        // scratch buffer and convert to Pins at the chunk boundary.
+        const PointMap = struct {
```
jparise, line 778: There's an opportunity to hoist up this type and share it between `PageListFormatter` and `PageFormatter`:

https://github.com/ghostty-org/ghostty/blob/16f2fdc90c2ca59ed870c7b3355a337b1fd650b2/src/terminal/formatter.zig#L834-L837

... but that will trigger a bunch of test code updates too, and I didn't want to make this initial PR too noisy.

If the `pin_offsets` idea looks good, I can add another commit here that shares the type, or we can look at it in a separate change.

jparise:
The initial motivation for this capability is a more efficient "slice of the entire screen that's currently visible" search for #12881.

jparise:
This isn't quite the right shape for the current version of #12881. I'm going to close this now and will revisit as a future optimization.

PR #12893, opened 2026-06-01
font: glyf outline decoder and rasterizer

mitchellh:
This adds a Glyf outline decoder and rasterizer.

So it turns out that FreeType and CoreText have very shitty APIs for raw Glyf table rasterization. CoreText as far as I can find can't do it at all. In both cases you have to create a synthetic font with just this entry and rasterize the glyph. And the code to do all that was WAYYYYYY complex such that this made way more sense.

We need this for the Glyph Protocol.

**AI disclosure:** Hand-written parser, rasterizer. AI assisted validation and test writing. I read the spec myself.

cc @qwerasd205


mitchellh:
Actually, just thought of a totally allocation-free way to do this. Costs more CPU but decoding outlines and rasterizing should be rare and cached (in an Atlas) so I think we can trade memory for CPU here.

**Edit:** The allocation-free path worked, but it was horribly complicated compared to just allocating. Given raterizing is rare, I think we just store the raw data and then reinitialize and allocate whenever we rasterize. And that's fine.

src/font/opentype/glyf.zig
```diff
@@ -15,6 +16,49 @@ const sfnt = @import("sfnt.zig");
 pub const Glyf = struct {
     data: []const u8,
 
+    /// A decoded glyph outline.
+    ///
+    /// The `countours` slice is the list of end point indices and
```
raphamorim: nit: contours

PR #12900, opened 2026-06-02
libghostty-vt: add options to configure default cursor's style and blink

noib3:
This PR adds 2 options to `libghostty-vt` to configure the style and blink status of the default cursor. They control how the terminal renders the cursor when a program doesn't request any explicit style or when it resets it to the terminal's default state by sending a DECSCUSR reset sequence (`CSI 0 q`).


mitchellh (approved):

mitchellh:
Approved pending CI.

PR #12902, opened 2026-06-02
core: send selection_changed notification

jparise:
The core had no signal to the apprt when the active selection changed, so a consumer (e.g. a screen reader) kept reading a stale selection until some unrelated query refreshed it.

This change adds a payload-less selection_changed action that's fired on a selection state transition. The apprt reads the current selection through the normal read path.

This consolidates selection state changes so the notification fires consistently: all sites route through setSelection rather than calling screen.select directly, including the mouse paths that previously bypassed it for clipboard timing.

The new setSelectionAndCopy extends setSelection with the additional 'copy_on_select' behavior.

On macOS, this posts .ghosttySelectionDidChange, which is debounced before posting a NSAccessibility .selectedTextChanged notification.

GTK has no consumer yet and no-ops the action.

See: #9932


jcollie (approved):
LGTM

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1070,6 +1073,27 @@ extension Ghostty {
             }
         }
 
+        private static func selectionChanged(
+            _ app: ghostty_app_t,
+            target: ghostty_target_s) {
+            switch target.tag {
+            case GHOSTTY_TARGET_APP:
+                Ghostty.logger.warning("selection changed does nothing with an app target")
+                return
```
bo2themax, line 1082: We should return a Boolean here otherwise, last time I checked there will be another waning log for not handling this event, if we don't

bo2themax (dismissed):
LGTM

bo2themax (approved):

mitchellh (approved):

mitchellh:
Appreciate the smaller PRs towards a11y improvements. Thank you.

PR #12907, opened 2026-06-03
terminal: saturate cursor subtraction in resizeCols

c0x0o:
PageList.resize takes the .lt branch when columns shrink, which calls resizeWithoutReflow (mutating self.rows to the new smaller value) and then resizeCols with the original opts.cursor.y. When both axes shrink in one call and the cursor sits at or past the new bottom row, the expression `self.rows - c.y - 1` underflows and panics in safety builds.

Use saturating subtraction; "remaining rows below cursor" is 0 once the cursor sits at or past the new bottom.

This problem is reported by [discussion#12905](https://github.com/ghostty-org/ghostty/discussions/12905)

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -1059,7 +1059,7 @@ fn resizeCols(
         break :cursor .{
             .tracked_pin = c.pin orelse try self.trackPin(p),
             .untrack = c.pin == null,
-            .remaining_rows = self.rows - c.y - 1,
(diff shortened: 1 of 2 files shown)


mitchellh:
Thank you. We should cover this with a unit test.

c0x0o:
Two unit tests are added: one for PageList itself, one for libvt c API(respect the reproduce PoC in discussion), PTAL @mitchellh

jcollie (approved):

PR #12938, opened 2026-06-06
renderer: glyph protocol

mitchellh:
This adds the renderer part of the glyph protocol, tying together all the other work we did.

CleanShot 2026-06-05 at 21 16 41@2x


mitchellh:
I fixed the issues. There is some hacky rasterization stuff because our placement logic isn't 100% spec compliant. I don't love it. I'm going to let this sit for a bit so I can sleep on it and consider some other approaches.

PR #12953, opened 2026-06-07
terminal: support click_events=2

alexbathome:
## Context 

Implements the OSC 133 `click_events=2` such that click events on the terminal are sent to the shell where the y coordinate is sent relative to the prompt pin. 

See https://sw.kovidgoyal.net/kitty/shell-integration/#click_events=1|2:~:text=click%5Fevents,xterm further detailed explanation. 

This should close https://github.com/ghostty-org/ghostty/issues/10865

### Testing

I did some basic manual testing here via the following: 
```
ghostty  main
❯ printf '\033]133;A;click_events=1\007' && cat
^[[<0;49;4M^[[<0;48;8M^[[<0;47;12M^C

ghostty  main  5s
❯ printf '\033]133;A;click_events=2\007' && cat
^[[<0;50;1M^[[<0;49;4M^[[<0;48;9M^[[<0;48;14M^C
```

#### Notes  	  

This is my first Ghostty PR. All code here is hand-rolled, AI was used to do "smart" code searches/point me in the right direction. 

I'm actively trying to learn Zig, Ghostty, and get more involved with the community here so I'm avoiding obtuse usage of AI where possible. 

Any feedback or tips are more than welcome! Thank you! :pray:

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4222,8 +4222,8 @@ fn maybePromptClick(self: *Surface) !bool {
             // For the event, we always send a left-click press event.
             // This matches what Kitty sends.
             const key: u8, const y: u32 = switch (v) {
-                .Absolute => .{ 1, pos_vp.y + 1 },
-                .Relative => .{ 2, pos_vp.y - prompt_pin.y + 1 },
+                .absolute => .{ 1, pos_vp.y + 1 },
+                .relative => .{ 2, pos_vp.y - prompt_pin.y + 1 },
             };
             var data: termio.Message.WriteReq.Small.Array = undefined;
             const resp = try std.fmt.bufPrint(
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4218,20 +4218,24 @@ fn maybePromptClick(self: *Surface) !bool {
         // Guarded at the start of this function
         .none => unreachable,
 
-        .click_events => {
+        .click_events => |v| {
             // For the event, we always send a left-click press event.
             // This matches what Kitty sends.
+            const key: u8, const y: u32 = switch (v) {
+                .Absolute => .{ 1, pos_vp.y + 1 },
+                .Relative => .{ 2, pos_vp.y - prompt_pin.y + 1 },
+            };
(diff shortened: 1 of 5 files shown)


pluiedev:
Vouched in #12954

src/Surface.zig
```diff
@@ -4218,20 +4218,24 @@ fn maybePromptClick(self: *Surface) !bool {
         // Guarded at the start of this function
         .none => unreachable,
 
-        .click_events => {
+        .click_events => |v| {
             // For the event, we always send a left-click press event.
             // This matches what Kitty sends.
+            const key: u8, const y: u32 = switch (v) {
+                .absolute => .{ 1, pos_vp.y + 1 },
+                .relative => .{ 2, pos_vp.y - prompt_pin.y + 1 },
```
pluiedev: We might want to avoid any potential overflows/underflows with this one. The surrounding logic doesn't scream to me that `pos_vp.y + 1 > prompt_pin.y` always holds.

pluiedev (commented):
Looks great overall except for one small thing

src/Surface.zig
alexbathome: Thanks @pluiedev ! 

I've updated this code to use [Zig's Saturated arithmetic](https://ziglang.org/documentation/0.10.1/#Operators:~:text=Saturating%20Addition%2E) to hopefully fix this issue 🙏 

I was considering breaking this out into a separate function and testing the max/min values here but: 
* I wasn't certain where the best place for a function (or a test - I noticed Surface.zig has no tests) 
* I thought it might be a bit overkill for a 2 character diff 🤔 

Let me know what you think! 🙏

pluiedev (approved):

src/Surface.zig
jcollie: The fact that Surface.zig doesn’t have any existing tests is even more reason to add tests for this change.

mitchellh: We've been slowly extracting more business logic out of Surface.zig to test but as it stands this is still a tricky one. I think for this diff this is fine. I'll take the saturated arithmetic.

PR #12956, opened 2026-06-08
lib-vt: add pwd_changed callback for OSC 7/9/1337

c0x0o:
Previously the libghostty-vt stream handler dropped .report_pwd as a no-op, so embedders never saw shell-reported cwd changes and the terminal's pwd field was never populated from escape sequences.

Wire the action to setPwd and expose a pwd_changed callback analogous to title_changed via GHOSTTY_TERMINAL_OPT_PWD_CHANGED. The payload is passed through unparsed; embedders read it with ghostty_terminal_get and decode any URI scheme themselves.

This is proposed in [discussion#12927](https://github.com/ghostty-org/ghostty/discussions/12927)


c0x0o:
I'v been test it for a while in our [terminal emulator](https://lingterm.app), which is based on libghostty-vt.

mitchellh (approved):
Looks great. Thanks.

Issue #12512, opened 2026-04-29
macOS: Search bar covers up matches


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9790

<div type='discussions-op-text'>

<sup>Originally posted by **joeshaw** December  2, 2025</sup>
### Issue Description

The search bar can completely occlude a search result and you'd never know it.  Here's an example as I search for "palette".  The currently-highlighted match is completely behind the bar.

<img width="585" height="118" alt="image" src="https://github.com/user-attachments/assets/b0843f1b-62ff-41fc-9bd3-20f1622d07e7" />

I understand that I can drag the bar around, but it would be nice if the viewport scrolled so that it didn't appear behind the search bar. I guess this might be tricky if the bar is at the bottom, but it's at least doable if it's at the top.

Developer's note:

You can also implement something like the "reveal animation" that most of the native apps have when searching.

### Expected Behavior

I expect Ghostty to behave more like a native Mac app with regard to search behavior.

### Actual Behavior

See above.

### Reproduction Steps

Mostly mentioned above.

### Ghostty Version

```text
Ghostty 1.3.0-main+34fd1dd0f

Version
  - version: 1.3.0-main+34fd1dd0f
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS 26.1 (Tahoe)

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
Default config
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

PR #12960, opened 2026-06-08
surface(macos): offset search overlay when occluding results

alexbathome:
# Context

This PR resolves issue #12512

When searching it is possible for the search result to be matched "under" the search bar itself. Leaving a confusing display to the user. 

This PR, modifies the `search_selected` action/`SearchSelected` message such that we can pass back the search results viewport coordinates. This allows us to calculate an "avoidance" offset in the Swift UI code and "offset" the Search Bar's location on the display revealing the matched result underneath. 

A more pragmatic approach might be to detect the intersection between the search bar's rect and the result's rect and then using the Swift anchor/corners to move the bar around the display. I opted to use this offset approach because I find it a bit more aesthetic. 

See example screen recording here: 

https://github.com/user-attachments/assets/a66412a1-4824-4129-ae78-61b1de069c12

# Notes

1. I am a bit uncertain about making a modification like this to the Action union because of how it impacts ABI. The comment guide explains that new actions should be appended to the end of the enum - however since this is a mutation of the existing `search_selected` key what is the preferred approach here? (The mutation definitely breaks ABI regardless) 
2. **AI USAGE**: I have never written swift before and I couldn't get the LSPs, or the build.nu to give me any real valuable output - I'll have to sit down and work through some Swift dev in my own time to gather that skill set, in this case most of **the Swift code was generously provided by claude**. If any other contributors are keen on Swift and don't mind reaching out to give me some pointers please do! 
3. I don't quite have a solid approach for it yet, but I was pondering a method to pass arbitrary Rect/Point data through the Surface message box as a :sparkles: _generic action_ :sparkles: perhaps? This would at least give us a path forward if there's ever the need for this sort of reactive UI to screen/terminal content - Someting to think about.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -872,10 +872,10 @@ typedef struct {
 // apprt.action.SearchSelected
 typedef struct {
   ssize_t selected;
-  uint32_t start_x;
-  uint32_t start_y;
-  uint32_t end_x;
-  uint32_t end_y;
+  ssize_t start_x;
+  ssize_t start_y;
+  ssize_t end_x;
+  ssize_t end_y;
 } ghostty_action_search_selected_s;
 
 // terminal.Scrollbar
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -872,6 +872,10 @@ typedef struct {
 // apprt.action.SearchSelected
 typedef struct {
   ssize_t selected;
+  uint32_t start_x;
+  uint32_t start_y;
+  uint32_t end_x;
+  uint32_t end_y;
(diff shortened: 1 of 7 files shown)


bo2themax:
Hi Alex, I’m already working on this. The terminal part is under review. Claude did the most part. You are welcome to leave a review at #12942.

alexbathome:
> Hi Alex, I’m already working on this. The terminal part is under review. Claude did the most part. You are welcome to leave a review at #12942.

Oh excellent work! Thanks @bo2themax ! I'll close this in favour of #12942

PR #12963, opened 2026-06-08
License: change copyright year to 2026

GautierPicon:
updates the copyright year of the license from 2024 to 2026


jcollie:
The copyright year should reflect the first year that the work was copyrighted, and do not need to be updated each year.

PR #12989, opened 2026-06-11
build: skip tip release when nothing release-relevant changed since last tip

bo2themax:
There's a [short discussion](https://discord.com/channels/1005603569187160125/1005603569711452192/1504521818495651910) about this in Discord, I don't any existing discussion or issues.

### AI Disclosure

Claude implemented this, changes are looking good to me, but I could be wrong.


8bittts:
or just use one of the many [alternatives](https://github.com/Uzaaft/awesome-libghostty), some of which always run on the `tip`. the version that i manage pulls the latest tip and releases it [almost](https://yen.chat) every day.

PR #12995, opened 2026-06-12
Surface: open file paths with trailing :line[:col] suffixes

jamesarch:
Fixes a long-standing paper cut discussed extensively in #1972: clicking a
file path that carries a `:line[:col]` suffix (the standard output format
of most compilers, e.g. `src/main.zig:42:10` from zig/clang, or
`file.scala:47:70` as reported [here](https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-2626961340))
does nothing today. The URL regex matches the full string including the
numeric suffix, so the path fails to resolve on disk and the click fails
silently (or hands a nonexistent path to the system opener).

With this change, when the clicked text doesn't resolve to a real file
as-is, we strip a trailing `:line[:col]` (validated to be all digits) and
try again, for both relative paths (resolved against the terminal pwd,
as before) and absolute paths. Exact matches take precedence, so files
whose names genuinely contain a numeric suffix still open correctly.
If neither resolves, behavior is unchanged.

The parsing helper returns the line/column components as well; they are
unused by the system opener today but are the natural parse result of
this format and are relevant to the custom-handler ideas discussed in
#9546.

Includes unit tests for the suffix parser. Tested manually on macOS
against compiler output and `grep -rn` output. The full test suite was
also run on Linux (`nix develop -c zig build test`); no new failures
versus unpatched main in the same environment, and the new unit test
passes on both platforms.

Disclosure: I used Claude Code to assist with this change, and I have
reviewed and tested every line personally.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -4429,7 +4429,28 @@ fn processLinks(self: *Surface, pos: apprt.CursorPos) !bool {
             const resolved_path = try self.resolvePathForOpening(str);
             defer if (resolved_path) |p| self.alloc.free(p);
 
-            const url_to_open = resolved_path orelse str;
+            // Compilers and tools commonly suffix file paths with
+            // `:line[:col]` (e.g. `src/main.zig:42:10`). If the text
+            // as-is doesn't resolve to a real file but the path with
+            // the suffix stripped does, open the stripped path so that
+            // the click does something useful instead of silently
+            // failing. Exact matches above take precedence so that
+            // files whose names actually contain a numeric suffix
+            // still open correctly.
+            const stripped_path: ?[]const u8 = stripped: {
+                if (resolved_path != null) break :stripped null;
+                const parsed = input.Link.parseFilePath(str);
+                if (parsed.line == null) break :stripped null;
+                if (std.fs.path.isAbsolute(parsed.path)) {
+                    std.fs.accessAbsolute(parsed.path, .{}) catch
+                        break :stripped null;
+                    break :stripped try self.alloc.dupe(u8, parsed.path);
+                }
+                break :stripped try self.resolvePathForOpening(parsed.path);
+            };
+            defer if (stripped_path) |p| self.alloc.free(p);
+
+            const url_to_open = resolved_path orelse stripped_path orelse str;
             try self.openUrl(.{ .kind = .unknown, .url = url_to_open });
         },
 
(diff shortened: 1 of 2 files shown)


jamesarch:
Gentle ping on this PR since it has been waiting for a few weeks.

The Vouch and Milestone checks are green now, but the `Test` and `Nix` workflows still show `action_required`, so I believe they need a maintainer to approve/run them.

Could a maintainer please approve the workflows or let me know if anything should change first? Thanks!

jamesarch:
Rebased onto current `main` (`0f166c80c`). The branch was 453 commits behind; the rebase applied cleanly with no conflicts and the diff is unchanged — 2 files, +89/−1.

**On why this gap still exists**

#1972 is closed as completed, and rightly so: #4713 shipped clickable paths in 1.1.0. But the `:line[:col]` case was reported inside that thread several times and never covered by it:

- [#1972 (comment)](https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-2627721738) — `file.scala:4` / `file.scala:4:10` open in iTerm, do nothing in Ghostty
- [#1972 (comment)](https://github.com/ghostty-org/ghostty/issues/1972#issuecomment-3211614857) — `app/folder/file.rb:1` **highlights on cmd+hover but the click does nothing**

That second description is exactly this PR's scope. Detection already works — the regex in `config/url.zig` matches the suffixed path — but `Surface.zig` then fails to resolve it and the click silently does nothing. This change adds the fallback in the open path only: exact match is still tried first, so a file genuinely named `foo:1` keeps working, and the suffix is only stripped when the literal path does not resolve.

Recent work in that thread (e.g. bkircher's regex fixes) is on the detection side and does not overlap.

**Verification**

- `git rebase origin/main` — clean, no conflicts
- `zig fmt --check src/input/Link.zig src/Surface.zig` — clean
- `zig ast-check` on both files — clean
- 5 unit tests for `parseFilePath` cover relative/absolute paths, `magnet:` URIs and non-numeric suffixes

I could not run `zig build test` locally: this machine only has Command Line Tools, and `build.zig` constructs the iOS xcframework before option parsing, so it fails with `DarwinSdkNotFound` even for `zig build --help`. So the above is the honest limit of what I verified — the test suite has not run against the rebase.

Which brings me back to the ask from July: `Test` and `Nix` are still `action_required`, so CI has never run on this branch at all. Could a maintainer approve the workflows? If they surface anything I will fix it promptly.

00-kat:
Closing as OP's been denounced/banned (https://github.com/ghostty-org/ghostty/pull/12984#issuecomment-5187314709).

PR #13008, opened 2026-06-13
gtk: remove deprecated style-hc.css resources, use media query

neheb:
The libadwaita warning says style-hc.css is deprecated and style.css with media queries should be used instead.

Move the .transparent rule from style-hc.css and
style-hc-dark.css into style.css wrapped in a
@media (prefers-contrast: more) query, then remove the separate high-contrast resource files and their entries in the gresource build.


jcollie:
We still support older libadwaita versions so we cannot use media queries yet. People that have the newer libadwaita will need to ignore that warning.

neheb:
OK.

Issue #12949, opened 2026-06-07
macOS: title bar tabs partially disappear when creating new tab from "Show All Tabs"


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/12944

<div type='discussions-op-text'>

<sup>Originally posted by **vwkd** June  6, 2026</sup>
### Issue Description

With the custom title bar, tabs disappear when creating a new tab from the "Show All Tabs" view.

https://github.com/user-attachments/assets/9d124e11-05cf-4538-8447-8515591f6717

### Expected Behavior

Tabs should not disappear.

### Actual Behavior

Tabs disappear.

### Reproduction Steps

1. Open Ghostty.
2. Create some tabs.
3. Right click on some tab and choose "Show All Tabs"
4. Click the big "+"

### Ghostty Version

```text
Ghostty 1.3.1

Version
  - version: 1.3.1
  - channel: stable
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS Tahoe 26.5.1 (25F80)

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
macos-titlebar-style = tabs
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

PR #13017, opened 2026-06-14
overrode function syncAppearance in order to update the tab bar as well

sahilkmishra:
#12949 
I'm unsure about the performance implications of overriding syncAppearance as opposed to other methods(ie setupKVO or setupTabGroupObservation) but I found that this shows the tab group correctly.

### DIFF
--- a/macos/Sources/Features/Terminal/Window Styles/TitlebarTabsTahoeTerminalWindow.swift
+++ b/macos/Sources/Features/Terminal/Window Styles/TitlebarTabsTahoeTerminalWindow.swift
@@ -54,7 +54,12 @@ class TitlebarTabsTahoeTerminalWindow: TransparentTitlebarTerminalWindow, NSTool
     // Called after new tab finishes adjusting and setupTabBar is called in order to prevent Tab Bar hiding/size bug that occurs with some interactions with Mac UI
     override func syncAppearance(_ surfaceConfig: Ghostty.SurfaceView.DerivedConfig) {
         super.syncAppearance(surfaceConfig)
-        setupTabBar()
+        DispatchQueue.main.async {
+            // HACK: wait a tick before doing anything, to avoid edge cases during startup... :/
+            // If we don't do this then on launch windows with restored state with tabs will end
+            // up with messed up tab bars that don't show all tabs.
+            self.setupTabBar()
+        }
     }
 
     override func becomeMain() {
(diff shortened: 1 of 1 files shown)


bo2themax:
!vouch

macos/Sources/Features/Terminal/Window Styles/TitlebarTabsTahoeTerminalWindow.swift
```diff
@@ -51,6 +51,11 @@ class TitlebarTabsTahoeTerminalWindow: TransparentTitlebarTerminalWindow, NSTool
         self.toolbar = toolbar
         toolbarStyle = .unifiedCompact
     }
+    override func syncAppearance(_ surfaceConfig: Ghostty.SurfaceView.DerivedConfig) {
+        super.syncAppearance(surfaceConfig)
+        setupTabBar()
+    }
+
```
bo2themax, line 64: SwiftLint failed here

```suggestion
    }
```

bo2themax (changes requested):

macos/Sources/Features/Terminal/Window Styles/TitlebarTabsTahoeTerminalWindow.swift
```diff
@@ -51,6 +51,11 @@ class TitlebarTabsTahoeTerminalWindow: TransparentTitlebarTerminalWindow, NSTool
         self.toolbar = toolbar
         toolbarStyle = .unifiedCompact
     }
+    override func syncAppearance(_ surfaceConfig: Ghostty.SurfaceView.DerivedConfig) {
+        super.syncAppearance(surfaceConfig)
+        setupTabBar()
```
bo2themax: Would be nice if we also added comments here explaining why this is necessary.

bo2themax (changes requested):

bo2themax:
I tried, although the change looks clean enough but the "first" second tab still has issues


https://github.com/user-attachments/assets/ea15dffc-1d22-4513-bc0b-49dc0d6c0548

I'm on macOS 27 btw, if your local run works fine(with the same flow) on Tahoe this could be a beta issue, then i'm willing to approve it now

sahilkmishra:
I get that behavior in Tahoe with the first second tab, I'll keep looking at the class later today. I don't know if using async here would be the right thing to do but I'll keep looking.

sahilkmishra:
Ok I used the async bit of code that's used in another portion of this class that's labelled hack but this resolved the first second tab error for me!

bo2themax (approved):
Works now, thanks

Issue #12852, opened 2026-05-29
Implement the Kitty Drag-and-Drop Protocol (OSC 72)


pluiedev:
Kovid strikes again with yet another new protocol: https://sw.kovidgoyal.net/kitty/dnd-protocol/

The protocol is not yet stable, but I figure we can start writing a parser and then work on integrating it with the apprt. As with OSC 5522, some kind of state machine would be needed to concatenate all the received packets together for large payload streams.

AnirudhG07:
Even [yazi #4005](https://github.com/sxyazi/yazi/pull/4005) now supports this feature for Kitty. Would be nice to see it coming in for Ghostty!

PR #13029, opened 2026-06-16
Kitty dnd parser

ajr-khll:
(#12852)
I opened a discussion to work on the new kitty dnd protocol and implementing it for Ghostty.  I was told to work on the parser but not to hook  up any actions to it yet. So, that's what I did! Largely based the format on kitty_clipboard_protocol.zig, and used Claude Opus 4.8  (Claude Code) for writing tests and some structural guidance early on. Would love to get started on adding actions as well!


src/terminal/osc/parsers/kitty_dnd_protocol.zig
```diff
@@ -0,0 +1,333 @@
...
+            'E' => .drag_offer_error,
+            'k' => .uri_list_data,
+            'q' => .query,
+            else => null,
+        };
+    }
+};
+
+/// Metadata keys defined by the protocol. Keys are case-sensitive: `x` and `X` are distinct.
+pub const Option = enum {
```
jcollie, line 83: I'd recommend adding some doc comments on what all of these options mean so that LSPs can display that info and you don't need to constantly refer back to the protocol spec.

src/terminal/osc/parsers/kitty_dnd_protocol.zig
```diff
@@ -0,0 +1,333 @@
...
+    /// The terminator used for this OSC, so any response can match it.
+    terminator: Terminator,
+
+    pub fn readOption(self: OSC, comptime key: Option) ?key.Type() {
+        return key.read(self.metadata);
+    }
+};
+
+/// Values for the `t` (event type) metadata key.
+pub const EventType = enum {
```
jcollie, line 28: Some doc comments would be nice here too.

jcollie (approved):
Be nice to have some more comments but otherwise this looks good. Didn't look at the spec, only looking at the Zig code here.

src/terminal/osc/parsers/kitty_dnd_protocol.zig
ajr-khll: Added doc comments throughout

mitchellh:
Looks like a great starting point thank you.

mitchellh (approved):

ajr-khll:
Just completed drag and drop *into* the terminal, for GTK. Yet to implement drag and drop out of the terminal. took some help from Claude again.

bo2themax (approved):
I don't see any Swift or header file changes, but I'll approve it just in case it blocks the merge in the future.

jcollie (changes requested):
I'm going to have to block this PR. You were asked to submit the protocol parsing _only_ but you've added the whole implementation as well. If you roll back the implementation changes we can get this approved, and then in a _separate_ PR you can submit the GTK implementation.

ajr-khll:
Rolled back to just the parser implementation. I'll make a separate PR for the GTK soon.

jcollie (approved):

PR #13037, opened 2026-06-17
Add macOS clipboard copy toast

jagoff:
## Summary

This adds macOS support for the existing `app-notifications = clipboard-copy` behavior.

GTK already shows an in-app Adwaita toast when Ghostty writes to the standard clipboard. macOS went through the same clipboard write path, but it did not surface any in-app feedback. This PR emits a macOS-side internal `didWriteClipboard` notification from the clipboard write callback and renders a short-lived SwiftUI overlay on the matching surface.

The macOS behavior intentionally follows the GTK semantics:

- only standard clipboard writes show feedback
- selection clipboard writes are ignored
- `app-notifications.clipboard-copy` gates the feedback
- non-empty text shows `Copied to clipboard`
- empty text shows `Cleared clipboard`
- the overlay uses a 3 second timeout, matching GTK's `adw.Toast` timeout

The overlay uses native SwiftUI material, system text styling, and primary foreground color so the typography and colors follow the macOS runtime instead of introducing a custom visual style.

## Changes

- Expose `app-notifications` through the macOS Swift config wrapper.
- Add a macOS internal `didWriteClipboard` notification.
- Post that notification when the macOS runtime writes to the standard clipboard.
- Render a transient clipboard feedback overlay in `SurfaceWrapper`.
- Add macOS config tests for default and disabled `clipboard-copy` notification behavior.
- Update config docs so `app-notifications` is no longer documented as GTK-only.

## Validation

- `git diff --check`

I could not run the full macOS test/build command in this local environment because `macos/build.nu` requires `nu`, and `xcodebuild` requires a full Xcode install while this machine only has Command Line Tools selected.

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -413,6 +413,19 @@ extension Ghostty {
                     guard let type = NSPasteboard.PasteboardType(mimeType: item.mime) else { continue }
                     pasteboard.setString(item.data, forType: type)
                 }
+
+                if location == GHOSTTY_CLIPBOARD_STANDARD {
+                    let text = contentArray.first(where: { $0.mime == "text/plain" })?.data ?? ""
+                    NotificationCenter.default.post(
+                        name: Ghostty.Notification.didWriteClipboard,
+                        object: surface,
+                        userInfo: [
+                            Ghostty.Notification.DidWriteClipboardMessageKey: text.isEmpty
+                                ? "Cleared clipboard"
+                                : "Copied to clipboard",
+                        ]
+                    )
+                }
                 return
             }
 
--- a/macos/Sources/Ghostty/Ghostty.Config.swift
+++ b/macos/Sources/Ghostty/Ghostty.Config.swift
@@ -189,6 +189,14 @@ extension Ghostty {
             return .init(rawValue: v)
         }
 
+        var appNotifications: AppNotifications {
+            guard let config = self.config else { return .all }
+            var v: CUnsignedInt = 0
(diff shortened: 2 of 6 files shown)


jagoff:
Follow-up note on visual parity with GTK:

I checked the GTK implementation after opening this. The GTK runtime uses `Adw.ToastOverlay` plus `Adw.Toast`, not custom Ghostty CSS. The libadwaita docs describe the toast node as containing a `label.heading`; Ghostty only sets the message and timeout.

This macOS implementation mirrors the behavior and uses native platform styling rather than trying to hardcode libadwaita colors. I updated the SwiftUI text style from `.caption.weight(.semibold)` to `.body.weight(.semibold)` so it is closer to libadwaita's `label.heading` semantics: default-size heading text, not a smaller caption.

jagoff:
Update after the latest push (`b5abd30`):

The current branch no longer tries to match GTK's visual timeout/color exactly. It keeps the GTK clipboard semantics:

- standard clipboard writes only
- selection clipboard writes ignored
- gated by `app-notifications.clipboard-copy`
- same user-facing messages: `Copied to clipboard` / `Cleared clipboard`

The macOS presentation is now intentionally platform-specific: a compact native SwiftUI overlay with muted orange text and a two-second display duration.

I could not update the PR description via `gh pr edit` because the current token lacks `updatePullRequest` access.

jagoff:
Update after latest push (`ba42dd4`): the macOS overlay text color is now muted green (`Color(red: 0.38, green: 0.58, blue: 0.32)`) instead of muted orange. Behavior is otherwise unchanged.

Issue #5685, opened 2025-02-11
macOS: quick terminal auto hiding dock not restored when keyboard used to switch spaces


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/5617

<div type='discussions-op-text'>

<sup>Originally posted by **criphc** February  6, 2025</sup>
the issue: dock not restored to not autohidden after closing quick terminal. 

how to reproduce: 
1. set dock to the right and "Automatically hide and show the Dock" to false
2. open quick terminal in a desktop space
3. switch space to a native fullscreen window with the quick terminal open
4. close the quick terminal
5. switch back to desktop space

example: 

https://github.com/user-attachments/assets/cbcd633b-bcd1-4b01-b0f5-c9a69f85f905

</div>

mitchellh:
Important detail for reproduction: you must be moving to a _fullscreen space_. If you move to a desktop space this bug does not occur.

liby:
After several attempts to fix this, including tracking fullscreen states, space changes, and adding various timing mechanisms, the problem remains stubbornly difficult to solve reliably. This has led me to question the fundamental design decision:

 - Should Quick Terminal affect the Dock's visibility at all? Given that Quick Terminal is not a full-screen window but rather a floating/overlay window, it seems questionable whether it should modify the system-wide Dock state.

Perhaps we should reconsider this design decision and let Quick Terminal coexist with the Dock without modifying its visibility? The complexity and unreliability of the current approach suggest this might be a better direction.

mitchellh:
> Perhaps we should reconsider this design decision and let Quick Terminal coexist with the Dock without modifying its visibility? The complexity and unreliability of the current approach suggest this might be a better direction.

This is what iTerm2 does, perhaps for this reason. I'd be fine looking into that. I think it looks a lot worse but if we can't fix this issue I agree a better approach is warranted.

I'll take a look too to see if there is anything to be done here.

ephemer:
@mitchellh @liby I am running into the weird behaviour that the quick terminal hides the dock and stumbled across this issue. I'm on the latest macOS 26.0 beta.

I find it unnecessary and distracting that simply opening the quick terminal hides the dock. This issue seems to be about the quick terminal not un-hiding the dock, but I'd like to make Quick Terminal not hide it in the first place. Searching for "dock" in the Option Reference only turns up one result regarding the dock icon (FWIW I'd rather not see any icon in the dock at all).

[...]

PR #13039, opened 2026-06-18
macos: quick-terminal - dock hide/unhide logic changes

exlee:
# Quick Terminal Dock / Space Handling

## Notes

- This PR addresses #5685

- ⚠️ Accessibility Permissions sometimes doesn't transfer well requiring deleting Ghostty from Settings Panel and re-adding again.
- Exposure of a logging function to Swift is an extra and could be separated
- I have another PR in preparation that allows enabling/disabling dock. It should be stacked on top of this PR though.

## LLM disclaimer

- Architecture (2-tiered state machine, state transitions, test shape) is end-to-end owned and developed by me
- Provisioned code was reviewed by me in multiple passes
- Edge cases problems were solved by me directly
- I tested manually repeatedly when working on PR on rebuilt instances; there are no edge cases known to me at this point
- **Codex/ChatGPT 5.5 contribution**: guided code generation, initial code analysis, code completion analysis, grooming contextual information visible below

### Initial State

Quick Terminal could hide the Dock when shown in a position that conflicts with
the Dock, but the restore path was tied mostly to quick-terminal toggle
and window focus events.

That left a few awkward Space-change cases:

- Keyboard Space changes could leave the Dock hidden because the expected focus
  or animation callbacks did not restore it.
- Restoring the Dock immediately on Space change could cause a visible Dock
  flash when the same Quick Terminal window was about to regain focus.
- Fullscreen Spaces made Dock visibility ambiguous: the Dock is hidden there
  even when user Dock auto-hide is off, so using "Dock is hidden" as the only
  signal could make Quick Terminal mutate user Dock settings at the wrong time.

### State Machine

The Dock logic now has two layers.

The first layer is the desired state:

- `shouldBeHidden == true`: Quick Terminal currently wants the Dock hidden.
- `shouldBeHidden == false`: Quick Terminal no longer wants to own Dock hiding.

The second layer is the managed side-effect state:

- `managedHidden == true`: Quick Terminal changed global Dock auto-hide and is
  responsible for restoring it later.
- `managedHidden == false`: Quick Terminal has not changed global Dock auto-hide.

The pure reducer is `QuickTerminalDockState`:

- If the Dock is visible and `shouldBeHidden` becomes true, transition to
  `hide` and mark `managedHidden`.
- If Quick Terminal previously managed hiding and `shouldBeHidden` becomes
  false on a normal Space, transition to `show` and clear `managedHidden`.
- If the Dock is already hidden by the user, Quick Terminal does not mark it as
  managed and does not restore it later.
- If the active Space is fullscreen, Quick Terminal never performs the `show`
  transition. It preserves `managedHidden` so the restore can happen later after
  returning to a normal Space.

```mermaid
flowchart LR
    request["setShouldHide(value) or apply()"]
    desired["Update / read shouldBeHidden"]
    hidden{"Dock hidden?\nDock.autoHide || fullscreen"}
    wantsHide{"shouldBeHidden?"}
    managed{"managedHidden?"}
    fullscreen{"fullscreen Space?"}

    hide["transition: hide\nmanagedHidden = true\nDock.autoHide = true"]
    show["transition: show\nmanagedHidden = false\nDock.autoHide = false"]
    defer["transition: none\nkeep managedHidden = true\nrestore later"]
    noop["transition: none\nno Dock side effect"]

    request --> desired --> wantsHide

    wantsHide -- true --> hidden
    hidden -- false --> hide
    hidden -- true --> noop

    wantsHide -- false --> managed
    managed -- false --> noop
    managed -- true --> fullscreen
    fullscreen -- false --> show
    fullscreen -- true --> defer
```

`HiddenDock` wraps that reducer and performs the AppKit/global Dock side effects
only when the reducer returns `hide` or `show`.

#### Added

- Debug logging for requested, applied, and skipped Dock transitions.
- Focus holding across Space changes: when Quick Terminal still owns focus /
  Dock hiding, it immediately refocuses the same Quick Terminal window instead
  of waiting for the delayed restore path.
- Fullscreen-safe Dock restore behavior: fullscreen Spaces are treated as
  read-only for Dock restore, so Quick Terminal never unhides the Dock while
  fullscreen is active.

#### Changed

- `HiddenDock` now delegates hide / show decisions to `QuickTerminalDockState`,
  a testable pure state machine.
- `HiddenDock` only performs AppKit/global Dock side effects when the reducer
  returns an explicit `hide` or `show` transition.
- `windowDidResignKey` no longer restores the Dock during transient focus loss
  from Space movement, avoiding a flash before the same Quick Terminal window is
  re-keyed.

#### Maintenance

- Added unit tests for the Dock reducer:
  - user-hidden Dock is not managed by Quick Terminal
  - normal Spaces can hide and show the Dock
  - fullscreen Spaces never show the Dock and defer restoration
  - pending `shouldBeHidden` is applied if the Dock later becomes visible

### Rationale

| Area | Rationale |
| --- | --- |
| Two-layer state machine | Quick Terminal needs to distinguish desired state from owned side effects. `shouldBeHidden` records whether Quick Terminal wants the Dock hidden; `managedHidden` records whether Quick Terminal actually changed global Dock auto-hide and must restore it later. This avoids interfering with users who already keep the Dock hidden. |
| Fullscreen defer | In fullscreen Spaces, macOS hides the Dock even when Dock auto-hide is disabled. Quick Terminal therefore never performs the show/unhide transition while fullscreen is active; it defers restoration until a normal Space is active again. |
| Focus holding | `quickterm -> Space change` can briefly produce a non-focused Quick Terminal window. If Quick Terminal still owns focus / Dock hiding, it immediately re-keys the same window instead of restoring the Dock and causing a flash. |
| Debug logging | Dock state bugs are hard to inspect from Cocoa alone. The logging bridge exposes QuickTerminal-specific Dock transitions through the existing Ghostty logger, making requested/applied/skipped transitions visible during manual testing. |

### DIFF
--- a/src/apprt/embedded.zig
+++ b/src/apprt/embedded.zig
@@ -23,6 +23,7 @@ const Config = configpkg.Config;
 const String = @import("../main_c.zig").String;
 
 const log = std.log.scoped(.embedded_window);
+const quick_terminal_log = std.log.scoped(.quick_terminal);
 
 pub const resourcesDir = internal_os.resourcesDir;
 
@@ -1393,6 +1394,11 @@ pub const CAPI = struct {
         if (builtin.target.os.tag.isDarwin()) {
             _ = Darwin;
         }
+    }
+
+    /// Log Quick Terminal debug messages from the embedding layer.
+    export fn ghostty_quick_terminal_debug_log(ptr: [*:0]const u8) void {
+        quick_terminal_log.debug("{s}", .{ptr});
     }
 
     /// Create a new app.
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1066,6 +1066,7 @@ GHOSTTY_API void ghostty_cli_try_action(void);
 GHOSTTY_API ghostty_info_s ghostty_info(void);
 GHOSTTY_API const char* ghostty_translate(const char*);
 GHOSTTY_API void ghostty_string_free(ghostty_string_s);
+GHOSTTY_API void ghostty_quick_terminal_debug_log(const char*);
 
 GHOSTTY_API ghostty_config_t ghostty_config_new();
 GHOSTTY_API void ghostty_config_free(ghostty_config_t);
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalController.swift
@@ -3,6 +3,12 @@ import Cocoa
 import SwiftUI
 import GhosttyKit
 
+private func quickTerminalDebugLog(_ message: String) {
+    message.withCString { ptr in
+        ghostty_quick_terminal_debug_log(ptr)
+    }
+}
+
 /// Controller for the "quick" terminal.
 class QuickTerminalController: BaseTerminalController {
     override var windowNibName: NSNib.Name? { "QuickTerminal" }
@@ -21,11 +27,16 @@ class QuickTerminalController: BaseTerminalController {
     // The active space when the quick terminal was last shown.
     private var previousActiveSpace: CGSSpace?
 
+    // The active space we last saw from NSWorkspace. This handles Space changes
+    // that don't reliably deliver key-window notifications, such as keyboard
+    // navigation between Spaces.
+    private var pendingActiveSpaceChange: CGSSpace?
+
     /// Cache for per-screen window state.
     let screenStateCache: QuickTerminalScreenStateCache
 
-    /// Non-nil if we have hidden dock state.
-    private var hiddenDock: HiddenDock?
+    /// Tracks the quick terminal's desired and managed Dock state.
+    private let hiddenDock = HiddenDock()
 
     /// The configuration derived from the Ghostty config so we don't need to rely on references.
     private var derivedConfig: DerivedConfig
@@ -91,6 +102,12 @@ class QuickTerminalController: BaseTerminalController {
             selector: #selector(windowDidResize(_:)),
             name: NSWindow.didResizeNotification,
             object: nil)
+
+        NSWorkspace.shared.notificationCenter.addObserver(
+            self,
+            selector: #selector(activeSpaceDidChange(_:)),
+            name: NSWorkspace.activeSpaceDidChangeNotification,
+            object: nil)
     }
 
     required init?(coder: NSCoder) {
@@ -101,9 +118,10 @@ class QuickTerminalController: BaseTerminalController {
         // Remove all of our notificationcenter subscriptions
         let center = NotificationCenter.default
         center.removeObserver(self)
+        NSWorkspace.shared.notificationCenter.removeObserver(self)
 
         // Make sure we restore our hidden dock
-        hiddenDock = nil
+        hiddenDock.setShouldHide(false, reason: "deinit")
     }
 
     // MARK: NSWindowController
@@ -162,7 +180,7 @@ class QuickTerminalController: BaseTerminalController {
         terminalViewContainer?.updateGlassTintOverlay(isKeyWindow: true)
 
         // Re-hide the dock if we were hiding it before.
-        hiddenDock?.hide()
+        hiddenDock.setShouldHide(position.conflictsWithDockOrientation, reason: "windowDidBecomeKey")
     }
 
     override func windowDidResignKey(_ notification: Notification) {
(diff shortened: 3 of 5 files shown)


src/apprt/embedded.zig
```diff
@@ -1395,6 +1396,11 @@ pub const CAPI = struct {
         }
     }
 
+    /// Log Quick Terminal debug messages from the embedding layer.
```
bo2themax, line 1399: You don't need add logging function here, just use OSLog like in other places in Swift.

bo2themax (changes requested):
First of all, it doesn't fix the issue for me using this branch. 

Then it lacks proper documentation about why in the code, it’s a lot of maintenance for us to understand it. 

And you removed the per-screen check without explaining why, it will now automatically hide the dock even if I open the quick terminal on a screen without a dock. 

I suggest you check all the lines you deleted or changed; you can blame the changes and check whether your changes would break anything.

Thanks for trying tho!

### NOTE

I sense your pr description is generated by AI which contains a lot noise, please take a look at our AI policy, and edit it yourself. It will be closed if it keeps like this.

src/apprt/embedded.zig
exlee: Damn GitHub notifications! I was browsing my open PRs and noticed change requests, will address this soon :)

exlee: Regarding the log: I was following convention setup in the `embedded.zig`, earlier there's:
```
const log = std.log.scoped(.embedded_window);
const quick_terminal_log = std.log.scoped(.quick_terminal);
```

On top of it since I'm not a maintainer I looked for prior art, there's not much OSLog logging, and most logging I found were in Zig with a scope. I can rewrite it, but would appreciate guidance (at least pointing to code sample to take pattern from).

exlee: Regarding the log: I was following convention setup in the `embedded.zig`, earlier there's:
```
const log = std.log.scoped(.embedded_window);
const quick_terminal_log = std.log.scoped(.quick_terminal);
```

On top of it since I'm not a maintainer I looked for prior art, there's not much OSLog logging, and most logging I found were in Zig with a scope. I can rewrite it, but would appreciate guidance (at least pointing to code sample to take pattern from).

exlee:
> First of all, it doesn't fix the issue for me using this branch.

That's interesting, because I was running for 2 months this branch and I hadn't had any issues whatsoever with it (I moved to my own terminal multiplexer since then, so not a Ghostty issue).

Could you be also more precise on why exactly it doesn't fix it? Too much hiding? Not enough hiding? Edge cases. "doesn't fix for me" could be slightly more informative 😄. 
 
> Then it lacks proper documentation about why in the code, it’s a lot of maintenance for us to understand it.

Could you point where it's underdocumented? I try not to be overly verbose when developing, and in my personal opinion many places have descriptive names enough.

Diagram and context I also find clear enough and is verbose for that exact purpose - to show that previously simple hide/restore logic after change takes into the consideration whether Dock was actually hidden or not (because you can't hide/restore on fullscreen etc.)

Of course I could put ASCII diagrams in the code itself, but I haven't seen prior art, so I decided not to do it with initial PR.

> And you removed the per-screen check without explaining why, it will now automatically hide the dock even if I open the quick terminal on a screen without a dock.

It was removed because I check for conflict in places where screen isn't available. (There are 3 checks, and was one where screen was in the scope). I missed the fact that actual screen is checked for the issue you pointed out. I'll update the PR.

> I suggest you check all the lines you deleted or changed; you can blame the changes and check whether your changes would break anything.

As I mentioned in original description and in the first paragraph - I did test this extensively and ran the build for 2 months. It worked for me without any side effects. While I agree that non checking the screen for Dock conflict is an omission I'd appreciate more information as I don't find "checking lines if you didn't break anything" constructive.

> ### NOTE
> I sense your pr description is generated by AI which contains a lot noise, please take a look at our AI policy, and edit it yourself. It will be closed if it keeps like this.

I did write that:
>> Codex/ChatGPT 5.5 contribution: guided code generation, initial code analysis, code completion analysis, **grooming contextual information visible below**

...so it's not really a secret ;-) 

---

Anyway, thanks for checking and pointing that out, I'll look into fixing the screen omission issue. I'll rebase the PR by the way in case it got too stale. And sorry again for the late reply. GitHub notifications failed me 😄

src/apprt/embedded.zig
bo2themax: `ghostty_quick_terminal_debug_log` this is just for macOS, you can just use Logger there in swift.

PR #13048, opened 2026-06-19
core: fix PageList.scroll using wrong type for row offset

dkinzler:
Changes `PageList.scroll` to use `usize` as the type for the number of remaining rows when trying to find a specific row in the list of pages. Previously `CellCountInt=u16` was used, but the row offset can be larger than `maxInt(u16)=65535`, in which case the scroll operation would fail.

Fixes #13000 where scrolling does not work correctly in the GTK app with a large scrollback history. E.g. nothing happens when you click in the middle of the scrollbar. Note that this problem only manifests at millions of lines of history instead of already when the row offset overflows `u16`. The reason for that is that GTK often emits multiple scroll adjustments, the first of which is usually close enough to the current position to not overflow `u16`. The other scroll adjustments are then handled by the fast path in `PageList.scroll` that works correctly.

#### AI Disclosure
No AI was used.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -2582,18 +2582,15 @@ pub fn scroll(self: *PageList, behavior: Scroll) void {
             if (n < midpoint) {
                 // Iterate forward from the first node.
                 var node_it = self.pages.first;
-                var rem: size.CellCountInt = std.math.cast(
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
Thank you!

PR #13053, opened 2026-06-20
input: modifier-aware keybind resolution and macOS fixes

qappell:
This PR corrects how Ghostty matches keybinds against keypresses by accounting for modifiers that the keyboard layout consumes to produce a character. Interest in this was spurred by trying to get `ctrl+shift+=` to work on Linux (discussion #7243).

The core resolution logic now checks the physical key and unshifted character with raw modifiers before matching against the produced character with consumed modifiers masked out. When the produced character differs from the unshifted character only in case, the raw modifiers are retained, so `ctrl+shift+a` correctly does not trigger `ctrl+a`. I also switched Unicode case folding from full to simple, since full case folding can expand a single codepoint to multiple. I improved the documentation in `src/config/Config.zig` to explain keybinds and the new flow meticulously.

While GTK exposes consumed modifiers directly on Linux, macOS lacks a direct API, so consumed modifiers are computed via a heuristic in `NSEvent.ghosttyKeyEvent`. `ghosttyCharacters` now correctly derives the produced character when Command or Control is active, and `MenuShortcutManager` matches the unshifted character with full modifiers first before falling back to the produced character with consumed modifiers stripped. Consumed bindings bypass `interpretKeyEvents` to avoid starting unfinished dead-key compositions, and a guard prevents AppKit from fuzzy-matching away a meaningful shift when matching menu equivalents. I also noticed that when a binding triggers a menu action during performKeyEquivalent, keyDown is bypassed, so I made Ghostty record the keypress separately to ensure it shows in the terminal inspector.

I verified that all default keybinds work as expected on macOS Tahoe and NixOS (both using Ghostty's Nix environment) with the US English layout, and I tested functionality across several other layouts, mainly European languages, which should be okay. Custom user-defined keybinds that worked before will most likely not be affected.

I used AI (mainly Claude Opus 4.8, Google Gemini, and Google AI Mode) to search the web for documentation, help with proper Swift and Zig syntax, and catch edge cases I missed. Nearly all the logic was written by myself.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1123,6 +1123,7 @@ GHOSTTY_API void ghostty_surface_set_color_scheme(ghostty_surface_t,
 GHOSTTY_API ghostty_input_mods_e ghostty_surface_key_translation_mods(ghostty_surface_t,
                                                                          ghostty_input_mods_e);
 GHOSTTY_API bool ghostty_surface_key(ghostty_surface_t, ghostty_input_key_s);
+GHOSTTY_API void ghostty_surface_record_inspector_key(ghostty_surface_t, ghostty_input_key_s);
 GHOSTTY_API bool ghostty_surface_key_is_binding(ghostty_surface_t,
                                                    ghostty_input_key_s,
                                                    ghostty_binding_flags_e*);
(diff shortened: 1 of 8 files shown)


pr-120:
I am having the same issue with this.

bo2themax:
Can you separate this into different prs, like one contains core/gtk only/macos only?

With this state I guess it would be difficult to review and merge.

MattFanto:
Any chance that this get merged please?

qappell:
Sorry, I've had a lot going on for a while. I do have a lot of fixes and updates to this PR and will hopefully push next week.

PR #13069, opened 2026-06-22
macOS: fix tabs frame on macOS 27 beta 2

bo2themax:
Also created a follow-up task to check whether this will survive the final release: #13070


mitchellh:
Looks good, should we make this a dynamic dispatch var/func and make a Golden Gate tabs subclass?

Will merge this as-is for now.

bo2themax:
> Looks good, should we make this a dynamic dispatch var/func and make a Golden Gate tabs subclass?
> 
> Will merge this as-is for now.

I think we could, but i don't know if it's worth to do it before public release. But i'll mention this in that task as well.

PR #13092, opened 2026-06-25
macos: fix SurfaceView notification retain cycle

mustafa0x:
## What

Fix a `SurfaceView` retain cycle caused by storing a Combine notification subscription on `self` while the publisher is also scoped with `object: self`.

The subscription now listens for the notification without a retained object filter and applies an equivalent weak identity filter in the Combine chain.

I also explicitly clear the related cancellables/timer in `deinit`.

## Why

Discussion: https://github.com/ghostty-org/ghostty/discussions/13090

In local A/B testing on macOS, closing a test window with 13 terminal surfaces left all 39 Ghostty-named IOSurfaces alive in the unpatched build. With this patch, the same workload dropped to zero Ghostty-named IOSurfaces after close.

```text
unpatched:
after-create    ghostty_iosurfaces=39
after-close     ghostty_iosurfaces=39

patched:
after-create    ghostty_iosurfaces=39
after-close     ghostty_iosurfaces=0
```

The suspected cycle is:

```text
SurfaceView
  -> accessibilitySelectionCancellable
  -> NotificationCenter publisher/subscription
  -> object self
  -> SurfaceView
```

The sink closure already captured `self` weakly, but the publisher's `object: self` argument appears to retain the object as part of the subscription.

## Test

```sh
mise x zig@0.15.2 -- zig build test
```

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
@@ -292,7 +292,11 @@ extension Ghostty {
             // A drag can emit multiple selection changes. Debounce so screen
             // readers hear one announcement once the selection settles.
             accessibilitySelectionCancellable = NotificationCenter.default
-                .publisher(for: .ghosttySelectionDidChange, object: self)
+                .publisher(for: .ghosttySelectionDidChange, object: nil)
+                .filter { [weak self] notification in
+                    guard let self else { return false }
+                    return (notification.object as AnyObject?) === self
+                }
                 .debounce(for: .milliseconds(100), scheduler: DispatchQueue.main)
                 .sink { [weak self] _ in
                     guard let self else { return }
@@ -386,6 +390,9 @@ extension Ghostty {
             let center = NotificationCenter.default
             center.removeObserver(self)
 
+            accessibilitySelectionCancellable = nil
+            searchNeedleCancellable = nil
+
             // Remove our event monitor
             if let eventMonitor {
                 NSEvent.removeMonitor(eventMonitor)
@@ -406,6 +413,7 @@ extension Ghostty {
 
             // Cancel progress report timer
             progressReportTimer?.invalidate()
+            progressReportTimer = nil
         }
 
         override func endSearch() {


macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
```diff
@@ -386,6 +390,9 @@ extension Ghostty {
             let center = NotificationCenter.default
             center.removeObserver(self)
 
+            accessibilitySelectionCancellable = nil
+            searchNeedleCancellable = nil
```
bo2themax, line 394: These are not necessary, but harmless

bo2themax:
!vouch

bo2themax (approved):

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
```diff
@@ -292,7 +292,11 @@ extension Ghostty {
             // A drag can emit multiple selection changes. Debounce so screen
             // readers hear one announcement once the selection settles.
             accessibilitySelectionCancellable = NotificationCenter.default
-                .publisher(for: .ghosttySelectionDidChange, object: self)
+                .publisher(for: .ghosttySelectionDidChange, object: nil)
```
mitchellh, line 295: A publisher doesn't retain `self`, as far as I'm aware. I don't have the docs in front of me or proof of this, but I always thought that NotificationCenter observers including the combine publisher do NOT retain self here.

This makes this plus the filter below unnecessary. I'd want evidence of the contrary before merging this.

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
mitchellh, line 394: This also shouldn't be necessary. Deinit will release this object which will release the combine subs.

mitchellh (changes requested):
I'm suspicious of all of this.

mitchellh:
This also looks like a fully AI written PR description @mustafa0x which is a violation of our policies, FYI, please respond with human knowledge going forward.

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
bo2themax: I did test this locally myself and checked the memory graph, there's a capture malloc. And you can add break point in deinit, before it never ran, with the change here, it ran after undo timeout(5s)

bo2themax: I wasn't sure myself before

mitchellh:
Okay, I'm wrong! `publish` **does** retain self. I couldn't find docs so I wrote this script to verify:

```swift
import Combine
import Foundation

final class Token {
    deinit { print("Token deinitialized") }
}

weak var weakToken: Token?
var publisher: NotificationCenter.Publisher?

// Create scope that will free token.
do {
    let token = Token()
    weakToken = token
    publisher = NotificationCenter.default.publisher(
        for: Notification.Name("TestNotification"),
        object: token
    )
}

print("Retained:", weakToken != nil)
publisher = nil
print("Released:", weakToken == nil)
```

mitchellh:
Replaced with #13094 with credit given to @mustafa0x Rewrote the PR and commit description manually to be a lot more reasonable lol.  Closing this.

mustafa0x:
> This also looks like a fully AI written PR description @mustafa0x which is a violation of our policies, FYI, please respond with human knowledge going forward.

Yes, in all honesty it was, and in all honesty I don't know enough Zig or Ghostty internals to have much useful human knowledge :)

However I was meticulous in guiding the agent in investigating and debugging and verifying claims. And after I had concrete results I opened a discussion that included a succinct diff, which @bo2themax kindly vouched.

But yeah, the bug is a killer. I have to restart Ghostty every ~5 days due to it. I asked you about it on twitter once, but you replied that you had never come across it.

mitchellh:
Appreciate you pointing it out. In the future, you can just be more direct with the message, hand-written works best. AI just likes to talk about a whole load of nonsense when a few sentences would've carried it across.

PR #13094, opened 2026-06-25
macos: avoid notification publisher retain cycle

mitchellh:
Turns out combine's `publisher(for:,object:)` retains the object! We verified this with a test script shown below. Fix this with a  manual filter. Found by @mustafa0x.

```
import Combine
import Foundation

final class Token {
    deinit { print("Token deinitialized") }
}

weak var weakToken: Token?
var publisher: NotificationCenter.Publisher?

// Create scope that will free token.
do {
    let token = Token()
    weakToken = token
    publisher = NotificationCenter.default.publisher(
        for: Notification.Name("TestNotification"),
        object: token
    )
}

print("Retained:", weakToken != nil)
publisher = nil
print("Released:", weakToken == nil)
```


bo2themax (approved):

mitchellh:
Note I also looked for this pattern anywhere else and this was the only culprit i could find

ghost:
hello, i didn't understand the code could sm1 pls explain?

PR #13130, opened 2026-07-01
ci: skip tip release when only non-artifact files change

bo2themax:
Ignore more files when releasing tip. Moved release-tip run to a separate script so we can test against it.

### AI Disclosure

Claude did this, I reviewed the changes and asked it run the tests locally before creating the pr.

### DIFF
--- a/.github/workflows/release-tip.yml
+++ b/.github/workflows/release-tip.yml
@@ -58,51 +58,61 @@ jobs:
           echo "build=$GHOSTTY_BUILD" >> $GITHUB_OUTPUT
           echo "commit=$GHOSTTY_COMMIT" >> $GITHUB_OUTPUT
           echo "commit_long=$GHOSTTY_COMMIT_LONG" >> $GITHUB_OUTPUT
-      - name: Check if tip already exists
+      # Classify what changed since the last tip. `build` matches any file that
+      # can reach the built artifact; `release_tip` re-adds release-tip.yml,
+      # which defines the build/tag/publish jobs. Everything else (all of
+      # .github plus docs and repo/lint/editor metadata) never changes the
+      # artifact. Skipped on manual dispatches, which always build.
+      - name: Detect changes since last tip
+        id: changes
+        if: github.event_name == 'workflow_run'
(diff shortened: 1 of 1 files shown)


.github/tip-release-ignore
```diff
@@ -0,0 +1,32 @@
+# Paths that never change the shipped app artifact. A push to main touching
+# ONLY these skips the tip release (see scripts/tip-release-skip.sh). Patterns
+# use shell case-glob semantics: * and ** both match across /.
+
+# CI, automation, repo governance.
+.github/**
```
jcollie: A change to `.github/workflows/release-tip.yml` should still trigger a build.

jcollie (changes requested):

.github/tip-release-ignore
bo2themax: thought about it before, do we need to immediately release the tip, after the yml change? I didn't think of any case when we need this.

bo2themax: Just like this pr for example, I don't see the reason why we need to release a new version 🤔

jcollie: The main one that I can think of is if you're changing the release process to pick a different version of XCode.

bo2themax: Can we manually release this in those cases?

jcollie: You could, but isn't the whole point of CI that it's all supposed to be automatic?

jcollie: Also, did you look at using `dorny/paths-filter` like we use in the `test` workflow?

bo2themax: > You could, but isn't the whole point of CI that it's all supposed to be automatic?

I get this, but i don't think changing Xcode requires an immediate release from CI. Anyway, i'll add it back for other reasons i don't know about.

bo2themax: > Also, did you look at using `dorny/paths-filter` like we use in the `test` workflow?

I can delete the scripts and reuse filter (now it seems running that test in every push doesn't worth that much)

jcollie (approved):
LGTM

PR #13153, opened 2026-07-02
macOS: only read fileURL from clipboard

### LINKED ISSUE #13152: macOS paste prefers NSURL over plain text, causing Transmit SFTP paths to paste as sftp:// URLs
Regression from: https://github.com/ghostty-org/ghostty/pull/4956

### Discussed in https://github.com/ghostty-org/ghostty/discussions/13133

<sup>Originally posted by **faridrashidi** July  1, 2026</sup>
### Issue Description

Opening this as a discussion per repository policy. I originally filed this as #13132, which was auto-closed because non-maintainers are asked to start with a discussion.

On macOS, copying a remote path from Panic Transmit with `Cmd+C` appears to put multiple representations on the pasteboard, including a plain text path and an `NSURL`/URL representation.

When pasting into Sublime Text, the plain path is pasted:

```text
/vf/users/rashidimehrabf2/nextneopi/TCGA/TCGA-2Y-A9GS
```

When pasting the same clipboard contents into Ghostty with `Cmd+V`, Ghostty pastes the URL representation instead:

```text
sftp://rashidimehrabf2@helix.nih.gov:22/vf/users/rashidimehrabf2/nextneopi/TCGA/TCGA-2Y-A9GS
```

This seems related to the macOS pasteboard representation ordering used by `getOpinionatedStringContents()` in `NSPasteboard+Extension.swift`, which currently tries `readObjects(forClasses: [NSURL.self])` first and only falls back to `.string` afterwards. For non-file URLs, it returns `absoluteString`, so an SFTP URL wins over the plain text path.

Relevant source:

```swift
if let urls = readObjects(forClasses: [NSURL.self]) as? [URL],
   urls.count > 0 {
    return urls
        .map { $0.isFileURL ? Ghostty.Shell.escape($0.path) : $0.absoluteString }
        .joined(separator: " ")
}

return self.string(forType: .string)
```

Related discussion: https://github.com/ghostty-org/ghostty/discussions/12519

### Expected Behavior

For normal terminal paste, Ghostty should prefer the plain text representation when one exists, or provide a way to paste plain text only.

In this case, `Cmd+V` should paste:

```text
/vf/users/rashidimehrabf2/nextneopi/TCGA/TCGA-2Y-A9GS
```

### Actual Behavior

Ghostty pastes:

```text
sftp://rashidimehrabf2@helix.nih.gov:22/vf/users/rashidimehrabf2/nextneopi/TCGA/TCGA-2Y-A9GS
```

### Reproduction Steps

1. On macOS, connect to an SFTP server in Panic Transmit.
2. Select a remote folder or file.
3. Press `Cmd+C` in Transmit. This is regular Copy, not `Shift+Cmd+C` / Copy URL.
4. Paste into Sublime Text and observe that the pasted value is the plain remote path.
5. Paste the same clipboard contents into Ghostty with `Cmd+V`.
6. Observe that Ghostty pastes the `sftp://user@host:port/...` URL instead of the plain path.

### Ghostty Logs

No logs observed yet. This appears to be deterministic pasteboard representation selection behavior.

### Ghostty Version

Unknown from CLI on the machine where I am filing this; `ghostty +version` was not available in PATH during discussion creation.

### OS Version Information

```text
ProductName:    macOS
ProductVersion: 26.5.1
BuildVersion:   25F80
```

Reproducible with normal paste behavior; no known clipboard-related Ghostty configuration is involved.

Source application: Panic Transmit, using regular `Cmd+C` on a remote SFTP item.

### Notes

This may be the same class of issue as #12519: Ghostty choosing a richer/non-text pasteboard representation when a plain text representation would be more appropriate for a terminal paste.</div>

bo2themax:
Fixes https://github.com/ghostty-org/ghostty/issues/13152, regression from https://github.com/ghostty-org/ghostty/pull/4956

### DIFF
--- a/macos/Sources/Helpers/Extensions/NSPasteboard+Extension.swift
+++ b/macos/Sources/Helpers/Extensions/NSPasteboard+Extension.swift
@@ -34,11 +34,11 @@ extension NSPasteboard {
 
     /// Gets the contents of the pasteboard as a string following a specific set of semantics.
     /// Does these things in order:
-    /// - Tries to get the absolute filesystem path of the file in the pasteboard if there is one and ensures the file path is properly escaped.
+    /// - Tries to get the absolute **filesystem path** of the file in the pasteboard if there is one and ensures the file path is properly escaped.
     /// - Tries to get any string from the pasteboard.
     /// If all of the above fail, returns None.
     func getOpinionatedStringContents() -> String? {
-        if let urls = readObjects(forClasses: [NSURL.self]) as? [URL],
+        if let urls = readObjects(forClasses: [NSURL.self], options: [.urlReadingFileURLsOnly: true]) as? [URL],
            urls.count > 0 {
             return urls
                 .map { $0.isFileURL ? Ghostty.Shell.escape($0.path) : $0.absoluteString }


bo2themax:
- [x] On second thought I should check whether I should delete the `isFileURL` check

PR #13151, opened 2026-07-02
fix(windows): fix MSVC shared library CRT linking in libghostty-vt

eunos-1128:
## Problem

Building `libghostty-vt` as a shared library on Windows with MSVC fails because `linkLibC()` 
only links `msvcrt.lib`, while shared DLL builds also require `vcruntime.lib` and `ucrt.lib`. 
This blocks downstream projects (e.g., `herdr`) from building against libghostty-vt on Windows.

## Solution

Explicitly link the full MSVC CRT chain for shared library builds:
- Link `libvcruntime`
- Add Windows SDK UCRT library path to linker search directories  
- Link `libucrt`

## Changes

Modified `src/build/GhosttyLibVt.zig` to:
1. Detect shared library builds with MSVC ABI
2. Dynamically resolve Windows SDK path and architecture-specific UCRT directory
3. Link missing CRT libraries

This change aligns with an existing fix in `GhosttyLib.zig`.

## Verification

✅ Build succeeds on Windows with MSVC

## Related

Addresses the Windows/MSVC build issue encountered when packaging `herdr` for conda-forge:
https://github.com/conda-forge/staged-recipes/pull/34029

PR #13164, opened 2026-07-03
Fix MSVC shared library CRT linking in libghostty-vt

eunos-1128:
Reprise https://github.com/ghostty-org/ghostty/pull/13151#event-27484705627 following I'm been vouched.


00-kat:
Superseded by #13151 (which was reopened).

PR #13171, opened 2026-07-03
fix: set max window clamp to current monitor size

yak3d:
This PR fixes #7984. The issue was that GTK would clamp the window itself based on the display it was opened on. We fix this by computing the size based on the current display and then implicitly setting the window size instead of relying on GTK to do it.

Claude Code w/ Opus 4.7 was used to investigate, fix and explain some of the Ghostty architecture to me.

### DIFF
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -1156,6 +1156,38 @@ pub const Window = extern struct {
         };
     }
 
+    fn toplevelComputeSize(
+        _: *gdk.Toplevel,
+        size: *gdk.ToplevelSize,
+        self: *Self,
+    ) callconv(.c) void {
+        // The compositor/quick terminal own the size in these states.
+        if (self.isMaximized() or
+            self.isFullscreen() or
+            self.isQuickTerminal()) return;
+
+        // If there's no GdkSurface yet these dimensions will be zero size which
+        // will make the window start out as small as possible. These checks ensure
+        // we don't start with a 0, 0 window size.
+        const gdk_surface = self.as(gtk.Native).getSurface() orelse return;
+        const w = gdk_surface.getWidth();
+        const h = gdk_surface.getHeight();
+        if (w <= 0 or h <= 0) return;
+
+        // GTK clamps the requested size to the compositor-reported bounds, which
+        // go stale when the window moves to a larger monitor. Only re-assert the
+        // current size when it exceeds those bounds; otherwise let GTK's default
+        // sizing win so a freshly-mapped window isn't forced to a tiny size.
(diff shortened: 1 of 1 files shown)


jcollie (changes requested):
This change makes it so that new windows (on Gnome 50) starts out as a minimally sized window. It ignores the `window-width` and `window-height` configs completely. Not sure what DE you're testing with but this change won't be accepted until that is remedied.

yak3d:
> This change makes it so that new windows (on Gnome 50) starts out as a minimally sized window. It ignores the `window-width` and `window-height` configs completely. Not sure what DE you're testing with but this change won't be accepted until that is remedied.

I noticed that, and thought it was just a bad config on my end. Sorry about that. I resolved it by only overriding gtk's clamp when the window exceeds the bounds so it should open at the correct size now.

src/apprt/gtk/class/window.zig
```diff
@@ -1156,6 +1156,35 @@ pub const Window = extern struct {
...
+        _: *gdk.Toplevel,
+        size: *gdk.ToplevelSize,
+        self: *Self,
+    ) callconv(.c) void {
+        // The compositor/quick terminal own the size in these states.
+        if (self.isMaximized() or
+            self.isFullscreen() or
+            self.isQuickTerminal()) return;
+
+        const gdk_surface = self.as(gtk.Native).getSurface() orelse return;
```
jcollie, line 1172: Would be good to have a comment explaining why this check is here.

src/apprt/gtk/class/window.zig
```diff
@@ -1156,6 +1156,35 @@ pub const Window = extern struct {
...
+        const w = gdk_surface.getWidth();
+        const h = gdk_surface.getHeight();
+        if (w <= 0 or h <= 0) return;
+
+        // GTK clamps the requested size to the compositor-reported bounds, which
+        // go stale when the window moves to a larger monitor. Only re-assert the
+        // current size when it exceeds those bounds; otherwise let GTK's default
+        // sizing win so a freshly-mapped window isn't forced to a tiny size.
+        var bounds_w: c_int = 0;
+        var bounds_h: c_int = 0;
```
jcollie: Small optimization, these should be initialized to `undefined`.

jcollie (changes requested):
Couple of small requests.

jcollie (approved):

UnsaltedScholar:
This also addresses: https://github.com/ghostty-org/ghostty/discussions/9421

Many thanks for this fix, this was a very annoying issue for me. (the upstream gtk [issue](https://gitlab.gnome.org/GNOME/gtk/-/work_items/7869) I created still hasn't been looked at after 8 months...)

PR #13173, opened 2026-07-03
kitty/gfx: add generation stamps, delete transmit time

mitchellh:
Add a generation counter to the kitty graphics image storage. Every content mutation (image transmit/replace, placement add, delete) assigns the storage a fresh stamp, and every image is stamped when it is added or replaced. 

This solves two problems: 

First, a retransmission of the same image ID with identical dimensions  was previously undetectable by anything comparing width, height, format,  and data length; the per-image stamp changes on every add/replace, so caches  keyed on it always see the change. Second, the dirty flag was the only  storage-wide signal, and it is also set by scrolling and resizing, which move  placements without changing contents. The generation is only bumped by content mutations, so an unchanged value means the placement set and all image data are identical and consumers can skip re-reading them, recomputing only placement geometry.

The generation replaces Image.transmit_time entirely: newest-image lookup by number, eviction ordering, and the renderer's texture staleness checks all key on it now. A monotonic counter strictly orders transmissions where Instant-based times could collide within clock resolution (the renderer previously assumed equal timestamps meant identical images), and this removes a syscall and an error path per transmission.

Delete commands now only mark a mutation (dirty flag and generation) when they actually remove something. A delete-all runs on every screen clear, so previously every ESC [ 2 J dirtied the image state even with no images stored. Eviction via setLimit also now marks the state dirty, which it previously did not.

### DIFF
--- a/example/c-vt-kitty-graphics/src/main.c
+++ b/example/c-vt-kitty-graphics/src/main.c
@@ -125,6 +125,17 @@ int main() {
   }
   printf("\nKitty graphics storage is available.\n");
 
+  /*
+   * The storage-wide generation changes on every image/placement
+   * mutation. Renderers can compare it against the value from the
+   * previous frame: if unchanged, placement iteration and image
+   * staleness checks can be skipped entirely.
+   */
+  uint64_t generation = 0;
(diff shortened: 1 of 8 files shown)


mitchellh:
I discovered this while working on some new downstream libghostty-vt renderers (multiple!). This makes it much more robust and performant to implement kitty graphics in those renderrs, and I adapted those to our own too here.

PR #13180, opened 2026-07-04
macos/applescript: derive key event text and codepoint

paaloeye:
Executing the  `send` key  **AppleScript** command is  _broken_ for printable characters. The generated `KeyEvent` lacks both _text representation_ and `unshiftedCodepoint`. Consequently, **Ghostty** ignores or failed to process letters and spaces sent via **AppleScript**.

Minimal repro (**doesn't** work as expected with `stable` and `tip`):

> [!WARNING]
> This sends `date` to all opened terminals/surfaces

```applescript
tell application "Ghostty"
    set allTerms to terminals

    repeat with t in allTerms
        send key "d" to t -- doesn't work 
        send key "a" to t -- ditto
        send key "t" to t -- ditto
        send key "e" to t -- ditto
        send key "enter" to t
    end repeat

   display dialog ("Broadcasted to " & (count of allTerms) & " terminal(s).")
end tell
```

This, however, **does** work (from [documentation](https://ghostty.org/docs/features/applescript#broadcast-a-command-to-every-terminal)):

```applescript
set cmd to "date"

tell application "Ghostty"
    set allTerms to terminals

    repeat with t in allTerms
        input text cmd to t
        send key "enter" to t
    end repeat

    display dialog ("Broadcasted to " & (count of allTerms) & " terminal(s).")
end tell
```

It has a different effect when one wants to broadcast on _character per character_ (e.g. synchronise 
two or more **Kakoune** or **NeoVim**, etc)

This PR fixes that by using  `UCKeyTranslate`  to retrieve the current keyboard layout, translate the simulated keycodes, and correctly populate these missing fields before sending the key event.

I have tested this locally using the **US International** keyboard layout as well as **US** proper, and it successfully translates the keycodes. I'd appreciate if someone with German or French layout test it.

I have also added some smoke tests to verify the AppleScript key event mapping. These are isolated in a separate commit; if you find them too noisy, I am happy to get rid of them.

An **AppleScript** `send` key command that cannot type standard letters is not particularly useful. Ensuring key events carry their respective text and codepoints is necessary for basic automation to function.

> [!NOTE]
> I am also happy to create a tracking issue if Ghosttty's workflow requires it.

#### Changes

  •  **ScriptKeyEventCommand.swift** : Translated key codes to character text and unshifted codepoints using  UCKeyTranslate .
  •  **KeyEventCommandTests.swift**: Added basic unit tests
  
### AI Disclosure
  
> [!IMPORTANT]
> This PR was prepared using **Antigravity CLI**. The AI helped to discover `UCKeyTranslate` Carbon API.

### DIFF
--- a/macos/Sources/Features/AppleScript/ScriptKeyEventCommand.swift
+++ b/macos/Sources/Features/AppleScript/ScriptKeyEventCommand.swift
@@ -1,4 +1,5 @@
 import AppKit
+import Carbon
 
 /// Handler for the `send key` AppleScript command defined in `Ghostty.sdef`.
 ///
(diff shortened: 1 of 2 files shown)


bo2themax:
I don't quite get what your use case is. Can you make an example of where current implementation doesn't work?

And be careful, this pr description already violates our ai policy.

paaloeye:
> Can you make an example of where current implementation doesn't work?

@bo2themax Of course, I'll trim the description and add a detailed explanation where current implementation breaks.
Cheers for the reply and my bad about violating AI policy!

paaloeye:
@bo2themax I've updated the description. Let me know if you have any questions. Cheers!

macos/Sources/Features/AppleScript/ScriptKeyEventCommand.swift
```diff
@@ -64,13 +65,67 @@ final class ScriptKeyEventCommand: NSScriptCommand {
...
+
         surface.sendKeyEvent(keyEvent)
 
         return nil
     }
 }
+
+/// extracted translation logic to map a Ghostty key to its generated text and unshifted codepoint,
+/// primarily so it can be unit-tested without instantiating an `NSScriptCommand`.
+struct ScriptKeyEventTranslator {
```
bo2themax, line 87: I don't think we need UCKeyTranslate here, ppl who use it should know which key they are trying to type.

bo2themax (changes requested):

macos/Sources/Features/AppleScript/ScriptKeyEventCommand.swift
bo2themax: KeymapDarwin.zig already using it

paaloeye: Let me clarify one thing.

Are you saying that the below should be able to _derive_ `.text` in KeymapDarwin.zig (via `UCKeyTranslate`)? 

```swift
let keyEvent = Ghostty.Input.KeyEvent(
    key: key,
    action: action,
    mods: mods
)
```

And my use case/repro _should_ work without the fix?

bo2themax:
Please see #13205, I don't like using `UCKeyTranslate` directly in swift. Your findings and ideas are legit tho, thanks.

paaloeye:
> Please see #13205, I don't like using `UCKeyTranslate` directly in swift. Your findings and ideas are legit tho, thanks.

Awesome! I'll test it in my setup.

PR #13182, opened 2026-07-04
terminal: add clipboard_set effect for OSC 52 clipboard writes

pearkes:
libghostty-vt was already parsing OSC 52 into the clipboard_contents action but the stream handler dropped, so there was no way to observe clipboard writes (in this case, a program using go-libghostty). This adds a clipboard_set effect following the existing bell/title_changed pattern and expose it through the C API as `GHOSTTY_TERMINAL_OPT_CLIPBOARD_SET`.

The callback receives the OSC 52 kind byte and the base64 payload exactly as received; decoding and kind interpretation are left to the embedder, matching how decoding is typically deferred. 

Note this intentionally does not deal with clipboard read requests given the security implications.

AI disclosure: Fable (via Claude Code) did the majority of the work here, I validated it on the client side and fully understand the pattern we're fitting into.

### DIFF
--- a/include/ghostty/vt/terminal.h
+++ b/include/ghostty/vt/terminal.h
@@ -82,7 +82,7 @@ extern "C" {
  * | `GHOSTTY_TERMINAL_OPT_SIZE`             | `GhosttyTerminalSizeFn`           | XTWINOPS size query (CSI 14/16/18 t)      |
  * | `GHOSTTY_TERMINAL_OPT_COLOR_SCHEME`     | `GhosttyTerminalColorSchemeFn`    | Color scheme query (CSI ? 996 n)          |
  * | `GHOSTTY_TERMINAL_OPT_DEVICE_ATTRIBUTES`| `GhosttyTerminalDeviceAttributesFn`| Device attributes query (CSI c / > c / = c)|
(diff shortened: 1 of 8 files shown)


trag1c:
!vouch

include/ghostty/vt/terminal.h
```diff
@@ -302,6 +285,34 @@ typedef struct {
...
+ * @param userdata The userdata pointer set via GHOSTTY_TERMINAL_OPT_USERDATA
+ * @param kind The OSC 52 selection kind byte
+ * @param data Pointer to the base64-encoded payload bytes
+ * @param len Length of the payload in bytes
+ *
+ * @ingroup terminal
+ */
+typedef void (*GhosttyTerminalClipboardSetFn)(GhosttyTerminal terminal,
+                                              void* userdata,
+                                              uint8_t kind,
```
mitchellh: The thing I don't like in this PR is this. This is some abstraction leakage here. Ideally we can reuse this callback for ANY type of clipboard set protocol, e.g. also supporting Kitty clipboard protocol and so on.

I'm not sure what the proper shape is but I'd block this for now on figuring that out and trying to come up with something more generic here...

mitchellh (changes requested):
One note which is really a public API shape. The implementation itself is fine.

mitchellh:
Okay made the change! Waiting on CI then will merge.

PR #13187, opened 2026-07-04
VisionOS Port Spike

avb:
Initial spike for porting Ghostty to visionOS


mitchellh:
It'd be more useful to get libghostty-vt compiling for visionOS rather than the whole app.

PR #13191, opened 2026-07-05
config: parse hex escapes as raw bytes, not codepoints

tasselx:
#13061

The string literal parser treated the value of every escape sequence as a Unicode codepoint and re-encoded it as UTF-8. For \x escapes of bytes >= 0x80 this double-encodes the byte: \xe6 (one byte of a UTF-8 multibyte sequence) became the two bytes C3 A6, corrupting the string into mojibake.

This broke the macOS "Open With > Ghostty" flow for files with non-ASCII names: the AppKit runtime passes the initial input through std.zig.stringEscape, which escapes every byte >= 0x80 as \xNN, and the parse on the other side then mangled the path so the shell failed with "no such file or directory".

Match the semantics of std.zig.string_literal.parseWrite: only \u{...} escapes produce a codepoint that is UTF-8 encoded; all other escapes produce a single raw byte that is copied through unchanged.

### DIFF
--- a/src/config/string.zig
+++ b/src/config/string.zig
@@ -20,16 +20,28 @@ pub fn parse(out: []u8, bytes: []const u8) ![]u8 {
             continue;
         }
 
+        // Only `\u{...}` escapes produce a codepoint that must be UTF-8
+        // encoded. All other escapes (`\x`, `\n`, ...) produce a single
+        // raw byte that must be copied as-is, otherwise `\x` escapes of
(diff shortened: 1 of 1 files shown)


mitchellh:
!vouch

tasselx:
Could a maintainer please label this with `os/macos` and `gui`?

This fixes the macOS "Open With > Ghostty" path mojibake issue reported in #13061. The fix is small and includes a regression test for `\xNN` byte escapes, so a review/merge when someone has time would be much appreciated. Thanks!

PR #13205, opened 2026-07-05
macOS: fix non control keys are not working for AppleScript

bo2themax:
`send key` only works for control keys like `enter` currently; this adds (fixes) the support for other keys listed as available. Found by @paaloeye in #13180

The core of this fix is relying on `UCKeyTranslate` to get the corresponding character and code point from a key code using `KeyboardLayout.character(for:modifiers:)`. 

ScriptKeyEventCommand now respects `macos-option-as-alt`, and attach `text`, `unshifted_codepoint` and `consumed_mods` under the same condition as a manual input events like in `performKeyEquivalent` and `localEventKeyDown`.

## AI Disclosure

Claude did the heavy lifting, I reviewed and rephrased some of the comments it generated. And ofc reviewed and tested myself.

### DIFF
--- a/macos/Ghostty.sdef
+++ b/macos/Ghostty.sdef
@@ -226,7 +226,7 @@
       </parameter>
     </command>
 
-    <command name="send key" code="GhstSKey" description="Send a keyboard event to a terminal.">
+    <command name="send key" code="GhstSKey" description="Send a keyboard event to a terminal. Keys are translated once through the current keyboard layout. Dead keys do not compose with subsequent keys.">
       <cocoa class="GhosttyScriptKeyEventCommand"/>
       <direct-parameter type="text" description="The key name (e.g. &quot;enter&quot;, &quot;a&quot;, &quot;space&quot;)."/>
       <parameter name="action" code="GKeA" type="input action" optional="yes" description="Press or release (default: press).">
--- a/macos/Sources/Features/AppleScript/ScriptKeyEventCommand.swift
+++ b/macos/Sources/Features/AppleScript/ScriptKeyEventCommand.swift
@@ -33,7 +33,9 @@ final class ScriptKeyEventCommand: NSScriptCommand {
         do {
             keyEvent = try Self.parse(
                 directParameter: directParameter,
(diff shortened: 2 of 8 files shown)


paaloeye (approved):
Well done! 

Tested it in my [harness](https://github.com/paaloeye/libgtty.nu) and it works as expected.

Test command: `gtty surface broadcast --engine swift --offset -3..-1`. 
One would need four surfaces and Nushell installed.

paaloeye:
I've been daily driving Ghostty with that — couldn't find any problems with it.
@bo2themax Is anything missing here? It wasn't trivial to discover using `send key` with anything other than `enter`, IMO it's worth updating the docs. I'm happy to give it go.

bo2themax:
> IMO it's worth updating the docs. I'm happy to give it go.

You're welcome to open a pr to the website once this is merged.

jparise (commented):
We learned a few things from #13888 that could be helpful here:

- The flaky tests probably needed to be `@MainActor`. The AppKit key translation routines want to run on the main thread.
- We could reuse/refactor the new `NSEvent` creation code in `KeyboardLayout` instead of duplicating it.
- Set `consumedMods` for Shift/Option translations to match real key events.
- Avoid synthetic text on releases/control keys: real `keyAction` filters C0 text and sends key-up without text. This PR currently bypasses that, potentially changing Enter/Tab and Kitty protocol behavior.

bo2themax:
> We learned a few things from #13888 that could be helpful here:
> 
> * The flaky tests probably needed to be `@MainActor`. The AppKit key translation routines want to run on the main thread.
> * We could reuse/refactor the new `NSEvent` creation code in `KeyboardLayout` instead of duplicating it.
> * Set `consumedMods` for Shift/Option translations to match real key events.
> * Avoid synthetic text on releases/control keys: real `keyAction` filters C0 text and sends key-up without text. This PR currently bypasses that, potentially changing Enter/Tab and Kitty protocol behavior.

Thanks, will take a look at these points next time I rebase this.

jparise (approved):

PR #13209, opened 2026-07-06
termio: pipeline pty reads to overlap parsing with draining (25 - 55% IO throughput increase)

mitchellh:
This replaces the single-threaded pty read loop on posix systems with a two-stage pipeline: a new `io-gather` thread drains the pty into a small ring of large buffers while the existing `io-reader` thread parses the previous batch concurrently.

After this, our IO throughput is now within noise (1 to 3%) of our VT parsing and processing throughput. This means that any VT performance improvements should raise IO throughput. We're completely bound by it now.

> [!IMPORTANT]
> 
> **Against all odds, we've found a massive (25% to 55%) performance improvement in Ghostty IO.** This is a subsystem we were happy to get 2% improvement gains in recent years. And the change is relatively simple and understandable, too.

The motivating discovery was actually found by Fable, but the resulting code was hand-written and hand-verified (in addition to model-verified as an extra check): on macOS the kernel tty output queue caps every read on the pty master at about 1 KiB regardless of the read buffer size. Instrumenting a pty with a 64 KB buffer while streaming a 6.49 MB file produced 6,337 reads where every read was exactly  1024 bytes. 

This immediately made me realize two things about the old loop that we've  had since like 2023 which called processOutput once per read: all per-call  overhead (terminal lock, render wakeup, cursor timestamp) was paid per kilobyte of bulk output, and the child could never run more than 1 KiB ahead of us, so while we parsed the child (e.g. `cat`) sat blocked on a full kernel queue instead of producing. 

In 2023, I justified this architecture by saying "reads are generally small" but I didn't understand then that reads are generally small because the kernel makes them small even if there is a lot of data.

To preserve latency for the more typical
small-reads-that-are-actually-small, sub-1 KB payloads deliver on the first EAGAIN with no added latency.

## Benchmarks

End-to-end throughput was measured by timing `cat file > /dev/ttysN` against a fresh app instance (M4 Max, macOS 26, ReleaseFast, medians of repeated interleaved A/B runs):

| stream                  | before       | after         | change  |
|-------------------------|--------------|---------------|---------|
| ascii.txt (6.5 MB)      | 91-92 MB/s   | 114-123 MB/s  | +25-30% |
| unicode.txt (8 MB)      | 116-117 MB/s | 180-183 MB/s  | +55%    |
| DOOM-Fire-Zig           | 530 fps      | 770 fps       | +45%    |

The pipeline now delivers the parse stage's full measured capacity (the parse thread is pegged while gather spends ~33% of a core, so any IO throughput improvements are now fully parser-bound).

**Linux note:** This needs to be verified on Linux. I think broadly architecture is better and should never be worse. But its possible some of the magic constants need to be tuned differently. Would love more testing there.

## Some Background on LLM Usage

As noted above, the original promising path was found by Fable 5. I decided to throw some money at analyzing our IO throughput, and went into it expecting minor improvements at the cost of unmaintainable special case optimizations (typical LLM results when given such tasks on an already-optimized path). 

I spun off a Fable 5 session (API pricing) prior to some holiday weekend (American, July 4th) festivities. When I came back late that night, it pointed this path out with some pretty questionable code/results.

The frustrating thing is that _I swear I tried this years ago_ and it didn't deliver results. Unfortunately, it was long enough ago (probably 2023) that I can't remember nor do I have any evidence. But, my brain had already clocked this possibility and blocked it out as "already tried and failed." The code in question that this PR ultimately touched has been basically unchanged for 2+ years.

As a second point, this is a **great example of what I love about LLMs**. I was not in a performance-hunting mood and I have other tasks I need to get to while at the computer, so improving Ghostty performance wasn't my current mode and I would not have worked on it anytime soon while at the computer. But, I try to keep an agent running all the time, and before my family came over for holiday afternoon, I decided today I would try budgeting $100 to Fable to focus on Ghostty's IO performance. Why not?

I came back, saw some questionable but interesting results, and decided it was worth some human hours to validate, understand, and work on this myself. It was worth it. 😄 

Anyways, the total API cost of this if you're curious: **$88.28**.

Remember, I **hand-wrote the code** so thats basically what it cost me to discover this path. Fable did produce its own solution (in about 3x the LoC change of this diff with non-idiomatic Zig styles, overly defensive guards, and simultaneously poor error handling). So, I guess I did pay for a solution. A bad one. Haha. But the problem it found was real and good.

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -1240,23 +1240,132 @@ const Subprocess = struct {
     }
 };
 
-/// The read thread sits in a loop doing the following pseudo code:
+/// The read thread works with a companion gather thread to form a two-stage
+/// pipeline that moves pty output into the terminal:
+///
+///   io-gather:  read()/poll() the pty into one of a few rotating
+///               buffers, batching bulk output.
+///   io-reader:  hand each filled buffer to processOutput (terminal
+///               lock, VT parse, state update, render scheduling).
+///
+/// This used to be a single serial loop (and still is on Windows):
 ///
 ///   while (true) { blocking_read(); exit_if_eof(); process(); }
 ///
-/// Almost all terminal-modifying activity is from the pty read, so
-/// putting this on a dedicated thread keeps performance very predictable
-/// while also almost optimal. "Locking is fast, lock contention is slow."
-/// and since we rarely have contention, this is fast.
+/// I found on macOS that the kernel tty output queue caps every read
+/// on the master at 1 KB no matter how large the read buffer is. This means
+/// that producers (e.g. `cat`) stall with the above architecture because
+/// there are windows in the `process()` part where we aren't draining
+/// the kernel pty fd.
 ///
-/// This is also empirically fast compared to putting the read into
-/// an async mechanism like io_uring/epoll because the reads are generally
-/// small.
+/// Instead, having a separate thread gather and drain the kernel pty
+/// into a rotating set of preallocated buffers minimizes this stall
+/// period to effectively zero: while the io-reader thread parses
+/// one batch, the gather thread is draining the kernel queue. There is
+/// still stalling (our VT parse is a bottleneck now), but we don't stall
+/// between them.
 ///
-/// We use a basic poll syscall here because we are only monitoring two
-/// fds and this is still much faster and lower overhead than any async
-/// mechanism.
+/// Interactive latency is preserved: a batch is delivered on the
+/// first EAGAIN unless the stream is saturated (>= 1 KiB gathered
+/// means the writer filled the kernel queue), in which case we bridge
+/// the producer's microsecond refill gaps with a short poll, bounded
+/// by a small total budget per batch that is well under a display
+/// frame. This means that small outputs that are more typical continue
+/// to be interactive.
+///
+/// We use basic poll/read syscalls here because we are only
+/// monitoring two fds and this is still much faster and lower
+/// overhead than any async mechanism.
 pub const ReadThread = struct {
+    /// The number of buffers rotated between the gather and parse
+    /// stages. The gather stage can run at most this many batches
+    /// ahead of the parse stage before it blocks, which (via the
+    /// kernel pty queue) is also what preserves flow control to the
+    /// child. Empirically chosen through measurements on an M4 Max.
+    /// Less than 4 there are minor slowdowns, above 4 there are no
+    /// improvements.
+    const buffer_count = 4;
+
+    /// The capacity of each gather buffer. One batch is also the unit
+    /// of work the parse stage does per terminal lock acquisition, so
+    /// this bounds both gather latency and lock hold time.
+    const buffer_capacity = 64 * 1024;
+
+    /// How many gathered bytes mark a stream as saturated. The macOS
+    /// kernel tty output queue hands the master at most about 1 KiB
+    /// per read, so gathering a full 1 KiB means the writer filled
+    /// the queue (a bulk stream worth briefly waiting on), while
+    /// anything smaller is an interactive trickle that must be
+    /// delivered with no added latency.
+    const bridge_threshold = 1024;
+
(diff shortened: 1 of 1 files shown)


mitchellh:
# vtebench - Ghostty Main vs this PR

CleanShot 2026-07-05 at 22 04 44@2x

# vtebench - Ghostty Main vs. this PR vs. Alacritty vs. Kitty (typical fast)

CleanShot 2026-07-05 at 22 05 55@2x

qappell:
@mitchellh Equally impressive gains on Linux for me (NixOS, Hyprland, Ryzen 5 9600X with iGPU) with just some quick, dirty tests:

### ASCII throughput w/ hyperfine
Before:
| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `clear; cat /tmp/ghostty-bench/ascii.txt > /dev/tty` | 120.0 ± 9.4 | 100.0 | 136.3 | 1.00 |

After:
| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `clear; cat /tmp/ghostty-bench/ascii.txt > /dev/tty` | 64.2 ± 4.6 | 56.6 | 73.4 | 1.00 |

### vtebench

Before:
```
  cursor_motion (1195 samples @ 1.22 MiB):
    8.09ms avg (90% < 8ms) +-0.65ms

  dense_cells (836 samples @ 1.29 MiB):
    11.17ms avg (90% < 11ms) +-0.87ms

  light_cells (1405 samples @ 1 MiB):
    6.72ms avg (90% < 7ms) +-0.83ms

  scrolling (26 samples @ 1 MiB):
    387.5ms avg (90% < 412ms) +-17.65ms

  scrolling_bottom_region (57 samples @ 1 MiB):
    175.54ms avg (90% < 184ms) +-6.25ms

  scrolling_bottom_small_region (58 samples @ 1 MiB):
    174.47ms avg (90% < 185ms) +-6.55ms

  scrolling_fullscreen (394 samples @ 1 MiB):
    24.88ms avg (90% < 31ms) +-5.27ms

  scrolling_top_region (60 samples @ 1 MiB):
    167.97ms avg (90% < 177ms) +-6.49ms

  scrolling_top_small_region (60 samples @ 1 MiB):
    168.87ms avg (90% < 178ms) +-7.48ms

  unicode (787 samples @ 1.06 MiB):
    12.25ms avg (90% < 18ms) +-3.61ms
```

After:
```
  cursor_motion (1340 samples @ 1.22 MiB):
    6.82ms avg (90% < 10ms) +-1.67ms

  dense_cells (967 samples @ 1.29 MiB):
    9.9ms avg (90% < 11ms) +-1.12ms

  light_cells (1649 samples @ 1 MiB):
    5.59ms avg (90% < 8ms) +-1.42ms

  scrolling (31 samples @ 1 MiB):
    322.97ms avg (90% < 327ms) +-6.45ms

  scrolling_bottom_region (78 samples @ 1 MiB):
    128.21ms avg (90% < 136ms) +-6.48ms

  scrolling_bottom_small_region (78 samples @ 1 MiB):
    127.79ms avg (90% < 137ms) +-6.91ms

  scrolling_fullscreen (654 samples @ 1 MiB):
    14.81ms avg (90% < 20ms) +-3.13ms

  scrolling_top_region (78 samples @ 1 MiB):
    128.15ms avg (90% < 137ms) +-6.85ms

  scrolling_top_small_region (78 samples @ 1 MiB):
    128.95ms avg (90% < 138ms) +-7.16ms

  unicode (1311 samples @ 1.06 MiB):
    7.07ms avg (90% < 11ms) +-2.49ms
```

### DOOM-fire-zig
Before: 
`mem: 51179B min / 209808B avg / 222138B max [ 352.45 fps ]`

After:
`mem: 50974B min / 203228B avg / 222937B max [ 532.48 fps ]`

mitchellh:
Awesome, thank you for confirming!

PR #13220, opened 2026-07-06
terminal: various VT processing optimizations (~1.5x to ~6x throughput increase)

mitchellh:
This is a series of five commits that optimizes VT processing throughput in various ways. Each commit is isolated, individually benchmarked, and carries a detailed commit message so please read each for details about each change.

After #13209 made IO fully parser-bound, these gains should translate directly into end-to-end IO throughput (until some other stage becomes the new bottleneck). Plain ASCII processing went from ~128 MB/s to ~725 MB/s. `time cat ascii_150MB.txt` went from 1.5s before 13209 to 1.2s on main to 566ms on this branch.

## The changes

1. **batch printed codepoint runs into direct row fills**. Profiling showed ~85% of plain-text time inside `Terminal.print`, re-answering the same questions (margins, modes, width, charset, style) for every single character. A new `print_slice` stream action delivers runs of decoded codepoints to `Terminal.printSlice`, which hoists the invariants and fills rows with a masked-compare + branch-free store loop, falling back to `print()` for anything complex. **Result: 2.2x–5.7x on ascii plus unicode text.**
2. **dispatch CSI finals directly from stream fast paths**. Every byte through `Parser.next` copies a ~240 byte `[3]?Action` and a typical CSI copied it twice. New `csi_entry/final` fast paths dispatch directly without the action array. **Result: +17-18% on CSI streams.**
3. **bulk-parse CSI parameter bytes at the slice level**. Parameter digits/separators are consumed in a tight slice loop with parser state in locals instead of re-entering the per-byte path. **Result: +29-41% on escape-heavy streams.**
4. **skip style map update when SGR leaves style unchanged**. Skip the release/hash/probe/use churn when an SGR attribute is a no-op. **Result: +4-7% on TUI-refresh patterns, -2-3% on adversarial random-color streams** (tradeoff detailed in the commit message). This one is more questionable, but willing to measure on real workloads.

## Benchmarks

Measured with `ghostty-bench +terminal-stream` (full terminal handler, 100 MB deterministic synthetic corpora, 120x80 terminal, M4 Max, macOS 26, ReleaseFast, hyperfine means of 10 runs, ~15 ms process startup included in all numbers). These are parser-stage numbers, not end-to-end app numbers.

| stream                     | before | after  | throughput       | change |
|----------------------------|--------|--------|------------------|--------|
| ascii (no newlines)        | 784 ms | 138 ms | 128 → 725 MB/s   | 5.7x   |
| ascii lines                | 833 ms | 198 ms | 120 → 505 MB/s   | 4.2x   |
| unicode mixed-script       | 779 ms | 320 ms | 128 → 313 MB/s   | 2.4x   |
| CJK (all wide)             | 424 ms | 126 ms | 236 → 794 MB/s   | 3.4x   |
| unicode, mode 2027 on      | 807 ms | 367 ms | 124 → 273 MB/s   | 2.2x   |
| CJK, mode 2027 on          | 495 ms | 198 ms | 202 → 505 MB/s   | 2.5x   |
| csi mix (SGR/CUP/EL/modes) | 648 ms | 414 ms | 154 → 242 MB/s   | 1.6x   |
| sgr fire (doom-fire-like)  | 495 ms | 303 ms | 202 → 330 MB/s   | 1.6x   |
| TUI redraw (repeat styles) | 642 ms | 291 ms | 156 → 344 MB/s   | 2.2x   |
| osc                        | 8.26 s | 8.20 s | (untouched path) | ~1.0x  |

**End-to-end note:** #13209 measured the parse thread pegged while the gather thread used ~33% of a core, so parser gains of this size may make gather (or the renderer lock) the new bottleneck for plain text before the full 5.7x shows up end to end. I'll take a look at that soon...

## LLM Notes

These findings were almost all found by Fable 5. I went through each change and simplified quite a lot, read every single line, re-ran verifications by hand. Fable in particular isn't good at writing elegant Zig code, so there's a lot of style stuff. Ultimately though, I understand all of this and feel comfortable with the changes.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -321,6 +321,422 @@ pub fn printRepeat(self: *Terminal, count_req: usize) !void {
     }
 }
 
+/// Print multiple codepoints to the terminal at once. This is
+/// semantically identical to calling `print` for each codepoint in
+/// order, but is much faster because it can batch cell writes and
+/// hoist per-codepoint checks out of the hot loop.
+///
+/// The codepoints must all be printable: it is illegal for any
+/// codepoint in this slice to be a C0 control character. Therefore,
+/// this should only be called as a result of a proper VT parser
+/// (like our own).
+///
+/// This is optimized for the common case: ASCII, soft-wrap, etc.
+/// Sequences of codepoints that require special handling (e.g. wide characters,
+/// grapheme clustering) are handled correctly but fall back to the
+/// slower per-codepoint path. They're less common and this is optimized
+/// for the aforementioned cases.
+pub fn printSlice(self: *Terminal, cps: []const u32) !void {
+    var i: usize = 0;
+    while (i < cps.len) {
+        // Try the fast-path print first. This will return the number of
+        // codepoints it consumed.
+        const consumed = try self.printSliceFast(cps[i..]);
+        if (consumed > 0) {
+            i += consumed;
+            continue;
+        }
+
+        // Consuming zero bytes means that the fast path can't handle
+        // the next codepoint or the terminal is in a state we can't
+        // fast-path. Fall back to the slow cp-by-cp print then try
+        // fast paths again.
+        try self.print(@intCast(cps[i]));
+        i += 1;
+    }
+}
+
+/// Attempt to print a prefix of `cps` using a batched fast path that
+/// writes cells directly. Returns the number of codepoints consumed.
+/// A return value of zero means the caller must print the first
+/// codepoint via the normal `print` path.
+///
+/// The fast path handles runs of narrow (width 1) and wide (width 2)
+/// codepoints being written to simple cells. Everything else (zero
+/// width codepoints, grapheme cluster continuations, insert mode,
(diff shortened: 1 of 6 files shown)


src/terminal/Terminal.zig
```diff
@@ -321,6 +321,422 @@ pub fn printRepeat(self: *Terminal, count_req: usize) !void {
...
+inline fn printSliceEligible(cp: u32, comptime width: PrintSliceWidth) bool {
+    assert(cp > 0xFF);
+    if (comptime build_options.kitty_graphics) {
+        if (cp == kitty.graphics.unicode.placeholder) return false;
+    }
+
+    return unicode.table.get(@intCast(cp)).width == comptime @as(u2, switch (width) {
+        .narrow => 1,
+        .wide => 2,
+    });
```
jparise, line 496: Maybe this here?

```suggestion
    return unicode.table.get(@intCast(cp)).width == comptime @as(u2, width.cellsPerCp());
```

PR #7732, opened 2025-06-29
Sprite Face Rework

qwerasd205:
- Large rework of how we draw sprite font glyphs, explained in the commit message and comments.
- Adds separated block sextants from symbols for legacy computing supplement
- Adds explicit underline cursor instead of using underline glyph for it, resolves #7651, supersedes #7685

Currently we support these glyphs with the sprite font:
```
[...]

PR #13225, opened 2026-07-06
Fix `adjust-cursor-height` regression

qwerasd205:
The `cursor-height` metric (and corresponding `adjust-`) was introduced by #3062 but the sprite face rework in #7732 accidentally removed the logic that it relied on. I've moved the logic to live inside of the sprite face itself (which, I think, was my plan while writing the rework, I just forgot to actually do it lol), and added a test for the height metric being respected and the re-centering being performed correctly.

This problem came to my attention because of #13221, which didn't go about doing the fix the right way, but did make me realize that it was a problem in the first place (since I had thought that I had already implemented this logic when doing the rework!)

### Verification
https://github.com/user-attachments/assets/4074690b-846e-442d-8ec0-91a34042f6eb

### DIFF
--- a/src/font/sprite/Face.zig
+++ b/src/font/sprite/Face.zig
@@ -565,6 +565,92 @@ test "sprite face render all sprites" {
     try std.testing.expect(!diff); // There should be no diffs from reference.
 }
 
+test "full height cursor sprites respect cursor height metric" {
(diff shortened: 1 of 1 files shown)


mitchellh:
Perfect. Thanks.

PR #13226, opened 2026-07-06
terminal: VT throughput optimizations from real-world dataset (~1.2x to ~3.4x)

mitchellh:
This is a series of seven commits that optimizes VT processing throughput. Each commit is isolated, individually benchmarked, and carries a detailed commit message so please read each for details about each change.

Whereas #13220 was driven by synthetic data, this series was driven by profiling a 2.6 GB recording of real terminal sessions from an asciinema data dump. Through this, I've been able to improve throughput processing the full dump from 276 to 342 MB/s on my system.

> [!NOTE]
>
> **LLM usage:** This series of work was largely driven by Fable 5 and the summaries below started as LLM-written. I've proofread (and mostly modified) every line of work and rewritten everything to be shorter and more in line with how I'd describe a change. Nothing here was unreviewed. I also threw away 3 sets of changes I didn't agree with the maintenance of, but did speed up things a bit.

## The changes

1. **decode ASCII inline in the SIMD scan for ESC**. Real streams call `utf8DecodeUntilControlSeq` on short runs (an escape every ~18 bytes), so ~9% of total time was simdutf setup plus its scalar tail paid per tiny run. The ESC scan and the UTF-8 to UTF-32 "decode" (a widening for ASCII) are now one pass. **Result: +5.4% on the real corpus.**
2. **handle CSI entry bytes inline in consumeUntilGround**. The `[` after ESC and the single `csi_entry` byte each paid a `nextNonUtf8` call, two to three calls for every one of the tens of millions of CSI sequences in the recording. Both transitions are now handled in the consumeUntilGround loop, so a typical CSI parses with no per-byte calls at all. **Result: +2.7% real corpus, +3.1% CSI-heavy stream.**
3. **fill style-only cell runs in bulk in printSliceFill**. The largest single item in the profile (~25%). The print fast path's two scans (run eligibility, simple-cell check) are early-exit search loops LLVM won't vectorize, and real traffic constantly lands in the general path because styled text overwrites cells styled differently (TUI redraws), paying a per-cell release/use pair. The scans are now vectorized and uniformly-styled runs are consumed wholesale: one vector scan, one releaseMultiple/useMultiple pair, one branch-free fill. **Result: +11% real corpus, +21% TUI redraw.**
4. **release style refs per run instead of per cell in clearCells**. Erasing styled rows released each cell's style reference one at a time even though styled cells overwhelmingly share one style per run (status bars, highlighted regions, solid rows). Runs now release with a single releaseMultiple. **Result: +1.1% real corpus, 2.1x on full-screen styled erase.**
5. **log unsupported-input messages once per distinct value**. The recording triggers ~120k warnings, dominated by a few messages some program re-emitted every frame ("unimplemented mode: 34"), each paying formatting plus a blocking writev while adding nothing beyond the first occurrence. A logUnsupportedOnce helper suppresses repeats per (call site, value) using a 64-byte lock-free table per site. **Result: +3.2% on the real corpus, system time halved.**

## Benchmarks

Measured with `ghostty-bench +terminal-stream` (full terminal handler, 120x80 terminal, M4 Max, macOS 26, ReleaseFast, hyperfine means of 6 runs, 64KiB read chunks). These are parser-stage numbers, not end-to-end app numbers.

| stream                        | before  | after   | throughput       | change |
|-------------------------------|---------|---------|------------------|--------|
| real 2.6 GB session recording | 9.441 s | 7.609 s | 276 → 342 MB/s   | 1.24x  |
| ascii (no escapes)            | 119 ms  | 84 ms   | 838 → 1186 MB/s  | 1.41x  |
| TUI redraw (rotating styles)  | 417 ms  | 293 ms  | 240 → 342 MB/s   | 1.42x  |
| styled paint + ED 2 erase     | 418 ms  | 124 ms  | 239 → 808 MB/s   | 3.38x  |
| csi mix (random-color SGR/CUP)| 695 ms  | 696 ms  | (adversarial)    | ~1.0x  |

Note the "csi mix" benchmark above was a generated adversarial input e.g. a worst-case input for the changes we made. It wasn't based in real-world data or expectations. But I asked for it to be done so we can verify we don't see regressions too much (and were able to verify we see basically none).

### DIFF
--- a/src/benchmark/TerminalStream.zig
+++ b/src/benchmark/TerminalStream.zig
@@ -112,11 +112,14 @@ fn step(ptr: *anyopaque) Benchmark.Error!void {
     // aren't currently IO bound.
     const f = self.data_f orelse return;
 
-    var read_buf: [4096]u8 align(std.atomic.cache_line) = undefined;
+    var read_buf: [64 * 1024]u8 align(std.atomic.cache_line) = undefined;
     var f_reader = f.reader(&read_buf);
     const r = &f_reader.interface;
 
-    var buf: [4096]u8 = undefined;
+    // This buffer size matches the read buffer size used by the
+    // real IO thread (see termio Exec.zig buffer_capacity) so that
+    // the benchmark exercises the stream with realistic chunk sizes.
+    var buf: [64 * 1024]u8 = undefined;
     while (true) {
         const n = r.readSliceShort(&buf) catch {
             log.warn("error reading data file err={?}", .{f_reader.err});
--- a/src/simd/vt.cpp
+++ b/src/simd/vt.cpp
@@ -198,6 +198,92 @@ size_t DecodeUTF8(const uint8_t* HWY_RESTRICT input,
   return static_cast<size_t>(out - output);
 }
 
+// Widen the N uint8 lanes of v into N uint32 values stored at out.
+// This is the UTF-8 to UTF-32 "decode" for ASCII bytes.
+template <class D>
+static HWY_INLINE void WidenAsciiStore(D d,
+                                       hn::Vec<D> v,
+                                       char32_t* HWY_RESTRICT out) {
+  uint32_t* HWY_RESTRICT out32 = reinterpret_cast<uint32_t*>(out);
+#if HWY_TARGET == HWY_SCALAR
+  // The scalar fallback target has single-lane vectors, which cannot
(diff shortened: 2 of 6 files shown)


mitchellh:
No plain-text throughput improvements here (tested through catting files). As expected, most of this work was towards actual sequences.

PR #13236, opened 2026-07-07
termio: bound POSIX read-ahead on non-Darwin

EriksRemess:
The pipelined POSIX pty reader uses multiple large gather buffers to avoid stalling on Darwin, where pty reads are capped around 1KiB. On Linux this can let bulk terminal producers run too far ahead of terminal parsing/rendering.

Frame-style terminal apps such as DOOM-fire can then report very high producer FPS while Ghostty displays stale frames from the buffered stream.

Keep the Darwin-tuned 4 x 64KiB pipeline, but reduce non-Darwin read-ahead to 2 x 8KiB so the pty can apply backpressure before multiple frames are queued.

AI disclosure: Codex gpt-5.5 xhigh helped.

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -1281,15 +1281,21 @@ pub const ReadThread = struct {
     /// stages. The gather stage can run at most this many batches
     /// ahead of the parse stage before it blocks, which (via the
     /// kernel pty queue) is also what preserves flow control to the
-    /// child. Empirically chosen through measurements on an M4 Max.
-    /// Less than 4 there are minor slowdowns, above 4 there are no
-    /// improvements.
-    const buffer_count = 4;
+    /// child.
+    ///
+    /// The Darwin value was empirically chosen through measurements on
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
This sounds conceptually right to me. As I noted in my PR I was already unsure about these buffers on Linux. 😄

mitchellh:
I'm willing to merge this as-is since we have no measurement on Linux anyways, but it'd be nice to get before/after on large ascii/unicode/doom fire results. too.

mitchellh:
> Frame-style terminal apps such as DOOM-fire can then report very high producer FPS while Ghostty displays stale frames from the buffered stream.

Also sorry let me push back on this: the state is stale but the steady-state FPS that doom fire reports _is accurate_. That is how much data we're processing, its just not up to date on what is rendered. Kind of like: if a hose is pushing 1 gal/hr 1 mile away and the spicket is also outputting water at 1 gal/hr, the throughput is still accurate despite the water being delayed. This is what DOOM-fire represents to me. Its not a render benchmark.

PR #13237, opened 2026-07-07
termio: don't bridge pty reads while the parser is idle

mitchellh:
Fixes a frame time regression reported with fortio's `fps -fire` benchmark (fortio.org/terminal/fps): frame times nearly doubled at typical grid sizes after #13209 (the pipelined pty reader), and were more than 5x worse at small grids.

## The problem

The `fps -fire` program follows a request/response pattern: write a frame, end it with a cursor position query (CSI 6n), and block until the reply arrives before starting the next frame. 

The gather stage treats any burst of 1 KiB or more as a saturated stream. When its spin retries come up dry, it sleeps in a 1ms poll expecting the writer's next refill so it can publish fewer, larger batches. But a frame-synced writer will never refill here: it is blocked waiting for a reply to a query that is sitting inside the very batch the gather stage is holding back. The poll always sleeps its full timeout, adding ~1.2ms to every round trip. Ouch!

## The fix

Sleeping for a refill gap is only free while the parse stage is busy, since the wait hides behind parse time. Once the parser is idle, every additional microsecond spent bridging is added straight to output latency. So:

1. When the spin retries exhaust and the parse stage is idle, deliver the batch immediately instead of polling.

2. When the parse stage is busy, use a `pipe2` pipe to allow the parser to notify the gather thread it is idle. In the middle of the `poll` loop and sleep, we can get interrupted immediately and deliver the batch.

The pipe is only written while the gather stage is actively polling, so an interactive terminal never pays the syscalls, and a saturated stream never idles the parser, preserving full batching and throughput for bulk output. Win, win, win!

## Benchmarks

| workload | pre-#13209 | main | this PR |
|---|---|---|---|
| fps -fire 80x24 | 0.262 ms / 3674 fps | 1.435 ms / 694 fps | 0.234 ms / 4106 fps |
 | fps -fire 160x45 | 1.012 ms / 975 fps | 1.823 ms / 545 fps | 0.701 ms / 1405 fps | 
| fps -fire 284x68 | 2.443 ms / 407 fps | 2.391 ms / 414 fps | 1.351 ms / 732 fps |
 | cat 19.3 MB | 0.204 s (95 MB/s) | 0.086 s (224 MB/s) | 0.082 s (236 MB/s) |

## LLM Notes

Bisect script written by hand and found the offending commit. Fable 5 found the likely cause. I manually came up with the proposed solution and wrote it out. Fable helped benchmark for me to verify my assumptions conceptually and in the real world. Commit message written by hand except for the benchmark results.

### DIFF
--- a/src/termio/Exec.zig
+++ b/src/termio/Exec.zig
@@ -1363,6 +1363,23 @@ pub const ReadThread = struct {
         tail: usize = 0,
         count: usize = 0,
 
+        /// Set by the gather stage (under the mutex) while it sleeps
+        /// in a bridge poll. The parse stage only writes to the idle
+        /// pipe when this is set, so an interactive terminal never
+        /// pays the pipe syscalls.
+        bridging: bool = false,
+
+        /// A self-pipe the parse stage uses to interrupt the gather
+        /// stage's bridge poll the moment it runs out of batches.
+        /// Bridging a refill gap is only free while the parse stage
+        /// is busy; once it goes idle, every additional microsecond
+        /// spent bridging is added straight to output latency. The
+        /// write end is used by the parse stage, the read end is
+        /// polled by the gather stage. -1 when unavailable, in which
+        /// case bridge polls are bounded by their timeout only.
+        idle_read_fd: posix.fd_t = -1,
+        idle_write_fd: posix.fd_t = -1,
+
         /// Set by the gather stage when the stream is over (quit
         /// signal, EOF, or pty error). The parse stage drains any
         /// remaining batches and then exits.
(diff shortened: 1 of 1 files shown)


src/termio/Exec.zig
```diff
@@ -1363,6 +1363,23 @@ pub const ReadThread = struct {
...
+        /// A self-pipe the parse stage uses to interrupt the gather
+        /// stage's bridge poll the moment it runs out of batches.
+        /// Bridging a refill gap is only free while the parse stage
+        /// is busy; once it goes idle, every additional microsecond
+        /// spent bridging is added straight to output latency. The
+        /// write end is used by the parse stage, the read end is
+        /// polled by the gather stage. -1 when unavailable, in which
+        /// case bridge polls are bounded by their timeout only.
+        idle_read_fd: posix.fd_t = -1,
+        idle_write_fd: posix.fd_t = -1,
```
mitchellh, line 1381: Note, I don't use optionals here on purpose. File descriptors are already positive if valid and optionals expand our needed size unnecessarily because of it.

PR #13238, opened 2026-07-07
fix(lib-vt): pass RGB color inputs by pointer

elias8:
Fixes C ABI compatibility issue on arm64 macOS by passing color by reference instead of by value. I ran into an [issue](https://github.com/elias8/libghostty/actions/runs/28889437162/job/85698207979?pr=93) while working on the Dart bindings, where tests passed on Linux and x86_64 macOS locally, but failed on arm64 macOS. The Dart FFI appears to lower the argument (3 byte struct) differently on arm64 macOS. So the change addresses that by changing the arguments to by reference and aligns with the preexisting APIs.

_AI disclosure: I used codex 5.5 to assist with debugging and making the changes. However, I have reviewed all changes manually and verified by running the tests both locally and on CI with the fix applied._

### DIFF
--- a/include/ghostty/vt/color.h
+++ b/include/ghostty/vt/color.h
@@ -88,8 +88,8 @@
  * ghostty_color_palette_generate(
  *     palette,
  *     &skip,
- *     background,
- *     foreground,
+ *     &background,
+ *     &foreground,
  *     true,
  *     palette);
  * @endcode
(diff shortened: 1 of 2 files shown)


mitchellh:
Oh!! I ran into this and couldn't figure out what was going on. Thank you.

PR #13245, opened 2026-07-08
terminal: return free-listed page memory to the OS

mitchellh:
This uses `madvise` flags on both macOS and Linux to mark our unused pool items as free memory and not show up in RSS (until they're actually used). We can do this since we do page-aligned/page-multiple allocations, heyo!

Background: The PageList page pool never returns memory to the OS: destroyed pages are zeroed and free-listed until the surface exits. Any operation that shrinks the page count (clearing scrollback, pruning churn, resets, reflow) therefore retains its high-water RSS forever. Clearing a full scrollback keeps all of it resident, which at the default 10MB scrollback-limit is 10MB per terminal of memory that can never be used again for anything else.

Lots of memes on the internet about this, and it turns out operating systems give us an answer for this (both Linux and macOS at least), so let's do it kids.

On Linux, madvise(MADV_DONTNEED) reclaims the pages immediately and guarantees zero-fill on the next touch, which also lets us skip the zeroing memset entirely, making destroy cheaper.

On macOS, we zero in place and mark the item MADV_FREE_REUSABLE, which removes it from the process footprint immediately. Reuse is paired with MADV_FREE_REUSE when the pool hands a buffer back out so that footprint accounting stays correct. The zero invariant required by page reuse holds either way: reusable page contents are either preserved (our zeroes) or reclaimed and zero-filled by the kernel.

Other platforms and test builds keep the existing memset behavior.

## LLM Notes

Fable 5 found the retention behavior while re-analyzing scrollback memory, wrote the change and tests, and verified the madvise semantics empirically with memory probes on macOS and a real Linux kernel, plus before/after throughput benchmarks on both. I reviewed the analysis, the diff, rewrote the code to be more idiomatic Zig, and wrote this commit message you're reading.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -480,7 +480,9 @@ fn initPages(
 
         const page_buf = if (pooled) buf: {
             try tw.check(.page_buf_std);
-            break :buf try pool.pages.create();
+            const buf = try pool.pages.create();
+            recommitPoolItem(buf);
+            break :buf buf;
         } else buf: {
             try tw.check(.page_buf_non_std);
             break :buf try page_alloc.alignedAlloc(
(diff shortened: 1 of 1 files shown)


mitchellh:
Btw, verified this on both macOS and Linux by filling scrollback and doing a clear screen. Without this PR, memory shows up in top/footprint/etc. After this PR, it doesn't.

PR #13250, opened 2026-07-08
macos: fix quick terminal restoring stale size after display reconnect

quinnypig:
"Why is my quick terminal not taking up the entire top of my docked Mac screen after I reconnect?" Boy howdy are you in the right PR. 

It turns out that the  quick terminal caches its last-closed window frame per display so it can restore the user's size when reopened. The cache entry was considered valid whenever the current screen was the same size *or larger* than when the frame was saved ("persist when screens grow"). This has led to a pattern that was simply maddening. To wit: that rule breaks across display changes. 

When an external display is disconnected and later reconnected at a different resolution (common after traveling with a laptop, do not even get me started on projectors) the same display can come back larger than when the frame was cached. The stale frame is still treated as valid and restored, so the quick terminal no longer fills the screen (it appears at a partial width/height). Because the cache is persisted, restarting Ghostty does not clear it, and the user is slowly driven mad. Welcome to madness; we have snacks.

This PR addresses this by treating a cached frame as valid when the screen geometry matches exactly (both backing scale factor and frame size). On any mismatch we drop the entry and fall back to the configured quick-terminal-size. Manual resizes are still remembered across toggles within a stable display configuration.

Fixes the regression reported in #12348.

AI disclaimer: I used AI for this. Of course I used AI for this, my code is terrible on a good day. Specifically, Claude Code, as well as a custom harness that has the curious tendency to write commit messages containing conspiracy theories about the code because I am history's greatest monster.  

Fight me!

### DIFF
--- a/macos/Sources/Features/QuickTerminal/QuickTerminalScreenStateCache.swift
+++ b/macos/Sources/Features/QuickTerminal/QuickTerminalScreenStateCache.swift
@@ -69,12 +69,12 @@ class QuickTerminalScreenStateCache {
         for screen in screens {
             guard let key = screen.displayUUID else { continue }
             if var entry = stateByDisplay[key] {
-                // Drop on dimension/scale change that makes the entry invalid
+                // Drop on dimension/scale change that makes the entry invalid. The
+                // saved frame is only meaningful for the exact geometry it was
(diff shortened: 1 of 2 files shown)


mitchellh:
Thanks Corey. The description sounds right and the code looks good. ❤️

PR #13253, opened 2026-07-08
Revert "termio: bound POSIX read-ahead on non-Darwin"

rockorager:
This reverts commit bed47168ca7f34fe0a27e9f13c46b8df97cb77ca.

This commit reduced the buffer size and count on non-Darwin. This introduced a ~20% perf loss on linux for high throughput loads on a questionable claim. Producers of bytes should not be pacing frames based on how fast they can write to the terminal.


EriksRemess:
back for DOOM-fire to produce slideshow and make Ghostty unresponsive. `fps -fire` also lost around 25% of fps due to revert :sad:

mitchellh:
Yes, we need to find the right solution for that. This was empirically shown by multiple people to degrade Linux perf.

PR #13260, opened 2026-07-09
macOS:  bundle a build-time-generated Help Book

bo2themax:
With this, you can now search config options, keybind actions, and cli help.

**Inspired by `+explain-config` and idea of HelpBook from @jparise!**

The diff might be overwhelming, but it doesn't affect terminals. The main part of this pr is basically HTML generation and bundling them into a [.help bundle](https://developer.apple.com/library/archive/documentation/Carbon/Conceptual/ProvidingUserAssitAppleHelp/authoring_help/authoring_help_book.html#//apple_ref/doc/uid/TP30000903-CH206-SW5). This was first introduced in OS X and is archived, but it still works with Tips.app on macOS 27.

|  |  |  |
|:---:|:---:|:---:|
| tips-badge-dark | tips-badge-light | image |

### AI Closure

Claude implemented most of it. I just learned how help books work and did my best to let Claude document. 

**TBH I didn't read every line of changes in those html/css/js files**, but I tested a lot of random pages; it looks good. 

I did review every Zig change after a few rounds of review from different Claude sessions; I couldn't see anything bad.


jparise (commented):
I think the `CApi.zig` side could use some work. It looks more intrusive that the feature probably warrants. Or maybe some of this code shouldn't be in `CAap.zig` itself.

The other pattern that could be interesting is generating the macOS help strings at build time (and producing a `.plist` or similar resource) and consuming it in the macOS app.

macos/Sources/App/macOS/AppDelegate+HelpSearch.swift
```diff
@@ -0,0 +1,231 @@
...
+
+            switch result.kind {
+            case .option:
+                if !result.valueType.isEmpty {
+                    alert.informativeText = "Type: \(result.valueType)"
+                }
+            case .keybind:
+                if let shortcut = self.ghostty.config.keyboardShortcut(for: result.name)?.description {
+                    alert.informativeText = "Keybind: \(shortcut)"
+                }
```
jparise: Maybe we could look at more distinct styling for these fields?

bo2themax: By "distinct" you mean using a same prefix?

bo2themax:
> I think the `CApi.zig` side could use some work. It looks more intrusive that the feature probably warrants. Or maybe some of this code shouldn't be in `CAap.zig` itself.
>
> The other pattern that could be interesting is generating the macOS help strings at build time (and producing a `.plist` or similar resource) and consuming it in the macOS app.

Plist generation is a good idea. I thought the changes in CApi.zig are already generating those entry in compile time.

macos/Sources/App/macOS/AppDelegate+HelpSearch.swift
jparise: Sorry, I meant more _visually_ distinct. And also using macOS conventions for [displaying the keys](https://support.apple.com/guide/mac-help/what-are-those-symbols-shown-in-menus-cpmh0011/mac).

bo2themax: Make senses! I forgot we already have this in command palette

jparise:
> > I think the `CApi.zig` side could use some work. It looks more intrusive that the feature probably warrants. Or maybe some of this code shouldn't be in `CAap.zig` itself.
> > The other pattern that could be interesting is generating the macOS help strings at build time (and producing a `.plist` or similar resource) and consuming it in the macOS app.
> 
> Plist generation is a good idea. I thought the changes in CApi.zig are already generating those entry in compile time.

Or perhaps generate a [Help Book](https://developer.apple.com/library/archive/documentation/Carbon/Conceptual/ProvidingUserAssitAppleHelp/authoring_help/authoring_help_book.html) directly. I haven't used that format before, though, so I'm not sure how it trades off against the control of doing something more custom and integrated.

src/build/GhosttyResources.zig
```diff
@@ -125,6 +132,42 @@ pub fn init(b: *std.Build, cfg: *const Config, deps: *const SharedDeps) !Ghostty
...
+    // Installed outside share/ghostty so the app bundle doesn't pick it up
+    // twice (the Xcode project references both directories). The host
+    // check is because hiutil is a macOS host tool.
+    if (cfg.target.result.os.tag == .macos and
+        b.graph.host.result.os.tag == .macos)
+    {
+        const run = b.addRunArtifact(build_data_exe);
+        run.addArg("+help-book");
+        const book_dir = run.addOutputDirectoryArg("Ghostty.help");
+        const lproj = book_dir.path(b, "Contents/Resources/en.lproj");
```
jparise, line 145: Is there a path to supporting multiple languages?

src/extra/help_book.zig
```diff
@@ -0,0 +1,473 @@
...
+
+/// Write the complete help book bundle contents into the given directory,
+/// which becomes the `Ghostty.help` bundle root.
+pub fn write(alloc: Allocator, dir: std.fs.Dir) !void {
+    var arena_state = std.heap.ArenaAllocator.init(alloc);
+    defer arena_state.deinit();
+    const arena = arena_state.allocator();
+
+    try dir.makePath("Contents/Resources/en.lproj");
+    try dir.writeFile(.{ .sub_path = "Contents/PkgInfo", .data = "BNDLhbwr" });
```
jparise: I think `BNDLhbwr` is magical enough to warrant a comment or constant.

src/extra/help_book.zig
```diff
@@ -0,0 +1,473 @@
...
+        while (end > 0 and (line[end] & 0xC0) == 0x80) end -= 1;
+        line = line[0..end];
+    }
+    for (line) |c| {
+        if (c == '`') continue;
+        try markdown.writeEscapedByte(w, c);
+    }
+}
+
+test "help book" {
```
jparise, line 361: These tests don't look great to me. They're pretty heavy on asserting specific string content.

jparise (commented):
Nice to see this working so well!

src/build/GhosttyResources.zig
bo2themax: We could, but I think since Ghostty itself is not fully localized on macOS yet, we can leave that. In the future, we can just add more `lproj` to support other languages.

jparise (approved):
The overall approach and output look good to me. I think we can keep iterating on the format itself based on usage and feedback.

I think the Zig side could use a more thorough review; I haven't spent much time looking at it, but I can if others are short on time.

jparise (commented):
The docstring reformatting changes in `src/cli/` all look good regardless. Maybe move them into their own PR that we can more quickly land?

src/build/GhosttyResources.zig
```diff
@@ -125,6 +132,42 @@ pub fn init(b: *std.Build, cfg: *const Config, deps: *const SharedDeps) !Ghostty
...
+        const wf = b.addWriteFiles();
+        _ = wf.addCopyDirectory(book_dir, "Ghostty.help", .{});
+        _ = wf.addCopyFile(lsm_out, "Ghostty.help/Contents/Resources/en.lproj/Ghostty.helpindex");
+        _ = wf.addCopyFile(cs_out, "Ghostty.help/Contents/Resources/en.lproj/Ghostty.cshelpindex");
+
+        const install_step = b.addInstallDirectory(.{
+            .source_dir = wf.getDirectory(),
+            .install_dir = .prefix,
+            .install_subdir = "help",
+        });
```
jparise, line 167: Will this clean up renamed/removed help pages from the target director?

bo2themax: Yeah, since `zig-out` is ignore and it will generate on every release. Locally you'll need to remove it tho.

PR #13261, opened 2026-07-09
termio: hand off state capture to renderer under hot output

EriksRemess:
When alternate-screen output saturates the terminal, the parser can repeatedly reacquire the terminal-state mutex before the renderer can snapshot state, causing frame bursts and stutters. Add a renderer-wait handoff so the parser waits briefly while the renderer is capturing, then proceed in small chunks to reduce lock contention.

Use an atomic render-pending flag to avoid redundant renderer wakeups, and update the style table growth path to reuse dead style IDs before triggering rehash when near capacity.

Everything here up for discussion, no merge expected. 

Assisted by Codex 5.5

### DIFF
--- a/src/renderer/State.zig
+++ b/src/renderer/State.zig
@@ -30,6 +30,15 @@ preedit: ?Preedit = null,
 /// need about the mouse.
 mouse: Mouse = .{},
 
+/// True while a renderer wakeup has been queued but not yet consumed.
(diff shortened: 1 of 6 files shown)


EriksRemess:
fixed by https://github.com/ghostty-org/ghostty/pull/13265

PR #13264, opened 2026-07-09
terminal: scrollback page compression (70 to 90% memory savings)

mitchellh:
This adds transparent compression for non-active/non-viewport scrollback pages, reducing physical memory for compressed pages by anywhere from 70% to 90%. Compression is obviously highly dependent on the shape of the data, but these are the numbers I got for normal scrollback.

Due to compression being available, I bumped the default scrollback limit from 10MB to 50MB. On average, a full scrollback still uses less memory than the prior limit due to the compression ratios.

## Demo

Here is a demo video showing me filling my scrollback and using the inspector so you can see the live compression/decompression activity and results:

https://github.com/user-attachments/assets/7b9d0383-42f7-47bf-8b3f-853e3f89549c

## Resident vs. Virtual Memory

This PR works by lowering _resident/physicalmemory, but doesn't touch _virtual_ memory.

Practically what this means is that users need to make sure they're looking at resident memory to see the change.

We use OS primitives like `MADV_DONTNEED` on Linux or `MADV_FREE_REUSABLE` on Darwin to discard our physical memory, but retain our virtual memory allocations. This is awesome because it means our decompression is infallible: the OS has already given us the memory, but it just remaps it at that point.

This is baked into the core implementation, so compression only works on systems that support an OS primitive to retain virtual mappings while discarding physical. Today, that is macOS and 64-bit Linux. Other operating systems have support we just haven't coded it up yet.

## Implementation

A major refactor had to happen to PageList to represent nodes as either resident or compressed. Pins typically accessed node content directly so we had to add a bunch of helpers to read metadata without decompression (but some access requires it).

Compression is relatively slow and its important we don't impact IO performance. So we support incremental compression passes and they only run when the terminal is idle (250ms timer on the render thread that resets on any activity). Benchmarks show zero regression in IO throughput on this change.

In order to maintain the no-libc invariant for libghostty-vt, we use a hand-written (an AI assisted optimization) LZ4 compression implementation. The performance and compression ratio is _okay_. Its a good first step for this. I think in the future I want to look at other implementations we can bring in based on build configuration. 

## Performance

Measured with a saved 7.3 MB corpus made by repeating `gh --help` output into a 120x80 terminal with a 50 MB scrollback limit on my machine:

| compression measurement | result |
|-------------------------|--------|
| pages compressed | 121 |
| raw page backing | 49.56 MB |
| encoded storage | 3.03 MB (6.11% of raw) |
| estimated physical memory savings | 46.53 MB (93.89%) |
| full compression | 30.3 ms total, 12.2 ms over the 18.1 ms no-op baseline (~101 µs/page) |
| incremental drain | 29.7 ms total, 11.6 ms over baseline (~96 µs/page) |
| compress and restore | 33.5 ms total, 3.2 ms over full compression (~26 µs/page to restore) |

The workload above is especially repetitive, so its 6.11% encoded ratio is better than the 10% to 30% expected for text-heavy terminal history in general. Steady-state throughput is unchanged within noise (`terminal-stream` benchmarks and manual `cat` timings). 

## libghostty-vt

The same caller-driven compression controls are exposed to Zig and C. 

Note that compression _is not automatic_ for libghostty users. Callers must determine their own definition of "idle" and when to compress and call our incremental callback APIs to perform the compression. Decompression is automatic and as-needed (and will trigger recompression-required flags so callers are aware).

## LLM Notes

This work was done in concert with Codex. I reviewed and rewrote/reshaped pretty much every change extensively, particularly PageList/Terminal. This PR message is written by hand, commit messages are LLM written but reviewed.


mitchellh:
The main blocker on this is I want to do a bigger corpus of differential testing plus fuzzing on the lz4 implementation to make sure its solid. I'm daily driving this and haven't seen any issues but that of course is extremely anecdotal. I'm very happy with the overall shape of it.

PR #13265, opened 2026-07-09
renderer: hand off state mutex to avoid starving frames

Uzaaft:
The mutex is unfair. From my understanding (after a brief convo with @rockorager), it's a race between the parse thread and the renderer thread. 

The parse thread is unlocking it then locking it faster than the renderer thread can get a lock/unlock. Under sustained pty output the parse thread never sleeps between batches, so the renderer's frame snapshot can starve for as long as the output lasts.

The fix here makes the parse thread voluntarily stay off the lock until the renderer has had one turn by introducing two atomics:

- `demand`: a waiter count. The renderer increments it before locking and decrements it after acquiring, so demand > 0 means "the renderer is queued on the lock or about to be."
- `handoff_gen`: a generation counter. The renderer bumps it (with a futex wake) after unlocking, meaning "a waiter completed one full turn."

At each batch boundary the parse thread checks demand (a single relaxed load in the common case, so this **should** cost nothing when the renderer isn't waiting). If a waiter exists, it futex-waits on handoff_gen until the renderer has taken and released the lock, bounded by a 1ms timeout, trying to ensure a lost wake can't stall IO. 

This is inspired from `parking_lot` form rust land, which has eventual fairness, but applied only at the one site that misbehaves.

## LLM Note

I used Fable 5 in Amp to write the code, but only after I identified the problem myself from previous experience with mutex fairness (and the convo above with rockorager). 

The diagnosis — the parse thread barging the unfair mutex and starving the renderer — came first.Fable then traced the exact loop in `Exec.zig/generic.zig` and implemented the handoff protocol. 

I've reviewed the changes.

### DIFF
--- a/src/renderer/State.zig
+++ b/src/renderer/State.zig
@@ -1,11 +1,14 @@
 //! This is the render state that is given to a renderer.
 
 const std = @import("std");
+const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
 const Inspector = @import("../inspector/main.zig").Inspector;
 const terminalpkg = @import("../terminal/main.zig");
 const inputpkg = @import("../input.zig");
 const renderer = @import("../renderer.zig");
+
+const State = @This();
 
 /// The mutex that must be held while reading any of the data in the
 /// members of this state. Note that the state itself is NOT protected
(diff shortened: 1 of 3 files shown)


src/renderer/State.zig
```diff
@@ -1,12 +1,15 @@
...
 
 const std = @import("std");
+const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
 const Inspector = @import("../inspector/main.zig").Inspector;
 const terminalpkg = @import("../terminal/main.zig");
 const inputpkg = @import("../input.zig");
 const renderer = @import("../renderer.zig");
 
+const State = @This();
```
mitchellh: Nit: I usually put this at the top of the file (under the `//!` comments)

mitchellh (changes requested):
Pretty simple. Any way you can run (or prompt an agent to run) benchmarks for us against a baseline to see if this affects IO throughput. Since this requires the renderer, the only way to do it is via UI automation. Fable is good at this, it has to launch ghostty, run `cat`, use `/dev/tty` (the proper one) to monitor, and measure its results.

Uzaaft:
> Pretty simple. Any way you can run (or prompt an agent to run) benchmarks for us against a baseline to see if this affects IO throughput. Since this requires the renderer, the only way to do it is via UI automation. Fable is good at this, it has to launch ghostty, run `cat`, use `/dev/tty` (the proper one) to monitor, and measure its results.

From fable's run:

image

> 
> - Two ReleaseFast/ReleaseLocal Ghostty.app builds on an M-series Mac (macOS 27): baseline = 7cb44fea3 (upstream main), handoff = d34b54e9b (this PR). Binaries verified distinct.
> - Each run launches the app for real (visible window, live renderer thread, 120×40 cells, default config otherwise) with command set to a script that warms up with 1MB, then times cat of a 1GiB base64 text file to the pty via /usr/bin/time -p. quit-after-last-window-closed ends the run.
> - 5 runs per build, alternating A/B to neutralize thermal drift. Page cache pre-warmed so disk isn't in the loop.
>

PR #13278, opened 2026-07-10
Misc runtime safety fixes

mitchellh:
All found by GPT 5.6. I'm still going through manual review of each one now and will remove or rewrite the ones that are pointless or unreachable (if any, I prompted it to ignore those too).


mitchellh:
Reviewed all of them. Removed 2 fixes that are assertions instead and affect performance. The last two are also sus so I'm doing some more checking on those. The rest are very good, especially given libghostty-vt!

mitchellh:
Yeah the last 2 were unsafe, removed them.

PR #13282, opened 2026-07-10
terminal: update page serial to be treated as a generation marker (correctness)

mitchellh:
This changes our `page_serial` semantics from monotonic min to a generation marker (any change means you should reload). This fixes a number of real bugs that search had.

In reality, this invariant was already broken and not true. So this PR comes to reality with that and fixes all our usage. The most visible failures were panics when selecting cached search results after partially erasing a history page or splitting a page, but the same bug class affected viewport search fingerprints and asynchronous render highlights.

It was mostly search, for now, but this type of fix is important for some upcoming work I'm playing around with regarding deferred reflow and disk offload.

## Page Serial Semantics

`Node.serial` is now explicitly a page generation. It changes whenever a node is allocated or reused, and whenever an in-place mutation changes the meaning or valid range of its row coordinates.

`page_serial_min` is renamed to `page_serial_epoch` to reflect its remaining purpose. Only a whole-list reset advances the epoch. A serial below the epoch is definitely stale and can be rejected in O(1); a serial at or above it is only potentially valid and is checked against the live list using its captured pointer and generation.

## Performance

The generation bump is constant work per node whose layout is already being changed. There is no additional page-list scan on normal row append or scrollback pruning.

I did add a new field to the render state, so I ran render state benchmarks we already have that shows a very slight slowdown but this is acceptable especially given recent speedups outweigh this significantly:

| render measurement | before | after | change |
|--------------------|--------|-------|--------|
| full rebuild | 2.915 us/update | 2.926 us/update | +0.27% |
| lock-held rebuild | 2.472 us/update | 2.489 us/update | +0.53% (within noise) |
| clean update | 19.03 ns/update | 19.30 ns/update | +1.43% (+0.27 ns) |
| one dirty row | 52.97 ns/update | 53.40 ns/update | +0.96% (within noise) |
| flattened highlight miss | 86.77 ns/update | 86.87 ns/update | +0.19% (within noise) |
| flattened highlight match | 88.38 ns/update | 90.26 ns/update | +2.00% (+1.88 ns) |

## LLM Notes

This work was done in concert with Codex. I reviewed and reshaped the serial model, implementation, tests, documentation, and performance analysis throughout. This PR message is hand-written.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -379,16 +379,17 @@ pages: List,
 /// going to risk it.
 page_serial: u64,
 
-/// A conservative lower bound on live page generations. This is an epoch for
-/// whole-list invalidation, not the generation of the first page or the exact
-/// minimum live generation. Page generations are not monotonic in list order:
-/// replacement and split operations can put a fresh generation before older
-/// live pages.
+/// The first serial in the current whole-list validity epoch. Only `reset`
+/// advances this value, immediately before rebuilding every page. It is not
+/// the generation of the first page or the exact minimum live generation.
+/// Page generations are not monotonic in list order: replacement and split
+/// operations can put a fresh generation before older live pages.
 ///
-/// A generation below this value is definitely invalid. A generation at or
-/// above it is only potentially valid and must still be checked against the
-/// live list with `nodeIsValid` before its coordinates are used.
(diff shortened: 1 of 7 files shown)


mitchellh:
Found a performance pitfall in search, fixing that then will reassess....

PR #13292, opened 2026-07-11
terminal: reduce page map tombstone probing

rockorager:
Page-local hyperlink and grapheme maps retain tombstones after entries
are removed. Repeated clear, redraw, and cell movement can therefore
turn operations on otherwise sparse maps into full probe-chain scans.

Use no-clobber insertion when moving managed cell data. Those call sites
already guarantee that the destination is absent, so searching the probe
chain for a duplicate is unnecessary.

Also port Zig's allocation-free `HashMapUnmanaged.rehash`. A canonical
get-or-put probe that encounters at least one eighth of the table as
tombstones rebuilds the map and retries before returning entry pointers.
Adapted lookups retain their old behavior because they lack the context
needed to hash resident keys.

Rehashing does not allocate, change page layout, or move Page, Row, or
Cell objects. It can move map entries, so get-or-put now explicitly
documents that existing key and value pointers may be invalidated.

## Performance

Both benchmarks used a ReleaseFast `ghostty-bench +terminal-stream`
build and Hyperfine with 5 warmups and 30 measured runs.

### Known-absent cell movement

A 4x128 terminal first touched and erased every cell with one stable
explicit OSC 8 link, populating the default 64-slot map with tombstones.
It then alternated ICH and DCH two million times to move one linked cell
through blanks.

| implementation | mean | standard deviation |
|---|---:|---:|
| clobbering insertion | 366.4 ms | 17.3 ms |
| no-clobber insertion | 119.2 ms | 19.4 ms |

The no-clobber path is **3.07 times faster**.

### Clear and redraw churn

A 24x80 alternate screen ran 100,000 frames. Each frame cleared the
screen and drew 32 linked cells on a changing row using one stable
explicit OSC 8 link.

| implementation | mean | standard deviation |
|---|---:|---:|
| before in-place rehash | 966.3 ms | 25.4 ms |
| after in-place rehash | 279.0 ms | 19.2 ms |

The rehashed version is **3.46 times faster**.

### Primary-screen controls

A control printed 100,000 linked lines on the primary screen. The
no-clobber comparison measured 1.500 s before and 1.475 s after. The
rehash comparison measured 1.470 s before and 1.461 s after. Both are
within noise, as expected for ordinary full-width scrolling.

## Testing

```
mise exec zig@0.15.2 -- zig build -Dapp-runtime=none test
mise exec zig@0.15.2 -- zig build -Dapp-runtime=none -Demit-bench -Doptimize=ReleaseFast
```

Hash-map coverage includes full occupancy, real hash placement,
fragmented canonical probes, adapted contexts, and complete tombstone
removal.

### DIFF
--- a/src/terminal/hash_map.zig
+++ b/src/terminal/hash_map.zig
@@ -646,6 +646,8 @@ fn HashMapUnmanaged(
             return null;
         }
 
+        /// The get-or-put family may rehash a fragmented table. Any key or
+        /// value pointers previously returned by this map may be invalidated.
         pub fn getOrPut(self: *Self, key: K) Allocator.Error!GetOrPutResult {
             if (@sizeOf(Context) != 0)
                 @compileError("Cannot infer context " ++ @typeName(Context) ++ ", call getOrPutContext instead.");
@@ -705,6 +707,7 @@ fn HashMapUnmanaged(
             var idx = @as(usize, @truncate(hash & mask));
 
             var first_tombstone_idx: usize = self.capacity(); // invalid index
+            var tombstones: Size = 0;
             var metadata = self.metadata.? + idx;
             while (!metadata[0].isFree() and limit != 0) {
                 if (metadata[0].isUsed() and metadata[0].fingerprint == fingerprint) {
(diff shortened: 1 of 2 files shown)


mitchellh:
Incorporated in #13294

PR #13310, opened 2026-07-12
add nushell complete attribute

ruoyouz:
Added correct attribute so nushell can generate completions for ghostty redefined commands (i.e., `ssh` and `sudo`).

Fixes #12008

### DIFF
--- a/src/shell-integration/nushell/vendor/autoload/ghostty.nu
+++ b/src/shell-integration/nushell/vendor/autoload/ghostty.nu
@@ -6,6 +6,8 @@ export module ghostty {
 
   # Wrap `ssh` with `ghostty +ssh` and translate the shell-integration
   # feature flags into command options.
+  # Add attribute so nushell completion can function correctly
+  @complete external
   export def --wrapped ssh [...args] {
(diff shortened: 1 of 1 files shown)


src/shell-integration/nushell/vendor/autoload/ghostty.nu
```diff
@@ -6,6 +6,8 @@ export module ghostty {
 
   # Wrap `ssh` with `ghostty +ssh` and translate the shell-integration
   # feature flags into command options.
+  # Add attribute so nushell completion can function correctly
```
jparise: Let's remove this comment (and the one below) before we consider merging this. I don't think it belongs in the function's documentation block.

jparise (changes requested):

src/shell-integration/nushell/vendor/autoload/ghostty.nu
ruoyouz: thank you for the feedback. i have removed both comments

jparise (approved):

jparise:
Thanks!

PR #13311, opened 2026-07-13
terminal/search: don't clear storage when updating fingerprint

vancluever:
Using `clearRetainingCapacity` as it was being used in `Fingerprint.update` produces undefined behavior; both the old "copy" and current version of the fingerprint would still be using the same storage.

This commit changes the function so that the storage is more clearly re-used, and shrunk at the end if need be.

### DIFF
--- a/src/terminal/search/viewport.zig
+++ b/src/terminal/search/viewport.zig
@@ -197,9 +197,8 @@ pub const ViewportSearch = struct {
         pub fn update(self: *Fingerprint, alloc: Allocator, pages: *PageList) Allocator.Error!bool {
             try self.entries.ensureTotalCapacity(alloc, pages.rows);
 
-            const old = self.entries.items;
+            const old_len = self.entries.items.len;
             var changed = false;
-            self.entries.clearRetainingCapacity();
 
             var i: usize = 0;
             var it = pages.pageIterator(.right_down, .{ .viewport = .{} }, null);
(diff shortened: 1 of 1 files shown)


vancluever:
I caught this doing the 0.16.0 update because the underlying storage in `ArrayList` is now being invalidated (using `@memset(self.items, undefined)`) whereas before only the length was being reset. :slightly_smiling_face:

src/terminal/search/viewport.zig
```diff
@@ -209,15 +208,20 @@ pub const ViewportSearch = struct {
...
-                self.entries.appendAssumeCapacity(entry);
+                if (i < self.entries.items.len) {
+                    self.entries.items[i] = entry;
+                } else {
+                    self.entries.appendAssumeCapacity(entry);
+                }
             }
 
-            return changed or i != old.len;
+            self.entries.shrinkRetainingCapacity(i);
```
jparise, line 223: Should we call this unconditionally, or only when `i != old_len`?

jparise (commented):
Ugh, thanks for this! I was so focused on making sure the backing _size_ was correct that I totally didn't realize the _contents_ had become defined.

jparise (approved):

src/terminal/search/viewport.zig
vancluever: I think this is fine - [`shrinkRetainingCapacity`](https://ziglang.org/documentation/0.16.0/std/#std.array_list.Aligned.shrinkRetainingCapacity) _does_ assert on `new_len <= old_len` so if there's issues, we'd likely catch it, otherwise it's effectively a no-op when the length is the same (which it would be if `new_len > old_len` since we append to the end in that event).

PR #13332, opened 2026-07-14
terminal: report modifyOtherKeys state via XTQMODKEYS

aymanbagabas:
Implements querying support for xterm `modifyOtherKeys` (XTQMODKEYS resource 4) via two forms:

- `CSI ? 4 m` — the XTQMODKEYS query.
- `DCS $ q > 4 m ST` — the DECRQSS form. `CSI ? 4 m` is technically a private `SGR` sequence (historically a superscript rendition), so DECRQSS is offered as a safer alternative per review feedback. Both forms are kept.

Ghostty implements mode 1 by default (ambiguous modified keys use the `CSI 27` numeric form) and mode 2 when enabled, but never responded to either query, so applications could not detect support or the active mode.

Both forms reply with the XTMODKEYS control `> 4 ; Pv m` (CSI-prefixed for `CSI ? 4 m`, wrapped in `DCS 1 $ r … ST` for DECRQSS), where `Pv` is `2` when mode 2 is active and `1` otherwise. `Pv` is never `0` because the mode 1 encoding is always active and cannot be disabled. The reply payload is shared via `Terminal.modifyOtherKeysReport` and unit tested.

Discussion: https://github.com/ghostty-org/ghostty/discussions/13331

## AI disclosure
Implemented with AI assistance (GitHub Copilot CLI, model Claude Opus 4.8). I reviewed, understand, and can explain all of the code and its interaction with the key encoder without AI aid, per the AI Usage Policy.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -3628,6 +3628,17 @@ pub fn printAttributes(self: *Terminal, buf: []u8) ![]const u8 {
     return writer.buffered();
 }
 
+/// Writes the XTMODKEYS report for modifyOtherKeys into buf and returns the
+/// written slice. This is the payload for XTQMODKEYS (both the `CSI ? 4 m`
+/// and `DCS $ q > 4 m ST` query forms) and is formatted as an XTMODKEYS
+/// control: `> 4 ; Pv m`. Ghostty's default encoder already emits the
+/// numeric form for ambiguous keys (mode 1), so the floor is 1; mode 2
+/// reports 2.
+pub fn modifyOtherKeysReport(self: *Terminal, buf: []u8) ![]const u8 {
+    const pv: u8 = if (self.flags.modify_other_keys_2) 2 else 1;
+    return try std.fmt.bufPrint(buf, ">4;{d}m", .{pv});
+}
+
 /// The modes for DECCOLM.
 pub const DeccolmMode = enum(u1) {
     @"80_cols" = 0,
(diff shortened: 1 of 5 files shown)


jcollie:
Initial read-through looks good but I'm on mobile so maybe I missed something.

src/terminal/stream_terminal.zig
```diff
@@ -559,6 +560,16 @@ pub const Handler = struct {
         self.writePty(resp);
     }
 
+    fn queryModifyOtherKeys(self: *Handler) void {
```
mitchellh, line 629: This should short-circuit if write_pty is not set.

src/termio/stream_handler.zig
```diff
@@ -994,6 +995,22 @@ pub const StreamHandler = struct {
         });
     }
 
+    pub fn queryModifyOtherKeys(self: *StreamHandler) !void {
+        log.debug("querying modifyOtherKeys", .{});
```
mitchellh: Pointless log.

src/terminal/stream.zig
```diff
@@ -1780,23 +1782,35 @@ pub fn Stream(comptime H: type) type {
                                 self.handler.vt(.modify_key_format, format);
                             },
 
+                            // XTQMODKEYS: query key modifier options. We only
+                            // support modifyOtherKeys (resource 4); `vim --clean`
+                            // sends `CSI ? 4 m`. Reply is `CSI > 4 ; Pv m`.
+                            '?' => switch (input.params.len) {
```
mitchellh, line 1810: We should have a couple unit test cases for this

mitchellh (changes requested):

aymanbagabas:
Thank you @jcollie @mitchellh for the review. Addressed the review comments and also added DECRQSS report support.

src/terminal/Terminal.zig
```diff
@@ -3493,7 +3493,16 @@ pub fn printAttributes(self: *Terminal, buf: []u8) ![]const u8 {
...
+/// written slice. This is the payload for XTQMODKEYS (both the `CSI ? 4 m`
+/// and `DCS $ q > 4 m ST` query forms) and is formatted as an XTMODKEYS
+/// control: `> 4 ; Pv m`. Ghostty's default encoder already emits the
+/// numeric form for ambiguous keys (mode 1), so the floor is 1; mode 2
+/// reports 2.
+pub fn modifyOtherKeysReport(self: *Terminal, buf: []u8) ![]const u8 {
+    const pv: u8 = if (self.flags.modify_other_keys_2) 2 else 1;
+    return try std.fmt.bufPrint(buf, ">4;{d}m", .{pv});
+}
 pub const DeccolmMode = enum(u1) {
```
jcollie, line 3643: Blank line between functions.

aymanbagabas:
@jcollie @mitchellh do you have any other notes or comments on this?

PR #10263, opened 2026-01-11
feat(gtk): Port undo-timeout feature to Linux/GTK

GitJuhb:
## Summary

- Ports the `undo-timeout` feature from macOS to the Linux/GTK runtime
- Allows users to undo closed tabs/splits within a configurable timeout (default 5 seconds)
- Uses `Ctrl+Shift+Z` keybinding (same as macOS `cmd+z`)

## Technical Details

Key implementation:
- `ExpiringUndoStack` holds closed surfaces for the timeout period
- GLib timers handle automatic expiration
- Surfaces are ref'd to stay alive until restored or expired
- Forces synchronous Core surface cleanup during shutdown to avoid assertion failures

The tricky part was handling GObject's asynchronous finalization vs Core App's synchronous assertions. Solution: `forceDeinitCore()` explicitly cleans up Core resources before GObject finalization runs.

## Test plan

- [x] Create split with `Ctrl+Shift+O`
- [x] Close split with `Ctrl+Shift+W`
- [x] Undo within timeout with `Ctrl+Shift+Z` - split is restored
- [x] Wait for timeout (default 5s) - undo no longer available
- [x] Close tab, undo - tab restored with surfaces
- [x] Clean shutdown with no assertion panics

## Note

This PR was developed using [Claude Code](https://claude.ai/claude-code). The contributor has minimal coding knowledge but verified the feature works as expected. Thank you to the Ghostty team for the well-structured codebase!

PR #13333, opened 2026-07-14
terminal: implement DECSTR soft reset (CSI ! p)

deblasis:
The CSI exclamation point p sequence, known as DECSTR or soft terminal reset, was logged as unimplemented and simply ignored. 
Because of that, any settings a program had applied, such as the scroll region, origin mode, SGR attributes, or character set, stayed in effect. 
Later output then appeared in the wrong position, with the wrong color, or in the wrong character set until a full reset with RIS or the reset command fixed it.DECSTR resets the terminal's programmable state back to power-up defaults according to DEC STD 070. This includes making the cursor visible, switching to replace mode, using absolute origin, turning autowrap on, applying default SGR, selecting ASCII character sets, setting the full scrolling region, removing protection, and clearing the saved cursor position. Unlike RIS, it leaves the screen contents untouched and does not move the cursor.

## Repro

```
printf '\e[4;9r\e[?6h\e[?7l\e[1;31;44m\e(0\e[!p\e[1;1Hplain-abc\n'
```

Drop the `\e[!p` to see today's behavior: the text lands mangled in the
leftover scroll region, DEC line-drawing charset, and SGR.

## Before / after

Same reproducer fed through the terminal model. Left: `CSI ! p` ignored, so
state leaks past the reset. Right: with this change.

| ignored (today) | soft reset (this PR) |
|---|---|
|  |  |

xterm, VTE, and conhost all implement this. Test included.

Note: DEC STD 070 lists autowrap as reset-off; this matches xterm/conhost in
resetting it to the default (on).

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -4420,6 +4420,45 @@ pub fn fullReset(self: *Terminal) void {
     self.flags.dirty.clear = true;
 }
 
+/// Soft terminal reset (DECSTR, `CSI ! p`). Resets programmable state (modes,
+/// SGR, charsets, margins, saved cursor) to defaults without touching the
+/// screen or cursor. This is the difference from `fullReset` (RIS).
+pub fn softReset(self: *Terminal) void {
+    const screen: *Screen = self.screens.active;
+
+    // Modes reset by DECSTR (DEC STD 070).
+    self.modes.set(.cursor_visible, true); // DECTCEM
+    self.modes.set(.insert, false); // IRM
+    self.modes.set(.origin, false); // DECOM
+    self.modes.set(.disable_keyboard, false); // KAM
+    self.modes.set(.cursor_keys, false); // DECCKM
+    self.modes.set(.keypad_keys, false); // DECNKM
+    // DECAWM: the spec resets this off, but we match xterm/conhost and leave
+    // the default (on).
+    self.modes.set(.wraparound, true);
+
+    // SGR back to default.
+    screen.cursor.style = .{};
+    screen.manualStyleUpdate() catch unreachable;
+
+    // Charsets (G0-G3) back to ASCII.
+    screen.charset = .{};
+
+    self.setProtectedMode(.off); // DECSCA
+
+    // Full scrolling region.
+    self.scrolling_region = .{
+        .top = 0,
+        .bottom = self.rows - 1,
+        .left = 0,
+        .right = self.cols - 1,
+    };
+
+    screen.cursor.pending_wrap = false;
+    screen.saved_cursor = null; // DECSC
+    self.previous_char = null; // REP
+}
+
 /// Returns true if the point is dirty, used for testing.
 fn isDirty(t: *const Terminal, pt: point.Point) bool {
     return t.screens.active.pages.getCell(pt).?.isDirty();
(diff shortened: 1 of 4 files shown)


src/terminal/Terminal.zig
```diff
@@ -3913,6 +3913,44 @@ pub fn fullReset(self: *Terminal) void {
     self.flags.dirty.clear = true;
 }
 
+/// Soft terminal reset (DECSTR, `CSI ! p`). Unlike `fullReset` (RIS), resets
+/// programmable state to power-up defaults without clearing the screen or
```
mitchellh: I think this comment needs some work, e.g. "power-up defaults" is confusing in the context of a non-hardware terminal. I'd be clearer about what the spec says (probably this) and what the practical implementation is for terminal emulators.

mitchellh (changes requested):
Seems reasonable to support, obviously not widely used but also low risk to do. I'd be curious to know what other terminals do with some verification (because LLMs will make that shit up).

deblasis:
Some extra context that I didn't want to land in the body of the commit in case this gets merged:
I bumped into this while experimenting with some conPTY "optimizations" on Windows, for which I have to build a solid framework to test and compare everything down to the cell-by-cell level among the other things.

Re: verification — every claim below is a permalink pinned to a commit SHA, quoting the line it rests on, so none of it has to be taken on faith.

| Source | Autowrap after DECSTR | Clears screen / moves cursor? | Evidence |
|---|---|---|---|
| **VT510 manual** *(the spec)* | **OFF** | — | ["Autowrap (DECAWM): No autowrap."](https://vt100.net/docs/vt510-rm/DECSTR.html) |
| xterm | ON | No / No | [restores `autoWrap` resource, default `True`](https://github.com/ThomasDickey/xterm-snapshots/blob/6380a3eaed857c182ea6cfa78cd706966b2628d0/charproc.c#L14510) |
| VTE | ON | No / No | [`reset(false, false)`](https://github.com/GNOME/vte/blob/e78e268f342631ce9e13e69be41578da8348dbdc/src/vteseq.cc#L6163) |
| conhost / Windows Terminal | ON | No / No | [`SetSystemMode(AutoWrap, true)`](https://github.com/microsoft/terminal/blob/922beefb83764646331662d6d15d70107d556402/src/terminal/adapter/adaptDispatch.cpp#L2991) |
| kitty | ON | No / No | [`self->modes = empty_modes`](https://github.com/kovidgoyal/kitty/blob/f47590533d7177daf0b74963f9d1b7581467af20/kitty/screen.c#L207-L208) (`.mDECAWM=true` at [L40](https://github.com/kovidgoyal/kitty/blob/f47590533d7177daf0b74963f9d1b7581467af20/kitty/screen.c#L40)) — added 2026-07-15 |
| **Ghostty (this PR)** | **ON** | **No / No** | matches all four |

**Takeaway:** the spec resets autowrap OFF, but every emulator that implements DECSTR resets it ON. This PR sides with the implementers. None of them clear the screen or move the cursor, matching us.

### xterm documents this deviation deliberately

The clearest statement of spec-vs-practice comes from [xterm itself](https://github.com/ThomasDickey/xterm-snapshots/blob/6380a3eaed857c182ea6cfa78cd706966b2628d0/charproc.c#L14308-L14317):

> Implement soft or hard (full) reset of the VTxxx emulation. There are a couple of differences from real DEC VTxxx terminals (**to avoid breaking applications which have come to rely on xterm doing this**):
>
> + autowrap mode should be reset (instead it's reset to the resource default).

So it's a deliberate compatibility choice, not an oversight — which is why I'd rather match it than the letter of the spec.

### Spec (VT510)

The [DECSTR table](https://vt100.net/docs/vt510-rm/DECSTR.html) lists DECAWM as "No autowrap", alongside DECTCEM "Cursor enabled", IRM "Replace mode", DECOM "Absolute", DECSCA "Normal (erasable by DECSEL and DECSED)", DECSTBM "Top margin = 1; bottom margin = page length", DECSC "Home position". (Citing the VT510 manual specifically; xterm cites DEC STD 070 p. 14-9 for the same sequence.)

### xterm — ON

[`CASE_DECSTR`](https://github.com/ThomasDickey/xterm-snapshots/blob/6380a3eaed857c182ea6cfa78cd706966b2628d0/charproc.c#L6147-L6149) → `VTReset(xw, False, False)` → `ReallyReset(full=False)`. The DECSTR branch does `bitcpy(&xw->flags, xw->initflags, WRAPAROUND | ...)`, and `initflags` is snapshotted from the [`autoWrap` resource, which defaults `True`](https://github.com/ThomasDickey/xterm-snapshots/blob/6380a3eaed857c182ea6cfa78cd706966b2628d0/charproc.c#L423). The RIS branch calls `CursorSet(screen, 0, 0, ...)`; the DECSTR branch doesn't, and nothing in the shared path clears or moves the cursor.

### VTE — ON (with a caveat worth flagging)

`Terminal::DECSTR` → `reset(false, false)` = `clear_tabstops=false, clear_history=false`. In [`Terminal::reset`](https://github.com/GNOME/vte/blob/e78e268f342631ce9e13e69be41578da8348dbdc/src/vte.cc#L10832) the block that resets `cursor.row`/`cursor.col` and the ring is entirely gated on `clear_history`, so a soft reset leaves screen and cursor untouched. Modes go through `m_modes_private.reset()` → `set_modes(m_default_modes)`, and DECAWM is in the [default-set list](https://github.com/GNOME/vte/blob/e78e268f342631ce9e13e69be41578da8348dbdc/src/modes.hh#L237):

```cpp
constexpr Private() : Self{eDEC_AUTOWRAP, eDEC_TEXT_CURSOR, ...}
```

Caveat, because it looks like a contradiction if you go looking: [`modes.py`](https://github.com/GNOME/vte/blob/e78e268f342631ce9e13e69be41578da8348dbdc/src/modes.py#L434) says `mode_WHAT('DEC_AUTOWRAP', 7, default=False, ...)` under a `# Default: reset` comment. That `default=` is dead metadata for writable modes — [`write_modes()`](https://github.com/GNOME/vte/blob/e78e268f342631ce9e13e69be41578da8348dbdc/src/modes.py#L1383-L1392) only consults it for non-writable `MODE_FIXED` entries; writable modes emit a bare `MODE(name, number)`. The effective default is the `modes.hh` constructor above.

### conhost / Windows Terminal — ON

[`AdaptDispatch::SoftReset()`](https://github.com/microsoft/terminal/blob/922beefb83764646331662d6d15d70107d556402/src/terminal/adapter/adaptDispatch.cpp#L2984-L3020):

```cpp
_api.SetSystemMode(ITerminalApi::Mode::AutoWrap, true); // Wrap at end of line.
```

No clear, no cursor move. It also resets only the *active* saved-cursor slot, matching xterm's alt-buffer behaviour (GH#19918) — same as this PR.

### kitty — ON, as of today

Added in [`0816716c`](https://github.com/kovidgoyal/kitty/commit/0816716c72e88c5d5ccfc89c8d07cde7bb3c4ac0) (2026-07-15T03:46Z), fixing [#10263](https://github.com/kovidgoyal/kitty/issues/10263), filed ~6h earlier. [`screen_soft_reset`](https://github.com/kovidgoyal/kitty/blob/f47590533d7177daf0b74963f9d1b7581467af20/kitty/screen.c#L229-L234) saves and restores the cursor around the reset, and the `linebuf_clear`/`historybuf_clear` are gated on `is_hard_reset` — so no clear, no move.

One honest caveat: kitty's soft reset is *broader* than the others (it also resets tabstops, colours, selections, and the parser), so "kitty matches" holds for autowrap/clear/cursor but they aren't identical implementations. Kovid's own commit message notes the sequence "is largely undefined ... Do what I feel is sensible in these cases."

Issue #13359, opened 2026-07-16
macOS 26.5.2: `undo-timeout = 0` makes closing a tab crash Ghostty (1.3.1 + tip)


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/13358

<div type='discussions-op-text'>

<sup>Originally posted by **cjbarber** July 16, 2026</sup>
### Issue Description

With `undo-timeout = 0` set, closing a tab takes down the whole app. Doesn't matter how I close it
(X button, cmd+w, close_tab, cmd+shift+w, undo of a close), every window dies at once with no
confirmation. An empty tab with nothing running still does it.

Blank config doesn't crash. Add `undo-timeout = 0` and it crashes. Delete it (back to the 5s default) and it's fine.

It started the day I updated macOS 15.7.7 to 26.5.2. That same 1.3.1 build had been fine for a while.
I can't remember if I already had `undo-timeout = 0` before the update, so I don't know if it needs
macOS 26 or crashes with that setting on any OS. Same crash on 1.3.1 stable and tip
(1.3.2-main-+73534c468).

Crash is EXC_BAD_ACCESS / KERN_INVALID_ADDRESS. The fault address is different garbage every time
but the crash site never moves. 

Stack from the minidump (tip):

```
Ghostty.App.action(_:target:action:) + 740
Surface.handleMessage + 2012
ghostty_app_tick + 488
closure #1 in static Ghostty.App.wakeup(_:) + 104
main + 140
```

Same stack on 1.3.1, different offsets (App.action + 688, Surface.handleMessage + 2336).
libswiftObservation and Combine are on the crashing thread too.

Sentry event IDs (local only, +crash-report can't upload them): ffa311d2-a522-45c5-a7b6-13a53121c72a
(tip), 262a194a-ab67-4e0b-4536-fdd788285076 (1.3.1). 

### Expected Behavior

Closing a tab closes that tab and nothing else.

### Actual Behavior

Closing a tab kills the whole app, every window and tab, including ones with processes running. No
confirmation dialog. A crash dump lands in ~/.local/state/ghostty/crash/.

### Reproduction Steps

1. Config with one line: undo-timeout = 0
2. Launch, cmd+t, close the new tab (X or cmd+w). Whole app crashes.
3. Delete the line and it never crashes.

### Ghostty Logs

```text
No log output at the crash. Stack and Sentry IDs are above.
```

### Ghostty Version

```text
[...]

PR #13364, opened 2026-07-17
macOS: free surface synchronously in deinit on main thread

bo2themax:
Since the renderer thread now emits scrollbar events on almost every frame, there's always a `.scrollbar` message for the dying surface in the app mailbox. 

The OS runtime seems to schedule `appTick` and `ghostty_surface_free` differently across macOS pre-26, 26 and 27. 

On macOS 26.x, `ghostty_app_free` happens after `App.scrollbar(_:target:v:)`, leaving `surface.userdata` pointing at a freed `SurfaceView`.

When `deinit` runs on the main thread, free the surface synchronously instead of detaching to a task. This fixes both crashes mentioned in https://github.com/ghostty-org/ghostty/pull/9512 and https://github.com/ghostty-org/ghostty/issues/13359.

### AI Disclosure 

I used Claude to analyze the backtrace, but the code is written and tested by myself.

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.Surface.swift
+++ b/macos/Sources/Ghostty/Ghostty.Surface.swift
@@ -24,6 +24,13 @@ extension Ghostty {
         }
 
         deinit {
+            guard !Thread.isMainThread else {
+                // The surface remains registered with the app and holds unretained
+                // userdata until it is freed. When already on the main thread, free
+                // it synchronously so teardown completes before we disappear.
+                ghostty_surface_free(surface)
+                return
+            }
             // deinit is not guaranteed to happen on the main actor and our API
             // calls into libghostty must happen there so we capture the surface
             // value so we don't capture `self` and then we detach it in a task.


jparise:
Seems related to #9760, too.

I'm curious if the approach here covers all of the complications I discovered while working on that one.

mitchellh (changes requested):
I'm not confident in this change. If I'm understanding properly, the problem is sometimes `ghostty_app_free` happens BEFORE all the surface frees? That's fundamentally it? If so, this doesn't seem right. If not, I don't understand what this is doing. It doesn't seem like a surface should have any business freeing the entire app (which owns N surfaces).

I need more info.

bo2themax:
> I'm not confident in this change. If I'm understanding properly, the problem is sometimes `ghostty_app_free` happens BEFORE all the surface frees? That's fundamentally it? If so, this doesn't seem right. If not, I don't understand what this is doing. It doesn't seem like a surface should have any business freeing the entire app (which owns N surfaces).
> 
> I need more info.

Sh*t, you're right, I used `ghostty_surface_free` in VM, but typed it wrong🫪, and still pasted in every comment...

And for the context for `undo-timeout=0` here are the sequence difference between 26.5.1 and 27 Beta 3 (using https://github.com/ghostty-org/ghostty/commit/0784f6b617f28453172eb0ca19548a63b69aa38b)

I assume the second `scrollbar_dirty => .scrollbar` arrives early when draining the mailbox.

image
image

For the previous undo crash mentioned in https://github.com/ghostty-org/ghostty/pull/9512, I reproduced it again the backtrace is exactly the same as `undo-timeout=0`, so I reverted it.

macos/Sources/Ghostty/Ghostty.Surface.swift
```diff
@@ -24,6 +24,21 @@ extension Ghostty {
         }
 
         deinit {
+            guard !Thread.isMainThread else {
+                // Free synchronously when it’s on the main thread.
+                // Since the renderer thread now emits scrollbar events
+                // on almost every frame, there's always a `.scrollbar` message
+                // for the dying surface in the app mailbox.
+                //
```
jparise: Even if this ends up being the right approach, I don't think we should frame the rationale like this. This is a point-in-time description of the current state rather than a lifecycle contract.

That's a (the?) reason why #9760 is as involved as it is: it avoids making assumptions about the mailbox, etc. and instead makes the contract "signal and wait for the apprt to finish" before we finalize the termination of the (macOS) app itself.

I acknowledge that the issue you're going after here (#13359, #13359) is in the tab+surface close path rather than the whole app termination path, but there are similarities.

bo2themax: Make sense to me! Do you have a suggestion how should we phrase this? I don't think I have better understanding of the whole lifecycle including the renderer and the mailbox🤔

macos/Sources/Features/Terminal/TerminalController.swift
```diff
@@ -503,14 +503,9 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
...
-                // for some people on macOS Tahoe this caused a crash and the queue
-                // fixes it.
-                // https://github.com/ghostty-org/ghostty/pull/9512
-                DispatchQueue.main.async {
-                    undoManager.disableUndoRegistration {
-                        target.closeTab(nil)
-                    }
+                // Close the tab when undoing
+                undoManager.disableUndoRegistration {
+                    target.closeTab(nil)
```
jparise: Let's move this into its own change. That will make it easier to attribute any new issues or regressions that the `deinit` change might introduce.

macos/Sources/Ghostty/Ghostty.Surface.swift
```diff
@@ -24,6 +24,21 @@ extension Ghostty {
...
+                // OS runtime seems to schedule `appTick` and `ghostty_surface_free` differently for pre-26, 26 and 27.
+                // On macOS 26.x, `ghostty_surface_free` happens after `App.scrollbar(_:target:v:)`,
+                // which results in `surface.userdata` pointing at a freed `SurfaceView`.
+                //
+                // This fixes both crashes mentioned in https://github.com/ghostty-org/ghostty/pull/9512
+                // and https://github.com/ghostty-org/ghostty/issues/13359
```
jparise:
```suggestion
                // The surface remains registered with the app and holds unretained
                // userdata until it is freed. When already on the main thread, free
                // it synchronously so teardown completes before we disappear.
```

jparise (approved):
I thought about it a bit more, and I think this approach will work and address the root issue.

I don't really like how it solidifies some assumptions around the macOS app's threads and lifecycle alongside the embedded apprt's though.

PR #13369, opened 2026-07-17
libghostty: mark semantic stream failures

mitchellh:
The libghostty-vt stream is made to be infallible: in the case of any error it just logs and moves on. That's because a terminal can't really... stop, under normal operations. But, under special operations (fuzzing, replays, etc.) it can and should stop!

Rather than make the operation fallible, its simply enough for me at least to know that something went wrong. This is a simple change that adds a simple flag that is flagged to true when such a scenario happens.

For normal Ghostty GUI operations, this isn't used at all. For libghostty consumers they can choose to read it if they want, but don't have to.

This also adds a C API to read it.


mitchellh:
AI was only used as a judge here and to help construct the C API test. Not used anywhere else.

PR #13370, opened 2026-07-17
terminal: move cursor defaults into terminal state

mitchellh:
Cursor defaults were duplicated across stream handlers and it was a pretty significant amount of simple and yet non-trivial logic to understand.

Store these on Terminal itself and have methods to route things like DECSCUSR through for consistent behaviors.

No AI usage here.


mitchellh:
Should've made this a draft. I'm still looking to see if I missed any corners here (e.g. cloning, C API, etc.)

PR #13384, opened 2026-07-19
config: update `background-blur` to reflect growing adoption

pluiedev:
`ext-background-effect-v1` is finally seeing major adoption throughout KWin, Mutter, cosmic-comp, Niri, etc. and we should update our comments on that.


jcollie (approved):
LGTM

00-kat (approved):

PR #13394, opened 2026-07-20
terminal: handle page capacity errors in cursorScrollAbove

mitchellh:
Re: #13160

When cursorScrollAbove rotates rows across a page boundary, the last row of the previous page is cloned into the destination page. That can cause capacity failures we didn't previously handle.

The error propagated out of the operation after rows had already been rotated,  leaving the page list half-mutated. Subsequent operations on the corrupted state can cause crashes since the state was incoherent.

Handle these errors the same way insertLines and deleteLines already do for their cross-page copies: increase the destination page capacity for the dimension that ran out and retry the row copy.

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -1035,36 +1035,107 @@ fn cursorScrollAboveRotate(
     self: *Screen,
     fresh_node: ?*PageList.List.Node,
 ) !void {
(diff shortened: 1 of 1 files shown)


mitchellh:
FYI I think `eraseRow` family has the same bug, but will post that in isolation if true. I'm looking around for other places we don't handle capacities properly.

PR #13397, opened 2026-07-20
terminal: handle page capacity errors in eraseRow

mitchellh:
Re: #13160 (related but not that issue)

PageList eraseRow and eraseRowBounded have the same issue previously fixed for cursorScrollAbove: when shifting rows up across a page boundary, the top row of the next page is cloned into the last row of the previous page, and that clone can fail if the destination page lacks capacity for the row's managed memory.

Handle the errors the same way the other cross-page copies do: increase the destination page capacity for the dimension that ran out and retry the row copy.

This type of logic was repeated EVERYWHERE so I extracted this into a helper in PageList and Screen. They're slightly different due to the extra accounting that Screen has to do for the cursor.

Don't know of any scenario this actually happened in the real world but it was trivially reproducible with tests.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -3636,6 +3636,36 @@ pub const IncreaseCapacity = enum {
     grapheme_bytes,
     hyperlink_bytes,
     string_bytes,
+
+    /// Returns the capacity dimension that must be increased for the
+    /// given row clone error to succeed on retry, or null if the page
+    /// only needs to be rehashed at its current capacity.
+    pub fn forCloneError(err: Page.CloneFromError) ?IncreaseCapacity {
+        return switch (err) {
+            // Rehash the sets
+            error.StyleSetNeedsRehash,
+            error.HyperlinkSetNeedsRehash,
+            => null,
+
+            // Increase style memory
+            error.StyleSetOutOfMemory,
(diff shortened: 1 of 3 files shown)


aaronlumen:
> [!TIP] 
> `\033;[2F`
> 

> [!WARNING]
> In terminal and console programming, the sequence **`\033[;2f`** is an ANSI escape code (or CSI code) used for cursor 
> positioning. It commands the terminal to move the text cursor to a specific coordinate on the screen.Breakdown of the 
> Code` **\033**`:   The C-style octal representation of the ASCII Escape character 
> (equivalent to hex \x[1]B).
> **`[:`** The Control Sequence Introducer (CSI), which tells the terminal an escape command is beginning.
> **`;[2]f`:** The command itself. The **f** stands for "Horizontal and Vertical Position" (HVP). 
> The numbers before and after the semicolon define the row and column, respectively.
> How It Works
> Because there is no number before the semicolon, the row defaults to 1 (the very top of the screen).
> The number 2 means the column.Together, **`\0[3]3[;2f`** moves the cursor to Row 1, Column 2 (the top line, second 
> character from the left).

> [!TIP]
> Note: This sequence functions identically to `**\033[;2H**` (Cursor Position).

PR #10253, opened 2026-01-10
fix(macos): fix fullscreen tab bar disappearing after monitor disconnect

neurosnap:
# Issue Description

On macOS Sequoia 15.x, when using Ghostty in native fullscreen mode with multiple tabs, disconnecting an external monitor causes the tab bar to disappear and show desktop background through a gap at the top of the window.

Affected `macos-titlebar-style` options: "native", "transparent", and "tabs"

# Reproduction Steps

- Open Ghostty with multiple tabs in fullscreen mode
- Connect an external monitor to computer
- Close the laptop lid (using external monitor only)
- Cycle through all tabs
- Open the laptop lid
- Unplug the external monitor
- Switch to a different tab

**Expected:** Tab bar remains visible with the new tab selected
**Actual:** Tab bar disappears, gap at top shows desktop background instead of terminal content

https://github.com/user-attachments/assets/a0d507cb-48be-4122-aee8-3a8c31c28e7b

# Root Cause

In native fullscreen with tabs, each tab is a separate NSWindow managed by an NSWindowTabGroup. When the display configuration changes (monitor disconnect), macOS sends didChangeScreenParametersNotification. However, only the currently visible tab's window updates its frame geometry to match the new screen. Background tabs retain stale frame data from the old monitor configuration.

When switching to a background tab after the screen change, AppKit restores the tab's stale geometry, which doesn't match the new screen dimensions. This causes the toolbar window (NSToolbarFullScreenWindow) containing the tab bar to be positioned incorrectly - pushed off the top of the screen.

# Solution

The fix programmatically cycles through all tabs when screen parameters change, briefly making each tab key to trigger AppKit's internal layout recalculation. This updates all tabs' geometry to match the new screen configuration before the user switches tabs.

I used ampcode (AI) to help me:
- Debug the issue
- Figure out root cause
- Come up with a strategy to fix it
- Apply the code changes

My primary job was to:
- QA all code changes against all `macos-titlebar-style` options on mac os
- Review code changes
- Submit report and PR

Reference: https://github.com/ghostty-org/ghostty/discussions/5389

PR #13407, opened 2026-07-21
macOS: sync appearance after screen parameters change

carldaws:
In #5389 there are multiple reports of tab bars appearing incorrectly after disconnecting an external monitor when Ghostty is in fullscreen.

BaseTerminalController handles the notfiication from AppKit about a screen change and, for non-fullscreen windows, it calls setFrame, which triggers windowDidMove, which calls fixTabBar. But for fullscreen windows, none of this happens, we skip clamping because macOS handles the sizing and the tab bar doesn't get fixed.

I've restructured the handler so that the clamping is conditional on the window being non-fullscreen and window.screen being defined, and regardless of the clamping, called syncAppearance at the end of the handler.

This has fixed the issue for me, reproduced locally using BetterDisplay to create a virtual second display and with my external monitor.

AI disclosure: I used GLM 5.2 and Claude Opus 4.8 in OpenCode to investigate the issue and the call chain, but I fully understand what is going on and wrote this PR and my commit message and the comment and the code myself.

### DIFF
--- a/macos/Sources/Features/Terminal/BaseTerminalController.swift
+++ b/macos/Sources/Features/Terminal/BaseTerminalController.swift
@@ -534,6 +534,13 @@ class BaseTerminalController: NSWindowController,
         guard let window else { return }
         guard window.isVisible else { return }
 
+        defer {
+            // Always resync appearance. A display change can leave window
+            // appearance in a stale state (e.g. a black/hidden tab bar):
+            // https://github.com/ghostty-org/ghostty/discussions/5389
+            syncAppearance()
+        }
+
         // We ignore fullscreen windows because macOS automatically resizes
         // those back to the fullscreen bounds.
         guard !window.styleMask.contains(.fullScreen) else { return }
@@ -1024,6 +1031,7 @@ class BaseTerminalController: NSWindowController,
     /// as we add new things:
     ///
     ///  - ``toggleBackgroundOpacity``
+    ///  - ``didChangeScreenParametersNotification``
     func syncAppearance() {
         // Purposely a no-op. This lets subclasses override this and we can call
         // it virtually from here.


bo2themax:
~~I don't see a clear way to reproduce this in the linked discussion, how did you reproduce and verify it? 🫪~~

Oki, I can reproduce it with default config now

carldaws:
@bo2themax yes it may be that having a config entry for background opacity might be required to reproduce, but if you've reproduce the bug with a default config then perhaps not.

In any case, for anyone else curious, with my config:

```
background-image = background.png
background-image-opacity = 0.20
background-image-fit = cover
theme = Nord
```

Open Ghostty on an external monitor (or a virtual one with BetterDisplay) and set it to fullscreen, then open a few tabs. If you disconnect the monitor, Ghostty will remain fullscreen in a workspace on your primary screen, if you navigate to that workspace everything looks okay, but as soon as you select another tab, the tab bar goes black / transparent. I usually bring it back by using command+N to switch to another tab.

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -534,42 +534,46 @@ class BaseTerminalController: NSWindowController,
         guard let window else { return }
         guard window.isVisible else { return }
```
bo2themax, line 536:
```suggestion

        defer {
            // Always resync appearance. A display change can leave window
            // appearance in a stale state (e.g. a black/hidden tab bar):
            // https://github.com/ghostty-org/ghostty/discussions/5389
            syncAppearance()
        }
```

you can just add defer here.

bo2themax (changes requested):
Looks good to me, only a code style change suggestion

macos/Sources/Features/Terminal/BaseTerminalController.swift
carldaws: Yes that makes more sense, done!

macos/Sources/Features/Terminal/BaseTerminalController.swift
```diff
@@ -534,42 +534,48 @@ class BaseTerminalController: NSWindowController,
...
-        // those back to the fullscreen bounds.
-        guard !window.styleMask.contains(.fullScreen) else { return }
+        defer {
+            // Always resync appearance. A display change can leave window
+            // appearance in a stale state (e.g. a black/hidden tab bar):
+            // https://github.com/ghostty-org/ghostty/discussions/5389
+            syncAppearance()
+        }
 
-        guard let screen = window.screen else { return }
```
bo2themax, line 541: The following changes should be reverted. Only `defer` would do the trick

bo2themax (changes requested):

macos/Sources/Features/Terminal/BaseTerminalController.swift
carldaws: Ah I see, thanks for your patience, I think I've got it right now

bo2themax:
Looks clean now, but I can still reproduce running this branch.

1. Create a new window
2. Enter fullscreen
3. Create new tabs
4. Disconnect the monitor.

It's important to create tabs in fullscreen to reproduce this

carldaws:
@bo2themax you're right, apologies, I think I must have opened the tabs before I moved Ghostty to the external monitor when trying to confirm my change. Tabs opened that way seem to work, which may be a clue. I'm still looking into it.

carldaws:
@bo2themax after extensive experimentation I think this might be an AppKit bug or similar. I tried to call syncAppearance and fixTabBar on monitor disconnect, I added logging to ensure didChangeScreenParametersNotification was firing, forced ghostty_surface_draw, checked the dimensions of the Metal surface to ensure they matched, hid the NSScrollPocket and more and still the issue persists.

Even though #10253 was reported on a different version of macOS and a different bug, I think it is the correct mechanism for handling this issue for now. I pulled that branch and have confirmed that the issue is resolved. I will close this PR.

PR #13411, opened 2026-07-22
fix: convert Windows selection timestamps to QPC ticks

elias8:
Fixes triple click selectoin on windows. Selection press timestamps are provided in nanoseconds, but `std.time.Instant` uses QPC ticks on windows. Converting between the units preserves the elapsed time to calculate presses correctly. Exposed while adding windows runner for the libghostty [binding](https://github.com/elias8/libghostty/actions/runs/29901534786/job/88863392044?pr=112#step:10:510).

_Note: used GPT 5.5 to identify the issue._

### DIFF
--- a/src/terminal/c/selection_gesture.zig
+++ b/src/terminal/c/selection_gesture.zig
@@ -738,6 +738,13 @@ fn clearWordBoundaryCodepoints(event: *EventWrapper, target: *[]const u21) void 
     target.* = &selection_codepoints.default_word_boundaries;
 }
 
+fn nanosecondsToQpcTicks(ns: u64) u64 {
+    var result: std.os.windows.LARGE_INTEGER = undefined;
+    _ = std.os.windows.ntdll.RtlQueryPerformanceFrequency(&result);
+    const frequency = @as(u64, @bitCast(result));
+    return @intCast(@as(u128, ns) * frequency / std.time.ns_per_s);
+}
+
 fn instantFromNs(ns: u64) SelectionGesture.Time {
     if (comptime builtin.target.cpu.arch == .wasm32 and
         builtin.target.os.tag == .freestanding)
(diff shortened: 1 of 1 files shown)


src/terminal/c/selection_gesture.zig
```diff
@@ -738,6 +738,11 @@ fn clearWordBoundaryCodepoints(event: *EventWrapper, target: *[]const u21) void
     target.* = &selection_codepoints.default_word_boundaries;
 }
 
+fn nanosecondsToQpcTicks(ns: u64) u64 {
+    const frequency = std.os.windows.QueryPerformanceFrequency();
```
jcollie: We are on the verge of merging the Zig 0.16 update, which has removed this particular API. Looks like it was replaced with:

https://ziglang.org/documentation/0.16.0/std/#std.os.windows.ntdll.RtlQueryPerformanceFrequency

Once the 0.16 PR is merged this will need to be updated.

jcollie (changes requested):
This looks good other than the 0.16 API updates that will be needed.

src/terminal/c/selection_gesture.zig
elias8: Confirmed. `QueryPerformanceFrequency` itself is a wrapper around `std.os.windows.ntdll.RtlQueryPerformanceFrequency`, which is available in 15.2. I'll update it so we don't have to wait for zig 0.16 upgrade.

jcollie:
Can you rebase on the latest main now that the Zig 0.16 PR has landed?

elias8:
> Can you rebase on the latest main now that the Zig 0.16 PR has landed?

main now uses `TimeStamp`, which doesn't seem to rely on QPC, so this PR may be outdated/unneccessary. I'm syncing the binding and will verify and update accordingly once done syncing.

elias8:
Closing as resolved. The [CI is green](https://github.com/elias8/libghostty/actions/runs/30025767316) after updating to the latest main.

PR #13414, opened 2026-07-22
gtk: improve split sizing

dkinzler:
This PR improves the way splits/surfaces are sized in the GTK app, which eliminates flickering and slightly improves performance.

Fixes #13328, #12709, #11187.
Related #8208 (closed) but some later comments mention flickering issues persisting.
Builds on top of the changes in #12698.

Previously an idle callback was used to sync the split ratio between the GTK widget tree and the split tree data structure that represents the split layout. The widget tree contains a `SplitTreeSplit` widget, which wraps a `GtkPaned` widget, for every split. During size allocation a `GtkPaned` widget first computes the initial position of the divider and thereby the size for its two children. We get notified of that position (and the max possible position) via the `propPosition/propMaxPosition` callbacks in `SplitTreeSplit` and set up an idle callback (the `onIdle` function) to update the position if it does not match the desired split ratio. Since the initial position is often not correct, especially in nested layouts or if the ratio is not 0.5, a surface will first be shown with the wrong size for a few frames until the idle callback runs and corrects the sizing. In nested layouts it might take multiple rounds of size allocation and idle callbacks until every surface gets the correct size. This causes flickering as widgets eventually snap to another size, which is especially noticeable if the layout changes quickly e.g. when resizing a split using keybinds.

To fix this, the divider position will now be corrected directly from the `propMaxPosition` callback, which runs during GTK size allocation, right after a `GtkPaned` computes the initial position and right before it uses the position to allocate sizes for its two children. With this change every surface will be sized correctly during the first round of size allocation. 
 The idle callback is still used to update the ratio in the split tree when a split is resized by manually dragging the divider in the UI. The logic to sync the split ratio was moved to the new `syncSplitRatio` function which is called from both `propMaxPosition` and `onIdle`.

This is kind of hacky, but I reviewed the GTK source code in detail to verify that this is safe (see the various code comments for more details). I also tested extensively on both Hyprland and KDE Plasma: creating deeply nested layouts, resizing with both keybinds and dragging dividers by hand, with multiple tabs, resizing the entire window, resizing entire subtrees to 0 and back. Everything seems to work fine.

For performance testing I used sysprof which can also collect GTK stats. When creating/deleting/resizing splits I can measure a slight but consistent increase in GTK FPS (+5 to 10) on my system. Other than that CPU usage and FPS seem to be the same before and after. I guess this makes sense, while we added a bit of work to the GTK loop during size allocation, we avoid surfaces being resized.

For the flickering, here's a side-by-side comparison. Left is before the changes, right is after.

https://github.com/user-attachments/assets/2a4f0b4b-e113-49b5-b0d7-d9e507a5a4ff

AI Disclosure: no AI was used.

### DIFF
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -1115,12 +1115,17 @@ const SplitTreeSplit = extern struct {
         /// Assumed to be correct.
         handle: Surface.Tree.Node.Handle,
 
+        /// Cache the split ratio. Looking it up is relatively expensive
+        /// because we have to first walk the widget tree to find the
+        /// SplitTree widget.
+        ratio: f64,
+
         /// Source to handle repositioning the split when properties change.
         idle: ?c_uint = null,
 
         /// Whether the max-position/position property of the gtk.Paned widget
         /// changed. We use these to distinguish between a resize and the user
-        /// manually moving the split divider. See the "on-idle" function.
+        /// manually moving the split divider. See the "onIdle" function.
         max_changed: bool = false,
         pos_changed: bool = false,
 
@@ -1144,6 +1149,7 @@ const SplitTreeSplit = extern struct {
         const self = gobject.ext.newInstance(Self, .{});
         const priv = self.private();
         priv.handle = handle;
+        priv.ratio = split.ratio;
 
         // Setup our paned fields
         const paned = priv.paned;
@@ -1154,6 +1160,17 @@ const SplitTreeSplit = extern struct {
             .vertical => .vertical,
         });
 
+        // Request min width/height 1 for surfaces to prevent them from
+        // becoming temporarily invisible in nested layouts. Otherwise
+        // gtk.Paned might mark a surface as not visible and unmap it
+        // for a single frame. See the comments below in propMaxPosition.
+        if (gobject.ext.isA(start_child, SurfaceScrolledWindow)) {
+            start_child.setSizeRequest(1, 1);
+        }
+        if (gobject.ext.isA(end_child, SurfaceScrolledWindow)) {
+            end_child.setSizeRequest(1, 1);
+        }
+
         // Signals and so on are setup in the template.
 
         return self;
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/class/split_tree.zig
```diff
@@ -1269,10 +1249,60 @@ const SplitTreeSplit = extern struct {
             };
             paned.setPosition(desired_pos);
         } else {
-            // If only position changed, this is a manual human update and
-            // we need to write our update back to the tree.
+            // Update ratio of the split tree node. This is the most dangerous
```
tristan957: Should we add some debug asserts here?

src/apprt/gtk/class/split_tree.zig
```diff
@@ -1163,49 +1180,13 @@ const SplitTreeSplit = extern struct {
...
-    // function will change both max-position and position and synchronously
-    // call our propMaxPosition and propPosition functions. I.e. when the
-    // widget is resized, it should not be possible for onIdle to run before
-    // we have been notified of both property changes.
-    fn onIdle(ud: ?*anyopaque) callconv(.c) c_int {
-        const self: *Self = @ptrCast(@alignCast(ud orelse return 0));
+    // Sync split ratio between the gtk.Paned widget and the split tree.
+    // If to_paned=true, update the gtk.Paned widget to match the ratio
+    // in the split tree and vice-versa for to_paned=false.
+    fn syncSplitRatio(self: *Self, to_paned: bool) void {
```
tristan957: I don't love `to_paned`. Maybe a well named enum would be easier to understand when reading the code?

dkinzler: Changed to use an enum, that's definitely better. Alternatively I also thought about just having a function that returns all the parameters (min, max, position, current_ratio etc.) and then doing the rest directly in propMaxPosition/onIdle, but that is also kind of awkward.

src/apprt/gtk/class/split_tree.zig
dkinzler: Can't hurt, added assertions for all the assumptions in the comment.

tristan957:
Can you fix the commit message formatting for the last two commits?

dkinzler:
Done, sorry about that.

tristan957:
Can you merge the commits where it makes sense? For instance if they are fixing up things in a previous commit.

dkinzler:
Of course, that makes sense. Done.

tristan957:
I added a milestone. I am not that familiar with the split tree code at the moment. It'll take me some time get at the bones of this change, but everything seems to make sense, and the video is definitely an improvement.

jcollie (changes requested):
This LGTM but it should be rebased on the latest `main` so that it's past the Zig 0.16 changes (the rebase is trivial).

jcollie (approved):
LGTM but let's see if @tristan957  has any further feedback.

tristan957:
> LGTM but let's see if @tristan957 has any further feedback.

No feedback. Feel free to merge!

PR #13419, opened 2026-07-22
pkg/apple-sdk: support Xcode 27 SDK headers

mitchellh:
Xcode 27's math.h uses the __need_infinity_nan protocol provided by matching Clang resource headers. Zig 0.16's bundled float.h predates that protocol, causing the bundled libc++ compilation to fail.

Overlay the SDK math.h through the Apple SDK libc include path and provide the missing infinity and NaN definitions. The compatibility header can be removed once Zig's bundled Clang headers support the protocol.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,6 +102,7 @@ jobs:
       - build-nix
       # - build-nix-macos
       - build-macos
+      - build-macos-xcode-27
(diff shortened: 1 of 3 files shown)


mitchellh:
Added an xcode 27 CI job.

bo2themax (approved):
🫶🏻

PR #13427, opened 2026-07-23
fix(lib-vt): link iOS dylib with Apple ld

elias8:
Addresses [libghostty issue #111](https://github.com/elias8/libghostty/issues/111): iOS apps using flterm/libghostty build successfully but are rejected by App Store because the packaged `ghostty.framework` lacks a valid `LC_ENCRYPTION_INFO_64` load command and appears not to have been built with Apple’s linker.

For physical iOS targets, this links the existing static archive into the dynamic library using Apple’s linker so the packaged framework contains the expected encryption information. Other targets continue using the existing build path.

Longterm this should be fixed in Zig’s Mach O linker. For now, I think this is the quickest way to resolve this. And allthough not specific to this, there are Mach O related issues in the Zig repo. 

Note: GPT 5.5 did the job mostly.

### DIFF
--- a/src/build/GhosttyLibVt.zig
+++ b/src/build/GhosttyLibVt.zig
@@ -102,6 +102,17 @@ pub fn initShared(
     b: *std.Build,
     zig: *const GhosttyZig,
 ) !GhosttyLibVt {
+    const target = zig.vt.resolved_target.?;
+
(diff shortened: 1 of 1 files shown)


mitchellh:
Thanks, replaced by https://github.com/ghostty-org/ghostty/pull/13430

PR #13452, opened 2026-07-25
build: fix static libghostty-vt linking on Windows

noib3:
This PR fixes static linking for libghostty-vt on Windows by propagating a couple of missing dependencies (discovered while running Neovim's Zig build, see [this CI run](https://github.com/neovim/neovim/actions/runs/30130848061/job/89604799965?pr=39773)).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -742,6 +742,10 @@ jobs:
       - name: Build libghostty-vt
         run: zig build -Demit-lib-vt
 
+      - name: Test static linking
+        working-directory: example/c-vt-static
+        run: zig build
+
   build-libghostty-windows-gnu:
     runs-on: namespace-profile-ghostty-windows
     timeout-minutes: 45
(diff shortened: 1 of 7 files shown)


mitchellh:
Thanks for updating our CI too. Let’s run it and see. Looks good thoigh

noib3:
Could you approve CI again?

noib3:
The build is still failing and I'm kinda out of ideas here. I asked Sol to give it a shot and it proposed a couple of fixes which I don't really understand bc I'm not familiar with MSVC or Windows in general. Should I push them or close this?

mitchellh:
Okay here's what we can do. This fixes the build, this adds CI to test it, is narrowly scoped to windows+msvc, and this unblocks Neovim. But, we don't fully understand the set of changes here. Neovim is likely to have a high number of real Windows/MSVC users that will validate these changes rather quickly (since we have no better way atm). So, I think the best course of action is to clearly comment/document the changes we don't have high understanding of, be honest about it, and merge it in.

I'm going to pull this down now to add those comments.

mitchellh:
Actually, once I pulled it down and looked at it closer, this all seemed sane to me. I think we're good.

PR #13454, opened 2026-07-25
deps: update translate-c backport

vancluever:
Was brought up as possibly being a build issue here: https://codeberg.org/vancluever/translate-c/pulls/1

I do think that this is the better approach and seems to be the close equivalent of the `.zig_ilb` option that's coming with `LazyPath` in 0.17.0 (which is how translate-c behaves there).

I was looking for something like this initially and I _think_ I might have passed over it to start with because it was a bit hard to determine the circumstances that `b.graph.zig_lib_directory` would be null, but upon further examination, I think such cases would be rare if they happened at all. Rather than default to the cwd in this event though I just get it to error out - that way we'll know if it ever is the case!

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -11,8 +11,8 @@
         // Aro/translate-c itself.
         .translate_c = .{
             .lazy = true,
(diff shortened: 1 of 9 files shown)


mitchellh:
Looks good, I'll merge but will want to update the mirrors at some point.

PR #13464, opened 2026-07-26
feat: implement global shortcuts through vicinae-hotkey-v1

aurelleb:
This PR provides an implementation for the [vicinae-hotkey-v1](https://github.com/vicinaehq/vicinae-wayland-protocols/tree/main/staging/vicinae-hotkey) protocol, as discussed [here](https://github.com/ghostty-org/ghostty/discussions/13453).

This is a wayland protocol that allows the client to dynamically negotiate global shortcuts with the compositor. Unlike the portal, the clients are free to bind, rebind, and unbind global shortcuts they reserve.

Here are a few advantages of using this over the global shortcut portal for ghostty specifically:

- `vicinae-hotkey-v1` lets the client know the state of its bindings at all time, if a global bind is not granted by the compositor the cllient is notified with a descriptive error message which is designed to help the user understand what the problem might be. In my implementation, I decided I would show a desktop notification to the user in case a global shortcut reservation fails.

- Global binds set in the config cannot drift from what is actually registered, cannot be unilaterally changed by the user in compositor settings, and do not pollute the global shortcut namespace permanently. Reservations are only active while ghostty is running.

- The protocol provides the client with an input serial that can be used to generate an `xdg_activation` token, allowing ghostty to steal focus when one of its global shortcut is used. Currently ghostty doesn't have an input serial to pass to `xdg_activation`. I didn't wire it for now, in order keep things simple. But I guess it will be a nice to have.

---

AI disclosure: most of the code was written by Fable 5 (Claude Code), as zig is not my primary language.

From an implementation perspective: I made it so that `vicinae-hotkey-v1` is used to manage global shortcuts over the portal when the global is advertised by the compositor. If it's not available, we fallback on the portal like before.

At this time the protocol is implemented by Hyprland (since [v0.56.0](https://github.com/hyprwm/Hyprland/pull/15010)) and there is an open PR for [niri](https://github.com/niri-wm/niri/pull/4145). There is also an official [wayland-protocols proposal](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/525).

As for the clients, for now the only implementer (that I know of) is [vicinae](https://github.com/vicinaehq/vicinae).

The idea would be to merge this `vicinae-hotkey-v1` implementation and then have it be superseded by the upstream version later, assuming it is turned into an official extension protocol following the wayland-protocols process.

PS: sorry for the diff noise about all the .po changes, I'm not sure whether this is intended or not, but I did regenerate the translations as asked

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -1471,6 +1471,14 @@ pub const Application = extern struct {
             self,
             .{},
         );
+
+        _ = GlobalShortcuts.signals.@"bind-failed".connect(
+            priv.global_shortcuts,
+            *Application,
+            globalShortcutBindFailed,
+            self,
+            .{},
+        );
     }
 
     fn activate(self: *Self) callconv(.c) void {
@@ -1677,6 +1685,41 @@ pub const Application = extern struct {
         self.core().performAllAction(self.rt(), action.*) catch |err| {
             log.warn("failed to perform action={}", .{err});
         };
+    }
+
+    /// May fire before any window exists, hence a desktop notification
+    /// rather than a toast.
+    fn globalShortcutBindFailed(
+        _: *GlobalShortcuts,
+        failure: *const GlobalShortcuts.BindFailed,
+        self: *Self,
+    ) callconv(.c) void {
+        var label_buf: [128]u8 = undefined;
+        var writer: std.Io.Writer = .fixed(&label_buf);
+        const label: []const u8 = label: {
+            const ok = key.labelFromTrigger(&writer, failure.trigger) catch false;
+            break :label if (ok) writer.buffered() else "?";
+        };
+
+        const detail: [*:0]const u8 = detail: {
+            if (failure.message[0] != 0) break :detail failure.message;
+            break :detail if (failure.revoked)
+                i18n._("The keybind was revoked by the system.")
+            else
+                i18n._("The keybind was denied by the system.");
+        };
+
+        var body_buf: [512]u8 = undefined;
+        const body = std.fmt.bufPrintZ(
+            &body_buf,
+            "{s}: {s}",
+            .{ label, detail },
+        ) catch return;
+
+        Action.desktopNotification(self, .app, .{
+            .title = std.mem.span(i18n._("Global keybind unavailable")),
+            .body = body,
+        });
     }
 
     fn actionReloadConfig(
--- a/src/apprt/gtk/key.zig
+++ b/src/apprt/gtk/key.zig
@@ -57,6 +57,16 @@ pub fn xdgShortcutFromTrigger(
     return slice[0 .. slice.len - 1 :0];
 }
 
+/// Returns the keysym (identical to a GDK keyval) for a trigger key,
+/// or null if the trigger has no keysym representation.
+pub fn keysymFromTrigger(trigger: input.Binding.Trigger) ?c_uint {
+    return switch (trigger.key) {
+        .physical => |k| keyvalFromKey(k),
+        .unicode => |cp| gdk.unicodeToKeyval(cp),
+        .catch_all => null,
+    };
+}
+
 fn writeTriggerKey(
     writer: *std.Io.Writer,
     trigger: input.Binding.Trigger,
@@ -325,6 +335,28 @@ test "xdgShortcutFromTrigger" {
         .mods = .{ .ctrl = true, .alt = true, .super = true, .shift = true },
         .key = .{ .unicode = 92 },
     })).?);
+}
+
+test "keysymFromTrigger" {
+    const testing = std.testing;
+
+    // Unicode keys use the keysym numbering (latin-1 maps directly).
+    try testing.expectEqual(@as(?c_uint, 'q'), keysymFromTrigger(.{
+        .mods = .{ .super = true },
+        .key = .{ .unicode = 'q' },
+    }));
+
+    // Physical keys map through our keymap.
+    try testing.expectEqual(@as(?c_uint, gdk.KEY_a), keysymFromTrigger(.{
+        .mods = .{},
+        .key = .{ .physical = .key_a },
+    }));
+
+    // catch_all has no keysym representation.
+    try testing.expectEqual(@as(?c_uint, null), keysymFromTrigger(.{
+        .mods = .{},
+        .key = .catch_all,
+    }));
 }
 
 test "labelFromTrigger" {
--- a/src/apprt/gtk/winproto/wayland/Hotkeys.zig
+++ b/src/apprt/gtk/winproto/wayland/Hotkeys.zig
@@ -106,7 +106,6 @@ fn bindOne(
 
     const entry = try self.alloc.create(Entry);
     errdefer self.alloc.destroy(entry);
-    try self.entries.ensureUnusedCapacity(self.alloc, 1);
 
     const hotkey = try manager.bind(
         keysym,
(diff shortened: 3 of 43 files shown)


src/apprt/gtk/class/application.zig
```diff
@@ -1679,6 +1687,41 @@ pub const Application = extern struct {
...
+        const body = std.fmt.bufPrintZ(
+            &body_buf,
+            "{s}: {s}",
+            .{ label, detail },
+        ) catch return;
+
+        Action.desktopNotification(self, .app, .{
+            .title = std.mem.span(i18n._("Global keybind unavailable")),
+            .body = body,
+        });
```
pluiedev, line 1722: I feel like this should instead use a toast or something... but we can change the UI side separately later

src/apprt/gtk/key.zig
```diff
@@ -57,6 +57,16 @@ pub fn xdgShortcutFromTrigger(
     return slice[0 .. slice.len - 1 :0];
 }
 
+/// Returns the keysym (identical to a GDK keyval) for a trigger key,
+/// or null if the trigger has no keysym representation.
+pub fn keysymFromTrigger(trigger: input.Binding.Trigger) ?c_uint {
+    return switch (trigger.key) {
+        .physical => |k| keyvalFromKey(k),
+        .unicode => |cp| gdk.unicodeToKeyval(cp),
+        .catch_all => null,
```
pluiedev, line 66: I'm not sure the `catch_all` path is truly reachable (correct me if I'm wrong)

src/apprt/gtk/winproto/wayland/Hotkeys.zig
```diff
@@ -0,0 +1,163 @@
...
+const Config = @import("../../../../config.zig").Config;
+const Binding = @import("../../../../input.zig").Binding;
+const key = @import("../../key.zig");
+const GlobalShortcuts = @import("../../class/global_shortcuts.zig").GlobalShortcuts;
+
+const log = std.log.scoped(.winproto_wayland_hotkeys);
+
+alloc: Allocator,
+app_id: [:0]const u8,
+entries: std.ArrayListUnmanaged(*Entry) = .empty,
```
pluiedev: We're in 0.16 so it's just std.ArrayList now

src/apprt/gtk/winproto/wayland/Hotkeys.zig
```diff
@@ -0,0 +1,163 @@
...
+    shortcuts: *GlobalShortcuts,
+    trigger: Binding.Trigger,
+    action: Binding.Action,
+) !void {
+    const keysym = key.keysymFromTrigger(trigger) orelse return error.NoKeysym;
+
+    var desc_buf: [256]u8 = undefined;
+    const description = std.fmt.bufPrintZ(&desc_buf, "{f}", .{action}) catch "";
+
+    const entry = try self.alloc.create(Entry);
```
pluiedev: Allocating each entry just creates more hassle than it's worth, simply call `addOne` and you'll get the pointer to the newly added entry for free. There's no need to ensureUnusedCapacity then immediately appendAssumeCapacity either.

src/apprt/gtk/winproto/wayland/Hotkeys.zig
```diff
@@ -0,0 +1,163 @@
...
+    event: vicinae.HotkeyV1.Event,
+    entry: *Entry,
+) void {
+    switch (event) {
+        .bound => log.debug("hotkey bound action={f}", .{entry.action}),
+
+        .denied => |v| {
+            log.warn("hotkey denied action={f} reason={} message={s}", .{
+                entry.action,
+                @intFromEnum(v.reason),
```
pluiedev: Why do `@intFromEnum` and sacrifice the perfectly legible names?

pluiedev (changes requested):
Overall looks pretty good, some nitpicks here and there

src/apprt/gtk/class/application.zig
aurelleb: we could, but if it's revoked later the ghostty window might not be focused, so if the window is in another workspace the user won't immediately know.

trag1c: I'm by no means familiar with the code, but when looking at other stuff I had a hunch that perhaps this should be using `performAction(_, .desktop_notification)` instead of `Action.desktopNotification` 🤔

trag1c (approved):
The `.po` changes look fine. Approving to unblock because of the `manager` team review request, though IDK how reasonable the punctuation is wrt the HIG.

src/apprt/gtk/key.zig
aurelleb: As  I understand it a binding like `keybind = global:catch_all=whatever` would reach that path, like it does in the portal implementation

src/apprt/gtk/winproto/wayland/Hotkeys.zig
aurelleb: The issue with that is that regrowing the array list would dangle the pointers which are passed to the listener callbacks. So I don't think it's a viable solution.

src/apprt/gtk/winproto/wayland/Hotkeys.zig
aurelleb: no good reason indeed, fixed

src/apprt/gtk/winproto/wayland/Hotkeys.zig
pluiedev: Hmmmm... to my understanding the list of global hotkeys is always replaced wholesale and never grows in between config reloads, so this shouldn't (IMO) be a problem

aurelleb: Yes but the array list is bound to grow when doing the initial population, since Hotkeys.zig doesn't know how many keybinds are going to be populated and thus cannot reserve the exact size.

Even assuming we could reserve the exact size, I would still not take the risk to pass an easily danglable pointer like that, it's a footgun for later

pluiedev: Hmm, fair point. I would encourage using an ArenaAllocator here actually to avoid needing to individually clean up each tiny allocation in case of e.g. error handling (and to make life easier in general)

aurelleb: Fair, I refactored it to use an arena allocator

src/apprt/gtk/class/application.zig
aurelleb: It's called like this in other places in the codebase so I thought I would just do the same.
Also as far as I can tell both are functionally equivalent

pluiedev (approved):
I think at this point I don't have any more major concerns. Thanks!!

src/apprt/gtk/class/application.zig
trag1c: > It's called like this in other places in the codebase

Where? I grepped for `desktopNotification` and only see the switch case in `performAction` and the definition of the action itself.

> Also as far as I can tell both are functionally equivalent

Yeah, but I would assume `performAction` is preferred as it's the apprt API for it and it seems to be used a few times in `Application` already.

Again though I'm not very familiar so maybe @pluiedev can chime in w/ more useful thoughts on whether this matters at all or not. 🫡

pluiedev: It doesn't make that much of a difference here

mitchellh:
Amazing! Super fast!

PR #13466, opened 2026-07-26
cli: fix list-themes preview lifecycle

jparise:
Start the vaxis event loop so the theme preview can receive terminal input, and retain its environment map for as long as vaxis may access it.

### DIFF
--- a/src/cli/list_themes.zig
+++ b/src/cli/list_themes.zig
@@ -241,6 +241,7 @@ const Preview = struct {
     allocator: std.mem.Allocator,
     should_quit: bool,
     tty: vaxis.Tty,
+    env_map: std.process.Environ.Map,
     vx: vaxis.Vaxis,
     mouse: ?vaxis.Mouse,
     themes: []ThemeListElement,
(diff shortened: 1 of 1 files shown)


src/cli/list_themes.zig
```diff
@@ -296,13 +297,22 @@ const Preview = struct {
...
         var loop: vaxis.Loop(Event) = .init(global.io(), &self.tty, &self.vx);
+        try loop.start();
+
         const writer = self.tty.writer();
+        defer {
+            loop.should_quit = true;
+            self.vx.deviceStatusReport(writer) catch {};
+            writer.flush() catch {};
+            if (loop.thread) |*thread| thread.await(global.io());
+        }
```
jparise: It seems like `defer loop.stop()` should be enough, so maybe this is something that should be considered in libvaxis upstream.

mitchellh:
Gotta run the proper update deps script to fix the build I think.

jparise:
> Gotta run the proper update deps script to fix the build I think.

Right. I don't have a nix install handy right now, so that's the holdup.

jparise:
> Gotta run the proper update deps script to fix the build I think.

The fix itself doesn't need the vaxis dep update after all, so I dropped it.

jcollie (approved):

PR #13473, opened 2026-07-27
Introduce `scrollback-limit-lines` to limit scrollback by lines instead of bytes

mitchellh:
This adds a new config `scrollback-limit-lines` to limit scrollback by lines instead of bytes. This also renames `scrollback-limit` to `scrollback-limit-bytes` to make it clear what it does but we have a compatibility entry so old configurations will continue to work, so its not breaking.

**This is NOT exclusive to `scrollback-limit-bytes`**. When both are set, then the _first limit reached_ is used. Since lines is affected by viewport size and bytes are affected by entries (more styles, more graphemes, etc.), they serve somewhat different purposes and it might be useful to set both.

The default remains 50MB of bytes, unlimited lines.

This is not exposed to libghostty yet. I have that coming as a follow up change.

### DIFF
--- a/src/benchmark/ScrollbackCompression.zig
+++ b/src/benchmark/ScrollbackCompression.zig
@@ -138,7 +138,7 @@ pub fn create(
         .terminal = try .init(global.io(), alloc, .{
             .rows = opts.@"terminal-rows",
             .cols = opts.@"terminal-cols",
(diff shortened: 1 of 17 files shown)


mitchellh:
I also ran our ghostty-bench on `+terminal-stream` and there was no regressions from a baseline. All good.

PR #13481, opened 2026-07-27
libghostty: scrollback limits can be changed at runtime

### LINKED ISSUE #13268: vt: support changing `max_scrollback` after terminal creation
Discussed in https://github.com/ghostty-org/ghostty/discussions/12770

This will require some core changes since its not currently runtime change-able and it has pruning/preloading implications. Shouldn't be too hard to do though. Required by Neovim.

mitchellh:
**This is an ABI breaking change for C libs.**

Fixes https://github.com/ghostty-org/ghostty/issues/13268

You can now change scrollback limits (bytes and lines) at runtime. 

This breaks the ABI but makes for a much more long-term pattern: `ghostty_terminal_new` now only takes viewport size, and you set scrollback configurations with the generic `ghostty_terminal_set`. The `GhosttyTerminalOptions` struct is fully removed. I think this will serve us much better over the long term.

### DIFF
--- a/include/ghostty/vt/terminal.h
+++ b/include/ghostty/vt/terminal.h
@@ -173,26 +173,6 @@ extern "C" {
  */
 
 /**
- * Terminal initialization options.
- *
- * @ingroup terminal
- */
-typedef struct {
-  /** Terminal width in cells. Must be greater than zero. */
-  uint16_t cols;
-
-  /** Terminal height in cells. Must be greater than zero. */
-  uint16_t rows;
-
-  /** Maximum number of lines to keep in scrollback history. */
-  size_t max_scrollback;
-
-  // TODO: Consider ABI compatibility implications of this struct.
-  // We may want to artificially pad it significantly to support
-  // future options.
-} GhosttyTerminalOptions;
-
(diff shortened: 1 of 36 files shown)


include/ghostty/vt/terminal.h
```diff
@@ -888,6 +868,40 @@ typedef enum GHOSTTY_ENUM_TYPED {
...
+  /**
+   * Set the maximum scrollback allocation in bytes.
+   *
+   * This is an estimate. Internally, libghostty only prunes bytes up 
+   * to a "page"-granularity. A page is the minimum allocated unit of
+   * grid space within Ghostty. A page at the time of writing these docs
+   * is about 400KB, so the byte limit will be within this delta.
+   *
+   * Lowering the limit immediately removes eligible complete historical
+   * pages. A value of zero disables scrollback and erases retained history.
```
pluiedev, line 887: I think it's worth mentioning how the two limits interact here, since otherwise people might assume they conflict somehow

mitchellh: Good call. Done.

PR #13489, opened 2026-07-27
font/sprite: update to z2d 0.12.1, use native path insetting

vancluever:
This change updates z2d to 0.12.1 and changes the sprite font path insetting functionality to use the new path offset abilities released in the update.

In addition, there has been a slight change to the drawing of E0B5 and its respective reflection; we now add a 1-pixel horizontal line segment to each end to force them to be perpendicular. This is because offsetting pre-expands the curves and ultimately causes the end segments of the curve itself to have slight non-horizontal angles, which produce small artifacts at the ends without the forced horizontal ends.


vancluever:
Note the test file changes are due to the standard AA perturbations as a result of the change in path calculations (now using offsetting) and the addition of the forced horizontal ends on the curve glyphs.

mitchellh (approved):
Awesome!

PR #13494, opened 2026-07-28
terminal: add visibility reports with GTK suspension tracking

rockorager:
## Summary

Applications cannot reliably determine whether an unfocused terminal is still
visible, so focus reports alone are insufficient for avoiding unnecessary
rendering.

This adds terminal visibility reporting by:

- implementing private mode 2033
- supporting `CSI ? 998 n` visibility queries and `CSI ? 999 ; Ps n` responses
- reporting effective visibility changes while mode 2033 is enabled
- preserving host-owned visibility state across terminal resets
- treating unknown visibility conservatively as potentially visible

On GTK 4.12 and newer, surface visibility now combines widget mapping with the
toplevel `suspended` state. This allows Ghostty to recognize windows hidden on
another workspace or otherwise known by the compositor to be non-visible.
Older GTK versions retain the existing conservative behavior.

## Testing

Added coverage for:

- mode 2033 support and enable/disable behavior
- explicit visibility queries
- immediate reports when enabling the mode
- visible and non-visible responses
- visibility persistence across terminal resets
- suppression of visibility queries in read-only mode

## AI disclosure

Amp assisted with the implementation, tests, commit messages, and this pull
request description. I reviewed the resulting changes and understand how they
interact with the terminal, termio, surface, and GTK visibility paths.

Implements: #13451 
Reference: https://rockorager.dev/misc/visibility-reports/

### DIFF
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -1374,9 +1374,31 @@ pub const Window = extern struct {
             }
         }
 
+        // Notify every displayed surface when the compositor changes the
+        // Wayland xdg_toplevel suspended state.
+        _ = gobject.Object.signals.notify.connect(
+            self.as(gtk.Window),
+            *Self,
+            propSuspended,
+            self,
+            .{ .detail = "suspended" },
+        );
+
         // When we are realized we always setup our appearance since this
         // calls some winproto functions.
         self.syncAppearance();
+    }
+
+    fn propSuspended(
+        _: *gtk.Window,
+        _: *gobject.ParamSpec,
+        self: *Self,
+    ) callconv(.c) void {
+        const tab = self.getSelectedTab() orelse return;
+        const tree = tab.getSurfaceTree() orelse return;
+
+        var it = tree.iterator();
+        while (it.next()) |entry| entry.view.updateOcclusion();
     }
 
     fn btnNewTab(_: *adw.SplitButton, self: *Self) callconv(.c) void {
--- a/src/terminal/stream_terminal.zig
+++ b/src/terminal/stream_terminal.zig
@@ -507,7 +507,22 @@ pub const Handler = struct {
                 buf[writer.end] = 0;
                 self.writePty(buf[0..writer.end :0]);
             },
+
+            .visibility => self.sendVisibilityReport(),
         }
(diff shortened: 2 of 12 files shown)


src/apprt/gtk/class/window.zig
```diff
@@ -1374,11 +1374,35 @@ pub const Window = extern struct {
             }
         }
 
+        // GTK exposes Wayland's xdg_toplevel suspended state starting in
+        // 4.12. Notify every displayed surface when the compositor changes it.
+        if (gtk_version.atLeast(4, 12, 0)) {
```
tristan957: Isn't our minimum GTK version much higher than 4.12?

jcollie: 4.14 is the minimum so I agree that we don't need this conditional.

https://ghostty.org/docs/linux#supported-gtkadwaita-versions

jcollie:
Not as part of this PR, but "someone" should really add documentation regarding all of the modes that we support to the Ghostty website.

tristan957 (approved):
I am approving the GTK part! I think the motivation for this mode makes a ton of sense. Getting others to buy in is a different story.

rockorager:
> Getting others to buy in is a different story.

I've got feedback from Kovid, dnkl, and rapha (rio) that are supportive of this mode. In my view of the terminal world, that's a pretty healthy agreement :)

src/terminal/stream_terminal.zig
```diff
@@ -507,9 +507,22 @@ pub const Handler = struct {
                 buf[writer.end] = 0;
                 self.writePty(buf[0..writer.end :0]);
             },
+
+            .visibility => self.sendVisibilityReport(),
         }
     }
 
+    fn sendVisibilityReport(self: *Handler) void {
```
mitchellh, line 515: This should have an early exit if `write_pty` is not set like others, just to avoid unnecessary wrk.

mitchellh (changes requested):
One request

src/terminal/stream_terminal.zig
rockorager: Done!

jcollie (approved):

PR #13498, opened 2026-07-28
docs: clarify macOS dependencies

vegerot:
reword: The doc said "macOS doesn't need any dependencies" and then immediately listed things you needed to install for macOS 😁.  This is just rewording the doc to be more consistent.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -40,14 +40,14 @@ here:
 
 ## Extra Dependencies
 
+### Linux
+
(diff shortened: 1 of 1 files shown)


mitchellh:
Fair enough, I meant like outside of standard build environments (e.g. I don't list the compiler an extra dep on Linux either). But I can see how this is confusing.

mitchellh:
!vouch

PR #13514, opened 2026-07-29
macOS: rename `download` to `install` for `auto-update`

bo2themax:
[automaticallyDownloadsUpdates](https://sparkle-project.org/documentation/api-reference/Classes/SPUUpdater.html#/c:objc(cs)SPUUpdater(py)automaticallyDownloadsUpdates) will actually install the update as well. `download` could be misleading.

There will be following prs to update current behaviour of when to show `installing` pill as well.

### AI Disclosure

Looks trivial but no AI involved, just a quick find and replace in VSCode


mitchellh (changes requested):
I prefer more descriptive names, so lets do `install`?

bo2themax:
> I prefer more descriptive names, so lets do `install`?

Sure, I actually did that before opening this, then thought on/off would be better

PR #13515, opened 2026-07-29
datastruct/circ_buf: fix metadata after shrinking

fallintoplace:
## Summary

- normalize circular-buffer metadata after every resize
- retain the oldest values when shrinking below the current length
- cover partial shrink, exact-length shrink, and empty-to-zero boundaries

## Root cause

`resize` rotated live values to index zero before reallocating, but only repaired `head` and `full` when capacity grew. Shrinking a partially filled buffer could therefore leave `head` beyond the new allocation and report a length greater than capacity. A later append could index outside the resized storage.

## Validation

- `zig fmt --check src/datastruct/circ_buf.zig`
- `zig test src/circ_buf_test.zig --test-filter 'CircBuf resize'` using a temporary import harness: 8 tests passed


jcollie (approved):
LGTM

PR #13519, opened 2026-07-30
inspector: add copy and export for terminal IO events

Uzaaft:
Adds "Copy" and "Export to file" buttons to the Terminal IO inspector so recorded VT events can be saved outside the app for sharing or analysis.

Found myself needing/wishing for this while I was debugging my tmux fork with libghostty-vt. 

Disclaimer: I haven't considered performance at all, so please lmk if there are anything here you would like me to optimize.

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -703,6 +703,11 @@ pub const Application = extern struct {
             .initial_size => return Action.initialSize(target, value),
 
             .inspector => return Action.controlInspector(target, value),
+            .export_terminal_io => return try Action.exportTerminalIO(
+                self,
+                target,
+                value,
+            ),
 
             .key_sequence => return Action.keySequence(target, value),
             .key_table => return Action.keyTable(target, value),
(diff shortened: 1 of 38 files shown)


bo2themax (approved):
Image

Uzaaft:
I'm gonna fix the build errors 🫡

src/apprt/gtk/class/application.zig
```diff
@@ -1997,6 +2002,99 @@ const Action = struct {
...
+                null,
+                null,
+                &gerr,
+            );
+            if (gerr) |err| {
+                defer err.free();
+                log.err(
+                    "failed to export terminal IO events err={s}",
+                    .{err.f_message orelse "(unknown)"},
+                );
```
tristan957, line 2088: This should probably also be an in app notification

trag1c (approved):
The `.po` changes look fine. Our v1.4 string freeze starts today (not formalized in a document yet) but I think it's fine to squeeze this one in.

src/apprt/gtk/class/application.zig
jcollie: That probably should be some sort of popup or similar in the inspector, not in Ghostty proper

src/inspector/widgets/termio.zig
```diff
@@ -93,6 +96,16 @@ pub const Stream = struct {
...
+
+                cimgui.c.ImGui_SameLineEx(0, cimgui.c.ImGui_GetStyle().*.ItemInnerSpacing.x);
+                if (cimgui.c.ImGui_Button("Copy")) {
+                    self.copyEvents(surface);
+                }
+
+                cimgui.c.ImGui_SameLineEx(0, cimgui.c.ImGui_GetStyle().*.ItemInnerSpacing.x);
+                if (cimgui.c.ImGui_Button("Export to file")) {
+                    self.exportEvents(surface);
+                }
```
jcollie, line 108: I find it a little distracting that these buttons appear/disappear depending on whether there are any events buffered. But that can be a fix for another day.

jcollie (approved):
There are some things that should be addressed in follow-up PRs but otherwise LGTM.

src/inspector/widgets/termio.zig
Uzaaft: @jcollie I can get behind that. I'll make them always be visible.

PR #13527, opened 2026-07-31
macos: add detach_split keybind action

gadgetman6:
## What

Adds a `detach_split` keybind action (macOS only) that moves the focused split into its own new window — the keyboard equivalent of grabbing the split's drag handle and dropping it outside the window.

Related: discussion #12126 (full surface mobility), which proposes `detach_split` with window/tab arguments. This implements the window case only, as a plain no-argument action.

## How

The macOS drag-and-drop system already implements the full detach operation in `BaseTerminalController.ghosttySurfaceDragEndedNoTarget`: removing the surface from its split tree, moving it (running program, contents, cwd) into a new window, and registering a "Move Split" undo group. That observer is registered for the controller's lifetime, not just during drags.

This PR plumbs a new `detach_split` binding action through the standard path (`input.Binding.Action` → `apprt.Action` → `GHOSTTY_ACTION_DETACH_SPLIT`), and the macOS handler posts the existing `.ghosttySurfaceDragEndedNoTarget` notification for the focused surface. No drop point in the notification means the new window gets default positioning, which `TerminalController.newWindow` already handles.

The action returns `false` (not performable) when the surface is not part of a split tree, matching `toggle_split_zoom` semantics. GTK lists it as unimplemented. A command palette entry ("Detach Split") is included.

## Testing

- `zig build` on macOS (Xcode 26.2, Zig 0.16.0), then manual testing of the debug build with `--keybind=cmd+shift+o=detach_split`: detaching a pane from a split moves it into its own window with the running program intact. The action is a no-op (returns unperformable) on a lone surface, and undo/redo come from the existing drag-path undo group.
- `ghostty +list-actions` and `+list-keybinds` show the new action; the command palette entry appears.

## AI disclosure

Per `AI_POLICY.md`: this change was implemented with Claude Code (Claude Fable 5). The implementation is AI-written under my direction; I reviewed the diff, built it locally, and manually tested the resulting app before submitting.


src/input/Binding.zig
```diff
@@ -643,6 +643,15 @@ pub const Action = union(enum) {
...
 
+    /// Detach the current split and move it into its own new window.
+    ///
+    /// The running program, contents, and working directory all move with
+    /// the split. If the focused surface is not part of a split tree (it
+    /// is already alone in its window), this does nothing.
+    ///
+    /// This is currently only supported on macOS.
+    detach_split,
+
```
jcollie, line 676: We should have a note indicating which version of Ghostty this first was available.

jcollie:
!vouch

jcollie (approved):
No comment on the swift changes but the core changes LGTM.

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1375,6 +1374,36 @@ extension Ghostty {
...
+                // If the window has no splits, there is nothing to detach.
+                guard controller.surfaceTree.isSplit else { return false }
+
+                // The no-target drag handler implements exactly this operation:
+                // remove the surface from its tree and move it into a new window.
+                NotificationCenter.default.post(
+                    name: .ghosttySurfaceDragEndedNoTarget,
+                    object: surfaceView
+                )
+                return true
```
bo2themax:
```suggestion
                guard
                    let surface = target.target.surface,
                    let surfaceView = self.surfaceView(from: surface),
                    let controller = BaseTerminalController.controller(owning: surfaceView),
                    // If the window has no splits, there is nothing to detach.
                    controller.surfaceTree.isSplit,
                    let window = controller.window
                else {
                    return false
                }

                let surfaceFrameInWindow = surfaceView.convert(surfaceView.frame, to: nil)
                let windowTopLeft = NSPoint(x: surfaceFrameInWindow.minX, y: surfaceFrameInWindow.maxY)
                NotificationCenter.default.post(
                    name: .ghosttySurfaceDragEndedNoTarget,
                    object: surfaceView,
                    userInfo: [
                        Foundation.Notification.Name.ghosttySurfaceDragEndedNoTargetPointKey: window.convertPoint(toScreen: windowTopLeft)
                    ]
                )
                return true
```

bo2themax (changes requested):
Added the position and changes some code style. Please rebase before applying.

gadgetman6:
Hey @jcollie @bo2themax thanks for the suggestions and help - totally missed the follow ups on this, will get it rebased so we can merge.

trag1c (changes requested):
Blocking on the string additions

src/input/Binding.zig
jcollie: Since this adds a translatable string and we are now in string freeze, this will have to wait to be merged until after 1.4 is released so this isn't quite correct now. Not sure what the version after 1.4 will be called.

gadgetman6:
@trag1c split the palette entry (and the `.pot` change) out into #14020, so this PR adds no new strings anymore -- it's keybind-only now, and `detach_split` is mapped to no palette commands in `command.zig`. #14020 is stacked on this one and can wait until after 1.4.

Ok to merge this one without the palette change?

jcollie:
> Ok to merge this one without the palette change?

No, we want to keep the command palette changes and the addition of the keybind in the same PR.

gadgetman6:
Ok, we can wait for the freeze to lift. Brought the palette entry and the `.pot` change back into this PR.

PR #13530, opened 2026-07-31
lib-vt: answer XTGETTCAP queries

fornwall:
Ghostty's full termio path answers XTGETTCAP from the static terminfo map, but `terminal/stream_terminal.zig`, which backs libghostty-vt, parses the same DCS request and then discards it. There is no XTGETTCAP effect either, so an embedder cannot restore the replies through the C API.

Programs query these over SSH instead of assuming the remote host has the client's terminfo entry. This matters more for an embedder than for the desktop app, which can install its entry on the remote through shell integration.

Answer the queries in `stream_terminal` the same way termio does: look up each requested key in the static terminfo map and write the reply to the pty, skipping the lookups entirely when no `write_pty` effect is set. The map now stores null-terminated responses so they can be handed straight to `write_pty` without copying. `terminal/dcs.zig` and the termio path are unchanged.

`TN` is handled separately. It names the terminfo entry the terminal runs as, so it has to agree with `TERM` — which is set in `termio/Exec.zig`, a layer libghostty-vt does not contain. The library never sees `TERM` and cannot answer on the embedder's behalf, and answering with Ghostty's own entry from the static map would misreport every embedder, so `TN` is intercepted before the map lookup. The name is instead configured through a new option, `GHOSTTY_TERMINAL_OPT_TERMINFO_NAME`: the string is copied into the terminal, names longer than 128 bytes are rejected, and while unset the query goes unanswered.

This is the first dependency from `src/terminal` on `src/terminfo`, so libghostty-vt now carries Ghostty's terminfo table: +16,023 bytes (+1.9%) on a `wasm32-freestanding` `ReleaseSmall` build.

---

AI usage: Created with claude code and opus 5. I have reviewed the code and made modifications where it made sense. I have also tested this end to end in an application.

### DIFF
--- a/include/ghostty/vt/terminal.h
+++ b/include/ghostty/vt/terminal.h
@@ -1210,6 +1210,22 @@ typedef enum GHOSTTY_ENUM_TYPED {
    * Input type: size_t*
    */
   GHOSTTY_TERMINAL_OPT_UNKNOWN_MAX_BYTES = 36,
+
+  /**
+   * Set the name of the terminfo entry this terminal runs as, reported
+   * in response to an XTGETTCAP query for "TN" (e.g. "xterm-256color").
+   *
+   * The string data is copied into the terminal. A NULL value pointer
+   * clears the name (equivalent to setting an empty string). A name
+   * longer than 128 bytes returns GHOSTTY_INVALID_VALUE.
+   *
+   * If this is unset then we don't report anything for an XTGETTCAP
+   * TN query, because we don't know what the embedding terminal around
+   * libghostty is advertising itself as.
+   *
+   * Input type: GhosttyString*
+   */
+  GHOSTTY_TERMINAL_OPT_TERMINFO_NAME = 37,
   GHOSTTY_TERMINAL_OPT_MAX_VALUE = GHOSTTY_ENUM_MAX_VALUE,
 } GhosttyTerminalOption;
 
--- a/src/terminal/c/terminal.zig
+++ b/src/terminal/c/terminal.zig
@@ -98,6 +98,9 @@ const TerminalWrapper = struct {
     /// We also need to store a temp dir path for some operations (e.g., kitty
     /// graphics). This provides stable storage for the API calls.
     tmp_dir_path: [max_path_bytes]u8,
+    /// The terminfo name reported for XTGETTCAP "TN". The stream handler holds
+    /// a slice into this.
+    terminfo_name_buf: [Handler.max_terminfo_name_bytes]u8,
     stream: Stream,
     effects: Effects = .{},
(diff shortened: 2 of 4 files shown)


fornwall:
I have not created any discussion and didn't see any existing for this - let me know if I should create one!

mitchellh (changes requested):
The problem statement is valid but the code quality in the solution is a bit of a mess that should be cleaned up.

include/ghostty/vt/terminal.h
```diff
@@ -90,6 +90,7 @@ extern "C" {
  * | `GHOSTTY_TERMINAL_OPT_PWD_CHANGED`      | `GhosttyTerminalPwdChangedFn`     | Pwd change via OSC 7 / OSC 9 / OSC 1337   |
  * | `GHOSTTY_TERMINAL_OPT_ENQUIRY`          | `GhosttyTerminalEnquiryFn`        | ENQ character (0x05)                      |
  * | `GHOSTTY_TERMINAL_OPT_XTVERSION`        | `GhosttyTerminalXtversionFn`      | XTVERSION query (CSI > q)                 |
+ * | `GHOSTTY_TERMINAL_OPT_TERMINFO_NAME`    | `GhosttyTerminalTerminfoNameFn`   | XTGETTCAP query for "TN" (DCS + q)        |
```
mitchellh: I thought about this and I don't think this should be a callback function. I think this should be an option with a string parameter that is documented as needing to be valid for the lifetime of the terminal.

mitchellh (changes requested):

include/ghostty/vt/terminal.h
fornwall: > I thought about this and I don't think this should be a callback function. I think this should be an option with a string parameter that is documented as needing to be valid for the lifetime of the terminal.

I have updated the PR.

I think it's better to copy than to borrow here - borrowing gives little for what it costs here, right? And copying aligns with the existing `GHOSTTY_TERMINAL_OPT_TITLE`, `GHOSTTY_TERMINAL_OPT_PWD` and `GHOSTTY_TERMINAL_OPT_KITTY_IMAGE_MEDIUM_TEMP_FILE` options.

Let me know what you think regarding borrowing vs copying.

fornwall:
This PR is ready for review. Compared to the initial version:

- Rebased against latest main resolving merge conflicts.
- Switched from callback function to option (note: copying instead of borrowing here - [motivation](https://github.com/ghostty-org/ghostty/pull/13530#discussion_r3695648896)).
- Moved away from violating the scope of `dcs.zig`, instead letting `stream_terminal.zig` do the work.
- General code cleanup and de-duplication.
- Updated PR description.

mitchellh:
Thanks, I gave you conflicts again. Besides that, this is looking good. I'm going to go in and do some idiomatic Zig cleanup but that's just subjective...

mitchellh (approved):
Pending CI

Issue #9393, opened 2025-10-29
`ssh-terminfo` doesn't cache if `~/.local/state/` doesn't exist


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/9359

<div type='discussions-op-text'>

<sup>Originally posted by **injust** October 26, 2025</sup>
### Issue Description

If you remove `~/.local/state/` (and presumably $XDG_STATE_HOME is unset), `ssh-terminfo` doesn't cache.

---

I grabbed this from https://github.com/ghostty-org/ghostty/discussions/8872:

> No errors are shown but if you manually run `ghostty +ssh-cache --add=example.com`

```
Error: Unable to add 'example.com' to cache. Error: error.FileNotFound
```

### Expected Behavior

Ghostty creates `~/.local/state/` if it doesn't exist.

### Actual Behavior

`ssh-terminfo` runs on every SSH connection.

### Reproduction Steps

1. Remove ~`/.local/state/`
2. SSH to any server twice

### Ghostty Version

```text
Ghostty 1.3.0-main+2f5c09c1b

Version
  - version: 1.3.0-main+2f5c09c1b
  - channel: tip
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS Sequoia 15.7.1

### Minimal Ghostty Configuration

```ini
shell-integration-features = ssh-terminfo
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

jparise:
I'll take a look at this given that I've been working on `+ssh-cache` recently.

satmandu:
Just wanted to add a note here that if the permissions on ~/.local/state are wrong, one gets an access denied message that doesn't actually tell the user where there is a permission problem.

e.g.,
```bash
ghostty +ssh-cache --add=user@host
Error: Unable to add 'user@host' to cache. Error: error.AccessDenied
```

[...]

PR #13533, opened 2026-07-31
cli: report ssh terminfo cache failures

jparise:
A state directory with the wrong permissions left the terminfo cache failing with errors that named no path, so there was nothing to act on:

    $ ghostty +ssh-cache --add=user@host
    Error: Unable to add 'user@host' to cache. Error: error.AccessDenied

Every +ssh-cache failure now names its cache file, and +ssh no longer swallows cache-related errors.

Error messages in these actions are also lowercased after the "Error: "
prefix and append the error with ": {t}" rather than a second "Error: ".

Ref: https://github.com/ghostty-org/ghostty/issues/9393#issuecomment-5145799368

### DIFF
--- a/src/cli/ssh-cache/DiskCache.zig
+++ b/src/cli/ssh-cache/DiskCache.zig
@@ -393,6 +393,16 @@ fn readEntries(
     return entries;
 }
 
+/// Whether an error from a cache operation reflects a problem with the cache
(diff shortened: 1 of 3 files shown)


mitchellh:
Thanks. The windows failures are real.

PR #13535, opened 2026-07-31
build: update iOS deployment target to 13

elias8:
The core dependencies used in libghostty-vt don't require newer iOS deployment targets, so we can use the older version to maximize compatibility with consumers targeting older iOS versions.

This is primarily for convenience (the same can be achieved with `-Dtarget` build option). It establishes the baseline for future prebuilt dylibs, which must be compiled for earlier supported iOS versions by consumers.

Initially applied the build option to the [libghostty binding](https://github.com/elias8/libghostty/pull/133), but then thought the default belongs here.

### DIFF
--- a/src/build/Config.zig
+++ b/src/build/Config.zig
@@ -692,9 +692,9 @@ pub fn osVersionMin(tag: std.Target.Os.Tag) ?std.Target.Query.OsVersion {
             .patch = 0,
         } },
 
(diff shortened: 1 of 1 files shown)


mitchellh:
Looks like our Metal shaders rely on higher versions. I know these aren't required for libghostty-vt so if you want to focus the change on that then feel free to try that, going to close this for now.

PR #13539, opened 2026-08-01
build: lower iOS deployment target version

elias8:
Reopening #13535.


mitchellh:
Actually I want to rethink this. Every source I can find online shows that iOS 13 usage is less than 0.5% and even iOS 17 usage is down to like 1%. I think our minimum is reasonable versus introducing something like this. Is there a specific reason you're trying to lower it? Or just because you can?

elias8:
It is based on actual [use case](https://github.com/elias8/libghostty/issues/127). Frameworks like flutter and RN also default to lower deployment targets. I currently work around this by passing `-Dtarget` when building the dart bindings but if we are going to build versioned dylibs, I think libghostty should target a lower version by default. Consumers can still build from source when needed but this would make common cases more convenient

PR #13543, opened 2026-08-01
libghostty(formatter): fix superfluous newline in html formatting

RoniJacobson:
In the html formatter every page is formatted in a div. When the div closes it causes a newline in the html rendering. In order to fix this a newline is now removed whenever the div is closed (if there are any newlines waiting to be rendered - as far as I could see in my testing there was always one).

I thought of trying to add a test but could not think of a way to do so without adding a massive blob of html into the file.

Before (orange was added by me to show where the div ends):
image

After:
image

No AI was used in this pr.

### DIFF
--- a/src/terminal/formatter.zig
+++ b/src/terminal/formatter.zig
@@ -1354,6 +1354,9 @@ pub const PageFormatter = struct {
                     closing.len,
                 );
             }
+            // Closing the div creates a newline in the output
+            // so make sure to create one less newline
+            if (blank_rows >= 1) blank_rows -= 1;
         }
 
         return .{ .rows = blank_rows, .cells = blank_cells };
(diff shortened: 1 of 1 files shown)


jcollie (changes requested):
This needs a test. Not sure what you mean by "massive blob" but that's what `@embedFile` is for. If it's only referenced in a test it won't affect the size of the production binary.

RoniJacobson:
I initially thought I would have to include the whole html page I showed in my screenshots for the test, but after taking another look I realized it was just a matter of using the `PageFormatter` itself for the test :)

jcollie (approved):
Approved pending CI

RoniJacobson:
Looks like ci has passed but the merge didn't work?

PR #13549, opened 2026-08-02
termio: fix doc comment grammar and PTY casing in message.zig

12ya:
### What
Comment-only cleanup in `src/termio/message.zig`:

- Fixes a subject-verb agreement error in the `Message` union's doc
  comment: "the number of messages ... are also very few" -> "is also
  very small"
- Capitalizes "pty" to "PTY" in several doc comments 

### Why
Caught while reading through `message.zig`. No functional changes,
doc comments only.


trag1c:
!vouch

mitchellh:
Sure, although we use pty/PTY casing all over the place, I don't want to see a ton of PRs to fix all this lo.

PR #13553, opened 2026-08-02
config: update scrollbar doc per current implementation

bo2themax:
We should also update the doc after #9865, which is discussed in https://github.com/ghostty-org/ghostty/discussions/9610

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -1438,9 +1438,9 @@ input: RepeatableReadableIO = .{},
 /// Valid values:
 ///
 ///   * `system` - Respect the system settings for when to show scrollbars.
-///     For example, on macOS, this will respect the "Scrollbar behavior"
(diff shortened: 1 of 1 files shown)


src/config/Config.zig
```diff
@@ -1438,9 +1438,9 @@ input: RepeatableReadableIO = .{},
 /// Valid values:
 ///
 ///   * `system` - Respect the system settings for when to show scrollbars.
-///     For example, on macOS, this will respect the "Scrollbar behavior"
-///     system setting which by default usually only shows scrollbars while
-///     actively scrolling or hovering the gutter.
+///     On macOS, we only shows scrollbars while actively scrolling or hovering
```
trag1c:
```suggestion
///     On macOS, we only show scrollbars while actively scrolling or hovering
```

mitchellh (approved):

PR #13567, opened 2026-08-03
crash: do not use global state

vancluever:
This removes use of global state from the crash reporting functionality (everything in `src/crash`).

This particularly ensures that there are no races on the system environment during the execution of the initialization thread that would possibly cause crashes, particularly in any (albeit unsupported) 3rd party integrations of libghostty-internal.

Ultimately, this pushes any coupling of I/O and environment to places that would more correctly interface with global state, such as the same-thread `global.init`, and the crash report CLI.

Note that similar de-coupling actions have been taken on XDG and home directory functionality, pushing their coupling points up the stack in a similar way.


vancluever:
Should fix #13547

mitchellh (approved):

PR #13577, opened 2026-08-03
core: avoid copying OSC 52 clipboard responses

jparise:
OSC 52 clipboard reads built their response in an allocated buffer and then passed it through Message.writeReq, which allocated a second copy for large responses.

Instead, transfer the allocated response directly using .write_alloc.

Small responses now retain their initial allocation until the IO thread consumes them instead of being copied inline and freed immediately. Their allocation count is unchanged, while large responses improve from two allocations to one. Both cases avoid the additional copy.


jcollie (approved):
LGTM

PR #13579, opened 2026-08-03
termio: free resources for discarded messages

jparise:
Messages can own allocated data or a derived config. Some paths (writer thread draining, mailbox shutdown with unread messages, and queue push failures) discarded messages without releasing those resources.

This change adds Message.deinit and uses it whenever a message is discarded.


jcollie (approved):

mitchellh (approved):
Good finds. 😄

PR #13586, opened 2026-08-03
gtk: use native blur on GTK 4.23.3+

### LINKED ISSUE #13581: GTK: Disable custom blur & use builtin blur on GTK >=4.23
See #13578. We have to manually disable our custom implementation since it would otherwise conflict with GTK's. Using GTK's own blur should be as easy as inserting a blur node into the Surface widget's snapshot.

pluiedev:
Finally, what was previously thought impossible, is now possible.
The blur region itself is far more accurate than what we can conjure up on our own, and in a much more finetuned and detailed way too.
Thank you, GTK devs!

Closes #13581

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -37,6 +37,7 @@ const InspectorWindow = @import("inspector_window.zig").InspectorWindow;
 const i18n = @import("../../../os/i18n.zig");
 const media = @import("../media.zig");
 const global = @import("../../../global.zig");
+const gtk_version = @import("../gtk_version.zig");
 
 const log = std.log.scoped(.gtk_ghostty_surface);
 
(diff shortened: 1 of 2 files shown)


tristan957 (approved):
This looks good. Are there any behavioral differences between this and what we currently do? Asking in the context of doc changes

pluiedev:
Basically none. The way this works is that GTK traverses the scenegraph and tries to find visible nodes with a blur value above 20, then it marks the visible region as being blurred and sends it to the Wayland compositor. This is pretty much what we already do, only much much *much* smarter since now it is per-widget rather than per-window. You can potentially have differently-blurred splits, in fact

pluiedev:
For reference this is how it looks with this PR:

image

Versus our previous manual approach:

image

I sure can't tell a difference

mitchellh (approved):

PR #13597, opened 2026-08-04
gtk: do not set bell ringing if already focused

lotheac:


tristan957:
Is this what we do on the macOS side? It is good to match behaviors. I'd also like to know if other terminals operate this way, specifically alacritty, kitty, and foot.

lotheac:
> Is this what we do on the macOS side? It is good to match behaviors. I'd also like to know if other terminals operate this way, specifically alacritty, kitty, and foot.

not sure what macOS does, but it might have had a similar bug in the past: https://github.com/ghostty-org/ghostty/issues/8487

note that the bell ringing state only affects the terminal title (🔔 is added); `ringBell()` is called regardless of focus and regardless of whether the bell state was already ringing. that part sets the window urgency, so really this PR is only about not adding the bell icon to the title if the terminal is already focused. I'm not sure if alacritty/kitty/foot even have any manipulation of the window title based on bell; I used alacritty previously and I don't think I ever saw that.

tristan957:
Thanks for getting back to me! I would like another maintainer to review the UX change.

pluiedev (approved):
I think this makes sense. If anybody feels strongly that the old behavior should be the case we can always change it back or add a setting

PR #13601, opened 2026-08-04
macos: defer transparent titlebar KVO rebinding

mitchellh:
Fixes #13386, based on https://github.com/mustafa0x/ghostty/commit/a8c090

Defer transparent-titlebar KVO rebinding to the next main-queue turn. Track the observed tab group so unchanged bindings are preserved.

Previously, a tab-group callback could invalidate and recreate its own observation before returning, leaving closed terminal windows registered with AppKit after the undo timeout. These windows accumulated titlebar and layer state, increasing memory use and WindowServer CPU with tab churn.

Validated with an AppDelegate change that sat and created/closed tabs in a loop, then counted weak controllers/windows/nsapp window.


bo2themax (approved):
Interesting 🫪

PR #13605, opened 2026-08-04
Fix mode 2031 reporting light when using a non-conditional theme despite a dark GTK color scheme

### LINKED ISSUE #13604: Mode 2031 reports do not work unless both a dark and a light theme are configured
Update:

I managed to find the source of the issue.

With an isolated Ghostty (debug build) instance launched using:

```fish
ghostty  --config-default-files=false --window-theme=dark
```

Ghostty's GTK log correctly reports:

```text
debug(gtk_ghostty_application): style manager changed scheme=.dark
```

However, querying the current color scheme with `CSI ? 996 n` returns:

```text
CSI ? 997 ; 2 n
```

which indicates light mode.

I was able to narrow this down further by configuring an actual conditional theme pair:

```ini
theme = light:Rose Pine Dawn,dark:Rose Pine
```

With that configuration, the same query correctly returns:

```text
CSI ? 997 ; 1 n
```

The issue appears to be in `Config.changeConditionalState`. The function only rebuilds the configuration when the changed state key is present in `_conditional_set`. A single, non-conditional theme such as:

```ini
theme = Kitty Default
```

does not add `.theme` to that set. As a result, changing the application color scheme from the default `.light` state to `.dark` is treated as irrelevant and the function returns `null`.

The surface and GTK runtime therefore know that the color scheme is dark, but the configuration copied into Termio retains its default `_conditional_state.theme = .light`. Since the mode-2031 response reads the theme from the Termio configuration, it reports light.

The following minimal patch treats the theme state as relevant even when the configured theme itself is not conditional, and updates the existing test accordingly:

```diff
diff --git a/src/config/Config.zig b/src/config/Config.zig
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -4416,7 +4416,7 @@
 
             // Conditional set contains the keys that this config uses. So we
             // only continue if we use this key.
-            if (self._conditional_set.contains(key) and !deepEqual(
+            if ((key == .theme or self._conditional_set.contains(key)) and !deepEqual(
                 @TypeOf(@field(self._conditional_state, field.name)),
                 @field(self._conditional_state, field.name),
                 @field(new, field.name),
@@ -10587,7 +10587,7 @@ test "changed" {
     try testing.expect(!source.changed(&dest, .@"font-size"));
 }
 
-test "changeConditionalState ignores irrelevant changes" {
+test "changeConditionalState preserves non-conditional theme state" {
     const testing = std.testing;
     const alloc = testing.allocator;
 
@@ -10600,9 +10600,12 @@ test "changeConditionalState ignores irrelevant changes" {
         try cfg.loadIter(alloc, &it);
         try cfg.finalize();
 
-        try testing.expect(try cfg.changeConditionalState(
-            .{ .theme = .dark },
-        ) == null);
+        var cfg2 = (try cfg.changeConditionalState(.{
+            .theme = .dark,
+        })).?;
+        defer cfg2.deinit();
+
+        try testing.expectEqual(.dark, cfg2._conditional_state.theme);
     }
 }
```

I changed the existing `changeConditionalState` test because it explicitly asserted the behavior responsible for the bug: a light-to-dark state change with a non-conditional theme was expected to return `null`. That assumption is no longer valid because the theme state is observable through mode 2031 even when it does not change any configured values. The revised test therefore verifies that the new configuration preserves `_conditional_state.theme = .dark`. 

The drawback is that this replaces the test coverage for the original optimization, where genuinely irrelevant conditional-state changes should still return `null`. If there is a better way to implement a new test for this, I would defer to that instead.

This was reproduced against commit 6687d60.

_Originally posted by @UnsaltedScholar in https://github.com/ghostty-org/ghostty/discussions/13448#discussioncomment-17889061_

UnsaltedScholar:
Fixes #13604

First reported in #13448 

Edit (mostly copied from my disclaimer edit in my other PR):
I truly apologize for not including this earlier, I was a tad overeager after my vouch request was approved and I was quite rushed and ended up creating this pull request on my phone during lunch and had several errands afterwards. 

AI Usage Disclaimer:
GPT 5.6 Sol was used to understand the project structure. I asked it to tell me where the mode 2031 theme switch occurred and it pointed me toward the function that I originally edited. Just to be sure I asked it to review my changes and it reminded me to change the test that was affected.

I apologize again for forgetting to add this, it won't happen again.

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -60,6 +60,11 @@ font_grid_set: font.SharedGridSet,
 last_notification_time: ?std.Io.Timestamp = null,
 last_notification_digest: u64 = 0,
 
+/// The current system color scheme. This is kept separate from conditional
+/// configuration state so mode 2031 remains correct for non-conditional
+/// themes.
+system_color_scheme: std.atomic.Value(apprt.ColorScheme) = .init(.light),
+
 /// The conditional state of the configuration. See the equivalent field
 /// in the Surface struct for more information. In this case, this applies
 /// to the app-level config and as a default for new surfaces.
@@ -392,6 +397,11 @@ pub fn colorSchemeEvent(
     rt_app: *apprt.App,
     scheme: apprt.ColorScheme,
 ) !void {
+    // Cache the system color scheme and notify every surface when it changes.
+    if (self.system_color_scheme.swap(scheme, .monotonic) != scheme) {
+        try self.performAllAction(rt_app, .system_color_scheme_changed);
+    }
+
     const new_scheme: configpkg.ConditionalState.Theme = switch (scheme) {
         .light => .light,
         .dark => .dark,
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -678,6 +678,7 @@ pub fn init(
             .size = size,
             .full_config = config,
             .config = try termio.Termio.DerivedConfig.init(alloc, config),
+            .system_color_scheme = &app.system_color_scheme,
             .backend = .{ .exec = io_exec },
             .mailbox = io_mailbox,
             .renderer_state = &self.renderer_state,
@@ -4756,9 +4757,6 @@ pub fn colorSchemeCallback(self: *Surface, scheme: apprt.ColorScheme) !void {
     // Setup our conditional state which has the current color theme.
     self.config_conditional_state.theme = new_scheme;
     self.notifyConfigConditionalState();
-
-    // If mode 2031 is on, then we report the change live.
-    self.queueIo(.{ .color_scheme_report = .{ .force = false } }, .unlocked);
 }
 
 pub fn posToViewport(self: Surface, xpos: f64, ypos: f64) terminal.point.Coordinate {
@@ -4845,6 +4843,14 @@ pub fn performBindingAction(self: *Surface, action: input.Binding.Action) !bool 
     }
 
     switch (action.scoped(.surface).?) {
+        .system_color_scheme_changed => {
+            // Termio suppresses this report unless mode 2031 is enabled.
+            self.queueIo(
+                .{ .color_scheme_report = .{ .force = false } },
+                .unlocked,
+            );
+        },
+
         .csi, .esc => |data| {
             // We need to send the CSI/ESC sequence as a single write request.
             // If you split it across two then the shell can interpret it
--- a/src/termio/Options.zig
+++ b/src/termio/Options.zig
@@ -1,5 +1,6 @@
 //! The options that are used to configure a terminal IO implementation.
 
+const std = @import("std");
 const xev = @import("../global.zig").xev;
 const apprt = @import("../apprt.zig");
 const renderer = @import("../renderer.zig");
@@ -16,6 +17,9 @@ full_config: *const Config,
 
 /// The derived configuration for this termio implementation.
 config: termio.Termio.DerivedConfig,
+
+/// The system color scheme used for mode 2031 reports.
+system_color_scheme: *const std.atomic.Value(apprt.ColorScheme),
 
 /// The backend for termio that implements where reads/writes are sourced.
 backend: termio.Backend,
--- a/src/termio/Termio.zig
+++ b/src/termio/Termio.zig
@@ -37,6 +37,9 @@ backend: termio.Backend,
 /// The derived configuration for this termio implementation.
 config: DerivedConfig,
 
+/// The system color scheme used for mode 2031 reports.
+system_color_scheme: *const std.atomic.Value(apprt.ColorScheme),
+
 /// The terminal emulator internal state. This is the abstract "terminal"
 /// that manages input, grid updating, etc. and is renderer-agnostic. It
 /// just stores internal state about a grid.
@@ -300,6 +303,7 @@ pub fn init(self: *Termio, alloc: Allocator, opts: termio.Options) !void {
         .alloc = alloc,
         .terminal = term,
         .config = opts.config,
+        .system_color_scheme = opts.system_color_scheme,
         .renderer_state = opts.renderer_state,
         .renderer_wakeup = opts.renderer_wakeup,
         .renderer_mailbox = opts.renderer_mailbox,
@@ -715,10 +719,11 @@ pub fn colorSchemeReportLocked(self: *Termio, td: *ThreadData, force: bool) !voi
     if (!force and !self.renderer_state.terminal.modes.get(.report_color_scheme)) {
         return;
     }
-    const scheme: terminalpkg.device_status.ColorScheme = switch (self.config.conditional_state.theme) {
-        .light => .light,
-        .dark => .dark,
-    };
+    const scheme: terminalpkg.device_status.ColorScheme =
+        switch (self.system_color_scheme.load(.monotonic)) {
+            .light => .light,
+            .dark => .dark,
+        };
 
     var buf: [terminalpkg.device_status.max_color_scheme_report_encode_size]u8 = undefined;
     var writer: std.Io.Writer = .fixed(&buf);
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -1280,6 +1280,10 @@ pub const Application = extern struct {
         const priv = self.private();
         priv.config.unref();
         priv.config = config.ref();
+
+        // Apply GTK appearance settings from the new config.
+        self.syncStyleManager();
+
         self.as(gobject.Object).notifyByPspec(properties.config.impl.param_spec);
 
         // Show our errors if we have any
@@ -1421,15 +1425,12 @@ pub const Application = extern struct {
         }
     }
 
-    /// Setup the style manager on startup. The primary task here is to
-    /// setup our initial light/dark mode based on the configuration and
-    /// setup listeners for changes to the style manager.
-    fn startupStyleManager(self: *Self) void {
+    /// Sync the configured window theme.
+    fn syncStyleManager(self: *Self) void {
         const priv = self.private();
         const config = priv.config.get();
-
-        // Setup our initial light/dark
         const style = self.as(adw.Application).getStyleManager();
+
         style.setColorScheme(switch (config.@"window-theme") {
             .auto, .ghostty => auto: {
                 const lum = config.background.toTerminalRGB().perceivedLuminance();
@@ -1442,6 +1443,16 @@ pub const Application = extern struct {
             .dark => .force_dark,
             .light => .force_light,
         });
+    }
+
+    /// Setup the style manager on startup. The primary task here is to
+    /// setup our initial light/dark mode based on the configuration and
+    /// setup listeners for changes to the style manager.
+    fn startupStyleManager(self: *Self) void {
+        const style = self.as(adw.Application).getStyleManager();
+
+        // Setup our initial light/dark.
+        self.syncStyleManager();
 
         // Setup color change notifications
         _ = gobject.Object.signals.notify.connect(
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -193,6 +193,9 @@ pub const Application = extern struct {
         /// only be set by the main loop thread.
         running: bool = false,
 
+        /// Whether the current configuration reload is a soft reload.
+        config_reload_soft: bool = false,
+
         /// The timer used to quit the application after the last window is
         /// closed. Even if there is no quit delay set, this is the state
         /// used to determine to close the app.
@@ -241,6 +244,11 @@ pub const Application = extern struct {
     pub fn default() *Self {
         const app = gio.Application.getDefault().?;
         return gobject.ext.cast(Self, app).?;
+    }
+
+    /// Returns whether the current configuration reload is soft.
+    pub fn configReloadIsSoft(self: *Self) bool {
+        return self.private().config_reload_soft;
     }
 
     /// Creates a new Application instance.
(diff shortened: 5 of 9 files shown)


tristan957 (approved):
@jcollie I'll let you merge this after your review since you were involved already.

jcollie (changes requested):
This "fixes" the issue, but I don't think that it's the correct fix. Mode 2031 color reports basically "cheated" and used the Ghostty config theme to cache the last known color scheme to avoid sending multiple DSRs to TUIs when the scheme hadn't changed. This change fixes the problem that DSRs were not sent when the system theme changes by always making the theme part of the conditional set, but that will force extraneous Ghostty config changes when the system theme changes but the user configured Ghostty for a single theme for both light and dark system themes.

I think that the proper fix is for the core app to maintain a cache of the current system theme and send out mode 2031 DSRs separately from reloading the config.

https://github.com/ghostty-org/ghostty/blob/48d85eaeb06ac9fc49073815bda5bac97de655ca/src/App.zig#L390-L413

UnsaltedScholar:
@jcollie 

Is it preferable to keep the mode 2031 color-scheme cache entirely at the app level, as you suggested, or would it be more beneficial to retain a per-surface cache so surfaces with different effective `window-theme` settings can report their own schemes correctly (assuming this is something Ghostty wants to allow)?

As a corollary to this, should I remove `conditional_state` from `Termio.DerivedConfig`, since its current use appears limited to mode 2031 reporting, or should it remain, even with the duplicated conditional state, for any potential Termio features in the future?

Something like this (this compiles/fixes the issue on linux, have no way to test MacOS currently):

Edit: This particular blurb is an example that was partially generated by AI (specifically the `report_changed` bool block and the added comments). This code is not intended as the final version (which will be written by me); it is solely meant to be a quick example implementation to confirm if I was thinking in the right direction.
```diff
diff --git a/src/App.zig b/src/App.zig
--- a/src/App.zig
+++ b/src/App.zig
@@ -8,6 +8,12 @@ font_grid_set: font.SharedGridSet,
 last_notification_time: ?std.Io.Timestamp = null,
 last_notification_digest: u64 = 0,

+/// The current color scheme reported by the application runtime. This is
+/// intentionally separate from the conditional configuration state because
+/// mode 2031 must track the runtime scheme even when no conditional
+/// configuration depends on it.
+color_scheme: apprt.ColorScheme,
+
 /// The conditional state of the configuration. See the equivalent field
 /// in the Surface struct for more information. In this case, this applies
 /// to the app-level config and as a default for new surfaces.
@@ -37,6 +43,7 @@ pub fn init(
         .surfaces = .empty,
         .mailbox = .{},
         .font_grid_set = font_grid_set,
+        .color_scheme = .light,
         .config_conditional_state = .{},
     };
 }
@@ -61,12 +68,16 @@ pub fn colorSchemeEvent(
     rt_app: *apprt.App,
     scheme: apprt.ColorScheme,
 ) !void {
+    // Cache the runtime color scheme independently of the configuration.
+    // New surfaces use this as their initial mode 2031 report value.
+    self.color_scheme = scheme;
+
     const new_scheme: configpkg.ConditionalState.Theme = switch (scheme) {
         .light => .light,
         .dark => .dark,
     };

-    // If our scheme didn't change, then we don't do anything.
+    // Only reload configuration when its conditional state changed.
     if (self.config_conditional_state.theme == new_scheme) return;

     // Setup our conditional state which has the current color theme.
     
diff --git a/src/Surface.zig b/src/Surface.zig
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -2,6 +2,7 @@
             .size = size,
             .full_config = config,
             .config = try termio.Termio.DerivedConfig.init(alloc, config),
+            .color_scheme = app.color_scheme,
             .backend = .{ .exec = io_exec },
             .mailbox = io_mailbox,
             .renderer_state = &self.renderer_state,
@@ -28,20 +29,37 @@ pub fn colorSchemeCallback(self: *Surface, scheme: apprt.ColorScheme) !void {
     crash.sentry.thread_state = self.crashThreadState();
     defer crash.sentry.thread_state = null;

+    // Update the runtime state used by mode 2031 independently of the
+    // conditional configuration. This state is protected by the renderer
+    // mutex because the IO thread reads it while encoding reports.
+    const report_changed: bool = report_changed: {
+        self.renderer_state.mutex.lockUncancelable(global.io());
+        defer self.renderer_state.mutex.unlock(global.io());
+
+        if (self.io.color_scheme == scheme) break :report_changed false;
+        self.io.color_scheme = scheme;
+        break :report_changed true;
+    };
+
+    // Report runtime scheme changes without requiring a config reload.
+    if (report_changed) {
+        self.queueIo(
+            .{ .color_scheme_report = .{ .force = false } },
+            .unlocked,
+        );
+    }
+
     const new_scheme: configpkg.ConditionalState.Theme = switch (scheme) {
         .light => .light,
         .dark => .dark,
     };

-    // If our scheme didn't change, then we don't do anything.
+    // Only reload configuration when its conditional state changed.
     if (self.config_conditional_state.theme == new_scheme) return;

     // Setup our conditional state which has the current color theme.
     self.config_conditional_state.theme = new_scheme;
     self.notifyConfigConditionalState();
-
-    // If mode 2031 is on, then we report the change live.
-    self.queueIo(.{ .color_scheme_report = .{ .force = false } }, .unlocked);
 }

 pub fn posToViewport(self: Surface, xpos: f64, ypos: f64) terminal.point.Coordinate {

diff --git a/src/termio/Options.zig b/src/termio/Options.zig
--- a/src/termio/Options.zig
+++ b/src/termio/Options.zig
@@ -6,6 +6,9 @@ full_config: *const Config,
 /// The derived configuration for this termio implementation.
 config: termio.Termio.DerivedConfig,

+/// The current runtime color scheme used for mode 2031 reports.
+color_scheme: apprt.ColorScheme,
+
 /// The backend for termio that implements where reads/writes are sourced.
 backend: termio.Backend,

diff --git a/src/termio/Termio.zig b/src/termio/Termio.zig
--- a/src/termio/Termio.zig
+++ b/src/termio/Termio.zig
@@ -4,6 +4,10 @@ backend: termio.Backend,
 /// The derived configuration for this termio implementation.
 config: DerivedConfig,

+/// The current runtime color scheme used for mode 2031 reports. This is
+/// protected by the renderer state mutex.
+color_scheme: apprt.ColorScheme,
+
 /// The terminal emulator internal state. This is the abstract "terminal"
 /// that manages input, grid updating, etc. and is renderer-agnostic. It
 /// just stores internal state about a grid.
@@ -32,7 +36,6 @@ pub const DerivedConfig = struct {
     osc_color_report_format: configpkg.Config.OSCColorReportFormat,
     clipboard_write: configpkg.ClipboardAccess,
     enquiry_response: []const u8,
-    conditional_state: configpkg.ConditionalState,

     pub fn init(
         alloc_gpa: Allocator,
@@ -53,7 +56,6 @@ pub const DerivedConfig = struct {
             .osc_color_report_format = config.@"osc-color-report-format",
             .clipboard_write = config.@"clipboard-write",
             .enquiry_response = try alloc.dupe(u8, config.@"enquiry-response"),
-            .conditional_state = config._conditional_state,

             // This has to be last so that we copy AFTER the arena allocations
             // above happen (Zig assigns in order).
@@ -77,6 +79,7 @@ pub const DerivedConfig = struct {
         .alloc = alloc,
         .terminal = term,
         .config = opts.config,
+        .color_scheme = opts.color_scheme,
         .renderer_state = opts.renderer_state,
         .renderer_wakeup = opts.renderer_wakeup,
         .renderer_mailbox = opts.renderer_mailbox,
@@ -110,7 +113,7 @@ pub fn colorSchemeReportLocked(self: *Termio, td: *ThreadData, force: bool) !voi
     if (!force and !self.renderer_state.terminal.modes.get(.report_color_scheme)) {
         return;
     }
-    const scheme: terminalpkg.device_status.ColorScheme = switch (self.config.conditional_state.theme) {
+    const scheme: terminalpkg.device_status.ColorScheme = switch (self.color_scheme) {
         .light => .light,
         .dark => .dark,
     };
```

jcollie:
> @jcollie
> 
> Is it preferable to keep the mode 2031 color-scheme cache entirely at the app level, as you suggested, or would it be more beneficial to retain a per-surface cache so surfaces with different effective `window-theme` settings can report their own schemes correctly (assuming this is something Ghostty wants to allow)?

This would be a cache of the system color scheme so app-level caching is more appropriate since the surfaces all belong to the same system.

> As a corollary to this, should I remove `conditional_state` from `Termio.DerivedConfig`, since its current use appears limited to mode 2031 reporting, or should it remain, even with the duplicated conditional state, for any potential Termio features in the future?

If it's not used to anything else, it should be removed.

> Something like this (this compiles/fixes the issue on linux, have no way to test MacOS currently):

Don't have the time to look at the diff in detail now, will take a look when the PR is updated, but that seems like the right approach.

UnsaltedScholar:
> This would be a cache of the system color scheme so app-level caching is more appropriate since the surfaces all belong to the same system.

> If it's not used to anything else, it should be removed.

The rework I just pushed should be correct in this regard now (and is much more minimal than the "example" I provided above). If there are any further problems, please let me know.

The only notable thing is that it defaults to dark mode, I think this is reasonable given most people use dark terminals, but if this is a problem I can change it.

src/termio/Options.zig
```diff
@@ -17,6 +18,9 @@ full_config: *const Config,
 /// The derived configuration for this termio implementation.
 config: termio.Termio.DerivedConfig,
 
+/// The system color scheme used for mode 2031 reports.
+color_scheme: *const std.atomic.Value(u8),
```
jcollie: This should be named `system_color_scheme` to emphasize that it's the system scheme.

src/termio/Termio.zig
```diff
@@ -37,6 +37,9 @@ backend: termio.Backend,
 /// The derived configuration for this termio implementation.
 config: DerivedConfig,
 
+/// The system color scheme used for mode 2031 reports.
+color_scheme: *const std.atomic.Value(u8),
```
jcollie: Ditto on the name.

src/termio/Termio.zig
```diff
@@ -300,6 +301,7 @@ pub fn init(self: *Termio, alloc: Allocator, opts: termio.Options) !void {
         .alloc = alloc,
         .terminal = term,
         .config = opts.config,
+        .color_scheme = opts.color_scheme,
```
jcollie: Ditto on the names.

src/App.zig
```diff
@@ -60,6 +60,12 @@ font_grid_set: font.SharedGridSet,
 last_notification_time: ?std.Io.Timestamp = null,
 last_notification_digest: u64 = 0,
 
+/// The current system color scheme. This is kept separate from conditional
+/// configuration state so mode 2031 remains correct for non-conditional
+/// themes.
+color_scheme: std.atomic.Value(u8) =
```
jcollie: Ditto on the name. Atomic values are not limited to primitive ints, `apprt.ColorScheme` can be used directly because it's an `enum` backed by an explicit integer.

src/Surface.zig
```diff
@@ -676,6 +676,7 @@ pub fn init(
             .size = size,
             .full_config = config,
             .config = try termio.Termio.DerivedConfig.init(alloc, config),
+            .color_scheme = &app.color_scheme,
```
jcollie: Ditto on the name.

src/App.zig
```diff
@@ -392,6 +398,9 @@ pub fn colorSchemeEvent(
     rt_app: *apprt.App,
     scheme: apprt.ColorScheme,
 ) !void {
+    // Cache the system color scheme independently of conditional config.
+    self.color_scheme.store(@intFromEnum(scheme), .monotonic);
```
jcollie: Pretty sure that we need to compare against the previous value and trigger all the surfaces to send a report, probably through `performAllAction`.

jcollie (changes requested):
The basics are looking good but I don't think that we're sending reports now.

src/termio/Options.zig
UnsaltedScholar: done

src/App.zig
UnsaltedScholar: Done, required changing the backing type of ColorScheme from u2 to u8, I got a compile error otherwise.

src/App.zig
UnsaltedScholar: Done

UnsaltedScholar:
Sorry for the delay. I made the changes you requested, however, when I was manually testing it I noticed some related problems that I fixed in the last two commits. 

For one, reloading the config from the right-click context menu after changing the window-theme setting didn't update everything correctly. The second issue I noticed when fixing the first was that reloading caused a duplicate toast/notification to appear when the color scheme actually needed to change (it didn't duplicate when switching from system to dark if the system theme was already dark).

If you guys would rather those last two things be addressed in a different PR I can do that as well, but they seemed relevant enough to what I was fixing here so I added them.

Minutia:
A third issue I noticed was that `window-theme = auto` doesn't really behave exactly as described, I think. I have background set to pure black and foreground to pure white, but the window theme still changes when I change KDE theme (it basically acts like a copy of the `system` option ignoring the other Ghostty config options). I didn't fix that as I thought it was beyond the scope of this PR.

The AI usage disclaimer from before applies, though I did also have it review the current changes to help make sure I didn't mess anything up in other parts of Ghostty.

jcollie:
> A third issue I noticed was that `window-theme = auto` doesn't really behave exactly as described, I think.

Yeah I think that the description of`window-theme=auto` isn't very good. But I think that it's better to start a discussion to come to a consensus before starting on a PR.

jcollie (approved):
This looks good enough for now. I've got some ideas for follow-ups later but let's get this in so we can get feedback.

gotenksIN:
spent my morning finding out why pi was loading light colour scheme despite window-theme being set to dark, rebuilt with this PR and its fixed now :D

PR #13606, opened 2026-08-04
fix synthetic font application 

UnsaltedScholar:
Fix synthetic font application for fonts without native italic/bold face but for which fontconfig returns non-italic/bold face on query.

First reported in: #13583

Edit:
I truly apologize for not including this earlier, I was a tad overeager after my vouch request was approved and I was quite rushed and ended up creating this pull request on my phone during lunch and had several errands afterwards. 

AI Usage Disclaimer:
GPT 5.6 Sol was used to understand the project structure. Specifically, I basically asked it to look at the codebase and tell me what files/functions were relevant to how Ghostty applies synthetic fonts. From there, I simply used ZLS to jump definitions until I found what I thought was the right spot.

### DIFF
--- a/src/font/discovery.zig
+++ b/src/font/discovery.zig
@@ -316,6 +316,61 @@ pub const Fontconfig = struct {
         pub fn next(self: *DiscoverIterator) fontconfig.Error!?DeferredFace {
             if (self.i >= self.fonts.len) return null;
 
+            slant: {
+                const roman: i32 =
+                    @intFromEnum(fontconfig.Slant.roman);
+                const italic: i32 =
+                    @intFromEnum(fontconfig.Slant.italic);
+                const oblique: i32 =
+                    @intFromEnum(fontconfig.Slant.oblique);
+
+                const requested: i32 = switch (self.pattern.get(.slant, 0) catch
+                    fontconfig.Value{ .integer = roman }) {
+                    .integer => |value| value,
+                    else => roman,
+                };
+
+                const matched: i32 = switch (self.fonts[self.i].get(.slant, 0) catch
+                    fontconfig.Value{ .integer = roman }) {
+                    .integer => |value| value,
+                    else => roman,
+                };
+
+                const requested_slanted: bool =
+                    requested == italic or requested == oblique;
+                const matched_slanted: bool =
+                    matched == italic or matched == oblique;
+
+                if (requested_slanted and !matched_slanted) return null;
+
+                break :slant;
+            }
+
+            weight: {
+                const regular: i32 =
+                    @intFromEnum(fontconfig.Weight.regular);
+                const bold_min: i32 =
+                    @intFromEnum(fontconfig.Weight.demibold);
+
+                const requested: i32 = switch (self.pattern.get(.weight, 0) catch
+                    fontconfig.Value{ .integer = regular }) {
+                    .integer => |value| value,
+                    else => regular,
+                };
+
+                const mismatch: bool = requested >= bold_min and
+                    switch (self.fonts[self.i].get(.weight, 0) catch
+                        fontconfig.Value{ .integer = regular }) {
+                        .integer => |value| value < bold_min,
+                        .range => false, // go with fontconfig's result
+                        else => true,
+                    };
+
+                if (mismatch) return null;
+
+                break :weight;
+            }
+
             // Get the copied pattern from our fontset that has the
             // attributes configured for rendering.
             const font_pattern = try self.config.fontRenderPrepare(


mitchellh (changes requested):
This code quality has major issues. It looks like undisclosed AI to me. If its AI, then I suspect you can learn yourself how to fix it.

UnsaltedScholar:
@mitchellh 
I apologize for forgetting to add the AI usage declaration (see edit in initial comment), but unfortunately the quality issues are a result of being a physicist who has learned to code from other physicists, not AI in this case (hopefully this didn't come across as sarcastic). The 22 lines I have added are all mine; I thought I was doing the correct thing by keeping my changes as small as possible (a habit I have learned from trying not to break my lab's 15+ year old c/c++ analysis software). 

I don't know the Ghostty team's internal design decisions beyond the comments provided in the code currently, so I wanted to keep my footprint minimal (or at least what I thought was minimal; granted, in retrospect, putting the changes where they are now probably wasn't the best idea).

I understand that it is not anyone's responsibility to teach me to code here. However, more specific guidance about your objection, whether it's my code placement, the "style" I used, or something else, would help me correct the changes. I'd rather not try to guess at the project's intended design constraints.

This all being said, I would appreciate some clarification on the following questions before I attempt to implement anything more robust:
1. Should Ghostty search the remaining descriptors if Fontconfig's first candidate for bold/italic is regular?
2. Is demibold a close enough match to a bold request (e.g. if, for some reason, a font has demibold but not bold as a named weight)?
3. Is Ghostty fine with allowing native oblique for italic requests (when native italic is missing)
4. For edge cases (e.g. missing properties), should Ghostty just accept what Fontconfig gives (i.e. how pedantic should I be in checking things fontconfig may throw)?

I really like Ghostty, and would still like to address this issue correctly, but I currently lack a full idea of the expected behavior.

mitchellh:
> The 22 lines I have added are all mine

If this the case, I will dedicate the time to perform a helpful review. When its AI I don't because the person on the other side often doesn't reciprocate, but in this case I shall. Stand by.

src/font/discovery.zig
```diff
@@ -316,6 +316,28 @@ pub const Fontconfig = struct {
         pub fn next(self: *DiscoverIterator) fontconfig.Error!?DeferredFace {
             if (self.i >= self.fonts.len) return null;
 
+            const requested_slant: fontconfig.Slant =
+                @enumFromInt((try self.pattern.get(.slant, 0)).integer);
```
mitchellh: enumFromInt can trigger runtime safety, its better here to use `std.meta.enumFromInt` (I think thats what its called) and handle the possibility that an unknown value exists. Same for matched_slant below.

src/font/discovery.zig
```diff
@@ -316,6 +316,28 @@ pub const Fontconfig = struct {
...
+            const matched_slant: fontconfig.Slant =
+                @enumFromInt((try self.fonts[self.i].get(.slant, 0)).integer);
+
+            const requested_weight =
```
mitchellh: I'd prefer to have an explicit type here `: T`

src/font/discovery.zig
```diff
@@ -316,6 +316,28 @@ pub const Fontconfig = struct {
...
+                (try self.pattern.get(.weight, 0)).integer;
+            const matched_weight = try self.fonts[self.i].get(.weight, 0);
+            const bold_min = @intFromEnum(fontconfig.Weight.demibold);
+            const weight_mismatch = requested_weight >= bold_min and
```
mitchellh: I would use a named block here like `weight: { }` and put requested_weight, matched_weight, and bold_min in that block (since its scoped, you can simplify var names aesthetically and just do `requested` and `matched`). Since these are just intermediaries used to produce this final value, it cleanly designates them as such.

src/font/discovery.zig
```diff
@@ -316,6 +316,28 @@ pub const Fontconfig = struct {
...
+                switch (matched_weight) {
+                    .integer => |weight| weight < bold_min,
+                    .range => false,
+                    else => true,
+                };
+
+            if ((requested_slant != .roman and matched_slant == .roman) or
```
mitchellh: You can move the conditional matching the slants up above the weight_mismatch calculation. Since if they fire first, calculating the weight mismatch is a waste of cycles.

mitchellh (changes requested):

UnsaltedScholar:
After more trial and error than anticipated, I managed to clean up the implementation following your guidance (even managed to remove the need for converting from int to enum entirely). I couldn't find any way to make it more compact and still handle all(?) the edge cases. 

Thank you for taking the time to provide comments on my initial implementation. I'm sorry it was subpar, but I shall keep your suggestions in mind for the future. 

Some notes:
- `bold_min` is somewhat arbitrary, but I chose demibold because it is the lightest weight that contains "bold" in the name
- italic and oblique are both accepted native fonts when italics are requested (only really applies if a font for some reason has oblique but not italics, which would be weird, but I guess possible)

PR #13610, opened 2026-08-04
macOS: show/search description when subtitle missing in CommandPalette

bo2themax:
image


clason:
Quick follow-up suggestion: Titles should always take precedence over description when filtering. (Typing `u` will now focus "Change Tab Title" instead of "Check for Updates", like before.)

bo2themax:
> Quick follow-up suggestion: Titles should always take precedence over description when filtering. (Typing `u` will now focus "Change Tab Title" instead of "Check for Updates", like before.)

Yeah, make sense! Thanks

Issue #2630, opened 2024-11-08
Move tab to a new window


mitchellh:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/2625

<div type='discussions-op-text'>

<sup>Originally posted by **jcollie** November  8, 2024</sup>
Currently (with libadwaita enabled) you can move a tab to another existing window by dragging the tab header from one window to another. I propose adding a keybind action and corresponding context menu entries to move a tab to a new (previously non-existent) window where it would become the first and only tab in that window.</div>

### Design

1. Add `move_tab_to_new_window` binding
2. Add the associated apprt action
3. Implement for GTK and/or macOS
We can support both platforms for this one.

kvz:
> Currently (with libadwaita enabled) you can move a tab to another existing window by dragging the tab header from one window to another

Just adding a datapoint, I could not get this to work on Omarchy/Hyprland. The tab will just immediately snap back to its original window after dragging it to another.

gwuen:
> Just adding a datapoint, I could not get this to work on Omarchy/Hyprland. The tab will just immediately snap back to its original window after dragging it to another.

Moving tabs only seems works between windows of the same instance. Depending on the way you start Ghostty, you might create new instances without realizing. This behavior is configurable with the `gtk-single-instance` option, with the default value being `detect`:

> If detect, Ghostty will assume true (single instance) unless one of the following scenarios is found:
>
> 1. TERM_PROGRAM environment variable is a non-empty value. In this case, we assume Ghostty is being launched from a graphical terminal session and you want a dedicated instance.
>
> 2. Any CLI arguments exist. In this case, we assume you are passing custom Ghostty configuration. Single instance mode inherits the configuration from when it was launched, so we must disable single instance to load the new configuration.

Moving a single tab to a *new window* should be unaffected though. This seems to be fiddly in general, but dragging it to the edge of the screen generally works for me.

jcollie:
> > Just adding a datapoint, I could not get this to work on Omarchy/Hyprland. The tab will just immediately snap back to its original window after dragging it to another.
> 
> Moving tabs only seems works between windows of the same instance. Depending on the way you start Ghostty, you might create new instances without realizing. This behavior is configurable with the `gtk-single-instance` option, with the default value being `detect`:

[...]

PR #13621, opened 2026-08-05
core,gtk: add move_tab_to_new_window action

pluiedev:
Implements most of #2630

This is in reality a really simple change and ideally we can get this out before the 1.4 string freeze


src/apprt/action.zig
```diff
@@ -351,6 +351,9 @@ pub const Action = union(Key) {
     /// otherwise the terminal-set title.
     copy_title_to_clipboard,
 
+    /// Move a tab to a new window.
```
bo2themax, line 354:
```suggestion
    /// Move a tab to a new window.
    ///
    /// Only implemented on Linux, but there's a native tab menu provided by macOS.
```

bo2themax (approved):
I'll put the same suggestions for another GTK action. 

Also, [swift side's suggestion](https://github.com/ghostty-org/ghostty/pull/10999/changes/91f02c03deb665866c3b08db53f57b5ddc8e93be#diff-3e22b9a043d8e220a6dab6c7ec0fa6155ab905e4e3d0ba534f295433547a2698R29)

src/apprt/action.zig
pluiedev: That's the wrong place to put this comment (this is internal, in `apprt`) but I'll add this to the user-facing side as well

00-kat (approved):

src/apprt/gtk/class/window.zig
```diff
@@ -604,6 +604,26 @@ pub const Window = extern struct {
         return tab_view.reorderPage(page, desired_pos) != 0;
     }
 
+    pub fn moveTabToNewWindow(self: *Self, surface: *Surface) bool {
+        const tab_view = self.private().tab_view;
+        const tab = ext.getAncestor(
+            Tab,
+            surface.as(gtk.Widget),
+        ) orelse return false;
+        const page = tab_view.getPage(tab.as(gtk.Widget));
```
jcollie, line 613: We should check to see if this is the only tab in the window and skip moving to a new window.

jcollie (approved):

jcollie (approved):

mitchellh (approved):

PR #13627, opened 2026-08-05
terminal/kitty: evict without scratch allocation

jparise:
Track each image's placement count in its existing metadata. This lets us use constant-time usage checks (rather than scans) during eviction.

Select the best candidate directly from storage on each eviction, preserving the existing priority order: unused status, transient hint, generation, then ID.

Since eviction no longer allocates, it can't fail, so callers no longer need to handle out-of-memory conditions.

### DIFF
--- a/src/terminal/kitty/graphics_image.zig
+++ b/src/terminal/kitty/graphics_image.zig
@@ -99,7 +99,7 @@ pub const LoadingImage = struct {
                 .height = t.height,
                 .compression = t.compression,
                 .format = t.format,
-                .usage = t.usage,
+                .metadata = .{ .transient = t.usage.transient },
             },
 
             .display = cmd.display(),
@@ -600,7 +600,18 @@ pub const Image = struct {
     format: command.Transmission.Format = .rgb,
     compression: command.Transmission.Compression = .none,
     data: Data = .{ .complete = "" },
-    usage: command.Transmission.Usage = .default,
+    metadata: packed struct(u32) {
+        /// The image's transient usage hint, used to prioritize eviction.
+        transient: bool = false,
+
+        /// Set this if the image was loaded without an ID or number. Such
+        /// images must not receive responses even though they currently get
+        /// IDs in the public range (which is bad!).
+        implicit_id: bool = false,
+
+        /// Number of placements referencing this image.
+        placement_count: u30 = 0,
+    } = .{},
 
     /// Unique, monotonically increasing stamp assigned each time an
     /// image is added to (or replaced in) an ImageStorage. A changed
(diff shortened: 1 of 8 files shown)


src/terminal/kitty/graphics_storage.zig
```diff
@@ -675,128 +689,73 @@ pub const ImageStorage = struct {
...
+                while (it.next()) |kv| {
+                    const img = kv.value_ptr;
+                    if (exclude_id != null and img.id == exclude_id.?) continue;
+
+                    const current: Candidate = .{
+                        .id = img.id,
+                        .generation = img.generation,
+                        .priority = (if (img.metadata.transient) @as(u2, 0) else @as(u2, 1)) +
+                            (if (img.metadata.placement_count > 0) @as(u2, 2) else @as(u2, 0)),
+                    };
```
jcollie: I think that this logic would be good to move to an `.init` on `Candidate`.

jcollie (approved):
One small nit but otherwise this looks great!

src/terminal/kitty/graphics_storage.zig
jparise: I actually had it that way for a bit and inlined it on my final review pass given it was the only caller. 🙂

I'm good with moving it back, especially because it moves the `priority` math closer to the comment that documents the encoding.

mitchellh:
Sorry I just committed a bunch of kitty fixes, can you rebase.

jparise:
> Sorry I just committed a bunch of kitty fixes, can you rebase.

No worries. Wasn't too hard of a merge.

src/terminal/kitty/graphics_image.zig
```diff
@@ -600,7 +600,18 @@ pub const Image = struct {
...
+        /// The image's transient usage hint, used to prioritize eviction.
+        transient: bool = false,
+
+        /// Set this if the image was loaded without an ID or number. Such
+        /// images must not receive responses even though they currently get
+        /// IDs in the public range (which is bad!).
+        implicit_id: bool = false,
+
+        /// Number of placements referencing this image.
+        placement_count: u30 = 0,
```
mitchellh, line 613: Is it theoretically possible for placement count to exceed 2^30? Does anything limit this?

mitchellh (changes requested):
A question

src/terminal/kitty/graphics_image.zig
jparise: I guard against that case with an `OutOfMemory` error result:

```zig
            img.metadata.placement_count = std.math.add(
                @TypeOf(img.metadata.placement_count),
                img.metadata.placement_count,
                1,
            ) catch {
                self.placements.removeByPtr(gop.key_ptr);
                return error.OutOfMemory;
            };
```

mitchellh: Ah there it is, I missed that. That's great.

mitchellh:
Okay, my review looks good. Doing one more (well two) through 5.6 and Fable and will report back.

mitchellh:
The machines bless you

jparise:
> The machines bless you

Praise be to the machine

PR #13634, opened 2026-08-05
macos: handled untrusted OSC8 hyperlinks more carefully

mitchellh:
OSC8 hyperlinks previously executed directly via the NSWorkspace opener so a malicious application can just do whatever it wanted and trick the user into opening something through Launch Services.

This PR notifies apprt of OSC8 hyperlinks so they can be handled specially. In this PR, I added macOS-specific handling of OSC8 through a variety of improvements:

  - Preview text is sanitized, so invisible Unicode characters now show.
  - Questionable-looking URLs require confirmation to open, but a user can confirm to open.
  - Very questionable or definitely unsafe URLs are blocked with an  alert that only allows the user to copy the link. The alert also notifies the user why.

This PR also adds an explicit `link-osc8` config (default true) that users can use to disable osc8 completely.

### Custom URL Schemes (Confirm)

CleanShot 2026-08-05 at 10 25 57@2x

### Invisible Characters (Block)

CleanShot 2026-08-05 at 10 26 44@2x

### Executable Target (Block)

CleanShot 2026-08-05 at 10 27 31@2x

### DIFF
--- a/macos/Sources/Helpers/UntrustedURL.swift
+++ b/macos/Sources/Helpers/UntrustedURL.swift
@@ -1,0 +1,262 @@
+import Foundation
+import UniformTypeIdentifiers
+
+/// A URL supplied by a source that should not be allowed to dispatch directly
+/// to the operating system, such as terminal output.
+struct UntrustedURL: Equatable {
+    enum DenialReason: Equatable {
+        case malformedURL
+        case unsafeCharacters
+        case invalidWebURL
+        case inaccessibleFile
+        case unsafeFile
+
+        var message: String {
+            switch self {
+            case .malformedURL:
+                "The target is not an absolute URL with a scheme."
+            case .unsafeCharacters:
+                "The target contains invisible or line-breaking characters."
+            case .invalidWebURL:
+                "The web target does not contain a valid host."
+            case .inaccessibleFile:
+                "The local target does not exist or is not a regular file or directory."
+            case .unsafeFile:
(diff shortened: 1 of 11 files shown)


macos/Sources/Helpers/UntrustedURL.swift
```diff
@@ -0,0 +1,262 @@
+import Foundation
+import UniformTypeIdentifiers
+
+/// A URL supplied by a source that should not be allowed to dispatch directly
+/// to the operating system, such as terminal output.
+struct UntrustedURL: Equatable {
+    enum DenialReason: Equatable {
```
jparise, line 7: This could also be shaped as `enum DenialError: Error, LocalizedError`, which is a little more expressive. What's here works well, too.

bo2themax (approved):

macos/Sources/Helpers/UntrustedURL.swift
bo2themax: catching then switch would be a pain tho🫪

tristan957:
Should we do something similar in the GTK apprt?

mitchellh:
Yeah, sorry from Discord:

> i think so, we shouldn't drop down to xdg-open for everything
i put the classification in the macos apprt rather than core though because we can do some macos specific stuff like query what app would open it
i think for GTK we can port the invisible unicode char detection + executable bit stat-ing + stat-failure-requires-confirmation

PR #13639, opened 2026-08-05
macos: tolerate display link creation failures

mitchellh:
Fixes #13293

Treat Core Video display link creation as optional when macOS has no active displays. The previous error path reported every creation failure as out of memory and aborted renderer initialization.

Tabs created while the session is locked now initialize normally and fall back to event-driven rendering without vsync.


mitchellh:
Found a bug here where we never restore the vsync in this path. I think we should probably grab the display link when our display ID changes, so doing that now.

PR #13654, opened 2026-08-05
core: avoid allocating for pwd change actions

jparise:
Surface.handleMessage allocated a null-terminated copy for every working directory update. OSC 7 values fit within the parser's 2 KiB fixed buffer, so use stack-fallback storage sized for that bound and its terminator.

The message type does not enforce the OSC bound, so an oversized future producer still falls back to the heap. performAction already borrows the value only for the duration of the call, preserving its existing lifetime.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -1073,9 +1073,10 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
         .pwd_change => |w| {
             defer w.deinit();
 
-            // We always allocate for this because we need to null-terminate.
-            const str = try self.alloc.dupeZ(u8, w.slice());
-            defer self.alloc.free(str);
+            var stack = std.heap.stackFallback(256, self.alloc);
+            const alloc = stack.get();
+            const str = try alloc.dupeZ(u8, w.slice());
+            defer alloc.free(str);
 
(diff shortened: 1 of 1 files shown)


src/Surface.zig
```diff
@@ -1073,9 +1073,13 @@ pub fn handleMessage(self: *Surface, msg: Message) !void {
         .pwd_change => |w| {
             defer w.deinit();
 
-            // We always allocate for this because we need to null-terminate.
-            const str = try self.alloc.dupeZ(u8, w.slice());
-            defer self.alloc.free(str);
+            var stack = std.heap.stackFallback(
+                terminal.osc.Parser.MAX_BUF + 1,
```
mitchellh: I would use a smaller number here that is more aligned with safety. Since we continue to call functions below the stack continues to get deep. If we max our stack here, we can screw apprt's who themselves may make a deep stack. So, we should be safe.

mitchellh (changes requested):

mitchellh (approved):
Aye

src/Surface.zig
jparise: Makes sense given we have the heap fallback, too. I originally did this all on the stack but didn't want to assume anything about the OSC parser's constants.

PR #13672, opened 2026-08-06
apprt: move command palette support check into zig

bo2themax:
Raised during conversation in #10999.

* Added two command files for macOS and GTK to check whether an action is supported. So new action added in the future will have to update these two to avoid drifting.

* Fixes the entries both lists were missing:
  * macOS: `show_on_screen_keyboard`, `move_tab_to_new_window`
  * GTK: `toggle_background_opacity`, `toggle_visibility`

## AI Disclosure

Claude implemented this, i reviewed and **tested on macOS**, gtk part looks simple enough.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -416,11 +416,13 @@ typedef struct {
   ghostty_input_mods_e mods;
 } ghostty_input_trigger_s;
 
+// input.Command
 typedef struct {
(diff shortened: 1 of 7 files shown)


tristan957 (approved):
GTK tree looks good.

jcollie (approved):

PR #13673, opened 2026-08-06
input: update toggle_maximize documentation

bo2themax:
## AI Closure

Claude found and changed it

### DIFF
--- a/src/input/Binding.zig
+++ b/src/input/Binding.zig
@@ -747,7 +747,7 @@ pub const Action = union(enum) {
 
     /// Maximize or unmaximize the current window.
     ///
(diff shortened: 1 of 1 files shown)


src/input/Binding.zig
```diff
@@ -747,8 +747,8 @@ pub const Action = union(enum) {
 
     /// Maximize or unmaximize the current window.
     ///
-    /// This has no effect on macOS as it does not have the concept of
-    /// maximized windows.
+    /// On macOS this zooms the window, which is the closest equivalent
```
tristan957:
```suggestion
    /// On macOS, this zooms the window, which is the closest equivalent
```

jcollie (approved):

Issue #12825, opened 2026-05-27
macOS Tahoe(26.5) switching tabs cause unexpected tabs flashing


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/12818

<div type='discussions-op-text'>

<sup>Originally posted by **Tunglies** May 26, 2026</sup>
### Issue Description

When switching tabs, tabs flashing and jumping outliners as unexpected.

### Expected Behavior

Switching as normally

### Actual Behavior

https://github.com/user-attachments/assets/a6060dc6-278f-400d-9b57-0e662519664e

### Reproduction Steps

1. Reload Ghostty with the config below
2. Open Ghostty with 2 tabs
3. Run `printf '\033]11;rgb:1e/1e/2e\007'` to change the background temporarily
4. Switch tabs

### Ghostty Logs

```text
No need
```

### Ghostty Version

```text
Ghostty 1.3.1

Version
  - version: 1.3.1
  - channel: stable
Build Config
  - Zig version   : 0.15.2
  - build mode    : .ReleaseFast
  - app runtime   : .none
  - font engine   : .coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue
```

### OS Version Information

macOS Tahoe(26.5)

### (Linux only) Display Server

None

### Minimal Ghostty Configuration

```ini
macos-titlebar-style = tabs
theme = Xcode Light
```

### I acknowledge that:

- [X] I have reviewed the FAQ and confirm that my issue is NOT among them.
- [X] I have searched the Ghostty repository (both open and closed Discussions and Issues) and confirm this is not a duplicate of an existing issue or discussion.
- [X] I have checked the "Preview" tab on all text fields to ensure that everything looks right, and have wrapped all configuration and code in code blocks with a group of three backticks (` ``` `) on separate lines.</div>

PR #13675, opened 2026-08-06
macos: sync appearance when new windows are created

### LINKED ISSUE #13324: macOS: background-opacity not applied properly
Related discussions: https://github.com/ghostty-org/ghostty/discussions/13322 and https://github.com/ghostty-org/ghostty/discussions/13320

It seems to be related to the same rc, so I created in one issue. If this is determined otherwise, one can mention them in separate commits.

This is a regression caused by some pr, 1.3.1 works fine

Reproduce steps:

1. Reload Ghostty with the following config
2. Close all existing window if there's any
3. Open a new window and observe how the background is not set properly

```ini
background-opacity = 0.7
macos-titlebar-style = hidden # window-decoration = false
# background-blur = macos-glass-regular # same happens with liquid glass
```

zenangst:
For new windows to get their appearance synced, we need to call `syncAppearance` after `super.showWindow(sender)`. All previous calls to `syncAppearance` on `TerminalWindow` will be ignored because the window needs to have `isVisible` set to `true`.

This regression was introduced by: 5368adcd29754939e6c283198ef6b1c122293815 

It added `.dropFirst()` to the `focusedSurface` appearance publishers in `TerminalController.swift` which removes the initial call of the subscription.

Fixes https://github.com/ghostty-org/ghostty/issues/13324

(landed on the same fix as @rasitakyol found here: https://github.com/ghostty-org/ghostty/pull/13341)

### DIFF
--- a/macos/Sources/Features/Terminal/TerminalController.swift
+++ b/macos/Sources/Features/Terminal/TerminalController.swift
@@ -1159,6 +1159,8 @@ class TerminalController: BaseTerminalController, TabGroupCloseCoordinator.Contr
         }
 
         super.showWindow(sender)
+
+        syncAppearance()
     }
 
     // Shows the "+" button in the tab bar, responds to that click.


bo2themax (approved):

bo2themax:
~~Changes look good, but I haven't run it myself yet.~~ LGTM

zenangst:
https://github.com/user-attachments/assets/18dba1c0-ef1f-4aa5-95f8-b48d574f9213

I tried switching tabs to check that #12825 didn't resurface.

zenangst:
Here is a video showing it using `NSSolariumWindowTabs` set to `true`:

https://github.com/user-attachments/assets/d837b8de-443a-495b-ad97-6d90dff8faff

zenangst:
Note that `showWindow` is not invoked when the user switches tabs, so in theory, it should not cause issues for the other fix.

jcollie:
!vouch

PR #13677, opened 2026-08-06
build: fix flatpak/snap, restore rpath opt, fix local gtk4-layer-shell

vancluever:
This fixes regressions in the flatpak/snap builds, and knock-on stuff that was discovered as as a result:

* Update the Zig versions in the flatpak/snap build configuration files.
* Restore the classic `-Dpatch-rpath` option, and add a new `-Dpatch-interp` option. This ensures that the snap can still use `-Dpatch-rpath` correctly.
* There seems to be an issue in Zig when parsing IPv6 addresses that leads to issues loading `resolv.conf` files; when trying to load a nameserver that has an IPv6 address with a numeric interface index as the scoped zone ID, Zig will try to resolve the interface as a name rather than just use the index. This is coming up in snap builds because the build process seems to, by default, use the exhaustive `/run/systemd/resolve/resolv.conf` file, versus the simpler stub (`stub-resolv.conf`) file. We work around this for the time being by linking the stub at the end of the Zig part, overwriting the link to the non-stub file. 
* Fixed `gtk4-layer-shell` packaging - the migration to external translate-c meant that non-system builds of the dependency were not handing the local `gtk4-layer-shell` headers over for translation. Now, instead, we've extracted the management of the `gtk4-layer-shell` source and `wayland-protocols` generation to a locally-cached object so that the source can be shared by both C translation and the library build in a way that is not coupled to any particular step.

### DIFF
--- a/flatpak/dependencies.yml
+++ b/flatpak/dependencies.yml
@@ -13,12 +13,12 @@ modules:
       - chmod a+x /app/zig/zig
     sources:
       - type: archive
-        sha256: 02aa270f183da276e5b5920b1dac44a63f1a49e55050ebde3aecc9eb82f93239
-        url: https://ziglang.org/download/0.15.2/zig-x86_64-linux-0.15.2.tar.xz
+        sha256: 70e49664a74374b48b51e6f3fdfbf437f6395d42509050588bd49abe52ba3d00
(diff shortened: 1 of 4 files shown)


mitchellh (approved):
This all looks fine to me, but want to see CI run of course.

kenvandine:
The snap changes look good, thanks!

PR #13692, opened 2026-08-08
gtk/wayland: clean up blur regions

gotenksIN:
Fix two resource leaks when the Wayland background blur region changes.

Temporary `wl_region` objects were destroyed only after an error. They are now always destroyed after the blur request. The previous cached blur region is also released before its replacement.

This does not change blur protocol selection. Ghostty continues to use `ext-background-effect-v1` before GTK 4.23.3 and defers blur handling to GTK 4.23.3 and later.

Testing:

- Verified blur on KDE Plasma 6.7.4 with GTK 4.22.4.
- Verified with `zig build -Doptimize=ReleaseFast`.

This change was developed with AI assistance and manually reviewed and tested.

### DIFF
--- a/src/apprt/gtk/winproto/wayland.zig
+++ b/src/apprt/gtk/winproto/wayland.zig
@@ -112,6 +112,11 @@ pub const App = struct {
 
 /// Per-window (wl_surface) state for the Wayland protocol.
 pub const Window = struct {
+    pub const BlurState = union(enum) {
+        ext: *ext.BackgroundEffectSurfaceV1,
+        kde: ?*org.KdeKwinBlur,
+    };
+
     apprt_window: *ApprtWindow,
 
     /// The Wayland surface for this window.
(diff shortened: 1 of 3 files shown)


tristan957:
@pluiedev wanna see if this is something we should take a look at?

pluiedev:
No, I just said that we don't want to maintain the KDE blur API unless absolutely necessary

gotenksIN:
The smaller part of the commit that actually fixed blur on plasma 6.7/gtk 4.22.4 with ext protocol is not reflected in this PR because it was closed, but it is available on the branch this PR is tracking, https://github.com/gotenksIN/ghostty/commit/454c5966eb2954ae46be9ecff38c02341fab60e9

pluiedev:
Oh, open that as a separate PR then.

gotenksIN:
https://github.com/ghostty-org/ghostty/pull/13699 new PR has been opened for it

PR #13706, opened 2026-08-09
core,gtk: add `drag-handle` config

pluiedev:
Fixes hundreds of complaints about the fact that drag handles cannot be hidden, on GTK at least.

I'm not sure if we ever made an issue for this? If you come across any discussions asking for this, please link them here :)

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -2345,6 +2345,30 @@ keybind: Keybinds = .{},
 /// Specified as either hex (`#RRGGBB` or `RRGGBB`) or a named X11 color.
 @"window-titlebar-foreground": ?Color = null,
 
+/// Controls when drag handles are shown over splits,
+/// allowing splits to be rearranged with mouse controls.
+///
+/// Valid values:
+///
+///  - `always`
+///
+///    Always display the drag handle, even when there's only one split.
+///
+///  - `auto` *(default)*
+///
+///    Automatically show and hide the drag handle. The handle is only
+///    shown when there are two or more splits present.
+///
+///  - `never`
+///
+///    Never show the drag handle. Splits then cannot be rearranged with
+///    mouse controls.
+///
+/// Available since: 1.4.0.
+///
+/// Currently only supported on Linux (GTK).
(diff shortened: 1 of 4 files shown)


src/config/Config.zig
```diff
@@ -2345,6 +2345,30 @@ keybind: Keybinds = .{},
...
+///
+/// Valid values:
+///
+///  - `always`
+///
+///    Always display the drag handle, even when there's only one split.
+///
+///  - `auto` *(default)*
+///
+///    Automatically show and hide the drag handle. The tab bar is only
```
jcollie: `tab bar` => `drag handle`

src/config/Config.zig
```diff
@@ -2345,6 +2345,30 @@ keybind: Keybinds = .{},
...
+///    shown when there are two or more tabs present.
```
jcollie: `tabs` => `splits`

jcollie (approved):
Couple of small changes, otherwise LGTM.

src/config/Config.zig
```diff
@@ -2345,6 +2345,30 @@ keybind: Keybinds = .{},
...
+///
+///  - `auto` *(default)*
+///
+///    Automatically show and hide the drag handle. The handle is only
+///    shown when there are two or more splits present.
+///
+///  - `never`
+///
+///    Never show the drag handle. Splits then cannot be rearranged with
+///    mouse controls.
```
bo2themax, line 2365: Why also disable mouse controls tho, I think it's easier to maintain to just hide/show the handle🫪

pluiedev: It's more like this: the drag handles are hidden -> therefore there are no mouse-driven ways to move splits. Maybe it's a bit unclearly written, do let me know

bo2themax: Ah ok, macOS handles it a bit differently, the drag was actually handled by another view, 3 dots just responds to the hover event.

I ran it locally tho, it's rather easy to disable the drag as well on macOS, so we can follow as the comment says

bo2themax (approved):

PR #13717, opened 2026-08-09
macOS: rework for #12712 and #13645

bo2themax:
`needleSelection` was introduced in #12712 to select all texts when syncing pasteboard, the crash happens most on macOS 15 in `readPasteboardNeedle`. It seems that `objectWillChange` fires differently there, and it's hard to reproduce on macOS 26/27. I think guaranteeing from ourside is enough, I believe SwiftUI already as its own when updating the binding.

**Confirmed with a simple example on macOS 15, it seems a SwiftUI issue🫪. So I changed the minimal macOS version for text selection to macOS 26. I don't see an elegant way to fix it.**

image

```swift
struct ContentView: View {
    @State private var text = ""
    @State private var selection: TextSelection?
    var body: some View {
        TextField("Search", text: $text, selection: $selection)
    }
}
```

### DIFF
--- a/macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
+++ b/macos/Sources/Ghostty/Surface View/OSSurfaceView.swift
@@ -117,17 +117,24 @@ extension Ghostty {
 
 extension Ghostty.OSSurfaceView {
     @MainActor class SearchState: ObservableObject {
+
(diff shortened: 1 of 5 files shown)


mitchellh:
I like this more than mine: https://github.com/ghostty-org/ghostty/pull/13716 Closed mine.

PR #13727, opened 2026-08-10
gtk: free hotkeys memory on app teardown

dkinzler:
In debug builds the DebugAllocator throws an error about leaked memory when you close Ghostty, if you have global keybinds in your config with a Wayland compositor that supports the vicinae-hotkey protocol. The cause is the `Hotkeys.entries` array list never actually being freed. Not really a problem because the list should be kept around until app teardown anyway, but not getting an error every time would be nice (even if you need a somewhat specific setup for this to even happen right now).

To fix this free the array list memory in Hotkeys.deinit with `ArrayList.clearAndFree`. As the existing comment on `deinit` already mentions, we can't use `ArrayList.deinit` because it leaves the list in an invalid state and `Hotkeys.clear` might still get called and use it.


jcollie (approved):
Approved, pending CI

pluiedev (approved):

PR #13729, opened 2026-08-10
terminal/kitty: use fixed table for control keys

Uzaaft:
Swap out the hash map backed by an arena with a fixed array.
Measurements from poop:
CPU Cycles: 173M → 171M (−1.2%)
Instructions: 519M → 517M (−0.5%)
Peak RSS: 11.9 → 11.8 MB
Cache misses: 356K → 312K (−12.4%)

### DIFF
--- a/src/terminal/kitty/graphics_command.zig
+++ b/src/terminal/kitty/graphics_command.zig
@@ -12,8 +12,22 @@ const log = std.log.scoped(.kitty_gfx);
 /// ignored, matching how unknown letter keys are ignored when building a
 /// command.
 ///
-/// For the value of this: if the value is a single printable ASCII character
-/// it is the ASCII code. Otherwise, it is parsed as a 32-bit unsigned integer.
+/// This is deliberately a dense value table plus a presence bitmap. There are
+/// only 52 possible keys(technically KIP has fever than 52 possible .. at the moment of writing), so a lookup is one machine-word bit test followed by
+/// an indexed load. The values are undefined until their presence bit is set.
+///
+/// Other alternatives were considered and should probably be revisited in the
+/// future. Some of them were:
+/// A `[52]?u32` table is 416 bytes versus 216 bytes for this layout, and
+/// its larger copies erased the smaller lookup code.
+/// `std.EnumMap` added optional-return lowering,
+/// `std.StaticBitSet(52)` kept the same size but generated larger code
+/// for the packed mask. Passing the table by pointer removed the copies
+/// but did not improve end-to-end throughput. Keep this representation
+/// unless a new benchmark shows an improvement in the zig compilers codegen.
+///
+/// For the value itself: if it is a single printable ASCII character it is the
+/// ASCII code. Otherwise, it is parsed as a 32-bit unsigned integer.
 const KV = struct {
     values: [52]u32 = undefined,
     present: u64 = 0,
--- a/src/terminal/kitty/graphics_command.zig
+++ b/src/terminal/kitty/graphics_command.zig
@@ -1,25 +1,47 @@
 const std = @import("std");
 const assert = @import("../../quirks.zig").inlineAssert;
 const Allocator = std.mem.Allocator;
-const ArenaAllocator = std.heap.ArenaAllocator;
 const simd = @import("../../simd/main.zig");
 const lib = @import("../lib.zig");
 
 const log = std.log.scoped(.kitty_gfx);
 
 /// The key-value pairs for the control information for a command. The
-/// keys are always single characters and the values are either single
-/// characters or 32-bit unsigned integers.
+/// recognized keys are always single ASCII letters and the values are either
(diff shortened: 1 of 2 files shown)


src/terminal/kitty/graphics_command.zig
```diff
@@ -1,25 +1,47 @@
...
+/// single characters or 32-bit unsigned integers. Unknown non-letter keys are
+/// ignored, matching how unknown letter keys are ignored when building a
+/// command.
 ///
 /// For the value of this: if the value is a single printable ASCII character
 /// it is the ASCII code. Otherwise, it is parsed as a 32-bit unsigned integer.
-const KV = std.AutoHashMapUnmanaged(u8, u32);
+const KV = struct {
+    values: [52]u32 = undefined,
+    present: u64 = 0,
```
jcollie, line 33: Did you try using a `?u32` so that you can use Zig nulls to determine if a value is present or not? Storage size should be the same as a `u64` and the intent is clearer I think.

Uzaaft:
Two things that might be worthwhile to consider: 
kip has what, 30-ish keys. I can reduce the table from +200 bytes down to around 128, but this will probably need a uglier switch statement and more upkeep. 

We can parse directly into command field and remove the KV struct. I havent measured this so idk if it's faster or not, but lmk if this is a direction you'd be interested in exploring and I can experiment with poop.

src/terminal/kitty/graphics_command.zig
Uzaaft: I tried it with one of my experiments:
image


The struct size is bigger, that's why I dropped it. Besides the size tradeoff I saw I'm a fan!

Uzaaft:
@pluiedev made me aware of [std.EnumMap](https://ziglang.org/documentation/master/std/#std.EnumMap) so I'll see if that's a better implementation.

Uzaaft:
So std.EnumMap turns out to have worse codegen.

src/terminal/kitty/graphics_command.zig
Uzaaft: @jcollie Would you still want me to implement the nullable variant over the current one given the size tradeoff? 
If not, I'll wire up some proper docs with a note about my experimentation and the decision made such that it's documented.

jcollie: Not if the nullable variant slows things down. One other experiment would be to use a `StaticBitSet` for `present` instead of hand-rolling the bit testing.

Uzaaft: 100s run with poop:
```
   Workload / metric           Hand u64 bitmap    StaticBitSet(52)   Delta
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   APC wall time                       13.8 ms             14.5 ms   +4.7%
  ─────────────────────────────────────────────────────────────────────
   APC RSS                             8.63 MB             8.63 MB   0.0%
  ─────────────────────────────────────────────────────────────────────
   APC cycles                            40.3M               42.4M   +5.3%
  ─────────────────────────────────────────────────────────────────────
   APC instructions                       148M                149M  +0.6%
  ─────────────────────────────────────────────────────────────────────
   APC cache references                  3.72M               3.85M  +3.5%
  ─────────────────────────────────────────────────────────────────────
   APC cache misses                      36.7K               37.9K  +3.4%
  ─────────────────────────────────────────────────────────────────────
   APC branch misses                      172K                173K  +0.3%
  ─────────────────────────────────────────────────────────────────────
   Terminal wall time                  34.9 ms             34.9 ms  −0.1%
  ─────────────────────────────────────────────────────────────────────
   Terminal RSS                        12.3 MB             12.4 MB  +0.4%
  ─────────────────────────────────────────────────────────────────────
   Terminal cycles                        167M                168M  +0.3%
  ─────────────────────────────────────────────────────────────────────
   Terminal instructions                  515M                515M   0.0%
  ─────────────────────────────────────────────────────────────────────
   Terminal cache references             5.36M               5.39M  +0.6%
  ─────────────────────────────────────────────────────────────────────
   Terminal cache misses                  326K                328K  +0.8%
  ─────────────────────────────────────────────────────────────────────
   Terminal branch misses                1.17M               1.17M   0.0%
```

jcollie: Looks like hand-rolled wins!

Uzaaft: @jcollie Looks like that. 
I do think at some point, if zig gets a better codegen story this should be revisited. Writing and comparing IR and ASM for variants of the same struct but in zig/unsafe rust shows better codegen on the rust side. 

I'll write some docs!

Uzaaft: Docs with what was tried during experiments added.

jcollie (approved):

jcollie (approved):

PR #13732, opened 2026-08-10
apprt/gtk: add WeakRef.deinit and use it at teardown sites

### LINKED ISSUE #13713: apprt/gtk: give WeakRef(T) a deinit, so "clear this in dispose" is visible at the use site
### Discussed in https://github.com/ghostty-org/ghostty/discussions/13712

<sup>Originally posted by **hakonhagland** August  9, 2026</sup>
### Suggestion

`WeakRef(T)` in `src/apprt/gtk/weak_ref.zig` exposes `set` and `get`. There is no `deinit` and no `clear`, so releasing one is spelled `set(null)` — which reads like an ordinary assignment rather than a mandatory teardown step. I would like to suggest adding a `deinit` that forwards to `g_weak_ref_clear`, purely so the requirement is visible on the type someone is already using.

```zig
/// Release this reference. MUST be called before the memory holding this
/// struct is freed, usually from the owner's `dispose`. The target keeps a
/// pointer to this GWeakRef so it can clear it at finalize; if this memory
/// is gone by then, the target walks into it.
pub fn deinit(self: *Self) void {
    self.ref.clear();
}
```

Nothing upstream is broken — every current use site clears correctly — and `set(null)` would keep working. This is only about where the rule lives.

### Why

The rule is already written down, in `src/apprt/gtk/class/inspector_window.zig`:

> You MUST clear all weak refs in dispose, otherwise it causes memory corruption on dispose on the TARGET (weak referenced) object. The only way we caught this is via Valgrind. Its not a leak, its an invalid memory read. In practice, I found this sometimes caused hanging!

I hit exactly that, in a downstream branch of my own, and it cost several days. I added a `WeakRef(Window)` to a dialog and did not clear it in `dispose`. Closing a window that had ever shown the dialog then blocked the GTK main loop forever, with this stack:

```
g_pointer_bit_lock_and_get (address=0x3c369e80)
_weak_ref_lock (weak_ref=0x3c369e80)
weak_ref_data_clear_list (object=[GhosttyWindow])
_object_unref_clear_weak_locations (object=[GhosttyWindow])
g_object_unref (old_ref = 1)
```

The locked word held `0x13` — the low bit set, and pointer bits of `0x12`, which is not an address. The `GWeakRef` had been freed and the memory reused, and whatever wrote there happened to leave the lock bit set, so nothing would ever release it. It is also nondeterministic for that reason, which is why it took a symbolised stack to find rather than a reproduction.

That is my mistake, not a Ghostty bug. But the comment that would have prevented it is in a file I had no reason to open, while the type I *was* using offered `set` and `get` and no hint that teardown was mandatory. A method named `deinit` on `WeakRef` itself would have put the rule where I was looking.

### Notes

- `gobject.WeakRef.clear` is already bound, so this is an addition rather than any new dependency. `set(null)` also unregisters; `clear` is the call the GLib docs name for a `GWeakRef` that is going away.
- Zig has no destructors, so this enforces nothing — it is documentation that happens to be executable. I do not want to oversell it.
- Happy to open a PR if this seems worth having; I have not, since CONTRIBUTING.md asks for a discussion first.

---

**AI disclosure per `AI_POLICY.md`:** I investigated the underlying incident with Claude Code — the agent captured and read the stack and drafted this text, and I reviewed it.
</div>

hakonhagland:
Fixes #13713.

`WeakRef(T)` offers `set` and `get`, so releasing one is spelled `set(null)` — indistinguishable from an ordinary assignment. The requirement that it *must* happen before the owning memory is freed lives in a comment in `class/inspector_window.zig`, which is not where somebody using the type is looking.

This adds `deinit`, forwarding to `g_weak_ref_clear` — the call GLib documents for a `GWeakRef` that is going away — and switches the dispose-time clears to it.

### What changed

- `weak_ref.zig`: new `deinit`, with the reasoning in its doc comment.
- `window.zig`, `split_tree.zig`, `application.zig`, `command_palette.zig`: the four dispose-time clears now call `deinit`.

`set(null)` is unchanged and still valid. The clear in `Application.handleReloadConfig` deliberately stays a `set(null)`: the object is alive there and the reference is reused, so it is a logical clear rather than teardown — which is the distinction the new name is meant to make visible.

### Why it is worth a method

Zig has no destructors, so this enforces nothing; it is documentation that happens to be executable. The concrete case is in #13713: I added a `WeakRef(Window)` in a downstream branch, did not clear it, and closing a window that had shown that dialog deadlocked the GTK main loop inside `weak_ref_data_clear_list` locking freed memory. Every upstream call site already gets this right — the point is only to put the rule where the next person will see it.

### Testing

`zig build test` passes. `zig fmt --check` clean. Built and used on Linux/GTK; the change is behaviourally identical to what was there, since `g_weak_ref_clear` and `g_weak_ref_set(NULL)` both unregister.

---

**AI disclosure per `AI_POLICY.md`:** I investigated the underlying incident with Claude Code and it drafted this change; I reviewed it.

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -1548,7 +1548,7 @@ pub const Application = extern struct {
             diag.close();
             diag.unref(); // strong ref from get()
         }
-        priv.config_errors_dialog.set(null);
+        priv.config_errors_dialog.deinit();
         if (priv.signal_source) |v| {
             if (glib.Source.remove(v) == 0) {
                 log.warn("unable to remove signal source", .{});
--- a/src/apprt/gtk/class/command_palette.zig
+++ b/src/apprt/gtk/class/command_palette.zig
@@ -609,7 +609,7 @@ const Command = extern struct {
         switch (priv.data) {
             .regular => {},
             .jump => |*j| {
-                j.surface.set(null);
+                j.surface.deinit();
             },
         }
 
--- a/src/apprt/gtk/class/split_tree.zig
+++ b/src/apprt/gtk/class/split_tree.zig
@@ -647,7 +647,7 @@ pub const SplitTree = extern struct {
 
     fn dispose(self: *Self) callconv(.c) void {
         const priv = self.private();
-        priv.last_focused.set(null);
+        priv.last_focused.deinit();
         if (priv.rebuild_source) |v| {
             if (glib.Source.remove(v) == 0) {
                 log.warn("unable to remove rebuild source", .{});
--- a/src/apprt/gtk/class/window.zig
+++ b/src/apprt/gtk/class/window.zig
@@ -1296,7 +1296,7 @@ pub const Window = extern struct {
             priv.handle_active_state_source = null;
         }
 
-        priv.command_palette.set(null);
+        priv.command_palette.deinit();
 
         if (priv.config) |v| {
             v.unref();
--- a/src/apprt/gtk/weak_ref.zig
+++ b/src/apprt/gtk/weak_ref.zig
@@ -22,6 +22,23 @@ pub fn WeakRef(comptime T: type) type {
             }
         }
 
+        /// Release this weak reference.
+        ///
+        /// You MUST call this before the memory holding this struct is freed,
+        /// which in practice means from the owner's `dispose`. The target keeps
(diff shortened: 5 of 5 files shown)


jcollie (approved):
Approved pending CI

PR #13734, opened 2026-08-10
config: fix open_config when the config directory is a symlink

### LINKED ISSUE #13714: open_config fails when the config directory is a symlink
### Discussed in https://github.com/ghostty-org/ghostty/discussions/13698

<sup>Originally posted by **hakonhagland** August  8, 2026</sup>
### Issue Description

Pressing <kbd>ctrl</kbd>+<kbd>,</kbd> (`open_config`) does nothing when the Ghostty configuration directory is a symlink, which is what you get by keeping the config in a dotfiles repository:

```
~/.config/ghostty -> ~/dotfiles/ghostty
```

No editor opens and no error is shown in the UI. The only sign of failure is a line on stderr:

```
warning(gtk_ghostty_application): error getting config file path: error.NotDir
```

This is a regression on `main`. It works in 1.3.1.

`openPath` in `src/config/edit.zig` creates the config directory before opening it:

```zig
try std.Io.Dir.cwd().createDirPath(global.io(), config_dir);
```

When a path component already exists, `createDirPath` checks that it is a directory *without* following symlinks — deliberately, to stop a dangling symlink causing an infinite loop. On Linux the check is `statx` with `AT.SYMLINK_NOFOLLOW`, so a symlink pointing at a perfectly good directory reports kind `.sym_link`, not `.directory`, and `createDirPath` returns `error.NotDir`. `openPath` treats that as fatal.

It looks like this arrived with the Zig 0.16 upgrade in e8525c0fd, which replaced `makePath` with `createDirPath`:

```diff
-        try std.fs.cwd().makePath(config_dir);
+        try std.Io.Dir.cwd().createDirPath(global.io(), config_dir);
```

Zig 0.15.2's [`makePath`](https://github.com/ziglang/zig/blob/0.15.2/lib/std/fs/Dir.zig#L1191) used `statFile` for the same check, which *does* follow symlinks, so the symlinked directory was accepted.

Worth flagging: this looks like a transient Zig regression rather than a permanent change. There is no released Zig 0.16.0 — the newest tag is 0.15.2, and I am building with a 0.16.0 beta snapshot. On Zig master the function has been [renamed back to `makePath`](https://github.com/ziglang/zig/blob/738d2be9d6b6ef3ff3559130c05159ef53336224/lib/std/Io/Dir.zig#L320) and now uses `statPath` with [`follow_symlinks` defaulting to `true`](https://github.com/ziglang/zig/blob/738d2be9d6b6ef3ff3559130c05159ef53336224/lib/std/Io/Dir.zig#L374-L376), so it follows symlinks again as 0.15.2 did. So this may well resolve itself whenever Ghostty next bumps Zig, and you may prefer to wait rather than carry a workaround.

`writeConfigTemplate` in `src/config/Config.zig` calls `createDirPath` on the same directory and looks like it would fail the same way, though I have not reproduced that one. Happy to look if useful.

I have a fix and can open a PR if you would like one.

### Expected Behavior

<kbd>ctrl</kbd>+<kbd>,</kbd> opens the configuration file in the default editor, as it does in 1.3.1 and as it does on `main` when the config directory is a real directory rather than a symlink.

### Actual Behavior

Nothing happens. No editor, no window, no error in the UI. `error.NotDir` is logged to stderr, and `open_config` returns before it ever reaches `xdg-open`.

### Reproduction Steps

1. Move the Ghostty config directory aside and replace it with a symlink: `mv ~/.config/ghostty ~/ghostty-config-test && ln -s ~/ghostty-config-test ~/.config/ghostty`
2. Start Ghostty from a terminal so stderr is visible.
3. Press <kbd>ctrl</kbd>+<kbd>,</kbd>.
4. No editor opens; stderr shows `error getting config file path: error.NotDir`.

Restore with:

```
rm ~/.config/ghostty && mv ~/ghostty-config-test ~/.config/ghostty
```

Reproduced on a clean build of `main` at `ec58fbc6a` with no local patches, built with Zig 0.16.0 (beta snapshot, since 0.16.0 is not yet released) as `build.zig.zon` requires. It does not depend on the contents of the config file — an empty config reproduces it.

### Ghostty Logs

```text
debug(app): mailbox message=open_config
warning(gtk_ghostty_application): error getting config file path: error.NotDir
```

### Ghostty Version

```text
Ghostty 1.3.2-HEAD-+ec58fbc6a

Version
  - version: 1.3.2-HEAD-+ec58fbc6a
  - channel: tip

Build Config
  - Zig version   : 0.16.0
  - build mode    : .Debug
  - app runtime   : .gtk
  - libxev        : io_uring
  - kernel version: 7.0.0-27-generic
  - desktop env   : other
```
```


### OS Version Information

Ubuntu 26.04 LTS

Wayland

GNOME Shell 50.1

```ini
# none — reproduces with an empty configuration file
```

_No response_

### I acknowledge that:

hakonhagland:
Fixes #13714.

`open_config` does nothing when the Ghostty configuration *directory* is a symlink, which is what you get by keeping your config in a dotfiles repository. No dialog, no editor, and the only trace is `error getting config file path: error.NotDir` on stderr.

### Cause

`openPath` creates the config directory before opening the file. When a path component already exists, `createDirPath` checks that it is a directory **without following symlinks** — `filePathKind` uses `statx` with `AT.SYMLINK_NOFOLLOW`, deliberately, so that a dangling symlink cannot cause an infinite loop. A symlink pointing at a perfectly good directory therefore reports kind `.sym_link` rather than `.directory`, and the call returns `error.NotDir` before anything reaches `xdg-open`.

This is not present in 1.3.1. It arrived with the Zig 0.16 migration in e8525c0fd, which substituted `std.fs.cwd().makePath` for `std.Io.Dir.cwd().createDirPath`. The migration is a faithful one-for-one change; the trap is that the two functions differ on exactly this edge case, because `makePath` checked with `statFile`, which follows symlinks.

### Fix

Skip the creation when the directory is already reachable. `access` follows symlinks, so the case that used to fail is now the case that needs no work, and a genuine failure to create still propagates from the unchanged `createDirPath` call.

### Why skipping is safe

`access` succeeding means the directory exists and is reachable, so there is nothing to create. `access` failing runs the original `createDirPath` under `try`, exactly as before.

There is one case where behaviour differs and it is worth stating: if the config-directory path exists but is **not** a directory, `access` succeeds, creation is skipped, and the failure surfaces one line later from `createFileAbsolute` instead. Both report an error; only the origin moves. That case is also unreachable in practice — with `$XDG_CONFIG_HOME/ghostty` as a regular file, Ghostty panics during startup long before any keybinding runs, and that panic reproduces on unmodified `main`, so it is a separate matter and out of scope here.

### Worth knowing before merging

**The Zig-side regression is transient.** 0.16.0 was never released; `main` builds against a beta snapshot. On Zig master the API has been renamed back to `makePath` and the existing-component check follows symlinks again, so this will most likely stop reproducing at the next Zig bump. The guard is cheap and correct against either Zig, but you may reasonably prefer to wait — I would rather say so than have you find out afterwards.

**`writeConfigTemplate` has the identical problem.** `Config.zig` takes `dirname` of the config file, i.e. the same symlinked directory, so a user with a symlinked config directory and no config file yet cannot get a template written either. I have deliberately not touched it here — happy to add it to this PR, or leave it for a separate one, whichever you prefer. Three other `createDirPath` sites are only affected if their own directories are symlinks, which is speculative, so I have left those alone.

### Testing

Reproduced on unmodified `main` with `ln -s ~/dotfiles/ghostty ~/.config/ghostty`. After the fix, `Ctrl+,` opens the editor and the log shows `mailbox message=open_config` with no error. `zig build test` passes; `zig fmt --check` clean.

Environment: Ubuntu 26.04, GNOME on Wayland, GTK 4.22.4, libadwaita 1.9.0.

---

**AI disclosure per `AI_POLICY.md`:** I diagnosed and patched this with Claude Code; I verified the reproduction on my own machine and reviewed and edited this description.

### DIFF
--- a/src/config/edit.zig
+++ b/src/config/edit.zig
@@ -27,9 +27,19 @@ pub fn openPath(alloc_gpa: Allocator) ![:0]const u8 {
     // Get the path we should open
     const config_path = try configPath(alloc_arena);
 
-    // Create config directory recursively.
+    // Create config directory recursively, unless it is already there.
+    //
+    // The existence check is not just an optimization. When a component
+    // already exists, `createDirPath` verifies that it is a directory without
+    // following symlinks, deliberately, so that a dangling symlink cannot
+    // cause an infinite loop. A config directory symlinked into a dotfiles
+    // repository therefore fails with `error.NotDir` even though it is a
+    // perfectly good directory.
     if (std.fs.path.dirname(config_path)) |config_dir| {
-        try std.Io.Dir.cwd().createDirPath(global.io(), config_dir);
+        const cwd = std.Io.Dir.cwd();
+        cwd.access(global.io(), config_dir, .{}) catch {
+            try cwd.createDirPath(global.io(), config_dir);
+        };
     }
 
     // Try to create file and go on if it already exists


tristan957:
Please do not post a wall of slop. If you are going to interact with a human, have some respect for that human by not sending word vomit to them.

tristan957:
Closing because we will take https://github.com/ghostty-org/ghostty/pull/13736.

PR #13736, opened 2026-08-10
config/edit: better handling of existing paths

vancluever:
This adds some better handling of existing paths when editing configuration files:

* If we've found an existing file we just skip any attempts to create files/dirs, and just return the path.

* If the path (including file) does not exist, we check to see if the directory exists first (possibly following symlinks). Directory creation happens normally after this (note that any intermediary symlinks in this process will still cause the process to fail, this is to prevent infinite loops, as per the comments in std.Io.Threaded.dirCreateDirPath).

### DIFF
--- a/src/config/edit.zig
+++ b/src/config/edit.zig
@@ -27,31 +27,52 @@ pub fn openPath(alloc_gpa: Allocator) ![:0]const u8 {
     // Get the path we should open
     const config_path = try configPath(alloc_arena);
 
-    // Create config directory recursively.
-    if (std.fs.path.dirname(config_path)) |config_dir| {
-        try std.Io.Dir.cwd().createDirPath(global.io(), config_dir);
+    if (!config_path.exists) {
+        if (std.fs.path.dirname(config_path.name)) |config_dir| check_dir: {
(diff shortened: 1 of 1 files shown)


vancluever:
Fixes #13714, supersedes #13734. Note that this differs from the latter via more thoughtful use of data we've already gathered (e.g., if we already know the file exists, we just skip that logic altogether).

src/config/edit.zig
```diff
@@ -27,31 +27,52 @@ pub fn openPath(alloc_gpa: Allocator) ![:0]const u8 {
...
+                error.PathAlreadyExists => {},
+                else => return err,
+            }
+        };
+    }
+
+    return try alloc_gpa.dupeZ(u8, config_path.name);
 }
 
+const configPathResult = struct {
```
jcollie: Type names should be PascalCase

jcollie (changes requested):
One change…

src/config/edit.zig
vancluever: Doh, got Go-brained for a minute there, good catch :rofl:

jcollie (approved):

Issue #13733, opened 2026-08-10
Unicode Hex Input produces �� instead of emoji on macOS


bo2themax:
### Discussed in https://github.com/ghostty-org/ghostty/discussions/13730

<div type='discussions-op-text'>

<sup>Originally posted by **richardgill** August 10, 2026</sup>
### Issue Description

I use macOS's Unicode Hex Input source to enter characters by typing their Unicode value while holding Option.

Characters represented by a single UTF-16 value work correctly. For example, holding Option, typing `2705`, and releasing Option produces ✅.

Characters represented by two UTF-16 values do not work correctly. Holding Option, typing `D83EDDF5`, and releasing Option produces `��` instead of 🧵. The same input produces 🧵 correctly in other macOS applications.

Keyboard: built-in MacBook Air keyboard  
Input source: macOS Unicode Hex Input

Instructions for enabling Unicode Hex Input on MacOS:  
https://servicedesk.uwgb.edu/TDClient/33/Portal/KB/Article/3648/Mac-Unicode-Hex-Input

This appears to be a remaining surrogate-pair edge case from the broader Unicode Hex Input problem discussed in #5306.

I have tested a proposed fix on my Mac:  
https://github.com/richardgill/ghostty/tree/fix/macos-split-surrogate-input

AI disclosure: I used the Pi coding agent to help investigate the behavior, develop and test the proposed fix, and edit this report. I reviewed the final text and code.

### Expected Behavior

Holding Option, typing D83EDDF5, and releasing Option should insert 🧵 once, matching the behavior of other macOS applications.

### Actual Behavior

Ghostty inserts two Unicode replacement characters, `��`, instead of 🧵.

### Reproduction Steps

1. Enable the macOS Unicode Hex Input source by following these instructions:  
   https://servicedesk.uwgb.edu/TDClient/33/Portal/KB/Article/3648/Mac-Unicode-Hex-Input
2. Use the minimal Ghostty configuration provided below.
3. Restart Ghostty and select Unicode Hex Input from the macOS input menu.
4. Hold Option, type `2705`, and release Option. Ghostty inserts ✅ correctly.
5. Hold Option, type `D83EDDF5`, and release Option.
6. Observe that Ghostty inserts `��` instead of 🧵.
7. Repeat the second input in another macOS application and observe that it inserts 🧵 correctly.

Feel free to tag me anytime if you would like me to test a fix on my machine.

### Ghostty Version

```text
Ghostty 1.3.1

Version
[...]

PR #13737, opened 2026-08-10
macOS: support decoding the surrogate pair with UnicodeHexInput

bo2themax:
Not yet a perfect fix for https://github.com/ghostty-org/ghostty/issues/13733

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
@@ -190,6 +190,8 @@ extension Ghostty {
 
         // This is set to non-null during keyDown to accumulate insertText contents
         private var keyTextAccumulator: [String]?
+        /// Temporary lead surrogate that's waiting for the trail
+        private var leadSurrogate: LeadSurrogate?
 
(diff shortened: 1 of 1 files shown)


mitchellh:
Yeah this is super weird. Ghostty understands raw UTF-8, can these not be UTF-8 encoded and let Ghostty decode the bytes?

bo2themax:
> Yeah this is super weird. Ghostty understands raw UTF-8, can these not be UTF-8 encoded and let Ghostty decode the bytes? 

I tried this as well, decoding surrogates correctly and send them to the surface, but no🤔. Might give Claude a go today.

bo2themax:
> Yeah this is super weird. Ghostty understands raw UTF-8, can these not be UTF-8 encoded and let Ghostty decode the bytes?

So I checked(with Claude), our UTF-8 handling in the lib doesn't cover this, given the fact that AppKit deliver them separately. Since a lone surrogate can't be represented in standard UTF8, we still need some extra handling in either apprt or lib to join them.

For now I think only change swift makes sense. If in the future we support windows upstream and run into the same issue there, we can move it to the lib then. 
> I also tried with that approach locally, it adds more complexity.

PR #13741, opened 2026-08-11
gtk: Fix +new-window -e <command> inheriting forced shell integration from the running GTK application. Fixes #12378.

### LINKED ISSUE #12378: Setting `shell-integration = nushell` breaks `ghostty +new-window -e` on Linux

PRIHLOP:
## Summary 
Fix `+new-window -e <command>` inheriting forced shell integration from the running GTK application.

Arguments after `-e` are now marked as an explicit command that requires shell detection. That marker is propagated with the existing command override to the new surface for both the current `+new-window` and `+new-tab` GTK paths. When the override is applied, forced shell integration becomes `detect`; an explicit `none` remains disabled.

This is scoped to the GTK `-e` path. It does not change the generic forced shell-integration behavior, so configured integration continues to support shell executables with non-standard names.

Fixes #12378.

## Root cause

The GTK IPC path replaced `config.command` for the new surface but retained the running application's `shell-integration` value. `termio` therefore treated an arbitrary explicit command such as Vim as the forced shell and appended shell-specific arguments.

## AI usage

I used Hermes Agent (with ChatGPT 5.6) to help inspect the codebase, trace the GTK command path, and draft parts of the implementation and test. I manually reviewed and edited the changes, validated the behavior and test results, and understand the affected code paths.

### DIFF
--- a/src/apprt/gtk/class/application.zig
+++ b/src/apprt/gtk/class/application.zig
@@ -12,6 +12,7 @@ const gtk = @import("gtk");
 
 const build_config = @import("../../../build_config.zig");
 const build_info = @import("../build/info.zig");
+const cli = @import("../../../cli.zig");
 const global = @import("../../../global.zig");
 const i18n = @import("../../../os/main.zig").i18n;
 const apprt = @import("../../../apprt.zig");
@@ -1854,6 +1855,7 @@ pub const Application = extern struct {
             null,
             if (overrides) |o| .{
                 .command = o.command,
+                .shell_integration = o.shell_integration,
                 .working_directory = o.working_directory,
                 .title = o.title,
             } else .none,
@@ -1934,6 +1936,7 @@ pub const Application = extern struct {
                 },
                 .{
                     .command = overrides.command,
+                    .shell_integration = overrides.shell_integration,
                     .working_directory = overrides.working_directory,
                     .title = overrides.title,
                 },
@@ -1946,6 +1949,7 @@ pub const Application = extern struct {
                 null,
                 .{
                     .command = overrides.command,
+                    .shell_integration = overrides.shell_integration,
                     .working_directory = overrides.working_directory,
                     .title = overrides.title,
                 },
@@ -1957,6 +1961,7 @@ pub const Application = extern struct {
 
     fn parseOverrides(arena_alloc: Allocator, arguments_it: *glib.VariantIter) (Allocator.Error || error{ValueRequired})!struct {
         command: ?configpkg.Command = null,
+        shell_integration: ?configpkg.Config.ShellIntegration = null,
         working_directory: ?[:0]const u8 = null,
         title: ?[:0]const u8 = null,
     } {
@@ -1965,6 +1970,9 @@ pub const Application = extern struct {
         var working_directory: ?[:0]const u8 = null;
         var title: ?[:0]const u8 = null;
         var command: ?configpkg.Command = null;
+        var parsed_shell_integration: struct {
+            @"shell-integration": ?configpkg.Config.ShellIntegration = null,
+        } = .{};
 
         const s_variant_type = glib.VariantType.new("s");
         defer s_variant_type.free();
@@ -2010,6 +2018,19 @@ pub const Application = extern struct {
                 command = cmd;
                 continue;
             }
+            if (std.mem.cutPrefix(u8, str, "--shell-integration=")) |v| {
+                cli.args.parseIntoField(
+                    @TypeOf(parsed_shell_integration),
+                    arena_alloc,
+                    &parsed_shell_integration,
+                    "shell-integration",
+                    std.mem.trim(u8, v, &std.ascii.whitespace),
+                ) catch |err| {
+                    log.warn("unable to parse shell integration {s}: {t}", .{ v, err });
+                    continue;
+                };
+                continue;
+            }
             if (std.mem.cutPrefix(u8, str, "--working-directory=")) |v| {
                 working_directory = arena_alloc.dupeZ(u8, std.mem.trim(u8, v, &std.ascii.whitespace)) catch |err| {
                     log.warn("unable to duplicate working directory: {t}", .{err});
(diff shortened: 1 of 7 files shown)


jcollie (changes requested):
I don't think that this is the right approach. The user should be able to specify `--shell-integration` on the `ghostty +new-window` or `ghostty +new-tab` CLI and have that passed to the running Ghosty instance. Then in the running Ghostty if the config has `shell-integration` set but `shell-integration` was not received from the CLI then the override logic is applied.

Also, need a _much_ better commit message. The merge message is good and could basically be copied to the commit message.

PRIHLOP (commented):
Okay, I did change as you mentioned.
I verified changes by E2E test and tested GUI manually by launching compiled binary file.
Could you please review new implementation?

src/apprt/gtk/class/application.zig
```diff
@@ -2010,6 +2015,13 @@ pub const Application = extern struct {
                 command = cmd;
                 continue;
             }
+            if (std.mem.cutPrefix(u8, str, "--shell-integration=")) |v| {
+                shell_integration = parseShellIntegration(v) orelse {
```
jcollie: Rather than a custom parsing function (which would require upkeep as new integrations are added) reuse `parseIntoField` from `src/cli/args.zig`

src/apprt/gtk/class/surface.zig
```diff
@@ -4387,9 +4395,49 @@ fn computeFraction(progress: u8) f64 {
...
+    applyShellIntegrationOverride(&config, true, .nushell);
+    try testing.expectEqual(.nushell, config.@"shell-integration");
+
+    applyShellIntegrationOverride(&config, true, .none);
+    try testing.expectEqual(.none, config.@"shell-integration");
+
+    config.@"shell-integration" = .nushell;
+    applyShellIntegrationOverride(&config, false, null);
+    try testing.expectEqual(.nushell, config.@"shell-integration");
+}
```
jcollie, line 4454: The tests for these functions should be kept next to the functions themselves.

src/apprt/gtk/class/surface.zig
```diff
@@ -3530,6 +3533,11 @@ pub const Surface = extern struct {
         if (priv.overrides.command) |c| {
             config.command = try c.clone(config._arena.?.allocator());
         }
+        applyShellIntegrationOverride(
+            &config,
+            priv.overrides.command != null,
+            priv.overrides.shell_integration,
+        );
         if (priv.overrides.working_directory) |wd| {
             const config_alloc = config.arenaAlloc();
```
jcollie, line 3539: If you move the call to `applyShellIntegrationOverride` inside the previous `if` statement the boolean parameter is unnecessary.

jcollie (changes requested):
Just a few small changes needed.

PRIHLOP (commented):
Thanks. The custom parser with `cli.args.parseIntoField` was replaced, and the tests with the helper was colocated, also the boolean parameter was removed. 
The command-specific detect fallback now only runs when a command is overridden. Explicit `--shell-integration` handling outside that condition was kept because the option is also valid without `-e/--command`(otherwise it would be silently ignored). Also was added a regression coverage for both standalone explicit values and command overrides.

jcollie (approved):

PR #13744, opened 2026-08-11
macos: defer OSC52 clipboard read confirmations until focused

mitchellh:
Fixes #10077

Clipboard read confirmations would immediately show a sheet which grabbed focus. This could be used for a bunch of dumb reasons, including  DoS attacks. But, it also caused focus/sheet loops for programs that did OSC52 on focus changes (which was seen via some Neovim configs!).

Now, if a surface is unfocused, we bell the surface and show the confirmation request on next focus. If the surface is not focused or another request comes in, we cancel the prior one.

This also fixes some memory management issues around clipboard requests that were likely small leaks (didn't verify the old bug, but verified the new code, and eyeballed the old).

To implement this, I decided to reorient the whole clipboard confirmation thing around state on SurfaceView (which simplifies memory management) and using Combine on BaseTerminalController to get notified.

### DIFF
--- a/macos/Sources/Features/ClipboardConfirmation/ClipboardConfirmationController.swift
+++ b/macos/Sources/Features/ClipboardConfirmation/ClipboardConfirmationController.swift
@@ -1,7 +1,6 @@
 import Foundation
 import Cocoa
 import SwiftUI
-import GhosttyKit
 
 /// This initializes a clipboard confirmation warning window. The window itself
 /// WILL NOT show automatically and the caller must show the window via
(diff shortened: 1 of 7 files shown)


bo2themax (approved):
I'm thinking we can move the confirmation to an overlay of the surface, so we don't need to check them in the window controller. But I haven't tried it yet, will open a follow up pr if it works better.

bo2themax (changes requested):
I tried a script that repeated request clipboard write/read, it still steals focus from another tab

https://github.com/user-attachments/assets/23a49f99-f74b-453c-bef4-992c54fb801d

mitchellh:
Oh strange, I had a similar script and it did not. let me try again.

mitchellh:
Oh its because this only protected OSC52 read, but we should probably just do this for all clipboard ops that require confirmation. I started with OSC52 due to the linked discussion.

mitchellh:
Fixed, pushed.

bo2themax (approved):

PR #10531, opened 2026-02-02
feat: add toggle_broadcast action for synchronized input

sconeman:
## Summary

Add a new keybindable action `toggle_broadcast` that broadcasts all keyboard input to all terminal splits simultaneously, similar to tmux's `synchronize-panes` or iTerm2's CMD+SHIFT+I feature.

## Changes

- Add `toggle_broadcast` keybindable action in `Binding.zig`
- Add `broadcast_enabled: bool` state field in `App.zig` (app-level)
- Add `broadcast_mode` apprt action for UI layer notification
- Update C ABI in `ghostty.h`
- Implement broadcast logic in `Surface.zig` `keyCallback()`
- Add command palette entry "Toggle Broadcast Mode"
- Add unit tests for parsing and enum values

## Features

- Global broadcast mode affects all surfaces in the application
- Skips read-only surfaces
- Skips surfaces with exited child processes
- Proper error handling and logging for allocation failures

## Usage

Add to config:
```ini
keybind = cmd+shift+i=toggle_broadcast
```

Then:
1. Open multiple splits
2. Press the keybind to enable broadcast mode
3. Type in one terminal - input appears in ALL terminals
4. Press keybind again to disable

## Test plan

- [x] `zig build test -Dtest-filter="broadcast"` - All tests pass
- [x] Manual testing with multiple splits
- [x] Verified read-only surfaces are skipped
- [x] Verified toggle on/off works correctly

PR #13745, opened 2026-08-11
macos,gtk: add broadcast support to sync across panes

dave92082:
## Summary

  This change adds synchronized input across terminal panes. Users asked for this function in #3227, #11797, and #12832. You can put panes into groups. You can make a maximum of 16 groups. When you send input to one
  pane in a group, all panes in that group get the same input. A colored border shows the panes in each group. The core code controls all groups. The macOS runtime and the GTK runtime use the same core code.
  Programs that use libghostty can also use this function through the binding-action API.

 ## Description

  An earlier change (#10531) had one global switch. This change uses groups in place of one global switch. Each group has a different border color. A group can contain panes from different splits, tabs, and windows.

  How to use the function

  - Click a pane while you hold the modifier keys. This click puts the pane into a group or removes it from its group. If the pane that has focus is in a group, the clicked pane goes into the same group. If not, the
  click makes a new group. The new group gets the lowest free group number.
  - Use the broadcast-group-click-mods option to set the modifier keys. The default is super+shift on macOS. The default is ctrl+shift on GTK. Set the option to false to disable the click function.
  - The modifier keys must be an exact match. Thus the click cannot change the other click functions. These functions stay the same: shift+click, cmd+click, ctrl+click, and ctrl+alt+drag.
  - Use the toggle_broadcast_group:N key binding to put the pane that has focus into group N. N can be 1 to 16. Use the same key binding again to remove the pane from the group. The default key bindings are for
  groups 1 to 10. On macOS the default is cmd+ctrl+<digit>. On the other platforms the default is ctrl+shift+<digit>. The digit 0 is group 10. The key bindings apply to the physical digit keys and the unicode digit
  keys. Thus keyboard layouts with shifted digits (for example, AZERTY) operate correctly.
  - Use the clear_broadcast_groups action to remove all groups at one time. This action is also in the command palette.
  - Use the broadcast-group-colors option to set the border colors. This option operates the same as the palette option. Each N=color value changes one color. You do not give the full list. Example:
  broadcast-group-colors = 3=#ff0000. The 16 default colors are in a special sequence. Two adjacent group numbers do not get almost equal colors.

  https://github.com/user-attachments/assets/3c372afd-fdac-4962-bc73-4dc1fef371f4

  AI disclosure

  Claude Code with the Fable 5 model helped with this change. I made the design for the user interaction. I read each change. I did the manual tests on real hardware.

  Vouched in #13742.


bo2themax (changes requested):
The results look good to me, but i would recommend separate changes to 2/3 prs(core, gtk, macOS). It's easier to review and more like to get merged incrementally.

I also don't like hard coding the shortcuts for enabling the broadcast selection. Maybe we should add a keybind for that.

jcollie (changes requested):
This is a great start, but I think that the control of the broadcast groups should be in the core app/surface and not in each individual runtime. That should reduce code duplication and ensure that the logic works the same across runtimes. It would also allow broadcast groups to be exposed to `libghostty-vt` users.

Minor nits:

- The actions should be called `toggle_broadcast_group` not `join_broadcast_group` to more accurately reflect what they are doing.
- There should be a action `clear_broadcast_groups` that would clear all broadcast groups at once so you don't have to remove them one by one.
- The border colors should not be hardcoded, but confgurable.

One the above things are addressed I'm sure I'll have further feedback but that's more than enough to get started with.

dave92082:
@bo2themax:
 1. I updated the shortcuts for clicking in to the groups to be configurable following the same config override pattern used elsewhere for key+mouse interactions.
 2. With the changes requested by @jcollie to centralize the functionality it makes much less sense to split the work up between 2/3 PR's. If you think that is still needed, let me know and perhaps I can stage Core on one and UI's on a second

@jcollie:
1. Great callout for moving the logic to core, I have re-worked it and the runtimes now share the Core implementation. 
2. Actions have been renamed as requested
3. clear_broadcast_group has been added and is available in the command palette
4. Borders colors are now configurable with a default of the same 10. If configured, up to 64 are allowed which would increase the number of allowable groups. Not sure anyone would have a need to go that high but would love to see a screenshot if they did :)

Other items I found and fixed while test driving it today for a while:
* Pasted input was not being sent to all broadcast group members. 
* Non visible input characters (Home, End, etc.) where not being sent to all broadcast group members

jcollie:
Still working my way through this, but here's a couple of initial thoughts:

- The `BroadcastGroup` seems over-engineered. Lots of `O(n²)` operations. I think that just adding `broadcast_group:  ?u8 = null` to the surface would work better. Sending a broadcast to a group is just an `O(n)` operation then. Plus eliminates many allocations. 
- Keeping track of the maximum number of configured colors seems like unnecessary work. We should pick a maximum number of groups (16? 64 I think is overkill) and just stick with that. We'd have to pick distinct-enough default colors for all of them though.
- We should re-use some of the machinery that is used for parsing palettes. That seems like a better way to configure the colors than what `ColorList` provides. This would also allow for overriding only specific colors rather than having to always specifying the whole list.
- On GTK, the broadcast border should be controlled through a template binding (see how the bell border or the readonly overlay are controlled). Unfortunately GTK doesn't allow you to add/remove CSS classes through templates so that won't work.

(No comment on the Swift code as I'm not a macOS dev).

dave92082:
@jcollie  Looking through the implementation and the codebase (long time user, first time contributor 😉), simplifying the BroadcastGroup as you suggest should make this overall change much smaller. I agree that 64 is overkill and that 16 should be more than enough for everyone. I've been daily driving this PR for a couple days now and haven't used more than 3. 

I'll work on incorporating the first three items tonight and will have to do some research for the GTK template binding though I'm not sure how fruitful that will be.

src/apprt/gtk/class/application.zig
```diff
@@ -2450,6 +2463,19 @@ const Action = struct {
...
     }
 
+    pub fn broadcastGroup(
+        target: apprt.Target,
+        value: apprt.action.BroadcastGroup,
+    ) bool {
+        switch (target) {
+            .app => return false,
+            .surface => |core| {
+                core.rt_surface.surface.updateBroadcastGroup(value);
```
pluiedev: Why does this need to be an apprt action if all this does is tying into core state? Doing this purely at the core level would make more sense.

src/apprt/action.zig
```diff
@@ -559,6 +566,16 @@ pub const MoveTab = extern struct {
...
 
+/// A surface's current broadcast input group membership.
+pub const BroadcastGroup = extern struct {
+    /// True when the surface is a member of a group.
+    member: bool,
+
+    /// The group's color index into `broadcast-group-colors`. Only
+    /// meaningful when member is true.
+    color: u8,
+};
```
pluiedev: This should be a `?u8` with an associated C struct:

```suggestion
pub const BroadcastGroup = struct {
    /// The group's color index into `broadcast-group-colors`.
    color: ?u8,
    
    // Sync with: ghostty_action_broadcast_group_s
    pub const C = extern struct {
        member: bool,
        color: u8,
    };

    pub fn cval(self: BroadcastGroup) C {
        return .{
            .member = self.color != null,
            .color = self.color orelse 0,
        };
    }
};
```

src/App.zig
```diff
@@ -557,6 +558,179 @@ pub fn findSurfaceByID(self: *const App, id: u64) ?*Surface {
...
+
+        break :group self.lowestUnusedBroadcastGroup() orelse return;
+    };
+    syncBroadcastGroup(rt_app, surface);
+}
+
+/// The lowest group number without any member surface, or null if all
+/// group numbers are taken.
+fn lowestUnusedBroadcastGroup(self: *const App) ?u8 {
+    var used = [_]bool{false} ** max_broadcast_groups;
```
pluiedev: Use `@splat` instead, or alternatively use `std.StaticBitSet`


```suggestion
    var used: [max_broadcast_groups]bool = @splat(false);
```

src/App.zig
```diff
@@ -557,6 +558,179 @@ pub fn findSurfaceByID(self: *const App, id: u64) ?*Surface {
...
+    for (self.surfaces.items) |rt_surface| {
+        if (rt_surface.core().broadcast_group) |group| used[group] = true;
+    }
+    for (used, 0..) |v, i| if (!v) return @intCast(i);
```
pluiedev: If you use `std.StaticBitSet` there's `findLastSet` which does this much more efficiently

```suggestion
    const last = bit_set.findLastSet() orelse return null;
    return @intCast(last + 1);
```

pluiedev (changes requested):

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1200,6 +1203,42 @@ extension Ghostty {
...
+                    return false
+
+                case GHOSTTY_TARGET_SURFACE:
+                    guard let surface = target.target.surface else { return false }
+                    guard let surfaceView = self.surfaceView(from: surface) else { return false }
+
+                    return MainActor.assumeIsolated {
+                        surfaceView.broadcastGroupColor = broadcastGroupColor(v)
+                        return true
+                    }
```
bo2themax: I wouldn't recommend use `MainActor.assumeIsolated` here, although in most of the cases it will be called on main thread. We don't want crashing the app if something weird happens.

```suggestion
                    Task { @MainActor in
                        surfaceView.broadcastGroupColor = broadcastGroupColor(v)
                    }
                    return true
```

macos/Sources/Ghostty/Surface View/SurfaceView.swift
```diff
@@ -200,6 +200,15 @@ extension Ghostty {
                     }
                 }
 
+                #if canImport(AppKit)
```
bo2themax: We removed iOS support just now, so these conditional checks won't be needed.

macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
```diff
@@ -683,6 +688,26 @@ extension Ghostty {
...
+            // configured modifiers must match exactly so that other modified
+            // clicks (shift+click extends selection, cmd+click opens links,
+            // ctrl+click is the context menu) are never shadowed by a
+            // superset combination. When the app isn't active the click is
+            // an ordinary activation click, so let it continue as normal.
+            if NSApp.isActive,
+               let surface = self.surface,
+               let clickMods = derivedConfig.broadcastGroupClickMods,
+               event.modifierFlags.intersection([.shift, .control, .option, .command]) == clickMods {
+                ghostty_surface_toggle_broadcast_group(surface)
```
bo2themax: Agreed with Leah, I'm thinking we can trigger the broadcast and select surfaces with a mouse event with modifiers configured. apprts are only responsible drawing/clearing borders when a surface is selected and deselected.

Then we probably won't need to change it here at all.

bo2themax (changes requested):

dave92082:
Not ignoring these comments today, lost power around 11am PST and it's still out. I'll get back to this tomorrow when it is (hopefully) back up

dave92082:
@jcollie @pluiedev @bo2themax , all PR comments should be addressed. I have tested on both mac and fedora

pluiedev (approved):
Thanks. No more concerns on the GTK side.

bo2themax (approved):

dave92082:
Updated PR description to reflect approved changes

bo2themax:
> Updated PR description to reflect approved changes

I don't think you need to add to the **how**s in the description; descriptions are more of why and additional notes. This would help others to understand the changes when blaming. That big chunk of text seems overwhelming to me.

src/apprt/action.zig
```diff
@@ -563,6 +570,26 @@ pub const MoveTab = extern struct {
     amount: isize,
 };
 
+/// A surface's current broadcast input group membership.
+pub const BroadcastGroup = struct {
+    /// The group's color index into `broadcast-group-colors`, or null
+    /// if the surface is not in a group.
+    color: ?u8,
```
jcollie: This should be called `group` and should be a `u4`.

src/apprt/action.zig
```diff
@@ -563,6 +570,26 @@ pub const MoveTab = extern struct {
...
+
+    // Sync with: ghostty_action_broadcast_group_s
+    pub const C = extern struct {
+        member: bool,
+        color: u8,
```
jcollie: This should be called `group` not color. Keep as `u8` since C doesn't support `u4`

src/apprt/gtk/css/style.css
```diff
@@ -134,6 +134,16 @@ label.resize-overlay {
...
 }
 
+/*
+ * Broadcast input groups. The border color comes from a
+ * broadcast-color-N class generated at runtime from the
+ * broadcast-group-colors configuration.
+ */
+.surface .broadcast-overlay {
+  border-width: 2px;
+  border-style: solid;
```
jcollie, line 144:
```suggestion
  border-style: solid;
  border-color: rgba(0, 0, 0, 0);
```

This eliminates an annoying flash of color when turning off a group caused by the custom color CSS class being removed faster than the revealer hides the border.

src/apprt/gtk/class/surface.zig
```diff
@@ -727,6 +729,11 @@ pub const Surface = extern struct {
         /// and the matching left mouse release should also be suppressed.
         suppress_left_mouse_release: bool = false,
 
+        /// The broadcast group color index currently shown by the border
+        /// overlay, so updateBroadcastGroup can remove the matching CSS
+        /// class when the group changes. Null when not in a group.
+        broadcast_color: ?u8 = null,
```
jcollie:
```suggestion
        broadcast_group: ?u4 = null,
```

src/apprt/gtk/class/application.zig
```diff
@@ -2473,6 +2486,19 @@ const Action = struct {
         }
     }
 
+    pub fn broadcastGroup(
```
jcollie:
```suggestion
    pub fn syncBroadcastGroup(
```

src/apprt/gtk/class/application.zig
```diff
@@ -2473,6 +2486,19 @@ const Action = struct {
...
+        target: apprt.Target,
+        value: apprt.action.BroadcastGroup,
+    ) bool {
+        switch (target) {
+            .app => return false,
+            .surface => |core| {
+                core.rt_surface.surface.updateBroadcastGroup(value);
```
jcollie:
```suggestion
                core.rt_surface.surface.syncBroadcastGroup(value);
```

src/apprt/gtk/class/surface.zig
```diff
@@ -1477,6 +1484,38 @@ pub const Surface = extern struct {
...
+        priv.broadcast_revealer.setRevealChild(@intFromBool(color != null));
+    }
+
+    /// The CSS class carrying the border color for the given broadcast
+    /// group color index. Must match the class names generated by the
+    /// application's runtime CSS.
+    fn broadcastColorClass(buf: []u8, color: u8) [:0]const u8 {
+        return std.fmt.bufPrintZ(buf, "broadcast-color-{d}", .{color}) catch
+            unreachable;
+    }
```
jcollie:
```suggestion
    fn broadcastColorClass(buf: []u8, group: u4) [:0]const u8 {
        return std.fmt.bufPrintZ(buf, "broadcast-color-{d}", .{group}) catch
            unreachable;
    }
```

src/apprt/action.zig
```diff
@@ -428,6 +434,7 @@ pub const Action = union(Key) {
         readonly,
         copy_title_to_clipboard,
         move_tab_to_new_window,
+        broadcast_group,
```
jcollie:
```suggestion
        sync_broadcast_group,
```

src/apprt/action.zig
```diff
@@ -357,6 +357,12 @@ pub const Action = union(Key) {
     /// Move a tab to a new window.
     move_tab_to_new_window,
 
+    /// The target surface's broadcast input group membership changed.
+    /// The apprt should update its visual indicator (e.g. a border in
+    /// the group's color). Group membership itself is managed entirely
+    /// by the core.
+    broadcast_group: BroadcastGroup,
```
jcollie:
```suggestion
    sync_broadcast_group: BroadcastGroup,
```

include/ghostty.h
```diff
@@ -620,6 +625,12 @@ typedef enum {
   GHOSTTY_GOTO_TAB_LAST = -3,
 } ghostty_action_goto_tab_e;
 
+// apprt.action.BroadcastGroup
+typedef struct {
+  bool member;
+  uint8_t color;
```
jcollie:
```suggestion
  uint8_t group;
```

include/ghostty.h
```diff
@@ -972,13 +983,15 @@ typedef enum {
   GHOSTTY_ACTION_READONLY,
   GHOSTTY_ACTION_COPY_TITLE_TO_CLIPBOARD,
   GHOSTTY_ACTION_MOVE_TAB_TO_NEW_WINDOW,
+  GHOSTTY_ACTION_BROADCAST_GROUP,
```
jcollie:
```suggestion
  GHOSTTY_ACTION_SYNC_BROADCAST_GROUP,
```

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -638,6 +638,9 @@ extension Ghostty {
             case GHOSTTY_ACTION_COPY_TITLE_TO_CLIPBOARD:
                 return copyTitleToClipboard(app, target: target)
 
+            case GHOSTTY_ACTION_BROADCAST_GROUP:
+                return broadcastGroup(app, target: target, v: action.action.broadcast_group)
```
jcollie:
```suggestion
                return toggleBroadcastGroup(app, target: target, v: action.action.broadcast_group)
```

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1139,6 +1142,42 @@ extension Ghostty {
                 return true
         }
 
+        private static func broadcastGroup(
```
jcollie:
```suggestion
        private static func syncBroadcastGroup(
```

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -1139,6 +1142,42 @@ extension Ghostty {
...
+        }
+
+        /// The border color for the given broadcast group membership, from
+        /// the broadcast-group-colors configuration.
+        @MainActor
+        private static func broadcastGroupColor(_ v: ghostty_action_broadcast_group_s) -> Color? {
+            guard v.member else { return nil }
+            guard let delegate = NSApp.delegate as? AppDelegate else { return nil }
+            let colors = delegate.ghostty.config.broadcastGroupColors
+            let idx = Int(v.color)
```
jcollie:
```suggestion
            let idx = Int(v.group)
```

src/apprt/gtk/class/application.zig
```diff
@@ -716,6 +716,8 @@ pub const Application = extern struct {
                 value,
             ),
 
+            .broadcast_group => return Action.broadcastGroup(target, value),
```
jcollie:
```suggestion
            .sync_broadcast_group => return Action.syncBroadcastGroup(target, value),
```

src/apprt/action.zig
jcollie:
```suggestion
    group: ?u4,
```

src/apprt/action.zig
jcollie:
```suggestion
        group: u8,
```

src/apprt/action.zig
```diff
@@ -563,6 +570,26 @@ pub const MoveTab = extern struct {
...
+    // Sync with: ghostty_action_broadcast_group_s
+    pub const C = extern struct {
+        member: bool,
+        color: u8,
+    };
+
+    pub fn cval(self: BroadcastGroup) C {
+        return .{
+            .member = self.color != null,
+            .color = self.color orelse 0,
```
jcollie:
```suggestion
            .member = self.group != null,
            .group = self.group orelse 0,
```

src/config/Config.zig
```diff
@@ -6961,6 +7011,56 @@ pub const Keybinds = struct {
...
+                try self.set.put(
+                    alloc,
+                    .{
+                        .key = .{ .physical = @field(
+                            inputpkg.Key,
+                            std.fmt.comptimePrint("digit_{u}", .{i}),
+                        ) },
+                        .mods = mods,
+                    },
+                    .{ .toggle_broadcast_group = (i - start) + 1 },
```
jcollie:
```suggestion
                    .{ .toggle_broadcast_group = (i - start) },
```

src/config/Config.zig
```diff
@@ -6961,6 +7011,56 @@ pub const Keybinds = struct {
...
+                );
+
+                try self.set.put(
+                    alloc,
+                    .{ .key = .{ .unicode = i }, .mods = mods },
+                    .{ .toggle_broadcast_group = (i - start) + 1 },
```
jcollie:
```suggestion
                    .{ .toggle_broadcast_group = (i - start) },
```

src/config/Config.zig
```diff
@@ -6961,6 +7011,56 @@ pub const Keybinds = struct {
...
+                );
+            }
+
+            // 0 joins group 10, matching the number row read left
+            // to right.
+            try self.set.put(
+                alloc,
+                .{ .key = .{ .physical = .digit_0 }, .mods = mods },
+                .{ .toggle_broadcast_group = 10 },
```
jcollie:
```suggestion
                .{ .toggle_broadcast_group = 9 },
```

src/config/Config.zig
```diff
@@ -6961,6 +7011,56 @@ pub const Keybinds = struct {
...
+            );
+            try self.set.put(
+                alloc,
+                .{ .key = .{ .unicode = '0' }, .mods = mods },
+                .{ .toggle_broadcast_group = 10 },
```
jcollie:
```suggestion
                .{ .toggle_broadcast_group = 9 },
```

src/config/Config.zig
```diff
@@ -8731,6 +8831,135 @@ pub const MiddleClickAction = enum {
...
+
+/// See broadcast-group-colors. This works like Palette: a fixed-size
+/// color table where individual entries can be overridden with
+/// `N=color` config values.
+pub const BroadcastGroupColors = struct {
+    const Self = @This();
+
+    /// The number of colors, which is also the maximum number of
+    /// broadcast groups that can exist at once.
+    pub const len = 16;
```
jcollie:
```suggestion
    pub const len = std.math.maxInt(u4) + 1;
```

src/config/Config.zig
```diff
@@ -8731,6 +8831,135 @@ pub const MiddleClickAction = enum {
...
+        .{ .r = 0x33, .g = 0xc7, .b = 0xde }, // 7: cyan
+        .{ .r = 0xdc, .g = 0x8a, .b = 0xdd }, // 8: pink
+        .{ .r = 0xb5, .g = 0x83, .b = 0x5a }, // 9: brown
+        .{ .r = 0x99, .g = 0xc1, .b = 0xf1 }, // 10: light blue
+        .{ .r = 0xa5, .g = 0x1d, .b = 0x2d }, // 11: dark red
+        .{ .r = 0x8f, .g = 0xf0, .b = 0xa4 }, // 12: light green
+        .{ .r = 0x61, .g = 0x35, .b = 0x83 }, // 13: dark purple
+        .{ .r = 0x9a, .g = 0x99, .b = 0x96 }, // 14: gray
+        .{ .r = 0x1a, .g = 0x5f, .b = 0xb4 }, // 15: dark blue
+        .{ .r = 0xff, .g = 0xbe, .b = 0x6f }, // 16: light orange
```
jcollie:
```suggestion
        .{ .r = 0xff, .g = 0x78, .b = 0x00 }, // 0: orange
        .{ .r = 0x35, .g = 0x84, .b = 0xe4 }, // 1: blue
        .{ .r = 0xf6, .g = 0xd3, .b = 0x2d }, // 2: yellow
        .{ .r = 0x91, .g = 0x41, .b = 0xac }, // 3: purple
        .{ .r = 0x33, .g = 0xd1, .b = 0x7a }, // 4: green
        .{ .r = 0xe0, .g = 0x1b, .b = 0x24 }, // 5: red
        .{ .r = 0x33, .g = 0xc7, .b = 0xde }, // 6: cyan
        .{ .r = 0xdc, .g = 0x8a, .b = 0xdd }, // 7: pink
        .{ .r = 0xb5, .g = 0x83, .b = 0x5a }, // 8: brown
        .{ .r = 0x99, .g = 0xc1, .b = 0xf1 }, // 9: light blue
        .{ .r = 0xa5, .g = 0x1d, .b = 0x2d }, // 10: dark red
        .{ .r = 0x8f, .g = 0xf0, .b = 0xa4 }, // 11: light green
        .{ .r = 0x61, .g = 0x35, .b = 0x83 }, // 12: dark purple
        .{ .r = 0x9a, .g = 0x99, .b = 0x96 }, // 13: gray
        .{ .r = 0x1a, .g = 0x5f, .b = 0xb4 }, // 14: dark blue
        .{ .r = 0xff, .g = 0xbe, .b = 0x6f }, // 15: light orange
```

src/config/Config.zig
```diff
@@ -8731,6 +8831,135 @@ pub const MiddleClickAction = enum {
...
+            u8,
+            std.mem.trim(u8, value[0..eql_idx], " "),
+            10,
+        ) catch return error.InvalidValue;
+        if (number < 1 or number > len) return error.InvalidValue;
+
+        self.colors[number - 1] = try Color.parseCLI(
+            std.mem.trim(u8, value[eql_idx + 1 ..], " "),
+        );
+    }
```
jcollie, line 8930:
```suggestion
    pub fn parseCLI(self: *Self, input: ?[]const u8) !void {
        const value = input orelse return error.ValueRequired;
        const eql_idx = std.mem.indexOfScalar(u8, value, '=') orelse
            return error.InvalidValue;

        const number = std.fmt.parseInt(
            u4,
            std.mem.trim(u8, value[0..eql_idx], " "),
            10,
        ) catch return error.InvalidValue;

        self.colors[number] = try Color.parseCLI(
            std.mem.trim(u8, value[eql_idx + 1 ..], " "),
        );
    }
```

src/config/Config.zig
```diff
@@ -8731,6 +8831,135 @@ pub const MiddleClickAction = enum {
...
+
+    /// Compare if two of our value are requal. Required by Config.
+    pub fn equal(self: Self, other: Self) bool {
+        return std.meta.eql(self, other);
+    }
+
+    /// Used by Formatter
+    pub fn formatEntry(self: Self, formatter: formatterpkg.EntryFormatter) !void {
+        var buf: [128]u8 = undefined;
+        for (self.colors, 1..) |v, number| {
```
jcollie:
```suggestion
        for (self.colors, 0..) |v, number| {
```

src/config/Config.zig
```diff
@@ -8731,6 +8831,135 @@ pub const MiddleClickAction = enum {
...
+            try formatter.formatEntry(
+                []const u8,
+                std.fmt.bufPrint(
+                    &buf,
+                    "{d}=#{x:0>2}{x:0>2}{x:0>2}",
+                    .{ number, v.r, v.g, v.b },
```
jcollie: The `Color` type has a [formatter](https://github.com/ghostty-org/ghostty/blob/3790fb78fecb3577dee30c40efe1ced3e3f0d9a1/src/config/Config.zig#L5649-L5658) that should be reused rather than rolling our own.

src/input/Binding.zig
```diff
@@ -673,6 +673,28 @@ pub const Action = union(enum) {
...
+    /// The argument is the group number from 1 to 16. Each group is
+    /// outlined with a distinct border color (see
+    /// `broadcast-group-colors`). A terminal that is already in the given
+    /// group is removed from it instead (toggle), and a terminal in a
+    /// different group is moved to the given group.
+    ///
+    /// Terminals can also be toggled in and out of groups by clicking
+    /// them with the modifiers set by `broadcast-group-click-mods`
+    /// (by default cmd+shift+click on macOS, ctrl+shift+click on Linux).
+    toggle_broadcast_group: u8,
```
jcollie:
```suggestion
    toggle_broadcast_group: u4,
```

src/App.zig
```diff
@@ -557,6 +558,176 @@ pub fn findSurfaceByID(self: *const App, id: u64) ?*Surface {
...
+    }
+
+    const number: u8 = group - 1;
+    surface.broadcast_group = if (surface.broadcast_group == number)
+        null
+    else
+        number;
+    syncBroadcastGroup(rt_app, surface);
+    return true;
+}
```
jcollie, line 621:
```suggestion
pub fn toggleNumberedBroadcastGroup(
    self: *App,
    rt_app: *apprt.App,
    surface: *Surface,
    group: u4,
) bool {
    _ = self;

    surface.broadcast_group = if (surface.broadcast_group == group)
        null
    else
        number;
    syncBroadcastGroup(rt_app, surface);
    return true;
}
```

src/Surface.zig
```diff
@@ -158,6 +158,12 @@ focused: bool = true,
 /// visible so reporting remains conservative.
 visible: bool = true,
 
+/// The broadcast input group this surface is a member of, or null if
+/// it isn't in a group. This is the zero-based group index (the
+/// user-facing group number is one higher), which doubles as the
+/// index into the broadcast-group-colors config. Managed by App.
+broadcast_group: ?u8 = null,
+
```
jcollie, line 165:
```suggestion
/// The broadcast input group this surface is a member of, or null if
/// it isn't in a group. Managed by App.
broadcast_group: ?u4 = null,

```

src/apprt/gtk/class/surface.zig
```diff
@@ -1477,6 +1484,38 @@ pub const Surface = extern struct {
...
+        var buf: [32]u8 = undefined;
+        if (priv.broadcast_color) |old| {
+            priv.broadcast_overlay.removeCssClass(broadcastColorClass(&buf, old));
+        }
+        if (color) |v| {
+            priv.broadcast_overlay.addCssClass(broadcastColorClass(&buf, v));
+        }
+        priv.broadcast_color = color;
+        priv.broadcast_revealer.setRevealChild(@intFromBool(color != null));
+    }
```
jcollie:
```suggestion
    /// Update the broadcast group border overlay to match the given
    /// membership reported by the core.
    pub fn updateBroadcastGroup(
        self: *Self,
        value: apprt.action.BroadcastGroup,
    ) void {
        const priv = self.private();
        const group = value.group;

        // The border color comes from a broadcast-color-N CSS class
        // that the application generates from broadcast-group-colors,
        // so we swap the class for our previous group with the class
        // for the new one.
        var buf: [32]u8 = undefined;
        for (0..std.math.maxInt(u4)+1) |old| {
            priv.broadcast_overlay.removeCssClass(broadcastColorClass(&buf, @intCast(u4, old)));
        }
        if (group) |v| {
            priv.broadcast_overlay.addCssClass(broadcastColorClass(&buf, v));
            priv.broadcast_revealer.setRevealChild(1);
        } else {
            priv.broadcast_revealer.setRevealChild(0);
        }
    }
```

src/apprt/gtk/class/surface.zig
jcollie: We don't want to duplicate the group state here, just remove any possible `broadcast-color-X` class any time we need to change.

macos/Sources/Ghostty/Ghostty.App.swift
```diff
@@ -638,6 +638,9 @@ extension Ghostty {
             case GHOSTTY_ACTION_COPY_TITLE_TO_CLIPBOARD:
                 return copyTitleToClipboard(app, target: target)
 
+            case GHOSTTY_ACTION_BROADCAST_GROUP:
+                return broadcastGroup(app, target: target, v: action.action.broadcast_group)
+
```
jcollie:
```suggestion
            case GHOSTTY_ACTION_SYNC_BROADCAST_GROUP:
                return syncBroadcastGroup(app, target: target, v: action.action.broadcast_group)

```

jcollie (changes requested):
OK, lotsa stuff here, but to summarize:

1. I really dislike the difference between 1-based group numbers and 0-based color indexes. Unifying everything on a 0-based index simplifies things, and lets us use a `u4` to represent the group. That should let the compiler catch any problems.
2. Action names should contain a verb so added `toggle` or `sync` in various places. Hopefully I caught all the necessary changes but the compiler may make a liar of me.
3. Added a `border-color` to the base CSS class which fixes a minor visual glitch.
4. The GTK surface shouldn't retain the group state, it's easy enough to just remove every custom CSS class as removing a non-existent CSS class is not an error.

src/apprt/gtk/css/style.css
dave92082: I was actually enjoying the flash on ungroup :) I get that it wasn't an intentional behavior though and agree it's probably not for everyone.

dave92082:
@jcollie All suggestions applied thanks for the thorough pass! A few small things came up along the way that weren't in the suggestions themselves:

  - The compiler did indeed make a liar of you in a couple spots 🙂  the toggleNumberedBroadcastGroup suggestion referenced a number variable that no longer exists so I used group, and the GTK one had a pre-0.11 @intCast(u4, old) in it.
  - The GTK surface method ended up as syncBroadcastGroup since that's what the application.zig suggestion called it
  - I also renamed the ghostty_action_u union member to sync_broadcast_group to match the tag like the other actions, so the Swift side reads action.action.sync_broadcast_group rather than the broadcast_group in the suggestion.
  - Fun one: switching the action payload to u4 turned up a gap in cli/args.zig — the tagged-union parser didn't know how to parse u4 fields so keybind parsing wouldn't compile. Added it to the int list, which is the one change outside the files you touched.

trag1c (changes requested):
Blocking on the new strings (we're at a string freeze for 1.4)

dave92082:
Updated the translation files so all checks should pass. (Did _not_ add translated strings)

jcollie:
I want to try this locally tonight before I approve, but this looks good!

jcollie (approved):
This is looking good to me, ready to merge after 1.4 is released.

PR #13748, opened 2026-08-11
macos: ignore -e arguments as open files

mitchellh:
Fixes #13319

AppKit treats existing positional arguments as documents, causing paths passed to a child command after -e to open an extra terminal surface.

We now process args ourselves during openFile callbacks to ignore  file paths after `-e`. There isn't a way to avoid this I can find because AppKit processes argc/argv from the main entrypoint and that can't be overridden.

### DIFF
--- a/macos/Sources/App/macOS/CommandLineOpenFileFilter.swift
+++ b/macos/Sources/App/macOS/CommandLineOpenFileFilter.swift
@@ -1,0 +1,56 @@
+import Foundation
+
+/// Filters the open-file events AppKit creates from command arguments following
+/// `-e`. Each matching event is consumed once so later requests to open the same
+/// file are handled normally.
+final class CommandLineOpenFileFilter {
+    private let workingDirectory: String
+    private var filesToIgnore: Set<String>
+
+    init(
+        arguments: [String],
+        workingDirectory: String,
+        fileExists: (String) -> Bool
+    ) {
+        self.workingDirectory = workingDirectory
+
+        guard let commandIndex = arguments.firstIndex(of: "-e") else {
+            self.filesToIgnore = []
+            return
+        }
+
+        // Ghostty treats every argument following `-e` as part of the child
+        // command. Existing paths in that suffix are the arguments AppKit can
+        // independently turn into open-file events.
+        self.filesToIgnore = Set(arguments[arguments.index(after: commandIndex)...]
+            .compactMap { argument in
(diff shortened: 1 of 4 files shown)


macos/Tests/CommandLineOpenFileFilterTests.swift
```diff
@@ -0,0 +1,64 @@
...
+                "/usr/bin/vim",
+                "./file.txt",
+                "../other.txt",
+                "missing.txt",
+            ],
+            workingDirectory: "/tmp/project",
+            fileExists: { existing.contains($0) }
+        )
+
+        #expect(!filter.shouldIgnore("/tmp/before.txt"))
```
mitchellh, line 36: Btw this is intentional right now. I couldn't find evidence in any terminal really how this should be handled. I think its harmless.

macos/Sources/App/macOS/CommandLineOpenFileFilter.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
+
+/// Filters the open-file events AppKit creates from command arguments following
+/// `-e`. Each matching event is consumed once so later requests to open the same
+/// file are handled normally.
+final class CommandLineOpenFileFilter {
+    private let workingDirectory: String
```
jparise, line 7: I think this is fine, but we should also consider using the `URL`-based file system paths more often.

jparise (approved):

macos/Sources/App/macOS/CommandLineOpenFileFilter.swift
mitchellh: Makes sense generally. Since the surrounding AppKit and FileManager APIs use path strings here, I’d prefer to keep this change as-is. I don't think URL buys us anything?

jparise: Not in this case because the inputs are already `String`, but the working directory could be a `URL` with internal conversion. (This is already doing `as NSString` to get the to older `NSString`-based path routines that are no longer available on `String` and have moved to `URL`.)

mitchellh: Okay fair enough, did this: https://github.com/ghostty-org/ghostty/commit/8c9fd7aa79c4d6cc768293fa6e3726162d00c618

PR #13767, opened 2026-08-12
Expose semantic prompt state through libghostty-vt

figelwump:
## Summary

- Add `GHOSTTY_TERMINAL_DATA_CURSOR_AT_PROMPT` to expose semantic prompt state through `libghostty-vt`.
- Reuse `Terminal.cursorIsAtPrompt()` rather than introducing separate prompt tracking.

## Motivation

Ghostty already tracks whether the cursor is at a semantic prompt through `Terminal.cursorIsAtPrompt()`. Exposing this query through `libghostty-vt` lets downstream terminal consumers use Ghostty's existing semantic prompt state without introducing separate tracking.

The query returns false when semantic prompt information is unavailable or the alternate screen is active.

Discussed in https://github.com/ghostty-org/ghostty/discussions/13755.

## Testing

- `zig build test-lib-vt`
- `zig build test-lib-vt -Dtest-filter="get cursor_at_prompt"`
- `zig build -Demit-lib-vt`

## AI disclosure

Codex assisted with adding the `libghostty-vt` query and test, addressing review feedback, running validation, and drafting this description. I reviewed and edited the final code and description, reviewed the validation results, and understand how the change interacts with Ghostty's semantic prompt state.

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -1134,7 +1134,6 @@ GHOSTTY_API ghostty_app_t ghostty_surface_app(ghostty_surface_t);
 GHOSTTY_API ghostty_surface_config_s ghostty_surface_inherited_config(ghostty_surface_t, ghostty_surface_context_e);
 GHOSTTY_API void ghostty_surface_update_config(ghostty_surface_t, ghostty_config_t);
 GHOSTTY_API bool ghostty_surface_needs_confirm_quit(ghostty_surface_t);
-GHOSTTY_API bool ghostty_surface_is_at_prompt(ghostty_surface_t);
 GHOSTTY_API bool ghostty_surface_process_exited(ghostty_surface_t);
 GHOSTTY_API void ghostty_surface_refresh(ghostty_surface_t);
 GHOSTTY_API void ghostty_surface_draw(ghostty_surface_t);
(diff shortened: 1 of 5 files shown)


mitchellh:
I'll accept the libghostty change but not the internal lib change.

bo2themax (approved):
header filer for macOS LGTM

mitchellh (changes requested):
Sorry forgot to hit the red button, my notes above.

figelwump:
Understood on not approving the internal lib change, but to clarify, `ghostty_surface_is_at_prompt` is the part Toastty actually uses. I added the libghostty counterpart based on the pre-PR feedback, but that API can’t query the terminal backing an existing `ghostty_surface_t` so it doesn't replace ghostty_surface_is_at_prompt for Toastty.

Is there another supported way for embedders using ghostty_surface_t to access this state, or is the expectation that apps like Toastty carry internal lib changes downstream? If the latter, all good, I’ll remove it and keep this PR scoped to libghostty-vt. (Checking on this because the internal lib addition would avoid requiring Toastty contributors to build against a downstream Ghostty fork.)

mitchellh:
> Understood on not approving the internal lib change, but to clarify, ghostty_surface_is_at_prompt is the part Toastty actually uses.

I understand that, but `libghostty-vt` is the only downstream-supported lib version of Ghostty. Otherwise, you're just maintaining a fork of our internal libraries and if we don't need it, I expect any fork to be their own maintainer.

> Is there another supported way for embedders using ghostty_surface_t to access this state, or is the expectation that apps like Toastty carry internal lib changes downstream?

Right, the latter, but I'll also note that I'm close to open sourcing a fully supported Swift framework for macOS and iOS on top of libghostty-vt that gives you something I think better than what our internal lib does (for reusability). This won't be an official Ghostty project, its a separate library on top of libghostty-vt that will be maintained by my company (Superlogical).

figelwump:
> Right, the latter, but I'll also note that I'm close to open sourcing a fully supported Swift framework for macOS and iOS on top of libghostty-vt that gives you something I think better than what our internal lib does (for reusability). This won't be an official Ghostty project, its a separate library on top of libghostty-vt that will be maintained by my company (Superlogical).

Oh nice, that's good to know, thanks.

figelwump:
Removed the internal lib changes

jcollie (approved):

PR #13779, opened 2026-08-13
gtk: implement GtkAccessibleText for the terminal surface

alex19EP:
This PR is a first step toward making Ghostty usable for blind users on Linux.

Screen reader accessibility was requested in
https://github.com/ghostty-org/ghostty/discussions/2351. I opened a vouch
request for this work in
https://github.com/ghostty-org/ghostty/discussions/13746, which jcollie
granted. The staged PR order described there starts with this
one.

### What this PR does

With this change the Orca screen reader can read the terminal viewport: the
visible text, the cursor position, hit-testing between screen points and text
offsets, per-row rectangles so flat review works, and change events as output
arrives so Orca can announce what changed.

### What it contains

- `src/apprt/a11y_text.zig` and `src/apprt/a11y_offsets.zig`: the snapshot
  walk and the offset math. Both are GTK-free and unit tested.
- `src/benchmark/A11yText.zig`: a benchmark for the snapshot walk and for the
  per-frame change probe.
- The `GtkAccessibleText` implementation on `GhosttySurface`.
- Application and window identity for AT-SPI, so the window is announced with
  a name and the right role instead of as an unnamed object.

### Performance

The per-frame work is a cheap change probe, not a rebuild: 6.7us at 30x80 and
17.6us at 60x200. A full snapshot rebuild costs 9.5us and 23.2us respectively
and only runs when the probe reports that the contents actually changed.

To reproduce, against a ReleaseFast build:

    ghostty-gen +styled --seed=42 --style-rate=0.3 | head -c 4000000 > corpus
    ghostty-bench +a11y-text --mode=probe --terminal-rows=30 \
        --terminal-cols=80 --loops=1000 --data=corpus

Those figures are hyperfine medians over 10 runs after 3 warmups.

The probe only runs once an AT client has queried the surface at least once,
and only on frames that surface actually renders, so an idle terminal and a
terminal nobody is inspecting both cost nothing.

I deliberately did not gate this on focus. With splits, that would silence a
build running in the other half of the window. Whether to speak an unfocused
terminal is the AT client's policy call, not ours to pre-empt by withholding
the events. There is a comment in `glareaRender` explaining this, and why
`org.a11y.Status.IsEnabled` is not the gate it looks like either.

### GTK versions

`get_extents` and `get_offset` were added to `GtkAccessibleTextInterface` in
GTK 4.16. `G_DEFINE_INTERFACE` allocates the vtable at the running GTK's
`sizeof`, so writing those two slots unconditionally would write past the
allocation on an older runtime. They are installed only when the running GTK
is 4.16 or newer. Everything else in the interface is 4.14, and the public
wrappers for those two null-check, so on 4.14 they simply do not participate.

### Testing

48 unit tests across the two new modules, 42 of them in `a11y_offsets.zig`.

The one that matters most is a 3000 iteration round-trip fuzz test. It replays
the change events we emit against a model of the AT client's own copy of the
text and asserts the two never diverge. That test is why I trust the diff.

Beyond the unit tests I have been driving this daily with Orca and a braille
display.

### One thing I cannot test myself

I am blind, so I cannot verify mouse behaviour.

The change I would most like a sighted reviewer to look at is
`src/apprt/gtk/ui/1.2/surface.blp`: focus and both event controllers move from
the inner `GLArea` up to `GhosttySurface`. That move is what makes the surface
an accessible focus target, but it is also the part of this diff most likely
to break things for sighted users. The four things need checking:

1. Clicking a surface focuses it.
2. Click-drag selection copies to PRIMARY.
3. Middle-click pastes.
4. Clicking an unfocused split focuses it.

### What is not in this PR

Selection, text attributes, and links, in that order.

### Test harness

I have an end-to-end harness on my `gtk-accessibility-poc` branch: pytest
driving a real Ghostty window over AT-SPI under Xvfb. 41 tests, 36 of them
passing against this branch; the 5 failures are all features deferred to the
later PRs (attributes, selection, set_caret).

I left it out here because it needs a display server, so it would have to be
opt-in and disabled in CI, the same shape as the existing macOS XCUITest
suite. Happy to submit it separately if you want it.

## AI Usage Disclosure

This change was developed with heavy assistance from Claude Code (Anthropic). I direct the design and behavior and Claude Code writes most of the Zig under my review. All code was reviewed, built, and tested by me before submission, including daily use with Orca and a braille display.

### DIFF
--- a/src/a11y/offsets.zig
+++ b/src/a11y/offsets.zig
@@ -1,0 +1,1337 @@
+//! Offset math over the flat accessibility text snapshot produced by
+//! `text.build`.
+//!
+//! Where `text.zig` produces the snapshot, this module navigates it:
+//! converting between byte offsets, UTF-8 codepoint offsets, terminal
+//! grid positions and widget-space points, and computing the minimal
+//! diff between two snapshots.
+//!
+//! It is deliberately free of GTK and of any surface state. Every
+//! function takes the snapshot text and returns numbers, so the offset
+//! arithmetic can be tested directly. That matters more here than
+//! anywhere else in the accessibility code: AT-SPI offsets are in
+//! codepoints, not bytes, and every historical crash in this subsystem
+//! has been a boundary or unit mix-up. Handing an orphan UTF-8
+//! continuation byte to the AT-SPI bridge makes `g_variant_new_string`
+//! return NULL and SIGSEGVs inside `g_variant_builder_add_value`; being
+//! off by a codepoint in the other direction silently truncates what a
+//! screen reader announces.
+//!
+//! Two units appear throughout and are never interchangeable:
+//!
+//!   - `usize` is an offset *into the snapshot*, either bytes or
+//!     codepoints. Parameter and field names say which (`..._byte`,
+//!     `..._cp`); an unsuffixed offset in a public signature is a
+//!     codepoint offset, because that is what AT-SPI speaks.
+//!   - `u32` is a *terminal grid* coordinate: a row or a column, counted
+//!     in cells.
+//!
+//! Nothing here uses C integer types. The apprt casts at the boundary
+//! where it hands numbers to GTK, which keeps that conversion in one
+//! reviewable place instead of spread across the arithmetic.
+
+const std = @import("std");
+const simd = @import("../simd/main.zig");
+
+/// Byte length of a UTF-8 codepoint given its leading byte. A total
+/// wrapper over `std.unicode.utf8ByteSequenceLength`: stray continuation
+/// bytes advance 1 so a scan over malformed input can't stall. These
+/// helpers are total (rather than using e.g.
+/// `std.unicode.utf8CountCodepoints`) because their call sites are AT
+/// callbacks with no useful error path, and the snapshots they read are
+/// built codepoint by codepoint and always well-formed.
+pub fn utf8CpLen(b: u8) usize {
+    return std.unicode.utf8ByteSequenceLength(b) catch 1;
+}
+
+/// Count UTF-8 codepoints in `s`, via simdutf when the build has SIMD
+/// enabled.
+pub fn utf8CpCount(s: []const u8) usize {
+    return simd.countUtf8(s);
+}
+
+/// Byte offset of the `cp_idx`-th codepoint in `s`. Clamps at `s.len`.
+pub fn utf8CpToByte(s: []const u8, cp_idx: usize) usize {
+    var i: usize = 0;
+    var c: usize = 0;
+    while (i < s.len and c < cp_idx) : (c += 1) i += utf8CpLen(s[i]);
+    return i;
+}
+
+/// The unchanged ends of two snapshots: the leading `prefix_len` bytes
+/// and the trailing `suffix_len` bytes are common to both, and everything
+/// between them was replaced. Both lengths land on UTF-8 codepoint
+/// boundaries in either snapshot.
+pub const PrefixSuffixDiff = struct {
+    prefix_len: usize,
+    suffix_len: usize,
+
+    /// Bytes each side has to be told about: what leaves `old` plus what
+    /// arrives from `new`.
+    pub fn cost(self: PrefixSuffixDiff, old_len: usize, new_len: usize) usize {
+        const unchanged = self.prefix_len + self.suffix_len;
+        return (old_len - unchanged) + (new_len - unchanged);
+    }
+};
+
+/// Compute the byte-wise common prefix and suffix of `old` and `new`,
+/// then pull both offsets back to UTF-8 codepoint boundaries.
+///
+/// Alignment matters because two multi-byte codepoints can share a
+/// leading byte (e.g. `│` and `├`, both starting with 0xE2 0x94), and a
+/// byte-wise compare can land inside a character.
+///
+/// Aligning against `old_text` alone also aligns `new_text`: the bytes at
+/// the two boundaries are equal by construction, and in well-formed UTF-8
+/// a non-continuation byte can only ever begin a codepoint. So a boundary
+/// that is valid in one snapshot is valid in the other.
+pub fn prefixSuffix(old_text: []const u8, new_text: []const u8) PrefixSuffixDiff {
+    var prefix: usize = 0;
+    const min_len = @min(old_text.len, new_text.len);
+    while (prefix < min_len and old_text[prefix] == new_text[prefix]) : (prefix += 1) {}
+
+    // Cap the suffix so it can't overlap the prefix (that would make the
+    // remove/insert ranges go negative).
+    var suffix: usize = 0;
+    const max_suffix = @min(old_text.len - prefix, new_text.len - prefix);
+    while (suffix < max_suffix and
+        old_text[old_text.len - 1 - suffix] == new_text[new_text.len - 1 - suffix]) : (suffix += 1)
+    {}
+
+    // Retreat each end onto a codepoint boundary. Both only ever shrink,
+    // so the no-overlap cap above still holds afterwards.
+    //
+    // The `prefix < old_text.len` guard matters when `old_text` is a
+    // prefix of `new_text`: `prefix == old_text.len` and indexing would go
+    // out of bounds, but end-of-buffer is already a codepoint boundary.
+    while (prefix > 0 and prefix < old_text.len and
+        isContinuation(old_text[prefix])) : (prefix -= 1)
+    {}
+    while (suffix > 0 and isContinuation(old_text[old_text.len - suffix])) : (suffix -= 1) {}
+
+    return .{ .prefix_len = prefix, .suffix_len = suffix };
+}
+
+/// Whether `b` is a UTF-8 continuation byte, i.e. a byte that cannot
+/// begin a codepoint.
+fn isContinuation(b: u8) bool {
+    return b & 0xC0 == 0x80;
+}
+
+/// Look for a whole-line shift: a K > 0 at a `\n` boundary of `old` such
+/// that `new[0..|old|-K] == old[K..]`. Returns 0 if no such K exists.
+/// Since `\n` is ASCII, every candidate K is already a UTF-8 codepoint
+/// boundary.
+///
+/// As written this finds an *upward* shift: K bytes left the top and the
+/// remainder moved up. The downward shift is the same relation with the
+/// snapshots exchanged, so `chooseDiff` finds it by calling this with the
+/// arguments swapped rather than by duplicating the scan.
+pub fn scrollK(old_text: []const u8, new_text: []const u8) usize {
+    var search: usize = 0;
+    while (std.mem.indexOfScalarPos(u8, old_text, search, '\n')) |i| {
+        search = i + 1;
+        const boundary = i + 1;
+        const remainder = old_text.len - boundary;
+        // A zero-byte remainder matches trivially at every trailing `\n`
+        // and would mask the prefix/suffix diff for every change where
+        // old_text ends in `\n`. Require a non-trivial middle.
+        if (remainder == 0) continue;
+        // Early rows have a remainder too long to fit in `new_text`. The
+        // remainder shrinks as the scan advances, so this is a skip and
+        // not a stop.
+        if (remainder > new_text.len) continue;
+        if (std.mem.eql(u8, new_text[0..remainder], old_text[boundary..])) {
+            // The first match is the smallest K, which is the one that
+            // describes the change in the fewest bytes.
+            return boundary;
+        }
+    }
+    return 0;
+}
+
+/// How to describe a viewport change to an AT client.
+pub const Diff = union(enum) {
+    /// Nothing to announce.
+    none,
+
+    /// The text shifted up by whole lines: `k` bytes left the top of the
+    /// viewport, everything below moved up to fill the gap, and new
+    /// content arrived at the bottom. This is the shape of command
+    /// output, and of scrolling forwards towards the prompt.
+    shift_up: usize,
+
+    /// The text shifted down by whole lines: `j` bytes of new content
+    /// arrived at the top and the tail fell off the bottom. This is the
+    /// shape of scrolling back into the scrollback.
+    shift_down: usize,
+
+    /// Anything else: replace the bytes between the common prefix and the
+    /// common suffix.
+    replace: PrefixSuffixDiff,
+};
+
+/// Pick the cheapest description of the change from `old_text` to
+/// `new_text`, measured in bytes the AT client has to be told about.
+///
+/// Cheapest-wins is what keeps the three shapes from poaching each
+/// other's cases. On a real whole-line shift, prefix/suffix covers nearly
+/// the whole viewport (every row moved), so a shift wins by a wide
+/// margin. But shift detection also matches spuriously on typing echo:
+/// typing `x` at a `$ ` prompt when earlier rows also end in `$ ` makes
+/// the last row of `old` a prefix of `new`, and there the shift is the
+/// expensive one. Neither side needs to know about the other; the byte
+/// count separates them.
+///
+/// A shift only wins by being strictly cheaper, so an exact tie with
+/// prefix/suffix takes the replacement. Between the two shift directions
+/// a tie goes to `shift_up`, which is by far the common case, since a terminal
+/// spends most of its life scrolling forwards.
+pub fn chooseDiff(old_text: []const u8, new_text: []const u8) Diff {
+    const ps = prefixSuffix(old_text, new_text);
+    const ps_total = ps.cost(old_text.len, new_text.len);
+
+    // Bytes off the top, plus whatever `new` grew past what survived.
+    //
+    // The subtractions below cannot underflow: a non-zero K from
+    // `scrollK(a, b)` carries the guarantee that `a.len - K <= b.len`,
+    // which is exactly what each one needs.
+    const k = scrollK(old_text, new_text);
+    const up_total = if (k > 0)
+        k + (new_text.len - (old_text.len - k))
+    else
+        std.math.maxInt(usize);
+
+    // The mirror image: `j` bytes onto the top of `new`, plus whatever of
+    // `old` fell off the bottom.
+    const j = scrollK(new_text, old_text);
+    const down_total = if (j > 0)
+        j + (old_text.len - (new_text.len - j))
+    else
+        std.math.maxInt(usize);
+
+    if (k > 0 and up_total < ps_total and up_total <= down_total) {
+        return .{ .shift_up = k };
+    }
+    if (j > 0 and down_total < ps_total) return .{ .shift_down = j };
+    if (ps_total > 0) return .{ .replace = ps };
+    return .none;
+}
+
+/// How many terminal columns each codepoint of a snapshot occupies.
+///
+/// The snapshot holds one codepoint per *character*, but a terminal row is
+/// a grid of *cells*, and the two do not correspond: a double-width
+/// character (CJK, an emoji in emoji presentation) is one codepoint across
+/// two columns, while the combining marks of a grapheme cluster are
+/// codepoints across no columns of their own. Equating codepoint index
+/// with column, which every function here used to do, puts a caret, a
+/// click, or a highlight one column left per wide character preceding it
+/// on the row.
+///
+/// Nothing in the text records that, so `text.build` records it here
+/// as it walks the cells. The widths therefore come from the same cell
+/// data the renderer draws from, rather than from a width table this
+/// module would have to keep in sync with the terminal's own idea of how
+/// wide a character rendered, which depends on the configured
+/// grapheme-width method and cannot be re-derived from the text alone.
+pub const CellWidths = struct {
+    /// Columns occupied by the i-th codepoint of the snapshot. Indices
+    /// past the end read as 1, so a short or absent slice degrades to one
+    /// column per codepoint instead of going out of bounds.
+    per_cp: []const u8 = &.{},
+
+    /// Widths for a snapshot whose cell data we don't have. Exact for
+    /// all-single-width text and off by a column per wide character
+    /// otherwise, so pass it only where an approximate answer beats no
+    /// answer at all.
+    pub const uniform: CellWidths = .{};
+
+    pub fn at(self: CellWidths, cp_idx: usize) u32 {
+        if (cp_idx >= self.per_cp.len) return 1;
+        return self.per_cp[cp_idx];
+    }
+};
+
+/// Where a row begins, in both units we need to walk it: bytes to index
+/// `text`, codepoints to index `CellWidths`.
+const RowStart = struct {
+    byte: usize,
+    cp: usize,
+};
+
+/// Locate the start of `row`, or null when the snapshot has fewer rows.
+fn rowStart(text: []const u8, row: u32) ?RowStart {
+    if (row == 0) return .{ .byte = 0, .cp = 0 };
+
+    var byte: usize = 0;
+    var cp: usize = 0;
+    var seen: u32 = 0;
+    while (byte < text.len) {
+        const is_newline = text[byte] == '\n';
+        byte += utf8CpLen(text[byte]);
+        cp += 1;
+        if (!is_newline) continue;
+        seen += 1;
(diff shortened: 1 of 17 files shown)


alex19EP:
hello. i addressed copilot revues and fixed typos found by CI. can someone approve workflow?

bo2themax:
> hello. i addressed copilot revues and fixed typos found by CI. can someone approve workflow?

Is the description of this written by you or AI? Looks suspicious and violating our AI policy.

Checks workflow approved.

src/apprt.zig
```diff
@@ -56,4 +56,6 @@ test {
     _ = runtime;
     _ = action;
     _ = structs;
+    _ = @import("apprt/a11y_text.zig");
```
bo2themax: I don't think we need to import them here. I also had this with ai generated changes before. 

You can move the related files down below with the benefit of zig0.16

alex19EP:
> > hello. i addressed copilot revues and fixed typos found by CI. can someone approve workflow?
> 
> Is the description of this written by you or AI? Looks suspicious and violating our AI policy.
> 
> Checks workflow approved.

Hello. by me. But I am a Russian speaker so I sometime using translator (deepl) service to translate my Russian text to English and fix some grammar. i can stop Doing that if it is suspicious.

src/apprt.zig
alex19EP: yep tested will remove them.

> You can move the related files down below with the benefit of zig0.16

This I don't understand Do you mean that I should change these files hierarchy? move them to another directory?

src/apprt/gtk/class/surface.zig
```diff
@@ -695,6 +713,38 @@ pub const Surface = extern struct {
 
         action_group: ?*gio.SimpleActionGroup = null,
 
+        // Accessibility state for GtkAccessibleText
```
pluiedev: IMO there's too much scattered state within the surface's private fields that should logically belong together. Make a new struct with all of the accessibility state and move all the related logic there

src/apprt/gtk/class/surface.zig
```diff
@@ -2005,6 +2055,20 @@ pub const Surface = extern struct {
         for (priv.key_tables.items) |s| alloc.free(s);
         priv.key_tables.deinit(alloc);
 
+        // Accessibility state. These use `std.heap.c_allocator` rather
+        // than `alloc` above because their contents are handed to GTK,
+        // which may still be holding them when we get here.
+        priv.ax_probe_buf.deinit(std.heap.c_allocator);
```
pluiedev: Don't use `std.heap.c_allocator` when we already have access to the application's overall allocator.

src/apprt/gtk/class/surface.zig
```diff
@@ -2622,6 +2686,708 @@ pub const Surface = extern struct {
...
+        // compile-time one: our bindings always declare these fields, but
+        // `G_DEFINE_INTERFACE` allocates the vtable at the running GTK's
+        // own `sizeof(GtkAccessibleTextInterface)`. Against a 4.14
+        // runtime the two slots are not part of that allocation and
+        // writing them scribbles past the end of it.
+        //
+        // Leaving them unset on 4.14 costs per-range geometry, so a
+        // screen reader's flat review has no coordinates to group rows by
+        // and falls back to reading the object as a whole. Degraded, but
+        // it reads.
```
pluiedev: Far too much LLM yapping.


```suggestion
        // `get_extents` and `get_offset` are 4.16 additions.
        // While they are accessible with current bindings, we need to make sure
        // we do not access invalid memory when running with older GTK versions
        // by writing to fields that don't yet exist.
```

src/apprt/gtk/class/surface.zig
```diff
@@ -2622,6 +2686,708 @@ pub const Surface = extern struct {
...
+    /// the same rect, all lines collapse into one "line" and flat review can't
+    /// step through them. We compute a per-line rect from the terminal's cell
+    /// dimensions: Y = row_index * cell_height, height = cell_height.
+    fn axGetExtents(
+        self_opaque: *gtk.AccessibleText,
+        start: c_uint,
+        end: c_uint,
+        extents: *graphene.Rect,
+    ) callconv(.c) c_int {
+        const self: *Self = @ptrCast(@alignCast(self_opaque));
```
pluiedev: Do a safe check instead.


```suggestion
        const self = gobject.ext.cast(Self, self_opaque) orelse return 0;
```

src/apprt/gtk/class/surface.zig
```diff
@@ -2622,6 +2686,708 @@ pub const Surface = extern struct {
...
+    /// gtk-xft-dpi font scale, which changes how large a cell *is* but not
+    /// which coordinate space it is measured in.
+    ///
+    /// At 1x the two spaces coincide, which is what kept this hidden: every
+    /// rect we reported was already correct, and the test harness runs under
+    /// Xvfb at 1x. On a scaled display an unconverted rect is inflated by the
+    /// scale factor, and Orca intersects each line's rect with the widget's
+    /// allocation to decide which lines are on screen. Inflated rows run out
+    /// of the box early, so flat review silently loses the bottom of the
+    /// viewport, including the row the prompt is on.
```
pluiedev: Wayyyyyyyyyyyyy too much yapping. You need to tell the LLM to output less text and condense the logic more.

src/apprt/gtk/class/surface.zig
```diff
@@ -2622,6 +2686,708 @@ pub const Surface = extern struct {
...
+        // to frame boundaries rather than a time-based TTL: a TTL would
+        // let the cache refresh mid-read (e.g. between
+        // `get_character_count` and `iter_line` during an Orca flat review
+        // pass), silently shifting offsets under the client and causing
+        // lines past the first "short" snapshot length to disappear.
+        if (priv.ax_cached_text != null) return priv.ax_cached_text;
+
+        const core_surface = priv.core_surface orelse return null;
+
+        const alloc = std.heap.c_allocator;
```
pluiedev: Again use application's allocator and not the C allocator

pluiedev (changes requested):
I don't know. I haven't come close to reviewing all of this but so far I'm pretty disappointed by the quality. I love that we're finally trying to get Ghostty working with Assistive Technology but this PR isn't really up to my standards.

src/apprt/gtk/class/surface.zig
jcollie: Definitely agree this needs to be moved to a new struct. `surface.zig` is already too big.

alex19EP: done

src/apprt/gtk/class/surface.zig
alex19EP: switched to Application.default.allocator

src/apprt/gtk/class/surface.zig
alex19EP: done.

I tried to strike a balance between the comments that are already in the codebase—which really help me, as a beginner, read the Zig code—and the fact that the AI does generate unnecessarily long comments. I addressed your specific comment and reviewed the rest of the comments again. I made changes where I thought it was appropriate.

src/apprt/gtk/class/surface.zig
alex19EP: done. Thank you for recommendation

src/apprt/gtk/class/surface.zig
alex19EP: fixed. and treamed the other comments too

src/apprt/gtk/class/surface.zig
alex19EP: done

alex19EP:
i think i Addressed all the review feedback for now. please tell me if I overlooked something. I'm also updated pull request description to be more transparent about AI usage.

src/apprt/gtk/class/surface_a11y.zig
```diff
@@ -0,0 +1,635 @@
...
+/// needs NUL termination (despite the `get_contents` docs) or it reads
+/// past the allocation. Invalid UTF-8 becomes an empty result, logged:
+/// glib would otherwise substitute "[Invalid UTF-8]", which Orca speaks.
+fn bytesNulTerm(slice: []const u8) *glib.Bytes {
+    if (std.unicode.utf8ValidateSlice(slice)) {
+        const alloc = Application.default().allocator();
+        const buf = alloc.alloc(u8, slice.len + 1) catch return emptyBytes();
+        defer alloc.free(buf);
+        @memcpy(buf[0..slice.len], slice);
+        buf[slice.len] = 0;
```
pluiedev:
```suggestion
        const buf = alloc.dupeSentinel(u8, slice, 0) catch return emptyBytes();
```

src/apprt/gtk/class/surface_a11y.zig
```diff
@@ -0,0 +1,635 @@
...
+        // g_bytes_new copies, so `buf` can be freed when this returns.
+        return glib.Bytes.new(buf.ptr, slice.len + 1);
+    }
+
+    // We slice on codepoint boundaries, so getting here means the
```
pluiedev: I'm not really sure how helpful this is? Invalid UTF-8 should be very very rare and even if this does get caught it's not very visible to the end user

src/a11y/offsets.zig
```diff
@@ -0,0 +1,1370 @@
...
+/// built codepoint by codepoint and are always well-formed.
+pub fn utf8CpLen(b: u8) usize {
+    if (b < 0xC0) return 1;
+    if (b < 0xE0) return 2;
+    if (b < 0xF0) return 3;
+    return 4;
+}
+
+/// Count UTF-8 codepoints in `s`.
+pub fn utf8CpCount(s: []const u8) usize {
```
pluiedev, line 49: Why not just https://ziglang.org/documentation/master/std/#std.unicode.utf8CountCodepoints? In fact this is reimplementing a *lot* of what `std.unicode` already does. We also already use `simdutf` for faster UTF-8 processing so it should be a good idea to expose a few more functions there to here

src/a11y/offsets.zig
```diff
@@ -0,0 +1,1370 @@
+//! Offset math over the flat accessibility text snapshot produced by
```
pluiedev, line 1: I'm not sure this belongs in `src/apprt` either. Most of this is apprt-independent and therefore should either live in `src/terminal` or a dedicated folder (`src/a11y`)

pluiedev (changes requested):
Thanks, it's looking a lot better now. Still a couple of issues, though

src/apprt/gtk/class/surface_a11y.zig
alex19EP: applyed thanks.

src/apprt/gtk/class/surface_a11y.zig
alex19EP: the check itself prevents a crash, cince invalid UTF-8 makes g_variant_new_string return NULL and the AT-SPI bridge segfaults on it. I hit this during development. But the big hex dump was endeed unneaded, it is now a one line warning and the helper is deleted.

src/a11y/offsets.zig
alex19EP: Indeed good ideas, thanks. Done both.

utf8CpLen is now a wrapper over std.unicode.utf8ByteSequenceLength. We cannot use the std counting functions like utf8CountCodepoints directly because they return errors and the accessibility callbacks don't have a good error path.

Regarding simdutf, we now use it for utf8CpCount when it is available (scalar fallback otherwise). I benchmarked it: 44-59x faster than the scalar loop on viewport sized buffers (20-83ns vs 0.9-4.9us). So definitely thank you for pushing me in the right direction here.

src/a11y/offsets.zig
alex19EP: Fully agree. moved to a new src/a11y package (main.zig, text.zig, offsets.zig). I picked src/a11y over src/terminal because the units here (AT-SPI codepoint offsets, widget space points) are accessibility concepts, not terminal ones, and later PRs in this series (links, attributes) will add more runtime independent logic there.

alex19EP:
Hello thank you for new feedback I think I addressed all of this.

pushed changes.

tristan957:
Part of me wonders whether we could split this PR up a bit? The g_set_prgname()/g_set_application_name(). The focus tracking. The accessibility roles. Then finally the wiring of the surface.

Thanks for taking this on. Accessibility is very important. Silly question, but I wonder if there are any integration tests that we can run to make sure this doesn't regress? Would it require launching an instance in a headless compositor and then querying the AT-SPI dbus APIs?

alex19EP:
> Part of me wonders whether we could split this PR up a bit? The g_set_prgname()/g_set_application_name(). The focus tracking. The accessibility roles. Then finally the wiring of the surface.

so more small commits? i can do that yep.

> Thanks for taking this on. Accessibility is very important.

as a blind person wich Relies on accessibility every day. I'm so happy to hear this.

> Silly question, but I wonder if there are any integration tests that we can run to make sure this doesn't regress? Would it require launching an instance in a headless compositor and then querying the AT-SPI dbus APIs?

yep that's how i creatyed this tests in my dev testing branch you can look on how it is done https://github.com/alex19EP/ghostty/tree/gtk-accessibility-poc/test/gtk-a11y

i am Happy to think about more cleaner solution for upstreaming this.

tristan957:
Regarding testing, I see that there are many AT-SPI related packages on PyPI. Maybe we should be build a little AT-SPI client (in zig?) that we can then use to query the application for related state.

Out of curiosity, have you found any problems with using tab to navigate the application at all, including the command palette?

alex19EP:
> Regarding testing, I see that there are many AT-SPI related packages on PyPI. Maybe we should be build a little AT-SPI client (in zig?) that we can then use to query the application for related state.

I am open to zig client creation. AT-SPI ships Atspi-2.0.gir, so zig-gobject can generate bindings for it just like the rest of the GTK stack, no hand written D-Bus needed.

one problem to keep in mind: my python harness goes through libatspi, which is the same client stack Orca itself uses. that already caught real client side quirks that a fresh client would not reproduce. so even with a zig client I would want the tests to stay close to what Orca actually does.

also the hard part is not really the client language. it is the environment: the a11y registry has to be started manually on a private bus, the app must have real input focus, and rendering i think needs software GL under a headless X server. that I think is the same reason the macOS XCUITest suite exists but is kept out of CI, but i am not sure. I would suggest we sort this out in a follow-up PR about the test
harness and not gate this one on it.

> Out of curiosity, have you found any problems with using tab to navigate the application at all, including the command palette?

yes. in the command palette every command row is its own tab stop. ten tab presses land on ten different list items, focus never leaves the list and never reaches the close button. with the full command list that is a lot of pressing. the rows also have an empty accessible name, the title label is just a
plain child inside the row, so a screen reader has nothing proper to announce when it lands there. orca has A heuristic to Overcome these type of problem and read proper Label but it is better to properly fix it of course.

Both are easy fixes and I plan to do it in the next pull requests.

src/a11y/text.zig
```diff
@@ -0,0 +1,438 @@
...
+    }
+};
+
+/// Walk the viewport of `screen` and append its text to `buffer`.
+///
+/// `buffer` is appended to, not cleared, so callers reusing a buffer must
+/// clear it themselves.
+pub fn build(
+    alloc: Allocator,
+    buffer: *std.ArrayList(u8),
```
pluiedev: More idiomatic Zig here would use a `*std.Io.Writer` so this function remains reusable for all buffer impls, but it's a fairly minor detail

pluiedev (approved):
I haven't checked the logic in src/a11y super closely but it looks pretty convincing. GTK looks clean. Thank you so much for enduring the previous rounds of reviews—we're thrilled to see Ghostty become more accessible and appreciate the work you've poured into making it happen :heart:

src/a11y/text.zig
alex19EP: done. I changed it as a new commit so you can review it in one piece Happy to rebase and Incorporate this commit to my main work after you take a look if you want.

And thank you for reviews and kind words.

alex19EP:
i split gtk changes to new commits as @tristan957 asked.

tristan957:
I think we should do some benchmarking on this. Christian Hergert added accessibility to VTE and he found there to be a drop in performance with accessibility enabled. The drop was large enough that accessibility is disabled by default in Ptyxis, if I remember correctly.

I see you already did some performance work, but it isn't clear if there is a degradation in performance. Maybe I cannot read!

alex19EP:
performance was one of my concerns from the start (Exactly because I'm a long time VTE user). that is why this PR ships a benchmark commit (ghostty-bench +a11y-text) next to the feature and why the change events are gated the way they are.

the cost has three levels:

1. no AT client: the per frame hook sets one boolean and returns. nothing is walked. the code only becomes active after a real AT-SPI client reads the surface once. so for users without a screen reader there is no measurable degradation, and no need for an off by default switch.

2. AT client attached, frame rendered but nothing changed. this is the number that matters most because it is paid on every rendered frame while orca is running, and it runs under the renderer mutex, so it is the one that could hurt io throughput. it is a text only walk into a retained buffer plus a memcmp. i used claude to measure it today with hyperfine, 100k iterations, noop mode subtracted: 4.0 us per frame at 30x80 and 12.6 us at 60x200. at 60 fps that is under 0.1 percent of the frame budget.

3. AT client attached and the frame changed: full rebuild with cell widths, 5.3 us at 30x80 and 15.2 us at 60x200, plus change events. the events are diffed to the smallest remove and insert pair, so a keystroke emits a few characters, not the whole viewport. emitting the whole viewport every frame is the classic way a terminal kills both performance and the screen reader, so that was gated from day one.

i asked claude to read And compare the VTE implementation With our accessibility implementation and here their reply verbatim:

> What VTE does (vteaccess-gtk4.cc, current master):
>
> _vte_accessible_text_init() is called from the widget's constructed() (in GTK4 builds with a11y compiled in) and connects to the terminal's contents-changed signal. Every emission of that signal runs vte_accessible_text_contents_snapshot(), which rebuilds the entire accessibility state:
>
> - the viewport text, via get_text_displayed_a11y(), which also fills a VteCharAttributes record (row, column, foreground and background color resolved per cell, underline, strikethrough) for every byte of the text (vte_char_attr_list_fill);
> - a byte-offset array with one entry per character;
> - a linebreak scan that touches one attribute record per character;
> - an O(n) loop over every character to locate the caret (vte_accessible_text_contents_find_caret);
> - a fresh GString and GBytes allocation each time.
>
> At 60x200 the attribute list alone is hundreds of kilobytes written per content change. The change events themselves are diffed to a minimal prefix/suffix range, so the event side is reasonable; the cost is in the snapshot that precedes the diff.
>
> The gate for all of this is vte_terminal_get_enable_a11y(), a config property whose default is true (vtegtk.cc, enable-a11y param spec). There is no check for whether an AT client is actually connected. GTK itself drops the event emission when no AT context is realized, but by that point the snapshot work has already been done. So with default settings, every VTE terminal pays
> the full snapshot cost on every content change even when nothing is listening on the accessibility bus. That is consistent with the performance drop Christian observed and with why a user-facing switch exists and Ptyxis ships with it off.
>
> How this PR differs:
>
> - The gate is client activity, not configuration. The per-frame hook stores one boolean until the first real AT-SPI read latches the surface active (frameRendered in surface_a11y.zig). No screen reader means no walk, ever, so no switch is needed.
> - The per-frame path while a reader is attached builds text only: no attributes, no color resolution, no per-character arrays, and no allocation in steady state (a retained probe buffer plus a memcmp).
> - The caret position falls out of the same walk that builds the text, instead of a separate O(n) search afterward.
> - Cell widths are recorded only on frames that actually changed, and text attributes are not collected at all in this PR. The follow-up that adds them computes them on demand inside the get_attributes query, where the AT asks about one position at a time, rather than per frame.
>
> The measured numbers above (4 to 13 us per unchanged frame) are the direct consequence of that structure: the per-frame work is a fraction of a percent of a 60 fps frame budget, and only exists while a screen reader is actually attached.

as a blind person who relies on accessibility every day, to be honest: i am not very comfortable with gating a11y behind a config option the way VTE does. it is obscure for the people who need it most, and IMHO it sends the wrong signal.

in defense of the client attachment approach: it is a well proven path. Chromium and Firefox both enable their accessibility engines only when an assistive technology client actually shows up, for the same reason.

alex19EP:
Is there something to do on my side or you just haven't had time to re-review yet? If time is a problem I certainly understand that and there is no pressure. I just want to ensure that there are now no blockers on my side.

pluiedev:
We're just blocked on a review from the terminal team as a matter of bureaucracy. I'll ask around if anyone is interested in a review

mitchellh:
I don't really love the root level a11y package. It has a bunch of logic in there that I think is generalizable at best and repeated at worst (like utf8 stuff, seems weird to just have rather generic utf8 stuff in there if it doesn't already exist elsewhere). 

I also think that this overlaps with our text formatter. It seems like its ultimately trying to achieve the same thing on the surface and I wonder if it would've been cleaner to retrofit a lot of the text caching and so on onto that. The macOS a11y work for example uses the formatter. It has bugs (in terms of missing formatter features), but it does the job. Feels sus to split the two.

I need to look deeper but thats my initial reaction. @alex19EP did you look into that at all already? Do you mind explaining rationale for not using those (even if its just that you didn't know they existed).

alex19EP:
hello @mitchellh thanks for looking into it.

yes, i looked at the formatter. it was the first thing i studied, through the macOS path (ghostty_surface_read_text -> dumpText -> selectionString -> ScreenFormatter). the snapshot walk should still be reference it in a comment: interior blanks are kept as spaces the same way ScreenFormatter keeps non trailing blanks.

the reason i did not build on it is that AT-SPI needs a different output contract than a formatted string:

1. line identity. in the snapshot, line i of the text is always viewport row i. an empty row still contributes its newline (there is a test for this). the formatter's plain mode elides blank rows until a non blank one follows and drops trailing blank rows completely, and selectionString hardcodes unwrap=true, so soft wrapped rows are joined into one line. all the offset math on top (per line extents, point to offset, caret routing, orca flat review reading "line n") relies on line == row.

2. units. AT-SPI speaks codepoint offsets and screen columns. the snapshot records one width byte per codepoint (2 for a wide cell, 0 for a combining mark, 1 otherwise), and only the cell walk can produce that, because it is what sees cell.wide. the formatter's answer to position mapping is PinMap/PointMap, which map byte to pin, are documented as a significant performance hit, answer the inverse question, and would still need byte to codepoint and pin to viewport row conversion on every AT query.

3. the caret. the cursor offset falls out of the same walk. the formatter can only express the cursor as a CUP escape in vt mode, so with plain output i would need an O(n) scan afterwards, which is exactly the VTE find_caret slow pattern from my previous comment.

4. the per frame probe writes into a retained buffer with no allocation while holding the renderer mutex. to be fair, this one is not fundamental: the formatter has a fast run path and could probably be made comparably fast (at least claude thinks so last time i explored it). the mismatch is the contract here.

retrofitting that onto the formatter means adding a11y only options to a general purpose tool used for copy, vt and html: keep blank rows, a widths sidecar, codepoint counting, a cursor out parameter. a self contained walk with its own tests and benchmark felt smaller and safer. but i am not attached to the split: if after a deeper look you would rather see this as a formatter mode (a "grid" emit variant), i am happy to rework it that way. the GTK caching layer does not care where the bytes come from.

on the utf8 point: agreed. the generic part is three small helpers at the top of offsets.zig  and utf8CpCount is just a wrapper over simd.countUtf8. i can move them into src/unicode right away. the rest of offsets.zig is a11y specific offset math (codepoint boundary aligned diffing, grid to offset mapping, granularity queries for get_string_at_offset). i checked and I think there is nothing similar elsewhere in the tree to reuse.

the package location is also not something i am attached to. the only hard requirement i Imagine. is that the walk stays outside the GTK class so it can be unit tested and benchmarked without a live GTK surface and an AT-SPI bus. root level src/a11y was my choice for that, but src/terminal or anywhere else works too. long term it could also serve the macOS side: the formatter string works there because VoiceOver pulls a whole string on demand, but for example accessibilityLine(for:) currently re-derives line structure by splitting that string on every query, which is the kind of work the snapshot plus offsets design exists to avoid.

If something isn't clear just... ask and I will try to reword it.

alex19EP:
@mitchellh  a friendly ping. Does anything apart from moving unicode-is needed on my side?

Issue #13766, opened 2026-08-12
Ghostty 1.4 Translation Tracking Issue


trag1c:
Hello translators! ❤️

This issue tracks the status of translations for the upcoming v1.4 release, planned for mid/late September. There are a lot more strings this time, mainly due to the entire command palette now being localized. With that in mind, partial translation PRs are perfectly fine!

As always, missing translations are not a release blocker, but the goal is to have as many locales up to date as possible. Localization team members are pinged as a heads-up, but **anyone can contribute and review translations.**

## Locales with missing translations

Locale  | Count | Maintainers                | PR
:-----: | :---: | :------------------------- | :---:
`ca`    | 181   | @ghostty-org/ca_ES         | #13776
`de`    | 181   | @ghostty-org/de_DE         | #13846
`ga`    | 181   | @ghostty-org/ga_IE         | #13837
`id`    | 181   | @ghostty-org/id_ID         | #13905
`it`    | 181   | @ghostty-org/it_IT         | #13996
`ja`    | 181   | @ghostty-org/ja_JP         | #14094
`lt`    | 181   | @ghostty-org/lt_LT         | #13777
`zh_CN` | 5     | @ghostty-org/zh_CN         | #13608
`be`    | 0     | @ghostty-org/be_BY         | #13771
`bg`    | 0     | @ghostty-org/bg_BG         | #13802
`es_AR` | 0     | @ghostty-org/es_AR         | #13784
`es_BO` | 0     | @ghostty-org/es_BO         | #13782
`es_ES` | 0     | @ghostty-org/es_ES         | #13800
`eu`    | 0     | @ghostty-org/eu_ES @ibaios | #14095
`fr`    | 0     | @ghostty-org/fr_FR         | #13971
`he`    | 0     | @ghostty-org/he_IL         | #13785
`hr`    | 0     | @ghostty-org/hr_HR         | #14019 #14194
`hu`    | 0     | @ghostty-org/hu_HU         | #13799
`kk`    | 0     | @ghostty-org/kk_KZ         | #13778
`ko_KR` | 0     | @ghostty-org/ko_KR         | #13815
`lv`    | 0     | @ghostty-org/lv_LV         | #11663
`mk`    | 0     | @ghostty-org/mk_MK         | #14088
`nb`    | 0     | @ghostty-org/nb_NO         | #13781
`nl`    | 0     | @ghostty-org/nl_NL         | #13853
[...]

PR #13785, opened 2026-08-13
i18n: update Hebrew translations for v1.4

slsrepo:
Update the Hebrew translations in po/he.po with the 181 new strings for v1.4, as requested in issue #13766 :)

### DIFF
--- a/po/he.po
+++ b/po/he.po
@@ -184,7 +184,7 @@ msgstr "כרטיסייה"
 #: src/apprt/gtk/ui/1.2/surface.blp:386 src/apprt/gtk/ui/1.5/window.blp:232
 #: src/apprt/gtk/ui/1.5/window.blp:339 src/input/command.zig:464
 msgid "Change Tab Title…"
-msgstr "שנה/י את כותרת הכרטיסייה…"
+msgstr "שינוי כותרת הכרטיסייה…"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:391 src/apprt/gtk/ui/1.5/window.blp:60
 #: src/apprt/gtk/ui/1.5/window.blp:110 src/apprt/gtk/ui/1.5/window.blp:237
@@ -203,7 +203,7 @@ msgstr "חלון"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:406 src/apprt/gtk/ui/1.5/window.blp:215
 msgid "Change Window Title…"
-msgstr "שנה/י את כותרת החלון…"
+msgstr "שינוי כותרת החלון…"
 
 #: src/apprt/gtk/ui/1.2/surface.blp:411 src/apprt/gtk/ui/1.5/window.blp:220
 #: src/input/command.zig:421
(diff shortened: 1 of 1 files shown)


po/he.po
```diff
@@ -392,710 +392,710 @@ msgstr "המפתחים של Ghostty"
...
 #: src/input/command.zig:155
 msgid "Copy to Clipboard"
-msgstr ""
+msgstr "העתקה ללוח ההעתקה"
 
 #: src/input/command.zig:156
 msgid ""
 "Copy the selected text to the clipboard in both plain and styled formats."
-msgstr ""
+msgstr "העתק/י את הטקסט הנבחר ללוח ההעתקה בפורמט טקסט פשוט ומעוצב."
```
CraziestOwl: I see you used 'שם הפעולה for most of them. Any reason not to use it in all of them?

slsrepo: The new command palette has entries for each action with titles for the actions that are being run and descriptions that contain a "Call to Action".
So wherever the string is the title of the action I used 'שם הפעולה' (for example 'שינוי כותרת המסוף') and when it is the description with the "Call to Action", I kept the currently used gender appropriate verbs (like 'שנה/י את כותרת המסוף').
The intention is for the titles to be short and simple, while keeping the description more personal and informative☺

slsrepo: Looking at the specific string you replied on, It seems an error slipped in, It should be either/or rather than and, so the end should be "בפורמטים של טקסט פשוט או מעוצב".

po/he.po
```diff
@@ -392,710 +392,710 @@ msgstr "המפתחים של Ghostty"
...
 
 #: src/input/command.zig:673
 msgid "Toggle whether mouse events are reported to terminal applications."
-msgstr ""
+msgstr "הפעל/י או בטל/י את הדיווח של אירועי עכבר לאפליקציות מסוף."
 
 #: src/input/command.zig:678
 msgid "Toggle Background Opacity"
-msgstr ""
+msgstr "שינוי של אטימות הרקע"
```
CraziestOwl, line 1061: maybe שקיפות instead of 'אטימות' 
in this context 'אטימות' doesn't sounds right to me.
I know they are technically  opposites ;p

po/he.po
CraziestOwl: sounds good to me. also good catch, I didn't noticed it haha.

po/he.po
slsrepo: Yeah, the accurate translation is 'אטימות', with 'שקיפות' being transparency, but I agree that since they are opposites using 'שקיפות' can still work while sounding better☺
The thing is, it breaks the description:
msgstr "שנה/י את אטימות הרקע של חלון שהתחיל כשקוף."
Double transparency sounds weird, should we change the ending?

CraziestOwl: You right it will be a bit awkward. But I think it sounds kind of weird either way 😅.
The original sentence is strange as it uses the word toggle, but also gives the starting state 

I think 'שנה/י את שקיפות הרקע ' might be enough. What do you think?

CraziestOwl: it seems to be on the toggle_background_opacity command:
```
        .toggle_background_opacity => comptime &.{.{
            .action = .toggle_background_opacity,
            .title = i18n.N_("Toggle Background Opacity"),
            .description = i18n.N_("Toggle the background opacity of a window that started transparent."),
        }},
```
Not sure why the added 'started transparent' ...

CraziestOwl: > it seems to be on the toggle_background_opacity command:
> 
> ```
>         .toggle_background_opacity => comptime &.{.{
>             .action = .toggle_background_opacity,
>             .title = i18n.N_("Toggle Background Opacity"),
>             .description = i18n.N_("Toggle the background opacity of a window that started transparent."),
>         }},
> ```
> 
> Not sure why the added 'started transparent' ...

doesn't matter I found the answer:
```
    /// Toggle the window background opacity. This only has an effect
    /// if the window started as transparent (non-opaque), and toggles
    /// it between fully opaque and the configured background opacity.
    toggle_background_opacity,
```

CraziestOwl: I consulted my Family haha, they suggested 'כהות' but I don't know if I would interrupt it as עכירות in that context

slsrepo: I think these fit in less and we should either keep it as is, or switch to 'שינוי שקיפות הרקע' and 'שנה/י את שקיפות הרקע' for the description.
I'm leaning more towards keeping it as is, to stay closer to the original English string.
Also, since the strings show one on top of the other, I think the description makes up for the title and make it sound ok.

What do you think?

CraziestOwl: Let's just leave it. While the title is confusing, you are right the description somewhat makes up for it

CraziestOwl (approved):

trag1c (approved):

PR #13787, opened 2026-08-13
macOS: avoid holding SurfaceView when sending notifications

bo2themax:
We shouldn't hold a closing surface view when sending notifications. This happens rarely, but it's the right thing to do.

### AI Disclosure

Found by Claude when judging other branches, I applied the changes myself.

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
@@ -1787,20 +1787,22 @@ extension Ghostty {
 
             // Note the callback may be executed on a background thread as documented
             // so we need @MainActor since we're reading/writing view state.
-            Task { @MainActor in
+            // We use [weak self] here because we don't want to extend the surface's
+            // lifetime when a notification is triggered right before the surface closes.
+            Task { @MainActor [weak self] in
                 do {
                     try await UNUserNotificationCenter.current().add(request)
 
                     // We need to keep track of this notification so we can remove it
                     // under certain circumstances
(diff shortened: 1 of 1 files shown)


macos/Sources/Ghostty/Surface View/SurfaceView_AppKit.swift
```diff
@@ -1787,20 +1787,22 @@ extension Ghostty {
...
+            // We use [weak self] here because we don't want to extend the surface's
+            // lifetime when a notification is triggered right before the surface closes.
+            Task { @MainActor [weak self] in
                 do {
                     try await UNUserNotificationCenter.current().add(request)
 
                     // We need to keep track of this notification so we can remove it
                     // under certain circumstances
-                    notificationIdentifiers.insert(uuid)
+                    self?.notificationIdentifiers.insert(uuid)
```
jparise, line 1798: Maybe `guard let self ...` here to simplify the rest of this block's calling patterns? Everything afterward is dependent on `self` being available.

bo2themax: We still need to remove the notification using uuid 3s later. I would imagine at least copy the check twice🙈

jparise: If that's the case, then isn't the `if self?.focused == true` condition going to fail and prevent the UUID-based removal from running anyway?

bo2themax: Oh yeah, you are right

PR #13789, opened 2026-08-13
pkg/wuffs: use C-only mirror of wuffs

### LINKED ISSUE #13788: Wuffs build fails on Nix nixpkgs-unstable because of ESET Antivirus flagging google wuffs test/data/ hippopotamus-bad-comment-length.jpeg as Trojan Win32/Exploit.MS04-028
### Discussed in https://github.com/ghostty-org/ghostty/discussions/13786

<sup>Originally posted by **KrappRamiro** August 13, 2026</sup>
# Issue
On systems where an ESET Antivirus inspects files as they're written to disk (e.g. ESET Endpoint with on-access protection), building Ghostty via Nix fails.

This happens because on the build, the [test/data/](https://github.com/google/wuffs/blob/main/test/data/) directory of **wuffs** is included. 
In that directory, there is a file called [hippopotamus-bad-comment-length.jpeg](https://github.com/google/wuffs/blob/main/test/data/artificial-jpeg/hippopotamus-bad-comment-length.jpeg) , a deliberately malformed JPEG created for testing.

ESET incorrectly flags that file as a Trojan (Win32/Exploit.MS04-028) , breaking the build.
 
From what i saw on [`pkg/wuffs/build.zig` line 78](https://github.com/ghostty-org/ghostty/blob/main/pkg/wuffs/build.zig#L78) , only [wuffs `release/c/`](https://github.com/google/wuffs/tree/main/release/c) folder is used.

That means the whole `test/data` folder is unnecesary, and could be excluded, fixing the issue.

# Suggested Fix

The Wuffs maintainer (@nigeltao) [suggested in google/wuffs#164](https://github.com/google/wuffs/issues/164#issuecomment-5260113571) that downstream consumers use [google/wuffs-mirror-release-c](https://github.com/google/wuffs-mirror-release-c) instead of the full [google/wuffs](https://github.com/google/wuffs) repo.

That mirror contains only the `release/c/` directory, which is what `pkg/wuffs/build.zig` uses

They said It's also smaller and changes less frequently.

Alternatively, the repackaged tarball at `deps.files.ghostty.org` could simply exclude the test/ directory when it's generated, since the build never reads from it (from what i could see in [`pkg/wuffs/build.zig`](https://github.com/ghostty-org/ghostty/blob/main/pkg/wuffs/build.zig)

# Relevant Info
Please see issue https://github.com/google/wuffs/issues/164 and also https://github.com/ghostty-org/ghostty/discussions/11976</div>

jcollie:
This prevents us from pulling in test images that trigger some anti-virus scanners. It's also smaller since it only has the necessary bits that we need.

This also updates to the latest release: 0.4.0-alpha.10.

### DIFF
--- a/build.zig.zon.json
+++ b/build.zig.zon.json
@@ -149,10 +149,10 @@
     "url": "https://deps.files.ghostty.org/wayland-protocols-258d8f88f2c8c25a830c6316f87d23ce1a0f12d9.tar.gz",
     "hash": "sha256-XO3K3egbdeYPI+XoO13SuOtO+5+Peb16NH0UiusFMPg="
   },
-  "N-V-__8AAAzZywE3s51XfsLbP9eyEw57ae9swYB9aGB6fCMs": {
+  "N-V-__8AAP5JWgCGP_AD0teWpa4krRvE9VPZzvviGdbmN4jI": {
     "name": "wuffs",
-    "url": "https://deps.files.ghostty.org/wuffs-122037b39d577ec2db3fd7b2130e7b69ef6cc1807d68607a7c232c958315d381b5cd.tar.gz",
-    "hash": "sha256-nkzSCr6W5sTG7enDBXEIhgEm574uLD41UVR2wlC+HBM="
+    "url": "git+https://github.com/google/wuffs-mirror-release-c.git?rev=v0.4.0-alpha.10#7411f488fe2e2c205c3d3b3d28638b7356522930",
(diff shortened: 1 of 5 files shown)


pluiedev (approved):
Makes sense

mitchellh:
Will fix the mirror later, thanks

KrappRamiro:
Thank you a lot!

@mitchellh Could you please `@ping` me when you fix the mirror later? Thank you very much

PR #13830, opened 2026-08-15
libghostty: attacking wasm binary size

mitchellh:
This should release our ReleaseFast bundle from 1.1MB to ~800KB. 

The major win is disabling logging in ReleaseFast wasm builds (~200KB).

The next is removing aggressive inlining in paths that don't make sense for performance. Verified with benchmarks on native to not affect anything really.

The third was really dumb: `var buf: [4096]u32 = @splat(c)` in LLVM releasefast for wasm was lowering to 4096 separate `i32.store`... like... 30KB of code. Replacing this with a for loop reduced by 30KB and made REP (a rare sequence) 11x faster lol.

### DIFF
--- a/src/lib_vt.zig
+++ b/src/lib_vt.zig
@@ -391,16 +391,15 @@ pub const std_options: std.Options = opts: {
     var options: std.Options = .{};
 
     if (builtin.target.cpu.arch.isWasm()) {
-        // Wasm builds we specifically want to optimize for space with small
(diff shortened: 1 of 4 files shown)


andrewrk:
Not LLVM's fault, Zig's fault. Sorry! Here's the issue to track: https://codeberg.org/ziglang/zig/issues/36521

PR #13840, opened 2026-08-15
feat: expand tildes in config theme path to HOME

preiter93:
When loading a theme from a path that includes a tilde:
```
theme="~/.cache/wal/colors-ghostty"
```
ghostty currently fails with the following error:
```
cannot include path separators unless it is an absolute path
```

This PR tries to expand the ~ of the path. If there is no ~ or expansion fails, it falls back to the unexpanded value.

### DIFF
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -9995,11 +9995,11 @@ pub const Theme = struct {
     /// Expand tilde paths in light/dark theme values.
     pub fn finalize(self: *Theme, alloc: Allocator) Allocator.Error!void {
         var buf: [std.fs.max_path_bytes]u8 = undefined;
-    
+
         const light = expandHome(self.light, &buf);
         if (!std.mem.eql(u8, light, self.light))
             self.light = try alloc.dupeZ(u8, light);
-    
+
         const dark = expandHome(self.dark, &buf);
         if (!std.mem.eql(u8, dark, self.dark))
             self.dark = try alloc.dupeZ(u8, dark);
--- a/src/config/Config.zig
+++ b/src/config/Config.zig
@@ -9973,9 +9973,32 @@ pub const Theme = struct {
         // Trim our value
         const trimmed = std.mem.trim(u8, input, cli.args.whitespace);
 
+        // Expand our value
+        var buf: [std.fs.max_path_bytes]u8 = undefined;
+        const expanded = expanded: {
+            if (!std.mem.startsWith(u8, trimmed, "~/"))
+                break :expanded trimmed;
+        
+            var environ_map = global.environMap() catch |err| {
+                log.warn("error getting environment map for theme path {s}: {}", .{ trimmed, err });
+                break :expanded trimmed;
+            };
+            defer environ_map.deinit();
+        
+            break :expanded internal_os.expandHome(
+                global.io(),
+                &environ_map,
+                trimmed,
+                &buf,
+            ) catch |err| {
+                log.warn("error expanding home directory in theme path {s}: {}", .{ trimmed, err });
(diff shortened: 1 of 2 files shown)


src/config/theme.zig
```diff
@@ -113,12 +115,40 @@ pub const LocationIterator = struct {
...
+            &buf,
+        ) catch |err| {
+            log.warn(
+                "error expanding home directory for theme path={s}: {}",
+                .{ theme_, err },
+            );
+            break :expanded theme_;
+        };
+    };
+
```
jcollie: This would be better handled inside `Config.Theme` when the config is parsed, rather than when the theme is opened.

https://github.com/ghostty-org/ghostty/blob/cecf81678e47f967b0354acada67e69d229f436b/src/config/Config.zig#L9940-L10045

jcollie (changes requested):
Tilde expansion should happen when the config is parsed, not when the theme is loaded.

src/config/theme.zig
preiter93: Sounds good, tbh I couldnt decide in which of the 2 places to put. Will move it. This might also make it easier to test.

preiter93: moved it to config @jcollie, the expand title is now duplicated in the config. its also in WorkingDirectory.finalize. Shold I move
```zig
fn expandHome(path: []const u8, buf: []u8) []const u8 {
    if (!std.mem.startsWith(u8, path, "~/"))
        return path;

    var environ_map = global.environMap() catch |err| {
        log.warn("error getting environment map for path {s}: {}", .{ path, err });
        return path;
    };
    defer environ_map.deinit();

    return internal_os.expandHome(
        global.io(),
        &environ_map,
        path,
        buf,
    ) catch |err| {
        log.warn("error expanding home directory in path {s}: {}", .{ path, err });
        return path;
    };
}
```

Into a helper in config and use it in WorkingDirectory and theme?

src/config/Config.zig
```diff
@@ -9973,9 +9973,32 @@ pub const Theme = struct {
...
+                global.io(),
+                &environ_map,
+                trimmed,
+                &buf,
+            ) catch |err| {
+                log.warn("error expanding home directory in theme path {s}: {}", .{ trimmed, err });
+                break :expanded trimmed;
+            };
+        };
+
```
jcollie: This will fail to expand tildes if the user specifies separate light & dark themes. This should definitely be broken out into some sort of helper so that it can also be reused with the implementation in `WorkingDirectory`

https://github.com/preiter93/ghostty/blob/c1217342958b90ed3a25413c1616dfd2dd8cd1bf/src/config/Config.zig#L9949-L9971

jcollie (changes requested):

src/config/theme.zig
jcollie: Yes, this definitely needs some sort of helper function.

src/config/Config.zig
preiter93: Alright, will look into how light/dark theme are specified tomorrow and edtract the helper

preiter93: @jcollie I've updated the PR. I added a finalize function like for working directories that is run after the parseCLI in Config.finalize. The one thing I am not sure is if the expandHome helper should log and fallback to original path if there is an error, or if it should return an error, and logging should happen on caller side. And also not sure where helpers should be put in this file, is there convention?

jcollie:
This is looking good, just need to fix up the formatting to please `zig fmt`.

preiter93:
done, feel free to merge when it goes through ci :)

jcollie (approved):

PR #13885, opened 2026-08-18
fix(lib-vt): report size when mode 2048 is enabled

elias8:
Mode 2048 requires terminals to report the current rows, columns, and pixel dimensions [when the mode is enabled](https://gist.github.com/rockorager/e695fb2924d36b2bcf1fff4a3704bd83#:~:text=When%20first%20enabled%2C%20the%20terminal%20MUST%20send%20a%20report%20of%20the%20current%20size.), then report updated geometry after later resizes. The native termio stream already queues the initial report, but the terminal stream only reported from `resize`, so libghostty consumers that enabled the mode after committing geometry received nothing until another resize.

This makes the generic handler match termio by requesting current geometry through the existing size callback and writing the encoded report through `write_pty` on every enable.

### DIFF
--- a/include/ghostty/vt/terminal.h
+++ b/include/ghostty/vt/terminal.h
@@ -86,13 +86,13 @@ extern "C" {
  *
  * | Option                                  | Callback Type                     | Trigger                                   |
  * |-----------------------------------------|-----------------------------------|-------------------------------------------|
- * | `GHOSTTY_TERMINAL_OPT_WRITE_PTY`        | `GhosttyTerminalWritePtyFn`       | Query responses written back to the pty   |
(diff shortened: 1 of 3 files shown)


include/ghostty/vt/terminal.h
```diff
@@ -86,13 +86,13 @@ extern "C" {
  *
  * | Option                                  | Callback Type                     | Trigger                                   |
  * |-----------------------------------------|-----------------------------------|-------------------------------------------|
- * | `GHOSTTY_TERMINAL_OPT_WRITE_PTY`        | `GhosttyTerminalWritePtyFn`       | Query responses written back to the pty   |
+ * | `GHOSTTY_TERMINAL_OPT_WRITE_PTY`        | `GhosttyTerminalWritePtyFn`       | VT query and mode reports written back to the PTY |
```
mitchellh: Unnecessary comment change

PR #13886, opened 2026-08-18
fix(kitty): Relax image placement maximum scroll-to-after rows

AnthonyZhOon:
While using `icat` consecutively on wallpapers I got overlapping images from `cursor=After`[^1] requests with vertically stacked terminal surfaces.
[^1]: `C=0` for a display command in the protocol

image

> Top: Ghostty @ tip 12967b68
> Bottom: Kitty 0.47.1
> `kitten icat --transfer-mode=stream -n ~/Pictures/wallpapers/S*`

To relax the terminal-height based limit, cap the scroll bound to be 16x the terminal rows

### DIFF
--- a/src/terminal/kitty/graphics_exec.zig
+++ b/src/terminal/kitty/graphics_exec.zig
@@ -279,7 +279,7 @@ fn display(
                 // the work so an untrusted row count can't make us spin.
                 const rows_to_move: usize = @min(
                     @as(usize, size.rows),
-                    @as(usize, terminal.rows),
+                    @as(usize, 16 *| terminal.rows),
                 );
                 for (0..rows_to_move) |_| terminal.index() catch |err| {
                     log.warn("failed to move cursor: {}", .{err});


src/terminal/kitty/graphics_exec.zig
```diff
@@ -279,7 +279,7 @@ fn display(
                 // the work so an untrusted row count can't make us spin.
                 const rows_to_move: usize = @min(
                     @as(usize, size.rows),
-                    @as(usize, terminal.rows),
+                    @as(usize, 16 *| terminal.rows),
                 );
```
AnthonyZhOon, line 283: This row bound was added in  https://github.com/ghostty-org/ghostty/pull/13630/changes/e5840bb9bacdaae78c42baa4b276f52eaa91fdbe to add a reasonable limit to the number of rolls scrolled by an image placement instead of the u32 image limit. But the terminal height can be too tight and cause this safety limit to interfere with normal usage.

mitchellh:
The fix isn't right but the issue is. I've got something else here: https://github.com/ghostty-org/ghostty/pull/13887

PR #13888, opened 2026-08-18
macos: translate physical menu shortcuts

jparise:
Translate printable physical keybindings through the current macOS keyboard layout before assigning menu key equivalents. Previously these bindings could not be represented because SwiftUI shortcuts are character-based, so actions such as `super+backquote` had no native menu shortcut.

Keep native keycodes as dispatch identity so translated display characters do not change physical semantics or precedence over Unicode bindings. Refresh shortcuts when the input source changes, and prevent AppKit from transforming equivalents that are already localized.

**AI Usage:** The approach was suggested by GPT 5.6 Sol, but I wrote most of the code and understand it all.

### DIFF
--- a/macos/Tests/Ghostty/NormalizedMenuShortcutKeyTests.swift
+++ b/macos/Tests/Ghostty/NormalizedMenuShortcutKeyTests.swift
@@ -74,6 +74,15 @@ struct NormalizedMenuShortcutKeyTests {
         #expect(a != b)
     }
 
+    @Test func physicalKeysUseKeyCodeIdentity() {
+        let configured = Key(keyEquivalent: "`", physicalKeyCode: 0x32, modifiers: .command)
+        let event = Key(physicalKeyCode: 0x32, modifiers: .command)
+        let unicode = Key(keyEquivalent: "`", modifiers: .command)
+
+        #expect(configured == event)
+        #expect(configured != unicode)
+    }
+
     @Test func differentModifiersAreNotEqual() {
         let a = Key(keyEquivalent: "c", modifiers: .command)
         let b = Key(keyEquivalent: "c", modifiers: .option)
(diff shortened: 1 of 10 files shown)


bo2themax (changes requested):
I believe this also relates to #13180 and then #13205. But as I said in the comments there in #13180, core already use [UCKeyTranslate](https://github.com/ghostty-org/ghostty/blob/55dac8fc47c239c54540fa30f3bea882d359d07f/src/input/KeymapDarwin.zig#L163). I wonder if there's a way to avoid this kinda duplication.

The changes looks good to me, but I haven't run it locally and want a second opinion on this.

jparise:
> I believe this also relates to #13180 and then #13205. But as I said in the comments there in #13180, core already use [UCKeyTranslate](https://github.com/ghostty-org/ghostty/blob/55dac8fc47c239c54540fa30f3bea882d359d07f/src/input/KeymapDarwin.zig#L163). I wonder if there's a way to avoid this kinda duplication.

That's good feedback. I remember seeing #13180 go by but should have revisited it more closely.

I revised this to use AppKit’s `characters(byApplyingModifiers:)`, avoiding the `UCKeyTranslate` duplication. Going further to reuse the code from core seems like it would net out to more overall complexity.

bo2themax:
> Going further to reuse the code from core seems like it would net out to more overall complexity.

Yeah, true🫪

bo2themax (approved):

macos/Tests/Ghostty/NormalizedMenuShortcutKeyTests.swift
```diff
@@ -13,11 +13,6 @@ struct NormalizedMenuShortcutKeyTests {
         #expect(key == nil)
     }
 
-    @Test func lowercasesKeyEquivalent() {
```
bo2themax, line 16: why is this test deleted tho?

jparise: The `keyEquivalent` property it used was removed in the refactor. The lowercasing behavior remains and is covered by the `uppercaseAndLowercaseWithShiftAreEqual()` and the config normalization tests, rather than inspecting the `Key`'s internal storage.

PR #13901, opened 2026-08-18
terminal: clear progress bar on full reset

fornwall:
Emit a `progress_report` remove effect from the `full_reset` arm, matching kitty and WezTerm which both clear progress on reset.

Previously, only the termio `StreamHandler` removed the progress bar on RIS (#10178) - but the terminal stream handler used by `libghostty-vt` did not, so an embedder's progress bar would outlive the reset.

### DIFF
--- a/src/terminal/stream_terminal.zig
+++ b/src/terminal/stream_terminal.zig
@@ -345,7 +345,12 @@ pub const Handler = struct {
             },
             .active_status_display => self.terminal.status_display = value,
             .decaln => try self.terminal.decaln(),
-            .full_reset => self.terminal.fullReset(),
+            .full_reset => {
+                self.terminal.fullReset();
+
+                // Clear the progress bar
+                self.progressReport(.{ .state = .remove });
+            },
             .start_hyperlink => try self.terminal.screens.active.startHyperlink(value.uri, value.id),
             .end_hyperlink => self.terminal.screens.active.endHyperlink(),
             .semantic_prompt => try self.terminal.semanticPrompt(value),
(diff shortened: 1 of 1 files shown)


src/terminal/stream_terminal.zig
```diff
@@ -345,7 +345,12 @@ pub const Handler = struct {
             },
             .active_status_display => self.terminal.status_display = value,
             .decaln => try self.terminal.decaln(),
-            .full_reset => self.terminal.fullReset(),
+            .full_reset => {
+                self.terminal.fullReset();
+
+                // Clear the progress bar
+                self.progressReport(.{ .state = .remove });
```
mitchellh, line 352: I think we should only do this if there was previously a progress bar state.

mitchellh (changes requested):

src/terminal/stream_terminal.zig
fornwall: 👍 Did this in b56c6d88f81bd68d36a9dcb84fa43c1455df53b0

fornwall: Thinking about it some more, adding this state tracking and diffing here is not 100% natural, and worth thinking through to make things consistent and deliberate.

- Embedders must already tolerate redundant removes - a terminal program that emits remove progress twice produces two remove callbacks, so that could be an argument for always firing remove also on reset.
  - OTOH it's not given that we should always emit this on reset. Reset is used a lot more than explicit progress bar resets, so the state diffing reduces noises.
  - Another direction could be diffing also on explicit removals (and expanding that behaviour to other effects).
- Title and pwd are cleared by reset, but no `title_changed` or `pwd_changed` callbacks are fired, so a callback driver embedder keeps the old title and pwd after resets.
  - Could do this in a follow up PR if desired.
  - Another approach in general could be to add a `full_reset` effect, instead of doing the fan out from full resets into multiple callbacks. But that means that the embedder needs to be aware of vt semantics - each embedder needs to know what a terminal full reset should mean, weakening the "we take care of terminal functionality" value from this library.

The balance (that affects how much state tracking should be done, and the granularity of callbacks) is how thin `libghostty-vt` should be. Perhaps it make sense with general state tracking and diffing? What do you think?

mitchellh: Agree, lets remove this. Sorry for the churn.

fornwall: 👍 Reverted in b7ee3ab6b321470e0638d02aabb5043efdbe797b

PR #13906, opened 2026-08-19
macOS: group settings menu in a separate group

bo2themax:
**Reverted https://github.com/ghostty-org/ghostty/pull/13664 with a second thought.** 

This keeps the consistency with other first-party apps like Terminal, Finder, and Music. It seems that most of the first-party apps show the standard icons on macOS 27 for "Settings...", "Find", "AutoFill" and etc., but in a separate group. 

image

I believe this still follows [HIG](https://developer.apple.com/design/human-interface-guidelines/menus#Icons).


jparise (approved):
Firefox groups this separately, too.

PR #13930, opened 2026-08-20
terminal/search: reuse page encoding buffer

jparise:
SlidingWindow previously allocated and freed an encoding writer for every page appended during a search. Retain one scratch writer for the window and clear it between pages so subsequent encodings can reuse its capacity.


jcollie (approved):

mitchellh (changes requested):
Not sure if this one has any justified performance associated with it to keep a high-water-mark buffer alongside the state at all times.

jparise:
> Not sure if this one has any justified performance associated with it to keep a high-water-mark buffer alongside the state at all times.

I think you're right. I thought this was only tied to the lifetime of a single `SlidingWindow`, but looking more closely I see that we could have several of them active.

PR #13936, opened 2026-08-20
terminal: clear recycled row metadata

fornwall:
Reset all content-derived row metadata whenever in-place scrolling, line insertion/deletion, or active-area erasure recycles Row storage after clearing its cells.

Previously, blank rows could inherit wrap, semantic prompt, or other state from discarded content.

Update the complete snapshot fixture because the previous golden serialized a stale wrap flag on the alternate screen.

---

AI usage: Created with claude code and opus 5. I have reviewed the code and made modifications where it made sense. I have also tested this end to end by a script reproducing user visible buggy behaviour being fixed.


mitchellh:
This touches some of the most performance sensitive parts of the codebase. THis needs to be paired with a significant amount of benchmarks. Also, the pre/postcondition here is misunderstood: we've historically zeroed rows when they're needed, not when they're discarded, as a performance choice. If you can prove otherwise, that's fine, but this needs a lot more diligence.

fornwall:
> Also, the pre/postcondition here is misunderstood: we've historically zeroed rows when they're needed, not when they're discarded, as a performance choice

In 7 of the 8 changed sites here **needed** and **discarded** are at the same instant - the recycled rows become immediately visible as a blank row in the active area. So there is no distinction there (and this PR does not change when zeroing of cells happens, it just adds missing zeroing of row metadata).

In one case (`eraseRows`) the rows are retired into unused page capacity, for possible use by `grow()` - and when erasing the active area, that reclaim happens directly afterward in the same function. And the zeroing of cells was already done here before this PR, and seems to be necessary and per design:

- [PageList.grow fast path](https://github.com/ghostty-org/ghostty/blob/1ffa77c90bc47e8200e2c2c88b34ef6745ffd834/src/terminal/PageList.zig#L3980-L3991) relies on that - no clearing of the re-exposed row there.
- [PageList.createPage](https://github.com/ghostty-org/ghostty/blob/1ffa77c90bc47e8200e2c2c88b34ef6745ffd834/src/terminal/PageList.zig#L4501-L4505) documents `The pool guarantees that buffers it hands out are zeroed`.
- Commit 9df9c999a ("fix(terminal): clear erased rows") states: "Clearing these rows is necessary to avoid memory corruption".

And once again, this PR does not change that - it only adds missing zeroing of row metadata to accompany the existing zeroing.

> This touches some of the most performance sensitive parts of the codebase. THis needs to be paired with a significant amount of benchmarks.

Note that no cell zeroing is added or moved here - the `@memset`/`clearCells` work done for all cells in a row happens in exactly the same places as before. What is added here is one 8-byte read-modify-write of the `Row` header per recycled row compared to the existing clearing of all `cols * 8` bytes in the row.

Benchmark data, 2 runs of three synthetic benchmarks, before and after:

- `primary-scroll` = plain LF scrolling (cursorDownScroll->grow)
- `region-scroll` = DECSTBM region (cursorScrollRegionUp fast path)
- `alt-scroll` = alt-screen scrolling (eraseRow recycle path)

10M scrolled lines each (every line forces a scroll, far denser than any real workload), two binaries interleaved ABBA within each of 20 rounds.

| Workload | Run | before (median) | after (median) | Δ median | Δ min |
|---|---|---|---|---|---|
| primary-scroll | 1 | 668.9 ms | 665.8 ms | −0.46% | −0.46% |
| primary-scroll | 2 | 677.4 ms | 673.5 ms | −0.58% | −0.59% |
| region-scroll | 1 | 670.4 ms | 666.4 ms | −0.60% | −0.29% |
| region-scroll | 2 | 678.6 ms | 675.2 ms | −0.51% | −0.24% |
| alt-scroll | 1 | 570.3 ms | 572.4 ms | +0.38% | +1.23% |
| alt-scroll | 2 | 574.3 ms | 581.3 ms | +1.22% | +2.37% |

`primary-scroll` and `region-scroll` happened to be faster (code-layout/alignment movement). The `alt-scroll` benchmark shows a `~1%` median cost - `~0.5 ns` ≈ 2 cycles per scrolled line, consistent with the expected cost of the added 8-byte read-modify-write.

**Setup:**
- AMD Ryzen 9 7950X3D, Ubuntu 26.04 LTS (kernel 7.0.0-29-generic)
- Both binaries built with Zig 0.16.0 via `zig build -Demit-bench -Doptimize=ReleaseFast -Dapp-runtime=none`
- The benchmark process was pinned to a reserved core with its SMT sibling offlined, the performance governor active, frequency boost disabled, and ASLR off (`setarch -R`). Each run was executed under `perf stat -e cpu-migrations` and counted 0 migrations, confirming the pinning held.
- Run using [this benchmark script](https://gist.github.com/fornwall/3844909168f5584aa4a696affd6bd6e2)

> If you can prove otherwise, that's fine, but this needs a lot more diligence.

Let me know if the above is not enough.

mitchellh:
I'm not going to read a bunch of LLM output in a comment and its against our community guidelines. (EDIT: more explicitly, LLM usage is fine, but not in conversation with other humans)

fornwall:
How about this (100% human written) short form, with details in the earlier comment (which is a mix between verbose human and verbose LLM written output):

- This PR does not change "zeroing at discard" vs "zeroing at usage" - existing code and design already zeroed cells at discard, and this PR changes to also zero (the relatively small) row metadata.
- This PR is needed for correctness.
- Benchmark shows (IMHO) acceptable ~1% increase in a worst case synthetic benchmark.

?

mitchellh:
No, sorry, I'm not convinced you understand it and instead are just relaying feedback back and forth and blindly trusting LLM output. Like, if I took you to a whiteboard and asked you to draw out the paths from memory where this is happening and explain the pros/cons of your approach versus alternates you've considered, I'm not sure you could do it.

If there is a correctness issue, I recommend opening a discussion at this point explaining the problem, and someone who has the broader awareness of how this whole system works can consider appropriate solutions.

fornwall:
For reference, opened a discussion at https://github.com/ghostty-org/ghostty/discussions/13940

PR #13960, opened 2026-08-21
terminal: reload cell pointers when print grows a page

ArneshBanerjee:
Fixes #11261.

`Terminal.print`'s grapheme path caches a `*Cell` for the previous cell and keeps using it after writing other cells. Writing the wide spacer tail calls `printCell`, which can grow the page to make room for the cursor hyperlink. Growing clones the page and frees the old one, so the cached pointer dangles and the following `appendGrapheme` writes into freed memory. The second test case in the issue reproduces it.

Rather than recomputing `prev` on every use, which is too expensive for this path, the fix records the cursor page identity before the spacer write and reloads the cell only if the page actually changed. Node pointer plus serial is used because nodes are pooled and a replacement can land on the same address. Nothing changes when the page does not grow.

Three other pointers in the same function were held across an operation that can replace a page, so they are now read through the cursor or a freshly resolved pin: the grapheme move after a wrap, the grapheme append loop, and `printCell`'s assert on a failed hyperlink write.

Tests:

- `Terminal: VS16 widening when the spacer tail grows the page` fills the page hyperlink map so the spacer tail is what forces growth. It crashes without the fix.
- `Terminal: grapheme transfer when widening wraps to the next line` covers the wrap path where the previous cell already holds grapheme data, which had no test before.

`zig build test` passes.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -1323,14 +1323,21 @@ pub fn print(self: *Terminal, c: u21) !void {
                                 const old_rac = old_pin.rowAndCell();
 
                                 if (new_pin.node == old_pin.node) {
-                                    new_pin.node.page().moveGrapheme(prev.cell, new_rac.cell);
-                                    prev.cell.content_tag = .codepoint;
+                                    new_pin.node.page().moveGrapheme(old_rac.cell, new_rac.cell);
+                                    old_rac.cell.content_tag = .codepoint;
                                     new_rac.cell.content_tag = .codepoint_grapheme;
                                     new_rac.row.grapheme = true;
                                 } else {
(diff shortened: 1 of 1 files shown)


jcollie:
!vouch

mitchellh:
Thanks, this looks sound. I brought it down locally and ran a bunch of benchmarks on it (`ghostty-bench +terminal-stream`) with specific workloads shaped like this and don't see any regressions at all, which was my primary issue. It looks like the introduction of node serials (which didn't exist when I opened that issue!!) makes it possible to fix this in an efficient way. Yay!

ArneshBanerjee:
@jcollie thanks so much for the vouch. I was going to open a discussion for it but had get off.

And thanks for the merge Mitchell!

PR #13963, opened 2026-08-21
libghostty: implement Kitty clipboard protocol read/write

mitchellh:
This implements the full Kitty clipboard protocol for libghostty-vt. libghostty users need to only have the clipboard read/write effect for this to work. This doesn't yet do mode 5522.


mitchellh:
I would love to get reads in too but that depends on if I can solve the clipboard read problem which stands on its own next to this. I tried at first to fit it in here but I think I need to take a step back and look at sovling it for OSC 52 first but with this in mind... 

So depending how that goes I might open that PR separately.

mitchellh:
Going to draft, I merged the clipboard read effect and I'm going to implement that in here too.

PR #13967, opened 2026-08-22
gtk: avoid physical fallback for XKB modifiers

tothedarktowercame:
Hi, I noticed that this XKB config line was causing problems for Ghosttty:

```
    key <DELE> {   [ ISO_Level3_Shift ] };
```

That's a valid reconfiguration of the "delete key" as a modifier, and the same worked fine in other terminal emulators (like Alacritty).  With Ghosttty, I was getting actual `<delete>` behaviour whenever I pressed the modifier key (although the modifier engaged after that).

This patch fixes it.  In line with your AI disclose policy: I made the patch with Codex Sol.  The logical fix is very succinct, and the bulk of the patch works around an underlying issue to do with GDK not recognising some modifiers as modifiers.  I've tested the implementation and the patch definitely resolves it!

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1378,19 +1378,11 @@ pub const Surface = extern struct {
                 if (entry.native == keycode) break :w3c entry.key;
             } else .unidentified;
 
-            // Consult the pre-remapped XKB keyval/keysym to get the (possibly)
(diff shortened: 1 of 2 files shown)


pluiedev (approved):
Separating out `remapKey` is also a good idea. Thanks!

jcollie (approved):

PR #13973, opened 2026-08-22
terminal: Kitty drag and drop protocol drop-only core logic and state machine

mitchellh:
This builds out the core logic and state machine for the Kitty drag and drop protocol for _drops only_. This hooks it into libghostty-vt's Zig API but it isn't available to the C API and it isn't hooked up to any Ghostty GUI. It isn't really recommended that Zig consumers integrate this yet because I'm sure the API will continue to change dramatically as I address the missing features: drag source, remote drops, etc.

The major thing this does it the core `src/terminal/kitty/dnd.zig` stuff with e2e tests extracted from Kitty's own `kitty_tets/dnd.py`. So this verifies that what we have so far is working properly.

### DIFF
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -91,6 +91,12 @@ mouse_shape: mouse.Shape = .text,
 /// Per-session Glyph Protocol registrations.
 glyph_glossary: glyph.Glossary = .empty,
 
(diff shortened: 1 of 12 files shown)


mitchellh:
Going to merge this since it touches no production consumers, but the limitations are such that we can't really hook this up to an apprt yet, but we can iterate and improve on it.

PR #13987, opened 2026-08-24
renderer/image: Fuse copying data and converting pixel format

AnthonyZhOon:
Profiling `mpv --vo=kitty --vo-kitty-use-shm <video>` shows up a copy and then swizzle in `prepImage`
This comes from the renderer copying the raw image data for ownership, then doing an rgb to rgba conversion to replace the copied data.
This change optimize this case by letting the format conversion read from the data source instead of a copy of it. 

image
image
> Above-Before: prepImage profiles a memcpy + swizzle
> Below-After: prepImage profiles just a swizzle

The existing data path for kitty images is:
```

Read tty for Kitty image transmission (srgb, srgba, or PNG bytes)
-> copy into Kitty graphics ImageStorage (cpu-owned)

Renderer updateFrame creates Image.Pending in renderer-owned CPU storage
-> sync Kitty ImageStorage with renderer-owned ImageMap
-> copy bytes into renderer.ImageMap (renderer-owned) 
-> convert ImageMap bytes to preferred GPU upload format

Renderer drawFrame uploads image data to the GPU
-> iterate through renderer.ImageMap
  -> for Pending image uploads
    -> convert the pixel format (no-op if already done), create the GPU-side texture and upload the data
```

# Note
Kitty graphics only supports RGB and RGBA data
The image file decode path uses a wuffs png and jpeg decode function configured to return RGBA 8-bit, so I think in practice we only ever upload RGB8 or RGBA8. And the gray-alpha and gray pixel formats aren't ever used.

# AI Disclosure
I didn't use any LLM assistance for this.

### DIFF
--- a/src/renderer/image.zig
+++ b/src/renderer/image.zig
@@ -710,11 +710,8 @@ pub const State = struct {
             return;
         }
 
-        // Copy the data so we own it.
-        const data = if (alloc.dupe(
-            u8,
-            pending.dataSlice(),
-        )) |v| v else |_| {
+        // Store it in the map
+        const new_image: Image = if (pending.convertCopy(alloc)) |v| v else |_| {
             if (!gop.found_existing) {
                 // If this is a new entry we can just remove it since it
                 // was never sent to the GPU.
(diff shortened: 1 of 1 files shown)


mitchellh:
Looks right. Thank you.

PR #13993, opened 2026-08-24
macOS: Kitty clipboard read support

mitchellh:
This adds Kitty clipboard protocol _read_ support to macOS. In the process, this also does most of the core termio, apprt, and Surface work so GTK is likely very easy to do, I just didn't have the machine on hand to test at the given moment. I will create an issue to follow up with that. 

This fully supports:

- Non-text data, like images! For this, we show an image preview.
- Per-program "remember"
- Showing the program name if given instead of generic "An application"

CleanShot 2026-08-24 at 08 37 45@2x

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -85,6 +85,29 @@ typedef struct {
   size_t len;
 } ghostty_clipboard_content_s;
 
+// The payload for completing a clipboard read request. See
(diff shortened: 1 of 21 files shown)


bo2themax (approved):

mitchellh:
After this I can fast follow with full mode 5522 (read) support which will enable Claude Code I think.

PR #13996, opened 2026-08-24
i18n: it_IT translations for 1.4

Misairuzame:


Atomk:
Thank you! Available for feedback when you are satisfied with the translations :)

Misairuzame:
Italian translation is now ready for review.
Note that I decided to change the translation of "Split" from "Divisione" to "Riquadro" as it seems to me like a more standard translation.
Any feedback is appreciated @Atomk  :smile:

po/it.po
```diff
@@ -270,27 +270,27 @@ msgstr "Esegui un comando…"
...
 
 #: src/apprt/gtk/class/application.zig:1780
 msgid "Global keybind unavailable"
-msgstr ""
+msgstr "Scorciatoia globale non disponibile."
 
 #: src/apprt/gtk/class/application.zig:2259
 msgid "Export Terminal IO Events"
-msgstr ""
+msgstr "Esporta eventi di I/O del terminale"
```
Atomk: While grammatically it makes sense, I think that "di" makes it sound a bit weird

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:708
 msgid "Ghostty"
-msgstr ""
+msgstr "Ghostty"
 
 #: src/input/command.zig:709
 msgid "Put a little Ghostty in your terminal."
-msgstr ""
+msgstr "Metti un piccolo Ghostty nel tuo terminale."
```
Atomk: If here "put" means "show", maybe "Mostra" would fit better than "metti"

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:696
 msgid "Redo"
-msgstr ""
+msgstr "Ripeti"
 
 #: src/input/command.zig:697
 msgid "Redo the last undone action."
-msgstr ""
+msgstr "Ripeti l'ultima azione annullata."
```
Atomk: This is a tricky one, I've seen this translated as "ripristina", in this case I think it might make the intention a bit clearer. Our "ripeti" is probably closer to "repeat" in English, it's very subtle.

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
+"Copia il titolo del terminale negli Appunti. Se il terminale non ha un "
+"titolo, questa azione non ha alcun effetto."
 
 #: src/input/command.zig:185
 msgid "Paste from Clipboard"
-msgstr ""
+msgstr "Incolla dagli Appunti"
 
 #: src/input/command.zig:186
 msgid "Paste the contents of the main clipboard."
```
Atomk, line 466: @trag1c what does "main clipboard" mean here? Is it the system clipboard or a Ghostty feature? I see there's also a "Paste the contents of the selection clipboard" string and I am a bit confused by what the clipboard is in this context. The translation hints at the system clipboard but it seems there's something else going on.

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:191
 msgid "Paste from Selection"
-msgstr ""
+msgstr "Incolla dalla selezione"
 
 #: src/input/command.zig:192
 msgid "Paste the contents of the selection clipboard."
-msgstr ""
+msgstr "Incolla i contenuti dagli Appunti di selezione."
```
Atomk, line 475: The translation here depends on whether this is the system clipboard, see the other comment about this.

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:197
 msgid "Start Search"
-msgstr ""
+msgstr "Inizia ricerca"
 
 #: src/input/command.zig:198
 msgid "Start a search if one isn't already active."
-msgstr ""
+msgstr "Inizia una ricerca se non è già attiva."
```
Atomk: Perhaps "se non è già in corso", for consistency with the "Termina la ricerca in corso" string

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 #: src/input/command.zig:209
 msgid "End Search"
-msgstr ""
+msgstr "Termina ricerca"
 
 #: src/input/command.zig:210
 msgid "End the current search if any and hide any GUI elements."
 msgstr ""
+"Termina la ricerca in corso, se presente, e nascondi "
+"gli eventuali elementi dell'interfaccia grafica."
```
Atomk: Just want to appreciate the commitment to keep this fully in Italian! 🇮🇹

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:243
 msgid "Clear Screen"
-msgstr ""
+msgstr "Pulisci schermo"
 
 #: src/input/command.zig:244
 msgid "Clear the screen and scrollback."
-msgstr ""
+msgstr "Pulisci lo schermo e la cronologia di scorrimento."
```
Atomk, line 549: Perhaps "schermo" could become "schermata", just to make a distinction between Ghostty's screen and the computer screen ("pulisci lo schermo" sounds funny)

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:280
 msgid "Scroll the screen down by a page."
-msgstr ""
+msgstr "Scorri lo schermo in giù di una pagina."
 
 #: src/input/command.zig:286
 msgid "Copy Screen to Temporary File and Copy Path"
-msgstr ""
+msgstr "Copia schermo in file temporaneo e copia percorso"
```
Atomk, line 601: This convinced me that "schermata" probably fits better in these strings

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:389
 msgid "Copy Selection as HTML to Temporary File and Open"
-msgstr ""
+msgstr "Copia selezione come HTML in file temporaneo e apri"
 
 #: src/input/command.zig:390
 msgid "Copy the selection contents as HTML to a temporary file and open it."
 msgstr ""
+"Copia i contenuti della selezione come HTML in un file temporaneo e aprilo."
```
Atomk: Is it intended for this to be on its own line?

po/it.po
Atomk, line 1150: If "put" means "show" then "mostra" would be a bit clearer (@trag1c any hint on what this does?)

po/it.po
```diff
@@ -397,710 +397,754 @@ msgstr "Sviluppatori di Ghostty"
...
 
 #: src/input/command.zig:691
 msgid "Undo the last action."
-msgstr ""
+msgstr "Annulla l'ultima azione."
 
 #: src/input/command.zig:696
 msgid "Redo"
-msgstr ""
+msgstr "Ripeti"
```
Atomk: This is a very tricky one, since our "ripeti" is probably closer to "repeat" in English, from this command's description it seems it cannot be used to repeat the last action (Google Docs actually lets you do this). I've also seen this translated as "ripristina" which might fit better, even though "ripeti" is so common that probably either of those is fine. Your call.

po/it.po
Atomk: Just want to appreciate the commitment to keep this fully in Italian! 🇮🇹

po/it.po
Atomk, line 763: Is it intended for this translation to add the empty string compared to the original? I found a few other similar cases.

Atomk (commented):
I just have a few nitpicks and a couple doubts on the intended meaning of the original strings, other than that everything is pretty much as good as it can be, great job!

po/it.po
00-kat: The main clipboard is the <kbd>ctrl</kbd>+<kbd>c</kbd>/<kbd>ctrl</kbd>+<kbd>v</kbd> one, and the selection clipboard is the middle click one. There's no Ghostty-specific clipboard (on Linux at least, I don't know about macOS).

po/it.po
00-kat: The selection clipboard isn't *the* system clipboard; they're both provided by the system but it's not the conventional clipboard of most OSs. (If you're not familiar with it: selecting text puts it in the selection clipboard, and middle clicking pastes from it; in essence middle click pastes the last selection.)

po/it.po
00-kat: It inserts a Ghost emoji as if you typed it:

https://github.com/user-attachments/assets/1d8a126f-c011-4dc0-8351-b5775ca60022

So “show” isn't exactly accurate. I'd honestly say “metti” is probably fine (going off [these definitions]).

[these definitions]: https://en.wiktionary.org/wiki/mettere#Italian

po/it.po
00-kat: Multiple adjacent strings are joined, so these are all equivalent:
```po
msgstr "foo bar baz lorem ipsum dolor sit"

msgstr "foo bar baz "
"lorem ipsum dolor sit"

msgstr ""
"foo bar baz lorem "
"ipsum dolor sit"
```

It's just done this way because it formats nicer when hardwrapped (gettext likes hardwrapping everything to some number of characters; if you don't do it it does it the next time the po file is updated anyway). You'll see this in long English strings too: https://github.com/ghostty-org/ghostty/blob/4dac203c3f6da1c3ec884e71b1cdc6a428b7a8e5/po/it.po#L647-L649

po/it.po
Atomk: Oh so it's intentionally vague as a sort of easter egg I suppose, thanks for the video and for even double-checking the definition! I now realize I could have downloaded the latest version to check this myself.
EDIT: I'm not able to mark this comment as resolved, I get error "failed to resolve thread"

po/it.po
Atomk: Oh I though the empty string had some special meaning, thanks for the explanation!

po/it.po
00-kat: > EDIT: I'm not able to mark this comment as resolved, I get error "failed to resolve thread"

Oh yeah that's an annoying thing about GitHub, only maintainers and the person who opened a PR can mark threads as resolved 🫩. Done it for you ^^.

po/it.po
Misairuzame: Yep, will do

po/it.po
Misairuzame: First of all I think I could replace all "Pulisci" with "Cancella", or maybe even "Ripulisci", wdyt?

About "schermo" and "schermata", I think "schermata" usually means "everything that's on the screen" (even if in this case it would be perfectly understandable that it's referring to the terminal) but I also found a couple sources that use "schermo" to mean "the terminal's screen", such as [Apple](https://support.apple.com/it-it/guide/terminal/trmld1f46097/2.14/mac/15.0) (buffer di schermo, schermo principale/alternativo), and [Wikipedia](https://it.wikipedia.org/wiki/Clear_(Unix)) ("clear è un comando Unix che viene utilizzato per ripulire lo schermo.").
So maybe "schermo" is the correct term in a technical sense, I'm not too sure :sweat_smile:. Open to any further discussion

po/it.po
Misairuzame: Very good call, I think "ripristina" is technically more correct:
[Microsoft](https://support.microsoft.com/it-it/office/foundations-experiences/undo-redo-or-repeat-an-action) [Wikipedia](https://it.wikipedia.org/wiki/Undo)
Will do!

po/it.po
Misairuzame: Thanks!!! :it: 
Btw, re-reading this translation I'm feeling like a good alternative could be:
"Termina la ricerca in corso, se attiva, e nascondi eventuali elementi dell'interfaccia grafica."
I tried to always translate "if any" with "se presente" for consistency but I feel like for this case "attiva" sounds better.
Also the "gli eventuali elementi" sounds better to me without "gli".
Wdyt?

po/it.po
Atomk: Thanks for checking, following your example I tried to look for references in official translations but had no luck with Windows Terminal, and VSCode says "Cancella terminale" which is pretty disappointing. Interestingly, I've found [an MS-DOS manual](https://ia800700.us.archive.org/27/items/msdos3.10olivetti/MS-DOS%203.10%20Olivetti.pdf) by Olivetti dated 1986, where the CLS command is described as "Cancella il contenuto dello schermo". It's a bit unfortunate that we don't have a more specific/unambiguous term but maybe it's just me being overly pedantic, since there's a clear precedent let's keep "schermo".

po/it.po
Atomk: I appreciated the consistency, though I agree that in this specific string "attiva" flows better. I also agree that "gli" does not add much value to the sentence, I was wondering though, if the mentioned GUI refers to the search GUI, we have a few even simpler options:
- "nascondi eventuali elementi dell'interfaccia grafica" (current version without "gli")
- "nascondi i relativi elementi dell'interfaccia grafica"
- "nascondi la relativa interfaccia grafica"
- "nascondi l'interfaccia di ricerca"

po/it.po
Misairuzame: Amazing find! Let's go with "schermo" for now then, but maybe in the future we can revisit.
I think I will change "Pulisci" though.

po/it.po
Misairuzame: I think it means to hide the search bar (and highlights, etc.):

https://github.com/user-attachments/assets/1321eba2-80bc-4b66-a529-95e3e557b22a

Atomk: Nice thanks, then any of the suggested options will be fine! I feel "nascondi l'interfaccia di ricerca" is the most unambiguous one despite the repetition of "ricerca", but I have no strong preference so feel free to choose another variation if you disagree.

po/it.po
Atomk: First time I hear about this, interesting. Then I think this translation can be fine, even though it sounds a bit weird,  "paste from the clipboard" and "past from the main clipboard" should always mean the same thing but since it's two different strings I guess the distinction is important in some cases... I can't think of anything to clarify the clipboard translations, let's keep them as they are.

po/it.po
Atomk: I was not familiar with it, thanks. A part of me wants this to just say "Paste the contents of the last selection" for the sake of simplicity and non-linux-users friendliness, but it makes sense for a technical tool to be precise. Let's keep this translation.

00-kat: > non-linux-users friendliness

I don't think it even works outside Linux.

po/it.po
00-kat: > since it's two different strings

The “main clipboard” one is the description that appears on hover, which is why it tries to be more elaborate than the command name.

Atomk: Oops I can't believe I missed that, for some reason I was convinced they were descriptions for two different commands, sorry 😅

Misairuzame:
In the end I chose to left "Pulisci" as the translation for "Clear" and i chose "nascondi i relativi elementi dell'interfaccia grafica" for the end_search command.
If there is no further feedback, @Atomk I thank you as always for your review and your time :smile: and I'd say this is ready to be merged.

Atomk (approved):

00-kat (approved):

Issue #12228, opened 2026-04-10
Zig 0.16 migration


pluiedev:
Zig 0.16 has landed on April 13, 2026, so it's time to do this all over again :)

In progress PR at https://github.com/ghostty-org/ghostty/pull/12388

The major breaking change this time is the new `std.Io` interface that will be part of any code that has to access basically any system interface (filesystem, threads, timers, etc.), and OS-specific APIs like `std.posix` has almost been completely removed. This will likely be an even greater challenge compared to Writergate due to how much we rely on `libxev` for these sorts of tasks. I would recommend trying to port the codebase to use a generic `std.Io` implementation, and polyfilling in the features we need but aren't quite in the standard library yet — then we could either try to run the whole program off of `std.Io.Evented`, or try to repackage `libxev` as yet another `std.Io` implementation.

These Zig dependencies need to be migrated:
- [x] `libvaxis` — merged PR: https://github.com/rockorager/libvaxis/pull/316
- [X] `z2d` — https://github.com/vancluever/z2d/pull/169
- [x] `zig-objc` — merged PR: https://github.com/mitchellh/zig-objc/pull/34
- [x] `zig-js` — merged PR: https://github.com/mitchellh/zig-js/pull/10
- [X] `uucode` — has switched to Zig 0.16 for its main branch
- [x] `libxev` — https://github.com/mitchellh/libxev/pull/220
- [x] `zig-wayland` — has switched to Zig 0.16 for its main branch
- [ ] `zf` — https://github.com/natecraddock/zf/pull/87
- [X] `gobject` — New [0.16-compatible release](https://github.com/ghostty-org/zig-gobject/releases/tag/0.8.0-2026-04-13-24-1)
- [X] `zon2nix` — merged PR: #12259

francisqureshi:
I'd love to help out with this, I've been wanting to start contributing on Ghostty for some time and I've been working on Zig Master on my own Zig projects that use zig-objc so can for sure help there as I have a simple fork at the moment.

pluiedev:
If you can make a PR to zig-objc that will be great, as I only have access to a Linux machine :)

pluiedev:
@jacobsandlund Looks like the API changed a lot with the Zig 0.16 branch. I'm not smart enough to figure out how to port our uucode setup to the new system — could you maybe lend a hand?

Update: never mind, I figured this out myself :)

francisqureshi:
[...]

PR #14012, opened 2026-08-25
cli: update list-themes preview sample to valid Zig 0.16 syntax

tacheraSasi:
The demo code shown in the `+list-themes` theme preview was stale from before the Zig 0.16 migration (context: #12228). It referenced `std.Io.getStdOut().writer()`, which never existed in any Zig release, and `pub fn main() !void`. This rewrites the rendered sample to valid Zig 0.16 idioms:

```zig
const std = @import("std");

pub fn main(init: std.process.Init) !void {
    var buf: [1024]u8 = undefined;
    var stdout = std.Io.File.stdout().writer(init.io, &buf);
    const w = &stdout.interface;

    var i: usize = 1;
    while (i <= 16) : (i += 1) {
        if (i % 15 == 0) {
            try w.writeAll("ZiggZagg\n");
        } else if (i % 3 == 0) {
            try w.writeAll("Zigg\n");
        } else if (i % 5 == 0) {
            try w.writeAll("Zagg\n");
        } else {
            try w.print("{d}\n", .{i});
        }
    }
    try w.flush();
}
```

The gutter line numbers, row offsets, and child window height were renumbered to match, and the zig version shown in the demo prompt line was updated from v0.13.0 to v0.16.0.

### DIFF
--- a/src/cli/list_themes.zig
+++ b/src/cli/list_themes.zig
@@ -1205,7 +1205,7 @@ const Preview = struct {
                         .x_off = x_off,
                         .y_off = next_start,
                         .width = width,
-                        .height = 24,
+                        .height = 28,
                     },
                 );
                 const bold: vaxis.Style = .{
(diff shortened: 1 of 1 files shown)


jcollie:
!vouch

jcollie (approved):
Approved pending CI

PR #14020, opened 2026-08-25
core: add detach_split command palette entry

gadgetman6:
## What

Adds the "Detach Split" command palette entry for the `detach_split` action from #13527.

## Why a separate PR

Split out of #13527 at @trag1c's request: the palette entry adds two translatable strings and Ghostty is in a string freeze for 1.4, so this is meant to land after 1.4.0 ships. It is stacked on #13527 (the first two commits are that PR, only the last commit is new); I'll rebase once #13527 merges.

## Testing

- `zig build` on macOS (Zig 0.16.0).
- `.pot` regenerated with `zig build update-translations`. Only the `.pot` is updated, no `.po` changes, matching how other string additions land.

## AI disclosure

Same as #13527: implemented with Claude Code (Claude Fable 5) under my direction; I reviewed the diff and built it locally.


jcollie:
Closing this, we want to keep the addition of the keybind and the command palette in the same PR. I know you're anxious to get the keybind into a release but this will just have to wait.

PR #14033, opened 2026-08-26
renderer: run DisplayLink only while a frame is pending

j-c-m:
A focused vsync window currently keeps CVDisplayLink running at the display cadence. Idle frames still hop the render thread for a Metal no-op presentLastTarget.

Start the link from updateFrame when cells_rebuilt or an animation wake is set, so Thread.drawFrame(false) still waits for vsync. Stop it after a draw that leaves nothing pending. Unfocused, occluded, and window-vsync=false are unchanged.

Idle focused Time Profiler (t ≥ 5s): renderer plus CVDisplayLink went from 0.80% of one core on main to 0.13% on this branch (6.3×; 257 ms / 32 s vs 29 ms / 23 s).

This is from https://github.com/ghostty-org/ghostty/discussions/14030#discussioncomment-18165199, this looks like a clean win with minimal change for focused window efficiency.

AI - extensive use of grok 4.6 and some Claude in the experiments, this small change was all human after the research.

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -1191,10 +1191,13 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                 };
             }
 
-            // If we're not visible, then we want to stop the display link
-            // because it is a waste of resources and we can move to pure
-            // change-driven updates.
-            if (self.visible and self.focused) {
+            // Keep the link running only while a focused, visible surface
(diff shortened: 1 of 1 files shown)


mitchellh:
Looks good, made some changes and replaced it with this https://github.com/ghostty-org/ghostty/pull/14035 kept you as coauthor

PR #14035, opened 2026-08-26
renderer: park DisplayLink while idle

mitchellh:
#14033

Pause the CVDisplayLink when there isn't any real work to do.

Start the link after updateFrame rebuilds cells, keep it running while cell changes or animations remain pending, and resync it from the no-redraw path to sleep it again. 

I also did some benchmark to measure the cost of starting/stopping the display link since this includes a lot more of that and I found that a continuously running link used 29 to 35 us of CPU per callback, while starting and stopping it for every frame used 103 to 121 us. So, in some pathological case this can be worse, but its still microseconds, and in the normal case this helps Ghostty sleep a lot more.

### DIFF
--- a/src/renderer/generic.zig
+++ b/src/renderer/generic.zig
@@ -1191,11 +1191,18 @@ pub fn Renderer(comptime GraphicsAPI: type) type {
                 };
             }
 
-            // If we're not visible, then we want to stop the display link
-            // because it is a waste of resources and we can move to pure
-            // change-driven updates.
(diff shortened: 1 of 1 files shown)


j-c-m:
Looks good to me. The removal of syncDisplayLink at 1913 also makes sense under the drawing case, I didn't even think to check the start & stop time, now that I see it this is a good tweak.

mitchellh:
Nothing is free, not even stopping. 😄

PR #14039, opened 2026-08-26
renderer: instance cell backgrounds and convert window bg on the CPU

j-c-m:
A smallish PR for [#1](https://github.com/j-c-m/jetty/blob/master/docs/jetty-cpu-gpu.md#1-one-opaque-cell-pass-largest-gpu-cut) without changing the rendering quality/design of Ghostty. Showed measurable improvement when tracing.

Cell backgrounds were two fullscreen triangles that converted and blended every pixel, including empty cells that should show the window color.

Convert the window color on the CPU and use it as the render-pass clear. Draw each cell background as a quad. Skip empty cells. Skip blending when every cell is opaque.

Metal System Trace vs main, ~50 GPU kicks/s, one dirty cell per frame: empty grid 538 µs → 173 µs per kick (3.1×). Full opaque grid 537 µs → 296 µs (1.8×).

AI: Yes, mostly grok 4.6 again with a little Claude.


j-c-m:
benchmark traced:

https://github.com/ghostty-org/ghostty/blob/1fdcb4104f145828904b6ef1d8261fbd445a73b5/src/benchmark/cell_bg_trace.py

qwerasd205:
I haven't, and probably won't have time for a proper review of this- but want to note for Mitchell- I strongly suspect this will introduce a regression in handling of the `alpha-blending` config option- and possibly another quite subtle one: I avoided converting color spaces on the CPU with our own math until now for a reason, if we implement our own math then we need to extract the color profile for the display, which is something we *can* do, but we do have to do it if we want to do this optimization, otherwise our colors will be off on displays that are calibrated and not just standard display P3.

j-c-m:
> I haven't, and probably won't have time for a proper review of this- but want to note for Mitchell- I strongly suspect this will introduce a regression in handling of the `alpha-blending` config option- and possibly another quite subtle one: I avoided converting color spaces on the CPU with our own math until now for a reason, if we implement our own math then we need to extract the color profile for the display, which is something we _can_ do, but we do have to do it if we want to do this optimization, otherwise our colors will be off on displays that are calibrated and not just standard display P3.

Agree. I closed this PR. Going to split it and test the impact of only the non loadColor commit to see if it's worth doing a proper display aware shared color profile across CPU & Shader.

PR #14051, opened 2026-08-27
macos: normalize action working directory paths

mitchellh:
Discussion #14048

Directory URLs no longer export a trailing slash through PWD, which keeps zsh's %1~ prompt expansion from resolving to an empty string.

A shared URL helper removes trailing separators while preserving the filesystem root and percent-decoding behavior. Tests cover normal, repeated, encoded, and root paths.

### DIFF
--- a/macos/Sources/Helpers/Extensions/URL+Extension.swift
+++ b/macos/Sources/Helpers/Extensions/URL+Extension.swift
@@ -3,10 +3,6 @@ import Foundation
 extension URL {
     /// The decoded path with trailing separators removed, except for the root path.
     var pathWithoutTrailingSlash: String {
-        var result = path(percentEncoded: false)
-        while result.count > 1 && result.hasSuffix("/") {
-            result.removeLast()
-        }
-        return result
+        FilePath(path(percentEncoded: false)).string
     }
 }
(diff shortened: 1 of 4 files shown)


macos/Sources/Helpers/Extensions/URL+Extension.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+extension URL {
+    /// The decoded path with trailing separators removed, except for the root path.
+    var pathWithoutTrailingSlash: String {
```
bo2themax, line 6: We can also use FilePath to remove it.

macos/Sources/Helpers/Extensions/URL+Extension.swift
```diff
@@ -0,0 +1,12 @@
...
+        var result = path(percentEncoded: false)
+        while result.count > 1 && result.hasSuffix("/") {
+            result.removeLast()
+        }
+        return result
```
bo2themax:
```suggestion
        FilePath(path(percentEncoded: false)).string
```

macos/Sources/Helpers/Extensions/URL+Extension.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
```
bo2themax, line 1:
```suggestion
import Foundation
import System
```

bo2themax (approved):

PR #14052, opened 2026-08-27
gtk,opengl: free us from the clutches of GtkGLArea

pluiedev:
GtkGLArea had numerous downsides that forced us to invent unsightly hacks in our renderer to work around them, most chiefly the fact that it holds its own GdkGLContext on the main thread (GL contexts are not at all thread-safe), forcing us to keep our GL calls on the main thread. It also does not interact well with triple-buffering and initialization is forced to be this sort of deferred song-and-dance since we need to wait for the GLArea to initialize its GL context before we can initialize the renderer, the core surface, and then most things in the GTK surface.

We instead invent our own custom widget named RenderSurface that takes simple DMABUFs and displays them. The task of obtaining a GL context falls to manual EGL bindings, since we also need EGL to export OpenGL textures into DMABUFs. We keep the EGL context solely on the render thread meaning that the main thread never concerns itself with rendering except when being notified that the renderer has pushed a new frame.
    
What makes this extra significant is that now the entire GTK apprt no longer depends on OpenGL in any way, shape or form. As long as it is being fed DMABUFs, it can render from whichever graphics API you want. This means we can add more backends based on OpenGL ES or more likely Vulkan rather painlessly in the future.

**AI disclosure**: I came up with the idea and let Pi implement most of the nitty-gritty details around EGL, as well as replumbing the renderer and cleaning up all the GTK-specific workarounds there. I then carefully vetted every line of code and spent roughly as much time reviewing as coding. Most of the documentation and all commit messages are in my own words.


mitchellh:
My first high-level pass: this looks great. I want to dig in a bit more but I did pass through every line of code manually here and looking WAY simpler overall. 

One thing that felt a bit weird was the message-based displayRealized/Unrealized. Why do that versus just locking the renderer and keeping that synchronous? I feel like you may not want to depend on queues before you allocate/remove GPU resources.

pkg/opengl/egl.zig
```diff
@@ -0,0 +1,295 @@
...
+//! and are also accessed via `c.egl*` — GLAD `#define`s each extension
+//! function name to its internal function-pointer global, so the same
+//! call syntax works for both core and extension functions.
+//!
+//! Call `load()` once before using any EGL extension functions.
+
+const std = @import("std");
+const c = @cImport({
+    @cInclude("glad/glad_egl.h");
+});
```
jcollie: This needs to be a `translate-c` step somewhere in the build process.

pluiedev:
Good point. I figure if you're unrealizing a surface anyway there's not much of a point trying to send stuff into a mailbox that's going to be destroyed very soon... Lemme change that

pkg/opengl/egl.zig
jcollie: This needs to be a `translate-c` step somewhere in the build process.

jcollie:
I've been running these patches on my daily driver for the past couple of days and no problems so far!

src/apprt/surface.zig
```diff
@@ -153,6 +153,9 @@ pub const Message = union(enum) {
     /// Selected search index change
     search_selected: ?usize,
 
+    /// Renderer pushed a new frame, redraw this surface.
+    redraw,
```
mitchellh, line 157: Do we need this anymore? I can't remember the context but I thought we only needed this for main-thread drawing. If we're not doing that anymore, is this necessary? I don't think macOS even uses this.

mitchellh (changes requested):
Mostly ready to accept this as-is now but curious on this.

src/apprt/surface.zig
pluiedev: This is used to tell GTK to redraw the render surface: https://docs.gtk.org/gtk4/method.Widget.queue_draw.html

~~I do wonder if we can do this more intelligently... Maybe something like https://docs.gtk.org/gdk4/class.FrameClock.html could help the widget redraw itself autonomously. Let me look into this more~~

This is a bad idea as GTK does not like being redrawn hundreds of times a second. Gonna leave this as is.

pluiedev:
Also I found a bug with background images so I'm not gonna hit merge just yet till I fix it

dkinzler:
I tried to test this on my system with a Nvidia card, but ran into some trouble. I use Arch + Hyprland + Nvidia Driver 580.178.04-1 with a GTX1060, the card is kind of old, but generally things work out of the box for various GPU-accelerated apps, including Ghostty.

Setting up the display with `eglGetDisplay` does not work. By default it selects the MESA driver and does software rendering and if I try to force the Nvidia driver with the `__EGL_VENDOR_LIBRARY_FILENAMES` env var the function returns null. Eventually I got it to work by using `eglGetPlatformDisplay` and some other hacks (see the [changes](https://github.com/dkinzler/ghostty/commit/a52e8473afb3636daa25d50a5f9cbc89c2fbb5d2) here, includes some more comments). As an additional data point, GTK uses `eglGetPlatformDisplayEXT` on my system (although that is not for headless rendering).

With those changes Ghostty works and uses the Nvidia GPU, but there is a lot of lag. See the following video. The input is delayed sometimes up to 1s and when running `ls` the output seems to appear slightly delayed and the prompt after very delayed. I am not sure if this is still related to the Nvidia driver or some other problem. CPU/GPU usage seems normal compared to Ghostty tip. One other thing I noticed is that it always selects OpenGL 4.3 instead of 4.6 (as on tip), although forcing 4.6 by increasing the minimum does not change anything regarding the lag. 


https://github.com/user-attachments/assets/59ea7473-eabb-4132-9bf7-48bac00aad2c



If you don't have access to a Nvidia system, this might be tough to work on, so I will investigate more myself and can of course test any ideas you might have.

Disclaimer: The code above was developed in collaboration with AI, while I kind of understand what's going on high-level, I don't know a lot about graphics programming.

pluiedev:
Ugh why is Nvidia like this

One side benefit is maybe we can do a bit of device selection here like with Metal, so in hybrid (dual-GPU) setups we can use the one that's more appropriate for this use case. We have to do this unconditionally on Vulkan anyways so having this for OpenGL is probably also a good idea

pluiedev:
OK so, bad news. Nvidia sucks. There's no EGL extension for exporting DMABUFs on Nvidia drivers so we have to add a fallback path that copies each frame into CPU memory instead (yikes). The reason why `eglGetDisplay` prefers Mesa is probably because Mesa's ICD is placed more towards the front than Nvidia's ICD...

image

jcollie:
> OK so, bad news. Nvidia sucks. There's no EGL extension for exporting DMABUFs on Nvidia drivers so we have to add a fallback path that copies each frame into CPU memory instead (yikes). The reason why `eglGetDisplay` prefers Mesa is probably because Mesa's ICD is placed more towards the front than Nvidia's ICD...

Oh, that _does_ suck.

AnthonyZhOon:
I tried this with a 3080 with drivers 610.57.04, I see OpenGL 4.6 being used by default
I'm not seeing MESA software rendering, at least it's using the GPU but I don't have integrated graphics.
I am seeing the latency spikes on key-inputs but it can render as fast as the main branch, so it seems like failing to notify the render thread on app events and just relying on the default low frame rate re-renders from cursor blink.
I still get 2k FPS on DOOM-fire with stable present at 160fps so I think pure rendering side is fine?

Opening the Inspector reliably causes a crash with stack trace
```
thread 41024 panic: attempt to use null value
/home/fedora/dev/ghostty-ref/pkg/opengl/draw.zig:9:28: 0x2d2c060 in clearColor (ghostty)
    glad.context.ClearColor.?(r, g, b, a);
                           ^
/home/fedora/dev/ghostty-ref/src/apprt/gtk/class/imgui_widget.zig:321:22: 0x2d2be7d in glAreaRender (ghostty)
        gl.clearColor(0x28 / 0xFF, 0x2C / 0xFF, 0x34 / 0xFF, 1.0);
                     ^
```

https://github.com/user-attachments/assets/0ca5fe45-970e-4b65-9674-874e9d6096f6

> Note that repeated inputs are smooth but it looks like switching between current frame and a past frame? There's also a deferred present for the last input

AnthonyZhOon:
Profiling with Nvidia Nsight Systems on Doom-Fire-Zig. I see the terminal's draw calls not being deferred to when demanded by gtk's draw.
image

image
> The GlFinish call is our terminal draw, the blue section in the main thread is `eglSwapBuffers` called by Gtk to VSync the draw
> Above: Main branch, the frame draw is deferred to near the end of the frame timing, it seems Gtk provides frame-pacing that is now not being used.
> Below: Current PR, multiple terminal draws are done within a single frame-timing. 
> This is due to changing from synchronised draws (single-buffer) to triple-buffering to not block draws on VSync

I suspect our latency bug comes this PR moving the BlitFrameBuffer for drawing the final terminal into a DMA texture being on the renderer thread and not being flushed (synced) before the DMA read in the main thread's Gtk GL context.

~~Otherwise the excess draw calls is a problem for this PR as well.~~
Edit: Noticed the extra draw calls are intentional from enabling triple buffering

AnthonyZhOon:
Here's a clearer example.
image
The OpenGL API call for BlitFrameBuffer made in the previous frame but the execution in OpenGL HW is deferred to the next frame's draw. The latency is from Gtk's idle refresh rate.
At this point, is it worth moving the OpenGL rendering to multiple threads?

AnthonyZhOon:
Here's a patch fix, move the gl.finish() till after we blit for presenting to the app surface.
```
diff --git a/src/renderer/opengl/Frame.zig b/src/renderer/opengl/Frame.zig
index 4128887e9..d7a3f3e44 100644
--- a/src/renderer/opengl/Frame.zig
+++ b/src/renderer/opengl/Frame.zig
@@ -56,7 +56,6 @@ pub inline fn renderPass(
 /// pushing the newly presented and exported frame to the frame queue.
 pub fn complete(self: *const Self, sync: bool) void {
     _ = sync;
-    gl.finish();

     // If there are any GL errors, consider the frame unhealthy.
     const health: Health = if (gl.errors.getError()) .healthy else |_| .unhealthy;
@@ -74,6 +73,7 @@ pub fn complete(self: *const Self, sync: bool) void {
         // Notify the surface that it should redraw
         _ = self.renderer.surface_mailbox.push(.redraw, .{ .forever = {} });
     }
+    gl.finish();

     // Report the health to the renderer.
     self.renderer.frameCompleted(health);
```
This is a latent bug in the OpenGL renderer. No more latency issues from trying it out

dkinzler:
I can confirm that the above patch fixes the lag for me, runs completely normal now.

> OK so, bad news. Nvidia sucks. There's no EGL extension for exporting DMABUFs on Nvidia drivers so we have to add a fallback path that copies each frame into CPU memory instead (yikes). 

Just to make sure, do you mean `EGL_MESA_image_dma_buf_export` by this? Because that extension is supported by my Nvidia driver (as reported by eglinfo and in Ghostty with `eglQueryString` after creating the display) and I don't get any errors in `Target.exportDmabuf`. Or maybe not in older driver versions?

pluiedev:
~~Ok I've opted to do this the more cursed and long-winded way. Hopefully it should work better on more drivers. cc @vancluever @Uzaaft for testing~~

Scratch that, this isn't working, let's fallback to the old approach. It's curious that Nvidia actually supports the Mesa extension, which didn't come up in my research. If it works well for you then I'll take your testimony

src/renderer/OpenGL.zig
```diff
@@ -42,20 +43,69 @@ alloc: std.mem.Allocator,
...
+egl_display: *gl.egl.Display,
+egl_context: *gl.egl.Context,
+
+pub fn init(alloc: Allocator, opts: rendererpkg.Options) !OpenGL {
+    try egl.load();
+
+    const display: *egl.Display = try .init(egl.c.EGL_DEFAULT_DISPLAY);
+
+    log.info("EGL vendor={s}", .{display.queryString(.vendor) orelse "(unknown)"});
+    log.info("EGL extensions={s}", .{display.queryString(.extensions) orelse "(unknown)"});
```
vancluever, line 55: Minor nit: I think this is line-broken on main right now; possibly coming through drivers and not necessarily Ghostty, but it might help with readability, e.g.:

```
EGL API version 1.5 found
 - Vendor: NVIDIA
 - Version: 1.5
 - Device: /dev/dri...
 - Client APIs: OpenGL_ES OpenGL
 - Extensions:
        ...
```

jcollie: Pretty sure that this is logged directly to stderr by GTK. There's not much we can do without an upstream code change.

vancluever: Yeah, just saying maybe some formatting on this line to mimic said logging might be in order. Not a big deal either way 🙂

AnthonyZhOon:
Noticed triple buffering is the intended OpenGL behaviour now, so drawing extra discarded frames is intentional, and should be the same as the Metal behaviour. (GTK4 does double-buffering, I wonder if there's triple buffering is worth it in practice for both renderers?)
In the future, we could optimize triggering a draw in `Termio.zig` for any read, even if it doesn't change visual state, could be changed in the future for power saving? Just something I noticed when mpv plays a 60fps video using kitty the terminal runs at 160fps (monitor refresh) and draws multiple times per present despite no state change I think.
```
fn processOutputLocked(self: *Termio, buf: []const u8) void {
    // Schedule a render. We can call this first because we have the lock.
    self.terminal_stream.handler.queueRender() catch unreachable;
```

mitchellh:
We're going to _send this_. 

Ghostty 1.4 is soon but if we can get some weeks for this to bake that'd be fantastic. Thanks @pluiedev

mitchellh:
I should add, if we get a ton of reports of Linux breaking I will revert it and we can fix up and go again another time.

neobrain:
> I should add, if we get a ton of reports of Linux breaking I will revert it and we can fix up and go again another time.

I'm just one data point, but this PR breaks startup for me: 49b95d8 throws errors like "MESA-EGL: warning: egl: failed to create dri2 screen" and ultimately crashes when `vkAllocateMemory` returns an error. d30379c (the latest version before this PR) is running fine. Both tested using your bundled nix flake.

Running with Lavapipe (software rendering) OR with `GSK_RENDERER=gl` works (though curiously rendering is garbled for certain window sizes after this PR).

This is running an Nvidia GPU (GB10, DGX Spark) on COSMIC (Wayland) with the proprietary GPU driver.

pluiedev:
Please open a new discussion instead of adding details here. Filling in the template is necessary for informed debugging.

PR #14054, opened 2026-08-28
terminal: mark the previous row dirty when clearing its spacer head

fornwall:
Erasing a wrapped wide character at the start of a row (`ECH` or `DCH`) also clears the spacer head it left at the end of the previous row, but that row was never marked dirty. With both rows visible, an incremental render kept the stale spacer head on screen until something unrelated redrew that row.

The clearing happens in the row-start branch of `splitCellBoundary`. `clearCells` doesn't do dirty tracking, and both callers only mark the cursor row, so mark the previous row at the point it's mutated.

The added dirty assertions fail without the fix.

## AI Disclaimer
Claude did the heavy lifting - identifying the root cause, generating code and description. I reviewed and iterated on it to move around and tweak tests, comments and reduce verboseness. Verified the end user visible behaviour improvement with a script that coloured the wide character, which made the stale rendering visible until a switch to the alt screen and back cleared it it.

### DIFF
--- a/src/terminal/Screen.zig
+++ b/src/terminal/Screen.zig
@@ -1949,6 +1949,11 @@ pub fn splitCellBoundary(
                         p_rac.row,
                         p_cells[p_row.node.cols() - 1 ..][0..1],
                     );
+
+                    // `clearCells` does not mark rows dirty, and our
+                    // callers only mark the cursor row, so mark the
+                    // previous row here.
+                    p_row.markDirty();
                 }
             }
         }
--- a/src/terminal/Terminal.zig
+++ b/src/terminal/Terminal.zig
@@ -9922,6 +9922,38 @@ test "Terminal: eraseChars wide char wrap boundary conditions" {
     }
 }
 
+test "Terminal: eraseChars clearing wrapped wide char marks spacer head row dirty" {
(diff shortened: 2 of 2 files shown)


mitchellh:
Spacer heads don’t participate or affect rendering so they don’t need to participate in dirty tracking. Or am I wrong

fornwall:
Spacer heads carry style, right? Standing at the end of line, print (with a background color) a wide character so it wraps -> the last column (carrying the spacer head) is rendered with the background color.

And if the wrapped wide char is erased that clears the last column to the erase background, meaning that the previous row needs re-rendering.

fornwall:
> Standing at the end of line, print (with a background color) a wide character so it wraps -> the last column (carrying the spacer head) is rendered with the background color.

It's not obvious if this is the correct behaviour - it means that printing a single background colored wide character at end of line colors three cells: the spacer head at end of the previous line, and the two wide char cells at the start of the next line.

Screenshot 2026-08-28 at 02 45 00


But a motivation for why this, the current behaviour (minus the missing dirty tracking), is correct/desired: Otherwise a long stretch of uniformly background colored content would have a hole in the last column here.

PR #14055, opened 2026-08-28
i18n: command palette - Focus and Untitled translation support

EriksRemess:
"Focus: {s}" and "Untitled" was not included in command palette translations. This PR adds support and placeholders in .po files for those strings.


pluiedev:
This will likely have to be deferred to the 1.5 release since the 1.4 string freeze is already in effect.

trag1c (changes requested):
what Leah said :) also the `POT-Creation-Date` timestamps should be updated

EriksRemess:
POT timestamp updated

00-kat (approved):
(Ack *changes*, but this won't be merged for a bit of course.)

PR #14059, opened 2026-08-28
Wrong repo, delete this

aflat:
When a session is reloaded, the window resize didn't work. It also didn't exit cleanly when you tried to exit one of the panes in a split pane window


aflat:
Woops, wrong repo, this shoudln't be merged

PR #14061, opened 2026-08-28
gtk: implement kitty clipboard protocol

### LINKED ISSUE #10549: Implement the Kitty Clipboard Protocol (OSC 5522)

jcollie:
This PR wires up the kitty clipboard protocol in the GTK apprt after all of the work to implement the protocol in the core and the macOS apprt. Adds an image preview to the confirmation dialog.

Since we are in string freeze for 1.4, two new strings added for the confirmation dialog were not marked for translation. Once 1.4 is released they can be marked for translation for 1.5.

Fixes #10549 

AI disclosure: Claude Fable 5 was used to develop this PR but the author has thoroughly reviewed the code and has rewritten sections.


src/apprt/gtk/class/clipboard_confirmation_dialog.zig
```diff
@@ -207,6 +224,35 @@ pub const ClipboardConfirmationDialog = extern struct {
...
+        // session grant for the password supplied by the program.
+        if (comptime can_remember) switch (req.*) {
+            .kitty_read, .kitty_write => {
+                // TODO: mark these strings for translation with
+                // i18n._() and regenerate the translation files.
+                priv.remember_choice.as(adw.PreferencesRow).setTitle(
+                    "Remember choice for this terminal session",
+                );
+                priv.remember_choice.as(adw.ActionRow).setSubtitle(
+                    "Future requests with the same password will be allowed",
```
vancluever, line 240: I think that _password_ here might be more of an implementation detail (I think that session ID/password combination and how this looks like will likely be something that varies from application to application), so I think we should change this to something more generalized.

Additionally, we might want to just *deny* all future requests for an application (e.g., just block it), so that needs to be taken into account too.

src/apprt/gtk/class/surface.zig
```diff
@@ -4129,6 +4167,445 @@ const Clipboard = struct {
...
+    /// into the other.
+    fn kittyImagePreview(contents: anytype) ?*gdk.Texture {
+        for (contents) |content| {
+            if (!std.mem.startsWith(u8, content.mime, "image/")) continue;
+            if (content.data.len == 0) continue;
+
+            const bytes = glib.Bytes.new(content.data.ptr, content.data.len);
+            defer bytes.unref();
+
+            // TODO: use glycin directly here so untrusted image data
```
vancluever, line 4273: This is a good idea and I think we should get this done sooner rather than later.

src/apprt/gtk/class/surface.zig
```diff
@@ -4129,6 +4167,445 @@ const Clipboard = struct {
...
+        state: apprt.ClipboardRequest,
+    ) void {
+        const kitty = state.kitty_write;
+        const surface = self.private().core_surface orelse {
+            kitty.destroy();
+            return;
+        };
+
+        surface.completeClipboardRequest(state, .{}) catch |err| switch (err) {
+            error.UnauthorizedPaste => showKittyConfirmation(
```
vancluever, line 4204: Minor nit: I know this is baked into the higher-level surface but since this is the error for both reads and writes (pastes and copies basically), we might want to look at updating the error name to reflect.

vancluever (changes requested):
Sent along some other feedback (crash logs namely) through other channels, other feedback inline here.

src/apprt/gtk/class/surface.zig
jcollie: Yes, a good idea but too big for this PR.

src/apprt/gtk/class/surface.zig
jcollie: I'll leave that for another PR, I'd rather not touch the core as part of the PR.

src/apprt/gtk/class/clipboard_confirmation_dialog.zig
jcollie: I've changed it to "Future requests from the same application will be allowed". Despite the fact that Kitty calls it a "password" it's really not and treating it as such is wrong.

vancluever:
So some of the "remember" boxes seem broken now; most of the MIME types don't display the switch (it's showing up for the multipart test but not the others, other than plain text which is displaying the one from the blueprint - more on that below).

The multi-part preview looks great - I think we could later add friendly names so that we're not just showing raw MIME types, e.g., "Plain Text" for text/plain, "HTML" for text/html, "IMAGE_TYPE image for image/..., etc, but yeah, later.

Images don't stretch anymore so that's good to see.

For the "remember" text/subtext, I think actually the text currently in the blueprint itself might actually be better for this. Main reason being is that, again, the token is an implementation detail and the user will likely not even know it exists (a compliant application will likely choose a random session key and always use it without any option from the user to use it or not).

image

Finally, selecting "deny" with remember turned on still does not remember the choice (so you will continue to get the consent dialog on future attempts).

jcollie:
@vancluever can you share what requests do/dont get the remember choice switch? The rules for when it shows up are complicated - OSC52 always gets it, but for OSC5522 it only gets it if _both_ a `name=` _and_ a `pw=` are in the OSC5522 request. If that condition is not met the remember choice switch is never shown for OSC5522 requests. The `test/clipboard-multipart.py` script explicitly sets both which is why you see it there.

This reliably shows the remember choice switch:
```
kitty +kitten clipboard --password text:1234 --human-name "Evil Corp" --mime
text/plain build.zig
```

As far as remembering denies for OSC5522 requests, that's something in the core - it only records allows, never denies. So without core changes (which would need to be mirrored on the macOS side) we _can't_ record denies.

There's also the fact that the Kitty password is a very weak mechanism. All a malicious program has to do is pick a new one and the user gets asked again. In fact, I'd bet that many non-malicious programs that implement OSC5522 won't bother to save the password somewhere stable so on every new invocation the user will be bothered again.

I think having the same text for OSC52 and OSC5522 isn't a great option as it hides the fact that OSC5522 grants are more granular _and_ have different conditions in which they get reset (OSC5522 grants don't get reset with config reloads, only when the terminal is reset).

I think that some of these are bigger design decisions than the current PR allows. I don't have a mac to test but I'd assume that some of these same concerns apply there.

vancluever:
Approving this with some final parting recommendations. I think that these should be addressed, but I think they've been surfaced enough that we are aware of them for future work.

> @vancluever can you share what requests do/dont get the remember choice switch? The rules for when it shows up are complicated - OSC52 always gets it, but for OSC5522 it only gets it if _both_ a `name=` _and_ a `pw=` are in the OSC5522 request. If that condition is not met the remember choice switch is never shown for OSC5522 requests. The `test/clipboard-multipart.py` script explicitly sets both which is why you see it there.

Right, yeah, I forgot about that part of the spec/implementation when I did the follow up. Checked again and this seems fine.

> As far as remembering denies for OSC5522 requests, that's something in the core - it only records allows, never denies. So without core changes (which would need to be mirrored on the macOS side) we _can't_ record denies.

Hmm, okay, yeah that should be addressed. Maybe someone doesn't want it allowed on a certain app for security/opsec reasons; not necessarily to stop a malicious application, but possibly to keep an app from sending data that might be sensitive to the clipboard, etc.

> There's also the fact that the Kitty password is a very weak mechanism. All a malicious program has to do is pick a new one and the user gets asked again. In fact, I'd bet that many non-malicious programs that implement OSC5522 won't bother to save the password somewhere stable so on every new invocation the user will be bothered again.

I think all that's really given in the spec as it stands is a transport, and while it's simple, I don't think it's inherently insecure (barring, well, the lack of encryption on the transport in a number of scenarios, but SSH et al exist, of course). I think that the real issues of security depend on implementation and use case. Consider the [section](https://sw.kovidgoyal.net/kitty/clipboard/#avoiding-repeated-permission-prompts) of the spec:

> The editor or other program using this facility should ideally use a password randomly generated at startup, such as a UUID4. However, terminals may implement permanent/stored passwords. Users can then configure terminal programs they trust to use these password.

This supplies a fairly solid example, namely an ephemeral secret that gets generated at startup, or a configurable shared secret. I'd expect any program that was unable to conform to even this simple reference implementation to not implement it at all. Server side (e.g., _our_ side) I think going forward it would be good for us to implement this in a robust and scalable way. This could be a simple key/value pair of trusted applications for now and their tokens, with more secure options as implementations evolve. For the time being, I think the session-based store is fine, but this is something I'd recommend for a follow-up.

Also:

> All a malicious program has to do is pick a new one and the user gets asked again.

I think this is a scenario where it would be on us to indicate that there is a saved secret mismatch and give a big bad warning about, something akin to an SSH host key mismatch. This is a scenario where there's a (possible) breach of trust and should be treated as such.

> I think having the same text for OSC52 and OSC5522 isn't a great option as it hides the fact that OSC5522 grants are more granular _and_ have different conditions in which they get reset (OSC5522 grants don't get reset with config reloads, only when the terminal is reset).

Right, but when is the average user of the terminal going to know about this or care? A consistent UX is important here, or people are going to be left wondering why they get one dialog and set of domain language for one program, a different one for another, or even different ones for the same program depending on what OSC is used (think even different editor extensions).

vancluever (approved):

PR #14069, opened 2026-08-29
main: fix inverted allow_stack_tracing condition

jcollie:
The Zig 0.16.0 update dropped the negation from the std default (!strip_debug_info), disabling stack traces in every unstripped build.

AI disclosure: Claude Fable was used to diagnose the problem and find the fix. Commit message was written by me.

Claude-Session: https://claude.ai/code/session_01QfzQME46DQXwMWa43bQaa3


pluiedev (approved):
Oh I was *so* confused why stack traces suddenly refuse to show up, lets go

PR #14070, opened 2026-08-29
terminal/kitty: accept shm names without a leading slash

j-c-m:
POSIX portable shm names start with '/'. Darwin shm_open requires that form, glibc accepts both, so any Linux client may omit the slash from the Kitty payload while still creating a valid object. mpv currently does this (mpv-player/mpv#18424), which is how the mismatch showed up: --vo-kitty-use-shm fails on macOS while q=2 hides the error.
                                                                                                         
Prepend '/' when the payload name does not already start with one.

### DIFF
--- a/src/terminal/kitty/graphics_image.zig
+++ b/src/terminal/kitty/graphics_image.zig
@@ -211,9 +211,13 @@ pub const LoadingImage = struct {
         }
 
         // Since we're only supporting posix then max_path_bytes should
-        // be enough to stack allocate the path.
-        var buf: [std.fs.max_path_bytes]u8 = undefined;
-        const pathz = std.fmt.bufPrintZ(&buf, "{s}", .{path}) catch return error.InvalidData;
+        // be enough to stack allocate the path. Extra byte is for a
+        // leading '/' when the client omitted it (Darwin requires it).
+        var buf: [std.fs.max_path_bytes + 1]u8 = undefined;
+        const pathz = if (path.len > 0 and path[0] != '/')
+            std.fmt.bufPrintZ(&buf, "/{s}", .{path}) catch return error.InvalidData
+        else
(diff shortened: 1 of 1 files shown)


jcollie (approved):

jcollie:
Hmmmm. Maybe we shouldn't be papering over bugs in mpv? Not sure how long it'll take mpv to merge the patch and cut a release (and that release make into distros).

j-c-m:
I think it is much more prevalent than that as Linux (glibc) accepts both forms, so anything that targets that may or may not have the leading slash.

j-c-m:
And a well written tui app will silently fallback from shm to a lower performance tty base64 dump, so it's a tricky one.

AnthonyZhOon:
How do kitty and other terminals behave for this case?

j-c-m:
> How do kitty and other terminals behave for this case?

https://github.com/kovidgoyal/kitty/pull/10422

Kitty has the same failure on macOS. Usually what happens is the TUI tries shm, fails, then does the tty base64 send as a failover. So it's difficult to notice as the app is just "slower".

mitchellh:
We must wait for Kitty’s behavior to match it. At the moment if this doesn’t work in Kitty then this is working properly in Ghostty

j-c-m:
Kitty has hardened the spec. So far I've only found mpv out of posix compliance, hopefully that patch lands. Closing this PR. Spec requires leading /, Ghostty is compliant.


https://github.com/kovidgoyal/kitty/commit/22042970cf3a4668d02a1a7bcccca778ec864c21

jcollie:
Kovid added/adopted some restrictions that we might want to enforce inside Ghostty before passing them to the OS:

image

mitchellh:
Done: https://github.com/ghostty-org/ghostty/pull/14080

j-c-m:
Mpv has merged the fix.

https://github.com/mpv-player/mpv/pull/18424

PR #13678, opened 2026-08-06
nix: update nixpkgs-unstable

jcollie:
Update nixpkgs-unstable to pick up fontconfig 2.18. Currently we are linking against 2.17 and you get errors like this on standard error when using config files meant for fontconfig 2.18:

  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 20: invalid constant used :
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 23: invalid constant used : monospace
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 42: invalid attribute 'xsi:nil'
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 43: invalid constant used :
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 46: invalid constant used : sans-serif
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 68: invalid attribute 'xsi:nil'
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 69: invalid constant used :

PR #14071, opened 2026-08-29
pkg/fontconfig: update to 2.18.3

jcollie:
If your distro ships 2.18-based config files, we get many errors like the following when we compile Ghostty against our vendored copy of Fontconfig 2.14.

  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 19: invalid attribute 'xsi:nil'
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 20: invalid constant used :
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 23: invalid constant used : monospace
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 42: invalid attribute 'xsi:nil'
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 43: invalid constant used :
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 46: invalid constant used : sans-serif
  Fontconfig warning: "/etc/fonts/conf.d/48-guessfamily.conf", line 68: invalid attribute 'xsi:nil'

Nixpkgs was updated in #13678 to pick up a newer Fontconfig when building against the system fontconfig.

Fontconfig no longer ships any generated headers as part of the official tarball. Generated them and added/updated them in the override subdir. Added new macros that are needed. Added new source files. Derive FC_VERSION_* macros from build.zig.zon version.

AI disclosure: Claude Fable 5 was used to develop this PR. Changes thorougly reviewed by the author.

Claude-Session: https://claude.ai/code/session_01QfzQME46DQXwMWa43bQaa3


vancluever:
Would be nice to have something in the `build.zig` so so that the generated files can be refreshed going forward, sounds like it could be a similar thing to wayland-protocols, etc.

vancluever:
PS: I actually think that the generation for wayland-protocols is actually handled in cache (see [here](https://github.com/ghostty-org/ghostty/blob/main/pkg/gtk4-layer-shell/build.zig#L133-L170) in the gtk4-layer-shell `build.zig`). Hence the `override` dir might actually be able to be removed in favor of this completely and the generated code can be kept out of VCS, and treated like an intermediary build artifact.

jcollie:
That would be great but that requires duplicating a lot of the meson build machinery. I think really the better question is do we really want to keep the ability to build fontconfig on platforms where we can't rely on dynamically linking to the system copy. That's a common cause of problems when people build/package fontconfig on platforms that do have a system copy of fontconfig.

vancluever:
My concern is that generating these files is not a process that has been captured anywhere, and is also seemingly complex judging from what you've mentioned. So when we go to upgrade this again to keep expectations in line, the generation process will need to be parsed out again; it seems like it would be worth the time to add port the necessary functionality for generation and maintain it going forward if need be.

If doing this is too much of a maintenance burden I'd definitely err on just removing static linking to fontconfig (I'm assuming that's what this would mean), given that any system that lacks a dynamic copy would (seemingly) have an ambiguous user experience anyway.

jcollie:
closing in favor of #14113

PR #14072, opened 2026-08-29
build: update Sparkle to 2.9.4

Svector-anu:
Update the macOS Sparkle dependency from 2.9.0 to 2.9.4.

This keeps the Swift package resolution and all tag/tip release workflow downloads aligned on the same version. Sparkle 2.9.2 included fixes for GHSA-g3hp-f6mg-559v and GHSA-hg88-v3cw-3qrh; 2.9.4 is the current stable release.

Validation:
- verified the 2.9.4 release contains `Sparkle-for-Swift-Package-Manager.zip`
- verified the lockfile revision matches the 2.9.4 tag
- `jq empty` on `Package.resolved`
- `git diff --check`

I could not run Xcode package resolution locally because the active developer directory is Command Line Tools rather than a full Xcode installation.

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -160,7 +160,7 @@ jobs:
 
       - name: Setup Sparkle
         env:
-          SPARKLE_VERSION: 2.9.0
+          SPARKLE_VERSION: 2.9.4
         run: |
           mkdir -p .action/sparkle
           cd .action/sparkle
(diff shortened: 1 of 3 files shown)


bo2themax:
!vouch

bo2themax (approved):
Waiting CI

bo2themax:
> 2.9.4 is the current stable release. 

This isn't quite right now, can you change it to 2.9.6?

mitchellh:
Ill just take this for now, but we can use a 2.9.6 update before 1.4

PR #14076, opened 2026-08-29
freestanding support

Uzaaft:


mitchellh (commented):
This all looks very reasonable so far. Maybe lets add a freestanding CI target with your esp board or something cross-compile to verify.

Uzaaft:
@mitchellh I'll look into adding `riscv32-freestanding-eabi` and `thumb-freestanding-eabi` compile steps to the CI/CD. Is it safe to assume that we probably want this to be outside of the current matrix of supported targets(win, mac, linux, and bsd? 

Also, should I look into running the tests as well? I fear that it might blow up the CI times since we'll probably have to use qemu.

.github/workflows/freestanding.yml
```diff
@@ -0,0 +1,51 @@
+on:
+  push: {}
+  pull_request: {}
+  workflow_dispatch: {}
+
+name: Freestanding
```
mitchellh: Lets add this to the existing test.yml

src/terminal/PageList.zig
```diff
@@ -538,12 +540,15 @@ fn initialCapacity(cols: size.CellCountInt) Capacity {
     return cap;
 }
 
-/// This is the page allocator we'll use for all our underlying
-/// VM page allocations.
-inline fn pageAllocator() Allocator {
+/// This is the allocator we'll use for all our underlying page allocations.
```
mitchellh: Improve this comment to explain what the param alloc is used for.

mitchellh (changes requested):

.github/workflows/test.yml
```diff
@@ -649,6 +649,45 @@ jobs:
           nix develop -c zig build -Demit-lib-vt \
             -Dtarget=${{ matrix.target }}
 
+  build-libghostty-vt-freestanding:
```
mitchellh, line 653: This needs to be added to the big list at the top of required tests to pass.

Uzaaft: Done.

mitchellh (approved):
Approved pending CI.

PR #14080, opened 2026-08-30
terminal/kitty: validate POSIX shared memory names

mitchellh:
Update shared memory name validation according to the new spec:

https://github.com/kovidgoyal/kitty/commit/22042970cf3a4668d02a1a7bcccca778ec864c21

### DIFF
--- a/src/terminal/kitty/graphics_image.zig
+++ b/src/terminal/kitty/graphics_image.zig
@@ -210,6 +210,15 @@ pub const LoadingImage = struct {
             return error.UnsupportedMedium;
         }
 
+        // POSIX shared memory names must begin with a slash, contain at
+        // least one character after it, contain no other slashes, and fit
+        // within NAME_MAX. Some shm_open implementations accept names
+        // without the leading slash, but the Kitty protocol does not.
+        if (!validSharedMemoryName(path, posix.NAME_MAX)) {
+            log.warn("invalid shared memory name", .{});
+            return error.InvalidData;
+        }
+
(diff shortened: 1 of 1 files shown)


src/terminal/kitty/graphics_image.zig
```diff
@@ -831,6 +840,16 @@ pub const Rect = struct {
     }
 };
 
+/// Returns whether a name follows the POSIX shared memory name format.
+fn validSharedMemoryName(name: []const u8, name_max: usize) bool {
+    if (name.len < 2 or name.len > name_max or name[0] != '/') return false;
+    for (name[1..]) |c| {
+        if (c == '/' or c == 0) return false;
+    }
```
jcollie, line 848: `std.mem.findAnyPos`?

jcollie (approved):
One small possible improvement, otherwise LGTM

mitchellh:
Too lazy to address the feedback, possibly an improvement but not a big deal lol.

j-c-m:
Ghostty is spec compliant, the client caveat is that on macOS, it actually limits you to 31 characters including the / for shm names. (Although NAME_MAX is 255, which doesn't include the / on Linux!). Ghostty still handles this, shm_open fails, Ghostty returns invalid data.

https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/sys/posix_shm.h#L53-L54
```
#define PSHMNAMLEN      31      /* maximum name segment length we bother with */
```

I am going to submit a doc update to Kitty.

PR #14081, opened 2026-08-30
terminal: fix living item over-count in RefCountedSet.addWithId

mitchellh:
Reported in https://github.com/ghostty-org/ghostty/discussions/14064

I validated this myself manually. The zero-ref branch of `addWithIdContext` incremented `living` unconditionally even if `upsert` resolved the value to an item that was already alive under a different ID.

This would cause `living` to be invalid for each time this happened and the downstream effect was that `count()` drifted. I couldn't find any crashing or invalid effect except that this caused requested style memory  to be over-provisioned.

cc @qwerasd205 since its ref counted set, but I did this work manually ❤️

### DIFF
--- a/src/terminal/ref_counted_set.zig
+++ b/src/terminal/ref_counted_set.zig
@@ -317,6 +317,19 @@ pub fn RefCountedSet(
 
             if (id < self.next_id) {
                 if (items[id].meta.ref == 0) {
+                    // Requested ID is dead, but if the value exists not under
+                    // another ID then ref count that and increase the ID.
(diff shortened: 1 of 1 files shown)


qwerasd205:
I'll give this a review today

qwerasd205 (approved):
The change looks good, it makes sense. I believe I have comments at a couple of the callsites for `addWithId` questioning whether it offers any performance benefit. I don't remember whether I ended up benchmarking its removal or not, but it might make sense to just drop this function altogether and use the normal add.

It does make sense to merge this now though, even if I do end up benchmarking and deciding that it's not worth having this function and removing it.

PR #14085, opened 2026-08-30
gtk: do not warn when gtk-xft-dpi is -1

mgsloan:
Before this change, ghostty frequently logs the following warning, even though a `gtk-xft-dpi` value of `-1` is valid and indicates default scaling.

```
warning(gtk_ghostty_surface): gtk-xft-dpi has invalid value (-1), using default
```

From [the gtk docs](https://docs.gtk.org/gtk4/property.Settings.gtk-xft-dpi.html):

> The font resolution, in 1024 * dots/inch.
>
> -1 to use the default value.

### DIFF
--- a/src/apprt/gtk/class/surface.zig
+++ b/src/apprt/gtk/class/surface.zig
@@ -1572,12 +1572,13 @@ pub const Surface = extern struct {
                 break :xft_scale 1.0;
             };
 
-            // Use a value of 1.0 for the XFT DPI scale if the setting is <= 0
-            // See:
-            // https://gitlab.gnome.org/GNOME/libadwaita/-/commit/a7738a4d269bfdf4d8d5429ca73ccdd9b2450421
+            // -1 specifies default scale. See:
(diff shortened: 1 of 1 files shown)


src/apprt/gtk/class/surface.zig
```diff
@@ -1572,12 +1572,13 @@ pub const Surface = extern struct {
                 break :xft_scale 1.0;
             };
 
-            // Use a value of 1.0 for the XFT DPI scale if the setting is <= 0
-            // See:
-            // https://gitlab.gnome.org/GNOME/libadwaita/-/commit/a7738a4d269bfdf4d8d5429ca73ccdd9b2450421
```
jcollie, line 1577: Let's keep this comment in.

jcollie (changes requested):

src/apprt/gtk/class/surface.zig
mgsloan: Done!  Thanks for reviewing

jcollie (approved):
Approved pending CI

PR #14088, opened 2026-08-30
Update mk localization for v1.4

andrejdaskalov:
Addressing #13766 for mk.


andrejdaskalov:
@marijagjorgjieva When reviewing this, can you focus specifically on line 1123 in `po/mk.po`? I may be missing some better phrasing. As I understand it, the command inserts a ghost emoji character into the terminal. But I think the original phrasing was both "Put a little (bit of) Ghostty in your terminal." and also "Put a little (small) Ghostty in your terminal."
The double phrasing is a bit lost in our translation, which focuses more on the literal functionality (inserting a ghost character). 

I may be wrong about the original phrasing though, so maybe some of the maintainers can weigh in if that is the case.

marijagjorgjieva (approved):

trag1c (approved):

PR #14108, opened 2026-09-01
gtk: clear surface notifications on focus and close

EriksRemess:
GTK notifications could remain visible after their originating surface was focused or closed. macOS already clears them in these cases.
This tracks notifications per surface and withdraws them on focus or close. Notifications sent while the surface is already focused are removed after three seconds, just like on macOS.
Repeated notifications with the same text replace each other, while notifications from different surfaces remain independent. Clicking a notification still focuses the originating surface.

AI disclosure: Codex 5.6 Sol helped with most of the implementation.


EriksRemess:
Related discussions: https://github.com/ghostty-org/ghostty/discussions/13956 https://github.com/ghostty-org/ghostty/discussions/13952 https://github.com/ghostty-org/ghostty/discussions/12786

pluiedev (changes requested):
I think this is a bit overengineered and there should be a way to implement the same logic but more easily. Calculating the hash manually shouldn't be required as long as we don't use *auto* hash maps and instead implement our own hash function. We should also use array hash maps instead as IIRC normal hash maps have some unbounded memory usage problems. Notification tracking should also be moved to a separate file as `surface.zig` is enormous enough as-is.

EriksRemess:
Moved desktop notification functionality out of surface.zig, simplified a bit.

src/apprt/gtk/class/surface.zig
```diff
@@ -1777,29 +1781,22 @@ pub const Surface = extern struct {
...
-        defer icon.unref();
-        notification.setIcon(icon.as(gio.Icon));
-
-        const pointer = glib.Variant.newUint64(core_surface.id);
-        notification.setDefaultActionAndTargetValue(
-            "app.present-surface",
-            pointer,
+        priv.desktop_notifications.send(
+            app.allocator(),
+            app.as(gio.Application),
```
pluiedev: Throughout the codebase we just assume there's one global Application object. Just fetch `Application.default()` whenever you need it.

src/apprt/gtk/desktop_notifications.zig
```diff
@@ -0,0 +1,370 @@
+//! Sends and clears desktop notifications associated with a single GTK surface.
+//! Access this object only on the GTK main thread and keep its address stable
+//! until deinit has cancelled its timers.
```
pluiedev: This is pretty much a given for the entirety of `src/apprt/gtk` and is therefore redundant. When prompting you want the LLM to emit fewer pointless comments like these.

src/apprt/gtk/desktop_notifications.zig
```diff
@@ -0,0 +1,370 @@
...
+    fn deinit(self: Key, alloc: Allocator) void {
+        alloc.free(self.title);
+        alloc.free(self.body);
+    }
+
+    fn hash(self: Key) u64 {
+        var hasher = std.hash.Wyhash.init(0);
+        hasher.update(self.title);
+        hasher.update("\x00");
+        hasher.update(self.body);
```
pluiedev: I'm not an expert on hashing algorithms but I don't see a reason why we should use NUL to separate the strings here. The standard library's autohasher instead hashes the length of the string after each string (see https://ziglang.org/documentation/master/std/#std.hash.auto_hash.hashPointer w/ deep or deep recursive hash strategy).

Also the whole point of the Context struct is that you don't have to define `hash` and `eql` on the type directly.

src/apprt/gtk/desktop_notifications.zig
```diff
@@ -0,0 +1,370 @@
...
+    focused: bool,
+    title: [:0]const u8,
+    body: [:0]const u8,
+) void {
+    assert(surface_id != 0);
+    if (self.surface_id == 0) {
+        self.surface_id = surface_id;
+    } else {
+        assert(self.surface_id == surface_id);
+    }
```
pluiedev: This is very strange. Is there any possible reason why the notif object would somehow be shared across different surfaces? Ideally the surface id would be set once the core surface initializes when the widget realizes and becomes immutable thereafter.

src/apprt/gtk/desktop_notifications.zig
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/// Store content and return any evicted entry for withdrawal by the caller.
+fn track(
+    self: *DesktopNotifications,
+    alloc: Allocator,
+    title: []const u8,
+    body: []const u8,
+) Allocator.Error!TrackResult {
+    const key: Key = .{ .title = title, .body = body };
```
pluiedev: Just accept a key as the 3rd argument

src/apprt/gtk/desktop_notifications.zig
```diff
@@ -0,0 +1,370 @@
...
+
+/// Withdraw delivered notifications and cancel their timers. GIO's freedesktop
+/// backend cannot withdraw a notification until its asynchronous Notify reply
+/// arrives. A notification still in flight may therefore outlive this call.
+pub fn clear(
+    self: *DesktopNotifications,
+    alloc: Allocator,
+    app: *gio.Application,
+) void {
+    while (self.pop()) |removed| self.clearRemoved(alloc, app, removed);
```
pluiedev: Why not just iterate over the map directly? Since `pop` is an ordered remove, this will cause 1/2 N^2 element moves.

src/apprt/gtk/desktop_notifications.zig
```diff
@@ -0,0 +1,370 @@
...
+
+/// GIO requires the same ID for sending and withdrawing. Including the surface
+/// ID keeps identical notifications from different surfaces independent.
+/// The fixed buffer fits the prefix, two 16-digit hex values, and a terminator.
+fn formatId(
+    buf: *[64]u8,
+    surface_id: u64,
+    key: Key,
+) [:0]u8 {
+    return std.fmt.bufPrintZ(
```
pluiedev, line 252: ~~I think `bufPrintZ` has been replaced by `std.mem.printSentinel` now (with sentinel = 0).~~

Never mind, this is a master branch thing.

pluiedev (changes requested):
Overall I'm not that impressed with the code quality here. See detailed review below.

Also I'm not convinced a hash map is necessarily the best data structure here? Assuming the number of notifications stays small, you might be better off using a [std.Deque](https://ziglang.org/documentation/0.16.0/std/#std.Deque) instead of fiddling with custom hashing algorithms etc.

Merge conflicts need to be addressed as well.

src/apprt/gtk/class/surface.zig
EriksRemess: now uses Application.default().

src/apprt/gtk/desktop_notifications.zig
EriksRemess: removed obvious GTK commentary, kept ownership and timer explanations

src/apprt/gtk/desktop_notifications.zig
EriksRemess: replaced custom hashing code with the standard library

src/apprt/gtk/desktop_notifications.zig
EriksRemess: moved it to surface initialization

src/apprt/gtk/desktop_notifications.zig
EriksRemess: done

src/apprt/gtk/desktop_notifications.zig
EriksRemess: deque draining is now linear

src/apprt/gtk/desktop_notifications.zig
EriksRemess: ok

PR #14112, opened 2026-09-01
macos: pulse osc 9;4 progress on a layer sibling

j-c-m:
This PR dramatically reduces the resources used by a window displaying an indeterminate progress bar on macOS. Previously Ghostty would use ~6% and WindowServer ~40% when display a `printf '\033]9;4;3'` (indeterminate) progress bar, this change reduces that to Ghostty negligible and WindowServer ~5%. It does help with other progress bars as well, but they simply are much less animated to begin with.

---

Indeterminate OSC 9;4 progress used a SwiftUI repeatForever overlay on top of the Metal surface. That kept WindowServer compositing the whole terminal at display refresh for as long as a tool reported progress.

The bar is now a 2px NSView sibling of the scroll view, drawn with the view's own layer. Indeterminate is a full-width opacity pulse instead of a bouncing chunk: it changes opacity in 24 steps at 12 Hz with no interpolation between ticks, so the layer is not animated at display refresh.

AI: Grok & Claude used, code review, cleanup and manual tests performed the old fashioned way.

### DIFF
--- a/macos/Sources/Ghostty/Surface View/SurfaceProgressBar.swift
+++ b/macos/Sources/Ghostty/Surface View/SurfaceProgressBar.swift
@@ -1,20 +1,29 @@
-import SwiftUI
+import AppKit
 
 /// The progress bar to show a surface progress report. We implement this from scratch because the
 /// standard ProgressView is broken on macOS 26 and this is simple anyways and gives us a ton of
 /// control.
-struct SurfaceProgressBar: View {
-    let report: Ghostty.Action.ProgressReport
+///
+/// A 2px sibling of the scroll view (not a Metal overlay) so WindowServer does
+/// not composite the terminal. Uses this view's own layer.
+final class SurfaceProgressBar: NSView {
+    static let hairlineHeight: CGFloat = 2
 
-    private var color: Color {
+    private var report: Ghostty.Action.ProgressReport?
+    private var timer: Timer?
+    private var pulse = PulsingProgressBar()
+
+    private var color: NSColor {
+        guard let report else { return .controlAccentColor }
         switch report.state {
-        case .error: return .red
-        case .pause: return .orange
-        default: return .accentColor
+        case .error: return .systemRed
+        case .pause: return .systemOrange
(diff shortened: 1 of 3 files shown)


macos/Sources/Ghostty/Surface View/SurfaceProgressBar.swift
```diff
@@ -1,20 +1,29 @@
-import SwiftUI
+import AppKit
 
 /// The progress bar to show a surface progress report. We implement this from scratch because the
 /// standard ProgressView is broken on macOS 26 and this is simple anyways and gives us a ton of
 /// control.
-struct SurfaceProgressBar: View {
-    let report: Ghostty.Action.ProgressReport
+///
+/// A 2px sibling of the scroll view (not a Metal overlay) so WindowServer does
```
jparise: I'd skip naming the size in the comment when it's controlled by a constant below.

bo2themax:
Hmmm, I don't know about this. Current performance looks harmless to me. And I don't think now is a good time to change behaviours.

j-c-m:
Same machine (M2 pro) , one minute apart. Bounce = current `repeatForever` overlay on main. Pulse = this PR.

| | bounce (main) | pulse (PR) |
|---|---:|---:|
| Ghostty CPU | 126 ms/s (~13% of a core) | 8 ms/s (~0.8%) |
| Ghostty wakeups | 146/s | 14/s |
| Ghostty GPU | 0.00 ms/s | 0.00 ms/s |
| WindowServer CPU | 528 ms/s (~53% of a core) | 59 ms/s (~6%) |
| WindowServer wakeups | 577/s | 252/s |
| WindowServer GPU | 10.8 ms/s | 1.8 ms/s |
| GPU HW residency | 36.4% | 3.8% |
| GPU power | 21 mW | 3 mW |

j-c-m:
> Hmmm, I don't know about this. Current performance looks harmless to me. And I don't think now is a good time to change behaviours.

The efficiency impact is dramatic, metrics now attached to this thread.

PR #14122, opened 2026-09-02
renderer: smooth scrollback

j-c-m:
I am opening this PR because native smooth scroll back with a trackpad feels and looks absolutely magical compared to the current chunky line based jumps.

I've been using it for a few days across both macOS and Linux in my DD Ghostty branch: https://github.com/j-c-m/ghostty/tree/j-c-m

This follows the pixel delta. Finger down, the grid moves with you. Lift, and the OS coast keeps going. You can rest between rows. Tap, and it stops. Same input path as Safari. The terminal finally uses it.

Default is smooth-scrollback = mouse, no-key, no-jump. Page keys, find, and jump-to-prompt stay discrete unless you turn key / jump on. Alternate screen, mouse reporting, and live PTY output stay discrete.

Depends on the fractional-row painting included in this PR.

It does merge conflict with https://github.com/ghostty-org/ghostty/pull/14053, but I have already resolved, if needed. It's easier to merge this commit first.

AI: Some Grok & Claude, but less than I would have liked.

### DIFF
--- a/src/Surface.zig
+++ b/src/Surface.zig
@@ -316,6 +316,7 @@ const DerivedConfig = struct {
     mouse_hide_while_typing: bool,
     mouse_reporting: bool,
     mouse_scroll_multiplier: configpkg.MouseScrollMultiplier,
+    smooth_scrollback: configpkg.Config.SmoothScrollback,
     mouse_shift_capture: configpkg.MouseShiftCapture,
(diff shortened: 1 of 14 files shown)


mitchellh:
This diff is too big and touches too many core areas for a maintainer to review without the submitter being an expert on the systems without AI.

PR #14128, opened 2026-09-02
surface: restore application mouse shape after modifier overrides

j-c-m:
This fixes a regression of 9a6469743 introduced in 6e8ed4e8b.

With this fix the mouse pointer will be restored to the previous (which could have be set to something else via OSC22), not a hardcoded .text or .default.

It also will change the cursor to a text selection if shift is held even without mouse tracking, I think this is the expected behavior when a cursor is set via OSC22. (Kitty additionaly, once you start selecting changes to text (I-beam), this would be a follow-up if we desire to behave like kitty with OSC22 pointers).

### DIFF
--- a/src/surface_mouse.zig
+++ b/src/surface_mouse.zig
@@ -31,23 +31,10 @@ over_link: bool,
 /// True if the mouse pointer is currently hidden.
 hidden: bool,
 
-/// Translates key state to mouse shape (cursor) state, based on a state
-/// machine.
-///
-/// There are 4 current states:
-///
-/// * text: starting state, displays a text bar.
-/// * default: default state when in a mouse tracking mode. (e.g. vim, etc).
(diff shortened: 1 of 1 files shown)


j-c-m:
Simple test on main, printf '\e]22;pointer\a' , will turn the mouse pointer finger, hold shift will turn I-beam, release, it should turn back to the pointer finger but doesn't, subsequent printf '\e]22;pointer\a' will also fail to set the pointer.

On this fix it will reset back to the OSC22 set pointer.

vancluever:
@j-c-m thanks for catching this.

I'm just going to piggyback on your commit here to just simplify some things; I think the whole "state machine" thing that I was envisioning didn't necessarily materialize, so I'm just going to update some of the code to reflect. Otherwise looks good!

vancluever (approved):

PR #14133, opened 2026-09-03
libghostty-vt: let the png decode hook report an allocation cap

### LINKED ISSUE #12163: libghostty-vt: decode png hook should support a `cap` on bytes returned
This lets callers over-allocate which may be more efficient in some cases.

cc @pluiedev

ArneshBanerjee:
Fixes #12163

GhosttySysImage only carried data_len, and the library freed the pixel buffer
with that length. The allocator needs the real allocation length to free, so a
decode callback had to allocate exactly the pixel data. A buffer rounded up to
a block size, or one reused across images, had no way to be described.

Adds data_cap, the number of bytes allocated for data, which may be larger than
data_len. The library frees data_cap bytes and still reads data_len bytes of
pixels. Leaving it at 0 means the allocation is exactly data_len, so existing
callbacks keep working unchanged. A non-zero cap below data_len is a broken
callback, so the buffer is freed and the image rejected rather than leaked.

On the Zig side sys.Image gains a capacity field and a deinit that frees with
it, so the capacity-aware free lives in one place instead of at each call site.

Tested with the full zig build test and zig build test-lib-vt suites, plus new
tests in src/terminal/c/sys.zig covering the over-allocated, zero-cap, and
undersized-cap paths.

### DIFF
--- a/src/terminal/c/sys.zig
+++ b/src/terminal/c/sys.zig
@@ -12,6 +12,7 @@ pub const Image = extern struct {
     height: u32,
     data: ?[*]u8,
     data_len: usize,
+    data_extra: usize,
 };
 
 /// C: GhosttySysDecodePngFn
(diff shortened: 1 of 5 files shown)


src/terminal/c/sys.zig
```diff
@@ -96,15 +97,37 @@ fn decodePngWrapper(
...
+    // is exactly the pixel data.
+    const capacity = if (out.data_cap == 0) out.data_len else out.data_cap;
+
+    // A cap smaller than the pixel data is a broken callback: the pixels
+    // it reported don't fit in the allocation it reported. Free what it
+    // said it allocated so we don't leak, then reject the image.
+    if (capacity < out.data_len) {
+        alloc.free(result_data[0..capacity]);
+        return error.InvalidData;
+    }
```
jcollie: I'm curious about this. Wouldn't it be better to store the amount of _additional_ memory that had been allocated? Then you can always just free `result_data[0 . .out.data_len + out.data_extra])`?

ArneshBanerjee: That is better. Switched to `data_extra`.

It makes the bad state unrepresentable. With a total cap, a value below `data_len` is meaningless but still expressible, so the wrapper had to check for it and reject. With the surplus there is nothing to check, and 0 naturally means no extra.

I kept a checked add, since both lengths come from the callback and a wrapping sum would free a too-short slice. On overflow it rejects without freeing.

Force-pushed as it reworks the same field.

PR #14137, opened 2026-09-03
terminal: make all page data structures treat zero as empty to avoid eagerly paging in mmap pages

mitchellh:
This updates all our page data structures so that the `0` value (literally `@memset(0)`) means empty. This way, when we initialize a new page via mmap (OS-guaranteed zeroed), we don't need to write to it, and don't trigger the kernel to physically map the memory.

From Ghostty 1.3.1, our empty terminal physical memory usage goes from 128 KB to 48 KB (#14130) to 16 KB (this PR). And even with an empty prompt written on my machine, it holds at 16KB, only increasing to two pages (32 KB) with 24 rows written.

Here are some measurements. 

| Per terminal                                          | Before (macOS) | After (macOS) | Before (Linux) | After (Linux) |
|-------------------------------------------------------|----------------|---------------|----------------|---------------|
| Page-list memory dirty, fresh                         | 48 KiB         | 16 KiB        | 24 KiB         | 8 KiB         |
| Page-list memory dirty, 24 visible rows written       | 64 KiB         | 32 KiB        | 36 KiB         | 20 KiB        |

Note macOS uses 16KB pages and Linux generally uses 4 KB pages. 

I ran `ghostty-bench +terminal-stream` main vs this branch and with every normal workload the results are within noise (sometimes faster sometimes slower).

**AI usage:** It was used as a judge/validator. The actual changes were me, commit messages and PR messages all me.

### DIFF
--- a/src/terminal/ref_counted_set.zig
+++ b/src/terminal/ref_counted_set.zig
@@ -212,6 +212,30 @@ pub fn RefCountedSet(
             @memset(table.ptr(base)[0..l.table_cap], 0);
             @memset(items.ptr(base)[0..l.cap], .{});
 
+            return initFromParts(table, items, l, context);
+        }
+
+        /// Like `init`, but for backing memory that the caller guarantees
+        /// is already zero-filled (e.g. fresh OS pages). This writes
+        /// nothing to the backing buffer, so the OS pages behind the table
+        /// and items stay untouched until the first `add`.
+        ///
+        /// Behavior is undefined if the backing memory is not zero.
+        pub fn initAssumeZeroed(base: OffsetBuf, l: Layout, context: Context) Self {
+            return initFromParts(
+                base.member(Id, l.table_start),
+                base.member(Item, l.items_start),
+                l,
+                context,
+            );
+        }
+
+        fn initFromParts(
+            table: Offset(Id),
+            items: Offset(Item),
+            l: Layout,
+            context: Context,
+        ) Self {
             return .{
                 .table = table,
                 .items = items,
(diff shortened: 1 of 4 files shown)


mitchellh:
cc @qwerasd205 a ref counted set change but it doesn't change any actual logic, just avoids a memset for the assume zero case, ref counted set was already empty-at-zero.

src/terminal/ref_counted_set.zig
```diff
@@ -726,7 +750,8 @@ pub fn RefCountedSet(
 
                 var psl_stats: [32]Id = @splat(0);
 
-                for (items[0..self.layout.cap], 0..) |item, id| {
+                // Only items below next_id have ever been written.
+                for (items[1..self.next_id], 1..) |item, id| {
```
qwerasd205, line 755: !!! No, not true! `next_id` goes down when dead items are "trimmed" from the end, and I believe there's some metadata that isn't fully zeroed out. Not sure if it's relevant to this function (haven't fully re-context-loaded on this and took a proper look at this logic here), but this comment isn't right, so if you're relying on that assumption for this change, it's a bad one!

vancluever (approved):
Very cool (again 🙂)!

(Notwithstanding qwer's comments ofc!)

src/terminal/ref_counted_set.zig
mitchellh: Ah bleh, I misunderstood that. No, the runtime doesn't depend on it, because the only issue is really is that with zero-construction `meta.bucket` is 0 and not maxInt. assertIntegrity probably always should've started at item 1 anyways since item 0 is reserved and not used and its bucket value doesn't matter afaict.

EDIT: To be extra clear: meta.bucket is 0 for the zero item which always "exists", and is fixed up for items `1..` on `insert`.

PR #14144, opened 2026-09-04
input: encode ctrl keys with modifyOtherKeys 2

mitchellh:
#7425

Control-modified characters use xterm's MOK2 encoding.

I incorrectly believed previously that MOK2 ctrl chars were still encoded  as C0 bytes. This is wrong. I'm going to do a more in depth audit if possible with every possible key combination against xterm to see where we diverge but this fixes this for now without regressing any tests.

Background: https://invisible-island.net/xterm/modified-keys.html

### DIFF
--- a/src/input/key_encode.zig
+++ b/src/input/key_encode.zig
@@ -379,42 +379,10 @@ fn legacy(
         return try writer.writeAll(sequence);
     }
 
-    // If we match a control sequence, we output that directly. For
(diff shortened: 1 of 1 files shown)


aaronlumen:
Ansi ctrl sequence is good 
\033[;2F
Reset terminal position for live score updates on small trade board for example.  They can be sneaky

PR #14145, opened 2026-09-04
input: improve xterm modifyOtherKeys 2 compatibility

mitchellh:
Follow-up to #14144

I wrote a harness that created all possible US-layout keyboard input combinations with xterm patch 411 and Ghostty main and compared their full encoding sequence. There are various miscompatibilities on purpose but these were definitely bugs I wanted to address first.

- Normal-mode numeric keypad keys now preserve their numeric output instead of falling through to generic MOK2 encoding. Application keypad behavior is unchanged.
- F13 through F25 now emit their xterm-compatible function-key sequences, including modifier parameters.
- Help and Context Menu now emit editing-key codes 28 and 29.
- Alt+Escape now emits `CSI 27;3;27~` under MOK2 while retaining the traditional `ESC ESC` encoding otherwise.

After these changes, 2,081 of 2,096 cases match xterm exactly. The remaining 15 differences are intentional.

### DIFF
--- a/src/input/function_keys.zig
+++ b/src/input/function_keys.zig
@@ -89,7 +89,7 @@ pub const keys = keys: {
     result.set(.page_up, pcStyle("\x1b[5;{}~") ++ .{Entry{ .sequence = "\x1B[5~" }});
     result.set(.page_down, pcStyle("\x1b[6;{}~") ++ .{Entry{ .sequence = "\x1B[6~" }});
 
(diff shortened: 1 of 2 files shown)


jcollie:
> After these changes, 2,081 of 2,096 cases match xterm exactly. The remaining 15 differences are intentional.

Did #14146 fix any of these differences. If not are they documented in the source code anywhere?

PR #14167, opened 2026-09-07
terminal/parser: preserve CSI subparameter separators

philocalyst:
CSI allows for BOTH `;` params AND `:` subparams. Prior to this, the parser
would discard sequences containing colons unless the last byte was `m`.
That mixed command policy into syntax parsing, and caused a little issue for my
later work on the multi-cursor protocol, which makes great use of subparams.

In this change:
- I introuce a `Parser.Params` so the bytewise parser and the stream *fast path*
  share accumulation, overflow, and separator handling;
- Preserve separator bits on every CSI dispatch rather than treating colons as
  a parser error;
- Keep 24 parameters inline and (addressing an earlier comment) growing allocator-backed streams only for larger
  batches. I borrow from Kitty's internal limit of 256, but I can't say there's a good reason to bound it, other than it's
  super bad terminal behavior.
  For now over-limit sequences as a unit in comparison to sendin a dispatch on a truncation.

A colon is only *meaningful* to a command that defines a subparameter (ECMA-48
5.4.2). One predicate, `Action.CSI.allowsSubparams`, says which finals those
are: `m` (SGR) and `CSI > ... SP q` (this protocol). Anything else carrying a
colon gets dropped whole with a cold warning. Read positionally, `CSI 1:2 H`
would be a cursor move to (1, 2). The bytewise state machine and the stream fast
path both call that one predicate, so they can't drift apart.

Separator bits live in a `StaticBitSet` next to the parameters. Two things keep
that off the hot path. `Params` remembers whether it saw a colon *at all*, so
`reset` skips clearing the 32-byte mask on nearly every sequence. And consumers
borrow the bit set by pointer instead of copying it into each dispatch.

SGR behavior is otherwise unchanged. Other CSI commands still get their param
values with no allowlist in the way; the consumers that actually need
subparameter structure read the separator bits.

Added tests that work on bytewise and stream parsing, mixed separators, separator
reset, inline and spilled storage, allocation failure, the 256-parameter
boundary, and that subparameters only reach the finals that define them.

AI DISCLOSURE: I used $llm(s) for research/hardening of implementation and differential testing against Kitty. I wrote the initial Zig code and then got cross-references and reviewed improved style/structure to match idioms (I'm a rust dev). All work is double + triple checked, but let me know if anything feels weird, want to get good at Zig and ideally write more of any future PRs by hand (to become a not ai driver with real skillz!!)

### DIFF
--- a/src/terminal/Parser.zig
+++ b/src/terminal/Parser.zig
@@ -123,6 +123,39 @@ pub const Action = union(enum) {
             };
         }
 
+        /// Iterates the parameters as semicolon-separated groups, where
+        /// a group is the maximal run of parameters joined by colons.
+        /// `1;2:3:4;5` yields `{1}`, `{2,3,4}`, and `{5}`.
+        ///
+        /// This saves protocols that structure their parameters with
+        /// colons, such as the kitty multiple cursors protocol, from
+        /// walking the separator bits themselves.
+        pub const GroupIterator = struct {
+            params: []const u16,
+            params_sep: *const SepList,
+
+            /// The parameter the next group starts at. Set this past a
+            /// leading parameter you have already consumed, such as an
+            /// operation code.
+            index: usize = 0,
+
+            pub fn next(self: *GroupIterator) ?[]const u16 {
+                if (self.index >= self.params.len) return null;
+
+                // A set bit at i means parameter i is followed by ':',
+                // so the group continues past it. The bound keeps a
+                // trailing bit from running off the end.
+                const start = self.index;
+                var end = start;
+                while (end + 1 < self.params.len and
+                    self.params_sep.isSet(end)) : (end += 1)
+                {}
+
+                self.index = end + 1;
+                return self.params[start .. end + 1];
+            }
+        };
+
         // Implement formatter for logging
         pub fn format(
             self: CSI,
--- a/src/terminal/Parser.zig
+++ b/src/terminal/Parser.zig
@@ -454,6 +454,46 @@ pub const Params = struct {
     }
 };
 
+test "Params: spills only after inline capacity" {
+    var params: Params = .empty;
+    params.alloc = testing.allocator;
+    defer params.deinit();
+
+    // Exactly `INLINE_PARAMS` still fits inline.
+    for (0..INLINE_PARAMS) |i| params.push(@intCast(i), false);
+    try testing.expectEqual(0, params.spill.items.len);
+
+    // One more moves the whole sequence to the heap, separators and all.
+    // `take` still hands back a single contiguous slice.
+    params.push(INLINE_PARAMS, true);
+    const values = params.take().?;
+    try testing.expectEqual(INLINE_PARAMS + 1, values.len);
+    for (values, 0..) |value, i| try testing.expectEqual(i, value);
+    try testing.expect(params.seps.isSet(INLINE_PARAMS));
+
+    // A reset keeps the spill capacity but not its contents. The next
+    // short sequence is served inline again.
+    params.reset();
+    params.push(1, false);
+    try testing.expectEqualSlices(u16, &.{1}, params.take().?);
+    try testing.expectEqual(0, params.spill.items.len);
+}
+
+test "Params: allocation failure drops the sequence" {
+    var failing: testing.FailingAllocator = .init(testing.allocator, .{
+        .fail_index = 0,
+    });
+    var params: Params = .empty;
+    params.alloc = failing.allocator();
+    defer params.deinit();
+
+    // A failed spill discards everything. Half a sequence is worse than
+    // none of it.
+    for (0..INLINE_PARAMS + 1) |i| params.push(@intCast(i), false);
+    try testing.expect(params.take() == null);
+    try testing.expect(failing.has_induced_failure);
+}
+
 pub fn init() Parser {
     var result: Parser = .{
         .state = .ground,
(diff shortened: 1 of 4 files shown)


philocalyst:
- https://github.com/philocalyst/ghostty/pull/2
- https://github.com/philocalyst/ghostty/pull/3
- https://github.com/philocalyst/ghostty/pull/4

The rest are above, sad and frustrated that I can't figure out gh PR stacks, but if you want to review ahead-of-time, give them a gander

jcollie:
Can you demonstrate that this is how `xterm` behaves? When it comes to terminal escape sequences `xterm` is our golden standard and we don't deviate from following what it does without good reason.

mitchellh:
In addition to the above request by @jcollie, please also please share:

- micro benchmark numbers of the CSI parser
- terminal-stream with normal CSI workloads
- adversarial terminal-stream numbers with colon-heavy workloads
- allocation impact in more detail, you explained some of this but i'm curious what the sizes were before, after, and what their failure behavior is

Although, if xterm doesn't support this, then al of this is moot and we will close.

philocalyst:
Great feedback, going to be a bit delayed in benchmarks until I get a green, so here's kind of my logic (source ranges below so you're welcome to form your own opinion). I should have been clearer about where this is an intentional extension rather than matching xterm...

So off the bat, xterm *does* support colon subparams, but that support is *command-specific*. It maintains their metadata and explictly for the purposes of SGR and keyboard-mdofiier handling. It's different from a general permission to accept colons on arbitrary CSI commands. [^1] Xterm does **not** implement Kitty’s multiple-cursors command: its `CSI >` table rejects the space intermediate in `CSI > … SP q`.[^2] Of course, Xterm doesn't do many other things Ghostty does, but in this case there's no behavior to reproduce, but it also doesn't implement some differential behavior either. 

Justifying the PR further, the goal is to make us capable of the Kitty multi-cursor extension, and perhaps future work they do over there, as this seems to be a route they like. Over in kitty-land, they dispatch that exact prefix/intermediate/final combination to its multiple-cursor handler, which consumes the preserved colon-joined groups.[^3] This PR is explictly linked to the rest in the stack, so it's goals are coupled with that protocol work, viewed in isolation I could see how this just looks like a weakening of definitions, but it's not.

This new gate keeps the prior `m` allowance and adds ONLY that exact multi-cursor ID. Colon-bearing CUP, such as `CSI 1:2H`, and colon-bearing ordinary cursor-style commands remain VETOED. [^4]

The one thing that is a little scary is the slot change. Prior to this, it was 24. Xterm has always had 30. Kitty goes all the way up to 256. I implemented a 24-inline-with-spill abstraction, so in the common case allocation behavior should be the same, subject to compiler optimizations (but nothing proves this is the right choice)[^1][^5]. Preserving the protocol’s grouping is necessary; but I'm open to discussion on that strategy there. This raising of the capacity also makes previously rejected long sequences acceptable, I don't know of some kind of evil attack that would leverage this, but who knows! 

Have mercy on my specialization here!

Again benchmarks to follow, found some odd behaviors in edge cases when I was researching in my follow-up PRs, but here I'm fairly confident in the work. I'll get benches + fixes up on their work *if* we're able to agree on getting this in (please yes)

[^1]: **xterm’s representation and command-specific handling:** [`ptyx.h:425–437`](https://github.com/ThomasDickey/xterm-snapshots/blob/9489b2056ee51fa9dd6a7087483b9b8f85d6a0c4/ptyx.h#L425-L437) includes the 30-slot limit and subparameter metadata; [`charproc.c:3496–3546`](https://github.com/ThomasDickey/xterm-snapshots/blob/9489b2056ee51fa9dd6a7087483b9b8f85d6a0c4/charproc.c#L3496-L3546) contains the complete subparameter gate and its exceptions. 

[^2]: **xterm rejects the extension’s identifier:** [`VTPrsTbl.c:4699–4745`](https://github.com/ThomasDickey/xterm-snapshots/blob/9489b2056ee51fa9dd6a7087483b9b8f85d6a0c4/VTPrsTbl.c#L4699-L4745) includes the `CSI >` table declaration and the space entry mapping to `CASE_CSI_IGNORE`. 

[^3]: **Kitty implements it:** the introducing implementation’s [`vt-parser.c:1345–1355`](https://github.com/kovidgoyal/kitty/blob/8f5dc42a612533efdbdbbe638e7a2a005c76599e/kitty/vt-parser.c#L1345-L1355) distinguishes multiple cursors from cursor style and version queries; the group decoder is at [`vt-parser.c:953–981`](https://github.com/kovidgoyal/kitty/blob/df00dd1cecd18589fc0ddba01139c3e248278806/kitty/vt-parser.c#L953-L981). 

[^4]: **Existing versus proposed Ghostty behavior:** baseline [`Parser.zig:373–399`](https://github.com/ghostty-org/ghostty/blob/da9e21602f918d47a46399c458937eff7c7a74ac/src/terminal/Parser.zig#L373-L399); this PR’s exact allowlist at [`Parser.zig:106–124`](https://github.com/philocalyst/ghostty/blob/79ffa1e9f4ba62bd1131c55213fb97052476df10/src/terminal/Parser.zig#L106-L124), enforced in [`Parser.zig:649–673`](https://github.com/philocalyst/ghostty/blob/79ffa1e9f4ba62bd1131c55213fb97052476df10/src/terminal/Parser.zig#L649-L673). 

[^5]: **Capacity and allocation are separate choices:** Kitty’s [256-field limit](https://github.com/kovidgoyal/kitty/blob/df00dd1cecd18589fc0ddba01139c3e248278806/kitty/vt-parser.c#L22) and [fixed arrays, lines 215–224](https://github.com/kovidgoyal/kitty/blob/df00dd1cecd18589fc0ddba01139c3e248278806/kitty/vt-parser.c#L215-L224); this PR’s [limits, lines 263–280](https://github.com/philocalyst/ghostty/blob/79ffa1e9f4ba62bd1131c55213fb97052476df10/src/terminal/Parser.zig#L263-L280) and [accumulation, spill, and failure handling, lines 377–454](https://github.com/philocalyst/ghostty/blob/79ffa1e9f4ba62bd1131c55213fb97052476df10/src/terminal/Parser.zig#L377-L454).

philocalyst:
also added an AI disclosure, didn't mean to leave that out guys :(

PR #14200, opened 2026-09-09
terminal: report alt screen modes from the active screen

### LINKED ISSUE #14199: DECRQM reports alt screen modes 47/1047/1049 independently of the active screen
### Discussed in https://github.com/ghostty-org/ghostty/discussions/14198

<sup>Originally posted by **aymanbagabas** September  9, 2026</sup>
### Issue Description

Modes 47, 1047 and 1049 all control one thing: whether the alternate screen is
active. Ghostty stores an independent bit per mode, and `setMode` writes that bit
before dispatching the screen switch, so the bit and the actual screen can
disagree. DECRQM then reports the stale bit.

Concretely, entering the alternate screen with `?47h` and leaving it with
`?1049l` leaves mode 47 reported as set while the primary screen is showing.
`requestMode` reads the raw bit (`modes.get(mode)` in
`src/termio/stream_handler.zig`), and nothing reconciles that bit against the
active screen, which is tracked separately as `screens.active_key`.

The practical effect is that an application cannot use DECRQM to learn whether
the alternate screen is active. That matters for anything that inspects or
restores terminal state: shells, editors, multiplexers and session managers. It
also means the three modes can report three different answers to the same
question.

For reference, xterm keeps no per-mode bit for these. It answers DECRQM for all
three from a single buffer flag, `screen->whichBuf`:
[`misc.c#L5610-L5618`](https://github.com/ThomasDickey/xterm-snapshots/blob/9489b2056ee5/misc.c#L5610-L5618).

### Expected Behavior

After leaving the alternate screen, querying mode 47 reports reset:

```
^[[?47;2$y
```

### Actual Behavior

Mode 47 reports set while the primary screen is showing:

```
^[[?47;1$y
```

### Reproduction Steps

1. Open Ghostty.
2. Enter the alternate screen with one mode and leave with another, then query
   the first mode:

```sh
printf '\033[?47h\033[?1049l\033[?47$p'
IFS= read -r -t 1 reply; printf '%s\n' "$reply" | cat -v
```

3. Observe the reply is `^[[?47;1$y` (set), although the primary screen is
   showing. It should be `^[[?47;2$y` (reset).

The same applies to any mix of the three. For example, `?47h` followed by
querying 1047 reports reset while the alternate screen is active.

### Ghostty Logs

```text
N/A
```

### Ghostty Version

```text
Ghostty 1.3.1

Version
  - version: 1.3.1
  - channel: stable
Build Config
  - Zig version   : 0.15.2
  - app runtime   : .none
  - font engine   : .coretext
  - libxev        : kqueue
```

### OS Version Information

macOS 26.6.2 (build 25G83)

None

_No response_

```ini
N/A
```

### Additional Relevant Information

There is a fix on this branch:

https://github.com/aymanbagabas/ghostty/tree/aymanbagabas/vt-altscreen-mode-report

It makes `screens.active_key` the single source of truth, so all three modes
report the screen that is actually active.

It also stops XTRESTORE applying DECSET side effects for these modes. Restoring
mode 1049 currently routes through `setMode`, so `?1049r` erases the screen;
restoring should only switch buffers. xterm does that here:
[`charproc.c#L8588-L8600`](https://github.com/ThomasDickey/xterm-snapshots/blob/9489b2056ee5/charproc.c#L8588-L8600).
Setting `?1049h` still erases unconditionally, which is unchanged.

A second, larger change is worth considering. `ModeEntry` currently assumes every
mode owns a bit in `ModePacked`. Letting an entry declare that it reflects
another mode's state would collapse these three onto one bit and one XTSAVE slot,
and would remove the reconciliation entirely. That renumbers the snapshot bit
registry, so it needs a snapshot format version bump and matching `snapshot.ksy`
changes, which is why the branch does not do it. Happy to take that route instead
if preferred.

### I acknowledge that:

- [X] I have searched the Ghostty documentation and confirm there doesn't exist a configuration option that solves my issue.

aymanbagabas:
Modes 47, 1047, and 1049 each owned an independent bit in ModePacked, but all three describe a single piece of state: whether the alternate screen is active. setMode writes the bit generically before dispatching to switchScreenMode, so the bit and the screen could disagree. Entering the alternate screen with ?47h and leaving it with ?1049l left mode 47 reported as set while the primary screen was showing, because DECRQM reads the raw bit and nothing reconciled it against the active screen.

switchScreen now converges the three bits on the screen it switched to, making screens.active_key the single source of truth and the bits a mirror with exactly one writer. This also covers callers that bypass mode dispatch, such as the tmux viewer and snapshot restore.

XTRESTORE no longer routes these modes through setMode. Applying DECSET semantics on restore erased the screen on ?1049r, whereas restoring should only switch buffers. Setting ?1049h still erases unconditionally, which is unchanged.

The complete-v1 snapshot fixture is regenerated. Its wire format is unchanged; only current_modes bits 16, 33, and 35 and the record crc32c differ, since that fixture's terminal is on the alternate screen and previously encoded an inconsistent state.

For reference, xterm answers DECRQM for all three modes from a single buffer flag rather than per-mode bits:
https://github.com/ThomasDickey/xterm-snapshots/blob/9489b2056ee5/misc.c#L5610-L5618

Fixes: https://github.com/ghostty-org/ghostty/issues/14199

## AI disclosure

Implemented with AI assistance (GitHub Copilot CLI, model Claude Opus 5). I reviewed, understand, and can explain all of the code and its interaction with the key encoder without AI aid, per the AI Usage Policy.


aymanbagabas:
@jcollie This fixes the issue the naive way, reusing the existing ModeEntries. Ideally, terminal modes would be derived from the internal state of the terminal rather than we keep track of all of them.

jcollie:
I'm not the final authority on terminal stuff but you've made a convincing argument, especially with the link to the xterm source. I'll have to look into it some more before I'm sure this is the best way.

PR #14203, opened 2026-09-10
build: refactoring our use of translate-c

vancluever:
This commit refactors our use of translate-c, in preparation for larger removal of `cImport` and better co-ordination between building of C dependencies and translation of headers.

The major update is the creation of an internal helper package that wraps our use of the external translate-c library. This allows us to not only have better shorthand and a data-driven, declarative approach to C translation (versus the otherwise more imperative approach), it also funnels the external dependency into a single package instead of spreading it out among what will be an increasingly larger amount of places as dependencies in `pkg/` get updated.

It also includes some refactors, namely to the harfbuzz package, which has had its individual settings refactored into helpers to allow for the settings to be better shared between translation and the build of the c-based static library.

Wuffs has also had a bit of a refactor too so that we don't generate a file with all of the macro defines in it - we just send these in as `-D` flags now.

### DIFF
--- a/pkg/translate-c/build.zig
+++ b/pkg/translate-c/build.zig
@@ -1,0 +1,218 @@
+//! This is a wrapper package for our use of translate-c. It provides helpers
+//! for short-hand addition of the translation of C files and headers, along
+//! with lower-level control of the process a la the standard external
+//! translate-c package.
+
+const std = @import("std");
+const apple_sdk = @import("apple_sdk");
+pub const Translator = @import("translate_c").Translator;
+
+/// Options for translation.
+pub const Options = struct {
+    /// Describes the specification for a single include file.
+    pub const IncludeFile = struct {
+        /// Describes the type of an include file.
+        pub const Type = enum {
+            /// A system include, included as `<file.h>`.
+            system,
+
+            /// A user-defined include, included as `"file.h"`.
+            user,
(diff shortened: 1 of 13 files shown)


pkg/translate-c/build.zig
```diff
@@ -0,0 +1,224 @@
...
+/// name the dependency anything else other than `translate_c`, change this
+/// to match.
+///
+/// When importing this dependency across multiple files within the same build
+/// instance, make sure you only set this once, preferably early on in your
+/// project's root `build` function. When using multiple build *instances*, you
+/// will want to change this once per instance (e.g., anywhere there is a root
+/// `build` function beside a build.zig.zon file that has has this as a
+/// dependency).
+pub var dependency_name = "translate_c";
```
vancluever: I'm assuming that each build would load this file in as an independent container so each one would have a different static var, but I'm not too sure. I'm probably going to review one more time before taking this out of draft, and if in doubt, I may move this to `Options` or just make it constant and say that you need to load the dependency in as `translate_c` completely or else things will break.

vancluever: Yeah, turns out that it's file-local and build instances don't affect this, so like if you have two `build` functions at different dependency layers, changing it in one will affect the other. Will move this to options.

PR #14204, opened 2026-09-10
terminal: consume unused row capacity when a page grows managed memory

mitchellh:
`Page.increaseCapacity` previously would only try to fit capacity increases for managed memory into remaining `std_size` space. This meant that every single initial increase would exceed std_size and force heap allocation, which is slow for a variety of reasons.

This commit changes `increaseCapacity` to consume unused row capacity if it has it in order to fit more managed memory. Practically: if a page has 350 row capacity, but has only written 40 rows, we'll shrink rows to make room for styles, or whatever.

If it can't fit the desired capacity even by shrinking rows to what is currently used, then the page goes non-standard with the existing capacity retained.

Benchmark is `ghostty-bench +terminal-stream` on a 128MB file of the given general corpus produced by `ghostty-gen`.

macOS (Apple Silicon, 16 KiB pages):

  | Corpus | Base     | This PR           |
  |--------|---------:|---------------:|
  | styled |   651 ms |   588 ms (-10%) |
  | utf8   |   716 ms |   683 ms  (-5%) |
  | mixed  | 5,419 ms | 5,674 ms  (+5%) |
  | osc    | 3,074 ms | 3,095 ms  (+1%) |

EDIT: I should add that the reason that mixed/osc are more expensive is because they don't have the projection optimization I did for graphemes+styles. Under the "laddering" case (double, double, double), this optimization is worse because repeatedly swapping items in the pool is actually more expensive than a heap allocation. I have some fixes for that coming, but I don't think these represent realistic workloads currently and that style-heavy workloads are more realistic.

If I'm wrong then well, its a one-line backout really.

### DIFF
--- a/src/terminal/PageList.zig
+++ b/src/terminal/PageList.zig
@@ -4165,6 +4165,13 @@ pub const IncreaseCapacityError = error{
 /// (we currently double every time) but callers shouldn't depend on that.
 /// The only guarantee is some amount of growth.
 ///
+/// Growing a managed dimension can push the layout past the standard
+/// page size, which would move the page out of the memory pool onto
+/// its own heap mapping. Before that happens, unused row capacity is
+/// traded for the requested memory: the new page keeps every row the
+/// page currently uses (`size.rows`) but its row capacity may shrink.
+/// Callers must not assume row capacity never decreases.
+///
 /// Adjustment can be null if you want to recreate, reclone the page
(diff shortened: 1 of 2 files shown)


mitchellh:
Actually stand by, let me get in some other work so I can get a more broad improvement. It isn't totally verified its just what I think so let me wait until I know for sure.

PR #14206, opened 2026-09-10
feat(cli): add option filter to `+show-config` action

korikhin:
This PR adds the ability to provide a list of configuration keys to `+show-config` action for output filtering.

Closes #2486.

### Testing

Built the app and walked through all the cases below, also running the action with various flag combinations
(`--docs`, `--default`, `--changes-only`, `--no-pager`) to confirm they retain their existing behaviour.

```sh
# No filters. Prints the full user config.
ghostty +show-config
ghostty +show-config --

# No match. Prints nothing (exit code 0).
ghostty +show-config -- non-existent-option

# Prints identical output. Entries deduplicated and order-independent.
# Non-existent names ignored, and so are empty or whitespace-only arguments.
ghostty +show-config -- font-family font-size
ghostty +show-config -- font-size non-existent-option font-family font-size '' ' '

# Prints default values.
ghostty +show-config --default -- font-family font-size
```

**AI Disclosure:** I asked Claude (Sonnet 5, free version) whether it is a good idea to introduce a `--` delimiter, because it strays from the original feature request, but simplifies parsing (CLI design decision). I also asked Claude to help me phrase the documentation comments. The code is hand-rolled.


korikhin:
Hi @jcollie! Skimmed through the changes once more and squashed everything. Hope I didn't miss anything.

korikhin:
Hey @jcollie, friendly ping on this one. Any thoughts — yay or nay?

jparise:
I'm -0 on this one, given that it's easy enough to pipe the existing output to e.g. `grep`.

I was actually after something a little different in https://github.com/ghostty-org/ghostty/issues/2486#issuecomment-2435123441, which turned into `+explain-config`.

korikhin:
Thanks for the feedback!

I completely overlooked the `+explain-config` action, which is neat as it also handles keybinds. But still, `+show-config` handles such flags as `--default`, `--docs`, and also **showing the values**. And `grep` is only useful on an option-per-line basis (`--docs=false`), because if you also want to see the documentation with something like

```sh
ghostty +show-config --docs | grep -C 10 font-family
```

it might be cut somewhere in the middle.

This gives the desired output:

```sh
ghostty +show-config --docs | awk -v RS='' '/font-family =/'
```

If you don't provide ` =`, some other stuff sneaks into the output because of the way `+show-config` handles new lines.
The `awk` matched `font-family` in the font variant options and yanked `font-size` as well.

<details>
<summary>Output (awk without " =")</summary>

```
# The font families to use.

...

# configuration files or on the CLI if you want to clear values set on the
# CLI.
font-family = JetBrainsMonoNL Nerd Font Mono             <- (we are looking for this)
font-family-bold = JetBrainsMonoNL Nerd Font Mono
font-family-italic = JetBrainsMonoNL Nerd Font Mono
font-family-bold-italic = JetBrainsMonoNL Nerd Font Mono
# Font size in points. This value can be a non-integer and the nearest integer
# pixel size will be selected. If you have a high dpi display where 1pt = 2px

... (this block has no "font-family" to match)

# environment (e.g. the GNOME display scale and large text settings).
font-size = 15
```

</details>

<details>
<summary>Untouched output (original)</summary>

```
# The font families to use.

...

# configuration files or on the CLI if you want to clear values set on the
# CLI.
font-family = JetBrainsMonoNL Nerd Font Mono

font-family-bold = JetBrainsMonoNL Nerd Font Mono
font-family-italic = JetBrainsMonoNL Nerd Font Mono
font-family-bold-italic = JetBrainsMonoNL Nerd Font Mono
# Font size in points. This value can be a non-integer and the nearest integer
# pixel size will be selected. If you have a high dpi display where 1pt = 2px

...

# environment (e.g. the GNOME display scale and large text settings).
font-size = 15
```

</details>

Anyway, you can always perform a search in the pager if you have one, or combine the actions together. Or make better use of `grep`/`awk`. But from a user's perspective, this is simply easier — I am just a user! I just wanted to take a stab at this.

After all, I think these actions (with this PR) don't compete with each other, but rather complement each other, because they have different use cases as per this [comment](https://github.com/ghostty-org/ghostty/issues/2486#issuecomment-2435123441) and their literal names.

IMHO, bonus point for the ability to provide a **list** of options + non-allocating.

Thank you for reading!

PR #14208, opened 2026-09-10
app: add window theme switching action

jparise:
Add an app-scoped set_window_theme action with dark, light, and system values so appearance can be changed from keybindings and generated command palette entries. Previously these controls had no shared action and were unavailable across application runtimes.

Apply the action as a session override through NSAppearance on macOS and AdwStyleManager on GTK. Keep overrides across config reloads, matching the existing window-decoration override behavior, without persisting configuration.

See #7221

### DIFF
--- a/include/ghostty.h
+++ b/include/ghostty.h
@@ -863,6 +863,13 @@ typedef struct {
   bool soft;
 } ghostty_action_reload_config_s;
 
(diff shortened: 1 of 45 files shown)


trag1c (changes requested):
Blocking on the strings

PR #14210, opened 2026-09-10
libghostty and renderer base layer to support pixel precise scrolling.

j-c-m:
This PR is the base layer to support pixel precise (smooth) scrolling. It does not change any existing behaviors yet, but supplies an inspector to manually manipulate the extra fractional row.

It is a smaller, hopefully more correct implementation aligned with the ideas in https://github.com/ghostty-org/ghostty/pull/14122.

When fractional scrolling (GHOSTTY_TERMINAL_OPT_SCROLL_ROW_FRAC) is enabled the render puts one extra document row under the viewport (if it exists) so the consuming renderer can fill the fractional gap.

The row iterator will hold this extra row when set, the default of 0 leaves existing behavior unchanged.

Add GHOSTTY_TERMINAL_OPT_SCROLL_ROW_FRAC to scrollViewport.

The renderer additions make it possible to manipulate and visualize this with the inspector, as content_offset is always zero now, (no callers to any new interface), it does not change the behavior.

AI, Grok and Claude were used, I wrote all the code in the runtime path. AI tools extended the limited tests I implemented with more complete coverage subject to my review.

### DIFF
--- a/include/ghostty/vt/render.h
+++ b/include/ghostty/vt/render.h
@@ -210,6 +210,10 @@ typedef enum GHOSTTY_ENUM_TYPED {
   /** All render-state colors in one sized struct (GhosttyRenderStateColors).
    *  Initialize the output with GHOSTTY_INIT_SIZED before querying. */
   GHOSTTY_RENDER_STATE_DATA_COLORS = 19,
+
+  /** Fractional part of the scroll row (double).
+   *  When non-zero, the row iterator yields one extra row below the viewport. */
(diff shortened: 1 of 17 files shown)


mitchellh:
FYI, we're now restricting non-maintainer open PRs to 2 PRs at a time. I arbitrarily chose two simple ones to keep open. If you prefer to prioritize others, feel free to close those and reopen others.

This is because a lot of my questions and feedback were the same and its counterproductive to repeat myself. I think we can learn in serial and open in serial. This also helps maintainers pace their review.

PR #14212, opened 2026-09-12
gtk/build: speed up repeated builds by letting the Zig build cache work

jcollie:
Repeated builds were penalized for three reasons:

1. The gresource XML embedded the absolute cache paths of the compiled
.ui files. It now uses relative paths, resolved against a `--sourcedir`
passed to glib-compile-resources.

2. Blueprints were compiled through a small Zig wrapper, and a Run step
hashes the bytes of the executable it runs. A Zig binary does not relink
to the same bytes, so after a branch switch that touched the wrapper
every .ui moved, the gresource compiler re-ran and the whole app
recompiled for identical output. blueprint-compiler is now run directly
and the wrapper is reduced to a single version check whose output
nothing reads.

3. The gresource pipeline was built once per artifact. It is now
memoized on the *std.Build.

Cold builds and builds after real blueprint changes are unaffected. A
build after the wrapper relinks goes from ~43s to ~7s on my system.

AI disclosure: Claude Opus 5 was used to pinpoint the source of the
cache poisoning and to prepare a patch. Claude Fable 5.1 measured it,
traced the remaining churn to the wrapper's relinks, and replaced a
custom build step with stock ones. The commit message is human-written
based on the author's understanding of the changes.

### DIFF
--- a/src/build/SharedDeps.zig
+++ b/src/build/SharedDeps.zig
@@ -983,14 +983,121 @@ pub fn addSimd(
     }
 }
 
-/// Creates the resources that can be prebuilt for our dist build.
-pub fn gtkNgDistResources(
-    b: *std.Build,
-) struct {
+pub const GtkNgResources = struct {
     resources_c: DistResource,
     resources_h: DistResource,
-} {
+};
+
+/// Memoized result of `gtkNgDistResources`, keyed on the `*std.Build`.
+/// The configure pass is single-threaded, so a file-scope map is enough.
+var gtk_ng_resources: std.AutoHashMapUnmanaged(*std.Build, GtkNgResources) = .empty;
+
+/// Creates the resources that can be prebuilt for our dist build.
+///
+/// Memoized because `add` calls this once per artifact that links GTK and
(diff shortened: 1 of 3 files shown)


pluiedev (commented):
Hmm, my first instinct would be that a custom step shouldn't be necessary to cache properly and we should look a bit more closely

mitchellh:
I agree with that assessment. Andrew has made it very clear that custom steps are not a reliable solution in downstream build files and would break caching and other things as they make changes.

jcollie:
OK, updated with a version that doesn't have a custom build step.

src/build/SharedDeps.zig
```diff
@@ -983,27 +983,50 @@ pub fn addSimd(
     }
 }
 
-/// Creates the resources that can be prebuilt for our dist build.
-pub fn gtkNgDistResources(
-    b: *std.Build,
-) struct {
+pub const GtkNgResources = struct {
```
mitchellh, line 986: Oh wow `gtk-ng` spotted.

PR #14216, opened 2026-09-12
macOS: implement move_tab_to_new_window

### LINKED ISSUE #2630: Move tab to a new window

pedronaugusto:
Closes #2630 for macOS. #13621 added the `move_tab_to_new_window` action and the GTK side; this does the same on macOS with AppKit.

With native tabs a tab is already a window, so the action forwards to AppKit's own `moveTabToNewWindow:`. It's a no-op when the window is alone in its tab group, same as the Window menu item. The "only implemented on Linux" note comes off the doc comment in `Binding.zig`.
Tested on macOS 26, from a keybind and from the command palette. I have no macOS 13-15 machine.

Full disclosure, written with Claude Code, I directed it, read every line, and tested the result.

### DIFF
--- a/macos/Sources/Ghostty/Ghostty.App.swift
+++ b/macos/Sources/Ghostty/Ghostty.App.swift
@@ -621,6 +621,9 @@ extension Ghostty {
             case GHOSTTY_ACTION_MOVE_TAB:
                 return moveTab(app, target: target, move: action.action.move_tab)
 
+            case GHOSTTY_ACTION_MOVE_TAB_TO_NEW_WINDOW:
+                return moveTabToNewWindow(app, target: target)
+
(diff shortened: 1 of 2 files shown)


bo2themax (approved):
LGTM

PR #14219, opened 2026-09-13
Kkp on off feature

rajangdavis:


rajangdavis:
My bad, I was trying to make this MR for [my fork](https://github.com/rajangdavis/ghostty) and it created it here...

PR #14226, opened 2026-09-13
build: honor `-Dstrip` for libghostty-vt and vendored SIMD libs

Uzaaft:
Nix closure of the GTK app, before -> after:

libghostty-vt.a    17M -> 3.0M
closure           3.2G -> 1017M
paths              353 -> 308


clason:
Related: https://github.com/neovim/neovim/issues/40622

Uzaaft:
Good lead, I need to think more about this😅

PR #14246, opened 2026-09-15
build: fully transition away from cImport/addTranslateC

vancluever:
This migrates all remaining uses of `cImport` (and `addTranslateC` for good measure) to using translate-c for C translation, ensuring that we are ready for when `cImport` is removed from the language, and also that all sources of C translation are using the same snapshot of the external package (when can then be updated when we need to fix something).

A couple of notes:

* A few options have been added to support the new translations, namely the ability to link libraries (passed through to `linkLibrary` on the Translator side) and whether or not to initialize default values (looks like `cImport` did this without a way to control it, but translate-c does not do it by default).

* Using the new library linking option actually simplifies the process of translating a number of the C packages as we have been shipping the necessary headers for these packages already with the applicable libraries. For some of the more complex translation processes though, we still include the appropriate directories directly.


jcollie (approved):
LGTM and good riddance!

mitchellh:
hell yeah.

PR #14260, opened 2026-09-16
terminal: insertLines tolerates a stale scroll region

MisterTea:
## Summary

A resize can leave `scrolling_region.bottom` past the currently materialized page rows, so `page_pin.down(rem-1)` is null. The previous code unwrapped that and crashed the IO thread. Log a warning and skip the reverse-index until the next layout sync.

This is general VT robustness, not tmux-specific.

## Test plan

- [x] `zig build test-lib-vt -Dtest-filter="insertLines tolerates stale"`

## AI disclosure

This change was prepared with Cursor. I reviewed the code and can explain the scroll-region / page-pin interaction.


src/terminal/Terminal.zig
```diff
@@ -2984,10 +2984,23 @@ pub fn insertLines(self: *Terminal, count: usize) void {
     // region. So we take whichever is smaller.
     const adjusted_count = @min(count, rem);
 
+    const bottom_pin =
+        self.screens.active.cursor.page_pin.down(rem - 1) orelse {
+            // A resize/layout update can transiently leave the scroll-region
+            // bottom beyond the currently materialized page rows. Ignore the
```
mitchellh, line 2990: Hm, I think we should instead verify this bug in resize and fix it there, rather than band-aid it elsewhere.

mitchellh (changes requested):

MisterTea:
This ended up being a bug with our e2e testing code The old test never resized. It just did scrolling_region.bottom = 99 and swallowed the null page_pin.down. That is undone.  Closing this PR

PR #14262, opened 2026-09-16
tmux: fix list-windows action use-after-free

MisterTea:
## Summary

`receivedListWindows` stored `Action.windows` as a slice into a temporary ArrayList that is freed on return. Logging that action (or otherwise touching the slice) use-after-freed arena state and crashed Ghostty immediately after `list-windows`.

Sync layouts into `self.windows` first, then point the action at the stable `self.windows.items` slice. Print window counts instead of dumping `Window` with `{any}`, which walks `ArenaAllocator` nodes.

This is independently useful today: control mode already logs viewer actions.

Related: #1935

## Test plan

- [x] Regression test fails without the fix due to mismatched temporary/viewer window pointers
- [x] `zig build test -Dtest-filter='session changed resets state'`
- [x] `zig build test -Dtest-filter=tmux`

## AI disclosure

This change was prepared with Cursor. I reviewed the slice lifetime and the `Action.format` change.


mitchellh (changes requested):
Please add a test that would've failed without this fix.

MisterTea:
Done

PR #14273, opened 2026-09-17
tmux: handle window-close and active window

MisterTea:
## Summary

- Parse `%window-close` / `%unlinked-window-close` and `%session-window-changed`.
- Emit an `active_window` viewer action so later layers can track tmux's current window.
- Parse `%exit` as soon as the notification line is seen.

`%output` is left escaped on purpose. A control client that is not reading pane bytes should not pay to decode octal sequences up front; callers that need the original stream can unescape later (see #12076).

This is stacked on #14262. GitHub will also show that commit until it merges, because we cannot target another PR's branch from a fork.

**Depends on:** #14262

CI / compile may fail until that PR is on `main`. That is expected.

## Test plan

- [ ] `zig build test-lib-vt -Dtest-filter="tmux"` after #14262 is merged
- [ ] Close a tmux window and change the active window; parser emits the new notifications (still log-only on `main` until later PRs)

## AI disclosure

This change was written with [Cursor](https://cursor.com). I directed the work, reviewed the diff, and am responsible for it.

Related: #1935


mitchellh (changes requested):
We purposely didn't unescape output because if there is a tmux control client that attaches without the intention of reading output then its wasted CPU that is easy to avoid. Is there a reason to unescape it immediately?

The other two changes are acceptable, if you split them out.

MisterTea:
Dropped the parse-time `%output` unescape. A client that is not reading pane bytes should not pay that CPU up front; decoding can happen later at use (as in #12076). This PR is now only `%window-close` / `%unlinked-window-close`, `%session-window-changed`, and the `active_window` viewer action.

PR #14274, opened 2026-09-17
tmux: parse @affinities from list-windows

MisterTea:
## Summary

- Include iTerm2's session `@affinities` option in `list-windows` so native window groups can be restored later.
- Encode the field as a tab-delimited `A`-prefixed sentinel so values containing spaces stay unambiguous and an unset option remains parseable.
- Store affinities on each parsed window and update fixtures.

This is the unique commit `653d01f6e` (`tmux: parse @affinities from list-windows`). GitHub will also show #14262 and #14273 until those merge, because this branch is stacked and we cannot target another PR's branch from a fork.

**Depends on:** #14273 (which depends on #14262)

CI / compile may fail until those PRs are on `main`. That is expected.

## Test plan

- [ ] `zig build test-lib-vt -Dtest-filter="tmux"` after #14273 is merged
- [ ] list-windows fixtures with and without `@affinities` still parse
- [ ] Space-containing affinity values round-trip through the tab-delimited field

## AI disclosure

This change was written with [Cursor](https://cursor.com). I directed the work, reviewed the diff, and am responsible for it.

Related: #1935


mitchellh:
Closing PRs that depend on other PRs. Just wait until we merge the dependent PR.

PR #14300, opened 2026-09-18
font: store glyph cache keys as packed u64

j-c-m:
This is a long time follow-up to 93dcb195 and something I originally missed in 8824256, as I am making another performance pass through the render pipeline.

93dcb195 stopped using the full RenderOptions as identity but left them in the key, but packed them for a temporary identity and hashes. This commit uses the true identity as the key and stops any temporary packing, making eql very cheap allowing us to use a simple fast hash that (mostly) relies on glyph ids to distribute accross the hash table.

I don't have a big headline doom-fire-zig fps number maybe 1% 766->774 FPS 120x40 window on my hardware but it does seem limited somewhere else. However, when profiling doom-fire-zig runs the renderGlyph leaf drops from 22% to 1.8%.

AI: I used Grok 4.6 to build harnesses for automated testing & profiling utilizing ghostty-bench, `cmatrix-b`, and doom-fire-zig (120x40) (not in this PR).  Initial profiling research and actual code produced by a dumb little human (so says the AI).


src/font/SharedGrid.zig
```diff
@@ -469,50 +469,47 @@ const CodepointKey = struct {
...
+    },
 
     const Context = struct {
         pub fn hash(_: Context, key: GlyphKey) u64 {
-            // Packed is a u64 but std.hash.int improves uniformity and
-            // avoids collisions in our hashmap.
-            const packed_key = Packed.from(key);
-            return std.hash.int(@as(u64, @bitCast(packed_key)));
+            const x: u64 = @bitCast(key);
+            return x ^ (x >> 32);
```
mitchellh: Why this?

PR #14301, opened 2026-09-18
font: store CodepointKey as a packed u64

j-c-m:
This is the same treatment to CodepointKey as #14300

vs main this increases doom-fire-zig fps by 3% (749->772).

In my tests these will stack, and in profiling this change moves a ~17% getIndex to a ~3% HashMap.


src/font/SharedGrid.zig
```diff
@@ -463,10 +463,36 @@ pub fn renderGlyph(
...
-    presentation: ?Presentation,
+    style: Style,
+    has_presentation: bool,
+    presentation: Presentation,
+    _padding: u27 = 0,
+
+    const Context = struct {
+        pub fn hash(_: Context, key: CodepointKey) u64 {
+            const x: u64 = @bitCast(key);
+            return x ^ (x >> 32);
```
mitchellh, line 476: Why this?

mitchellh (changes requested):

mitchellh (changes requested):
Same as the other PR, just a simple question. This is simple enough and makes sense otherwise.

src/font/SharedGrid.zig
j-c-m: It's faster than std.hash.int (it is actually the first fold in std.hash.int) , but also why I had to move codepoint to the front of the packed structure. The distribution seems fine once you fold in the style & presentation so the same codepoint with different styles don't cluster together.

The lookup here is the hot path, and the eql is extremely cheap so even in the case we need to walk hash slot we get 3 or so walks before its more expensive than a "better" hash. In profiling in just general terminal workloads I did not see this happening.

PR #14279, opened 2026-09-17
gtk,opengl: cleanup & fixes for #14052

pluiedev:
Fixes most issues mentioned in #14243, #14254 and elsewhere e.g. on Discord. Will add more fixes if they can be solidly reproduced.

Please review each commit individually.

PR #14304, opened 2026-09-19
gtk,opengl: fix misplaced gl.finish() when exporting frames

AnthonyZhOon:
We must call `gl.finish()` after the present GL calls and before adding the exported frame handle to the readable LatestFrame slot that the app-thread reads.
After the fixes in #14279 I could reliably reproduce fcitx5 with mozc japanese triggering out of order frame states. I tracked this down to the unsynchronised present logic. I believe the bug was just triggered by rapid redraws allowing the app to read data from unsynchronised DMA buffers.
Now rendering *should* be smooth 

Fixes https://github.com/ghostty-org/ghostty/discussions/14243#discussioncomment-18459289


mitchellh:
This looks right to me, but will let @pluiedev provide her input.

pluiedev (approved):

jcollie (approved):

PR #14318, opened 2026-09-20
gtk: context menu "Copy Link" option

thomasfedb:
Implements #2107 for GTK only.

Adds a "Copy Link" item to the surface context menu that copies the URL of the link under the mouse. Works for OSC 8 hyperlinks and regex-matched links, and is hidden when there is no link.

Tested manually on Fedora 44 with an OSC 8 link and plain text URLs.

The `po/` changes are from `zig build update-translations`.
First pass written with Claude Code (Claude Fable 5.1), reviewed and reprompted by me. Tested by hand.


jcollie (changes requested):
Haven't looked at the code yet, but since we're in a string freeze right now this will have to wait until after 1.4.

po/be.po
```diff
@@ -757,7 +785,9 @@ msgstr "Скапіяваць вылучэнне як ANSI-паслядоўнас
 msgid ""
 "Copy the selection contents as ANSI escape sequences to a temporary file and "
 "open it."
-msgstr "Скапіяваць змест вылучэння як кіруючыя паслядоўнасці ANSI ў часовы файл і адкрыць яго."
+msgstr ""
+"Скапіяваць змест вылучэння як кіруючыя паслядоўнасці ANSI ў часовы файл і "
+"адкрыць яго."
```
jcollie, line 790: For some reason a bunch of strings got reformatted (as far as I can tell they didn't change). Perhaps it's a different version of gettext, or maybe a setting from your system that's leaking in? Whatever the case we need track that down and fix it, we don't want this sort of churn - it makes it difficult to review the code.

thomasfedb: Tried with gettext 1.0 also with the same result. Perhaps `update-translations` just hasn't been run for a while? Happy to remove those changes from this PR.

trag1c (changes requested):
Blocking on the strings as jcollie said

thomasfedb:
> Blocking on the strings as jcollie said

@trag1c Got a rough idea when the freeze will be over? Sorry, not familiar with the release cadence.

pluiedev:
After 1.4 releases. Current estimate is some time in the first half of October.

PR #14326, opened 2026-09-20
renderer: Vulkan

EriksRemess:
Had quick idea to see if I can use Vulkan instead of OpenGL as renderer for Linux build.
Need to build with -Drenderer=vulkan
Mostly done by Codex 5.6 Sol in High mode as an experiment, probably unmaintainable.


00-kat:
xref https://discord.com/channels/1005603569187160125/1551322174881665065 on the [Ghostty Discord](https://discord.gg/ghostty).

PR #14332, opened 2026-09-21
gtk: streamline surface overrides

neoto:
While investigating the EGL context being created twice (is this known and/or expected?), I noticed that the `overrides` argument was being repeated in a bunch of places. I couldn't help but do something about it so here I am.

This essentially just moves things around, placing related logic in a nicer box. I haven't changed anything when it comes to functionality.

Feel free to close if this isn't something worthwhile.

CC @jcollie


jcollie (approved):
LGTM

PR #14336, opened 2026-09-22
gtk,imgui: Fix broken widget by allowing a GLES context to be used

AnthonyZhOon:
```
info(opengl): loaded OpenGL 4.3
warning(gtk_ghostty_imgui_widget): GLArea for Dear ImGui widget failed to realize: Unable to create a GL context
warning(gtk_ghostty_imgui_widget): Dear ImGui context not initialized
```
Despite logs showing loaded OpenGL, api tracing ghostty showed OpenGL ES being used, the imgui widget was failing to initialize a context because we did not support GLES for the imgui widget.

Still not sure what's going on with our OpenGL api selection but this gets the inspector working in this scenario for me.

# AI Disclosure
GPT Astra-Light in the desktop ChatGPT wrote the intial code, I deleted unnecessary changes and looked up the imgui function being called to understand the change, and tested the result.


jcollie (approved):
This looks good to me. Tried it out on my system although I'm not sure that it was broke before. I'll wait to hear from some others before merging.

AnthonyZhOon:
With `GTK_DEBUG=interactive ghostty` if the GL backend is GL ES then without providing an ES version string to the imgui context init it fails to create a GL context which was exposed after we allowed gles-api on the main app.

PR #14371, opened 2026-09-23
terminal/c: expose the application-requested mouse shape

toppk:
This commit was generated with extensive usage of AI (openai codex and claude code).

Adds GHOSTTY_TERMINAL_DATA_MOUSE_SHAPE to ghostty_terminal_get, returning a public GhosttyMouseShape enum, so C frontends can read the pointer shape the terminal already tracks for OSC 22 without parsing OSC themselves.
I am doing this so that I can handle OSC22 requests from applications for my c based terminal. Without this I do not see these commands.

There was some discussion on discord whether to make it lower level by not using the enum and just sending the string. I didn't do that because my usecase is a normal c based terminal, and I think the curated enum that libghostty provides should be used by normal users, as well as the fact that this is already how zig land operates.

I did mention this in my vouch request: https://github.com/ghostty-org/ghostty/discussions/14365 but there's no other applicable issue or discussion on this subject on github. Although the discussion here: https://github.com/ghostty-org/ghostty/discussions/12477 proposes a more low level api that might be worth looking into at some point, I think that's complementary.


toppk:
pinging @pluiedev who had some ideas about this approach.

PR #14375, opened 2026-09-24
Support resizing the window with CSI 8 t

shreeve:
Relates to #1115, following my vouch request (#14372).

This implements the xterm window manipulation sequence `CSI 8 ; rows ; columns t`
(XTWINOPS), which lets a program request a terminal size in cells. My use case is
a TUI that wants a sensible starting size instead of asking the user to resize.
It's off by default behind a new `vt-window-resize-allowed` option, following
`title-report` and `vt-kam-allowed`.

**Security.** Since any program, including one over SSH, could send this:

- it's disabled unless the user opts in;
- requests below 40 columns × 10 rows are raised to that size, so a program can't
  shrink the window to hide its output (@jcollie's concern in #14372);
- requests larger than the screen are clamped to the screen;
- it's ignored in splits, multi-tab windows, the quick terminal, and when the
  window manager owns the size (fullscreen, maximized, tiled).

**GNOME.** This works on GNOME Wayland and X11. A floating GTK 4 window can resize
itself by calling `gtk_window_set_default_size()` while mapped: GTK uses the new
default size unless the compositor has fixed the size, and mutter accepts a size
committed by the client. gnome-terminal handles CSI 8 t the same way on GTK 4
(`terminal_window_update_size()`), including skipping maximized, fullscreen, and
tiled windows. One Ghostty-specific catch: the window's compute-size handler
re-asserts the current size when it exceeds the compositor bounds, which blocked
shrinking a window that fills the screen, so it steps aside for a requested resize.

Both platforms resize the window by the difference between the requested and
current surface size, so the titlebar, headerbar, and tab bar are accounted for.
A zero or omitted parameter leaves that dimension untouched.

**Testing.** Parser tests added; the full test suite passes on macOS and Linux.
On a macOS release build and on GNOME Shell 50.1 / mutter 50.1 / GTK 4.22
(headless, Wayland and Xwayland), verified via `stty size`: exact grids for full,
partial, minimum, and oversized requests; refusals for splits, tabs, maximized,
tiled, and fullscreen, and recovery once undone; and exact sizes at monitor
scales 1.25, 1.5, and 2, with 1.25 text scaling, and with window padding.

libghostty-vt parses the sequence but ignores it; exposing it to embedders as an
effect (like `title_changed`) could be a follow-up.

AI disclosure: I used Claude Code to help research, write, and test this change.
I've reviewed all of it and can answer questions about it.


shreeve:
resize

